Issue #16120, opened 2018-10-28
[Ionic 4.0.0-beta.13] <ion-reorder-group> not working for <ion-item-sliding> with <ion-item-options>


conorseed:
**Ionic version:**  (check one with "x")
(For Ionic 1.x issues, please use https://github.com/ionic-team/ionic-v1)
[ ] **2.x**
[ ] **3.x**
[X] **4.x**

**I'm submitting a ...**  (check one with "x")
[X] bug report
[ ] feature request

**Current behavior:**
A list of `<ion-item-sliding>` refuses to reorder when `<ion-reorder>` element is placed within the required `<ion-item>` element of an `<ion-item-sliding>`, with an `<ion-item-options>` element present within the same `<ion-item-sliding>`.

<img width="365" alt="screen shot 2018-10-29 at 11 17 23 am" src="https://user-images.githubusercontent.com/13262289/47622650-41b2fc00-db6c-11e8-9928-8c89235d6054.png">
 

**Expected behavior:**
Should be able to drag+drop to reorder `<ion-item-sliding>` with `<ion-item-options>`.

**Steps to reproduce:**
Use this code, try and re-order list. The first two items with be moveable. The third won't be. Once you've tried to move the third item, all of the items will then be non-movable.

```
<ion-list lines="full">
      <ion-reorder-group disabled="false" (ionItemReorder)="moveChecklist($event)">

      <ion-item-sliding>
        <ion-item href="/" routerDirection="forward" detail>
          <ion-reorder slot="start"></ion-reorder>
          <ion-label text-wrap>
            <h2>Moveable</h2>
          </ion-label>
        </ion-item>
      </ion-item-sliding>

      <ion-item-sliding>
          <ion-item href="/" routerDirection="forward" detail>
            <ion-reorder slot="start"></ion-reorder>
            <ion-label text-wrap>
              <h2>Moveable</h2>
            </ion-label>
          </ion-item>
        </ion-item-sliding>

      <ion-item-sliding>
        <ion-item detail>
          <ion-reorder slot="start"></ion-reorder>
          <ion-label text-wrap>
            <h2>Non moveable (ion-item-options present)</h2>
          </ion-label>
        </ion-item>
        <ion-item-options side="end">
            <ion-item-option color="light">
              <ion-icon name="clipboard" slot="top"></ion-icon>
            </ion-item-option>
        </ion-item-options>
      </ion-item-sliding>

    </ion-reorder-group>
  </ion-list>
```

```
[...]

PR #17568, opened 2019-02-21
fix(reorder): increase gesture priority

sbannigan:

Allow `ion-item-sliding` items to be reordered inside an `ion-reorder-group`.

- Increase the `gesturePriority` of `ion-reorder-group` from `100` to `110`.
- Add a test case for `ion-item-sliding` in the `reorder-group` basic test.

I'm not sure if there is a reason this was lower priority than `ion-item-sliding` before, but this seems to resolve the issue.

**Ionic Version**: 4.0.2

**Fixes**: #16120

### DIFF
--- a/core/src/components/reorder-group/test/basic/index.html
+++ b/core/src/components/reorder-group/test/basic/index.html
@@ -34,12 +34,22 @@
             <ion-reorder slot="end"></ion-reorder>
           </ion-item>
 
-          <ion-item button onclick="openAlert()">
-            <ion-label>
-              Item 2 (default ion-reorder)
-            </ion-label>
-            <ion-reorder slot="end"></ion-reorder>
-          </ion-item>
+          <ion-item-sliding>
+            <ion-item button onclick="openAlert()">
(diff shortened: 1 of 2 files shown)


core/src/components/reorder-group/test/basic/index.html
```diff
@@ -34,12 +34,22 @@
...
-          </ion-item>
+          <ion-item-sliding>
+            <ion-item button onclick="openAlert()">
+              <ion-label>
+                Item 2 (default ion-reorder with ion-item-sliding)
+              </ion-label>
+              <ion-reorder slot="end"></ion-reorder>
+            </ion-item>
+            <ion-item-options side="start">
+              <ion-item-option onClick="favorite(item)">Favorite</ion-item-option>
```
liamdebeasi: These functions don't exist (favorite, unread, etc). Can you remove them?

sbannigan: Ah yeah good catch! Removed 😄

Issue #17557, opened 2019-02-21
Ion-icon with slot attribute wrong color in ion-item with color attribute on android


Spinnenzunge:
# Bug Report
Slotted ion-icon does not apply the contrast color of the color variable that is set on ion-item on Android.

**Ionic version:**

[x] **4.x**

**Current behavior:**
If I set the color attribute on an ion-item, the ion-icon with slot="start" does not apply the contrast color of my set color on Android.

**Expected behavior:**
Contrast color of my color variable is set to slotted ion-icon on Android the same it is on iOS.

**Steps to reproduce:**
```
<ion-item color="primary" lines="none">
   <ion-icon src="assets/icon/icon.svg" slot="start"></ion-icon>
   <ion-label>Label</ion-label>
</ion-item>
```

**Related code:**
Following CSS is probably interfering.

```
::slotted(ion-icon) {
    color: rgba(var(--ion-text-color-rgb,0,0,0),.54);
    font-size: 24px;
}
```

**Ionic info:**

```
Ionic:

   ionic (Ionic CLI)             : 4.10.3 (/usr/local/lib/node_modules/ionic)
   Ionic Framework               : @ionic/angular 4.0.1
   @angular-devkit/build-angular : 0.12.4
   @angular-devkit/schematics    : 7.2.4
   @angular/cli                  : 7.2.4
   @ionic/angular-toolkit        : 1.4.0

System:

   NodeJS : v10.14.1 (/usr/local/bin/node)
   npm    : 6.7.0
   OS     : macOS Mojave
```

liamdebeasi:
Hi there,

Thanks for the issue! This has been merged (via #17585) and will be in the next release of Ionic.

PR #17585, opened 2019-02-22
fix(item): slotted ion-icon receives custom color

liamdebeasi:
Fixes a bug where a slotted ion-icon in ion-item was not receiving the color contrast value in MD mode

- Remove manually setting color for slotted icon in MD

**Ionic Version**:

**Fixes**: #17557


brandyscarney:
Instead of using `:not` could we just update the icon color? something like:

```
:host(.ion-color) ::slotted(ion-icon) {
  color: current-color(contrast);
}
```

My concern is that the current rule is so specific that someone won't be able to override a slotted `ion-icon`. I didn't test this theory though. 🙂

PR #17612, opened 2019-02-26
feat(skeleton-text): adds animated, support for border-radius and height

brandyscarney:
- moves most of the CSS to the host element so that it's easier to customize the skeleton text
- adds auto inherit for border-radius
- moves height to 100% so it will naturally fill an avatar and thumbnail
- adds animated property defaulted to `false`
- removes mode specific styles as they're the same
- changes background to an rgba making it easier to customize
- still works outside of a container element
- adds usage examples for javascript and angular
- moves the width to the parent element making it easier to style
- adds the ability to style the width using CSS
- uses the global theme text color as the background if a custom background is not supplied before defaulting

| `master` | `feat-skeleton-text-animated` |
| ------| ------|
| localhost_3333_src_components_skeleton-text_test_basic iphone 6_7_8 2 | localhost_3333_src_components_skeleton-text_test_basic iphone 6_7_8 3 |


brandyscarney:
Published a dev build with this if anyone wants to try it out early: `4.1.0-dev.201902272232.d66b12b`

manishrajmishra:
Is there any way so we can change the degree of inclination of loading effect from vertical to some degree ?

Issue #15434, opened 2018-08-31
this.platform.is('tablet') not working properly


smxdevst:
# Bug Report

```
Ionic:

   ionic (Ionic CLI)  : 4.1.1 (/usr/local/lib/node_modules/ionic)
   Ionic Framework    : ionic-angular 3.9.2
   @ionic/app-scripts : 3.1.11

Cordova:

   cordova (Cordova CLI) : 8.0.0
   Cordova Platforms     : android 7.0.0, ios 4.5.5
   Cordova Plugins       : cordova-plugin-ionic-webview 1.2.1, (and 25 other plugins)

System:

   ios-deploy : 1.9.2
   NodeJS     : v8.9.4 (/usr/local/bin/node)
   npm        : 6.0.1
   OS         : macOS High Sierra
   Xcode      : Xcode 9.4.1 Build version 9F2000
```

**Describe the Bug**
On my phone Google Pixel 2 XL `this.platform.is('tablet')` giving true

**Steps to Reproduce**
Add `this.platform.is('tablet')` on any page on Google Pixel 2 XL

**Expected Behavior**
`this.platform.is('tablet')` should return false.

yuankunluo:
The platforms are: Android,iOS, windows, browser, windowsphone

smxdevst:
@yuankunluo 

https://ionicframework.com/docs/api/platform/Platform/

![d0ddu5xe4b8](https://user-images.githubusercontent.com/13336971/44960517-fe167b80-aeb5-11e8-8310-f579091f03f0.jpg)

daem0ndev:
Here’s another resource I found for detecting pixel 2 xl: https://stackoverflow.com/questions/49078882/media-query-to-differentiate-google-pixel-1-2-vs-google-pixel-xl-vs-google-pixel

liamdebeasi:
Hi there,

This issue has been resolved via #17630 and will be in the next release of Ionic.

Thanks for the issue!

PR #17630, opened 2019-02-27
fix(platform): account for larger tablets

liamdebeasi:
This PR fixes an issue where `platform.is('tablet')` would return `false` for larger tablets such as the iPad Pro and the Galaxy View.

Note: There are still edge cases this PR misses, but the purpose of this PR is to bring support to some of the more common platforms (large iOS and Android tablets, specifically). I think the ideal fix requires some re-thinking of how we do platform detection, so I've created a feature request for it here: #17631

- Add additional checks to `isTablet` to see if platform is iPad or Android tablet. Fall back to old detection method if not.

**Ionic Version**:

**Fixes**: #15434


brandyscarney:
Going to tag @adamdbradley for review on this one as I know there were some issues with tablet sizes before.

Issue #17012, opened 2019-01-09
Ionic 4.0.0 RTL issues


abennouna:
This issue is to collect and keep track of RTL bugs vs. **master** so we can make sure to properly update and fix them.

Below are:
- a list of some utilities presenting issues in RC.0,
- and a list of components with tests included in the test suite.

## Legend
| Symbol  | Description |
| ------------- | ------------- |
| :white_check_mark: | Tested. Found compliant. |
| :ballot_box_with_check:  | Tested. Initially not compliant or Regression. Solved by PR specified in description. |
| :question:  | Not necessarily an issue, or not a core-specific issue, or design decision. |
| :warning:  | Not tested. Help needed. |
| :x:  | Tested. Not _completely_ compliant. Should list the defects. |

# Fixed issues

## CSS Utilities
  - [x] :white_check_mark: Text Alignment: `text-start`, `text-end`, `text-justify`, & their responsive versions
  - [x] :ballot_box_with_check: Float Elements: `float-start`, `float-end`, & their responsive versions (regression: https://github.com/ionic-team/ionic/issues/17012#issuecomment-455384111) (PR #18315 solves issue in Chrome)
  - [x] :white_check_mark: Element Padding: `padding-start`, `padding-end`
  - [x] :white_check_mark: Element Margin: `margin-start`, `margin-end`
  - [x] :white_check_mark: Flex Container Properties: `justify-content-start`, `justify-content-end`

## Grid
  - [x] :white_check_mark: Offsetting columns (iOS/MD) (PR: #16702)
  - [x] :white_check_mark: Push and pull (iOS/MD) (PR: #16702)

## Components
- [x] :white_check_mark: action-sheet
- [x] alert:
  - [x] :white_check_mark: **scroller with radio in MD**: position correctly (see https://gyazo.com/d1602f045193380be6d02432a7e61878)
- [x] :white_check_mark: anchor
- [x] :white_check_mark: avatar
- [x] :white_check_mark: badge
- [x] :white_check_mark: button
  - [x] :white_check_mark: **slotted icon in button in iOS**
  - [x] :white_check_mark: **slotted icon in button in MD**
- [x] :white_check_mark: buttons
- [x] :white_check_mark: card
- [x] :white_check_mark: card-header
- [x] :white_check_mark: checkbox
- [x] :white_check_mark: chip
- [x] :white_check_mark: content
- [x] datetime
[...]

PR #17645, opened 2019-03-01
fix(popover): update animation origin in RTL/MD

abennouna:
In case the popover position is adjusted because it is off screen, the RTL animation should be inverted as well.

- Original main-axis origin fix was flawed in case of RTL.
- Invert the animation’s main-axis origin in case the popover’s position is adjusted.

**Ionic Version**:
4.x

**Fixes**: #17012 + #17381


brandyscarney:
Looks great, thank you! Good to merge once the build passes. 👍

Issue #17659, opened 2019-03-02
Spinner icons: incorrect default spinners in MD


abennouna:
# Bug Report

**Ionic version:**
[x] **4.x**

**Current behavior:**
The default spinner icon uses the iOS platform default (**`lines`**), even in MD mode, for the following components:
- **`refresher`**
- **`infinite-scroll`**

**Expected behavior:**
In MD mode, the default spinner icon should be **`crescent`**.

**Steps to reproduce:**
Use one of these components in MD mode, without specifying an icon.

**Related code:**

```
<ion-refresher slot="fixed" (ionRefresh)="doRefresh($event)">
  <ion-refresher-content></ion-refresher-content>
</ion-refresher>

<ion-infinite-scroll threshold="100px" (ionInfinite)="loadData($event)">
  <ion-infinite-scroll-content></ion-infinite-scroll-content>
</ion-infinite-scroll>
```

**Other information:**
I'm submitting a PR for it.

**Ionic info:**

```
Ionic:

   ionic (Ionic CLI)             : 4.10.3 (~/.nvm/versions/node/v10.12.0/lib/node_modules/ionic)
   Ionic Framework               : @ionic/angular 4.1.0-dev.201902272232.d66b12b
   @angular-devkit/build-angular : 0.13.4
   @angular-devkit/schematics    : 7.3.4
   @angular/cli                  : 7.3.4
   @ionic/angular-toolkit        : 1.4.0

System:

   NodeJS : v10.12.0 (~/.nvm/versions/node/v10.12.0/bin/node)
   npm    : 6.4.1
   OS     : macOS High Sierra
```

PR #17660, opened 2019-03-02
fix(spinner): fix default spinner logic for relevant components

abennouna:
Currently, the default spinner icon is using the iOS platform default (**`lines`**), even in MD mode, for the following components:
- **`refresher`**
- **`infinite-scroll`**

In addition, the default **`loading`** spinner icon logic is not consistent with the above components. It does not try to get the `spinner` icon from `config` before falling back to platform default.

Finally, some test templates using the `infinite-scroll` component are not correctly using the loading _spinner_ and _text_ attributes.

- In the **`loading`** component, fall back to `spinner` icon config before platform default.
- In both **`refresher`** & **`infinite-scroll`** components, use platform default as final fallback.
- In **`infinite-scroll`** test templates, and the _basic_ **`virtual-scroll`** test template, fix loading _spinner_ and _text_ attributes.

**Ionic Version**: 4.x

**Fixes**: #17659


brandyscarney:
Thank you!

PR #17681, opened 2019-03-05
fix(split-pane): use proper grammar in warnings

kensodemann:

The former message said: split pane could not found any main node

That was grammatically incorrect and it made it sound like the split pane itself was doing the searching rather than being the entity that did not have a properly specified child "main" element.

On "cannot" vs. "can not" they are both technically correct according to Webster, but "cannot" seems to be preferred in this case with "can not" being preferred in cases where the "not" modifies another word, such as "can not only do this but also..."

- Fix the wording of the warning messages

**Ionic Version**:
4.x


brandyscarney:
Thanks Ken!

Issue #17587, opened 2019-02-22
ion-datetime default value is not local


kyleabens:
# Bug Report

**Ionic version:**

[x] **4.0.2**

**Current behavior:**
ion-datetime currently defaults to todays date if none is selected but does not give an option to apply an offset.  Because of this, the time is not local.

**Expected behavior:**

ion-datetime default value should be local.

**Ionic info:**

```
Ionic:

   ionic (Ionic CLI)             : 4.10.3 (/usr/local/lib/node_modules/ionic)
   Ionic Framework               : @ionic/angular 4.0.2
   @angular-devkit/build-angular : 0.12.4
   @angular-devkit/schematics    : 7.3.1
   @angular/cli                  : 7.3.1
   @ionic/angular-toolkit        : 1.3.0

Cordova:

   cordova (Cordova CLI) : 8.1.2 (cordova-lib@8.1.1)
   Cordova Platforms     : android 7.1.4, ios 4.5.5
   Cordova Plugins       : cordova-plugin-ionic-keyboard 2.1.3, cordova-plugin-ionic-webview 2.2.5, (and 14 other plugins)

System:

   ios-deploy : 2.0.0
   NodeJS     : v10.15.0 (/usr/local/bin/node)
   npm        : 6.4.1
   OS         : macOS Mojave
   Xcode      : Xcode 10.1 Build version 10B61
```

bkarv:
Beat me to it, yes same issue the default time is not local it seems like it is defaulting to the GMT time.

liamdebeasi:
Hi there,

Thanks for opening an issue with us! Would you be able to provide an example of what you are seeing versus what you are expecting? Thanks!

ingljo:
We are experiencing the same issue. 
Steps to reproduce:

Create ion-datetime. Set format to MMM DD, YYYY HH:mm.
Set client timezone to for example GMT+2.
When selecting date at 10:00 in local time, the picker hours for the default datetime is 08:00. 
Expected hours for default value should be 10:00 (local time).

liamdebeasi:
Hey there! 

I am working on a PR (#17656) that should fix this issue. If someone wouldn't mind testing this out on their end as well, that would be great. Thanks!

liamdebeasi:
Hi there,

This fix will be in the next release of Ionic. I have published a nightly build of Ionic with this fix (`npm install @ionic/core@dev`). Feel free to test it out and provide feedback if you have the time.

Thanks!

[...]

Issue #16723, opened 2018-12-13
bug(ion-datetime) format as MM only


Newbie012:
# Bug Report

**Ionic version:**

[x] **4.x**

**Current behavior:**

If you pick a month in `<ion-datetime displayFormat="MM"></ion-datetime>`, it won't select it. Once you try to select it again, the value will be displayed, but it won't patch the value.

**Expected behavior:**

Choosing a month should be one time, and it should patch the control value.

**Steps to reproduce:**

 - run `<ion-datetime displayFormat="MM"></ion-datetime>`.
 - pick a month.
 - see occurance.

**Related code:**
```
<ion-datetime displayFormat="MM"></ion-datetime>
```

```
insert short code snippets here
```

**Other information:**

**Ionic info:** 

```
insert the output from ionic info here
```

liamdebeasi:
Hi there,

Thanks for opening an issue with us. I am going to close this for now, but we will be re-evaluating the `ion-datetime` functionality and bugs soon during which we will take a closer look at this issue.

Thanks for using Ionic!

liamdebeasi:
Hi there,

This has been fixed via #17706 and will be in the next release of Ionic. I have published a nightly build of Ionic with this fix (`npm install @ionic/core@dev`). Feel free to test it out and provide feedback if you have the time.

Thanks!

Issue #16436, opened 2018-11-23
Ionic 4: DateTime - Start picker on today's date


bkarv:
#  Feature Request

**Ionic Info**
Run `ionic info` from a terminal/cmd prompt and paste the output below.

```
Ionic:

   ionic (Ionic CLI)             : 4.3.0 (/usr/local/lib/node_modules/ionic)
   Ionic Framework               : @ionic/angular 4.0.0-beta.15
   @angular-devkit/build-angular : 0.9.0-beta.3
   @angular-devkit/schematics    : 0.9.0-beta.3
   @angular/cli                  : 6.1.5
   @ionic/angular-toolkit        : not installed

Cordova:

   cordova (Cordova CLI) : 7.1.0
   Cordova Platforms     : browser 5.0.4
   Cordova Plugins       : cordova-plugin-ionic-keyboard 2.1.3, cordova-plugin-ionic-webview 2.2.0, (and 9 other plugins)

System:

   ios-deploy : 1.9.1
   NodeJS     : v10.10.0 (/usr/local/bin/node)
   npm        : 6.4.1
   OS         : macOS High Sierra
```

**Describe the Bug**
In DateTime when a 'max' is set, the date picker starts at at max value, which makes it a poor UI experience. It should start at current date/time or property to supply a starting point.

**Steps to Reproduce**
In Ionic 4 set Datetime 'max' to a large number and picker will start their.

**Expected Behavior**
Start at today's date or allow us to supply a starting pointl

**Additional Context**
<img width="476" alt="screen shot 2018-11-23 at 11 24 20 pm" src="https://user-images.githubusercontent.com/10624315/48943452-5b7f0d80-ef77-11e8-8917-1fe60f45ee8a.png">
.

bryce13950:
Have you tried adding an initial value to this as todays date?

bkarv:
I was trying to avoid dirtying the form. I can do a work around by having a viewValue and my dataValue. The viewValue is what the date time uses whilst the dataValue updates if the OK button is pressed which now can be controlled though the custom picker options.

However it’s seems a bit of work for something that logically should be available. If you look at angular material date picker which is a good comparison given it’s maturity and popularity, they have a whole section dedicated to setting the initial view. 

https://material.angular.io/components/datepicker/overview

bryce13950:
Yeah it's definitely not ideal, but I am just trying to find a way that you can get around it for the time being. We will let you know when it is fixed!

[...]

PR #17706, opened 2019-03-06
fix(datetime): default to local date

liamdebeasi:
This PR fixes an issue where datetimes are not localized to the user's timezone. This PR should fix a number of datetime issues.

- Create a UTC-like string that has the hours localized to the user's timezone _before_ generating columns in picker
- Set timezone offset before passing to `convertDataToISO` to ensure that function does not attempt to convert timezones by itself
- Added more tests

**Ionic Version**:

**Fixes**: #17587 #16436 #16723

### DIFF
--- a/.github/PROCESS.md
+++ b/.github/PROCESS.md
@@ -1,11 +1,12 @@
 # Process
 
-This document is to describe the internal process that the Ionic team uses for issue management and project planning.
+This document is to describe the internal process that the Ionic team uses for issue management, project planning and the development workflow.
 
 ## Table of contents
  * [Project Boards](#project-boards)
  * [Managing Issues](#managing-issues)
  * [Workflow](#workflow)
+ * [Releasing](#releasing)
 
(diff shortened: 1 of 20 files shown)


dominic-ks:
Hi,

Can I just ask, what version have these fixes gone into? I'm still experiencing this and I'm not sure how to check if I have the right version or not.

Thanks,

liamdebeasi:
This was fixed in Ionic 4.2. There was another bug regarding `null` values that was fixed in Ionic 4.4. You can view all changes for each Ionic version here: https://ionicframework.com/docs/release-notes

MKhowaja:
Is there a way to go back to the behavior before this change? Ie ignore the user's device timezone

MKhowaja:
@liamdebeasi One idea would be to guard the existing behaviour behind a prop that defaults to the new behaviour. This way there would atleast be a possibility of backwards compatibility.

PR #17708, opened 2019-03-07
Update item.tsx to detect if ion-item is clickable

jayordway:

Currently, the ion-item component does not by default show the detail arrow if in ios mode and the onclick attribute is present. If the item is clickable and in ios mode, the arrow should appear on the right side of the ion-item.

Correct the item.tsx file to look for the onclick attribute and show the arrow when clickable

**Ionic Version**:

4.1.1

**Fixes**: #
No bug created currently


brandyscarney:
Thanks for the PR! So the reason we didn't do this and instead added the `button` attribute is because users could add an event listener to listen to a click and we couldn't find a way to grab that event listener. While I see the purpose of adding this, I'm not sure that we want to make it only cover certain cases as it may be more confusing that you would only add the `button` attribute when you're using an event listener. See the following SO post: https://stackoverflow.com/questions/11455515/how-to-check-whether-dynamically-attached-event-listener-exists-or-not

jayordway:
@brandyscarney The documentation for ion-item is not correct then:

"If true, a detail arrow will appear on the item. Defaults to false unless the mode is ios and an href, onclick or button property is present."

Without this check, the details arrow doesn't show by default when in ios mode and an "onclick" is present.

brandyscarney:
@jayordway Correct. I believe that documentation was correct at one point until we realized it was confusing that it didn't work for event listeners. I'm all for ideas on how we can improve this. Maybe we just document that adding a click event listener won't automatically convert the item to a button but make the more common use case (click attributes) automatically use a button? 

Thoughts @liamdebeasi ?

jayordway:
I agree with correcting the documentation, otherwise folks will be confused when using onclick and ios, while not seeing the detail arrow.

liamdebeasi:
Yep, that sounds right. There isn't a clean way of detecting click events for vanilla JS as well as all frameworks. We should update the documentation to make this clearer.

liamdebeasi:
Hi there,

I created a PR that updates the documentation to highlight the correct `detail` property usage: https://github.com/ionic-team/ionic/pull/17861

Feel free to comment on it with any feedback you may have.

Thanks!

liamdebeasi:
Hi there,

The detail property usage changes have been merged (https://github.com/ionic-team/ionic/pull/17861), so I am going to close this.

Thank you for all of your work on this - we appreciate it!

Issue #17717, opened 2019-03-07
Add ticks attribute to range


sethlilly:
# Feature Request

**Ionic version:**

- [x] **4.x**

**Describe the Feature Request**

When using `ion-range` with `snaps` set to `true`, ticks are displayed; however, when using the `snaps` attribute the display of ticks may not always be desired.

**Describe Preferred Solution**

Add a `ticks` attribute (default: `true`) to `ion-range` to show/hide ticks when `snaps` is set to `true`.

**Describe Alternatives**

Because this is styled in shadow DOM, there really isn't a good way to override the display of these elements from outside the component aside from setting a CSS variable but that only results in hiding the ticks. The solution above prevents the ticks from being rendered in the first place, resulting in a cleaner DOM.

**Related Code**

...

**Additional Context**

PR forthcoming.

d0x:
Nice Idea with the Ticks Attribute! The scss variables to disable the ticks which @sethlilly  mentioned are:
```
        --bar-background: transparent;
        --bar-background-active: transparent;
        --ion-color-step-250: transparent;
```
(at least for ionic 4.0.2. Before that it was `--ion-background-color-step-250: transparent` instead of ` --ion-color-step-250`)

PR #17718, opened 2019-03-07
feat(range): add ticks attribute

sethlilly:
When using `range` with `snaps` set to `true`, ticks are displayed; however, when using the `snaps` attribute the display of ticks may not always be desired. This PR adds a `ticks` attribute to `range`.

- Add `ticks` attribute to `range`

**Ionic Version**:

- [x] 4.x

**Fixes**: #17717


core/src/components/range/range.tsx
```diff
@@ -105,6 +105,12 @@ export class Range implements ComponentInterface {
    */
   @Prop() step = 1;
 
+  /**
+   * If `false`, tick marks are not displayed.
```
brandyscarney:
```suggestion
   * If `true`, tick marks are displayed based on the step value.
```

This is super nitpicky but we try to always start each boolean with `If true` 🙂 It will require a npm build though.

sethlilly:
Thanks, @brandyscarney and @liamdebeasi!

Issue #17754, opened 2019-03-12
Texts disappear using slot=fixed in ion-content


yordis:
# Bug Report

**Ionic version:**

[x] **4.x**

**Current behavior:**

Using `slot="fixed"` makes all my text to disappear `except` the texts inside the buttons.

**Expected behavior:**

The texts shouldn't desappear.

**Steps to reproduce:**

1. Create a page
2. Add the code snippet below
3. Notice that the content disappears

**Related code:**

```
<ion-header>
  <ion-toolbar>
    <ion-buttons slot="start">
      <ion-back-button></ion-back-button>
    </ion-buttons>
    <ion-title>Sign Up</ion-title>
  </ion-toolbar>
</ion-header>

<ion-content>
  <div slot="fixed">
    <div></div>
    <div class="sign-up-buttons">
      <ion-button size="large" color="primary">Sign up with Email</ion-button>
      <ion-button size="large" class="facebook-button">
        <ion-icon slot="start" name="logo-facebook"></ion-icon>
        Sign up with Facebook</ion-button
      >
      <ion-button fill="outline" size="large" color="dark">
        <ion-icon slot="start" name="logo-google"></ion-icon>
        Sign up with Google</ion-button
      >
      <ion-button margin-top fill="clear">
        Already have an account?<ion-text color="primary"
          >&nbsp; Log in</ion-text
        >
      </ion-button>
    </div>
    <div class="legal-links">
      <p>By signing up, you agree to our <br /></p>
      <p>
        <ion-button fill="clear">
          Terms of Service
        </ion-button>
        and
        <ion-button fill="clear">
          Privacy Policy
        </ion-button>
      </p>
    </div>
  </div>
</ion-content>
```

**Other information:**

**Ionic info:**

```
➜ ionic info

Ionic:

   ionic (Ionic CLI)             : 4.11.0 (/Users/ubi/.config/yarn/global/node_modules/ionic)
   Ionic Framework               : @ionic/angular 4.1.1
   @angular-devkit/build-angular : 0.13.5
   @angular-devkit/schematics    : 7.3.5
   @angular/cli                  : 7.3.5
   @ionic/angular-toolkit        : 1.4.0

System:

   NodeJS : v11.6.0 (/Users/ubi/.asdf/installs/nodejs/11.6.0/bin/node)
   npm    : 6.5.0-next.0
   OS     : macOS Mojave
```

brandyscarney:
[...]

PR #17797, opened 2019-03-15
fix (ion-content): move slot="fixed" inside wrapper div

akshit-wwstay:
Fix the visibility of the content when passed through slot "fixed". 

- Move slot="fixed" inside the div wrapper, fixes the issue

**Ionic Version**: 4.10.3

**Fixes**: #17754


brandyscarney:
Thank you for the PR! This fix will break the FAB functionality:

| `master` | `akshit-wwstay:ion-content-slot-fix` |
| ---| ---|
| localhost_3333_src_components_fab_test_basic(iPhone 6_7_8) (1) | localhost_3333_src_components_fab_test_basic(iPhone 6_7_8) (2) |

The fixed content needs to be outside of the scrolling div as it should not scroll with the page. I believe the solution to this bug may be to add the following css to `content.scss` but it would require further testing:

```
::slotted([slot="fixed"]) {
  position: absolute;
}
```

🙂

DavidWiesner:
The problem is a missing z-index. When adding a z-index to the fixed slot, the text is visible again

dalemugford:
> The problem is a missing z-index. When adding a z-index to the fixed slot, the text is visible again. By default the fixed slot is behind the scroll content. So instead of adding a div there should be set a z-index on the fixed slot or the fixed slot should move above the scroll slot.

I'm finding the reverse— divs with `slot="fixed"` are _always_ on top of `ion-content`, with no way to make them _below_ ion-content. This prevents UI choices like having a parallax header with a fixed image behind content that can scroll overtop.

brandyscarney:
Thanks again for the PR! I added a PR here that should fix this: https://github.com/ionic-team/ionic/pull/19867

I'm going to close this as it has been merged. 🙂

PR #17799, opened 2019-03-15
fix(virtual-scroll): don't crash with empty cells list

wwoods:
In some conditions combining e.g. ion-infinite-scroll with ion-virtual-scroll, the cells list can be empty, meaning accessing `this.cells[this.cells.length - 1]` throws an error.

Correct the line to return zero when the cell list is empty rather than throwing an error.

-
-
-

**Ionic Version**: latest.

**Fixes**: #


liamdebeasi:
Hi there,

Does this PR fix a particular issue? If so, could you add it to the PR description?

Thanks!

wwoods:
Hi Liam, it does not fix an issue filed on GitHub, but an issue I've found
in my own project.  Figured it was simpler to just make a PR than both an
issue and a PR.

liamdebeasi:
Ok no prob. Do you have some steps to reproduce the original issue? This will help when testing out this PR. Thanks!

wwoods:
Sure - I can probably get around to setting up a MWE repo, but it might be
a couple weeks.  If that's too long, I'd imagine the following would work:

1. Create a default Ionic project with any template.
2. Replace the home page with a virtual scroll with divs, based on an items
list that is initially empty.
3. In ngOnInit, set a timeout for e.g. 500ms and add something to the items
list in the callback.

liamdebeasi:
Hi @wwoods,

The steps that you provided do not seem to trigger the issue. Can you re-verify that the steps you sent are correct?

I need to be able to verify the issue before we can get this merged in.

Thanks!

brandyscarney:
Thanks!

Issue #17801, opened 2019-03-15
ionic4/angular: overlay `getTop` can return `undefined`


Pierrci:
# Bug Report

**Ionic version:**

[x] **4.x**

**Current behavior:**
In the angular package, the return type of the `getTop` method in `OverlayBaseController` is `Promise<Overlay>`

**Expected behavior:**
The return type should be `Promise<Overlay | undefined>`

**Steps to reproduce:**
Use any component extending `OverlayBaseController` and call the `getTop` method without any alert currently present. With the `AlertController` for example, returned object will be of type `Promise<HTMLIonAlertElement>`, while it should be `Promise<HTMLIonAlertElement | undefined>`.

**Related code:**

```typescript
import { Component, OnInit } from '@angular/core';
import { AlertController } from '@ionic/angular';

@Component({
  selector: 'app-home',
  templateUrl: 'home.page.html',
  styleUrls: ['home.page.scss'],
})
export class HomePage implements OnInit {
  constructor(private alertController: AlertController) {}

  async ngOnInit() {
    const undefinedAlert = await this.alertController.getTop(); // typed as HTMLIonAlertElement
    alert(`type of alert is ${typeof undefinedAlert}`); // typeof undefinedAlert === 'undefined'
	undefinedAlert.dismiss(); // still possible because of the typing, will throw an error
  }
}
```

**Other information:**

**Ionic info:**

```
Ionic:

   ionic (Ionic CLI)             : 4.10.3 (/usr/local/lib/node_modules/ionic)
   Ionic Framework               : @ionic/angular 4.1.1
   @angular-devkit/build-angular : 0.13.6
   @angular-devkit/schematics    : 7.2.4
   @angular/cli                  : 7.3.6
   @ionic/angular-toolkit        : 1.4.0

System:

   NodeJS : v11.6.0 (/usr/local/bin/node)
   npm    : 6.5.0
   OS     : macOS Mojave
```

Pierrci:
Just submitted the related PR

brandyscarney:
It seems the PR was merged so I'm going to close this, thanks for the issue!

PR #17802, opened 2019-03-15
fix(angular): overlay.getTop can return undefined

Pierrci:
Add `undefined` to the return type of `getTop` method in `OverlayBaseController`

**Ionic Version**: 4.x

**Fixes**: #17801


mhartington:
Thanks for the PR!

PR #17826, opened 2019-03-19
tests(): increase timeout for fab and menu tests

liamdebeasi:
This PR adds more timeouts for the FAB and Menu tests to reduce the likelihood of slowdowns on CI causing screenshot diffs to be generated

- Wait for click event/render to propagate for FABs
- Wait for menu element to be ready before interacting with it

**Ionic Version**:

**Fixes**: #


brandyscarney:
The FAB tests look great. Are the menu tests supposed to be open though?

liamdebeasi:
Yes they are. I think my timeout change was too small -- let me revert

Issue #17851, opened 2019-03-21
Floating ion-label overlay ion-select value


eulerchavez:
# Bug Report

**Ionic version:**

[x] **4.x**

**Current behavior:**
Floating ion-label overlay ion-select value

**Expected behavior:**
Floating ion-label should not overlay ion-select value

![image](https://user-images.githubusercontent.com/18647380/54786094-86a7ff80-4bed-11e9-83ba-37bf9fd90f08.png)

**Related code:**

![image](https://user-images.githubusercontent.com/18647380/54786251-06ce6500-4bee-11e9-936f-7bd316fa5eaf.png)

![image](https://user-images.githubusercontent.com/18647380/54786270-1057cd00-4bee-11e9-9835-f40c1806e3b7.png)

**Ionic info:**

![image](https://user-images.githubusercontent.com/18647380/54786195-dd153e00-4bed-11e9-8d50-efa210f58210.png)

liamdebeasi:
Hi there,

Thanks for the issue. What browser/version are you seeing this issue on? Additionally, which version of Ionic are you using?

Thanks!

eulerchavez:
> Hi there,
> 
> Thanks for the issue. What browser/version are you seeing this issue on? Additionally, which version of Ionic are you using?
> 
> Thanks!

Hi,

I'm using Google Chrome, Version 73.0.3683.86 (64 bits) and Ionic 4.1.2

Greetings

liamdebeasi:
Hi there,

Thanks for the follow up! would you be able to provide a repository with the code required to reproduce this issue?

Thanks!

mburger81:
Hello we have the same issue! What you habe to say the problem is only on init with pres selecting select. For example on a editing form mask. In our opinion this is a timing problem with async select loading.

mburger81:
Hello we have the same issue! What you habe to say the problem is only on init with pres selecting select. For example on a editing form mask. In our opinion this is a timing problem with async select loading.

If you are in this situation you can select manually a value and after that the floating label is moving on top

mburger81:
This bug can be related to #15716 and could be fixed with #17860

liamdebeasi:
Hi there,

This issue has been resolved and will be available in the next release of Ionic. I have pushed a nightly build in case you would like to try it out and provide feedback. (`npm i @ionic/core@dev @ionic/angular@dev`)

Thanks for your patience as we worked to resolve this issue!

mburger81:
[...]

Issue #15716, opened 2018-09-22
Ion-Select not selected after API Call


abhilashbr92:
# Bug Report

**Describe the Bug**
I have a select box which display usernames. The list of usernames is got from a API call.

I have a Form Control which is assigned to the select box.

I have value in the form control but the item is displayed in the select box.

I have made a demo in stackblitz below.

https://stackblitz.com/edit/github-igwscn-8kylvc?file=src%2Fapp%2Fapp.component.ts

**Steps to Reproduce**
1. In app.component.ts there is a variable called delay which has value 3000.Now you see the select box has not selected any value
2. Change delay value to 0 and save.Now you see the select box has value selected.

**Expected Behavior**
With delay value 3000, after the username list is returned the name should be displayed in the box.
The issue doesn't exist if I use v3 ion-select or HTML5 select. It is occuring after updating app to v4.

vlapo:
Have a same issue. It looks like `this.text` property is not updated properly after selected option is loaded. 
https://github.com/ionic-team/ionic/blob/master/core/src/components/select/select.tsx#L189

abhilashbr92:
Hello Ionic team, any update on this. When can we expect the issue to be fixed?

abhilashbr92:
@vlapo Have you found any alternative about this issue?

vlapo:
Not at all. I think it is simple problem with render selected text after load options. But it has to be fixed in stencil component and I do not have experience with it.

andleer:
Providing another example. This is not API bound per se but simply time based.

My example has 3 separate `<ion-select>` elements each with a different delay for loading the respective bound arrays. The selection `<ion-select [(ngModel)]="selectedWidgetId">` is loaded before any delay is introduced. It doesn't seem to matter if this is set before or after the binding array is set. Each of the 3 controls is bound after a delay. The 10ms delay seems to never cause the issue. The 65 is intermittent and the 200 always causes the issue.

As the OP stated, this bug results in the selected element not being shown on the UI but it does immediately appear when you click on the item to change the selection and the popup does show the correct default selection.

I do wonder if this is simply a time issue or if there is some other event (hook?) occurring before the binding array is completed that causes the issue.

This repo contains a simple, single page that demonstrates the issue.
https://github.com/andleer/ionic4-select-bug/

[...]

PR #17860, opened 2019-03-22
fix(select): fix options being added async while overlay is open

liamdebeasi:
Fixes an issue where select options were not being updated when loaded asynchronously when the overlay was already open

- Repopulate options/inputs for the appropriate overlay when needed

**Ionic Version**:

**Fixes**: #15716 #17851


brandyscarney:
@liamdebeasi I added some more examples to the async test of the difference interfaces and upped the time. Code looks good to me, great fix!

PR #17861, opened 2019-03-22
docs(item): update item detail property usage

liamdebeasi:
Updates the documentation for item to better explain how to use the `detail` prop

- Update detail arrows documentation


brandyscarney:
I think instead of documenting how the detail property works - we should document what a "clickable" item is, how you can achieve it, and maybe refer to detail arrow as a part of that. Meaning:

1) how does an item become clickable? `href` property or `button` property where adding the `href` property will render a child `<a>` and `button` will render a child `<button>`.
2) how do these properties (or the act of being clickable) affect the item component? 
  - adds ripple effect for `md` mode, activated highlight for `ios` mode
  - adds the detail arrow by default to clickable `ios` items, or by setting the `detail` property

core/src/components/item/readme.md
```diff
@@ -2,10 +2,13 @@
 
 Items are elements that can contain text, icons, avatars, images, inputs, and any other native or custom elements. Generally they are placed in a list with other items. Items can be swiped, deleted, reordered, edited, and more.
 
+## Clickable Items
+
+An item is considered "clickable" if it has an `href` or `button` property set. Clickable items have a few visual differences that indicate they can be interacted with. For example, a clickable item receives the ripple effect on Material Design mode, has a highlight when activated, and has a [detail arrow](/#detail-arrows.
```
brandyscarney:
```suggestion
An item is considered "clickable" if it has an `href` or `button` property set. Clickable items have a few visual differences that indicate they can be interacted with. For example, a clickable item receives the ripple effect upon activation in `md` mode, has a highlight when activated in `ios` mode, and has a [detail arrow](/#detail-arrows) by default in `ios` mode.
```

Issue #17848, opened 2019-03-21
Support safe-area-inset for normal content


rossholdway:
# Bug Report

**Ionic version:**

[x] **4.1.2**

**Current behavior:**
When viewing content on a device that uses the CSS env safe-area-inset-* such as the iPhone X, there is no support to contain general content within an element that applies `safe-area-inset-left` or `safe-area-inset-right` other than in specific items such as ion-item. Therefore content displayed in a landscape orentation on devices that use safe-area-inset is obscured.

**Expected behavior:**
I'd expect content to not be obscured behind device screen features such as the iPhone notch.

**Steps to reproduce:**
You can see this issue in the ionic-conference-app. Simply run the app in an emulator and rotate it's orentation to landscape. Some content is obscured.

![Screenshot 2019-03-21 at 15 14 15](https://user-images.githubusercontent.com/2728159/54762521-33817d00-4bec-11e9-962f-476a0d093d19.png)

**Other information:**
A possible fix could be to extend the grid to apply `safe-area-inset-left` and `safe-area-inset-right` by default.

I'd be happy to work though this and submit a pull request but would like to know community / your thoughts first! Thanks.

rossholdway:
It'd be nice if something like this worked `--ion-grid-padding: 5px calc(5px + env(safe-area-inset-right)) 5px calc(5px + env(safe-area-inset-left));` or even expose as individual vars, setting the default to use safe-area-inset for left and right.

daem0ndev:
Looked into this one, adding the safe area padding to the grid fixes the `Food` label, but the Breakfast and time/location are sibling h1 and p elements that dont get the padding since they are not in the ion-grid.

really the developer should be intentional about when and how they are compensating for safe area padding. e.g. simply adding the padding to content makes it hard for list items to properly fill the horizontal space as the padding should be inset to the child and not the parent.

that being said, adding a rule like this should be pretty safe and would correct the example markup in the ionic conference app: 

```
ion-content > [padding] {
  padding-left: calc(env(safe-area-inset-left) + 16px);
  padding-right: calc(env(safe-area-inset-right) + 16px);
}
```

PR #17873, opened 2019-03-25
fix(grid): account for horizontal safe area insets

rossholdway:
Prevents content in the grid from being obscured by device features such as the iPhone X screen notch by applying horizontal safe-area-inset padding to the `ion-grid` component.

This uses the ionic global `--ion-safe-area-right` and `--ion-safe-area-left` to apply the padding, allowing the user to override the safe area, if required.

Apply `safe-area-inset-*` to the right and left padding of the grid.

**Ionic Version**: 4.1.2

**Fixes**: #17848

### DIFF
--- a/core/src/components/grid/grid.mixins.scss
+++ b/core/src/components/grid/grid.mixins.scss
@@ -27,7 +27,7 @@
     @include media-breakpoint-up($breakpoint) {
       $padding: map-get($paddings, $breakpoint);
 
-      @include padding($padding);
+      @include padding($padding...);
     }
   }
 }
--- a/core/src/components/grid/grid.vars.scss
+++ b/core/src/components/grid/grid.vars.scss
@@ -5,22 +5,30 @@
 // --------------------------------------------------
 
 /// @prop - The padding for the grid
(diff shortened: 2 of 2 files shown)


brandyscarney:
Thank you for the PR! The only issue I can see with this is if a user has a grid inside of something (such as an `ion-item`, `ion-toolbar` or any other component that automatically does this). It would add too much padding in that case. I wonder if we should just add some CSS utilities that will check for safe-areas?

sean-perkins:
Hello 👋 thanks for submitting this proposed PR! After discussing with the team, we have decided to close this PR without merging. 

The proposed changes could have unexpected side-effects with other components that react to safe area padding. Certain components like `ion-toolbar` have consistent expected implementation patterns, but with `ion-grid`, the component could be used in a design that does not want/need the additional safe area padding.

I think we need to do additional information gathering on developer pain points here and evaluate if this is:
- An implementation detail that developers are responsible for;
- A new API or property that allows developers to easily apply safe area padding;
- CSS utilities/etc.

Appreciate the work here, let me know if you have any questions!

PR #17875, opened 2019-03-25
feat(input): hide spinner on input type number

paulstelzer:
Hide spinner on input type number

**Ionic Version**:

**Fixes**: #


manucorporat:
What's your use case? do you need it visible at some point?

paulstelzer:
@manucorporat It should be disabled at any time, we do not want any spinners there

brandyscarney:
Thank you for the PR @paulstelzer! So if this were a shadow component it would be solved using [CSS Shadow Parts](https://www.w3.org/TR/css-shadow-parts-1/) which should arrive early 2020, but since this is a scoped component it can actually be done now with CSS. Here's a Codepen example: https://codepen.io/brandyscarney/pen/oNNMazp?editors=0110

I'm going to close this as we try to avoid adding component properties where CSS will work, but we appreciate it!

Issue #15181, opened 2018-08-15
ion-back-button does not work correctly together with angular router navigate option "replaceUrl" (V4)


kkbear1985:
# Bug Report

**Ionic Info**
Run `ionic info` from a terminal/cmd prompt and paste the output below.

```
Ionic:

   ionic (Ionic CLI)          : 4.0.1 
   Ionic Framework            : @ionic/angular 4.0.0-beta.2
   @angular-devkit/core       : 0.7.3
   @angular-devkit/schematics : 0.7.3
   @angular/cli               : 6.1.3
   @ionic/ng-toolkit          : 1.0.6
   @ionic/schematics-angular  : 1.0.4

Cordova:

   cordova (Cordova CLI) : 7.0.1
   Cordova Platforms     : none

System:

   Android SDK Tools : 26.0.2
   NodeJS            : v8.11.3 (C:\Program Files\nodejs\node.exe)
   npm               : 6.2.0
   OS                : Windows 7

Environment:

   ANDROID_HOME : D:\Android\sdk

```

**Describe the Bug**
ion-back-button does not work correctly together with angular router replaceUrl option

**Steps to Reproduce**
if i navigate from "/home" to "/login" and then navigate to "/pageC" by setting:
this.router.navigate(["/pageC"], {replaceUrl:true});

the browser back button correctly take me back to "/home" instead of "/login "
but ion-back-button still take me back to "/login"

**Related Code**
If you are able to illustrate the bug with an example, please provide a sample application via an online code collaborator such as [StackBlitz](https://stackblitz.com), or [GitHub](https://github.com).

**Expected Behavior**
ion-back-button should take me back to "/home" instead of "/login " while the angular router option replaceUrl is set to "true"

**Additional Context**
List any other information that is relevant to your issue. Stack traces, related issues, suggestions on how to fix, Stack Overflow links, forum links, screenshots, OS if applicable, etc.

daem0ndev:
I am also experiencing this same issue. In doing further investigation, I believe the issue is that when doing a router.navigate with replaceUrl: true, it is not updating the url / fullPath within the StackController of the ion router outlet. 

In my scenario my page has child routing, and on enter it sets a default child route programatically. So for example navigating to `/home` will automatically do a relative navigation to `/home/child` using replaceUrl: true. Now from here if I navigate to `/another-page` the back button takes me to `/home` instead of `/home/child` which is very similar to the op's issue.

[...]

PR #17879, opened 2019-03-26
fix(angular): account for replaceUrl option

liamdebeasi:
There was an issue where our StackController was not taking into account `replaceUrl`. So when going from page A->page B->page C (replace url), page B was still in the stack. Going back via `ion-back-button` from page C would go back to page B, whereas going back from page C via the browser back button would go back to page A.

- Check for `replaceUrl` on current navigation when inserting the new view

**Ionic Version**:

**Fixes**: #15181


KillerCodeMonkey:
damn i just thought of a PR and now i have to see this one :D.

Nice thing ;)

mhartington:
LGTM :shipit:

PR #17885, opened 2019-03-26
[Vue] Update ion-range usage

michaeltintiuc:
As per https://github.com/ionic-team/ionic/issues/17878

The usage docs aren't clear on how to set default values for ion-range with dual-knobs


michaeltintiuc:
@liamdebeasi thanks for the comments, addressed!

brandyscarney:
Thanks @michaeltintiuc! 🙂

Issue #16954, opened 2019-01-04
ActivatedRoute is not updated correctly when using ion-router-outlet


hakimio:
# Bug Report

**Ionic version:**
4.0.0-rc.0

**Current behavior:**
Getting correct activated route using Angular ActivatedRoute service only works the first time when "ion-router-outlet" is used and router NavigationEnd event occurs.

**Expected behavior:**
Getting correct activated route using Angular ActivatedRoute service works every time NavigationEnd event occurs and "ion-router-outlet" is used.

**Steps to reproduce:**
Try to get activated route every time Angular router fires NavigationEnd event.

**Related code:**
Following code correctly returns data object of the activated route when Angular "router-outlet" is used but only works the first time when Ionic "ion-router-outlet" is used:

```typescript
import {Component, OnInit} from '@angular/core';
import {ActivatedRoute, NavigationEnd, Router} from '@angular/router';
import {filter, map, mergeMap} from 'rxjs/operators';

@Component({
    selector: 'app-foo',
    templateUrl: './foo.component.html',
    styleUrls: ['./foo.component.scss']
})
export class FooComponent implements OnInit {

    constructor(
        private router: Router,
        private activatedRoute: ActivatedRoute
    ) {}

    ngOnInit() {
        this.router.events
            .pipe(
                filter((event) => event instanceof NavigationEnd),
                map(() => this.activatedRoute),
                map((route) => {
                    while (route.firstChild) {
                        route = route.firstChild;
                    }
                    return route;
                }),
                mergeMap((route) => route.data)
            )
            .subscribe(
                (routeData) => {
                    console.log(routeData);
                }
            );
    }

}
```

**Other information:**
daem0ndev explains the issue quite well in [this comment](https://github.com/ionic-team/ionic/issues/16736#issuecomment-448408927).
This issue is quite important and it should be blocker for stable version 4 release.

As a workaround "IonicRouteStrategy" can be removed from app module providers until the issue is fixed.

**Related issues:**
[ActivatedRoute.queryParams break after navigation](https://github.com/ionic-team/ionic/issues/16736)
[StackController Page reuse uses old Page data](https://github.com/ionic-team/ionic/issues/16516)

**Ionic info:** 
```
Ionic:

   ionic (Ionic CLI)             : 4.6.0
   Ionic Framework               : @ionic/angular 4.0.0-rc.0
   @angular-devkit/build-angular : 0.11.4
   @angular-devkit/schematics    : 7.1.4
[...]

Issue #16736, opened 2018-12-14
[4.0.0.beta-17] ActivatedRoute.queryParams break after navigation


h4rm:
# Bug Report

**Ionic version:**
4.0.0.beta-17 and 4.0.0.beta-18

**Current behavior:**
Subscriptions on ActivatedRoute.queryParams do not fire after navigating away from and returning to that page. This is particullary problematic for base pages of tabs that are cached. The problem does not occur on sub-pages because they are re-initialized i.e. onInit fires again and again (e.g. on speaker-detail in ionic-conference app).

**Expected behavior:**
Navigation should not interfere with the queryParams subscription.

**Steps to reproduce:**
1. Clone ionic-conference-app
2. Add code below
3. Test with "Test" button on speakers list. Observe console log of queryParams. Navigate to speaker detail and back. Try "Test" button again and observe no console output.
4. Test same with Test button speaker detail page. Navigate from base to detail and back and to detail again - logs are fired.

**Related code:**

`speaker-list.html`
```...
<ion-content class="outer-content">
   <!-- TEST BUTTON -->
  <ion-button (click)="test()">Test</ion-button>

  <ion-list>
...
```

in `speaker-list.ts` modify
```
...
export class SpeakerListPage implements OnInit {
  speakers: any[] = [];

  constructor(
    public actionSheetCtrl: ActionSheetController,
    public confData: ConferenceData,
    public inAppBrowser: InAppBrowser,
    private route: ActivatedRoute,
    private router: Router
  ) {}

  ngOnInit() {
    console.log('Init');
    this.route.queryParams.subscribe(test => console.log(test));
  }

  test() {
    this.router.navigate(['./'], { queryParams: { test: Math.random() }, relativeTo: this.route });
  }
...
```

**Ionic info:** 
```
Ionic:

   ionic (Ionic CLI)             : 4.5.0 (C:\Users\Ferdinand\AppData\Roaming\npm\node_modules\ionic)
   Ionic Framework               : @ionic/angular 4.0.0-beta.17
   @angular-devkit/build-angular : 0.10.7
   @angular-devkit/schematics    : 7.1.3
   @angular/cli                  : 7.1.3
   @ionic/angular-toolkit        : 1.2.0

Cordova:

   cordova (Cordova CLI) : 8.1.2 (cordova-lib@8.1.1)
   Cordova Platforms     : android 7.1.4
[...]

Issue #16534, opened 2018-11-30
[4.0.0-beta.17] routerLink not working after first successful forward/back navigation


h4rm:
# Bug Report

**Ionic Info**
```
Ionic:

   ionic (Ionic CLI)             : 4.4.0 (/Users/ben/.nvm/versions/node/v10.13.0/lib/node_modules/ionic)
   Ionic Framework               : @ionic/angular 4.0.0-beta.17
   @angular-devkit/build-angular : 0.11.0
   @angular-devkit/schematics    : 7.1.0
   @angular/cli                  : 7.1.0
   @ionic/angular-toolkit        : 1.2.0

Cordova:

   cordova (Cordova CLI) : 8.0.0
   Cordova Platforms     : none
   Cordova Plugins       : no whitelisted plugins (0 plugins total)

System:

   ios-deploy : 1.9.2
   ios-sim    : 5.0.6
   NodeJS     : v10.13.0 (/Users/ben/.nvm/versions/node/v10.13.0/bin/node)
   npm        : 6.4.1
   OS         : macOS
   Xcode      : Xcode 9.4.1 Build version 9F2000
```
**Describe the Bug**
After navigating to another route with routerLink (e.g. relative navigation in tabs via ['../', 'session', session.id]) and after going back via ionic back button, which works as intended, the original routerLink is not working anymore.

**Steps to Reproduce**
1. Clone ionic-conference-app
2. (Change @ionic/angular to beta17)
3. in `schedule.html` of the schedule, switch href with `[routerLink]="['../', 'session', session.id]"`
4. Navigate to session detail and back, try again

**Expected Behavior**
Navigation should still work.
Maybe this has to do with your recent changes @manucorporat.

eduboxgithub:
I also found this bug.

It also happens when you navigate to a details page with NavController.

For example: 
`this.NavController.navigateForward(['../details', item.id], true, { relativeTo: this.ActivatedRoute })`

When you press the back button on details page it goes back, but you cannot navigate anymore to the details page.

I started a new sidemenu project, added a details page and linked a click event to a button on a item from the list. This navigation one happens once.

You can download the source to test it: https://bit.ly/2FQKiWc

manucorporat:
Does it work in beta.16?

eduboxgithub:
I thought it was working in the previews versions, because I was using the Angular Router directly.

[...]

PR #17888, opened 2019-03-26
fix(router-outlet): support relative router links

daem0ndev:
This PR resolves an issue where relative router links break after a forward and back navigation to a component that has already been created in the ion router outlet stack.

- Create activated route proxy for each component instance activated in the outlet
- Update the underlying observables on the proxy so consumers need not be aware of implementation details

**Ionic Version**:
4.x

**Fixes**: #16534, #16736, #16954


mhartington:
Merged! Thanks @daem0ndev for all your work on this and doing the investigation.

rgolea:
What version will this launch on?

liamdebeasi:
This will be in the next release of Ionic. I am not certain of the version yet.

rgolea:
Great! Some minor breaking changes I've got as I just tested it, is that if you were using something like the example below, you will get `route.params` as undefined and it will get instanciated afterwards. You will have to translate that code inside `ngOnInit` (no biggie, just to write it down in breaking changes):
```
@Component({
  ....
})
export class MyComponent(){
    public id$ = this.route.params.pipe(map(params => params.id));

    constructor(
         private readonly route: ActivatedRoute
    ){}
}
```

mhartington:
I wouldn't consider this a breaking change actually, as the example above is really an anti-pattern/not-recommended. We can make a note, but it should have been avoided to begin with.

rgolea:
@mhartington sure! Just a note will go a long way. Thank you for the fix @daem0ndev!

daem0ndev:
@rgolea @mhartington I created this PR to resolve the breaking change. I do agree its breaking since in a vanilla angular app, you can bind to the observables prior to ngOnInit. https://github.com/ionic-team/ionic/pull/17914

PR #17914, opened 2019-03-30
fix(angular): make route observables available on constructor

daem0ndev:
The fix introduced in https://github.com/ionic-team/ionic/pull/17888 also broke certain expectations as observables on ActivatedRoute were no longer available at the constructor but only after ngOnInit, this could cause many breaking changes for users that expect to use the observables as they would in a vanilla angular app.

- Create observable proxies upon component construction

**Ionic Version**: 4.x

**Fixes**:


rgolea:
You sir, hats off to you my friend!

Issue #17424, opened 2019-02-07
slide effect not working in ionic 4.10.2 


thiagomatrix:
# Bug Report

**Ionic version:**

[x] **4.x**

**Current behavior:**

**Expected behavior:**

**Steps to reproduce:**

**Related code:**

```
**in arquive.ts**

  slideOpts= {
    effect: 'fade',
    loop: true,
    speed:1500,
  };

  slidesDidLoad(slides: IonSlides) {  
    slides.startAutoplay();
  }

**in arquive.html**

<ion-slides #background (ionSlidesDidLoad)="slidesDidLoad(background)" [options]="slideOpts">
 <ion-slide class="welcome-slide">
 <h2>Slide 1</h2>
  </ion-slide>

  <ion-slide style="background-color: blue">
    <h2>Slide 2</h2>
  </ion-slide>

  <ion-slide style="background-color: red">
    <h2>Slide 3</h2>
  </ion-slide>
</ion-slides>


```

**Other information:**

**Ionic info:** 

```
Not working in ionic 4.10.2 slide effect

only show effect flip, not set out effect.

```

Nazirovich:
It's not working... I really hope they solve this issue asap... The fade effect on the sliders is much needed for my work.

LuizTokuhara:
Yes, ion-slides are kind of buggy, only custom effect is working... if you put 2 or more sliders on same page it act strange too, some can loop and others not.

yshnv:
ion-slides properties not working correctly.
Unable to change effects of slides.

edgarelc:
Same issue here... None of the effects in the swiper API works at all ("slide", "fade", "cube", "coverflow" or "flip"). The only one that works is "slide" which is the default one.

microwin168:
I have same issue.Will this issue be fixed in plan?

drew-r:
I'm finding something is calling toString on the options object before it is passed to swiper.
From `normalizeOptions`:
![image](https://user-images.githubusercontent.com/3891910/53375723-ef37ef80-3953-11e9-813a-e81194a35cee.png)

microwin168:
@liamdebeasi Is this a bug?  I need this function so much.
.Will this issue be fixed in plan?
Thanks for reply!

microwin168:
@thiagomatrix  Have you resolve this issue?

liamdebeasi:
Hi there,

[...]

PR #16616, opened 2018-12-06
fix(slides): add effect plugins for swiper

CFT-Chris:
#### Short description of what this resolves:
Missing Coverflow, Cube, Fade and Flip swiper effects for ``ion-slides``

#### Changes proposed in this pull request:

- Add needed plugins to swiper builder

**Ionic Version**: 4.0.0-beta.17

**Fixes**: #15857

PR #17959, opened 2019-04-04
fix(slides): expose interface to provide custom animations

liamdebeasi:
This PR exposes the interface to allow users to tap into certain callbacks in Swiper. The result is that users can provide their own custom animations, allowing them to use fade, cube, flip, coverflow, etc, without us having to introduce the code into the framework code base.

- Ensure that users can tap into swiper callbacks but _not_ override the callbacks Ionic uses to make `ion-slides` work
- Added documentation on how to add custom animations
- Add demos for all 4 additional animations
- Fixed usage for slides

**Fixes**: #17424

This PR would allow #16616 to be closed as well.


kalicki:
Hey @liamdebeasi 

I made a video! Check this bug: https://send.firefox.com/download/4eb61d0b2a643f83/#KHRL1rCut418GnvbZgheCA

liamdebeasi:
Can you post the code you are using for that? Thanks!

kalicki:
Sorry =/
Actually this is with the current slides of v4.2. Can you guarantee that this does not happen in your PR ?! 

I'll test your PR

kalicki:
Simple code:

```
<ion-slides pager="false" [options]="slideOptions" #tagSlides>
        <ion-slide *ngFor="let tag of tags">
          <ion-card class="tag-image" mode="ios" (click)="goPage(tag)">
            <img src="{{tag.image}}" alt="" />
          </ion-card>
          <h2>{{tag.title}}</h2>
        </ion-slide>
      </ion-slides>
```

```
slideOptions = {
    loop: false,
    initialSlide: 0,
    slidesPerView: 1,
    centeredSlides: false,
    spaceBetween: 10,
    effect: 'coverflow',
    coverflowEffect: {
      rotate: 0,
      stretch: 0,
      depth: 250,
      modifier: 1.93,
      slideShadows: false
    }
  };
```

liamdebeasi:
Hi there,

If you have found a bug in `ion-slides` in the latest stable, please open a new bug report here: https://github.com/ionic-team/ionic/issues/new?assignees=&labels=&template=bug_report.md&title=bug%3A+

This PR is to expose interfaces for custom animations and does not fix any other bugs.

Thanks!

brandyscarney:
I'm a bit confused by the usage:

1) Do I have to pass effect now? What's the point of passing `slide` if the default is slide and `flip` if I still have to define the effect?
2) Where can I find more animation examples? I know there are links to the framework tests but there are some problems with this
    - what if we get rid of those tests? 
    - what if we want something that isn't listed?

I think it would be better to just show the usage for them instead of having it in a test and then link to where it came from (probably swiper docs?)

liamdebeasi:
1. Yea for some of them there is corresponding CSS so you'll need to pass in the effect name. There's no point in passing in "slide" if it's the default.

2. Good question - still trying to figure out a good place for these. We provided the remaining animations that are bundled by default in Swiper, but beyond that that's on the user to come up with. For the actual code, I got it from the Swiper repo so unless we host the code somewhere, we're just going to be pointing users to some github repo.

liamdebeasi:
Took a look again and seems like this was due to us always explicitly saying to use the `slide` effect internally. I updated it so you don't need to provide any `effect` value and swiper will just default to `slide` if the animation handlers are not overriden.

brandyscarney:
Friendly reminder, don't forget about this comment: https://github.com/ionic-team/ionic/pull/16616#issuecomment-480042124

core/src/components/slides/readme.md
```diff
@@ -15,6 +15,419 @@ http://www.idangero.us/
 
 Licensed under MIT
 
+### Custom Animations
+
+By default, Ionic ships with the `slide` animation effect built in. Custom animations can be provided via the `options` prop. Examples for other animations can be found below.
```
brandyscarney:
```suggestion
By default, Ionic slides use the built-in `slide` animation effect. Custom animations can be provided via the `options` property. Examples of other animations can be found below.
```

CFT-Chris:
I'm a bit late to this party, but I have seen the changes on the docs.

Has anyone looked at [my suggestion](https://github.com/ionic-team/ionic/issues/16887#issuecomment-450424820) to just add a `swiperlib` attribute so that all swiper plugins can be exposed to "advanced" users looking for them?  I think the current stream of changes to `ion-slides` is akin to running on a treadmill: every new feature someone wants from the Slider library won't be readily available to someone using `ion-slides` until another PR is pushed.

Me personally, I need the "navigation" arrows, and I have a [branch](https://github.com/ionic-team/ionic/compare/master...CFT-Chris:slides-navigation) ready that proposes the required changes to `ion-slides` to make them work. I've seen others looking for "parallax" effects, virtual slides, etc. and of course this PR addressed those looking for the flip/coverflow/fade/cube effects.

I'm not sure if it's a great developer experience to have devs [copy/paste chunks of swiper code](https://ionicframework.com/docs/api/slides) into their slides options to make them work.

If we allow devs to import their own Swiper library (custom built with the features they want) and have `ion-slides` consume that library per my [suggestion here](https://github.com/ionic-team/ionic/issues/16887#issuecomment-450424820), then all Swiper options would be available to them with all the added Ionic benefits that come with `ion-slides`.

Thoughts?

Issue #17854, opened 2019-03-22
Ionic 4 - styling ion-toast for full width, problem with left/right value of .toast-wrapper


entom:
**Ionic version:**
[x] **4.x**

**I'm submitting a ...** 
[ ] bug report
[x] feature request

**Current behavior:**
Ion-toast is showing on position left: 8px, right: 8px.

**Expected behavior:**
I would like to show ion-toast on full width (100% width) on screen, I can set --width: 100%;
But there is left 8px setting which i'm not able to overwrite with any settings. .toast-wrapper is in shadow-root.

**Steps to reproduce:**
Show toast on screen, try to make it full width.

**Related code:**
```
.toast-wrapper {
    left: 8px;
    right: 8px;
...
}
```

**Other information:**

**Ionic info:** 
```
Ionic:

   ionic (Ionic CLI)             : 4.12.0 (/usr/local/lib/node_modules/ionic)
   Ionic Framework               : @ionic/angular 4.1.2
   @angular-devkit/build-angular : 0.13.6
   @angular-devkit/schematics    : 7.2.4
   @angular/cli                  : 7.3.6
   @ionic/angular-toolkit        : 1.5.0

Cordova:

   cordova (Cordova CLI) : 8.0.0
   Cordova Platforms     : android 8.0.0
   Cordova Plugins       : cordova-plugin-ionic-keyboard 2.1.3, cordova-plugin-ionic-webview 3.1.2, (and 33 other plugins)

System:

   NodeJS : v11.12.0 (/usr/local/bin/node)
   npm    : 6.9.0
   OS     : macOS Mojave
   Xcode  : Xcode 10.1 Build version 10B61
```

entom:
I'm working with it.

PR #17961, opened 2019-04-04
fix(toast): allow to change position start/end of toast

entom:
This allows user to set start and/or end position of toast-wrapper by using --start, --end.

-
-
-

**Ionic Version**:
4.2.0
**Fixes**: #17854


entom:
Ok, no problem. I have created some tests.

liamdebeasi:
Hey there!

Sorry for the delay in reviewing this. We had an update to toast that we were trying to get merged in, and unfortunately there were some merge conflicts created as a result.

Would you be able to resolve the conflicts? After that I can take another look.

Thanks! 🙂

entom:
Ok, no problem. Thanks for response. I have resolved conflicts and merged changes.

brandyscarney:
Thanks for the PR! We made some changes that I just wanted to make aware, the big one being renaming the following:

```
left -> start
right -> end
```

The main reason for this is because in RTL applications the `left` is actually the `right`, so in many places we are using `start` and `end` now for RTL support. 🙂 

I made a couple changes to the implementation but this is good to go! We can merge it once the build passes.

brandyscarney:
Merged, thank you!!

Issue #17899, opened 2019-03-27
ion-item-sliding programatic open method


karaggeorge:
# Feature Request

**Ionic version:**
[x] **4.x**

**Describe the Feature Request**

`ion-item-sliding` currently has no way of opening programatically. The components methods include `close`, but there's no `open`. 

That's usually fine for mobile, but when rendering on desktop, it's not natural to click and drag, so it's really often that a programatic way to open it is required.

I've seen a ton of work arounds trying to mimic the internal behavior, but this should probably be an api method that exists on the component. Even if there's no animation it literally just translates the item to the right position.

**Describe Preferred Solution**

Offer a `open` method on the slide item

**Describe Alternatives**

Trying to manually mimic the behavior, which is different for each ionic version, and doesn't seem to be possible in Ionic 4

shreeshbhat:
@brandyscarney : I would like to pick this up if this is beginner friendly.
How should I go about tackling this?
I am planning to add open() method in this file : 
https://github.com/ionic-team/ionic/blob/master/core/src/components/item-sliding/item-sliding.tsx

Not sure about implementation though.

shreeshbhat:
Yep, I would like to work on this and submit a pull request.

untilbit:
I need this to migrate from Ionic 3 to Ionic 4.

In v3 i used something like this to open it:

    const swipeAmount = 100;
    itemSliding.startSliding(swipeAmount);
    itemSliding.moveSliding(swipeAmount);
    itemSliding.setElementClass('active-options-right', true);
    itemSliding.setElementClass('active-swipe-right', true);
    item.setElementStyle('transition', null);
    item.setElementStyle('transform', 'translate3d(-' + swipeAmount + '%, 0px, 0px)');

shreeshbhat:
@brandyscarney : Should I make changes to list.tsx as well to open all closed sliding items?

shreeshbhat:
@brandyscarney : Thanks for the prompt response.
I saw in the folder test/preview/index.html, there was a closeOpened() method in list, that was calling item-sliding's method.
Should I create a separate test for open() in basic or standalone?

> The open method will likely need to close open sliding items when it gets called.

This open method should behave like a toggle?

brandyscarney:
[...]

PR #17964, opened 2019-04-05
feat(item-sliding): add open method

shreeshbhat:
Adds a feature method to programmatically open an item-sliding component

- Fetches ion-item-options using querySelector
- Sets ion-item-options' display to flex, so that we can store its width value
- Sets openSlidingItem from this.el, so that closeOpened() method will work 
- Resets ion-item-options' display
- Use setOpenAmount() method to move the item-sliding component to the specified width

This change takes care of closing item-sliding component when we click away from the component.
Also, closeOpened() method works across a list of item-sliding component after using open() method.

**Ionic Version**: 4.2.0
**Fixes**:  #17899


shreeshbhat:
I tried smoothing the transition while opening by using the following code before setting display to flex : 
```options.style.transition = 'transform 0.6s cubic-bezier(.36, .66, .04, 1)';```
It didn't help. Current behavior is it just opens up spontaneously.

**Comment by shreeshbhat** [comment]
@brandyscarney @liamdebeasi  : Please review and let me know if any changes are required.

**Comment by shreeshbhat** [comment]
Ah, good catch. How should it behave on multiple open() calls if it is already opened?

**Comment by liamdebeasi** [comment]
If the item is already open, calling `open()` again should do nothing.

**Comment by shreeshbhat** [comment]
Sounds good. Will make that change and try to get easing animation done tonight.
Also, I see that you added a merge commit. 
Would you like me to rebase from master and squash it down to one commit in PR?

**Comment by liamdebeasi** [comment]
Sounds good re: changes. Don't worry about rebasing -- we squash the commits when we merge into `master`👍

**Comment by shreeshbhat** [comment]
@brandyscarney @liamdebeasi : 
Updated PR to include easing transition as well as multiple click on open() behavior.

**Comment by shreeshbhat** [comment]
@liamdebeasi : This is ready for review. Please let me know if any changes are required.

**Comment by untilbit** [comment]
Waiting for the merge of this feature

File: core/src/components/item-sliding/item-sliding.tsx [hunk 1]
```diff
@@ -116,6 +116,24 @@ export class ItemSliding implements ComponentInterface {
     return Promise.resolve(this.getSlidingRatioSync());
   }
 
+  /**
+   * Open the sliding item.
+   */
+  @Method()
+  async open() {
+    if (this.openAmount !== 0 && this.item !== null) {
```
liamdebeasi:
```suggestion
    if (this.openAmount === 0 && this.item !== null) {
```

I think this should be checking if `openAmount` is 0, otherwise the item will never open if closed.

core/src/components/item-sliding/item-sliding.tsx
```diff
@@ -116,6 +116,24 @@ export class ItemSliding implements ComponentInterface {
...
+      const options = this.el.querySelector('ion-item-options');
+      if (options) {
+        this.item.style.transition = 'transform 500ms cubic-bezier(0.36, 0.66, 0.04, 1)';
```
liamdebeasi: You shouldn't need to manually apply this transition style here. Try setting `this.state = SlidingState.Enabled`. That _should_ fix it

shreeshbhat: Will make this change & test it out tonight.

shreeshbhat: This change didn't work. :(

shreeshbhat: I will keep the transition style change as it is since it is working. Please let me know if that's okay?

core/src/components/item-sliding/item-sliding.tsx
shreeshbhat: Ah yes, I agree, it should be an equals check here.
Making this change now.

shreeshbhat:
>
> After fixing the issue I noted in my other comment, if I click the open and close buttons really quickly, the item-sliding element seems to get pushed all the way offscreen
>

That's an interesting behavior. I am finding it hard to figure out why this is happening.
Looks like when you are trying to open a sliding item in the middle of closing it, it gets the options.offsetWidth as 734. Otherwise, the width is 146. 
Also, this huge change in offsetWidth changes depends on window size.

I tried clearing timeout if it exists using below code at the start of open(), it didn't help :
```
    if (this.tmr !== undefined) {
      clearTimeout(this.tmr);
      this.tmr = undefined;
    }
```
@liamdebeasi : How should I proceed now?

shreeshbhat:
I think this issue has something to do with setTimeout() being called inside setOpenAmount().
Is the expected behavior that if we quickly call open() in the middle of closing it, we should do nothing during that open() call?

shreeshbhat:
I have added a delayed open() method call for dirty state.
If you want, I can ignore such quick method open call in middle of close().
Waiting for your feedback.

liamdebeasi:
Hey there! I pushed a few updates to your branch:

1. Added the ability to specify a side when calling `open` (start or end). Right now this doesn't support RTL, but it's part of a bigger RTL overhaul: https://github.com/ionic-team/ionic/issues/17012.
2. Fixed a bug with "expandable" options
3. Removed the setTimeout on open -- I don't think it's needed anymore (things seem to be working fine with out it)

Let me know if you run into any issues/have any feedback!

shreeshbhat:
Ooh, this looks more complicated than I thought.
Could you please explain more about non-expandable buttons?

liamdebeasi:
When calling `open` on an item with buttons that expand, the expanded button seems to cover up the non-expanding buttons:

| Wrong | Correct |
| ------- | ------- |
| image| image

shreeshbhat:
Ah thanks for clarifying. I wasn't aware of difference between expanding and non-expanding buttons.
Will test it out tonight and let you know if I can figure out what's happening.

liamdebeasi:
I just pushed an update that resolves the expandable button issue. Should be good now -- feel free to test and let me know if you run into any news issues.

Thanks!

core/src/components/item-sliding/readme.md
```diff
@@ -692,6 +692,22 @@ Type: `Promise<number>`
...
 
+### `open(side?: string) => Promise<void>`
+
+Open the sliding item.
+
+#### Parameters
+
+| Name   | Type     | Description |
+| ------ | -------- | ----------- |
+| `side` | `string` |             |
```
brandyscarney: Can we document the description? It should be like the following in the code:

```
  /**
   * Open the sliding item.
   *
   * @param side The side of the options to open. Defaults to `"end"`.
   */
```

Or however you want to describe it. It could just be 

> If a side is not provided, it will open the options on the `"end"` side.

shreeshbhat: Yep, I can update the PR with this change tonight.

liamdebeasi: I just took care of it. Thank you though 🙂

brandyscarney:
This is coming along nicely!

@liamdebeasi I tested it out by opening each item in a page one after another and I am concerned by two things:

- The previously opened item isn't closing
- When there's only one side it does nothing

I know I said we should default to the `end` side - but I wonder if we shouldn't have it behave like menu-toggle?

```
  /**
   * Optional property that maps to a Menu's `menuId` prop.
   * Can also be `start` or `end` for the menu side.
   * This is used to find the correct menu to toggle.
   *
   * If this property is not used, `ion-menu-toggle` will toggle the
   * first menu that is active.
   */
```

Where it looks for the first active option and opens that one, unless side is provided. I'm not sure here, but the first one I think is an issue. Other than that though this is working nicely!

untilbit:
This feature wasn't included in Ionic 4.3. When can we expect it?

brandyscarney:
This looks great! Good job @shreeshbhat and @liamdebeasi !

shreeshbhat:
It was mostly @liamdebeasi 's changes that helped fix issues. 
Thanks for taking the time to review and make changes. 🎉

Issue #17822, opened 2019-03-19
ion-item-sliding: options of the opposite direction of sliding is shown too


SherifMoShalaby:
# Bug Report

**Ionic version:**

[x] **4.0.2**

**Current behavior:**

adding an ion-item-sliding to an ion-item with ion-item-options both sides, when sliding the item from left to right (or vise versa) to the end of the item, the right icons shown up while it should not.
so now both options from both sides will be shown in any swiping direction

**Expected behavior:**

when swiping from left to right or vise versa, only icons from the side i`m sliding should be shown

**Steps to reproduce:**

1. create new project with version 4.0.2.

2. in app.component.html add the sample example found in ionic documentation regards ion-item-sliding 

3. swipe from left to right or vise versa, issue will be generated.

**Other information:**

the issue was not presented in version 4 beta 7

**Ionic info:**

```
Ionic:

   ionic (Ionic CLI)             : 4.10.3 (C:\Users\sshalaby\AppData\Roaming\npm\node_modules\ionic)
   Ionic Framework               : @ionic/angular 4.0.2
   @angular-devkit/build-angular : 0.12.4
   @angular-devkit/schematics    : 7.3.2
   @angular/cli                  : 7.3.1
   @ionic/angular-toolkit        : 1.4.0

Cordova:

   cordova (Cordova CLI) : 8.1.2 (cordova-lib@8.1.1)
   Cordova Platforms     : not available
   Cordova Plugins       : not available

System:

   NodeJS : v10.15.2 (C:\Program Files\nodejs\node.exe)
   npm    : 6.9.0
   OS     : Windows 10
```

liamdebeasi:
Hi there,

Thanks for the issue. Is this the behavior you are describing?

![zf3gqstXst](https://user-images.githubusercontent.com/2721089/54690061-01daba00-4af7-11e9-8d4b-2875e3d783bf.gif)

SherifMoShalaby:
> Hi there,
> 
> Thanks for the issue. Is this the behavior you are describing?
> 
> ![zf3gqstXst](https://user-images.githubusercontent.com/2721089/54690061-01daba00-4af7-11e9-8d4b-2875e3d783bf.gif)

Yes.

brandyscarney:
The issue is most likely somewhere in this:

```
.item-sliding-active-slide {
  @include rtl() {
    &.item-sliding-active-options-start ion-item-options:not(.item-options-end) {
      width: 100%;

      visibility: visible;
[...]

PR #17986, opened 2019-04-06
fix(item-sliding): hide opposite side options

masimplo:
This resolves an issue where both ion-item-options are show when sliding an ion-item-sliding item. 

- adds a css rule to hide the opposite item-options when sliding item is open
- adds an item in basic tests to demonstrate the issue and fix
-

**Ionic Version**:

**Fixes**: #17822

Video demonstrating issue and fix:
[video](https://www.youtube.com/watch?v=rJYvsZGI6kw)


masimplo:
Hi, 
thanks for looking into this. You are correct, it was a bit tricky to reproduce as it does not happen in the emulator but on desktop. Nice catch!
Just pushed the proposed fix, which works like a charm.

liamdebeasi:
Hey there! Per @brandyscarney's comment https://github.com/ionic-team/ionic/issues/17822#issuecomment-479046156 I've updated the CSS a little bit to fix the issue. Looks like the `visibility: hidden` got removed during our v3->v4 transfer, so I added it back.

liamdebeasi:
Merged. Thank you! 🎉

Issue #17767, opened 2019-03-13
Make Ion-Item --border-width available for all 4 directions


Spinnenzunge:
# Feature Request
Make ion-item css variable `--border-width` available for all directions:
```
--border-width-top
--border-width-right
--border-width-bottom
--border-width-left
```

**Ionic version:**
[x] **4.x**

**Describe the Feature Request**
Border width for all four directions for easier custom styles

**Describe Preferred Solution**
```
--border-width-top
--border-width-right
--border-width-bottom
--border-width-left
```

masimplo:
I just opened a PR for this, but the fact that --border-width would no longer be supported and thus introduce a breaking change with this PR haunted me. So I tried setting a list of widths as border-width and it seems to work as demonstrated here: https://stackblitz.com/edit/typescript-6twsu6?file=style.css

@Spinnenzunge did you try setting the border-width like:

```css
--border-width: 10px 0px 20px 5px;
```

Spinnenzunge:
That works, but the four different variables makes it way easier to overwrite in subclasses etc. I could generate my own --border-width-direction variables and assign to --border-width but that would kinda be what i requested in the feature request

masimplo:
I see. However breaking the current API and having to declare for variables instead of one when wanting to set a border seems like something a lot of people would dislike. Do you have an example where setting the current variable with four parameters would not work? Is it when you have 2 sides defined and you want to define one more, without redefining previous ones?

I am trying to find a way to support both border-width and border-side-width at the same time, but my CSS variable knowledge is not that extensive. [stackoverflow question](https://stackoverflow.com/questions/55615332/css-custom-properties-variables-for-box-model)
In case you think there is no other way to achieve what you need and @brandyscarney says it is ok to break the API I will be happy to reopen the pull request.

brandyscarney:
Thanks for looking into this @masimplo! We can't introduce any breaking changes to the API without a major version release, and since we just released `v4` we're not looking to release a major version anytime soon (at the earliest should there be 6 months between any major version releases).

For the original issue - would a possible solution be to create your own variables for this? Take the following example:

```
ion-item {
  --border-top-width: 2px;
  --border-end-width: 3px;
  --border-bottom-width: 2px;
  --border-start-width: 6px;
  
  --inner-border-width: 0;
  
  --border-width: var(--border-top-width) var(--border-end-width) var(--border-bottom-width) var(--border-start-width);
  --border-color: red;
}

ion-item:first-child {
[...]

PR #17990, opened 2019-04-07
feat(item): improve border-width granularity

masimplo:
Introduce 4 different CSS variables for the ion-item border-width instead of 1.

```css
--border-width-top
--border-width-right
--border-width-bottom
--border-width-left
```

- Add the new variables to the item's css to support more granular control of the border width
- This will probably be a breaking change as border-width is no longer available. Not sure how you handle such changes to the API, but couldn't find an elegant CSS solution for supporting both.

**Ionic Version**: 4.2.0

**Fixes**: #17767

### DIFF
--- a/core/api.txt
+++ b/core/api.txt
@@ -489,7 +489,10 @@ ion-item,css-prop,--background-activated
 ion-item,css-prop,--border-color
 ion-item,css-prop,--border-radius
 ion-item,css-prop,--border-style
-ion-item,css-prop,--border-width
+ion-item,css-prop,--border-width-bottom
+ion-item,css-prop,--border-width-left
+ion-item,css-prop,--border-width-right
+ion-item,css-prop,--border-width-top
 ion-item,css-prop,--box-shadow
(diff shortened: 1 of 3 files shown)


masimplo:
Not sure if this PR should be merged as mentioned in my comment in the issue this is supposed to resolve. https://github.com/ionic-team/ionic/issues/17767#issuecomment-480643982

masimplo:
I have confirmed that setting like `--border-width: 10px 0px 20px 5px;` is equivalent to setting the distinct properties so this PR should be closed and not merged. Maybe the docs could be improved to point this out.

masimplo:
Reopening this, as there is an ongoing discussion and it might be one of the two possible solutions (otherwise it can be turned in the other solution quite easily)

brandyscarney:
I'm going to close this per my comment here: https://github.com/ionic-team/ionic/issues/17767#issuecomment-482191164

Thank you though! 🙂

Issue #18002, opened 2019-04-09
bug: Clear input on ion-input causes focus loss of field


RHinderiks:
# Bug Report

**Ionic version:**
[x] **4.2.0**

**Current behavior:**
When using clearInput on a ion-input a button appears to clear the input field. When clicking this button the focus is lost from the field. 

**Expected behavior:**
Keep focus so a user can keep on typing in the same field they were just typing in.

**Steps to reproduce:**
Make a page with a ion-input field
Add clearInput to the field
Type in field
Press on x to delete content

**Related code:**

```
<ion-input clearInput></ion-input>
```

**Ionic info:**

```
Ionic:

   ionic (Ionic CLI)             : 4.12.0 (/usr/local/lib/node_modules/ionic)
   Ionic Framework               : @ionic/angular 4.2.0
   @angular-devkit/build-angular : 0.13.8
   @angular-devkit/schematics    : 7.3.8
   @angular/cli                  : 7.3.8
   @ionic/angular-toolkit        : 1.5.0

Cordova:

   cordova (Cordova CLI) : 8.1.2 (cordova-lib@8.1.1)
   Cordova Platforms     : android 8.0.0, ios 5.0.0
   Cordova Plugins       : cordova-plugin-ionic-keyboard 2.1.3, cordova-plugin-ionic-webview 4.0.1, (and 20 other plugins)

System:

   Android SDK Tools : 26.1.1 (/Users/rh/Library/Android/sdk)
   ios-deploy        : 1.9.4
   ios-sim           : 7.0.0
   NodeJS            : v8.11.4 (/Users/rh/.nvm/versions/node/v8.11.4/bin/node)
   npm               : 6.7.0
   OS                : macOS Mojave
   Xcode             : Xcode 10.2 Build version 10E125
```

PR #18004, opened 2019-04-09
fix(input): prevent default behaviour on clear text input button

konnectappdev:
## Pull request checklist

- [X] Docs have been added / updated (for bug fixes / features)

## Pull request type
- [X] Bugfix

When clear-input is used on a input field it loses focus on that field whenever the input is cleared

Issue Number: #18002

## What is the new behavior?
A check has been added to prevent default behaviour of the clear-input button, This causes the input field to lose focus. A check is added to only apply this logic whenever the clear-input button is shown since this function is also used for clear-on-edit but would make no sense to apply in that scenario.

- [X] No


core/src/components/input/input.tsx
```diff
@@ -307,7 +307,11 @@ export class Input implements ComponentInterface {
     }
   }
 
-  private clearTextInput = () => {
+  private clearTextInput = (ev?: Event) => {
+    if (this.clearInput && !this.readonly && !this.disabled && ev) {
+      ev.preventDefault();
```
liamdebeasi:
```suggestion
      ev.preventDefault();
      ev.stopPropagation();
```

liamdebeasi:
Merged. Thank you! 🎉

PR #18188, opened 2019-05-02
feat(menu-button): add css variables for padding

### LINKED ISSUE #18187: feat: ion-menu-button should use --padding-start variable
I think the ion-menu-button should use the --padding-start variables instead of hardcoding the padding like here:

https://github.com/ionic-team/ionic/blob/master/core/src/components/menu-button/menu-button.md.scss

JustDoItSascha:
fixes #18187

### DIFF
--- a/core/api.txt
+++ b/core/api.txt
@@ -573,6 +573,10 @@ ion-menu-button,prop,color,string | undefined,undefined,false,false
 ion-menu-button,prop,menu,string | undefined,undefined,false,false
 ion-menu-button,prop,mode,"ios" | "md",undefined,false,false
 ion-menu-button,css-prop,--color
+ion-menu-button,css-prop,--padding-bottom
+ion-menu-button,css-prop,--padding-end
+ion-menu-button,css-prop,--padding-start
+ion-menu-button,css-prop,--padding-top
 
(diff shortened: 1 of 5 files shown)


brandyscarney:
Thanks! A few things:

1. Could you move the variables so they are in the main `menu-button` scss file and remove them from the individual mode files (since they are the same): https://github.com/ionic-team/ionic/blob/master/core/src/components/menu-button/menu-button.scss#L25

2. Could you make sure the defaults are defined in each of the mode files like the following for iOS:

```
:host {
  --padding-top: 0;
  --padding-end: 5px;
  --padding-bottom: 0;
  --padding-start: 5px;
}
```

and MD would use `8px` instead of `5px`. 

3. Could you make sure that the variables are documented here: https://github.com/ionic-team/ionic/blob/master/core/src/components/menu-button/menu-button.scss#L8

A good reference is back-button: https://github.com/ionic-team/ionic/blob/master/core/src/components/back-button/back-button.scss#L28-L31

After that please run `npm run build` as the documentation will be generated. 🙂 Let me know if you have any questions, thank you!

brandyscarney:
I went ahead and made the changes, thanks!

PR #18189, opened 2019-05-02
fix(menu-button): use css variables for padding

### LINKED ISSUE #18187: feat: ion-menu-button should use --padding-start variable

JustDoItSascha:
fixes #18187


brandyscarney:
Thank you for the PR! Could you combine this PR with #18188 please?

Issue #18327, opened 2019-05-21
bug: ion-card[button] white space at bottom of button cards


studioromeo:
# Bug Report

**Ionic version:**
[x] **4.4**

**Current behavior:**
When the `<ion-card button>` element has a button attribute it adds a `<button>` element around the cards contents. On Chrome (and possibly other browsers) buttons are displayed inline and this causes the `<button>` element to not fill the `<ion-card>` element as it honours the white space between the `<button>` and closing `</ion-card>` tag. This is most noticeable when we place an image as the cards contents as we'll see a blank line before the end of the card (see below for an example)

<img width="332" alt="Screenshot 2019-05-21 at 22 30 54" src="https://user-images.githubusercontent.com/846587/58132236-ade47280-7c18-11e9-8422-8d84a333edc7.png">

**Expected behavior:**
Setting the display property of the `<button>` element to `block` ensures the `<button>` fills the available space and ignores any white space between the tags.

<img width="335" alt="Screenshot 2019-05-21 at 22 31 04" src="https://user-images.githubusercontent.com/846587/58132399-1895ae00-7c19-11e9-9100-0302ed4c12a8.png">

**Steps to reproduce:**
Using the below code we should see the issue described

```html
<ion-card button>
  <img src="...">
</ion-card>
```

**Other information:**
Pull request: https://github.com/ionic-team/ionic/pull/18328

Thanks!

liamdebeasi:
Hi there,

This issue has been resolved via https://github.com/ionic-team/ionic/pull/18328 and will be in the next release of Ionic.

Thanks!

PR #18328, opened 2019-05-21
fix(card): remove white space from bottom of card

studioromeo:
## Pull request checklist

## Pull request type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

Issue Number: #18327

## What is the new behavior?
Sets the button element to be a block element so it ignores any white space between the `</button>` and `</ion-card>` tags. I've added a test example of this and from testing manually I couldn't see any regressions.

- [ ] Yes
- [x] No

## Other notes
I don't think this is related to this PR but in the spirit of being honest. I wasn't able to get e2e tests to run on my machine (even on the master branch) so I haven't been able to test this using the e2e suite. If anyone can help me here that'd be cool but otherwise I'm hoping the CI will save me. An example of one of the tests that fails is below:

```
 FAIL  src/components/select/test/basic/e2e.ts (6.289s)
  ● select: basic

    TypeError: Cannot read property 'waitForVisible' of null
```

Node: v10.15.3
npm: 6.4.1


core/src/components/card/card.scss
```diff
@@ -70,6 +70,8 @@
   @include padding(0);
   @include margin(0);
 
+  display: block;
```
liamdebeasi: The issue here stems from `img` being displayed `inline` by default. Can you apply `display: block` to the `img` element instead of `:host` here?

studioromeo: Hm, I'm not sure to be honest. I may be misreading your comment but the `` tag is already displayed as a block level element (https://github.com/ionic-team/ionic/blob/master/core/src/components/card/card.scss#L44-L48).

The issue currently is we have content like this
```html
<ion-card>
  <button>
    <img src=".." />
  </button>
</ion-card>
```

The `button` element being inline block will render the space between it's closing tag and the closing ion-card type which results in a gap between the bottom of the button and the bottom of the card.

`<ion-card>`
Screenshot 2019-05-22 at 15 15 00

`<button>`
Screenshot 2019-05-22 at 15 14 54

If we make just the wrapping button a block level element it'll ignore the white space between closing tags and always match the height of the card which is what we want I think?

liamdebeasi: Huh interesting... We had a similar issue on `ion-img`, and the solution there was to apply `display` to the `img` element. I took a look at the code and what you have is probably fine. 👍

liamdebeasi:
Merged. Thank you!

studioromeo:
Awesome! Thanks 😄

core/src/components/card/test/button/index.html
```diff
@@ -68,6 +68,10 @@
         </ion-card-content>
       </ion-card>
 
+      <ion-card button>
+        <img src="https://images.unsplash.com/photo-1483354483454-4cd359948304?dpr=1&auto=format&fit=crop&w=1000&q=80&cs=tinysrgb&ixid=dW5zcGxhc2guY29tOzs7Ozs%3D">
```
adamdbradley: We should avoid actual image requests here since end to end testing may or may not have downloaded the image in time for the screenshot, causing our tests to fail often. With many other tests we use a small datauri.

studioromeo: @adamdbradley thats fair. I was copying from the existing examples but I can update this in a future PR if you like?

liamdebeasi: @studioromeo I'm going through our tests now to update this, so I'll make sure to update the card test as well 🙂

studioromeo: Ok thanks @liamdebeasi !

PR #18374, opened 2019-05-23
test(components): update all CSS attributes to classes

brandyscarney:
## Pull request checklist

## Pull request type

 

 

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes
- [x] Other (please describe): This updates all of our tests to use the class versions of the CSS attributes

Issue Number: N/A

## What is the new behavior?

-
-
-

- [ ] Yes
- [x] No


studioromeo:
Hey @brandyscarney. Is this the recommended way to apply these styles over attributes? Thanks!

brandyscarney:
Hey @studioromeo! Yes, we are recommending using classes over attributes moving forward. I updated the [documentation](https://ionicframework.com/docs/layout/css-utilities) awhile ago but I wanted to update our codebase in case anyone was referencing it. While the attributes still exist and will continue to work the same, our goal is to eventually remove them with a major version release. As we add more CSS utilities (such as [display](https://github.com/ionic-team/ionic/blob/master/core/src/css/display.scss#L19) that was recently added) we will be adding them as classes only. The reason behind moving to classes is because in React, Stencil, or any framework using JSX and TypeScript there will be type errors thrown if you try to use a CSS attribute as it is looking for it to be a property of that element. Rather than have users use CSS attributes for certain frameworks and classes for others, we're recommending classes for consistency as they will work across frameworks without issue. Let me know if you have any questions, thanks!

studioromeo:
Thanks @brandyscarney that's good to know! 😄

Issue #18399, opened 2019-05-28
feat: permanent cancel button on ion-searchbar


aballet:
# Feature Request

**Ionic version:**

[x] **4.x**

**Describe the Feature Request**

Being able to keep the cancel button of the ion-searchbar visible, even when there's no focus on the searchbar. Comportment of the button doesn't change.

**Describe Preferred Solution**

A simple attribute would do the trick: `<ion-searchbar permanentCancelButton></ion-searchbar>`.

**Additional Context**

Use case: A user clicks on a search icon in the toolbar; the toolbar is dismissed to give space to the invoked searchbar. Now the user needs a way to go back to the previous toolbar if the searchbar is not focused.

liamdebeasi:
Hi there,

Thanks for the feature request. Can you explain your use case a bit more? I'm not sure I fully understand the benefit of having the `Cancel` button always visible even if the input is not focussed.

Thanks!

aballet:
Sure @liamdebeasi ! Have a look at the following GIF, it's showing the latest comportment of the Facebook search bar that I was trying to reproduce. Imagine this with the actual implementation in Ionic: we click on the search icon, and end up with the expanded search bar. If the user clicks on anything but the bar, it will lose its focus, hence leaving the user with no way of getting back to the previous screen with the search icon (except clicking again on the bar to let the cancel button appear, but it's really bad in terms of UX). Hence the permanent cancel button that would be needed in that case.

![O6yKjmASymi063IR5vLngBf8KLJXYGb7ZbKYvl6S](https://user-images.githubusercontent.com/17181237/58577412-b0f9e700-8246-11e9-992f-456cdd4caf8d.gif)

liamdebeasi:
Hi there,

Thanks for the follow up! The behavior in the GIF is actually just autofocusing the searchbar when it appears, which is why the cancel button is visible.

I am going to close this issue, but I think this functionality could be accomplished with the following feature request: https://github.com/ionic-team/ionic/issues/17745.

Thanks for using Ionic!

aballet:
Hi @liamdebeasi, I think you closed it too soon. Let me explain:

Agreed, it is autofocus, and that's why the cancel button is visible. However, when we loose the focus (like in this new GIF, the user simply clicks on the background), the cancel button disappears, leaving the user with the search bar, but no way to come back to the other interface except clicking on the bar again, and hitting cancel.

[...]

PR #18401, opened 2019-05-28
feat(searchbar): add permanent cancel button attribute

aballet:
## Pull request checklist

## Pull request type

 

 

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes

The cancel button on `ion-searchbar` is only displayed when `showCancelButton` is set to `true`, and the search bar is focused.
See feature request #18399.

## What is the new behavior?

An attribute `permanentCancelButton` can be added to `ion-searchbar`. This allows the cancel button to stay even when the search bar is not focused. The `showCancelButton` attribute must still be set to enable the cancel button. 
A use case is described in the feature request.

- [ ] Yes
- [x] No

## Other information

A more elegant way of doing this would be to add a unique `cancelButton` attribute than may take different values : `off` (default), `onFocus` (same as showCancelButton), `permanent`. However, this introduces a breaking change for whoever uses the cancel button or the `(ionCancel)` event.


masimplo:
> A more elegant way of doing this would be to add a unique cancelButton attribute than may take different values : off (default), onFocus (same as showCancelButton), permanent. However, this introduces a breaking change for whoever uses the cancel button or the (ionCancel) event.

How about if cancelButton attribute supports both values, so that it does not break the current api, but enhance it. The type could be boolean or one of the above values. All you would have to do then is fallback to the boolean check in your ifs.

aballet:
I understand, it's better indeed. I didn't think about it this way, thank you!
I'll work on that and commit the changes in a few hours.

aballet:
Hey @masimplo, I tried your proposal.
If we use the attribute without any value, ie. `<ion-searchbar showCancelButton></ion-searchbar>`, the value of showCancelButton is now an empty string `""` instead of `true` when union types weren't used. It's still a breaking change.

masimplo:
Hey @aballet,

thanks for taking the time to try this out. I just tried it as well and got the same behavior you are describing. It might be this is hitting a stencil limitation/bug as it should identify the existence of the attribute as a true value, even for union types. 
Haven't played much with stencil so I have no clue if this can be overcome currently (although ion-split-pane has a similar property `when': string | boolean;` but since there is no test for it, I am not sure it is even working properly).

I was pretty sure this was going to work, sorry for leading you down the wrong path with my suggestion.

liamdebeasi:
Hi there,

Thanks for opening this PR! We appreciate all the hard work you put into it.

I am going to close this PR because we think this functionality could be achieve via a different feature. Please see the issue thread here for an explanation: https://github.com/ionic-team/ionic/issues/18399

Thanks!

aballet:
Hi,
The feature is slightly different though, I added some more explanation in #18399 because it could really be useful in some cases.

liamdebeasi:
So I'm wondering if the initial behavior should be a little different. On iOS Settings the searchbar does not display the cancel button until it has been focused. Additionally, once you hit "cancel" the "cancel" button goes away but the searchbar still remains.

I think it would be good to build this feature with that native iOS functionality in mind. In doing so we _should_ be able to cover the native behavior as well as the behavior that was displayed in the Facebook example in this original issue.

I'll try to dive into the code today and see if I can get something working. Thanks!

liamdebeasi:
Hey there,

I pushed a few changes to your branch. The notable changes are:

1. Renamed `permanent-cancel-button` to `persist-cancel-button`. I felt `permanent` implied that users could never get rid of the cancel button.
2. Exposed `showCancel` and `hideCancel` methods. (should these be called `showCancelButton` and `hideCancelButton` instead?)
3. Added in logic to only show the persisted cancel button if the user has focused the searchbar at least once. Additionally, the cancel button will be dismissed if the user taps the cancel button.

I believe all of this brings the feature inline with what the behavior on iOS is (see the Settings app searchbar for an example) and also lets developers replicate the behavior Facebook has on their app.

I still need to work on adding some automated tests.

Let me know if you have any feedback!

aballet:
Hi,

This is really great, thank you!

I slightly edited the documentation to reflect the latest changes.
Also, I think that the comportment of the `persistCancelButton` attribute should be the same in the md mode... or that the doc should state clearly that it is iOS only, otherwise it is misleading.
I chose the first approach with the latest commit; please tell me what you think, I'm not used to material design.

If you'd like me to draft some tests, please let me know.

manucorporat:
Not a fan of adding two methods and one prop to support this feature request, this is all new API and code to maintain. We should be careful how this APIs will play a role in the future.

I would rather prefer to expose a `forceShowCancel` than two methods.

liamdebeasi:
Thanks for the feedback @manucorporat! I agree, but in this case I think we are going to need these methods. In order to get the desired behavior devs are going to need to be able to programmatically show/hide the cancel button. (See the search in the Twitter or Settings apps on iOS).

core/src/components/searchbar/searchbar.tsx
```diff
@@ -101,9 +115,14 @@ export class Searchbar implements ComponentInterface {
   @Prop() searchIcon = 'search';
 
   /**
-   * If `true`, show the cancel button.
+   * Sets the behavior for the cancel button.
+   * Setting to `true` shows the cancel button on focus.
+   * Setting to `false` disables the cancel button.
+   * Setting to `persist` shows the cancel button on focus
```
brandyscarney:
```suggestion
   * Setting to `"persist"` shows the cancel button on focus
```

core/src/components/searchbar/searchbar.tsx
```diff
@@ -101,9 +115,14 @@ export class Searchbar implements ComponentInterface {
...
+   * but does not hide it until the user clicks it
+   * or calls the `hideSearch` method.
```
brandyscarney:
```suggestion
   * or calls the `hideCancel` method.
```

core/src/components/searchbar/searchbar.tsx
liamdebeasi: GitHub won't let me directly commit these suggestions for some reason, so I've made them manually and pushed an update.

liamdebeasi:
Hi there,

I discussed a few changes with @brandyscarney and @manucorporat. We would like to do the following:

1. Do not expose `showCancel` or `hideCancel` methods.
2. Do not expose a `persist-cancel-button-prop`.
3. Extend the `show-cancel-button` prop to accept three new values:
  a. `always`: Always shows the cancel button regardless of focus state.
  b. `never`: Never show the cancel button (equivalent to `false`).
  c. `focus`: Only show the cancel button when the searchbar is focused  (equivalent to `true`).
4. Deprecate the `true` and `false` values.

I decided to make this as a separate PR since these changes are very different from the original feature proposal. I will be sure to add you as a co-author 🙂. You can follow the new PR here: https://github.com/ionic-team/ionic/pull/18507

Thanks!

liamdebeasi:
Thanks for the PR! We appreciate all of the hard work you put into this. As per my previous comment, we are going to close this PR in favor of https://github.com/ionic-team/ionic/pull/18507. You will be added as a co-author to that commit.

Thanks!

PR #18468, opened 2019-06-05
fix(angular): pass navigation options to router when using NavController

MarkChrisLevy:
- [X] Bugfix
- [ ] Feature
- [ ] Build related changes

When navigateForward or navigateBack is called with url as string or UrlTree, navigation options are not passed to angular router. When url is an array, options are passed properly.

Issue Number: #18469

## What is the new behavior?
Navigation options must be always passed to angular router.

- [ ] Yes
- [X] No


masimplo:
This is an angular issue. Cannot find the particular issue number right now, but I came upon it a couple weeks back.

liamdebeasi:
Thanks for the PR! Yep, that issue is related @masimplo. `navigateByUrl` will ignore most of the `NavigationExtras` that modify the url (which is why there is that extra case for handling `fragment` and `queryParams` when creating the UrlTree).

angular/src/providers/nav-controller.ts
```diff
@@ -202,7 +202,7 @@ export class NavController {
         urlTree.fragment = options.fragment;
       }
 
-      return this.router!.navigateByUrl(urlTree);
+      return this.router!.navigateByUrl(urlTree, options);
```
liamdebeasi:
```suggestion
      /**
       * `navigateByUrl` will still apply `NavigationExtras` properties
       * that do not modify the url, such as `replaceUrl` which is why 
       * `options` is passed in here.
       */
      return this.router!.navigateByUrl(urlTree, options);
```

liamdebeasi: Since this caveat about navigateByUrl is a little confusing/hard to remember, let's add an extra comment that just makes it very clear what we are doing and why.

MarkChrisLevy: @liamdebeasi Great, thanks!

liamdebeasi: 👍 Also you'll need to commit this yourself since I do not have permission to write to your branch.

liamdebeasi:
Hey there,

You will need to run `npm run lint` to get these tests to pass.

liamdebeasi:
Thank you! 🎉

PR #18478, opened 2019-06-06
feat(app): Keyboard open and close events

### LINKED ISSUE #18467: feat: Fire keyboard open/close events via Visual Viewport API
# Feature Request

**Ionic version:**

[x] **4.x**

Ionic should provide events that devs can subscribe to that allow them to determine when the on-screen keyboard has opened/closed.

This is similar to the events provided in the native Keyboard plugin, but by using Web APIs we would be able to provide a similar functionality to web apps and PWAs.

Cordova/Capacitor users would be able to use this feature in order to reduce their dependency on native code.

A good use case to consider is a chat application. The chat message bar is typically at the bottom of the screen, and tapping it would open the keyboard and cause the entire view to shift up. Ionic provides a way of disabling this, but the input would still be below the keyboard. Having an event like `ionKeyboardOpen` would allow developers to translate the chat bar up so that it sits just above the keyboard.

Using the [Visual Viewport API](https://developer.mozilla.org/en-US/docs/Web/API/VisualViewport), detect when the on-screen keyboard appears and closes.

Requirements:
1. The `ionKeyboardOpen` event should fire when the on-screen keyboard is presented. The event should contain a property with the keyboard height.
2. The `ionKeyboardClosed` event should fire when the on-screen keyboard is dismissed.
3. Feature detection should be used so browsers that do not have this API will not encounter errors. At the time of writing, the main targets are iOS 13+ and Chrome WebView 62+ (so ideally this would be Android 5.0+ assuming devices running these have auto-updated WebViews).

**Describe Alternatives**

iOS <= 12 and Chrome WebView <= 61 will need to keep using the native Keyboard plugin as they do not fully support the Visual Viewport API.

**Related Code**

Should just be a matter of doing something similar to:

```
window.visualViewport.addEventListener('resize', (e) => {
  // check to see if keyboard opened
  // if so, calculate keyboard height and fire ionKeyboardOpen
  // check to see if keyboard closed
  // if so, fire ionKeyboardClose
  // otherwise, do nothing
});
```

**Additional Context**

iOS 12 supports the Visual Viewport API when enabling it via Experimental Features; however, the `resize` event does not fire when the keyboard is opened/closed. Additionally, Chrome 61 has the Visual Viewport API enabled, but it does not have a `resize` event.

liamdebeasi:
blocks https://github.com/ionic-team/ionic-docs/pull/1387
blocks https://github.com/ionic-team/ionic-react-hooks/pull/9

## Pull request checklist

## Pull request type

 

 

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes

Issue Number: resolves #18467

## What is the new behavior?

- Devs can now subscribe to `ionKeyboardDidOpen` and `ionKeyboardDidClose` events to adjust inputs/content relative to the on-screen keyboard.

Note: Chrome 62+ and iOS 13+ only

- [ ] Yes
- [x] No


core/src/components/app/app.tsx
```diff
@@ -28,6 +28,8 @@ export class App implements ComponentInterface {
         if (config.getBoolean('hardwareBackButton', isHybrid)) {
           import('../../utils/hardware-back-button').then(module => module.startHardwareBackButton());
         }
+
+        import('../../utils/keyboard').then(module => module.startKeyboardAssist(window));
```
adamdbradley, line 32: Can you add a typeof window !== 'undefined' check here?

PR #18493, opened 2019-06-09
fix(ion-tabs): preserve route navigation extras when changing tabs

jmannau:
## Pull request checklist

## Pull request type

 

 

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

When using ion-tabs in an angular app, changing tabs doesn't restore the previous route navigation extras, such as Query Params or URL Fragments.

I haven't checked this behaviour in the react or vue distributions.

Issue Number: https://github.com/ionic-team/ionic/issues/18717

## What is the new behavior?

This is a bug fix so that when changing tabs, any previous route navigation extras are restored.

Please note, this PR is only for @ionic/angular and not React or Vue projects.

- [ ] Yes
- [x] No


jmannau:
Hi Ionic Team, you're doing an amazing job and I know are absolutely flat out. I'm just wondering if I can do anything else to help review this pull request so it can be approved and merged? Thanks a million!

liamdebeasi:
Thanks for the PR! We appreciate all the work you put into this.

I think my main concern is changing what `getLastUrl` returns. This PR changes the return from a `string` to a `RouteView` which might be considered a breaking change for users that directly access `IonRouterOutlet` in their apps.

I am in favor of fixing this issue, but it might have to wait until the next major release of Ionic as we only do breaking changes in major releases. I will post in this thread when I have more to share.

Thanks!

jmannau:
Thanks for the feedback!! Would it help if I updated the PR to include a new api that returns the RouteView, leaving getLastUrl to return a string? This would not introduce a breaking change and fix the bug at the same time. 

I’m happy to do this if you think it would work. 

Cheers,

James

yannbf:
> Thanks for the PR! We appreciate all the work you put into this.
> 
> I think my main concern is changing what `getLastUrl` returns. This PR changes the return from a `string` to a `RouteView` which might be considered a breaking change for users that directly access `IonRouterOutlet` in their apps.
> 
> I am in favor of fixing this issue, but it might have to wait until the next major release of Ionic as we only do breaking changes in major releases. I will post in this thread when I have more to share.
> 
> Thanks!

@liamdebeasi what about #18740, which fixes the same problem however does not introduce a breaking change?

liamdebeasi:
Sorry for the delay in replying!

What I'm thinking is we can add this change, but add it as a new method and update internal references to use this method. We would still have the old method, but we can mark it as deprecated. This will let us avoid a breaking change while still fixing the issue.

jmannau:
Hi @liamdebeasi I made time to look at this again and have restored `getLastUrl` to  return a string for backward compatibility and added a new method `getLastRouteView` that returns the `RouteView`. This should remove any breaking changes. The `ion-tab` directive now uses `getLastRouteView` internally.

liamdebeasi:
Thanks for making the changes! I found a small bug in the PR. Here are steps to reproduce:

1. Load Angular test app in `src/angular/test/test-app` in your browser.
2. Click "Tabs test".
3. Click "Go to Page 2 with Query Params".
4. Click "Tab Two". Notice that the query params/fragments disappear. (This is correct)
5. Click "Tab One". Notice that the query params/fragments re-appear. (This is correct)
6. Click the back button in the top left. Notice that the query params/fragments disappear. (This is correct)
7. Click "Go to Page 2 with Query Params".
8. Click "Tab Two".
9. Click "Tab One".
10. Click "Tab One" again. Notice that you go back a page as in step 6, but the query params/fragments do not disappear. (This part appears to be incorrect).

Other than that, this PR appears to be working properly.

jmannau:
Hi @liamdebeasi I spent some today reviewing this carefully. I have updated the PR to take into account going to the root tab view and respecting the url extras. I have added documentation https://github.com/ionic-team/ionic/pull/18493/files#diff-891ce68da9d3204cb1a1870a225e5350R69 to describe different scenarios wrt tab navigation.

angular/src/directives/navigation/ion-router-outlet.ts
```diff
@@ -241,6 +241,20 @@ export class IonRouterOutlet implements OnDestroy, OnInit {
     return active ? active.url : undefined;
   }
 
+  /**
+   * Returns the RouteView of the active page of each stack.
+   */
+  getLastRouteView(stackId?: string): RouteView | undefined {
```
mhartington: Is this meant to be part of the public API? If not, can we tag it with `@private` so it doesn't come up in completion?

angular/src/directives/navigation/ion-router-outlet.ts
```diff
@@ -241,6 +241,20 @@ export class IonRouterOutlet implements OnDestroy, OnInit {
...
+    return this.stackCtrl.getLastUrl(stackId);
+  }
+
+  /**
+   * Returns the root view in the tab stack.
+   */
+  getRootView(stackId?: string): RouteView | undefined {
```
mhartington: Mark as private

angular/src/directives/navigation/stack-controller.ts
```diff
@@ -191,6 +191,11 @@ export class StackController {
     return views.length > 0 ? views[views.length - 1] : undefined;
   }
 
+  getRootUrl(stackId?: string) {
```
mhartington: Mark as private

angular/src/directives/navigation/ion-router-outlet.ts
jmannau: I have marked this as `@internal` similar to other framework methods. I can't mark it as `private` as it is used by external classes [ion-tabs.ts](https://github.com/ionic-team/ionic/pull/18493/files#diff-891ce68da9d3204cb1a1870a225e5350R102)

angular/src/directives/navigation/ion-router-outlet.ts
jmannau: Similarly, i have added `@internal` as it used by [ion-tabs.ts](https://github.com/ionic-team/ionic/pull/18493/files#diff-891ce68da9d3204cb1a1870a225e5350R94)

angular/src/directives/navigation/stack-controller.ts
jmannau: marked as `@internal`

angular/src/directives/navigation/ion-router-outlet.ts
mhartington: sounds good! Thanks for making the change

liamdebeasi:
PR has been merged. Thank you!

JeremyLopez:
> PR has been merged. Thank you!

Is there a version of `@ionic/angular` that contains this fix?  I'm running `4.11.10` right now and selecting a new tab does not show this behavior.  Or perhaps I'm missing something...

jmannau:
The fix was only released with v5.0. Sorry.

Issue #18555, opened 2019-06-17
bug: no styling on disabled "icon-only ion-button" in ion-toolbar


jeroenklein:
# Bug Report

**Ionic version:**
[x] **4.5.0**

**Current behavior:**
If the ``disabled`` attribute on the ``icon-only ion-button`` is set programatically or via HTML, there is no styling to reflect this. The button cannot be clicked, but there is no indicator to show the disabled status.

**Expected behavior:**

On ``V4.4.2`` everything is working fine. See https://codepen.io/jeroenklein/pen/BgoOqY

**Steps to reproduce:**

https://codepen.io/jeroenklein/pen/GzpRrQ

**Related code:**

https://codepen.io/jeroenklein/pen/GzpRrQ

**Other information:**

Related issues:
https://github.com/ionic-team/ionic/issues/18542

KillerCodeMonkey:
should be fixed with: #18560

PR #18509, opened 2019-06-11
feat(button): implement opacity for disabled button

Ivnosing:
## Pull request checklist

Please check if your PR fulfills the following requirements:
- [x] Tests for the changes have been added (for bug fixes / features)
- [x] Docs have been reviewed and added / updated if needed (for bug fixes / features)
- [x] Build (`npm run build`) was run locally and any changes were pushed
- [x] Lint (`npm run lint`) has passed locally and any fixes were made for failures

## Pull request type

 

 

Please check the type of change your PR introduces:
- [ ] Bugfix
- [x] Feature
- [ ] Code style update (formatting, renaming)
- [ ] Refactoring (no functional changes, no api changes)
- [ ] Build related changes
- [ ] Documentation content changes
- [ ] Other (please describe): 

## What is the current behavior?

Opacity could not be set for disabled buttons.

Issue Number: fixes #16965 

## What is the new behavior?

Opacity of both default and disabled buttons can be overwritten in the host component with the CSS variable `--opacity`.

```scss
:host(.button-disabled) {
  --opacity: .5;
  pointer-events: none;
}
```

## Does this introduce a breaking change?

- [ ] Yes
- [x] No

## Other information

```html
<ion-button style="--opacity: 1" disabled>Disabled opacity: 1</ion-button>
```

PR #18560, opened 2019-06-17
fix(Button): default opacity for disabled clear buttons

KillerCodeMonkey:
Clear buttons to not get disabled state, when attribute disabled is set.

```HTML
<ion-header>
  <ion-toolbar>
    <ion-title>
      Ionic Blank
    </ion-title>
  </ion-toolbar>
</ion-header>

<ion-content>
  <div class="ion-padding">
    The world is your oyster.
    <p>If you get lost, the <a target="_blank" rel="noopener" href="https://ionicframework.com/docs/">docs</a> will be your guide.</p>
  </div>
</ion-content>

<ion-footer>
  <ion-toolbar>
    <ion-buttons>
      <ion-button fill="clear" color="primary" disabled>clear disabled</ion-button>
      <ion-button fill="clear" color="primary">clear enabled</ion-button>
      <ion-button fill="solid" color="primary" disabled>solid disabled</ion-button>
      <ion-button fill="solid" color="primary">solid enabled</ion-button>
      <ion-button fill="outline" color="primary" disabled>outline disabled</ion-button>
      <ion-button fill="outline" color="primary">outline enabled</ion-button>
    </ion-buttons>
  </ion-toolbar>
</ion-footer>
```

Bildschirmfoto 2019-06-17 um 13 36 50

as you can see the clear buttons do not have the default disabled opacity set ;).

In ionic v4.4:

Bildschirmfoto 2019-06-17 um 13 46 04

https://github.com/ionic-team/ionic/pull/18509#issuecomment-502646901

## Pull request type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

Since 4.5.0 the `opacity: .5;` does not work for clear buttons.

Issue Number:
introduced with PR #18509
#18555

## What is the new behavior?
opacity is set to the default `opacity: .5;` ion-buttons with the `fill="clear"` and disabled attributes.

- [ ] Yes
- [x] No

## Other information
- works in v4.4


brandyscarney:
Thank you for the PR! Unfortunately this fix would not allow for users to override the disabled opacity because it would not be inherited in the `.button-native` from the host element. I believe the fix for this issue is to remove the following line here: https://github.com/ionic-team/ionic/blob/master/core/src/components/button/button.md.scss#L63

KillerCodeMonkey:
ah okay. i only checked the changes, which introduced the problem. I am on it

KillerCodeMonkey:
@brandyscarney i think the `$button-md-clear-opacity` scss var is obsolete then?

it is only used in the md variables.scss where it is defined

KillerCodeMonkey:
@brandyscarney is there a trick to get `yarn start` working? After merging the master it fails with

`
Error: invalid outputTarget type "experimental-dist-module". Valid target types: angular, dist, docs,
           docs-json, docs-custom, stats, www at config.outputTargets.forEach.outputTarget
`

brandyscarney:
@KillerCodeMonkey I am not sure, I only ever use `npm start`. Yes it looks like the scss var is obsolete.

brandyscarney:
@KillerCodeMonkey Oh you may need to run a fresh `npm i` if you haven't lately. We just merged some big changes in with Stencil yesterday.

KillerCodeMonkey:
But i Leave IT there If Somebody needs the Default value from the Framework
(which could possibly Change)

There are other unused vars in the buttons vars scss.

PS: i Made a freshinstall an installed sass AS global dep.

Brandy Carney <notifications@github.com> schrieb am Fr., 21. Juni 2019,
17:11:

simonhaenisch:
Will this be merged or fixed soon? I thought you have screenshot diff tests to catch these kind of regressions?

BTW a workaround is to use `fill="none"` which is not valid and breaks the colors but other than that looks the same as `"clear"` (https://codepen.io/simonhaenisch/pen/KjWejq?editors=1000).

KillerCodeMonkey:
@brandyscarney it should be fixed. i tested it and everything runs with success :).

for removing obsolete sass vars i would create a nother PR if your team would appreciate it.

brandyscarney:
> I thought you have screenshot diff tests to catch these kind of regressions?

We do, but the disabled button in the clear test is towards the bottom of the page and the way our screenshot tool is set up it is being cut off by the browser height. We need to move the button to a new test where it will be in the visible screenshot or increase the browser height, but then we run the risk of this happening again: https://screenshot.ionicframework.com/data/tests/34dfc3c/src/components/button/test/clear/

simonhaenisch:
I see. Couldn't you utilize sth like [`scrollIntoView`](https://developer.mozilla.org/en-US/docs/Web/API/Element/scrollIntoView) (or `window.scrollTo(0, document.body.scrollHeight)`) to get a different scroll position for a second screenshot and then do two screenshot diffs?

Anyway, not really related to this PR (: This will be fixed with the 4.6 release, right?

PR #18577, opened 2019-06-19
chore: upgrade to angular 8

mhartington:


angular/package.json
```diff
@@ -49,41 +49,38 @@
...
-    "rxjs": ">=6.2.0",
-    "zone.js": "^0.8.26"
+    "@angular/common": "~8.0.0",
+    "@angular/compiler": "~8.0.0",
+    "@angular/core": "~8.0.0",
+    "@angular/forms": "~8.0.0",
+    "@angular/platform-browser": "~8.0.0",
+    "@angular/platform-browser-dynamic": "~8.0.0",
+    "@angular/router": "~8.0.0",
+    "rxjs": "~6.4.0",
```
paulstelzer: I would suggest not to be so strict. I think    `"rxjs": ">=6.4.0" `would be better (ionic-native expects 6.5.0 or newer).

Same for Angular. I think `"@angular/common": ">=7.2.1"`, would be better if Angular 7 is still supported, otherwise `"@angular/common": ">=8.0.0"`

danielsogl: Good Point. Angular 8 is backwards compatible to angular 7 and should be allowed.

mhartington: For rxjs, `~6.4.0` will still allow for 6.5. Please see the semver check below.

https://jubianchi.github.io/semver-check/#/~6.4.0/6.5.0

For angular, it is not backwards compatible. There are breaking changes between 7 and 8, which is why the versions are set the way they are.

paulstelzer: @mhartington no, 6.5.0 does not satisfy constraint ~6.4.0 - you have to choose ^6.4.0 - then it will satisfy 6.5.0

PR #18620, opened 2019-06-25
feat(router-link): add router-link and deprecate anchor

brandyscarney:
## Pull request checklist

## Pull request type

 

 

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes

Issue Number: N/A

## What is the new behavior?

-
-
-

- [ ] Yes
- [x] No


brandyscarney:
> Add `ion-router-link` to excludeComponents in stencil.config.ts for the angular bindings generation.
> 
> Move ion-anchor and ion-router-link to the routing bundle in stencil.config.ts!

Done!

simonhaenisch:
Has `ion-anchor` been deprecated or different behavior? I get an error with using `target` on `ion-anchor` now:

> Object literal may only specify known properties, and 'target' does not exist in type 'IonAnchor'.

brandyscarney:
It was renamed from `ion-anchor` to `ion-router-link` to make it better known that. `ion-router-link` is a new component that didn't exist prior to `4.6.0`. I added `download` and `target` to `ion-anchor` in the following PR: https://github.com/ionic-team/ionic/pull/17883 

I removed `download` and `target` from it in this PR after discussing with @manucorporat, prior to any release. I'm not sure how you were using it before? It was not in `4.5.0` or `4.6.0`. If you need `target` with it, could you please tell us more about your use case? Thanks!

simonhaenisch:
@brandyscarney thanks for explaining (:

Before, we were using it like

```tsx
<ion-anchor href="/some/page" target="_blank">Some page</ion-anchor>
```

(which worked in 4.5.0)

I'm generally against opening links in new tabs by default, and I know about tabnabbing (which is not an issue in our case because the `href` is safe), but it's a requirement for one of our components to open links in a new tab by default. 🤷🏻‍♂️

Just to be clear, we're not opening a third-party url, just a route of our app (like a link to a user profile), which is why we were using `ion-anchor`, or rather `ion-router-link` now.

brandyscarney:
@simonhaenisch It's interesting that that broke. Are you using Stencil or is it just Ionic Core?

I believe the idea is that the `ion-router-link` would only be used for navigating with the Ionic Router and being able to change the `router-direction` which would not be needed if opening in a new tab or not navigating to a URL within the app, but @manucorporat would know more on this. I'm up for learning more and discussing the use cases of this though as I want to make sure we are covering all use cases for it and it isn't confusing.

simonhaenisch:
Using it in a Stencil + Ionic PWA (`@stencil/core@1.1.2` and `@ionic/core@4.6.0`). 

One thing that the router also takes care of (afaik?) is to deal with the `/#` in the url depending on `useHash`. By using `ion-anchor`/`ion-router-link`s instead of normal `a`s it's easy to change that setting without breaking anything. That's why I believe that at least the `target` prop should be implemented (`download` is a different story). I can make a PR for that if you want.

manucorporat:
@simonhaenisch I might be missing something, but i don't see how target might be useful at all. Ion-router does not understand target at all. ion-router-link is just a declarative API to call:
```
router.push(url)
```
instead of performing an actual page reload.

target="_black" means a new tab (completely new page and context) will load. I have reviewed some of the equivalent APIs in React Router, Angular and Stencil and I don't see "target" either.

Why you can't just use `<a target="">`? I just don't understand the use case

simonhaenisch:
Just for context, the use case is a table component where each row links to a document, but at the end of each row there's an icon that will link to the same document but open it in a new tab.



I would not use `target="_blank"` but I don't have the final say over it and apparently a lot of our users don't know how else to open links in new tabs 🤷🏻‍♂️ (Just in a recent user testing session, two out of two users didn't get that they had to fill in their email and password in the login form, because there's placeholders `you@domain.com` and `***` on the inputs.)

---

Anyway, I guess it's just a matter of consistency. In general, I would not use an `a` for any app routing (as it would cause a reload) and always use `ion-router-link` instead. However in this one case, where I also want to route to a certain part of the app, but launch it in a new tab, it's kind of inconvenient having to switch to an `a` tag.

```tsx
interface Td {
  content: string | VNode;
  href?: string;
  target?: string; // can be set but doesn't have to be
}
```

```tsx
tds.map(({ content, ...linkProps}) => (
  <td>
    <ion-router-link {...linkProps}>{content}</ion-router-link>
  </td>
));
```

This template would become unnecessarily complex

```jsx
tds.map(({ content, href, target}) => (
  <td>
    {target ? (
      <a href={href} target={target}>{content}</a>
    ) : (
      <ion-router-link href={href}>{content}</ion-router-link>
    )}
  </td>
));
```

---

Looking at React Router, they have actually implemented it. See [this comment](https://github.com/ReactTraining/react-router/issues/2188#issuecomment-146366561). The way they do it is: if `target` is set, just let the browser handle the link (rather than using `router.push` in our case). I think that's a rather good approach.

brandyscarney:
Thanks for the thorough response @simonhaenisch! We are going to add the `target` to the `ion-router-link`. One question though: have you been using the `@ionic/core@one` builds to develop your app? If so - is it possible that you added in the target to the `ion-anchor` after this PR (https://github.com/ionic-team/ionic/pull/17883) went in and then it broke from being removed? Or are you sure it worked in `4.5.0`? I tried it out in a Stencil project with `4.5.0` and saw the same error:

Screen Shot 2019-06-28 at 2 02 38 PM

simonhaenisch:
@brandyscarney ah yes, I was using the `@one` tag, which indeed was a dev build. Sry, false alarm about it being broke then 🙈

Thanks!

Issue #18585, opened 2019-06-20
bug: Cannot change AM/PM input in ion-datetime


godspeed529:
# Bug Report

**Ionic version:**
[x] **4.x**

**Current behavior:**
Cannot change AM/PM input in ion-datetime element.

**Expected behavior:**
Should be able to change AM/PM input in ion-datetime element.

**Steps to reproduce:**

1. cd into cloned github project directory
2. run ng serve 
3. open website in browser
4. click rendered ion-datetime element
5. try to change am/pm input 
6. see "Error parsing date: "[object Object]". Please provide a valid ISO 8601 datetime format: https://www.w3.org/TR/NOTE-datetime" in console

**Related code:**

https://github.com/godspeed529/datetime-error-parsing-date

**Other information:**
This regression was introduced in v4.5.0

**Ionic info:**

```
Ionic:

   ionic (Ionic CLI)             : 4.12.0 (/usr/local/lib/node_modules/ionic)
   Ionic Framework               : @ionic/angular 4.5.0
   @angular-devkit/build-angular : 0.13.9
   @angular-devkit/schematics    : 7.3.9
   @angular/cli                  : 7.3.9
   @ionic/angular-toolkit        : 1.5.1

System:

   NodeJS : v10.14.1 (/usr/local/bin/node)
   npm    : 6.9.0
   OS     : macOS Mojave
```

adamduren:
This regression makes uses of ionic with `ion-datetime` unusable. We had to issue an update reverted to `@ionic/angular@4.4.2`. Will this be hotfixed outside of the normal 2 week release cycle?

godspeed529:
@brandyscarney can this issue get some love also.  It is blocking our 4.6.x upgrade.

liamdebeasi:
Thanks for the issue! This isn't a regression, it's actually a bug that was uncovered when we fixed another bug oddly enough 🙂. 

We are looking into a fix, and I will update this thread when I have more to share.

Thanks!

adamduren:
@liamdebeasi I'm not sure I understand. I don't have this issue on 4.4.2 so in that case it appeared to be working as intended. If there was another bug that was fixed but as a result this happened then I would call that a regression.

liamdebeasi:
In an older version of Ionic (probably around 4.4.2) we fixed a bug where changing the selected month/day/year etc did not refresh the picker values according to the `max` and `min` constraints. 

[...]

PR #18684, opened 2019-07-01
fix(datetime): Cannot change AM/PM in ion-datetime

topalavlad:
## Pull request checklist

## Pull request type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

Users cannot change AM/PM in `ion-datetime`

Issue Number: #18585

## What is the new behavior?
Users can manually change AM/PM in `ion-datetime`

- [ ] Yes
- [x] No


core/src/components/datetime/datetime-util.ts
```diff
@@ -378,7 +388,7 @@ export function getValueFromFormat(date: DatetimeData, format: string) {
     return (date.hour! < 12 ? 'am' : 'pm');
   }
   if (format === FORMAT_hh || format === FORMAT_h) {
-    return (date.hour! > 12 ? date.hour! - 12 : date.hour);
+    return (date.hour! > 12 ? date.hour! - 12 : (date.hour === 0 ? 12 : date.hour));
```
topalavlad: Without setting the hour to 12 when the user selects 12AM~00:mm, the hour scrolls to the first value in the list because it cannot find 0.

liamdebeasi:
This looks great! There are still a handful of bugs surrounding max/min but it looks like they exist in `master` too, so I'm in favor of merging this in and then addressing those other issues separately.

core/src/components/datetime/datetime-util.ts
```diff
@@ -324,6 +324,16 @@ export function updateDate(existingData: DatetimeData, newData: any): boolean {
         (existingData as any)[key] = newData[key].value;
       }
       return true;
+    } else if (newData.ampm) {
+      newData.hour = {
```
liamdebeasi: Are we able to do `newData.hour.value` here to get rid of an extra layer?

core/src/components/datetime/datetime-util.ts
```diff
@@ -324,6 +324,16 @@ export function updateDate(existingData: DatetimeData, newData: any): boolean {
...
+        value: newData.hour
+          ? newData.hour.value
+          : (newData.ampm.value === 'pm'
+            ? (existingData.hour! < 12 ? existingData.hour! + 12 : existingData.hour!)
```
liamdebeasi: Can we add some comments here to explain what's going on for future devs?

core/src/components/datetime/datetime-util.ts
topalavlad: I tried but couldn't see a nice way to do it because when users change AM/PM `newData.hour` is undefined. I'll take another look.

liamdebeasi: Ahhh good point ok never mind then 👍

topalavlad: Actually, I replaced one of the ternary ifs and I think it's more readable now.

liamdebeasi:
Actually looks like this is a separate issue unrelated to your changes 🤔

topalavlad:
@liamdebeasi yes, it's a different thing. I actually left a comment with that issue on the fixed issue. Unfortunately it isn't trivial to fix since the AM/PM state is not saved on the DateTimeData and you cannot tell without it if the user selected 12 or 00, 5 or 17 and so on.

I'll open a separate issue for it.

adamduren:
@liamdebeasi will this be included in the next release?

pawan-logiciel:
@liamdebeasi please confirm when this will be updated ?

core/src/components/datetime/datetime-util.ts
```diff
@@ -324,6 +324,21 @@ export const updateDate = (existingData: DatetimeData, newData: any): boolean =>
         (existingData as any)[key] = newData[key].value;
       }
       return true;
+    } else if (newData.ampm) {
+      if (!newData.hour) {
+        // Even though in the picker column hour values are between 1 and 12, the hour value is actually normalized
+        // to [0, 23] interval. Because of this when changing between AM and PM we have to update the hour so it points
+        // to the correct HH hour
+        newData.hour = {
+          value: newData.hour
```
liamdebeasi: If `newData.hour` is guaranteed to be undefined/null, why do we need this additional check?

topalavlad: You're right, it's unnecessary. I'll remove it.

liamdebeasi:
Thank you!

PR #18708, opened 2019-07-03
fix(img): don't render img without src

RafaelKr:
## Pull request checklist

## Pull request type

 

 

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

When the ion-img component is added to the DOM, the img is rendered with an `undefined` src.
There is a short delay between the initial render and the IntersectionObserver/setTimeout kicking in. While this delay an ugly "broken image" is shown:

image

## What is the new behavior?

The img-tag isn't rendered until the src is set, which is fixed with this PR.

- [x] Yes
- [ ] No

## Other information

I checked if this fix is working by editing the corresponding file in my node_modules folder. I did this PR directly on github, this is why I didn't check any of the requirements-checkboxes.

@brandyscarney any chance this will be included in the 4.6.1 release?


RafaelKr:
I just found that in another file you're using `return undefined;`. https://github.com/ionic-team/ionic/blob/24840d4d9945507b92835a7c866910bd71114e01/core/src/components/virtual-scroll/virtual-scroll.tsx#L425
Would this be a better? And why?

brandyscarney:
Thanks for the PR! Is this a regression in `4.6.0`? Is there an issue for it?

RafaelKr:
No and no

brandyscarney:
We're trying to focus only on getting regression fixes in for 4.6.1. If you could provide a way to reproduce the bug here or an issue with more details we could look into it for 4.6.2. 🙂

brandyscarney:
@RafaelKr Can you please provide a way for me to reproduce this? I don't see `ion-img` rendering without a src here: https://codepen.io/brandyscarney/pen/KOMdYo

RafaelKr:
@brandyscarney Okay, I just had some time to create a reproduction.
And I found out, that it isn't related to `ion-img`, because `img` behaves the same.

The problem occurs, if you set `null` as src.
In my case I used an Angular safe navigation operator (`item.img?.src`), this automatically returns `null`, while `img` is not set.
Minimal: reproduction: https://github.com/RafaelKr/ionic-ion-img-bug

In my specific case, I created a `loading-img` component, which uses ion-skeleton-text, while the image is loading. This can be found in branch `feat/broken-loading-img`: https://github.com/RafaelKr/ionic-ion-img-bug/tree/feat/broken-loading-img

I could fix this easily by rendering the `ion-img` only, when a src is set. See branch `feat/fixed-loading-img`: https://github.com/RafaelKr/ionic-ion-img-bug/tree/feat/fixed-loading-img

Conclusion: I think this PR should be closed, to keep the same behavior for `ion-img` and the simple `img` element.

Btw: Do you think it would be nice to have this `loading-img` component as part of Ionic? In my main project I also have an `@Input() lazy = false` to toggle between `ion-img` and `img`, which I removed for simplicity here.

brandyscarney:
Thanks for getting back to me @RafaelKr! We try to keep our components mostly independent of each other, and leave the logic for how they interact up to the individual developer. I think this is a case where there are many different implementations possible. For example, here is a JavaScript implementation: https://codepen.io/brandyscarney/pen/dxXxGP

Issue #16002, opened 2018-10-18
MenuControl swipeEnable problem


csotomon:
# Bug Report

**Ionic Info**
Run `ionic info` from a terminal/cmd prompt and paste the output below.

```
Ionic:

   ionic (Ionic CLI)             : 4.2.1 (/Users/camilo/.nvm/versions/node/v8.11.1/lib/node_modules/ionic)
   Ionic Framework               : @ionic/angular 4.0.0-beta.13
   @angular-devkit/build-angular : 0.8.5
   @angular-devkit/schematics    : 0.8.5
   @angular/cli                  : 6.2.5
   @ionic/angular-toolkit        : 1.0.0

Cordova:

   cordova (Cordova CLI) : 8.0.0
   Cordova Platforms     : none
   Cordova Plugins       : no whitelisted plugins (0 plugins total)

System:

   Android SDK Tools : 26.1.1 (/Users/camilo/Library/Android/sdk/)
   ios-deploy        : 2.0.0
   ios-sim           : 7.0.0
   NodeJS            : v8.11.1 (/Users/camilo/.nvm/versions/node/v8.11.1/bin/node)
   npm               : 6.4.1
   OS                : macOS High Sierra
   Xcode             : Xcode 10.0 Build version 10A255
```

**Describe the Bug**
In a page, i want to disable the menu swipe action, calling this.menuController.swipeEnable(false); buy i get the following error:  

```
core.js:1673 ERROR Error: Uncaught (in promise): TypeError: Cannot read property 'apply' of undefined
TypeError: Cannot read property 'apply' of undefined
    at util.js:20
    at ZoneDelegate.push../node_modules/zone.js/dist/zone.js.ZoneDelegate.invoke (zone.js:388)
    at Object.onInvoke (core.js:3820)
    at ZoneDelegate.push../node_modules/zone.js/dist/zone.js.ZoneDelegate.invoke (zone.js:387)
    at Zone.push../node_modules/zone.js/dist/zone.js.Zone.run (zone.js:138)
    at zone.js:872
    at ZoneDelegate.push../node_modules/zone.js/dist/zone.js.ZoneDelegate.invokeTask (zone.js:421)
    at Object.onInvokeTask (core.js:3811)
    at ZoneDelegate.push../node_modules/zone.js/dist/zone.js.ZoneDelegate.invokeTask (zone.js:420)
    at Zone.push../node_modules/zone.js/dist/zone.js.Zone.runTask (zone.js:188)
[...]

PR #18806, opened 2019-07-16
fix(menu-controller): add swipeGesture proxy

kensodemann:
## Pull request checklist

## Pull request type

 

 

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

Currently, using `MenuController.swipeEnable()` causes a `Cannot read property 'apply' of undefined` error as `swipeEnable` does not exist on the component being proxied.

Issue Number: #16002

## What is the new behavior?

- `swipeEnable()` still exists on the controller, but proxies `swipeGesture`
- `swipeGesture()` has been added to the controller

It would probably be best to deprecate `swipeEnable()` and remove it in a future version, but I am not sure how to do that.

- [ ] Yes
- [x] No


angular/src/providers/menu-controller.ts
```diff
@@ -61,7 +61,17 @@ export class MenuController {
    * @return Returns the instance of the menu, which is useful for chaining.
    */
   swipeEnable(shouldEnable: boolean, menuId?: string): Promise<HTMLIonMenuElement> {
-    return proxyMethod(CTRL, this.doc, 'swipeEnable', shouldEnable, menuId);
+    return proxyMethod(CTRL, this.doc, 'swipeGesture', shouldEnable, menuId);
```
manucorporat: I would call swipeGesture() h
```suggestion
    console.warn('MenuController.swipeEnable is deprecated. Use MenuController.swipeGestute() instead');
    return this.swipeGestute(shouldEnable, menuId);
```

angular/src/providers/menu-controller.ts
```diff
@@ -61,7 +61,17 @@ export class MenuController {
    * @return Returns the instance of the menu, which is useful for chaining.
```
manucorporat: Add a 
```suggestion
   * @return Returns the instance of the menu, which is useful for chaining.
   * @deprecated Use swipeGesture() instead.
```

angular/src/providers/menu-controller.ts
manucorporat: cc @brandyscarney maybe you have an input on this, how to add deprecation warning

PR #18827, opened 2019-07-18
feat(fab-button): optional slot for close icon

peterpeterparker:
## Pull request checklist

## Pull request type

- [X] Feature

Currently `ion-fab-button` have a reference on the ionicons, using the `<ion-icon name="close"/>` as unique close button for the fab (in case it would be use in a container with `fab-list`).

## Issue Number:

https://github.com/ionic-team/ionic/issues/18442

## What is the new behavior?
Instead of a unique reference respectively an `ion-icon`, the suggestion is to provide an optional `<slot name="close"/>` to let the user be able to provide his/her own icon for the close action. 

As commented in the PR, majority were for a non breaking change, therefore the default ion-icon close should remain as default option.

- [X] No


core/src/components/fab-button/fab-button.tsx
```diff
@@ -153,7 +153,7 @@ export class FabButton implements ComponentInterface, AnchorInterface, ButtonInt
         onClick={(ev: Event) => openURL(this.href, ev, this.routerDirection)}
       >
         <span class="close-icon">
-          <ion-icon name="close" lazy={false}></ion-icon>
+          <slot name="close"></slot>
```
topalavlad: Why not use the previous `ion-icon` as default? Without it, you're forcing people to change their code on the next upgrade.

peterpeterparker: Watching for changes to detect if a slot if provided in order to not display a default icon seems an overhead to me

Furthermore, I think the the component should be "ionicons reference free" respectively it should not be mandatory to fetch the ionicons in order to use that component

But fill free to improve the PR

topalavlad: > Watching for changes to detect if a slot if provided

You don't have to watch for changes. Anything inside the `slot` declaration will be rendered if no `slot` is provided. I agree with the "reference free" but to me it seems unnecessary to introduce a breaking change.

I had created a [PR](https://github.com/ionic-team/ionic/pull/18519) actually but I think yours is good because it has the documentation changes as well.

manucorporat: We can't merge PRs that introduce breaking changes, @topalavlad solution should work great:

```suggestion
          <slot name="close">
            <ion-icon name="close" lazy={false}></ion-icon>
          </slot>
```

peterpeterparker: @manucorporat my bet is that in any case, one day, this breaking change will happen to remove the dependency on the ion-icons. For example in order to be able to ship this component as a standalone one.

That being said, I've modified the PR as you and @topalavlad suggested.

topalavlad: `ionicons` should not be a dependency after this change for `fab-button` if people fill the slot, right? 
It's also probably about when the breaking changes happen, not if they ever do. Probably a major version would be suitable for removing ionicons.

manucorporat: ionicons is a different package already but an important part of ionic, i am fine to allow a system to use different icons, but ionions and ionic are part of the same project.

core/src/components/fab-button/readme.md
```diff
@@ -4,6 +4,13 @@ Floating Action Buttons (FABs) represent the primary action in an application. B
 
 If the FAB button is not wrapped with `<ion-fab>`, it will scroll with the content. FAB buttons have a default size, a mini size and can accept different colors:
 
+The below chart has a description of each slot.
```
manucorporat: The documentation of slots is part of stencil:
```
 * @slot - Content is placed between the named slots if provided without a slot.
```
add a comment like this in the component itself, like this: https://github.com/ionic-team/ionic/blob/master/core/src/components/toolbar/toolbar.tsx#L10-L14

Then run `stencil build` it will update the readme

thomasbrus:
@peterpeterparker Why was this closed? :) I still experience the issue that the close icon is not showing. For now I've solved it as follows (might anyone come across this):

```tsx
const MyFabButton = () => {
  const fabButtonRef = useRef<HTMLIonFabButtonElement>(null);

  const handleClick = () => {
    const fabButton = fabButtonRef.current!;
    const buttonInner = fabButton.shadowRoot!.querySelector('.button-inner') as HTMLSpanElement;
    buttonInner.style.opacity = '1';
    buttonInner.style.transform = fabButton.activated ? 'scale(0.8) rotate(45deg)' : '';
  };

  return (
    <IonFabButton ref={fabButtonRef} onClick={handleClick}>
      <IonIcon icon={add} size="large" />
    </IonFabButton>
  );
};```

peterpeterparker:
@thomasbrus I think I closed it because I had the feeling that it would never been solved and I replaced the use of `<ion-fab-button/>` by custom `<button/>` and styling.

PR #18841, opened 2019-07-19
feat(events): add unsubscribeId method

dylanvdmerwe:
## Pull request checklist

## Pull request type
- [ ] Bugfix
- [x] Feature
- [ ] Build related changes

As per the docs, you currently subscribe to an event as follows:
```typescript
 events.subscribe('user:created', (user, time) => {
    console.log('Welcome', user, 'at', time);
  });
```

The problem here is that this does not make it easy to unsubscribe from this event. In this way, users may not know that they need to unsubscribe from the event, which could cause memory leaks. Especially if using the arrow syntax.

## What is the new behavior?
As per the `window.setTimeout` and `window.setInterval` methods, this PR makes the `subscribe` return a number to identify the subscription. 

So with an interval you would use the following pattern:
```typescript
let timerId : number; 

ngOnInit() {
   this.timerId = window.setInterval(() => console.log('interval'), 1000);
}

ngOnDestroy() {
   window.clearInterval(this.timerId);
}
```

In the same sense, now the following pattern can be used for events to clean up:
```typescript
let eventId : number;

ngOnInit() {
 this.eventId = this.events.subscribe('user:created', (user, time) => {
    console.log('Welcome', user, 'at', time);
  });
}

ngOnDestroy() {
  this.events.unsubscribeId(this.eventId);
}
```

- [x] Yes
- [ ] No

The only main change in this PR is that fact that you can only register one `EventHandler` object at the time now instead of multiple. This is the only interface change.

There is only one new method `unsubscribeId`.

### DIFF
--- a/angular/src/providers/events.ts
+++ b/angular/src/providers/events.ts
@@ -4,21 +4,31 @@ export type EventHandler = (...args: any[]) => any;
 @Injectable({
   providedIn: 'root',
 })
-export class Events {
-  private c = new Map<string, EventHandler[]>();
+export class AppEvents {
+  private c = new Map<string, EventHandlerHolder[]>();
 
   /**
    * Subscribe to an event topic. Events that get posted to that topic will trigger the provided handler.
    *
    * @param topic the topic to subscribe to
    * @param handler the event handler
+   *
+   * @return random id of the registered handler to unsubscribe
    */
-  subscribe(topic: string, ...handlers: EventHandler[]) {
-    let topics = this.c.get(topic);
+  subscribe(topic: string, handler: EventHandler): number {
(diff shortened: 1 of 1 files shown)


manucorporat:
I don't think we want to keep adding features to Events, seems like there are better APIs built into angular or external packages to handle this kind of tasks.

liamdebeasi:
Thanks for the PR! We appreciate all of the hard work you put into developing this. The Events service in Ionic Angular has been deprecated and will be removed in the next major release of Ionic. As a result, we have decided not to add any new functionality to the Events service.

I am going to close this PR, but we will be providing guidance on how to integrate the Events service alternatives into your app soon. Thanks! 🙂

PR #18842, opened 2019-07-19
WIP: Modal swipe to close and card presentation style

mlynch:
This adds a new card presentation style for modals which is the new default in iOS 13: 

It also adds a swipe-to-close gesture for both fullscreen and card presented modals.

Note: the new card modal is only available on iOS since Android does not use this style of UI from what we can tell.

Things that aren't right yet that we need to figure out:

- [ ] Where to put the `.ion-page-card` and `.ion-page-card-reset` classes which need to be applied to the presenting view (so, outside of the scoped modal classes). They are currently in `core.scss`
- [x] Linting is failing, haven't bothered to fix that yet

image


core/src/css/core.scss
```diff
@@ -127,3 +127,12 @@ html.plt-ios.plt-hybrid, html.plt-ios.plt-pwa {
     --ion-safe-area-right: env(safe-area-inset-right);
   }
 }
+
+
+.presenting-view-card {
```
mlynch: I hate that these are here. They are meant to be applied to a view/page that is presenting a modal, so we can apply a cool effect to scale it down. They are also iOS only.

Is there a per-platform CSS file for pages/views to put these in?

Livijn:
Yes, please. This would be lit!

mlynch:
One other open question here is if we want to support the modal stacking. For example, if you open a modal and the modal opens another modal (and so on and so forth), the modals "stack". I think we should but that is going to require some kind of stacking system and more work.

mlynch:
Made some tweaks to better support stacking

PR #18918, opened 2019-07-26
feat(animation): Ionic Animation utility

liamdebeasi:
## Pull request checklist

## Pull request type

 

 

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes

Issue Number: N/A

## What is the new behavior?

-
-
-

- [ ] Yes
- [ ] No


core/src/utils/animation/animation-utils.ts
```diff
@@ -0,0 +1,83 @@
...
+    unRegTrans = () => {
+      el.removeEventListener('webkitAnimationEnd', onTransitionEnd, opts);
+      el.removeEventListener('animationend', onTransitionEnd, opts);
+    };
+  }
+
+  return unregister;
+};
+
+export const supportsWebAnimations = (): boolean => {
```
manucorporat: since it's const, we could do:
```
export const supportsWebAnimations = /*@__PURE__*/(() => {
return !!(window as any).Animation;
})();
```
so only runs once

core/src/utils/animation/animation-utils.ts
```diff
@@ -0,0 +1,83 @@
...
+export const generateKeyframeString = (name: string | undefined, keyframes: any[] = []): string => {
+  if (name === undefined) { console.warn('A name is required to generate keyframes'); }
+
+  const keyframeString = [`@keyframes ${name} {`];
+
+  keyframes.forEach(keyframe => {
+    const offset = keyframe.offset;
+
+    const frameString = [];
+    for (const property in keyframe) {
```
manucorporat: Object.keys()?

core/src/utils/animation/animation-utils.ts
```diff
@@ -0,0 +1,83 @@
...
+  stylesheet.id = stylesheetId;
+  stylesheet.appendChild(document.createTextNode(keyframeString));
+
+  const rootNode = (element.getRootNode() as any);
+  const styleContainer = (rootNode.head || rootNode);
+
+  const existingStylesheet = rootNode.querySelector(`#${stylesheetId}`);
+  if (existingStylesheet) { return; }
+
+  styleContainer.appendChild(stylesheet);
```
manucorporat: if we appendChild() maybe we should call the funcion add() instead of create()? just a though

core/src/utils/animation/animation.ts
```diff
@@ -0,0 +1,816 @@
...
+  const cleanUpElements = () => {
+    if (supportsWebAnimations()) {
+      webAnimations.forEach(animation => {
+        animation.cancel();
+      });
+
+      webAnimations = [];
+    } else {
+      elements.forEach(element => {
+        element.style.removeProperty('animation-name');
```
manucorporat: wonder if we can do something that minifies better

core/src/utils/animation/animation.ts
```diff
@@ -0,0 +1,816 @@
...
+
+    return generatePublicAPI();
+  };
+
+  /**
+   * Remove CSS class from this animation's elements
+   * before the animation begins.
+   */
+  const beforeRemoveClass = (className: string | string[] | undefined): Animation => {
+    beforeRemoveClasses = addClassToArray(beforeRemoveClasses, className);
```
manucorporat: wonder if we can use something like
```ts
beforeRemoveClasses.push(...parseClassname(className))
```

core/src/utils/animation/animation.ts
```diff
@@ -0,0 +1,816 @@
...
+  };
+
+  const getFill = (): 'auto' | 'none' | 'forwards' | 'backwards' | 'both' | undefined => {
+    if (_fill !== undefined) { return _fill; }
+    if (parentAnimation) { return parentAnimation.getFill(); }
+
+    return undefined;
+  };
+
+  const getDirection = (): 'normal' | 'reverse' | 'alternate' | 'alternate-reverse' | undefined => {
```
manucorporat: this `'normal' | 'reverse' | 'alternate' | 'alternate-reverse'` is used many places, let's create a TS type: 

KeyframeDirection? maybe?

core/src/utils/animation/animation.ts
```diff
@@ -0,0 +1,816 @@
...
+   */
+  const afterClearStyles = (propertyNames: string[] = []): Animation => {
+    for (const property of propertyNames) {
+      afterStylesValue[property] = '';
+    }
+
+    return generatePublicAPI();
+  };
+
+  const getFill = (): 'auto' | 'none' | 'forwards' | 'backwards' | 'both' | undefined => {
```
manucorporat: make `'auto' | 'none' | 'forwards' | 'backwards' | 'both' ` a type

core/src/utils/animation/animation.ts
```diff
@@ -0,0 +1,816 @@
...
+    return generatePublicAPI();
+  };
+
+  const parent = (animation: Animation): Animation => {
+    parentAnimation = animation;
+
+    return generatePublicAPI();
+  };
+
+  const addElement = (el: Element | Element[] | Node | Node[] | NodeList | undefined | null): Animation => {
```
manucorporat: Node | Node[] shold be enough, Element is already a Node (i think)

core/src/utils/animation/animation.ts
```diff
@@ -0,0 +1,816 @@
...
+    if (el != null) {
+      const nodeList = el as NodeList;
```
manucorporat: this does not look safe? why do we treat it as a NodeList?

core/src/utils/animation/animation.ts
```diff
@@ -0,0 +1,816 @@
...
+        webAnimations.forEach(animation => {
+          animation.currentTime = animation.effect.getComputedTiming().delay + (getDuration()! * step);
+          animation.pause();
+        });
+      } else {
+        const animationDuration = `-${getDuration()! * step}ms`;
+
+        elements.forEach(element => {
+          if (_keyframes.length > 0) {
+            (element as HTMLElement).style.animationDelay = animationDuration;
```
manucorporat: maybe we only need to accept HTMLElement in first instance?

core/src/utils/animation/animation.ts
```diff
@@ -0,0 +1,816 @@
...
+    }
+
+    return generatePublicAPI();
+  };
+
+  const fromTo = (property: string, fromValue: any, toValue: any): Animation => {
+    return from(property, fromValue).to(property, toValue);
+  };
+
+  const generatePublicAPI = (): Animation => {
```
manucorporat: ok! maybe we need some `function` instead of arrow functions, this way we can return `this` instead of calling `generatePublicAPI` everything, ideally, we shold be able to reuse the same object

```
const a = b.from();
a === b
```

liamdebeasi:
Noticing one bug where the `animationend` event does not fire after updating an in-progress animation with CSS animations. Web Animations appears to be working fine

PR #18948, opened 2019-07-30
docs(nav): typos

rtpHarry:
## Pull request checklist

## Pull request type

 

 

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes

Issue Number: N/A

## What is the new behavior?

-
-
-

- [ ] Yes
- [ ] No


core/src/components/nav/readme.md
```diff
@@ -1,7 +1,8 @@
 # ion-nav
 
-Nav is a standalone component for loading arbitrary components and pushing to new components on to the stack.
-Unlike RouterOutlet, Nav is not tied to a particular router. Meaning that if we load a Nav component, and push other components to the stack, they will not affect the apps overall router. This fits use cases where you could have a modal, which needs it's own sub-navigation, but not make it tied to the apps URL.
+Nav is a standalone component for loading arbitrary components and pushing new components on to the stack.
+
+Unlike RouterOutlet, Nav is not tied to a particular router. Meaning that if we load a Nav component, and push other components to the stack, they will not affect the apps overall router. This fits use cases where you could have a modal, which needs it's own sub-navigation, but without making it tied to the apps URL.
```
brandyscarney:
```suggestion
Unlike Router Outlet, Nav is not tied to a particular router. This means that if we load a Nav component, and push other components to the stack, they will not affect the app's overall router. This fits use cases where you could have a modal, which needs its own sub-navigation, without making it tied to the apps URL.
```

rtpHarry: I accepted as you are vastly more experienced, but isn't RouterOutlet referring to a component class?

brandyscarney: When talking about the components outside of their element/tag name I like to keep them separated, like for `<ion-picker-controller>` we don't say "PickerController" but rather Picker Controller. Item dividers as well. We're really just talking about the `ion-router-outlet` or `IonRouterOutlet` component, and `RouterOutlet` is not an exposed class.

This is what I am trying to avoid: https://ionicframework.com/docs/api/list-header

I think we should just call them what they are in terms of components and leave the code examples to the usage. 

Take a look at how iOS does it for their segment: https://developer.apple.com/design/human-interface-guidelines/ios/controls/segmented-controls/. The class is a `UISegmentedControl` but they just refer to it by what it is. If anything we should be using lower case references to other components.

rtpHarry: Thank you for the detailed clarification, I will follow this format moving forward.

brandyscarney:
Thanks 🙂

PR #18980, opened 2019-08-02
feat(searchbar): add inputmode property

5im0n:
## Pull request checklist

## Pull request type

 

 

- [ ] Bugfix
- [X] Feature
- [ ] Build related changes

Ionic searchbar component can now have a `inputmode` property.
The default value is set to `search`. https://developer.mozilla.org/fr/docs/Web/HTML/Attributs_universels/inputmode

- [ ] Yes
- [X] No


core/src/components/searchbar/searchbar.tsx
```diff
@@ -81,6 +81,12 @@ export class Searchbar implements ComponentInterface {
    */
   @Prop() disabled = false;
 
+  /**
+   * A hint to the browser for which keyboard to display.
+   * This attribute applies when the value of the type attribute is `"text"`, `"password"`, `"email"`, or `"url"`. Possible values are: `"verbatim"`, `"latin"`, `"latin-name"`, `"latin-prose"`, `"full-width-latin"`, `"kana"`, `"katakana"`, `"numeric"`, `"tel"`, `"email"`, `"url"`.
+   */
+  @Prop() inputmode = 'search';
```
manucorporat: If there is a well-known set of values why not define it in the type itself:
```suggestion
  @Prop() inputmode: 'verbatim', 'latin', 'latin-name', 'latin-prose', 'full-width-latin', 'kana', 'katakana', 'numeric', 'tel', 'email', 'url' = 'search';
```

5im0n: It make sence, but I think we should review the set of values.
As we can read in the [HTML spec](https://html.spec.whatwg.org/multipage/interaction.html#input-modalities:-the-inputmode-attribute) somes attributes like `latin`, `latin-prose`... doens't exist.

As the spec say, i suggest: 

`@Prop() inputmode: 'none' | 'text' | 'tel' | 'url' | 'email' | 'numeric' | 'decimal' | 'search' = 'search';`

We should update the `inputmode` for the `ion-input` component as well !

manucorporat: agreed, it should be based in what the spec says, and updating ion-inputs would be great but i would update ion-input in a different PR though, it might be a breaking change

5im0n: I push a new commit to set the default values that match the specs.

Do you want that a create a new PR to do the same thing on the `ion-input` ?

Issue #18940, opened 2019-07-30
bug: toggle button background on ios


strbam:
# Bug Report

**Ionic version:**
[x] **4.7.1**
[x] **4.7.0**

**Current behavior:**
Background color for toogle button is not set on ios devices.

**Expected behavior:**
Parameter ```--background```  sets toggle background. Similar behavior as ```--background-checked``` is expected.

**Steps to reproduce:**
1) Create ionic project
2) Add toggle button ``` <ion-toggle> ```
3) Use styling code that is below
4) Deploy to ios device

**Related code:**

```
:root {
    --ion-background-color: #00b98b;
}

ion-toggle {
    --background: orange;
    --handle-background: navy;

    --background-checked: darkgreen;
    --handle-background-checked: khaki;
}
```

**Result on iPad**:
![ipad](https://user-images.githubusercontent.com/53467212/62111075-17839b80-b2b0-11e9-9b8b-955342e8cb9c.jpg)

**Result on Safari**:
![safari](https://user-images.githubusercontent.com/53467212/62111076-17839b80-b2b0-11e9-8788-f98b1e5e9af2.jpg)

**Other information:**

**Ionic info:**
```
Ionic:

   ionic (Ionic CLI)             : 4.10.2 (/usr/local/lib/node_modules/ionic)
   Ionic Framework               : @ionic/angular 4.7.1
   @angular-devkit/build-angular : 0.801.2
   @angular-devkit/schematics    : 8.1.2
   @angular/cli                  : 8.1.2
   @ionic/angular-toolkit        : 2.0.0

Cordova:

   cordova (Cordova CLI) : 9.0.0 (cordova-lib@9.0.1)
   Cordova Platforms     : ios 5.0.1
   Cordova Plugins       : cordova-plugin-ionic-keyboard 2.1.3, cordova-plugin-ionic-webview 4.1.1, (and 4 other plugins)

System:

   ios-deploy : 1.9.4
   ios-sim    : 8.0.1
   NodeJS     : v11.9.0 (/usr/local/bin/node)
   npm        : 6.9.0
   OS         : macOS Mojave
   Xcode      : Xcode 10.3 Build version 10G8
```

PR #19012, opened 2019-08-06
fix(toggle): change background to use CSS variable

sethlilly:

Background color for `ion-toggle` is now properly set in iOS mode.

- Change `.toggle-icon` property `background-color` to `background` for consistency with `checked` state and the `.toggle-inner` `background` property
- Change `background` property to use CSS variable

**Ionic Version**:

- [x] 4.7.0
- [x] 4.7.1

**Fixes**: #18940


brandyscarney:
Thanks for the PR! This is good to merge when the build passes. :tada:

brandyscarney:
Thank you again!

PR #19066, opened 2019-08-10
Revert "feat(spinner): add circular spinner for MD default"

brightest-dazes:
Reverts ionic-team/ionic#19052


brandyscarney:
Thank you for the PR, however we will not be reverting this as it is a feature that was added to get us closer to the Material Design spec. Thanks for understanding!

PR #19144, opened 2019-08-21
docs(item): wrap usage used directive instead of class

rtpHarry:
I'm not sure if this is correct but I have created the PR ready for merge if this is an issue. Please review before accepting.

I thought that the new approved way for Ionic 4 was to use `ion-text-wrap` over the `text-wrap` directive.

As I understand it this is a general policy as not all platforms support directives.

It seems the `text-wrap` directive was still lurking in the usage examples.

### DIFF
--- a/core/src/components/item/usage/angular.md
+++ b/core/src/components/item/usage/angular.md
@@ -68,14 +68,14 @@ List Items
   </ion-item>
 
   <ion-item>
-    <ion-label text-wrap>
+    <ion-label class="ion-text-wrap">
     Multiline text that should wrap when it is too long
     to fit on one line in the item.
     </ion-label>
   </ion-item>
 
   <ion-item>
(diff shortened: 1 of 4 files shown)


brandyscarney:
Yes, this is correct! Thank you! Could you please run `npm run build` and commit the changes? It should make some automatic updates to the readme file. 🙂

rtpHarry:
I'm on Window 10. It's converted all the include paths in the docs from `/` to `\`:

image

Is this going to mess everything up if I try to commit it?

rtpHarry:
Done some digging and it seems that its currently an issue with Stencil on Windows:

https://github.com/ionic-team/stencil/issues/1708

rtpHarry:
Made a patch for my install, and also found some more references once I had it as a local repo.

I think this is good to go now?

PR #19382, opened 2019-09-18
fix(ion-content): Position fixed slot relatively

matthewdenobrega:
## Pull request checklist

## Pull request type

 

 

- [X] Bugfix
- [ ] Feature
- [ ] Build related changes

Issue Number: #17754

## What is the new behavior?

The issue here was that the .inner-scroll class on main positions it absolutely, and main fills the whole screen. The fixed slot was not positioned (ie static) which means that it was hidden by main, even when main didn't have any content. Positioning the fixed slot relatively resolves this, and as far as I can see won't break anything, as the slot has no size when it does not have content.

Positioning the slot itself relatively didn't work, presumably because it is display:contents - so I put a wrapper. I'm happy to tweak this if there is a more elegant solution. I'm also not 100% sure on the naming, this is my first contribution to ionic.

- [ ] Yes
- [X] No

### DIFF
--- a/core/src/components/content/content.scss
+++ b/core/src/components/content/content.scss
@@ -119,6 +119,10 @@
   position: relative;
 }
 
+.fixed-slot-position {
+  position: relative;
+}
+
 .transition-effect {
   position: absolute;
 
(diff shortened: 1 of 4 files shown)


brandyscarney:
Thanks again for the PR! I am going to close this as this proposed solution breaks fab, however I did find what I think works for all cases and created a PR here: https://github.com/ionic-team/ionic/pull/19867

I'll add you as a co-author!

PR #19395, opened 2019-09-19
refactor(card): convert card to shadow

adamdbradley:


manucorporat:
There quite a lot of CSS that will stop working, which might be fine, but in that case it should be removed

PR #19505, opened 2019-09-30
fix(component): fix icon type

manucorporat:
Fixes https://github.com/ionic-team/ionic/issues/19503


liamdebeasi:
Looks like this was already fixed in https://github.com/ionic-team/ionic-framework/pull/19588

PR #19097, opened 2019-08-14
feat: ionCheck and ionUncheck for ion-checkbox

mahnuh:
## Pull request checklist

Please check if your PR fulfills the following requirements:
- [ ] Tests for the changes have been added (for bug fixes / features)
- [x] Docs have been reviewed and added / updated if needed (for bug fixes / features)
- [x] Build (`npm run build`) was run locally and any changes were pushed
- [x] Lint (`npm run lint`) has passed locally and any fixes were made for failures

## Pull request type

Please check the type of change your PR introduces:
- [ ] Bugfix
- [x] Feature
- [ ] Code style update (formatting, renaming)
- [ ] Refactoring (no functional changes, no api changes)
- [ ] Build related changes
- [x] Documentation content changes
- [ ] Other (please describe): 

## What is the current behavior?

Issue Number: #19096 

## What is the new behavior?

- added ionCheck event to ion-checkbox which will be emitted when the checkbox is being checked
- added ionUncheck event to ion-checkbox which will be emitted when the checkbox is being unchecked

## Does this introduce a breaking change?

- [ ] Yes
- [x] No

## Other information

--

PR #19509, opened 2019-10-01
Feat ion toggle toggle event

mahnuh:
## Pull request checklist

## Pull request type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes

The `ionChange` event will be emitted as soon as the toggle's value changes, whether that is because the toggle is being toggled manually or the value changes otherwise, programatically for example. At the moment from the `ionChange` event there is no way to determine if the value change happened manually or automatically.
This is basically the same issue as mentioned in PR #19097 for ion-checkbox.

## What is the new behavior?

This PR adds the `ionToggle` event which is only emitted when the toggle actually is being toggled by the user.

- [ ] Yes
- [x] No

## Other information

--

### DIFF
--- a/angular/src/directives/proxies.ts
+++ b/angular/src/directives/proxies.ts
@@ -909,13 +909,14 @@ export declare interface IonToggle extends Components.IonToggle {}
 @Component({ selector: 'ion-toggle', changeDetection: ChangeDetectionStrategy.OnPush, template: '<ng-content></ng-content>', inputs: ['checked', 'color', 'disabled', 'mode', 'name', 'value'] })
 export class IonToggle {
   ionChange!: EventEmitter<CustomEvent>;
+  ionToggle!: EventEmitter<CustomEvent>;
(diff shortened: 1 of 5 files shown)


manucorporat:
I am not convinced about adding random APIs across inputs, we need to solve the problem for all inputs at once, and ship a consistent API for all components.

Maybe it could be a prop of ionChange itself, like "userInteraction" or "trigger"

mahnuh:
This actually would be great and solve so many problems. Should I open an issue for that?

liamdebeasi:
@manucorporat I'm wondering if the "trigger" would be good to add to all events for interactive components? I could see this being useful on `ion-checkbox` or `ion-radio` as well.

tricki:
This event should probably also be emitted when the state is [changed by dragging](https://github.com/ionic-team/ionic/blob/master/core/src/components/toggle/toggle.tsx#L139).

This might also be useful for `ion-input`, in which case "trigger" might not be the ideal name.

PR #19519, opened 2019-10-01
feat(datetime): configurable timezone for display

daem0ndev:
## Pull request checklist

## Pull request type

 

 

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes

Currently there is no way to override the behavior of the ion-datetime converting values to the users local timezone.

Issue Number: https://github.com/ionic-team/ionic/issues/19401

## What is the new behavior?

- Provide ability to specify the timezone for display on ion-datetime using the new `displayTimezone` property

- [ ] Yes
- [x] No


core/src/components/datetime/datetime.tsx
```diff
@@ -81,6 +81,12 @@ export class Datetime implements ComponentInterface {
    */
   @Prop() displayFormat = 'MMM D, YYYY';
 
+  /**
+   * The timezone to use for display purposes only. By default it will display values
+   * in the user's timezone.
+   */
+  @Prop() displayTimezone?: string;
```
liamdebeasi: What are the accepted values here? Maybe link to https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/toLocaleString

core/src/components/datetime/test/basic/index.html
```diff
@@ -142,6 +142,18 @@
...
 
+      <ion-list>
+        <ion-item>
+          <ion-label>Display UTC 00:00 in Local Timezone (default behavior)</ion-label>
+          <ion-datetime display-format="MMM DD, YYYY HH:mm" value="2020-01-01T00:00:00Z"></ion-datetime>
+        </ion-item>
+
+        <ion-item>
+          <ion-label>Display UTC 00:00 in UTC (display-timezone = 'utc')</ion-label>
+          <ion-datetime display-format="MMM DD, YYYY HH:mm" value="2020-01-01T00:00:00Z" display-timezone="utc""></ion-datetime>
```
liamdebeasi: Can we add another test that has something other than utc as well?

liamdebeasi:
```suggestion
          <ion-datetime display-format="MMM DD, YYYY HH:mm" value="2020-01-01T00:00:00Z" display-timezone="utc"></ion-datetime>
```

core/src/components/datetime/datetime.tsx
```diff
@@ -326,7 +332,15 @@ export class Datetime implements ComponentInterface {
              * there can be 1 hr difference when dealing w/ DST
              */
             const date = new Date(convertDataToISO(this.datetimeValue));
-            this.datetimeValue.tzOffset = date.getTimezoneOffset() * -1;
+
+            // If a custom display timezone is provided, use that tzOffset value instead
+            if (typeof(this.displayTimezone) !== 'undefined' && this.displayTimezone.length > 0) {
+              const utcDateTime = new Date(date.toLocaleString('en-US', { timeZone: 'utc' }));
```
liamdebeasi: This should be fine to default to `en-US` here right, since we're just manipulating these dates internally for now right?

core/src/components/datetime/datetime.tsx
```diff
@@ -326,7 +332,15 @@ export class Datetime implements ComponentInterface {
              * there can be 1 hr difference when dealing w/ DST
              */
             const date = new Date(convertDataToISO(this.datetimeValue));
-            this.datetimeValue.tzOffset = date.getTimezoneOffset() * -1;
+
+            // If a custom display timezone is provided, use that tzOffset value instead
+            if (typeof(this.displayTimezone) !== 'undefined' && this.displayTimezone.length > 0) {
```
liamdebeasi:
```suggestion
            if (this.displayTimezone !== undefined && this.displayTimezone.length > 0) {
```

core/src/components/datetime/datetime.tsx
daem0ndev: exactly, its simply used to determine the offsets only

daem0ndev:
> Not sure if this is a bug related to your changes, but the "Display UTC 00:00 in UTC (display-timezone = 'utc')" picker doesn't seem to be showing `2020` as a year.
> 
> image

I dont think its related because the picker above which does not use the displayTimezone does not show the 2020 either

core/src/components/datetime/test/basic/index.html
daem0ndev: done!

daem0ndev:
Made the changes :) thanks!

core/src/components/datetime/datetime-util.ts
```diff
@@ -242,12 +242,13 @@ export const parseDate = (val: string | undefined | null): DatetimeData | undefi
...
- * Converts a valid UTC datetime string
- * To the user's local timezone
+ * Converts a valid UTC datetime string to JS Date time object.
+ * By default uses the users local timezone, but an optional
+ * timezone can be provided.
  * Note: This is not meant for time strings
  * such as "01:47"
  */
-export const getLocalDateTime = (dateString: any = ''): Date => {
+export const getDateTime = (dateString: any = '', timeZone: any = ''): Date => {
```
liamdebeasi:
```suggestion
export const getDateTime = (dateString: any = '', timeZone?: string): Date => {
```

core/src/components/datetime/datetime-util.ts
```diff
@@ -284,14 +285,23 @@ export const getLocalDateTime = (dateString: any = ''): Date => {
       date.getMilliseconds()
     )
   );
+
+  if (timeZone && timeZone.length > 0) {
```
liamdebeasi: Doing the previous suggestion above, we can then do:
```suggestion
  if (timeZone !== undefined) {
```

No big performance benefit here, just a simpler way of doing it. Plus, if the user does input an empty string/invalid value for the timezone, the browser should throw an error. Thoughts?

core/src/components/datetime/datetime-util.ts
```diff
@@ -284,14 +285,23 @@ export const getLocalDateTime = (dateString: any = ''): Date => {
...
+    const tzDateTime = new Date(localDateTime.toLocaleString('en-US', { timeZone }));
+    const offset = utcDateTime.getTime() - tzDateTime.getTime();
+    return new Date(date.getTime() - offset);
+  }
+
+  return localDateTime;
 };
 
-export const updateDate = (existingData: DatetimeData, newData: any): boolean => {
+export const updateDate = (existingData: DatetimeData, newData: any, displayTimezone: any): boolean => {
```
liamdebeasi:
```suggestion
export const updateDate = (existingData: DatetimeData, newData: any, displayTimezone?: string): boolean => {
```

core/src/components/datetime/datetime.tsx
```diff
@@ -81,6 +81,14 @@ export class Datetime implements ComponentInterface {
    */
   @Prop() displayFormat = 'MMM D, YYYY';
 
+  /**
+   * The time zone to use for display purposes only. See
```
liamdebeasi:
```suggestion
   * The timezone to use for display purposes only. See
```

core/src/components/datetime/datetime-util.ts
```diff
@@ -284,14 +285,23 @@ export const getLocalDateTime = (dateString: any = ''): Date => {
       date.getMilliseconds()
     )
   );
+
+  if (timeZone && timeZone.length > 0) {
+    const utcDateTime = new Date(localDateTime.toLocaleString('en-US', { timeZone: 'utc' }));
+    const tzDateTime = new Date(localDateTime.toLocaleString('en-US', { timeZone }));
+    const offset = utcDateTime.getTime() - tzDateTime.getTime();
```
liamdebeasi: Another thought: Let's try moving these three lines to a util function in `datetime-util.ts`. I see we are using this code in two different places, so moving it to the util would be good.

Something like:

```typescript
const getTimezoneOffset = (date: Date, timezone: string) => {
  const utcDateTime = new Date(localDateTime.toLocaleString('en-US', { timeZone: 'utc' }));
  const tzDateTime = new Date(localDateTime.toLocaleString('en-US', { timeZone }));
  return utcDateTime.getTime() - tzDateTime.getTime();
}
```

core/src/components/datetime/datetime-util.ts
daem0ndev: I only did the any because thats how `dateString` was done and tried to be consistent :) will update today

liamdebeasi:
Merged. Thank you!

PR #19544, opened 2019-10-03
docs: publish Stencil JsonDocs types with @ionic/docs

camwiegert:
This adds a copy task to the Stencil config that copies Stencil's `JsonDocs` types to the docs package directory as `types.d.ts` and adds that to the files array in `package.json` so it can be published.

```js
{
  type: 'copy',
  dir: '../docs',
  copy: [
    {
      src: '../node_modules/@stencil/core/dist/declarations/docs.d.ts',
      dest: 'types.d.ts'
    }
  ]
}
```

The idea is to bundle the types as they were when the package was published so they can be used alongside the JSON data.

### DIFF
--- a/core/stencil.config.ts
+++ b/core/stencil.config.ts
@@ -94,6 +94,16 @@ export const config: Config = {
       file: '../docs/core.json'
     },
     {
+      type: 'copy',
+      dir: '../docs',
+      copy: [
+        {
+          src: '../node_modules/@stencil/core/dist/declarations/docs.d.ts',
+          dest: 'types.d.ts'
+        }
+      ]
+    },
+    {
       type: 'dist-hydrate-script'
     },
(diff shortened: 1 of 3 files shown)


camwiegert:
@brandyscarney @liamdebeasi I'm not sure how to test that this change works with the release process. Any ideas? My desired outcome is that `types.d.ts` is published alongside `core.json` in the docs package.

manucorporat:
How are the types supposed to be used?
types.d.ts does not have a default export, like you would expect in a json file.

manucorporat:
I think we can solve this problem much better at the stencil level, i would say not merge.

adamdbradley:
Great idea! We'll be able to do essentially the same thing, but from within the stencil build, since rebuilds may change types, and then others would be able take advantage of this. Thanks

Issue #19580, opened 2019-10-08
bug: IonTabButton href showing on browser as undefined when not defined


diogomateus:
**Bug Report**

**Ionic version:**
"@ionic/core": "^4.10.0",
"@ionic/react": "^4.10.0-rc.3",

**Current behavior:**
Having IonTabButton component and not passing an `href` prop, it's always returning undefined. This makes the page reload even if an onClick is passed. On documentation it says that href can be `string | undefined`.

**Expected behavior:**
Having IonTabButton component and its prop href is not defined, I expect the href not to show on the html.

**Ref:**
<img width="208" alt="Screenshot 2019-10-08 at 15 18 59" src="https://user-images.githubusercontent.com/4015038/66404375-3b660a80-e9e0-11e9-8b62-7a39c5a46cbd.png">

elylucas:
Hi @diogomateus,

Could you provide a snippet of what your <IonTabButton> looks like? Curious whats going on inside of the onClick. 

Also, it is not advised to avoid the href, as behind the scenes the href is set dynamically to forward to the correct page that was last left on that particular tab. Trying to navigate with an onClick will always take you to the same page.

elylucas:
Fixed in 4.11.0, thanks

PR #19588, opened 2019-10-08
fix(react): handle undefined attributes better, fixes #19563, #19580

### LINKED ISSUE #19563: bug: React IonRange is disabled even if `disabled === undefined`
# Bug Report

**Ionic version:**
"@ionic/core": "^4.10.0",
"@ionic/react": "^4.10.0-rc.3",

**Current behavior:**

Having `IonRange` component and it's prop `disabled` to which I pass variable that can be `type boolean | undefined`,
-> when the variable is equal to `false` IonRange is **enabled**
-> when the variable is equal to `true` IonRange is **disabled**
-> when the variable is equal to `undefined` IonRange is **enabled**

**Expected behavior:**

Having `IonRange` component and it's prop `disabled` to which I pass variable that can be `type boolean | undefined`,
-> when the variable is equal to `false` IonRange have to be  **enabled**
-> when the variable is equal to `true` IonRange have to be **disabled**
-> when the variable is equal to `undefined` IonRange have to be **disabled** because `undefined` is a falsy value

**Steps to reproduce:**

**Related code:**
```JSX
<div className={rangeModifier(urgencyIndex)}>
      <IonRange
        disabled={readOnly}
        max={urgency && urgency.size}
        min={1}
        mode='ios'
        onIonChange={onChange}
        snaps
        step={1}
        value={urgencyIndex}
      />
    </div>
```

elylucas:
## Pull request checklist

## Pull request type

 

 

- [X] Bugfix
- [ ] Feature
- [ ] Build related changes

Issue Number: #19563, #19580

## What is the new behavior?

-
-
-

- [ ] Yes
- [ ] No


packages/react/src/components/utils/attachEventProps.ts
```diff
@@ -19,7 +19,8 @@ export const attachEventProps = (node: HTMLElement, newProps: any, oldProps: any
         syncEvent(node, eventNameLc, newProps[name]);
       }
     } else {
-      if (typeof newProps[name] === 'object') {
+      const propType = typeof newProps[name];
+      if (propType === 'object' || propType === 'undefined') {
         (node as any)[name] = newProps[name];
       } else {
         node.setAttribute(camelToDashCase(name), newProps[name]);
```
manucorporat: Do we need to setAttribute at all? convert a prop name to dash case might not  work, we should always set property:
```
if (name in node) {
    node[name] = newProps[name];
}
```

Also, if our code adds attributes, we need to also add code to  remove them (I don't think we should do that at all), we should use properties 99%  of the times. In fact, we remove pure CSS attributes for this reason, also JSX  will complain about any non-prop passed

manucorporat: On a side note, why is this function called `attachEventProps` seems like it's doing a lot more than that

SatoAV:
Should this PR also fix this issue?
https://github.com/ionic-team/ionic/issues/19546

Very much hoping it does :)

PR #19658, opened 2019-10-14
fix(header): collapsible header works in tabs

liamdebeasi:
## Pull request checklist

## Pull request type

 

 

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

Issue Number: https://github.com/ionic-team/ionic/issues/19640

## What is the new behavior?

- Header now selects correct tab
- Also reduced the number of DOM reads/rights by moving the toolbar border animation to the intersection observer

- [ ] Yes
- [x] No

### DIFF
--- a/core/src/components/header/header.tsx
+++ b/core/src/components/header/header.tsx
@@ -60,10 +60,7 @@ export class Header implements ComponentInterface {
     if (!canCollapse && this.collapsibleHeaderInitialized) {
       this.destroyCollapsibleHeader();
     } else if (canCollapse && !this.collapsibleHeaderInitialized) {
-      const tabs = this.el.closest('ion-tabs');
-      const page = this.el.closest('ion-app,ion-page,.ion-page,page-inner');
-
-      const pageEl = (tabs) ? this.el.closest('ion-tab') : (page) ? page : null;
+      const pageEl = this.el.closest('ion-app,ion-page,.ion-page,page-inner');
       const contentEl = (pageEl) ? pageEl.querySelector('ion-content') : null;
 
       await this.setupCollapsibleHeader(contentEl, pageEl);
--- a/core/src/components/header/header.tsx
+++ b/core/src/components/header/header.tsx
@@ -2,7 +2,7 @@ import { Component, ComponentInterface, Element, Host, Prop, h, readTask, writeT
 
 import { getIonMode } from '../../global/ionic-global';
 
-import { cloneElement, createHeaderIndex, handleContentScroll, handleToolbarIntersection, setHeaderActive } from './header.utils';
+import { cloneElement, createHeaderIndex, handleContentScroll, handleToolbarBorderIntersection, handleToolbarIntersection, setHeaderActive } from './header.utils';
 /**
  * @virtualProp {"ios" | "md"} mode - The mode determines which platform styles to use.
  */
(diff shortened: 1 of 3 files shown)


core/src/components/header/header.tsx
```diff
@@ -57,14 +57,13 @@ export class Header implements ComponentInterface {
...
     const canCollapse = (hasCollapse && getIonMode(this) === 'ios') ? hasCollapse : false;
-
     if (!canCollapse && this.collapsibleHeaderInitialized) {
       this.destroyCollapsibleHeader();
     } else if (canCollapse && !this.collapsibleHeaderInitialized) {
       const tabs = this.el.closest('ion-tabs');
       const page = this.el.closest('ion-app,ion-page,.ion-page,page-inner');
 
-      const pageEl = (tabs) ? tabs : (page) ? page : null;
+      const pageEl = (tabs) ? this.el.closest('ion-tab') : (page) ? page : null;
```
DavidStrausz: `ion-tab` is not a component thats available, so `this.el.closest('ion-tab')` will always return `null` here.

core/src/components/header/header.tsx
```diff
@@ -101,21 +100,17 @@ export class Header implements ComponentInterface {
...
-        this.intersectionObserver = new IntersectionObserver(toolbarIntersection, { threshold: 0.25, rootMargin: `-${mainHeaderHeight}px 0px 0px 0px` });
+
+        /**
+         * Handle interaction between toolbar collapse and
+         * showing/hiding content in the primary ion-header
+         * as well as progressively showing/hiding the main header
+         * border as the top-most toolbar collapses or expands.
+         */
+        const toolbarIntersection = (ev: any) => { handleToolbarIntersection(ev, mainHeaderIndex, scrollHeaderIndex); };
+        this.intersectionObserver = new IntersectionObserver(toolbarIntersection, { threshold: [0.25, 0.3, 0.4, 0.5, 0.6, 0.7, 0.8, 0.9, 1], rootMargin: `-${mainHeaderHeight}px 0px 0px 0px` });
```
DavidStrausz: The `IntersectionObserver` API is available as of iOS 12.2, will there be a fallback for earlier versions of iOS?

core/src/components/header/header.tsx
liamdebeasi: Whoops, `ion-tab` is a component in vanilla JS but not in `@ionic/angular`. Will fix, thanks.

core/src/components/header/header.tsx
liamdebeasi: Probably not - The polyfill for this is not great. I'll add some feature detection to disable it so users don't get errors. If unsupported, would you be fine with `collapse="condense"` on the header and `collapse="true"` on buttons being ignored?

DavidStrausz: OK I see, I have to check how the look and feel is on a device where its disabled but my first thought would be to hide the collapsable - big - header entirely on devices that do not support it and just to display the small header.

DavidStrausz: So I tried the current behaviour on an iOS 11 device, currently only the large title in the collapsible header is visible and simply scrolls out of view, but because of the missing IntersectionObserver API (which throws an error of currently) the small title is never displayed. I think this is acceptable but having it the other way round (i.e. small title in header is always visible, large title never shows) would be nicer as it was the default until iOS 13 anyway.

And its now working inside tabs too!

liamdebeasi: Ok cool. I'll do the feature detection in a separate PR so we can get this bug fix out. 👍

PR #19670, opened 2019-10-15
refactor(ionicons): update to ionicons v5

adamdbradley:
🎉


core/api.txt
```diff
@@ -951,7 +951,7 @@ ion-searchbar,prop,disabled,boolean,false,false,false
 ion-searchbar,prop,inputmode,"decimal" | "email" | "none" | "numeric" | "search" | "tel" | "text" | "url",'search',false,false
 ion-searchbar,prop,mode,"ios" | "md",undefined,false,false
 ion-searchbar,prop,placeholder,string,'Search',false,false
-ion-searchbar,prop,searchIcon,string,'search',false,false
+ion-searchbar,prop,searchIcon,string | undefined,undefined,false,false
```
brandyscarney: Why is this undefined?

core/package.json
```diff
@@ -30,7 +30,7 @@
     "loader/"
   ],
   "dependencies": {
-    "ionicons": "^4.6.3",
+    "ionicons": "^5.0.0-11",
```
brandyscarney: Should we be updating to a pre-release? I'd like to use a final `5.0` officially

core/src/components/back-button/back-button.tsx
```diff
@@ -54,20 +54,32 @@ export class BackButton implements ComponentInterface, ButtonInterface {
...
-    return this.icon != null ? this.icon : config.get('backButtonIcon', 'arrow-back');
+  get backButtonIcon() {
+    const icon = this.icon || config.get('backButtonIcon');
+    if (icon != null) {
+      // icon set on the component, next from the config
+      return icon;
+    }
+
+    if (this.mode === 'ios') {
+      // default md back button icon
```
brandyscarney:
```suggestion
      // default ios back button icon
```

core/src/components/back-button/back-button.tsx
```diff
@@ -54,20 +54,32 @@ export class BackButton implements ComponentInterface, ButtonInterface {
    */
   @Prop() type: 'submit' | 'reset' | 'button' = 'button';
 
-  private get backButtonIcon() {
-    return this.icon != null ? this.icon : config.get('backButtonIcon', 'arrow-back');
+  get backButtonIcon() {
+    const icon = this.icon || config.get('backButtonIcon');
+    if (icon != null) {
+      // icon set on the component, next from the config
```
brandyscarney:
```suggestion
      // icon is set on the component or by the config
```

core/src/components/back-button/back-button.tsx
```diff
@@ -54,20 +54,32 @@ export class BackButton implements ComponentInterface, ButtonInterface {
...
+      // default md back button icon
+      return 'chevron-back';
+    }
+
+    // default md back button icon
+    return 'arrow-back-sharp';
   }
 
-  private get backButtonText() {
+  get backButtonText() {
```
brandyscarney: Why are we removing `private` from all of these?

core/src/components/icon/readme.md
```diff
@@ -2,10 +2,6 @@
 
 # ion-icon
 
-Icons can be used on their own as a standalone component, or inside of another component.
```
brandyscarney: Should we add this line back? It still holds true.

core/src/components/icon/readme.md
```diff
@@ -2,10 +2,6 @@
...
-
-Ionicons provide platform continuity out of the box in Ionic by rendering a different icon based on the device the app is running on. For example, by setting the icon `name` to `alarm` the icon will automatically use the `ios-alarm` on an iOS device, and the `md-alarm` for any device running Material Design. This allows the developer to write the markup once while Ionic handles displaying the appropriate icon based on the mode.
```
brandyscarney: Maybe since this has been removed we should give them an example of how to set the icon by mode?

core/src/components/searchbar/searchbar.tsx
```diff
@@ -412,8 +414,8 @@ export class Searchbar implements ComponentInterface {
   render() {
     const animated = this.animated && config.getBoolean('animated', true);
     const mode = getIonMode(this);
-    const clearIcon = this.clearIcon || (mode === 'ios' ? 'ios-close-circle' : 'md-close');
-    const searchIcon = this.searchIcon;
+    const clearIcon = this.clearIcon || (mode === 'ios' ? 'close-circle' : 'close-sharp');
+    const searchIcon = this.searchIcon || (mode === 'ios' ? 'search-outline' : 'search-sharp');
```
brandyscarney: Is there a way to set this in the prop so that the users can see the defaults?

core/api.txt
brandyscarney: Nevermind see my more recent comment on the searchIcon prop

core/src/components/back-button/back-button.tsx
adamdbradley: so i could unit test them. But doesn't actually matter and they're not made public since `@Method` is on them.

core/src/components/icon/readme.md
adamdbradley: 👍

PR #19778, opened 2019-10-29
Update input.tsx

eduardoroliveira:
Please add file type to input

## Pull request checklist

## Pull request type

 

 

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes

Issue Number: N/A

## What is the new behavior?

-
-
-

- [ ] Yes
- [x] No

### DIFF
--- a/core/src/components/input/input.tsx
+++ b/core/src/components/input/input.tsx
@@ -93,7 +93,7 @@ export class Input implements ComponentInterface {
    * Possible values: `"none"`, `"text"`, `"tel"`, `"url"`,
    * `"email"`, `"numeric"`, `"decimal"`, and `"search"`.
    */
-  @Prop() inputmode?: 'none' | 'text' | 'tel' | 'url' | 'email' | 'numeric' | 'decimal' | 'search';
+  @Prop() inputmode?: 'none' | 'text' | 'tel' | 'url' | 'email' | 'numeric' | 'decimal' | 'search' | 'file';
 
   /**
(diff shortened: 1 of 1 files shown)


liamdebeasi:
Thanks for the PR! `"file"` is not a valid value for `inputmode`. See https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/inputmode. Did you mean to change the `type` property instead?

eduardoroliveira:
The type field documented to ion-input says it supports: `"date" | "email" | "number" | "password" | "search" | "tel" | "text" | "time" | "url"`. My suggestion was to expand to also support type `file`, as documented in here `https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/file`.

Thanks

liamdebeasi:
Ok in that case you will need to do two things:

1. Change the `type` property to support `"file"` and revert your `inputmode` change.
2. Run `npm run lint.fix && npm run build` from inside the `core` directory.

I need to check with the team to see if this is something we want to add. We do not currently have proper styles for `input[type="file"]`.

liamdebeasi:
Thanks for the PR! We appreciate the work you put into creating this. After discussing with the team, we have decided not to accept this PR.

Our concern is that even if we accept `type="file"`, the interface that is rendered will still be incorrect when it comes to presenting a file upload field. We feel this will cause confusion as the interface that is rendered will still look like a text box.

I am going to close this PR, but if you are interested in seeing a file upload component/interface in Ionic please feel free to create a feature request. Thanks!

PR #19893, opened 2019-11-11
feat(select): add --placeholder-opacity, expose shadow parts

liamdebeasi:
## Pull request checklist

## Pull request type

 

 

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes

Issue Number: https://github.com/ionic-team/ionic/issues/17446

## What is the new behavior?

- Added `--placeholder-opacity and `--placeholder-color` css variables
- Added shadow parts for relevant elements

- [ ] Yes
- [x] No

### DIFF
--- a/core/src/components/select/select.tsx
+++ b/core/src/components/select/select.tsx
@@ -471,10 +471,10 @@ export class Select implements ComponentInterface {
           'select-disabled': disabled,
         }}
       >
-        <div class={selectTextClasses}>
+        <div class={selectTextClasses} part="select-text">
           {selectText}
         </div>
-        <div class="select-icon" role="presentation">
+        <div class="select-icon" role="presentation" part="select-icon">
           <div class="select-icon-inner"></div>
         </div>
         <button
--- a/core/src/components/select/select.tsx
+++ b/core/src/components/select/select.tsx
@@ -474,7 +474,7 @@ export class Select implements ComponentInterface {
         <div class={selectTextClasses} part="select-text">
           {selectText}
         </div>
-        <div class="select-icon" role="presentation" part="select-icon">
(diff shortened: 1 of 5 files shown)


core/src/components/select/select.tsx
```diff
@@ -471,10 +471,10 @@ export class Select implements ComponentInterface {
           'select-disabled': disabled,
         }}
       >
-        <div class={selectTextClasses}>
+        <div class={selectTextClasses} part="select-text">
```
brandyscarney: Do we want parts to be prefixed with select? In Segment we just add `part="indicator"` since the way you will target would be like:

```
ion-segment::part(indicator)
```

```suggestion
        <div class={selectTextClasses} part="text">
```

core/src/components/select/select.tsx
```diff
@@ -471,10 +471,10 @@ export class Select implements ComponentInterface {
...
           {selectText}
         </div>
-        <div class="select-icon" role="presentation">
+        <div class="select-icon" role="presentation" part="select-icon">
```
```suggestion
        <div class="select-icon" role="presentation" part="icon">
```

core/src/components/select/select.tsx
liamdebeasi: 👍  I removed the prefix

brandyscarney:
Approved once changes are made

gonewayword:
Is this still not working at all for anyone else? I cannot get the color or opacity to change no matter what. I've applied these style variables to the element, parent elements, etc, and no matter what it does not change. Even when the DOM shows it as being there, it's still getting overriden by:

.select-placeholder {
    color: currentcolor;
    opacity: 0.33;
}

Tried targeting that specifically too, nothing either.

PR #19932, opened 2019-11-17
fix(role): cancel role not close

rtpHarry:
I think this is correct?

## Pull request checklist

Please check if your PR fulfills the following requirements:
- [ ] Tests for the changes have been added (for bug fixes / features)
- [ ] Docs have been reviewed and added / updated if needed (for bug fixes / features)
- [ ] Build (`npm run build`) was run locally and any changes were pushed
- [ ] Lint (`npm run lint`) has passed locally and any fixes were made for failures

## Pull request type

 

 

Please check the type of change your PR introduces:
- [x] Bugfix
- [ ] Feature
- [ ] Code style update (formatting, renaming)
- [ ] Refactoring (no functional changes, no api changes)
- [ ] Build related changes
- [ ] Documentation content changes
- [ ] Other (please describe): 

## What is the current behavior?

Issue Number: N/A

## What is the new behavior?

-
-
-

## Does this introduce a breaking change?

- [ ] Yes
- [ ] No

PR #19944, opened 2019-11-19
fix(role): cancel role not close

rtpHarry:
Second attempt at #19932 which was auto closed when 5.0 branch was removed.

@brandyscarney


brandyscarney:
Thank you!!

Issue #20017, opened 2019-11-29
bug: animation not working on devices that require -webkit- prefix


DavidWiesner:
# Bug Report
On old devices the animation (page and overlay transition) did not work.

**Ionic version:**

[x] **4.x**

**Current behavior:**
When I run an app (e.g.: default ionic starter app) the page transition does not work on older devices (e.g.: android 5 with chrome 37). I discovered this devices will require an prefix to get the animation work. Even worst an popover will not be appear because its `popover-wrapper` has an initial opacity of zero. Without animation the popup will not appear.

**Expected behavior:**
Animations should also work on older devices.

**Steps to reproduce:**

1. `ionic start newApp blank`
2. `cd ./newApp`
3. `ionic serve --host=0.0.0.0 -- --disable-host-check`
4. open this page with your ip on an older device browser and navigate to a second page
5. no animation is triggered

**Related code:**

The `setStyleProperty`, `removeStyleProperty` and `createKeyframeStylesheet` should add an prefix (e.g.: `-webkit-`) for older devices if required. 

https://github.com/ionic-team/ionic/blob/d8161bc1297fe735b86cfe75ba135ef8aa8f09a5/core/src/utils/animation/animation-utils.ts#L2-L8

https://github.com/ionic-team/ionic/blob/d8161bc1297fe735b86cfe75ba135ef8aa8f09a5/core/src/utils/animation/animation-utils.ts#L80

**Other information:**
- Android 5.0 Chrome 37

**Ionic info:**
```
Ionic:

   ionic (Ionic CLI)             : 4.10.2 (/usr/local/lib/node_modules/ionic)
   Ionic Framework               : @ionic/angular 4.11.5
   @angular-devkit/build-angular : 0.801.3
   @angular-devkit/schematics    : 8.1.3
   @angular/cli                  : 8.1.3
   @ionic/angular-toolkit        : 2.1.1

System:

   NodeJS : v10.15.3 (/usr/local/Cellar/node@10/10.15.3/bin/node)
   npm    : 6.4.1
   OS     : macOS
```

liamdebeasi:
Thanks for the issue. You are correct, we need some prefixing for older browsers. I will look at your PR shortly.

liamdebeasi:
Thanks for the issue. This has been resolved via https://github.com/ionic-team/ionic/pull/20020 and will be available in an upcoming release of Ionic Framework.

PR #20020, opened 2019-11-29
fix(animation): support css animation on older devices

DavidWiesner:
## Pull request checklist

## Pull request type

 

 

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

Issue Number: #20017

## What is the new behavior?

- a function was added to detect if and what animation prefixes are required for the current browser  
- prefixes were added on two utility functions `setStyleProperty` and `removeStyleProperty ` 
- prefixes were added to the keyframe stylesheet generation `createKeyframeStylesheet `

- [ ] Yes
- [x] No


core/src/utils/animation/animation-utils.ts
```diff
@@ -1,10 +1,28 @@
...
+        }
+      }
+    }
+  }
+  return animationPrefix;
+};
 
 export const setStyleProperty = (element: HTMLElement, propertyName: string, value: string | null) => {
-  element.style.setProperty(propertyName, value);
+  element.style.setProperty(getAnimationPrefix(element) + propertyName, value);
```
liamdebeasi: Since the `setStyleProperty` and `removeStyleProperty` functions are used for other things not related to animation props, I think it makes sense to only apply the prefix if `propertyName.includes('animation')`.

DavidWiesner: Ok that's right. I will use `startsWith`

core/src/utils/animation/animation-utils.ts
```diff
@@ -1,10 +1,28 @@
+let animationPrefix: string | null = null;
+
+export const getAnimationPrefix = (ele: HTMLElement): string => {
+  if (animationPrefix === null) {
```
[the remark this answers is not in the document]
DavidWiesner: I had to add an `any` cast for the styles to prevent typescripts `Expression is always true` error.

core/src/utils/animation/animation-utils.ts
DavidWiesner: Typescript do not allow me to test the `propertyName` to be undefined or null without any cast. Therefore I did not test for an `undefined` or `null` on the `propertyName`.  Typescript prevent that the `propertyName` can be null, but I do not know this is true for any usecase in a javascript environment. So my question is should I add an null/undefined check for the `propertyName`?

liamdebeasi: You can assume that `propertyName` is always a string and never undefined.

core/src/utils/animation/animation-utils.ts
```diff
@@ -1,10 +1,26 @@
...
+  }
   element.style.setProperty(propertyName, value);
 };
 
 export const removeStyleProperty = (element: HTMLElement, propertyName: string) => {
-  element.style.removeProperty(propertyName);
+  if (propertyName.startsWith('animation')) {
+    propertyName = getAnimationPrefix(element) + propertyName;
+  }
+  element.style.removeProperty(getAnimationPrefix(element) + propertyName);
```
liamdebeasi: This looks like it will prepend the vendor prefix twice.

Test case:
`propertyName = 'animation-name'`

`if (propertyName.startsWith('animation')) {` returns true, so `propertyName = '-webkit-animation-name'.

And then you remove the property by doing `'-webkit-' + propertyName`, which will end up being `'-webkit--webkit-animation-name'`.



Might just be easiest to do something like:

```
const prefix = propertyName.startsWith('animation') ? getAnimationPrefix(element) : '';
element.style.removeProperty(`${prefix}${propertyName}`);
```
```

Comment (DavidWiesner) [reply core/src/utils/animation/animation-utils.ts]: Oh I'm really sorry to waste your time. I forgot to remove that.

Comment (liamdebeasi) [reply core/src/utils/animation/animation-utils.ts]: No worries - thanks!

File: core/src/utils/animation/animation-utils.ts [hunk 4]
```diff
@@ -1,9 +1,25 @@
+export const getAnimationPrefix = (el: HTMLElement): string => {
```
liamdebeasi: I think we can replace the implementation with the following:

```
const getAnimationPrefix = (): string => {
  const el = document.documentElement;
  const supportsUnprefixed = (el.style as any).animationName !== undefined;
  const supportsWebkitPrefix = (el.style as any).webkitAnimationName !== undefined;
  
  return (!supportsUnprefixed && supportsWebkitPrefix) ? '-webkit-' : '';
};

const animationPrefix: string = getAnimationPrefix();
```

This prevents us from a) having to pass in an element and b) having to check for a prefix every time we call `setStyleProperty` or `removeStyleProperty`.

core/src/utils/animation/animation-utils.ts
```diff
@@ -1,9 +1,25 @@
...
+
+  if (!supportsUnprefixed && supportsWebkitPrefix) {
+    return '-webkit-';
+  }
+
+  return '';
+};
 
 export const setStyleProperty = (element: HTMLElement, propertyName: string, value: string | null) => {
+  if (propertyName.startsWith('animation')) {
```
liamdebeasi: I'll leave this up to you, but I typically try to avoid modifying the values of function parameters (I treat them as if they are `const` except for very specific use cases).

So instead you could just do:

```
element.style.setProperty(
  propertyName.startsWith('animation') ? animationPrefix + propertyName : propertyName, 
  value
);
```

Again, totally up to you though.

core/src/utils/animation/animation-utils.ts
```diff
@@ -1,9 +1,25 @@
...
+    propertyName = getAnimationPrefix(element) + propertyName;
+  }
   element.style.setProperty(propertyName, value);
 };
 
 export const removeStyleProperty = (element: HTMLElement, propertyName: string) => {
+  if (propertyName.startsWith('animation')) {
```
liamdebeasi: Same thing as my previous comment re: modifying the values of function parameters.

core/src/utils/animation/animation-utils.ts
```diff
@@ -69,6 +85,7 @@ export const getStyleContainer = (element: HTMLElement) => {
 
 export const createKeyframeStylesheet = (keyframeName: string, keyframeRules: string, element: HTMLElement): HTMLElement => {
   const styleContainer = getStyleContainer(element);
+  const keyframePrefix = getAnimationPrefix(element);
```
liamdebeasi:
```suggestion
```

Making the first suggested change means that we can just grab the value of `animationPrefix`.

core/src/utils/animation/animation-utils.ts
```diff
@@ -77,7 +94,7 @@ export const createKeyframeStylesheet = (keyframeName: string, keyframeRules: st
 
   const stylesheet = (element.ownerDocument || document).createElement('style');
   stylesheet.id = keyframeName;
-  stylesheet.textContent = `@keyframes ${keyframeName} { ${keyframeRules} } @keyframes ${keyframeName}-alt { ${keyframeRules} }`;
+  stylesheet.textContent = `@${keyframePrefix}keyframes ${keyframeName} { ${keyframeRules} } @${keyframePrefix}keyframes ${keyframeName}-alt { ${keyframeRules} }`;
```
liamdebeasi: Same as previous comment. We can just grab the value of the variable now.

```suggestion
  stylesheet.textContent = `@${animationPrefix}keyframes ${keyframeName} { ${keyframeRules} } @${animationPrefix}keyframes ${keyframeName}-alt { ${keyframeRules} }`;
```

core/src/utils/animation/animation-utils.ts
DavidWiesner: I disagree with your suggestion. With this const variable is not possible to use ssr. In the first version of this pr I used an cache value. But at the end it's just a comparison to undefined. this is no expensive operation or not a more expensive operation than test for a cache value.

liamdebeasi: Ionic Animations wasn't compatible with SSR to begin with (see: https://github.com/ionic-team/ionic/blob/master/core/src/utils/animation/animation.ts#L67-L68). It's something on the list to address, but I think it's fine for now.

liamdebeasi: I'd just rather not constantly recheck the same value when we know the vendor prefix isn't going to change mid-session.

core/src/utils/animation/animation-utils.ts
DavidWiesner: I can change this.

core/src/utils/animation/animation-utils.ts
DavidWiesner: See my comments above

core/src/utils/animation/animation-utils.ts
liamdebeasi: What if we went back to the old way of passing in `el` to `getAnimationPrefix`, but then we cache the vendor prefix value and return it if it's already defined? That way we aren't doing additional unnecessary checks and we don't stray further from SSR support.

core/src/utils/animation/animation-utils.ts
```diff
@@ -24,12 +24,25 @@ const convertHyphenToCamelCase = (str: string) => {
   return str.replace(/-([a-z])/ig, g => g[1].toUpperCase());
 };
 
+let animationPrefix: string | null = null;
+
+export const getAnimationPrefix = (el: HTMLElement): string => {
+  if (animationPrefix === null) {
```
liamdebeasi: See comment above.
```suggestion
  if (animationPrefix === undefined) {
```

core/src/utils/animation/animation-utils.ts
```diff
@@ -24,12 +24,25 @@ const convertHyphenToCamelCase = (str: string) => {
   return str.replace(/-([a-z])/ig, g => g[1].toUpperCase());
 };
 
+let animationPrefix: string | null = null;
```
liamdebeasi: Ok sorry last thing -- we typically prefer undefined over null for consistency
```suggestion
let animationPrefix?: string;
```

core/src/utils/animation/animation-utils.ts
DavidWiesner: I do not know what the suggestion will do or prevent? How the animationPrefix can be undefined?  It is null when not initial assigned otherwise an empty string or '-webkit-'

core/src/utils/animation/animation-utils.ts
liamdebeasi: Just wanted to clarify: there's no bug here, but we usually prefer using `undefined` over `null`where applicable so it would be good to remain consistent.

DavidWiesner: My typescript (3.7.2) did not like a global variable to be marked as optional with `?`. 
Error:(27, 5) TS7034: Variable 'animationPrefix' implicitly has type 'any' in some locations where its type cannot be determined.
Error:(27, 20) TS1005: ',' expected.
Error:(27, 21) TS1134: Variable declaration expected.

liamdebeasi: Oh strange... ok `string | undefined` is fine. thanks!

liamdebeasi:
Merged. Thank you!

PR #20037, opened 2019-12-04
feat(refresher): iOS native refresher

liamdebeasi:
Dev build: `5.0.0-dev.201912061730.b8359e0`

needs review: the `contentId` prop. This was added so we can translate the content to not bounce back over the refresher without interfering with the user's styling (if we translated the inner scroll, users would see a white space where the content once was).

## Pull request checklist

## Pull request type

 

 

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes

Issue Number: https://github.com/ionic-team/ionic/issues/18664

## What is the new behavior?

- Added native iOS PTR behavior

- [ ] Yes
- [x] No


liamdebeasi:
Ok iOS refresher now falls back to the non-native refresher on desktop. I tried playing with the native refresher on desktop Safari, but the experience just didn't really feel too great. Even the "native" refresher in News on macOS doesn't really feel smooth.

`pulling-icon` and `refreshing-spinner` both default to `lines` for iOS. However, when using Desktop Safari (and thus falling back to the non-native refresher) the non-native refresher looks odd as a result (it will show a paused spinner, that rotates 180 degrees so it's a little awkward). Any thoughts on this?

Still not sure about how we can get around adding the `content-id` prop without breaking theming for users.

brandyscarney:
@liamdebeasi Maybe we just document that Safari / browsers on desktop running iOS mode won't use the native refresher & therefore will look "off". If they want it to look nice on both they need to do platform detection to change the value?

liamdebeasi:
Yeah a bunch of usage examples would probably be good here for that. I think running in iOS mode on desktop is a valid use case, so it would be ideal if looked right. Maybe we can have the refresher icon default to the old icon only on desktop iOS mode?

Issue #20218, opened 2020-01-17
feat: Add "context" property to IonReorderGroup


Program4GitHub:
# Feature Request

**Ionic version:** 5.4.13
**@ionic/react version:** 4.11.8

**Describe the Feature Request**

I'm using Ionic React, and I want to create an 'IonReorderGroup' inside a 'div' that has a scrollbar.
However, as I saw in Ionic's core code, the 'autoScroll' function works only with the 'IonContent' that wraps the Group.
Therefore, I want the 'IonReorderGroup' tag to have an optional prop (called 'context', for instance). 
In the 'context' prop I will be able to pass the ref of the 'div' element (or the div's id), and then the autoScroll function will affect the div's scrollbar when I drag an element.

**Related Code**

Solving this problem is pretty easy. All it requires is to add the 'context' property, then changing the following lines in the path:
> node_modules\@ionic\core\dist\collection\components\reorder-group\reorder-group.js
```
const contentEl = this.el.closest('ion-content');
```
Will become
```
var contentEl;
if (this.context !== null) contentEl = this.context;
else contentEl = this.el.closest("ion-content");
```
Or Something similar...

**Additional Context**

An example of a use case:
On the left, there is a 'div' that contains an 'IonReorderGroup'.
Because the list is long, the 'div' has a styled scrollbar.
However, when I drag an item, there is no autoScroll, so there are glitches.
With a 'context' property, such as `context={divElement}` or `context="divID"` this problem can be easily solved.
![image](https://user-images.githubusercontent.com/38870042/72578907-2f8fbb80-38e0-11ea-9353-6cc2e6cb7a54.png)

liamdebeasi:
Hello, thanks for the issue! We are currently exploring ways to improve our components when working with non `ion-content` scrollable areas. The particular use case given was for virtual scroll where scrolling is on a child element rather than `ion-content`, but we think the proposed API could work for your use case as well.

Discussions are happening on https://github.com/ionic-team/ionic-framework/pull/24426, and we hope to add this feature in the future.

To be clear, we are designing this feature with the virtual scroll use case in mind, so I would not expect perfect feature parity when using a non `ion-content` scrollable area. However, I think the API that is being proposed should work well enough such that the issue mentioned here no longer happens.

I am going to close this as well as the associated PR in favor of this new API. Please feel free to follow #24426 for updates on this new API. Thanks!

PR #20234, opened 2020-01-18
fix(reorder-group): Add 'context' property

Program4GitHub:
## Pull request checklist

## Pull request type

 

 

- [x] Feature

Issue Number: #20218

## What is the new behavior?

The context property allows autoScroll in reorder-group, even if it was created inside a small div (a div that has a limited height, maybe with a styled scrollbar - run "reorder-group context test" to see a use case).
Now, you can just add: context="{divID}", which does all the work for you!
You are welcome to change the property name to a better one :)

- [x] No

## Other information

An example of a use case:
On the left, there is a 'div' that contains an 'IonReorderGroup'.
Because the list is long, the 'div' has a styled scrollbar.
However, when I dragged an item, there was no autoScroll, so there were glitches.
Now, with the 'context' property, this problem can be easily solved.
image


liamdebeasi:
Thanks for the PR! We appreciate all of the hard work you put into creating this. Unfortunately, we are unable to accept this PR. Please see https://github.com/ionic-team/ionic-framework/issues/20218#issuecomment-1032901279 for recommended alternatives and plans for future work. Thanks!

PR #20273, opened 2020-01-23
feat(react): add Ionic Animations wrapper (experimental)

liamdebeasi:
## Pull request checklist

## Pull request type

 

 

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes

React users would need to use the raw animations API to use Ionic Animations.

## What is the new behavior?

- Added an experimental wrapper component. Does not currently support everything (I.e. adding child animations, etc).
- Users can still access the raw animation object for more advanced features.

- [ ] Yes
- [x] No

### DIFF
--- a/packages/react/src/components/CreateAnimation.tsx
+++ b/packages/react/src/components/CreateAnimation.tsx
@@ -53,8 +53,6 @@ export interface CreateAnimationProps {
  /**
   * QUESTIONS
   * How do I update props on the fly? (i.e. toggle play/pause)
-  * Is there a better way to setup the animation
-  * other than individually checking each prop?
   * Should the before/after hooks accept arrays of
   * callback in the core method?
   * How do I accept multiples elements even if they aren't immediately nested?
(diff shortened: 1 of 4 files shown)


packages/react/src/components/CreateAnimation.tsx
```diff
@@ -0,0 +1,159 @@
...
+  progressStart?: { forceLinearEasing: boolean, step?: number };
+  progressStep?: { step: number };
+  progressEnd?: { playTo: 0 | 1 | undefined, step: number, dur?: number };
+}
+
+export class CreateAnimation extends React.Component<CreateAnimationProps> {
+  private nodes: Map<number, HTMLElement> = new Map();
+  animation?: Animation;
+
+  updateAnimation(props: any) {
```
elylucas: these seems to only get ran once on mount but having the name updateAnimation makes it feel like it gets updated each render, consider a rename, maybe like setupAnimation.

packages/react/src/components/CreateAnimation.tsx
```diff
@@ -0,0 +1,159 @@
...
+    const animation = this.animation;
+    if (animation === undefined) { return; }
+
+    if (this.nodes.size > 0) {
```
elylucas: this method of adding the child node to the animation probably won't work if someone ever adds children to the animation dynamically. Might be a future change as thats probably an edge case, but sooner or later someone will file a bug :)

PR #20278, opened 2020-01-23
fix(components): fix issues with button states

brandyscarney:
- [x] Update action sheet colors to match spec (md)
- [x] Remove activated background on buttons (md - uses ripple)
- [x] Update Segment Button to use proper indicator color when inside of a toolbar
- [x] Add back input highlight on focus

fixes https://github.com/ionic-team/ionic/issues/20276


brandyscarney:
Ready for review

brandyscarney:
These changes break custom, without having background steps:

Screen Shot 2020-01-24 at 5 15 24 PM

PR #20314, opened 2020-01-27
fix(ios): translucent toolbar blur no longer obscures entering page title

liamdebeasi:
## Pull request checklist

## Pull request type

 

 

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

Issue Number: https://github.com/ionic-team/ionic/issues/19158

## What is the new behavior?

- Added `'.header-background` where the backdrop filter will be applied to.
- Updated animation to translate the toolbar bg instead of fade out when the header is translucent

- [ ] Yes
- [x] No

### DIFF
--- a/core/src/components/header/header.ios.scss
+++ b/core/src/components/header/header.ios.scss
@@ -9,10 +9,6 @@
 }
 
 @supports (backdrop-filter: blur(0)) {
-  .header-translucent-ios {
-    backdrop-filter: $header-ios-translucent-filter;
-  }
-  
   .header-translucent-ios ion-toolbar {
     --opacity: .8;
     --backdrop-filter: #{$header-ios-translucent-filter};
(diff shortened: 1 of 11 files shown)


core/src/components/header/header.ios.scss
```diff
@@ -9,13 +9,15 @@
 }
 
 @supports (backdrop-filter: blur(0)) {
-  .header-translucent-ios {
+  .header-background {
```
brandyscarney: Wouldn't this apply to MD too?

liamdebeasi: It's in the header.ios.scss file, so it should be fine, right? Or do I still need `.header-translucent-ios .header-background`?

brandyscarney: I thought that scoped used to do that, but I'm seeing it work fine. I wonder why we specify the class + mode in all of the CSS files though instead of just using `.header-translucent` or `ion-header` then.

PR #20332, opened 2020-01-29
feat(calendar): Adding calendar component

corysmc:
Here's a first pass at the ion-calendar... Still has a ways to go, but wanted to get an initial code review before continuing - to fix bugs/issues that are likely present :) 

- [X] Swipe R/L To go to next/previous month
- [X] Props to pass in selectedDates, disabledDates, disabledBeforeDate, disabledAfterDate, disabledDays (weekdays), availableDates
- [ ] Remove date-fns dependency
- [ ] Fix flash on older devices - when swiping between months, there's an awkward flash as the animation ends and the current month is re-rendered.
- [ ] Add internationalization options - pass in month names, and weekday names
- [ ] RTL support
- [ ] Add a vertical scrolling calendar option
- [ ] accessibility: hover states, focus states
- [ ] add css variables for color, padding, etc.

**Related issues**
- Ionic Components - Desktop support https://github.com/ionic-team/ionic/issues/15500


core/src/components/calendar/calendar.tsx
```diff
@@ -0,0 +1,555 @@
...
+    md: 'calendar.md.scss'
+  }
+})
+export class Calendar implements ComponentInterface {
+  private monthsWrapper!: HTMLElement;
+  private today: Date = new Date();
+  private gesture?: Gesture;
+  private dragging = false;
+
+  @Element()
```
brandyscarney: Can you make sure all of these are on one line please? We follow this guide: https://stenciljs.com/docs/style-guide

core/src/components/calendar/calendar.tsx
```diff
@@ -0,0 +1,555 @@
...
+                      {this.getMonthTitle(month)}
+                      <ion-icon
+                        size="small"
+                        name={this.showYears ? 'caret-up' : 'caret-down'}
+                        slot="end"
+                      />
+                    </ion-button>
+                  </div>
+                  <div />
+                  <ion-button
```
brandyscarney: Can this be a native button? I can help with this. We usually don't use `ion-button` when building components because it pulls in all of the button CSS to this component. See `ion-menu-button`: https://github.com/ionic-team/ionic/blob/master/core/src/components/menu-button/menu-button.tsx#L76-L105

AndreasGassmann:
Great to see someone is working on this, that will be very helpful :).

I've briefly looked at the code without testing it, but it looks like ranges are not supported, right? Is this something that is planned? I really like the kiwi.com date picker on desktop, but it also works well on mobile (although the feedback/snappiness could be better there).

Yohandah:
I don't know how this is being implemented for the iOS version but with the recent release of iOS 14 there is an official "calendar component", maybe it should match its style ? 

https://developer.apple.com/design/human-interface-guidelines/ios/controls/pickers/

image

There's also a "popover" mode like Material 

image

liamdebeasi:
Thanks for the PR, and apologies for the delay! We really appreciate all of the effort you put into building this.

In Ionic Framework v6 we will be launching a brand new `ion-datetime` component which will feature a calendar picker. As a result, we are unable to accept this PR. We definitely encourage you to give the new `ion-datetime` component a try and let us know if you run into any bugs or think of any features that are missing.

For more information on Framework v6, please see: https://ionicframework.com/blog/announcing-the-ionic-framework-v6-beta/

PR #20454, opened 2020-02-12
fix(angular): avoid initializing components multiple times

liamdebeasi:
## Pull request checklist

## Pull request type

 

 

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

Issue Number: https://github.com/ionic-team/ionic/issues/20356

## What is the new behavior?

This particular code was added to account for the scroll position not being restored (https://github.com/ionic-team/ionic/issues/14737). It looks like this is no longer an issue in ng8+, so we can safely remove this. ng9 introduced a breaking change in how to determines where to insert DOM nodes, which is causing our code to move DOM elements around more than it needs to.

"Embedded views (such as ones created by *ngFor) are now inserted in front of anchor DOM comment node (e.g. ) rather than behind it as was the case previously. In most cases this does not have any impact on rendered DOM. In some cases (such as animations delaying the removal of an embedded view) any new embedded views will be inserted after the embedded view being animated away. This difference only last while the animation is active, and might alter the visual appearance of the animation. Once the animation is finished the resulting rendered DOM is identical to that rendered with View Engine."

https://angular.io/guide/ivy-compatibility

- [ ] Yes
- [x] No

### DIFF
--- a/angular/src/directives/navigation/stack-controller.ts
+++ b/angular/src/directives/navigation/stack-controller.ts
@@ -240,9 +240,6 @@ export class StackController {
     if (enteringEl && enteringEl !== leavingEl) {
       enteringEl.classList.add('ion-page');
       enteringEl.classList.add('ion-page-invisible');
-      if (enteringEl.parentElement !== containerEl) {
-        containerEl.appendChild(enteringEl);
-      }
 
       if ((containerEl as any).commit) {
         return containerEl.commit(enteringEl, leavingEl, {


liamdebeasi:
I have tested v5 using that old scroll position reproduction and it looks like the scroll position is still be preserved properly.

liamdebeasi:
Not the right approach. Closing for now until I can discuss with the team.

PR #20495, opened 2020-02-14
fix(segment): only emit ionChange when user releases pointer from screen

liamdebeasi:
## Pull request checklist

## Pull request type

 

 

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

Issue Number: https://github.com/ionic-team/ionic/issues/20257 https://github.com/ionic-team/ionic/issues/20500

## What is the new behavior?

- `ionChange` is now emitted when the user removes their pointer from screen (And if the value has changed)
- This moves segment back in line with the iOS spec, so it is being considered a bug instead of a breaking change.

- [ ] Yes
- [x] No

### DIFF
--- a/angular/src/directives/control-value-accessors/value-accessor.ts
+++ b/angular/src/directives/control-value-accessors/value-accessor.ts
@@ -12,7 +12,7 @@ export class ValueAccessor implements ControlValueAccessor {
   constructor(protected el: ElementRef) {}
 
   writeValue(value: any) {
-    this.el.nativeElement.value = this.lastValue = value == null ? '' : value;
+    this.el.nativeElement.value = this.lastValue = value;
     setIonicClasses(this.el);
   }
 
--- a/angular/src/directives/control-value-accessors/value-accessor.ts
+++ b/angular/src/directives/control-value-accessors/value-accessor.ts
@@ -12,7 +12,14 @@ export class ValueAccessor implements ControlValueAccessor {
   constructor(protected el: ElementRef) {}
 
   writeValue(value: any) {
-    this.el.nativeElement.value = this.lastValue = value;
+    /**
+     * TODO for Ionic 6:
+     * Change `value == null ? '' : value;`
+     * to `value`. This was a fix for IE9, but IE9
+     * is no longer supported; however, this change
+     * is potentially a breaking change
+     */
+    this.el.nativeElement.value = this.lastValue = value == null ? '' : value;
     setIonicClasses(this.el);
   }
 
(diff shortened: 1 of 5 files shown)


core/src/components/segment/segment.tsx
```diff
@@ -53,15 +57,27 @@ export class Segment implements ComponentInterface {
   @Watch('value')
   protected valueChanged(value: string | undefined) {
     if (this.didInit) {
-      this.ionChange.emit({ value });
+      this.ionSelect.emit({ value });
+      if (!this.pointerDown) {
```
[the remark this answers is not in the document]
liamdebeasi: Thanks!

angular/src/directives/control-value-accessors/value-accessor.ts
```diff
@@ -12,6 +12,13 @@ export class ValueAccessor implements ControlValueAccessor {
   constructor(protected el: ElementRef) {}
 
   writeValue(value: any) {
+    /**
+     * TODO for Ionic 6:
+     * Change `value == null ? '' : value;`
+     * to `value`. This was a fix for IE9, but IE9
+     * is no longer supported; however, this change
+     * is potentially a breaking change
+     */
```
juanmiguelbesada: It is possible to trigger a deprecation warning if value == null? (console.warn)

core/src/components/segment/segment.tsx
```diff
@@ -51,17 +54,29 @@ export class Segment implements ComponentInterface {
   @Prop({ mutable: true }) value?: string | null;
 
   @Watch('value')
-  protected valueChanged(value: string | undefined) {
-    if (this.didInit) {
-      this.ionChange.emit({ value });
+  protected valueChanged(value: string | undefined, oldValue: string | undefined | null) {
+    this.ionSelect.emit({ value });
+    if (oldValue !== '' || this.didInit) {
```
juanmiguelbesada: Why is needed to check for oldValue?

liamdebeasi: When binding values in Angular using `ngModel`, `ion-segment` value is initialized to `null` and then to the `ngModel` value. We had an old fix in place, but it was causing issues in https://github.com/ionic-team/ionic/issues/20500.

angular/src/directives/control-value-accessors/value-accessor.ts
liamdebeasi: There isn't anything to deprecate here as we are just updating the default value for `null`. In Ionic 6 the `null` default would just be changing from `''` to `null`. We will make note of it in the breaking changes document.

Issue #17248, opened 2019-01-24
feat: Support custom icons for `ion-select`


cicixiaoyan:
### Prerequisites

- [X] I have read the [Contributing Guidelines](https://github.com/ionic-team/ionic-framework/blob/main/.github/CONTRIBUTING.md#creating-an-issue).
- [X] I agree to follow the [Code of Conduct](https://ionicframework.com/code-of-conduct).
- [X] I have searched for [existing issues](https://github.com/ionic-team/ionic-framework/issues) that already report this problem, without success.

### Ionic Framework Version

- [x] v4.x
- [x] v5.x
- [X] v6.x
- [ ] Nightly

### Current Behavior

Developers are unable to customize the icon that is displayed with `ion-select`. Developers can change the color of the icon by using CSS:

```css
ion-select::part(icon) {
  color: red;
}
```

But this does not allow the flexibility for developers to customize the icon used.

Ideally, developers would like a property or slot that can be used to customize the icon that is used.

```ts
@Prop() toggleIcon = chevronDown;
```

```html
<ion-select [toggleIcon]="arrowDown"></ion-select>
```

This icon could then be styled using CSS shadow parts, as the existing implementation is today.

FazioNico:
looking forward for

fer-fullstackdev:
I need this feature right now. Please let me know when can you do this??
Thanks

dylan-westbury:
To explain further. 

Currently ion-select icon sits in the shadow-root, so there is no way of disabling or editing it. 

With our use case, we use the detail on ion-items, and floating labels for ion-inputs, ion-selects, etc.

For ion-select, it looks like this because we can't disable the select-icon.

![Image of select-icon](https://i.imgur.com/pWHcJUj.png)

Here it is in the html of the above ion-select

![html of select-icon](https://i.imgur.com/VoYbqCA.png)

fer-fullstackdev:
Hello, @dylan-westbury 
Thanks for the explanation. 
When can you implement this feature then?

gaiapuffo:
Sorry but you create a framework where you can't change the color of select, to change the width of the popup is a disaster, default you can write 8 characters for the options so you have to change css and you have to make this operation in every application! This is a complete disaster, if you put out a framework, ok a some bugs but not that you can make the base and all personalization are a disaster...and there are 1000 bugs...You can make a framework with bugs ok, but not in this way the are too bugs

root14:
[...]

PR #20605, opened 2020-02-24
fix(select): add icon-inner & placeholder part

brandyscarney:
## Pull request checklist

## Pull request type

 

 

- [x] Bugfix
- [x] Feature
- [ ] Build related changes

This can be considered a feature or a fix, due to the fact it adds a new part, but we haven't documented any of these yet so I'm not sure which way to lean here.

You cannot change the opacity of the select icon or target the placeholder color with the current parts.

Placeholder - Due to the fact that the placeholder and value share the same div, you would have to be able to target it based on the `select-placeholder` class which does not work with parts. 

Select Icon - The opacity was set on the inner icon, so you can change the color using `::part(icon)` but it will still have the default `.33` opacity. I added a part for `icon-inner` since that is where the actual arrow is created, and also moved opacity so it is set on the parent wrapper which fixes this problem. 

Issue Number: #17248 

## What is the new behavior?

- adds `icon-inner` and `placeholder` parts
- adds custom test for styling the select parts
- remove opacity in icon-inner and set it on the icon wrapper
- align items center (vertically) - this makes alignment look nicer especially when font size changes

- [ ] Yes
- [x] No

### DIFF
--- a/core/src/components/select/select.scss
+++ b/core/src/components/select/select.scss
@@ -21,6 +21,8 @@
   display: flex;
   position: relative;
 
+  align-items: center;
+
   font-family: $font-family-base;
 
   overflow: hidden;
--- a/core/src/components/select/select.scss
+++ b/core/src/components/select/select.scss
@@ -87,7 +87,5 @@ button {
 
   color: currentColor;
 
-  opacity: inherit;
-
   pointer-events: none;
 }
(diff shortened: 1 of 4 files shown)


brandyscarney:
@liamdebeasi Let me know your thoughts on this implementation

PR #20729, opened 2020-03-09
feat(toast): add white-space variable for toast message

liamdebeasi:
## Pull request checklist

## Pull request type

 

 

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

Issue Number: https://github.com/ionic-team/ionic/issues/20727

## What is the new behavior?

- Changed the wrapping for the toast message from `pre-wrap` to `pre-line`.

- [ ] Yes
- [x] No

### DIFF
--- a/core/api.txt
+++ b/core/api.txt
@@ -1237,6 +1237,7 @@ ion-toast,css-prop,--max-width
 ion-toast,css-prop,--min-height
 ion-toast,css-prop,--min-width
 ion-toast,css-prop,--start
+ion-toast,css-prop,--white-space
 ion-toast,css-prop,--width
 
 ion-toggle,shadow
--- a/core/src/components/toast/readme.md
+++ b/core/src/components/toast/readme.md
@@ -250,6 +250,7 @@ Type: `Promise<void>`
 | `--min-height`    | Minimum height of the toast                                                                            |
 | `--min-width`     | Minimum width of the toast                                                                             |
 | `--start`         | Position from the left if direction is left-to-right, and from the right if direction is right-to-left |
+| `--white-space`   | How white space inside the toast message should be handled                                             |
 | `--width`         | Width of the toast                                                                                     |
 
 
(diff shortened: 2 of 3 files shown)


brandyscarney:
Why not use `"normal"`?

liamdebeasi:
We can use `normal`, but `pre-line` is the closest to `pre-wrap` in terms of behavior (minus collapsing whitespace) so I was concerned about introducing unexpected breaking changes by using `normal`.

RafaelKr:
Would it be an option to make it customizable defaulting to `pre-wrap` for now and switching to `normal` when 5.1.0 is released?

brandyscarney:
No matter what we change it to it may be an unexpected change to users. We're currently respecting all white-space, so take the following message:

```
message: `
  Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor <b>incididunt</b> ut labore et dolore magna aliqua. Ut enim ad minim veniam, quis nostrud exercitation <b>ullamco</b> laboris nisi ut aliquip ex ea commodo consequat.

`,
```

In `master`, using `pre-wrap`:

toast-pre-wrap

Changing to `pre-line`:

toast-pre-line

Changing to `normal`:

toast-normal

Judging by the Material Design specs I don't think we should be respecting the beginning or ending lines:

Screen Shot 2020-03-10 at 12 22 07 PM

They don't have any two-line examples but they do use `white-space: normal`: https://material-components.github.io/material-components-web-catalog/#/component/snackbar

RafaelKr:
Wouldn't it make more sense to change the default to `normal` when it goes into 5.1.0 anyways?

brandyscarney:
@RafaelKr Minor releases are for adding features. Fixing an issue to match the spec would be considered a bug fix and can go into a patch release. We only consider breaking changes things that remove from or change the public API, which would be a major release. This file is a general overview of our API: https://github.com/ionic-team/ionic/blob/master/core/api.txt

In this case we are going to be adding a CSS variable to the public API so it can be changed but leaving it as it is for now.

RafaelKr:
@brandyscarney Oh right, it could break things for others. Then I'm fine with the current way how to do it :)
Maybe something to put onto the v6 Milestone. Although it's probably still a long time until v6. 
Do you have a release cycle or are things ready when they're ready?

brandyscarney:
@RafaelKr Our normal release cycle is every 2 weeks at a minimum for feature (minor) releases. Patch releases we were releasing on the same cycle, but with Ionic 5 being such a big release we have been trying to get them out faster to get fixes in. At the moment we don't have big features planned so our main focus has been on fixes. This document is a little outdated since we were doing minor releases sooner: https://ionicframework.com/docs/intro/versioning

PR #20790, opened 2020-03-16
fix(content): apply --offset-top and --offset-bottom values correctly

liamdebeasi:
## Pull request checklist

## Pull request type

 

 

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

Issue Number: https://github.com/ionic-team/ionic/issues/20735

## What is the new behavior?

- Added a MutationObserver to resize content

In some cases, `ion-content` is rendered/hydrated before the root document element is hydrated. Because of https://github.com/ionic-team/ionic/blob/c15ae7cbf58065086816ca352c0a5ce981bb25b5/core/src/css/structure.scss#L25-L27 the root document element will have `display: none`, causing the `ion-content` to be hidden. This also means its `offsetParent` will be `null` and the element will have dimensions of 0x0. This is causing the `--offset-top` and `--offset-bottom` variables to not be set with the correct values.

To remedy this, I have created a MutationObserver that watched for attribute changes on `document.documentElement` (usually `html`). The callback will recalculate the offset variable and then remove the observer if the root document element has the `hydrated` class applied.

In the event that `ion-content` is hydrated after the root document element, we still have the `this.resize()` call in `componentDidLoad`. If the root document element is already hydrated, the MutationObserver will not be created.

- [ ] Yes
- [x] No


core/src/components/content/content.scss
```diff
@@ -29,6 +29,7 @@
   --offset-top: 0px;
   --offset-bottom: 0px;
   --overflow: auto;
+  background: red;
```
brandyscarney: ooooof

liamdebeasi: LOL I thought changes weren't building so I added that 😭

PR #20899, opened 2020-03-27
fix(slides): slides no longer breaks when Angular Ivy enabled

liamdebeasi:
## Pull request checklist

## Pull request type

 

 

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

Issue Number: https://github.com/ionic-team/ionic/issues/20356

## What is the new behavior?

- Synchronously destroy swiper instances to avoid race conditions in multiple connected/disconnected callbacks

- [ ] Yes
- [x] No

## Other information

short breakdown of the issue for future reference:

1. Component is added as a sibling to ion-router-outlet.
2. connectedCallback is called. Swiper is initialized.
3. Ionic moves the component to be a child of ion-router-outlet.
4. disconnectedCallback is called. Work begins to tear down Swiper.
5. connectedCallback is called. Swiper is initialized.

The problem lies in step 4. The code to destroy swiper is called after an asynchronous call to get the Swiper instance. In other words, while disconnectedCallback is called before the second connectedCallback, the actual code to destroy the swiper instance is called after the second connectedCallback.

### DIFF
--- a/core/src/components/slides/slides.tsx
+++ b/core/src/components/slides/slides.tsx
@@ -24,6 +24,7 @@ export class Slides implements ComponentInterface {
   private readySwiper!: (swiper: SwiperInterface) => void;
   private swiper: Promise<SwiperInterface> = new Promise(resolve => { this.readySwiper = resolve; });
   private syncSwiper?: SwiperInterface;
+  private didInit = false;
 
   @Element() el!: HTMLIonSlidesElement;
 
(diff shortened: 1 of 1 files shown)


core/src/components/slides/slides.tsx
```diff
@@ -147,15 +148,26 @@ export class Slides implements ComponentInterface {
...
+
+    /**
+     * We need to synchronously destroy
+     * swiper otherwise it is possible
+     * that it will be left in a
+     * destroyed state if connectedCallback
+     * is called multiple times
+     */
+    const swiper = this.syncSwiper;
+    if (swiper !== undefined) {
```
liamdebeasi: One potential issue:

In the event that slides is connected and then disconnected before swiper is fully initialized, Swiper will not actually be destroyed here. Need to investigate more, but will garbage collection pick this up for us?

PR #21049, opened 2020-04-15
fix(searchbar): screen readers correctly announce the cancel button text

liamdebeasi:
## Pull request checklist

## Pull request type

 

 

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

Issue Number: https://github.com/ionic-team/ionic/issues/21013

## What is the new behavior?

- Added `aria-hidden` to the search outline icon since it is purely decorative
- Added `aria-hidden` to the div that contains the cancel icon/text. Since either text or an icon is displayed, I didn't think it made sense to add special logic to each one. I added an `aria-label` to the parent button that defaults to the value of `cancelButtonText`. There is no breaking change because `cancelButtonText` defaults to "Cancel", which is what the aria-label value was hardcoded to previously.

- [ ] Yes
- [x] No

### DIFF
--- a/core/src/components/searchbar/searchbar.tsx
+++ b/core/src/components/searchbar/searchbar.tsx
@@ -420,6 +420,7 @@ export class Searchbar implements ComponentInterface {
   }
 
   render() {
+    const { cancelButtonText } = this;
     const animated = this.animated && config.getBoolean('animated', true);
     const mode = getIonMode(this);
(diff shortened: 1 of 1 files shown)


terrancle:
Hello,

To be able to translate the aria-label = "Reset" button as well as the aria-label = "Search text" into several languages, could you add properties?
For example resetButtonText, as well as searchText?
In fact, I wish I could do like cancelButtonText which now works great for VoiceOver!

PR #21224, opened 2020-05-06
fix(modal): modal gets stuck when swiping too fast

julian-baumann:
## Pull request checklist

## Pull request type

 

 

- [X] Bugfix
- [ ] Feature
- [ ] Build related changes

Modal throws an error when swiping down too fast. This fix should prevent "newStepValue" from being "NaN" because the velocity or step is 1, greater than 1, 0 or less than 0. 

And also the modal gets stuck when swiping up very fast, so the step is becoming less than zero and it would just "return".

ezgif com-video-to-gif

## What is the new behavior?

Fixes the issues described above

- [ ] Yes
- [x] No


core/src/components/modal/test/swipe-to-close/e2e.ts
```diff
@@ -0,0 +1,11 @@
+import { testModal } from '../test.utils';
+
```
liamdebeasi, line 2: The `swipe-to-close` tests are not necessary. We have tests that do the same thing in the `spec` directory.

core/src/components/modal/gestures/swipe-to-close.ts
```diff
@@ -39,16 +39,18 @@ export const createSwipeToCloseGesture = (
...
   const onEnd = (detail: GestureDetail) => {
-    const velocity = detail.velocityY;
-    const step = detail.deltaY / height;
-    if (step < 0) { return; }
+    let velocity = detail.velocityY;
+    if (velocity <= 0) {
+      velocity = 0.0001; // Fixes modal stuck when swiping up from the bottom very fast
+    }
+
+    const step = getStep(detail.deltaY, height);
```
liamdebeasi, line 53: We can just use the `clamp` function here instead:
```suggestion
    const velocity = detail.velocityY;
    const step = clamp(0.0001, detail.deltaY / height, 0.9999);
```

This will ensure that the "ending" values are always in-bounds

core/src/components/modal/gestures/swipe-to-close.ts
```diff
@@ -98,3 +100,8 @@ export const createSwipeToCloseGesture = (
 const computeDuration = (remaining: number, velocity: number) => {
   return clamp(400, remaining / Math.abs(velocity * 1.1), 500);
 };
+
+const getStep = (deltaY: number, height: number) => {
+  const step = (height > 0) ? deltaY / height : 0.0001;
+  return Math.max(0.0001, Math.min(0.9999, step)); // Prevent step to be >= 1 or <= 0, cause this would throw an error or the modal would get stuck
+};
```
liamdebeasi, line 107:
```suggestion
```

No longer needed if we use the `clamp` function (see previous comments)

core/src/components/modal/gestures/swipe-to-close.ts
```diff
@@ -39,16 +39,18 @@ export const createSwipeToCloseGesture = (
   };
 
   const onMove = (detail: GestureDetail) => {
-    const step = detail.deltaY / height;
-    if (step < 0) { return; }
+    const step = getStep(detail.deltaY, height);
```
liamdebeasi, line 42: We can just use the `clamp` function here instead:


```suggestion
    const step = clamp(0.0001, detail.deltaY / height, 0.9999);
```

This will ensure that the "ending" values are always in-bounds

core/src/components/modal/gestures/swipe-to-close.ts
```diff
@@ -39,16 +39,18 @@ export const createSwipeToCloseGesture = (
...
   };
 
   const onEnd = (detail: GestureDetail) => {
-    const velocity = detail.velocityY;
-    const step = detail.deltaY / height;
-    if (step < 0) { return; }
+    let velocity = detail.velocityY;
+    if (velocity <= 0) {
+      velocity = 0.0001; // Fixes modal stuck when swiping up from the bottom very fast
+    }
```
liamdebeasi, line 51:
```suggestion
    const velocity = detail.velocityY;
```

I don't think we need to clamp the velocity. Swiping is working fine for me with just the `clamp`.

julian-baumann: You were right. I removed it and pushed a new version.

liamdebeasi:
Merged. Thank you!

Issue #20501, opened 2020-02-15
feat: Swipeable Modal add correct animation for iPhones with edge-to-edge display and fix statusbar color (bug)


hansPeter055:
# Feature Request

**Ionic version:**

[x] **5.x**

**Describe the Feature Request**

On the iPhone X or any iPhone with an edge-to-edge display, in ionic the new iOS 13 style modal presents in a kind of wrong animation. The edges should start transform from the iPhone X round edges.

How it is in Ionic:
![ezgif com-video-to-gif-4](https://user-images.githubusercontent.com/58850496/74591931-bbe6e880-501c-11ea-8bd0-77cd4bcce8bc.gif)

**Describe Preferred Solution**

How it should be:
![ezgif com-video-to-gif-5](https://user-images.githubusercontent.com/58850496/74591940-d5883000-501c-11ea-91ff-59ff8e9280db.gif)

It would be nice if anyone could add this animation when on an iPhone with an edge-to-edge display to start transforming the border-radius of the presenting element with the correct iPhone edge radius. And it should also change the statusbar color from black to white (in light mode).

Also in dark mode the background of the presenting element should transition to a dark gray.

HansKrywaa:
+1

liamdebeasi:
Thanks, I agree that this should be automatic. As a temporary workaround, you can toggle the status bar color yourself when you open/close a modal.

liamdebeasi:
Changing this to a feature request and this is going to require an update to Capacitor. For now, developers can manually toggle the status bar styles themselves.

JayJay666:
I have a same issue, and annoying me.

Ionic 5 introduced Swipe to Close Modal on this link and on video toggling working perfect: https://ionicframework.com/blog/announcing-ionic-5/ but now its not working and status bar must be change manually.

Would be handy if can set the automatic change in modal property config or disable automatic and toggle maually when I want.

liamdebeasi:
> I have a same issue, and annoying me.
> 
> Ionic 5 introduced Swipe to Close Modal on this link and on video toggling working perfect: https://ionicframework.com/blog/announcing-ionic-5/ but now its not working and status bar must be change manually.
> 
> Would be handy if can set the automatic change in modal property config or disable automatic and toggle maually when I want.

The code that is running in the video changed the status bar manually as well. Capacitor v3 gives us the ability to do this via the updated Status Bar plugin: https://capacitorjs.com/docs/v3/apis/status-bar#style.

[...]

PR #21225, opened 2020-05-06
feat(modal): add initial border radius

julian-baumann:
## Pull request checklist

## Pull request type

 

 

- [ ] Bugfix
- [X] Feature
- [ ] Build related changes

On iPhone X and up there is the "wrong" border transition.

See [this issue #20501]("https://github.com/ionic-team/ionic/issues/20501")

## What is the new behavior?

Now checks if the bottom safe area isn't 0px which would mean there is a home-indicator and therefore it is an iPhone with edge to edge display and then adds the desired border-radius.

- [ ] Yes
- [X] No


liamdebeasi:
Thanks for the PR! What issue does this address? It seems to be different than the one linked in your PR description.

julian-baumann:
Sorry, maybe I gave a bit to lees information.

This is the current behavior:
gif

This is what this PR fixes:
gif

It just adds the right border radius and doesn't fix the statusbar color "bug"

core/src/components/modal/test/swipe/e2e.ts
```diff
@@ -0,0 +1,11 @@
+import { testModal } from '../test.utils';
```
liamdebeasi, line 1: These tests are redundant - we can just use the `spec` tests instead.

core/src/components/modal/animations/ios.enter.ts
```diff
@@ -52,14 +52,23 @@ export const iosEnterAnimation = (
       const toPresentingScale = SwipeToCloseDefaults.MIN_PRESENTING_SCALE;
       const finalTransform = `translateY(${modalTransform}) scale(${toPresentingScale})`;
 
+      let initialBorderRadius = '0px';
+      const isIPhone = /iPhone/.test(navigator.userAgent);
+      const safeArea = getComputedStyle(document.documentElement).getPropertyValue('--ion-safe-area-bottom');
```
liamdebeasi, line 57: I would like to stay away from using`getComputedStyle` in this context. It will likely cause jank on lower end iPhones. Is there any way we can do this without forcing a reflow like this?

julian-baumann: You are right, that was not a very good approach. I pushed a new version where it is now calculation the screen ratio. It seems every iPhone today with an edge to edge display has a screen ratio about 0.46 (px). I put this value in an array so that if any new iPhone released with a different ratio, we can simply add it.

liamdebeasi:
Thanks for the PR! We appreciate all of the hard work you put into doing this. After discussing with the team, we have decided not to merge this PR.

While I agree that this PR brings the card modal closer to the native iOS modal spec, it brings tradeoffs that do not make the code easily maintainable.

The ratio value used here is hardcoded, but the ratio on new iPhones is subject to change at any time. Apple could release a new phone with a smaller notch and this could potentially break (Allegedly this may happen https://www.theverge.com/2020/4/13/21218806/apple-iphone-12-2020-redesign-notch-cameras-specs). I realize we could just update this array with whatever the new ratio is, but I do not think it is a good idea to add code that we will need to potentially race to update any time a new phone comes out.

I spent some time investigating other ways of accomplishing this border radius check, and I am not sure there is a good way of making this work currently. We try to make our components as close to the native spec as possible, so if this feature becomes easily doable in the future we will be happy to revisit this. Thank you!

PR #21339, opened 2020-05-19
fix(transition): ios transition works on devices without IO support

### LINKED ISSUE #21221: bug: gracefully fail if IntersectionObserver is not available
# Bug Report

**Ionic version:**
[x] **5.1**

**Current behavior:**
When using collapsible headers on iOS devices running iOS versions <12.2 without `IntersectionObserver` support an error is thrown:

`ReferenceError: Can't find variable: IntersectionObserver`

Which I also leads to another error while trying to animate the large title during page transition, which prevents the app from working correctly:

`TypeError: null is not an object (evaluating 'clonedTitleEl.innerText = largeTitleEl.innerText')`

**Expected behavior:**
As there is no reasonable polyfill available for the `IntersectionObserver` API the exception should at least be handled so it just fails silently and also the animation should make sure `clonedTitleEl` and `largeTitleEl` exist before accessing properties on them.

**Steps to reproduce:**
Run any app using collapsible headers on a device without `IntersectionObserver` support.

**Related code:**
https://github.com/ionic-team/ionic/blob/794c3d4e9608968d8d5ef8b77f9b024fa0131572/core/src/components/header/header.tsx#L117

https://github.com/ionic-team/ionic/blob/8a02b28efeca81c25176ff52508b4005441e8314/core/src/utils/transition/ios.transition.ts#L182

DavidStrausz:
## Pull request checklist

## Pull request type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

The following error is thrown on every page transition if using collapsible headers on a device without IntersectionObserver support:

`TypeError: null is not an object (evaluating 'clonedTitleEl.innerText = largeTitleEl.innerText')`

Issue Number: fixes #21221 
## What is the new behavior?
The part of the the page transition where the cloned element would be needed is skipped in case the IO API is not available.

- [ ] Yes
- [x] No


DavidStrausz:
@liamdebeasi Hm I'm not sure we're 100% talking about the same thing here. I agree it's useless to setup the collapse animation without IO support. The root problem is that the animation logic is completly separated from the component (in the iOS transition), and it is accessing the cloned title and back button elements for example:

here:
https://github.com/ionic-team/ionic/blob/6b159a0e913169ded8e829ee37668bb0e694ce93/core/src/utils/transition/ios.transition.ts#L188

or here:
https://github.com/ionic-team/ionic/blob/6b159a0e913169ded8e829ee37668bb0e694ce93/core/src/utils/transition/ios.transition.ts#L131

These are only created if the collapsible header is setup - but as it is not setup anymore (this was fixed by you in #21222) this leads to a TypeError.

At first I wanted to skip creating the large title animation at all, but the return values are used further down the pipe to determine if we're animating forward or backward:

https://github.com/ionic-team/ionic/blob/6b159a0e913169ded8e829ee37668bb0e694ce93/core/src/utils/transition/ios.transition.ts#L303

https://github.com/ionic-team/ionic/blob/6b159a0e913169ded8e829ee37668bb0e694ce93/core/src/utils/transition/ios.transition.ts#L524

So yeah, I it's not the prettiest fix, but it does the job.

liamdebeasi:
Oh hmm... Maybe it should still create the cloned title regardless of IO support: https://github.com/ionic-team/ionic/blob/master/core/src/components/header/header.tsx#L129-L132.

The cloned elements are really more for the large title transition than it is for the collapsable header. If we did that, then the check in the iOS transition file should not be necessary.

DavidStrausz:
@liamdebeasi Yeah sure, that should work as well, I'll move the cloning parts before the early exit and revert the check in the transition tomorrow as soon as I'm back at my dev machine! Or feel free to close my PR if you want to do it yourself :)

liamdebeasi:
I'll stick with this PR -- will take another look when the changes are made, but I think we should be good to merge after.

DavidStrausz:
Great! Thanks for the feedback so far, I‘ll let you know when I‘m done!

DavidStrausz:
@liamdebeasi This is ready for review again :)

liamdebeasi:
Merged. Thank you!

Issue #21290, opened 2020-05-13
feat: Support for HTML5 features for Alert Inputs


cerkiner:
# Feature Request

**Ionic version:**
[x] **5.x**

**Describe the Feature Request**
It is necessary to add special attributes such as maxLength, inputmode, pattern, id etc. on inputs while using AlertController.

**Describe Preferred Solution**

Have new options for inputs to set custom attributes while creating alerts;

```
alert.addInput({
      name: 'pin',
      placeholder: 'Pin Code',
      type: 'password',
      maxLength: 4,
      inputmode: 'numeric',
      ...
});
```

**Describe Alternatives**

```
alert.present().then(v => {
   let input = (document.querySelector('ion-alert input') as any);
   if (input) {
      input.setAttribute('maxlength','4');
      input.setAttribute('inputmode','numeric');
      ...
   }
});
```

However, this is just a hacky workaround and it gets complicated really quickly with multiple inputs. I think Ionic team should include a feature to add custom attributes and other native HTML5 features for inputs.

liamdebeasi:
Thanks for the issue. I am going to close this as a duplicate of https://github.com/ionic-team/ionic/issues/21202.

cerkiner:
Oh okay. Sorry for the duplicate. I was working on this issue/improvement for some time. I checked when I started but I didn't re-check when I created the issue.

Issue #7819, opened 2016-08-22
Need a new option for inputs at the alertController component


navid045:
#### Short description of the problem:

We can't add some custom attributes for inputs in alertController.
#### What behavior are you expecting?

I think it is necessary that we could add custom attributes like maxlength, id, style and etc to inputs in alertController.

I suggest we have a new option in inputs named "attr" that can set custom attributes for inputs. for example like this:

```
let alert = this.alertCtrl.create({
    title: 'Login',
    inputs: [
      {
        name: 'test',
        placeholder: 'Test name',
        attr: { maxlength: 6, id: "test", style:"color:red; font-size: 14px;" }
      }
    ],
    buttons: [
      {
        text: 'Cancel',
        role: 'cancel',
        handler: data => {
          console.log('Cancel clicked');
        }
      }
    ]
  });
  alert.present();
}
```

**Which Ionic Version?** 1.x or 2.x
2.x
#### Plunker that shows an example of your issue

**For Ionic 1 issues** - http://plnkr.co/edit/Xo1QyAUx35ny1Xf9ODHx?p=preview

**For Ionic 2 issues** - http://plnkr.co/edit/me3Uk0GKWVRhZWU0usad?p=preview

**Run `ionic info` from terminal/cmd prompt:** (paste output below)
Your system information:

Cordova CLI: 6.2.0
Gulp version:  CLI version 3.8.11
Gulp local:
Ionic CLI Version: 2.0.0-beta.31
Ionic App Lib Version: 2.0.0-beta.17
OS: Windows 8
Node Version: v4.4.0

jgw96:
Hello! We are moving our feature requests to a new feature request [doc](https://docs.google.com/document/d/1vZ339ibia2DA9JVSfO5eGSvej868vooex61xJ5Za8YM/edit). I have moved this feature request to the doc and because of this I will be closing this issue for now. Thanks for using Ionic!

jgw96:
Hello all! While this is an awesome feature request it is not something that we plan on doing anytime soon. Because of this I am going to move this to our internal feature tracking repo for now as it is just "collecting dust" here. Once we decide to implement this I will move it back. Thanks everyone for using Ionic!

jgw96:
This issue was moved to driftyco/ionic-feature-requests#54

nkaredia:
```
alertoptions = {
   title: 'sometitle',
   inputs: [{name: 'someName', id: 'someID'}]
}
```
```
alert.present().then(v => {
[...]

PR #21365, opened 2020-05-24
feat(alert): alert input custom attributes

### LINKED ISSUE #21202: feat: AlertController inputs should be having minlength, maxlength options for text fields
# Feature Request

**Ionic version:**
[x] **5.x**

As sometimes I need to show alert user and asking him for inputs alert `AlertInput` got options for min/max and seems those working for numeric to date fields, but what if I want to limit the length of the text that user provide, so I think maxlength and minlength should be added.

Just adding these to `AlertInput` and apply the change to the rendered input.

**Additional Context**
This is been requested in the forum and the solution provided with "hack" but seems is not working for some reason.
https://forum.ionicframework.com/t/ionic-unable-to-fix-maxlength-for-prompt-input-of-alert-controller/119472

cerkiner:
## Pull request checklist

## Pull request type

 

 

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes

No custom attributes can be added for the inputs in the alert controller. Maxlength, pattern, class etc. attributes can be greatly useful when added to the inputs in the alert controller. I think we should be able to add such features.

Issue Number: resolves #21202, #21290, #7819

## What is the new behavior?

This change introduces 'cssClass' and 'attributes' properties for inputs.

```
let alert = this.alertCtrl.create({
    inputs: [
      {
        ...
        cssClass: 'anyClass',
        attributes: { 
                inputmode: 'decimal',
                minlength: 1,
                maxlength: 4,
                autocomplete: 'off',
                ...
        }
        ...
      }
    ]
  });
  alert.present();
}
```

- [ ] Yes
- [x] No


cerkiner:
npm run build process in core directory upgraded slides/swiper/swiper.bundle.js. 
I have committed this change separately in https://github.com/ionic-team/ionic/pull/21365/commits/3e9919c719bb83e1d184627ba85328658dcdd28f

I am not sure if it belongs here.

cerkiner:
I rebased all to current master to resolve the issue. I removed that commit and updated the branch. Now, all checks have passed and the change looks file.

core/src/components/alert/alert.tsx
```diff
@@ -404,23 +406,22 @@ export class Alert implements ComponentInterface, OverlayInterface {
...
-                  }}
+                  class={inputClass(i)}
                 />
               </div>
             );
           } else {
             return (
               <div class="alert-input-wrapper">
                 <input
+                  {...i.attributes as AlertInputAttributes}
```
liamdebeasi, line 424: Let's move this line below `class={inputClass(i)}`. This will allow users to override attributes such as `value` and `id` either in the top level `AlertInput`, or in `attributes`.

For example:

```javascript
{
  attributes: {
    value: 'Hello'
  }
}
```

The above will not do anything since `value={i.value}` is set after `{...i.attributes as AlertInputAttributes}`.

liamdebeasi: Also be sure to do this for the textarea on as well: https://github.com/ionic-team/ionic/pull/21365/files#diff-14f90b10a83b2339d7a4078841868b3eR409

cerkiner: I made the change. Later while I was manually testing I had some issues with 3 attributes; onInput, class and disabled. I merged them all for an optimum solution, please and comment on the end result.

cerkiner:
The latest change lets the user override all attributes except onInput, class and disabled. The reason behind this is overriding these attributes can cause unexpected issues.

- In the case of overriding onInput, values are not synced properly. Fixing this was essential for passing value changes on dismiss.

- In the case of overriding class, input losses ionic designs. Fixing this was important to provide a better look.

- In the case of overriding disabled, dom disabled state and ionic disabled classes were mixed. Fixing this was important to reflect a better look and functionality.

Users are still able to set these attributes but they are not overridden completely. They are merged for an optimum solution, please check and comment on the end result.

core/src/components/alert/alert.tsx
cerkiner: By the way, changes are all in seperate commits, I can revert easily based on your decision

core/src/components/alert/test/translucent/index.html
```diff
@@ -158,6 +158,16 @@
...
+          },
+          {
+            name: 'name8',
+            type: 'password',
+            placeholder: 'Advanced Attributes',
+            cssClass: 'specialClass',
+            attributes: {
+              maxlength: 4,
+              inputmode: 'decimal'
+            }
```
liamdebeasi, line 170:
```suggestion
```

This is the same test as in `basic`, so I don't think we need this again.

core/src/components/alert/test/standalone/index.html
```diff
@@ -178,6 +178,16 @@
...
+          },
+          {
+            name: 'name8',
+            type: 'password',
+            placeholder: 'Advanced Attributes',
+            cssClass: 'specialClass',
+            attributes: {
+              maxlength: 4,
+              inputmode: 'decimal'
+            }
```
liamdebeasi, line 190:
```suggestion
```

This is the same test as in `basic`, so I don't think we need this again.

core/src/components/alert/alert.tsx
```diff
@@ -406,13 +408,14 @@ export class Alert implements ComponentInterface, OverlayInterface {
...
                   value={i.value}
-                  onInput={e => i.value = (e.target as any).value}
                   id={i.id}
-                  disabled={i.disabled}
                   tabIndex={0}
-                  class={{
-                    'alert-input': true,
-                    'alert-input-disabled': i.disabled || false
+                  {...i.attributes as AlertTextareaAttributes}
+                  disabled={i.disabled || (i.attributes && i.attributes.disabled)}
```
liamdebeasi, line 414: The `class` and `onInput` explanations make sense. I'm not sure about the `disabled` one though.

For example, if I did:

```javascript
{
  placeholder: 'Outer Placeholder',
  attributes: {
    placeholder: 'Inner Placeholder'
  }
}
```

`'Inner Placeholder'` would win in this case. But if I did:

```javascript
{
  disabled: true,
  attributes: {
    disabled: false
  }
}
```

`true` would win which is not what I would expect. Can we change it so that the `disabled` in `attributes` wins? This would bring it in line with the rest of the values.

Also, you are checking truthiness here which may lead to unexpected results. Even if we did:
```javascript
(i.attributes && i.attributes.disabled) || i.disabled
```

This would simplify down to: `false || true` which would always return `true`. We should probably do this instead:

```javascript
disabled={i.attributes && i.attributes.disabled !== undefined ? i.attributes.disabled : i.disabled}
```

cerkiner:
- Unnecessary tests are removed. (I copied the same input in all tests only because I didn't fully understand the purposes of those.)

- About disabled property, I got your point, I made a change to ensure higher specificity always wins.

```
{
  disabled: true,
  attributes: {
    disabled: false
  }
}
```
Results in false.

- I merged the latest master to the branch, resolved the conflicts.

core/src/components/alert/alert.tsx
```diff
@@ -532,6 +533,15 @@ export class Alert implements ComponentInterface, OverlayInterface {
   }
 }
 
+const inputClass = (input: AlertInput): CssClassMap => {
+  return {
+    'alert-input': true,
+    'alert-input-disabled': ((input.attributes && input.attributes.disabled !== undefined) ? input.attributes.disabled : input.disabled) || false,
```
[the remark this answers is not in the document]
cerkiner: done!! it makes sense, I felt sorry for not doing it in the previous commit.  😄

cerkiner:
resolved the conflicts for the latest master branch  👍

PR #21581, opened 2020-06-19
fix(virtual-scroll): fix top offset calculation

lydemann:
## Pull request checklist

## Pull request type

 

 

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

Virtual scroll will calculate the offset top by traversing the parents until it meets the ion content element.
This is a problem as `offsetTop` is not using the parent element but the `offsetParent` element thus possibly counting the same offsetParent multiple times and returning a wrong offset. That means, it will recount the same `offsetParent` multiple times for each non-positioned element nested between `ion-content` and `ion-list`.
Issue Number: N/A

## What is the new behavior?

- Calculation of offset is now done by traversing the offsetParents until it finds the ion content
- This ensures the correct offset top regardless of how many nested components there are in-between the ion-content

- [ ] Yes
- [x] No

### DIFF
--- a/core/src/components/virtual-scroll/virtual-scroll.tsx
+++ b/core/src/components/virtual-scroll/virtual-scroll.tsx
@@ -266,7 +266,7 @@ export class VirtualScroll implements ComponentInterface {
     let node: HTMLElement | null = el;
     while (node && node !== contentEl) {
       topOffset += node.offsetTop;
-      node = node.parentElement;
+      node = node.offsetParent as HTMLElement;
     }
     this.viewportOffset = topOffset;
     if (scrollEl) {


kensodemann:
@lydemann - do you have a simple reproduction repo we can use in order to get out your PR? Something simple created from one of our starters should be sufficient.

lydemann:
> @lydemann - do you have a simple reproduction repo we can use in order to get out your PR? Something simple created from one of our starters should be sufficient.

Sure, here you go:
https://github.com/lydemann/virtual-scroll-bug-demo

You can see how it is not rendering items for the whole ion-content height when within many nested divs.

liamdebeasi:
Merged. Thank you!

Issue #21346, opened 2020-05-20
bug: ion-fab overlaps iPhone status bar


LekisS:
# Bug Report

**Ionic version:**

**5.1.1**

**Current behavior:**

When I put an `ion-fab` with `slot="fixed"`, the FAB icons overlap the iOS status bar.

**Expected behavior:**

The FAB icons should be positioned lower to avoid overlapping the iOS status bar.

**Steps to reproduce:**

- Create a blank page with `ion-content`
- Put a `ion-fab` button directly in it with `slot="fixed"`
- Build on iOS 13.4.x (reproduced on iPhone 6S, iPhone X, iPhone 11)

**Related code:**

```
<ion-content has-bouncing="false">
  <ion-fab vertical="top" horizontal="start" slot="fixed">
    <ion-menu-button>
      <ion-fab-button color="secondary" size="small">
          <img src="../../assets/icons/menu.svg">
        </ion-fab-button>
    </ion-menu-button>
  </ion-fab>
  <ion-fab vertical="top" horizontal="end" slot="fixed">
    <ion-fab-button size="small">
      <img src="../../assets/icons/filter.svg">
    </ion-fab-button>
  </ion-fab>
  </ion-fab>
</ion-content>
```

**Other information:**

Screenshot on an iPhone 6S with iOS 13.4.1 :

![bug](https://user-images.githubusercontent.com/28733/82470553-d02bfd00-9ac5-11ea-8f2a-28318e11b236.png)

**Ionic info:**

```
Ionic:

   Ionic CLI                     : 6.9.2 (/Users/tom/.config/yarn/global/node_modules/@ionic/cli)
   Ionic Framework               : @ionic/angular 5.0.7
   @angular-devkit/build-angular : 0.803.26
   @angular-devkit/schematics    : 8.3.26
   @angular/cli                  : 8.3.26
   @ionic/angular-toolkit        : 2.2.0

Cordova:

   Cordova CLI       : 9.0.0 (cordova-lib@9.0.1)
   Cordova Platforms : android 8.1.0, browser 6.0.0, ios 5.1.1
   Cordova Plugins   : cordova-plugin-ionic-keyboard 2.2.0, cordova-plugin-ionic-webview 4.1.3, (and 6 other plugins)

Utility:

   cordova-res (update available: 0.14.0) : 0.13.1
   native-run                             : 1.0.0

System:

   Android SDK Tools : 26.1.1 (/Users/tom/Library/Android/sdk)
[...]

PR #21778, opened 2020-07-17
fix(fab-button): on iphone X and higher devices i.e, with notch the b…

pritamsoni-hsr:
…utton fix for #21346

## Pull request checklist

## Pull request type

 

 

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

Issue Number: #21346

## What is the new behavior?

- for the fab-button replaced fixed bottom position of 10px to adjust according on mobile devices

- [ ] Yes
- [x] No

### DIFF
--- a/core/src/components/fab/fab.scss
+++ b/core/src/components/fab/fab.scss
@@ -35,7 +35,7 @@
 // --------------------------------------------------
 
 :host(.fab-vertical-top) {
-  top: $fab-content-margin;
+  top: max(#{$fab-content-margin}, var(--ion-safe-area-top, 0px));
 }
 
 :host(.fab-vertical-top.fab-edge) {
@@ -44,7 +44,7 @@
 
 
 :host(.fab-vertical-bottom) {
-  bottom: $fab-content-margin;
+  bottom: max(#{$fab-content-margin}, var(--ion-safe-area-bottom, 0px));
 }
 
 :host(.fab-vertical-bottom.fab-edge) {


pritamsoni-hsr:
The problem appears on the recent full screen devices, so for devices which are still based on an older browser, this problem won't appear. 
One solution could be to keep the old style along with the new ones, and an old browser will invalidate the `max(...)` ones and the new ones will override the property with `max(...)`

liamdebeasi (commented):
Thanks for the PR! Unfortunately, this change would cause a regression when using a fixed `ion-fab` in a page that uses `ion-header`.

Given the following query:

```css
top: max(#{$fab-content-margin}, var(--ion-safe-area-top, 0px));
```

This would evaluate to:

```css
top: max(10px, 44px); // 44px is the top safe area value on my iPhone 11
```

This would cause `44px` to always win out, adding additional padding when using `ion-fab` on a page with `ion-header`. The same logic applies for `ion-fab` and `ion-footer`.

In terms of a CSS-only fix for top `ion-fab` elements, we could do something like:

**fab.scss**
```css
:host(.fab-vertical-top) {
  top: var(--ion-safe-area-top);
}
```

**global.scss**
```css
ion-header + ion-content ion-fab.fab-vertical-top {
  top: $fab-content-margin;
}
```

This would not work for `ion-footer`. Maybe @brandyscarney has ideas on how to properly solve this?

Prosayya (approved):

amandaesmith3:
Hello! I wanted to circle back to this. The team and I discussed this internally, and we've decided not to accept this PR. The issue this is meant to fix is still present in the latest version of Ionic, and we're absolutely interested in fixing it, but the approach needed would be different from the one taken here due to the regression Liam described above.

If this is still something you're interested in tackling, I encourage you to open a new PR that addresses these concerns. We appreciate your work!

Issue #21871, opened 2020-08-04
bug: scroll assist cancels first click event


iphilgood:
# Bug Report

**Ionic version:**
[ ] **4.x**
[x] **5.x**

**Current behavior:**
An `ion-item` in a `ion-list` needs two clicks/taps to trigger the click event. This only happens on iOS (Safari/WebKit).

**Expected behavior:**
The event should trigger on the first click.

**Steps to reproduce:**

1. Create an `ion-list`
1. Within the `ion-list` create an `ion-item` with a `(click)` event handler
1. Within the `ion-item` create an `ion-label`
1. Within the `ion-item` create an `ion-input`
1. Within the `ion-item` create an `ion-button`

**Related code:**

This is the markup with which you can reproduce the issue. You can also find a GIF at the bottom of this issue.

```html
<ion-list>
  <ion-item (click)="onTitleClicked()" button detail="false">
    <ion-label position="floating">
      Title
    </ion-label>
    <ion-input [value]="title" readonly></ion-input>
    <ion-button
      *ngIf="title"
      slot="end"
      fill="clear"
      size="small"
      (click)="resetTitle($event)"
    >
      <ion-icon name="close"></ion-icon>
    </ion-button>
  </ion-item>
</ion-list>
```

I tried to narrow it down to the following, but the issue still persists.

```html
<ion-list>
  <ion-item (click)="onTitleClicked()" button detail="false">
    <ion-label>
      Title
    </ion-label>
    <ion-input [value]="title"></ion-input>
  </ion-item>
</ion-list>
```

You can find a sample application on [my GitHub](https://github.com/iphilgood/ionic-ion-item-click-bug).

**Other information:**
I found out that this change of behaviour was introduced with version `v5.0.6`. I've checked the [changes between the versions `v5.0.5` and `v5.0.6`](https://github.com/ionic-team/ionic-framework/compare/v5.0.5...v5.0.6) but I didn't find any suspicious code 🤔 

![ionic-bug](https://user-images.githubusercontent.com/126632/89261377-3c7ebd00-d62e-11ea-9fd0-7b3a16316a45.gif)

**Ionic info:**

```
Ionic:

   Ionic CLI                     : 6.11.0 (/Users/phil/.asdf/installs/nodejs/12.17.0/.npm/lib/node_modules/@ionic/cli)
   Ionic Framework               : @ionic/angular 5.3.1
   @angular-devkit/build-angular : 0.901.12
[...]

PR #21875, opened 2020-08-04
fix(input): await async event registration on iOS

iphilgood:
## Pull request checklist

## Pull request type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

To trigger a `(click)` event on an `ion-item`, several clicks were necessary. For further information see the issue below.

Issue Number: #21871 

## What is the new behavior?

Now you only need to click once to trigger the `(click)` event on an `ion-item`. async methods are now awaited and the event listeners get registered properly.

- [ ] Yes
- [x] No

## Other information

Please have a look at issue #21871.

### DIFF
--- a/core/src/utils/input-shims/input-shims.ts
+++ b/core/src/utils/input-shims/input-shims.ts
@@ -11,7 +11,7 @@ const SCROLL_ASSIST = true;
 const SCROLL_PADDING = true;
 const HIDE_CARET = true;
 
-export const startInputShims = (config: Config) => {
+export const startInputShims = async (config: Config) => {
   const doc = document;
   const keyboardHeight = config.getNumber('keyboardHeight', 290);
   const scrollAssist = config.getBoolean('scrollAssist', true);
@@ -77,11 +77,11 @@ export const startInputShims = (config: Config) => {
   // At this point we need to look for all of the inputs not registered yet
   // and register them.
   for (const input of inputs) {
-    registerInput(input);
+    await registerInput(input);
   }
 
-  doc.addEventListener('ionInputDidLoad', ((ev: InputEvent) => {
-    registerInput(ev.detail);
+  doc.addEventListener('ionInputDidLoad', (async (ev: InputEvent) => {
+    await registerInput(ev.detail);
   }) as any);
 
   doc.addEventListener('ionInputDidUnload', ((ev: InputEvent) => {


core/src/utils/input-shims/input-shims.ts
```diff
@@ -77,11 +77,11 @@ export const startInputShims = (config: Config) => {
...
   // and register them.
   for (const input of inputs) {
-    registerInput(input);
+    await registerInput(input);
   }
 
-  doc.addEventListener('ionInputDidLoad', ((ev: InputEvent) => {
-    registerInput(ev.detail);
+  doc.addEventListener('ionInputDidLoad', (async (ev: InputEvent) => {
+    await registerInput(ev.detail);
```
liamdebeasi, line 84:
```suggestion
  doc.addEventListener('ionInputDidLoad', ((ev: InputEvent) => {
   registerInput(ev.detail);
```

Adding async/await here doesn't do anything since there isn't anything that gets run after `registerInput`.

iphilgood: @liamdebeasi I think I messed up 😄 I tried to run your suggestion but the issue still exists. Even with my suggestion from this PR. The funny thing is, when I'm removing the following lines, the click on the inputs work as expected:

```ts
doc.addEventListener('ionInputDidLoad', ((ev: InputEvent) => {
  registerInput(ev.detail);
}) as any);
```

Can you provide some additional information to this event registration? Is it really needed? This issue has a really big impact on our customers and I'm looking for a way to fix it.

liamdebeasi: Yes, we have an internal utility called "scroll assist" which tries to automatically scroll inputs into view when focused. This helps prevent inputs from being covered by the keyboard. In order to do so, we need to "register" each input.

There are other utilities that use this input registration, but scroll assist is the biggest one.

I tried your code reproduction in the GitHub issue, and I was only able to reproduce this in WebKit/Safari. This leads me to believe that the bug is in WebKit, not Ionic Framework. I need to spend some time isolating the issue, but we should be able to get to the bottom of things one way or another. From what I noticed, Safari does not even dispatch a click event, so adding async/await here would not resolve the issue anyways.

iphilgood: Thanks for your explanation. Okay, got it.

Yes, it only happens on iOS. As I tried to find out what the issue was, it all made sense because the input shim only gets loaded on iOS. Do you already have an idea, why this worked prior v5.0.6? 🤔 

If there's anything I can do, please let me know. I'm happy to help 👍

iphilgood:
Closing this PR because it doesn't really solve the underlying problem.

PR #21886, opened 2020-08-06
docs:(input): add available autocapitalize property options

lwpamihiranga:
## Pull request checklist

## Pull request type

 

 

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes

Issue Number: Fixes https://github.com/ionic-team/ionic-docs/issues/1511

## What is the new behavior?

-
-
-

- [ ] Yes
- [x] No


core/src/components/input/input.tsx
```diff
@@ -41,6 +41,8 @@ export class Input implements ComponentInterface {
 
   /**
    * Indicates whether and how the text value should be automatically capitalized as it is entered/edited by the user.
+   * Default options are: `"off"`, `"characters"`, `"words"`, and `"sentences"`.
+   * For more information, see [Autocapitalize for mobile](https://developers.google.com/web/updates/2015/04/autocapitalize)
    */
   @Prop() autocapitalize = 'off';
```
liamdebeasi, line 47: Rather than explicitly list the values in the documentation, we can make it a type here instead. The docs will automatically pull the values in, and users will also get added TypeScript checking too.

```suggestion
  @Prop() autocapitalize: AutocapitalizeTypes = 'off';
```

You would add the type definition here: https://github.com/ionic-team/ionic-framework/blob/master/core/src/interface.d.ts#L40

```
export type AutocapitalizeTypes = 'off' | 'none' | 'on' | 'sentences' | 'words' | 'characters';
```

liamdebeasi: For an example of how this looks, check out `inputmode` or `enterkeyhint` in the ion-input properties: https://ionicframework.com/docs/api/input#properties

core/src/components/input/input.tsx
```diff
@@ -41,6 +41,8 @@ export class Input implements ComponentInterface {
 
   /**
    * Indicates whether and how the text value should be automatically capitalized as it is entered/edited by the user.
+   * Default options are: `"off"`, `"characters"`, `"words"`, and `"sentences"`.
+   * For more information, see [Autocapitalize for mobile](https://developers.google.com/web/updates/2015/04/autocapitalize)
```
liamdebeasi, line 45: This change would not be necessary if we added a type

liamdebeasi (dismissed):

core/src/components/input/input.tsx
lwpamihiranga: @liamdebeasi Thanks for pointing that out

lwpamihiranga:
@liamdebeasi  I did add the type that you have mentioned in the review. But after that, it gives and build error.

```
[ ERROR ]  TypeScript: ./src/components.d.ts:2849:15
           Interface 'HTMLIonInputElement' cannot simultaneously extend types 'IonInput' and 'HTMLStencilElement'.Named
           property 'autocapitalize' of types 'IonInput' and 'HTMLStencilElement' are not identical.

   L2848:  };
   L2849:  interface HTMLIonInputElement extends Components.IonInput, HTMLStencilElement {
   L2850:  }
```

Is there any changes needed to be done in the Stencil package also? 🤔

brandyscarney:
Thank you for the PR! I did some further research into this and I don't think we'll be able to update this type due to it having a conflict to the global property of `HTMLElement`: https://github.com/microsoft/TypeScript/blob/ad96a52cc6451be4ee3b48078da875a9eb2faa49/src/lib/dom.generated.d.ts#L6596

```typescript
interface HTMLElement extends Element, DocumentAndElementEventHandlers, ElementCSSInlineStyle, ElementCSSInlineStyle, ElementContentEditable, GlobalEventHandlers, HTMLOrSVGElement {
    accessKey: string;
    readonly accessKeyLabel: string;
    autocapitalize: string;
    dir: string;
    draggable: boolean;
    ...
}
```

I think updating just the comment like your first commit for now should be sufficient. I can make these updates. 🙂

brandyscarney:
@liamdebeasi is this good to go now?

liamdebeasi (approved):

lwpamihiranga:
@brandyscarney @liamdebeasi Let's merge this 😊

liamdebeasi:
Merged. Thank you!

PR #21905, opened 2020-08-10
fix(fab-button) : Undo the contain:strict

### LINKED ISSUE #21772: bug: small ion-fab-button with a ion-fab-list has a wrong click event area
# Bug Report

**Ionic version:**

[x] **5.x**

**Current behavior:**

If a `ion-fab` has a `ion-fab-button size="small"` with a `ion-fab-list`, the area of the `ion-fab` will trigger the ion-fab-list even if the `ion-fab-button` isn't pressed. Which makes the click event trigger without actually pressing the `ion-fab-button`. I encountered this issue while binding a `(click)` event to the `ion-fab-button`

**Expected behavior:**

The `ion-fab-list` should be triggered only when the `ion-fab-button` is pressed, and not the `ion-fab` area itself.
Or the small `ion-fab-button` area should take the whole ion-fab area (so it's triggered when pressing the `ion-fab` area).

**Related code:**
- Press the red area of the first fab (normal size) : the `ion-fab-button` is triggered (ripple effect, click event, pointer is changing, ...), the `ion-fab-list` as well
- Press the red area of the second fab (small size) : the `ion-fab-list` is triggered but not the `ion-fab-button` (no ripple effect, no click event, pointer doesn't change, ...)

https://stackblitz.com/edit/ionic-v5-bug-fab-size

Gwe-n:
Fixes #21772

## Pull request checklist

## Pull request type

 

 

- [X] Bugfix
- [ ] Feature
- [ ] Build related changes

Issue Number: #21772 

## What is the new behavior?

The problem comes from the `contain:strict` which prevents `.button-native` from fitting with `ion-fab-button`. I only cancelled it in the `.fab-button-small` context. 

- [ ] Yes
- [X] No

### DIFF
--- a/core/src/components/fab-button/fab-button.scss
+++ b/core/src/components/fab-button/fab-button.scss
@@ -205,6 +205,12 @@
   height: #{$fab-small-size};
 }
 
+:host(.fab-button-small) .button-native {
+  position: static;
+
+  contain: none;
+}
+
 // FAB Close Icon
 // --------------------------------------------------
 


liamdebeasi:
Thanks for the PR and apologies for the delay. We appreciate the work you put into creating this.

After discussing with the team, we have decided not to move forward with this PR. Being able to click outside of the fab button to expand/collapse the fab list seems like a bug. We think that it would be better to move the click listener from `ion-fab` to `ion-fab-button` to avoid this problem entirely. `ion-fab-button` would then call a method on `ion-fab` to expand/collapse the fab list.

I am going to close this PR, however we do have plans to address this issue in a separate PR. Thank you!

PR #21956, opened 2020-08-22
add trigger part for checkbox and toggle

yacut:
I would like to hide trigger button for checkbox or toggle component if I need the controlled component behaviour:

```tsx
// index.tsx
import React from "react";
import { IonContent, IonPage, IonToggle, IonItem, IonLabel, IonItemDivider } from "@ionic/react";
import classes from "./index.module.css";

export class ToggleExamples extends React.Component {
    checked: boolean = false;

    handleItemClick = () => {
        setTimeout(() => this.checked = !this.checked, 2000);
    }

    public render() {
        return (
            <IonPage>
                <IonContent>
                    <IonItemDivider>Controlled Toggle</IonItemDivider>
                    <IonItem onClick={this.handleItemClick}>
                        <IonLabel>Checked: {JSON.stringify(this.checked)}</IonLabel>
                        <IonToggle checked={this.checked} className={classes.checkbox} />
                    </IonItem>
                </IonContent>
            </IonPage>
        );
    }
}
```

```css
// index.module.css
checkbox::part(trigger) {
  display: none;
}
```
## Pull request checklist

## Pull request type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes

- Click on item always switch the toggle/checkbox

## What is the new behavior?

- Click on item will only switch the toggle/checkbox when `checked` variable changed

- [ ] Yes
- [x] No

### DIFF
--- a/core/src/components/checkbox/checkbox.tsx
+++ b/core/src/components/checkbox/checkbox.tsx
@@ -173,6 +173,7 @@ export class Checkbox implements ComponentInterface {
           onBlur={this.onBlur}
           disabled={this.disabled}
           ref={btnEl => this.buttonEl = btnEl}
+          part="trigger"
         >
         </button>
       </Host>
--- a/core/src/components/checkbox/readme.md
+++ b/core/src/components/checkbox/readme.md
@@ -279,6 +279,7 @@ export class CheckboxExample {
 | ------------- | ------------------------------------------------- |
 | `"container"` | The container for the checkbox mark.              |
 | `"mark"`      | The checkmark used to indicate the checked state. |
+| `"trigger"`   | The trigger button of the checkbox.               |
 
 
 ## CSS Custom Properties
--- a/core/src/components/toggle/readme.md
+++ b/core/src/components/toggle/readme.md
@@ -446,10 +446,11 @@ export class ToggleExample {
 
 ## Shadow Parts
 
-| Part       | Description                                                   |
-| ---------- | ------------------------------------------------------------- |
-| `"handle"` | The toggle handle, or knob, used to change the checked state. |
-| `"track"`  | The background track of the toggle.                           |
+| Part        | Description                                                   |
+| ----------- | ------------------------------------------------------------- |
+| `"handle"`  | The toggle handle, or knob, used to change the checked state. |
+| `"track"`   | The background track of the toggle.                           |
+| `"trigger"` | The trigger button of the toggle.                             |
 
 
 ## CSS Custom Properties
--- a/core/src/components/toggle/toggle.tsx
+++ b/core/src/components/toggle/toggle.tsx
@@ -214,6 +214,7 @@ export class Toggle implements ComponentInterface {
           disabled={disabled}
           ref={btnEl => this.buttonEl = btnEl}
           aria-hidden="true"
+          part="trigger"
         >
         </button>
       </Host>


liamdebeasi:
Thanks for the PR. Can you please describe your use case for this feature?

yacut:
@liamdebeasi It's very simple. I want the `this.handleItemClick` operation to be done by clicking on the item, not on toggle, so that the user does not have to aim at the toggle trigger controller and just click on the item.

liamdebeasi:
Hmm I guess I am not sure as to how adding a CSS Shadow Part here would help with that? Are you trying to style some part of the checkbox/toggle?

yacut:
@liamdebeasi Could you take a closer look at the example I described above? Especially the css part...
```css
checkbox::part(trigger) {
  display: none;
}
```
If I hide the trigger, it will not respond to clicks. And this is exactly what I want to achieve.

liamdebeasi:
Just to confirm I understand the problem you are trying to solve: You want to prevent people from being able to click checkbox/toggle to trigger it without disabling the checkbox/toggle entirely?

liamdebeasi:
Ok thanks for helping me understand! I think there are easier ways to do this. For example, you can set `pointer-events: none` on `ion-toggle` and `ion-checkbox` which will prevent users from being able to click them without using the `disabled` prop.

That being said, I think doing this will lead to a bad user experience as users are going to expect that these components are interactive. If you need to prevent users from clicking these components, I strongly recommend using the `disabled` prop instead as it will give a visual indicator to your users that the components are not interactive.

Do either of these solutions work for you use case?

yacut:
@liamdebeasi And again, please take a closer look at the example I described above. I have controlled `ion-toggle` and onClick event on the `ion-item`. The value of the `ion-toggle` will be changed, when user clicks on the `ion-item`, so the whole component is interactive for users. I don't know what a bad user  experience you mean.

Speaking about `pointer-events` property, it is not available at least for React users:
IncidentUpdateNotifications_tsx_—_ilert-mobile

If you can suggest other ways to solve this problem, I will be happy to look into them.

liamdebeasi:
> Speaking about pointer-events property, it is not available at least for React users:

`pointer-events` is CSS, not a component property. So you would do:

```css
ion-toggle {
  pointer-events: none;
}
```

> And again, please take a closer look at the example I described above. I have controlled ion-toggle and onClick event on the ion-item. The value of the ion-toggle will be changed, when user clicks on the ion-item, so the whole component is interactive for users. I don't know what a bad user experience you mean.

Can you please create a code example (GitHub repo, StackBlitz, CodeSandbox, etc) of the issue you are running into?

yacut:
> pointer-events is CSS, not a component property. So you would do:
> 
> ion-toggle {
>   pointer-events: none;
> }

Sorry, my bad.

Here is an example: https://codesandbox.io/s/ionic-react-tooglecheckbox-issue-forked-8i92z
I think your solution with `pointer-events: none;` is working, so you can close the pull request.

Thank you for the response.

liamdebeasi:
Great, glad to hear `pointer-events: none` solves the issue. I am going to close this PR, thanks!

yacut:
I also noticed that everything works also without css. Maybe in these three months that the PR exists a lot has changed in Ionic components. 😄

Issue #21982, opened 2020-08-26
bug: interactive items in ion-item unexpectedly being focused with delegatesFocus


larsblumberg:
# Bug Report

**Ionic version:**
[ ] **4.x**
[x] **5.x**

**Current behavior:**
On Safari in macOS or iOS, `IonInput` (`ion-input`) steals focus when its value gets changed programmatically.

**Expected behavior:**

`ion-input` doesn't take the focus.

**Steps to reproduce:**
The following code snippet reproduces the effect that when an `ion-input`’s value got changed progammatically, here via a button click, then the `ion-input` will receive focus automatically as an effect. This effect is only observable within an `ion-item` container.

Apart from the fact that this might be unwanted and thus should be controllable by the programmer, it also comes with the side effect of the virtual keyboard popping up since the input receives focus. That’s only acceptable if the user sets focus explicitely on the input. However in this case, the input surprisingly "steals" the focus.

**Related code:**

```jsx
const Test = () => {
  const [value, setValue] = useState("")
  return <IonItem>
    <IonInput value={value} onIonChange={e => setValue(e.detail.value)} />
    <IonButton onClick={() => setValue("foo")}>foo</IonButton>
  </IonItem>
}
```

Full demo project available here: https://github.com/larsblumberg/ion-input-focus-stealing-on-safari

**Other information:**

Effect only observable when input is contained within an `ion-item` and only in Safari on macOS and iOS.

**Ionic info:**

```
Ionic:
   Ionic CLI       : 6.11.1 (/Users/lars/.config/yarn/global/node_modules/@ionic/cli)
   Ionic Framework : @ionic/react 5.2.2

Capacitor:
   Capacitor CLI   : 2.0.2
   @capacitor/core : 2.2.0

Utility:
   cordova-res : not installed
   native-run  : not installed

System:
   NodeJS : v14.4.0 (/usr/local/Cellar/node/14.4.0/bin/node)
   npm    : 6.14.4
   OS     : macOS Catalina
```

liamdebeasi:
Thanks for the issue. Can you reproduce this in a blank Ionic starter app and provide a link to the GitHub repo?

larsblumberg:
Here's a full Ionic app which reproduces this issue on Safari in macOS and iOS: https://github.com/larsblumberg/ion-input-focus-stealing-on-safari

larsblumberg:
[...]

PR #22049, opened 2020-09-08
fix(input): only focus the first input / textarea when clicking on the parent item

### LINKED ISSUE #22032: bug: iOS: Same as #21509 but with ion-textarea
I make it very short: This is the same bug as #21509 which was fixed in ionic/core 5.3.2, but if there's an `ion-textarea` instead of `ion-input`, clicking on `ion-item` doesn't properly focus the textarea.

The solution is to `querySelector()` `ion-textarea` as well: https://github.com/ionic-team/ionic-framework/blob/master/core/src/components/item/item.tsx#L192

cc @brandyscarney

### LINKED ISSUE #22037: bug: [@ionic/angular@5.3.2 bug] always focus at first ion-input inside ion-item w/ multiple inputs
**Current behavior:**

I have switch 5.3.1 , 5.3.2 tested, It only happens in 5.3.2
If there are multiple ion-inputs inside ion-item, you have no idea to input data on ion-inputs except 1st one ex:
It always auto focus on 1st ion-input!

# Bug Report
Ionic 5
```
<ion-item>
    <ion-input></ion-input>  <!-- 1st, focus always goes here -->
    <ion-input></ion-input>  <!-- 2nd, click here jump to 1st -->
    <ion-input></ion-input>  <!-- 3rd, click here jump to 1st-->
    <ion-input></ion-input>  <!-- ... , click here jump to 1st-->
    <ion-input></ion-input>  <!-- ... , click here jump to 1st-->
</ion-item>
```

**Expected behavior:**

should correctly  focus each ion-input individually
**Steps to reproduce:**

create templat like above provided codes

brandyscarney:
:warning: dev build: `5.4.0-dev.202009081606.5ddd37f` :warning:

## Pull request checklist

## Pull request type

 

 

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

Issue Number: fixes #22037 resolves #22032

## What is the new behavior?

- Check if the event target is an ion-item before focusing the first input, thus if another input exists it won't change focus
- Add support for focusing ion-textarea if that exists as the first input

- [ ] Yes
- [x] No

### DIFF
--- a/core/src/components.d.ts
+++ b/core/src/components.d.ts
@@ -2440,6 +2440,10 @@ export namespace Components {
          */
         "enterkeyhint"?: 'enter' | 'done' | 'go' | 'next' | 'previous' | 'search' | 'send';
         /**
+          * This is required for a WebKit bug which requires us to blur and focus an input to properly focus the input in an item with delegatesFocus. It will no longer be needed with iOS 14.
+         */
+        "fireFocusEvents": boolean;
+        /**
           * Returns the native `<textarea>` element used under the hood.
          */
         "getInputElement": () => Promise<HTMLTextAreaElement>;
@@ -2480,7 +2484,11 @@ export namespace Components {
          */
         "rows"?: number;
         /**
-          * Sets focus on the specified `ion-textarea`. Use this method instead of the global `input.focus()`.
+          * Sets blur on the native `textarea` in `ion-textarea`. Use this method instead of the global `textarea.blur()`.
+         */
+        "setBlur": () => Promise<void>;
+        /**
+          * Sets focus on the native `textarea` in `ion-textarea`. Use this method instead of the global `textarea.focus()`.
          */
         "setFocus": () => Promise<void>;
         /**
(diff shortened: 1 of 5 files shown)


liamdebeasi (approved):

brandyscarney:
Open the test at `/src/components/item/test/inputs` in Chrome

1) If you clicked on something after opening, refresh so nothing has focus
2) Click on the label “Multiple inputs” — see that the first input “Input 1" has focus
3) Click on “Input 2” — see that it has focus
4) Click on the label “Multiple inputs” again - see that the second input remains focused

Repeat in Safari, except on step 4 it will focus the first input again. This is due to Safari changing the `activeElement` to the first input when you click the item label, even though it doesn't change the focus. Do we revert back to where they both behave the same? cc @liamdebeasi

liamdebeasi:
Agreed re: following iOS 14 behavior on iOS.

Also, this fix should also account for other focusable elements too not just `ion-input` and `ion-textarea`. For example, if you have an input and a button, clicking the button would cause the input to be focused which is not always correct. See: https://github.com/ionic-team/ionic-framework/issues/21982 It seems to be the same issue where you click the second input, but the first input gets focused.

PR #22087, opened 2020-09-14
feat(segment): add swipeGesture property to allow for disabling of the swipe gesture

### LINKED ISSUE #22048: feat: allow users to disable swipe gesture on segment
# Feature Request

**Ionic version:**

[x] **5.x**

Related to #21251.
This issue has been closed as the issue matched "the native iOS spec.", which is fine. But as @lincolnthree already mentioned in the comments on his issue, this behavior does not seem to match the native Android spec.

I feel like we can and should give Ionic users the option to choose wether they want the `ionChange` event to be triggered on this swipe gesture or not. It's basically already disabled in the `scrollable=true` mode, so the code is already there (I guess). The problem with the `scrollable` mode is that it isn't always desirable because of the scroll bar.

If we were to add this feature, I think we should default to the current behavior for backwards compatibility.

In the `ion-segment` component, there should be an option `changeOnSwipe` or `triggerChangeOnSwipe` (or something similar like that).

The default value should be true, to keep it the same as it is now.

When this is set to false, the behavior seen in #21251 should not trigger an `ionChange` event. Also, I think swiping between different `ion-segment-button`s should be disabled in this case.

**Describe Alternatives**

I tried using the `scrollable` attribute, and tried hiding the scroll bar, but this is a bad and hacky "solution".

**Related Code**

**Additional Context**

TakumaKira:
closes #22048

## Pull request checklist

## Pull request type

 

 

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes

Issue Number: #22048

## What is the new behavior?

- By setting `changeOnSwipe` property to false, developers can disable to select ionSegmentButton inside on swiping/dragging.

- [ ] Yes
- [x] No


brandyscarney:
Could we call this `swipeGesture` to match the naming of `ion-menu` and `ion-nav`?

```
<ion-menu swipeGesture="false" contentId="nav"> </ion-menu>
```

brandyscarney:
Thank you for this by the way! The code looks good on initial review. Requested myself and Liam to test it. 🙂

core/src/components/segment/segment.tsx
```diff
@@ -49,6 +49,16 @@ export class Segment implements ComponentInterface {
    */
   @Prop() scrollable = false;
 
+  /**
+   * If `false`, the gesture to drag the indicator between the buttons will be disabled.
```
liamdebeasi, line 53: Can we re-word this so that it explain it more "positively"? Something like "If `true`, users will be able to swipe between segment buttons to activate them".

core/src/components/segment/segment.tsx
```diff
@@ -373,10 +382,18 @@ export class Segment implements ComponentInterface {
 
   private onClick = (ev: Event) => {
     const current = ev.target as HTMLIonSegmentButtonElement;
+    /**
+     * If current element is ion-segment then that means
+     * user tried to swipe to other ion-segment-button,
+     * and we should not update the checked button.
+     */
+    if (current.tagName === 'ION-SEGMENT') {
```
liamdebeasi, line 390: Can you clarify why this is needed?

liamdebeasi (commented):
A few small changes, but otherwise this looks really good. Nice job!

core/src/components/segment/segment.tsx
TakumaKira: I will change it to yours!

core/src/components/segment/segment.tsx
TakumaKira: What point should I clarify? Without these lines, the animation doesn't work if the user tried to swipe to then click another button.

liamdebeasi: I am mostly curious what issue this fixed by adding that line

TakumaKira: I see🙂 The change or comment itself is ok?

liamdebeasi: Yeah change + comment looks good, thanks!

core/src/components/segment/test/basic/index.html
```diff
@@ -229,6 +232,14 @@
           }
         }
 
+        function toggleSwipeGesture() {
+          const ionSegmentElement = document.querySelectorAll('ion-segment');
+          for (var i = 0; i < ionSegmentElement.length; i++) {
+            const swipeGesture = ionSegmentElement[i].swipeGesture === false ? false : true;
+            ionSegmentElement[i].swipeGesture = swipeGesture;
```
liamdebeasi, line 239:
```suggestion
            ionSegmentElement[i].swipeGesture = !ionSegmentElement[i].swipeGesture;
```

I noticed this wasn't toggling properly, so we can just do this instead.

liamdebeasi (changes requested):
One small test change, then I think we are good to merge!

liamdebeasi (approved):
This looks great, thanks for taking care of this!

liamdebeasi:
@TakumaKira One last thing -- can you sync this branch with the latest `master`? We had a few build issue with node 15/npm 7 that have since been resolved, but you will need to grab the latest from `master` to get the fixes.

TakumaKira:
@liamdebeasi

> @TakumaKira One last thing -- can you sync this branch with the latest `master`? We had a few build issue with node 15/npm 7 that have since been resolved, but you will need to grab the latest from `master` to get the fixes.

I think I'm done ;) I rebased the branch on the current `master`(commit d77a9d5).

liamdebeasi:
Merged. Thank you!

ZaLiTHkA:
I just had to apply the hacky workaround as [posted in this comment](https://github.com/ionic-team/ionic-framework/issues/22161#issuecomment-698310038) on #22161 (which led me to #22048, which led me here), but I had to do this in _a current app_ with `@ionic/angular@6.1.1` and `@angular/core@13.3.3`.. has this issue perhaps returned in recent builds?

one intersting observation though: the "segment selection change on page scroll" issue only appears to happen on `ion-segment` elements that _do_ have `scrollable` enabled. I'm not sure if this was a requirement of the original problem.. so could the behaviour I'm seeing here be caused by something else instead?

PR #22113, opened 2020-09-18
fix(radio): tab only to the checked (or first if none exists) radio for a11y compliance

### LINKED ISSUE #21743: bug: ion-radio-group does not follow a11y best practices
# Bug Report

**Ionic version:**

[ ] **4.x**
[x] **5.x**

**Current behavior:**

ion-radio-group doesn't follow wai aria practices. As a result the user will have to use the tab key itself to navigate between the group and the radio buttons with in the list.

**Expected behavior:**
The aria practice example is provided in the link below :

https://www.w3.org/TR/2017/WD-wai-aria-practices-1.1-20170628/examples/radio/radio-1/radio-1.html#:~:text=The%20role%3D%22radio%22%20attribute,tabindex%3D%22%2D1%22%20.

Essentially:

- tabs should move between `ion-radio-group`s
- if a radio button is "on" that one should be selected when the group is tabbed into
- if no button is "on" the first button should be selected when the group is tabbed into
- arrows should be used to navigate within the group, navigation should be circular from end to start or start to end going backwards

**Steps to reproduce:**
A sample application which demonstrates the issue is in the repo: https://github.com/mahiii85/ionic-test.git

branch: review/ion-set-one

brandyscarney:
Dev build: `5.4.0-dev.202009222045.375f0aa`

## Pull request checklist

## Pull request type

 

 

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

Issue Number: fixes #21743

## What is the new behavior?

-
-
-

- [ ] Yes
- [x] No


core/src/components/radio-group/radio-group.tsx
```diff
@@ -64,6 +94,35 @@ export class RadioGroup implements ComponentInterface {
     }
   }
 
+  @Listen('keydown', { target: 'document' })
+  onKeydown(ev: any) {
```
liamdebeasi, line 98: Can we do `if (ev.target && !this.el.includes(ev.target)) { return; }`? That way if we have 10 radio groups, we aren't going to do 10 `querySelectorAll` calls. It would be good to make this a single listener, but not sure that's possible since this callback seems dependent on the radio group it is in.

core/src/components/radio/test/basic/index.html
```diff
@@ -20,6 +20,26 @@
     </ion-header>
 
     <ion-content id="content" class="radio-test outer-content">
+      <ion-list>
```
liamdebeasi, line 23: If you are on the "Fruits (Group w/ values)" radio group and you tab to the first element, should it uncheck the disabled radio when you use the arrow keys to go to the new radio?

Steps:

1. In the group, "Grape" should be disabled and checked.
2. Tab to group. "Apple" should be focused but unchecked.
3. Press the right arrow key. "Cherry" should be focused and now checked.

brandyscarney: I would say so, since that's how it works when there are no checked radios. 

1. Tab to the Pizza Toppings (Unchecked Group)
2. See that Pepperoni isn't selected
3. Use arrow keys, see that it selects

I think having a radio that is disabled and checked is really weird UX. They can't get back to the radio since it's disabled, so we just have to pretend it doesn't exist.

brandyscarney:
@liamdebeasi Can you take a look at this commit: https://github.com/ionic-team/ionic-framework/pull/22113/commits/1b8a9d7cb82904a66be8b08c13bd9d9fa58666da

I figured out that the reason this wasn't working when used in a select popover is because it was still grabbing the first input with `.ion-focusable` even though the tabindex was negative.

liamdebeasi (approved):

PR #22131, opened 2020-09-21
perf(vue): improve build with smaller bundle sizes

liamdebeasi:
## Pull request checklist

## Pull request type

 

 

- [ ] Bugfix
- [ ] Feature
- [x] Build related changes

Issue Number: N/A

## What is the new behavior?

- Added rollup to improve bundling

- [ ] Yes
- [x] No


aparajita:
Hey @liamdebeasi, how's it going with this? I just looked at a bundle analysis and it looks like no tree shaking is happening. My app uses only page, content, input, and button, and I'm seeing just about every other possible component in the bundle, including the swiper lib. There must be a way to avoid this.

Bundle analysis is attached.

[report.html.zip](https://github.com/ionic-team/ionic-framework/files/5504329/report.html.zip)

liamdebeasi:
We are looking into other ways of improving, but unfortunately this is really just a limitation in Webpack with its treeshaking capabilities. We are going to be using a Custom Elements build in Ionic Vue soon that should help -- we plan on doing a public beta for this so keep an eye on https://twitter.com/Ionicframework if you want to test it.

aparajita:
Hmm.... have you looked into using rollup for a production build? Its tree shaking is supposed to be better.

liamdebeasi:
Ionic Vue uses rollup internally. The problem is that Vue itself uses Webpack in all of its starter apps, so that is what most developers are going to have. There is a rollup plugin for Vue but it doesn't fully support Vue 3 yet: https://github.com/vuejs/rollup-plugin-vue/issues

aparajita:
This seems strange, because Vue 3 itself was specifically rewritten to tree shake better with webpack. Perhaps there is some chain or cycle of dependencies in Ionic/Vue that is preventing tree shaking.

liamdebeasi:
It doesn't have anything to do with what Vue 3 does. Inside of Stencil, there is an `import` statement with a path variable for lazy loading. Webpack makes a regex out of the path variable and bundles anything that matches the regex, which is all Ionic Framework chunks. The problem is that Webpack is "blindly" doing this without verifying whether or not a particular chunk is actually being used in the app.

The custom elements build will let us avoid the lazy loading completely, so we are hopeful that this will resolve the issue.

aparajita:
> The custom elements build will let us avoid the lazy loading completely, so we are hopeful that this will resolve the issue.

Look forward to it.

PR #22219, opened 2020-09-30
chore(github): add pull request labeler for packages

brandyscarney:
This adds the Github workflow to automatically label PRs based on the files changed, removing the code that is done by ionitron. I have tested it out in a sample repository here: https://github.com/brandyscarney/testing-labeler/pull/3

It seems to be behaving as desired, removing labels when the changes are reverted and adding them based on the file path.

Screen Shot 2020-09-30 at 1 23 23 PM

### DIFF
--- a/.github/labeler.yml
+++ b/.github/labeler.yml
@@ -1,0 +1,22 @@
+# This is used with the label workflow which
+# will triage pull requests and apply a label based on the
+# paths that are modified in the pull request.
+#
+# For more information, see:
+# https://github.com/actions/labeler
+
+'package: core':
+  - core/**/*
+
+'package: angular':
+  - angular/**/*
+  - packages/angular-server/**/*
+
+'package: react':
+  - packages/react/**/*
+  - packages/react-router/**/*
+
+'package: vue':
+  - vue/**/*
+  - packages/vue/**/*
+  - packages/vue-router/**/*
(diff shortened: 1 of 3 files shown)


.github/labeler.yml
```diff
@@ -0,0 +1,22 @@
...
+  - packages/angular-server/**/*
+
+'package: react':
+  - packages/react/**/*
+  - packages/react-router/**/*
+
+'package: vue':
+  - vue/**/*
+  - packages/vue/**/*
+  - packages/vue-router/**/*
```
liamdebeasi, line 22: Minor thing, but are we able to do like `packages/vue-*`? Just in case we have `@ionic/vue-server` for SSR we won't need to update this

liamdebeasi (approved):

.github/labeler.yml
brandyscarney: I tested it out and it does seem to work, so I can make that change. Should we do the same for react/angular? Or leave it?

brandyscarney: Should I remove the first check for vue? Are we deleting all of the old stuff?

liamdebeasi: Yeah you can remove the first check -- it will be removed when Ionic Vue launches

liamdebeasi: might as well do it for the other one in case we add something

Issue #22324, opened 2020-10-15
bug: Previous page is not drawn when navigating back in history


virusv:
# Bug Report

**Ionic version:**
[ ] **4.x**
[x] **5.x**

**Current behavior:**
When you press the back button, the URL address changes, but the previous page is not displayed

**Expected behavior:**
Display the previous page

**Steps to reproduce:**
install ionic vue beta
`ionic start myApp blank --type vue --tag vue-beta`

**packages:**
```json
"@ionic/vue": "5.4.0-rc.3",
"@ionic/vue-router": "5.4.0-rc.3",
"core-js": "^3.6.5",
"vue": "^3.0.0-0",
"vue-router": "^4.0.0-beta.13"
```

Create 3 routes:
```js
const routes: Array<RouteRecordRaw> = [
    {
        path: '/',
        name: 'home',
        component: Home
    },
    {
        path: '/products/',
        name: 'products',
        component: Products
    },
    {
        path: '/products/:id/view',
        name: 'product',
        component: ProductSingle,
    }
];
```

Go from the first route to the second with query parameters:
```js
this.$router.push({
    name: 'products',
    query: {
        cat: 1,
        order: 'price',
    }
});
```

then go to 3rd page:
```js
this.$router.push({
    name: 'product',
    params: {
        id: 1,
    }
});
```

And try to go back:
`$router.back()` or using the `ion-back-button` component, the address bar changes, but the previous component is not rendered.

**Related code:**
https://github.com/virusv/bug-ionic-vue-backpage

**Other information:**
The bug exists in previous versions @ionic/vue && @ionic/vue-router (rc.1, rc.2 and -dev.*)

**Ionic info:**

```
Ionic:

   Ionic CLI       : 6.11.8-testing.0 (C:\Users\Ivan Nalivayko\AppData\Roaming\npm\node_modules\@ionic\cli)
   Ionic Framework : @ionic/vue 5.4.0-rc.3

Capacitor:

   Capacitor CLI   : 2.4.2
   @capacitor/core : 2.4.2

Utility:

   cordova-res : not installed
   native-run  : not installed

System:

   NodeJS : v14.8.0 (C:\Program Files\nodejs\node.exe)
   npm    : 6.14.7
   OS     : Windows 10
```

liamdebeasi:
Thanks for the issue. Can you try the following dev build and let me know if it resolves the issue?

```
[...]

PR #22350, opened 2020-10-19
fix(vue): going back with query params now goes to correct view

liamdebeasi:
## Pull request checklist

## Pull request type

 

 

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

Issue Number: resolves https://github.com/ionic-team/ionic-framework/issues/22324

## What is the new behavior?

- Use the history listener to just grab the action/direction since it fires before `router.beforeEach` and let the `router.beforeEach` callback handle calling the ion router method.

- [ ] Yes
- [x] No

### DIFF
--- a/packages/vue-router/src/router.ts
+++ b/packages/vue-router/src/router.ts
@@ -59,7 +59,7 @@ export const createIonRouter = (opts: IonicVueRouterOptions, router: Router) => 
      */
     currentNavigationInfo = {
       action: info.type,
-      direction: info.direction === '' ? 'none' : info.direction
+      direction: info.direction === '' ? 'forward' : info.direction
     };
   });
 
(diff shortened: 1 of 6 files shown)


packages/vue-router/src/router.ts
```diff
@@ -39,8 +45,23 @@ export const createIonRouter = (opts: IonicVueRouterOptions, router: Router) =>
...
+     * additional information as to the type
+     * of navigation (forward, backward, etc).
+     *
+     * We can use this to better handle the
+     * `handleHistoryChange` call in
+     * router.beforeEach
+     */
+    currentNavigationInfo = {
+      action: info.type,
+      direction: info.direction === '' ? 'none' : info.direction
```
liamdebeasi, line 62: This `"none"` assumption might not be correct. Need to test this and see if we should use `"forward"` instead.

liamdebeasi: Since we default to `"forward"` in all other scenarios, I am going to do the same here.

Issue #21509, opened 2020-06-11
bug: ion-item padding prevents child ion-input focus


ehorodyski-ionic:
# Bug Report

**Ionic version:**

[ ] **4.x**
[x] **5.x**

**Current behavior:**

While working with a community member, it was observed that when a floating label and `ion-input` are nested in an `ion-item`, users can tap the left padding of the item and it will:

1. Animate the floating label to the top
2. Present the device keyboard
3. *Not* focus the input field

This behavior was captured for iOS with Ionic Angular 5.2.0.

**Expected behavior:**

When tapping the padding, it should place the input field in focus in addition to the other actions performed.

**Steps to reproduce:**

1. Create an `ion-item` with an `ion-label` and `ion-input` as children.
2. Set the `position` of the `ion-label` to `floating`.
3. Open the app on an iOS device/simulator.
4. Tap the far-left of the `ion-item`.

Please find a GIF of this behavior towards the bottom of this Issue.

**Related code:**

Reproduction Application: https://github.com/ehorodyski-ionic/ionic-floating-label-focus-issue

**Other information:**

![ezgif-1-05eaea0df48e](https://user-images.githubusercontent.com/62819287/84413496-b3737880-abde-11ea-83e2-88dab1520e3f.gif)

**Ionic info:**

```
Ionic:

   Ionic CLI                     : 6.10.0 
   Ionic Framework               : @ionic/angular 5.2.0
   @angular-devkit/build-angular : 0.901.8
   @angular-devkit/schematics    : 9.1.8
   @angular/cli                  : 9.1.8
   @ionic/angular-toolkit        : 2.2.0

Capacitor:

   Capacitor CLI   : 2.2.0
   @capacitor/core : 2.2.0

Utility:

   cordova-res (update available: 0.14.0) : 0.11.0
   native-run                             : 1.0.0

System:

   NodeJS : v12.16.1 (/usr/local/bin/node)
   npm    : 6.14.4
   OS     : macOS Catalina
```

torgnywalin:
I have the same issue if clicking to the right of the label text (over the input).
Clicking anywhere on the ion-item should focus the input. 

I have ionic version 5.2.1. Testet on iOS.

kevingrabher:
Could be a duplicate of #17694 ?

liamdebeasi:
[...]

PR #22352, opened 2020-10-19
fix(item): only add click events to clickable items

### LINKED ISSUE #22011: bug: ion-item being read as clickable by screen reader
**Ionic version:**

[ ] **4.x**
[x] **5.x** (Seems to have been introduced with 5.3.2)

**Current behavior:**

ion-item when used in a page is read as clickable. This was first noticed with the release of version 5.3.2

**Steps to reproduce:**
Can be verified from:

https://ionicframework.com/docs/demos/api/item/index.html?ionic:mode=ios

**Other information:**

ion-items-clicable

brandyscarney:
Dev build: `5.5.0-dev.202010301640.9336349`

## Pull request checklist

## Pull request type

 

 

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

Issue Number: fixes #22011

## What is the new behavior?

Correctly announce items for screen readers, don't announce clickable when it is not

- [ ] Yes
- [x] No


kensodemann (approved):
Tested in a couple different apps. Does not break anything that I have found.
I was not able to test in NVDA to see if it fixed the issue there, but had no issues in the apps I tried it in.

LGTM, I am fine with merging it.

liamdebeasi (commented):
My only concern about this is that we are making the click handler not reactive. If an item loads with no inputs then adds an input after a delay, this delegatesFocus function will never fire.

I can't imagine this is a common use case, but this PR would break that behavior.

brandyscarney:
I fixed Liam's concerns with adding an input later on. The event listener is now added in the `componentDidUpdate` only if it has not already been added. Dev build for testing `5.5.0-dev.202010301640.9336349`. We need to test out that it doesn't break the fix for https://github.com/ionic-team/ionic-framework/issues/21509.

liamdebeasi (approved):
Other than the removeEventListener change we talked about offline, this looks good to go

PR #22449, opened 2020-11-06
fix(refresher): ios native refresher now works in side menu

liamdebeasi:
## Pull request checklist

## Pull request type

 

 

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

Sometimes the scrollEl's clientHeight is zero because the containing element is hidden. This results in the refresher firing too early.

## What is the new behavior?

- Added a default MAX_PULL value in the event the content height is 0.

- [ ] Yes
- [x] No

### DIFF
--- a/core/src/components/refresher/refresher.tsx
+++ b/core/src/components/refresher/refresher.tsx
@@ -165,7 +165,15 @@ export class Refresher implements ComponentInterface {
   private async setupiOSNativeRefresher(pullingSpinner: HTMLIonSpinnerElement, refreshingSpinner: HTMLIonSpinnerElement) {
     this.elementToTransform = this.scrollEl!;
     const ticks = pullingSpinner.shadowRoot!.querySelectorAll('svg');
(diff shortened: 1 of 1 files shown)


EinfachHans:
The title of this PR is not correct i think 😃 And what about assigning `MAX_PULL` later? Where we knew the `scrollEl` must be visible. For example when the gesture starts?

liamdebeasi:
Our PR titles and commit messages are designed to explain a concrete bug that a commit fixes, so just saying "ensure MAX_PULL is not 0" is not enough. In this case, this particular bug was discovered in a menu, so the title is going to reference the menu.

liamdebeasi:
I'd rather not do DOM reads every time the user puts their finger/pointer down on the screen, but I can look into an alternative.

EinfachHans:
Of course a fixed value is also not bad, but currently i like that it is based on the scroll height 😃 Maybe just save the scroll height on the first time the gesture starts if this is an option

PR #22464, opened 2020-11-10
force web attribute for ios/android

aboalwi:
## Pull request checklist

## Pull request type

 

 

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes
- [x] Other (please describe): 
Having more options.

It's automatically decide to use native or web refresher based on pre-defined conditions.

Issue Number: N/A

## What is the new behavior?
Can always use web ion-refresher by using forceWebIos/forceWebAndroid attributes.

- [ ] Yes
- [x] No

## Other information
Having ios scrolling issues because of bounce so we basically disabled bounce completely in ios. This caused ion-refresher to stop working this will help us keep going with our project.


liamdebeasi:
Thanks for the PR. What problem does this PR solve? The refresher should automatically detect which variant to use based upon the mode of your application.

aboalwi:
The problem is ion-refresher native implementation never works on iOS when bounce/rubber band is disabled. This forces ion-refresher to use the web implementation instead. 
I had to disable iOS bounce for scroll issues like: When a video is playing inside a list and user scroll to top or bottom. Or an ion-range is being dragged and scroll to top.

liamdebeasi:
Thanks for the clarification. The native iOS implementation relies on the rubber band effect to work, but there's an easier way to accomplish what you are trying to do.

Setting the `pulling-icon` property on `ion-refresher-content` to an arrow icon (see https://ionicons.com/ for options) will force the refresher to use the web implementation instead on iOS mode. For example, you could do something like this:

```html
<ion-refresher>
  <!-- this would be `pullingIcon="arrow-down"` instead in Ionic Angular -->
  <ion-refresher-content pulling-icon="arrow-down"></ion-refresher-content>
</ion-refresher>
```

Can you give that a try and let me know if it resolves the issue?

aboalwi:
Yep it works! Thanks!

Issue #21646, opened 2020-06-29
bug: labels are not read for ion-toggle


kensodemann:
# Bug Report

**Ionic version:**

[ ] **4.x**
[x] **5.x**

**Current behavior:**
The label for toggle button is not read. (it’s read as blank).
https://ionicframework.com/docs/api/toggle

kensodemann:
This sort of works.

**mobile**: works fine for me on both Android and iOS devices
**web**: confirmed using Apple VoiceOver on my Mac while in Chrome that the label is not read either when tabbing in to the toggle, rather it says "Button"

kensodemann:
Others have tried on Windows using  NVDA reader and Windows narrator with the same result.

PR #22402, opened 2020-10-28
fix(checkbox): use a native input to fix a11y issues with axe and screen readers

brandyscarney:
Dev build: `5.5.0-dev.202011111928.7c95d89` 🚨  This dev build includes https://github.com/ionic-team/ionic-framework/pull/22477 for testing of both at the same time

:warning: Depends on https://github.com/ionic-team/ionic-framework/pull/22352 :warning: 

Fixes #21644
Fixes #20517
Fixes #17796

If this ends up working well for Checkbox, this can also be used for the following issues:
- https://github.com/ionic-team/ionic-framework/issues/21552
- https://github.com/ionic-team/ionic-framework/issues/21646
- https://github.com/ionic-team/ionic-framework/issues/21548

# How this updates a11y and screen readers

## Axe

No longer throws errors for aria attributes or buttons without content

## NVDA

Tested in Edge and Chrome, reads as:

```
check box  checked    Default
check box  not checked    Primary
check box  checked    Tertiary
check box  checked    Success
check box  checked    Warning
check box  checked    Dark
check box  checked    Danger
check box  checked    Light
check box  checked    Medium
```

Checkbox must be in a form to use space to select it with NVDA running, otherwise you need to press `Ins` + `space` to enter forms mode, then you can use space to select.

## VoiceOver

On Mac:

- Chrome: properly reads uncheck or check, space bar works, incorrectly announces as `checkbox inside of a checkbox` (this is a compromise we have to make in order for it to work properly in Safari)
- Safari: properly reads uncheck or check, space bar works, correctly announces as being on a checkbox

On iOS:

Properly selects checkbox item, double tap will uncheck or check checkbox, skips over disabled checkboxes entirely

Issue #21552, opened 2020-06-16
bug: ion-toggle and ion-select generates empty buttons


kensodemann:
# Bug Report

**Ionic version:**

[ ] **4.x**
[x] **5.x**

**Current behavior:**
Axe chrome extension highlights the "empty button" error for ion-toggle button and ion-select

**Expected behavior:**
A clean report.

Refer to item 2 in the document shared in Slack for further details.
https://ionicteam.slack.com/archives/C7CGXAVTL/p1592319616186200

PR #22477, opened 2020-11-11
fix(toggle): properly announce toggle as switch with a label on screen readers

brandyscarney:
Dev build: `5.5.0-dev.202011111928.7c95d89`

:warning: Depends on https://github.com/ionic-team/ionic-framework/pull/22402 :warning:

fixes https://github.com/ionic-team/ionic-framework/issues/21646 references #21552

Added a switch from the example here: https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/Switch_role

Found our Toggle behaves the same way in macOS and iOS Safari, see the following:

Screen Shot 2020-11-12 at 12 30 56 PM
Screen Shot 2020-11-12 at 12 31 59 PM


kensodemann:
Testing on my iPad, this announces toggle as checkbox
On Android, though, it does announce it as a switch.

Issue #19250, opened 2019-09-02
feat: Slides: add elements outside `.swiper-wrapper`


tricki:
# Feature Request

**Ionic version:**
[x] **4.x**

**Describe the Feature Request**
Adding a way to add a parallax background to a slider based on total slider progress. This requires an element that is a direct child of .swiper-container (outside .swiper-wrapper).
[Relevant Swiper docs](https://idangero.us/swiper/api/#parallax)

**Describe Preferred Solution**
Add a new named slot outside of `.swiper-wrapper`.

**Describe Alternatives**
Moving the parallax element manually after load.

**Related Code**
```
<ion-content>
  <ion-slides>
    <div
      slot="parallax"
      class="bg-parallax"
      style="background-image:url(https://loremflickr.com/320/240/)"
      data-swiper-parallax="-23%"
    ></div>
    <ion-slide>
      Slide 1
    </ion-slide>
    <ion-slide>
      Slide 2
    </ion-slide>
  </ion-slides>
</ion-content>
```

**Additional Context**
Previous requests:
https://github.com/ionic-team/ionic/issues/13584
https://github.com/ionic-team/ionic/issues/15330

liamdebeasi:
Thanks for the issue. We recently exposed the full Swiper API in Ionic 4.9.0. Does having access to the full API solve your issue?

tricki:
I don't think so since the parallax DOM element needs to be outside of [the slot](https://github.com/ionic-team/ionic/blob/master/core/src/components/slides/slides.tsx#L498).

From [the Swiper docs](https://idangero.us/swiper/api/#parallax): 

> There are two types of parallax elements supported: [1] Direct child elements of swiper-container. Parallax effect for such elements will depend on total slider progress. Useful for parallax backgrounds

A slot outside of `.swiper-wrapper` would also be useful for adding things like prev/next buttons, especially with the new full API access.

MarkChrisLevy:
@liamdebeasi Can you please take a look at #22598 - it is very small change, but very needed ;-)

liamdebeasi:
Thanks for the issue! With the release of Ionic 6, we made the decision to deprecate `ion-slides` in favor of using Swiper.js directly.

Moving forward, `ion-slides` will only receive critical security fixes, and the component will be removed in Ionic 7. As a result, I am going to close this issue.

We have prepared migration guides for each of the 3 JavaScript frameworks we support, so developers can get started migrating right away.

[...]

Issue #15330, opened 2018-08-25
Slides: add elements outside `.swiper-wrapper`


tricki:
# Support Question

I'm trying to add a parallax background to a slider (based on total slider progress). This requires an element that is a direct child of `.swiper-container` (outside `.swiper-wrapper`).

I thought I could just add another slot to `Slides` at the correct position. But that doesn't work because a `<slot>` element is inserted which doesn't mirror the original attributes. And Swiper only checks for direct children of `.swiper-container`.

![image](https://user-images.githubusercontent.com/523154/44618492-05ff6b80-a877-11e8-86bc-9c307c22aa4c.png)

Relevant Swiper code (`$el === $('.swiper-container')`):

    $el.children('[data-swiper-parallax], [data-swiper-parallax-x], [data-swiper-parallax-y]')
      .each((index, el) => {
        swiper.parallax.setTransform(el, progress);
      });

Source: https://github.com/nolimits4web/swiper/blob/master/src/components/parallax/parallax.js#L56

Is there a way I can insert an element into the Shadow DOM without a `<slot>` element being inserted? Or is there a way to mirror the attributes of the "slotted" element (`<div slot="container" data-swiper-parallax="-23%"/>`) to the `<slot>` element?

PR #22598, opened 2020-11-28
feat(slides): add container slot

MarkChrisLevy:
## Pull request checklist

## Pull request type

 

 

- [ ] Bugfix
- [X] Feature
- [ ] Build related changes

All child elements of ion-slides are placed within swiper-wrapper - that is correct for ion-slide, but it makes impossible to add elements like custom pagination or other swiper plugins, which should be placed directly within swiper-container.

Issue Number:
#19250
#15330

## What is the new behavior?

If elements other than ion-slide must be added to ion-slides, then it should be placed in "container" slot.

- [ ] Yes
- [X] No

### DIFF
--- a/core/src/components/slides/readme.md
+++ b/core/src/components/slides/readme.md
@@ -429,15 +429,16 @@ const slideOpts = {
 };
 ```
 
-## Container slot
-Swiper allows to add non slide children, which can be used to add additional functionality (e.g. pagination) - to add that kind of child element use named slot "container".
+## Container slots
+Swiper allows to add non slide children, which can be used to add additional functionality (e.g. pagination) - to add that kind of child element use named slot "top" or "bottom".
 ```html
 <ion-slides>
+  <div slot="top">...</div>
   <ion-slide>a</ion-slide>
   <ion-slide>b</ion-slide>
   <ion-slide>c</ion-slide>
   <ion-slide>d</ion-slide>
-  <div slot="container">...</div>
+  <div slot="bottom">...</div>
 </ion-slides>
 ```
 
--- a/core/src/components/slides/readme.md
+++ b/core/src/components/slides/readme.md
@@ -429,6 +429,18 @@ const slideOpts = {
 };
 ```
 
+## Container slot
+Swiper allows to add non slide children, which can be used to add additional functionality (e.g. pagination) - to add that kind of child element use named slot "container".
+```html
+<ion-slides>
+  <ion-slide>a</ion-slide>
+  <ion-slide>b</ion-slide>
+  <ion-slide>c</ion-slide>
+  <ion-slide>d</ion-slide>
+  <div slot="container">...</div>
+</ion-slides>
+```
+
 <!-- Auto Generated Below -->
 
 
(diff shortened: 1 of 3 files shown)


core/src/components/slides/slides.tsx
```diff
@@ -521,6 +521,9 @@ export class Slides implements ComponentInterface {
         <div class="swiper-wrapper">
           <slot></slot>
         </div>
+        
+        <slot name="container"></slot>
```
liamdebeasi, line 525: In addition to adding the slot, we need to update the component file so this is automatically listed on our docs site. See https://github.com/ionic-team/ionic-framework/blob/master/core/src/components/item/item.tsx#L12-L14 for an example.

core/src/components/slides/readme.md
```diff
@@ -429,6 +429,18 @@ const slideOpts = {
 };
 ```
 
+## Container slot
+Swiper allows to add non slide children, which can be used to add additional functionality (e.g. pagination) - to add that kind of child element use named slot "container".
```
liamdebeasi, line 433: Can we also add an example of this in `/test/basic/index.html`?

core/src/components/slides/slides.tsx
liamdebeasi, line 525: Would it be useful to have a slot that is rendered before `.swiper-wrapper`? In that case, it may be useful to have the slots named `top` and `bottom`.

liamdebeasi (commented):
Thanks for doing this! Added a few comments.

core/src/components/slides/slides.tsx
MarkChrisLevy: Sure! I did that - added slot top and bottom, but maybe (just maybe), better naming would be "before" and "after", which means "before/after" slides? If you think that top/bottom is ok then I'm fine with that.

liamdebeasi: "before" and "after" could work too. I'm not 100% sold on the "top" and "bottom" naming. Maybe @brandyscarney has ideas?

MarkChrisLevy: Done

MarkChrisLevy: @brandyscarney Any comment regarding naming the slots?

MarkChrisLevy: @brandyscarney Knock knock ;-)

MarkChrisLevy:
@brandyscarney Sorry for bothering, but it would be nice to have your voice and maybe merge this PR.

liamdebeasi:
Thanks for the PR, and apologies for the delay! We really appreciate all of the effort you put into building this.

In Ionic Framework v6, the Ionic Team has decided to deprecate `ion-slides` in favor of using Swiper.js directly. There are numerous benefits to this approach, and we think this will be better for developers in the long run. One of the benefits is that you can add custom pagination and use other Swiper plugins without any hassle.

The existing `ion-slides` component will only receive critical fixes (such as security fixes) moving forward. As a result, we are unable to accept this PR.

We have a migration guide on our new docs site which goes more into detail about the benefits of this change and how to migrate: https://beta.ionicframework.com/docs/api/slides#migration

Issue #22602, opened 2020-12-01
bug: ionic vue wrong route instance provided to props function


liamdebeasi:
See: https://forum.ionicframework.com/t/does-ionic-vue-support-router-props/200067/8

Culprit is here: https://github.com/ionic-team/ionic-framework/blob/master/packages/vue/src/components/IonRouterOutlet.ts#L384

Need to provide injectedRoute like RouterView does: https://github.com/vuejs/vue-router-next/blob/master/src/RouterView.ts#L95

liamdebeasi:
Resolved via https://github.com/ionic-team/ionic-framework/pull/22605.

PR #22605, opened 2020-12-01
fix(vue): pass in correct route to props function

liamdebeasi:
## Pull request checklist

## Pull request type

 

 

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

Issue Number: resolves https://github.com/ionic-team/ionic-framework/issues/22602

## What is the new behavior?

- Correct injectedRoute is now passed into props fn
- Props fn is now only called on initial render

- [ ] Yes
- [x] No

### DIFF
--- a/packages/vue/src/components/IonRouterOutlet.ts
+++ b/packages/vue/src/components/IonRouterOutlet.ts
@@ -377,13 +377,34 @@ export const IonRouterOutlet = defineComponent({
          * IonRouterOutlet does not support named outlets.
          */
         const routePropsOption = c.matchedRoute?.props?.default;
+
+        /**
+         * Since IonRouterOutlet renders multiple components,
+         * each render will cause all props functions to be
+         * called again. As a result, we need to cache the function
+         * result and provide it on each render so that the props
+         * are not lost when navigating from and back to a page.
+         * When a component is destroyed and re-created, the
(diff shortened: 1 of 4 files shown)


packages/vue/src/components/IonRouterOutlet.ts
```diff
@@ -374,22 +376,22 @@ export const IonRouterOutlet = defineComponent({
...
-            ...routeProps
-          }
+        const routePropsOption = c.matchedRoute?.props?.default;
+        const routeProps = routePropsOption
+          ? routePropsOption === true
+            ? c.params
+            : typeof routePropsOption === 'function'
+
+            // Only call props function on initial render
+            ? !c.ionPageElement && routePropsOption(injectedRoute)
```
liamdebeasi, line 386: This isn't quite right. If you navigate from a page that has a props function then go back, the props will be cleared.

liamdebeasi: Two options here are:

1. Allow the props fn to be called whenever the component is rendered.
2. Cache the results of the props function in the view stack item.

liamdebeasi: I think it would be annoying to have the function called every time. Leaning towards option 2.

PR #22641, opened 2020-12-08
Custom elements bundle

claviska:
This adds the `dist-custom-elements-bundle` to the framework so users can import components [as described in the Stencil docs](https://stenciljs.com/docs/custom-elements) as opposed to the default lazy loader.

It also moves types from `src/interface.d.ts` to `/src/index.ts`.


claviska:
Closing this in favor of #22665 because this would require us to upgrade Angular's TypeScript which would be a breaking change for some users.

PR #22665, opened 2020-12-11
Custom elements bundle (no type changes)

claviska:
This adds the dist-custom-elements-bundle to the framework so users can import components as described in the Stencil docs as opposed to the default lazy loader.


liamdebeasi:
Closing in favor of https://github.com/ionic-team/ionic-framework/pull/22863.

Issue #22148, opened 2020-09-23
feat: update component behaviors/styles for latest ios


Yohandah:
# Feature Request

**Ionic version:**

[x] **5.x**

**Describe the Feature Request**
iOS 14 introduced new changes and it would be great if Ionic implemented those.

- [ ] spinner style has changed: `ion-spinner` doesn't match.
![85353330-cfb7c500-b4bc-11ea-9a41-776aaac8f0f6](https://user-images.githubusercontent.com/15921402/93994513-42705f00-fd90-11ea-9189-fdb951f7cdf5.gif)

- [ ] when long pressing on back button (top left corner), you can see the nav history
![image](https://user-images.githubusercontent.com/15921402/93994725-7ea3bf80-fd90-11ea-9284-7217c9a39bc3.png)

- [ ] there's a new popover and `ion-popover` doesn't match its style
![120036687_343660223537895_5145024124145916449_n](https://user-images.githubusercontent.com/15921402/93995355-3042f080-fd91-11ea-9e4b-2b7708d6c4de.jpg)

- [ ] new UIDatePicker and `ion-datetime` doesn't match its style
![image](https://user-images.githubusercontent.com/15921402/94027433-1fa77000-fdbb-11ea-9d85-83c9cd7dee76.png)

![image](https://user-images.githubusercontent.com/15921402/93995730-aba4a200-fd91-11ea-866e-e653111bcea3.png)

BerkeAras:
Another component:
**Toasts:**
Toast are now rounder in iOS 14:
![image](https://user-images.githubusercontent.com/36629890/96368394-f2e53f00-1153-11eb-8b42-a68251546d51.png)

liamdebeasi:
Hi everyone,

We are going to be breaking these tasks out into separate issues and closing this big issue out.

Toast: https://github.com/ionic-team/ionic-framework/issues/22781
Spinner: https://github.com/ionic-team/ionic-framework/issues/22782
Refresher: https://github.com/ionic-team/ionic-framework/issues/22783
DateTime: https://github.com/ionic-team/ionic-framework/issues/16630 (Most of the specs for this are being tracked internally, but we are planning big improvements to datetime that should resolve most of the outstanding issues and update the designs)

Regarding the nav history and context menu components, we do not have plans to implement those at this time.

Thanks!

PR #22676, opened 2020-12-14
feat(toast): Change toast styles to latest iOS 14 styles

BerkeAras:
## Pull request checklist

## Pull request type

 

 

- [ ] Bugfix
- [X] Feature
- [ ] Build related changes

Issue Number: #22148

## What is the new behavior?

- The border-radius of toasts on iOS are now rounded (like iOS 14 style).
- The new property 'centeredText' has been added (iOS 14 style too).

- [ ] Yes
- [X] No


core/src/components.d.ts
```diff
@@ -2535,6 +2535,10 @@ export namespace Components {
           * An array of buttons for the toast.
          */
         "buttons"?: (ToastButton | string)[];
+        /**
+          * If `true`, the text of the toast will be centered.
+         */
+        "centeredText": boolean;
```
liamdebeasi, line 2541: This prop should not be needed. If we just set `text-align: center` on `:host` in `toast.ios.scss` that should be enough. Users can override it if they want the text aligned another way.

liamdebeasi (changes requested):
Thanks for the PR. Does this also make the toast appear at the top by default?

Just as a heads up: We are adding iOS 14 style updates in the next major release of Ionic Framework, so this will not get merged until then.

core/src/components.d.ts
liamdebeasi: Actually not sure how this affects buttons, but either way we shouldn't need a prop to do this.

BerkeAras: @liamdebeasi Buttons are centered too.

BerkeAras: @liamdebeasi See commit dc709a5.
The property **centeredText** has been removed. Text will be centered automatically and the font will be bold on iOS.

core/api.txt
```diff
@@ -1259,7 +1259,6 @@ ion-title,css-prop,--color
 ion-toast,shadow
 ion-toast,prop,animated,boolean,true,false,false
 ion-toast,prop,buttons,(string | ToastButton)[] | undefined,undefined,false,false
-ion-toast,prop,centeredText,boolean,false,false,false
```
BerkeAras: @liamdebeasi Property has been removed

core/src/components.d.ts
liamdebeasi: > @liamdebeasi Buttons are centered too.

Do you know if they are centered on native iOS 14 apps?

core/src/components/toast/toast.ios.scss
```diff
@@ -22,6 +22,10 @@
   display: block;
   position: absolute;
 
+  font-weight: 600;
```
liamdebeasi, line 25: Can we move this to a sass variable in `toast.ios.vars.scss`?

BerkeAras (commented):
@liamdebeasi Yes, toasts are centered on native iOS 14 too:
image

core/src/components/toast/toast.ios.scss
BerkeAras: @liamdebeasi What about chaning the font-weight of the header to 600 too?
It does not look good if the header is thinner than everything else.

appct:
Excellent. Thank you.

liamdebeasi:
Thanks for the issue! We appreciate the work you put into creating this PR. After discussion with the team, we have decided not to merge the PR.

We did some investigation and found that this notification on iOS is used in privacy-focused system notifications. In other words, iOS sends this notification, not the app you are running. For example, this notification will appear when an app reads from your clipboard, camera roll, etc. It also appears when peripherals such as AirPods are connected.

We feel that updating our `ion-toast` style to mimic this system notification would lead to a confusing user experience and also risk apps getting rejected from the iOS App Store given the privacy-focused nature of the native notifications.

I am going to close this PR and the associated issue, but please let me know if you have any questions. Thank you!

Issue #22694, opened 2020-12-18
bug: ion-skeleton-text animation is not correct


danielehrhardt:
# Bug Report

**Ionic version:**

[ ] **4.x**
[x] **5.x**

**Current behavior:**
If the skeleton text is used on a device like a Tablet it does flicker in the Middle of the Animation
![captured](https://user-images.githubusercontent.com/19570043/102628234-a44f2380-4149-11eb-8f66-3228667830ff.gif)

**Expected behavior:**
Smooth Animation no flickering.

**Steps to reproduce:**
`<ion-skeleton-text [animated]="true" ></ion-skeleton-text>`
Use this on a FullHD Browser window. 

**Related code:**
https://stackblitz.com/edit/skeleton-flickering

**Ionic info:**

```
   Ionic CLI                     : 6.12.2 (C:\Users\Daniel\AppData\Roaming\npm\node_modules\@ionic\cli)
   Ionic Framework               : @ionic/angular 5.5.1
   @angular-devkit/build-angular : 0.1100.3
   @angular-devkit/schematics    : 11.0.3
   @angular/cli                  : 11.0.3
   @ionic/angular-toolkit        : 3.0.0

Capacitor:

   Capacitor CLI   : 2.4.4
   @capacitor/core : 2.4.4

Utility:

   cordova-res : 0.15.2
   native-run  : not installed

System:

   NodeJS : v14.15.1 (C:\Program Files\nodejs\node.exe)
   npm    : 6.14.9
   OS     : Windows 10
```

patrickbussmann:
Same issue here (on same project) 😀
(but with MacBook Pro Safari & Chrome)

liamdebeasi:
Thanks for the issue. What device are you testing this on? I cannot reproduce this on my end.

danielehrhardt:
> Thanks for the issue. What device are you testing this on? I cannot reproduce this on my end.

You can see it in Google Chrome, Firefox, Edge. 
https://skeleton-flickering.stackblitz.io

danielehrhardt:
If it is above the 800px 
![image](https://user-images.githubusercontent.com/19570043/102630943-7a97fb80-414d-11eb-8049-d79c6c2a4f30.png)

liamdebeasi:
Thanks for the browser info. What physical device are you testing this on? (MacBook Pro, iPad, etc)

I tested this in Chrome on a MacBook Pro (using the internal display as well as an external display) and cannot reproduce the issue.

[...]

PR #22697, opened 2020-12-18
fix(skeleton-text): animation no longer jumps on large skeleton text elements

danielehrhardt:
## Pull request checklist

- [ X] Docs have been reviewed and added / updated if needed (for bug fixes / features)
- [ X] Build (`npm run build`) was run locally and any changes were pushed
- [X ] Lint (`npm run lint`) has passed locally and any fixes were made for failures

## Pull request type

 

 

- [X ] Bugfix
- [ ] Feature
- [ ] Build related changes

https://github.com/ionic-team/ionic-framework/issues/22694

Issue Number: #22694

## What is the new behavior?
- Smooth Animation even if the width is more then 800px

- [ ] Yes
- [X ] No

### DIFF
--- a/core/src/components/skeleton-text/skeleton-text.scss
+++ b/core/src/components/skeleton-text/skeleton-text.scss
@@ -56,13 +56,12 @@ span {
     left: 0;
     transform: translateX(-100%);
     background: linear-gradient(
-      90deg,
-      transparent 0,
-      $skeleton-text-background 20%,
-      $skeleton-text-background-animated 60%,
-      transparent
+      to right,
+      $skeleton-text-background 8%,
+      $skeleton-text-background-animated 18%,
+      $skeleton-text-background 33%
     );
-    animation: shimmer 2s infinite;
+    animation: shimmer 1s infinite;
     content: "";
   }
 }
--- a/core/src/components/skeleton-text/skeleton-text.scss
+++ b/core/src/components/skeleton-text/skeleton-text.scss
@@ -46,24 +46,31 @@ span {
 
 :host(.skeleton-text-animated) {
   position: relative;
+  overflow: hidden;
 
-  background: linear-gradient(to right, $skeleton-text-background 8%, $skeleton-text-background-animated 18%, $skeleton-text-background 33%);
-  background-size: 800px 104px;
-  animation-duration: 1s;
-  animation-fill-mode: forwards;
-  animation-iteration-count: infinite;
-  animation-name: shimmer;
-  animation-timing-function: linear;
+  &::after {
+    position: absolute;
+    top: 0;
+    right: 0;
+    bottom: 0;
+    left: 0;
+    transform: translateX(-100%);
+    background: linear-gradient(
+      90deg,
+      transparent 0,
+      $skeleton-text-background 20%,
+      $skeleton-text-background-animated 60%,
+      transparent
+    );
+    animation: shimmer 2s infinite;
+    content: "";
+  }
 }
 
 /* stylelint-disable property-blacklist */
 @keyframes shimmer {
-  0% {
-    background-position: -468px 0
-  }
-
   100% {
-    background-position: 468px 0
+    transform: translateX(100%);
   }
 }
 /* stylelint-enable property-blacklist */
--- a/core/src/components/skeleton-text/skeleton-text.scss
+++ b/core/src/components/skeleton-text/skeleton-text.scss
@@ -46,23 +46,29 @@ span {
 
 :host(.skeleton-text-animated) {
   position: relative;
+
   overflow: hidden;
 
   &::after {
     position: absolute;
+    /* stylelint-disable property-blacklist */
     top: 0;
     right: 0;
     bottom: 0;
     left: 0;
+    /* stylelint-enable property-blacklist */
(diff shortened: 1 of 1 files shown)


core/src/components/skeleton-text/skeleton-text.scss
```diff
@@ -46,24 +46,31 @@ span {
...
+    left: 0;
+    transform: translateX(-100%);
+    background: linear-gradient(
+      90deg,
+      transparent 0,
+      $skeleton-text-background 20%,
+      $skeleton-text-background-animated 60%,
+      transparent
+    );
+    animation: shimmer 2s infinite;
```
liamdebeasi, line 65:
```suggestion
    animation: shimmer 1s infinite;
```

The old animation was 1s long

liamdebeasi (changes requested):
Thanks for the PR! Two things I noticed:

1. The animation should be 1s, not 2s.
2. The "shimmer" seems to be larger than it was before.

I attached a comparison below (you may need to open the GIFs in a new tab to see the full size). The left GIF has your changes, and the right GIF is what is currently in `master`.

| branch | `master` |
|--------|----------|
| NEW | OLD |

core/src/components/skeleton-text/skeleton-text.scss
```diff
@@ -46,24 +46,31 @@ span {
 
 :host(.skeleton-text-animated) {
   position: relative;
-
-  background: linear-gradient(to right, $skeleton-text-background 8%, $skeleton-text-background-animated 18%, $skeleton-text-background 33%);
-  background-size: 800px 104px;
```
liamdebeasi, line 51: Are we able to add this back? The shimmer seems to be wider as a result of this removal.

danielehrhardt:
I changed the animation speed. 
I also changed the background gradient. But the old one somehow look wired. 
I attached two GIFs. 
Maybe you could ask your Team, what you prefer, send me a quick comment which one you want, so i can update the PR. 
1.
captured
2.
captured (1)

danielehrhardt:
@liamdebeasi ??

danielehrhardt:
@liamdebeasi Could you please check my last Question?

liamdebeasi:
> @liamdebeasi Could you please check my last Question?

Sorry, most of the team was out the past few weeks for winter holidays, so I didn't see your question.

Ideally this PR would keep the visual effect of the gradient exactly the same as shown in the `master` column in https://github.com/ionic-team/ionic-framework/pull/22697#pullrequestreview-555800059. Is there any reason why we can't keep the old linear-gradient we had set?

liamdebeasi:
Thanks, I took a look and some of the issues we discussed previously still persist (mainly the GIFs in https://github.com/ionic-team/ionic-framework/pull/22697#pullrequestreview-555800059).

I think the solution here is simpler than originally expected. The background-size of the skeleton text is set to `800px`, but the shimmer animation moves the background a total of `936px` (468 * 2). Changing the shimmer animation to go from `-400px` to `400px` seems to fix the animation on my end.

Can you give that a shot and let me know if it looks good to you? So the animation should look something like this:

```css
@keyframes shimmer {
  0% {
    background-position: -400px 0
  }

  100% {
    background-position: 400px 0
  }
}
```

If it looks good then I think we can revert the changes made here and just include the CSS Animation change.

danielehrhardt:
> Thanks, I took a look and some of the issues we discussed previously still persist (mainly the GIFs in [#22697 (review)](https://github.com/ionic-team/ionic-framework/pull/22697#pullrequestreview-555800059)).
> 
> I think the solution here is simpler than originally expected. The background-size of the skeleton text is set to `800px`, but the shimmer animation moves the background a total of `936px` (468 * 2). Changing the shimmer animation to go from `-400px` to `400px` seems to fix the animation on my end.
> 
> Can you give that a shot and let me know if it looks good to you? So the animation should look something like this:
> 
> ```css
> @keyframes shimmer {
>   0% {
>     background-position: -400px 0
>   }
> 
>   100% {
>     background-position: 400px 0
>   }
> }
> ```
> 
> If it looks good then I think we can revert the changes made here and just include the CSS Animation change.

In general on lager Screens the Shimmer appears two times because of the fixed Size of 800px. But it looks good to me. 
Should i change the PR or do you edit it directly?

liamdebeasi:
> In general on lager Screens the Shimmer appears two times because of the fixed Size of 800px.

I think this is intentional, but it might be worth revisiting the designs for this in the future.

> But it looks good to me. Should i change the PR or do you edit it directly?

If you could make the change that would be great. I do not have write access to this PR.

danielehrhardt:
> > In general on lager Screens the Shimmer appears two times because of the fixed Size of 800px.
> 
> I think this is intentional, but it might be worth revisiting the designs for this in the future.
> 
> > But it looks good to me. Should i change the PR or do you edit it directly?
> 
> If you could make the change that would be great. I do not have write access to this PR.

Changed the Code Back to Original with the 400px change.

danielehrhardt:
@liamdebeasi Could you check and merge?

liamdebeasi (approved):

liamdebeasi:
Merged. Thank you!

PR #22701, opened 2020-12-18
feat(breadcrumbs): add breadcrumbs component

### LINKED ISSUE #22770: feat: Breadcrumbs Component
# Feature Request

**Ionic version:**

[ ] **4.x**
[x] **5.x**

Add a Breadcrumbs component for desktop

Examples:
- https://material-ui.com/components/breadcrumbs/
- https://vuetifyjs.com/en/components/breadcrumbs/#usage
- https://www.lightningdesignsystem.com/components/breadcrumbs/

**Related Code**

**Additional Context**

image

brandyscarney:
## Pull request checklist

## Pull request type

 

 

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes

Issue Number: resolves #22770

TODO:
- [ ] iOS custom separator changes separator icon color on focus 
- [ ] Collapsed indicator as a div


ocarreterom:
Any screenshot showing the final result? 🤗

BerkeAras:
@brandyscarney Thanks for your afford :D
Could you add some screenshots?
https://screenshot.ionicframework.com/8888328

BerkeAras:
Are there any Screenshots yet?

core/src/components/breadcrumbs/test/basic/index.html
```diff
@@ -0,0 +1,238 @@
...
+          Prime Lenses
+        </ion-breadcrumb>
+        <ion-breadcrumb>
+          Product Info
+        </ion-breadcrumb>
+      </ion-breadcrumbs>
+
+      <p>Breadcrumbs: collapsed middle</p>
+
+      <ion-breadcrumbs>
```
liamdebeasi, line 200: When I set `color="danger"` on `ion-breadcrumbs`, it only gets applied to "Product Info", not "Home". Is this expected? When I set `color="danger"` on the `ion-breadcrumb` for "Home", it works as expected.

core/src/components/breadcrumbs/breadcrumbs.scss
```diff
@@ -0,0 +1,46 @@
...
+// Breadcrumbs
+// --------------------------------------------------
+
+:host {
+  /**
+   * @prop --background: Background of the breadcrumbs
+   * @prop --color: Text color of the breadcrumbs
+   */
+  @include font-smoothing();
+  @include padding($breadcrumbs-padding-top, $breadcrumbs-padding-end, $breadcrumbs-padding-bottom, $breadcrumbs-padding-start);
```
liamdebeasi, line 12: We chatted about this earlier, but we should adjust this padding when used in `ion-toolbar`. Optionally remove altogether in favor of having devs use the `ion-padding` class when used in `ion-content`.

liamdebeasi (commented):
This looks great! Is it possible to customize color/background-color of the three dots? We might need a shadow part for this on `ion-breadcrumbs`.

core/src/components/breadcrumbs/test/basic/index.html
brandyscarney: @liamdebeasi Yeah Ben wanted it to only change the color of the active breadcrumb. But if you apply it to individual breadcrumbs it would change their color.

BerkeAras:
@ocarreterom 
localhost_3333_src_components_breadcrumbs_test_basic(iPhone X)

appct:
Excellent. Thank you.

liamdebeasi (commented):
This is looking great! Some observations that we might want to address:

1. Hovering over a disabled breadcrumb still changes its color on iOS mode.
2. We might want to consider doing `display: none` on the collapsed breadcrumbs rather than removing them from the DOM altogether because there is a flicker when adding them back:

https://user-images.githubusercontent.com/2721089/113343718-2ed52080-92fe-11eb-8233-2d928a97fb5f.mov

3. The tab ordering on "collapsed middle / expand on click" seems odd. This is the tab order I am receiving: `Re-collapse button --> ... Icon --> Home`. I would expect the the `Home` breadcrumb to be tabbed to before the `... Icon`.
4. On an iPhone, holding down on one of the breadcrumbs causes a space next to one of the divider icons to be selected. I believe we have `user-select: none` set, so not sure why this would be happening.
5. Should the left edge of the "Home" text line up with the title here? The `padding-left` values are the same but look like there is padding on the inner `a` in `ion-breadcrumb`.
image

I did not test screen readers, but I can do that if you want me to.

core/src/components/breadcrumb/breadcrumb.tsx
```diff
@@ -0,0 +1,198 @@
...
+  /**
+   * Emitted when the breadcrumb loses focus.
+   */
+  @Event() ionBlur!: EventEmitter<void>;
+
+  /**
+   * Emitted when the collapse property changes.
+   * @internal
+   */
+  @Event() ionCollapsed!: EventEmitter<StyleEventDetail>;
```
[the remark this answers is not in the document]
BerkeAras: Yeah, why not. Use `ionCollapse` and `ionCollapseClick` 👍

core/package-lock.json
```diff
@@ -16071,7 +16071,9 @@
           "integrity": "sha1-w1se+RjsPJkPmlvFe+BKrOxcgRY=",
           "dev": true,
           "requires": {
-            "is-descriptor": "^0.1.0"
+            "is-accessor-descriptor": "^0.1.6",
```
liamdebeasi, line 16074: Can we revert these package-lock changes if they aren't needed?

core/src/components.d.ts
```diff
@@ -5,8 +5,10 @@
...
 import { IonicSafeString } from "./utils/sanitization";
+import { RouterDirection } from "./components/router/utils/interface";
+import { AnimationBuilder as AnimationBuilder1 } from "./utils/animation/animation-interface";
```
liamdebeasi, line 11: Do you know why this is getting imported again? Maybe due to the package-lock changes?

core/src/components/breadcrumb/breadcrumb.scss
```diff
@@ -0,0 +1,139 @@
+@import "./breadcrumb.vars";
+
+// Breadcrumb
+// --------------------------------------------------
+
+:host {
+ /**
+   * @prop --color: Color of the breadcrumb
```
liamdebeasi, line 8:
```suggestion
   * @prop --color: Text color of the breadcrumb
```

core/src/components/breadcrumb/breadcrumb.scss
```diff
@@ -0,0 +1,139 @@
...
+
+:host(.ion-focused) .button-native {
+  border: 1px solid var(--color);
+  // color: var(--color-focused);
+
+  // &::after {
+  //   background: var(--background-focused);
+
+  //   opacity: var(--background-focused-opacity);
+  // }
```
liamdebeasi, line 105: Is this still needed?

core/src/components/breadcrumb/breadcrumb.tsx
```diff
@@ -0,0 +1,212 @@
...
+        </TagType>
+        { showCollapsedIndicator &&
+            <button
+              part="collapsed-indicator"
+              onClick={() => this.collapsedIndicatorClick()}
+              class={{
+                'breadcrumbs-collapsed-indicator': true,
+              }}
+            >
+              <ion-icon name="ellipsis-horizontal" lazy={false}></ion-icon>
```
liamdebeasi, line 196: We should opt to use the ES6 import instead so that icons work properly with React and Vue. In other components we provided the string name here and then manually bundled the icons used in the main chunk of Ionic React and Vue, but if we do the ES6 import here it _should_ allow us to not have to manually bundle the icons and allow treeshaking to work better.

```suggestion
              <ion-icon icon={ellipsisHorizontal} lazy={false}></ion-icon>
```
`ellipsisHorizontal` is imported from `ionicons/icons`.

core/src/components/breadcrumb/breadcrumb.tsx
```diff
@@ -0,0 +1,212 @@
...
+            </button>
+        }
+        { showSeparator &&
+          <span class="breadcrumb-separator" part="separator">
+            <slot name="separator">
+              { mode === 'ios'
+                ? <ion-icon name="chevron-forward-outline" lazy={false}></ion-icon>
```
liamdebeasi, line 203: We should opt to use the ES6 import instead so that icons work properly with React and Vue. In other components we provided the string name here and then manually bundled the icons used in the main chunk of Ionic React and Vue, but if we do the ES6 import here it _should_ allow us to not have to manually bundle the icons and allow treeshaking to work better.

```suggestion
              <ion-icon icon={chevronForwardOutline} lazy={false}></ion-icon>
```
`chevronForwardOutline` is imported from `ionicons/icons`.

core/src/components/breadcrumb/breadcrumb.md.scss
```diff
@@ -0,0 +1,92 @@
...
+
+  box-shadow: 0px 1px 2px rgba(0, 0, 0, 0.2), 0px 2px 8px rgba(0, 0, 0, 0.12);
+}
+
+
+// Breadcrumb Slotted Icons
+// ------------------------------------------
+
+::slotted(ion-icon) {
+  color: #7d8894;
```
liamdebeasi, line 59: Along with the other hardcoded hex values, can we update this to work with the `color` property?

liamdebeasi (changes requested):
Overall this is looking really good. I left a few notes but 2 main things:

1. Looks like this branch is out of date with `next`. I see a few unrelated commits getting pulled in.
2. Most of the color values are hardcoded hex values which will not work with our theming system.

Another thing we might want to consider is if it makes sense for the breadcrumb component to determine its state by querying the parent breadcrumbs component similar to what we do for ion-radio: https://github.com/ionic-team/ionic-framework/blob/master/core/src/components/radio/radio.tsx#L125-L129

Or does it make sense to keep this top-down approach where the breadcrumbs component delegates the state to the child breadcrumb components?

core/src/components.d.ts
brandyscarney: I don't. Deleting it and re-building adds it every time and I reverted the package-lock changes.

liamdebeasi (approved):

Issue #22719, opened 2020-12-26
bug: ios collapse header should not be able to be between two states


EinfachHans:
# Bug Report

**Ionic version:**

[ ] **4.x**
[x] **5.x**

**Current behavior:**

Currently, the ios large (collapse) header is able to be between the two states when releasing the pointer via scroll. It then looks like: 

![image](https://user-images.githubusercontent.com/20899975/103143412-a733b080-4715-11eb-90ec-70c57934c8c6.jpeg) 

**Expected behavior:**

As far as i see, native it switches either to the expanded or collapsed state, determined by how far the transition was done when releasing.

EinfachHans:
Github Repo: https://github.com/EinfachHans/ionic-collapsable-header

Just the starter Repo with a List to make the content scrollable. Here an image of how it can look after releasing the finger:

<img width="393" alt="Bildschirmfoto 2020-12-29 um 00 26 44" src="https://user-images.githubusercontent.com/20899975/103248830-a6e12280-496c-11eb-9403-3bb57eb499bf.png">

ocarreterom:
I made a first attempt to this feature. https://github.com/ionic-team/ionic-framework/pull/22737

mikelhamer:
Is this still relevant?

EinfachHans:
@mikelhamer yes, it is

mikelhamer:
@EinfachHans thanks. I reached out on #22737 to try to see where that PR left off. From the comments it seemed like a solution was in place, but then the PR was deleted.

PR #22737, opened 2021-01-05
fix(content): collapsible large title now snaps into place in ion-content

ocarreterom:
## Pull request checklist

## Pull request type

 

 

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

Issue Number: https://github.com/ionic-team/ionic-framework/issues/22719

## What is the new behavior?

This is a first attempt to add header scroll snap.

- [ ] Yes
- [ ] No

## Other information

- A css-only solution using `scroll-snap` API.
- All the content (except ion-header) should be wrapped in a container (`scroll-snap-content`), otherwise each children will snap.

### DIFF
--- a/core/src/components/content/content.tsx
+++ b/core/src/components/content/content.tsx
@@ -130,6 +130,11 @@ export class Content implements ComponentInterface {
       : forceOverscroll;
   }
 
+  private shouldScrollSnap() {
+    return (getIonMode(this) === 'ios' || isPlatform('ios')) &&
+      !!this.el.querySelector('ion-header[collapse="condense"]')
+  }
+
   private resize() {
     if (this.fullscreen) {
       readTask(() => this.readDimensions());
@@ -330,7 +335,7 @@ export class Content implements ComponentInterface {
             'scroll-x': scrollX,
             'scroll-y': scrollY,
             'overscroll': (scrollX || scrollY) && forceOverscroll,
-            'scroll-snap': isPlatform('ios')
+            'scroll-snap': this.shouldScrollSnap()
           }}
           ref={el => this.scrollEl = el!}
           onScroll={(this.scrollEvents) ? ev => this.onScroll(ev) : undefined}
--- a/core/src/components/content/content.tsx
+++ b/core/src/components/content/content.tsx
@@ -329,7 +329,8 @@ export class Content implements ComponentInterface {
             'inner-scroll': true,
             'scroll-x': scrollX,
             'scroll-y': scrollY,
-            'overscroll': (scrollX || scrollY) && forceOverscroll
+            'overscroll': (scrollX || scrollY) && forceOverscroll,
+            'scroll-snap': isPlatform('ios')
           }}
           ref={el => this.scrollEl = el!}
           onScroll={(this.scrollEvents) ? ev => this.onScroll(ev) : undefined}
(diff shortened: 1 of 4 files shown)


ocarreterom:
Questions:
- Should it be an optional feature? `<ion-content snap={true}>`?
- If it is enabled by default, should `<ion-content>` add the wrapper container automatically?

BerkeAras:
@ocarreterom Thank you for your contributing to Ionic Framework 🥳. It is a great idea to add a scroll-snap function! In my opinion, this should be **given by default**. What about the attribute `no-snap` to disable it?

core/src/components/content/content.tsx
```diff
@@ -329,7 +329,8 @@ export class Content implements ComponentInterface {
             'inner-scroll': true,
             'scroll-x': scrollX,
             'scroll-y': scrollY,
-            'overscroll': (scrollX || scrollY) && forceOverscroll
+            'overscroll': (scrollX || scrollY) && forceOverscroll,
+            'scroll-snap': isPlatform('ios')
```
liamdebeasi, line 333: Can we change this so it only activates when the collapsible large title is in use?

liamdebeasi (commented):
Thanks for the PR! As for whether or not this should be a feature, I think this is something that should be enabled automatically, but only with the collapsible large title. I consider this to be more of a bug fix than a feature since it should have been there to begin with.

core/src/components/content/test/scroll-snap/index.html
```diff
@@ -0,0 +1,39 @@
...
+    </ion-header>
+
+    <ion-content fullscreen>
+      <ion-header collapse="condense">
+        <ion-toolbar>
+          <ion-title size="large">Content - Scroll Snap</ion-title>
+        </ion-toolbar>
+      </ion-header>
+      <!-- It is required to group up the content -->
+      <div class="ion-padding scroll-snap-content">
```
liamdebeasi, line 29: Will developers need to add the "scroll-snap-content" class?

liamdebeasi:
Also let's make sure we test this on iOS 12 devices as well. CSS Scroll Snapping is still a bit buggy on WebKit, especially on older versions. I recall running into issues when I initially tried this when building the collapsible large title.

core/src/components/content/test/scroll-snap/index.html
liamdebeasi: > All the content (except ion-header) should be wrapped in a container (scroll-snap-content), otherwise each children will snap.

Just saw this in the original post. Is there a way we can do this without adding the class? The collapsible large title already takes a bit of work to wire up, so I would like to avoid adding more to that.

Maybe we can target only by `ion-header[collapse="condense"]` in the CSS?

ocarreterom: The CSS class is not required. I can move to a generic selector `::slotted(*)`, but a wrapper container is still required.

> Maybe we can target only by ion-header[collapse="condense"] in the CSS?

That could be controlled by `'scroll-snap': ionHeaderCondenseExists && isPlatform('ios')` but not resolves the needed of a wrapper.
`<ion-content>` children need `scroll-snap-align: start` since its default value is `scroll-snap-align: none` and it does not work as we want.

mikelhamer:
@ocarreterom I am taking a look at #22719 trying to get a feel for where we are with everything. Was there a reason this PR got deleted and not merged in as a solution? Anything I can do to help

ocarreterom:
Hi @mikelhamer I suppose I remove the fork due to inactivity in this PR. I don't know why this PR was not merged. Feel free to get my code (if you think its ok) and open a new PR 🤗

Issue #21291, opened 2020-05-13
bug: ion refresher not hiding on IOS


olaisen81:
# Bug Report

**Ionic version:**

[x] **5.x**

**Current behavior:** The refresher is not hiding in IOS until I tap anywhere on the screen, it hides itself as normal behavior if you touch outside of the ionic list component. If you tap down from ones of the item in the ion list it doesn't hide.
 
**Expected behavior:** The refresher should hide after calling event.target.complete() function

**Steps to reproduce:** I run the refresher and when I finish loading the content I call event.target.complete() function. The bug reproduce if you tap down slowly, i notice that if i tap down fast it works. I can't reproduce it on a new sidemenu projects, here it works, incredible. It seems like the problem is due to list fetching from DB, with static list it works.
![91385BCA-96B8-4E91-8D1A-B7C6B1FE2D4C](https://user-images.githubusercontent.com/20971348/81927079-571f3800-95e3-11ea-9ecf-dc4356853add.gif)

**Other information:**

**Ionic info:**

Ionic CLI                     : 6.9.1 (/usr/local/lib/node_modules/@ionic/cli)
   Ionic Framework               : @ionic/angular 5.1.0
   @angular-devkit/build-angular : 0.803.26
   @angular-devkit/schematics    : 8.3.26
   @angular/cli                  : 8.3.26
   @ionic/angular-toolkit        : 2.2.0

olaisen81:
**Code reproduction:**

**page.component.html**
```
<ion-header [translucent]="true">
    <ion-toolbar>
        <ion-buttons slot="start">
            <ion-menu-button></ion-menu-button>
        </ion-buttons>
        <ion-title>{{'interfaces.menu.notifiche' | translate}}</ion-title>
    </ion-toolbar>
</ion-header>
<ion-content [fullscreen]="true">
    <ion-refresher slot="fixed" (ionRefresh)="doRefresh($event)">
        <ion-refresher-content></ion-refresher-content>
    </ion-refresher>
    <ion-header collapse="condense">
        <ion-toolbar>
            <ion-title size="large">{{'interfaces.menu.notifiche' | translate}}</ion-title>
        </ion-toolbar>
    </ion-header>
    <ion-list>
        <ion-list-header lines="inset">
            <ion-label color="primary">{{'interfaces.menu.gestione' | translate}}</ion-label>
        </ion-list-header>
        <ion-item-sliding *ngFor="let notifica of this.listaNotifiche">
[...]

PR #22820, opened 2021-01-26
fix(refresher): end pointer event if pulling farther than pullMax

### LINKED ISSUE #22819: bug: removing target from DOM during gesture causes onEnd to not fire
# Bug Report

**Ionic version:**

[ ] **4.x**
[x] **5.x**

**Current behavior:**

When using `ion-card` and switching loading states while refreshing, scroll down after refresher completed triggers the refresh again.

**Expected behavior:**

Content is scrollable and refresh isn't triggered again.

**Steps to reproduce:**

* Open the example app
* Pull refresher down until refresh is triggered automatically (**!Important Tap on the card and pull down**)
* wait until refresher is closed
* swipe content from bottom to top
=> refresh is triggered again

example video: https://user-images.githubusercontent.com/7203239/105836301-4544c000-5fcd-11eb-9a91-b189a7c2aab4.mov

**Related code:**
Sample application: https://github.com/Falingorn/bug-ionic-refresher-example/tree/bug/refresher-snapback-refresh

**Other information:**

**Ionic info:**

```
Ionic:

   Ionic CLI : 6.12.3

Utility:

   cordova-res (update available: 0.15.2) : 0.15.1
   native-run (update available: 1.3.0)   : 1.0.0

System:

   NodeJS : v14.15.4
   npm    : 6.14.10
   OS     : macOS Catalina
```

Falingorn:
closes #22819

## Pull request checklist

## Pull request type

 

 

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

Issue Number: #22819 

## What is the new behavior?

Refresher doesn't call `beginRefresh` anymore if refresher is pulled farther than pullMax, instead state is set and gesture is finished manually. This results in the same refreshing behavior as pulling between pullMin and pullMax and additionally ends the gesture correctly.

- [ ] Yes
- [x] No

### DIFF
--- a/core/src/components/refresher/refresher.tsx
+++ b/core/src/components/refresher/refresher.tsx
@@ -612,8 +612,9 @@ export class Refresher implements ComponentInterface {
     }
 
     if (deltaY > this.pullMax) {
-      // they pulled farther than the max, so kick off the refresh
-      this.beginRefresh();
+      // they pulled farther than the max, so end the related gesture, which will kick off the refresh
+      this.state = RefresherState.Ready;
+      this.gesture?.end();
       return;
     }
 
--- a/core/src/utils/gesture/index.ts
+++ b/core/src/utils/gesture/index.ts
@@ -231,6 +231,9 @@ export const createGesture = (config: GestureConfig): Gesture => {
     destroy() {
       gesture.destroy();
       pointerEvents.destroy();
+    },
+    end() {
+      pointerUp(undefined);
     }
   };
 };
@@ -305,6 +308,7 @@ export type GestureCallback = (detail: GestureDetail) => boolean | void;
 export interface Gesture {
   enable(enable?: boolean): void;
   destroy(): void;
+  end(): void;
 }
 
 export interface GestureConfig {


liamdebeasi:
Thanks for the PR! I am not sure this is something we are able to accept at the moment. Our docs state that the current behavior is intended, so changing it would be a breaking change:

The docs describe `pullMax` as "The maximum distance of the pull until the refresher will automatically go into the refreshing state." (https://ionicframework.com/docs/api/refresher#properties).

If we are to change the behavior of `ion-refresher`, it would be helpful to understand why the current behavior is causing issues. Could you provide some more clarification around this?

Falingorn:
Hi @liamdebeasi,

you're welcome! 🙂 
I think, the behavior isn't affected in any way.  The `ion-refresher` works in the same way as before.
The `beginRefresh()` function is triggered through the `pointerUp` event (instead to call it directly), which then calls `onEnd()` function as the normal pull-to-refresh mechanic would do. https://github.com/ionic-team/ionic-framework/pull/22820/files#diff-b29397aad7f5dbb9aef05f5ce05a3bc7df6ae3858de670c6c4dcca0db23bc6a5R624-R626

> If we are to change the behavior of ion-refresher, it would be helpful to understand why the current behavior is causing issues. Could you provide some more clarification around this?
There already is a video in the linked issue. Maybe it will help to understand why the current behavior is causing issues. Link: https://user-images.githubusercontent.com/7203239/105836301-4544c000-5fcd-11eb-9a91-b189a7c2aab4.mov

appct:
Excellent. Thank you.

Falingorn:
Hi @liamdebeasi,

I have a question regarding the native refresher implementation and would really appreciate your help. For what reason does this condition exist? https://github.com/ionic-team/ionic-framework/blob/d5980354fa1672839fa7e89e8fc91cfa524e3f03/core/src/components/refresher/refresher.tsx#L463-L466

While implementing a native refresher I stumbled upon this bug https://github.com/ionic-team/ionic-framework/issues/21291. The workaround didn't work as expected, so I dug a bit deeper and found the lines mentioned above.
I prepared a solution, but the refresher is closed even if the pointer isn't removed and I'd like to know why this behaviour exists before I update my PR with the new solution. 😅 
Until that I made another branch with the new solution, so you can take a look at it here. https://github.com/Falingorn/ionic-framework/commit/0f204ad1baf179d62f5f93bd077ac2bd7f1dc467

Thanks in advance for your time and help!

liamdebeasi:
That line fixed an issue where the refresher was not properly dismissed when calling the `complete` method synchronously (I.e. `ionRefresh` is fired and the `complete` method is immediately called).

One thing that would be good is if you could help me figure out how to reproduce this issue in the first place as it does not reproduce for me. In https://github.com/ionic-team/ionic-framework/issues/22819 what version of iOS are you testing on?

Falingorn:
Hi @liamdebeasi,
thank you for the explanation and sorry for the delayed answer. I updated the sample application from #22819, which you can find here https://github.com/Falingorn/bug-ionic-refresher-example.

> In #22819 what version of iOS are you testing on?

Tested it again with iOS 14.6 and it's still reproducable. Here is a video which shows the native refresher bug. https://user-images.githubusercontent.com/7203239/123406723-7db3ee80-d5ab-11eb-98ae-5984b3a5dcdb.mov


**Steps to reproduce**
* Open the example app
* Tap on the card, hold and pull down until refresh is triggered automatically (**!Important Pull down while holding the card and not below**)
* wait until refresher is closed

ionic refresher:
* swipe content from bottom to top, as shown in the attached video in #22819
=> refresh is triggered again

native iOS refresher:
=> the refresher icon stays visible, as shown in the video

Hope that is understandable and will help to reproduce the issue.

Falingorn:
Hi @liamdebeasi,
Could you already reproduce the issue for you?

liamdebeasi:
Yes I can reproduce the issue. The reason why this bug is happening is that when an event's target is removed from the DOM, events will still be targeted at it but may not bubble anymore. So what's happening is `ion-card` is the target when you start the gesture, but the moment you remove `ion-card` from the DOM, the events no longer bubble.

As a result, this causes the `touchend` event to never bubble up to the listeners on `ion-content` resulting in `onEnd` not being fired.

As you noted in https://github.com/ionic-team/ionic-framework/pull/22820#issuecomment-868387447 there is another bug with the native refresher not hiding. This is caused by the same underlying problem of `onEnd` not firing because events no longer bubble.

Given that this impacts both types of refreshers and potentially other gestures, I am not sure that the proposed solution is a good long term solution, but I will try to look into how we can address this problem in other ways.

liamdebeasi:
So my first attempt at another solution is to have our gesture utility listen for the `touchend` event on the `event.target` that we receive from `touchstart`. This would allow us to clean up the gesture even if the target is removed from the DOM.

I _believe_ that the target will not change between touch start and touch end so this should be safe thing to do. I need to do some more testing on other gesture in Framework, but can you try the following dev build and let me know if it resolves the issue?

```
npm install @ionic/angular@5.7.0-dev.202107062210.635ac19
```

Falingorn:
Hi @liamdebeasi!

Indeed, the `@ionic/angular@5.7.0-dev.202107062210.635ac19` package resolves the issues. That's awesome, thank you so much for investigating into this! 🎉 🙏 

Just fyi, what I noticed with this dev version is the following error:
```
TypeError: el[add] is not a function. (In 'el[add](eventName, callback, listenerOpts)', 'el[add]' is undefined)
vendor.js:116843:10
```
This seems to be related to the `addEventListener` function in `GESTURE_CONTROLLER, createGesture`.

Again, thank you so much and have a great day!

liamdebeasi:
Thanks! Do you know how I can reproduce the error in the sample app? It might be that the `target` is undefined in some scenarios.

liamdebeasi:
I am going to close this in favor of https://github.com/ionic-team/ionic-framework/pull/23713, but I gave you co-author credit in the commit. Thanks for all the help!

mrahn24:
HI @liamdebeasi !

Sorry that I didn't answer, I was on vacation 😅 . Thanks for the better solution and the credits as well! 💪

PR #22863, opened 2021-02-03
feat(custom-elements): add experimental custom elements build 

adamdbradley:
`@ionic/core/components` package will be included in the npm distribution.


core/package.json
```diff
@@ -24,19 +24,20 @@
...
+    "components/",
     "css/",
+    "dist/",
     "hydrate/",
     "loader/"
   ],
   "dependencies": {
-    "ionicons": "^5.1.2",
+    "@stencil/core": "^2.4.0",
+    "ionicons": "^5.5.0-1",
```
liamdebeasi, line 35: Any chance there's a stable release of ionicons we can use?

liamdebeasi (commented):
Any chance you can create a dev build so we can do some tests in conference apps before we merge?

core/package.json
adamdbradley: Yeah thinking we can release 5.5.0 once we know the dev build is good to go

core/README.md
```diff
@@ -40,6 +40,24 @@ The `@ionic/core` package can by used in simple HTML, or by vanilla JavaScript w
 * [@ionic/angular](https://www.npmjs.com/package/@ionic/angular)
 
 
+## Custom Elements Build (Experimental)
```
vicb, line 43: Is there anything else to do beside calling `initialize()` and registering the components ?

I am trying this feature of 5.6 and have a few issues:
-The body is initially hidden[1],
- The modals, alert, ... would not render - I am using the `modalController` from `@ionic/core/components` and registering `ion-modal` (`IonModal`) + `ion-backdrop`,
- I have a `<ion-select>` that is not rendering (supposed to be a popover) 

[1] because of 
```
html:not(.hydrated) body {
    display: none;
}
```

vicb: Thanks for your super fast answer.

I have added the class to my html for now.

I'll dig a little bit in the overlay issue trying to find the root cause and I'll create an issue with a repro.

PR #22932, opened 2021-02-16
feat(range): add support for custom pin text

### LINKED ISSUE #22864: feat: format the pin text for ion-range
# Feature Request

**Ionic version:**
[x] **5.x**

Currently the pin always displays an integer value:

https://github.com/ionic-team/ionic-framework/blob/487349f02a41344db2478735d27bf79f2a1c99b3/core/src/components/range/range.tsx#L558

This is not very flexible and they are a lot of times when you would want something else:
- `35%`,
- `1.2`,
- `42px`,
- ...

I think that adding a `pin` slot with the current label as the default value would be a great solution:

```
{pin && <div class="range-pin" role="presentation" part="pin"><slot name="pin">{Math.round(value)}</slot></div>}
```

The current usages would still work the same by using the default value:

```
    <ion-range pin min="20" max="80" step="2">
      <ion-icon size="small" slot="start" name="sunny"></ion-icon>
      <ion-icon slot="end" name="sunny"></ion-icon>
    </ion-range>
```

But you can customize the text by doing

```
    <ion-range pin min="20" max="80" step="2" @ionChange=${e => this.value = e.detail.value}>
      <ion-icon size="small" slot="start" name="sunny"></ion-icon>
      <ion-icon slot="end" name="sunny"></ion-icon>
      <ion-text slot="pin">${this.value}%</ion-text>
    </ion-range>
```

Edit:
I have quickly hacked this approach and it seems to work well.

There is one caveat: `@ionChange=${e => this.value = e.detail.value}` could be debounced (by adding debounce="<ms>") and then the value would not be updated as quickly as the pin is moved.

Adding a `ionRawChange` event might do the trick (this event would not be debounced).

**Describe Alternatives**

One alternative could be to pass a formatter callback as property: `(value: number) => string`.

I think that the slotted approach is nicer.

**Additional Context**

If this feature makes sense and is accepted, I would be happy to work on the implementation.

Thanks for the great framework !

vicb:
## Pull request checklist

## Pull request type

 

 

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes

Issue Number: closes #22864

The pin text is fixed and set to `Math.round(value)`

## What is the new behavior?

- Added a `"pin"` slot to customize the pin text, i.e. `<span slot=pin>{format(value)}</slot>`.
- Added a `ionRawChange` event on top of the existing `ionChange` event as you want the pin text to immediately update to the current value. 

My use case is to display 1 fractional digit as in the added example (i.e. "1.3").
It could also be useful to format currency ("$123") or percentage ("42%").

- [ ] Yes
- [x] No

No:
- The pin slot defaults to the text before this CL (`Math.round(value)`),
- The `ionRawChange` is purely additional.

## Other information

This could also be implemented in a simpler manner by using a `formatPin: (number) => string` property.

It would be a little simpler both to implement and to use. It might be a little less flexible but would probably be sufficient. Let me know if you want me to implement that instead.

### DIFF
--- a/angular/src/directives/proxies.ts
+++ b/angular/src/directives/proxies.ts
@@ -562,13 +562,14 @@ export declare interface IonRange extends Components.IonRange {
 @Component({ selector: "ion-range", changeDetection: ChangeDetectionStrategy.OnPush, template: "<ng-content></ng-content>", inputs: ["color", "debounce", "disabled", "dualKnobs", "max", "min", "mode", "name", "pin", "snaps", "step", "ticks", "value"] })
 export class IonRange {
   ionChange!: EventEmitter<CustomEvent>;
+  ionRawChange!: EventEmitter<CustomEvent>;
   ionFocus!: EventEmitter<CustomEvent>;
   ionBlur!: EventEmitter<CustomEvent>;
   protected el: HTMLElement;
   constructor(c: ChangeDetectorRef, r: ElementRef, protected z: NgZone) {
     c.detach();
     this.el = r.nativeElement;
-    proxyOutputs(this, this.el, ["ionChange", "ionFocus", "ionBlur"]);
+    proxyOutputs(this, this.el, ["ionChange", "ionRawChange", "ionFocus", "ionBlur"]);
   }
 }
 export declare interface IonRefresher extends Components.IonRefresher {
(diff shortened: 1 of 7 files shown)


liamdebeasi:
Thanks for the PR! I am going to close this in favor of https://github.com/ionic-team/ionic-framework/pull/22972 as it adds an API for customizing the pin text via the `pinFormatter` property.

Issue #22936, opened 2021-02-17
bug(router): beforeEnter is not triggered for initial mavigation


vicb:
# Bug Report

**Ionic version:**

[ ] **4.x**
[x] **5.5.4**

**Current behavior:**

I have a simple project with two routes `/foo` and `/bar`.

When I navigate to them via `<ion-item button href=...>` the `beforeEnter` hook is called as expected.

However when I visit navigate to my project by typing the url `http://<site>/bar` then the `beforeEnter` hook is not called.

I think this behavior is not expected ? 
At least it could be a pain if you want to lazy-load the component in the `beforeEnter` hook (or check authentication, ...)

**Expected behavior:**

The `beforeEnter` hook should be called when directly navigating to the url, i.e. `http://<site>/bar`

**Steps to reproduce:**

- Navigate to https://runkit.com/vicb/ionic-router
- Open the console
- Click on go to bar / go to foo - you should see `beforeEnter` logs in the console
- Manually update the url to end in "/foo" or "/bar",
- The content of `<ion-nav>` is updated as expected but the `beforeEnter` is not triggered (no logs).

**Related code:**

<!-- If you are able to illustrate the bug or feature request with an example, please provide a sample application via one of 
https://runkit.com/vicb/ionic-router

liamdebeasi:
Thanks for the issue. Can you try the following dev build and let me know if it resolves the issue?

Using NPM:
```
npm install @ionic/core@5.6.0-dev.202102181507.9d44979
```

Using CDN:
```
<script type="module" src="https://cdn.jsdelivr.net/npm/@ionic/core@5.6.0-dev.202102181507.9d44979/dist/ionic/ionic.esm.js"></script>
<script nomodule src="https://cdn.jsdelivr.net/npm/@ionic/core@5.6.0-dev.202102181507.9d44979/dist/ionic/ionic.js"></script>
<link rel="stylesheet" href="https://cdn.jsdelivr.net/npm/@ionic/core@5.6.0-dev.202102181507.9d44979/css/ionic.bundle.css" />
```

vicb:
@liamdebeasi  Thanks a lot for the super quickly fixing it.

I have tested it in the attached runkit and in my app and it works great ! (no more need for the workaround I mentioned).

I left a couple of minor comments on the PR.

liamdebeasi:
[...]

PR #22943, opened 2021-02-18
fix(router): guards are now run when landing directly on a page

liamdebeasi:
## Pull request checklist

## Pull request type

 

 

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

Issue Number: resolves https://github.com/ionic-team/ionic-framework/issues/22936

## What is the new behavior?

- Ensure that guards are run when `ion-router` loads

- [ ] Yes
- [x] No

### DIFF
--- a/core/src/components/router/router.tsx
+++ b/core/src/components/router/router.tsx
@@ -61,7 +61,17 @@ export class Router implements ComponentInterface {
     await waitUntilNavNode();
     console.debug('[ion-router] found nav');
 
-    await this.onRoutesChanged();
+    const canProceed = await this.runGuards(this.getPath());
+    if (canProceed !== true) {
+      if (typeof canProceed === 'object') {
+        const path = parsePath(canProceed.redirect);
+        const queryString = canProceed.redirect.split('?')[1];
+        this.setPath(path, ROUTER_INTENT_NONE, queryString);
+        await this.writeNavStateRoot(path, ROUTER_INTENT_NONE);
+      }
+    } else {
+      await this.onRoutesChanged();
+    }
   }
 
   componentDidLoad() {
--- a/core/src/components/router/test/guards/basic.e2e.ts
+++ b/core/src/components/router/test/guards/basic.e2e.ts
@@ -1,0 +1,16 @@
+import { newE2EPage } from '@stencil/core/testing';
+
+test('router: guards - guards should be run on initial load', async () => {
+  const page = await newE2EPage({
+    url: '/src/components/router/test/guards#/guard-initial-page?ionic:_testing=true'
+  });
+
+  await page.waitForChanges();
+
+  await checkUrl(page, '#/child/1');
+});
+
+const checkUrl = async (page, url: string) => {
+  const getUrl = await page.url();
+  expect(getUrl).toContain(url);
+}
--- a/core/src/components/router/test/guards/index.html
+++ b/core/src/components/router/test/guards/index.html
@@ -80,9 +80,26 @@
       }
     }
 
+    class GuardInitialPage extends HTMLElement {
+      connectedCallback() {
+        this.innerHTML = `
+        <ion-header>
+          <ion-toolbar>
+            <ion-buttons>
+              <ion-back-button></ion-back-button>
+            </ion-buttons>
+            <ion-title>Guard Initial Page</ion-title>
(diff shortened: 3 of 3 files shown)


core/src/components/router/router.tsx
```diff
@@ -61,7 +61,17 @@ export class Router implements ComponentInterface {
     await waitUntilNavNode();
     console.debug('[ion-router] found nav');
 
-    await this.onRoutesChanged();
+    const canProceed = await this.runGuards(this.getPath());
+    if (canProceed !== true) {
+      if (typeof canProceed === 'object') {
+        const path = parsePath(canProceed.redirect);
+        const queryString = canProceed.redirect.split('?')[1];
```
vicb, line 68: What about including everything after the first `?` instead ?

It should be equivalent in most cases but it would avoid users shooting themselves if they include a non-escape `?` in the query string of in the fragment i.e `site.com?value=ab?cd#de?f`.

core/src/components/router/router.tsx
```diff
@@ -61,7 +61,17 @@ export class Router implements ComponentInterface {
     await waitUntilNavNode();
     console.debug('[ion-router] found nav');
```
vicb, line 62: The comment is off-topic to the issue but I noticed that the router has some console traces that could not be de-activated.

Is it intended ?

liamdebeasi: My guess is that they were added during the development of Ionic Framework v4 3+ years ago and never removed.  While they should not show up in production builds, we probably don't need them in development builds either.

core/src/components/router/router.tsx
liamdebeasi: Fixed!

core/src/components/router/router.tsx
```diff
@@ -132,13 +133,14 @@ export class Router implements ComponentInterface {
     console.debug('[ion-router] URL pushed -> updating nav', url, direction);
 
     let path = parsePath(url);
-    let queryString = url.split('?')[1];
+    let queryString = url.substring(url.indexOf('?') + 1);
```
vicb, line 136: I think this is wrong is `url.indexOf('?') == -1`

vicb: What about adding a `let targetUrl` which would be equal to `url` or `redirect` and call `parsePath` and compute the `queryString` only once before `setPath` ?

liamdebeasi: If `indexOf` returns `-1`, then it would be `url.substring(0)` which is just `url`.

vicb: > If `indexOf` returns `-1`, then it would be `url.substring(0)` which is just `url`.

Yeah I think this is a problem. It should be empty in this case, no ?

vicb: Also, would it make sense to propagate fragments ? `/route#fragment`

liamdebeasi: Oh hah I got my variables mixed up. Yes it should just be `undefined`.

core/src/components/router/router.tsx
vicb: Are you going to fix it in the PR or do you want me to submit an other PR specifically for that ?

liamdebeasi: I can remove it in this.

core/src/components/router/router.tsx
vicb: Thinking more about it handling fragments is not that simple... as they could be routes (if `use-hash=true`).

But then maybe there is a problem with the handling of query string, what if `site.com?dark=1#/route` ?

vicb:
@liamdebeasi do you have a plan to move this forward ?

The 2 related issues (#22936, #22944) are blocking me from using the router and I would love to help in order to have them solved.

Thanks.

liamdebeasi:
Yes, but I paused this to focus on wrapping up our 5.6.0 release. Once 5.6.0 ships I will finish this.

vicb:
Great, thanks @liamdebeasi 

I'm happy to help with testing and anything else. Do not hesitate to ping me if needed.

liamdebeasi:
Closing in favor of https://github.com/ionic-team/ionic-framework/pull/23123.

PR #22957, opened 2021-02-22
fix(progress-bar): use correct theme colors

### LINKED ISSUE #20098: bug: Progress bar doesn't use theme colors
# Bug Report

**Ionic version:**

[x] **4.x**

**Current behavior:**

Using an `ion-progress-bar` with a dark theme results in a white progress bar.

**Expected behavior:**

Progress bar would pick up the theme colors.

**Steps to reproduce:**

1. Create an app with a progress bar
2. Use the dark theme
3. Progress bar is a white strip

**Related code:**
I had to do the following to fix it:
```css
ion-progress-bar {
  --buffer-background: var(--ion-background-color);
  --background: var(--ion-background-color);
  --progress-background: var(--ion-color-step-400);
}
```

It seems like that should be the default.

**Ionic info:**

```
Ionic:

   Ionic CLI       : 5.4.12 
   Ionic Framework : @ionic/react 4.11.6

Capacitor:

   Capacitor CLI   : 1.2.1
   @capacitor/core : 1.3.0

Utility:

   cordova-res : not installed
   native-run  : not installed

System:

   NodeJS : v12.11.0 
   npm    : 6.13.4
   OS     : macOS Mojave
```

liamdebeasi:
TODO: Give co-author credit to domske

## Pull request checklist

## Pull request type

 

 

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

Issue Number: resolves #20098

## What is the new behavior?

- Don't place a white background on the buffer bar to allow theming to work properly

- [ ] Yes
- [x] No

## Other information

| Before | After |
| ---| ---|
| before | after |

### DIFF
--- a/core/src/components/progress-bar/progress-bar.scss
+++ b/core/src/components/progress-bar/progress-bar.scss
@@ -119,7 +119,7 @@
 .buffer-circles.buffer-circles-reversed {
   /* stylelint-disable property-blacklist */
   right: unset;
-  left: 0;
+  left: 2px;
 
   background-position: right;
   /* stylelint-enable property-blacklist */
@@ -127,7 +127,7 @@
 
 .buffer-circles {
   /* stylelint-disable property-blacklist */
-  right: 0;
+  right: -8px;
 
   left: unset;
   /* stylelint-enable property-blacklist */
--- a/core/src/components/progress-bar/progress-bar.scss
+++ b/core/src/components/progress-bar/progress-bar.scss
@@ -10,7 +10,7 @@
   * @prop --progress-background: Color of the progress bar
   * @prop --buffer-background: Color of the buffer bar
   */
-  --background: #{ion-color(primary, base, 0.2)};
+  --background: #{ion-color(primary, base, 0.3)};
   --progress-background: #{ion-color(primary, base)};
   --buffer-background: var(--background);
   display: block;
@@ -27,7 +27,7 @@
 
 :host(.ion-color) {
   --progress-background: #{current-color(base)};
-  --buffer-background: #{current-color(base, 0.2)};
+  --buffer-background: #{current-color(base, 0.3)};
 }
 
 // indeterminate has no progress-buffer-bar, so it will be added to the host
@@ -72,17 +72,10 @@
 }
 
 .progress-buffer-bar {
-  // It's currently here because --buffer-background has an alpha
-  // Otherwise the buffer circles would be seen through
-  background: #fff;
+  background: var(--buffer-background);
 
-  z-index: 1; // Make it behind the progress
-
-  &:before {
-    background: var(--buffer-background);
-
-    content: "";
-  }
+  // Make it behind the progress
+  z-index: 1;
 }
 
 // MD based animation on indeterminate type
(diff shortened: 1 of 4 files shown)


brandyscarney:
The dots appear pretty visibly on the regular buffer bar (but not reversed):

brandyscarney (approved):

domske:
Hmm, something went wrong. In my original commit, the dots are moved synchron with the progress bar. And the dots container was overflow hidden. So that the dots are smoothly appear and disappear on the edges. But with the current state it looks like this (The dots shines through. And are not animated sync.)

demo

Brandy, your GIF: is that an optical illusion or are the dots going in the wrong direction?

liamdebeasi:
Thanks, I forgot a transition. I will resolve that here: https://github.com/ionic-team/ionic-framework/pull/22964

Because we are animating `transform` and `width`, they won't be perfectly in sync, but the overlap/gap should not be nearly as noticeable after my PR.

For the dots disappearing, that has always been there. The issue is just more pronounced because the dots no longer go behind the buffer. One idea is to animate background position rather than transform, but there may be some performance issues associated with that.

domske:
Ok, but if the wrapper is overflow hidden, the dots will also not behind the buffer. (as in my original commit). 
demo
<sub>(This is from Vuetify, not from Ionic)</sub>

But I not yet viewed my orginal and the current state of the source code.  So please don't get confused by me. ^^

domske:
btw. the live reload does not update changes in scss. I have to restart on every change by `npm start`. Changes in jsx works. Is there a trick?

brandyscarney:
@domske Yes we run into similar issues with the stencil build, I believe it is this issue https://github.com/ionic-team/stencil/issues/2635

As a workaround if you make a change in `progress-bar.scss` you can update `progress-bar.ios.scss` to trigger the style changes.

domske:
@brandyscarney Thanks, It works. But updating the `.(ios|md).scss` collapses the styles of the site and I have to press the refresh button of the browser. So both solutions *sigh* anyway ...

domske:
Back to the issue: Don't set width / height on e.g. 100% when using left/right/top/bottom. It's a style conflict / overwrite. 
Example: 

```scss
left: -10px;
right: -10px;
width: 100%;
```
will not work. The element is now left: -10px and width 100% instead of 100% + 20px.

liamdebeasi:
I wonder if it makes sense to keep the container that was in the old PR, set `overflow: hidden` and just change the width of that container. It might make for a nicer effect.

The reason why we changed the PR was because it was animating `left` which has some performance implications. I'll see if there's a better way to do this later this week.

domske:
The only idea with the overflow-hidden container, is that the dots can move behind the progress bar without shining through.
So, without this container - and only change the width, results in a aprupt disapearing dot background. Because you cannot move some pixels forward without overlay the progress bar. ...
I test some other solution (simpler) at this moment. But it's a bit tricky this little component. ^^
Where do you benchmark the style performance? I know that transform and other 3d accelerated stuff is better. Aber is there a way to measure things like `left` vs `translateX` or `scaleX` vs `width`, etc?

liamdebeasi:
Yeah, I like that overflow hidden container idea.


Re: benchmarking style performance-- this is a shameless plug but I go over the tools I use here: https://www.youtube.com/watch?v=A5a15NrcGoM

Using `left` probably isn't the end of the world performance-wise but we prefer to stick with `transform` where possible.

domske:
So fixed #22965 😓 But feel free to modify.

liamdebeasi:
Thank you! I will take a look at the PR.

Issue #23029, opened 2021-03-09
bug(overlay): alert/modal not showing with element build (5.6.0)


vicb:
# Bug Report

**Ionic version:**

[x] **5.6.0**

**Behavior:**

The following code should display an alert with the custom element build introduced in 5.6.0

```js
  const alert = await alertController.create({...});
  await alert.present();
```

However `alert` is undefined and `await alert.present();` errors.

The root cause seems to be that `IonAlert.componentOnReady` is undefined then

```js
const componentOnReady = (el, callback) => {
  if (el.componentOnReady) {
    el.componentOnReady().then(callback);
  }
  else {
    callback();
  }
};
```

the callback is called from the else branch from `createOverlay`

```js
const createOverlay = (tagName, opts) => {
  /* tslint:disable-next-line */
  if (typeof customElements !== 'undefined') {
    return customElements.whenDefined(tagName).then(() => {
      const element = document.createElement(tagName);
      element.classList.add('overlay-hidden');
      // convert the passed in overlay options into props
      // that get passed down into the new overlay
      Object.assign(element, opts);
      // append the overlay element to the document body
      getAppRoot(document).appendChild(element);
      return new Promise(resolve => componentOnReady(element, resolve));
    });
  }
  return Promise.resolve();
};
```

resulting in the inner `Promise` being resolved to `undefined`.

See https://stackblitz.com/edit/js-7d4zxa?file=index.js for a repro.

An alert should be shown when clicking the button.

**Other information:**

- If you select the `<ion-alert>` via the dev tools and call its `present()` method then the alert is shown as expected.
- the same issue exists with the `modalController`.

liamdebeasi:
Thanks for the issue. Can you try the following dev build and let me know if it resolves the issue?

```
npm install @ionic/core@5.7.0-dev.202103121659.630848a
```

liamdebeasi:
Thanks for the issue. This has been resolved via https://github.com/ionic-team/ionic-framework/pull/23039, and a fix will be available in an upcoming release of Ionic Framework.

PR #23039, opened 2021-03-12
fix(): overlays now present correctly when using custom elements build

liamdebeasi:
## Pull request checklist

## Pull request type

 

 

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

Issue Number: resolves https://github.com/ionic-team/ionic-framework/issues/23029

## What is the new behavior?

- Element ref is passed in callback
- We now wait in an raf before calling callback with CE build

- [ ] Yes
- [x] No

### DIFF
--- a/core/src/utils/helpers.ts
+++ b/core/src/utils/helpers.ts
@@ -9,7 +9,7 @@ export const componentOnReady = (el: any, callback: any) => {
   if (el.componentOnReady) {
     el.componentOnReady().then(callback);
   } else {
-    callback();
+    raf(() => callback(el));
   }
 }
 
--- a/core/src/utils/helpers.ts
+++ b/core/src/utils/helpers.ts
@@ -19,7 +19,7 @@ declare const requestAnimationFrame: any;
  */
 export const componentOnReady = (el: any, callback: any) => {
   if (el.componentOnReady) {
-    el.componentOnReady().then((el: any) => callback(el));
+    el.componentOnReady().then((resolvedEl: any) => callback(resolvedEl));
   } else {
     raf(() => callback(el));
   }
--- a/core/src/utils/helpers.ts
+++ b/core/src/utils/helpers.ts
@@ -5,9 +5,21 @@ import { Side } from '../interface';
 declare const __zone_symbol__requestAnimationFrame: any;
 declare const requestAnimationFrame: any;
 
+/**
+ * Waits for a component to be ready for
+ * both custom element and non-custom element builds.
+ * If non-custom element build, el.componentOnReady
+ * will be used.
+ * For custom element builds, we wait a frame
+ * so that the inner contents of the component
+ * have a chance to render.
+ *
+ * Use this utility rather than calling
+ * el.componentOnReady yourself.
+ */
 export const componentOnReady = (el: any, callback: any) => {
   if (el.componentOnReady) {
-    el.componentOnReady().then(callback);
+    el.componentOnReady().then((el: any) => callback(el));
   } else {
     raf(() => callback(el));
   }
--- a/core/src/utils/test/ready.spec.ts
+++ b/core/src/utils/test/ready.spec.ts
@@ -9,7 +9,10 @@ describe('componentOnReady()', () => {
     });
 
     const component = document.createElement('hello-world');
-    componentOnReady(component, () => done())
+    componentOnReady(component, (el) => {
+      expect(el).toEqual(component);
+      done();
+    })
   });
 
   it('should correctly call callback for a lazy loaded component', async (done) => {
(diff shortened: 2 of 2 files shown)


core/src/utils/helpers.ts
```diff
@@ -9,7 +9,7 @@ export const componentOnReady = (el: any, callback: any) => {
   if (el.componentOnReady) {
     el.componentOnReady().then(callback);
   } else {
-    callback();
+    raf(() => callback(el));
```
vicb, line 12: Hey @liamdebeasi,
Thanks for the fix, it is working well in my app (using 5.7.0-dev.202103121659.630848a).

May I suggest adding a comment and a test ?
Thanks

core/src/utils/helpers.ts
```diff
@@ -9,7 +9,7 @@ export const componentOnReady = (el: any, callback: any) => {
   if (el.componentOnReady) {
     el.componentOnReady().then(callback);
```
vicb, line 10: Do you want to update this code too for consistency with below ?

```js
el.componentOnReady().then(() => callback(el));
```

core/src/utils/helpers.ts
liamdebeasi: Great glad it's working! Yeah I still need to write tests and add some docs for this. Just wanted to get this code up for some initial testing.

core/src/utils/helpers.ts
liamdebeasi: `el.componentOnReady()` returns `el` already, so we don't need to pass it in ourselves.

vicb: Sure.
However:
- I think it could make the code easier to understand,
- I generally prefer being explicit and list the parameters. So maybe `.then((el) => callback(el))`. Being implicit makes it easy to shot yourself in the foot sometimes. Might be personal preferences.

liamdebeasi: Yep fair point - I updated the code.

core/src/utils/test/ready.spec.ts
```diff
@@ -0,0 +1,33 @@
+import { componentOnReady } from '../helpers';
+
+describe('componentOnReady()', () => {
+  it('should correctly call callback for a custom element', async (done) => {
+    customElements.define('hello-world', class extends HTMLElement {
+      constructor() {
```
vicb, line 6: Is this empty constructor needed ?

core/src/utils/test/ready.spec.ts
```diff
@@ -0,0 +1,33 @@
...
+        super();
+      }
+    });
+
+    const component = document.createElement('hello-world');
+    componentOnReady(component, () => done())
```
vicb, line 12: Do you want to check that the component is passed to the callback (same below).

core/src/utils/test/ready.spec.ts
liamdebeasi: Yeah when you extend HTMLElement you need to call `super()`, so the constructor is needed.

vicb: Thanks.
I meant to double check but I forgot. TIL.

core/src/utils/test/ready.spec.ts
```diff
@@ -0,0 +1,33 @@
...
+  });
+
+  it('should correctly call callback for a lazy loaded component', async (done) => {
+    customElements.define('hello-world', class extends HTMLElement {
+      constructor() {
+        super();
+      }
+
+      componentOnReady() {
+        return new Promise((resolve) => {
```
vicb, line 22: Maybe there should be a check that `componentOnReady()` is called too ?
(The current test would pass even if not)

core/src/utils/test/ready.spec.ts
```diff
@@ -0,0 +1,42 @@
...
+  it('should correctly call callback for a custom element', (done) => {
+    customElements.define('hello-world', class extends HTMLElement {
+      constructor() {
+        super();
+      }
+    });
+
+    const component = document.createElement('hello-world');
+    componentOnReady(component, (el) => {
+      expect(el).toEqual(component);
```
vicb, line 13: I would use `toBe` instead of `toEqual` here (and below) ?

core/src/utils/test/ready.spec.ts
```diff
@@ -0,0 +1,42 @@
...
+
+describe('componentOnReady()', () => {
+  it('should correctly call callback for a custom element', (done) => {
+    customElements.define('hello-world', class extends HTMLElement {
+      constructor() {
+        super();
+      }
+    });
+
+    const component = document.createElement('hello-world');
```
vicb, line 11: nit: use "eager-component" / "lazy-component" instead of "hello-world" ?

core/src/utils/test/ready.spec.ts
liamdebeasi: Ah yeah I always forget about that distinction

core/src/utils/test/ready.spec.ts
liamdebeasi: The lazy vs non lazy difference is already noted in the name of the test, so not sure this change would make much of a difference.

vicb:
Thanks a lot for working on this @liamdebeasi 
The PR looks great 👍

PR #23083, opened 2021-03-22
test(cypress): migrate Angular tests to Cypress

brandyscarney:
✅ Migrates Angular test app to Cypress

Verified test fails properly on CircleCI when it should:

Screen Shot 2021-03-26 at 2 07 55 PM


angular/test/test-app/e2e/src/router-link.spec.js
```diff
@@ -0,0 +1,196 @@
+const { hasUncaughtExceptionCaptureCallback } = require("process");
```
liamdebeasi, line 1:
```suggestion
```

angular/test/test-app/e2e/src/slides.spec.js
```diff
@@ -0,0 +1,46 @@
+const { hasUncaughtExceptionCaptureCallback } = require("process");
```
liamdebeasi, line 1:
```suggestion
```

angular/test/test-app/e2e/src/modal.spec.js
```diff
@@ -0,0 +1,42 @@
...
+    cy.wait(800);
+
+    cy.get('app-modal-example h2').should('have.text', '123');
+    cy.get('app-modal-example h3').should('have.text', '321');
+
+    cy.get('#onWillDismiss').should('have.text', 'false');
+    cy.get('#onDidDismiss').should('have.text', 'false');
+
+    cy.get('#close-modal').click();
+    cy.wait(800);
```
liamdebeasi, line 17: Rather than add these artificial waits, we should really be checking if the modal no longer exists. This could be abstracted to a Cypress command:

```js
cy.get('ion-modal#my-id').should('not.exist');
```

I have similar commands in the Vue test app: https://github.com/ionic-team/ionic-framework/blob/master/packages/vue/test-app/tests/e2e/support/commands.js

Cypress has built-in waiting, so it should work even if there's an animation to dismiss the modal. In the event that CI is slow, the modal may not dismiss after `800ms` which would result in a test failure here.

Same goes for any of the other `cy.wait` commands -- we should prefer to check the existence/non-existence of a component rather than a fixed wait.

liamdebeasi (changes requested):

PR #23152, opened 2021-04-06
fix(react): support React refs (object and callback)

TuckerWhitehouse:
React refs can be an object, or a callback - callbacks are
commonly used by 3rd party libraries for combining multiple refs.
https://reactjs.org/docs/refs-and-the-dom.html#callback-refs

## Pull request checklist

## Pull request type

 

 

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

Issue Number: #23153

Providing a callback ref breaks internal assumptions that forwardedRefs will always be in the object format.

## What is the new behavior?

- Internal ref usage has been decoupled from forwardedRefs.
- Refs may be provided in object or callback format.

- [ ] Yes
- [x] No

### DIFF
--- a/packages/react/src/components/__tests__/createComponent.spec.tsx
+++ b/packages/react/src/components/__tests__/createComponent.spec.tsx
@@ -27,13 +27,23 @@ describe('createComponent - events', () => {
 });
 
 describe('createComponent - ref', () => {
-  test('should pass ref on to web component instance', () => {
+  test('should pass ref on to web component instance (RefObject)', () => {
     const ionButtonRef: React.RefObject<any> = React.createRef();
     const IonButton = createReactComponent<JSX.IonButton, HTMLIonButtonElement>('ion-button');
 
     const { getByText } = render(<IonButton ref={ionButtonRef}>ButtonNameA</IonButton>);
     const ionButtonItem = getByText('ButtonNameA');
     expect(ionButtonRef.current).toEqual(ionButtonItem);
+  });
+
+  test('should pass ref on to web component instance (RefCallback)', () => {
+    let current
+    const ionButtonRef: React.RefCallback<any> = value => current = value;
+    const IonButton = createReactComponent<JSX.IonButton, HTMLIonButtonElement>('ion-button');
+
+    const { getByText } = render(<IonButton ref={ionButtonRef}>ButtonNameA</IonButton>);
+    const ionButtonItem = getByText('ButtonNameA');
+    expect(current).toEqual(ionButtonItem);
   });
 });
 
--- a/packages/react/src/components/createControllerComponent.tsx
+++ b/packages/react/src/components/createControllerComponent.tsx
@@ -1,7 +1,7 @@
 import { OverlayEventDetail } from '@ionic/core';
 import React from 'react';
 
-import { attachProps } from './utils';
+import { attachProps, setRef } from './utils';
 
 interface OverlayBase extends HTMLElement {
   present: () => Promise<void>;
@@ -30,7 +30,7 @@ export const createControllerComponent = <
 
   type Props = OptionsType &
     ReactControllerProps & {
-      forwardedRef?: React.RefObject<OverlayType>;
+      forwardedRef?: React.ForwardedRef<OverlayType>;
     };
 
   class Overlay extends React.Component<Props> {
@@ -73,9 +73,7 @@ export const createControllerComponent = <
       if (this.props.onDidDismiss) {
         this.props.onDidDismiss(event);
       }
-      if (this.props.forwardedRef) {
-        (this.props.forwardedRef as any).current = undefined;
-      }
+      setRef(this.props.forwardedRef, null)
     }
 
     async present(prevProps?: Props) {
@@ -106,9 +104,7 @@ export const createControllerComponent = <
       // Check isOpen again since the value could have changed during the async call to controller.create
       // It's also possible for the component to have become unmounted.
       if (this.props.isOpen === true && this.isUnmounted === false) {
-        if (this.props.forwardedRef) {
-          (this.props.forwardedRef as any).current = this.overlay;
-        }
+        setRef(this.props.forwardedRef, this.overlay)
         await this.overlay.present();
       }
     }
--- a/packages/react/src/components/utils/index.tsx
+++ b/packages/react/src/components/utils/index.tsx
@@ -18,13 +18,32 @@ export const createForwardRef = <PropType, ElementType>(
 ) => {
   const forwardRef = (
     props: IonicReactExternalProps<PropType, ElementType>,
-    ref: React.Ref<ElementType>
+    ref: React.ForwardedRef<ElementType>
   ) => {
     return <ReactComponent {...props} forwardedRef={ref} />;
   };
   forwardRef.displayName = displayName;
 
   return React.forwardRef(forwardRef);
+};
+
+export const setRef = (ref: React.ForwardedRef<any> | React.Ref<any> | undefined, value: any) => {
+  if (typeof ref === 'function') {
+    ref(value)
+  } else if (ref != null) {
+    // Cast as a MutableRef so we can assign current
+    (ref as React.MutableRefObject<any>).current = value
+  }
+};
+
+export const mergeRefs = (
+  ...refs: (React.ForwardedRef<any> | React.Ref<any> | undefined)[]
+): React.RefCallback<any> => {
+  return (value: any) => {
+    refs.forEach(ref => {
+      setRef(ref, value)
+    })
+  }
 };
 
 export * from './attachProps';
--- a/packages/react/src/components/IonIcon.tsx
+++ b/packages/react/src/components/IonIcon.tsx
@@ -22,7 +22,7 @@ interface IonIconProps {
 }
 
 type InternalProps = IonIconProps & {
-  forwardedRef?: React.RefObject<HTMLIonIconElement>;
+  forwardedRef?: React.ForwardedRef<HTMLIonIconElement>;
 };
 
 class IonIconContainer extends React.PureComponent<InternalProps> {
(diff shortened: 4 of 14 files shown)


packages/react/src/components/createControllerComponent.tsx
```diff
@@ -107,7 +111,11 @@ export const createControllerComponent = <
       // It's also possible for the component to have become unmounted.
       if (this.props.isOpen === true && this.isUnmounted === false) {
         if (this.props.forwardedRef) {
-          (this.props.forwardedRef as any).current = this.overlay;
+          if (typeof this.props.forwardedRef === 'function') {
```
liamdebeasi, line 114: Can we extract this to a shared function in the `utils` file? Something like:

```js
const setRef = (ref, value = null) => {
  if (typeof ref === 'function') {
    ref(value);
  } else {
    ref.current = value
  }
};
```

That way we can use this in both `createControllerComponent.tsx` and `createOverlayComponent.tsx`.

packages/react/src/components/utils/index.tsx
```diff
@@ -27,6 +27,21 @@ export const createForwardRef = <PropType, ElementType>(
...
+export const mergeRefs = (
+  ...refs: (React.ForwardedRef<any> | React.Ref<any> | undefined)[]
+) => {
+  return (value: any) => {
+    refs.forEach(ref => {
+      if (typeof ref === 'function') {
+        ref(value)
+      } else if (ref != null) {
+        // Cast as a MutableRef so we can assign current
+        (ref as React.MutableRefObject<any>).current = value
```
liamdebeasi, line 39: We can probably use the helper function here as well (we may need to cast as MutableRef in the helper function)

packages/react/src/components/utils/index.tsx
```diff
@@ -27,6 +27,21 @@ export const createForwardRef = <PropType, ElementType>(
   return React.forwardRef(forwardRef);
 };
 
+export const mergeRefs = (
+  ...refs: (React.ForwardedRef<any> | React.Ref<any> | undefined)[]
+) => {
+  return (value: any) => {
+    refs.forEach(ref => {
```
liamdebeasi, line 34: Do we need to return anything here?

liamdebeasi (changes requested):
Great job! PR will be approved once the changes noted below are made.

Here is a dev build if you would like to test this in your app:

```
npm install @ionic/react@5.7.0-dev.202104151952.ec15e01 @ionic/react-router@5.7.0-dev.202104151952.ec15e01
```

packages/react/src/components/utils/index.tsx
TuckerWhitehouse: I don't believe react has any requirements or expectations for the return value of a callback ref, but I will double-check and confirm.

TuckerWhitehouse: Confirmed the docs referenced above show an example with no return value, and the typescript types show the ForwardedRef function format and RefCallback should return void.

TuckerWhitehouse:
Hey @liamdebeasi - I believe all feedback has been addressed in the updated commit, please let me know if there are any other required changes -- thanks for reviewing this!

packages/react/src/components/__tests__/createComponent.spec.tsx
```diff
@@ -27,14 +27,24 @@ describe('createComponent - events', () => {
...
+  test('should pass ref on to web component instance (RefObject)', () => {
     const ionButtonRef: React.RefObject<any> = React.createRef();
     const IonButton = createReactComponent<JSX.IonButton, HTMLIonButtonElement>('ion-button');
 
     const { getByText } = render(<IonButton ref={ionButtonRef}>ButtonNameA</IonButton>);
     const ionButtonItem = getByText('ButtonNameA');
     expect(ionButtonRef.current).toEqual(ionButtonItem);
   });
+
+  test('should pass ref on to web component instance (RefCallback)', () => {
```
liamdebeasi, line 39: We should add a test here to test that `onIonChange` is called at most once at a time.

liamdebeasi (changes requested):
I tested an updated dev build (`5.7.0-dev.202104161346.589a1b8`) and it looks like the `onIonChange` handler is being fired twice now. I am testing this with the CodeSandbox found in https://github.com/ionic-team/ionic-framework/issues/23153.

If I remove the `ref` or use a non-callback ref then the `onIonChange` handler only fires once. My guess is maybe it has something to do with the `mergeRefs` function.

Steps to reproduce:

1. Open CodeSandbox on issue I linked to with the dev build installed.
2. Type `a`. Observe that the console logs `onChange a`.
3. Type `b`. Observe that the console logs `onChange ab` twice.

packages/react/src/components/__tests__/createComponent.spec.tsx
TuckerWhitehouse: Hey @liamdebeasi - I'm unsure how to go about introducing this test. I looked throughout the project and couldn't find any examples where the web components were being rendered and their behaviors verified. If you have any pointers on how to get everything set up, I'd be more than happy to add this!

TuckerWhitehouse:
I've pushed another code change to ensure the refs remain stable between renders (I don't believe this is the cause of the issue described above, but it was another issue I found while debugging).

That being said, I'm not sure the double onIonChange is actually caused by these changes (it appears to happen with both dev builds as well). I tried to recreate an example without react-hook-form and was unable to, which makes me lean towards this being a bug in that library.

--

For reference, I tried various combinations of object refs, react refs, inline functions, inline functions that set refs, inline functions that returned refs, etc, and was unable to produce the double event behavior seen with the react-hook-form ref.

packages/react/src/components/__tests__/createComponent.spec.tsx
liamdebeasi: I haven't actually tested this, but I was thinking something like the following:

```jsx
let current;
const ionInputRef: React.RefCallback<any> = value => current = value;
const changeCallback = jest.fn();
const IonInput = createReactComponent<JSX.IonInput, HTMLIonInputElement>('ion-input');

const { getBy } = render(<IonInput id="my-input" onIonChange={changeCallback} ref={ionInputRef}></IonInput>);

const ionInputItem = getBy('#my-input');
ionInputItem.value = 'abc';
expect(changeCallback).toBeCalledTimes(1);
ionInputItem.value = 'abcxyz';
expect(changeCallback).toBeCalledTimes(2);
```

liamdebeasi:
Hmm ok I will take another look. If the issue is not in Ionic React then we probably do not need to add the other test I mentioned.

packages/react/src/components/__tests__/createComponent.spec.tsx
TuckerWhitehouse: Directly assigning to `.value` doesn't seem to trigger the `onIonChange` -- From what I could tell, this setup doesn't actually render the `ion-input` WebComponent.

I tried to use `fireEvent.change(ionInput, { target: { value: '' } }` but that throws "The given element does not have a value setter".

liamdebeasi: 🤔  Weird. Let's hold off on this I posted the suggestion before I saw https://github.com/ionic-team/ionic-framework/pull/23152#issuecomment-821281532, so I'll do some more digging and see if we even need to do this.

liamdebeasi (approved):
Ok so looks like if you type "a", the `react-hook-form` package sets that as the input's default value. Then when you type "ab", it sets the value to the default value and then sets it to "ab" for some reason. The actual Ionic event still shows "ab" as the value, but I agree with your original suggestion that this looks like a bug in the `react-hook-form`.

Code looks good to go. Thanks for putting up this PR!

liamdebeasi (commented):
One more question -- do we need to make these `React.ForwardedRef` as well?

IonTabButton: https://github.com/ionic-team/ionic-framework/blob/8e0e5da7407adecb7471b3a6b0ac059337761355/packages/react/src/components/navigation/IonTabButton.tsx#L11

IonBackButton: https://github.com/ionic-team/ionic-framework/blob/8e0e5da7407adecb7471b3a6b0ac059337761355/packages/react/src/components/navigation/IonBackButton.tsx#L16

TuckerWhitehouse:
Hey @liamdebeasi -- I updated the two components to use `React.Ref` instead of `React.RefObject`.
From what I can tell, neither of these components are currently setup to forward the ref to the inner component, so the ref is just going to point to the instance of the class.

I could probably update them to use the `createForwardRef` but that is a change in behavior for refs on those elements and I'm not sure if it makes sense to introduce what could be a breaking change with the other (non-breaking) support for callback refs.

liamdebeasi:
That all makes sense. I agree we should not introduce a breaking change with this. Thanks!

liamdebeasi:
Merged. Thank you very much for the PR!

TuckerWhitehouse:
Hey @liamdebeasi - Thanks for reviewing and helping to get this merged, looking forward to this being included in an upcoming release :D (now I just need to sort out that react-hook-form bug... haha)

liamdebeasi:
We just shipped v5.6.5 a few hours ago that includes this fix 😄

liamdebeasi:
@TuckerWhitehouse  Hey there, this seems to have caused a regression in https://github.com/ionic-team/ionic-framework/issues/23287. When an Ionic component re-renders, the ref value becomes `null`. I am working to determine how to fix, but wanted to check in and see if you had any ideas.

TuckerWhitehouse:
@liamdebeasi I believe the code referenced in that ticket is using `createRef` where they should be using `useRef` -- The `createRef` is going to create a new ref object on each render and React will set the ref to null whenever the value passed in as a `ref` changes (referential integrity and all).

liamdebeasi:
Ah good catch! In this case, should we be calling that `mergeRefs` function in `componentDidUpdate` rather than once in the `constructor`?

TuckerWhitehouse:
Yup, `createRef` is the util for classes, and `useRef` is the util for functions -- Looks like switching to `useRef` resolved the issue in #23287 :)

Issue #23002, opened 2021-03-03
feat: 


NLueg:
# Feature Request

**Ionic version:**

[x] **5.5.5**

**Describe the Feature Request**
Right now, if I add an ion-input inside an ion-item, it takes over the click function of the ion-item. If we add `button=false` on an ion-item, the expected result is that items placed within, such as a ion-input, will not make the ion-item work like a button.

**Describe Preferred Solution**
It should be possible to deactivate the `clickListener` of the `ion-item` in some cases. To be more precise this if should be modified with a new property or using the `button` attribute:  https://github.com/ionic-team/ionic-framework/blob/943e3f6ae37ecc56f21168f057dde77a05e4e144/core/src/components/item/item.tsx#L161.

**Describe Alternatives**
There is currently no other way to implement this behavior. 
On Android devices, it helps to call `stopPropagation` and `preventDefault` in the child-element that was clicked. But on iOS the click event of the ion-item is called anyway so in the following example the button event is triggered, but the `ion-input` is focused which is quite bad behavior. 

**Related Code**
```html
<ion-item>
  <ion-label slot="start"> {{ labelText }}</ion-label>
  <ion-input
    type="number"
    [(ngModel)]="amount"
    [placeholder]="placeholder"
  ></ion-input>
  <ion-button
    (click)="minusAmount($event)"
  >
    <ion-icon slot="icon-only" name="remove-circle-outline"></ion-icon>
  </ion-button>
  <ion-button (click)="plusAmount($event)">
    <ion-icon slot="icon-only" name="add-circle-outline"></ion-icon>
  </ion-button>
</ion-item>
``` 

```ts
    $event.stopPropagation();
    $event.preventDefault();
   ...
  }

  minusAmount($event: MouseEvent) {
    $event.stopPropagation();
    $event.preventDefault();
   ...
  }
```

**Additional Context**

https://github.com/ionic-team/ionic-framework/issues/19590
https://forum.ionicframework.com/t/clickable-button-inside-an-ion-item-whereas-the-ion-item-itself-is-also-clickable/79262/3

liamdebeasi:
Thanks for the issue. Can you please clarify the issue you are running into?

NLueg:
Sure. I made two gifs from android and ios which should show the bad behaviour of an ion-item with an ion-button and ion-input inside it. 
[...]

PR #23157, opened 2021-04-07
fix(item): adds delegate click property

NLueg:
## Pull request checklist

## Pull request type

 

 

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

Currently, a click on an item always delegates the focus to the first input inside the item.

Issue Number: #23002

## What is the new behavior?

- It's now possible to prevent the click-delegation by the introduced property

- [ ] Yes
- [x] No

### DIFF
--- a/core/src/components/item/item.tsx
+++ b/core/src/components/item/item.tsx
@@ -113,6 +113,11 @@ export class Item implements ComponentInterface, AnchorInterface, ButtonInterfac
    */
   @Prop() type: 'submit' | 'reset' | 'button' = 'button';
 
+  /**
+   * Specifies if the first input should be focused if the ion-item is clicked
+   */
+  @Prop() delegateClick: boolean = true;
+
   @Listen('ionColor')
   labelColorChanged(ev: CustomEvent<string>) {
     const { color } = this;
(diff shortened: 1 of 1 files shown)


liamdebeasi:
Thanks for the PR! I am not sure this is the correct solution to the issue. The first issue is that this PR does not disable the native delegatesFocus behavior built into the shadow root, so the behavior would persist on iOS. The second issue is being able to disable this delegatesFocus behavior means that tapping the `ion-label` would no longer focus the `ion-input` which might not be desirable.

I did some more digging and found that WebKit does not consider buttons to be focusable, so things are working as intended regarding https://github.com/ionic-team/ionic-framework/issues/23002. That being said, I do not think the behavior is desirable. I need to discuss with the team what our other options are.

NLueg:
@liamdebeasi thank you for your fast response. Actually, this PR fixes the issue for iOS (at least for the Safari). I checked it before. But I understand the other arguments.
Besides this PR I have no idea how the issue can be fixed. It's currently not possible to put a clickable like a button inside an `ion-item` with an `input` or `ion-input`.

PR #23160, opened 2021-04-07
fix(content): change main child element to div when parent is menu

### LINKED ISSUE #22808: bug: a11y ion-menu with ion-content duplicates main element
# Bug Report

**Ionic version:**
[ ] **4.x**
[x] **5.x**

**Current behavior:**
In an Ionic app that uses `ion-split-pane` (and/or is based off of the **sidemenu** template), duplicate `main` elements are created, one for each `ion-content` component. In the **sidemenu** template, `ion-content` is used within the split pane menu, as well as within the body of the page. 

This causes the app to fail accessibility checks, such as [axe browser extension](https://www.deque.com/axe/browser-extensions/). The issues axe finds are:

- Main landmark must not be contained in another landmark
- Document must not have more than one main landmark
- Ensures landmarks are unique

**Expected behavior:**
An app built off of the **sidemenu** (or any other) template should not fail an accessibility check due to duplicate `main` landmarks.

**Steps to reproduce:**
1. Create a new Ionic app with the **sidemenu** template, or clone [ionicAngular](https://github.com/Ksan8/ionicAngular) (a minimal Ionic Angular app)
2. Install the [axe browser extension](https://www.deque.com/axe/browser-extensions/) if not already installed
3. Run `ionic serve` from the app root to open the app in the browser
4. Open DevTools and click on the axe tab
5. Choose "Scan all of my page"
6. Notice the issues listed above appear in the axe results

**Related code:**

Sample Ionic Angular app: https://github.com/Ksan8/ionicAngular

**Other information:**
One workaround is to replace or remove the `ion-content` element within `ion-menu` in the split pane, and instead use a wrapper div or other parent component (such as `ion-list`). Adding a bit of CSS (`{flex-grow: 1; flex-shrink: 1}` will allow it to fill the rest of the side menu. A bit more tweaking may be needed in order to add other missing styling.

Other suggestions would be welcome.

**Ionic info:**
```
Ionic:

   Ionic CLI                     : 6.12.3 (/Users/keziawineberg/.nvm/versions/node/v12.13.1/lib/node_modules/@ionic/cli)
   Ionic Framework               : @ionic/angular 5.5.2
   @angular-devkit/build-angular : 0.1000.8
   @angular-devkit/schematics    : 10.0.8
   @angular/cli                  : 10.0.8
   @ionic/angular-toolkit        : 2.3.3

Capacitor:

   Capacitor CLI   : 2.4.6
   @capacitor/core : 2.4.6

Utility:

   cordova-res                          : 0.15.2
   native-run (update available: 1.3.0) : 1.2.2

System:

   NodeJS : v12.13.1 (/Users/keziawineberg/.nvm/versions/node/v12.13.1/bin/node)
   npm    : 6.14.10
   OS     : macOS Big Sur
```

Ionic team edit:

```
render() {
  const TagType = this.el.closest('ion-menu') !== null ? 'div' : 'main' as any;

  return (
    <Host>
      <TagType>
        <slot></slot>
      </TagType>
    </Host>
  );
}
```

willmartindev:
## Pull request checklist

## Pull request type

 

 

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

Issue Number: resolves #22808

## What is the new behavior?

- When `ion-content` is a child of `ion-menu`, child content is now wrapped in a `div` rather a `main` element.

- [ ] Yes
- [x] No

### DIFF
--- a/core/src/components/content/content.tsx
+++ b/core/src/components/content/content.tsx
@@ -307,6 +307,7 @@ export class Content implements ComponentInterface {
     const { scrollX, scrollY } = this;
     const mode = getIonMode(this);
     const forceOverscroll = this.shouldForceOverscroll();
+    const TagType = this.el.closest('ion-menu') !== null ? 'div' : 'main' as any;
     const transitionShadow = (mode === 'ios' && config.getBoolean('experimentalTransitionShadow', true));
 
     this.resize();
@@ -324,19 +325,19 @@ export class Content implements ComponentInterface {
         }}
       >
         <div id="background-content" part="background"></div>
-        <main
+        <TagType
           class={{
             'inner-scroll': true,
             'scroll-x': scrollX,
             'scroll-y': scrollY,
             'overscroll': (scrollX || scrollY) && forceOverscroll
           }}
-          ref={el => this.scrollEl = el!}
-          onScroll={(this.scrollEvents) ? ev => this.onScroll(ev) : undefined}
+          ref={(el: HTMLElement) => this.scrollEl = el!}
+          onScroll={(this.scrollEvents) ? (ev: UIEvent) => this.onScroll(ev) : undefined}
           part="scroll"
         >
           <slot></slot>
-        </main>
+        </TagType>
 
         {transitionShadow ? (
           <div class="transition-effect">
--- a/core/src/components/content/content.tsx
+++ b/core/src/components/content/content.tsx
@@ -307,7 +307,7 @@ export class Content implements ComponentInterface {
     const { scrollX, scrollY } = this;
     const mode = getIonMode(this);
     const forceOverscroll = this.shouldForceOverscroll();
-    const TagType = this.el.closest('ion-menu') !== null ? 'div' : 'main' as any;
+    const TagType = this.el.closest('ion-menu, ion-popover, ion-modal') !== null ? 'div' : 'main' as any;
     const transitionShadow = (mode === 'ios' && config.getBoolean('experimentalTransitionShadow', true));
 
     this.resize();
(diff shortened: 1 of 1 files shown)


core/src/components/content/content.tsx
```diff
@@ -307,6 +307,7 @@ export class Content implements ComponentInterface {
     const { scrollX, scrollY } = this;
     const mode = getIonMode(this);
     const forceOverscroll = this.shouldForceOverscroll();
+    const TagType = this.el.closest('ion-menu') !== null ? 'div' : 'main' as any;
```
liamdebeasi, line 310:
```suggestion
    const TagType = this.el.closest('ion-menu, ion-popover, ion-modal') !== null ? 'div' : 'main' as any;
```

We should account for this in popovers and modals as well since those are not "main" landmarks/they are temporary.

liamdebeasi (changes requested):

core/src/components/content/content.tsx
```diff
@@ -307,6 +307,7 @@ export class Content implements ComponentInterface {
     const { scrollX, scrollY } = this;
     const mode = getIonMode(this);
     const forceOverscroll = this.shouldForceOverscroll();
+    const TagType = this.el.closest('ion-menu, ion-popover, ion-modal') !== null ? 'div' : 'main' as any;
```
liamdebeasi, line 310: Two things here:

1. This probably doesn't need to be reactive and might result in a small performance hit. We can just do this query selector in `connectedCallback` and cache the result:

```tsx
private isMainContent = true;

...

connectedCallback() {
  this.isMainContent = this.el.closest('ion-menu, ion-popover, ion-modal') === null;
}

...

const { isMainContent } = this;

const TagType = isMainContent ? 'main' : 'div';
```

`connectedCallback` is called when `ion-content` is added to the DOM.

2. If possible, try to avoid using the `any` type.

liamdebeasi (changes requested):

PR #23194, opened 2021-04-15
fix(radio-group): enable unselect with spacebar

### LINKED ISSUE #22734: bug: Cannot unselect radio button with space bar with allowEmptySelection true
# Bug Report

**Ionic version:**

[ ] **4.x**
[x] **5.x**

**Current behavior:**

Using an ion-radio-group with allowEmptySelection set to "true", a user is able to unselect via touch and via mouse click, but cannot unselect via spacebar. This causes issues for a11y.

https://ionicframework.com/docs/api/radio-group#properties

**Expected behavior:**
When allowEmptySelection is "true" the user should be able to both select and unselect with a spacebar.
When allowEmptySelection is "false" (or unspecified), the user should only be able to select via spacebar.

**Related code:**

https://github.com/ionic-afl/ionic-v5-observations

willmartindev:
## Pull request checklist

## Pull request type

 

 

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

Issue Number: resolves #22734 

## What is the new behavior?

- Users can now unselect radio buttons with spacebar if `allowEmptySelection` is enabled on the radio group

- [ ] Yes
- [x] No


core/src/components/radio-group/radio-group.tsx
```diff
@@ -144,7 +144,9 @@ export class RadioGroup implements ComponentInterface {
       // space bar on top of a selected radio (only applies
       // to radios in a select popover)
```
liamdebeasi, line 145:
```suggestion
      // space bar on top of a selected radio
```

Remove that extra comment since it is no longer true

liamdebeasi (changes requested):
Looks good, but we need to add a test for this before we can merge. I would prefer that we do a spec test if possible since they run a bit faster than E2E tests.

You can add a `radio-group.spec.ts` file in `radio-group/test`. See https://github.com/ionic-team/ionic-framework/blob/next/core/src/components/accordion/test/accordion.spec.ts for examples of how I write the tests. Also see https://stenciljs.com/docs/unit-testing for the Stencil testing API for `newSpecPage`.

We should add a test for when the user presses spacebar on a radio group that has `allowEmptySelection` and a test for when the user presses spacebar on a radio group that does not have `allowEmptySelection`.

core/src/components/radio-group/test/radio-group.e2e.ts
```diff
@@ -0,0 +1,88 @@
...
+    await page.keyboard.press('Space');
+  } else if (selectionMethod === 'mouse') {
+    await page.click(selector);
+  }
+  
+  await page.waitForChanges();
+
+  const radioGroup = await page.find(`ion-radio#${radioButtonId} >>> input`);
+  const checked = await radioGroup.getProperty('checked');
+  return checked;
```
liamdebeasi, line 21: Any reason why we can't just look at the `checked` property on the host `ion-radio` element instead of diving into the shadow dom?

liamdebeasi: Discussed offline to keep test as is. `checked` is a state not a prop.

PR #23224, opened 2021-04-22
fix(react): remove hardware back button event listener when NavManager is unmounted

### LINKED ISSUE #23170: bug: Hardware back event triggered twice (or more) if an IonReactRouter is unmounted and re-mounted
# Bug Report

**Ionic version:**

[ ] **4.x**
[x] **5.x**

**Current behavior:**
Hardware back button triggers the back event multiple times.

**Expected behavior:**
Hardware back button only triggers event once

**Steps to reproduce:**
Create a React app with two IonReactRouters.

For example, one router for logged out users and one for logged in users. The logged out router is mounted when the user is logged out, and upon login the new router is mounted. Only one router is mounted at any given moment, but users who log in or out will cause more than one router to be created during the app lifetime.

For example: `{isLoggedIn ? <AuthedApp/> : <UnauthedApp/>}` where each component has it's own routing.

Navigate to at least two pages in the second router and then press back. Ionic will take you back two pages instead of one.

**Related code:**

I think the issue is that the ionBackButton event is [registered but never removed](https://github.com/ionic-team/ionic-framework/blob/master/packages/react/src/routing/NavManager.tsx#L71) when the router is unmounted. It should be removed when `componentWillUnmount` is called.

Also seeing ` Can't perform a React state update on an unmounted component.` from IonReactRouter which further points to this being the issue.

**Other information:**

**Ionic info:**

```
Ionic:

   Ionic CLI       : 6.12.3 (/usr/local/lib/node_modules/@ionic/cli)
   Ionic Framework : @ionic/react 5.6.3

Capacitor:

   Capacitor CLI   : 2.4.7
   @capacitor/core : 2.4.7

Utility:

   cordova-res (update available: 0.15.3) : 0.15.2
   native-run (update available: 1.3.0)   : 1.2.2

System:

   NodeJS : v14.15.0 (/usr/local/Cellar/node@14/14.15.0/bin/node)
   npm    : 6.14.8
   OS     : macOS Big Sur

```

willmartindev:
## Pull request checklist

## Pull request type

 

 

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

Issue Number: resolves #23170 

## What is the new behavior?

- The hardware back button on Android will no longer trigger multiple times when `IonReactRouter` is unmounted and remounted.

- [ ] Yes
- [x] No


packages/react/src/routing/NavManager.tsx
```diff
@@ -67,14 +67,22 @@ export class NavManager extends React.PureComponent<NavManagerProps, NavContextS
...
-          this.nativeGoBack();
-          processNextHandler();
-        });
-      });
-    }
+    this.handleIonBackButton = this.handleIonBackButton.bind(this);
+  }
+
+  componentDidMount() {
+    document.addEventListener('ionBackButton', this.handleIonBackButton);
```
liamdebeasi, line 74: Two things here:

1. I would prefer if we kept this `addEventListener` in the constructor to avoid a race condition between when the instance is created and when it is mounted in the DOM (even as unlikely as it may be).
2. We need to keep the `typeof document !== 'undefined'` checks for both `addEventListener` and `removeEventListener` as both `document` and `window` will be undefined when doing SSR. Without this check an error will be thrown and SSR will fail.

liamdebeasi (changes requested):
Other than my comment below this looks really great!

packages/react/src/routing/NavManager.tsx
```diff
@@ -67,14 +67,22 @@ export class NavManager extends React.PureComponent<NavManagerProps, NavContextS
...
+  }
+
+  componentWillUnmount() {
+    document.removeEventListener('ionBackButton', this.handleIonBackButton)
+  }
+
+  handleIonBackButton(e: any) {
```
liamdebeasi, line 81: Minor thing, but can we call this `handleHardwareBackButton`? I want to avoid any confusion between HBB and the `ion-back-button` component.

liamdebeasi (approved):
Assuming what we discussed on Slack is fine, this PR looks good to go.

PR #23227, opened 2021-04-23
fix(vue): overlay events can now be listened for without the "on" prefix

liamdebeasi:
## Pull request checklist

## Pull request type

 

 

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

Currently Vue users need to listen to overlay events using an `on` prefix which is not very Vue-like. This syntax was carried over from Ionic React, but the `on` prefix is more suitable in a React environment.

## What is the new behavior?

- Users can now do `@didPresent` instead of `@onDidPresent`
- Deprecated `@onDid` and `@onWill` events

- [ ] Yes
- [x] No

## Other information

Only impacts overlay components

### DIFF
--- a/core/src/components/action-sheet/readme.md
+++ b/core/src/components/action-sheet/readme.md
@@ -418,7 +418,7 @@ Developers can also use this component directly in their template:
     header="Albums"
     css-class="my-custom-class"
     :buttons="buttons"
-    @onDidDismiss="setOpen(false)"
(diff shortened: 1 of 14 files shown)


liamdebeasi:
Forgot to mention that I am considering this a fix instead of a feature since it really should have been `@didDismiss` from the beginning. the `@onDidDismiss` syntax was an oversight that I should have caught previously.

PR #23231, opened 2021-04-23
feat(popover): popover can now be used inline

liamdebeasi:
## Pull request checklist

## Pull request type

 

 

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes

- Popover currently is only available via popoverController

## What is the new behavior?

- Popover is now available to use inline in your component template
- Popover is now a shadow component

- [x] Yes
- [] No

## Other information

In order for slots to work properly across framework integrations, `ion-popover` needs to be converted to shadow. As a result, I exposed a few shadow parts so that users can still style certain internals.

More robust documentation and usage examples will be added in the desktop popover PR. I wanted to split the inline popover feature out to its own PR to make it easier to review. I will be doing the same thing for `ion-modal` after popover is done in prep for our `ion-datetime` work.

A few notes on things I added:

- The `isOpen` property is used in Ionic Vue and Ionic React to programmatically open/close a popover. I moved it to the core component so that Angular and Vanilla JS users can use this feature as well. 
- The `willPresent`, `didPresent`, `willDismiss`, and `didDismiss` events are shorthands used in Ionic Vue and Ionic React so devs don't need to keep writing `ionPopover` every time they want to listen for a popover lifecycle event. I moved those to the core component too so that Angular and Vanilla JS users can use these as well.

The above 2 notes also help reduce the amount of framework-specific code we need to write an maintain.

- Another noteworthy change is that framework-specific delegates are no longer required when writing popover inline. This is because when you write `ion-popover` inline, the framework takes care of rendering it, so the popover is already created within the JS Framework's application context. Framework delegates are still needed for `popoverController` since those are initially created outside of the JS Framework's app context.


willmartindev (approved):
Looking really solid! Great work.

brandyscarney (approved):

Issue #23254, opened 2021-05-03
bug: Ionic Vue Router throwing error when its running on server in SSR mode.


mohammadsiddiqui:
# Bug Report

**Ionic version:**

[ ] **4.x**
[x] **5.x**

**Current behavior:**

Running ionic vue with SSR with ionic router, the @ionic/vue-router throws an error about history is not defined when it is running on server.

**Expected behavior:**

The error should not trigger.

**Steps to reproduce:**

clone the repo https://github.com/mohammadsiddiqui/vue-ionic-ssr and build it and run it in SSR mode and look at the server console.
**Related code:**

```
insert short code snippets here
```

**Other information:**

**Ionic info:**

![image](https://user-images.githubusercontent.com/24387047/116839842-b054f280-abcb-11eb-89d2-f20f5858ed59.png)

```
insert the output from ionic info here
```

liamdebeasi:
Thanks for the issue and PR! Could your provide some more steps to reproduce? I am trying to run the app in SSR mode, but I do not get the `history is not defined` error.

mohammadsiddiqui:
In the repo, I forgot to add back vue-router dependency as I was testing without ionic-router which I did now. you can run the same now.
Node: 12.16.3
NPM: 6.14.5

I have recorded a video demonstrating the issue, please find the link here. 

https://player.vimeo.com/video/544647326

![image](https://user-images.githubusercontent.com/24387047/116908124-74a54180-ac3a-11eb-980a-f36d2e8c4fbf.png)

liamdebeasi:
Thanks, that makes sense. I left a comment on your PR.

liamdebeasi:
Also here is a dev build of your changes if you would like to test them in your app:

```
npm install @ionic/vue@5.7.0-dev.202105031742.e2a0ece @ionic/vue-router@5.7.0-dev.202105031742.e2a0ece
```

liamdebeasi:
Thanks for the issue. This has been resolved via https://github.com/ionic-team/ionic-framework/pull/23255, and a fix will be available in an upcoming release of Ionic Framework.

PR #23255, opened 2021-05-03
fix(vue): use correct history mode when doing ssr to avoid errors

mohammadsiddiqui:
if this runs in SSR mode, we are getting history is undefined! so I just initialized the history variable from opts.

## Pull request checklist

## Pull request type

 

 

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

Issue Number: #23254 

## What is the new behavior?

- No error
- SSR works great

- [ ] Yes
- [x] No

## Other information
I have just added a line to initialize history variable before using it, so on the server it doesn't break

### DIFF
--- a/packages/vue-router/src/router.ts
+++ b/packages/vue-router/src/router.ts
@@ -40,6 +40,7 @@ export const createIonRouter = (opts: IonicVueRouterOptions, router: Router) => 
      * but we can check to see if the latest routing action
      * was a replace action by looking at the history state.
      */
+    const history = opts.history;
     const replaceAction = history.state.replaced ? 'replace' : undefined;
     handleHistoryChange(to, action || replaceAction, direction);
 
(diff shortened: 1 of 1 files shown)


packages/vue-router/src/router.ts
```diff
@@ -40,6 +40,7 @@ export const createIonRouter = (opts: IonicVueRouterOptions, router: Router) =>
      * but we can check to see if the latest routing action
      * was a replace action by looking at the history state.
      */
+    const history = opts.history;
     const replaceAction = history.state.replaced ? 'replace' : undefined;
```
liamdebeasi, line 44:
```suggestion
     * We need to use opts.history rather than window.history
     * because window.history will be undefined when using SSR.
     */
    const replaceAction = opts.history.state.replaced ? 'replace' : undefined;
```

Can we just add a comment here saying what this is doing? I tried adding it myself, but I am not able to push to your branch.

liamdebeasi (commented):
Looks good! Just need to add the comment I mentioned below.

packages/vue-router/src/router.ts
mohammadsiddiqui: @liamdebeasi thank you I have committed your suggestion.

liamdebeasi (approved):
Thanks for making this PR! I will get this merged once tests pass.

liamdebeasi:
Merged. Thank you again!

PR #23355, opened 2021-05-24
cleanup(router): Cleanup code, add comments

vicb:
## Pull request checklist

## Pull request type

 

 

- [ ] Bugfix
- [ ] Feature
- [x] Code style update (formatting, renaming)
- [ ] Build related changes

- [ ] Yes
- [x] No

## Other information

A few minor changes that make the code easier to read (IMO).

I am wondering there is a need for the chain abstraction ?
Every single branch of the route tree is converted to a list of "chain". 
It might be as simple (also simplify the code and be a little more efficient) to work with the tree directly.

@liamdebeasi any thoughts on that ?

### DIFF
--- a/core/src/components/router/utils/parser.ts
+++ b/core/src/components/router/utils/parser.ts
@@ -1,6 +1,16 @@
 import { RouteChain, RouteNode, RouteRedirect, RouteTree } from './interface';
 import { parsePath } from './path';
 
+const readProp = (el: HTMLElement, prop: string): string | null | undefined => {
+  if (prop in el) {
+    return (el as any)[prop];
+  }
+  if (el.hasAttribute(prop)) {
+    return el.getAttribute(prop);
+  }
+  return null;
+};
+
 export const readRedirects = (root: Element): RouteRedirect[] => {
   return (Array.from(root.children) as HTMLIonRouteRedirectElement[])
     .filter(el => el.tagName === 'ION-ROUTE-REDIRECT')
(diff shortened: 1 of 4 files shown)


core/src/components/router/utils/path.ts
```diff
@@ -27,33 +48,20 @@ export const chainToPath = (chain: RouteChain): string[] | null => {
...
+const removePrefix = (prefix: string[], path: string[]): string[] | null => {
   if (prefix.length > path.length) {
     return null;
   }
   if (prefix.length <= 1 && prefix[0] === '') {
     return path;
   }
   for (let i = 0; i < prefix.length; i++) {
-    if (prefix[i].length > 0 && prefix[i] !== path[i]) {
+    if (prefix[i] !== path[i]) {
```
vicb, line 64: `length == 0` could only happen for when `prefix = ['']` - other empty parts are filtered out in `parsePath()`. This case is already accounted for on l. 60.

_I also removed the `export` to make sure this function is only called internally by the router_

vicb:
Also quick question on https://github.com/ionic-team/ionic-framework/blob/bc2f1b89449cdc139cb10d28159822d269e5b554/core/src/components/router/utils/matching.ts#L134-L148

Note that `Math.pow(1, level);` is always `== 1`.

I guess the intent was to do `score += Math.pow(level, 1);` or `score += level;` ? If that is the case should the line below be `score += Math.pow(level, 2);` ?

core/src/components/router/utils/path.ts
liamdebeasi: That sounds good. While we are changing this function, could we add a few tests?

core/src/components/router/utils/parser.ts
```diff
@@ -17,46 +27,33 @@ export const readRoutes = (root: Element): RouteChain[] => {
...
 
-export const readRouteNodes = (root: Element, node = root): RouteTree => {
+export const readRouteNodes = (node: Element): RouteTree => {
   return (Array.from(node.children) as HTMLIonRouteElement[])
     .filter(el => el.tagName === 'ION-ROUTE' && el.component)
     .map(el => {
-      const component = readProp(el, 'component');
-      if (component == null) {
-        throw new Error('component missing in ion-route');
-      }
```
liamdebeasi, line 27: Do we still need this error?

core/src/components/router/utils/path.ts
```diff
@@ -9,6 +10,26 @@ export const generatePath = (segments: string[]): string => {
   return '/' + path;
 };
 
+const generateUrl = (segments: string[], useHash: boolean, queryString?: string) => {
```
liamdebeasi, line 13: Can we add some tests for `generateUrl`?

liamdebeasi (changes requested):
This looks great! Just need some tests and we will be good to go.

liamdebeasi:
Regarding https://github.com/ionic-team/ionic-framework/pull/23355#issuecomment-847353459, I am not sure what the original intent was. Are you running into issues with it?

core/src/components/router/utils/parser.ts
vicb: Look at the `filter()` on l 22 above, `el.component` is guaranteed not to be null then `readProp(el, 'component')` could not be null

core/src/components/router/utils/path.ts
vicb: That a common debate: should we test "private" stuff (in langs that support private). The general consensus is no.

The function is kind of private (I removed the `export`) and is call only from `readPath` that is extensively tested.

If we want to test `removePrefix` then we would have to `export` it, sending the signal that it could be used anywhere and then it would be good to keep the `> 0` because we can not make any assumptions on what the arguments would be.

So I would say that:
1) it's not worth adding the `export`
2) this function is tested via `readPath`
3) I would even inline this in `readPath` but I don't know your policy about it.

core/src/components/router/utils/path.ts
vicb: I would make a similar comment as I did for `removePrefix`:

- I would prefer not to add an `export` for test only,
- The function is tested via `writePath`.

core/src/components/router/utils/path.ts
liamdebeasi: We can keep it as is. I am fine having it get tested via `readPath`.

vicb:
> Regarding [#23355 (comment)](https://github.com/ionic-team/ionic-framework/pull/23355#issuecomment-847353459), I am not sure what the original intent was. Are you running into issues with it?

No issues.
I am trying to understand the matching process which is an interesting part of the router and would explain some things that look weird - my intent is to add documentation about that when I am comfortable enough with the matching.

When I was looking at it, I figured that `Math.pow(1, level); ` was a non-sense because it is just `1`. So I was wondering if the original intent was not to have `level` instead ?

liamdebeasi (approved):
Nice job! 👍

vicb:
> I can ask some team members to try and find out what the original intent was.

Yep, it would be nice to figure out. No rush for for now, we can deal with that later.

PR #23443, opened 2021-06-10
refactor(all): update required browser, framework, and mobile platform versions for v6

liamdebeasi:
## Pull request checklist

## Pull request type

 

 

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes

Framework v6 is going to better define exactly which browser, JS framework, and mobile platform versions it supports as it has been a source of confusion in the past.

Additionally, we are dropping support for iOS 11 and iOS 12.

- [x] Yes
- [ ] No

### DIFF
--- a/BREAKING.md
+++ b/BREAKING.md
@@ -26,8 +26,7 @@ This is a comprehensive list of the breaking changes introduced in the major ver
 - [Vue](#vue)
   * [Tabs Config](#tabs-config)
   * [Overlay Events](#overlay-events)
-  * [Minimum Required Version](#minimum-required-version)
-
+- [Browser and Platform Support](#browser-and-platform-support)
 
 
 ### Components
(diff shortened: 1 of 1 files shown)


EinfachHans:
What changes cause ios 11 & 12 to drop? Purely of interest. 😊

liamdebeasi:
The primary reason is usage. 93% of all iPhones are running iOS 13+ and 88% of all iPads are running iPadOS 13+ (see: https://developer.apple.com/support/app-store/). We typically provide `n-2` support for mobile platforms, and with iOS 15 coming out later this year that would leave us with iOS 13, iOS 14, and iOS 15.

The secondary reason is we want to encourage users to use the latest software so they can get all the new Web APIs and performance improvements. For example `IntersectionObserver`, which is used in our new `ion-datetime`, is not supported in iOS 12.0. Another example is https://bugs.webkit.org/show_bug.cgi?id=222187 which is fixed in iOS 15 and brings some solid performance improvements for components that use CSS Variables.

PR #23505, opened 2021-06-24
feat(): ion-hide-when-dark and ion-show-when-dark

eduardoRoth:
Added two classes to easily show/hide content when using `prefers-color-scheme: dark` in your application.

- [ ] Bugfix
- [X] Feature
- [ ] Build related changes

- [ ] Yes
- [X] No

### DIFF
--- a/core/src/css/display.scss
+++ b/core/src/css/display.scss
@@ -30,3 +30,16 @@
     }
   }
 }
+
+// Adds classes for displaying content based on prefers-color-scheme
+.ion-hide-when-dark {
+  @media (prefers-color-scheme: dark) {
+    display: none;
+  }
+}
+.ion-show-when-dark {
+  display: none;
+  @media (prefers-color-scheme: dark) {
+    display: block;
+  }
+}


liamdebeasi:
Thanks for the PR! Could you please describe what the use case is for this feature?

eduardoRoth:
Mainly for `` tags where I would need to have different images depending if the user has dark mode or not (mainly logos).

I know that I could use `background-image` css but for some cases it'd be better (simpler?) to just have the `img` tag:

Dark mode:
image

Light mode:
image

liamdebeasi:
Thanks! One of the concerns I have with this PR is that it still would cause both `img` sources to be loaded. 

For example: https://codepen.io/liamdebeasi/pen/ExWqyMo

If you look in the Network tab in dev tools, you should see that both the light mode and dark mode images are loaded.

The functionality you want to accomplish is already built into the browser with the `picture` element. It lets you define multiple image sources and show them based upon a media query: https://codepen.io/liamdebeasi/pen/NWpQreq

In that example, only one image is loaded at a time.

The other concern I have with this PR is that it makes the assumption that everyone wants to use `display: block` when showing the image, but some people may want `grid` or `flex`.

eduardoRoth:
> Thanks! One of the concerns I have with this PR is that it still would cause both `img` sources to be loaded.
> 
> For example: https://codepen.io/liamdebeasi/pen/ExWqyMo
> 
> If you look in the Network tab in dev tools, you should see that both the light mode and dark mode images are loaded.
> 
> The functionality you want to accomplish is already built into the browser with the `picture` element. It lets you define multiple image sources and show them based upon a media query: https://codepen.io/liamdebeasi/pen/NWpQreq
> 
> In that example, only one image is loaded at a time.
> 
> The other concern I have with this PR is that it makes the assumption that everyone wants to use `display: block` when showing the image, but some people may want `grid` or `flex`.

My bad! For all this time I didn't remember the `picture` tag, so this PR can be deleted 😅

Thanks for your time reviewing this 😄

PR #23537, opened 2021-06-30
docs(loading): show how to use dismiss method returns from useIonLoading for react

omar2205:
The docs don't say how to use dismiss(), this shows that the `useIonLoading` hook returns the present and dismiss functions.

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes

The docs mention quickly using dismiss() to close the IonLoading, but it doesn't say how to access this function.

Issue Number: N/A

## What is the new behavior?
- Just adds the dismiss as a second return in the react example.

- [ ] Yes
- [x] No

### DIFF
--- a/core/src/components/loading/usage/react.md
+++ b/core/src/components/loading/usage/react.md
@@ -7,17 +7,18 @@ import { IonButton, IonContent, IonPage, useIonLoading } from '@ionic/react';
 interface LoadingProps {}
 
 const LoadingExample: React.FC<LoadingProps> = () => {
-  const [present] = useIonLoading();
+  const [present, dismiss] = useIonLoading();
   return (
     <IonPage>
       <IonContent>
         <IonButton
           expand="block"
-          onClick={() =>
+          onClick={() => {
             present({
-              duration: 3000,
+              message: 'Loading...',
             })
-          }
+            setTimeout(dismiss, 3000)
+          }}
         >
           Show Loading
         </IonButton>
--- a/core/src/components/loading/usage/react.md
+++ b/core/src/components/loading/usage/react.md
@@ -1,17 +1,12 @@
 ```tsx
 /* Using with useIonLoading Hook */
 
-import React from "react";
-import { IonButton, IonContent, IonPage, useIonLoading } from "@ionic/react";
+import React from 'react';
+import { IonButton, IonContent, IonPage, useIonLoading } from '@ionic/react';
 
 interface LoadingProps {}
 
 const LoadingExample: React.FC<LoadingProps> = () => {
-  /**
-   * The recommended way of dismissing is to use the `dismiss` property
-   * on `IonLoading`, but the `dismiss` method returned from `useIonLoading`
-   * can be used for more complex scenarios.
-   */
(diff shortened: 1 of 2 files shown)


liamdebeasi (changes requested):
Thanks for the PR! To make this change, you should add the `dismiss` line to `core/src/components/loading/usage/react.md` and then run `npm run build` to have that get added to `readme.md`. The part of the `readme.md` file you edited is automatically generated and should not be edited manually.

core/src/components/loading/usage/react.md
```diff
@@ -7,17 +7,18 @@ import { IonButton, IonContent, IonPage, useIonLoading } from '@ionic/react';
...
+  const [present, dismiss] = useIonLoading();
   return (
     <IonPage>
       <IonContent>
         <IonButton
           expand="block"
-          onClick={() =>
+          onClick={() => {
             present({
-              duration: 3000,
```
liamdebeasi, line 18: I would rather we keep the original `duration: 3000` property. I don't want users thinking the recommended way to dismiss `IonLoading` is to use a `setTimeout` manually when there's a property built in that will do that.

core/src/components/loading/usage/react.md
```diff
@@ -7,17 +7,18 @@ import { IonButton, IonContent, IonPage, useIonLoading } from '@ionic/react';
 interface LoadingProps {}
 
 const LoadingExample: React.FC<LoadingProps> = () => {
-  const [present] = useIonLoading();
+  const [present, dismiss] = useIonLoading();
```
liamdebeasi, line 10:
```suggestion
  /**
   * The recommended way of dismissing is to use the `dismiss` property
   * on `IonLoading`, but the `dismiss` method returned from `useIonLoading` 
   * can be used for more complex scenarios.
   */
  const [present, dismiss] = useIonLoading();
```

liamdebeasi (changes requested):
Looks great. I added a few comments below which should help make the docs a bit clearer.

omar2205:
>   /**
>    * The recommended way of dismissing is to use the `dismiss` property
>    * on `IonLoading`, but the `dismiss` method returned from `useIonLoading` 
>    * can be used for more complex scenarios.
>    */

Did you mean "duration" property?

core/src/components/loading/usage/react.md
liamdebeasi: I think GitHub may have accidentally marked this as resolve, but I would still like to see this issue resolved.

One idea is that maybe we could add an additional example underneath the one that uses the `duration` property to show how you can manually dismiss the loading component? You could use your `setTimeout` example there and add a comment above the `setTimeout` that says something like: 

```The `dismiss` method can also be used to dismiss `ion-loading` after a long running task such as a network request.```

omar2205: we could add a comment to mention it
`setTimeout(dismiss, 2000) // simulate a dynamic dismissal of the IonLoading`

liamdebeasi (approved):
Great job!

liamdebeasi:
Merged. Thank you!

PR #23576, opened 2021-07-04
feat(picker-column): add 'showName' property

Giulio-91:
add the possibility tho display the column name by setting the showName column property to true

## Pull request checklist

## Pull request type

 

 

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes

Issue Number: N/A

## What is the new behavior?

- now it is possible to show the column name by setting the showName column property to true

- [ ] Yes
- [x] No

### DIFF
--- a/core/src/components/picker-column/picker-column.ios.scss
+++ b/core/src/components/picker-column/picker-column.ios.scss
@@ -25,6 +25,19 @@
   pointer-events: none;
 }
 
+.picker-name {
+  height: $picker-ios-option-height;
+  
+  transform-style: preserve-3d;
+
+  color: inherit;
+  
+  font-size: $picker-ios-option-font-size;
+
+  line-height: $picker-ios-option-height;
+  pointer-events: none;
+}
+
 .picker-opt {
   @include padding($picker-ios-option-padding-top, $picker-ios-option-padding-end, $picker-ios-option-padding-bottom, $picker-ios-option-padding-start);
   @include margin(0);
--- a/core/src/components/picker-column/picker-column.md.scss
+++ b/core/src/components/picker-column/picker-column.md.scss
@@ -25,6 +25,20 @@
   pointer-events: none;
 }
 
+.picker-name {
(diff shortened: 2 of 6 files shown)


liamdebeasi:
Thanks for the PR! Is there an associated issue for this PR?

Giulio-91:
> Thanks for the PR! Is there an associated issue for this PR?

No. Should I create it?

liamdebeasi:
If there isn't one already that's fine. Could you please explain the use case for this feature?

Giulio-91:
> If there isn't one already that's fine. Could you please explain the use case for this feature?

Yes of course. I needed to show the column name of the picker.

So I added the "showName" property in the 'PickerColumn' interface. If it is set to true, the column name is shown on the top of the column.

I also added an example in the test script.

liamdebeasi:
Thanks for the clarification and apologies for the delay. What is the benefit of showing the column name instead of using `prefix` or `suffix`? In practice, the `prefix` and `suffix` properties are what describe the columns. Here's an example from the stock iOS clock app:

IMG_8516

I think the addition of this API is a bit confusing since it lets you describe the column in two different ways. Here is a screenshot of the example you added: 

image

The "hours" suffix describes the column, so having `showName` here is redundant in my opinion.

liamdebeasi:
Thanks for the PR! I discussed this with the team and we are unable to accept your PR at this time. We feel that the existing `suffix` and `prefix` properties already provide the functionality that the proposed `showName` property would provide.

As I mentioned in https://github.com/ionic-team/ionic-framework/pull/23576#issuecomment-883373658, native iOS apps use the suffix and prefix features to assign labels to picker columns. Adding the `showName` property to Ionic Framework would allow developers describe a column in two different ways which we feel may lead to a confusing user experience.

In the future we recommend opening an issue first before creating a PR as it lets us discuss the scope of a feature before any programming work begins. Please let me know if you have any questions. Thank you!

PR #23585, opened 2021-07-06
fix(datetime): add keyboard year navigation

### LINKED ISSUE #18122: feat: Use arrows to change day/month/year HH:SS in datetime
# Feature Request

**Ionic version:**

[x] **4.3.0**

Currently `ion-datetime` control is accessible through keyboard in non optimal way. To change day/month user should use `TAB` key to go to target value and then press `space` key to select it.

The issue with this is that if user needs to change just value in 2nd column (s)he needs to `TAB` skip all the values in 1st one (e.g. 31 days for example). This is very daunting experience. It's also does not help that while tab-jumping though values the current column is not scrolled so user can't see currently focused item when it's out of visible window.

The better solution would be to use left-right keys to change current columns and up-down keys to navigate values. The space still can be used to select focused value to be current.

The tab key should be used/reserved to navigation between controls. So when in datetime control and `tab` pressed the datetime should close and next control should get focus.

**Describe Alternatives**

Also it would be super cool to support entering date with number keys.

**Related Code**

http://plnkr.co/edit/li6pVumtGoVGrhjTbAqg?p=preview

**Additional Context**

### LINKED ISSUE #21553: bug: Ion-datetime component is not accessible for the user using keyboard
# Bug Report

**Ionic version:**

[ ] **4.x**
[x] **5.x**

**Current behavior:**
The screen reads the ion-datetime component as a button. It’s not reading the associate aria-label or the <ion-label> for it. Also, the date selector is not accessible using only keyboard. It does not receive the focus by default.

Issue is reproducible from the sample implementation also
https://ionicframework.com/docs/demos/api/datetime/index.html?ionic:mode=ios

Refer to item 4 in the document shared in Slack for further details.
https://ionicteam.slack.com/archives/C7CGXAVTL/p1592319616186200

willmartindev:
## Pull request checklist

## Pull request type

 

 

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

Issue Number: 
- resolves #21553 
- resolves #18122

## What is the new behavior?

- Users can now navigate years in `ion-datetime` by pressing (`Shift` + `PageUp`) or (`Shift` + `PageDown`).
- Added accessibility section to `ion-datetime` docs.

- [ ] Yes
- [x] No

## Other information

Keyboard behavior based on https://www.w3.org/TR/wai-aria-practices/examples/dialog-modal/datepicker-dialog.html

### DIFF
--- a/core/src/components/datetime/utils/manipulation.ts
+++ b/core/src/components/datetime/utils/manipulation.ts
@@ -277,7 +277,7 @@ export const getPreviousYear = (refParts: DatetimeParts) => {
  * Given DatetimeParts, generate the next year.
  */
 export const getNextYear = (refParts: DatetimeParts) => {
-  return changeYear(refParts, +1);
+  return changeYear(refParts, 1);
 }
 
 /**
--- a/core/src/components/datetime/utils/manipulation.ts
+++ b/core/src/components/datetime/utils/manipulation.ts
@@ -270,14 +270,14 @@ const changeYear = (refParts: DatetimeParts, yearDelta: number) => {
  * Given DatetimeParts, generate the previous year.
  */
 export const getPreviousYear = (refParts: DatetimeParts) => {
-  return changeYear(refParts, +1);
+  return changeYear(refParts, -1);
 }
 
 /**
  * Given DatetimeParts, generate the next year.
  */
 export const getNextYear = (refParts: DatetimeParts) => {
-  return changeYear(refParts, -1);
+  return changeYear(refParts, +1);
 }
 
 /**
--- a/core/src/components/datetime/utils/manipulation.ts
+++ b/core/src/components/datetime/utils/manipulation.ts
@@ -256,6 +256,30 @@ export const getNextMonth = (refParts: DatetimeParts) => {
   return { month, year, day };
 }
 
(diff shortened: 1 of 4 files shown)


core/src/components/datetime/utils/manipulation.ts
```diff
@@ -256,6 +256,30 @@ export const getNextMonth = (refParts: DatetimeParts) => {
...
+ */
+export const getPreviousYear = (refParts: DatetimeParts) => {
+  return changeYear(refParts, -1);
+}
+
+/**
+ * Given DatetimeParts, generate the next year.
+ */
+export const getNextYear = (refParts: DatetimeParts) => {
+  return changeYear(refParts, +1);
```
liamdebeasi, line 280:
```suggestion
  return changeYear(refParts, 1);
```

Minor style thing, but the positive here is implied

liamdebeasi (changes requested):
Other than a small style comment this looks good to go

core/src/components/datetime/utils/manipulation.ts
willmartindev: thanks!

liamdebeasi (approved):
Nice job!

PR #23145, opened 2021-04-05
fix(segment, segment-button): use correct tablist and tab roles for screen readers

willmartindev:
## Pull request checklist

Please check if your PR fulfills the following requirements:
- [x] Tests for the changes have been added (for bug fixes / features)
- [x] Docs have been reviewed and added / updated if needed (for bug fixes / features)
- [x] Build (`npm run build`) was run locally and any changes were pushed
- [x] Lint (`npm run lint`) has passed locally and any fixes were made for failures

## Pull request type

 

 

Please check the type of change your PR introduces:
- [x] Bugfix
- [ ] Feature
- [ ] Code style update (formatting, renaming)
- [ ] Refactoring (no functional changes, no api changes)
- [ ] Build related changes
- [ ] Documentation content changes
- [ ] Other (please describe): 

## What is the current behavior?

Issue Number: resolves #23059 

## What is the new behavior?

- `ion-segment` now uses `role="tablist"`
- `ion-segment-button` now uses `role="tab"` and updates `aria-selected` based on its selected state
- `ion-segment-button` now passes the `aria-controls` attribute to its child button element

## Does this introduce a breaking change?

- [ ] Yes
- [x] No

PR #23590, opened 2021-07-08
feat(segment): add keyboard navigation, add selectOnFocus property to control selection follow focus behavior

### LINKED ISSUE #23520: bug: segment should be usable with arrow keys
# Feature Request

**Ionic version:**
[ ] **4.x**
[ ] **5.x**
[x] **6.x**

Would like the ability to arrow key left and right between Segment buttons. 
Bonus feature request would be able to use Enter Key to activate the focused Segment Button.

Closest I can find for this behavior is with the Angular Material Tabs.

image

[https://material.angular.io/components/tabs/overview#tab-group-basic](https://material.angular.io/components/tabs/overview#tab-group-basic)

willmartindev:
## Pull request checklist

## Pull request type

 

 

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

Issue Number: resolves #23520 

## What is the new behavior?

- `ion-segment` now responds properly to keyboard navigation.

- [ ] Yes
- [x] No


core/src/components/segment-button/segment-button.tsx
```diff
@@ -87,15 +87,7 @@ export class SegmentButton implements ComponentInterface, ButtonInterface {
...
-    if (this.disabled) { return -1; }
-
-    const hasTabIndex = this.el.hasAttribute('tabindex');
-
-    if (hasTabIndex) {
-      return this.el.getAttribute('tabindex');
-    }
-
-    return 0;
+    return this.checked && !this.disabled ? 0 : -1;
```
liamdebeasi, line 90: Have we verified that this does not regress https://github.com/ionic-team/ionic-framework/pull/23145?

core/src/components/segment/segment.tsx
```diff
@@ -431,12 +433,74 @@ export class Segment implements ComponentInterface {
...
+      case 'Enter':
+        ev.preventDefault();
+        current = this.getSegmentButton('focused');
+      default:
+        break;
+    }
+
+    const previous = this.checked || current;
+    if (previous && current) {
+      this.checkButton(previous, current);
```
liamdebeasi, line 485: Rather than check the segment button, I think we should focus it and let the user decide whether or not they want to activate it.

This would bring our behavior in line with the MD spec: https://material-components.github.io/material-components-web-catalog/#/component/tabs

core/src/components/segment/segment.tsx
```diff
@@ -431,12 +433,74 @@ export class Segment implements ComponentInterface {
...
+    }
+  }
+
   render() {
     const mode = getIonMode(this);
     return (
       <Host
         role="tablist"
         onClick={this.onClick}
+        onKeyup={this.onKeyUp}
```
liamdebeasi, line 503: Minor style thing, but can we use the `@Listen` decorator here? This is what we do in radio group, and I would like to keep it consistent: https://github.com/ionic-team/ionic-framework/blob/d527414e109c9cc6ae43171911f89a43b2f0691f/core/src/components/radio-group/radio-group.tsx

core/src/components/segment/segment.tsx
```diff
@@ -431,12 +433,74 @@ export class Segment implements ComponentInterface {
...
+        break;
+    }
+
+    const previous = this.checked || current;
+    if (previous && current) {
+      this.checkButton(previous, current);
+    }
+  }
+
+  private handleFallbackFocus() {
```
liamdebeasi, line 489: What is the purpose of this?

core/src/components/segment-button/segment-button.tsx
willmartindev: Yeah, I found no regression!

Additionally, this follows the `tabIndex` pattern shown here: https://www.w3.org/TR/wai-aria-practices-1.1/examples/tabs/tabs-1/tabs.html#sc1_label

core/src/components/segment/segment.tsx
willmartindev: The following link suggests defaulting to automatic selection unless any associated subsequent actions might cause a network request/delay: https://www.w3.org/TR/wai-aria-practices-1.1/#kbd_selection_follows_focus

I think we are good to go either way though, based on what assumptions we want to make/not make. 

Idea: we could also fairly easily expose a `@Prop() keyboardBehavior: focus | select` on `ion-segment` if we want to leave both possibilities open.

core/src/components/segment/segment.tsx
willmartindev: Each `ion-segment-button` sets their `tabIndex` to `-1` if they are not selected. That way, the `selected` button will always receive focus on tab.

This function fixes an edge case where *none* of the segment buttons are selected, since that is possible if the user doesn't set a `value`. In that case, all of the elements would have `tabIndex={-1}`. If that situation is met, this function makes the first button tab-able.

core/src/components/segment/segment.tsx
willmartindev: Adding the prop was a quick change, so I went ahead and added it with another commit. Let me know if you think it should stay or go. 👌

core/src/components/segment/segment.tsx
```diff
@@ -431,6 +439,76 @@ export class Segment implements ComponentInterface {
...
+        ev.preventDefault();
+        current = this.getSegmentButton('last');
+        break;
+      case ' ':
+      case 'Enter':
+        ev.preventDefault();
+        current = this.getSegmentButton('focused');
+        keyboardBehavior = 'select';
+      default:
+        break;
```
liamdebeasi, line 493: If we add `delegatesFocus: true` to the `ion-segment` component decorator we should not have to do this as the browser will handle this for us.

liamdebeasi (changes requested):

core/src/components/segment/segment.tsx
```diff
@@ -431,6 +439,76 @@ export class Segment implements ComponentInterface {
     this.checked = current;
   }
 
+  private getSegmentButton = (selector: 'first' | 'last' | 'next' | 'previous' | 'focused'): HTMLIonSegmentButtonElement | null => {
+    const buttons = this.getButtons().filter(button => !button.disabled);
+    const currIndex = buttons.findIndex(button =>
+      (this.keyboardBehavior === 'select' && this.checked)
+      ? button === this.checked
+      : button.classList.contains('ion-focused'
+    ));
```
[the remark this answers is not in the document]
liamdebeasi: Actually thought of a better solution:

```suggestion
    const currIndex = buttons.findIndex(button => button === document.activeElement || button === this.checked);
```

This returns the index if the button we are looking at is either the focused element or the checked/selected element. This also avoids iterating through all the classes on each button.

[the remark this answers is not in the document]
willmartindev: We need to keep the logic for checking whether `selectOnFocus` is true or false, but I changed the `classList.contains` to `document.activeElement` as per your suggestion. 

Also, I changed it to an `if` statement and added a comment for readability.

core/src/components/segment/segment.tsx
```diff
@@ -431,6 +439,76 @@ export class Segment implements ComponentInterface {
...
+        current = this.getSegmentButton('focused');
+        keyboardBehavior = 'select';
+      default:
+        break;
+    }
+
+    const previous = this.checked || current;
+    if (keyboardBehavior === 'focus') {
+      current?.focus();
+    } else if (keyboardBehavior === 'select' && previous && current) {
```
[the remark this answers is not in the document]
willmartindev: Tabbing into a segment with no value set will not select the first item, but any subsequent arrow key presses will select on focus. This is intended behavior, such that a user quickly tabbing through items does not accidentally modify the selected state.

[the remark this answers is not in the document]
liamdebeasi: Makes sense!

core/src/components/segment/segment.tsx
```diff
@@ -431,6 +442,83 @@ export class Segment implements ComponentInterface {
...
+        current = this.getSegmentButton('first');
+        break;
+      case 'End':
+        ev.preventDefault();
+        current = this.getSegmentButton('last');
+        break;
+      case ' ':
+      case 'Enter':
+        ev.preventDefault();
+        current = this.getSegmentButton('focused');
```
liamdebeasi, line 499:
```suggestion
        current = document.activeElement;
```

If an `ion-segment-button` is focused then it will be `document.activeElement`. If segment is not focused then this keydown handler will not fire anyways since it is attached to `ion-segment`.

core/src/components/segment/segment.tsx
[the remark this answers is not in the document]
liamdebeasi: But if `selectOnFocus: true`, isn't the current button focused too?

liamdebeasi (changes requested):
This is looking great! A few small comments, but overall this is a really solid PR.

core/src/components/segment/segment.tsx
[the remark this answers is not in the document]
willmartindev: Ah! I was seeing weird behavior when trying your suggestion, but it was fixed by just checking for `button === document.activeElement` rather than `button === document.activeElement || button === this.checked`. 

The checked button is only relevant if it is also the `activeElement`, so checking for it was a redundant holdover from my initial commit. 

Thanks!

core/src/components/segment/segment.tsx
liamdebeasi: Thinking about this more I don't think we need to make the change. `delegatesFocus` can sometimes be a bit finicky on iOS so I would rather not spend time debugging that.

core/src/components/segment/readme.md
```diff
@@ -8,6 +8,20 @@ Their functionality is similar to tabs, where selecting one will deselect all ot
...
+### Keyboard Navigation
+
+On recieving focus, `ion-segment` will pass focus to the selected `ion-segment-button` element. The component also has full keyboard support for navigating between `ion-segment-button` elements. The following table details what each key does:
+
+| Key                | Function                                                       |
+| ------------------ | -------------------------------------------------------------- |
+| `ArrowRight`       | Selects the next focusable element.                            |
+| `ArrowLeft`        | Selects the previous focusable element.                        |
+| `Home`             | Selects the first focusable element.                           |
+| `End`              | Selects the last focusable element.                            |
```
liamdebeasi, line 22: This is a little confusing because whether or not it "selects" depends on the value of `selectOnFocus`. Maybe we just say that it focuses the element rather than selects it.

Above the table you could note that "By default, keyboard navigation will only focus `ion-segment-button` elements, but you can use the `selectOnFocus` property to ensure that they get selected on focus too."

core/src/components/segment/segment.tsx
```diff
@@ -16,7 +16,9 @@ import { createColorClasses, hostContext } from '../../utils/theme';
     ios: 'segment.ios.scss',
     md: 'segment.md.scss'
   },
-  shadow: true
+  shadow: {
+    delegatesFocus: true
+  }
```
liamdebeasi, line 21: If we add delegatesFocus do we still need the explicit "Enter" key listener below?

willmartindev: Yes, if I remove the `Enter` listener then nothing happens on `Enter` keyup. 

I am guessing this is because all selection logic is handled in the parent `ion-segment` with an `ionSelect` event, rather than the child button's `onClick` or `onKeyUp`.

core/src/components/segment/segment.tsx
```diff
@@ -431,6 +439,74 @@ export class Segment implements ComponentInterface {
...
+
+    switch (selector) {
+      case 'first':
+        return buttons[0];
+      case 'last':
+        return buttons[buttons.length - 1];
+      case 'next':
+        return buttons[currIndex + 1] || buttons[0];
+      case 'previous':
+        return buttons[currIndex - 1 ] || buttons[buttons.length - 1];
```
liamdebeasi, line 454:
```suggestion
        return buttons[currIndex - 1] || buttons[buttons.length - 1];
```

core/src/components/segment/segment.tsx
```diff
@@ -431,6 +439,74 @@ export class Segment implements ComponentInterface {
     this.checked = current;
   }
 
+  private getSegmentButton = (selector: 'first' | 'last' | 'next' | 'previous' | 'focused'): HTMLIonSegmentButtonElement | null => {
```
liamdebeasi, line 442:
```suggestion
  private getSegmentButton = (selector: 'first' | 'last' | 'next' | 'previous'): HTMLIonSegmentButtonElement | null => {
```

liamdebeasi (approved):
Great job! Once the builds pass this is good to merge. Be sure to mark this as a feature since we are adding an API. 

I updated the PR title, but you may have to set the commit message yourself when merging.

PR #23596, opened 2021-07-08
feat(toast): add icon property to show icon at start of toast content

### LINKED ISSUE #23524: feat: toast with icon
# Feature Request

**Ionic version:**
[ ] **4.x**
[x] **5.x**
[ ] **6.x**

A toast with an icon would be nice. The buttons can already have an icon. But the message itself should have an optional icon at slot start.

https://ionicframework.com/docs/api/toast

image
<sup>Just a photo montage.</sup>

**Related Code**

```ts
const toast = await this.toastController.create({
  header: 'Toast header',
  message: 'Click to Close',
  icon: 'bug', // <- new property
});
```

domske:
## Pull request checklist

fixes #23524

## Pull request type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes

The toast does not allow to display an icon. 

## What is the new behavior?

You can now optionally specify an icon to show with the toast message.

**md**

demo-md

**ios**

demo-ios

- [ ] Yes
- [x] No

## Other information

I added linter ignore blacklist property comment to allow `padding` and `margin-left`. Feel free to change it. I don't know what you prefer.

Should be available in Ionic 5.x


EinfachHans:
Maybe better to use `icon` instead of `name` Property to allow custom icons?

domske:
@EinfachHans 

You can already use custom icons with the `name` attribute.
Just place your svg icons to the Ionic icons svg build folder and you can easily use both (build-in and custom icons) as name.

For example in Angular. You can automatically copy your custom icons with the build. 

**angular.json**

`projects {} > app {} > architect {} > build {} > options {} > assets []`:

```json
{
  "glob": "**/*.svg",
  "input": "src/assets/custom-icons",
  "output": "./svg"
}
```

Place e.g. an icon like `custom-my-icon.svg` to your src assets and you can use the ion-icon (or the toast icon prop) like: 

```html
<ion-icon name="custom-my-icon"></ion-icon>
```

We could replace `name` with the `icon` prop. But does it also work with my solution above? I have to test it... Because I use the `name` solution. It's nicer (names) than src-paths.

[Icon Source](https://github.com/ionic-team/ionicons/blob/master/src/components/icon/icon.tsx)

We probably should also complete the `ion-icon` documentation. Because I cannot find the `icon` prop. 
Only in the source code. See [docs](https://ionic.io/ionicons/usage).

EinfachHans:
As far as i understand the `icon` prop could also take for example `"assets/..."` and determinates if its a asset icon or ionicon. Also the toast button icon uses the `icon` Property

domske:
Ok I replaced the `name` property with `icon`. Not yet tested in runtime / productive app. But I think it will work since the button icon also uses this property.

core/src/components/toast/toast.scss
```diff
@@ -141,7 +150,7 @@
   z-index: 0;
 }
 
-.toast-icon {
+.toast-button-icon {
```
liamdebeasi, line 153:
```suggestion
.toast-icon,
.toast-button-icon {
```

These styles should also apply to `.toast-icon` here so we do not need to duplicate it above.

core/src/components/toast/toast.scss
```diff
@@ -120,6 +120,15 @@
...
 }
 
+.toast-icon {
+  /* stylelint-disable property-blacklist */
+  margin-left: 15px;
+
+  padding: 10px;
+   /* stylelint-enable property-blacklist */
+
+  font-size: 1.4em;
```
liamdebeasi, line 130:
```suggestion
  @include margin(0, 0, 0, 15px);
  @include padding(10px, 10px, 10px, 10px);
```

We need to include the mixins here to account for RTL mode. Currently, the styles are not consistent between LTR and RTL modes, but the mixins will handle this switch for us.

Also, no need to add the font size here as it should apply in the same block as `.toast-button-icon` (see my other comment)

core/src/components/toast/toast.tsx
```diff
@@ -281,6 +287,10 @@ export class Toast implements ComponentInterface, OverlayInterface {
           <div class="toast-container" part="container">
             {this.renderButtons(startButtons, 'start')}
 
+            {this.icon !== undefined &&
+              <ion-icon class="toast-icon" part="icon" icon={this.icon}></ion-icon>
```
liamdebeasi, line 291: We need to document this shadow part up top. See above the `@Component` decorator. We need to add something like this:

```
@part icon - The icon that appears next to the toast content.
```

core/src/components/toast/toast.tsx
```diff
@@ -106,6 +106,12 @@ export class Toast implements ComponentInterface, OverlayInterface {
    */
   @Prop() animated = true;
 
+  /**
+   * Name of the icon to display. See `ion-icon`.
```
liamdebeasi, line 110:
```suggestion
   * The name of the icon to display. See `ion-icon`.
```

core/src/components/toast/toast.tsx
liamdebeasi, line 291:
```suggestion
              <ion-icon class="toast-icon" part="icon" icon={icon} lazy={false} aria-hidden="true"></ion-icon>
```

The icon does not serve a functional purpose, so I don't think we need to announce it to screen readers. We can add `lazy` here so that the icon is loaded when the `ion-toast` is first created, not when the toast comes into view.

Let's also add `icon` to the first line in the `render` function so that we don't need to keep accessing `this`.

liamdebeasi (changes requested):
Thanks for the PR! I have included comments regarding changes which are required in order for this feature to get merged.

Additionally, can you please add some usage examples in the `usage` directory? You will need to update the `angular.md`, `javascript.md`, `react.md`, `stencil.md`, and `vue.md` files, run `npm run build`, and then push that result.

I think we can add the icon example inside of the existing `presentToastWithOptions` function in these usage files.

domske:
Ok, change requests pushed...

I just wonder:

- Why `this.icon` on the first line? It's a simple short path and equal to the the other properties like `this.message` and `this.header`. It's the same style.

- Why `lazy={false}` and not `lazy="false"` since `aria-hidden="true"`? Ok, the type is different (boolean and string) but does this matter here?

Nothing important. I'm just wondering.

liamdebeasi:
Sorry I had two different PRs open and was referencing the wrong one 🤦 . Ignore the `this.icon` comment.

The `ariaHidden` property expects a string: https://github.com/ionic-team/ionicons/blob/master/src/components/icon/icon.tsx#L38
But the `lazy` property expects a boolean: https://github.com/ionic-team/ionicons/blob/master/src/components/icon/icon.tsx#L82

Since the `ariaHidden` property eventually gets set as the `aria-hidden` attribute, we need to accept a string since the `aria-hidden` attribute cannot accept a boolean (it's just plain HTML at that point)

domske:
Ok, please let me know if there are any pending change requests. Or did I miss something?

willmartindev (changes requested):
Looks good! Just a small docs change from me to reflect the switch from `name` to `icon`.

liamdebeasi (approved):
Great job!. We are good to merge when the tests finish.

liamdebeasi:
Merged. Thank you!

PR #23686, opened 2021-07-26
feat(datetime): add hourCycle property

### LINKED ISSUE #23661: feat: Override option to force 12-hour/24-hour format in ion-datetime
### Prequisites

The feature would allow for `ion-datetime` (v6) to take in a property to force its time display format, to either 12-hour or 24-hour time, regardless of the user's locale/region. 

### Describe the Use Case

Currently developing an app for businesses to track employee time and there's a client-level preference to show all times within the app in 12-hour or 24-hour format. Some clients need their times specifically in 24-hour time, such as hospitals.

- Add a "timeFormat" property which by default is null or undefined (listens to locale/region)
- Has optional "12h" and "24h" as input to override locale and force one format or the other.

### Describe Alternatives

An alternative could be bringing back `displayFormat` as was implemented in Ionic <= 5 for `ion-datetime` which took in  values, such as `"H:mm"`, `"h:mm A"`, etc. Perhaps renaming to `timeFormat` to be more specific.

### Related Code

```
<ion-datetime
    ...
    timeFormat="24h"
>
</ion-datetime>
```

or the alternative

```
<ion-datetime
    ...
    timeFormat="H:mm"
>
</ion-datetime>
```

### Additional Information

_No response_

EinfachHans:
resolves #23661

## Pull request checklist

## Pull request type

 

 

- [ ] Bugfix
- [X] Feature
- [ ] Build related changes

Currently the date-time uses the locale to check if it should use a 12 or 24 hour format

Issue Number: #23661 

## What is the new behavior?

- `timeFormat` prop added which could be '12h' or '24h' (or undefined)
- If set: Use this format
- if not set: Use locale like before

- [ ] Yes
- [X] No


core/src/components/datetime/datetime.tsx
```diff
@@ -1397,7 +1403,8 @@ export class Datetime implements ComponentInterface {
    * should just be the default segment.
    */
   private renderTime(mode: Mode) {
-    const use24Hour = is24Hour(this.locale);
+    const { timeFormat } = this;
+    const use24Hour = timeFormat != null ? timeFormat === '24h' : is24Hour(this.locale);
```
liamdebeasi, line 1407: We are going to need to revamp the `is24Hour` function a bit since you can also encode the hour cycle inside of the locale string. (See: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/Locale/hourCycle#adding_an_hour_cycle_via_the_locale_string)

My current thinking is pass `this.hourCycle` and `this.locale` into `is24Hour`. The internals of the function would look something like this:

```typescript
export const is24Hour = (locale: string, hourCycle?: 'h23' | 'h12') => {
  /**
   * If developer has explicitly enabled h23 time
   * then return early and do not look at the system default.
   */
  if (hourCycle !== undefined) {
    return hourCycle === 'h23';
  }

  /**
   * If hourCycle was not specified, check the locale
   * that is set on the user's device. We first check the
   * Intl.DateTimeFormat hourCycle option as developers can encode this
   * option into the locale string. Example: `en-US-u-hc-h23`
   */
  const date = new Date('5/18/2021 00:00');
  const formatted = new Intl.DateTimeFormat(locale, { hour: 'numeric' });
  const options = formatted.resolvedOptions();
  if (options.hourCycle !== undefined) {
    return options.hourCycle === 'h23';
  }
  
  /**
   * If hourCycle is not specified (either through lack
   * of browser support or locale information) then fall
   * back to this slower hourCycle check.
   */
  const parts = formatted.formatToParts(date);
  const hour = parts.find(p => p.type === 'hour');

  if (!hour) {
    throw new Error('Hour value not found from DateTimeFormat');
  }

  return hour.value === '00';
}
```

The support for `options.hourCycle` is somewhat new, but as time goes on we would be able to rely more and more on this instead of our fallback hack.

This would need some more revamping if we are to support `h11` and `h24` formats, but I think sticking with `h12` and `h23` is fine for now.

Also please add tests to the `helpers.spec.ts` file for this.

I know things get a bit tricky when you are dealing with different hour formats, so let me know if you would like any assistance on this change.

liamdebeasi (changes requested):
Thanks for working on this! This is a great start. I would like to see this PR move more in line with what the JavaScript internationalization APIs use so that the API is more future proof. I left changes below which are required to merge this PR.

Please let me know if you have any questions about anything I wrote.

core/src/components/datetime/datetime.tsx
```diff
@@ -318,6 +318,12 @@ export class Datetime implements ComponentInterface {
    */
   @Prop() showDefaultTimeLabel = true;
 
+  /**
+   * The time format of the `ion-datetime`. If no value is set, this is
+   * specified by the current locale.
+   */
+  @Prop() timeFormat?: '24h' | '12h' | null;
```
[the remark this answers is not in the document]
EinfachHans: Yes that correct, i named it '24h' because i think this sounds more logical for me as a user instead of `23` 🤔 Also the function is called `use24hour` so i thought it would be the correct naming. I understand the official hourCycle Reference but is this understandable for the users?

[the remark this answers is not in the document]
liamdebeasi: I knew what you meant, but I also have only ever worked in `h12` and `h23` time. I think this may be confusing for developers who are looking `h24` support. Additionally, if we do add `h24` support this API becomes more confusing. Documentation is going to be crucial here, so I can help write that.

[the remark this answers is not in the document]
liamdebeasi: Also given that these are the values used in browser APIs, I think these values will become more recognizable over time.

core/src/components/datetime/datetime.tsx
EinfachHans: I get this error by my idea when trying to adjust the function: 

**TS2339: Property 'hourCycle' does not exist on type 'ResolvedDateTimeFormatOptions'**

and found this: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/DateTimeFormat/resolvedOptions

Are you sure this is available via the `formatted.resolvedOptions()`?

core/src/components/datetime/datetime.tsx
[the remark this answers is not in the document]
EinfachHans: Yes accepted 😃 Over time would be better if this matches the browser API

core/src/components/datetime/datetime.tsx
EinfachHans: Or do we have to use this there instead?

```
let locale = new Intl.Locale("fr-FR-u-hc-h23");
console.log(locale.hourCycle);
```

liamdebeasi: MDN is out of date it looks like: https://caniuse.com/mdn-javascript_builtins_intl_datetimeformat_datetimeformat_options_parameter_options_hourcycle_parameter

EinfachHans: Trying this

`let locale = new Intl.Locale("fr-FR-u-hc-h23");`

also says **TS2339: Property 'Locale' does not exist on type 'typeof Intl'**

liamdebeasi: TypeScript doesn't recognize it, but it should: https://github.com/microsoft/TypeScript/issues/34399 

In that case, we could extend `ResolvedDateTimeFormatOptions` and cast the result as the custom interface:

```typescript
/**
 * Typescript 4.x does not recognize hourCycle as a valid option.
 *  See https://github.com/microsoft/TypeScript/issues/34399.
 */
interface DatetimeFormatOptions extends ResolvedDateTimeFormatOptions {
  hourCycle?: 'h11' | 'h12' | 'h23' | 'h24'
}

const options = formatted.resolvedOptions() as DatetimeFormatOptions;
```

EinfachHans: Okay i will add this

EinfachHans: We use the check here to determinate if we should use AM/PM or not and the hour values right?
Shouldn't we check then:

```
if (options.hourCycle !== undefined) {
    return options.hourCycle === 'h23' || options.hourCycle === 'h24';
  }
```

as the `h24` also should not show AM/PM currently?

liamdebeasi: I think for the purposes of this PR we can support `h12` and `h23` for now. I am not opposed to supporting `h11` and `h24` but it seems like `h12` and `h23` are the more common formats that Ionic developers are looking to use.

liamdebeasi: If you want to add support for `h11` and `h24` definitely feel free to add it, but that's not a requirement for this feature to be merged.

EinfachHans: Yes, i don't mean to add `h11` and `h24` in general, i meant if someone would pass a local like `fr-FR-u-hc-h24` the above implementation would show AM/PM

EinfachHans: This would be easily changed by added this to the check here:

```ts
/**
 * If hourCycle was not specified, check the locale
 * that is set on the user's device. We first check the
 * Intl.DateTimeFormat hourCycle option as developers can encode this
 * option into the locale string. Example: `en-US-u-hc-h23`
 */
const formatted = new Intl.DateTimeFormat(locale, { hour: 'numeric' });
const options = formatted.resolvedOptions() as DatetimeFormatOptions;
if (options.hourCycle !== undefined) {
  return options.hourCycle === 'h23' || options.hourCycle === 'h24';
}
```

EinfachHans:
Requested Changes pushed, i think if you want to have some specific docs and explanation here it would be easier if you directly push into this pr or give me the text as i'm not the best in writing understandable docs in english 😃

willmartindev (approved):
Really nice work on this PR.

liamdebeasi (approved):
Great job!

liamdebeasi:
Merged. Thank you!!

Issue #23703, opened 2021-07-30
bug:  ion-img can sometimes not load in on a lower-end CPU device due to Intersection Observer API behavior


Jacob-McKay:
### Prequisites

- [X] I have read the [Contributing Guidelines](https://github.com/ionic-team/ionic-framework/blob/main/.github/CONTRIBUTING.md#creating-an-issue).
- [X] I agree to follow the [Code of Conduct](https://ionicframework.com/code-of-conduct).
- [X] I have searched for [existing issues](https://github.com/ionic-team/ionic-framework/issues) that already report this problem, without success.

### Ionic Framework Version

- [ ] v4.x
- [X] v5.x
- [ ] v6.x

### Current Behavior

Sometimes (on a lower-end mobile device, or chrome browser with CPU throttling enabled), the ion-img will not display even when it is scrolled into view.  This is because the Intersection Observer API may pass in multiple items to the callback where the code assumes only one item (where the last item is the current value stating the img is in the viewport).

### Expected Behavior

If the ion-img has been supplied a src, and has been scrolled into view, I expect the image to load and be visible to the user.

### Steps to Reproduce

Linked is an experiment you can run with CPU Throttling enabled in your code, it's kind of what my app is doing, it has a bunch of images hooked up to the Intersection Observer API and you will see logged to the console the times that there is more than one item in the callback.

### Code Reproduction URL

https://codepen.io/Jacob-McKay/pen/LYympzR

liamdebeasi:
Thanks for the issue. Can you provide some additional steps to reproduce the issue? I tried your CodePen on a low end device with CPU throttling enabled. I was able to get multiple entries per IO callback, but the first entry (index zero) was the entry that stated the img is in the viewport, so I was not able to reproduce the exact behavior you described.

liamdebeasi:
I am looking at this a bit more and I think your solution in https://github.com/ionic-team/ionic-framework/pull/23704 is correct. The Intersection Observer entries go from oldest to newest, so the last entry should represent the latest state of the `ion-img`.

liamdebeasi:
Thanks for the issue. This has been resolved via https://github.com/ionic-team/ionic-framework/pull/23704, and a fix will be available in an upcoming release of Ionic Framework.

PR #23704, opened 2021-07-30
fix(img): correctly determine when to load image when scrolling quickly on slower devices

Jacob-McKay:
Fix for this issue: https://github.com/ionic-team/ionic-framework/issues/23703

There is a rare race condition you may experience on devices with slower CPU where the Intersection Observer API supplies multiple items to the callback, instead of only a single element which the previous code assumed it would get.  This changes takes the last item (which will be the most current information).

## Pull request checklist

## Pull request type

 

 

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

Issue Number: [23703](https://github.com/ionic-team/ionic-framework/issues/23703)

## What is the new behavior?

- Instead of keying off of the first item in the callback to the ion-img Intersection Observer API callback, take the last, most recent one to determine whether the img is on screen and should load

- [ ] Yes
- [X] No

### DIFF
--- a/core/src/components/img/img.tsx
+++ b/core/src/components/img/img.tsx
@@ -61,7 +61,7 @@ export class Img implements ComponentInterface {
       this.removeIO();
       this.io = new IntersectionObserver(data => {
         /**
-         * On slower devices, it is possible for an intersection observer entry to contain multiple 
+         * On slower devices, it is possible for an intersection observer entry to contain multiple
          * objects in the array. This happens when quickly scrolling an image into view and then out of
          * view. In this case, the last object represents the current state of the component.
          */
--- a/core/src/components/img/img.tsx
+++ b/core/src/components/img/img.tsx
@@ -60,7 +60,7 @@ export class Img implements ComponentInterface {
       'isIntersecting' in window.IntersectionObserverEntry.prototype) {
       this.removeIO();
       this.io = new IntersectionObserver(data => {
-        if (data[data.length-1].isIntersecting) {
+        if (data[data.length - 1].isIntersecting) {
           this.load();
           this.removeIO();
         }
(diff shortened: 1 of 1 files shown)


core/src/components/img/img.tsx
```diff
@@ -60,10 +60,7 @@ export class Img implements ComponentInterface {
       'isIntersecting' in window.IntersectionObserverEntry.prototype) {
       this.removeIO();
       this.io = new IntersectionObserver(data => {
-        // because there will only ever be one instance
-        // of the element we are observing
-        // we can just use data[0]
-        if (data[0].isIntersecting) {
+        if (data[data.length - 1].isIntersecting) {
```
liamdebeasi, line 63:
```suggestion
        /**
         * On slower devices, it is possible for an intersection observer entry to contain multiple 
         * objects in the array. This happens when quickly scrolling an image into view and then out of
         * view. In this case, the last object represents the current state of the component.
         */
        if (data[data.length - 1].isIntersecting) {
```

liamdebeasi (changes requested):

liamdebeasi (approved):
Thank you! This is good to merge when the tests pass.

core/src/components/img/img.tsx
```diff
@@ -60,10 +60,12 @@ export class Img implements ComponentInterface {
...
-        // because there will only ever be one instance
-        // of the element we are observing
-        // we can just use data[0]
-        if (data[0].isIntersecting) {
+        /**
+         * On slower devices, it is possible for an intersection observer entry to contain multiple 
+         * objects in the array. This happens when quickly scrolling an image into view and then out of
+         * view. In this case, the last object represents the current state of the component.
+         */
+        if (data[data.length - 1].isIntersecting) {
```
liamdebeasi, line 68:
```suggestion
        /**
         * On slower devices, it is possible for an intersection observer entry to contain multiple
         * objects in the array. This happens when quickly scrolling an image into view and then out of
         * view. In this case, the last object represents the current state of the component.
         */
        if (data[data.length - 1].isIntersecting) {
```

Jacob-McKay:
Ope, thought I that was a 'complete' button

willmartindev (approved):

liamdebeasi:
Merged. Thank you!

Issue #23777, opened 2021-08-17
bug: ion-nav animation always reset by router


MarkChrisLevy:
### Prequisites

- [X] I have read the [Contributing Guidelines](https://github.com/ionic-team/ionic-framework/blob/main/.github/CONTRIBUTING.md#creating-an-issue).
- [X] I agree to follow the [Code of Conduct](https://ionicframework.com/code-of-conduct).
- [X] I have searched for [existing issues](https://github.com/ionic-team/ionic-framework/issues) that already report this problem, without success.

### Ionic Framework Version

- [ ] v4.x
- [X] v5.x
- [ ] v6.x

### Current Behavior

When ion-nav has animation prop set to custom transition, the transition is never used when page change is triggered by ion-router.

### Expected Behavior

When ion-nav has animation prop, the custom transition should be used instead of default (mode specific) transition.

### Steps to Reproduce

1. Define router with some routes
2. For ion-nav set custom transition (e.g. when in md mode set animation=iosTransitionAnimation from "@ionic/core" package)
3. Trigger page change with ion-route-link or other component that uses ion-router for page changes

### Ionic Info

5.6.13

MarkChrisLevy:
PR is comming...

liamdebeasi:
Thanks! Looks like `animationBuilder` needs to be set after `...opts` here: https://github.com/ionic-team/ionic-framework/blob/main/core/src/components/nav/nav.tsx#L875

This is the same thing as what we did for `ion-router-outlet`: https://github.com/ionic-team/ionic-framework/blob/main/core/src/components/router-outlet/route-outlet.tsx#L230-L231

MarkChrisLevy:
@liamdebeasi exactly, I will make a PR in a moment

MarkChrisLevy:
@liamdebeasi Done :-)

liamdebeasi:
Thanks for the issue. This has been resolved via https://github.com/ionic-team/ionic-framework/pull/23779, and a fix will be available in an upcoming release of Ionic Framework.

PR #23779, opened 2021-08-17
fix(nav): custom animation is no longer overridden

MarkChrisLevy:
opts will always have animationBuilder prop when used with ion-router and this fix will make sure, that animationBuilder will be set with appropriate value

## Pull request checklist

## Pull request type

 

 

- [X] Bugfix
- [ ] Feature
- [ ] Build related changes

Issue Number: #23777

## What is the new behavior?
Use custom transition if set on ion-nav and page change triggered by ion-router.

- [ ] Yes
- [X] No

### DIFF
--- a/core/src/components/nav/nav.tsx
+++ b/core/src/components/nav/nav.tsx
@@ -872,14 +872,15 @@ export class Nav implements NavOutlet {
       mode,
       showGoBack: this.canGoBackSync(enteringView),
       baseEl: this.el,
-      animationBuilder: this.animation || opts.animationBuilder || config.get('navAnimation'),
       progressCallback,
       animated: this.animated && config.getBoolean('animated', true),
 
       enteringEl,
       leavingEl,
 
-      ...opts
+      ...opts,
+      
+      animationBuilder: opts.animationBuilder || this.animation || config.get('navAnimation')
     };
     const { hasCompleted } = await transition(animationOpts);
     return this.transitionFinish(hasCompleted, enteringView, leavingView, opts);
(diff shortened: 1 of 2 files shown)


liamdebeasi (approved):
Thank you! This is good to merge when tests pass.

core/src/components/nav/nav.tsx
```diff
@@ -872,14 +872,15 @@ export class Nav implements NavOutlet {
...
       progressCallback,
       animated: this.animated && config.getBoolean('animated', true),
 
       enteringEl,
       leavingEl,
 
-      ...opts
+      ...opts,
+
+      animationBuilder: opts.animationBuilder || this.animation || config.get('navAnimation')
```
liamdebeasi, line 883: Any reason why we switched the order here?

MarkChrisLevy: Transition passed in opts (e.g. it came from router) should take precedence over nav transition. Simple example - I set transition for nav, but in specific scenario, when moving from view A to B I need other transition, which can be passed to push() or set in router-link. Previous impl would always pick nav's transition.

liamdebeasi: Good catch! In that case, could we also make that same change here: https://github.com/ionic-team/ionic-framework/blob/main/core/src/components/router-outlet/route-outlet.tsx#L191

MarkChrisLevy: Done

liamdebeasi:
Merged. Thank you!

Issue #21039, opened 2020-04-14
feat: add bottom drawer


BerkeAras:
# Feature Request

**Ionic version:**

[x] **4.x**
[x] **5.x**

**Describe the Feature Request**
Please add a PopUp Footer like on Facebook or Slack

**Describe Preferred Solution**
Creating a Component (for Angular, React, Vue and JS) with a Pop-up Footer (See image)

**Describe Alternatives**
No Alternative yet.

**Related Code**
[https://jsfiddle.net/yznuotav/](https://jsfiddle.net/yznuotav/) It only works on Mobile (Or DevTools with Simulation)

**Additional Context**
[
![image](https://user-images.githubusercontent.com/36629890/79254951-cd663880-7e85-11ea-8312-8ddc35531a26.png)
](url)

roman-rr:
@BerkeAras 
I made especially for ionic applications, and do regular updates. 
https://github.com/roman-rr/cupertino-pane

Have also pull request with ion-drawer component which i guess will be merged if have many requests
#20408

BerkeAras:
Oh my god, that looks so great! Thanks! You’ve got a GitHub Star!!!

HansKrywaa:
Any Updates to this? I really really want this Feature and i'm not that happy with the existing cupertino-pane.

BerkeAras:
No updates at the moment. Why dont you like it?

HansKrywaa:
I wrote down some things here: https://github.com/roman-rr/cupertino-pane/issues/41

roman-rr:
Thank you for issues @HansKrywaa
I will review your requests in plugin repo.

BerkeAras:
With the release of iOS 14 at the WWDC bottom drawer are much more used.
Now it is in much more apps, or the homescreen.
I think, bottom drawers should be a part of ionic.

HansKrywaa:
Milestone 😍💪🏼 @BerkeAras

lincolnthree:
Wow @BerkeAras - This is impressive. Seriously I rarely see such a well-implemented/smooth animation with touch support. 

@liamdebeasi I'm sure you've thought of this already, but this could approach even be used to enhance the action-sheet such that it uses touch support to dismiss.

I known I don't get a vote in this, but this gets two of my votes.

liamdebeasi:
Thanks! This is something we are actively working on.

BerkeAras:
That is not my code. Thanks to @roman-rr for the great project

[...]

PR #23828, opened 2021-08-26
feat(modal): add bottom sheet support

liamdebeasi:
## Pull request checklist

## Pull request type

 

 

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes

Issue Number: resolves https://github.com/ionic-team/ionic-framework/issues/21039

## What is the new behavior?

- Add support for bottom sheet within `ion-modal`

- [ ] Yes
- [x] No


core/src/components/modal/usage/react.md
```diff
@@ -153,3 +153,47 @@ In most scenarios, setting a ref on `IonRouterOutlet` and passing that ref's `cu
...
+      </IonReactRouter>
+    </IonApp>
+  )
+};
+
+...
+
+interface HomePageProps {
+  router: HTMLIonRouterOutletElement | null;
+}
```
willmartindev, line 179: Is this content needed to explain the usage of `ion-modal`? The shorter the code block the better.

core/src/components/modal/usage/react.md
```diff
@@ -153,3 +153,47 @@ In most scenarios, setting a ref on `IonRouterOutlet` and passing that ref's `cu
...
+
+const Home: React.FC<HomePageProps> = ({ router }) => {
+  const [showModal, setShowModal] = useState(false);
+  
+  return (
+    ...
```
willmartindev, line 185:
```suggestion
    // ...
```

core/src/components/modal/usage/react.md
```diff
@@ -153,3 +153,47 @@ In most scenarios, setting a ref on `IonRouterOutlet` and passing that ref's `cu
...
+    
+    <IonModal
+      isOpen={showModal}
+      initialBreakpoint={0.5}
+      breakpoints={[0, 0.5, 1]}
+      onDidDismiss={() => setShowModal(false)}>
+      <p>This is modal content</p>
+    </IonModal>
+    
+    ...
```
willmartindev, line 195:
```suggestion
    // ...
```

willmartindev (commented):
Awesome work here! A couple small comments/questions from me.

core/src/components/modal/modal.scss
```diff
@@ -107,3 +110,35 @@
     --height: #{$modal-inset-height-large};
   }
 }
+
+// Sheet Modal
+// --------------------------------------------------
+
+.modal-handle {
```
willmartindev, line 117: Thoughts on adding a `grab`/`grabbing` cursor here?

willmartindev:
The backdrop isn't clickable until I grab and drag the modal for the first time:

https://user-images.githubusercontent.com/17113462/131285938-25dbe0d5-2113-41b9-b36a-747ccdda0a7c.mov

core/src/components/modal/modal.tsx
```diff
@@ -75,6 +80,30 @@ export class Modal implements ComponentInterface, OverlayInterface {
...
+   */
+  @Prop() breakpoints?: number[];
+
+  /**
+   * A decimal value between 0 and 1 that indicates the
+   * initial point the modal will open at when creating a
+   * sheet modal. This value must also be listed in the
+   * `breakpoints` array.
+   */
+  @Prop() initialBreakpoint?: number;
```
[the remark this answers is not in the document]
liamdebeasi: I think making the `initialBreakpoint` an index rather than a value would be confusing since the indices `0` and `1` could be confused for the a breakpoint values.

We could remove the requirement that `initialBreakpoint` is a value in the `breakpoints` array, but then we would have to figure out what the behavior of the initial breakpoint is: 

- Should `initialBreakpoint` be a point that users can snap to initially but can't snap to during the interaction (I.e. `initialBreakpoint` is not automatically added to the `breakpoints` array)
- Should `initialBreakpoint` be a pointer users can snap to during the interaction (I.e. `initialBreakpoint` is automatically added to the `breakpoints` array.

I would probably opt for the first option as users can get the second option by adding `initialBreakpoint` as a value in `breakpoints`.

All that said, I'm not sure there is a scenario where it makes sense to start at breakpoint `0.5`, but never let users snap back to that value. Thoughts?

core/src/components/modal/modal.scss
liamdebeasi: The bottom sheet is pretty mobile-focused so I think I would want to see the desktop use case first.

core/src/components/modal/modal.tsx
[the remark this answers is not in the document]
willmartindev: Yeah, regarding your last statement, I can't think of a reason why one would want an initial breakpoint that isn't accessible in subsequent interactions. That being said, I think the first bullet is preferable to the second bullet. 

Since you already have a console warning implemented it might be fine not to change anything, just wanted to bring this up to hear your thoughts. 🙏

willmartindev (approved):

Issue #23897, opened 2021-09-08
bug: angular outer tabs receive tab button click events for nested tabs


EinfachHans:
### Prequisites

- [X] I have read the [Contributing Guidelines](https://github.com/ionic-team/ionic-framework/blob/main/.github/CONTRIBUTING.md#creating-an-issue).
- [X] I agree to follow the [Code of Conduct](https://ionicframework.com/code-of-conduct).
- [X] I have searched for [existing issues](https://github.com/ionic-team/ionic-framework/issues) that already report this problem, without success.

### Ionic Framework Version

- [ ] v4.x
- [X] v5.x
- [ ] v6.x

### Current Behavior

I have the following page setup:

```
- Base Page (Ion Tabs)
  - Page 1 (Ion Tabs)
    - Page 1.1
    - Page 1.2
    - Page 1.3
  - Page 2
  - Page 3
```

The i use nested tabs. This works in general, but i get an error when i change on of the inner tabs:
```
Error: Uncaught (in promise): Error: Cannot match any routes. URL Segment: '...'
```

I debugged a bit and found that the `select()` function of the **IonTabs** is called two times, one for each Tabs, i can see that on the outlet.

### Expected Behavior

The IonTabs should only executed the `select` method for Tabs belongs to this.

### Steps to Reproduce

See page structure above

EinfachHans:
Here is a Repo: https://github.com/EinfachHans/ionic-nested-tabs-bug

Switching the outer tabs (bottom) not cause the error to be shown in console

Switching the inner tabs (top under outer tab 1) cause the error in the console to be shown

liamdebeasi:
I get the following error when trying to run your app:

```
[ng] ./src/app/tab1/tab1-routing.module.ts:12:36-70 - Error: Module not found: Error: Can't resolve './subtab1/subtab1.module' in '/Users/liamdebeasi/Ionic/ionic-nested-tabs-bug/src/app/tab1'
[ng] ./src/app/tab1/tab1-routing.module.ts:16:36-70 - Error: Module not found: Error: Can't resolve './subtab2/subtab2.module' in '/Users/liamdebeasi/Ionic/ionic-nested-tabs-bug/src/app/tab1'
[ng] ./src/app/tab1/tab1-routing.module.ts:20:36-70 - Error: Module not found: Error: Can't resolve './subtab3/subtab3.module' in '/Users/liamdebeasi/Ionic/ionic-nested-tabs-bug/src/app/tab1'
[...]

Issue #19918, opened 2019-11-15
bug: 


znotdead:
# Bug Report

**Ionic version:**

[x] **4.11.5**

**Current behavior:**
```this.tabs.getTab is not a function```

**Expected behavior:**
`getTab` should return tab component

**Steps to reproduce:**
on page with tabs:
`@ViewChild(IonTabs, { static: false }) tabs: IonTabs;`
And run this code:
```
const currentTab = this.tabs.getSelected();
console.log(this.tabs.getTab(currentTab));
```
This gives error: ```ERROR TypeError: "this.tabs.getTab is not a function"```
In console:
```error TS2339: Property 'getTab' does not exist on type 'IonTabs'.```

According  to docs: https://ionicframework.com/docs/api/tabs :
```
getTab
--
Get a specific tab by the value of its tab property or an element reference.
getTab(tab: string \| HTMLIonTabElement) => Promise<HTMLIonTabElement \| undefined>
```

**Related code:**

```
const currentTab = this.tabs.getSelected();
console.log(this.tabs.getTab(currentTab));
```

**Other information:**

**Ionic info:**

```
Ionic:

   Ionic CLI                     : 5.4.2
   Ionic Framework               : @ionic/angular 4.11.5
   @angular-devkit/build-angular : 0.803.8
   @angular-devkit/schematics    : 8.3.8
   @angular/cli                  : 8.3.19
   @ionic/angular-toolkit        : 2.0.0

Cordova:

   Cordova CLI       : 9.0.0 (cordova-lib@9.0.1)
   Cordova Platforms : ios 5.0.1
   Cordova Plugins   : cordova-plugin-ionic-keyboard 2.1.3, cordova-plugin-ionic-webview 4.1.0, (and 8 other plugins)

Utility:

   cordova-res : not installed
   native-run  : not installed

System:

   Android SDK Tools : 26.1.1 
   NodeJS            : v12.13.0 
   npm               : 6.12.0
   OS                : Linux 4.15
```

liamdebeasi:
Thanks for the issue. What are you hoping to be able to do with `getTab`? We weren’t clear of the benefit of it in Angular apps, so we didn’t expose it there. We’re willing to add it but would like to understand the use case first.

znotdead:
I have a page with tabs.
Situation: Page1 has opened Tab1.
When navigate away from Page1 there are lifecycle events fired up.
[...]

PR #23902, opened 2021-09-08
fix(angular): nested tabs now go to correct page

liamdebeasi:
## Pull request checklist

## Pull request type

 

 

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

Issue Number: resolves https://github.com/ionic-team/ionic-framework/issues/23897

Each tabs instance listens for `ionTabButtonClick`, so if you click a tab button in a child tabs instance, the parent tabs instance will also see that event and try to navigate. But since the event was really only meant for the child tab, the parent tab will try to navigate to a non-existent route.

## What is the new behavior?

- When a tabs instance gets the ionTabButtonClick event it will prevent the event  from bubbling so any tabs higher up do not also receive this event and navigate to the wrong url

- [ ] Yes
- [x] No

### DIFF
--- a/angular/src/directives/navigation/ion-tabs.ts
+++ b/angular/src/directives/navigation/ion-tabs.ts
@@ -86,10 +86,20 @@ export class IonTabs {
    *   b. If the last route view doesn't exist, then navigate
    *      to the default tabRootUrl
    */
-  @HostListener('ionTabButtonClick', ['$event.detail.tab'])
-  select(tab: string) {
+  @HostListener('ionTabButtonClick', ['$event'])
+  select(ev: CustomEvent) {
+    const tab = ev.detail.tab;
     const alreadySelected = this.outlet.getActiveStackId() === tab;
     const tabRootUrl = `${this.outlet.tabsPrefix}/${tab}`;
+
+    /**
+     * If this is a nested tab, prevent the event
+     * from bubbling otherwise the outer tabs
+     * will respond to this event too, causing
+     * the app to get directed to the wrong place.
+     */
+    ev.stopPropagation();
+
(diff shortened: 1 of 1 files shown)


willmartindev (approved):

kirillgroshkov:
FYI this broke our app. Also, it's inconsistent with the documentation: https://ionicframework.com/docs/api/tabs

Docs say that argument of `select` can be `string`. But in this PR (and in released `5.8.0`) it only accepts `CustomEvent` and breaks if you pass `string as any` to it.

We'll temporary pin to `5.7.0` until it's fixed.

image

liamdebeasi:
See https://github.com/ionic-team/ionic-framework/issues/23952. We have a proposed fix in https://github.com/ionic-team/ionic-framework/issues/23952#issuecomment-923010019. It would be really helpful if you could test the dev build and let us know if it fixes the issue. Thanks!

liamdebeasi:
Also the methods you see on the docs are for the vanilla JS `ion-tabs` component, not the `ion-tabs` component in Angular (Yes it is confusing). We are working on addressing this inconsistency in https://github.com/ionic-team/ionic-framework/issues/19918, but the `select` method not working with `HTMLIonTabElement` in Angular is not an issue caused by this PR.

kirillgroshkov:
> See #23952. We have a proposed fix in [#23952 (comment)](https://github.com/ionic-team/ionic-framework/issues/23952#issuecomment-923010019). It would be really helpful if you could test the dev build and let us know if it fixes the issue. Thanks!

I unfortunately don't have time to test it, but looking at the PR code I *expect* that it'll fix it 👍

Issue #22925, opened 2021-02-15
feat: extend CustomEvent types for ionic components


Spinnenzunge:
# Feature Request
Create typing for Ionics various EventEmitter that extend ES CustomEvent.

**Ionic version:**

[x] **5.x**

**Describe the Feature Request**

Right now if for example I want to safe type the event parameter passed to a callback function for the `ionInfinite` event emitter and I type it as CustomEvent, I get a ts compiler error as soon as I call `event.target.complete()` because `complete()` is not a valid method for the target object in `CustomEvent`.

As a fallback I have to type my passed event as `any` for it to work.

**Describe Preferred Solution**

Create types for ionics various EventEmitter that extend ES CustomEvent.

**Describe Alternatives**

None

**Related Code**

Template:
```
<ion-infinite-scroll threshold="100px" (ionInfinite)="loadData($event)">
```

Controller:
```
loadData(event: CustomEvent): void {
  event.target.complete();
}
```

liamdebeasi:
Thanks, I think we could do this. Something like:

```ts
interface InfiniteScrollEvent extends CustomEvent {
  target: HTMLIonInfiniteScrollElement;
}
```

for components that actually provide a payload (such as Searchbar):

```ts
interface SearchbarEventDetail {
  value: string | undefined;
}

interface SearchbarEvent extends CustomEvent {
  target: HTMLIonSearchbarElement;
  detail: SearchbarEventDetail;
}
```

Spinnenzunge:
Sounds great!

AliAlmutawakel:
Also, I think this might be related.

$event in angular is always of type `Event`

![image](https://user-images.githubusercontent.com/17656592/121729643-ef6c4280-caab-11eb-82d7-87e4f8497123.png)

when it's really of type `CustomEvent`. See this snippet of code from Ionic codebase:
```javascript
@ProxyCmp({ inputs: ["allowEmptySelection", "name", "value"] })
@Component({ selector: "ion-radio-group", changeDetection: ChangeDetectionStrategy.OnPush, template: "<ng-content></ng-content>", inputs: ["allowEmptySelection", "name", "value"] })
export class IonRadioGroup {
  ionChange!: EventEmitter<CustomEvent>; // <---- Emits CustomEvent not Event
  protected el: HTMLElement;
  constructor(c: ChangeDetectorRef, r: ElementRef, protected z: NgZone) {
    c.detach();
    this.el = r.nativeElement;
[...]

PR #23956, opened 2021-09-20
feat(all): add CustomEvents types to components that let you call methods on the target

liamdebeasi:
## Pull request checklist

## Pull request type

 

 

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes

Issue Number: resolves https://github.com/ionic-team/ionic-framework/issues/22925

Calling methods on Ionic components in events was tricky because devs would do `ev.target.complete()`, but `target` is of type `EventTarget`, so TypeScript would always complain that `complete()` did not exist.

## What is the new behavior?

- Added Custom Event types that type the `target` field properly for components that a) have methods on them and b) emit public events (so internal events like `ionStyle` were excluded)
- I excluded overlay components because devs typically do not call `dismiss` after hearing `ionActionSheetWillPresent`. (If this is a use case for a developer reading this, please create a feature request! We would love to learn more)
- For developers reading this, if you need one of these types added to a component the emits events but has no method, please create a feature request! We excluded types for these components because it did not really address the needs in https://github.com/ionic-team/ionic-framework/issues/22925.
-

- [x] Yes
- [ ] No

Technically a breaking change, but the interface that was removed has not been used since v4 as the behavior for radio group changed in v5. If developers used it today in v5 they would likely get type errors.


willmartian (approved):
Nice work! 🚀 🚀 🚀

Issue #23842, opened 2021-08-28
bug: v6 datetime, ionBlur and ionFocus not working


MaNic1980:
### Prequisites

- [X] I have read the [Contributing Guidelines](https://github.com/ionic-team/ionic-framework/blob/main/.github/CONTRIBUTING.md#creating-an-issue).
- [X] I agree to follow the [Code of Conduct](https://ionicframework.com/code-of-conduct).
- [X] I have searched for [existing issues](https://github.com/ionic-team/ionic-framework/issues) that already report this problem, without success.

### Ionic Framework Version

- [ ] v4.x
- [ ] v5.x
- [X] v6.x

### Current Behavior

I run in an issue with "focus in" and "focus out" of the new "ion-datetime" component.
I have an "div" around the "ion-datetime" and check if the focus is in or outside.
If i click on the days or time, my focus is in like expected, but on the month arrows or the year button, the focus got lost.

https://user-images.githubusercontent.com/66515430/131230483-73d88e3c-aa3a-4a8e-9193-2f134abce962.mp4

 

### Expected Behavior

I expected in the whole "ion-datetime" component, my focus is in.

### Steps to Reproduce

Click inside the day area and then on the month arrows and year button to check the focus.

```
<ion-label>{{label}}</ion-label>
<div tabindex="1" (focusin)="focusin()" (focusout)="focusout()">
    <ion-datetime></ion-datetime>
</div>
```

```
label: string;
focusin() {
    this.label="focusin";
  }

focusout() {
    this.label="focusout";
}
```

### Code Reproduction URL

https://stackblitz.com/edit/ionic-angular-v5-5kjnbu?file=src%2Fapp%2Fapp.component.html

### Ionic Info

Ionic:

   Ionic CLI                     : 6.17.0 (/usr/local/lib/node_modules/@ionic/cli)
   Ionic Framework               : @ionic/angular 6.0.0-beta.4
   @angular-devkit/build-angular : 12.1.4
   @angular-devkit/schematics    : 12.2.3
   @angular/cli                  : 12.1.4
   @ionic/angular-toolkit        : 4.0.0

Cordova:

   Cordova CLI       : 10.0.0
   Cordova Platforms : ios 6.2.0
   Cordova Plugins   : cordova-plugin-ionic-keyboard 2.2.0, cordova-plugin-ionic-webview 4.2.1, (and 4 other plugins)

Utility:

   cordova-res : not installed globally
   native-run  : 1.4.0

System:

[...]

PR #23980, opened 2021-09-24
fix(datetime): add ionBlur/ionFocus events to whole component

amandaesmith3:
## Pull request checklist

## Pull request type

 

 

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

Issue Number: Resolves https://github.com/ionic-team/ionic-framework/issues/23842

Events for `ionBlur` and `ionFocus` aren't firing when clicking in and out of `ion-datetime`.

## What is the new behavior?

- `ion-datetime` fires `ionFocus` and `ionBlur` events when clicking in and out of the component host.
- `ionFocus` and `ionBlur` events fired from components inside `ion-datetime`'s shadow DOM are now prevented from bubbling up, to avoid duplicate events appearing at the top level.

- [ ] Yes
- [x] No

## Other information

Even with this fix, clicking the `ion-button`s in `ion-datetime` (prev/next month, and buttons in the footer) will cause the datetime to lose focus in Safari (only). This does not apply to tabbing to the buttons instead -- focus is correctly maintained. There are two Webkit bugs driving the click issue:
- [22261](https://bugs.webkit.org/show_bug.cgi?id=22261): Clicking a button doesn't emit a focus event. Closed as "won't fix."
- [229895](https://bugs.webkit.org/show_bug.cgi?id=229895): Clicking a button causes focus to drop entirely (which causes the `ion-datetime` to emit a blur event).

### DIFF
--- a/core/src/components/button/button.tsx
+++ b/core/src/components/button/button.tsx
@@ -124,12 +124,12 @@ export class Button implements ComponentInterface, AnchorInterface, ButtonInterf
   /**
    * Emitted when the button has focus.
    */
-  @Event({ composed: false }) ionFocus!: EventEmitter<void>;
+  @Event() ionFocus!: EventEmitter<void>;
 
   /**
    * Emitted when the button loses focus.
    */
-  @Event({ composed: false }) ionBlur!: EventEmitter<void>;
+  @Event() ionBlur!: EventEmitter<void>;
 
   componentWillLoad() {
     this.inToolbar = !!this.el.closest('ion-buttons');
--- a/core/src/components/button/button.tsx
+++ b/core/src/components/button/button.tsx
@@ -124,12 +124,12 @@ export class Button implements ComponentInterface, AnchorInterface, ButtonInterf
   /**
    * Emitted when the button has focus.
    */
-  @Event() ionFocus!: EventEmitter<void>;
+  @Event({ composed: false }) ionFocus!: EventEmitter<void>;
 
   /**
    * Emitted when the button loses focus.
    */
-  @Event() ionBlur!: EventEmitter<void>;
+  @Event({ composed: false }) ionBlur!: EventEmitter<void>;
 
   componentWillLoad() {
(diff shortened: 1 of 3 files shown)


core/src/components/button/button.tsx
```diff
@@ -124,12 +124,12 @@ export class Button implements ComponentInterface, AnchorInterface, ButtonInterf
...
    * Emitted when the button has focus.
    */
-  @Event() ionFocus!: EventEmitter<void>;
+  @Event({ composed: false }) ionFocus!: EventEmitter<void>;
 
   /**
    * Emitted when the button loses focus.
    */
-  @Event() ionBlur!: EventEmitter<void>;
+  @Event({ composed: false }) ionBlur!: EventEmitter<void>;
```
liamdebeasi, line 132: Do we still need these?

amandaesmith3: Yeah, they prevent duplicate blur/focus events from firing on non-Safari browsers. (Explained in the second "new behavior" bullet)

amandaesmith3:
Just noticed I forgot to include "on Safari" in the other info section, oops. Fixed now!

core/src/components/button/button.tsx
liamdebeasi: I pushed a commit that should resolve this issue without changing `composed` on `ion-button` events. My main concern with the previous approach is that this is a breaking change to the behavior of a public API, and we try to make those only when absolutely necessary.

The code is at https://github.com/ionic-team/ionic-framework/pull/23980/commits/4eddaeb97b85055c5bdd57ff385d83872aeda98e - let me know your thoughts.

amandaesmith3: Aha, yeah that looks great 👍 I initially had the thought of using `stopPropagation`, but couldn't think of a clean way to do it that wouldn't involve adding listeners to a bunch of different spots. I'll update the PR description, and keep that in mind!

liamdebeasi (approved):

Issue #23857, opened 2021-09-01
bug: desktop select always focuses first item, does not account for selected item


farisfaisalthena:
### Prequisites

- [X] I have read the [Contributing Guidelines](https://github.com/ionic-team/ionic-framework/blob/main/.github/CONTRIBUTING.md#creating-an-issue).
- [X] I agree to follow the [Code of Conduct](https://ionicframework.com/code-of-conduct).
- [X] I have searched for [existing issues](https://github.com/ionic-team/ionic-framework/issues) that already report this problem, without success.

### Ionic Framework Version

- [ ] v4.x
- [ ] v5.x
- [X] v6.x

### Current Behavior

When using something like `(ngModel)` on a popover interface in `ion-select`, there will have two highlighted value one is the correct one and one will always be the first one. Screenshot as shown below

<img width="718" alt="Screenshot 2021-09-01 at 3 07 54 PM" src="https://user-images.githubusercontent.com/61012396/131627783-1ecec9cd-e291-4453-93c3-58b1747207bd.png">

### Expected Behavior

When a value is already choosen from `ion-select` the first value from the `ion-select` should not be highlighted as well.

### Steps to Reproduce

```html
<h2>Hover Issue</h2>
  <ion-item lines="inset">
    <ion-label>Time (Select One)</ion-label>
    <ion-select interface="popover" [ngModel]="15">
      <ion-select-option [value]="5">5</ion-select-option>
      <ion-select-option [value]="10">10</ion-select-option>
      <ion-select-option [value]="15">15</ion-select-option>
    </ion-select>
  </ion-item>
```

### Code Reproduction URL

https://github.com/farisfaisalthena/starter-repo

amandaesmith3:
Thanks for the report, I've resolved this via https://github.com/ionic-team/ionic-framework/pull/23991 and the fix will be available in a future release of Ionic.

PR #23991, opened 2021-09-28
fix(select): focus selected item in popovers

amandaesmith3:
## Pull request checklist

## Pull request type

 

 

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

Issue Number: resolves https://github.com/ionic-team/ionic-framework/issues/23857

When opening a popover-style select, the first item is focused, rather than the one that's currently selected.

## What is the new behavior?

- `select` component now focuses the selected option when opening a popover, or the first option if nothing is selected.
- `select` focusing is now handled at the component level instead of going through an unrelated overlays API.
- Tests added for `select` to ensure the first item is focused when the select has no value, and then when something is selected, that item is focused instead.
- Programmatic focusing pulled into helper function for both `select` and `overlays` to keep things DRY.

- [ ] Yes
- [x] No

### DIFF
--- a/core/src/components/select/select.tsx
+++ b/core/src/components/select/select.tsx
@@ -2,7 +2,7 @@ import { Component, ComponentInterface, Element, Event, EventEmitter, Host, Meth
 
 import { getIonMode } from '../../global/ionic-global';
 import { ActionSheetButton, ActionSheetOptions, AlertInput, AlertOptions, CssClassMap, OverlaySelect, PopoverOptions, SelectChangeEventDetail, SelectInterface, SelectPopoverOption, StyleEventDetail } from '../../interface';
-import { findItemLabel, getAriaLabel, renderHiddenInput } from '../../utils/helpers';
+import { findItemLabel, focusElement, getAriaLabel, renderHiddenInput } from '../../utils/helpers';
 import { actionSheetController, alertController, popoverController } from '../../utils/overlays';
 import { hostContext } from '../../utils/theme';
 import { watchForOptions } from '../../utils/watch-options';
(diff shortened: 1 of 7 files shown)


liamdebeasi (changes requested):
This is a great start! I think there is some room for simplification here. My primary concern with this PR is it modifies the behavior of `focusFirstDescendant` which is used to trap focus in overlays as an accessibility feature.

In hindsight, we probably should not have had `ion-select` use this API but I think we can correct that and tackle this issue all at once.

Can we try the following and see how it goes?

1. All we really need inside of `focusFirstDescendant` is the focus logic (I.e. the code block inside of `if (firstInput)`. We can extract this to a helper function to use in `ion-select`. Create this function in `utils/helpers` called `focusElement` that has the focus logic.

2. From here, the item that we need to focus is very specific to `ion-select`, so we should have `ion-select` be responsible for focusing the element. Inside of the `this.interface === 'popover'` check, let's try the following approach:

```js
const { childOpts, value } = this;
const indexOfSelected = childOpts.map(o => o.value).indexOf(value);

await (overlay as HTMLIonPopoverElement).present(event as PointerEvent);

const opt = // select correct item

if (opt) {
  focusElement(opt);
}
````

Feel free to play with the code on this a bit, but this approach should work. This lets us avoid additonal changes in `popover.tsx` as well as the changes to `focusFirstDescendant` in `utils/overlays`.

File: core/src/components/select/select.tsx [hunk 1]
```diff
@@ -179,11 +179,18 @@ export class Select implements ComponentInterface {
...
+    await overlay.present();
+
+    // focus selected option for popovers
     if (this.interface === 'popover') {
-      await (overlay as HTMLIonPopoverElement).presentFromTrigger(event, true);
-    } else {
-      await overlay.present();
+      let indexOfSelected = this.childOpts.map(o => o.value).indexOf(this.value);
+      indexOfSelected = indexOfSelected > -1 ? indexOfSelected : 0; // default to first option if nothing selected
+      const selectedEl = this.overlay.querySelector<HTMLElement>(`.select-interface-option:nth-child(${indexOfSelected + 1})`);
```
liamdebeasi, line 188:
```suggestion
      const selectedEl = overlay.querySelector<HTMLElement>(`.select-interface-option:nth-child(${indexOfSelected + 1})`);
```

Minor perf thing, but we already have a local reference to overlay so no need to access it in the `this` scope.

core/src/components/select/test/basic/e2e.ts
```diff
@@ -48,12 +48,28 @@ test('select: basic', async () => {
...
+  expect(popoverOption1).toHaveClass('ion-focused');
+
+  let popoverOption2 = await popover.find('.select-interface-option:nth-child(2)');
+  await popoverOption2.click();
+  await page.waitForTimeout(500);
+
+  await select.click();
+  popover = await page.find('ion-popover');
+  await popover.waitForVisible();
+  await page.waitForTimeout(250);
```
liamdebeasi, line 68: Why do we use `waitForTimeout` here?

amandaesmith3: I just included it since the other time the popover is opened has one too (on line 53); I assumed it was necessary to make sure the animation is finished 🤔

Issue #23768, opened 2021-08-13
feat: Better keyboard input support on web for ion-datetime


bchehraz:
### Prequisites

- [X] I have read the [Contributing Guidelines](https://github.com/ionic-team/ionic-framework/blob/main/.github/CONTRIBUTING.md#creating-an-issue).
- [X] I agree to follow the [Code of Conduct](https://ionicframework.com/code-of-conduct).
- [X] I have searched for [existing issues](https://github.com/ionic-team/ionic-framework/issues) that already include this feature request, without success.

### Describe the Feature Request

On web/desktop, it's not obvious that you must use the mouse-wheel or up-down arrow keys in order to select a time in the time input fields of `ion-datetime` (v6). It would be nice if a user could type numbers into the time fields as input.

### Describe the Use Case

The app I'm building is for a wide range of users in terms of age & demographic and so I could use as much support as I can get in terms of usability. I want to use the same `ion-datetime` component on both web and mobile, but I just want the web version of the component have better support to allow for typing numbers into the time input fields because it's just not obvious that the user must scroll or use up-down keys. 

### Describe Preferred Solution

Allow for an override when on web to allow keyboard input (0-9, ":") to fill in the time fields. Perhaps this should be an override option to override the normal behavior, so I can decide that if the user is on web, to override the behavior and make it input from keyboard.

### Additional Information

I haven't tried the component since beta v1 or 2 but looking at the beta docs, I haven't any reason to believe this is a feature yet, but if it is and I'm just missing it, please let me know. Or if this feature doesn't seem reasonable or if I'm missing the point with the component, let me know how I can accomodate my users. Thanks

willmartindev:
Thank you for the feature request, @bchehraz. I think this would be a good addition; let me discuss it with the team.

bchehraz:
Thank you. Keeping my fingers crossed 🤞

liamdebeasi:
Thanks for the issue. This has been resolved via https://github.com/ionic-team/ionic-framework/pull/23996, and a fix will be available in an upcoming release of Ionic Framework.

PR #23996, opened 2021-09-29
fix(datetime): time picker uses new iOS 15 style

liamdebeasi:
## Pull request checklist

## Pull request type

 

 

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes

Issue Number: resolves https://github.com/ionic-team/ionic-framework/issues/23768

## What is the new behavior?

- Adds keyboard support to time picker for typing in values
- Uses new iOS 15 wheel style

- [ ] Yes
- [x] No

### DIFF
--- a/core/src/components/datetime/utils/format.ts
+++ b/core/src/components/datetime/utils/format.ts
@@ -1,11 +1,7 @@
 import { DatetimeParts } from '../datetime-interface';
 
 const get12HourTime = (hour: number) => {
-  if (hour < 13) {
-    return hour;
-  }
-
-  return hour - 12;
+  return hour % 12 || 12;
 }
 
 const getFormattedAMPM = (ampm?: string) => {
--- a/core/src/components/datetime/utils/format.ts
+++ b/core/src/components/datetime/utils/format.ts
@@ -1,5 +1,32 @@
 import { DatetimeParts } from '../datetime-interface';
 
+const get12HourTime = (hour: number) => {
+  if (hour < 13) {
+    return hour;
+  }
+
+  return hour - 12;
+}
+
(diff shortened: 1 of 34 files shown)


core/src/components/datetime/utils/format.ts
```diff
@@ -1,5 +1,28 @@
...
+}
+
+const getFormattedAMPM = (ampm?: string) => {
+  if (ampm === undefined) { return ''; }
+
+  return ampm.toUpperCase();
+}
+
+export const getFormattedTime = (refParts: DatetimeParts, use24Hour: boolean): string => {
+  if (refParts.hour === undefined || refParts.minute === undefined) { return 'Invalid Time'; }
```
amandaesmith3, line 14: Should this also warn if the user tries to pass a logistically invalid time, like `25` for the hour?

amandaesmith3 (changes requested):
This looks great! I only found a couple nitpicky things; honestly I could see this being shipped as-is.

core/src/components/datetime/utils/format.ts
liamdebeasi: We could, but the `refParts` that gets passed in is created via users selecting options in datetime that we create so there's no user generated content there. I believe datetime will error out if you pass in an invalid ISO 8601 string into the `value` field for user generated content. I can add a check if you think it would be valuable though.

amandaesmith3: Nah, if the only way to enter a custom time is through an ISO string, we should be good 👍 Just wanted to make sure.

amandaesmith3 (approved):

willmartian (approved):

Issue #22853, opened 2021-02-02
bug: ionic angular, karma tests not waiting for ionic to load on live-reloads


bohndthi:
# Bug Report

**Ionic version:**
[x] **5.x**

**Current behavior:**
If a page contains an ion-segment, the karma tests throws an long error message:

```
Chrome Headless 88.0.4324.50 (Linux x86_64): Executed 2 of 111 (1 FAILED) (0 secs / 0.339 secs)
ERROR: TypeError: config.get is not a function
TypeError: config.get is not a function
    at addEventListener (http://localhost:9876/_karma_webpack_/webpack:/node_modules/@ionic/core/dist/esm/helpers-90f46169.js:29:1)
    at SegmentButton.connectedCallback (http://localhost:9876/_karma_webpack_/webpack:/node_modules/@ionic/core/dist/esm/ion-segment_2.entry.js:375:23)
    at safeCall (http://localhost:9876/_karma_webpack_/webpack:/node_modules/@ionic/core/dist/esm/index-e806d1f6.js:1261:1)
    at fireConnectedCallback (http://localhost:9876/_karma_webpack_/webpack:/node_modules/@ionic/core/dist/esm/index-e806d1f6.js:1687:1)
    at initializeComponent (http://localhost:9876/_karma_webpack_/webpack:/node_modules/@ionic/core/dist/esm/index-e806d1f6.js:1650:1)
    at http://localhost:9876/_karma_webpack_/webpack:/node_modules/@ionic/core/dist/esm/index-e806d1f6.js:1750:1
    at ZoneDelegate.invoke (http://localhost:9876/_karma_webpack_/webpack:/node_modules/zone.js/dist/zone-evergreen.js:368:1)
    at ProxyZoneSpec.push.QpwO.ProxyZoneSpec.onInvoke (http://localhost:9876/_karma_webpack_/webpack:/node_modules/zone.js/dist/zone-testing.js:300:1)
    at ZoneDelegate.invoke (http://localhost:9876/_karma_webpack_/webpack:/node_modules/zone.js/dist/zone-evergreen.js:367:1)
```

**Expected behavior:**
no long error messages
**Steps to reproduce:**

- Create new project (angular)
- add an ion-segment to a page
- run the karma unit test for the page

**Ionic info:**

@ionic/angular: 5.5.3

liamdebeasi:
[...]

PR #24004, opened 2021-10-01
fix(angular): config.get not a function in karma (#22853)

studioromeo:
## Pull request checklist

## Pull request type

 

 

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

When angular bootstraps it sets `Ionic.config` regardless of whether it is set or not. This is fine in normal applications such as serving in the browser where angular is only initialised once. 

However in karma test suites angular is initialised on every test case but the window object persists between all test cases which causes ionic/angular to overwrite the config with something that is not compatible with the core components. This has been reported in #22853 as config.get throwing warnings when using ion-segment but I guess the issue could be for any components that use the config object.

Issue Number: 22853

## What is the new behavior?
Now during startup ionic/angular checks to see if `Ionic.config` is set. If it is it does not do anything with that variable. This seems to solve the problem described in #22853 and I've not noticed any adverse effects when stepping through the event listeners in [helpers](https://github.com/ionic-team/ionic-framework/blob/25eb8cdf98fe455433ca6185e89d9e1223a6d3ae/core/src/utils/helpers.ts#L71) that were previously failing because the config had been overwritten with another object

-
-
-

- [ ] Yes
- [x] No

## Other information

When inspecting Ionic.config on a second test case

Before | After
--- | ---
Screenshot 2021-10-01 at 15 30 04 | Screenshot 2021-10-01 at 15 28 48

PS: I wasn't able to find any tests for this part and running npm test gave me `angular no tests yet`. If theres some tests I need to write let me know and also how I should run the test suite.

### DIFF
--- a/angular/src/app-initialize.ts
+++ b/angular/src/app-initialize.ts
@@ -11,10 +11,14 @@ export const appInitialize = (config: Config, doc: Document, zone: NgZone) => {
     if (win && typeof (window as any) !== 'undefined') {
       const Ionic = win.Ionic = win.Ionic || {};
 
-      Ionic.config = {
-        ...config,
-        _zoneGate: (h: any) => zone.run(h)
-      };
+      // Only set the config key/value if one does not already exist
+      // Subsequent initilizations (eg in karma) should not overwrite the config if exists already in the window
+      if (!Ionic.config) {
+        Ionic.config = {
+          ...config,
+          _zoneGate: (h: any) => zone.run(h)
+        };
+      }
 
       const aelFn = '__zone_symbol__addEventListener' in (doc.body as any)
         ? '__zone_symbol__addEventListener'


liamdebeasi:
Thanks for the PR! We appreciate the work you put into making this.

I pushed a commit to your branch with an alternative way of fixing this issue. Ionic React and Ionic Vue call an `initialize` function exported from Ionic Core which lets you initialize (or re-initialize) the config without destroying the old `Config` object. This is a relatively new function, so Ionic Angular was never using it. Tests seem to work fine with this change. Additionally, developers will be able to change configs in between tests without any errors.

Once the build passes I will merge this PR. Thanks again!

liamdebeasi:
Here is a dev build if you are interested in testing:
```
npm install @ionic/angular@5.9.0-dev.202110012020.519054b
```

liamdebeasi:
Merged. Thank you!

studioromeo:
Hey. I gave the dev build a test and all seemed to work as expected, thanks for the merge 🥳

Issue #24049, opened 2021-10-08
bug: ion-label seems invalid (red color) with angular validators when loading view 


enrique-lozano:
### Prerequisites

- [X] I have read the [Contributing Guidelines](https://github.com/ionic-team/ionic-framework/blob/main/.github/CONTRIBUTING.md#creating-an-issue).
- [X] I agree to follow the [Code of Conduct](https://ionicframework.com/code-of-conduct).
- [X] I have searched for [existing issues](https://github.com/ionic-team/ionic-framework/issues) that already report this problem, without success.

### Ionic Framework Version

- [ ] v4.x
- [ ] v5.x
- [X] v6.x

### Current Behavior

Well, the truth is that this bug is difficult for me to explain. Simply by using the angular validators, if we make a form with a required input when loading the page, the label appears in red, but the bottom line does not. Once we focus on the input if the bar below is colored red as it should be, since the user has not yet entered anything, as well as when the focus is lost if the input is empty. However, as I said, when loading the app that label should not be red.

This is something that did not happen with ionic 5.8

I attach code with this behavior

### Expected Behavior

The text on the label should have a normal color

### Steps to Reproduce

See the attached code

### Code Reproduction URL

https://github.com/enriqueloz88/label-bug-ionic6

### Ionic Info

Ionic:

   Ionic CLI                     : 6.17.1 (C:\Users\kikel\AppData\Roaming\npm\node_modules\@ionic\cli)
   Ionic Framework               : @ionic/angular 6.0.0-rc.0 (C:\Users\kikel\Documents\test-input\node_modules\@ionic\angular)
   @angular-devkit/build-angular : 12.1.4 (C:\Users\kikel\Documents\test-input\node_modules\@angular-devkit\build-angular)
   @angular-devkit/schematics    : 12.1.4 (C:\Users\kikel\Documents\test-input\node_modules\@angular-devkit\schematics)
   @angular/cli                  : 12.1.4 (C:\Users\kikel\Documents\test-input\node_modules\@angular\cli)
   @ionic/angular-toolkit        : 4.0.0 (C:\Users\kikel\Documents\test-input\node_modules\@ionic\angular-toolkit)

Capacitor:

   Capacitor CLI      : not installed
   @capacitor/android : not installed
   @capacitor/core    : 3.2.4 (C:\Users\kikel\Documents\test-input\node_modules\@capacitor\core)
   @capacitor/ios     : not installed

Utility:

   cordova-res : 0.15.3
[...]

PR #23354, opened 2021-05-24
feat(item): add helper text, error text, counter, shape, and fill mode

willmartindev:
## Pull request checklist

Please check if your PR fulfills the following requirements:
- [x] Tests for the changes have been added (for bug fixes / features)
- [x] Docs have been reviewed and added / updated if needed (for bug fixes / features)
- [x] Build (`npm run build`) was run locally and any changes were pushed
- [x] Lint (`npm run lint`) has passed locally and any fixes were made for failures

## Pull request type

 

 

Please check the type of change your PR introduces:
- [ ] Bugfix
- [x] Feature
- [ ] Code style update (formatting, renaming)
- [ ] Refactoring (no functional changes, no api changes)
- [ ] Build related changes
- [ ] Documentation content changes
- [ ] Other (please describe): 

## What is the current behavior?

Issue Number: resolves #19619 

## What is the new behavior?

- Updated `ion-item` nested text inputs to match the current Material Design specification.
- Added ability to display helper text.
- Added ability to display error text.
- Added ability to display an input length counter. 
- Added support for two fill modes, `solid` and `outline`.
- Added `shape` property.

## Does this introduce a breaking change?

- [ ] Yes
- [x] No

## Other information

Material Design specification for text fields: https://material.io/components/text-fields

PR #24072, opened 2021-10-14
fix(label): only apply error appearance when control is touched

sean-perkins:
## Pull request checklist

## Pull request type

 

 

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

Currently the Material design appearance of an invalid `ion-label` in an `ion-item`, is applying the error appearance; even though the control has not been interacted with.

Issue Number: #24049

## What is the new behavior?

Fixes the change introduced within this [PR](https://github.com/ionic-team/ionic-framework/pull/23354).

- Aligns v6 label appearance rules to match 5.8.*
- Only applies the invalid error state when the control is touched/interacted with.

- [ ] Yes
- [x] No

## Other information

Display state of item when `ion-invalid` is initially set.

|Before|After|
|----|----|
|Screen Shot 2021-10-13 at 11 23 22 PM |Screen Shot 2021-10-13 at 11 23 00 PM|

I'm unsure of the implications with React/Vue. As far as I know, they don't have a form integration coupled with Ionic. This change is technically only necessary for Angular implementations, unless React/Vue are manually managing `ion-touched` on the `ion-item`. 

I could see it being advantageous to leave the web component alone and approach fixing this issue from the framework side, so the value accessor would instead apply `ion-invalid` when the control is invalid and dirty. Let me know if that would be a better fit to align with React/Vue as well.

### DIFF
--- a/core/src/components/label/label.md.scss
+++ b/core/src/components/label/label.md.scss
@@ -137,8 +137,8 @@
   color: #{current-color(base)};
 }
 
-:host-context(.ion-invalid).label-stacked:not(.ion-color),
-:host-context(.ion-invalid).label-floating:not(.ion-color) {
+:host-context(.ion-invalid.ion-touched).label-stacked:not(.ion-color),
+:host-context(.ion-invalid.ion-touched).label-floating:not(.ion-color) {
   color: var(--highlight-color-invalid);
 }
 


liamdebeasi (commented):
This looks great! When you have your environment setup, could you create a development build of Ionic Angular and test this in the example app provided in https://github.com/ionic-team/ionic-framework/issues/24049.

Also let's make sure that this does not break form validation colors when the form is submitted.

sean-perkins:
> This looks great! When you have your environment setup, could you create a development build of Ionic Angular and test this in the example app provided in #24049.
> 
> Also let's make sure that this does not break form validation colors when the form is submitted.

Confirmed in #24049 that on initial load with Android/Material design, the form label no longer displays an invalid state, until after interaction/touch.

This resolves the spec requirement for [user input patterns](https://material.io/archive/guidelines/patterns/errors.html#errors-usage).

However, after submit no error indications are present, unless in Angular a developer marks all controls as touched on submit.

```ts
onSubmit() {
    // form is the form group variable
    this.form.markAllAsTouched();
}
```

To my knowledge, framework has historically no implementation around form submissions or Angular's implementation of `ng-submitted` class. We could possibly ship Angular's framework implementation with a directive attached to `[formGroup],[formGroupName]` that would allow us to bind to submit and pass that context to the individual controls in the group.

liamdebeasi (approved):

Issue #23910, opened 2021-09-09
bug: datetime month and year picker not initialized properly on ios 14


wangzhengjiang:
### Prequisites

- [X] I have read the [Contributing Guidelines](https://github.com/ionic-team/ionic-framework/blob/main/.github/CONTRIBUTING.md#creating-an-issue).
- [X] I agree to follow the [Code of Conduct](https://ionicframework.com/code-of-conduct).
- [X] I have searched for [existing issues](https://github.com/ionic-team/ionic-framework/issues) that already report this problem, without success.

### Ionic Framework Version

- [ ] v4.x
- [ ] v5.x
- [X] v6.x

### Current Behavior

when using ion-datetime v6 on iphone, choice months, program will always return previous month of the selected month. 

### Expected Behavior

it should return the month of selected.

### Steps to Reproduce

1. created a new app using V6
2. add ion-datetime in the pages
3. deploy to iphone. 
4. run in xcode, when choice month, program will return previous month instead of the month of selected. 

### Ionic Info

Ionic:

   Ionic CLI                     : 6.17.0 (/usr/local/lib/node_modules/@ionic/cli)
   Ionic Framework               : @ionic/angular 6.0.0-beta.5
   @angular-devkit/build-angular : 12.1.4
   @angular-devkit/schematics    : 12.1.4
   @angular/cli                  : 12.1.4
   @ionic/angular-toolkit        : 4.0.0

Capacitor:

   Capacitor CLI      : 3.2.0
   @capacitor/android : 3.2.2
   @capacitor/core    : 3.2.0
   @capacitor/ios     : 3.2.2

Utility:

   cordova-res                          : not installed globally
   native-run (update available: 1.4.1) : 1.4.0

System:

   NodeJS : v14.17.5 (/usr/local/bin/node)
   npm    : 7.22.0
   OS     : macOS Big Sur

liamdebeasi:
Thanks for the issue. Can you reproduce this in an Ionic starter app and provide a link to the repo?

wangzhengjiang:
I created a app, can reproduce the error on iphone. 

https://github.com/wangzhengjiang/testdatetime

liamdebeasi:
Thanks for the reply. I do not see any error logged in the console, and the dates appear to be returning correctly for me. Can you please clarify the steps to reproduce the issue?

wangzhengjiang:
[...]

PR #24142, opened 2021-11-01
fix(datetime): resolve month and year jumping issue on ios

liamdebeasi:
## Pull request checklist

## Pull request type

 

 

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

Issue Number: resolves https://github.com/ionic-team/ionic-framework/issues/23910

## What is the new behavior?

- Instead of using `display: none` we hide the main content offscreen and set `visibility: hidden` to avoid the additional Intersection Observer callback. I did a brief paint test and there should be no significant changes to painting in terms of performance.

- [ ] Yes
- [x] No

### DIFF
--- a/core/src/components/datetime/datetime.scss
+++ b/core/src/components/datetime/datetime.scss
@@ -90,7 +90,12 @@
 
   position: absolute;
 
-  opacity: 0;
+  /**
+   * Use visibility instead of
+   * opacity to ensure element
+   * cannot receive focus
+   */
+  visibility: hidden;
   pointer-events: none;
 }
 
--- a/core/src/components/datetime/datetime.scss
+++ b/core/src/components/datetime/datetime.scss
@@ -73,11 +73,25 @@
   display: flex;
 }
 
+/**
+ * Safari 14 has an issue where Intersection
+ * Observer is incorrectly fired when
+ * unhiding the calendar content.
+ * To workaround this, we set the opacity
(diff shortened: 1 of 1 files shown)


core/src/components/datetime/datetime.scss
```diff
@@ -73,11 +73,25 @@
...
+ * of the content to 0 and hide it offscreen.
+ * TODO: This is fixed in Safari 15+, so remove
+ * when Safari 14 support is dropped.
+ */
 :host(.show-month-and-year) .calendar-next-prev,
 :host(.show-month-and-year) .calendar-days-of-week,
 :host(.show-month-and-year) .calendar-body,
 :host(.show-month-and-year) .datetime-time {
-  display: none;
+  @include position(null, null, null, -99999px);
```
amandaesmith3, line 89: It's still possible to tab to the calendar body like this, so you have to hit tab several times to get from the month/year dropdown to the pickers. Does `visibility: hidden` still get around the intersection observer issue? Otherwise we might need to involve the JS too and add `tabindex="-1"` or something.

amandaesmith3 (changes requested):

core/src/components/datetime/datetime.scss
liamdebeasi: Visibility works. Good catch!

amandaesmith3 (approved):

Issue #23728, opened 2021-08-04
bug: v6, modal core delegate needs to account for modals placed at the root, hardware back button not working with modal


d00ML0rDz:
### Prequisites

- [X] I have read the [Contributing Guidelines](https://github.com/ionic-team/ionic-framework/blob/main/.github/CONTRIBUTING.md#creating-an-issue).
- [X] I agree to follow the [Code of Conduct](https://ionicframework.com/code-of-conduct).
- [X] I have searched for [existing issues](https://github.com/ionic-team/ionic-framework/issues) that already report this problem, without success.

### Ionic Framework Version

- [ ] v4.x
- [ ] v5.x
- [X] v6.x

### Current Behavior

When setting up a fresh Ionic project and installing the v6 beta, the hardware back button on android doesn't navigate back when switching between the tabs if an IonModal has been added outside of IonReactRouter.

An example of why you might have the Modal outside of the ReactRouter is it may need to be opened on whatever page the user is on.

This was previously working fine in Ionic v5.

### Expected Behavior

The hardware back button should navigate back through the history.

### Steps to Reproduce

1. Pull the repo attached noticing the changes to the App.tsx underneath the IonReactRouter tag, where an IonModal has been added:
```
<IonModal isOpen={true}>
    Test
</IonModal>
```

2. Build and deploy the app to an android device. You should see that using the hardware back button does not go back when navigating between the 3 tabs and the modal does not display.

### Code Reproduction URL

https://github.com/d00ML0rDz/ionic-back-button-modal

### Ionic Info

Ionic:

   Ionic CLI       : 6.16.1 (C:\Users\jammi\AppData\Roaming\npm\node_modules\@ionic\cli)
   Ionic Framework : @ionic/react 6.0.0-beta.2

Capacitor:

   Capacitor CLI      : 3.1.2
   @capacitor/android : 3.1.2
   @capacitor/core    : 3.1.2
   @capacitor/ios     : not installed

Utility:

   cordova-res : 0.15.3
   native-run  : 1.4.0

System:

   NodeJS : v14.15.2 (C:\Program Files\nodejs\node.exe)
   npm    : 6.14.9
   OS     : Windows 10

d00ML0rDz:
[...]

PR #24165, opened 2021-11-04
fix(overlays): declarative modals now work properly with the hardware back button

liamdebeasi:
## Pull request checklist

## Pull request type

 

 

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes

This is a partial fix for https://github.com/ionic-team/ionic-framework/issues/23728

The hardware back button (HBB) listener for overlays always dismissed the last-presented (I.e. top most) overlay. The problem was that it did this regardless of whether or not that overlay was actually presented. So if you created and presented modal A, then created but did NOT present modal B, pressing the HBB would try to dismiss modal B, leaving modal A open.

This has been the behavior in Ionic for years, but it was never a big issue until the declarative overlays were introduced in Ionic 6. With the declarative overlays, the overlays are always in the DOM and so users started running into this issue more often.

Note: I will fix the other issue with the core delegate in a separate PR.

## What is the new behavior?

- Both the hardware back button and Esc listener only accounts for the top-most open overlay when trying to dismiss. To do this we exclude any overlay with the `.overlay-hidden` class. This class is added when an overlay is initially created, and is removed when the `present()` method is called. The class is then re-added when the `dismiss()` method is called.

- [ ] Yes
- [x] No

### DIFF
--- a/core/src/components/action-sheet/action-sheet.tsx
+++ b/core/src/components/action-sheet/action-sheet.tsx
@@ -252,6 +252,7 @@ export class ActionSheet implements ComponentInterface, OverlayInterface {
           [mode]: true,
 
           ...getClassMap(this.cssClass),
+          'overlay-hidden': true,
           'action-sheet-translucent': this.translucent
         }}
         onIonActionSheetWillDismiss={this.dispatchCancelHandler}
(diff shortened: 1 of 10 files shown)


amandaesmith3 (approved):
Looks good to me!

PR #24174, opened 2021-11-05
fix(modal, popover): opening modal and popover now works even if overlay was added to ion-app directly

liamdebeasi:
## Pull request checklist

## Pull request type

 

 

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

Issue Number: resolves https://github.com/ionic-team/ionic-framework/issues/23728

## What is the new behavior?

- Modal now opens correctly when placed directly inside of ion-app
- Shadows/backdrop opacities are correctly set for the first open modal
- This changes also fixes a "gotcha" I missed when building declarative modals. If you are creating a modal via `document.createElement`, then the modal is no longer automatically removed from the DOM when the modal is dismissed. Impact on this is low since most developers use `modalController`.

Note: I couldn't think of a better way of doing this fix without the `hasController` internal prop, so if you have other ideas definitely let me know!

- [ ] Yes
- [x] No

### DIFF
--- a/core/src/components/modal/modal.tsx
+++ b/core/src/components/modal/modal.tsx
@@ -325,10 +325,10 @@ export class Modal implements ComponentInterface, OverlayInterface {
      * If using overlay inline
      * we potentially need to use the coreDelegate
      * so that this works in vanilla JS apps.
-     * If a user has already placed the overlay
-     * as a direct descendant of ion-app or
-     * the body, then we can assume that
-     * the overlay is already in the correct place.
+     * If a developer has presented this component
+     * via a controller, then we can assume
+     * the component is already in the
+     * correct place.
      */
     const parentEl = this.el.parentNode as HTMLElement | null;
     const inline = this.inline = parentEl !== null && !this.hasController;
--- a/core/src/components/modal/modal.tsx
+++ b/core/src/components/modal/modal.tsx
@@ -61,6 +61,9 @@ export class Modal implements ComponentInterface, OverlayInterface {
   @Element() el!: HTMLIonModalElement;
 
   /** @internal */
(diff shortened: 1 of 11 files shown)


core/src/components/modal/modal.tsx
```diff
@@ -328,7 +331,7 @@ export class Modal implements ComponentInterface, OverlayInterface {
      * the overlay is already in the correct place.
```
amandaesmith3, line 331: This comment should be updated to reflect the new check.

core/src/utils/overlays.ts
```diff
@@ -54,7 +54,7 @@ export const createOverlay = <T extends HTMLIonOverlayElement>(tagName: string,
        * Convert the passed in overlay options into props
        * that get passed down into the new overlay.
        */
-      Object.assign(element, { ...opts });
+      Object.assign(element, { ...opts, hasController: true });
```
amandaesmith3, line 57: Does the `hasController` prop need to be added to other overlay components besides `modal` and `popover`? Or is the issue specific to those two? This function gets used in all overlays from what I can tell.

amandaesmith3 (changes requested):
Had a minor change and a question below. I couldn't think of a better way than the `hasController` prop either; since you can theoretically insert a modal anywhere, we can't rely on location in the DOM, and there isn't an inherent difference between modal elements of different types besides the new prop. So I think we're good there.

core/src/utils/overlays.ts
liamdebeasi: Not yet - the other overlays can't be used declaratively at the moment. Once they are we should add the prop though.

amandaesmith3 (approved):

Issue #24187, opened 2021-11-09
bug: v6 ion-textearea with item fill outline + autogrow makes contents grow outside of container


jesperbjerke:
### Prerequisites

- [X] I have read the [Contributing Guidelines](https://github.com/ionic-team/ionic-framework/blob/main/.github/CONTRIBUTING.md#creating-an-issue).
- [X] I agree to follow the [Code of Conduct](https://ionicframework.com/code-of-conduct).
- [X] I have searched for [existing issues](https://github.com/ionic-team/ionic-framework/issues) that already report this problem, without success.

### Ionic Framework Version

- [ ] v4.x
- [ ] v5.x
- [X] v6.x

### Current Behavior

When using the new `fill="outline"` on an item that contains an ion-textearea with `auto-grow="true"` makes the contents of the ion-textarea grow outside of the container. Removing the `fill="outline"` makes it work properly again.

### Expected Behavior

Contents should be contained within the textearea, like when not using `fill="outline"` 

### Steps to Reproduce

```html
<ion-item mode="md" fill="outline">
    <ion-label position="floating">This is my label</ion-label>
    <ion-textarea auto-grow="true"></ion-textarea>
</ion-item>
```

### Ionic Info

Ionic:

   Ionic CLI                     : 6.18.0 (/usr/local/lib/node_modules/@ionic/cli)
   Ionic Framework               : @ionic/angular 6.0.0-rc.2
   @angular-devkit/build-angular : 12.2.13
   @angular-devkit/schematics    : 12.2.13
   @angular/cli                  : 12.2.13
   @ionic/angular-toolkit        : 5.0.3

Capacitor:

   Capacitor CLI      : 3.3.1
   @capacitor/android : 3.3.1
   @capacitor/core    : 3.3.1
   @capacitor/ios     : 3.3.1

Utility:

   cordova-res : not installed globally
   native-run  : 1.5.0

System:

   NodeJS : v16.13.0 (/usr/local/bin/node)
   npm    : 8.1.2
   OS     : macOS Monterey

sean-perkins:
@jesperbjerke thanks for reporting this issue!

On an initial look, it's the combination of three separate factors that are causing this issue:
1. Using an `ion-item` with `fill="outline"`
2. Using an `ion-textarea` with `auto-grow="true"`
3. Using an `ion-label` with `position="floating"`

The ion-label is the main factor, as adjusting the position value to any value other than `floating` or removing the label, results in a proper display.

[...]

PR #24202, opened 2021-11-11
fix(textarea): floating label with autogrow textareas

sean-perkins:
## Pull request checklist

## Pull request type

 

 

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

Both inputs and textareas that are slotted through an `ion-item` using `fill="outline"` and an `ion-label` with `position="floating"` are always translated upwards by 25% of the element's height. With a textarea using auto grow, the contents of the textarea will translate outside of the container.

Issue Number: #24187

## What is the new behavior?

Within MD regardless of input or textarea, the control container will translate upwards of 14px, to offset the animation distance of the floating label. This allows auto grow textarea to display consistently.

- [ ] Yes
- [x] No

## Other information

Split PR #24196 to focus on the overflow issue.

### DIFF
--- a/core/scripts/screenshot/ionic.js
+++ b/core/scripts/screenshot/ionic.js
@@ -11,9 +11,7 @@ class IonicConnector extends ScreenshotConnector {
     const ws = fs.createWriteStream(this.masterBuildFilePath);
     const p = `/data/builds/master.json?ts=${Date.now()}`;
     await this.downloadToStream(ws, p);
-    this.logger.debug('downloadToStream successful');
     const masterBuild = await this.getMasterBuild();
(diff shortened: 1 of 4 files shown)


liamdebeasi (changes requested):
This change looks good, but we need to add a test. You can add a test to the `item/test/inputs` directory.

In here we need an outlined item with a floating label and an autogrowing textarea with enough text to make the original issue reproduce. We can then add a test to `e2e.ts` that takes a screenshot of this behavior. Might make sense to land your autogrow fix first otherwise the textarea won't be sized properly on load.

liamdebeasi (approved):
Tested in Chrome, Safari, and Firefox -- all look good 👍

PR #24203, opened 2021-11-11
fix(modal): card modal shadow now shows up correctly on ipad

liamdebeasi:
## Pull request checklist

## Pull request type

 

 

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

With the new inline modals, we need to make sure that subsequent modals do not have a boxshadow/backdrop shown otherwise the screen will get progressively darker. This rule should only apply to non-card and non-sheet modals, but the selector was a bit too broad. The end result was boxshadows were hidden on all card modals on iPad.

Also there was a box shadow being applied to both the wrapper and shadow element which made the shadow appear to stack.

We did not have visual regression tests for these kinds of stacked modals, so this issue slipped through.

## What is the new behavior?

- Shadows are now shown correctly on the card modal on iPad.
- Removed double box shadow.
- Added visual regression tests.

- [ ] Yes
- [x] No

### DIFF
--- a/core/src/components/modal/modal.ios.scss
+++ b/core/src/components/modal/modal.ios.scss
@@ -72,6 +72,10 @@
     transition: all 0.5s ease-in-out;
   }
 
+  :host(.modal-card) .modal-wrapper {
+    box-shadow: none;
+  }
+
   :host(.modal-card) .modal-shadow {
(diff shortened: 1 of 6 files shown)


sean-perkins (approved):
Lgtm 👍

Issue #21242, opened 2020-05-08
bug: Autogrow property in IonTextArea doesnt look good


andirsun:
# Bug Report

**Ionic version:**

[x] **5.1**

**Current behavior:**
When I initialize the value of a text area the label doesnt look good 

**Expected behavior:**
Need to auto expand if I set property autogrod = true

**Steps to reproduce:**

Set a value to textArea component

**Related code:**

```
<IonItem>
            <IonLabel position="stacked">Beneficios</IonLabel>
            <IonTextarea 
              className="inputs"
              placeholder={benefits}
              autoGrow={true}
              maxlength={400}     
              value={benefits}
              rows={6} cols={20}
              onIonChange={e => setBenefits(e.detail.value!)}>
            </IonTextarea>
 </IonItem>
```

**Other information:**

![localhost_8100_home(iPhone X)](https://user-images.githubusercontent.com/22928302/81361426-facc8d80-90a3-11ea-9724-8a9c2ea1d995.png)

**Ionic info:**

```
Ionic:

   Ionic CLI       : 6.8.0 (/Users/timugo/.nvm/versions/node/v12.16.0/lib/node_modules/@ionic/cli)
   Ionic Framework : @ionic/react 5.1.0

Capacitor:

   Capacitor CLI   : 2.1.0
   @capacitor/core : 2.1.0

Utility:

   cordova-res : not installed
   native-run  : not installed

System:

   NodeJS : v12.16.0 (/Users/timugo/.nvm/versions/node/v12.16.0/bin/node)
   npm    : 6.14.4
   OS     : macOS Mojave
```

andirsun:
@liamdebeasi 
- Clone https://github.com/andirsun/bugTextArea
- npm install
- see the IonTextArea cant see correctly

brandyscarney:
I was able to reproduce it in a Codepen *randomly* - if you keep refreshing eventually you'll see it: https://codepen.io/brandyscarney/pen/JjYabVa

andirsun:
> I was able to reproduce it in a Codepen _randomly_ - if you keep refreshing eventually you'll see it: https://codepen.io/brandyscarney/pen/JjYabVa

Yes, in the web browser the autogrow propertie works fine, but in phone view the input doesnt recognize the size of the preloaded text.

HansKrywaa:
Still appears. Any Updates to this?

andirsun:
> Still appears. Any Updates to this?

No yet :(

[...]

PR #24205, opened 2021-11-11
fix(textarea): textarea with autogrow will size to its contents

sean-perkins:
## Pull request checklist

## Pull request type

 

 

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

Resize behavior currently is using the scroll height of the control, to resize the outer container. This results in flaky behavior where the textarea will not correctly grow to its contents.

Issue Number: #24793, #21242

## What is the new behavior?

- Textarea sizing is controlled through a pseudo element with the same visual appearance as the textarea (only hidden)
- Pseudo element reflows naturally, without manually assigning container height: (inspired from: https://css-tricks.com/the-cleanest-trick-for-autogrowing-textareas/).
- Re-introduces screenshot tests for auto-grow behavior

- [ ] Yes
- [x] No


Lxstr:
Hi @sean-perkins.  I would really like to try this solution but I am having trouble using a forked @ionic/core in my stenciljs project. I did manage a while ago as I was having trouble with textarea [https://github.com/ionic-team/ionic-framework/issues/21242](url) but for the life of me can't figure out how I did it. I have looked at the contributing guide but it's vague. And I've tried npm link but got confused.

So far I have 

- forked ionic-framework
- cloned to laptop
- merged feat/textarea-auto-grow to main
- $ cd /core
- $ npm run build
- removed anything in .gitignore starting with core/
- commited the build
- pushed to remote (https://github.com/Lxstr/ionic-framework.git)

From here how would I use this build in a way that I could easily swap back and forth with the real ionic release? 

Ideally have both installed in my project and just import the one I want to test in the appropriate component.

Thanks in advance

Lxstr:
I've tested the new textarea and the bugs I was previously experiencing seem to be solved!! Awesome!

However, I do notice issues with the padding. It seems like the ::after elements is creating so much padding that it seems like an entire extra line is added at the bottom of the textarea that is un-useable.

I think the expectation would be little to no padding that the user can then add if required on textareas?

Thanks again for the fix and looking forward to the merge!

Screen Shot 2022-01-10 at 5 20 55 pm
Screen Shot 2022-01-10 at 5 34 14 pm

sean-perkins:
This PR is on hold until we finish our Playwright integration (being able to test across different browsers). Will rebase/resolve merge conflicts at that time.

anagstef:
Hello everyone! This PR indeed fixes all the issues we have with textarea and autogrow. Thank you for resolving them!!!

Is this getting prioritised anytime soon? I see that some Playwright PRs are merged. 

Would love to help if I can. Thank you again! :)

sean-perkins:
Since Playwright was merged in with our 6.1.0 work, this PR is ready to be worked on and updated for 6.2.0. I cannot commit to a timeline on 6.2, since we just released, but you can expect this feature to be included in those changes.

core/src/components/textarea/test/basic/index.html
```diff
@@ -68,11 +68,10 @@
             <ion-textarea clear-on-edit="true"></ion-textarea>
           </ion-item>
 
-          <!-- TODO: Re-add auto grow with PR#24205 -->
-          <!-- <ion-item>
+          <ion-item>
             <ion-label color="primary">Autogrow</ion-label>
             <ion-textarea auto-grow="true"></ion-textarea>
-          </ion-item> -->
+          </ion-item>
```
amandaejohnston, line 74: Is this needed? Seems like the `autogrow` test covers it, and that'd be one less thing to worry about in the `basic` test.

sean-perkins: I think we can remove it. It doesn't add any test coverage that isn't already handled by the existing auto grow tests.

amandaejohnston (approved):

core/src/components/textarea/textarea.scss
```diff
@@ -76,17 +76,37 @@
...
 
 .textarea-wrapper {
+  display: grid;
+
   min-width: inherit;
   max-width: inherit;
   min-height: inherit;
   max-height: inherit;
+
+  &::after {
```
liamdebeasi, line 86: Can we add comments here as to why this technique works?

core/src/components/textarea/textarea.scss
```diff
@@ -103,6 +123,8 @@
   resize: none;
   appearance: none;
 
+  overflow: hidden;
```
liamdebeasi, line 126: Does this fix the scrollbar showing when zooming in/out?

liamdebeasi (changes requested):

liamdebeasi:
There is also some strange flickering on iOS 13:
https://user-images.githubusercontent.com/2721089/179781340-6312fdb3-2cea-4095-9f6b-7f1b0097d729.mov

core/src/components/textarea/textarea.scss
sean-perkins: This is to resolve the vertical scrollbar showing on Firefox. It may resolve other issues with the scrollbar as well though.

sean-perkins:
@liamdebeasi good find. Looks like iOS 13 has some rendering difficulties with `grid-area` with multiple elements in a single container. 

To work around this, the reflected pseudo element will be the only element affecting the layout. The native textarea will be absolutely positioned and fill the container's height when the textarea is using auto grow. 

This appears to resolve the iOS 13 rendering bug and maintains auto grow behavior.

liamdebeasi (commented):
It looks like this may actually be a scroll assist issue with duplicating the input while we adjust the scroll position

https://user-images.githubusercontent.com/2721089/181077447-e500effe-ea17-40f9-b9f2-ac5cb4e841fb.mp4

Setting the following "fixes" the issue:

```js
window.Ionic = {
    config: {
      scrollAssist: false
    }
  }
```

sean-perkins:
Ok I have no clue how my previous "fix" was working, because when running again I ran into the same behavior you noted. It was caused by the scroll assist. 

Added a workaround that increases the height of the cloned input (textarea) to the height of the container (`100%`) when used with auto grow. 

Confirmed on iOS 13 simulator that this resolves the white box rendering.

core/src/components/textarea/textarea.scss
```diff
@@ -136,6 +157,12 @@
   pointer-events: none;
 }
 
+:host([auto-grow]) .cloned-input {
+  // Remove when deprecating iOS 13 support
+  // Workaround for iOS 13 to avoid rendering a white box
+  // when scroll assist is enabled and a user taps on a textarea with auto-grow.
```
liamdebeasi, line 163: I was also able to reproduce this issue on iOS 15, so it may not be iOS 15-specific.

core/src/components/textarea/test/basic/index.html
```diff
@@ -67,12 +67,6 @@
             <ion-label color="primary">Clear on Edit</ion-label>
             <ion-textarea clear-on-edit="true"></ion-textarea>
           </ion-item>
-
-          <!-- TODO: Re-add auto grow with PR#24205 -->
-          <!-- <ion-item>
-            <ion-label color="primary">Autogrow</ion-label>
-            <ion-textarea auto-grow="true"></ion-textarea>
-          </ion-item> -->
```
liamdebeasi, line 75: Why was this disabled?

liamdebeasi (approved):
Good to go when the comments below are addressed. Great job!

core/src/components/textarea/test/basic/index.html
sean-perkins: When migrating the Puppeteer tests to Playwright, I ran into issues with consistently capturing the screenshots because of auto-grow. We discussed disabling those auto grow screenshots until this PR was to be merged. 

When Amanda reviewed this PR, she called out that this test is effectively a duplicate of behavior captured in the other auto grow tests, so we elected for removing it entirely from the basic tests.

Issue #24241, opened 2021-11-17
bug: Datetime UI does not update the value after it has changed programmatically.


NormanV41:
### Prerequisites

- [X] I have read the [Contributing Guidelines](https://github.com/ionic-team/ionic-framework/blob/main/.github/CONTRIBUTING.md#creating-an-issue).
- [X] I agree to follow the [Code of Conduct](https://ionicframework.com/code-of-conduct).
- [X] I have searched for [existing issues](https://github.com/ionic-team/ionic-framework/issues) that already report this problem, without success.

### Ionic Framework Version

- [ ] v4.x
- [ ] v5.x
- [X] v6.x

### Current Behavior

Datetime UI does not show that the value has changed after it changed programmatically. This issue is unique for version 6.

### Expected Behavior

The Datetime UI should be updated when its value changes programmatically.

### Steps to Reproduce

- Start a new app with a blank template.
- add a Datetime component with its value bound to a component property.
- asynchronously update the value of the component property.
- observe that the UI does not show that the value has changed.

### Code Reproduction URL

https://github.com/NormanV41/ionic-issues-demo/tree/datetime-issue-v6

### Ionic Info

[WARN] Error loading @capacitor/ios package.json: Error: Cannot find module '@capacitor/ios/package'
       
       Require stack:
       - /usr/lib/node_modules/@ionic/cli/lib/project/index.js
       - /usr/lib/node_modules/@ionic/cli/lib/index.js
       - /usr/lib/node_modules/@ionic/cli/index.js
       - /usr/lib/node_modules/@ionic/cli/bin/ionic
[WARN] Error loading @capacitor/android package.json: Error: Cannot find module '@capacitor/android/package'
       
       Require stack:
       - /usr/lib/node_modules/@ionic/cli/lib/project/index.js
       - /usr/lib/node_modules/@ionic/cli/lib/index.js
       - /usr/lib/node_modules/@ionic/cli/index.js
       - /usr/lib/node_modules/@ionic/cli/bin/ionic

Ionic:

   Ionic CLI                     : 6.18.0 (/usr/lib/node_modules/@ionic/cli)
   Ionic Framework               : @ionic/angular 6.0.0-rc.3
   @angular-devkit/build-angular : 12.1.4
   @angular-devkit/schematics    : 12.1.4
   @angular/cli                  : 12.1.4
   @ionic/angular-toolkit        : 4.0.0

Capacitor:

   Capacitor CLI      : 3.2.5
[...]

Issue #20016, opened 2019-11-28
bug: prevent ion-datetime selected value to be updated


loopezz:
# Bug Report

**Ionic version:**

[x] **4.x**

**Current behavior:**
Updating the value while the picker is presented does also update the selected value within the picker.

**Expected behavior:**
The selected value should not be updated due to bad UX. It resets the user input.
- Do update current datetime which is visible in the UI list;
- Don't update current selected value while the picker is presented

**Steps to reproduce:**
Just select a custom time.

**Related code:**
https://stackblitz.com/edit/ionic-v4-angular-tabs-qnrfsw

```
insert short code snippets here
```

**Other information:**
Expected behavior is the same behavior as selecting datetime on iOS.

**Ionic info:**

```
insert the output from ionic info here
```

PR #24244, opened 2021-11-18
fix(datetime): update active calendar display when value changes

sean-perkins:
## Pull request checklist

## Pull request type

 

 

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

When programmatically changing the value of `ion-datetime`, the visual display will not correctly reflect the updated value.

Issue Number: #24241

## What is the new behavior?

Whenever the value is modified, the `ion-datetime` will correctly display the updated value.

- [ ] Yes
- [x] No


liamdebeasi (commented):
I think this behavior should be discussed further. The concern I have with this change is if a user is interacting with the datetime on May 2020 and the value changes to June 2021, the entire view jumps to June 2021.

We had this behavior in the old Ionic 5 picker and some felt that it was a bad UX: https://github.com/ionic-team/ionic-framework/issues/20016

Perhaps a middle ground could be to update the selected value without jumping the entire view to the new selected date? In this case, we could just update `this.activeParts` and leave `this.workingParts` alone. One thing I do notice is the `ion-picker-column` elements currently look at `workingParts` for the selected value. This might need to be changed to account for this.

sean-perkins:
Notes from aside conversation:

1. We will update this PR (the `ion-datepicker`) to update the view for the selected value to reflect the new current date. This interaction will not scroll the view and the user will remain in place. 
2. In post 6.0, we will look into options that allow users to customize the scrolling behavior.

High level notes regarding different platforms/libraries around this topic:
- iOS (native): will force scroll the user to the updated date when the value is changed
- Android (flutter/dart/etc.): does not allow manipulation of the date picker reference, once it's opened all properties are readonly
- Web (datetime-local) on device: selected date does not update in the calendar view when programmatically changed. User does not scroll to updated date.
- Vuetify, Quasar, MUI: transitions to the current date, but if you are actively interacting with the time picker it will not. If you are interacting with the primary date picker, it will jump to the current date.

sean-perkins:
@liamdebeasi small tweak here after our conversation.

I found that when the presentation mode is `time` or `date-time`, it's justifiable to force the user's time picker to the current value. The calendar view shouldn't scroll based on previous conversation.

The time picker however, has no visual distinction between the current value and the active value being selected in the design spec. The path of least resistance is to update the time picker's representation immediately so that the user can see the current value and then modify from there.

Open to your thoughts/feedback 👍

liamdebeasi (commented):
Code looks good. Can we add an E2E test to verify this fix?

core/src/components/datetime/test/set-value/e2e.ts
```diff
@@ -0,0 +1,37 @@
...
+
+    await page.$eval('ion-datetime', (elm: any) => {
+      elm.value = '2021-11-25T12:40:00.000Z';
+    });
+
+    await page.waitForChanges();
+
+    const activeTime = await page.find('ion-datetime >>> .time-body');
+
+    expect(activeTime).toEqualText('12:40 PM');
```
liamdebeasi, line 34: Can we force the locale to be "en-US" on `ion-datetime`? If this runs on CI but the locale on the machine is different this test may fail.

core/src/components/datetime/datetime.tsx
```diff
@@ -291,6 +297,22 @@ export class Datetime implements ComponentInterface {
    */
   @Watch('value')
   protected valueChanged() {
+    if (this.hasValue()) {
+      /**
+       * Update the value of the cloned reference to `activeParts` to synchronize
+       * the date display on the current render cycle.
+       */
+      const { month, day, year, hour, minute } = parseDate(this.value);
```
liamdebeasi, line 305: Can we clarify this to explain how we want the selected values to change, but we do not want the user's calendar view to be forced to the selected month if they are looking at a different month?

core/src/components/datetime/datetime.tsx
```diff
@@ -93,6 +93,12 @@ export class Datetime implements ComponentInterface {
   private minParts?: any;
   private maxParts?: any;
 
+  /**
+   * Duplicate reference to `activeParts` that does not rigger a re-render of the component.
```
liamdebeasi, line 97:
```suggestion
   * Duplicate reference to `activeParts` that does not trigger a re-render of the component.
```

Let's also add an explanation of when/why this is used

liamdebeasi (changes requested):

core/src/components/datetime/datetime.tsx
```diff
@@ -1213,12 +1243,13 @@ export class Datetime implements ComponentInterface {
...
    ) {
-    const { color, workingParts } = this;
+    const { color, activePartsClone: timeParts } = this;
+
     return (
       <ion-picker-internal>
         <ion-picker-column-internal
           color={color}
-          value={workingParts.hour}
+          value={timeParts.hour}
```
liamdebeasi, line 1252: Small nitpick, but this object isn't just time parts anymore, so maybe we should call it `activePartsClone`?

liamdebeasi (approved):
Other than the variable name, this looks great. Nice job!

core/src/components/datetime/datetime.tsx
sean-perkins: This should be good now 👍  I just went ahead and removed the rename and used destructing assignment for all variables in the render function.

Issue #24070, opened 2021-10-13
bug: time picker in ion-datetime does not let me select AM/PM using keyboard input


bchehraz:
### Prerequisites

- [X] I have read the [Contributing Guidelines](https://github.com/ionic-team/ionic-framework/blob/main/.github/CONTRIBUTING.md#creating-an-issue).
- [X] I agree to follow the [Code of Conduct](https://ionicframework.com/code-of-conduct).
- [X] I have searched for [existing issues](https://github.com/ionic-team/ionic-framework/issues) that already report this problem, without success.

### Ionic Framework Version

- [X] v6.x

### Current Behavior

In terms of accessibility on Web, I'm able to tab through to select a time using the time picker in `ion-datetime` by typing in values e.g. `12`, `tab`, `35`, but using the keyboard I am not able to switch between AM/PM (typing in A or P doesn't work, and neither does up/down arrow keys)

It also seems strange that I can use the scroll-wheel to select both times and AM/PM, but the up/down keys don't work for any of the input fields. This to me is a bug but if I need to put in a feature request instead, let me know and I will gladly do that!

In this screen recording, I'm using tab + keyboard to input into the component. When I reach the AM/PM section, I'm unable to switch from AM to PM using `A`, `P`, or up/down arrow keys so I'm stuck.
https://user-images.githubusercontent.com/7946204/137190396-f6c9036a-e3b9-48f1-ab54-95a60139124e.mov

### Expected Behavior

Using a combination of number input and `tab` key, I should be able to type in any input into the numbers for the time picker, not just numbers, but also `A` or `P` to specify AM or PM when I get to the AM/PM selection.

Using some combination of keyboard input, I should be able to safely input a time as well as the AM/PM value into the time picker of the `ion-datetime` component. I should be able to type the letter `A` or `P` to go to AM or PM.

In addition, if I'm able to use the scroll wheel on a mouse to go scroll through the time or AM/PM selection, I should also be able to use up-down keyboard arrow keys for additional accessibility.

### Steps to Reproduce

1. Simply generated a new ionic project, deleted some unnecessary code in the starter project, ran `ionic serve`
2. Added the following code into `home.html`. I happened to be testing out the ion-datetime inside a modal, but this issue will happen regardless of being in a modal.
```
  <ion-button id="open-modal">Open Datetime Modal</ion-button>
  <ion-modal trigger="open-modal" style="--height: 386px; --width: 351px;">
    <ng-template>
      <ion-content>
[...]

PR #24251, opened 2021-11-22
fix(datetime): keyboard navigation of time picker columns

sean-perkins:
## Pull request checklist

## Pull request type

 

 

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

Users cannot select individual increments of time or the am/pm value, when using only their keyboard in an `ion-datetime`. Users can only use the `space` key, to jump increments of 5. The `ion-popover` has opinionated keyboard event handlers that cancels the default browser behavior.

Issue Number: #24070

## What is the new behavior?

Users can utilize the up/down keys to navigate through the picker columns for hour, minutes and am/pm within an `ion-datetime`. Users can use the `home` and `end` keys to jump to the start or end of the list.

Introduces `cancelBrowserKeyboardEvents` option on `ion-popover` that allows implementations to disable the default behavior of canceling the browser events. This allows the picker column to support the browser behavior of scroll snapping with keyboard navigation.

<video src="https://user-images.githubusercontent.com/13732623/142899828-852106e0-662d-4787-8518-431ce4d09cf5.mp4"></video>

- [ ] Yes
- [x] No


liamdebeasi (commented):
Were you able to figure out why the keyboard was not working in a popover? Scroll containers with scroll snapping have native support for selecting items with the arrow keys, so this functionality should be built into the browsers/we should not need to rebuild it ourselves.

sean-perkins:
> Were you able to figure out why the keyboard was not working in a popover? Scroll containers with scroll snapping have native support for selecting items with the arrow keys, so this functionality should be built into the browsers/we should not need to rebuild it ourselves.

@liamdebeasi the popover implementation for keyboard interactions is preventing the arrow navigation.

https://github.com/ionic-team/ionic-framework/blob/next/core/src/components/popover/popover.tsx#L388

Specifically the `preventDefault()` - https://github.com/ionic-team/ionic-framework/blob/55db38ddc541c2632c7d3e4e4c9400ff5b5dfe8c/core/src/components/popover/utils.ts#L375

Do you recall the intention behind that? Looks like up/down/home are prevented, but left and right are not.

https://github.com/ionic-team/ionic-framework/blob/55db38ddc541c2632c7d3e4e4c9400ff5b5dfe8c/core/src/components/popover/utils.ts#L420

core/src/components/popover/popover.tsx
```diff
@@ -208,6 +208,15 @@ export class Popover implements ComponentInterface, PopoverInterface {
...
   @Prop() isOpen = false;
 
+  /**
+   * If `true`, the popover will cancel browser keyboard event bindings to prevent
+   * scroll behavior in a popover using a list of items.
+   *
+   * Disable this behavior to allow the contents of the popover to handle their own
+   * keyboard interactions (i.e. when using CSS scroll snap).
+   */
+  @Prop() cancelBrowserKeyboardEvents = true;
```
liamdebeasi, line 218: What do you think about making this a prop that disables the keyboard events when `true`? I think if developers want the browser keyboard events to be handled normally, it's likely they don't want Ionic getting in the way either. I can't think of a use case where we would want both the browser and Ionic to handle the keyboard events.

sean-perkins: Makes sense to me! Verified locally that all the keyboard navigation built-ins work the same after disabling the popover's keyboard event handlers.

I went with `disableKeyboardEvents` for the property, open to suggestions. We don't appear to have a definite pattern for enable vs. disable language for properties.

liamdebeasi: I try to name public APIs using positive language, that way it's easier to reason about when negating values. For example `!popoverEl.enableKeyboardEvent` is easier to reason about than `!popoverEl.disableKeyboardEvent`

core/src/components/popover/popover.tsx
```diff
@@ -208,6 +208,17 @@ export class Popover implements ComponentInterface, PopoverInterface {
    */
   @Prop() isOpen = false;
 
+  /**
+   * If `true` the popover will not register it's own keyboard event handlers.
```
liamdebeasi, line 212:
```suggestion
   * If `true` the popover will not register its own keyboard event handlers.
```
```suggestion
   * If `true` the popover will not register it's own keyboard event handlers.
```

core/src/components/popover/popover.tsx
```diff
@@ -208,6 +208,17 @@ export class Popover implements ComponentInterface, PopoverInterface {
...
+   * This allows the contents of the popover to handle their own keyboard interactions.
+   *
+   * If `false`, the popover will register it's own keyboard event handlers for
```
liamdebeasi, line 215:
```suggestion
   * If `false`, the popover will register its own keyboard event handlers for
```
```suggestion
   * If `false`, the popover will register it's own keyboard event handlers for
```

liamdebeasi (approved):
The code looks good minus a few comment typos. Thoughts on making this a private property for now? I know we discussed about some hypothetical use cases developers may run into, but I think I would prefer to wait for devs to come to us with their use cases before we commit to having this be part of the public API.

appy-one:
I've just created a new issue about keyboard navigation in a textarea in popovers (#24633), I though I'd share this here because it is affected by the same `preventDefault`s in overlays.ts

Issue #24139, opened 2021-11-01
bug: SetupConfig does not work with Ionic React (v6)


BerkeAras:
### Prerequisites

- [X] I have read the [Contributing Guidelines](https://github.com/ionic-team/ionic-framework/blob/main/.github/CONTRIBUTING.md#creating-an-issue).
- [X] I agree to follow the [Code of Conduct](https://ionicframework.com/code-of-conduct).
- [X] I have searched for [existing issues](https://github.com/ionic-team/ionic-framework/issues) that already report this problem, without success.

### Ionic Framework Version

- [ ] v4.x
- [ ] v5.x
- [X] v6.x

### Current Behavior

Ionic setupConfig does not work with React. Just no behavior, no response, no error.

### Expected Behavior

SetupConfig should override the existing default configuration.

### Steps to Reproduce

Create new ionic react app, update to Ionic v6 [(how can I update to the beta?)](https://github.com/ionic-team/ionic-framework/blob/next/BETA.md), try setupConfig.

### Additional Information

Code Example:

```js
import { useEffect } from 'react';
import { Capacitor } from '@capacitor/core';
import { Keyboard } from '@capacitor/keyboard';
import { IonApp, IonRouterOutlet, IonSplitPane, setupConfig } from '@ionic/react';
import { IonReactRouter } from '@ionic/react-router';
import { Redirect, Route } from 'react-router-dom';

/* Core CSS required for Ionic components to work properly */
import '@ionic/react/css/core.css';

/* Basic CSS for apps built with Ionic */
import '@ionic/react/css/normalize.css';
import '@ionic/react/css/structure.css';
import '@ionic/react/css/typography.css';

/* Optional CSS utils that can be commented out */
import '@ionic/react/css/padding.css';
import '@ionic/react/css/float-elements.css';
import '@ionic/react/css/text-alignment.css';
import '@ionic/react/css/text-transformation.css';
import '@ionic/react/css/flex-utils.css';
import '@ionic/react/css/display.css';


/* Theme variables */
import './theme/variables.scss';

/* Pages */
import PageHome from './pages/default/Home/Home';

setupConfig({ mode: 'md' });

const App = () => {
.........
```

amandaesmith3:
Thanks for the report; confirming I can replicate 👍 I'm indeed only seeing this with React (not Angular or Vue) and confirmed that `setupConfig` itself isn't working, rather than it being any particular property.

[...]

PR #24254, opened 2021-11-22
feat(react): add setupIonicReact function

liamdebeasi:
## Pull request checklist

## Pull request type

 

 

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes

Issue Number: resolves https://github.com/ionic-team/ionic-framework/issues/24139

With the Custom Elements build, we now need to call `initialize()`: https://github.com/ionic-team/ionic-framework/blob/next/packages/react/src/components/index.ts#L202. The problem is that this call also sets up the Ionic config. Given that Ionic config is not designed to be reactive, subsequent calls to `setupConfig` had no effect.

## What is the new behavior?

- After discussing with the team, we feel the best path forward is a (small) breaking change. Developers will be required to add `setupIonicReact` to their `App.tsx`/`App.jsx` file. This function will call `initialize` and allow developers to pass in a config, so it can be used instead of `setupConfig`.
- We explored non-breaking change options, but we think this change is the best option long term.
- I also noticed that setupConfig was being exported from Angular/React/Vue packages when it should not be. Developers should be going through the main entry points and not calling `setupConfig` on their own. If they need to eject from the normal process, they should call `initialize` instead.

- [x] Yes
- [ ] No

## Other information

Migration:

1. Import `setupIonicReact` from '@ionic/react';
2. In `App.tsx`, call `setupIonicReact();`. Config should be passed here instead of `setupConfig`.
3. Remove any calls to `setupConfig`.

### DIFF
--- a/BREAKING.md
+++ b/BREAKING.md
@@ -29,14 +29,15 @@ This is a comprehensive list of the breaking changes introduced in the major ver
 - [Config](#config)
   * [Transition Shadow](#transition-shadow)
 - [Angular](#angular)
-  * [Config Provider](#config-provider)
+  * [Config](#config-1)
 - [Vue](#vue)
+  * [Config](#config-2)
   * [Tabs Config](#tabs-config)
   * [Tabs Router Outlet](#tabs-router-outlet)
   * [Overlay Events](#overlay-events)
   * [Utility Function Types](#utility-function-types)
 - [React](#react)
-  * [App Initialization](#app-initialization)
+  * [Config](#config-3)
 - [Browser and Platform Support](#browser-and-platform-support)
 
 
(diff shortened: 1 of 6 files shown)


amandaesmith3:
We'll also need to update the docs when this is merged: https://beta.ionicframework.com/docs/react/config

BREAKING.md
```diff
@@ -168,6 +170,31 @@ The `experimentalTransitionShadow` config option has been removed. The transitio
 
 The `Config.set()` method has been removed. See https://ionicframework.com/docs/angular/config for examples on how to set config globally, per-component, and per-platform.
 
+### React
```
amandaesmith3, line 173: Should we also note that `setupConfig` is no longer exported for Angular and Vue?

amandaesmith3 (changes requested):
LGTM other than the BREAKING.md comment 👍 There's also updating the starter apps and docs, but those are other repos so we're good here otherwise.

liamdebeasi:
Starters have been updated in the `ionic6` branch: https://github.com/ionic-team/starters/compare/ionic6?expand=1
Docs have been updated in to the `react-config` branch: https://github.com/ionic-team/ionic-docs/compare/react-config?expand=1

amandaesmith3 (approved):

developer82:
Hi,
I'm using the following:
```
"@ionic/react": "^6.0.1",
"@ionic/react-router": "^6.3.2",
```

And setting the following:
```
setupIonicReact({
    swipeBackEnabled: false
});
```

And it still doesn't work. Swipe back still working. The reason I wish to disable it is because it's so buggy. I have a tab navigation and if clicking the same tab it swipes back to it over and over and when swiping back sometimes the animation "jumps" and I get white screens.

Issue #24168, opened 2021-11-05
bug: ion-content scroll element undefined in Vue onMounted hook


a-Leong:
### Prerequisites

- [X] I have read the [Contributing Guidelines](https://github.com/ionic-team/ionic-framework/blob/main/.github/CONTRIBUTING.md#creating-an-issue).
- [X] I agree to follow the [Code of Conduct](https://ionicframework.com/code-of-conduct).
- [X] I have searched for [existing issues](https://github.com/ionic-team/ionic-framework/issues) that already report this problem, without success.

### Ionic Framework Version

- [ ] v4.x
- [ ] v5.x
- [X] v6.x

### Current Behavior

When called in mounted lifecycle hook, IonContent scroll methods raise errors related to scroll element evaluating to `undefined`. `getScrollElement` resolves to `undefined` as well. E.g.:
```console
Uncaught (in promise) TypeError: Cannot read properties of undefined (reading 'scrollHeight')
    at HTMLElement.scrollToBottom (ion-content.js?0289:149)
    at scrollToBottom (Home.vue?4912:21)
    at eval (Home.vue?4912:26)
    at callWithErrorHandling (runtime-core.esm-bundler.js?5c40:6701)
    at callWithAsyncErrorHandling (runtime-core.esm-bundler.js?5c40:6710)
    at Array.hook.__weh.hook.__weh (runtime-core.esm-bundler.js?5c40:1964)
    at flushPostFlushCbs (runtime-core.esm-bundler.js?5c40:6902)
    at render (runtime-core.esm-bundler.js?5c40:4840)
    at mount (runtime-core.esm-bundler.js?5c40:3173)
    at Object.app.mount (runtime-dom.esm-bundler.js?830f:1574)
```

The IonContent element itself is not undefined the time of this invocation, only its scroll element appears to be.

### Expected Behavior

Scroll methods run successfully and `getScrollElement` resolves to the scroll element for an IonContent component in onMounted. This is the working behavior in v5.x. 

### Steps to Reproduce

Make a call to any scroll method on an IonContent component in the onMounted lifecycle hook of a Vue component. An error is raised when the component mounts. Invocations of scroll methods from user-triggered events work as expected. Adding a sufficiently large timeout before invoking scroll methods within onMounted causes methods to work as expected as well.

Edit:
[...]

PR #24255, opened 2021-11-22
fix(content): ensure scrollEl is always available in scroll methods

amandaesmith3:
## Pull request checklist

## Pull request type

 

 

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

Issue Number: Resolves https://github.com/ionic-team/ionic-framework/issues/24168

## What is the new behavior?

- `ion-content` methods for manual scrolling will wait for component ready if `scrollEl` isn't ready yet. This ensures that if those methods are called in certain lifecycle hooks (such as Vue's `onMounted`), `scrollEl` is defined and the method works properly.

- [ ] Yes
- [x] No

### DIFF
--- a/core/src/components/content/content.tsx
+++ b/core/src/components/content/content.tsx
@@ -175,17 +175,6 @@ export class Content implements ComponentInterface {
     }
   }
 
-  // scroll methods can be called in lifecycle hooks where scrollEl isn't ready
-  // (ex: Vue onMounted)
-  // ensure scrollEl is present
-  private async awaitScrollEl() {
-    if (!this.scrollEl) {
-      await new Promise(resolve => componentOnReady(this.el, resolve));
-    }
-
-    return Promise.resolve(this.scrollEl);
-  }
-
   /**
    * Get the element where the actual scrolling takes place.
    * This element can be used to subscribe to `scroll` events or manually modify
@@ -195,7 +184,13 @@ export class Content implements ComponentInterface {
    * and `scrollToPoint()` to scroll the content into a certain point.
    */
   @Method()
-  getScrollElement(): Promise<HTMLElement> {
+  async getScrollElement(): Promise<HTMLElement> {
+    // if this gets called in certain early lifecycle hooks (ex: Vue onMounted),
+    // scrollEl won't be defined yet, so wait for it to load in
+    if (!this.scrollEl) {
+      await new Promise(resolve => componentOnReady(this.el, resolve));
+    }
+
     return Promise.resolve(this.scrollEl!);
   }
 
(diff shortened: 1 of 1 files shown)


liamdebeasi (commented):
Are we able to use the `componentOnReady` helper instead? https://github.com/ionic-team/ionic-framework/blob/main/core/src/utils/helpers.ts#L20

We do something similar here: https://github.com/ionic-team/ionic-framework/blob/next/core/src/components/reorder-group/reorder-group.tsx#L59

amandaesmith3:
@liamdebeasi My understanding is that `componentOnReady` can't be used since the element isn't available at all when the hook runs -- the `el` being passed here would just be `undefined`.

liamdebeasi:
The Stencil component is available, but the inner contents of the component have not been rendered yet. If you look at the stack trace in https://github.com/ionic-team/ionic-framework/issues/24168#issue-1045393891 the code is calling `scrollToBottom` on the `ion-content` component. In this case the method is available, but the `scrollEl` element (inside the shadow root) is not defined yet.

Ping me on Slack if you want to step through the code together

amandaesmith3:
Ahh, I was thinking waiting for `scrollEl` to be ready; sounds like you mean the content el itself (`this.el`). Got it 👍

core/src/components/content/content.tsx
```diff
@@ -168,12 +169,23 @@ export class Content implements ComponentInterface {
...
+        updateScrollDetail(this.detail, this.scrollEl!, ts, shouldStart);
         this.ionScroll.emit(this.detail);
       });
     }
   }
 
+  // scroll methods can be called in lifecycle hooks where scrollEl isn't ready
+  // (ex: Vue onMounted)
+  // ensure scrollEl is present
+  private async awaitScrollEl() {
```
liamdebeasi, line 181: It might make sense to do this in the `getScrollElement` method instead so that those using the `getScrollElement` method get this fix too.

From there, we can do the following in methods like `scrollToBottom`:

```js
const scrollEl = await this.getScrollElement();
```

core/src/components/content/content.tsx
```diff
@@ -183,8 +184,14 @@ export class Content implements ComponentInterface {
    * and `scrollToPoint()` to scroll the content into a certain point.
    */
   @Method()
-  getScrollElement(): Promise<HTMLElement> {
-    return Promise.resolve(this.scrollEl);
+  async getScrollElement(): Promise<HTMLElement> {
+    // if this gets called in certain early lifecycle hooks (ex: Vue onMounted),
+    // scrollEl won't be defined yet, so wait for it to load in
```
liamdebeasi, line 189: Super minor style thing, but can we do multi line comments?
```suggestion
    /**
     * If this gets called in certain early lifecycle hooks (ex: Vue onMounted),
     * scrollEl won't be defined yet with the custom elements build, so wait for it to load in.
     */
```

liamdebeasi (approved):
Great job! We will need to verify this behavior in the app provided in the linked issue, but otherwise this is good to go.

PR #24302, opened 2021-12-01
fix(accordion): improve functionality with nested accordions

liamdebeasi:
## Pull request checklist

## Pull request type

 

 

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

Issue Number see internal ticket and https://twitter.com/schlimmson/status/1466015536253480960

The global accordion styles are too general and result in nested accordion groups getting impacts by the state of a parent accordion group. For example, if a parent accordion group is expanded, the icons in a nested accordion group will all be in the expanded state, even if the accordions themselves are not expanded

There was also an issue with keyboard navigation where you could jump between accordion groups by pressing the arrow keys.

## What is the new behavior?

- Styles targeting icons only apply to the direct descendant in the "header" slot.
- Styles targeting accordions only target direct descendants of accordion groups
- Arrow keys now stay within a single accordion group. To jump to the next/prev group users can use Tab or Shift + Tab
- I also moved some spec tests to E2E due to limitations in Sizzle, a library stencil uses under the hood. Sizzle does not support `:scope` so it was causing errors when the new query selector got run.

- [ ] Yes
- [x] No

### DIFF
--- a/core/src/components/accordion-group/accordion-group.tsx
+++ b/core/src/components/accordion-group/accordion-group.tsx
@@ -199,7 +199,7 @@ export class AccordionGroup implements ComponentInterface {
    */
   @Method()
   async getAccordions() {
-    return Array.from(this.el.querySelectorAll('ion-accordion'));
+    return Array.from(this.el.querySelectorAll(':scope > ion-accordion')) as HTMLIonAccordionElement[];
   }
 
   render() {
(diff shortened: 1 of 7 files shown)


amandaesmith3 (commented):
Setting `disabled` on the parent accordion group causes the whole thing (including the inner group) to *look* disabled, but the inner group can still be interacted with. Is that intentional? It feels like the inner group should be disabled too, but I'm open to other thoughts 🤔

amandaesmith3 (approved):

Issue #24245, opened 2021-11-18
bug: Angular - Argument of type 'Event' is not assignable to parameter of type 'CustomEvent<XXX>'.  


JumBay:
### Prerequisites

- [X] I have read the [Contributing Guidelines](https://github.com/ionic-team/ionic-framework/blob/main/.github/CONTRIBUTING.md#creating-an-issue).
- [X] I agree to follow the [Code of Conduct](https://ionicframework.com/code-of-conduct).
- [X] I have searched for [existing issues](https://github.com/ionic-team/ionic-framework/issues) that already report this problem, without success.

### Ionic Framework Version

- [ ] v4.x
- [X] v5.x
- [X] v6.x

### Current Behavior

Wrong definition type breaks the app.

### Expected Behavior

Having the right definition type

### Steps to Reproduce

- Create a new ionic project: `ionic start test-ionic` 
  - Framework: Angular
  - Starter template: tabs
 
Copy/paste the code sample (html and ts) from https://ionicframework.com/docs/api/reorder-group in `tab1.page.(html|ts)`. Then in your IDE you'll notice the given error: 

>  Argument of type 'Event' is not assignable to parameter of type 'CustomEvent<ItemReorderEventDetail>'. Type 'Event' is missing the following properties from type 'CustomEvent<ItemReorderEventDetail>': detail, initCustomEventngtsc(2345)

Try to build the app: `ionic build --prod`, it won't build (same error).

It does this with other events as well...

Same issue occurs on ionic 6 rc3

### Code Reproduction URL

https://github.com/JumBay/ionic-event-type-issue

### Ionic Info

$ ionic info
[WARN] Error loading @capacitor/ios package.json: Error: Cannot find module '@capacitor/ios/package'

       Require stack:
       - /usr/local/lib/node_modules/@ionic/cli/lib/project/index.js
       - /usr/local/lib/node_modules/@ionic/cli/lib/index.js
       - /usr/local/lib/node_modules/@ionic/cli/index.js
       - /usr/local/lib/node_modules/@ionic/cli/bin/ionic
[WARN] Error loading @capacitor/android package.json: Error: Cannot find module '@capacitor/android/package'

       Require stack:
       - /usr/local/lib/node_modules/@ionic/cli/lib/project/index.js
       - /usr/local/lib/node_modules/@ionic/cli/lib/index.js
       - /usr/local/lib/node_modules/@ionic/cli/index.js
[...]

PR #24314, opened 2021-12-03
fix(angular): strict template event bindings

sean-perkins:
## Pull request checklist

## Pull request type

 

 

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

The Angular bindings for the web components are incorrectly type checked by Angular Language Service with `strictTemplates` enabled. The previous usage of `fromEvent`, resulted in the types of `addEventListener` being applied to output events when using template event binding: `(ionSlideDidChange)="doSlideChange($event)"` instead of `CustomEvent`. 

Issue Number: #24245 

## What is the new behavior?

The Angular bindings no longer need to use `fromEvent` to create an observable from the event source. They instead generate with Angular's `@Output()` decorator, which correctly associates the type checking for the output events.

- [x] Yes
- [ ] No

This change will prevent a non-standard pattern for binding to output events in class implementations, but alternative patterns exist to provide the same solution.

Before this would work:
```ts
@ViewChild(IonSlides, { static: true }) slides: IonSlides;

this.slides.ionSlideDidChange.subscribe(() => {
  console.log('slideDidChange');
});
```

After, implementers will need to use `rxjs` to convert the DOM event:
```ts
import { fromEvent } from 'rxjs';

@ViewChild(IonSlides, { static: true, read: ElementRef }) slides: IonSlides;

fromEvent(this.slides, 'ionSlideDidChange').subscribe(() => {
  console.log('slideDidChange');
});
```

Note: In both implementations the subscriptions are always hot and should be tore down later. This is the downside of this non-standard pattern. With the template binding syntax, Angular will automatically tear down the subscription for you.

I spoke with Mike & other GDEs and this pattern is very rare.

## Other information

Dependent on this PR to adjust the generated proxy output: https://github.com/ionic-team/stencil-ds-output-targets/pull/209

Will need a prerelease version to validate outside of locally packing.

Also this PR will need to rebase with the changes after #24313 (likely rebasing the PR in stencil-ds-output-targets).


hakimio:
Is anyone ever going to review this?

sean-perkins:
@hakimio the team has done an internal review of this issue and the PR suggestion that I made.

The problem with my approach is that it causes a breaking change in a pattern that existing Ionic Angular applications could be making use of (there's no way to validate). Merging this change as-is, would only make sense in a major semantic version (i.e Ionic v7).

Instead, we are still diagnosing internally and trying to find a way to satisfy the strict template binding change within Angular while still supporting both versions of event binding:
```html
<ion-input (ionChange)="onChange($event)"></ion-input>
```

```ts
@ViewChild(IonInput) input;

input.ionChange.subscribe(event => {

});
```

hakimio:
The fact is that no Angular developer uses this pattern and there is no point in supporting it...

liamdebeasi:
@sean-perkins We do not have this scheduled in our current sprint. Can we close this for now? 

As Sean noted, we are looking to make these changes in a major release of Ionic to account for any potential breaking changes. We are still discussing the best way to go about addressing this issue.

Issue #24223, opened 2021-11-16
bug: ion-router can set wrong url after nav changes


cjorasch:
### Prerequisites

- [X] I have read the [Contributing Guidelines](https://github.com/ionic-team/ionic-framework/blob/main/.github/CONTRIBUTING.md#creating-an-issue).
- [X] I agree to follow the [Code of Conduct](https://ionicframework.com/code-of-conduct).
- [X] I have searched for [existing issues](https://github.com/ionic-team/ionic-framework/issues) that already report this problem, without success.

### Ionic Framework Version

- [ ] v4.x
- [x] v5.x
- [X] v6.x

### Current Behavior

If the same component is used to render more than one route then the router can set the wrong url after navigation.

For example, when the user clicks the `<ion-back-button>` the url can be left unchanged or can be cleared entirely.

Any code that looks at the url will be using an incorrect value, the browser history will be incorrect, refreshing the page will change the content, etc.

### Expected Behavior

Navigating back should result in the router setting the correct url.

### Steps to Reproduce

If the following routes are defined:
```tsx
<ion-router useHash>
    <ion-route url="/:s1/:s2/:s3" component="my-page" />
    <ion-route url="/:s1/:s2" component="my-page" />
    <ion-route url="/:s1" component="my-page" />
    <ion-route url="" component="my-page" />
</ion-router>
```

And the user navigates through the following steps:
```
1. start at "/"
2. navigate to "/a"
3. navigate to "/b"
4. click on the back button
5. the page content for "/a" is displayed
6. the url shows "/a/b"
```

If the routes are defined in reverse order:
```tsx
<ion-router useHash>
    <ion-route url="" component="my-page" />
    <ion-route url="/:s1" component="my-page" />
    <ion-route url="/:s1/:s2" component="my-page" />
    <ion-route url="/:s1/:s2/:s3" component="my-page" />
</ion-router>
```

And the user navigates through the following steps:
```
1. start at "/"
2. navigate to "/a"
3. navigate to "/b"
4. click on the back button
5. the page content for "/a" is displayed
6. the url shows "/"
```

After clicking on the ionic back button, `<ion-nav>` correctly transitions to the previous page and then calls `router.navChanged()` to update the url path.

This goes through this process:
```ts
// router.tsx - .navChanged()
const chain = routerIDsToChain(ids, routes);

// chain now matches the first route with the same component name
// it does not use any of the path values to determine the best match.
// In the above examples it will always match the first route that is defined

[...]

PR #24315, opened 2021-12-04
fix(router): pop to previous route with params

sean-perkins:
## Pull request checklist

## Pull request type

 

 

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

The function used to determine the correct path to resolve for `ion-router` does not take route parameters into consideration. It only compares the route ids, with the first definition always taking priority over all other routes. This means that if a router has multiple routes sharing the same component, the first definition will always be the route that is determined when running the matching algorithm.

i.e.:
```html
<ion-route url="/" component="my-page" />
<ion-route url="/:s1" component="my-page" />
<ion-route url="/:s1/:s2" component="my-page" />
<ion-route url="/:s1/:s2/:s3" component="my-page" />
```

`/` will always resolve as the path when determining the route to navigate to when going back to `my-page` component.

Issue Number: #24223

## What is the new behavior?

When calculating the best matched route in the chain, routes that both match the active route's id (page component match) and match the definition of route params will be weighted higher.

- [ ] Yes
- [x] No

## Other information

The behavior before this PR (URL resets on pop or back interaction):

<video src="https://user-images.githubusercontent.com/13732623/144538545-67ea11de-c181-45ce-a65b-c5157b06051c.mp4"></video>

The behavior after this PR:

<video src="https://user-images.githubusercontent.com/13732623/144538308-4864e5d8-8bb6-4da8-9330-b27f966bb750.mp4"></video>

### DIFF
--- a/core/src/components/router/utils/matching.ts
+++ b/core/src/components/router/utils/matching.ts
@@ -32,16 +32,60 @@ export const findRouteRedirect = (path: string[], redirects: RouteRedirect[]) =>
   return redirects.find(redirect => matchesRedirect(path, redirect));
 };
 
-export const matchesIDs = (ids: string[], chain: RouteChain): number => {
+export const matchesIDs = (ids: Pick<RouteID, 'id' | 'params'>[], chain: RouteChain): number => {
   const len = Math.min(ids.length, chain.length);
-  let i = 0;
-  for (; i < len; i++) {
-    if (ids[i].toLowerCase() !== chain[i].id) {
+
+  let score = 0;
+
+  for (let i = 0; i < len; i++) {
+    const routeId = ids[i];
+    const routeChain = chain[i];
+    // Skip results where the route id does not match the chain at the same index
+    if (routeId.id.toLowerCase() !== routeChain.id) {
       break;
     }
+    if (routeId.params) {
+      const routeIdParams = Object.keys(routeId.params);
+      /**
+       * Only compare routes with the chain that have the same number of parameters.
+       */
+      if (routeIdParams.length === routeChain.path.length) {
+        /**
+         * Maps the route's params into a path based on the path variable names,
+         * to compare against the route chain format.
+         *
+         * Before:
+         * ```ts
+         * {
+         *  params: {
+         *    s1: 'a',
+         *    s2: 'b'
+         *  }
+         * }
+         * ```
+         *
+         * After:
+         * ```ts
+         * [':s1',':s2']
+         * ```
+         */
+        const pathWithParams = routeIdParams.map(key => `:${key}`);
+        for (let j = 0; j < pathWithParams.length; j++) {
+          // Skip results where the path variable is not a match
+          if (pathWithParams[j].toLowerCase() !== routeChain.path[j]) {
+            break;
+          }
+          // Weight path matches for the same index higher.
+          score++;
+        }
+
+      }
+    }
+    // Weight id matches
+    score++;
   }
-  return i;
-};
+  return score;
+}
 
 export const matchesPath = (inputPath: string[], chain: RouteChain): RouteChain | null => {
   const segments = new RouterSegments(inputPath);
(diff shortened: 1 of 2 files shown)


core/src/components/router/utils/matching.ts
```diff
@@ -32,16 +32,58 @@ export const findRouteRedirect = (path: string[], redirects: RouteRedirect[]) =>
...
+       * After:
+       * ```ts
+       * [':s1',':s2']
+       * ```
+       */
+      const pathWithParams = Object.keys(routeId.params).map(key => `:${key}`);
+      /**
+       * Only compare routes with the chain that have the same number of parameters.
+       */
+      if (pathWithParams.length === routeChain.path.length) {
```
amandaesmith3, line 71: Does it make sense to do this check higher up, perhaps at the same time as the `id` check that `break`s out of the loop?

amandaesmith3 (commented):
I had a couple thoughts, but I trust your judgment on whether changes actually need to be made; I might be misunderstanding the logic.

core/src/components/router/utils/matching.ts
```diff
@@ -32,16 +32,58 @@ export const findRouteRedirect = (path: string[], redirects: RouteRedirect[]) =>
...
-  for (; i < len; i++) {
-    if (ids[i].toLowerCase() !== chain[i].id) {
+
+  let score = 0;
+
+  for (let i = 0; i < len; i++) {
+    const routeId = ids[i];
+    const routeChain = chain[i];
+    // Skip results where the route id does not match the chain at the same index
+    if (routeId.id.toLowerCase() !== routeChain.id) {
```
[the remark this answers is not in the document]
liamdebeasi: Similar to this, do we even need to call `toLowerCase`? I thought ID is a number?

core/src/components/router/utils/matching.ts
liamdebeasi: +1 to this, it would be nice to only do the `Object.keys().map()` evaluation if we need to.

core/src/components/router/utils/matching.ts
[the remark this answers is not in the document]
sean-perkins: The reason that the router chain doesn't need to convert the id to lowercase, is that when the nodes are parsed, we are doing that work there: https://github.com/ionic-team/ionic-framework/blob/main/core/src/components/router/utils/parser.ts#L37

I was mistaken regarding the `id` being a number in our discussion. The `id` is equivalent to the `component` value on the route definition.

core/src/components/router/utils/matching.ts
sean-perkins: I'm unsure if this check could move outside the bounds of checking `params` existing, since it's dependent on the number of keys on the params object. 

We could however shave a little off the processing time by checking the length prior to mapping the params. It'd be an extremely micro change, but happy to make that tweak 👍

liamdebeasi: Yeah if it's not too hard, could we move the Object.keys check inside of the body of the if?

liamdebeasi (approved):
Other than the `Object.keys` change, this PR looks good. Can we also change the commit subject to be something more descriptive? Maybe something like `fix(router): popping route now accounts for route params`.

Might be good to have the issue author verify the fix in their app if they have not already.

Great job!

amandaesmith3 (approved):
Ditto 👍

core/src/components/router/utils/matching.ts
```diff
@@ -32,16 +32,60 @@ export const findRouteRedirect = (path: string[], redirects: RouteRedirect[]) =>
...
+         *
+         * After:
+         * ```ts
+         * [':s1',':s2']
+         * ```
+         */
+        const pathWithParams = routeIdParams.map(key => `:${key}`);
+        for (let j = 0; j < pathWithParams.length; j++) {
+          // Skip results where the path variable is not a match
+          if (pathWithParams[j].toLowerCase() !== routeChain.path[j]) {
```
vicb, line 75: Doesn't this assume that all segments of `routeChain.path[j]` are parameter (i.e. `:param`) ?
Would it work if some of the segments are fixed string ? (i.e. `/fixed/:s1/:s2`)

vicb:
@sean-perkins Is the test app shown in the "Other information" section of the PR description available somewhere ? Shouldn't it be added as an e2e test ?

PR #24337, opened 2021-12-07
fix(vue): tabs no longer get unmounted when navigating back to a tabs context

### LINKED ISSUE #24332: bug: vue, undefined error when leaving tabs and coming back
### Prerequisites

### Ionic Framework Version

- [ ] v4.x
- [X] v5.x
- [ ] v6.x

### Current Behavior

When navigating, occasionally you'll get an error when exiting a page and entering into a new page

### Expected Behavior

No javascript errors

### Steps to Reproduce

It's very random, but the repo I have made shows it. I have also made a video demostrating this:
https://www.youtube.com/watch?v=4SvtlCTgn7k

### Code Reproduction URL

https://github.com/tylerclark/NavigationIssue

### Ionic Info

_No response_

### Additional Information

_No response_

sean-perkins:
## Pull request checklist

## Pull request type

 

 

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

Attempting to modify the `.ion-page-hidden` class for the transitioning element (typically the leaving element), causes an occasional JavaScript exception to throw when the element is already destroyed.

Issue Number: Resolves #24332

## What is the new behavior?

Aligns the Vue implementation for `IonRouterOutlet` against additional safety checks that exist in the React implementation.

- [ ] Yes
- [x] No

### DIFF
--- a/packages/vue/src/components/IonRouterOutlet.ts
+++ b/packages/vue/src/components/IonRouterOutlet.ts
@@ -234,6 +234,11 @@ export const IonRouterOutlet = defineComponent({
 
 See https://ionicframework.com/docs/vue/navigation#ionpage for more information.`);
       }
+      if (enteringViewItem === leavingViewItem) return;
+
+      if (!leavingViewItem && prevRouteLastPathname) {
+        leavingViewItem = viewStacks.findViewItemByPathname(prevRouteLastPathname, id, usingDeprecatedRouteSetup);
+      }
 
       /**
        * If the entering view is already
@@ -247,15 +252,15 @@ See https://ionicframework.com/docs/vue/navigation#ionpage for more information.
        * the previous tabs page is incorrectly
        * unmounted since it would automatically
        * unmount the previous view.
+       *
+       * This should also only apply to entering and
+       * leaving items in the same router outlet (i.e.
+       * Tab1 and Tab2), otherwise this will
+       * return early for swipe to go back when
+       * going from a non-tabs page to a tabs page.
        */
-      if (isViewVisible(enteringEl)) {
+      if (isViewVisible(enteringEl) && !isViewVisible(leavingViewItem.ionPageElement)) {
         return;
-      }
-
-      if (enteringViewItem === leavingViewItem) return;
-
-      if (!leavingViewItem && prevRouteLastPathname) {
-        leavingViewItem = viewStacks.findViewItemByPathname(prevRouteLastPathname, id, usingDeprecatedRouteSetup);
       }
 
       fireLifecycle(enteringViewItem.vueComponent, enteringViewItem.vueComponentRef, LIFECYCLE_WILL_ENTER);
--- a/packages/vue/src/components/IonRouterOutlet.ts
+++ b/packages/vue/src/components/IonRouterOutlet.ts
@@ -14,6 +14,10 @@ import { AnimationBuilder, LIFECYCLE_DID_ENTER, LIFECYCLE_DID_LEAVE, LIFECYCLE_W
 import { matchedRouteKey, routeLocationKey, useRoute } from 'vue-router';
 import { fireLifecycle, generateId, getConfig } from '../utils';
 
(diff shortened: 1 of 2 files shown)


sean-perkins:
@liamdebeasi would be cool if we could sneak this into our patch release for v5. There's no functional behavior side effects, just wrapping of logic in appropriate bound checks. 

I've confirmed with the reproduction app in the issue (you'll have to remove their `@fortawesome` private NPM packages in the package.json if you want to confirm as well).

Downside of this bug existing, is the number of junk exceptions it'll log in implementations making use of an error logging service like Sentry.

liamdebeasi (changes requested):
I don't think this change fixes the underlying issue. There appear to be two issues here:

1. The Home tab is getting unmounted when it should not be.
2. When going back from a non-tabs page to a tab page, the inner router outlet (inside of `ion-tabs`) thinks we are also transitioning from the Home page to the More page. In this case, there should be no transition because it is the outlet outlet that resolves the issue.

I would prefer we resolve these two issues as opposed to patching over them.

I found some concrete steps to reproduce the issue:

1. Load app. You start on `/more`.
2. Click the "Home" tab. You are now on `/home`.
3. Click the "Sessions" tab. You are now on `/sessions`.
4. Click the "More" tab. You are now on `/more`.
5. Click "Terms of Use". You are now on `/agreement-terms`.
6. Click the ion-back-button. You are now on `/more` again, but the Home tab has been unmounted.
7. Click "Privacy Policy". You are now on `/agreements-privacy`.
8. Click the ion-back-button. You are now on `/more` again, but there is the `classList` error because the inner router outlet is trying to transition from the Home tab to the More tab, but the Home tab has been unmounted.

I think if we can figure out why the inner router outlet is transitioning in the first place, then we should be able to fix both of these issues.

liamdebeasi:
The defect seems to be around here: https://github.com/ionic-team/ionic-framework/blob/main/packages/vue/src/components/IonRouterOutlet.ts#L79-L84

Outlet 2 (the tabs outlet) is calling `setupViewItem` even when `previousMatchedRouteRef` is `undefined`. Doing some digging, but adding an undefined check here may fix the issue.

liamdebeasi:
The big issue here is that Home is being unmounted when leaving a non-tab page and coming back to a tab page.

Here is an E2E test I wrote we can use to verify the fixed behavior:

```js
it('should not unmount tab 1 when leaving tabs context', () => {
    cy.visit('http://localhost:8080/tabs');
    cy.ionPageVisible('tab1');

    // Dynamically add tab 4 because tab 3 redirects to tab 1
    cy.get('#add-tab').click();

    cy.get('ion-tab-button#tab-button-tab4').click();
    cy.ionPageHidden('tab1');
    cy.ionPageVisible('tab4');

    cy.get('ion-tab-button#tab-button-tab2').click();
    cy.ionPageHidden('tab4');
    cy.ionPageVisible('tab2');

    cy.get('[data-pageid="tab2"] #routing').click();
    cy.ionPageVisible('routing');
    cy.ionPageHidden('tabs');

    cy.ionBackClick('routing');
    cy.ionPageDoesNotExist('routing');
    cy.ionPageVisible('tabs');
    cy.ionPageVisible('tab2');
    cy.ionPageHidden('tab1');
  });
```

liamdebeasi:
Another thing to try is maybe we can check to see if the `currentMatchedRouteRef` matches whatever the route for the associated active view is: https://github.com/ionic-team/ionic-framework/blob/main/packages/vue/src/components/IonRouterOutlet.ts#L83

Note this is different than comparing to `previousMatchedRouteRef` as this value will be `undefined` when navigating away from tabs (even though the active tab inside of tabs will still be considered active)

sean-perkins:
This looks good on my end 👍  verified with the reproduction app that the error is no longer thrown.

amandaesmith3 (approved):
Looks good to me as well 👍 We may want to update the PR description with the new behavior.

Issue #24230, opened 2021-11-17
bug: opening modal quickly results in white screen


riderx:
### Prerequisites

- [X] I have read the [Contributing Guidelines](https://github.com/ionic-team/ionic-framework/blob/main/.github/CONTRIBUTING.md#creating-an-issue).
- [X] I agree to follow the [Code of Conduct](https://ionicframework.com/code-of-conduct).
- [X] I have searched for [existing issues](https://github.com/ionic-team/ionic-framework/issues) that already report this problem, without success.

### Ionic Framework Version

- [ ] v4.x
- [ ] v5.x
- [X] v6.x

### Current Behavior

if you click multiple time on action sheet button at one point it become full white and you cannot do anything anymore

### Expected Behavior

it shouldn't fail

### Steps to Reproduce

you can go to this codesanbox here :
https://codesandbox.io/s/vue-3-ionic-vuex-forked-rdckf?file=/src/App.vue

and click on the cross on top right corner multiple time.

### Code Reproduction URL

https://codesandbox.io/s/vue-3-ionic-vuex-forked-rdckf?file=/src/App.vue

### Additional Information

Here a video to see the problem

https://user-images.githubusercontent.com/4084527/142094136-10f725d2-a5eb-4d83-abd2-a98cf4a9a4fc.mp4

amandaesmith3:
Thanks for the issue, I've resolved this via https://github.com/ionic-team/ionic-framework/pull/24380 and the fix will be available in a future release of Ionic 👍

liamdebeasi:
Reopening this. The initial fix caused some expected issues (see https://github.com/ionic-team/ionic-framework/issues/24460). We will look into a better long term fix and reply here when we have more information.

For context, we prioritized fixing #24460 as we considered that to be higher priority. That fix will be in Ionic 6.0.3. As a result, I recommend staying on 6.0.2 until this thread's issue gets resolved. Thanks for your patience!

liamdebeasi:
Hi there,

This appears to be resolved on the latest version of Ionic 6, so I am going to close this. If you are still running into an issue, please let me know. Thanks!

PR #24380, opened 2021-12-13
fix(modal): fix timing issue when rapidly closing and opening controller modal

amandaesmith3:
## Pull request checklist

## Pull request type

 

 

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

Issue Number: Resolves https://github.com/ionic-team/ionic-framework/issues/24230

If a controller modal is re-opened immediately after dismissing, the call to `detachComponent` that unmounts the user's modal content doesn't trigger until after the present animation (due to the `await this.currentTransition` line above it). This causes the modal to present, and then immediately unmount its content.

This issue only happens with controller modals; inline modals are unaffected. It also doesn't happen in core, only frameworks like Vue.

## What is the new behavior?

The `detachComponent` call is only made for inline modals. It isn't needed for controller modals anyway, because by that point, the entire modal (including the user's element) has already been removed from the DOM.

- [ ] Yes
- [x] No

## Other information

- Repo for testing (Vue): https://github.com/amandaesmith3/sandbox-ionic-vue/tree/FW-349
- The buttons in the header will open a controller or inline modal.
- If the browser is switched to responsive view, the X that closes the modal will be in the same spot as the rightmost header button. This lets you rapidly click to trigger the bug.
- To try the other modal type, switch the position of the two header buttons in the code, so that the one you want to try is on the right.
- For controller modals, you need at minimum 3 clicks -- present, dismiss, present. The bug happens if the present happens soon enough after the dismiss.
- For inline modals, the bug will not trigger.
- Screencast of bug: https://user-images.githubusercontent.com/90629384/145836560-5daabce8-b76c-4d48-be91-e89db29095e8.mp4
- After fix: https://user-images.githubusercontent.com/90629384/145836646-79671d76-836d-4f4e-bf9e-c576469f80e8.mp4


sean-perkins (commented):
Everything looks great from the reproduction app/solving the issue, nice work!

Anything we can do to solve the issue of these inserted comments not being cleaned up (they get inserted next to `ion-modal` in the DOM)?
```
<!--teleport start-->
<!--teleport end-->
<!--teleport start-->
<!--teleport end-->
```

I believe that's a responsibility of the framework delegate that may not be called anymore with the change for controller-based modals.

core/src/components/modal/modal.tsx
```diff
@@ -494,7 +494,18 @@ export class Modal implements ComponentInterface, OverlayInterface {
...
-      await detachComponent(delegate, this.usersElement);
+
+      /**
+       * If the modal is presented through a controller, we don't need to detach
+       * since the el was already removed during the `dismiss` call above. Skipping
+       * this step also prevents an issue where rapdily dismissing right after
+       * presenting could cause `detachComponent` to be called after the present
+       * finished, blanking out the newly opened modal.
+       */
+      if (this.inline) {
```
liamdebeasi, line 505: See other comment

liamdebeasi (commented):
It looks like there is a race condition in https://github.com/ionic-team/ionic-framework/blob/main/core/src/components/modal/modal.tsx#L495 where we the present method is allowed to proceed even though the dismiss method is not complete because we only await the animation.

I _think_ what I was trying to do at the time was implement a locking system where only one method can be called at a time, and subsequent calls wait for the previous method to finish. Unfortunately, I only leveraged the actual transition, forgetting that there are other async calls that we need to wait for.

This patch does resolve the behavior noted in the issue, but it does not resolve the underlying race condition, so I was hoping we could resolve the race condition before we merge.

-----

In terms of a different approach to fixing this, I am thinking we implement what the `router-outlet` component has currently: https://github.com/ionic-team/ionic-framework/blob/main/core/src/components/router-outlet/route-outlet.tsx#L240-L249

This creates a lock for a single `commit` call and is unlocked only when the `unlock` callback is called: https://github.com/ionic-team/ionic-framework/blob/main/core/src/components/router-outlet/route-outlet.tsx#L134

This would allow us to both resolve the race condition and the issue that reported this. Also, because popover uses the same `currentTransition` logic that the modal does, this race condition is probably there too. Happy to discuss more. Thoughts?

amandaesmith3:
The lock/unlock method didn't work; Liam and I spent a bunch of time debugging and couldn't find the source of the problem. We'll go with the original strategy for now. I've created an internal ticket to try and solve the race condition directly in the long term. Will add more details there.

PR #24401, opened 2021-12-14
fix(angular): defer control value accessor write until after hydration

### LINKED ISSUE #23809: bug: ion-label not floating inside ion-item with select input on ion-modal dialog
### Prequisites

### Ionic Framework Version

- [ ] v4.x
- [X] v5.x
- [ ] v6.x

### Current Behavior

ion-label not floating inside ion-item with select input on ion-modal dialog, label overlay the select field value. 
Selecting a new value fix the overlay bug, however, it comes back reopening the dialog. 
The bug not exists on modeless views.

### Expected Behavior

Floating ion-label not overlay on select fields in both normal and ion.modal view. 

### Steps to Reproduce

Create a modal and open a modal with an ion-label with floating prop with an ion-select.

`<ion-item>
      <ion-label color="primary" position="floating">Posizione
      </ion-label>
      <ion-select multiple="false" formControlName="nascSesso"
        [(ngModel)]="this.entita.Collaboratore.Anagrafica.NascSesso"
        [cancelText]="this.translateService.instant('component.select.cancelText')" ok-text="Ok">
        <ion-select-option *ngFor="let item of this.tipoSesso | keys" [value]="item.key">
          {{ 'Youpass.model.anagrafica.EnumSesso.' + item.value | translate }}</ion-select-option>
      </ion-select>
    </ion-item>`

### Code Reproduction URL

https://github.com/olaisen81/ion-label-issue

### Additional Information

_No response_

sean-perkins:
## Pull request checklist

## Pull request type

 

 

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

Stencil hydrated web components can delay the execution timing for the `connectedCallback()` on first initialization. This results in the Angular control value accessors writing the value to the element before it is initialized. 

In the event of using Angular forms with a form control that has an existing value; this will mean that the control will be initialized without a value. The `emitStyle()` from our Ionic form controls will update the `ion-item` appearance, without the appropriate `item-has-value` class. Then, before `attributeChangedCallback()` is registered, the value will be assigned to the web component. This results in the inner control displaying with a value, but the side-effect behavior that `ion-item` is out of sync.

This wasn't an issue before https://github.com/angular/angular-cli/commit/22e0208a9ee6257213b3bf93ac61a2c3d4ac9504, as likely the connectedCallback behavior in Stencil was benefiting from the defect in Zone.js.

Issue Number: Resolves #23809

## What is the new behavior?

The Ionic custom control value accessors will delay the `writeValue` operation (assigning the value to the element), until after both the custom element is defined and the `componentOnReady` life cycle method has completed. 

This aligns the Ionic web components with native web component behavior that the operations will occur in this order for each initialization of the component:
1. `connectedCallback()`
2. `controlValueAccessor.#writeValue()`

Updates the `removeEventListener` implementation for Stencil hydrated web components to make use of the monkey-patch method, just as the `addEventListener` operates. 

- [ ] Yes
- [x] No

### DIFF
--- a/angular/src/directives/control-value-accessors/value-accessor.ts
+++ b/angular/src/directives/control-value-accessors/value-accessor.ts
@@ -1,11 +1,29 @@
-import { AfterViewInit, ElementRef, Injector, OnDestroy, Directive, HostListener } from '@angular/core';
-import { ControlValueAccessor, NgControl } from '@angular/forms';
+import {
+  AfterViewInit,
+  ElementRef,
+  Injector,
+  OnDestroy,
+  Directive,
+  HostListener,
+  Input,
+  forwardRef,
+  Provider,
+} from '@angular/core';
+import { ControlValueAccessor, NgControl, NG_VALUE_ACCESSOR } from '@angular/forms';
 import { Subscription } from 'rxjs';
 
 import { raf } from '../../util/util';
 
+export const valueAccessorProvider = (token: any): Provider => ({
+  provide: NG_VALUE_ACCESSOR,
+  useExisting: forwardRef(() => token),
+  multi: true,
+});
+
 @Directive()
 export class ValueAccessor implements ControlValueAccessor, AfterViewInit, OnDestroy {
+  @Input() ngModel: any;
+
   private onChange: (value: any) => void = () => {
     /**/
   };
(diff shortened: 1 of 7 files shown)


sean-perkins:
Looks like `ngModel` implementation behaves differently than `NgControl`, with it's own set of oddities (https://github.com/angular/angular/issues/14988). 

Need to dig a little deeper to make this PR satisfy both requirements.

angular/src/directives/control-value-accessors/value-accessor.ts
```diff
@@ -1,11 +1,29 @@
...
 
+export const valueAccessorProvider = (token: any): Provider => ({
+  provide: NG_VALUE_ACCESSOR,
+  useExisting: forwardRef(() => token),
+  multi: true,
+});
+
 @Directive()
 export class ValueAccessor implements ControlValueAccessor, AfterViewInit, OnDestroy {
+  @Input() ngModel: any;
```
sean-perkins, line 25: If we wanted to keep our base class clean of this moneky-patch, we could create new directives with selectors for `ngModel` vs. `formControl/formControlName`. It would create x2 our value accessor files, but lowers nested complexity.

sean-perkins:
Discovered a flaw with this while triaging another issue and checking this fix against it. 

The `ngModel` change, actually fixes an issue where form controls are emitting `ionChange` the 2nd+ time it's rendered (i.e. a datetime in an inline modal) when opened (#24452). This is because the value is now being assigned before the component fully mounts.

It however regresses the form control behavior to do what `ngModel` was previously doing; as the write value is deferred until after the component is mounted. This results in our `@Listen` callbacks for value change to fire and re-emit the same value as a change.

I need to dig further into this, it may be a race condition between web component initialization and when Angular decides to write the value prop.

Issue #24377, opened 2021-12-13
bug: ion-page render problem


outdoorfish:
### Prerequisites

- [X] I have read the [Contributing Guidelines](https://github.com/ionic-team/ionic-framework/blob/main/.github/CONTRIBUTING.md#creating-an-issue).
- [X] I agree to follow the [Code of Conduct](https://ionicframework.com/code-of-conduct).
- [X] I have searched for [existing issues](https://github.com/ionic-team/ionic-framework/issues) that already report this problem, without success.

### Ionic Framework Version

- [ ] v4.x
- [ ] v5.x
- [X] v6.x

### Current Behavior

1. if route change faster than router transition, ion-page will not render
2. some situations, ion-page render incorrect

### Expected Behavior

ion-page render correct

### Steps to Reproduce

1. change route faster than router transition

https://user-images.githubusercontent.com/92348687/145746694-cf9cacd0-392e-4c92-aa66-0915055fe4ac.mov

2. push route, go back, change nested route(replace)

https://user-images.githubusercontent.com/92348687/145746712-a2734e54-8ddd-46fc-8811-ece3655538af.mov

### Code Reproduction URL

https://github.com/outdoorfish/ionic-ion-page-issue/tree/master/ionic-ion-page-issue

liamdebeasi:
Thanks for the issue. I am going to close this as a duplicate of https://github.com/ionic-team/ionic-framework/issues/24226.

Issue #23873, opened 2021-09-03
bug: routing breaks after specific sequence of navigations and clicking on ion-back-button


luxterful:
### Prequisites

- [X] I have read the [Contributing Guidelines](https://github.com/ionic-team/ionic-framework/blob/main/.github/CONTRIBUTING.md#creating-an-issue).
- [X] I agree to follow the [Code of Conduct](https://ionicframework.com/code-of-conduct).
- [X] I have searched for [existing issues](https://github.com/ionic-team/ionic-framework/issues) that already report this problem, without success.

### Ionic Framework Version

- [ ] v4.x
- [X] v5.x
- [ ] v6.x

### Current Behavior

After following a specific order of navigation steps the ion-back-button changes the path but the back animation does not happen and the view stays the same.

### Expected Behavior

The ion-back-button should trigger the animation and change the view

### Steps to Reproduce

Create a blank ionic vue project and create 3 views.

Add those Views to the router and redirect root to **view 2**

The views have buttons to go to the next view
View 1 -> View 2 -> View 3

The views also have an ion-back-button.

The App starts on view 2

1. go to view 3
2. go back
3. go back
4. go to view 2
5. go to view 3
6. go back
7. go back 

Now the URL shows the path to view 1 but the last animation did not happen and we are still on view 2

![ezgif com-gif-maker](https://user-images.githubusercontent.com/26602700/131980272-58dbfdd7-61c5-43ae-97a7-15fd9ebbe9ee.gif)

### Code Reproduction URL

https://github.com/luxterful/ionic-back-routing-issue

### Ionic Info

Ionic:

   Ionic CLI       : 6.16.3 (/home/lukas/.nvm/versions/node/v14.16.1/lib/node_modules/@ionic/cli)
   Ionic Framework : @ionic/vue 5.7.0

Capacitor:

   Capacitor CLI      : 3.2.2
   @capacitor/android : not installed
   @capacitor/core    : 3.2.2
   @capacitor/ios     : not installed

Utility:

   cordova-res : 0.15.3
   native-run  : 1.4.1

System:

   NodeJS : v14.16.1 (/home/lukas/.nvm/versions/node/v14.16.1/bin/node)
   npm    : 7.11.2
   OS     : Linux 5.8

saragonclapps:
I have the same issue and try many ways to force update but doesn't execute, the event router end trigger without  any error.

tigohenryschultz:
[...]

PR #24409, opened 2021-12-15
fix(vue-router): wrong location histories when replace

### LINKED ISSUE #24226: bug: ion-router-outlet breaks when a routing command (such as replace) occurs during a transition.
### Prerequisites

### Ionic Framework Version

- [ ] v4.x
- [X] v5.x
- [ ] v6.x

### Current Behavior

This may be related to #23873  but I'm not sure so I've logged it separately.

I can pretty easily do a set of router commands that breaks the router.

```
router.replace("page1");
// User wait for router animation to finish
router.push("page2");
// User wait for router animation to finish
router.go(-1);
setTimeout(() => {
    router.replace("page3");
}, 50);
```

**Problem 1:**

The code above breaks the page, so it appears no further route modification really does anything. I have an example project of this.

**Problem 2:**

For the last 2 commands, I can't just do this without a setTimeout and expect the router to obey. The router should listen to subsequent commands, but it does not. It stops modifying the route after `router.go(-1)` if the setTimeout is not in place.

```
router.go(-1);
router.replace("page3");
```

### Expected Behavior

My app has a tab setup but whenever you click a tab it should pop the history stack to undo any history pushes that were done inside of a single tab.

Example:
"Tab 1" opens "Page 1". "Page 1" has a button you can click that pushes "Page 2" to the router history.
"Tab 2" opens "Page 3", but in order to do so it must go back in the history 1 step to undo the "Page 2" push, then replace the "Page 1" history state with "Page 3".
This way, tapping on "Tab 2" no matter where you have previously navigated in the application results in the same history state as if you had tapped on it from just starting the application.

### Steps to Reproduce

```
git clone https://github.com/mattsteve/ionic-vue-router-go-broken
npm install
npm run serve
```

Visit `localhost:8080`.

When the app starts it should be on Page 1. Click the "Go to next page" button.

image

At the bottom of the screen click the "Page 3" button.

image

Now at the bottom of the screen, click the "Page 1" button.

image

The URL changes, but you are not visually routed to Page 1.

### Code Reproduction URL

https://github.com/mattsteve/ionic-vue-router-go-broken

### Ionic Info

```
Ionic:

   Ionic CLI : 5.4.16

Utility:

   cordova-res : not installed
   native-run  : not installed

System:

   NodeJS : v12.16.1
   npm    : 6.13.4
   OS     : Linux 5.3


  ────────────────────────────────────────────────────────────

           Ionic CLI update available: 5.4.16 → 6.7.0
                                
     The package name has changed from ionic to @ionic/cli!
                                
             To update, run: npm uninstall -g ionic
                 Then run: npm i -g @ionic/cli

  ────────────────────────────────────────────────────────────
```

### Additional Information

_No response_

yoyo930021:
## Pull request checklist

## Pull request type

 

 

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

The tab page will not render with enteringViewItem === leavingViewItem

Issue Number: #24226 

## What is the new behavior?

The location histories will correct, so enteringViewItem !== leavingViewItem.
The tab page will render.

- [ ] Yes
- [x] No

## Other information

Fixed #24226 

Do I need to add test ?

### DIFF
--- a/packages/vue-router/src/router.ts
+++ b/packages/vue-router/src/router.ts
@@ -238,7 +238,7 @@ export const createIonRouter = (opts: IonicVueRouterOptions, router: Router) => 
           const lastRoute = locationHistory.getCurrentRouteInfoForTab(routeInfo.tab);
           routeInfo.pushedByRoute = lastRoute?.pushedByRoute;
         } else if (routeInfo.routerAction === 'replace') {
-          const currentRouteInfo = locationHistory.last();
+          const currentRouteInfo = locationHistory.current(initialHistoryPosition, currentHistoryPosition);
 
           /**
            * If going from /home to /child, then replacing from


yoyo930021:
The blank page is a different bug.

liamdebeasi (changes requested):
Thanks for the PR! We appreciate your contribution to Ionic. Before this gets merged, could you please write a test for this?

tigohenryschultz:
In my opinion it looks like your change would just ignore another lower level problem. 

const last = () => locationHistory[locationHistory.length - 1];

The last would still reference the wrong viewItem? Shoulnd't the last item return the last item correctly, and it not was the one that is missing not de-mounted properly?

I'm currently tracking a similar bug where the prevRouteLastPathname points to the wrong component because something was replaced and not demounted.

Also doesn't help that none of the types.ts properties does not have comments so it is hard to tell what is used for what / why. Like why would you need to know the previous routes last path name, which  in some cases could be 2 routes ago. 

You would think that calling replace should demount the last item.

You can see a similar issue here, the circle in red is the page I navigated to by 'replace' but the current page is not demounted because prevRouteLastPathname  is a mess

image

The IonRouterOutlet.ts/router.ts/locationHistory.ts is a mess. Should be heavily refactored/commented.

In some cases multiple places can trigger a mount on a component, some things are async and await and others are not awaited properly.

Maybe I am wrong but it is hard to tell anything with a codebase that is not commented.

liamdebeasi:
I took a quick look and noticed that the `routeInfo` object when clicking the back button in Step 7 in https://github.com/ionic-team/ionic-framework/issues/23873#issue-987585140 shows that the current path name is "/page2" and the previous route name is "/page2". So  in other words, it is trying to go back to the same page again. I need to do a deeper dive on why that is getting set.

Regarding location history, the location history is our way of keeping track of navigation items separate from the browser. There are a few reasons for this:

1. The browser does not expose the browser history for (good) security reasons. For context, we only store information regarding route information within the Ionic app. No personal information is stored and this information is not persisted when you close/reload the page.
2. The browser navigation is linear whereas tabs navigation in mobile apps is not linear. So sometimes we need to fight against the browser APIs which gets a bit messy.
3. Ionic Vue supports router.go, meaning you could jump from view 10 to view 2. As a result, we need to preserve the history here so we can traverse it without adding new entries. The "last" item in the location history is not necessarily the "current" item because of this. See https://github.com/ionic-team/ionic-framework/blob/main/packages/vue-router/src/locationHistory.ts#L125-L133 for more context.

Definitely agree that the routing files could do with more commenting, though I'm not sure a refactor would solve much. The routing integrations are pretty robust, but it's mainly these edge cases that we are still trying to work through.

I'll update here/on the issue thread when I have a fix for folks to test. Thanks!

tigohenryschultz:
> I took a quick look and noticed that the `routeInfo` object when clicking the back button in Step 7 in [#23873 (comment)](https://github.com/ionic-team/ionic-framework/issues/23873#issue-987585140) shows that the current path name is "/page2" and the previous route name is "/page2". So in other words, it is trying to go back to the same page again. I need to do a deeper dive on why that is getting set.
> 
> Regarding location history, the location history is our way of keeping track of navigation items separate from the browser. There are a few reasons for this:
> 
> 1. The browser does not expose the browser history for (good) security reasons. For context, we only store information regarding route information within the Ionic app. No personal information is stored and this information is not persisted when you close/reload the page.
> 2. The browser navigation is linear whereas tabs navigation in mobile apps is not linear. So sometimes we need to fight against the browser APIs which gets a bit messy.
> 3. Ionic Vue supports router.go, meaning you could jump from view 10 to view 2. As a result, we need to preserve the history here so we can traverse it without adding new entries. The "last" item in the location history is not necessarily the "current" item because of this. See https://github.com/ionic-team/ionic-framework/blob/main/packages/vue-router/src/locationHistory.ts#L125-L133 for more context.
> 
> Definitely agree that the routing files could do with more commenting, though I'm not sure a refactor would solve much. The routing integrations are pretty robust, but it's mainly these edge cases that we are still trying to work through.
> 
> I'll update here/on the issue thread when I have a fix for folks to test. Thanks!

All the classes purposes make sense but as far as how the code is tangled, seems some responsibilities should be moved away/restructured. Or possibly different patterns implemented  for readability/less bugs, like fail first, or events.

I'm quite deep in debugging the https://github.com/ionic-team/ionic-framework/issues/23873#issue-987585140 but I am not sure if some of the stuff I am seeing are bugs. Like we were going from our /settings page to /settings/about page and it was trigger this:

`          viewStacks.mountIntermediaryViews(id, leavingViewItem, delta);`

Where it is a one-off for mounting views, and in this case it didn't do anything. The mounting was still performed #24329 

`    const setupViewItem = async (matchedRouteRef: any) => {`:

Why have two places mount code? I know the mountIntermediaryViews was commented but it still makes no sense.

For the entire block of code:
```

 if (routerAction === 'replace') {
                    leavingViewItem.mount = false;
                    leavingViewItem.ionPageElement = undefined;
                    leavingViewItem.ionRoute = false;
                }
                else if (!(routerAction === 'push' && routerDirection === 'forward')) {
                    const shouldLeavingViewBeRemoved = routerDirection !== 'none' && leavingViewItem && (enteringViewItem !== leavingViewItem);
                    if (shouldLeavingViewBeRemoved) {
                        leavingViewItem.mount = false;
                        leavingViewItem.ionPageElement = undefined;
                        leavingViewItem.ionRoute = false;
                        viewStacks.unmountLeavingViews(id, enteringViewItem, delta);
                    }
                }
                else {
                    viewStacks.mountIntermediaryViews(id, leavingViewItem, delta);
                }
```

Why would the leaving view not be mounted, if the user keeps navigating forward, don't demount? Like /settings -> /settings/a -> /settings/a/b? Then if the user starts using the ion-back-button would these be 'demounted'?

 Is there a place we can ask code questions?

Why do you inline demount the leaving view but not the enteringViewItem that uses ' viewStacks.unmountLeavingViews', shouldn't the leavingViewItem be passed to 'viewStacks.unmountLeavingViews'?

I would imagine the demounting should be performed when the item is removed from the locationHistory stack, not from the IonRouterOutlet code. Now things can get messy on what is mounted/not mounted/in the location stack/not in the location stack.

liamdebeasi:
`setupViewItem` is for setting up the view for a new route that we have not seen yet. This is for components that are not yet in the DOM and not yet registered with Vue. This function creates the view item, mounts it, and then proceeds with a transition.

`mountIntemediaryViews` is primarily for accounting for `router.go`. When you go from view item 1 to view item 10, you need to remount views 2-9 so that the swipe to go back gesture can occur.  `unmountLeavingViews` is for the same thing, just in the opposite direction (I.e. going from item 10 to item 1). We do not mount the previous views when starting the back transition because you would get a rendering flicker on lower end devices and the view would appear to "pop in".

---- 

I can understand the desire to get these lingering questions answered, but I'd prefer the focus of this thread be kept on the PR and the code changes contained within the PR. For other general questions about Ionic, you can ask them on the [Ionic Forums](https://forum.ionicframework.com/).

tigohenryschultz:
> `setupViewItem` is for setting up the view for a new route that we have not seen yet. This is for components that are not yet in the DOM and not yet registered with Vue. This function creates the view item, mounts it, and then proceeds with a transition.
> 
> `mountIntemediaryViews` is primarily for accounting for `router.go`. When you go from view item 1 to view item 10, you need to remount views 2-9 so that the swipe to go back gesture can occur. `unmountLeavingViews` is for the same thing, just in the opposite direction (I.e. going from item 10 to item 1). We do not mount the previous views when starting the back transition because you would get a rendering flicker on lower end devices and the view would appear to "pop in".
> 
> I can understand the desire to get these lingering questions answered, but I'd prefer the focus of this thread be kept on the PR and the code changes contained within the PR. For other general questions about Ionic, you can ask them on the [Ionic Forums](https://forum.ionicframework.com/).

My only questions would be code specific, which it looks like you are responsible for 99% of. Perhaps I can make a pull request and ask questions/add your answers as comments to commit to the codebase?

setupViewItem should still just create the view item and viewStacks should mount it, then setupViewItem can continue with the transition. There shouldn't be mounting code in 10 places.

yoyo930021:
Ha, I want to know what after in PR's? Continue to add tests? Or trash?

I found at least four problems in the router.
1. #23873
> The going-back will break routing
2. #24226
3. The first problem in #24377
>  The screen will be blank when the quick page switches.
4. The second problem in #24377
> When push new page and go back after, replace page will render the wrong page.

Is the fourth problem the same as the second problem? or the same as the third problem?

This PR only can solve the fourth problem.

mhartington:
Stepping in here. 
I'd like to remind @yoyo930021 and @tigohenryschultz that Ionic has a code of conduct.

https://github.com/ionic-team/ionic-framework/blob/main/CODE_OF_CONDUCT.md#our-standards

We expect everyone involved to be professional and respectful. 
This will count as a warning towards both of you.

Locking this PR for now.

sean-perkins:
Closing this in favor of PR #24433. Appreciate the time that everyone has put into this and look forward to constructive collaboration on that PR and future works. Thanks all & happy holidays 🎉

Issue #24208, opened 2021-11-11
bug: datetime should not scroll to month is entire month is disabled


himanshusukhpal:
### Prerequisites

- [X] I have read the [Contributing Guidelines](https://github.com/ionic-team/ionic-framework/blob/main/.github/CONTRIBUTING.md#creating-an-issue).
- [X] I agree to follow the [Code of Conduct](https://ionicframework.com/code-of-conduct).
- [X] I have searched for [existing issues](https://github.com/ionic-team/ionic-framework/issues) that already report this problem, without success.

### Ionic Framework Version

- [ ] v4.x
- [ ] v5.x
- [X] v6.x

### Current Behavior

Disabled months shouldn't get populated or be scrolled to

![image](https://user-images.githubusercontent.com/10000755/141270827-1b519f3b-94ad-4a6d-8b5e-32907a56ba17.png)

Even the official docs are showing this problem 

![image](https://user-images.githubusercontent.com/10000755/141271880-227ea80d-c3f2-4031-b0a7-a89f3b338d6d.png)

### Expected Behavior

The left or right arrow/swipe ability that if lead to a disabled month can be disabled for the current month

![image](https://user-images.githubusercontent.com/10000755/141271029-a21bee75-29a9-4bdf-8799-b33a2d06e4ec.png)

### Steps to Reproduce

Ion-Datetime calendar's default functionality

Yura13:
@sean-perkins @liamdebeasi 
After this fix appears a new issue.
Previously I've set a date to 19 January and saved it.  In my app, I have logic that I can change the date only during the current month. I can't change the date from 19 January to 9 February because navigation to next month doesn't work but the next month button is enabled.
![image](https://user-images.githubusercontent.com/17030431/153172754-7d554637-5621-402e-a8bc-7ddf4772a15f.png)

PR #24421, opened 2021-12-15
fix(datetime): prevent navigating to disabled months

sean-perkins:
## Pull request checklist

## Pull request type

 

 

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

Users can navigate to months outside of the min/max ranges, through either the arrow buttons or through swiping. Additionally the date time is always defaulting the date to today, when a value is not specified; which doesn't always respect min/max ranges.

- Swipe gestures will lock up on a desktop browser using `mode="ios"`

Issue Number: #24208, #24482

## What is the new behavior?

- The previous/next arrow buttons are disabled depending on the min/max value and the current month
- Swipe gestures are disabled for disabled months (months outside of the min/max range)
- Swipe gestures will not lock-up on desktop browsers using `mode="ios"` (threshold is 1 on desktop, [0.7, 1] on mobile touch devices)

- [ ] Yes
- [x] No

## Other information

To test these changes, you can leverage the existing `minmax` test:
1. `npm run start` (from core)
2. Visit: http://localhost:3333/src/components/datetime/test/minmax
3. Validate the next/previous buttons are disabled 
4. Validate the swipe gestures are disabled (enable device preview to test this)

|Before|After|
|---|---|
|<video src="https://user-images.githubusercontent.com/13732623/146264523-331fee71-4e05-44e6-af7f-2c4021b9e40c.mp4"></video>|<video src="https://user-images.githubusercontent.com/13732623/146264208-3ff01f66-e6e9-4820-a48c-d8533b7a67d9.mp4"></video>|

Special attention in before/after:
- The default date is Today (Dec, 15) in the before, even though the min/max is September to October.

Note: Above videos are outdated. The user can swipe to see the disabled month, but the container will not snap. Can re-record videos if necessary.


core/src/components/datetime/datetime.tsx
```diff
@@ -930,9 +930,40 @@ export class Datetime implements ComponentInterface {
...
+
+      if (todayParts.month < this.minParts?.month && todayParts.year <= this.minParts?.year || todayParts.month > this.maxParts?.month && todayParts.year >= this.maxParts?.year) {
+        /**
+         * Today's date is either below the minimum date range or exceeds the maximum
+         * date range.
+         *
+         * Take the minimum date value to default the user at the lower bounds
+         * of the date range.
+         */
+        dateParts = {
```
sean-perkins, line 950: I'm not completely happy with this implementation. I think defaulting a selected date should be a feature behind a property and not default behavior. 

Regardless, with the aspect of defaulting the date; both `min` and `max` support partial datetime format. So you can have an instance of `min="2021-09"`. Our date parts need the additional pieces to render the datetime accurately.

This logic attempts to remedy that, by taking the lowest value when looking at a scenario when "today" is either before the min range or after the max range. It will apply "good defaults" for the date parts with the data we have.

For instance if you are looking at:
- Today - 12/15/2021
- Min: 09/2021
- Max 10/2021

It will set the default value at the lower range, which is 09/01/2021.

core/src/components/datetime/test/minmax/e2e.ts
```diff
@@ -1,6 +1,7 @@
 import { newE2EPage } from '@stencil/core/testing';
+import { expectFiles } from '@stencil/core/testing/testing-utils';
```
amandaesmith3, line 2: Do we still need this import?

amandaesmith3 (changes requested):

core/src/components/datetime/datetime.tsx
liamdebeasi: Some libraries do not seem to do this:

Vuetify: https://codepen.io/liamdebeasi/pen/vYexwPK?editors=1010
Quasar: https://codepen.io/liamdebeasi/pen/RwLpmdG?editors=1010

It does look like Angular Material has this behavior, however: https://stackblitz.com/edit/angular-yon8wf?file=src%2Fapp%2Fdatepicker-min-max-example.ts

I think it might be confusing to have a dev enter "12/15/2021" and have the date show up as "09/01/2021" without warning. Happy to discuss more on this though.

liamdebeasi (changes requested):
Swiping to the min month on iOS causes the IntersectionObserver to do something strange. Maybe we need to have the IO get disabled/return early when at bounds?

https://user-images.githubusercontent.com/2721089/146397000-60d03d62-0489-4ee9-86fa-aa9c15692439.MP4

core/src/components/datetime/datetime.tsx
```diff
@@ -988,6 +1019,14 @@ export class Datetime implements ComponentInterface {
...
   }
 
+  private isPrevMonthDisabled = () => {
+    return this.workingParts.month <= this.minParts?.month && this.workingParts.year <= this.minParts?.year;
+  }
+
+  private isNextMonthDisabled = () => {
+    return this.workingParts.month >= this.maxParts?.month && this.workingParts.year >= this.maxParts?.year;
+  }
+
```
liamdebeasi, line 1029: We might want to consider moving this to the `state.ts` utility class. This has other functions that determine whether or not a day/month is disabled.

Additionally, the datetime file is already pretty large, so might be good to reduce the amount of new stuff we put in there.

core/src/components/datetime/datetime.tsx
liamdebeasi: Tried checking what Flutter does and the datepicker seems to crash when the current value is out of range. Put the code sample from https://api.flutter.dev/flutter/material/showDatePicker.html into the DartPad demo: https://dartpad.dev/embed-flutter.html?gh_owner=material-components&gh_repo=material-components-flutter&gh_path=docs/components/dartpad/date_pickers/regular&gh_ref=develop

From there, change the value on value/min/max dates to be what we have in this sample.

sean-perkins:
The swipe gesture Liam called out still remains. Doing more discovery on my end.

I'm able to replicate the same type of behavior in Safari Web, where you cannot navigate to the previous month. It seems that in Safari, the state variable for working parts gets updated twice in one cycle in the IO. This issue isn't present in Chrome.

amandaesmith3:
One last bug I noticed (sorry 😅): on iOS, if I use the "value inside bounds" test as-is, I'm unable to change the month via the dropdown/picker. It just moves back to the month I had selected after a short delay. Swiping between months on the calendar view and using the prev/next arrow buttons both work okay.

sean-perkins:
> If I set the `min` and `max` to something that excludes today and leave the `value` off, it still defaults to today, showing a disabled month. The PR description still mentions that this should be fixed, but I don't see the fix anymore in the changes; was it removed by mistake or does the description just need updating? I see there was discussion about other libraries not handling this behavior.

Updated the PR description, thanks for catching that! I had originally accounted for that behavior, but reverted as it changes current behavior fo the datetime. We should likely sync up and discuss/compare alternatives. Setting the date when unset to today has spawned other reported issues as well.

> One last bug I noticed (sorry 😅): on iOS, if I use the "value inside bounds" test as-is, I'm unable to change the month via the dropdown/picker. It just moves back to the month I had selected after a short delay. Swiping between months on the calendar view and using the prev/next arrow buttons both work okay.

Will take a look!

sean-perkins:
@amandaesmith3 are you seeing different behavior than this (commented out other tests):

<video src="https://user-images.githubusercontent.com/13732623/151409680-98b88868-dbe1-41d0-bd9e-0f8e043eb36d.mp4"></video>

May need OS version, could be a Safari 14 vs. 15 thing.

amandaesmith3:
Here's what I'm getting: https://user-images.githubusercontent.com/90629384/151412886-2a56aa14-b9be-4144-bace-55ebb2f6df6a.mp4

This is on iOS 14.7.1. Sorry, should've mentioned that 🤔

sean-perkins:
Can confirm, there's an issue explicitly with Safari 14. 

Updating the working parts when the `ion-picker-column-internal` emits `ionChange`, causes 2 render cycles. Rendering once with the selected month/year and then re-rendering with the old value.

sean-perkins:
@amandaesmith3 should be good for another test! I verified with Safari 14 on desktop. Also verified that this change doesn't regress behavior on iOS 15/Mobile Safari 15.

amandaesmith3 (approved):
Looks great to me now, awesome work! 🎉

core/src/components/datetime/utils/state.ts
```diff
@@ -102,3 +141,103 @@ export const getCalendarDayState = (
     ariaLabel: generateDayAriaLabel(locale, isToday, refParts)
   }
 }
+
+/**
+ * Given a working date, an optional minimum date range,
+ * and an optional maximum date range; determine if the
+ * previous navigation button is disabled.
+ */
+export const isPrevMonthDisabled = (
```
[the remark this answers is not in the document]
liamdebeasi: Also having `isMonthDisabled` might be useful if we ever expand upon disabled states in future releases

core/src/components/datetime/datetime.tsx
```diff
@@ -963,9 +974,9 @@ export class Datetime implements ComponentInterface {
   }
 
   componentWillLoad() {
-    this.processValue(this.value);
     this.processMinParts();
     this.processMaxParts();
+    this.processValue(this.value);
```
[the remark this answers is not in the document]
sean-perkins: Based on the current changes, it doesn't need to be, but I would recommend it stays in this position. Originally I had logic that manipulated the defaulted date value for this portion:
```
const valueToProcess = value || getToday();
```

that would take in consideration the min & max parts (which need to be set before this method was called). That work was ultimately reverted, since we need to investigate and discuss how we wish to handle dates outside bounds and if we wish to keep setting the default day (#24456).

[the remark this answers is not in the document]
liamdebeasi: Sounds good, let's keep it the way you have it.

core/src/components/datetime/datetime.tsx
```diff
@@ -724,7 +733,8 @@ export class Datetime implements ComponentInterface {
          */
         if (mode === 'ios') {
           const ratio = ev.intersectionRatio;
-          const shouldDisable = Math.abs(ratio - 0.7) <= 0.1;
+          // `maxTouchPoints` will be 1 in device preview, but > 1 on device
+          const shouldDisable = Math.abs(ratio - 0.7) <= 0.1 && navigator.maxTouchPoints > 1;
```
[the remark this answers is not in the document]
sean-perkins: It will no longer lock-up on browser, but it will animate early (at the 70% threshold mark). Let me see if adding the `maxTouchPoints` check when determining the IO threshold works both on device & web browser.

core/src/components/datetime/utils/state.ts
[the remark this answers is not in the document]
sean-perkins: This eliminated so much duplicate code. Awesome recommendation. Totally missed the manipulation utils existing to ease these calculations. 

Ended up making the API:
```ts
const isMonthDisabled(refParts, { minParts, maxParts }: { minParts?, maxParts? }) {
  ...
}
```

That way you don't need to supply `undefined` as a parameter for the next month check:
```ts
isMonthDisabled(nextMonth, undefined, maxParts);
// vs.
isMonthDisabled(nextMonth, { maxParts });
```

& this aligns with prev month:
```ts
isMonthDisabled(prevMonth, { minParts });
```

core/src/components/datetime/utils/state.ts
```diff
@@ -102,3 +142,81 @@ export const getCalendarDayState = (
...
+  }
+}) => {
+  // If the min date is set and the year is less than the min year.
+  if (minParts && minParts.year > refParts.year) {
+    return true;
+  }
+  // If the max date is set and the year is greater than the max year.
+  if (maxParts && maxParts.year < refParts.year) {
+    return true;
+  }
```
liamdebeasi, line 163: This can be simplified to

```js
if (isYearDisabled(refParts, minParts, maxParts)) {
  return true;
}
```

core/src/components/datetime/utils/state.ts
```diff
@@ -102,3 +142,81 @@ export const getCalendarDayState = (
...
+    day: number | null
+  },
+  minParts?: {
+    month?: number,
+    year: number
+  },
+  maxParts?: {
+    month?: number,
+    year: number
+  }) => {
```
liamdebeasi, line 196: We do we use this instead of `DatetimeParts`?

core/src/components/datetime/datetime.tsx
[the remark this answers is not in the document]
sean-perkins: Confirmed on device and web browser that using this check for assigning the threshold values on the intersection observers looks good.

core/src/components/datetime/utils/state.ts
```diff
@@ -102,3 +142,81 @@ export const getCalendarDayState = (
...
+  // If the min date is set and the year is the same as the min year,
+  // but the month is less than the min month.
+  if (minParts && minParts.year === refParts.year && minParts.month !== undefined && minParts.month > refParts.month) {
+    return true;
+  }
+  // If the max date is set and the year is the same as the max year,
+  // but the month is greater than the max month.
+  if (maxParts && maxParts.year === refParts.year && maxParts.month !== undefined && maxParts.month < refParts.month) {
+    return true;
+  }
```
liamdebeasi, line 173: The `isBefore` and `isAfter` utils in `utils/comparison.ts` handle this for us I think:

```js
if (isBefore(refParts, minParts) {
  return true;
}

if (isAfter(refParts, maxParts)) {
  return true;
}
```

core/src/components/datetime/datetime.tsx
```diff
@@ -714,6 +717,12 @@ export class Datetime implements ComponentInterface {
           return;
         }
 
+        const { month, year, day } = refMonthFn(this.workingParts);
+
+        if (isMonthSwipeDisabled(month, year, this.workingParts, this.minParts, this.maxParts)) {
```
liamdebeasi, line 722: Could this just be `isMonthDisabled`? Not sure I understand the purpose of this function.

core/src/components/datetime/utils/state.ts
sean-perkins: The `DatetimeParts` type is a bit of a beast. It's being used for a lot of utility functions that only require the `month` or `year`. As a result, we are actually overriding Typescript's ability to do type inference and using the postfix `!` operator to remove the `null` & `undefined` type. 

Really these functions should be as open as possible, only requiring the date parts that is uses or is dependent on based on inner function calls. 

I can refactor these to use `DatetimeParts`, but it makes the API usage something like:
```ts
isNextMonthDisabled({ month: 10, year: 2021, day: undefined })
```

liamdebeasi: Ahh I see. Ok let's keep it as you have it 👍

sean-perkins: I could see it making sense being consistent with `Datetimeparts` for the type here, but re-evaluating across the implementation if simpler types could exist for the utils. 

If you're good with that, I can create a tech debt task.

liamdebeasi: Yeah I see your point about `DatetimeParts` being excessive for something like this. Tech debt to explore a better way to do this would be good.

core/src/components/datetime/datetime.tsx
sean-perkins: Let me re-evaluate it based on the recent changes to addressing the Webkit IO issue. 

Originally it existed, because I was using `display: none` for the `calendar-month-disabled` class and we needed to render the working month, even if it was disabled (for instances where your current date value was outside of the min/max bounds). 

That may not exist anymore.

sean-perkins: Ok cool, it's needed for one case with rendering, so that you cannot free scroll the container when all months are disabled, but a day/week within the working month is enabled. Instead of having a util for that single case, I migrated it to the `isMonthDisabled` util and added the check in the `datetime.tsx` render.

core/src/components/datetime/datetime.tsx
```diff
@@ -1114,6 +1145,13 @@ export class Datetime implements ComponentInterface {
                 items={years}
                 value={workingParts.year}
                 onIonChange={(ev: CustomEvent) => {
+                  // Due to a Safari 14 issue we need to destroy
+                  // the intersection observer before we update state
+                  // and trigger a re-render.
+                  if (this.destroyCalendarIO) {
+                    this.destroyCalendarIO();
```
liamdebeasi, line 1152: What happens if we remove this? I tried testing in iOS 14 and I did not notice anything break.

liamdebeasi (approved):
The functionality is working well and the code looks good. I noticed one visual glitch I was hoping we could try and fix:

https://user-images.githubusercontent.com/2721089/151858491-86608d04-af29-4122-814d-39cf1ab444c0.MP4

When swiping quickly on iOS, it looks like the scrolling is reset somehow. I can reproduce this on iOS 14+15. I do not consider this to be blocking, but it would be great if we could fix it before merging.

core/src/components/datetime/datetime.tsx
sean-perkins: This specific code was to catch an issue when using the calendar month picker and changing months through that UI would re-render and cause the intersection observer to update the month back to the previous month. 

I noticed it prior by going from October -> September, then expanding the month picker and changing between months.

Was from this comment: https://github.com/ionic-team/ionic-framework/pull/24421#issuecomment-1023476866 (only applicable in Safari < 15).

Will prioritize time later tonight/early tomorrow to address the other quick scroll issue.

sean-perkins: We have identified this as a webkit bug, new issue being tracked here: #24683.

PR is ready to merge in the interim.

Issue #24291, opened 2021-12-01
bug: ion-item > inputs are all disabled if a child of ion-item is


me-to-you93:
### Prerequisites

- [X] I have read the [Contributing Guidelines](https://github.com/ionic-team/ionic-framework/blob/main/.github/CONTRIBUTING.md#creating-an-issue).
- [X] I agree to follow the [Code of Conduct](https://ionicframework.com/code-of-conduct).
- [X] I have searched for [existing issues](https://github.com/ionic-team/ionic-framework/issues) that already report this problem, without success.

### Ionic Framework Version

- [ ] v4.x
- [X] v5.x
- [ ] v6.x

### Current Behavior

Let's pretend a ion-item have some input childs like ion-checkbox or ion-input. If one of the child input is [disabled]="true", the whole ion-item is disabled too.

### Expected Behavior

Disabling one child should not propagate to other children

### Steps to Reproduce

Just have a ion-item with some disabled child inputs and some not.

Ionic:

   Ionic CLI                     : 5.4.16 (C:\Users\gmartin\AppData\Roaming\npm\node_modules\ionic)
   Ionic Framework               : @ionic/angular 5.9.1
   @angular-devkit/build-angular : 12.2.4
   @angular-devkit/schematics    : 12.2.4
   @angular/cli                  : 12.2.4
   @ionic/angular-toolkit        : 4.0.0

Cordova:

   Cordova CLI       : 10.0.0 (cordova-lib@10.1.0)
   Cordova Platforms : not available
   Cordova Plugins   : cordova-plugin-ionic-keyboard 2.2.0, cordova-plugin-ionic-webview 5.0.0, (and 18 other plugins)

Utility:

   cordova-res : 0.15.3
   native-run  : 1.5.0

System:

   Android SDK Tools : 26.1.1 (C:\Users\gmartin\AppData\Local\Android\Sdk)
   NodeJS            : v16.13.0 (C:\Program Files\nodejs\node.exe)
   npm               : 8.1.0
   OS                : Windows 10

### Additional Information

Therefore this was already fixed in this issue : https://github.com/ionic-team/ionic-framework/issues/18670
 and should be available since 4.6.1

amandaesmith3:
Thanks for the report, I'm able to reproduce. Looks like there's been a regression since https://github.com/ionic-team/ionic-framework/pull/18696; the extra inputs look interactable, but nothing happens when clicking them.

me-to-you93:
[...]

PR #24423, opened 2021-12-15
fix(radio): fix radio not showing checked state when not in a group

amandaesmith3:
## Pull request checklist

## Pull request type

 

 

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

Issue Number: https://github.com/ionic-team/ionic-framework/issues/24291

If an `ion-radio` is not part of an `ion-radio-group`, clicking it does not apply the `radio-checked` class, making it look like nothing happened. However, the native `input` *does* get correctly set to `checked=true`.

This bug looks to be masquerading as the one described in the issue linked above. While having multiple inputs in an `ion-item` and disabling one of them does *not* currently disable the other inputs, using an `ion-radio` as the enabled input can make it look that way.

## What is the new behavior?

The `checked` property of the `ion-radio` component is updated on click to match the native input. This is then picked up by the existing `@Watch('checked')`, which applies the proper styling.

- [ ] Yes
- [x] No

## Other information

Dev build posted on the original issue to be definitively sure that the source was this bug.

### DIFF
--- a/core/src/components/item/test/disabled/index.html
+++ b/core/src/components/item/test/disabled/index.html
@@ -116,6 +116,24 @@
         </ion-item>
 
         <ion-item>
+          <ion-radio disabled slot="start"></ion-radio>
+          <ion-checkbox slot="start"></ion-checkbox>
+          <ion-label>Radio + Checkbox</ion-label>
+        </ion-item>
+
+        <ion-item>
+          <ion-radio slot="start"></ion-radio>
+          <ion-checkbox disabled slot="start"></ion-checkbox>
+          <ion-label>Radio + Checkbox, disabled input second</ion-label>
+        </ion-item>
+
+        <ion-item>
+          <ion-checkbox slot="start"></ion-checkbox>
+          <ion-radio disabled slot="start"></ion-radio>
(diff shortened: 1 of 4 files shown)


sean-perkins (approved):
Looks good 👍

liamdebeasi (commented):
Does the issue reported in https://github.com/ionic-team/ionic-framework/issues/24291 only trigger with the use of `ion-radio`? If so, I'd like to discuss whether using `ion-radio` without a group is a use case we should support.

The docs currently state that radios should be used inside of `ion-radio-group`: https://ionicframework.com/docs/api/radio

amandaesmith3:
@liamdebeasi As far as I can tell, yes. I posted a dev build for the OP to try out in case there's an edge case I missed, but I tried every combination I could think of and only saw the issue described in this PR.

Our tests do have items for radios outside groups, so I figured that meant they were supported 🤔 There's also code in the component specifically to check whether it's in a group.

liamdebeasi (approved):
The best practice right now is to use radios in a group, but even the native radio input works as a standalone, so I think this is a good fix to merge to better align with the native web behavior. Great job!

Issue #24386, opened 2021-12-13
bug: content shadow styles causing performance issues at scale


tlaverdure:
### Prerequisites

- [X] I have read the [Contributing Guidelines](https://github.com/ionic-team/ionic-framework/blob/main/.github/CONTRIBUTING.md#creating-an-issue).
- [X] I agree to follow the [Code of Conduct](https://ionicframework.com/code-of-conduct).
- [X] I have searched for [existing issues](https://github.com/ionic-team/ionic-framework/issues) that already report this problem, without success.

### Ionic Framework Version

- [ ] v4.x
- [X] v5.x
- [X] v6.x

### Current Behavior

This issue attempts to address performance issues with `ion-content`, a continuation of #24359.

There are some performance issues with `ion-content` as more child DOM nodes are added. Any interaction with the pointer causes `Styles Invalidated` and `Styles Recalculated` events to occur that are very slow. I'm not sure if this is an issue with:
* How webkit optimizes shadow DOM styling
* If `will-change: scroll-position` is not performing well here.
https://github.com/ionic-team/ionic-framework/blob/03dd3729332da31c1541e5f62084d12a852231f6/core/src/components/content/content.scss#L107
* If webkit is recomputing math like in situations like these:
https://github.com/ionic-team/ionic-framework/blob/03dd3729332da31c1541e5f62084d12a852231f6/core/src/components/content/content.scss#L168-L169

### Expected Behavior

The styles required for `ion-content` should not introduce performance issues as the number of DOM nodes increases within the component's slot.

### Steps to Reproduce

In Safari, using developer tools you should be able to use the "Timeline" to inspect the CPU usage of this page after it has loaded:

https://angular-ivy-7he8g8.stackblitz.io

By just moving my mouse and clicking around CPU is maxed out. All related to styles and painting on the main thread.

https://user-images.githubusercontent.com/1731025/145899511-181cb2b4-89bd-45da-aaca-19e98bd2e9d5.mov

Now if I delete the `<style>` tag within the `<ion-content>` shadow DOM, the overall CPU usage is much less.

https://user-images.githubusercontent.com/1731025/145899583-013d680a-0ae5-48db-84aa-84081b103b8c.mov

### Code Reproduction URL

https://stackblitz.com/edit/angular-ivy-7he8g8

### Ionic Info

latest

liamdebeasi:
[...]

Issue #23437, opened 2021-06-09
feat: improve component compatibility with virtual scrolling solutions


liamdebeasi:
# Feature Request

**Ionic version:**

[x] **5.x**

**Describe the Feature Request**

At the moment, components such as `ion-refresher`, `ion-infinite-scroll`, and the collapsible large title rely on users scrolling on `ion-content`. As a result, when scrolling in a virtual scroll container, these components do not work as expected. While we cannot account for every single virtual scrolling solution, I think we can expose an API that lets users have more control over which scrollable container these components listen on.

This impacts solutions like CDK Scroller, Virtuoso, vue-virtual-scroller, etc.

For developers that would like to see this feature added to the Framework, please react to this post with a thumbsup/"+1" reaction. Thanks!

ghenry22:
It would be great if you could also validate that ion-reorder works with these alternative scrollers.  It is currently broken when used with ion-virtual-scroll but the bug reporting this was closed "because virtual-scroll will be deprecated".

mikkimichaelis:
I am using the following code which incorporates ion-virtual-scroll and ion-infinite-scroll.  Yes it has some issues but atm I can live with them and otherwise things seem to work just fine.  I found this thread from a console log regarding ion-virtual-scroll deprecation.  Is the OP stating that ion-virtual-scroll and ion-infinite-scroll do not work together?

"@ionic/angular": "^5.8.4",
"@angular/core": "~11.2.0",

`<ion-content fullscreen="true">
  <ion-refresher slot="fixed" [pullFactor]="this.settingsService.environment.ionRefresher.pullFactor" [pullMin]="this.settingsService.environment.ionRefresher.pullMin" [pullMax]="this.settingsService.environment.ionRefresher.pullMax" (ionRefresh)="doRefresh($event)">
    <ion-refresher-content>
    </ion-refresher-content>
  </ion-refresher>
  <ion-list>
    <ion-virtual-scroll [items]="dataService.searchMeetings$ | async">
      <ng-container *virtualItem="let meeting">
        <meeting-listing [meeting]="meeting"></meeting-listing>
      </ng-container>
    </ion-virtual-scroll>
  </ion-list>
  <ion-infinite-scroll [threshold]="threshold" (ionInfinite)="loadData($event)">
    <ion-infinite-scroll-content loadingSpinner="bubbles" loadingText="Loading more meetings...">
    </ion-infinite-scroll-content>
  </ion-infinite-scroll>
[...]

PR #24426, opened 2021-12-16
feat: allow custom content implementation with [ion-content] selector

tlaverdure:
## Pull request checklist

## Pull request type

 

 

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes

Allows a custom ion-content implementation to be used. I'm suggesting this new feature because the current implementation of `<ion-content>` requires shadow DOM and CSS variables. Webkit currently does not perform well with large amounts of DOM nodes within `<ion-content>` because of how it applies CSS variables when using shadow DOM.

Issue Number: #24386

## What is the new behavior?

- Allows an element to take the place of the `<ion-content>` component by using the attribute `ion-content`
- The developer is responsible for implementation of the `IonComponent` interface to make use of Ionic utilities and behaviors from other components
- The developer is responsible for adding the necessary styles

```html
<div id="ion-content">
  <div id="background-content"></div>
  <div id="inner-scroll"></div>
</div>
```

```javascript
const content = document.getElementById('content');

content.getScrollElement = async function() {
  return this.querySelector('#inner-scroll');
};

...
```

- [ ] Yes
- [X] No

### DIFF
--- a/core/src/components/content/test/custom/index.html
+++ b/core/src/components/content/test/custom/index.html
@@ -1,0 +1,201 @@
+<!DOCTYPE html>
+<html lang="en" dir="ltr">
+
+<head>
+  <meta charset="UTF-8">
+  <title>Content - Custom</title>
+  <meta name="viewport" content="width=device-width, initial-scale=1.0, minimum-scale=1.0, maximum-scale=1.0, user-scalable=no">
+  <link href="../../../../../css/ionic.bundle.css" rel="stylesheet">
+  <link href="../../../../../scripts/testing/styles.css" rel="stylesheet">
+  <script src="../../../../../scripts/testing/scripts.js"></script>
+  <script nomodule src="../../../../../dist/ionic/ionic.js"></script>
+  <script type="module" src="../../../../../dist/ionic/ionic.esm.js"></script>
+</head>
+
+<body>
+  <ion-app mode="ios">
+    <ion-header collapse="fade">
+      <ion-toolbar>
+        <ion-buttons slot="start">
+          <ion-button onclick="scrollToTop()">
+            <ion-icon slot="icon-only" name="arrow-up"></ion-icon>
+          </ion-button>
+        </ion-buttons>
+
+        <ion-title onclick="statusTap()">Content - Custom</ion-title>
+
+        <ion-buttons slot="end">
+          <ion-button onclick="scrollToBottom()">
+            <ion-icon slot="icon-only" name="arrow-down"></ion-icon>
+          </ion-button>
+        </ion-buttons>
+      </ion-toolbar>
+    </ion-header>
+
+    <div ion-content style="height: 100%; position: relative; overflow: hidden;" id="content">
+      <div id="background-content"></div>
+      <div style="position: absolute; left: 0; top: 0; width: 100%;">
+        <ion-refresher slot="fixed" id="ion-refresher">
+          <ion-refresher-content></ion-refresher-content>
+        </ion-refresher>
+      </div>
+      <div id="inner-scroll">
+        <f style="height: 50%"></f>
+        <f></f>
+        <f></f>
+        <f></f>
+        <f></f>
+        <f></f>
+        <f></f>
+        <f></f>
+        <ion-list>
+          <ion-item>
+            <ion-label>Input</ion-label>
+            <ion-input></ion-input>
+          </ion-item>
+        </ion-list>
+
+        <f></f>
+
+        <ion-infinite-scroll>
+        </ion-infinite-scroll>
+      </div>
+    </div>
+
+    <ion-footer collapse="fade">
+      <ion-toolbar>
+        <ion-title>Footer</ion-title>
+
+        <ion-buttons slot="end">
+          <ion-button>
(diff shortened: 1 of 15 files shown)


tlaverdure:
@liamdebeasi hope you had a great holiday. Could you please take a look at this when you get a chance? Thanks!

liamdebeasi:
Thanks! I am looking to do some discovery soon for how this feature can be used to solve https://github.com/ionic-team/ionic-framework/issues/23437 as well. I will follow up here when I have more information to share.

sean-perkins (commented):
What do we think about creating a constant/util for the content selector? It's a relatively simple selector, but to encourage that new features reliant on querying `ion-content` don't only target the element tag, having a constant would better enforce that pattern.

i.e.:
```ts
export const ION_CONTENT_SELECTOR = 'ion-content, [ion-content]';
```

```ts
.querySelector(ION_CONTENT_SELECTOR)
```

I would also be open to a util that takes the responsibility of the query & closest:
```ts
const ION_CONTENT_SELECTOR = 'ion-content, [ion-content]';

export const findIonContent = (el: HTMLElement) => el.querySelector(ION_CONTENT_SELECTOR);

export const findClosestIonContent = (el: HTMLElement) => el.closest(ION_CONTENT_SELECTOR);
```

tlaverdure:
Great feedback. I can get that added.

tlaverdure:
@sean-perkins Let me know if that looks good. Thanks

core/src/components/content/test/custom/index.html
```diff
@@ -0,0 +1,201 @@
...
+            <ion-icon slot="icon-only" name="help-circle-outline"></ion-icon>
+          </ion-button>
+        </ion-buttons>
+      </ion-toolbar>
+    </ion-footer>
+
+    <script>
+      const content = document.getElementById('content');
+
+      content.getScrollElement = async function() {
```
sean-perkins, line 80: I need to put a little extra thought into the API design around this contract. 

In an example like Angular, it's not as intuitive to override/set the `getScrollElement` method to satisfy existing logic that references that function in a variety of implementations.

i.e.:
```html
<cdk-virtual-scroll-viewport ion-content>
  <ion-refresher slot="fixed">
    <ion-refresher-content></ion-refresher-content>
  </ion-refresher>
  ...
</cdk-virtual-scroll-viewport>
```

We aren't able to easily specify the scroll element, so the call inside `ion-refresher` to get the scroll element will throw an exception. 

Some current thoughts rolling around are:
- Expose a scroll element prop for each element coupled to `ion-content` functionality 
- Default scroll element to `getScrollElement()` or the `ion-content` host if the function isn't available
- Introduce an API method to set the scroll element and re-bind private `scrollEl` references

Just not currently excited that this is really copy/paste across implementations. Will explore other patterns/thoughts. We could potentially auto-detect the scroll target from the `ion-content` host, but that may not always be a good fit for certain implementations.

sean-perkins: Also we don't need to stress doing this as part of this PR 🙂  just more discussion details around it that the Ionic team will need to be aware of as it relates to this change.

tlaverdure: Yeah I totally feel like this PR is more of a "bring your own ion-content" implementation. For my use case, this would work as I'm ready to just create a new content component in my app and copy paste the necessary methods. 

I'm game for whatever you guys think would work best. Ultimately, I'm just trying to avoid shadow DOM and the issues that CSS variables cause for Safari. 🙂

sean-perkins: Ok, spent some time exploring an alternative pattern built on what you're doing in this PR, available here: https://github.com/ionic-team/ionic-framework/compare/FW-342

Explored with just a class name `.ion-content` (we will likely use a more specific name), but I believe we may decide to allow you to customize this selector from the `IonicConfig`, so that in your own implementation you could choose to target `[ion-content]` or whatever. 

Did a few optimizations to find the scroll target for you when not using `ion-content`. We're advancing this feature proposal to an internal review.

Solves two larger issues with one effort:
1. Webkit performance issue you are tracking.
2. Ability to use more of the Ionic eco-system in a virtual scroll context.

tlaverdure: Your implementation looks great 👍🏾. What you did in `content.utils.ts` is key!

tlaverdure: @sean-perkins just checking if there is any update?

tlaverdure:
Hey @liamdebeasi @sean-perkins, are you guys still considering https://github.com/ionic-team/ionic-framework/compare/FW-342 as the implementation for this?

sean-perkins:
@tlaverdure yes, we are planning on following something very similar to that implementation for 6.1.0. The global selector value (`.ion-content`) is subject to change from that discovery work.

tlaverdure:
Sweet! 🎉 

Looking forward to that release. Thanks for the update!

sean-perkins:
Closing in favor of https://github.com/ionic-team/ionic-framework/pull/24883#issuecomment-1059465412. Thank you!

Issue #24432, opened 2021-12-17
bug: router.replace eventually breaks routing


tigohenryschultz:
### Prerequisites

- [X] I have read the [Contributing Guidelines](https://github.com/ionic-team/ionic-framework/blob/main/.github/CONTRIBUTING.md#creating-an-issue).
- [X] I agree to follow the [Code of Conduct](https://ionicframework.com/code-of-conduct).
- [X] I have searched for [existing issues](https://github.com/ionic-team/ionic-framework/issues) that already report this problem, without success.

### Ionic Framework Version

- [ ] v4.x
- [X] v5.x
- [ ] v6.x

### Current Behavior

After navigating around using router.push and router.replace, eventually a view will linger.

### Expected Behavior

Correct rendering of leaving/entering view items.

### Steps to Reproduce

Navigating from View A1 -(replace)-> B/1 -(push)-> B/2  -(back button)-> B/1 -(replace)-> A1 -(replace)-> B1

B1 never gets render and is glitched onto A1.

<img src="https://user-images.githubusercontent.com/75696293/146278422-06e4fd09-3541-4919-83ba-c6bcfcb27863.png" width="350px" />

### Ionic Info

Ionic:

   Ionic CLI : 6.16.3

Utility:

   cordova-res : not installed globally
   native-run  : not installed globally

System:

   NodeJS : v14.17.3
   npm    : 6.14.13
   OS     : Windows 10

sean-perkins:
Partially related to: #24226

We may decide to group the Vue routing issues together to reduce individual & overlapping tasks for us internally.

tigohenryschultz:
> Partially related to: #24226
> 
> We may decide to group the Vue routing issues together to reduce individual & overlapping tasks for us internally.

Since I have a good understand of routing now I don't mind tackling the issues/tickets regarding it, they're pretty high priority issues for the company I work for and we don't want any negative app reviews from routes breaking so I've been given the go ahead to work on this.

tigohenryschultz:
Working on this issue in this pull request: https://github.com/ionic-team/ionic-framework/pull/24433

liamdebeasi:
@tigohenryschultz Do you have a reproduction case for this router.replace issue? I have a fix for https://github.com/ionic-team/ionic-framework/issues/24226 and would like to test to see if my patch also fixes this issue.

Note that this patch does not fix https://github.com/ionic-team/ionic-framework/issues/23873 since that example does not use replace.

[...]

PR #24433, opened 2021-12-17
fix(vue-router): Prev route last pathname incorrect

tigohenryschultz:
## Pull request checklist

## Pull request type

 

 

- [X] Bugfix
- [ ] Feature
- [X] Code style update (formatting, renaming)
- [X] Refactoring (no functional changes, no api changes)
- [ ] Build related changes
- [X] Documentation content changes

Issue Number: https://github.com/ionic-team/ionic-framework/issues/24432

## What is the new behavior?

- New RouteInfo property replaceRoute to track these replacedRoutes
- Proper viewStacks.mount and unmount calling
- Override entire locationHistory item when replacing

- [] Yes
- [x] No

## Other information

I will be testing the router.go logic to see if anything not tested broke because I did change how replace worked in locationHistory, code review would be great for now.


yoyo930021:
I try to use your PR. It works well. It's better for #24409 .
But the #24409 is locked, so I can close it. =_=

I migrate your PR to the v6 version in this branch.
https://github.com/yoyo930021/ionic-framework/tree/fix-route-v6

packages/vue-router/src/router.ts
```diff
@@ -212,59 +221,77 @@ export const createIonRouter = (opts: IonicVueRouterOptions, router: Router) =>
...
+      if (isPushed) {
+        /**
+         * Push/Forward occurs whenever navigating forward, page -> page/a -> page/b
+         *
+         * Going to page/a and page/b would be push events
+         */
+        nextRouteInfo.lastPathname = leavingRouteInfo.pathname;
+        nextRouteInfo.tab = leavingRouteInfo.tab;
+        //@TODO: Check if bug, why would pathname be blank?
+        nextRouteInfo.pushedByRoute = (leavingRouteInfo.pathname !== '') ? leavingRouteInfo.pathname : undefined;
```
sean-perkins, line 244: I believe you can create an instance where `pathname` is `''`, when using `history.pushState`/`history.replaceState`, but also I cannot immediately think of an instance where that's appropriate. From the docs, it appears it happens when there is no path (https://developer.mozilla.org/en-US/docs/Web/API/Location/pathname).

sean-perkins (commented):
We will want to clean-up the commented out code & format any inline code documentation to match existing format, but that can come at a later date.

If there's an opportunity to write tests to cover functional changes to prevent the behavior regressing in the future, I would encourage it.

Other Core members will provide feedback and insight as they are able. Thanks for putting this PR together 👍

packages/vue-router/src/router.ts
tigohenryschultz: I'll add this comment to explain the blank scenario in code

tigohenryschultz: I saw the first route has path = '' before we redirect it to the desired page too. Also I did break tabs so I will fix that and remove commented out code and add test cases for my bug scenario and the tab scenario. Pretty sure there are still bugs when navigating around with push,replace,go,back in complex scenarios so I will see what I find. 


@sean-perkins When going between tabs on the same `ion-tab-bar` do we want each new tab to be pushed onto locationHistory or should the tab replace the previous tab if they keep clicking between tabs? The code base seems to handle tabs in unique ways and I was wondering if we could just have them behave as replacements, that should simplify things.

tigohenryschultz:
Last commit fixes spamming tabs and clicking the ion back button taking you to the last non tab page, but using android device back button only goes -1 to the last tab, or whatever is the last route

packages/vue-router/src/router.ts
```diff
@@ -265,8 +265,11 @@ export const createIonRouter = (opts: IonicVueRouterOptions, router: Router) =>
         nextRouteInfo.pushedByRoute = leavingRouteInfo?.pathname;
         nextRouteInfo.lastPathname = leavingRouteInfo?.pathname;
       } else if (nextRouteInfo.routerAction === 'push' && isNewTab) {
-        // const lastTabRouteInfo = locationHistory.getCurrentRouteInfoForTab(nextRouteInfo.tab);
-        nextRouteInfo.pushedByRoute = (leavingRouteInfo.pathname !== '') ? leavingRouteInfo.pathname : undefined;
+        if (isLeavingRouteTab) {
```
czzplnm, line 268: My thoughts behind this were if any last route was a tab, use its pushedByRoute(which would always be a non-tab route) else use the lastPathname. So if the user pushes ion-back-button it would go to a non-tab page, but android device back would just go back -1

tigohenryschultz, line 268: My thoughts behind this were if any last route was a tab, use its pushedByRoute(which would always be a non-tab route) else use the lastPathname. So if the user pushes ion-back-button it would go to a non-tab page, but android device back would just go back -1

tigohenryschultz:
Route.replace is working great, tab pages ion-back-button and device back button working great, navigating back to tabs from non-tab page sets the proper children to render, last issue is undefined viewItem when using router.go from another ticket: https://github.com/ionic-team/ionic-framework/issues/24432

yoyo930021:
Hi, I found some bugs in `@ionic/vue` PR.
Sometimes route will be broken, but only `@ionic/vue-router` will well.

Another problem:
When pushing -> replacing -> pushing -> pushing -> pop backing -> pop backing -> pop backing,
The leavingRouteInfo will be not found. the leaving view isn't destroyed.
I fix it in this commit https://github.com/yoyo930021/ionic-framework/commit/1c10f71a26c49a129379b319f5447f83cdb2a2e0 .

yoyo930021:
When pushing -> replacing -> pushing -> pop backing -> replacing -> pop backing, the route will wrong.
I fix it in https://github.com/yoyo930021/ionic-framework/commit/6602fbb75f7488557ab4f9935607ab990e3dab1d

tigohenryschultz:
> When pushing -> replacing -> pushing -> pop backing -> replacing -> pop backing, the route will wrong. I fix it in [yoyo930021@6602fbb](https://github.com/yoyo930021/ionic-framework/commit/6602fbb75f7488557ab4f9935607ab990e3dab1d)

I manually added your changes here: https://github.com/ionic-team/ionic-framework/pull/24433/commits/d6c7ed26a1338f259ff22604728d2fb215dbae62

Did I not push the change in?

tigohenryschultz:
Digging into the cypress tests I noticed that clicking tabs creates an ion-page in another ion-router-outlet:

image

Which the tests will probably find the wrong 'tab2secondary'?

@sean-perkins does this seem odd?

tigohenryschultz:
@liamdebeasi Problems I am not sure how to solve, every IonRouterOutlet watches the route change and every IonRouterOutlet tries to handlePageTransition. 

When going from say http://localhost:8080/tabs/tab1 to http://localhost:8080/tabs-secondary/tab1 shouldn't all 3 ion-router-outlets handle their transitions? Only the 1st and 3rd outlet would have the tabs-secondary content and the 1st and 2nd would have the first tabs content but all 3 ion-router-outlets should have entering/leaving transitions?

http://localhost:8080/tabs/tab1 ->  http://localhost:8080/tabs-secondary/tab1
Outlet 1 transition away pageid=tabs
Outlet 2 transition away pageid=tab1

Outlet 1 transition into pageid=tabs-secondary
Outlet 3 transition into pageid=tab1-secondary

But the ion-router-outlet code doesn't handle this at all it only expects the root outlet to handle the transition. 

Question is: Should this be refactored and simplified? Each outlet is handling the route change/transition instead of one place. The ion-router-outlets should behave more as render outlets but one place should still control all transitions/watching the route change. Keep tracking of the root path fixes some of the issues but if you want real browser behavior like router.go(-10) (back works) and router.go(10)(forward is broken since tabs are hacked in some cases) then this will need to be refactored some.

tigohenryschultz:
What I think could work as a workaround is handling the 'setupViewItem call for each outlet and determining if the viewItem you find based on the path/outletId is entering/leaving and triggering transitions there. That way all parent/nested outlets all get their life cycles and transitions. I'm kind of doing this in my last push

tigohenryschultz:
The only change from old behavior is clicking the back button does not remove/unmount the viewItem. The only time that occurs is when 'replace' is fired. 

The reason for this is if you have a 'forward' and 'back' button like a browser you would expect to be able to go forward or back.

packages/vue-router/src/router.ts
```diff
@@ -62,6 +66,13 @@ export const createIonRouter = (opts: IonicVueRouterOptions, router: Router) =>
   const initialHistoryPosition = opts.history.state.position as number;
   let currentHistoryPosition = opts.history.state.position as number;
 
+  const currentHistoryIndex = () => {
+    let index = (currentHistoryPosition) - initialHistoryPosition;
+    console.log('Current index is: ', index);
```
sean-perkins, line 71: Let's clean-out any temporary testing logs, comments, commented out code, etc.

packages/vue-router/src/router.ts
```diff
@@ -62,6 +66,13 @@ export const createIonRouter = (opts: IonicVueRouterOptions, router: Router) =>
   const initialHistoryPosition = opts.history.state.position as number;
   let currentHistoryPosition = opts.history.state.position as number;
 
+  const currentHistoryIndex = () => {
```
sean-perkins, line 69: We should be able to inline this:
```
const currentHistoryIndex = () => currentHistoryPosition - initialHistoryPosition;
```

packages/vue-router/src/router.ts
```diff
@@ -212,59 +239,83 @@ export const createIonRouter = (opts: IonicVueRouterOptions, router: Router) =>
           }
         }
       }
+      /**
+       * If there is no route id, assign one, else use incoming route id for location history lookup
+       */
+      if (incomingRouteParams.hasOwnProperty('id') === false) {
```
sean-perkins, line 245: We tend to do undefined checks and object-based notation,.
```
if (typeof incomingRouteParams.id === 'undefined') {
  incomingRouteParams.id = generateId('routeInfo');
}
```

packages/vue-router/src/router.ts
```diff
@@ -212,59 +239,83 @@ export const createIonRouter = (opts: IonicVueRouterOptions, router: Router) =>
...
+        nextRouteInfo.routerAnimation = lastRouteInfo?.routerAnimation;
       }
 
-      routeInfo.position = currentHistoryPosition;
-      routeInfo.delta = delta;
+      //
+      nextRouteInfo.matchedPath = location.matched[0].path;
+      nextRouteInfo.position = currentHistoryPosition;
+      //why store the delta of this route info?
+      // nextRouteInfo.delta = delta;
```
sean-perkins, line 314: Delta is stored for the page transition logic (introduced in this PR: #24056). Liam can probably describe the exact implementation reasoning in more detail.

packages/vue/test-app/package.json
```diff
@@ -13,16 +14,17 @@
     "sync": "sh ./scripts/sync.sh"
   },
   "dependencies": {
-    "@ionic/vue": "5.6.3",
-    "@ionic/vue-router": "5.6.3",
+    "@ionic/vue": "5.9.3",
```
sean-perkins, line 17: We will want to rebase against main so these target v6.

sean-perkins (changes requested):
Was only able to review about 1/2 so far. Will circle back when I get another chunk of time.

packages/vue-router/src/router.ts
tigohenryschultz: The only thing I would say about this is debugging isn't possible. 

Short hand is fun and all for rapidly writing code but if you ever need to go back and debug, then you have to refactor it.

packages/vue-router/src/router.ts
tigohenryschultz: I restructured the IonRouterOutlet to be able to handle all transitions from all outlets. Since it is impossible from one outlet to find all the entering/leaving viewItems you have to let all outlets handle transitions. So I think I may have removed the requirement for this.

The use case was if a user went from page/a  -> tab/1 -> tab/2 -> page/b. Since the tabs were on a different outlet, page/b could not find the tab/2 to navigate away from, so life cycles never fired and it was never hidden, the outlet it was in was hidden.

packages/vue/test-app/package.json
tigohenryschultz: The test app always uses the scripts to run the tests so it will move whatever the parent vue/vue-router have. You want me to have the test-app reference 6.0.2?

liamdebeasi:
Hello @yoyo930021 and @tigohenryschultz, 

I have proposed an alternative fix for this issue and attached a dev build for testing in https://github.com/ionic-team/ionic-framework/issues/24432#issuecomment-1030203775.

I would really appreciate it if you could test the dev build in your apps and let me know if it resolves the issue that was reported in #24432.

If it does not resolve the issue, please send over a sample application I can use to reproduce the issue, and I will be more than happy to take another look to see if we can better address the issue. Thanks!

liamdebeasi:
Hi there,

Thank you for your PR! We appreciate all the hard work you and @yoyo930021 put into creating your respective solutions, as well as your eagerness to improve Ionic. I want to provide some updates regarding this PR as well as [#24432](https://github.com/ionic-team/ionic-framework/issues/24432), the issue this PR was created to address.

I have proposed an alternative solution in https://github.com/ionic-team/ionic-framework/pull/24721 which fixes the issue in #24432. The fix has been merged into the `main` branch and will be available in an upcoming release of Ionic. I gave both of you co-author credit in the commit because your efforts helped to inform the solution that was merged in. I have also provided a dev build for this fix in https://github.com/ionic-team/ionic-framework/issues/24432#issuecomment-1030203775. Please feel free to test it out.

As for this PR, we are unable to accept this pull request. There are a couple reasons for this:

1. The PR includes many different changes that effect different functions of the router. While this was likely the result of fixing one issue and approaching another, we prefer PRs that address a single issue at a time. PRs that address many issues, should have very limited affected changes.
2. The PR changes some fundamentals with how routing in Ionic works that we are not comfortable accepting.

The changes here appear to make Ionic routing behave more like how routing works in a web browser than in a native mobile app. Ionic routing is designed to behave like the latter, so the changes proposed here are not a good fit for Ionic. We are looking to provide an experience that behaves like a native mobile application, not a mobile web app.

I understand you may still have questions or changes you wish to see made in Ionic. I encourage you to open issues where each issue is dedicated to a particular concern you have in Ionic. From there, we can discuss and answer any questions you may have.

I am going to close this PR, but please feel free to open up separate issues for each concern you have. Thanks!

tigohenryschultz:
Thanks for working hard on these issues @liamdebeasi 

However, we'll still have to use our custom build until all other routing issues are resolved but appreciate the effort.

One of the fixes in my pull request that is still an issue in master is that all tabs now have a proper life cycle. To accomplish this every ion-router-outlet will need to be able to handle route changes, transition animations also had to refactored to accommodate that. I know previously you didn't want to solve that problem and suggested watching the route object, is this the same position you hold? If not I can create a separate ticket/test case and detail the issue.

Also I deprecated prevRoute_lastPathname as it was a hack used to remember which tab to hide when navigating back into a nested ion-router-outlet tab.

Could I submit a pull request with just additional test cases for the above scenarios?

liamdebeasi:
Hey there,

Sorry for the delay! I got a new computer but did not realize my email was not properly configured, so I never saw this comment (which explains why my inbox was empty all the time 😄 )

Just so I understand, you want the individual tab items to fire lifecycle events even when navigating to and from tabs? If so, I think that makes sense to do. Could you create a separate ticket as you mentioned so we can better understand the issue? Thanks!

liamdebeasi:
Also just as a heads up: We are planning to consolidate a lot of our routing code across Angular, React, and Vue to make the routing integration easier to maintain. Part of that consolidation may result in some of these improvements you are hoping to see too.

Issue #24449, opened 2021-12-20
bug: ion-popover doesnt have backdrop in tree-shaken build (react & vue)


JohnDeved:
### Prerequisites

- [X] I have read the [Contributing Guidelines](https://github.com/ionic-team/ionic-framework/blob/main/.github/CONTRIBUTING.md#creating-an-issue).
- [X] I agree to follow the [Code of Conduct](https://ionicframework.com/code-of-conduct).
- [X] I have searched for [existing issues](https://github.com/ionic-team/ionic-framework/issues) that already report this problem, without success.

### Ionic Framework Version

- [ ] v4.x
- [ ] v5.x
- [X] v6.x

### Current Behavior

in production, tree-shaken build, (`npm run build && npx serve -s build`) 
popover are not dismissable by clicking on the backdrop. (no backdrop seems to show up)
its also possible to scroll while the popup is open.
I noticed this behavior when upgrading from v5 to v6

https://user-images.githubusercontent.com/24187269/146756242-067557e2-17e2-4596-8375-f79e56709606.mov

hosted production build: https://8179360d.ionic-popover-bug.pages.dev/

### Expected Behavior

the backdrop should show up and it should be dismissable by clicking on it.
it also shouldn't be possible to scroll away while the popover is open.
interestingly enough everything seems to work fine in dev mode (`npm start`)

https://user-images.githubusercontent.com/24187269/146757472-e6def746-9153-4377-9e10-ba1b37485a84.mov

### Steps to Reproduce

The only step to reproduce is to use popover in any way with react ionic and build to production.
just clone reproduction repo and create and serve a production build to encounter the issues.

### Code Reproduction URL

https://github.com/JohnDeved/ionic-popover-bug

### Ionic Info

```
Ionic:

   Ionic CLI       : 5.4.16 (/Users/undefined/.npm/_npx/864a9e3c2cd0cf50/node_modules/ionic)
   Ionic Framework : @ionic/react 6.0.1

Utility:

   cordova-res : not installed
   native-run  : not installed

System:

   NodeJS : v16.13.1 (/Users/undefined/.nvm/versions/node/v16.13.1/bin/node)
   npm    : 8.1.2
   OS     : macOS Big Sur
```

sean-perkins:
I believe this is a result of nested custom elements not being defined in overlays. This has been resolved here #24439 and will be available in the next release of Ionic.

You can use the dev build to verify this resolves your issue:
```bash
[...]

PR #24453, opened 2021-12-20
fix(react): backdrop for inline modal/popover overlay

sean-perkins:
## Pull request checklist

## Pull request type

 

 

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

In the React target, inline overlays (`ion-popover` and `ion-modal`) will not define their nested children custom elements (`ion-backdrop`), when in a production tree-shaken build.

Issue Number: #24449

## What is the new behavior?

Inline overlays will share the same custom element definition logic as the regular overlays. This correctly defines `ion-backdrop` for popover/modal inline overlays.

- [ ] Yes
- [x] No

### DIFF
--- a/core/src/utils/overlays.ts
+++ b/core/src/utils/overlays.ts
@@ -35,9 +35,6 @@ const createController = <Opts extends object, HTMLElm extends any>(tagName: str
     },
     async getTop(): Promise<HTMLElm | undefined> {
       return getOverlay(document, tagName) as any;
-    },
-    defineCustomElements() {
-      return registerOverlayComponents(tagName, customElement, childrenCustomElements);
     }
   };
 };
--- a/core/src/utils/overlays.ts
+++ b/core/src/utils/overlays.ts
@@ -35,6 +35,9 @@ const createController = <Opts extends object, HTMLElm extends any>(tagName: str
     },
     async getTop(): Promise<HTMLElm | undefined> {
       return getOverlay(document, tagName) as any;
+    },
+    defineCustomElements() {
+      return registerOverlayComponents(tagName, customElement, childrenCustomElements);
     }
   };
 };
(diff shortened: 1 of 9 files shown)


sean-perkins:
I'd love to wait until after the New Year to discuss this with the team 🙂 . I could see use deciding not to expose a new public method on the controller API and instead exporting pure constants for just the popover & modal custom element definitions.

packages/react/src/components/IonPopover.tsx
```diff
@@ -1,9 +1,8 @@
-import { JSX } from '@ionic/core/components';
-import { IonPopover as IonPopoverCmp } from '@ionic/core/components/ion-popover.js';
+import { JSX, popoverController } from '@ionic/core/components';
 
 import { createInlineOverlayComponent } from './createInlineOverlayComponent'
 
 export const IonPopover = /*@__PURE__*/ createInlineOverlayComponent<
   JSX.IonPopover,
   HTMLIonPopoverElement
->('ion-popover', IonPopoverCmp);
```
liamdebeasi, line 9: If you import `defineCustomElements` instead and call that function, it should define both the overlay and the backdrop. See https://github.com/ionic-team/ionic-framework/blob/main/packages/vue/src/components/IonApp.ts#L3-L7 as an example

liamdebeasi (changes requested):

sean-perkins:
Created a dev build of the latest to test with: `6.0.1-dev.1641412400.d3fd57e` & re-verified everything works as expected with the reproduction app: https://github.com/JohnDeved/ionic-popover-bug.

core/src/utils/overlays.ts
```diff
@@ -59,27 +59,33 @@ export const prepareOverlay = <T extends HTMLIonOverlayElement>(el: T) => {
...
-      /**
-       * If the parent element has nested usage of custom elements,
-       * we need to manually define those custom elements.
-       */
-      if (childrenCustomElements) {
-        for (const customElementDefinition of childrenCustomElements) {
-          if (!window.customElements.get(customElementDefinition.tagName)) {
-            window.customElements.define(customElementDefinition.tagName, customElementDefinition.customElement);
-          }
+  if (typeof window !== 'undefined' && typeof window.customElements !== 'undefined') {
```
liamdebeasi, line 64: This line is redundant. If the window SSR check in `createOverlay` fails, then this code won't get run.

liamdebeasi (changes requested):
The original issue says it impacts Vue as well, so we should make this change in https://github.com/ionic-team/ionic-framework/blob/main/packages/vue/src/components/Overlays.ts#L12-L18 as well

Also, my understanding of the issue is this impacts all overlay components, not just modal and popover. If that is the case, components like action sheet also need to be updated: https://github.com/ionic-team/ionic-framework/blob/main/packages/react/src/components/IonActionSheet.tsx#L6

sean-perkins:
Alright this should look better now. Didn't realize the overlays generation script was local to this repo. Simplified not needing to update the output targets. 

Kept the commits to follow the changes, but will squash and rework to better explain it affects react & vue targets.

liamdebeasi (approved):
Great job!

Issue #2150, opened 2014-09-06
Need to double click options in a select tag on WP


rbarriuso:
On Windows Phone (which is not currently supported by Ionic), the `select` tag with nested `options` opens a selection menu which appears twice, that is, the user needs to select an option, then the menu appears again and the user is forced to select an option again.
The following project based on the tabs starter template reproduces the issue on Windows Phone 8.1 and 8.0 (at least): http://we.tl/ktrI2mjrMk

Probably it's an Ionic issue because I couldn't reproduce it only using Cordova or only using AngularJS.

Ionitron:
Greetings, @rbarriuso!

My sensors indicate that this issue has been inactive for a while.
Due to this, the issue will be closed in 7 days.

If you believe the issue has not been fixed, please add a comment.
This will notify us so we can assist you, and will prevent the issue from being automatically closed.
Feel free to include the following information about the issue, so that we may assist you:

- Can you provide a detailed description of this issue?
- How would a robot go about replicating this issue?
- What platform does this happen on? (iOS, Android, all, etc.)
- What platform versions does this happen on? (Android 4.2, iOS 7, all, etc.)
- Is this an issue with the Ionic Framework, or Ionic CLI?

Thank you for allowing me to assist you.

lorenzyannick:
It's not fixed, just hacked !

Because I've many select under select list (profile page)

SELECT
SELECT
SELECT
INPUT
INPUT
SELECT
SELECT
SELECT

the tap-disabled works but it disabled the scroll also, so I've to start touchscroll outside select element !
So I have to start touch event over INPUT
Make the scroll behaviour a real pain :)

Make sense ?

lorenzyannick:
![wp_ss_20150121_0006](https://cloud.githubusercontent.com/assets/9976897/5842181/245976ea-a1a3-11e4-87d3-f588422506ea.png)

kashban:
It's still there. While the select box works ok when using the mouse on both IE 11 and Windows 8.1 Tablets, it does not when only being touched. Only the text of the current element is selected, just like a text input field, but the drop down list isn't opened. 

Pretty much like @lorenzyannick illustrated.

When tapping and holding on the select box and sliding sideways the menu comes up. 

I am going to try to debug this on a touch device once I have it connected to my developer PC.

VeldMuijz:
I've tried using this:
```
if (ionic.Platform.isWindowsPhone()) {
[...]

PR #24498, opened 2021-12-29
docs(ion-tabs): move slots property from ion-tabs to ion-tab-bar

dtarnawsky:
## Pull request checklist

## Pull request type

 

 

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes

The ion-tab page has a slot property documented but it doesn't affect the ion-tab component.
Issue Number: 2150

## What is the new behavior?

The ion-tab-bar page has a slot property documented (with properties for top and bottom).

- [ ] Yes
- [x] No

### DIFF
--- a/core/src/components/tab-bar/readme.md
+++ b/core/src/components/tab-bar/readme.md
@@ -164,6 +164,15 @@ export default defineComponent({
 | `translucent` | `translucent`  | If `true`, the tab bar will be translucent. Only applies when the mode is `"ios"` and the device supports [`backdrop-filter`](https://developer.mozilla.org/en-US/docs/Web/CSS/backdrop-filter#Browser_compatibility).                                                 | `boolean`             | `false`     |
 
 
+## Slots
+
+| Slot       | Description                                                           |
+| ---------- | --------------------------------------------------------------------- |
+|            | Content is placed between the named slots if provided without a slot. |
+| `"bottom"` | Content is placed at the bottom of the screen.                        |
+| `"top"`    | Content is placed at the top of the screen.                           |
+
+
 ## CSS Custom Properties
 
 | Name           | Description               |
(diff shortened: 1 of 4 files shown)


liamdebeasi (commented):
Thanks for the PR! What issue is "2150"? Issue #2150 on GitHub does not seem to be related to tabs.

Additionally, IonTabs does have top and bottom slots: https://github.com/ionic-team/ionic-framework/blob/main/core/src/components/tabs/tabs.tsx#L190-L195

IonTabBar does not have the top and bottom slots: https://github.com/ionic-team/ionic-framework/blob/main/core/src/components/tab-bar/tab-bar.tsx#L100

While developers do write `slot="bottom"` on `ion-tab-bar`, the slot API itself exists on IonTabs. `slot="bottom"` defines how IonTabBar should be displayed within IonTabs.

dtarnawsky:
Hey @liamdebeasi , for docs issues they are written up in the ionic-docs repo so issue 2150 is found at: [https://github.com/ionic-team/ionic-docs/issues/2150](https://github.com/ionic-team/ionic-docs/issues/2150). I'm not sure why we do this but that's a different story.

The reason for the PR is that as a developer you write the slot property on ion-tab-bar (not on ion-tabs). The underlying implementation doesn't really matter when you are trying to write code based on the docs. This is where new developers are getting confused because what is written in the docs doesn't match the code that they write and the only way to get it right is to find sample code that works and copy that (it's tribal rather than learned knowledge).

So yeah ion-tab-bar doesn't have "slot" in its implementation but you need to specify it to affect the UI and ion-tabs does have a "slot" in its implementation but you can't use it because its not hooked up to ion-tabs.

liamdebeasi:
I think the issue here is a misunderstanding of how slots work. The slots exist inside of `ion-tabs` and accept any arbitrary content, though the most common usage is to add `ion-tab-bar`. I think it would be confusing to state that only `ion-tab-bar` uses `slot="top"` or `slot="bottom"` when anything can go inside of these slots.

Maybe this PR could be updated to add some information on how to position `ion-tab-bar`? This documentation could go inside of the readme for `ion-tab-bar`.

PR #23341, opened 2021-05-21
feat(modal): modals can now be used inline

liamdebeasi:
## Pull request checklist

Please check if your PR fulfills the following requirements:
- [x] Tests for the changes have been added (for bug fixes / features)
- [x] Docs have been reviewed and added / updated if needed (for bug fixes / features)
- [x] Build (`npm run build`) was run locally and any changes were pushed
- [x] Lint (`npm run lint`) has passed locally and any fixes were made for failures

## Pull request type

 

 

Please check the type of change your PR introduces:
- [ ] Bugfix
- [x] Feature
- [ ] Code style update (formatting, renaming)
- [ ] Refactoring (no functional changes, no api changes)
- [ ] Build related changes
- [ ] Documentation content changes
- [ ] Other (please describe): 

## What is the current behavior?

Issue Number: resolves https://github.com/ionic-team/ionic-framework/issues/20117 resolves https://github.com/ionic-team/ionic-framework/issues/20263

This rework is also import for our upcoming `ion-datetime` component as it will allow developers to easily present datetime in an overlay:

```html
<ion-button id="openDatetime">Open Datetime</ion-button>

<ion-modal trigger="openDatetime">
  <ion-datetime></ion-datetime>
</ion-modal>
```

## What is the new behavior?

- Modals can now be written inline, allowing the instances to be reused as well.

## Does this introduce a breaking change?

- [x] Yes
- [ ] No

PR #24521, opened 2022-01-06
fix(angular): inline overlay dynamic template data

### LINKED ISSUE #24502: bug: Angular Inline Sheet Modal are not updated
### Prerequisites

### Ionic Framework Version

- [ ] v4.x
- [ ] v5.x
- [X] v6.x

### Current Behavior

CommonModule Angular directives for example ngFor or ngIf inside Inline Sheet Modal are not updated. I think because of ng-template. For example, you can't do a search and update the list.

### Expected Behavior

Items should be updated.

### Steps to Reproduce

Create an Inline Sheet Modal with a list with items and a ngFor, update the array.

### Code Reproduction URL

https://github.com/Marius-Romanus/ionic-bug-modals-24502
https://stackblitz.com/edit/ionic-angular-v5-g9hhgc?file=src%2Fapp%2Fapp.component.ts

### Ionic Info

Ionic:

   Ionic CLI : 6.18.1

Utility:

   native-run  : 1.5.0

System:

   NodeJS : v17.3.0
   npm    : 8.3.0
   OS     : Windows 10

### Additional Information

_No response_

sean-perkins:
## Pull request checklist

## Pull request type

 

 

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

Angular inline modals and popovers do not update their template contents when change detection occurs. This results in structural directives (`*ngIf`, `*ngFor`), pipes and bindings to not update after the component is initialized. 

Issue Number: Resolves #24502

## What is the new behavior?

Inline modals and popovers will correctly re-render their template contents when change detection occurs.

- [ ] Yes
- [x] No

## Other information

|Before|After|
|----|----|
|<video src="https://user-images.githubusercontent.com/13732623/148325100-59b5d45b-bb3a-436c-8f43-b5e998bb0cc3.mp4"></video>|<video src="https://user-images.githubusercontent.com/13732623/148324849-4457dd97-5e88-4c78-bf45-2104937e6498.mp4"></video>|


angular/src/directives/overlays/modal.ts
```diff
@@ -103,7 +103,6 @@ export class IonModal {
   protected el: HTMLElement;
 
   constructor(c: ChangeDetectorRef, r: ElementRef, protected z: NgZone) {
-    c.detach();
```
liamdebeasi, line 106: Does this need to be done for inline popovers as well? https://github.com/ionic-team/ionic-framework/blob/main/angular/src/directives/overlays/popover.ts#L107

Additionally, how does this impact change detection with a controller (if it impacts it at all)?

sean-perkins: Good catch, this did affect popovers as well. I've updated the PR to resolve and test both. I've also reworked the commit since this is more about inline overlays and less specific to just modals.

Tested with controllers & attached change detection and things are working as expected. I cannot notice an impact.

amandaesmith3: I was wondering why this call was added in the first place, and traced it back to the inline modal feature PR: https://github.com/ionic-team/ionic-framework/pull/23341/files#diff-d123478d8d877b48ad22b0a74610bcd649b94548e5e5c56a173bb3506d506e88R25 @liamdebeasi Can you think of why this was added originally? I'm wondering if there was some edge case bug this was meant to fix 🤔

liamdebeasi: `c.detach()` is what all the other components do, which is why I added it https://github.com/ionic-team/ionic-framework/blob/main/angular/src/directives/proxies.ts#L26

amandaesmith3: Huh... do we have our own change detection set up or something? 🤨

sean-perkins: How we "wrap" our web components into Angular is sorta special. We basically target the same selector tag as our web component, but then use Angular's content projection with `ng-content` to slot the existing web component inside the wrapping Angular component; since ng-content isn't an actual DOM node this results in `ion-modal` equalling our web component node. 

We then typically detach from all of Angular's change detection, because the updating of the web component should be driven through our implementation in Stencil and independent to anything in Angular. We don't want to cause an unwanted repaint of the web component. We only pass through inputs from our proxy to the web component node, so that the web component can decide when a re-render is necessary. 

Overlays are a special condition (inline most specifically), because we are reliant on using `ng-template` since we are using template outlets (Angular feature). In Ivy, they fixed `ng-template` to have its own change detection reference to render changes inside just that template context. When we "detach" we are opting out of that behavior. With controller implementations the custom Angular component the implementer creates; would drive the change detection of the template (which wouldn't be detached by default).

For all our other UI components, opting out makes sense. We aren't needing the benefits of functionality that Angular brings with what the web component needs.

liamdebeasi (approved):
Nice job! Also make sure commit subject follows our guidelines too

angular/src/directives/overlays/modal.ts
amandaesmith3: So _that's_ what those are for here! Thanks for the in-depth explanation 👍 If controller modals/popovers still work as expected (i.e. no extra repaints) then this sounds good to me.

amandaesmith3 (approved):

angular/src/directives/overlays/modal.ts
sean-perkins: Adding a notion doc for now in our internal documents.

PR #24533, opened 2022-01-07
fix(vue): correct route is replaced when using router.replace

liamdebeasi:
## Pull request checklist

## Pull request type

 

 

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

Issue Number: resolves https://github.com/ionic-team/ionic-framework/issues/24226

Previously I had thought that issue was related to ion-router-outlet not waiting for the previous transition to finish. This finding was incorrect. The real issue was that the replace logic inside of `IonRouter` was replacing the wrong route.

Currently, IonRouter gets the last item in the Location History array for the route that should be replaced: https://github.com/ionic-team/ionic-framework/blob/77f8412b746222793cd9d17f12f50d512ab5e886/packages/vue-router/src/router.ts#L241

This line (incorrectly) assumes that the last item in the Location History is the current route. Consider the following example:

History: `/page1` --> `/page2`.

If you do `router.go(-1)`, you will now be on `/page1`. If you were to then do `router.replace('/page3'`) one would expect Ionic to replace `/page1` with `/page3`. However, since it was grabbing the last item in the Location History, Ionic was replacing `/page2` with `/page3`.

## What is the new behavior?

- Router now gets the current route according to calculated deltas rather than grabbing the last item in the array.

- [ ] Yes
- [x] No

### DIFF
--- a/packages/vue-router/src/router.ts
+++ b/packages/vue-router/src/router.ts
@@ -238,7 +238,19 @@ export const createIonRouter = (opts: IonicVueRouterOptions, router: Router) => 
           const lastRoute = locationHistory.getCurrentRouteInfoForTab(routeInfo.tab);
           routeInfo.pushedByRoute = lastRoute?.pushedByRoute;
         } else if (routeInfo.routerAction === 'replace') {
-          const currentRouteInfo = locationHistory.last();
+
+          /**
+           * When replacing a route, we want to make sure we select the current route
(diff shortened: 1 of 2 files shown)


sean-perkins (approved):
Looks good to me 👍

amandaesmith3 (approved):

Issue #24545, opened 2022-01-10
bug: ModalController.getTop() method returns modals that are closed


carstenbaumhoegger:
### Prerequisites

- [X] I have read the [Contributing Guidelines](https://github.com/ionic-team/ionic-framework/blob/main/.github/CONTRIBUTING.md#creating-an-issue).
- [X] I agree to follow the [Code of Conduct](https://ionicframework.com/code-of-conduct).
- [X] I have searched for [existing issues](https://github.com/ionic-team/ionic-framework/issues) that already report this problem, without success.

### Ionic Framework Version

- [ ] v4.x
- [ ] v5.x
- [X] v6.x

### Current Behavior

I have an Ionic app that displays Datetime-Pickers in Modals.

With the switch to Ionic 6 I changed my code from this solution:
```
<ion-datetime (ionChange)="dateTimeChange($event)"
             displayFormat="DD.MM.YYYY"
             [value]="startAt"
             [min]="minStartAt"
             [max]="maxStartAt">
</ion-datetime>
```

to this one (as explained in the documentation [here](https://ionicframework.com/docs/api/datetime#usage)):
```
<ion-button fill="clear" id="open-start-date-input">
  <ion-icon icon="calendar"></ion-icon>
  <ion-input>&nbsp;{{ startAt | date }}</ion-input>
</ion-button>
<ion-modal class="datetime" trigger="open-start-date-input" show-backdrop="false">
  <ng-template>
    <ion-datetime (ionChange)="dateTimeChange($event)"
                  presentation="date"
                  [min]="minStartAt"
                  [max]="maxStartAt"
                  [value]="startAt"
                  [showDefaultButtons]="true">
    </ion-datetime>
  </ng-template>
</ion-modal>
```

Now when I want to close my "main" modal by calling `(await this.modalController.getTop()).dismiss()` the already closed inner Datetime Modal get's returned by `this.modalController.getTop()` so the "main" modal won't close.

### Expected Behavior

`ModalController.getTop()` only returns modals that are currently displayed.
Modals, that are not shown aren't returned.

### Steps to Reproduce

Migrating Datetimes in Modals from the Ionic 4/5 approach to Ionic 6 and try to dismiss a Modal via the `ModalController.getTop()` approach

### Code Reproduction URL

- https://github.com/sean-perkins/gh-24545-modal-top

### Ionic Info

Ionic:

[...]

Issue #19111, opened 2019-08-16
bug: controller getTop() method returns overlays that are not presented


damirarh:
# Bug Report

**Ionic version:**

[x] **4.7.4**

**Current behavior:**
The call to `LoadingController.getTop()` returns the most recently created loading overlay even if it hasn't been presented/opened yet.

**Expected behavior:**
According to the [docs](https://ionicframework.com/docs/api/loading-controller#methods),, the call to `LoadingController.getTop()` should only return a loading overlay if it has already been presented/opened.

**Steps to reproduce:**
1. Create a loading overlay and present it.
2. Create another loading overlay, but don't present it.
3. `LoadingController.getTop()` returns the second overlay instead of the first one.
4. `LoadingController.dismiss()` doesn't dismiss the open loading overlay and returns false.

**Related code:**

https://github.com/damirarh/ioni4-loading-ctrl

```typescript
this.loadingCtrl.create({message: 'presented'}).then(loader => {
  loader.present().then(() => {
    this.loadingCtrl.create({message: 'created'}).then(() => {
      this.loadingCtrl.getTop().then(topLoader => {
        console.log(topLoader.message); // outputs 'created' instead of 'presented'
        this.loadingCtrl.dismiss().then(dismissed => {
          console.log(dismissed); // doesn't dismiss the open loading overlay and outputs 'false'
        });
      });
    });
  });
});
```

**Other information:**
This behavior also makes it impossible to reliably close all opened loading overlays without having a reference to each opened instance (i.e. reimplement the behavior of `dismissAll` in Ionic 3) because there's no way to get to the previously opened loading overlay using the `LoadingController` if an opened loading overlay was created afterward. I [blogged](http://www.damirscorner.com/blog/posts/20190816-DismissingAllLoadingOverlaysInIonic4.html) about my attempt at implementing `dismissAll` in Ionic 4.

**Ionic info:**

```
Ionic:

   Ionic CLI                     : 5.2.2 (C:\Users\damir\AppData\Roaming\npm\node_modules\ionic)
   Ionic Framework               : @ionic/angular 4.7.4
   @angular-devkit/build-angular : 0.801.3
   @angular-devkit/schematics    : 8.1.3
   @angular/cli                  : 8.1.3
   @ionic/angular-toolkit        : 2.0.0

Utility:

   cordova-res : not installed
   native-run  : 0.2.7

System:

   NodeJS : v10.16.0 (C:\Program Files\nodejs\node.exe)
[...]

PR #24547, opened 2022-01-10
fix(overlays): getOverlay return visible overlay

sean-perkins:
## Pull request checklist

## Pull request type

 

 

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

The logic for `getTop` does not take into consideration hidden overlays. With usages of an inline modal within a modal instance using a controller; you are unable to dismiss the modal using `dismiss` or `getTop` without an id.

Issue Number: #19111, #24545

## What is the new behavior?

Only visible overlays are returned in `getOverlay`, used by the `getTop` logic.

- [ ] Yes
- [x] No

## Other information

Reproduction app: https://github.com/sean-perkins/gh-24545-modal-top

- Install dev build: 
```bash
npm i @ionic/angular@6.0.1-dev.1641853396.482dba5
npm i @ionic/core@6.0.1-dev.1641853396.482dba5
```
- Run app
- Open modal with the button
- Dismiss modal with the button

### DIFF
--- a/core/src/utils/overlays.ts
+++ b/core/src/utils/overlays.ts
@@ -120,6 +120,8 @@ export const focusFirstDescendant = (ref: Element, overlay: HTMLIonOverlayElemen
   }
 };
 
+const isOverlayHidden = (overlay: Element) => overlay.classList.contains('overlay-hidden');
+
 const focusLastDescendant = (ref: Element, overlay: HTMLIonOverlayElement) => {
   const inputs = Array.from(ref.querySelectorAll(focusableQueryString)) as HTMLElement[];
   let lastInput = inputs.length > 0 ? inputs[inputs.length - 1] : null;
@@ -291,7 +293,7 @@ export const connectListeners = (doc: Document) => {
 
     // handle back-button click
     doc.addEventListener('ionBackButton', ev => {
-      const lastOverlay = getTopOpenOverlay(doc);
+      const lastOverlay = getOverlay(doc);
       if (lastOverlay && lastOverlay.backdropDismiss) {
         (ev as BackButtonEvent).detail.register(OVERLAY_BACK_BUTTON_PRIORITY, () => {
           return lastOverlay.dismiss(undefined, BACKDROP);
@@ -302,7 +304,7 @@ export const connectListeners = (doc: Document) => {
     // handle ESC to close overlay
     doc.addEventListener('keyup', ev => {
       if (ev.key === 'Escape') {
-        const lastOverlay = getTopOpenOverlay(doc);
+        const lastOverlay = getOverlay(doc);
         if (lastOverlay && lastOverlay.backdropDismiss) {
           lastOverlay.dismiss(undefined, BACKDROP);
         }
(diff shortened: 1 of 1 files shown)


core/src/utils/overlays.ts
```diff
@@ -327,31 +329,8 @@ export const getOverlays = (doc: Document, selector?: string): HTMLIonOverlayEle
...
-
-  return;
-}
-
 export const getOverlay = (doc: Document, overlayTag?: string, id?: string): HTMLIonOverlayElement | undefined => {
-  const overlays = getOverlays(doc, overlayTag);
+  const overlays = getOverlays(doc, overlayTag).filter(o => !isOverlayHidden(o));
   return (id === undefined)
     ? overlays[overlays.length - 1]
     : overlays.find(o => o.id === id);
```
amandaesmith3, line 336: If you try to find an overlay by ID and it's closed, you'll get `undefined`, which feels weird 🤔 Maybe only the `overlays[overlays.length - 1]` should be filtered to non-hidden overlays?

sean-perkins: That's really good feedback. Implementers may have a use case of getting an overlay by id before it's presented. I made that change and also added an e2e test.

angular/test/test-app/e2e/src/overlay-nested.spec.ts
```diff
@@ -0,0 +1,17 @@
+describe('Overlay: Nested', () => {
+  beforeEach(() => {
+    cy.visit('/overlay-nested');
+  });
+
+  it('should dismiss the top overlay', () => {
```
liamdebeasi, line 6: Is there a reason why this test has to be an Angular-specific test? I feel like we could write this in the core E2E test given that it's an @ionic/core API. There are some existing overlay E2E tests if you are interested: https://github.com/ionic-team/ionic-framework/tree/223f36f6adacf8adce47cee4809a60c94a9e0efa/core/src/utils/test/overlays

core/src/utils/overlays.ts
```diff
@@ -327,34 +329,13 @@ export const getOverlays = (doc: Document, selector?: string): HTMLIonOverlayEle
...
 export const getOverlay = (doc: Document, overlayTag?: string, id?: string): HTMLIonOverlayElement | undefined => {
   const overlays = getOverlays(doc, overlayTag);
-  return (id === undefined)
-    ? overlays[overlays.length - 1]
-    : overlays.find(o => o.id === id);
+  if (id === undefined) {
+    const visibleOverlays = overlays.filter(o => !isOverlayHidden(o));
+    return visibleOverlays[visibleOverlays.length - 1];
+  }
+  return overlays.find(o => o.id === id);
```
liamdebeasi, line 338: This behavior seems confusing to me. If there is no ID, then this will exclude hidden overlays. However, if there is an ID, then this will include hidden overlays. Having the include/exclude behavior determined by the ID does not seem right.

Perhaps there should be a parameter that lets you indicate whether or not you want to include all overlays regardless of hidden state? While I do consider this to be a bug fix as the original behavior seems wrong, the new behavior could be considered breaking by some, so adding this may be a good compromise.

liamdebeasi (changes requested):
This is off to a great start 🚀 ! The angular test comment is not blocking, though it seems like we could simplify that by making it a test in core rather than angular.

core/src/utils/overlays.ts
sean-perkins: Added a parameter to either include or exclude hidden overlays. Default behavior is to include hidden overlays to be a non-breaking change. Let me know if that aligns with how you were thinking. 

Would love to consider moving to an object-based parameter for configuring the `getOverlay` method instead of individual parameters for each option in v7.

angular/test/test-app/e2e/src/overlay-nested.spec.ts
sean-perkins: Swapped it out with the Core test, good suggestion 👍

core/src/utils/overlays.ts
```diff
@@ -328,33 +330,25 @@ export const getOverlays = (doc: Document, selector?: string): HTMLIonOverlayEle
...
+  if (includeHidden) {
+    return (id === undefined)
+      ? overlays[overlays.length - 1]
+      : overlays.find(o => o.id === id);
+  } else {
+    const visibleOverlays = overlays.filter(o => !isOverlayHidden(o));
+    return (id === undefined)
+      ? visibleOverlays[visibleOverlays.length - 1]
+      : visibleOverlays.find(o => o.id === id);
+  }
```
amandaesmith3, line 351: This could be simplified a bit by having a `filteredOverlays` reference that either gets filtered or not depending on `includeHidden`, then using that for either case. That way you don't need to duplicate the rest of the logic.

amandaesmith3 (approved):

liamdebeasi (approved):
Great job! One idea for the commit subject could be `fix(overlays): getTop now returns the top-most presented overlay` or something like that. `getOverlay return visible overlay` likely does not mean much to a dev since `getOverlay` is internal.

Issue #24405, opened 2021-12-15
bug: The year / month selected in the year / month picker will be reflected in the calendar with a one-month shift.


tatsuyasenbom:
### Prerequisites

- [X] I have read the [Contributing Guidelines](https://github.com/ionic-team/ionic-framework/blob/main/.github/CONTRIBUTING.md#creating-an-issue).
- [X] I agree to follow the [Code of Conduct](https://ionicframework.com/code-of-conduct).
- [X] I have searched for [existing issues](https://github.com/ionic-team/ionic-framework/issues) that already report this problem, without success.

### Ionic Framework Version

- [ ] v4.x
- [ ] v5.x
- [X] v6.x

### Current Behavior

# Safari on iPhone iOS 15.1
* In Safari on iOS, after the ion-datetime embedded in ion-popover is displayed, the year / month selected in the year / month picker will be reflected in the calendar with a one-month shift.

https://user-images.githubusercontent.com/87968771/146101658-5c7cff37-9617-4e5d-b32e-ccb31797beb1.mp4

# Google Chrome on Android 12 or Windows10
* However, it will be reflected correctly on Google Chrome on Android or Windows.

https://user-images.githubusercontent.com/87968771/146101870-ee8c0f4d-673e-485f-b51c-56f27592abf3.mp4

### Expected Behavior

* I want you to improve the behavior of Safari on iOS.

### Steps to Reproduce

1. Run the code below to build.
    ```ts
    <ion-label>{{ report_date | date:"yyyy年MM月dd日 (EEE)" }}</ion-label>
    <ion-popover [isOpen]="isPopover" arrow="false" style="--offset-y: -35%;">
      <ng-template>
        <ion-datetime (ionChange)="onChangeDate($event)" showDefaultTitle="true" presentation="date">
          <div slot="title">報告日</div>
        </ion-datetime>
      </ng-template>
    </ion-popover>
    ```
2. I did the same operation as the video I uploaded to Current Behavior.

### Code Reproduction URL

* [add ion-datetime on ion-popover](https://github.com/tatsuyasenbom/ionic-angular-sample/commit/572165a25ec97044eb2a2a2920143664f23fa1b1)

### Ionic Info

```powershell
Ionic:

   Ionic CLI                     : 6.18.1 (C:\Users\Administrator\AppData\Roaming\npm\node_modules\@ionic\cli)
   Ionic Framework               : @ionic/angular 6.0.1
[...]

PR #24554, opened 2022-01-11
fix(datetime): showing calendar grid no longer causes month to switch on ios 15

liamdebeasi:
## Pull request checklist

## Pull request type

 

 

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

Issue Number: resolves https://github.com/ionic-team/ionic-framework/issues/24405

This fix is a monkey patch and should be removed when iOS 14 support is dropped in the future. There was a bug in iOS 14 where the month would move back by 1 month when switching from the month/year picker to the calendar grid. A fix was created in https://github.com/ionic-team/ionic-framework/pull/24142. This fix inadvertently introduced the same issue, but only on iOS 15 and with the month moving forward by 1.

I investigated other ways to fix the iOS 14 issue, but nothing was sufficient so I opted for this hacky (and temporary!) approach. The code will apply the iOS 14 fix if it detects a WebKit browser that does not support `aspect-ratio`. `aspect-ratio` was added in Safari 15: https://webkit.org/blog/11989/new-webkit-features-in-safari-15/

The code will apply the normal `display: none` behavior if it detects either a non-WebKit browser or a WebKit browser that supports aspect-ratio (I.e. Safari 15+).

## What is the new behavior?

- The patch that fixes a bug in iOS 14 is now only applied to iOS 14, not all browsers.

- [ ] Yes
- [x] No

## Other information

Dev build: 6.0.2-dev.1641931632.6a7e2d1

### DIFF
--- a/core/src/components/datetime/datetime.scss
+++ b/core/src/components/datetime/datetime.scss
@@ -79,24 +79,49 @@
  * unhiding the calendar content.
  * To workaround this, we set the opacity
  * of the content to 0 and hide it offscreen.
- * TODO: This is fixed in Safari 15+, so remove
- * when Safari 14 support is dropped.
+ *
+ * -webkit-named-image is something only WebKit supports
+ * so we use this to detect general WebKit support.
+ * aspect-ratio is only supported in Safari 15+
+ * so by checking lack of aspect-ratio support, we know
+ * that we are in a pre-Safari 15 browser.
+ *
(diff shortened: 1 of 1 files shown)


sean-perkins (approved):
Looks good on Safari 15, nice find with the aspect-ratio check.

amandaesmith3 (approved):

Issue #24563, opened 2022-01-12
bug: Karma incorrectly reports [main] attribute deprecation


davidgeary:
### Prerequisites

- [X] I have read the [Contributing Guidelines](https://github.com/ionic-team/ionic-framework/blob/main/.github/CONTRIBUTING.md#creating-an-issue).
- [X] I agree to follow the [Code of Conduct](https://ionicframework.com/code-of-conduct).
- [X] I have searched for [existing issues](https://github.com/ionic-team/ionic-framework/issues) that already report this problem, without success.

### Ionic Framework Version

- [ ] v4.x
- [ ] v5.x
- [X] v6.x

### Current Behavior

When running the tests on an Ionic 6 application containing an `ion-menu` component, the following warning is sometimes issued, even though the related template does not use a `main` attribute:

``` sh
WARN: '[DEPRECATED][ion-menu] Using the [main] attribute is deprecated, please use the "contentId" property instead:
BEFORE:
  <ion-menu>...</ion-menu>
  <div main>...</div>

AFTER:
  <ion-menu contentId="main-content"></ion-menu>
  <div id="main-content">...</div>
'
```

Sometimes, the tests will also issue the error/warning:

``` sh
ERROR: 'Menu: must have a "content" element to listen for drag events on.'

WARN: 'split pane does not have a specified main node'
```

### Expected Behavior

No warning should be issued.

### Steps to Reproduce

Create a new Ionic starter project containing a menu and then just run the tests, e.g.

``` sh
ionic start foo sidemenu --type=angular --no-git
cd .\foo
ng test
```

### Ionic Info

Ionic:

   Ionic CLI                     : 6.18.1 (C:\Users\David\AppData\Roaming\nvm\v16.13.0\node_modules\@ionic\cli)
   Ionic Framework               : @ionic/angular 6.0.2
   @angular-devkit/build-angular : 13.1.2
   @angular-devkit/schematics    : 12.0.1
   @angular/cli                  : 13.1.2
   @ionic/angular-toolkit        : 5.0.3

Capacitor:

   Capacitor CLI      : 3.3.4
   @capacitor/android : 3.3.4
   @capacitor/core    : 3.3.4
   @capacitor/ios     : 3.3.4

Utility:

   cordova-res : not installed globally
   native-run  : 1.5.0

System:

   NodeJS : v16.13.0 (C:\Program Files\nodejs\node.exe)
   npm    : 8.1.0
   OS     : Windows 10

### Additional Information

[...]

PR #24565, opened 2022-01-12
fix(menu): remove main attribute that was supposed to removed in v5

liamdebeasi:
## Pull request checklist

## Pull request type

 

 

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

Issue Number: resolves https://github.com/ionic-team/ionic-framework/issues/24563
The main attribute was removed in Ionic 5 but did not actually get removed in Ionic 5 despite being noted in the breaking changes: https://github.com/ionic-team/ionic-framework/blob/main/BREAKING.md#menu

## What is the new behavior?

-  Removed the menu attribute support for real this time. 
- I also updated some of the checks below since the linter pointed out that they were not allowed.

- [ ] Yes
- [x] No (Technically a change in behavior, but support for this was officially dropped in Ionic 5, so developers should not have been using it anyway)

### DIFF
--- a/core/src/components/menu/menu.tsx
+++ b/core/src/components/menu/menu.tsx
@@ -172,20 +172,9 @@ export class Menu implements ComponentInterface, MenuI {
     const el = this.el;
     const parent = el.parentNode as any;
 
-    if (this.contentId === undefined) {
-      console.warn(`[DEPRECATED][ion-menu] Using the [main] attribute is deprecated, please use the "contentId" property instead:
-BEFORE:
-  <ion-menu>...</ion-menu>
-  <div main>...</div>
-
-AFTER:
-  <ion-menu contentId="main-content"></ion-menu>
-  <div id="main-content">...</div>
-`);
-    }
     const content = this.contentId !== undefined
       ? document.getElementById(this.contentId)
-      : parent && parent.querySelector && parent.querySelector('[main]');
+      : undefined;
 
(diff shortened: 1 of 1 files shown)


core/src/components/menu/menu.tsx
```diff
@@ -169,26 +169,11 @@ export class Menu implements ComponentInterface, MenuI {
...
-
-AFTER:
-  <ion-menu contentId="main-content"></ion-menu>
-  <div id="main-content">...</div>
-`);
-    }
     const content = this.contentId !== undefined
       ? document.getElementById(this.contentId)
-      : parent && parent.querySelector && parent.querySelector('[main]');
+      : null;
```
liamdebeasi, line 174: I opted for null because `getElementById` returns `null` is no element was found.

core/src/components/menu/menu.tsx
```diff
@@ -169,26 +169,11 @@ export class Menu implements ComponentInterface, MenuI {
...
 
-    if (!content || !content.tagName) {
-      // requires content element
+    if (content === null) {
```
liamdebeasi, line 176: If `content` is defined as an element, then `tagName` is defined: https://developer.mozilla.org/en-US/docs/Web/API/Element/tagName. Nothing on MDN indicates tagName could be the empty string (which is falsy in this case)

sean-perkins (approved):

Issue #24571, opened 2022-01-12
bug: `ion-toast` prevents focusing `ion-input` while visible


papagei-ma:
### Prerequisites

- [X] I have read the [Contributing Guidelines](https://github.com/ionic-team/ionic-framework/blob/main/.github/CONTRIBUTING.md#creating-an-issue).
- [X] I agree to follow the [Code of Conduct](https://ionicframework.com/code-of-conduct).
- [X] I have searched for [existing issues](https://github.com/ionic-team/ionic-framework/issues) that already report this problem, without success.

### Ionic Framework Version

- [ ] v4.x
- [ ] v5.x
- [X] v6.x

### Current Behavior

When a toast is being displayed the input elements (type='text', because checkboxes, selects and buttons work properly) are not working. I checked that they are not disabled, but somehow the click event is not captured. The element keep working programmatically properly, as I set a timer and changed the value of the bound variable every second, and the value gets emitted to the element. But no user interactivity is available 

### Expected Behavior

One can type in the input element

### Steps to Reproduce

1. Have an ion-input or an html input type='text' on your page.
2. Create a toast for a reasonable amount of time maybe 5 seconds, through 
` const toast = await this.toastController.create({
      message: 'Your settings have been saved.',
      duration: 2000
    });
    toast.present();`
3. While the toast is showing, try to type anything in the input element.

### Code Reproduction URL

https://github.com/papagei-ma/ion-toast-v6

### Ionic Info

Ionic:

   Ionic CLI                     : 6.16.3 (/opt/homebrew/lib/node_modules/@ionic/cli)
   Ionic Framework               : @ionic/angular 6.0.2
   @angular-devkit/build-angular : 13.0.4
   @angular-devkit/schematics    : 13.0.4
   @angular/cli                  : 13.0.4
   @ionic/angular-toolkit        : 5.0.3

### Additional Information

Checked on Chrome and safari, bug appears on both.

sean-perkins:
Thanks for reporting this! 

On an initial look, it looks like the keyboard focus trap we use for overlays is being applied to toasts (when it shouldn't).

papagei-ma:
@sean-perkins thanks for the quick reply, do you mean you introduced it just recently, because I know for sure, It was working just fine, before v6

PR #21716, opened 2020-07-09
fix(overlays): trap focus inside overlay components except toast

liamdebeasi:
## Pull request checklist

Please check if your PR fulfills the following requirements:
- [x] Tests for the changes have been added (for bug fixes / features)
- [ ] Docs have been reviewed and added / updated if needed (for bug fixes / features)
- [x] Build (`npm run build`) was run locally and any changes were pushed
- [x] Lint (`npm run lint`) has passed locally and any fixes were made for failures

## Pull request type

 

 

Please check the type of change your PR introduces:
- [x] Bugfix
- [ ] Feature
- [ ] Code style update (formatting, renaming)
- [ ] Refactoring (no functional changes, no api changes)
- [ ] Build related changes
- [ ] Documentation content changes
- [ ] Other (please describe): 

## What is the current behavior?

Issue Number: fixes https://github.com/ionic-team/ionic-framework/issues/21647

## What is the new behavior?

- All overlay components (except ion-toast) now have their focus trapped inside the overlay component.
- This is to bring it in line with WCAG recommendations: https://www.w3.org/TR/UNDERSTANDING-WCAG20/keyboard-operation-trapping.html

## Does this introduce a breaking change?

- [ ] Yes
- [x] No

PR #24572, opened 2022-01-13
fix(toast): allow input focus while toast is visible

sean-perkins:
## Pull request checklist

## Pull request type

 

 

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

Keyboard focus is trapped when the toast overlay is active. This prevents focusing form controls on the view. 

Issue Number: #24571

## What is the new behavior?

Keyboard focus is not trapped for overlays presented for `IonToast`. 

- [ ] Yes
- [x] No

### DIFF
--- a/core/src/components/toast/toast.tsx
+++ b/core/src/components/toast/toast.tsx
@@ -140,7 +140,9 @@ export class Toast implements ComponentInterface, OverlayInterface {
   @Event({ eventName: 'ionToastDidDismiss' }) didDismiss!: EventEmitter<OverlayEventDetail>;
 
   connectedCallback() {
-    prepareOverlay(this.el);
+    prepareOverlay(this.el, {
+      trapKeyboardFocus: false
+    });
   }
 
   /**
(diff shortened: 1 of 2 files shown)


amandaesmith3:
I found a few comments in `src/utils/overlays.ts` mentioning that toast should be excluded from focus trapping, so it does seem this change is how it's supposed to be; I have to wonder how it regressed 🤔 I had a couple questions about how this interacts with screen readers:

1. What about toasts that don't have a duration, remaining on screen until manually dismissed? Should focus trapping be maintained for those?
2. Does this change prevent screen readers from announcing the toast when it appears?

sean-perkins:
Agreed! I saw the same comments (#21716) and will do a further dive to see if I can find the exact regression. I know we have a few other issues around forcing the focus trap on overlays and some developers have asked for more direct access to toggling that behavior. This change currently is an internal API to configure that option, that would eventually allow us to support a binding on the overlay components to disable focus trap (for specific needs).

Currently, I believe this PR direction may be ideal regardless of the point of regression, as we were likely always adding a focus listener on the toast, instead of conditionally applying that logic for overlays that actually need it.

sean-perkins:
I still need to verify screen reader behavior, on my list for early tomorrow!

liamdebeasi (approved):

sean-perkins:
@amandaesmith3 verified with VO on mac, the toast is still announced correctly either before you input text, or after typing and opening a toast.

amandaesmith3 (approved):
Awesome, thanks for checking on that 👍

Issue #24575, opened 2022-01-14
bug: `ion-item` slot="error" not visible in Safari


luongtu996:
### Prerequisites

- [X] I have read the [Contributing Guidelines](https://github.com/ionic-team/ionic-framework/blob/main/.github/CONTRIBUTING.md#creating-an-issue).
- [X] I agree to follow the [Code of Conduct](https://ionicframework.com/code-of-conduct).
- [X] I have searched for [existing issues](https://github.com/ionic-team/ionic-framework/issues) that already report this problem, without success.

### Ionic Framework Version

- [ ] v4.x
- [ ] v5.x
- [X] v6.x

### Current Behavior

The new slot "error" does not work on my ipad 8, and iphone 7.
i found 
```
::slotted([slot=error])
```
allways display:none;

<img width="286" alt="Screenshot 2022-01-14 at 15 41 54" src="https://user-images.githubusercontent.com/26574116/149484684-49ef888e-1174-4459-a65b-a6fe993de2f1.png">

### Expected Behavior

Items with the error slot should show up as expected.

### Steps to Reproduce

Just try on real device
```
 <ion-item class="ion-invalid">
          <ion-label position="stacked">Name</ion-label>
          <ion-input></ion-input>
          <ion-note slot="error">Test</ion-note>
 </ion-item>
```

### Ionic Info

```
Ionic:

   Ionic CLI                     : 6.18.1 (.../.nvm/versions/node/v14.17.1/lib/node_modules/@ionic/cli)
   Ionic Framework               : @ionic/angular 6.0.2
   @angular-devkit/build-angular : 13.1.2
   @angular-devkit/schematics    : 12.2.14
   @angular/cli                  : 13.1.2
   @ionic/angular-toolkit        : 5.0.3

Capacitor:

   Capacitor CLI      : 3.3.4
   @capacitor/android : 3.3.4
   @capacitor/core    : 3.3.4
   @capacitor/ios     : 3.3.4

Utility:

   cordova-res (update available: 0.15.4) : 0.15.3
   native-run                             : 1.5.0

System:

   NodeJS : v14.17.1 (.../.nvm/versions/node/v14.17.1/bin/node)
   npm    : 7.19.0
   OS     : macOS Monterey
```

vildhjarta8:
+1. Not working for me.

https://github.com/ionic-team/ionic-framework/issues/24445#issuecomment-1013035382

luongtu996:
i have to force css for item

```
.html
[...]

PR #24579, opened 2022-01-14
fix(item): error slot visible in Safari

sean-perkins:
## Pull request checklist

## Pull request type

 

 

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

The error slot of the `ion-item` is never visible in Safari. 

Issue Number: #24575

## What is the new behavior?

Adjusts the style ordering so that the error slot display overrides correctly on all browser targets.

- [ ] Yes
- [x] No

## Other information

You can verify this change by running the Core project:
- http://localhost:3333/src/components/item/test/bottom
- Toggle the "Toggle Error" control
- See that the error slot becomes visible

### DIFF
--- a/core/src/components/item/item.scss
+++ b/core/src/components/item/item.scss
@@ -433,6 +433,12 @@ button, a {
   display: none;
 }
 
+::slotted([slot="error"]) {
+  display: none;
+
+  color: var(--highlight-color-invalid);
+}
+
 :host(.item-interactive.ion-invalid) ::slotted([slot="error"]) {
   display: block;
 }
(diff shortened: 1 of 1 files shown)


amandaesmith3 (approved):
LGTM 👍 I wonder if this'll fix some of the other error slot issues being reported too.

sean-perkins:
@amandaesmith3 hopefully a few! I Another active discussion is that implementers expect the error text to only display after the control has ben interacted with, instead of when the control is only invalid. This isn't initially how we designed the feature, so we would need to regroup on what this means and if there's a better API pattern that would allow customizing that default.

Issue #20201, opened 2020-01-14
bug: range does not work when document is rtl but component is forced to ltr


ashishgehlot:
Hello Devs, 

I was trying to use ion-slider for RTL application but its not working as expected. In LTR mode its working fine. In RTL mode, Knob gets stick in right side and can not change by dragging also value change appears in reverse mode like Dragging from Right to left but animate from left to right.

Here I am adding Codepen project to where I reproduced this issue. Hope it will help you to understand issue. 
[https://codepen.io/ashishgehlot95/pen/RwNBbpr?editors=1010]

brandyscarney:
Thanks for the issue! The codepen you added is using a very old version of Ionic, please make sure you're using latest, which is `4.11.8` at the time of this writing. Also, I believe you need to add the `dir` on the `html` element: https://codepen.io/brandyscarney/pen/gObjgzK?editors=1010

@abennouna could you confirm this?

abennouna:
The range issue has been solved in a previous version of Ionic (4.0.2).

I understand that the pen demonstrates an issue when using the `dir=rtl` on the ion-range element itself.

As a matter of fact, the RTL fix involved checking the direction on the document itself, and does not account for other parent elements, or the slider itself.

So if this is a mixed-direction document, that may be an issue.

ashishgehlot:
Yes, I am using mixed-direction. 

<html dir="rtl"> 

<ion-range min="1000" max="2000" mode="ios" step="100" pin="true" dir="ltr" snaps="true" ticks="true" color="secondary"></ion-range>

If I remove dir attribute from ion-range works fine but with attribute it breaks .

https://codepen.io/ashishgehlot95/pen/mdyajMd?editors=1010

ashishgehlot:
Any update on this issue ?

FahmiChaar:
Any option to force component direction to LTR when my app direction is RTL ?

3zzy:
Its still broken :(

sean-perkins:
Hello everyone :wave: this has been fixed in #24597 and will be available in the next release of Ionic.

PR #24597, opened 2022-01-19
fix(range): setting dir on ion-range to enable rtl mode now supported

sean-perkins:
## Pull request checklist

## Pull request type

 

 

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

`ion-range` only supports RTL when the direction is set at the document level.

Issue Number: #20201, Tech Debt

## What is the new behavior?

- `ion-range` will respect `dir` specified on the host element (overrides the document value)
- `ion-range` uses the `isRTL` utility

- [ ] Yes
- [x] No

### DIFF
--- a/core/src/components/range/range.tsx
+++ b/core/src/components/range/range.tsx
@@ -433,8 +433,10 @@ export class Range implements ComponentInterface {
     const barStart = `${ratioLower * 100}%`;
     const barEnd = `${100 - ratioUpper * 100}%`;
 
-    const start = isRTL(this.el) ? 'right' : 'left';
-    const end = isRTL(this.el) ? 'left' : 'right';
+    const rtl = isRTL(this.el);
+
+    const start = rtl ? 'right' : 'left';
+    const end = rtl ? 'left' : 'right';
 
     const tickStyle = (tick: any) => {
       return {
@@ -501,7 +503,7 @@ export class Range implements ComponentInterface {
             part="bar-active"
           />
 
-          {renderKnob(isRTL(this.el), {
+          {renderKnob(rtl, {
             knob: 'A',
             pressed: pressedKnob === 'A',
             value: this.valA,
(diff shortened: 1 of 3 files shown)


amandaesmith3 (approved):

core/src/components/range/range.tsx
```diff
@@ -432,10 +433,8 @@ export class Range implements ComponentInterface {
     const barStart = `${ratioLower * 100}%`;
     const barEnd = `${100 - ratioUpper * 100}%`;
 
-    const doc = document;
-    const isRTL = doc.dir === 'rtl';
-    const start = isRTL ? 'right' : 'left';
-    const end = isRTL ? 'left' : 'right';
+    const start = isRTL(this.el) ? 'right' : 'left';
+    const end = isRTL(this.el) ? 'left' : 'right';
```
liamdebeasi, line 437: Minor optimization, but we can cache `isRTL(this.el)` to a variable so we don't need to call it twice.

core/src/components/range/range.tsx
```diff
@@ -502,7 +501,7 @@ export class Range implements ComponentInterface {
             part="bar-active"
           />
 
-          { renderKnob(isRTL, {
+          {renderKnob(isRTL(this.el), {
```
liamdebeasi, line 504: Same thing here. The calls in this function can cached within the function context

liamdebeasi (approved):
Two minor comments, but otherwise looks good. Let's make sure the commit subject follows the guidelines too. Maybe something like `fix(range): setting dir on ion-range to enable rtl mode now supported`?

liamdebeasi:
Can we also check to see if this is fixed? https://github.com/ionic-team/ionic-framework/issues/20201

sean-perkins:
> Can we also check to see if this is fixed? #20201

@liamdebeasi yeah, that issue will be resolved after I add the utility check for one more area I missed 😬

liamdebeasi:
Nice!!

Ornel21:
npm run build

Issue #19884, opened 2019-11-10
bug: ion-input not updating value if space trimmed equivalent of value changes


mattsteve:
# Bug Report

**Ionic version:**
[x] **4.11.4**

**Current behavior:**
When setting an `ion-input`'s value to a trimmed version of its previous value, the input does not recognize the value change. Therefore, there is no way to enforce input value trimming while typing inside an `ion-input` or after leaving an `ion-input`.

**Expected behavior:**
Setting the value of an `ion-input` should accept any string and update accordingly.

**Steps to reproduce:**
1. Initialize an ion-input on the page.
2. Listen to the ionInput event.
3. Inside the inputInput event, set the input's value to a trimmed version of its own value.

**Related code:**

**JS FIDDLE:** https://jsfiddle.net/kpe14jwz/3/

```HTML
<ion-item>
        <ion-label position="floating">Type here</ion-label>
        <ion-input id="ionInput"></ion-input>
</ion-item>
```

```Javascript
document.getElementById('ionInput')
.addEventListener('ionInput', (e) => {
	e.target.value = e.target.value.trim();
});
```

**Ionic Info:**

N/A. Working out of jsfiddle with CDN.

cjorasch:
If your goal is to trim the value (vs. remove all whitespace) then one potential issue with this approach is that the user will not be able to enter spaces in the middle of content.  For example, trying to enter "John Smith" results in "JohnSmith" because the space is trimmed as the user is typing.  This happens in the native example in your fiddle.

Changing the value during input can create some strange problems if you type quickly because of async event handling and debouncing.  For example, if you type `this and` quickly into the native example on Safari you get `this'd`.  Maybe because of some autocorrect features?

Normalizing is safer to do on blur or when handling the form content.

mattsteve:
Irregardless of the specific example, you can see how this would cause problems when trying to implement features that modify user input while typing such as input masks.

Same problem when preventing the user from typing anything but numbers:
https://jsfiddle.net/pk5uq1eL/

If this behaviour works in a native input, I expect it to work in ion-input. Async event handling should have nothing to do with accepting a new value. The last value that was set should be the value that the input uses.

liamdebeasi:
Thanks for the issue. This appears to be a regression that appeared in Ionic 4.6.0, so I will mark this as a bug.

Example in Ionic 4.5.0: https://codepen.io/liamdebeasi/pen/RwwBYop
[...]

PR #24606, opened 2022-01-20
fix(input): ion-input accepts any string value

sean-perkins:
## Pull request checklist

## Pull request type

 

 

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

Updating the value of the `ion-input` during the `ionInput` event callback, will not be reflected in the control. This means you cannot have functionality such as trimming white spaces from an input. 

Issue Number: #19884

## What is the new behavior?

The native input is updated when the attribute change callback is fired for the `value` prop. This forces the input element to always be in sync with the value of the `ion-input` web component.

This allows implementations such as preventing whitespace in an `ion-input`.

- [ ] Yes
- [x] No

### DIFF
--- a/core/src/components/input/input.tsx
+++ b/core/src/components/input/input.tsx
@@ -231,6 +231,17 @@ export class Input implements ComponentInterface {
    */
   @Watch('value')
   protected valueChanged() {
+    if (this.nativeInput) {
+      /**
+       * Assigning the native input's value on attribute
+       * value change, allows `ionInput` implementations
+       * to override the control's value.
+       *
+       * Used for patterns such as input trimming (removing whitespace),
+       * or input masking.
+       */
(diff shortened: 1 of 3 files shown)


amandaesmith3 (approved):
LGTM! It's interesting that the `valueChanged` handler has a comment implying this already should've been happening, but I couldn't find anywhere else `nativeInput.value` is updated. Must've been removed at some point 🤔

liamdebeasi (commented):
This issue regressed in 4.6, which was when Ionic was updated to the Stencil 1.0.0 release. That update cause quite a few issues, so this could be left over from that change.

Otherwise, I don't see anything else in the 4.6 changelog that would cause this behavior change.

liamdebeasi (approved):

Issue #24588, opened 2022-01-17
bug: datetime locale with h23 does not respect max range


OlivierChirouze:
### Prerequisites

- [X] I have read the [Contributing Guidelines](https://github.com/ionic-team/ionic-framework/blob/main/.github/CONTRIBUTING.md#creating-an-issue).
- [X] I agree to follow the [Code of Conduct](https://ionicframework.com/code-of-conduct).
- [X] I have searched for [existing issues](https://github.com/ionic-team/ionic-framework/issues) that already report this problem, without success.

### Ionic Framework Version

- [ ] v4.x
- [ ] v5.x
- [X] v6.x

### Current Behavior

I set min and max values for date-time.
- min value is 18:30
- max value is 20:40
- initial value is set to 20:30

I carefully provide values that don't include any timezone, to keep things simple. It works fine for min and initial value. It doesn't for max.
Max value seems to be exactly 3 hours later than the provided value (one can set anything up to 23:40). Users should be prevented from setting anything above 20:40.

Example:
https://stackblitz.com/edit/angular-ivy-cyzzpl?file=src/app/app.component.ts

As far as I remember, this used to work fine on Ionic 5 (but I had to adapt the code for the new date format anyway).

### Expected Behavior

The min value should be exactly the one provided and users should not be able to pass it.

### Steps to Reproduce

Set a max value and try to define a value that goes beyond this max value.

### Code Reproduction URL

https://stackblitz.com/edit/angular-ivy-cyzzpl?file=src/app/app.component.ts

### Ionic Info

Ionic:

   Ionic CLI                     : 6.18.1 (/Users/o.chirouze/.nvm/versions/node/v14.18.2/lib/node_modules/@ionic/cli)
   Ionic Framework               : @ionic/angular 6.0.2
   @angular-devkit/build-angular : 12.2.14
   @angular-devkit/schematics    : 12.2.14
   @angular/cli                  : 12.2.14
   @ionic/angular-toolkit        : 4.0.0

Capacitor:

   Capacitor CLI      : 3.3.2
   @capacitor/android : 3.3.2
   @capacitor/core    : 3.3.2
   @capacitor/ios     : 3.3.2

Cordova:

   Cordova CLI       : not installed
   Cordova Platforms : not available
   Cordova Plugins   : not available

Utility:

   cordova-res : not installed globally
   native-run  : 1.5.0

System:

[...]

PR #24476, opened 2021-12-22
fix(datetime): time picker format with hourCycle h23

sean-perkins:
## Pull request checklist

Please check if your PR fulfills the following requirements:
- [x] Tests for the changes have been added (for bug fixes / features)
- [ ] Docs have been reviewed and added / updated if needed (for bug fixes / features)
- [x] Build (`npm run build`) was run locally and any changes were pushed
- [x] Lint (`npm run lint`) has passed locally and any fixes were made for failures

## Pull request type

 

 

Please check the type of change your PR introduces:
- [x] Bugfix
- [ ] Feature
- [ ] Code style update (formatting, renaming)
- [ ] Refactoring (no functional changes, no api changes)
- [ ] Build related changes
- [ ] Documentation content changes
- [ ] Other (please describe): 

## What is the current behavior?

The time picker display ignores the value of hour cycle. The time picker will also restrict all hour selections to the minimum minute range; even though the time is above the min range.

i.e.: When the minimum time is 19:30, you are unable to select any value under 20:30 when the hour is 20.

Issue Number: Resolves #24474

## What is the new behavior?

- Time picker will return the hours back in 24 hour format when hourCycle is h23
- Time picker allows minute selection above the min range, when the hour is above the min hour (if specified)

## Does this introduce a breaking change?

- [ ] Yes
- [x] No

## Other information

|Before|After|
|---|---|
|<video src="https://user-images.githubusercontent.com/13732623/147148436-637d0bfb-6437-4583-be96-26f1513886b2.mp4"></video>| <video src="https://user-images.githubusercontent.com/13732623/147146337-f0d23717-ec29-41ad-9cf8-278236c85a0d.mp4"></video>|

PR #24610, opened 2022-01-20
fix(datetime): datetime locale with h23 will respect max time range

sean-perkins:
## Pull request checklist

## Pull request type

 

 

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

Locales that have the hour cycle in 24 hour format, will not respect the max time range specified. This means if you set the max to `20:40`, they are still able to select `21, 22, 23` for hour values.

Issue Number: #24588

## What is the new behavior?

- The time picker will respect the max hour range when the locale is in a 24 hour format
- Applies the same logic for restricting minutes used here #24476 

- [ ] Yes
- [x] No

### DIFF
--- a/core/src/components/datetime/utils/data.ts
+++ b/core/src/components/datetime/utils/data.ts
@@ -219,7 +219,7 @@ export const generateTime = (
       if (maxParts.hour !== undefined) {
         processedHours = processedHours.filter(hour => {
           const convertedHour = refParts.ampm === 'pm' ? (hour + 12) % 24 : hour;
-          return convertedHour <= maxParts.hour!;
+          return (use24Hour ? hour : convertedHour) <= maxParts.hour!;
         });
         isPMAllowed = maxParts.hour >= 13;
       }
--- a/core/src/components/datetime/test/data.spec.ts
+++ b/core/src/components/datetime/test/data.spec.ts
@@ -257,5 +257,35 @@ describe('generateTime()', () => {
         expect(minutes.length).toEqual(60);
       });
     });
+
+    it('should respect the min & max bounds', () => {
+      const refValue = {
+        day: undefined,
+        month: undefined,
+        year: undefined,
+        hour: 20,
+        minute: 30
+      }
+
+      const minParts = {
+        day: undefined,
+        month: undefined,
+        year: undefined,
+        hour: 19,
+        minute: 30
+      }
+
+      const maxParts = {
+        day: undefined,
+        month: undefined,
+        year: undefined,
+        hour: 20,
+        minute: 40
+      };
+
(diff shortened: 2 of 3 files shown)


core/src/components/datetime/utils/data.ts
```diff
@@ -219,7 +219,7 @@ export const generateTime = (
       if (maxParts.hour !== undefined) {
         processedHours = processedHours.filter(hour => {
           const convertedHour = refParts.ampm === 'pm' ? (hour + 12) % 24 : hour;
```
liamdebeasi, line 221: Can you explain why this works? If you are using 24 hour time, `refParts.ampm` should be `undefined`, which means `convertedHour` would be `hour`.

sean-perkins: The `workingParts` (used as `refParts` in that util function) has `ampm` defined. Likely as a result of when we call `processValue` on initialization of the datetime component:
```ts
...
this.workingParts = {
  ...
  ampm: hour >= 12 ? 'pm' : 'am'
}
```

Where as the hour cycle only takes into consideration the locale or if the developer has hard-set the hour cycle.

So this case exists when the developer hasn't specified the hour cycle manually, but the locale requires it.

liamdebeasi: Ok thanks. Would it make more sense to not have `ampm` be optional in then? https://github.com/ionic-team/ionic-framework/blob/main/core/src/components/datetime/datetime-interface.ts#L21

liamdebeasi (approved):

core/src/components/datetime/utils/data.ts
sean-perkins: Yeah, I think that makes sense for now. I'll confirm we aren't using that interface for any partial date object comparisons.

amandaesmith3 (approved):

core/src/components/datetime/utils/data.ts
sean-perkins: Looks like we are in a few places:
- https://github.com/ionic-team/ionic-framework/blob/main/core/src/components/datetime/datetime.tsx#L1182 
- https://github.com/ionic-team/ionic-framework/blob/main/core/src/components/datetime/datetime.tsx#L563
- https://github.com/ionic-team/ionic-framework/blob/main/core/src/components/datetime/datetime.tsx#L567

I'll create a tech debt task to have more strict explicit types.

Agular:
Applies the same logic for restricting minutes used here fix(datetime): time picker format with hourCycle h23 #24476

Was this implemented? Since I am using ionic 6.0.4 and the minutes are not updated when choosing a time with the hour less than the hour of the max time. This was recently fixed for min value.

PR #24621, opened 2022-01-21
chore(angular): upgrade test app to Angular 13

sean-perkins:
## Pull request checklist

Please check if your PR fulfills the following requirements:
- [ ] Tests for the changes have been added (for bug fixes / features)
- [ ] Docs have been reviewed and added / updated if needed (for bug fixes / features)
- [x] Build (`npm run build`) was run locally and any changes were pushed
- [x] Lint (`npm run lint`) has passed locally and any fixes were made for failures

## Pull request type

 

 

Please check the type of change your PR introduces:
- [ ] Bugfix
- [ ] Feature
- [ ] Code style update (formatting, renaming)
- [ ] Refactoring (no functional changes, no api changes)
- [ ] Build related changes
- [ ] Documentation content changes
- [x] Other (please describe): Tech Debt

## What is the current behavior?

Issue Number: N/A

## What is the new behavior?

- Upgrades Angular test app to Angular 13 dependencies
- Remove deprecated `entryComponents`
- Remove deprecated polyfill comments

## Does this introduce a breaking change?

- [ ] Yes
- [x] No

Issue #24586, opened 2022-01-17
bug: routerlink with null throws error in angular 13


marcovmun:
### Prerequisites

- [X] I have read the [Contributing Guidelines](https://github.com/ionic-team/ionic-framework/blob/main/.github/CONTRIBUTING.md#creating-an-issue).
- [X] I agree to follow the [Code of Conduct](https://ionicframework.com/code-of-conduct).
- [X] I have searched for [existing issues](https://github.com/ionic-team/ionic-framework/issues) that already report this problem, without success.

### Ionic Framework Version

- [ ] v4.x
- [ ] v5.x
- [X] v6.x

### Current Behavior

If I use routerLink with value null, I get console error.
```
<ion-item [routerLink]="null"></ion-item>
``` 

**Error**
```
core.mjs:6449 
        
       ERROR TypeError: Cannot read properties of null (reading 'root')
    at DefaultUrlSerializer.serialize (router.mjs:878:51)
    at Router.serializeUrl (router.mjs:4757:35)
    at RouterLinkDelegateDirective.updateTargetUrlAndHref (ionic-angular.js:3764:79)
    at RouterLinkDelegateDirective.ngOnInit (ionic-angular.js:3752:14)
    at callHook (core.mjs:2518:1)
    at callHooks (core.mjs:2487:1)
    at executeInitAndCheckHooks (core.mjs:2438:1)
    at selectIndexInternal (core.mjs:8368:1)
    at ɵɵadvance (core.mjs:8351:1)
```

I think it has something to do with:
> You can now disable the navigation of a routerLink by passing undefined and null. Previously the routerLink directive used to accept these two values as equivalent to an empty string.

https://update.angular.io/?l=3&v=12.0-13.0

### Expected Behavior

No error is visible in console and routerlink is disabled.

### Steps to Reproduce

1. Use Ionic 5.0.2 and angular 13
2. add view with [routerLink]="null" in an element
3. Run project and look in console.

sean-perkins:
@marcovmun thanks for reporting this. I don't believe we've updated our internal `RouterLinkDelegate` to support that new feature change in v13 of Angular. 

I'll need to look into all aspects of what assigning `null` to the routerLink includes, so that we have alignment with the expected implementation.

PR #24622, opened 2022-01-21
fix(angular): routerLink with null value works with Angular 13

sean-perkins:
## Pull request checklist

## Pull request type

 

 

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

Ionic Angular apps on Angular 13 will throw an exception when using `[routerLink]="null"` (new feature with Angular 13 to disable the router link).

Issue Number: #24586

## What is the new behavior?

- `[routerLink]="null"` now works with Ionic apps with Angular 13

- [ ] Yes
- [x] No

## Other information

Branched as a dependent of #24621, as we should upgrade the test app before patching fixes specific to Angular 13 without a reproduction app. Will update the base branch once that PR is reviewed, approved and merged.


angular/src/directives/navigation/router-link-delegate.ts
```diff
@@ -41,7 +41,7 @@ export class RouterLinkDelegateDirective implements OnInit, OnChanges, OnDestroy
   }
 
   private updateTargetUrlAndHref() {
-    if (this.routerLink) {
+    if (this.routerLink?.urlTree) {
```
liamdebeasi, line 44: Am I correct in assuming that when `[routerLink]="null"` there is no urlTree created by Angular? Also any backwards incompatible changes to be aware of for Angular 12?

sean-perkins: Correct, in Angular <= 12, `[routerLink]="null"` was evaluated as `routerLink=""`, but in Angular 13, it is explicitly null. This change was introduced into Angular to allow disabling the router link behavior in app.

Since we are using dependency injection for accessing the router link instance, it should always be defined as long as the `RouterModule` is imported (previous check), but with this change it will make sure the router link is enabled before applying our logic. 

I did verify with the test-app on v12, that the behavior remained the same.

liamdebeasi: Sounds great!

liamdebeasi (approved):
Great job on this! 🚀

amandaesmith3 (approved):

PR #24627, opened 2022-01-23
refactor(router): code cleanup.

vicb:
The main thing is to rename path to segments when the router deals with an array of segments.
The word "path: is reserved for the string version to remove confusion.

## Pull request checklist

- [X] Build (`npm run build`) was run locally and any changes were pushed - changes not pushed as not related to this PR.

## Pull request type

- [ ] Bugfix
- [ ] Feature
- [X] Refactoring (no functional changes, no api changes)
- [ ] Build related changes

- [ ] Yes
- [X] No

### DIFF
--- a/core/src/components/router/utils/dom.ts
+++ b/core/src/components/router/utils/dom.ts
@@ -3,6 +3,14 @@ import { componentOnReady } from '../../../utils/helpers';
 
 import { ROUTER_INTENT_NONE } from './constants';
 
+/**
+ * Activates the passed route chain.
+ *
+ * There must be exactly one outlet per route entry in the chain.
+ *
+ * The methods calls setRouteId on each of the outlet with the corresponding route entry in the chain.
+ * setRouteId will create or select the view in the outlet.
+ */
 export const writeNavState = async (
   root: HTMLElement | undefined,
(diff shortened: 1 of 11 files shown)


core/src/components/router/utils/dom.ts
```diff
@@ -3,6 +3,12 @@ import { componentOnReady } from '../../../utils/helpers';
 
 import { ROUTER_INTENT_NONE } from './constants';
 
+// Activates the passed route chain.
```
sean-perkins, line 6: These comment blocks will need to be converted to multi-line: `/**`. For whatever reason, VSCode doesn't detect these code blocks for the tooltip hover interaction otherwise. 

This applies to most root-level comments. Local variables that aren't exported don't _have to_ have multi-line, but it is a project convention we should try to follow.

vicb: I completely overlook hiver interaction, sorry.

One other benefit of JSDoc is that it makes it easier to document parameters and return values. I'll keep improving the doc blocks in the follow up PRs.

(I have pushed an updated version)

sean-perkins (approved):
Appreciate the changes here, great work 👍

PR #24642, opened 2022-01-24
fix(modal): native keyboard will dismiss when bottom sheet is dragged

### LINKED ISSUE #23878: bug: sheet modal gesture breaks when interacting with inputs/keyboard on ios
### Prequisites

### Ionic Framework Version

- [ ] v4.x
- [ ] v5.x
- [X] v6.x

### Current Behavior

I just tested the new Sheet Modal. First: Great work 🎉 Here is a Collections of things i mentioned:
- Similar to #23561: My IDEA (InteliJ) seems to not know the new properties like `breakpoints` or `initialBreakpoint`. Maybe missing in the angular component declaration?
- The `handle` can not be set to **false** for Sheet Modals: https://github.com/ionic-team/ionic-framework/blob/9fa373c6efeceb39913f1c158cda383c38cba29c/core/src/components/modal/modal.tsx#L550
- The handle itself doesn't looks nice in combination with IonTitle - Looks like it needs a bit more spacing here 
- The Backdrop seems to be a bit buggy yet, see video: After drag finished it changes:

- When i have an Input inside the Sheet Modal and focus it the Keyboard appears and the Modal pushs up a bit. After losing the focus of the input and the keyboard is gone, the breakpoints do not work anymore and i'm kind of in a free mode where i can stop the Modal wherever i want
- Where we already talk about that: This `freeMode` would be a cool feature
- Also it would be great to be able to disable the possibility to drag the Modal and add the possibility to change the current breakpoint programatically.

### Expected Behavior

Everything is described above

### Steps to Reproduce

Version: `6.0.0-beta.5`

### Code Reproduction URL

_No response_

### Ionic Info

_No response_

### Additional Information

_No response_

sean-perkins:
## Pull request checklist

## Pull request type

 

 

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

The swipe gesture for the modal bottom sheet will trigger when a form element (input, textarea, etc.) receives focus and opens the native keyboard. This causes the sheet gesture to become unresponsive and free scroll. 

Issue Number: Resolves #23878

## What is the new behavior?

- Swipe gestures will be disabled when the keyboard is immediately open (form element has focus).
- Swipe gestures will be re-enabled on the next frame (after the keyboard has repositioned the view)
- The swipe gesture for the sheet modal will dismiss any active keyboards (this aligns with native behavior, see: Maps sheet implementation on iOS)

- [ ] Yes
- [x] No

## Other information

|Before|After|
|---|---|
|<video src="https://user-images.githubusercontent.com/13732623/150849156-0a8ed91f-4c5e-49ea-a1ef-ab88bc31815d.mp4"></video>|<video src="https://user-images.githubusercontent.com/13732623/150849184-5f07511f-1cad-452e-8518-16e61139a7f1.mp4"></video>|

Behavior is easier to feel on device than from a video:
1. Update the sheet example test to render an `<ion-input></ion-input>` for each item
2. Run Core
3. Open the test on your mobile device
4. Present the modal sheet and play with the gesture, breakpoints and keyboard focus

### DIFF
--- a/core/src/components/modal/gestures/sheet.ts
+++ b/core/src/components/modal/gestures/sheet.ts
@@ -1,6 +1,7 @@
 import { Animation } from '../../../interface';
 import { GestureDetail, createGesture } from '../../../utils/gesture';
 import { clamp, raf } from '../../../utils/helpers';
+import { KEYBOARD_DID_OPEN } from '../../../utils/keyboard/keyboard';
 import { getBackdropValueForSheet } from '../utils';
 
 export const createSheetGesture = (
@@ -42,6 +43,27 @@ export const createSheetGesture = (
   const backdropAnimation = animation.childAnimations.find(ani => ani.id === 'backdropAnimation');
   const maxBreakpoint = breakpoints[breakpoints.length - 1];
 
+  /* tslint:disable-next-line */
+  if (typeof window !== 'undefined') {
+    const keyboardOpenCallback = () => {
+      /**
+       * When the native keyboard is opened and the webview
+       * is resized, the gesture implementation will become unresponsive
+       * and enter a free-scroll mode.
+       *
+       * When the keyboard is opened, we disable the gesture for
+       * a single frame and re-enable once the contents have repositioned
+       * from the keyboard placement.
+       */
+      gesture.enable(false);
+      raf(() => {
+        gesture.enable(true)
+      });
+    };
+
+    window.addEventListener(KEYBOARD_DID_OPEN, keyboardOpenCallback);
+  }
+
   /**
    * After the entering animation completes,
    * we need to set the animation to go from
@@ -78,6 +100,15 @@ export const createSheetGesture = (
 
     if (currentBreakpoint === 1 && content) {
       return false;
+    }
+
+    /* tslint:disable-next-line */
+    if (typeof document !== 'undefined') {
+      const activeElement = baseEl.ownerDocument.activeElement as HTMLElement;
+      if (activeElement.matches('input,ion-input,textarea,ion-textarea')) {
+        // Dismisses the open keyboard when the sheet drag gesture is started.
+        activeElement.blur();
+      }
     }
 
     return true;
@@ -188,11 +219,11 @@ export const createSheetGesture = (
           }
         }
 
-      /**
-       * This must be a one time callback
-       * otherwise a new callback will
-       * be added every time onEnd runs.
-       */
+        /**
+         * This must be a one time callback
+         * otherwise a new callback will
+         * be added every time onEnd runs.
+         */
       }, { oneTimeCallback: true })
       .progressEnd(1, 0, 500);
 
--- a/core/src/components/modal/gestures/sheet.ts
+++ b/core/src/components/modal/gestures/sheet.ts
@@ -43,24 +43,24 @@ export const createSheetGesture = (
   const backdropAnimation = animation.childAnimations.find(ani => ani.id === 'backdropAnimation');
   const maxBreakpoint = breakpoints[breakpoints.length - 1];
 
+  const keyboardOpenCallback = () => {
+    /**
+     * When the native keyboard is opened and the webview
+     * is resized, the gesture implementation will become unresponsive
+     * and enter a free-scroll mode.
+     *
+     * When the keyboard is opened, we disable the gesture for
+     * a single frame and re-enable once the contents have repositioned
+     * from the keyboard placement.
+     */
+    gesture.enable(false);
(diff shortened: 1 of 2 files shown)


EinfachHans:
@sean-perkins amazing!! Looks good so far 😊 Dev build would be awesome to test it in a dev app

core/src/components/modal/gestures/sheet.ts
```diff
@@ -42,6 +43,27 @@ export const createSheetGesture = (
...
+       * a single frame and re-enable once the contents have repositioned
+       * from the keyboard placement.
+       */
+      gesture.enable(false);
+      raf(() => {
+        gesture.enable(true)
+      });
+    };
+
+    window.addEventListener(KEYBOARD_DID_OPEN, keyboardOpenCallback);
```
amandaesmith3, line 64: It looks like this only gets called once when the modal is presented, but do we need to remove the listener when it's dismissed? 🤔

core/src/components/modal/gestures/sheet.ts
```diff
@@ -42,6 +43,27 @@ export const createSheetGesture = (
   const backdropAnimation = animation.childAnimations.find(ani => ani.id === 'backdropAnimation');
   const maxBreakpoint = breakpoints[breakpoints.length - 1];
 
+  /* tslint:disable-next-line */
```
amandaesmith3, line 46: Why the tslint disables? (Here and on L105)

core/src/components/modal/gestures/sheet.ts
```diff
@@ -80,6 +102,15 @@ export const createSheetGesture = (
...
 
+    /* tslint:disable-next-line */
+    if (typeof document !== 'undefined') {
+      const activeElement = baseEl.ownerDocument.activeElement as HTMLElement;
+      if (activeElement.matches('input,ion-input,textarea,ion-textarea')) {
+        // Dismisses the open keyboard when the sheet drag gesture is started.
+        activeElement.blur();
+      }
+    }
+
```
amandaesmith3, line 113: Putting this in `canStart` feels a little weird; it seems like it should go in `onStart` or somewhere else that more explicitly indicates that we're starting a gesture.

core/src/components/modal/gestures/sheet.ts
sean-perkins: The linter errors out that the check is always true, but that isn't the case in SSR. We use this approach in almost all cases where we are checking if the document or window is available.

core/src/components/modal/gestures/sheet.ts
sean-perkins: That's a good point. I had similar concessions, but wasn't able to find anywhere in gestures that we expose appending additional behavior to teardown. Let me see if there's a way to extend the `destroy` method of the gesture (which is called in the dismiss callback from the modal).

core/src/components/modal/gestures/sheet.ts
amandaesmith3: Ah, gotcha 👍 I wasn't getting an error on my end after removing the disable comment, so I wasn't sure. I'm seeing it in other places though, like you said.

core/src/components/modal/gestures/sheet.ts
sean-perkins: Agreed 🤔  let me confirm that doing this behavior while the gesture is starting, doesn't cause any other issues. Then I'll move it 👍

amandaesmith3 (approved):
Looks great, nice work 👍

sean-perkins:
> @sean-perkins amazing!! Looks good so far 😊 Dev build would be awesome to test it in a dev app

@EinfachHans The dev build version is `6.0.4-dev.1643056868.324d279`.

i.e.:
```bash
npm i @ionic/angular@6.0.4-dev.1643056868.324d279
```

EinfachHans (approved):
@sean-perkins works as expected! 😊

core/src/components/modal/gestures/sheet.ts
```diff
@@ -93,6 +115,17 @@ export const createSheetGesture = (
       contentEl.scrollY = false;
     }
 
+    /* tslint:disable-next-line */
+    if (typeof document !== 'undefined') {
+      const activeElement = baseEl.ownerDocument.activeElement as HTMLElement;
+      if (activeElement.matches('input,ion-input,textarea,ion-textarea')) {
```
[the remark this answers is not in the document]
ccapndave: I'm using `contenteditable` (with Lexical) and this does indeed break!

Issue #24069, opened 2021-10-13
feat: sheet style modal should change breakpoint on handle click


EinfachHans:
### Prerequisites

- [X] I have read the [Contributing Guidelines](https://github.com/ionic-team/ionic-framework/blob/main/.github/CONTRIBUTING.md#creating-an-issue).
- [X] I agree to follow the [Code of Conduct](https://ionicframework.com/code-of-conduct).
- [X] I have searched for [existing issues](https://github.com/ionic-team/ionic-framework/issues) that already include this feature request, without success.

### Describe the Feature Request

Check out the sheet style modal in the native ios maps app.

If you click on the handle the sheet dynamically moves to the next breakpoint (in a loop).

### Describe the Use Case

Use case is to match native 😃 

### Describe Preferred Solution

I would prefer a defaultly **true** setted property.

Also we maybe have to think about what happens with sheet style modals that have a `0` breakpoint

Issue #23917, opened 2021-09-12
feat: sheet modal programmatically move to a breakpoint


EinfachHans:
### Prequisites

- [X] I have read the [Contributing Guidelines](https://github.com/ionic-team/ionic-framework/blob/main/.github/CONTRIBUTING.md#creating-an-issue).
- [X] I agree to follow the [Code of Conduct](https://ionicframework.com/code-of-conduct).
- [X] I have searched for [existing issues](https://github.com/ionic-team/ionic-framework/issues) that already include this feature request, without success.

### Describe the Feature Request

I have some feature requests for the new v6 Sheet Modal that all kinda belongs to each other:

## Auto calculated height
It would be great to make the modal have the height of the content, not more, not less. The library we are currently using (https://github.com/roman-rr/cupertino-pane) has such an option.

## Disable Drag Gesture
It would be great do be able to dynamically enable/disable the Drag Gesture

## Manual move to Breakpoint
It would be great to be able to have a method on the modal like `moveToBreakpoint()`, where i can dynamically move to a breakpoint.

### Describe the Use Case

These three requests should work all together, image the following use case:

I want to show a Sheet Modal that has only a checkbox and a button below.
This should have the height of the content (Request No 1).
Dragging on this Sheet should be disabled (Request No 2).
If the users selects (or unselects) the checkbox some more content will be shown. After that i want to dynamically move the Breakpoint (Request No 3).
Also this `moveToBreakpoint()` method should of course accept numbers for the breakpoint, but also accepts something like `'auto'` to adjust the height on the current content (which changes with the checkbox)

 I hope i described it understandable 😃

liamdebeasi:
Can you provide a code example of how this currently works using the Cupertino pane?

EinfachHans:
See [here](https://github.com/roman-rr/cupertino-pane#common-configuration): There is the option **fitHeight**

liamdebeasi:
Sorry, my last comment was confusing. What I was asking for is a code example of how this is working with regards to your use case. I am interested in seeing how you are applying the feature from Cupertino Pane to address the use case that you have.

EinfachHans:
I create the pane with 

```
this.drawer = new CupertinoPane('#drawerPane', {
  showDraggable: false,
  fitHeight: true
});
```

It offers a function [`calcFitHeight()`](https://github.com/roman-rr/cupertino-pane#calcfitheight) which i can use after the content changes (the checkbox in my example).
[...]

PR #24648, opened 2022-01-25
feat(modal): add ability to programmatically set current sheet breakpoint

EinfachHans:
## Pull request checklist

## Pull request type

 

 

- [ ] Bugfix
- [X] Feature
- [ ] Build related changes

Developers cannot programmatically move a sheet modal breakpoint with an API.

Issue Number: #23917 

## What is the new behavior?

- `setCurrentBreakpoint(breakpoint: number)` to programmatically change the breakpoint of a sheet modal.
- `getBreakpoint()` to get the active breakpoint value for a sheet modal.
- `ionModalBreakpointWillChange/breakpointWillChange` event when the breakpoint is about to change.
- `ionModalBreakpointDidChange/breakpointDidChange` event when the breakpoint changes.
- extract Code to animate the gesture to a specific breakpoint to be used for the new method and the existing usage of the gesture
- after this is merged #24069 is easy as well, i can create a PR for this as well

- [ ] Yes
- [X] No


EinfachHans:
Also a `breakpointDidChange` Event would make sense here - any advice on how to add this?

core/src/components/modal/modal.tsx
```diff
@@ -560,6 +582,43 @@ export class Modal implements ComponentInterface, OverlayInterface {
     return eventMethod(this.el, 'ionModalWillDismiss');
   }
 
+  /**
+   * Move a sheet style modal to a specific breakpoint.
+   */
+  @Method()
+  async setBreakpoint(breakpoint: number): Promise<void> {
```
sean-perkins, line 589: How do you feel about renaming this public API to `setActiveBreakpoint`/`setCurrentBreakpoint`? It would be more explicit that it's effecting the current breakpoint.

sean-perkins (changes requested):
Have one suggested change for the public API naming for setting the breakpoint value.

Reviewed most of the changes here and they align with our internal feature design document. I would like to introduce a set of e2e tests on this new behavior, mainly to verify correct event emission and a few screenshot tests. I'll clone your fork this week and experiment with your changes to verify everything works correctly and will suggest a few tests on the new behavior.

Thanks for the great work here!

core/src/components/modal/modal.tsx
```diff
@@ -560,6 +582,43 @@ export class Modal implements ComponentInterface, OverlayInterface {
...
+    if (!this.isSheetModal || !this.breakpoints!.includes(breakpoint) || this.currentBreakpoint === breakpoint) {
+      return
```
sean-perkins, line 591: ~ Can we add a console warning here to inform developers when they are trying to set a breakpoint value on the unsupported cases?

```ts
if (!this.isSheetModal) {
  console.warning('setCurrentBreakpoint is only supported on sheet modals.');
  return;
}
if (!this.breakpoints!.includes(breakpoint)) {
  console.warn(`Attempted to set invalid breakpoint value ${breakpoint}. Please double check that the breakpoint value is part of your defined breakpoints.`);
  return;
}
if (this.currentBreakpoint === breakpoint) {
  return;
}
``

File: core/src/components/modal/modal.tsx (hunk 1 above)
Comment (sean-perkins) [inline core/src/components/modal/modal.tsx:589]: ~ What if we made this function return `Promise<boolean>`? Returning `false` when the value is not set and returning true when it completes successfully?

Comment (EinfachHans) [reply core/src/components/modal/modal.tsx:589]: For what case would this help? 🤔 i don't know if this is important to know in a running application, i guess this will only occour in active development and the warnings should be enough then?

Comment (EinfachHans) [reply core/src/components/modal/modal.tsx:589]: Or `moveToBreakpoint`? what do you like most?

Comment (sean-perkins) [reply core/src/components/modal/modal.tsx:589]: Agreed, let's go with `setCurrentBreakpoint`. Also aligns with the internal behaviors of the function.

Comment (sean-perkins) [reply core/src/components/modal/modal.tsx:589]: My thought was around at the time of calling the function, being able to know if it actually changed the breakpoint value or not. 

Since `setBreakpoint` will skip/do nothing if the breakpoint matches the current breakpoint, there's no way for a developer to wire logic that is dependent on cases where it did nothing. 

I think this isn't required though. With the addition of the new events, developers can minimally wire logic that responds to when the breakpoint value is changed.

**Comment by EinfachHans** [comment]
@sean-perkins Changes applied 😊

**Review by sean-perkins — COMMENTED** [review]
We are getting close! Had one question. 

I need to sync with the team tomorrow, trying to figure out the best way to test/validate a breakpoint is transitioned to correctly. Currently added coverage for events emitting/breakpoint values being set correctly.

Also due to the way we overwrite the implementation of `ion-modal` in Angular, I need to add a small set of e2e tests to our Angular test app to validate event emissions. I should be able to finish this piece tonight.

**Review by sean-perkins — APPROVED** [review]
Did a fresh review based on the current state and I do not see any outstanding changes. I am going to add the team for review. They may come back with changes or additional suggestions. 

I opted for not documenting the new API extensively in our usage docs, since it's a public method API and not really a usage pattern. I'm open to ideas on how to document it, if we feel there is value in doing so. Currently I could not justify it over vs. people looking at the generated API docs and seeing `getBreakpoint` and `setCurrentBreakpoint`.

Thanks for the heavy lifting on this feature and being available for quick changes + answering questions via Discord. Really appreciate the help on this. This is such a cool feature!

**Review by liamdebeasi — CHANGES_REQUESTED** [review]

**Review by amandaesmith3 — CHANGES_REQUESTED** [review]

File: core/src/components/modal/modal.tsx [hunk 3]
```diff
@@ -270,6 +277,13 @@ export class Modal implements ComponentInterface, OverlayInterface {
     }
   }
 
+  @Watch('breakpoints')
+  breakpointsChanged(breakpoints: number[] | undefined) {
+    if (typeof breakpoints !== 'undefined') {
+      this.breakpoints = breakpoints.sort((a, b) => a - b);
+    }
+  }
+
```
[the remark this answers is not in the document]
sean-perkins: Correct, we want to sort the breakpoints in the event they are passed in an incorrect ordering. I do need to execute this function from `componentDidLoad`, good catch! 

As for the breakpoints not being mutable, is that a factor here? My understanding of that usage was to trigger re-renders from value changes, which we do not want to do here. 

I think I can also remove the `@Watch` decorator from this function for now. We will have a follow-up discussion post-merge, if breakpoints should be dynamic and if we need to re-create gestures, etc. So in the interim, that logic is not doing anything.

core/src/components/modal/gestures/sheet.ts
```diff
@@ -5,6 +5,32 @@ import { getBackdropValueForSheet } from '../utils';
...
+   *
+   * For breakpoint changes as a result of dynamically setting the value,
+   * this value should be the difference between the new and old breakpoint.
+   * For example:
+   * - breakpoints: [0, 0.25, 0.5, 0.75, 1]
+   * - Current breakpoint value is 1.
+   * - Setting the breakpoint to 0.25.
+   * - The offset value should be 0.75 (1 - 0.25).
+   */
+  breakpointOffset: number;
```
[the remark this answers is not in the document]
sean-perkins: This value is dynamic based on the execution context. When using `setCurrentBreakpoint`, we need to transition from the current breakpoint to the specified breakpoint. 

For the gesture behavior with the `canDismiss`, it needs to clamp to the nearest breakpoint. Since the value is different, it cannot be calculated internal to the `moveSheetToBreakpoint` function and needs to passed in/defined when calling the function.

core/src/components/modal/modal.tsx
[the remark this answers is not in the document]
amandaesmith3: I believe that the breakpoints being immutable would mean that the `@Watch` would never trigger under normal circumstances. If you're removing it, then we're good to go there 👍

core/src/components/modal/gestures/sheet.ts
```diff
@@ -113,6 +140,7 @@ export const createSheetGesture = (
      * allow for scrolling on the content.
      */
     const content = (detail.event.target! as HTMLElement).closest('ion-content');
+    currentBreakpoint = getCurrentBreakpoint()!;
```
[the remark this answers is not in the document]
sean-perkins: In all instances current breakpoint should be defined. This appears to be a side-effect of the types set on the current breakpoint. We can likely add a check before creating the sheet gesture to avoid the modifier though.

amandaesmith3 (approved):
This LGTM once those two comments are addressed 👍

core/src/components/modal/modal.tsx
```diff
@@ -623,6 +650,43 @@ export class Modal implements ComponentInterface, OverlayInterface {
     return eventMethod(this.el, 'ionModalWillDismiss');
   }
 
+  /**
+   * Move a sheet style modal to a specific breakpoint.
```
liamdebeasi, line 654: Can we add a note that the breakpoint must be defined in your `breakpoints` array?

core/src/components/modal/modal.tsx
```diff
@@ -623,6 +650,43 @@ export class Modal implements ComponentInterface, OverlayInterface {
...
+      return;
+    }
+
+    if (this.moveSheetToBreakpoint) {
+      this.moveSheetToBreakpoint({
+        breakpoint,
+        breakpointOffset: 1 - this.currentBreakpoint!,
+        canDismiss: this.canDismiss !== undefined && this.canDismiss !== true && this.breakpoints![0] === 0,
+      });
+    }
```
liamdebeasi, line 679:
```suggestion
    const { currentBreakpoint, moveSheetToBreakpoint } = this;

    if (currentBreakpoint === breakpoint) {
      return;
    }

    if (moveSheetToBreakpoint) {
      moveSheetToBreakpoint({
        breakpoint,
        breakpointOffset: 1 - this.currentBreakpoint!,
        canDismiss: this.canDismiss !== undefined && this.canDismiss !== true && this.breakpoints![0] === 0,
      });
    }
```

core/src/components/modal/modal.tsx
```diff
@@ -270,6 +277,12 @@ export class Modal implements ComponentInterface, OverlayInterface {
     }
   }
 
+  breakpointsChanged(breakpoints: number[] | undefined) {
+    if (typeof breakpoints !== 'undefined') {
```
liamdebeasi, line 281: Any particular reason for this? Doing `breakpoints !== undefined` should be sufficient

core/src/components/modal/modal.tsx
```diff
@@ -237,6 +239,11 @@ export class Modal implements ComponentInterface, OverlayInterface {
    */
   @Event({ eventName: 'ionModalDidDismiss' }) didDismiss!: EventEmitter<OverlayEventDetail>;
 
+  /**
+   * Emitted after the modal breakpoint has changed.
+   */
+  @Event() ionBreakpointDidChange!: EventEmitter<{breakpoint: number}>;
```
liamdebeasi, line 245: Do we need to expose an `ModalCustomEvent` interface with the target set to `HTMLIonModalElement`? 

Similar to what infinite scroll and other components do: https://github.com/ionic-team/ionic-framework/blob/main/core/src/components/infinite-scroll/readme.md#interfaces

core/src/components/modal/test/sheet/e2e.ts
```diff
@@ -117,3 +117,86 @@ test('input should not be focusable when backdrop is active', async () => {
...
+
+      const ionBreakpointDidChangeSpy = await modal.spyOnEvent('ionBreakpointDidChange');
+
+      await modal.callMethod('setCurrentBreakpoint', 0.5);
+      await modal.waitForEvent('ionBreakpointDidChange');
+
+      expect(ionBreakpointDidChangeSpy).toHaveReceivedEventTimes(1);
+    });
+
+    it('should emit ionBreakpointDidChange when breakpoint is set to 0', async () => {
```
liamdebeasi, line 189: Can we add a test to check that `ionBreakpointDidChange` is called when swiping down to breakpoint 0?

core/src/components/modal/test/sheet/e2e.ts
```diff
@@ -117,3 +117,86 @@ test('input should not be focusable when backdrop is active', async () => {
...
+
+  describe('setting the breakpoint to the same value', () => {
+
+    it('should do nothing', async () => {
+      const modal = await openModal(page, '#sheet-modal');
+
+      await modal.callMethod('setCurrentBreakpoint', 0.25);
+      await page.waitForChanges();
+
+      const breakpoint = await modal.callMethod('getCurrentBreakpoint');
```
liamdebeasi, line 170: Do we need `waitForChanges()`? I don't know if the test would catch this bug if it regressed. Might be worth intentionally breaking this to test.

core/src/components/modal/test/sheet/e2e.ts
```diff
@@ -117,3 +117,86 @@ test('input should not be focusable when backdrop is active', async () => {
...
+        }
+      });
+
+      modal = await openModal(page, '#sheet-modal');
+
+      await modal.callMethod('setCurrentBreakpoint', 0.01);
+      await page.waitForChanges();
+    });
+
+    it('should not change the breakpoint', async () => {
```
liamdebeasi, line 150: Can we add tests for the get/set methods on non-sheet modals?

liamdebeasi (changes requested):

core/src/components/modal/test/sheet/e2e.ts
sean-perkins: Confirmed this test isn't covering the behavior. I'm going to end up removing it, as we are validating other invalid inputs and I'll add a test to confirm that the breakpoint value updates when using a correct value.

core/src/components/modal/modal.tsx
```diff
@@ -95,7 +97,7 @@ export class Modal implements ComponentInterface, OverlayInterface {
    * array must be the value of the `initialBreakpoint` property.
    * For example: [0, .25, .5, 1]
    */
-  @Prop() breakpoints?: number[];
+  @Prop({ mutable: true }) breakpoints?: number[];
```
liamdebeasi, line 100: Should we have an internal `sortedBreakpoints` array instead of setting `mutable: true` here? We don't necessarily need to re-render the modal whenever the breakpoints are sorted.

core/src/components.d.ts
```diff
@@ -5294,6 +5302,10 @@ declare namespace LocalJSX {
           * Emitted after the modal has presented. Shorthand for ionModalWillDismiss.
          */
         "onDidPresent"?: (event: CustomEvent<void>) => void;
+        /**
+          * Emitted after the modal breakpoint has changed.
+         */
+        "onIonBreakpointDidChange"?: (event: CustomEvent<ModalBreakpointChangeEventDetail>) => void;
```
liamdebeasi, line 5308: Do we need to create a `ModalCustomEvent` interface that encapsulates this?

liamdebeasi (approved):
Good to merge once my other two comments are addressed.

Great job @EinfachHans and @sean-perkins! This is a great feature addition to Ionic.

core/src/components/modal/modal-interface.ts
```diff
@@ -33,4 +33,12 @@ export interface ModalAnimationOptions {
...
 }
 
-export interface ModalAttributes extends JSXBase.HTMLAttributes<HTMLElement> {}
+export interface ModalAttributes extends JSXBase.HTMLAttributes<HTMLElement> { }
+
+export interface ModalBreakpointChangeEventDetail {
+  breakpoint: number;
+}
+
+export interface ModalCustomEvent extends CustomEvent {
```
liamdebeasi, line 42: Ok sorry last comment: Can we also document this in the readme? See the infinite scroll readme for an example of how we document it.

Issue #24652, opened 2022-01-25
bug: router-outlet getRouteId() should returns the params


vicb:
### Prerequisites

- [X] I have read the [Contributing Guidelines](https://github.com/ionic-team/ionic-framework/blob/main/.github/CONTRIBUTING.md#creating-an-issue).
- [X] I agree to follow the [Code of Conduct](https://ionicframework.com/code-of-conduct).
- [X] I have searched for [existing issues](https://github.com/ionic-team/ionic-framework/issues) that already report this problem, without success.

### Ionic Framework Version

- [ ] v4.x
- [ ] v5.x
- [X] v6.x

### Current Behavior

`getRouteId()` should [returns the params](https://github.com/ionic-team/ionic-framework/blob/7d5c6afd180b0023de82f621ccab2421e61752ea/core/src/components/router-outlet/route-outlet.tsx#L154-L159) to be consistent with [`ion-nav`](https://github.com/ionic-team/ionic-framework/blob/7d5c6afd180b0023de82f621ccab2421e61752ea/core/src/components/nav/nav.tsx#L436-L444)

It will also enable fixing a bug (I'll create a separate one)

### Expected Behavior

Returns the params passed to `setRouteId()`.

### Steps to Reproduce

Call `getRouteId()` params are nor returned.

I'll create a PR with tests.

liamdebeasi:
Fixed via https://github.com/ionic-team/ionic-framework/pull/24656

PR #24645, opened 2022-01-24
fix(router): matchIDs now correctly accounts for all parameters

vicb:
## Pull request checklist

Please check if your PR fulfills the following requirements:
- [X] Tests for the changes have been added (for bug fixes / features)
- [X] Build (`npm run build`) was run locally and any changes were pushed
- [X] Lint (`npm run lint`) has passed locally and any fixes were made for failures

## Pull request type

 

 

Please check the type of change your PR introduces:
- [X] Bugfix
- [ ] Feature
- [ ] Code style update (formatting, renaming)
- [ ] Refactoring (no functional changes, no api changes)
- [ ] Build related changes
- [ ] Documentation content changes
- [ ] Other (please describe): 

## What is the current behavior?

Issue Number: fixes #24641

## What is the new behavior?

`<ion-route>` parameters and fixed segments are taken into account when matching IDs.

## Does this introduce a breaking change?

- [ ] Yes
- [X] No

PR #24656, opened 2022-01-26
Router outlet 1

vicb:
## Pull request checklist

## Pull request type

 

 

- [X] Bugfix
- [ ] Feature
- [ ] Build related changes

`getRouteId()` does not set the params

Issue Number: #24652

## What is the new behavior?

`getRouteId()` sets the params 

- [ ] Yes
- [X] No

## Other information

This PR builds on top of #24645, I'll rebase when it is merged.


sean-perkins (approved):
The code modifications look good, great work. My comment is non-blocking and more of a discussion point that other members of the team can add to, if they feel it needs a change.

core/src/components.d.ts
```diff
@@ -2243,7 +2243,7 @@ export namespace Components {
          */
         "animated": boolean;
         /**
-          * By default `ion-nav` animates transition between pages based in the mode (ios or material design). However, this property allows to create custom transition using `AnimateBuilder` functions.
+          * By default `ion-router-outlet` animates transition between pages based in the mode (ios or material design). However, this property allows to create custom transition using `AnimateBuilder` functions.
```
sean-perkins, line 2246: While technically the `ion-nav` component is the responsible implementation layer for performing the transitions between pages: https://github.com/ionic-team/ionic-framework/blob/main/core/src/components/nav/nav.tsx#L506

I do see how this documentation as a public API is confusing to implementers and the `ion-router-outlet` would be a more appropriate scope. 

I question how valuable the comment is altogether, as any consumer of the API is likely familiar with Ionic's support for iOS and Material design, as well as familiar with the routing transitions, if they are trying to access a property to customize them. The documentation could be improved with details surrounding how to build animations for routing instead of referencing existing behavior.

vicb: I'm all for changing this comment to "Allows to create custom transition" only.

What I did here was just to replace `ion-nav` by `ion-router-outlet`. However note that the standalone `ion-router-outlet` does not animate.

amandaesmith3 (approved):

core/src/components.d.ts
vicb: There is no reason to mention `ion-nav` (or anything related to it) here. You can use either a `ion-nav` or a `ion-router-outlet` as your outlet but there is absolutely no relation between them.

sean-perkins: I'm in favor of updating the comment to:
> This property allows to create custom transition using `AnimateBuilder` functions.

Also, I intended to start this comment thread in `router-outlet.tsx` instead of the auto-generated output. We'll need to update there & re-run the build command once more.

amandaesmith3: Sounds good to me 👍

liamdebeasi (approved):

liamdebeasi:
Merged. Thank you!

Issue #24489, opened 2021-12-27
bug: React-hook-Form v7 for ionic-react 6 IonInput props is not assignable


joeldacuma:
### Prerequisites

- [X] I have read the [Contributing Guidelines](https://github.com/ionic-team/ionic-framework/blob/main/.github/CONTRIBUTING.md#creating-an-issue).
- [X] I agree to follow the [Code of Conduct](https://ionicframework.com/code-of-conduct).
- [X] I have searched for [existing issues](https://github.com/ionic-team/ionic-framework/issues) that already report this problem, without success.

### Ionic Framework Version

- [ ] v4.x
- [ ] v5.x
- [X] v6.x

### Current Behavior

IonInput is not assigning types of props using register function of react-hook-form

### Expected Behavior

props of register function using react-hook-form should be assignable to IonInput

### Steps to Reproduce

import { useForm } from 'react-hook-form';

  const { control, handleSubmit, register, setValue } = useForm({
    defaultValues: {
      email: ''
    }
  });

  const {ref, onChange, ...props} = register('email', {
    required: true,
  })

```
<IonItem>
  <IonLabel>Email</IonLabel>
      <IonInput  {...props} />         
  </IonItem>
```

### Ionic Info

Ionic:

   Ionic CLI       : 6.18.1 (/Users/joeldacuma/.nvm/versions/node/v14.17.0/lib/node_modules/@ionic/cli)
   Ionic Framework : @ionic/react 6.0.1

Capacitor:

   Capacitor CLI      : 2.4.6
   @capacitor/android : 2.4.6
   @capacitor/core    : 2.4.6
   @capacitor/ios     : 2.4.6

Utility:

   cordova-res : not installed globally
   native-run  : not installed globally

System:

   NodeJS : v14.17.0
   npm    : 6.14.13
   OS     : macOS Monterey

sean-perkins:
Hello @joeldacuma thanks for reporting this issue.

Can you help explain to me the reasoning behind why `ion-input` should have the exact type signature as `react-hook-form`?
 
In your pseudo code you have left out what `props` is tied to, so I'm having a difficult time following the example. 

An example repository would be great to visualize the issue.

joeldacuma:
> Hello @joeldacuma thanks for reporting this issue.
> 
> Can you help explain to me the reasoning behind why `ion-input` should have the exact type signature as `react-hook-form`?
> 
[...]

PR #24657, opened 2022-01-26
fix(input): min/max compatibility with react-hook-form

sean-perkins:
## Pull request checklist

## Pull request type

 

 

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

The type signature of the `min` and `max` properties on `IonInput` do not align with the typings for `input` from Stencil's public runtime, which clashes with `react-hook-form` usage when using this syntax:

```jsx
<IonInput {...register('email')} type="email"></IonInput>
```

Issue Number: #24489

## What is the new behavior?

- `min` has a type of `string | number`
- `max` has a type of `string | number`
- Type signatures of `IonInput` now align with the JSX extended types from Stencil public runtime and `react-hook-form`

- [ ] Yes
- [x] No


liamdebeasi (commented):
A few concerns:

1. This changes the public API so would need to be released in a 6.x release.
2. Where do you see that the `max` and `min` attributes can be numbers in the HTML input spec? Looking at the spec, they appear to be strings and the browser converts them to numbers: https://html.spec.whatwg.org/multipage/input.html#the-min-and-max-attributes

sean-perkins:
> A few concerns:
> 
> 1. This changes the public API so would need to be released in a 6.x release.
> 2. Where do you see that the `max` and `min` attributes can be numbers in the HTML input spec? Looking at the spec, they appear to be strings and the browser converts them to numbers: https://html.spec.whatwg.org/multipage/input.html#the-min-and-max-attributes

Sorry HTML spec was probably the wrong specification. The min/max typings through JSX & Stencil are typed as `number | string`:
- https://github.com/sean-perkins/stencil/blob/main/src/declarations/stencil-public-runtime.ts#L952
- https://github.com/sean-perkins/stencil/blob/main/src/declarations/stencil-public-runtime.ts#L949

This aligns with the types that react-hook-router uses:
- https://github.com/react-hook-form/react-hook-form/blob/master/src/types/form.ts#L151-L152

I think it's fine to hold this until a 6.x, but no other alternative exists. Implementers are currently having to do the following with react-hook-form and Ionic React:
```jsx
<IonInput {...register('email') as any} type="email"></IonInput>
```

I'll update the description to better reflect where these type changes come from.

amandaesmith3 (approved):
This LGTM. I think it's closer to a bug fix than a feature re: patch or minor release, but I could be persuaded 🤷‍♀️

liamdebeasi (approved):
Oh I see. In that case, yes I agree this is a fix. It should have been that way to begin with since it is how Stencil's types are configured.

PR #24664, opened 2022-01-26
docs(router-outlet): clarify usage when not using a framework

vicb:
## Pull request checklist

## Pull request type

 

 

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes
- [X] Documentation content changes

## What is the new behavior?

The router outlet can be used without any framework (see the [test app](https://github.com/ionic-team/ionic-framework/tree/main/core/src/components/router-outlet/test/basic)) so the document was misleading and this PR removes the part saying otherwise.

- [ ] Yes
- [X] No

## Other information

The documentation is still not quite right because the standalone version of the router outlet is quite different from the versions available for the different framework. The standalone version is a simple outlet that only can load a component.

I am wondering how to document components that are different standalone vs framework. Do you have guidelines/examples ? In the meantime I think this PR is still valuable as I think most users would use the router outlet when using the ionic router because it's much more lightweight than a ion-nav.

The parts of the docs that are not applicable to the standalone version:

> but contains the logic for providing a stacked navigation

This is not true for the standalone router-outlet. It might be true for the framework version ?

The document events `ionView...` do not exists on the standalone router outlet.

> Although router outlet has methods for navigating around

This is not true for the standalone version of the router outlet.

### DIFF
--- a/core/src/components/router-outlet/readme.md
+++ b/core/src/components/router-outlet/readme.md
@@ -1,11 +1,8 @@
 # ion-router-outlet
 
-Router outlet is a component used in routing within an Angular, React, or Vue app. It behaves in a similar way to Angular's built-in router outlet component and Vue's router view component, but contains the logic for providing a stacked navigation, and animating views in and out.
-
-> Note: this component should only be used with Angular, React, and Vue projects. For vanilla or Stencil JavaScript projects, use [`ion-router`](../router) and [`ion-route`](../route).
+The router outlet behaves in a similar way to Angular's built-in router outlet component and Vue's router view component, but contains the logic for providing a stacked navigation, and animating views in and out.
 
 Although router outlet has methods for navigating around, it's recommended to use the navigation methods in your framework's router.
-
 
 ## Life Cycle Hooks
 
--- a/core/src/components/router/readme.md
+++ b/core/src/components/router/readme.md
@@ -7,9 +7,9 @@ The router is a component for handling routing inside vanilla and Stencil JavaSc
 Apps should have a single `ion-router` component in the codebase.
 This component controls all interactions with the browser history and it aggregates updates through an event system.
 
-`ion-router` is just a URL coordinator for the navigation outlets of ionic: `ion-nav` and `ion-tabs`.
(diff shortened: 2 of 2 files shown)


sean-perkins:
I'll add comments on the details I'm familiar with, welcome any corrections or additional comments from the team.

> I am wondering how to document components that are different standalone vs framework. Do you have guidelines/examples ?

We don't currently have a standard approach to documenting the difference between the framework router integrations and the core `ion-router-outlet` implementation. It appears we have historically documented each independently ([React](https://ionicframework.com/docs/react/navigation), [Vue](https://ionicframework.com/docs/vue/navigation)). This is likely because when implementing, developers are approaching from their framework and aren't starting at `ion-router-outlet` (although this is speculation). There are also certain components we simply override in a framework (i.e. Tabs) that isn't very clear to the developer. 

All this being said, we are open to suggestions on improvements or references to other examples where it's done well to draw inspiration from. I think code-level documentation should be geared towards maintainers understanding the connections and differences between the implementations and the docs can be a separate approach of either individual to each framework or another approach.

> but contains the logic for providing a stacked navigation

You are correct, the `ion-router-outlet` core implementation does not provide stacked navigation, but the wrappers in each framework implementation, implement their own version of this behavior. We have intentions to splitting out a `@ionic/router` package (name to be determined, but as an example), to centralize all the common logic between the framework routing integrations. 

> The document events ionView... do not exists on the standalone router outlet.

You are correct here as well. I think this docs example needs further clarification. The section for the events should be explicit that it relates to the framework wrappers and isn't available in the vanilla web component implementation. 

> Although router outlet has methods for navigating around, it's recommended to use the navigation methods in your framework's router.

I'm pretty confident this is referring to `ion-nav`, which could be confusing in the `ion-router-outlet` documentation. I would be ok with removing this line, as any framework developer is likely to use their framework's routing methods prior to using our custom API. Providing this documentation seems to provide information that could confuse implementers.

liamdebeasi (approved):

sean-perkins (approved):

amandaesmith3 (approved):

Issue #24666, opened 2022-01-27
bug: Parent outlet flickers when using nested router outlet


sean-perkins:
### Prerequisites

- [X] I have read the [Contributing Guidelines](https://github.com/ionic-team/ionic-framework/blob/main/.github/CONTRIBUTING.md#creating-an-issue).
- [X] I agree to follow the [Code of Conduct](https://ionicframework.com/code-of-conduct).
- [X] I have searched for [existing issues](https://github.com/ionic-team/ionic-framework/issues) that already report this problem, without success.

### Ionic Framework Version

- [ ] v4.x
- [ ] v5.x
- [X] v6.x

### Current Behavior

When using nested router outlets, the non-active outlet can flicker and render before the desired outlet is mounted.

For example with:
```tsx
<IonApp>
    <IonReactRouter>
      <Route path="/" component={Main} />
      <Route path="/overlay-hooks" component={OverlayHooks} />
      <Route path="/overlay-components" component={OverlayComponents} />
      <Route path="/tabs" component={Tabs} />
    </IonReactRouter>
  </IonApp>
```

When navigating directly to `/tabs`, `Main` will render for a brief moment and flicker before the `Tabs` component with it's own router outlet is rendered.

<video src="https://user-images.githubusercontent.com/13732623/151271387-b1e58716-5041-427c-9203-bd1430f75cce.mp4"></video>

This is a separate issue from the flickering issue when navigating between tabs.

### Expected Behavior

When navigating directly to `/tabs`, the `Main` component route should never display. The `Tabs` component should render without a flicker.

### Steps to Reproduce

Outlined above.

PR #24667, opened 2022-01-27
fix(react-router): remove page transition flicker on outlet mounting

sean-perkins:
## Pull request checklist

## Pull request type

 

 

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

On initial load, when using nested router outlets with React, another outlet can briefly display before the enter view & outlet has mounted.

Issue Number: #24666

## What is the new behavior?

- Page transitions that are deferred as a result of the outlet not being rendered yet, happen immediately after the outlet is mounted.
- Add note about using a deprecated private API from react-router (`computedMatch`) that is no longer available in v6.
- Removes `tabs` attribute when cloning the page element
  - Cannot find a single reference to why we have this code. Looks like carry over from the Angular router integration, which also has no functional code tied to it.

- [ ] Yes
- [x] No

## Other information

|Before|After|
|---|---|
|<video src="https://user-images.githubusercontent.com/13732623/151272903-e969a074-51cb-4e7d-8daa-c621c12d22b1.mp4"></video>|<video src="https://user-images.githubusercontent.com/13732623/151272877-170d1949-6e74-452f-8b0e-de2b416fa3ad.mp4"></video>

Note: I am refreshing the browser in both examples.

### DIFF
--- a/packages/react-router/src/ReactRouter/IonRouteInner.tsx
+++ b/packages/react-router/src/ReactRouter/IonRouteInner.tsx
@@ -9,6 +9,12 @@ export class IonRouteInner extends React.PureComponent<IonRouteProps> {
         path={this.props.path}
         exact={this.props.exact}
         render={this.props.render}
+        /**
+         * `computedMatch` is a private API in react-router v5 that
+         * has been removed in v6.
+         *
+         * This needs to be removed when we support v6.
+         */
         computedMatch={(this.props as any).computedMatch}
       />
     );
--- a/packages/react-router/src/ReactRouter/IonRouteInner.tsx
+++ b/packages/react-router/src/ReactRouter/IonRouteInner.tsx
@@ -14,6 +14,8 @@ export class IonRouteInner extends React.PureComponent<IonRouteProps> {
          * has been removed in v6.
          *
          * This needs to be removed when we support v6.
+         *
+         * TODO: FW-647
          */
         computedMatch={(this.props as any).computedMatch}
       />
--- a/packages/react-router/src/ReactRouter/StackManager.tsx
+++ b/packages/react-router/src/ReactRouter/StackManager.tsx
@@ -29,6 +29,8 @@ export class StackManager extends React.PureComponent<StackManagerProps, StackMa
     isInOutlet: () => true,
   };
 
+  private pendingPageTransition = false;
+
   constructor(props: StackManagerProps) {
     super(props);
     this.registerIonPage = this.registerIonPage.bind(this);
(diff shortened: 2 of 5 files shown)


packages/react-router/src/ReactRouter/StackManager.tsx
```diff
@@ -59,7 +62,18 @@ export class StackManager extends React.PureComponent<StackManagerProps, StackMa
...
+      /**
+       * The route outlet has not mounted yet. We need to wait for it to render
+       * before we can transition the page.
+       *
+       * Set a flag to indicate that we should transition the page after
+       * the component has updated.
+       */
+      this.pendingPageTransition = true;
+
+      this.setState({
```
sean-perkins, line 74: This may be redundant. Open to feedback for those that are more familiar with how React handles internal state changes.

rgoldiez: @sean-perkins - are you tossing `routeInfo` into state as a way to get another render so that `pendingPageTransition` can be evaluated?  Would it be clearer  to put `pendingPageTransition` into state (instead of an instance variable) and use it from state?

liamdebeasi: The `componentOnReady` helper may be of use here: https://github.com/ionic-team/ionic-framework/blob/main/core/src/utils/helpers.ts#L70

sean-perkins: I was trying to maintain as much previous code behavior, of passing the context of `routeInfo` from the original function call, back when the function should be handled. 

Although, I don't believe `routeInfo` can change between the time React component is first constructed and the first render cycle. So I believe the `setState` is just re-assigning the same value and can be removed. I did brief testing of removing that call and tests passed and behavior seemed appropriate. 

The important piece is just catching when a page transition happens on a component that hasn't mounted (typically from routing directly to a view in a nested outlet) and re-handling that when the component is available.

packages/react-router/src/ReactRouter/IonRouteInner.tsx
```diff
@@ -9,6 +9,12 @@ export class IonRouteInner extends React.PureComponent<IonRouteProps> {
         path={this.props.path}
         exact={this.props.exact}
         render={this.props.render}
+        /**
+         * `computedMatch` is a private API in react-router v5 that
+         * has been removed in v6.
+         *
+         * This needs to be removed when we support v6.
+         */
         computedMatch={(this.props as any).computedMatch}
```
amandaesmith3, line 18: Do we have a Jira ticket for this yet?

sean-perkins: Good call, creating a spike for this now. Will update the comment to include the task 👍

amandaesmith3 (approved):

packages/react-router/src/ReactRouter/StackManager.tsx
sean-perkins: I don't believe `componentOnReady` will be applicable here. The most common state of this happening, was when `routerOutlet` (the element ref) was `undefined`. 

From testing, it does not appear to be needed in the `componentDidUpdate` life cycle 👍

liamdebeasi (approved):

Issue #24107, opened 2021-10-22
bug: react, custom elements build flicker when switching tabs


milanharia:
### Prerequisites

- [X] I have read the [Contributing Guidelines](https://github.com/ionic-team/ionic-framework/blob/main/.github/CONTRIBUTING.md#creating-an-issue).
- [X] I agree to follow the [Code of Conduct](https://ionicframework.com/code-of-conduct).
- [X] I have searched for [existing issues](https://github.com/ionic-team/ionic-framework/issues) that already report this problem, without success.

### Ionic Framework Version

- [ ] v4.x
- [ ] v5.x
- [X] v6.x

### Current Behavior

When switching between tabs, for a split second a blank page loads and then the page mounts. Consequently, you can see a flashing effect as the content all appears from a blank page. This bug does not happen in beta 5.

### Expected Behavior

I expect the tab switching to happen instantly as it does in previous versions. I do not expect a blank page to momentarily appear.

### Steps to Reproduce

This issue occurs when you nest IonRouterOutlet's which is necessary when you want to only render the tab bar on certain pages (for example do not render on a login page).

### Code Reproduction URL

https://github.com/milanharia/tab-flashing-effect

### Ionic Info

Ionic:

   Ionic CLI       : 6.17.1 (/Users/milan/.nvm/versions/node/v14.17.6/lib/node_modules/@ionic/cli)
   Ionic Framework : @ionic/react 6.0.0-rc.0

Capacitor:

   Capacitor CLI      : 3.2.5
   @capacitor/android : not installed
   @capacitor/core    : 3.2.5
   @capacitor/ios     : 3.2.5

Utility:

   cordova-res : not installed globally
   native-run  : 1.5.0

System:

   NodeJS : v14.17.6 (/Users/milan/.nvm/versions/node/v14.17.6/bin/node)
   npm    : 7.24.2
   OS     : macOS Big Sur

### Additional Information

https://user-images.githubusercontent.com/76167627/138437686-e37f7b48-cb6c-42d5-a8b0-cf81eb581b41.mp4

sean-perkins:
@milanharia following up on your issue, this appears to be related to the usage of nested `IonRouterOutlet`.

When using nested `IonRouterOutlet` it is encouraged to use the `ionPage` property on the nested outlet, to help the router better transition the pages: https://github.com/ionic-team/ionic-framework/issues/20597#issuecomment-648468733

If you update your example repository `App.tsx` to:
```ts
...
[...]

PR #24672, opened 2022-01-28
fix(react): nested router outlets will not reanimate entered views

sean-perkins:
## Pull request checklist

## Pull request type

 

 

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

When navigating within a nested router outlet, the page containing the router outlet is transitioning when the view is already visible. The most common instance of this, is when using nested router outlets in combination with IonTabs. When navigating between the tabs, the page containing IonTabs is transitioning, resulting in a flickering display.

Issue Number: #24107

## What is the new behavior?

The logic for determining when a page transition should occur has been amended to account for the conditions that apply when you are either within a single router outlet or within a nested router outlet.

Navigating between tabs within a nested router outlet will no longer flicker, as all subsequent routing will not transition the page containing the router outlet.

For example, when navigating between `/tabs/tab1` to `/tabs/tab2`, only the view elements for `tab1` and `tab2` will transition. The view element containing the `/tabs` path and router outlet will not transition, as that view is already active in the viewport.

- [ ] Yes
- [x] No

## Other information

This change should have a noticeable performance improvement for React applications using nested router outlets. Prior to this PR, the page element was transitioning and causing multiple repaints, when the view was already active. This likely wasn't caught prior to the custom elements build, as Stencil's hydrated (`.hydrated`) class was overriding the styling from `ion-page-invisible`. 

<video src="https://user-images.githubusercontent.com/13732623/151607200-2d2c5ed4-555a-4fff-b9d2-65f1d4531d10.mp4"></video>


sean-perkins:
Looks like I need to address some issues with regards to failing tests. Need to validate that the criteria they are validating is accurate to expected behavior vs. testing the previous incorrect behavior.

sean-perkins:
Ok, only one outstanding test is failing, which I will address in the morning.

> [1]   2) Nested Outlets 2
> [1]        /nested-outlet2 > 
> [1]     Go to list from Home IonItem click >
> [1]     Item#1 IonItem Click >
> [1]     Item page should be visible > 
> [1]     Back >
> [1]     Home page should be visible

I believe this test is covering incorrect behavior. Now when clicking the "Back >" interaction, you will be transitioned to the list view that you clicked "Item #1 IonItem Click >" from, instead of going back two views to the Home page 🎉

sean-perkins:
Verified in the reproduction app (from the attached issue) and the Ionic React conference app.

sean-perkins:
Performance tests comparing v6.0.4 against this PR.

Navigating between tabs in a nested outlet (navigating from /tab-1, to /tab-2 and back to /tab-1)

|v6.0.4|PR|Device/Env|
|---|---|---|
|  |  |M1 Macbook Air 16gb ram / Chrome|
|  | | iPhone XS Max iOS 15.2.1 / Mobile Safari |

Noticeable improvements:
- 66% render improvement in chrome browser
- 51% painting improvement in mobile Safari

amandaesmith3 (approved):
Awesome work, and thanks again for the live walkthrough! 🎉

packages/react-router/src/ReactRouter/StackManager.tsx
```diff
@@ -97,16 +99,64 @@ export class StackManager extends React.PureComponent<StackManagerProps, StackMa
...
+           * we are either transitioning using parameterized routes to the same view
+           * or a parent router outlet is re-rendering as a result of React props changing.
+           *
+           * If the route data does not match the current path, the parent router outlet
+           * is attempting to transition and we cancel the operation.
+           */
+          if (enteringViewItem.routeData.match.url !== routeInfo.pathname) {
+            return;
+          }
+        }
```
liamdebeasi, line 127: Would it make sense to only check if `enteringViewItem.routeData.match.url !== routeInfo.pathname`?

If `enteringViewItem === leavingViewItem` but `enteringViewItem.routeData.match.url === routeInfo.pathname` then that would mean you went from `/page/1` to `/page/1` which I don't think is possible.

liamdebeasi (approved):
Great job! One non-blocking comment.

packages/react-router/src/ReactRouter/StackManager.tsx
sean-perkins: This specific check is to catch the difference in behavior between changing between tabs vs. pushing to another route that uses the same component. 

With tabs and our reliance on React's re-rendering as a result of props changing to trigger route transition, the `/tabs` component will reach this check when navigating between `/tabs/tab1` to `/tabs/tab2`. But, since the entering view item's rout date is `/tabs` and not `/tabs/tab1`, it will return early (skipping the unnecessary transition). 

If pushing to the same detail page, the entering and leaving view item will be the same, but the entering view item's route will be `/page/1` and the route info will be `/page/1`, so the transition will occur. 

I'm a little worried about opening up this check without the entering/leaving comparison, since this is the only case that is currently attempting to transition when it shouldn't. Ideally all of this goes away, if we move off React's re-rendering/prop changes for handling route transitions and follow closer suite to the Vue implementation.

liamdebeasi: Makes sense, let's keep it as is 👍

shadowzack:
I have tried using version 6.0.4 but still got the same problem

sean-perkins:
@shadowzack the problem still exists in v6.0.4. The comparison screenshots are comparing the incorrect behavior vs. the corrected behavior. 

You need to upgrade to 6.0.5 or later: https://github.com/ionic-team/ionic-framework/blob/main/CHANGELOG.md#605-2022-02-02

shadowzack:
I have just tried version 6.0.5 and the latest 6.0.13, all have the same problem

shadowzack:
I have tried to build a test app with the latest ionic core 6.0.13, and the app works as indented with no flickering.
can an inner IonRouterOutlet inside another IonRouterOutlet effect the transition?
i have both tabs and a side menu just like the ionic react conference app, but even after removing the side menu the flickering still exists.
any ideas on why this is happing?

sean-perkins:
If I recall, I verified this issue against the conference app and noticed an improvement to the transition between tab items (nested outlets). The only issue currently that I am aware of this is causing flicker transitions is fallback routes: https://github.com/ionic-team/ionic-framework/issues/24855.

If you are able to pull in the similarities between your application and a new project, to isolate the reproduction, I can give a better recommendation. Nested outlets are unfortunately complex to handle page transitions and mounting/unmounting with how we replicate native page transitions.

shadowzack:
<IonApp>
    <IonReactRouter>
      <IonSplitPane contentId="main">
        <Menu />
      {/*   <IonRouterOutlet id="main"> */}
          <Route path="/" exact>
            <Redirect to="/login" />
          </Route>

          <Route path="/login" exact>
            <Login />
          </Route>

          <IonTabs>
            <IonRouterOutlet>
              <Route path="/notifications" exact>
                <TabNotifications />
              </Route>
              <Route path="/home" exact>
                <TabHome />
              </Route>
              <Route path="/chats" exact>
                <TabChats />
              </Route>
              <Route path="/operations" exact>
                <TabList />
              </Route>
              <Route path="/profile" exact>
                <TabProfile />
              </Route>
              <Route path="/support" exact>
                <TabSupport />
              </Route>
            </IonRouterOutlet>

            <IonTabBar slot="bottom">
              <IonTabButton tab="tabProfile" href="/profile">
                <IonIcon icon={personOutline} />
                <IonLabel>החשבון שלי</IonLabel>
              </IonTabButton>
              <IonTabButton tab="tabNotifications" href="/notifications">
                <IonIcon icon={notificationsOutline} />
                <IonLabel>התראות</IonLabel>
              </IonTabButton>
              <IonTabButton tab="tabHome" href="/home">
                <IonIcon icon={homeOutline} />
                <IonLabel>דף הבית</IonLabel>
              </IonTabButton>
              <IonTabButton tab="tabOperations" href="/operations">
                <IonIcon icon={listOutline} />
                <IonLabel>פעולות</IonLabel>
              </IonTabButton>
              <IonTabButton tab="tabChats" href="/chats">
                <IonIcon icon={chatbubbleOutline} />
                <IonLabel>הודעות</IonLabel>
              </IonTabButton>
            </IonTabBar>
          </IonTabs>
        {/* </IonRouterOutlet> */}
      </IonSplitPane>
    </IonReactRouter>
  </IonApp>


this is the code above, after running the code without the outer IonRouterOutlet the tabs transition without flickering. i think this might be related to the fact that the side menu and the tabs have the same paths(for example home using tabs and also using the side menu)

shadowzack:
I have downloaded the ionic-react-conference-app, even without adding anything just running the code you can see the flickering when moving between tabs. can you try to fix it I think fixing it will help me and the anyone stuck in this problem.

shadowzack:
I have migrated back to ionic 5 until this issue is resolved, TIP - you can migrate back using this command:-
npm install @ionic/react@v5-lts @ionic/react-router@v5-lts

PR #24681, opened 2022-02-01
fix(angular, react,  vue): overlays no longer throw errors when used inside tests

### LINKED ISSUE #24549: bug: action sheet controller referencing ActionSheet before initialized
### Prerequisites

### Ionic Framework Version

- [ ] v4.x
- [ ] v5.x
- [X] v6.x

### Current Behavior

Updated from ionic 5 to 6 and throws an error that has been reported by several users.

### Expected Behavior

Should work on as it did before the update

### Steps to Reproduce

Start the app and try to use a select on any page.

### Code Reproduction URL

https://github.com/popcorn245/ionic6-select-bug

### Ionic Info

Ionic:

   Ionic CLI : 5.4.16

Utility:

   cordova-res : not installed
   native-run  : not installed

System:

   NodeJS : v16.13.1
   npm    : 8.1.2
   OS     : Windows 10

### Additional Information

Related to this issue.

https://github.com/ionic-team/ionic-framework/issues/24448#event-5852991367

### LINKED ISSUE #24590: bug: Angular 13 Jest Tests fail while opening overlays: Cannot set property '$lazyInstance$' of undefined (present is not a function)
### Prerequisites

### Ionic Framework Version

- [ ] v4.x
- [ ] v5.x
- [X] v6.x

### Current Behavior

The error occurs when an alert or modal will be presented or any other function is called (e.g. onDidDismiss). Everything works fine when testing it in the browser, but if we run a jest test the following error is thrown:

```
Cannot set property '$lazyInstance$' of undefined
TypeError: Cannot set property '$lazyInstance$' of undefined
    at Object.registerInstance (/Users/.../node_modules/@ionic/core/dist/cjs/index-8138ba57.js:1988:90)
    at new Alert (/Users/.../node_modules/@ionic/core/dist/cjs/overlays-b91eb626.js:432:13)
    at Function.invokeTheCallbackFunction.construct (/Users/.../node_modules/jsdom/lib/jsdom/living/generated/CustomElementConstructor.js:27:30)
    at createElement (/Users/...node_modules/jsdom/lib/jsdom/living/helpers/create-element.js:204:33)
    at DocumentImpl.createElement (/Users/...node_modules/jsdom/lib/jsdom/living/nodes/Document-impl.js:719:12)
    at Document.createElement (/Users/.../node_modules/jsdom/lib/jsdom/living/generated/Document.js:261:60)
    at /Users/.../node_modules/@ionic/core/dist/cjs/overlays-b91eb626.js:4073:32
    at processTicksAndRejections (internal/process/task_queues.js:93:5)
```

We dig into this issue and see that some methods are not defined. We can create an alert with ```alert.create({})```
but methods like `present, onDidDismiss` are not defined.

### Expected Behavior

The modals or alerts can be used in jest test like in the browser. We can't reproduce this issue with ionic 5.

### Steps to Reproduce

We have setup a very plain Angular 13 project with the angular cli. We added our Jest configuration to run the unit tests with Jest and added @ionic/angular 6.0.2

Everything is working fine except for ```alert.create().then(alert => alert.present())```

### Code Reproduction URL

https://github.com/tobiloeb/ionic6-jest-issue

### Ionic Info

Ionic:

   Ionic CLI : 5.4.16

Utility:

   cordova-res : not installed
   native-run  : not installed

System:

   NodeJS : v14.16.1
   npm    : 6.14.12
   OS     : macOS Big Sur

### Additional Information

First we thought it is a jest problem, but everything is working fin except for the ionic overlays. And we found a very similar issue for ionic/vue: #24503 
In our case we already import the AlertController from '@ionic/angular' and not from core.

liamdebeasi:
## Pull request checklist

## Pull request type

 

 

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

Issue Number: resolves #24549, resolves #24590

In order to get custom elements defined when using controllers, we import the source component which would trick Stencil into including the compiled components in the overlays file. From there, we would register those compiled components with the Custom Elements Registry.

Unfortunately, this resulted in two issues:

1. Action Sheet was being initialized after it was referenced by its controller (https://github.com/ionic-team/ionic-framework/issues/24549).
2. Registering the components this way assumed that everyone was using the custom elements build. This resulted in controllers breaking for Angular users (https://github.com/ionic-team/ionic-framework/issues/24590). When the controller was created, the custom elements version of the overlay was created before the lazy loaded version was created. However, Stencil was expecting the lazy loaded version as it contained additional information it needed.

## What is the new behavior?

- Rather than try and build on the previous bandaid, I feel it would be better to have the React and Vue integrations register their over overlay components.
- This allows Angular to work with lazy loaded and React/Vue to work with custom elements.
- When Angular switches over to use custom elements, it will need to do something similar to what is done for React/Vue in this PR.
- The one tradeoff is people using @ionic/core and custom elements build will need to import and define the overlay component before they can use the controller. However, this aligns with what is expected when using core + CE build anyways. Developers need to call `initialize()` when using core directly, but this will be called for them when using a framework integration.

- [ ] Yes
- [x] No

## Other information

Dev Build: 6.0.5-dev.1643732746.a377f15

### DIFF
--- a/core/README.md
+++ b/core/README.md
@@ -50,7 +50,10 @@ Below is an example of importing `ion-badge`, and initializing Ionic so it is ab
 import { defineCustomElement } from "@ionic/core/components/ion-badge.js";
 import { initialize } from "@ionic/core/components";
 
+// Initializes the Ionic config and `mode` behavior
 initialize();
+
+//  Defines the `ion-badge` web component
 defineCustomElement();
 ```
 
@@ -64,7 +67,10 @@ For example, if you wanted to use `ion-modal`, you would do the following:
 import { defineCustomElement } from "@ionic/core/components/ion-modal.js";
 import { initialize } from "@ionic/core/components";
 
+// Initializes the Ionic config and `mode` behavior
 initialize();
+
+//  Defines the `ion-modal` and child `ion-backdrop` web components.
 defineCustomElement();
 ```
 
--- a/core/README.md
+++ b/core/README.md
@@ -44,19 +44,49 @@ The `@ionic/core` package can be used in simple HTML, or by vanilla JavaScript w
 
 In addition to the default, self lazy-loading components built by Stencil, this package also comes with each component exported as a stand-alone custom element within `@ionic/core/components`. Each component extends `HTMLElement`, and does not lazy-load itself. Instead, this package is useful for projects already using a bundler such as Webpack or Rollup. While all components are available to be imported, the custom elements build also ensures bundlers only import what's used, and tree-shakes any unused components.
 
-Below is an example of importing `ion-toggle`, and initializing Ionic so it's able to correctly load the "mode", such as Material Design or iOS. Additionally, the `initialize({...})` function can receive the Ionic config.
+Below is an example of importing `ion-badge`, and initializing Ionic so it is able to correctly load the "mode", such as Material Design or iOS. Additionally, the `initialize({...})` function can receive the Ionic config.
 
 ```typescript
-import { IonBadge } from "@ionic/core/components/ion-badge";
+import { defineCustomElement } from "@ionic/core/components/ion-badge.js";
 import { initialize } from "@ionic/core/components";
 
 initialize();
-
-customElements.define("ion-badge", IonBadge);
+defineCustomElement();
 ```
 
-Notice how `IonBadge` is imported from `@ionic/core/components/ion-badge` rather than just `@ionic/core/components`. Additionally, the `initialize` function is imported from `@ionic/core/components` rather than `@ionic/core`. All of this helps to ensure bundlers do not pull in more code than is needed.
+Notice how we import from `@ionic/core/components` as opposed to `@ionic/core`. This helps bundlers pull in only the code that is needed.
 
+In the example above, you could also import `IonBadge` and call `customElements.define('ion-badge', IonBadge)`, but we recommend importing the `defineCustomElement` function and calling that instead. This `defineCustomElement` function will automatically define the component as well as any child components that may be required.
+
+For example, if you wanted to use `ion-modal`, you would do the following:
+
+```typescript
+import { defineCustomElement } from "@ionic/core/components/ion-modal.js";
+import { initialize } from "@ionic/core/components";
+
+initialize();
+defineCustomElement();
+```
+
+The `defineCustomElement` function will define `ion-modal`, but it will also define `ion-backdrop`, which is a component that `ion-modal` uses internally.
+
+### Using Overlay Controllers
+
+When using an overlay controller, developers will need to define the overlay component before it can be used. Below is an example of using `modalController`:
+
+```typescript
+import { defineCustomElement } from '@ionic/core/components/ion-modal.js';
+import { initialize, modalController } from '@ionic/core/components';
+
+initialize();
+defineCustomElement();
+
+const showModal = async () => {
+  const modal = await modalController.create({ ... });
+  
+  ...
+}
+```
 
 ## How to contribute
 
--- a/core/README.md
+++ b/core/README.md
@@ -59,7 +59,7 @@ defineCustomElement();
 
 Notice how we import from `@ionic/core/components` as opposed to `@ionic/core`. This helps bundlers pull in only the code that is needed.
 
-In the example above, you could also import `IonBadge` and call `customElements.define('ion-badge', IonBadge)`, but we recommend importing the `defineCustomElement` function and calling that instead. This `defineCustomElement` function will automatically define the component as well as any child components that may be required.
+The `defineCustomElement` function will automatically define the component as well as any child components that may be required.
 
 For example, if you wanted to use `ion-modal`, you would do the following:
 
--- a/packages/react/src/hooks/useController.ts
+++ b/packages/react/src/hooks/useController.ts
@@ -12,7 +12,8 @@ interface OverlayBase extends HTMLElement {
 
 export function useController<OptionsType, OverlayType extends OverlayBase>(
   displayName: string,
-  controller: { create: (options: OptionsType) => Promise<OverlayType> }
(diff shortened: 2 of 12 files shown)


core/README.md
```diff
@@ -44,19 +44,49 @@ The `@ionic/core` package can be used in simple HTML, or by vanilla JavaScript w
...
+```typescript
+import { defineCustomElement } from "@ionic/core/components/ion-modal.js";
+import { initialize } from "@ionic/core/components";
+
+initialize();
+defineCustomElement();
 ```
 
-Notice how `IonBadge` is imported from `@ionic/core/components/ion-badge` rather than just `@ionic/core/components`. Additionally, the `initialize` function is imported from `@ionic/core/components` rather than `@ionic/core`. All of this helps to ensure bundlers do not pull in more code than is needed.
+The `defineCustomElement` function will define `ion-modal`, but it will also define `ion-backdrop`, which is a component that `ion-modal` uses internally.
```
amandaesmith3, line 71: Is this still the case? Looks like that piece was removed from overlays.ts, but maybe I'm missing how it's being done now.

packages/react/src/hooks/useController.ts
```diff
@@ -25,6 +26,9 @@ export function useController<OptionsType, OverlayType extends OverlayBase>(
       if (overlayRef.current) {
         return;
       }
+
+      defineCustomElement();
```
amandaesmith3, line 30: Does this need to be done for every present, or can it be called when the controller is created instead? (Ditto for useOverlay)

amandaesmith3 (commented):
Had a couple questions below :eyes: Also, might be good to set up a tech debt ticket for Angular if you haven't done so already.

core/README.md
liamdebeasi: Yes. When you use an overlay controller in Vue/React, we automatically call `defineCustomElement` for the overlay you are trying to use. We do not need to do this for Angular as that integration still uses the lazy loaded version of Ionic.

If you are using @ionic/core directly, you will need to import and call `defineCustomElement` yourself before using the controller. I mention this a bit further down in the `core/README.md` file.

liamdebeasi:
We have a ticket for the Angular custom elements build in FW-191, so I made sure to mention this change in that ticket as well.

core/README.md
amandaesmith3: Ah sorry, I'm wondering about the piece about `ion-backdrop`. Are children components also still handled? I know Sean had to add those manually to the `define` calls to fix some reference errors.

liamdebeasi: Yeah - The `defineCustomElement` function for `ion-modal` will define `ion-backdrop`. Here is what the `defineCustomElement` function for `ion-modal` looks like:

```js
function defineCustomElement$3() {
  if (typeof customElements === "undefined") {
    return;
  }
  const components = ["ion-modal", "ion-backdrop"];
  components.forEach(tagName => { switch (tagName) {
    case "ion-modal":
      if (!customElements.get(tagName)) {
        customElements.define(tagName, Modal);
      }
      break;
    case "ion-backdrop":
      if (!customElements.get(tagName)) {
        defineCustomElement$9();
      }
      break;
  } });
}

```

packages/react/src/hooks/useController.ts
liamdebeasi: My one concern is that this may introduce a sideeffect, but I can definitely check and get back to you.

core/README.md
amandaesmith3: Aha, got it. Makes sense 👍

sean-perkins (changes requested):

packages/react/src/hooks/useIonActionSheet.ts
```diff
@@ -1,4 +1,5 @@
 import { ActionSheetButton, ActionSheetOptions, actionSheetController } from '@ionic/core/components';
+import { defineCustomElement } from '@ionic/core/components/ion-modal.js';
```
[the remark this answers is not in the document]
liamdebeasi: Fixed, thanks!

core/README.md
```diff
@@ -44,19 +44,49 @@ The `@ionic/core` package can be used in simple HTML, or by vanilla JavaScript w
...
 
+### Using Overlay Controllers
+
+When using an overlay controller, developers will need to define the overlay component before it can be used. Below is an example of using `modalController`:
+
+```typescript
+import { defineCustomElement } from '@ionic/core/components/ion-modal.js';
+import { initialize, modalController } from '@ionic/core/components';
+
+initialize();
```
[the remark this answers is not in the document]
liamdebeasi: Done

core/README.md
```diff
@@ -44,19 +44,49 @@ The `@ionic/core` package can be used in simple HTML, or by vanilla JavaScript w
...
+import { defineCustomElement } from "@ionic/core/components/ion-badge.js";
 import { initialize } from "@ionic/core/components";
 
 initialize();
+defineCustomElement();
+```
+
+Notice how we import from `@ionic/core/components` as opposed to `@ionic/core`. This helps bundlers pull in only the code that is needed.
+
+In the example above, you could also import `IonBadge` and call `customElements.define('ion-badge', IonBadge)`, but we recommend importing the `defineCustomElement` function and calling that instead. This `defineCustomElement` function will automatically define the component as well as any child components that may be required.
```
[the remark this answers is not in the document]
liamdebeasi: Agreed -- I removed reference to `IonBadge` in favor of `defineCustomElement`

packages/react/src/hooks/useController.ts
liamdebeasi: Moving the `defineCustomElement` call to the root of the useOverlay/useController function does not appear to cause a side effect, so I made that change.

No functional difference, but the overhead of calling `defineCustomElement` multiple times has been removed.

amandaesmith3 (approved):

sean-perkins (approved):

PR #24697, opened 2022-02-02
fix(modal, popover): quickly opening modal/popover no longer presents duplicates

liamdebeasi:
## Pull request checklist

## Pull request type

 

 

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

Given a button that calls `modalController.create` and `modalController.present`, double clicking it currently results in 2 modals being presented. This is technically the correct behavior, but it is undesirable to some developers.

In Ionic 5, the modals/popovers would receive `pointer-events` the moment the presenting animation began, so this double modal behavior never happened. On the second click, the modal would receive the click causing the 2nd modal to not get presented.

In Ionic 6, modals/popovers did not receive pointer events by default causing the behavior to reproduce more easily. The reason why it changed was due to a workaround to get inline modals to behave properly. However,

## What is the new behavior?

- Modals/popovers receive pointer events by default. When a modal/popover is initially added to the DOM, it gets ` display: none` so in that case pointer events are not received. This behavior only applies when the modal/popover is set to be presented.

- [ ] Yes
- [x] No

## Other information

Note: I wasn't able to create an automated test that wasn't flaky. If you have ideas for how I can better automate a test for this, let me know!

### DIFF
--- a/core/src/components/modal/modal.scss
+++ b/core/src/components/modal/modal.scss
@@ -48,13 +48,6 @@
   outline: none;
 
   contain: strict;
-
-  pointer-events: none;
-}
-
-:host(.modal-interactive) .modal-wrapper,
-:host(.modal-interactive) ion-backdrop {
-  pointer-events: auto;
 }
 
 :host(.overlay-hidden) {
--- a/core/src/components/modal/modal.tsx
+++ b/core/src/components/modal/modal.tsx
@@ -587,7 +587,7 @@ export class Modal implements ComponentInterface, OverlayInterface {
 
     const showHandle = handle !== false && isSheetModal;
     const mode = getIonMode(this);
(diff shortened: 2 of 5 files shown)


sean-perkins (approved):
Changes look good 👍  cannot think of a way to simulate that wouldn't be flaky, besides possibly opening a modal/popover, manually changing it's appearance to be not visible and then simulating the pointer events into the state that would align with double clicking the button that opens a modal.

amandaesmith3 (approved):

agencia-estado:
still happen in the first time i click fast to open a popover. After that, it works

amandaesmith3:
@agencia-estado Please open a new issue with a repo we can use to reproduce the problem.

Issue #24699, opened 2022-02-02
bug: Action sheet safe area is incorrect on MD device


sean-perkins:
### Prerequisites

- [X] I have read the [Contributing Guidelines](https://github.com/ionic-team/ionic-framework/blob/main/.github/CONTRIBUTING.md#creating-an-issue).
- [X] I agree to follow the [Code of Conduct](https://ionicframework.com/code-of-conduct).
- [X] I have searched for [existing issues](https://github.com/ionic-team/ionic-framework/issues) that already report this problem, without success.

### Ionic Framework Version

- [ ] v4.x
- [ ] v5.x
- [X] v6.x

### Current Behavior

While safe area is applied to the action sheet to make the last action in the group accessible on a MD device, the background container does not extend to the bevel of the screen.

Safe area was first introduced in this PR #24176

![Screen Shot 2022-02-02 at 1 12 42 PM](https://user-images.githubusercontent.com/13732623/152212940-5bb08008-edb6-44ea-851b-7fae72a3b9f2.png)

### Expected Behavior

Safe area should be applied to the last group in the action sheet, to push the background to the bottom edge and push the action sheet group to be accessible.

![Screen Shot 2022-02-02 at 1 13 42 PM](https://user-images.githubusercontent.com/13732623/152213103-5ef90a27-8cc8-43bc-bbb5-8eefaf1562cd.png)

### Steps to Reproduce

1. Visit the Action Sheet docs: https://ionicframework.com/docs/api/action-sheet
2. Toggle MD
3. Open the action sheet
4. See that the action sheet does not go to the bottom of the screen

### Code Reproduction URL

https://ionicframework.com/docs/api/action-sheet

PR #24700, opened 2022-02-02
fix(action-sheet): background includes safe area margin

sean-perkins:
## Pull request checklist

## Pull request type

 

 

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

The safe area is being applied to the action sheet container, which does not have the background appearance applied to it. This pushes the container up, but leaves an empty space between the last group in the action sheet and the bottom bevel of the screen.

Issue Number: #24699

## What is the new behavior?

Safe are padding is applied to the last group of the action sheet, which has the background appearance applied to it. This will extend the container's visual appearance to the bottom bevel on an Android device.

- [ ] Yes
- [x] No

### DIFF
--- a/core/src/components/action-sheet/action-sheet.scss
+++ b/core/src/components/action-sheet/action-sheet.scss
@@ -67,7 +67,6 @@
 
 .action-sheet-wrapper {
   @include position(null, 0, 0, 0);
-  @include margin(var(--ion-safe-area-top, 0), auto, 0, auto);
   @include transform(translate3d(0, 100%, 0));
 
   display: block;
--- a/core/src/components/action-sheet/action-sheet.scss
+++ b/core/src/components/action-sheet/action-sheet.scss
@@ -67,7 +67,7 @@
 
 .action-sheet-wrapper {
   @include position(null, 0, 0, 0);
-  @include margin(var(--ion-safe-area-top, 0), auto, var(--ion-safe-area-bottom, 0), auto);
+  @include margin(var(--ion-safe-area-top, 0), auto, 0, auto);
   @include transform(translate3d(0, 100%, 0));
 
   display: block;
--- a/core/src/components/action-sheet/action-sheet.ios.scss
+++ b/core/src/components/action-sheet/action-sheet.ios.scss
@@ -22,6 +22,13 @@
   text-align: $action-sheet-ios-text-align;
 }
 
+// iOS Action Sheet Wrapper
+// ---------------------------------------------------
+
+.action-sheet-wrapper {
+  @include margin(var(--ion-safe-area-top, 0), auto, var(--ion-safe-area-bottom, 0), auto);
+}
+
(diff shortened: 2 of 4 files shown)


core/src/components/action-sheet/action-sheet.scss
```diff
@@ -67,7 +67,7 @@
 
 .action-sheet-wrapper {
   @include position(null, 0, 0, 0);
-  @include margin(var(--ion-safe-area-top, 0), auto, var(--ion-safe-area-bottom, 0), auto);
+  @include margin(var(--ion-safe-area-top, 0), auto, 0, auto);
```
liamdebeasi, line 70: This will break the safe area for iOS. The action sheet on iOS should move above the safe area, while the MD action sheet's content should move above the safe area, but the content should extend below the safe area.

I think what was wrong in my previous patch was that iOS safe area should use margin, but MD safe area should use padding.

It might make sense to revert https://github.com/ionic-team/ionic-framework/pull/24176/files and then add 

```
.action-sheet-wrapper {
  @include padding(var(--ion-safe-area-top, 0), auto, var(--ion-safe-area-bottom, 0), auto);
}
```

Or something to the safe are in MD mode

liamdebeasi (changes requested):

core/src/components/action-sheet/action-sheet.scss
sean-perkins: Good catch 👍  Although the padding/margin cannot be applied to the `action-sheet-wrapper` in MD, as it will increase the container size (good), but won't increase the container space the background color is applied to (bad). 

Likely we will have to do specific styles in iOS vs. MD stylesheets. As iOS pushes the space from the bottom, but doesn't have a background applied.

sean-perkins:
Still need to verify on device, will re-request reviews after I get a chance to do that. Few meetings first ⏳

sean-perkins:
Ok, confirmed this maintains iOS visual behavior and fixes the MD display. We may want to consider visual regression tests in the future that simulate safe areas 🤔

liamdebeasi:
We should already have a safe area test: https://github.com/ionic-team/ionic-framework/blob/main/core/src/components/action-sheet/test/basic/index.html#L50-L55

The visual regression URL isn't super visible anymore, but here's the diff for your latest commit: https://screenshot.ionicframework.com/7315e01/bf6bf00

liamdebeasi:
Oh scratch that, we don't have an E2E test that actually opens the action sheet. Could you add one?

sean-perkins:
> Oh scratch that, we don't have an E2E test that actually opens the action sheet with added safe area. Could you add one?

@liamdebeasi looks like the `/app/test/safe-area/` covers this behavior 🤔

amandaesmith3 (approved):

liamdebeasi (approved):

Issue #24709, opened 2022-02-03
bug: wrong type hint for Popover Controller alignment prop  


treighmawaka:
### Prerequisites

- [X] I have read the [Contributing Guidelines](https://github.com/ionic-team/ionic-framework/blob/main/.github/CONTRIBUTING.md#creating-an-issue).
- [X] I agree to follow the [Code of Conduct](https://ionicframework.com/code-of-conduct).
- [X] I have searched for [existing issues](https://github.com/ionic-team/ionic-framework/issues) that already report this problem, without success.

### Ionic Framework Version

- [ ] v4.x
- [ ] v5.x
- [X] v6.x

### Current Behavior

When creating a Popover Controller, the `alignment` option of the `popoverController.create()` method has the wrong type hint.  

### Expected Behavior

The `PopoverOptions.align` prop should be `PopoverOptions.alignment` as [noted in the docs ](https://ionicframework.com/docs/api/popover#alignment-1)

### Steps to Reproduce

```ts
async function openPopover(ev: Event) {
  const popover = await popoverController.create({
   // ...
    align: "center", // ❌ this is wrong, but passes type validation
    //alignment: "center", // ✅ this is valid, but triggers a type error
  });

  await popover.present();
}
```

### Code Reproduction URL

https://stackblitz.com/edit/vitejs-vite-pudsq6?file=src/App.vue

### Ionic Info

Ionic:

   Ionic CLI : 6.16.3

Utility:

   cordova-res                          : 0.15.3
   native-run (update available: 1.5.0) : 1.3.0

System:

   NodeJS : v16.13.1
   npm    : 8.1.2
   OS     : macOS Big Sur

  ─────────────────────────────────────────────────

     Ionic CLI update available: 6.16.3 → 6.17.1
          Run npm i -g @ionic/cli to update

sean-perkins:
@treighmawaka thanks for reporting this issue!

I've updated the title, since this problem stems from the interfaces from our core package.

PR #24711, opened 2022-02-03
fix(popover): use alignment with popover options

sean-perkins:
## Pull request checklist

## Pull request type

 

 

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

The types for `PopoverOptions` specify using `align` to set the `alignment` prop on `ion-popover` when using the popover controller. This does not actually occur. 

Issue Number: #24709

## What is the new behavior?

Updates the `PopoverOptions` to match the `ion-popover` prop name. This means the type will recommend using `alignment` and no longer error on the developer.

https://github.com/ionic-team/ionic-framework/blob/2a27befe463832b9ca7709ba22421abbdaa4cfa4/core/src/components/popover/popover.tsx#L193

- [ ] Yes
- [x] No

Note: The type was incorrect, so developers could be referencing the incorrect type that does no functional behavior.

### DIFF
--- a/core/src/components/popover/popover-interface.ts
+++ b/core/src/components/popover/popover-interface.ts
@@ -29,7 +29,7 @@ export interface PopoverOptions<T extends ComponentRef = ComponentRef> {
   dismissOnSelect?: boolean;
   reference?: PositionReference;
   side?: PositionSide;
-  align?: PositionAlign;
+  alignment?: PositionAlign;
   arrow?: boolean;
 
   trigger?: string;
--- a/core/src/components/popover/readme.md
+++ b/core/src/components/popover/readme.md
@@ -107,7 +107,7 @@ interface PopoverOptions {
   dismissOnSelect?: boolean;
   reference?: PositionReference;
   side?: PositionSide;
-  align?: PositionAlign;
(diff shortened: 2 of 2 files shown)


core/src/components/popover/popover-interface.ts
```diff
@@ -29,7 +29,7 @@ export interface PopoverOptions<T extends ComponentRef = ComponentRef> {
   dismissOnSelect?: boolean;
   reference?: PositionReference;
   side?: PositionSide;
-  align?: PositionAlign;
+  alignment?: PositionAlign;
```
liamdebeasi, line 32: You will also need to update the interface in `README.md` so the changes get reflected on the docs: https://ionicframework.com/docs/api/popover#interfaces

liamdebeasi (changes requested):

core/src/components/popover/popover-interface.ts
sean-perkins: Good catch 👍  wonder how common this is (referencing public interfaces manually). May be something we consider using MDX with a component to raw-load a source file or something 🤔

liamdebeasi (approved):

PR #24719, opened 2022-02-04
fix(router): fix router push with relative path

### LINKED ISSUE #24718: bug: ion-router push relative navigation is broken
### Prerequisites

### Ionic Framework Version

- [ ] v4.x
- [ ] v5.x
- [X] v6.x

### Current Behavior

The push method of the router is supposed to support relative navigation (i.e `./some/path` and `../some/path`) but the implementation assumes:
- `useHash` = false,
- `root` = /

For example calling `push('./relative')` on `http://localhost:3334/src/components/router/test/basic#/two/three/hello` navigates to `http://localhost:3334/src/components/router/test/basic#/src/components/router/test/relative`.

### Expected Behavior

Calling `push('./relative')` on `http://localhost:3334/src/components/router/test/basic#/two/three/hello` navigates to `http://localhost:3334/src/components/router/test/basic#/two/three/relative`.

### Steps to Reproduce

Call `push()` with a relative when `useHash` is true or `root` is not `/`

### Code Reproduction URL

_No response_

### Ionic Info

bash: ionic: command not found :)

### Additional Information

The problem comes from the fact that there is a confusion between a path and a url.

I have the fix ready and will submit a PR shortly.

vicb:
## Pull request checklist

- [X] Build (`npm run build`) was run locally and any changes were pushed - only changes related to this PR

## Pull request type

 

 

- [X] Bugfix
- [ ] Feature
- [ ] Build related changes

Router.push() is buggy when using a relative path and `useHash` = true or `root` != /

Issue Number: fixes #24718 

## What is the new behavior?

Router.push works with relative paths

- [ ] Yes
- [X] No

### DIFF
--- a/core/src/components/router/router.tsx
+++ b/core/src/components/router/router.tsx
@@ -118,18 +118,21 @@ export class Router implements ComponentInterface {
   }
 
   /**
-   * Navigate to the specified URL.
+   * Navigate to the specified path.
    *
-   * @param url The url to navigate to.
+   * @param path The path to navigate to.
    * @param direction The direction of the animation. Defaults to `"forward"`.
    */
   @Method()
-  async push(url: string, direction: RouterDirection = 'forward', animation?: AnimationBuilder) {
-    if (url.startsWith('.')) {
-      url = (new URL(url, window.location.href)).pathname;
+  async push(path: string, direction: RouterDirection = 'forward', animation?: AnimationBuilder) {
+    if (path.startsWith('.')) {
+      const currentPath = this.previousPath ?? '/';
+      // Convert currentPath to an URL by pre-pending a protocol and a host to resolve the relative path.
+      const url = new URL(path, `https://host/${currentPath}`);
+      path = url.pathname + url.search;
     }
 
-    let parsedPath = parsePath(url);
+    let parsedPath = parsePath(path);
 
     const canProceed = await this.runGuards(parsedPath.segments);
     if (canProceed !== true) {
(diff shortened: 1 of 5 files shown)


core/src/components/router/router.tsx
```diff
@@ -118,18 +118,19 @@ export class Router implements ComponentInterface {
...
    * @param direction The direction of the animation. Defaults to `"forward"`.
    */
   @Method()
-  async push(url: string, direction: RouterDirection = 'forward', animation?: AnimationBuilder) {
-    if (url.startsWith('.')) {
-      url = (new URL(url, window.location.href)).pathname;
+  async push(path: string, direction: RouterDirection = 'forward', animation?: AnimationBuilder) {
+    if (path.startsWith('.')) {
+      const currentPath = this.previousPath ?? '/';
+      path = (new URL(path, `https://host/${currentPath}`)).pathname;
```
sean-perkins, line 130: I'm unfamiliar to this syntax. Can you explain a little of what this does/how this works? Also, would this negatively impact/cause issues with developers who are not serving their content through https?

vicb: URL [second parameter](https://developer.mozilla.org/en-US/docs/Web/API/URL#usage_notes) is a url but we have a path which is only the path in the router a tree.

So basically we are creating a "fake" url just for the purpose of resolving a relative path. To transform a path to a url, you need to add a protocol (`http(s)://`) and a host (`host`).

We only read back the pathname and the search parts so the protocol and host could be anything, this would change the outcome.

Does this make sense ?

sean-perkins: Perfect, thanks for the explanation that makes a lot of sense! I think it could be helpful to include an inline comment regarding this.

vicb: Done

liamdebeasi (approved):

amandaesmith3 (approved):

PR #24721, opened 2022-02-04
fix(vue): replacing routes now updates location state correctly

liamdebeasi:
## Pull request checklist

## Pull request type

 

 

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

Issue Number: resolves https://github.com/ionic-team/ionic-framework/issues/24432

There are a few related issues:

1. When the incoming route had `routeAction === 'replace'` Ionic Vue would determine the leaving location by getting the second to last item in location history. This was fine when popping would remove items from location history automatically (I.e. what Ionic React does), but when you support `router.go(n)` you cannot do that. As a result, the `locationHistory.previous()` call was not always returning the correct previous route based on where the user was in location history.

2. The `routeInfo.tab === undefined` check was actually a bandaid for an issue I had run into when calling the old `updateByHistoryPosition` method (See: https://github.com/ionic-team/ionic-framework/blob/a09d7d4ab6dd0d90204015eaaf232ed190753c56/packages/vue-router/src/router.ts#L294). However, it was later determined in https://github.com/ionic-team/ionic-framework/commit/0b18260da64334d8211c5a0cd806f7416274fc5e that  this was not the correct solution for the problem that it tried to solve. Now that we have the correct solution in place, this bandaid is no longer needed.

3. When re-writing history, Ionic Vue never considered the case where you wanted to replace the current state of the route with another route and as a result, the location history was not being updated/wiped appropriately. This worked fine when you navigated backwards and then did a `router.replace`, but if you just did a `router.replace` on its own, location history did not update properly.

## What is the new behavior?

- Removed the `.previous` method in favor of calling `.current` which takes into account your location within history. I also removed the `previous` method in general as it is no longer used in the codebase.
- Removed the `routeInfo.tab` bandaid.
- Added a check to account for when doing `router.replace` in order to determine if we should wipe the location history.
- Unrelated to the PR, but there was a stray `cy.wait` that was not needed. I removed that.

- [ ] Yes
- [x] No

### DIFF
--- a/packages/vue-router/__tests__/locationHistory.spec.ts
+++ b/packages/vue-router/__tests__/locationHistory.spec.ts
@@ -69,15 +69,12 @@ describe('Location History', () => {
     expect(first.pathname).toEqual('/tabs/tab1/child/1');
   });
 
-  it('should correctly get current and previous routes', () => {
+  it('should correctly get last route', () => {
     locationHistory.add({ pathname: '/home' });
     locationHistory.add({ pathname: '/login' });
 
     const current = locationHistory.last();
     expect(current.pathname).toEqual('/login');
-
-    const previous = locationHistory.previous();
-    expect(previous.pathname).toEqual('/home');
   });
 
   it('should correctly determine if we can go back', () => {
--- a/packages/vue-router/src/locationHistory.ts
+++ b/packages/vue-router/src/locationHistory.ts
@@ -134,7 +134,6 @@ export const createLocationHistory = () => {
     const index = currentHistory - initialHistory;
     return locationHistory[index] || last();
   }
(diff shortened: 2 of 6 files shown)


amandaesmith3 (approved):
This LGTM 👍 Re: clearing out tabs, I'm hesitant to touch more than needed at once with router stuff; LMK if your investigation turns up anything but I'd want a strong repro/use case before forming opinions.

liamdebeasi:
Agreed. I couldn't think of a use case where this is relevant. I am in favor of leaving it alone for now until we get feedback from the community.

sean-perkins (approved):
Looks good to me. Agreed the discussion regarding leaving tabs alone for now.

Issue #24706, opened 2022-02-03
bug: last item in list is covered in a inline modal if header is also present


mburger81:
### Prerequisites

- [X] I have read the [Contributing Guidelines](https://github.com/ionic-team/ionic-framework/blob/main/.github/CONTRIBUTING.md#creating-an-issue).
- [X] I agree to follow the [Code of Conduct](https://ionicframework.com/code-of-conduct).
- [X] I have searched for [existing issues](https://github.com/ionic-team/ionic-framework/issues) that already report this problem, without success.

### Ionic Framework Version

- [ ] v4.x
- [ ] v5.x
- [X] v6.x

### Current Behavior

Opening an inline modal which contains a header and also a list of items are not able to scroll to the bottom of the list. Last item in list is covered and not shown.
If I remove the `ion-header` part from the modal anything is fine

Look at the picture you can see that at the bottom of the modal an item is cut, its the item with the number 16

![image](https://user-images.githubusercontent.com/3778892/152363108-b4d81181-a1b7-4fe6-bd6c-dbe2888d2f1f.png)

```
<ion-modal [isOpen]="true">
  <ng-template>
    <ion-header translucent>
      <ion-toolbar>
        <ion-title>Modal Content</ion-title>
        <ion-buttons slot="end">
          <ion-button>Close</ion-button>
        </ion-buttons>
      </ion-toolbar>
    </ion-header>
    <ion-content fullscreen>
      <ion-list>
        <ion-item>
          <ion-avatar slot="start">
          </ion-avatar>
          <ion-label>
            <h2>Gollum</h2>
            <p>Sneaky little hobbitses!</p>
          </ion-label>
        </ion-item>
        <ion-item>
          <ion-avatar slot="start">
          </ion-avatar>
          <ion-label>
            <h2>Frodo</h2>
            <p>Go back, Sam! I'm going to Mordor alone!</p>
          </ion-label>
        </ion-item>
        <ion-item>
          <ion-avatar slot="start">
          </ion-avatar>
          <ion-label>
            <h2>Samwise</h2>
            <p>What we need is a few good taters. 1</p>
          </ion-label>
        </ion-item>
        <ion-item>
          <ion-avatar slot="start">
          </ion-avatar>
          <ion-label>
            <h2>Samwise</h2>
            <p>What we need is a few good taters. 2</p>
          </ion-label>
        </ion-item>
........
        <ion-item>
          <ion-avatar slot="start">
          </ion-avatar>
[...]

PR #24723, opened 2022-02-04
fix(modal): inline modals inherit ion-page styling

sean-perkins:
## Pull request checklist

## Pull request type

 

 

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

Inline modals do not inherit CSS classes that are forwarded when the modal is presented. This results in implementations that make use of an `ion-header`, `ion-content` with overflow scroll, inside an inline modal, are unable to reach the end of the contents.

Issue Number: #24706

## What is the new behavior?

Inline modals now inherit CSS classes that are forwarded when the modal is presented (aligning with the controller implementation).

This results in the user being able to correctly scroll to the end of the contents, when using the above scenario.

- Inline modal templates are appended to a new parent element that inherits the CSS classes
- Inline modal template is replaced with the new parent element

- [ ] Yes
- [x] No

## Other information

This only appears to affect modal implementations (can be seen in our inline modal example when adding overflow). Was not able to replicate the same issue with popover usages.

### DIFF
--- a/core/src/utils/framework-delegate.ts
+++ b/core/src/utils/framework-delegate.ts
@@ -96,7 +96,6 @@ export const CoreDelegate = () => {
       cssClasses.forEach(c => el.classList.add(c));
       // Move each child from the original template to the new parent element.
       el.append(...BaseComponent.children);
-      
       // Append the new parent element to the original parent element.
       BaseComponent.appendChild(el);
     }
--- a/core/src/utils/framework-delegate.ts
+++ b/core/src/utils/framework-delegate.ts
@@ -92,14 +92,13 @@ export const CoreDelegate = () => {
     } else {
       // If there is no component, then we need to create a new parent
       // element to apply the css classes to.
-      const el: HTMLElement = BaseComponent.ownerDocument && BaseComponent.ownerDocument.createElement('div');
+      const el = BaseComponent.ownerDocument && BaseComponent.ownerDocument.createElement('div');
       cssClasses.forEach(c => el.classList.add(c));
       // Move each child from the original template to the new parent element.
-      while (BaseComponent.hasChildNodes()) {
-        el.appendChild(BaseComponent.firstChild);
-      }
+      el.append(...BaseComponent.children);
+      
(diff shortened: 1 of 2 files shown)


amandaesmith3 (approved):

core/src/utils/framework-delegate.ts
```diff
@@ -89,6 +89,17 @@ export const CoreDelegate = () => {
...
+      // If there is no component, then we need to create a new parent
+      // element to apply the css classes to.
+      const el: HTMLElement = BaseComponent.ownerDocument && BaseComponent.ownerDocument.createElement('div');
+      cssClasses.forEach(c => el.classList.add(c));
+      // Move each child from the original template to the new parent element.
+      while (BaseComponent.hasChildNodes()) {
+        el.appendChild(BaseComponent.firstChild);
+      }
+      // Append the new parent element to the original parent element.
+      (BaseComponent as HTMLElement).appendChild(el);
```
liamdebeasi, line 102: Another way to do this without the while loop and typecasting:

```suggestion
      const el = BaseComponent.ownerDocument && BaseComponent.ownerDocument.createElement('div');
      cssClasses.forEach(c => el.classList.add(c));
      // Move each child from the original template to the new parent element.
      el.append(...BaseComponent.children);
      
      // Append the new parent element to the original parent element.
      BaseComponent.appendChild(el);
```

Because `el` is of type `Element`, we can use [Element.append()](https://developer.mozilla.org/en-US/docs/Web/API/Element/append) which accepts multiple nodes to insert.

In my (very unscientific) tests I did not see any major performance difference: https://codepen.io/liamdebeasi/pen/GROWRvV

I'll leave it up to you, but this code seems a bit cleaner since it does not require typecasting.

liamdebeasi (approved):
One non-blocking comment

core/src/utils/framework-delegate.ts
sean-perkins: I like your approach more. I'll update it!

PR #24730, opened 2022-02-06
chore(nav): code cleanup

vicb:
## Pull request checklist

## Pull request type

 

 

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes

This PR do not change the behavior.

It tries to make the code of `<ion-nav>` easier to understand by:
- simplifying some code,
- adding some comments,
- there is also a bit of renaming and formatting (for consistency).

- [ ] Yes
- [X] No

### DIFF
--- a/core/src/components/nav/nav.tsx
+++ b/core/src/components/nav/nav.tsx
@@ -92,16 +92,11 @@ export class Nav implements NavOutlet {
   @Event({ bubbles: false }) ionNavDidChange!: EventEmitter<void>;
 
   componentWillLoad() {
-    this.useRouter =
-      !!document.querySelector('ion-router') &&
-      !this.el.closest('[no-router]');
+    this.useRouter = document.querySelector('ion-router') !== null && this.el.closest('[no-router]') === null;
 
     if (this.swipeGesture === undefined) {
       const mode = getIonMode(this);
-      this.swipeGesture = config.getBoolean(
-        'swipeBackEnabled',
-        mode === 'ios'
-      );
+      this.swipeGesture = config.getBoolean('swipeBackEnabled', mode === 'ios');
     }
 
     this.ionNavWillLoad.emit();
@@ -126,13 +121,13 @@ export class Nav implements NavOutlet {
       view._destroy();
     }
 
+    // Release swipe back gesture and transition.
     if (this.gesture) {
       this.gesture.destroy();
       this.gesture = undefined;
     }
-
-    // release swipe back gesture and transition
-    this.transInstr.length = this.views.length = 0;
+    this.transInstr.length = 0;
+    this.views.length = 0;
     this.destroyed = true;
   }
 
@@ -153,14 +148,7 @@ export class Nav implements NavOutlet {
     opts?: NavOptions | null,
     done?: TransitionDoneFn
   ): Promise<boolean> {
-    return this.queueTrns(
-      {
-        insertStart: -1,
-        insertViews: [{ component, componentProps }],
-        opts
-      },
-      done
-    );
+    return this.insert(-1, component, componentProps, opts, done);
   }
 
   /**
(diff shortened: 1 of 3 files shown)


core/src/components/nav/nav.tsx
```diff
@@ -906,25 +889,36 @@ export class Nav implements NavOutlet {
...
+   * @param view The view to insert.
+   * @param index The index where to insert the view.
+   */
   private insertViewAt(view: ViewController, index: number) {
     const views = this.views;
     const existingIndex = views.indexOf(view);
     if (existingIndex > -1) {
-      // this view is already in the stack!!
-      // move it to its new location
+      // The view already in the stack, removes it.
```
liamdebeasi, line 904: It looks like this moves the item in the stack rather than removes it altogether. Can we update the comment?

core/src/components/nav/nav.tsx
```diff
@@ -607,25 +579,33 @@ export class Nav implements NavOutlet {
...
-    this.runTransition(ti);
+    this.runTransitions(ti);
     return true;
   }
 
-  private async runTransition(ti: TransitionInstruction) {
+  /**
+   * Executes all the transition instruction from the queue.
+   *
+   * @param ti The transition instructions.
```
liamdebeasi, line 606: Can we make this more descriptive or remove altogether? There isn't any information in here that I could not already deduce from the type name.

core/src/components/nav/nav.tsx
```diff
@@ -503,22 +477,20 @@ export class Nav implements NavOutlet {
...
-  // 5. _transitionInit(): initializes the visual transition if required and schedules it to run
-  // 6. _viewAttachToDOM(): attaches the enteringView to the DOM
-  // 7. _transitionStart(): called once the transition actually starts, it initializes the Animation underneath.
-  // 8. _transitionFinish(): called once the transition finishes
-  // 9. _cleanup(): syncs the navigation internal state with the DOM. For example it removes the pages from the DOM or hides/show them.
+  /**
+   * Adds a navigation stack change to the queue and schedules it to run.
+   *
+   * @param ti The instructions.
+   * @param done Done function.
```
liamdebeasi, line 484: Same thing here -- the comment does not provide any additional information that I could not already deduce from the type names below. Can we either make this more descriptive or remove?

core/src/components/nav/nav.tsx
```diff
@@ -607,25 +579,33 @@ export class Nav implements NavOutlet {
...
     return true;
   }
 
-  private async runTransition(ti: TransitionInstruction) {
+  /**
+   * Executes all the transition instruction from the queue.
+   *
+   * @param ti The transition instructions.
+   */
+  private async runTransitions(ti: TransitionInstruction) {
```
liamdebeasi, line 608: Any particular reason for this change? This function does run a single transition, so maybe keeping the singular form would be best. I realize it does call `nextTrns` which causes this to be run again with another transition, but the body of this function only deals with a single transition at a time.

core/src/components/nav/nav.tsx
vicb: I have clarified the comment.

And I think there is probably a bug here as you probably want to insert at index - 1 if `existingIndex` is before `index`.
I don't want to fix this in this refactoring PR and need to come with a test case.

core/src/components/nav/nav.tsx
vicb: Removed

core/src/components/nav/nav.tsx
vicb: Removed

core/src/components/nav/nav.tsx
vicb: Agreed. I changed the name when I was looking how transitions were chained but the singular is probably ad good.

vicb:
Thanks for the review @liamdebeasi. 
I have updated with your feedback.

liamdebeasi (approved):

sean-perkins (approved):

liamdebeasi:
Merged. Thank you!

vicb:
Thanks, I am going to send the next one.

Issue #24669, opened 2022-01-27
bug: ion-input duplicate input values and delete key does not work with Japanese keyboard


tatsuyasenbom:
### Prerequisites

- [X] I have read the [Contributing Guidelines](https://github.com/ionic-team/ionic-framework/blob/main/.github/CONTRIBUTING.md#creating-an-issue).
- [X] I agree to follow the [Code of Conduct](https://ionicframework.com/code-of-conduct).
- [X] I have searched for [existing issues](https://github.com/ionic-team/ionic-framework/issues) that already report this problem, without success.

### Ionic Framework Version

- [ ] v4.x
- [ ] v5.x
- [X] v6.x

### Current Behavior

* In ion-input on iOS 15.2 x Ionic 6.0.4, two characters are entered with one tap, and the delete key does not work.
    * https://user-images.githubusercontent.com/87968771/151362623-b765c51c-c8b2-40b9-bc65-47bce33929c3.mp4
* This bug does not occur on the latest Google Chrome on Android 12 or Windows 11.
 

### Expected Behavior

* I want that the bugs reported in the Current Behavior will be improved.

### Steps to Reproduce

* It can be reproduced by any of the following build methods.
    * `ionic serve --address xxx.xxx.xxx.xx`
    * `ionic biuld --prod`

### Code Reproduction URL

https://github.com/tatsuyasenbom/ionic-angular-sample/commit/ae4184a8099691e0b3f6c8f3982ca7f24e964e8b

### Ionic Info

```
Ionic:

   Ionic CLI                     : 6.18.1 (C:\Users\Administrator\AppData\Roaming\npm\node_modules\@ionic\cli)        
   Ionic Framework               : @ionic/angular 6.0.4
   @angular-devkit/build-angular : 13.2.0
   @angular-devkit/schematics    : 13.0.4
   @angular/cli                  : 13.0.4
   @ionic/angular-toolkit        : 5.0.3

Capacitor:

   Capacitor CLI      : 3.3.3
   @capacitor/android : not installed
   @capacitor/core    : 3.3.3
   @capacitor/ios     : not installed

Utility:

   cordova-res : 0.15.4
   native-run  : 1.5.0

System:

   NodeJS : v14.17.0 (C:\Program Files\nodejs\node.exe)
   npm    : 6.14.13
   OS     : Windows 10
```

tatsuyasenbom:
In Ionic 6.0.3, this bug did not occur.

[...]

PR #24735, opened 2022-02-08
fix(input): IME composition mode

sean-perkins:
## Pull request checklist

## Pull request type

 

 

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

IMEs (input method editors) intercept keyboard events and conflict with the patch in #24606, since both are assigning a value to the native input, resulting in duplicate values on keypress. This also breaks the back button behavior (observed in Japanese IME). 

Issue Number: #24669

## What is the new behavior?

To prevent conflicts with IMEs and preserve the patch from the PR that introduced the new issue, it is recommended that we bind event listeners to `compositionstart` and `compositionend` to detect when a composition event from an IME is occurring and skip writing the value to the native input control while the event is in flight (inspired from: https://github.com/ng-select/ng-select/pull/1274)

- Usage of Japanese IME does not result in duplicate values
- Japanese IME usage of the back button/delete key is restored

- [ ] Yes
- [x] No

## Other information

While triaging this issue, a Stencil bug was discovered (https://github.com/ionic-team/stencil/issues/3235) that prevents using the JSX bindings for the composition events. 

To verify this issue, you can temporarily add a Japanese keyboard to your iOS device and open the basic input test in the browser. A guide on how to add the keyboard is available here: https://techwiser.com/use-japanese-keyboard-iphone.

### DIFF
--- a/core/src/components/input/input.tsx
+++ b/core/src/components/input/input.tsx
@@ -262,11 +262,12 @@ export class Input implements ComponentInterface {
   }
 
   componentDidLoad() {
-    if (this.nativeInput) {
-      // TODO: Update to JSX bindings when Stencil resolves bug with:
+    const nativeInput = this.nativeInput;
+    if (nativeInput) {
+      // TODO: FW-729 Update to JSX bindings when Stencil resolves bug with:
       // https://github.com/ionic-team/stencil/issues/3235
-      this.nativeInput.addEventListener('compositionstart', this.onCompositionStart);
-      this.nativeInput.addEventListener('compositionend', this.onCompositionEnd);
+      nativeInput.addEventListener('compositionstart', this.onCompositionStart);
+      nativeInput.addEventListener('compositionend', this.onCompositionEnd);
     }
   }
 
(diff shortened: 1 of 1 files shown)


sean-perkins:
The community member that reported the issue attached to this PR, commented that the dev build of this PR resolves their issue: https://github.com/ionic-team/ionic-framework/issues/24669#issuecomment-1032203212

core/src/components/input/input.tsx
```diff
@@ -260,12 +261,25 @@ export class Input implements ComponentInterface {
...
   disconnectedCallback() {
     if (Build.isBrowser) {
       document.dispatchEvent(new CustomEvent('ionInputDidUnload', {
         detail: this.el
       }));
     }
+    if (this.nativeInput) {
+      this.nativeInput.removeEventListener('compositionstart', this.onCompositionStart);
+      this.nativeInput.removeEventListener('compositionEnd', this.onCompositionEnd);
+    }
```
liamdebeasi, line 282: Minor thing, but requesting `this.nativeInput` causes some extra code to be run in Stencil to get the value, so caching the value allows us to skip that code from running additional times.

```suggestion
    const nativeInput = this.nativeInput;
    if (nativeInput) {
      nativeInput.removeEventListener('compositionstart', this.onCompositionStart);
      nativeInput.removeEventListener('compositionEnd', this.onCompositionEnd);
    }
```

liamdebeasi: Also if we can do the same thing for the addEventListener above that'd be great

liamdebeasi (changes requested):

amandaesmith3 (approved):
LGTM once Liam's comments are addressed 👍

core/src/components/input/input.tsx
sean-perkins: Can do 👍  I wasn't aware that accessing private members in Stencil had an overhead. Is that because it's bound to a `ref` or just in general?

liamdebeasi: I believe anything that is declared as a variable on the component instance has this slight overhead.

So basically anything like this https://github.com/ionic-team/ionic-framework/blob/main/core/src/components/input/input.tsx#L21-L25

Issue #24727, opened 2022-02-06
bug: [iOS] ion-input cursor jumps to end on input


tetsuwanadam:
### Prerequisites

- [X] I have read the [Contributing Guidelines](https://github.com/ionic-team/ionic-framework/blob/main/.github/CONTRIBUTING.md#creating-an-issue).
- [X] I agree to follow the [Code of Conduct](https://ionicframework.com/code-of-conduct).
- [X] I have searched for [existing issues](https://github.com/ionic-team/ionic-framework/issues) that already report this problem, without success.

### Ionic Framework Version

- [ ] v4.x
- [ ] v5.x
- [X] v6.x

### Current Behavior

On iOS 15.3, in an ion-input with existing text, if you set the cursor at the beginning or in the middle of the existing text and type a character, the cursor jumps to the end of the input.

### Expected Behavior

As you type the cursor should stay after the typed character, regardless of where in the input text you are typing.

### Steps to Reproduce

In an ion-input with existing text, move the cursor to the beginning or middle of the text and type a character.
I believe this issue arose either with the iOS 15.3 update or possibly the Ionic Vue 6.0.4 update.

### Code Reproduction URL

https://github.com/tetsuwanadam/ionic-test/tree/ion-input-ios-cursor-issue

### Ionic Info

Ionic:

   Ionic CLI       : 6.18.1 (/usr/local/lib/node_modules/@ionic/cli)
   Ionic Framework : @ionic/vue 6.0.5

Capacitor:

   Capacitor CLI      : 3.4.0
   @capacitor/android : not installed
   @capacitor/core    : 3.4.0
   @capacitor/ios     : 3.4.0

Utility:

   cordova-res : not installed globally
   native-run  : 1.5.0

System:

   NodeJS : v16.1.0 (/usr/local/bin/node)
   npm    : 7.24.1
   OS     : macOS Big Sur

liamdebeasi:
Thanks! I can reproduce this behavior in iOS 15. This appears to be a bug in iOS, not Ionic.

The good news is this issue appears to be addressed in the iOS 15.4 beta. I will update this thread when that version of iOS has shipped.

sean-perkins:
@liamdebeasi this appears to be a regression from this PR that I authored: https://github.com/ionic-team/ionic-framework/pull/24606

[...]

PR #24736, opened 2022-02-08
fix(input): cursor position does not jump to end

sean-perkins:
## Pull request checklist

## Pull request type

 

 

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

Setting the cursor position to the middle of an `ion-input` with an existing value and typing a new character, would result in the cursor position to reset to the end of the input.

Issue Number: #24727

## What is the new behavior?

When assigning the native input value, persist the cursor position so that the cursor stays in place and no longer jumps to the end of the line.

- [ ] Yes
- [x] No

### DIFF
--- a/core/src/components/input/input.tsx
+++ b/core/src/components/input/input.tsx
@@ -240,7 +240,10 @@ export class Input implements ComponentInterface {
        * Used for patterns such as input trimming (removing whitespace),
        * or input masking.
        */
-      this.nativeInput.value = this.getValue();
+       const { selectionStart, selectionEnd } = this.nativeInput;
+       this.nativeInput.value = this.getValue();
+       // Set the cursor position back to where it was before the value change
+       this.nativeInput.setSelectionRange(selectionStart, selectionEnd);
     }
     this.emitStyle();
     this.ionChange.emit({ value: this.value == null ? this.value : this.value.toString() });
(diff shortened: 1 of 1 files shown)


core/src/components/input/input.tsx
```diff
@@ -240,7 +240,10 @@ export class Input implements ComponentInterface {
        * Used for patterns such as input trimming (removing whitespace),
        * or input masking.
        */
-      this.nativeInput.value = this.getValue();
+       const { selectionStart, selectionEnd } = this.nativeInput;
+       this.nativeInput.value = this.getValue();
+       // Set the cursor position back to where it was before the value change
+       this.nativeInput.setSelectionRange(selectionStart, selectionEnd);
```
liamdebeasi, line 246: Can you make a tech debt ticket to remove this when we drop support for iOS 15.3? The issue may have appeared as a result of one of your other changes, but the behavior is definitely a bug in iOS given that it reproduces in 15.3 but not in 15.4.

liamdebeasi (changes requested):

core/src/components/input/input.tsx
sean-perkins: Yes, can do!

amandaesmith3 (approved):

Issue #24757, opened 2022-02-09
bug: datetime month navigation blocked with min attribute


antoine92190:
### Prerequisites

- [X] I have read the [Contributing Guidelines](https://github.com/ionic-team/ionic-framework/blob/main/.github/CONTRIBUTING.md#creating-an-issue).
- [X] I agree to follow the [Code of Conduct](https://ionicframework.com/code-of-conduct).
- [X] I have searched for [existing issues](https://github.com/ionic-team/ionic-framework/issues) that already report this problem, without success.

### Ionic Framework Version

- [ ] v4.x
- [ ] v5.x
- [X] v6.x

### Current Behavior

When I set a min date, for example `min="2022-02-09"`, the datetime picker opens correctly, and January is correctly disabled.
But if I navigate to march, I cannot go back to February.

https://user-images.githubusercontent.com/16256207/153254693-e2508753-aa73-4c89-96c3-79aa42f543b7.mov

### Expected Behavior

We should be able to go back to previous month as it is and was not disabled.

### Steps to Reproduce

- Add a min date to the datetime picker component
- Open the datetime picker
- Navigate to the next month
- Try to navigate back to the previous month => Not possible

### Ionic Info

Ionic:

   Ionic CLI                     : 6.12.1 (/usr/local/lib/node_modules/@ionic/cli)
   Ionic Framework               : @ionic/angular 6.0.7
   @angular-devkit/build-angular : 12.2.4 (/Users/antoine/Dev/agorapulse/platform/frontend/node_modules/@angular-devkit/build-angular)
   @angular-devkit/schematics    : 11.2.18
   @angular/cli                  : 12.2.4 (/Users/antoine/Dev/agorapulse/platform/frontend/node_modules/@angular/cli)
   @ionic/angular-toolkit        : 3.1.1

Capacitor:

   Capacitor CLI   : 3.2.4
   @capacitor/core : 3.2.4

Cordova:

   Cordova CLI       : not installed
   Cordova Platforms : not available
   Cordova Plugins   : not available

Utility:

   cordova-res : not installed
   native-run  : 1.5.0

System:

   NodeJS : v14.17.0 (/usr/local/bin/node)
   npm    : 6.14.13
   OS     : macOS
   Xcode  : Xcode 13.2.1 Build version 13C100

sean-perkins:
@antoine92190 thanks for reporting this issue! I can replicate the same behavior in our `minmax` test file.

[...]

PR #24759, opened 2022-02-09
fix(datetime): navigate to month within min range

sean-perkins:
## Pull request checklist

## Pull request type

 

 

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

When providing a `min` value to `ion-datetime` that includes a day, the month of the `min` value will be disabled and once you have navigated away from it, you will be unable to return to it. 

This is because the `isBefore`/`isAfter` checks additionally compare the day value, when min/max includes it. This is correct for date comparisons, but isn't a 1:1 with rules surrounding disabling months.

Issue Number: #24757

## What is the new behavior?

Users can navigate to min/max months, when the `min`/`max` value includes a day constraint. 

i.e.: If the `min` is `2022-02-09`, users should still be able to navigate to February, since days 10-28 are enabled.  

- [ ] Yes
- [x] No

### DIFF
--- a/core/src/components/datetime/datetime.tsx
+++ b/core/src/components/datetime/datetime.tsx
@@ -1238,6 +1238,9 @@ export class Datetime implements ComponentInterface {
       year,
       day: null
     }, {
+      // The day is not used when checking if a month is disabled.
+      // Users should be able to access the min or max month, even if the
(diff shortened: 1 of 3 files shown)


liamdebeasi (approved):
Can we add comments to the code as to why we are passing in `day: null`? Otherwise this looks good.

Issue #24584, opened 2022-01-17
bug: IonSelect w/ multiple={true} yields incorrectly formatted value in onIonChange


ebk46:
### Prerequisites

- [X] I have read the [Contributing Guidelines](https://github.com/ionic-team/ionic-framework/blob/main/.github/CONTRIBUTING.md#creating-an-issue).
- [X] I agree to follow the [Code of Conduct](https://ionicframework.com/code-of-conduct).
- [X] I have searched for [existing issues](https://github.com/ionic-team/ionic-framework/issues) that already report this problem, without success.

### Ionic Framework Version

- [ ] v4.x
- [ ] v5.x
- [X] v6.x

### Current Behavior

When a change is made in an IonSelect with `multiple={true}`, it triggers onIonChange 3 times with the following behavior:

Trigger 1) Values as array
Trigger 2) Values as comma-separated string
Trigger 3) Values as array

Obviously, this is very bad as it should only trigger once, and the 2nd trigger violates the Array<T> exected type.

### Expected Behavior

onIonChange should only trigger once, and should include only an array of values, not a comma-separated string.

### Steps to Reproduce

This happens with any IonSelect with multiple={true}. See example project below.

### Code Reproduction URL

https://stackblitz.com/edit/react-ne3qq8

### Ionic Info

Ionic:

   Ionic CLI       : 6.18.1 (/Users/evan/.config/yarn/global/node_modules/@ionic/cli)
   Ionic Framework : @ionic/react 6.0.2

Capacitor:

   Capacitor CLI      : 3.3.4
   @capacitor/android : 3.3.4
   @capacitor/core    : 3.3.4
   @capacitor/ios     : 3.3.4

Utility:

   cordova-res (update available: 0.15.4) : 0.15.3
   native-run                             : 1.5.0

System:

   NodeJS : v16.13.2 (/usr/local/bin/node)
   npm    : 8.1.2
   OS     : macOS Catalina

sean-perkins:
Thanks for reporting this issue! I'm seeing the same problems within your reproduction.

Similar issues have been marked as a duplicate of #20106, but I will confirm with the team on Tuesday (Monday is a holiday).

I'm currently doing internal exploration to rewrite how form controls behave and can add this test case to the list to see if it's resolved with the change to how we notify value change events. 

Regardless of the duplicate emission, the `1,2` string value is absolutely wrong and should never be emitted.

amandaesmith3:
[...]

PR #24763, opened 2022-02-10
fix(select): fix infinite loop when using array

adamduren:
## Pull request checklist

## Pull request type

 

 

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

When using multiple and passing an array of objects ionChange is emitted multiple times. The array is eventually stringified and the component becomes unusable.

Issue Number: #24584

## What is the new behavior?

The array is not stringified and the component behaves as expected.

-
-
-

- [ ] Yes
- [x] No

### DIFF
--- a/packages/react/src/components/proxies.ts
+++ b/packages/react/src/components/proxies.ts
@@ -72,6 +72,14 @@ import { defineCustomElement as defineIonToggle } from '@ionic/core/components/i
 import { defineCustomElement as defineIonToolbar } from '@ionic/core/components/ion-toolbar.js';
 import { defineCustomElement as defineIonVirtualScroll } from '@ionic/core/components/ion-virtual-scroll.js';
 
+const excludeValueProp = (_: any, props: any) => {
+  const newProps = {
+    ...props
+  };
+  delete newProps['value'];
+  return newProps;
+}
+
 export const IonAccordion = /*@__PURE__*/createReactComponent<JSX.IonAccordion, HTMLIonAccordionElement>('ion-accordion', undefined, undefined, defineIonAccordion);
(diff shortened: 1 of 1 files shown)


packages/react/src/components/proxies.ts
```diff
@@ -72,6 +72,14 @@ import { defineCustomElement as defineIonToggle } from '@ionic/core/components/i
 import { defineCustomElement as defineIonToolbar } from '@ionic/core/components/ion-toolbar.js';
 import { defineCustomElement as defineIonVirtualScroll } from '@ionic/core/components/ion-virtual-scroll.js';
 
+const excludeValueProp = (_: any, props: any) => {
```
sean-perkins, line 75: This file is unfortunately autogenerated, so any changes here will be overwritten every time the `core` package is built.

Instead, we will need to exclude `ion-select` from being auto generated here: https://github.com/ionic-team/ionic-framework/blob/main/core/stencil.config.ts#L72

and then create a new file in `packages/react/src/components/` called `IonSelect.tsx` that is basically:

```tsx
import type { JSX } from '@ionic/core/components';
import { defineCustomElement  } from '@ionic/core/components/ion-select.js';

import { createReactComponent } from './react-component-lib';

const excludeValueProp = (_: any, props: any) => {
  const newProps = {
    ...props
  };
  delete newProps['value'];
  return newProps;
}

export const IonSelect = /*@__PURE__*/createReactComponent<JSX.IonSelect, HTMLIonSelectElement>('ion-select', undefined, excludeValueProp, defineCustomElement);
```

and finally export it from `packages/react/src/components/index.ts`.

sean-perkins (changes requested):

packages/react/src/components/proxies.ts
adamduren: Since React reflects props to attributes does it make sense for the `render` function to filter out any prop with an object value to avoid this potential kind of issue elsewhere?

adamduren:
@sean-perkins anymore thoughts on my responses?

Issue #24774, opened 2022-02-11
bug: the ion-refresher create a conflict with the fixed elements in IOS


loicparent:
### Prerequisites

- [X] I have read the [Contributing Guidelines](https://github.com/ionic-team/ionic-framework/blob/main/.github/CONTRIBUTING.md#creating-an-issue).
- [X] I agree to follow the [Code of Conduct](https://ionicframework.com/code-of-conduct).
- [X] I have searched for [existing issues](https://github.com/ionic-team/ionic-framework/issues) that already report this problem, without success.

### Ionic Framework Version

- [ ] v4.x
- [X] v5.x
- [ ] v6.x

### Current Behavior

**When**
The bug appears when I add an ion-refresher in the ion-content when I have also some fixed elements in the page. (For example, if I use fixed tabs in the page). This bug appears only in IOS (builded app).

**What**
The fixed elements doesn't be relative to the document but to the parent element ion-content.

**Why**
Because the ion-refresher add a translateZ that cause a conflict with the position fixed.

### Expected Behavior

The refresher doesn't have any impact on the element with a fixed position.

### Steps to Reproduce

1. Create a page
2. Add the component ion-content
3. Inside this component, add a component ion-refresher.
4. In the ion-content, add a new element: `<div style="width:100%;height:30px;background-color:red;position:fixed;top:0;left:0;"></div>`
5. Add enough content to have the scrollbar
6. Open the builded app in ios.
7. Scroll in this page and see that the fixed element doesn't stay fixed.

### Ionic Info

Ionic:

   Ionic CLI                     : 6.18.1 (/Users/loicparent/.nvm/versions/node/v14.15.0/lib/node_modules/@ionic/cli)
   Ionic Framework               : @ionic/angular 5.9.2
   @angular-devkit/build-angular : 12.2.14
   @angular-devkit/schematics    : 12.2.14
   @angular/cli                  : 12.2.14
   @ionic/angular-toolkit        : 3.1.1

Capacitor:

   Capacitor CLI      : 3.3.3
   @capacitor/android : 3.3.3
   @capacitor/core    : 3.3.3
   @capacitor/ios     : 3.3.3

Cordova:

   Cordova CLI       : not installed
   Cordova Platforms : not available
   Cordova Plugins   : not available

Utility:

   cordova-res : 0.15.4
   native-run  : 1.5.0

System:

   ios-sim : 8.0.2
[...]

PR #24784, opened 2022-02-14
fix(refresher): refresher can be used inside any fixed slot container

liamdebeasi:
## Pull request checklist

## Pull request type

 

 

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

Issue URL: resolves https://github.com/ionic-team/ionic-framework/issues/24774

Currently you cannot place multiple items in the fixed slot if using an `ion-refresher`. This is because refresher expects `slot="fixed"` to be placed on the refresher host and errors if you do not do that.

This is desirable when you have other things be `position: fixed` along with the refresher. There is a bug in WebKit with how `position: fixed` content is handled in non-body scrollable containers, so the only way to get `position: fixed` content to work properly is to put it in the `fixed` slot.

## What is the new behavior?

- Allows developers to wrap refresher in a fixed slot and use it that way. Devs will likely need to set `width: 100%` on the container.

- [ ] Yes
- [x] No

### DIFF
--- a/core/src/components/refresher/refresher.tsx
+++ b/core/src/components/refresher/refresher.tsx
@@ -414,8 +414,8 @@ export class Refresher implements ComponentInterface {
   }
 
   async connectedCallback() {
-    if (this.el.getAttribute('slot') !== 'fixed') {
-      console.error('Make sure you use: <ion-refresher slot="fixed">');
+    if (this.el.closest('[slot="fixed"]') === null) {
+      console.error('<ion-refresher> must be used inside of slot="fixed" with <ion-content>.');
       return;
     }
 


liamdebeasi:
I tried to make an E2E test that catches errors when refresher is used improperly, but I could not find a way to get the error listener setup in time (the errors are fired on load). Any ideas for how I can test this is appreciated!

sean-perkins (approved):

amandaesmith3 (approved):

liamdebeasi:
It turns out that the use case for this request was not something that was actually occurring in a real world app, so I do not think it makes sense to merge this right now. See https://github.com/ionic-team/ionic-framework/issues/24774#issuecomment-1040054769 for more context.

-------

For anyone who finds this PR and wants to see this code merged: Please open a new issue with a code reproduction. We are more than happy to consider your use case.

As a policy, we do not merge code that does not address a use case that is actively occurring in an application.

Issue #23865, opened 2021-09-02
bug: Disabled button still clickable on Safari


josh-byster:
### Prequisites

- [X] I have read the [Contributing Guidelines](https://github.com/ionic-team/ionic-framework/blob/main/.github/CONTRIBUTING.md#creating-an-issue).
- [X] I agree to follow the [Code of Conduct](https://ionicframework.com/code-of-conduct).
- [X] I have searched for [existing issues](https://github.com/ionic-team/ionic-framework/issues) that already report this problem, without success.

### Ionic Framework Version

- [ ] v4.x
- [X] v5.x
- [ ] v6.x

### Current Behavior

Hi, it seems like for `IonButton` elements with little text (e.g. 1 character), setting the `disabled` property to true after it has been clicked does not prevent subsequent `onClick` events.
For example, in my minimal reproducible example, I have 2 buttons—one with more text than the other. If I click either button, they change color, as I expect (I set `disabled` to be true after one press).
However, I can keep clicking the second button and the `onClick` event keeps firing. This doesn't seem to happen when I click the first button after it becomes disabled.

![image](https://user-images.githubusercontent.com/8664074/131779282-1b2f4ff4-4bf1-48c0-a858-9cb8a103204d.png)

![image](https://user-images.githubusercontent.com/8664074/131779072-3f55e284-878e-4a33-8069-6ea79ba36d25.png)

![image](https://user-images.githubusercontent.com/8664074/131779053-0b371cb2-3ea2-4e45-a3b4-14c01fc84dae.png)

### Expected Behavior

I expect that any time the user clicks on a disabled button, the `onClick` event doesn't fire.

### Steps to Reproduce

Open in Safari, as this does not seem to happen in Chrome. I'm using Version 14.1.1 (16611.2.7.1.4). This also happens on iOS 14.17.1 Safari.

1. Click on "Hello World" or "2".
2. You will see a popup that a button was pressed. Both buttons will now become disabled.
3. Click "2". 
4. You will see a popup that the button was pressed. This should not happen.
5. Click "Hello World". 
6. On my machine, nothing pops up, though I can't say why this is.

### Code Reproduction URL

https://github.com/josh-byster/ionic-button-disable-bug-report/blob/main/src/App.tsx

### Ionic Info

```
Ionic:

   Ionic CLI       : 5.4.16 (/usr/local/lib/node_modules/ionic)
   Ionic Framework : @ionic/react 5.7.0

Utility:

   cordova-res : not installed
[...]

PR #24803, opened 2022-02-16
fix(button): ensure button text doesn't trigger pointer events when button is disabled

amandaesmith3:
## Pull request checklist

## Pull request type

 

 

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

Under certain conditions, the button text in an `ion-button` will trigger pointer events (hover styling, click listeners) even when the button is disabled. So far, it looks like these conditions are:

- Safari only (issue reporter replicated in 14, I confirmed in 15)
- The button was set to `disabled` programmatically -- if the button was disabled from the start, the bug doesn't show
- The button text can only contain numbers (issue reporter said it's short text only, but this seems like a coincidence; I'm able to replicate the bug with any text length as long as there aren't any letters)

The clickable area of the disabled button is only on the text. Clicking the padding around the text does not trigger the bug. Additionally, opening the dev tools will sometimes stop the bug from happening until the page is refreshed.

Steps to replicate, using the "Button - Basic" test:

1. Click "Toggle Disabled" at the bottom of the page. Note that the "12345" button goes from enabled to disabled.
2. Click "12345" on the text itself. Note that the click listener fires and logs to the console. (Also note that hover styling is applied to the button.)
3. Click the "12345" button in the area around the text. Note that the click listener does not fire.

Issue URL: Resolves https://github.com/ionic-team/ionic-framework/issues/23865

## What is the new behavior?

The button text now explicitly blocks pointer events. While CSS inheritance should make this redundant, there seems to be a quirk in Safari preventing the rule from propagating under all conditions.

- [ ] Yes
- [ ] No
- [x] Unsure

Current fix seems overly broad and may cause as-of-yet unknown side effects.

## Other information

While the current fix works, for whatever reason, this does not:
```
:host(.button-disabled) .button-inner {
  pointer-events: none;
}
```

Adding a class to `.button-inner` when the `disabled` prop is `true` and applying the CSS indirectly that way also didn't work.

Everything about the conditions that trigger this bug says "Webkit issue" to me, but I haven't been able to replicate with a native `<button>` in a way that only happens on Webkit. My attempts that were of most interest were:

1. `<div><button><span>12345</span></button></div>` with the `button` set to `disabled` and a click listener on the `div`. This showed a similar behavior to the bug -- clicking the button text, i.e. the `span`, would fire the click listener. However, this also replicated in Chrome, so it seems to be an unrelated quirk of event bubbling.
2. The same HTML setup as above, but with the `button` in a shadow DOM, and `pointer-events: none` set on the `div` rather than `disabled` on the `button`. The bug did not trigger in Safari or Chrome.

### DIFF
--- a/core/src/components/button/button.scss
+++ b/core/src/components/button/button.scss
@@ -209,6 +209,10 @@
   z-index: 1;
 }
 
+.button-inner {
+  pointer-events: none;
+}
+
 
 // Button Icons
 // --------------------------------------------------
--- a/core/src/components/button/button.scss
+++ b/core/src/components/button/button.scss
@@ -202,15 +202,12 @@
   flex-shrink: 0;
   align-items: center;
   justify-content: center;
+  pointer-events: none;
 
   width: 100%;
   height: 100%;
 
   z-index: 1;
-}
-
-.button-inner {
-  pointer-events: none;
 }
 
 
--- a/core/src/components/button/test/basic/index.html
+++ b/core/src/components/button/test/basic/index.html
@@ -10,7 +10,7 @@
   <script src="../../../../../scripts/testing/scripts.js"></script>
   <script nomodule src="../../../../../dist/ionic/ionic.js"></script>
   <script type="module" src="../../../../../dist/ionic/ionic.esm.js"></script></head>
-  
+
 <body>
   <ion-app>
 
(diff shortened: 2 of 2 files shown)


amandaesmith3:
After internal discussion, we've tracked this down to a Webkit issue that's already been resolved as of Safari 15.2. Will post details on the original issue in a bit. Closing this out.

Issue #24826, opened 2022-02-21
bug: datetime intersection observers get disconnected when changing presentation programmatically


amandaesmith3:
### Prerequisites

- [X] I have read the [Contributing Guidelines](https://github.com/ionic-team/ionic-framework/blob/main/.github/CONTRIBUTING.md#creating-an-issue).
- [X] I agree to follow the [Code of Conduct](https://ionicframework.com/code-of-conduct).
- [X] I have searched for [existing issues](https://github.com/ionic-team/ionic-framework/issues) that already report this problem, without success.

### Ionic Framework Version

- [ ] v4.x
- [ ] v5.x
- [X] v6.x

### Current Behavior

When `ion-datetime` has its `presentation` prop changed programmatically, the intersection observers driving month selection seem to be getting disconnected. This causes all sorts of buggy behavior.

Screencast: https://user-images.githubusercontent.com/90629384/138342256-4571c211-d08b-473f-a96a-055e2a4886c8.mp4

### Expected Behavior

Month selection should function normally after changing the `presentation`.

### Steps to Reproduce

1. Visit the existing `display` test for `datetime`.
2. Change the "Presentation" dropdown to something else.
3. Try to change the month via the arrow buttons.

### Code Reproduction URL

http://localhost:3333/src/components/datetime/test/display

### Ionic Info

```
[WARN] You are not in an Ionic project directory. Project context may be missing.

Ionic:

   Ionic CLI : 6.17.1

Utility:

   cordova-res : 0.15.3
   native-run  : 1.4.1

System:

   NodeJS : v16.13.0
   npm    : 8.1.0
   OS     : Windows 10
```

amandaesmith3:
@arnotixe That sounds like something else; could you open a separate issue?

amandaesmith3:
Resolved via: https://github.com/ionic-team/ionic-framework/pull/24828

PR #24828, opened 2022-02-21
fix(datetime): reinit behavior on presentation change

amandaesmith3:
## Pull request checklist

## Pull request type

 

 

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

When the `presentation` prop on `datetime` is changed programmatically, behaviors such as the intersection observers and keyboard listeners are disconnected.

Issue URL: Resolves https://github.com/ionic-team/ionic-framework/issues/24826

## What is the new behavior?

- Behavior is re-inited when the presentation prop is changed.
- Several `calendarBodyRef` uses were updated to avoid being set to `null` due to a Stencil bug.

- [ ] Yes
- [x] No

### DIFF
--- a/core/src/components/datetime/datetime.tsx
+++ b/core/src/components/datetime/datetime.tsx
@@ -425,10 +425,10 @@ export class Datetime implements ComponentInterface {
      * the date that is currently selected, otherwise
      * there can be 1 hr difference when dealing w/ DST
      */
-    const date = new Date(convertDataToISO(this.workingParts));
-    this.workingParts.tzOffset = date.getTimezoneOffset() * -1;
+    const date = new Date(convertDataToISO(this.activeParts));
+    this.activeParts.tzOffset = date.getTimezoneOffset() * -1;
 
-    this.value = convertDataToISO(this.workingParts);
+    this.value = convertDataToISO(this.activeParts);
 
     if (closeOverlay) {
       this.closeParentOverlay();
--- a/core/src/components/datetime/datetime.tsx
+++ b/core/src/components/datetime/datetime.tsx
@@ -197,6 +197,8 @@ export class Datetime implements ComponentInterface {
    */
   @Prop() presentation: 'date-time' | 'time-date' | 'date' | 'time' | 'month' | 'year' | 'month-year' = 'date-time';
 
+  private prevPresentation: string | null = null;
+
   /**
    * The text to display on the picker's cancel button.
    */
@@ -657,7 +659,11 @@ export class Datetime implements ComponentInterface {
   }
 
   private initializeCalendarIOListeners = () => {
-    const { calendarBodyRef } = this;
+    /**
+     * Stencil sometimes sets calendarBodyRef to null on rerender, even though
+     * the element is present. Query for it manually as a fallback.
+     */
+    const calendarBodyRef = this.calendarBodyRef || this.el.shadowRoot?.querySelector('.calendar-body');
     if (!calendarBodyRef) { return; }
 
     const mode = getIonMode(this);
@@ -876,7 +882,17 @@ export class Datetime implements ComponentInterface {
     }
   }
 
-  componentDidLoad() {
+  componentDidRender() {
+    /**
+     * Only init all the logic if the presentation has changed.
+     * Note that this will include the initial render.
+     */
+    if (this.presentation === this.prevPresentation) {
+      return;
+    } else {
+      this.prevPresentation = this.presentation;
+    }
+
     /**
      * If a scrollable element is hidden using `display: none`,
      * it will not have a scroll height meaning we cannot scroll elements
--- a/core/src/components/datetime/datetime.tsx
+++ b/core/src/components/datetime/datetime.tsx
@@ -94,6 +94,7 @@ export class Datetime implements ComponentInterface {
 
   private destroyCalendarIO?: () => void;
   private destroyKeyboardMO?: () => void;
+  private destroyOverlayListener?: () => void;
 
   private minParts?: any;
   private maxParts?: any;
@@ -878,7 +879,7 @@ export class Datetime implements ComponentInterface {
    * listener. This is so that we can re-create the listeners
    * if the datetime has been hidden/presented by a modal or popover.
    */
-  private destroyListeners = () => {
+  private destroyInteractionListeners = () => {
     const { destroyCalendarIO, destroyKeyboardMO } = this;
 
     if (destroyCalendarIO !== undefined) {
@@ -945,7 +946,7 @@ export class Datetime implements ComponentInterface {
       const ev = entries[0];
       if (ev.isIntersecting) { return; }
 
-      this.destroyListeners();
+      this.destroyInteractionListeners();
 
       writeTask(() => {
         this.el.classList.remove('datetime-ready');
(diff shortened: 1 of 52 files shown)


core/src/components/datetime/datetime.tsx
```diff
@@ -657,7 +659,11 @@ export class Datetime implements ComponentInterface {
   }
 
   private initializeCalendarIOListeners = () => {
-    const { calendarBodyRef } = this;
+    /**
+     * Stencil sometimes sets calendarBodyRef to null on rerender, even though
+     * the element is present. Query for it manually as a fallback.
+     */
```
amandaesmith3, line 665: I'm admittedly not sure why this happens 🤔 It's easiest to see if you log the `el` passed in when the `ref` is set in `renderCalendarBody`. I was able to consistently trigger it going between `date-time` and `time-date` presentations.

sean-perkins: This might just be a quirk in the timing of the `componentDidRender` vs. `componentDidLoad`. Although it is strange if that behavior is happening between subsequent renders.

amandaesmith3: I thought that too, but I tried doing this:
```tsx
await new Promise(resolve => setTimeout(resolve, 2000));
console.log(this.calendarBodyRef)
```
and it was still `null`. You can also log the `el` in `renderCalendarBody` to see every time it updates; it'll fire twice, with the second one setting the ref to `null`.

sean-perkins: This looks like something that Stencil is doing under the hood when disposing a cached node. When the presentation changes, `renderDatetime` will return a different case and that must trigger behavior to tear down view refs before rendering the new case.

I think your fallback is appropriate considering this behavior (as long as it is the instance of the new `.calendar-body` and not the previously rendered instance).

core/src/components/datetime/datetime.tsx
```diff
@@ -197,6 +197,8 @@ export class Datetime implements ComponentInterface {
    */
   @Prop() presentation: 'date-time' | 'time-date' | 'date' | 'time' | 'month' | 'year' | 'month-year' = 'date-time';
 
+  private prevPresentation: string | null = null;
```
sean-perkins, line 200: ~ Doesn't look like we have a strong stance on this, but organizing private variables at the top of the class is a useful practice. If we moved this, we should probably move `todayParts` as well.

sean-perkins (approved):

liamdebeasi (changes requested):
Also, can we add E2E tests to verify this new behavior?

core/src/components/datetime/datetime.tsx
amandaesmith3: I'll grab some time with him to talk through it, and mark this as resolved for now 👍

core/src/components/datetime/datetime.tsx
```diff
@@ -959,6 +968,21 @@ export class Datetime implements ComponentInterface {
...
+   * so we need to re-init behavior with the new elements.
+   */
+  componentDidRender() {
+    if (this.prevPresentation === null) {
+      this.prevPresentation = this.presentation;
+    }
+
+    if (this.presentation === this.prevPresentation) { return; }
+    this.prevPresentation = this.presentation;
+    this.destroyListeners();
```
liamdebeasi, line 982: `destroyListeners` never destroys the overlay listener. Can we add this? otherwise we will have multiple `willPresent` callbacks firing.

core/src/components/datetime/datetime.tsx
```diff
@@ -959,6 +968,21 @@ export class Datetime implements ComponentInterface {
...
+  /**
+   * When the presentation is changed, all calendar content is recreated,
+   * so we need to re-init behavior with the new elements.
+   */
+  componentDidRender() {
+    if (this.prevPresentation === null) {
+      this.prevPresentation = this.presentation;
+    }
+
+    if (this.presentation === this.prevPresentation) { return; }
```
liamdebeasi, line 980: Can we cache the values of prevPresentation and presentation? Every time we access `this.[variable]` we incur a small perf hit in Stencil, so caching the value lets us avoid those extra hits.

Also can we return in the `null` check? If we make it into the body of the block, it's just going to return on line 980 anyways, so better to avoid the additional equality check.

```js
const { prevPresentation, presentation } = this;

if (prevPresentation === null) {
  this.prevPresentation = presentation; 
  return;
}

if (prevPresentation === presentation) { return; }

this.prevPresentation = presentation;

this.destroyListeners();
this.initializeListeners();
```

liamdebeasi (changes requested):

liamdebeasi:
I am noticing the following error getting thrown during the E2E tests:

```js
TypeError: Cannot read properties of undefined (reading 'initializeCalendarIOListeners')
    at initializeListeners (ion-datetime.entry.js:1659)
    at Datetime.componentDidRender (ion-datetime.entry.js:1749)
    at safeCall (index-765cc965.js:1580)
    at postUpdateComponent (index-765cc965.js:1481)
    at postUpdate (index-765cc965.js:1416) undefined
```

Perhaps this is why the test is failing?

amandaesmith3:
I'm still working on it; there are a few different issues that need addressing. I'll re-request review when it's ready again.

core/src/components/datetime/datetime.tsx
```diff
@@ -408,6 +415,13 @@ export class Datetime implements ComponentInterface {
    */
   @Event() ionStyle!: EventEmitter<StyleEventDetail>;
 
+  /**
+   * Emitted when the working parts change. Used for testing
+   * purposes, e.g. to ensure month selection works.
+   * @internal
+   */
+  @Event() ionWorkingPartsDidChange!: EventEmitter<DatetimeParts>;
```
amandaesmith3, line 423: This should replace the `InitMonthDidChangeEvent` function completely. I'll make a separate PR to update the other tests.

amandaesmith3:
Alright, should finally be good to go now! Whew.

core/src/components/datetime/datetime.tsx
amandaesmith3: I wasn't able to find one, but I'm open to ideas. The old util (`InitMonthDidChangeEvent`) fits the bill and uses a mutation observer to indirectly watch for the month to change, but it doesn't account for the month element being re-rendered (and thus the reference to the observed element being lost) when `presentation` is changed.

sean-perkins: Spent some time exploring an alternative way to test this behavior. Here's a branch that shows how we can drop the internal event from `ion-datetime`: https://github.com/ionic-team/ionic-framework/tree/test/FW-844.

It will apply a mutation observer to dispatch a custom event once the calendar body is manipulated. It will also re-attach the observer to the correct element node after the presentation is changed. 

If you like this approach, feel free to take any parts of it/merge it into your working branch.

amandaesmith3: @sean-perkins Thanks for that! I cleaned it up and brought it in 👍 cc @liamdebeasi

packages/vue/src/proxies.ts
```diff
@@ -295,7 +295,8 @@ export const IonDatetime = /*@__PURE__*/ defineContainer<JSX.IonDatetime>('ion-d
   'ionChange',
   'ionFocus',
   'ionBlur',
-  'ionStyle'
+  'ionStyle',
+  'ionWorkingPartsDidChange'
```
liamdebeasi, line 299: Re-running `npm run build` should get rid of this

liamdebeasi (approved):
Good to go other than my one comment

sean-perkins (approved):
🎉

Issue #24821, opened 2022-02-21
bug: datetime min attribute breaks the minutes auto selection


antoine92190:
### Prerequisites

- [X] I have read the [Contributing Guidelines](https://github.com/ionic-team/ionic-framework/blob/main/.github/CONTRIBUTING.md#creating-an-issue).
- [X] I agree to follow the [Code of Conduct](https://ionicframework.com/code-of-conduct).
- [X] I have searched for [existing issues](https://github.com/ionic-team/ionic-framework/issues) that already report this problem, without success.

### Ionic Framework Version

- [ ] v4.x
- [ ] v5.x
- [X] v6.x

### Current Behavior

Two issues which are probably related:
- First issue: when adding a min attribute, the minutes are changing even when they don't need to

https://user-images.githubusercontent.com/16256207/154915684-7767d3fb-2c53-4ee7-997c-628cf29f38d9.mov

- Second issue: when scrolling to minimum available hour, the minutes change to its minimum value, but is not taken into account for validation

https://user-images.githubusercontent.com/16256207/154916057-66494e5f-1226-48c3-bfe3-f8e08ddfe157.mov

### Expected Behavior

- The minutes should keep the same value unless it's in a disabled range
- The validated minutes value should always be updated when changing hours

### Steps to Reproduce

- Set `min` value at the current date and time
- Go back and forth between the first available hours and any other hour

### Ionic Info

```
Ionic:

   Ionic CLI                     : 6.18.1 (/Users/antoinedupont/.npm-packages/lib/node_modules/@ionic/cli)
   Ionic Framework               : @ionic/angular 6.0.8-dev.1644432569.937206e
   @angular-devkit/build-angular : 12.2.4 (/Users/antoinedupont/dev/agorapulse/platform/frontend/node_modules/@angular-devkit/build-angular)
   @angular-devkit/schematics    : 11.2.18
   @angular/cli                  : 12.2.4 (/Users/antoinedupont/dev/agorapulse/platform/frontend/node_modules/@angular/cli)
   @ionic/angular-toolkit        : 3.1.1

Capacitor:

   Capacitor CLI      : 3.2.4
   @capacitor/android : 3.2.4
   @capacitor/core    : 3.2.4
   @capacitor/ios     : 3.2.4

Cordova:

   Cordova CLI       : not installed
   Cordova Platforms : not available
   Cordova Plugins   : not available

Utility:

   cordova-res : not installed globally
   native-run  : 1.5.0

System:

[...]

PR #24829, opened 2022-02-21
fix(datetime): persist minutes column on hour change

sean-perkins:
## Pull request checklist

## Pull request type

 

 

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

When using a `min` value on `ion-datetime` and changing the hour value, the state of the component is updated and re-renders. This causes the minutes column to become disconnected from the actual selected minutes value. This results in the displayed time value being different than the column picker display.

Issue URL: #24821

## What is the new behavior?

When the picker column is re-rendered, if the value is out of bounds or not scrolled correctly in the view, the picker will attempt to set the active item or scroll the active item into view.

- [ ] Yes
- [x] No

### DIFF
--- a/core/src/components/datetime/datetime.tsx
+++ b/core/src/components/datetime/datetime.tsx
@@ -83,6 +83,7 @@ export class Datetime implements ComponentInterface {
   private inputId = `ion-dt-${datetimeIds++}`;
   private calendarBodyRef?: HTMLElement;
   private popoverRef?: HTMLIonPopoverElement;
+  private minutesPickerColumnRef?: HTMLIonPickerColumnInternalElement;
   private clearFocusVisible?: () => void;
   private overlayIsPresenting = false;
 
@@ -1345,7 +1346,7 @@ export class Datetime implements ComponentInterface {
     ampmItems: PickerColumnItem[],
     use24Hour: boolean
   ) {
-    const { color, activePartsClone, workingParts } = this;
+    const { color, activePartsClone, workingParts, minutesPickerColumnRef } = this;
 
     return (
       <ion-picker-internal>
@@ -1364,10 +1365,20 @@ export class Datetime implements ComponentInterface {
               hour: ev.detail.value
             });
 
+            if (minutesPickerColumnRef) {
+              raf(() => {
+                // When the time picker is re-rendered, the minutes columns does not
+                // scroll to the correct position. This is a workaround to ensure
+                // that the minutes column displays the correct value.
+                minutesPickerColumnRef!.scrollActiveItemIntoView();
+              });
+            }
+
             ev.stopPropagation();
           }}
         ></ion-picker-column-internal>
         <ion-picker-column-internal
+          ref={ref => this.minutesPickerColumnRef = ref}
           color={color}
           value={activePartsClone.minute}
           items={minutesItems}
--- a/core/src/components/datetime/datetime.tsx
+++ b/core/src/components/datetime/datetime.tsx
@@ -83,7 +83,6 @@ export class Datetime implements ComponentInterface {
   private inputId = `ion-dt-${datetimeIds++}`;
   private calendarBodyRef?: HTMLElement;
   private popoverRef?: HTMLIonPopoverElement;
-  private minutesPickerColumnRef?: HTMLIonPickerColumnInternalElement;
   private clearFocusVisible?: () => void;
   private overlayIsPresenting = false;
 
@@ -1346,7 +1345,7 @@ export class Datetime implements ComponentInterface {
     ampmItems: PickerColumnItem[],
     use24Hour: boolean
   ) {
-    const { color, activePartsClone, workingParts, minutesPickerColumnRef } = this;
+    const { color, activePartsClone, workingParts } = this;
 
     return (
       <ion-picker-internal>
@@ -1365,20 +1364,10 @@ export class Datetime implements ComponentInterface {
               hour: ev.detail.value
             });
 
-            if (minutesPickerColumnRef) {
-              raf(() => {
-                // When the time picker is re-rendered, the minutes columns does not
-                // scroll to the correct position. This is a workaround to ensure
-                // that the minutes column displays the correct value.
-                minutesPickerColumnRef!.scrollActiveItemIntoView();
-              });
-            }
-
             ev.stopPropagation();
           }}
         ></ion-picker-column-internal>
         <ion-picker-column-internal
-          ref={ref => this.minutesPickerColumnRef = ref}
           color={color}
           value={activePartsClone.minute}
           items={minutesItems}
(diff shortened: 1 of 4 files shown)


core/src/components/datetime/datetime.tsx
```diff
@@ -1364,10 +1365,20 @@ export class Datetime implements ComponentInterface {
               hour: ev.detail.value
             });
 
+            if (minutesPickerColumnRef) {
+              raf(() => {
+                // When the time picker is re-rendered, the minutes columns does not
+                // scroll to the correct position. This is a workaround to ensure
+                // that the minutes column displays the correct value.
+                minutesPickerColumnRef!.scrollActiveItemIntoView();
```
liamdebeasi, line 1373: Are we able to solve this inside of `ion-picker-column-internal`? Seems to be specific to this component's rendering, not datetime.

sean-perkins: Tried solving from inside `ion-picker-column-internal`, but even though the watch callback for value is called on the re-render, the double state variable change (`activeParts` and `workingParts`), seems to cause issues. I tried also request animation frame wrapping the scroll active item within the picker and that also did not work.

It seems like the behavior happening is:
- Hour column is changed
- `workingParts` are updated 
  - re-render in progress
  - Assigns new value to `ion-picker-column-internal`
  - Scroll active item into view fires
 - `activeParts` are updated
   - re-render in progress
   - Conflicts with change from previous re-render / `ion-picker-column-internal` scroll position is reset

Open to alternatives here, the two state variables being updated in a single call is a problem in general that might require a larger discovery.

amandaesmith3 (approved):

core/src/components/datetime/datetime.tsx
liamdebeasi: Doing the following inside of `picker-column-internal` fixes the issue on my end:

```js
componentDidRender() {
    this.scrollActiveItemIntoView();
  }
```

We would likely want to do some additional checks here to make sure we don't do an excessive amount of scroll requests here.

This is the test case I used:

```html
<ion-datetime min="2022-02-24T10:15:00" value="2022-02-24T10:16:00"></ion-datetime>
```

1. Open time picker. Observe that it starts on 10:16.
2. Select 11 in the hour column. Observe that the time is 11:16.

-----

One thing we should probably figure out after we deal with the above is what to do if the value that you had selected is no longer available. If you were on 11:10 and then moved the hour to be 10. The minutes are now out of bounds -- should we select the minimum value?

liamdebeasi: Here is a video of the fix in action:
https://user-images.githubusercontent.com/2721089/155559699-16a971df-7e4b-46b3-9b6e-6deb5893447a.mov

sean-perkins: This is awesome, thanks! I had not thought of hooking into the render lifecycle. I'm going to experiment with this and some other changes I was exploring around minute values no longer being available (as a result of min/max ranges). 

I'll need to think more on what the picker should do with values that are no longer available. For time selections, I think it could make sense that it scrolls the wheel to the zero position (whatever the minimum bounds is). For more generic options, such as a picker wheel of colors, defaulting to the zero position seems incorrect and a blank starting point would make sense. This could be controlled based on what options you render though (passing in a blank option if you don't want a default selection for out of bounds values). 

I'll also need to make sure that if a time value (i.e. minute) is unavailable and it scrolls the active item state to `00` for instance, that it would actually commit a value change. Otherwise users would incorrectly believe their minute value is `00`, even though the active value is something not visible. It would also require users to scroll to the next item and then to the previous, to lock in `00`.

core/src/components/picker-column-internal/picker-column-internal.tsx
```diff
@@ -118,6 +118,26 @@ export class PickerColumnInternal implements ComponentInterface {
...
+          // Scroll the active item into the view if after rendering
+          // it is still not visible
+          this.scrollActiveItemIntoView();
+        }
+      } else if (items[0].value !== value) {
+        // If the picker column does not have an active item and the current value
+        // does not match the first item in the picker column, that means
+        // the value is out of bounds. In this case, we assign the value to the
+        // first item to match the scroll position of the column.
+        this.value = items[0].value;
```
sean-perkins, line 136: Need to see where Ryan lands on the infinite loop/render Stencil PR. This works for now without any noticeable issues besides the developer warning from Stencil when programmatically assigning a value. 

This does not solve cases where developers assign a value that is less than the min or greater than the max. Since the date time does not enforce or sanitize values, they can still have values that are unavailable to the picker column. This really isn't a problem worth solving now, likely an implementation detail for implementers.

core/src/components/picker-column-internal/picker-column-internal.tsx
```diff
@@ -118,6 +118,26 @@ export class PickerColumnInternal implements ComponentInterface {
...
+        if (el.scrollTop !== this.getScrollTopPositionForElement(activeItem)) {
+          // Scroll the active item into the view if after rendering
+          // it is still not visible
+          this.scrollActiveItemIntoView();
+        }
+      } else if (items[0].value !== value) {
+        // If the picker column does not have an active item and the current value
+        // does not match the first item in the picker column, that means
+        // the value is out of bounds. In this case, we assign the value to the
+        // first item to match the scroll position of the column.
```
liamdebeasi, line 135: Can we use multi line comments in places where we have comments on multiple lines?

core/src/components/picker-column-internal/picker-column-internal.tsx
```diff
@@ -118,6 +118,26 @@ export class PickerColumnInternal implements ComponentInterface {
...
+    const { activeItem, el, items, isColumnVisible, value } = this;
+
+    if (isColumnVisible) {
+      if (activeItem) {
+        if (el.scrollTop !== this.getScrollTopPositionForElement(activeItem)) {
+          // Scroll the active item into the view if after rendering
+          // it is still not visible
+          this.scrollActiveItemIntoView();
+        }
+      } else if (items[0].value !== value) {
```
liamdebeasi, line 131: Need to account for the empty case where `items` is `[]`

```suggestion
      } else if (items[0]?.value !== value) {
```

core/src/components/picker-column-internal/picker-column-internal.tsx
```diff
@@ -118,6 +118,26 @@ export class PickerColumnInternal implements ComponentInterface {
     }
   }
 
+  componentDidRender() {
+    const { activeItem, el, items, isColumnVisible, value } = this;
+
+    if (isColumnVisible) {
+      if (activeItem) {
+        if (el.scrollTop !== this.getScrollTopPositionForElement(activeItem)) {
```
liamdebeasi, line 126: Can we add this check inside of `centerPickerItemInView` instead? 

That way other code that calls `centerPickerItemInView` can benefit from this check and we only need to call `getScrollTopPositionForElement` once.

sean-perkins: Updated, let me know if this aligns with what you were thinking.

liamdebeasi (approved):

Issue #24817, opened 2022-02-18
bug: Swipe gesture (left to right) on overlay menu not working after showing a modal


MawsFr:
### Prerequisites

- [X] I have read the [Contributing Guidelines](https://github.com/ionic-team/ionic-framework/blob/main/.github/CONTRIBUTING.md#creating-an-issue).
- [X] I agree to follow the [Code of Conduct](https://ionicframework.com/code-of-conduct).
- [X] I have searched for [existing issues](https://github.com/ionic-team/ionic-framework/issues) that already report this problem, without success.

### Ionic Framework Version

- [ ] v4.x
- [ ] v5.x
- [X] v6.x

### Current Behavior

Hello,

I have a splitpane with a menu that I can open with a swipe gesture (left to right) . In the content of the splitpane I have a button that opens a modal. When I open the modal and I close it (manually or with the backdrop), I can't open the menu with the swipe gesture anymore.

### Expected Behavior

I would like to still be able to open the menu with the swipe gesture even after closing the modal.

So what I expect (Without the workaround I found) :
- [Go here](https://dashboard.ionicframework.com/preview/134aecb4/p5ozjlwzv6)
- Swipe from left to right to show the menu
- Swipe right to left to hide it
- Click on "open modal"
- Click on "close modal and remove show-modal class
- Be able to swipe the menu like here

### Steps to Reproduce

- [Go here](https://dashboard.ionicframework.com/preview/134aecb4/p5ozjlwzv6)
- Swipe from left to right to show the menu
- Swipe right to left to hide it
- (Until here the menu works fine)
- Click on "open modal"
- Close it by any mean (button "close modal" or backdrop or swipe down)
- Try to open the menu with the swipe gesture
- => Not working anymore

### Code Reproduction URL

https://github.com/MawsFr/test-modal-with-menu-swipe-gesture

### Ionic Info

```
[WARN] Error loading @capacitor/ios package.json: Error: Cannot find module
       '@capacitor/ios/package'
       
       Require stack:
       -
       /Users/maws/.nvm/versions/node/v12.18.3/lib/node_modules/@ionic/cli/lib/project/index.js
       -
       /Users/maws/.nvm/versions/node/v12.18.3/lib/node_modules/@ionic/cli/lib/index.js
       -
       /Users/maws/.nvm/versions/node/v12.18.3/lib/node_modules/@ionic/cli/index.js
       -
[...]

PR #24846, opened 2022-02-24
fix(modal): re-enable swipe gestures when modal is dismissed

sean-perkins:
## Pull request checklist

## Pull request type

 

 

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

When a modal is opened, the `.show-modal` class is applied. This class prevents the gesture for opening `ion-menu`. The class is never removed when a modal is dismissed, so when the modal is closed, users are unable to open menus through swipe gestures. 

Issue URL: #24817

## What is the new behavior?

When the modal is dismissed, the `.show-modal` class is removed from the host element. This allows the menu to be opened through a swipe gesture when the modal is dismissed.

- [ ] Yes
- [x] No

## Other information

Confirmed in the reproduction app with dev build `6.0.10-dev.1645727760.16b0a89`.

### DIFF
--- a/core/src/components/menu/test/focus-trap/e2e.ts
+++ b/core/src/components/menu/test/focus-trap/e2e.ts
@@ -12,7 +12,7 @@ test('menu: focus trap with overlays', async () => {
 
   const ionDidOpen = await page.spyOnEvent('ionDidOpen');
   const ionModalDidPresent = await page.spyOnEvent('ionModalDidPresent');
-  const ionModalDidDismiss= await page.spyOnEvent('ionModalDidDismiss');
+  const ionModalDidDismiss = await page.spyOnEvent('ionModalDidDismiss');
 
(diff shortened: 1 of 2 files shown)


liamdebeasi (approved):
Code looks good, but we should add a test if we can. I know this only happens with gestures + Puppeteer and gestures are flaky so we might need to loop back to this in the future.

sean-perkins:
> Code looks good, but we should add a test if we can. I know this only happens with gestures + Puppeteer and gestures are flaky so we might need to loop back to this in the future.

Sounds good, happy to take a stab at it and see if there's an approach that isn't as flaky.

Also, the reporter confirmed the behavior is resolved in their app: https://github.com/ionic-team/ionic-framework/issues/24817#issuecomment-1050166588 👍

amandaesmith3 (approved):

Issue #24747, opened 2022-02-08
bug: disabled item-sliding interactive after calling close()


Mathieu-COSYNS:
### Prerequisites

- [X] I have read the [Contributing Guidelines](https://github.com/ionic-team/ionic-framework/blob/main/.github/CONTRIBUTING.md#creating-an-issue).
- [X] I agree to follow the [Code of Conduct](https://ionicframework.com/code-of-conduct).
- [X] I have searched for [existing issues](https://github.com/ionic-team/ionic-framework/issues) that already report this problem, without success.

### Ionic Framework Version

- [ ] v4.x
- [ ] v5.x
- [X] v6.x

### Current Behavior

As expected users can not open options of a disabled ion-item-sliding.
Then somewhere is the code I call .close() on this disabled ion-item-sliding.
After that users can open the options.

### Expected Behavior

As expected users can not open options of a disabled ion-item-sliding.
Then somewhere is the code I call .close() on this disabled ion-item-sliding.
After that users can still **not** open the options.

### Steps to Reproduce

Call .close() on a disabled ion-item-sliding then try to open the options you will see that it opens. 

### Code Reproduction URL

https://github.com/Mathieu-COSYNS/ionic_bug_report_ion-item-sliding

### Ionic Info

Ionic:

   Ionic CLI       : 6.18.1 (/home/mc/.local/share/npm/lib/node_modules/@ionic/cli)
   Ionic Framework : @ionic/react 6.0.5

Capacitor:

   Capacitor CLI      : 3.4.0
   @capacitor/android : not installed
   @capacitor/core    : 3.4.0
   @capacitor/ios     : not installed

Utility:

   cordova-res : not installed globally
   native-run  : 1.5.0

System:

   NodeJS : v14.17.0 (/home/mc/.config/nvm/versions/node/v14.17.0/bin/node)
   npm    : 6.14.13
   OS     : Linux 5.4

PR #24847, opened 2022-02-24
fix(item-sliding): close() will maintain disabled state

sean-perkins:
## Pull request checklist

## Pull request type

 

 

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

When calling the `close()` API, it will always re-enable the swipe gesture regardless of the disabled state. This is undesired behavior, as it will make disabled items interactive. 

Issue URL: #24747

## What is the new behavior?

When closing items, the API will set the state of the gesture to the opposite of the disabled state. This is consistent with the behavior of the disabled changed behavior. This keeps disabled items disabled from swiping.

- [ ] Yes
- [x] No

## Other information

Writing a test for this behavior was unsuccessful, open to input/suggestions. Tried to write an extension of the interactive test that disables all items, closes all items and then attempts to swipe to check if there is an active item. Doing this test, I could not get the test case to fail (with the old and current code). Exploring the test in the browser, I was unable to get the same behavior.

```
test('item-sliding: should remain disabled when programmatically closed', async () => {
  const page = await newE2EPage({
    url: '/src/components/item-sliding/test/interactive?ionic:_testing=true'
  });

  const disableBtn = await page.find('#disable-btn');
  const closeBtn = await page.find('#close-btn');

  await disableBtn.click();

  await page.waitForChanges();

  await closeBtn.click();

  await page.waitForChanges();

  await openItemSliding('#item-0', page);

  const item = await page.find('#item-0');

  expect(item.classList.contains('item-sliding-active-slide')).toBe(false);

});
```

### DIFF
--- a/core/src/components/item-sliding/item-sliding.tsx
+++ b/core/src/components/item-sliding/item-sliding.tsx
@@ -408,7 +408,7 @@ export class ItemSliding implements ComponentInterface {
         this.state = SlidingState.Disabled;
         this.tmr = undefined;
         if (this.gesture) {
-          this.gesture.enable(true);
+          this.gesture.enable(!this.disabled);
         }
       }, 600) as any;
 


liamdebeasi (approved):
Code looks good. It might be tough to test this due to the 600ms setTimeout. Even if you do `page.waitFor(600)` there's no guarantee that your click event will fire at the right time.

sean-perkins:
@liamdebeasi dang that is totally what is affecting the ability to write tests on this. 

I did confirm manually with a modification to the test template to disable & close all, then manually verify gestures.

|Issue|PR|
|---|---|
|<video src="https://user-images.githubusercontent.com/13732623/155743966-16b515f4-0007-49fd-8823-1368cec89023.mp4"></video>|<video src="https://user-images.githubusercontent.com/13732623/155744255-27e756a8-935c-438a-96d2-dea4bd541690.mp4"></video>|

We could explore internal props to configure timeout intervals, if we determined test coverage was valuable here.

Issue #22868, opened 2021-02-04
bug: Ion input not binding value to the state instead binding to what's been entered.


Richa-Singh-1:
# Bug Report

**Ionic version:**
[x] 5.x

**Current behavior:**
For ion input value in input tag is shown based on what is entered(event.target.value) instead of what's been set through value binding via state.
Video : https://drive.google.com/file/d/1CHxSKay65hKG--5yWN-6yXa7FzYkv5L0/view?usp=sharing

**Expected behavior:**
Value should be shown based on what's been set through state .

**Steps to reproduce:**
Please run below commands in  the project :https://github.com/Richa-Singh-1/IonInput-Issue
npm i
npm run start

**Related code:**

```
const App: React.FC = () => {
  const [input, setInput] = useState('');

  const handleInputChange = (val) => {
    console.log(' even.detail.value  ', val);
    const input = val.replace(/\D/g, "");
    console.log('input  ', input);
    setInput(input);
  }
  return (
    <IonApp>
      <IonPage>
        <IonItem>
          <IonInput
            inputmode="numeric"
            maxlength={3}
            value={input}
            placeholder="Enter Input"
            onIonChange={e => handleInputChange(e.detail.value)}>
          </IonInput>

        </IonItem>
      </IonPage>
    </IonApp>
  );
};
```

**Other information:**

**Ionic info:**

```
Ionic:

   Ionic CLI       : 6.12.3 (/usr/local/lib/node_modules/@ionic/cli)
   Ionic Framework : @ionic/react 5.4.0

Capacitor:

   Capacitor CLI   : 2.4.2
   @capacitor/core : 2.4.2

Utility:

   cordova-res : not installed
   native-run  : not installed

System:

   NodeJS : v12.19.0 (/usr/local/bin/node)
   npm    : 6.14.8
   OS     : macOS Catalina
```

liamdebeasi:
Thanks for the issue. Can you please double check that your repo is set to public? I cannot access it.

Richa-Singh-1:
@liamdebeasi  updated

Richa-Singh-1:
@liamdebeasi any updates on the issue?

amitkharka:
I'm also facing the same issue. Looks like Stencil **input** components is not behaving as expected.  
https://github.com/ionic-team/ionic-framework/blob/master/core/src/components/input/input.tsx

dtarnawsky:
[...]

PR #24858, opened 2022-03-01
fix(react): re-render component from non-React capture phase listener

sean-perkins:
## Pull request checklist

## Pull request type

 

 

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

In situations where developers are updating a React component's value using `useState`, within an event listener from a change event; if the value matches the previous value, the component will not correctly repaint and the user can enter invalid input.

An example of this would be replacing an input's value to only accept numeric digits. If a user was to enter "12b", the React's component value would be "12", but the native input element within our `ion-input` would paint "12b"; allowing users to continue to enter invalid characters visibly, but the value would be formatted.  

This is caused by this bug in React: #13424.

Issue URL: #22868

## What is the new behavior?

- `ion-textarea`, `ion-input` and `ion-searchbar` are no longer auto-generated component wrappers/proxies
- `ion-textarea`, `ion-input` and `ion-searchbar` will add an event listener for `ionInput` that compares the React component's state value against the current web component value to determine when the component is out of sync.
- When the component is out of sync, it will update the React component's state to match the web component's value.

Note: `queueMicrotask` is required to allow the event to dispatch and any event handlers to complete, but will update the component's state prior to render. Otherwise the control's value will be set to its last evaluated value, instead of the updated/formatted value. Alternatively `requestAnimationFrame` could be used, but `queueMicrotask` will execute earlier in the task queue. 

You can test with this reproduction app: https://github.com/sean-perkins/react-capture-phase-demo
- `main` branch is of latest @ionic/react
- `pr` branch is of dev build

The reproduction app showcases a standard form using `useState` as well as a react-hook-form example.

- [ ] Yes
- [x] No

## Other information

|Issue|PR|
|---|---|
|<video src="https://user-images.githubusercontent.com/13732623/156257630-1d5e1ed3-ce2e-46b1-a9cf-fa64dbdeb249.mp4"></video>|<video src="https://user-images.githubusercontent.com/13732623/156257859-27ae941e-167b-4179-ab1c-a22e26e0f1e4.mp4"></video>


sean-perkins:
Need to check how this affects implementations using packages such as `react-hook-forms`. Need to potentially disable the comparison behavior with regards to `ionChange` event handlers. It currently prevents the component from re-rendering with a value, if you `setState` in the `ionChange` event phase.

sean-perkins:
Updated the original PR description to include a reproduction app to test and verify the behavior between latest `@ionic/react` and the dev build (`6.0.10-dev.11646429436.2d71c72`): https://github.com/sean-perkins/react-capture-phase-demo.

In the demo, you can see the behavior between a standard HTML form using `useState` as well as with `react-hook-form`. I also showcase the difference in behavior between mutating a control's value in the `ionChange` event vs. `ionInput` event. My recommendation is that developers use `ionInput` for these mutations, as you will see the key is briefly painted prior to the value being set & re-rendered. 

Video recordings:

Standard HTML form:
|Main|PR|
|---|---|
|<video src="https://user-images.githubusercontent.com/13732623/156849055-598dbf5b-1dc1-42de-9e05-909e70df1321.mp4"></video>|<video src="https://user-images.githubusercontent.com/13732623/156849371-aa4160fc-c322-4ca1-8b5d-7ba54c989e65.mp4"></video>|


React hook form:
|Main|PR|
|---|---|
|<video src="https://user-images.githubusercontent.com/13732623/156849218-10bbfeec-5e52-4458-b3a5-867eaea1174d.mp4"></video|<video src="https://user-images.githubusercontent.com/13732623/156849462-cd64dc60-69f8-413b-95dc-0c104aed3cd7.mp4"></video>|

packages/react/test-app/src/pages/components/Components.tsx
```diff
@@ -0,0 +1,21 @@
...
+
+interface ComponentsProps {}
+
+const Components: React.FC<ComponentsProps> = () => {
+  return (
+    <IonRouterOutlet>
+      <Redirect from="/components" to="/components/input" exact />
+      <Route path="/components/input" component={InputComponent} />
+      <Route path="/components/textarea" component={TextareaComponent} />
+      <Route path="/components/searchbar" component={SearchbarComponent} />
```
amandaesmith3, line 16: These three pages are basically the same, save for the component used; could they be merged?

amandaesmith3 (changes requested):

amandaesmith3:
This would also be a good one to post a dev build on the original issue if possible.

packages/react/test-app/src/pages/components/Components.tsx
amandaesmith3: Works for me 👍

amandaesmith3 (approved):

packages/react/src/components/createValueTrackedComponent.tsx
```diff
@@ -0,0 +1,133 @@
...
+
+  const displayName = dashToPascalCase(tagName);
+  const ReactComponent = class extends React.Component<StencilReactInternalProps<ElementType>> {
+    componentEl!: ElementType;
+
+    setComponentElRef = (element: ElementType) => {
+      this.componentEl = element;
+    };
+
+    updateValueIfChanged = (ev: any) => {
```
liamdebeasi, line 46: Rather than implement this at the React level, have we considered having `ion-input` wait for `ionChange` to fire before `ionInput` is fired? Something like this:

```diff
private onInput = (ev: Event) => {
    const input = ev.target as HTMLInputElement | null;
    if (input) {
      this.value = input.value || '';

+    await waitForIonChangeToFire();
    }
    this.ionInput.emit(ev as InputEvent);
}
```

This ensures that the value is committed to `ion-input` before any formatting is done. It also ensures that `ionChange` fires before `ionInput`, which is aligned with native `inputs` where `change` fires before `input`. Currently, `ionInput` fires before `ionChange`.

Eric also noted that this impacts Ionic Angular, so implementing this in core should fix the issue there too: https://github.com/ionic-team/ionic-framework/issues/22868#issuecomment-897059622

liamdebeasi (commented):
The following line links to the wrong GitHub issue: `This is caused by this bug in React: https://github.com/ionic-team/ionic-framework/issues/13424.`. Can we fix?

packages/react/src/components/createValueTrackedComponent.tsx
sean-perkins: I like the idea of trying to address this problem across both, my concern in doing so was changing existing behavior/causing a breaking change. 

If we feel that event ordering wouldn't meet that criteria, I can investigate this approach more!

sean-perkins:
We are going to close out this PR for now and re-evaluate the resolution after v7 discovery takes place. While this fix does patch the very specific issue in React, it is still present in other framework outputs and introduces implementation changes that are difficult to gauge side effects for.

We will keep the branch around for reference and future reflection. 

Thanks!

PR #24863, opened 2022-03-02
Update readme.md

flroent:
Extra <script> tag


liamdebeasi:
Thanks for the PR! The file you edited it autogenerated, so the changes will get overwritten. Can you make the change in `components/loading/usage/vue.md` and then run `npm run build` instead?

liamdebeasi:
Closing in favor of https://github.com/ionic-team/ionic-framework/pull/24865

Issue #24878, opened 2022-03-04
bug: initial value in ion-datetime with "time" presentation not shown


sburnicki:
### Prerequisites

- [X] I have read the [Contributing Guidelines](https://github.com/ionic-team/ionic-framework/blob/main/.github/CONTRIBUTING.md#creating-an-issue).
- [X] I agree to follow the [Code of Conduct](https://ionicframework.com/code-of-conduct).
- [X] I have searched for [existing issues](https://github.com/ionic-team/ionic-framework/issues) that already report this problem, without success.

### Ionic Framework Version

- [ ] v4.x
- [ ] v5.x
- [X] v6.x

### Current Behavior

ion-datetime with presentation="time" and an initial value shows the wrong value ("12:00 AM") until you interact with the component (click in the video).
https://user-images.githubusercontent.com/5701519/156773032-af44b34f-cf27-48e3-8bc5-86c513303cd5.mp4

### Expected Behavior

The datetime component  with presentation="time" should initially show the time passed via "value" attribute.

### Steps to Reproduce

* add a datetime component `<ion-datetime presentation="time" [value]="'16:23'"></ion-datetime>`
* go to the page with the component
* the component shows "12:00 AM" instead of "4:23 PM"

### Code Reproduction URL

https://github.com/sburnicki/ionic-bug-datetime-time-initial-value

### Ionic Info

```
Ionic:

   Ionic CLI                     : 6.18.1 
   Ionic Framework               : @ionic/angular 6.0.10
   @angular-devkit/build-angular : 13.0.4
   @angular-devkit/schematics    : 13.0.4
   @angular/cli                  : 13.0.4
   @ionic/angular-toolkit        : 5.0.3

Capacitor:

   Capacitor CLI      : 3.4.2
   @capacitor/android : not installed
   @capacitor/core    : 3.4.2
   @capacitor/ios     : not installed

Utility:

   cordova-res : not installed globally
   native-run  : 1.5.0

System:

   NodeJS : v16.13.1 
   npm    : 8.1.2
   OS     : Linux 5.11
```

liamdebeasi:
Thanks for the issue. I can reproduce this behavior. This behavior changed in https://github.com/ionic-team/ionic-framework/commit/aacb58a3224e3cc51c731d0c9aa52f52c9276692 according to `git bisect`. cc @sean-perkins

sburnicki:
Thanks for the insanely fast response and fix.
[...]

PR #24879, opened 2022-03-04
fix(datetime): time picker now scrolls to correct value

liamdebeasi:
## Pull request checklist

## Pull request type

 

 

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

Issue URL: resolves https://github.com/ionic-team/ionic-framework/issues/24878

The problem is that `scrollActiveItemIntoView` is called before `isColumnVisible` is set to `true`: https://github.com/ionic-team/ionic-framework/blob/main/core/src/components/picker-column-internal/picker-column-internal.tsx#L100

This did not used to be a problem, but with the introduction of the `isColumnVisible` check in https://github.com/ionic-team/ionic-framework/blob/main/core/src/components/picker-column-internal/picker-column-internal.tsx#L152, this caused the initial scroll to no longer happen.

## What is the new behavior?

- Moved `isColumnVisible` to be set to `true` before calling the scroll operation.

- [ ] Yes
- [x] No

## Other information

I opted not to add a new test as we do already have a test for this, but the test suite that checks is not working. I am working on fixing the test suite with my Playwright migration.

### DIFF
--- a/core/src/components/picker-column-internal/picker-column-internal.tsx
+++ b/core/src/components/picker-column-internal/picker-column-internal.tsx
@@ -91,6 +91,7 @@ export class PickerColumnInternal implements ComponentInterface {
       const ev = entries[0];
 
       if (ev.isIntersecting) {
+        this.isColumnVisible = true;
         /**
          * Because this initial call to scrollActiveItemIntoView has to fire before
          * the scroll listener is set up, we need to manage the active class manually.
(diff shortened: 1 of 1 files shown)


amandaesmith3 (approved):

sean-perkins (approved):
Awesome, thanks for fixing this 👍

PR #24883, opened 2022-03-04
feat(content, reorder-group, header, footer, infinite-scroll, refresher): allow custom scroll target

sean-perkins:
## Pull request checklist

## Pull request type

 

 

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes

The following components are dependent on an implementation using `ion-content`, which is not compatible in implementations that use a custom scroll container, such as virtual-scroll.

Affected components:
- `ion-header` - fade effect and large titles (iOS)
- `ion-footer` - fade effect (iOS)
- `ion-infinite-scroll` - scrolling
- `ion-reorder-group` - ability to drag/reorder items
- `ion-refresher` - ability to pull to refresh

Issue URL:  #23437

## What is the new behavior?

Developers can make use of a new global CSS class selector: `.ion-content-scroll-host` to specify to implementations using the above components/features, which container is the scroll target to attach listeners to. 

Developers using a custom scroll containers should additionally consider disabling scrolling on the host `ion-content` to prevent duplicate scroll containers (`scroll-y="false"` for vertical scrolling and `scroll-x="false"` for horizontal scrolling).

- `ion-header` fade effect can be used with a custom scroll container
- `ion-header` large/collapsible title can be used with a custom scroll container
- `ion-footer` fade effect can be used with a custom scroll container
- `ion-infinite-scroll` can be used with a custom scroll container (emitting `ionInfinite` when the container reaches threshold)
- `ion-reorder-group` can be used with a custom scroll container (allows reordering elements within target container)
- `ion-refresher` can be used with a custom scroll container (preserves iOS and MD animation effects without needing to specify a custom `#background-content` element)

All of the above features can be used against virtual scroll implementations.

- [ ] Yes
- [x] No

## Other information

There were many areas of the implementation that were lacking e2e coverage of functionality, the following components had tests filled in, to cover the behavior changed with this PR: `ion-infinite-scroll`, `ion-refresher`, `ion-footer` (fade effect is exclusive to iOS).


sean-perkins:
@tlaverdure this will officially implement the functionality desired by your PR (#24426) with a static selector `.ion-content-scroll-host`. This selector name is still dependent on the team's review and could change if a better name is suggested. 

The intention of this feature is not to support ejecting from `ion-content` implementations, but it is possible with the resulting feature. If you decide to do so, you become responsible for all styling concerns that `ion-content` is currently handling for you as well as opinionated implementations such as the `#background-content` element for pull-to-refresh animations.

The intended use case is for better supporting virtual scroll with our components. With those implementations, you can disable the `ion-content` scroll container, but benefit from all internal styles and not needing to re-create the `#background-content` element for pull-to-refresh implementations. 

A dev build to test with is available: `6.0.10-dev.11646422799.5f745c5`.

I am going to close out your PR now that this feature is officially in our active work. I appreciate your time with that discovery as well as the dialog for refining the feature. If you have additional feedback or have an opportunity to test with the dev build and have concerns/issues, please let me know!

tlaverdure:
Thanks @sean-perkins! I'm going to test it out next week and let you know how it goes.

core/src/utils/content/index.ts
```diff
@@ -0,0 +1,130 @@
...
+export const scrollByPoint = (el: HTMLElement, x: number, y: number, durationMs: number) => {
+  if (el.tagName.toLowerCase() === ION_CONTENT_ELEMENT_SELECTOR) {
+    const content = el as HTMLIonContentElement;
+    return content.scrollByPoint(x, y, durationMs);
+  }
+  return Promise.resolve(el.scrollTo({
+    top: y,
+    left: x,
+    behavior: durationMs > 0 ? 'smooth' : 'auto'
+  }));
```
amandaesmith3, line 121: The comment references `scrollToPoint`, and the use of `scrollTo` would mirror that behavior, but the actual function is called and uses `scrollByPoint`, which is slightly different. (Check their implementations in `ion-content`.) It looks like `scrollByPoint` was your intention here since it's used in the `scroll-assist` shim, so the comment and fallback should be updated.

amandaesmith3 (changes requested):

core/src/utils/content/index.ts
```diff
@@ -0,0 +1,130 @@
...
+/**
+ * Selector used for implementations reliant on `<ion-content>` for scroll event changes.
+ *
+ * Developers should use the `.ion-content-scroll-host` selector to target the element emitting
+ * scroll events. With virtual scroll implementations this will be the host element for
+ * the scroll viewport.
+ */
+const ION_CONTENT_SELECTOR = `${ION_CONTENT_ELEMENT_SELECTOR}, ${ION_CONTENT_CLASS_SELECTOR}`;
+
+const hasOverflowScroll = (node: Element | null) => {
```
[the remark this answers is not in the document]
sean-perkins: I would feel much more comfortable type casting from: 
```ts
const scrollContainer = getScrollContainer(el as HTMLElement);
```

and removing the type casting elsewhere. Ideally, functions should have narrow types of the explicit things they are dependent on. Otherwise they come highly coupled/harder to use over time. 

Datetime is an example of this, with the `DatetimeParts` type being required on all the utils, even if the util only needs a month to execute.

core/src/utils/content/index.ts
```diff
@@ -0,0 +1,130 @@
...
+  if (node === null) {
+    return false;
+  }
+  const isElement = node instanceof HTMLElement;
+  const overflowY = isElement && getComputedStyle(node).overflowY;
+  // Element is scrollable if there is overflow or if the overlay is explicit
+  const isScrollable = overflowY !== 'visible' && overflowY !== 'hidden';
+  // The element is the node if it's scrollable or the scroll height is larger than the client height
```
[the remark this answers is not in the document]
sean-perkins: The element being checked by `hasOverflowScroll`, is a match/the element with overflow scroll if it is explicitly scrollable or if the scroll height is larger than the client height.

core/src/utils/content/index.ts
```diff
@@ -0,0 +1,130 @@
...
+ * For `ion-content` the scroll target will be the result
+ * of the `getScrollElement` function.
+ *
+ * For custom implementations it will be the element host
+ * or a selector within the host, if supplied through `scrollTarget`.
+ */
+export const getScrollElement = async (el: Element) => {
+  await new Promise(resolve => componentOnReady(el, resolve));
+
+  if (el?.tagName.toLowerCase() === ION_CONTENT_ELEMENT_SELECTOR) {
```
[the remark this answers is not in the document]
sean-perkins: This check is very explicit to calling custom methods exposed on `ion-content`. I do not think we would want to open that up to other selectors, otherwise it will throw exceptions.

core/src/utils/content/index.ts
sean-perkins: Very good catch! Updated the comment and the fallback implementation for the `HTMLElement`, to use scroll by point behavior.

core/src/utils/content/index.ts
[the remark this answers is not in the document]
amandaesmith3: Works for me 👍

core/src/utils/content/index.ts
```diff
@@ -0,0 +1,130 @@
...
+ * returning the inner scroll element.
+ *
+ * For `ion-content` the scroll target will be the result
+ * of the `getScrollElement` function.
+ *
+ * For custom implementations it will be the element host
+ * or a selector within the host, if supplied through `scrollTarget`.
+ */
+export const getScrollElement = async (el: Element) => {
+  await new Promise(resolve => componentOnReady(el, resolve));
```
liamdebeasi, line 53: We only need to do this if the element is `ion-content`. We should move this inside of the `if` block below.

core/src/utils/content/index.ts
```diff
@@ -0,0 +1,130 @@
...
+  // Element is scrollable if there is overflow or if the overlay is explicit
+  const isScrollable = overflowY !== 'visible' && overflowY !== 'hidden';
+  // The element is the node if it's scrollable or the scroll height is larger than the client height
+  if (isScrollable && node.scrollHeight >= node.clientHeight) {
+    return true;
+  }
+  return false;
+}
+
+const getScrollContainer = (node: Element) => {
```
liamdebeasi, line 30: I don't get why we need to do all of this work. Since users are instructed to place `.ion-content-scroll-host` on their scrollable element, isn't querying for that class enough?

liamdebeasi: I also have concerns over the usage of `getComputedStyle` in `hasOverflowScroll` as that will compute all of the styles for the element again.

core/src/utils/content/index.ts
```diff
@@ -0,0 +1,130 @@
...
+export const findClosestIonContent = (el: Element) => {
+  return el.closest<HTMLElement>(ION_CONTENT_SELECTOR);
+}
+
+/**
+ * Scrolls to the top of the element. If an `ion-content` is found, it will scroll
+ * using the public API `scrollToTop` with a duration.
+ */
+export const scrollToTop = (el: HTMLElement, durationMs: number): Promise<any> => {
+  if (el.tagName.toLowerCase() === ION_CONTENT_ELEMENT_SELECTOR) {
```
liamdebeasi, line 97: We do this check in multiple places. Can we make this a utility function instead?

core/src/utils/content/index.ts
```diff
@@ -0,0 +1,130 @@
...
+/**
+ * Scrolls by a specified X/Y distance in the component. If an `ion-content` is found, it will scroll
+ * using the public API `scrollByPoint` with a duration.
+ */
+export const scrollByPoint = (el: HTMLElement, x: number, y: number, durationMs: number) => {
+  if (el.tagName.toLowerCase() === ION_CONTENT_ELEMENT_SELECTOR) {
+    const content = el as HTMLIonContentElement;
+    return content.scrollByPoint(x, y, durationMs);
+  }
+  return Promise.resolve(el.scrollTo({
```
liamdebeasi, line 117: Can we use [Element.scrollBy](https://developer.mozilla.org/en-US/docs/Web/API/Element/scrollBy) instead?

liamdebeasi (changes requested):

core/src/utils/content/index.ts
sean-perkins: Agreed, I think we can cut a lot of complexity here by removing this util and requiring the implementation to correctly place the class on the scrollable element. This was to help patch "detaching ion-content", but after our team discussion, we can remove those pieces.

core/src/utils/content/index.ts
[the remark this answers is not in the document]
sean-perkins: Ended up removing all this extra querying for scroll containers. Wasn't needed after our discussion around detaching ion-content 🎉

sean-perkins:
@tlaverdure I wanted to update you on the status of this PR, based on decisions the team has discussed privately. 

The changes for querying that would support detaching from `ion-content`, ended up conflicting with existing Dx patterns we have implemented for `ion-refresher`, to require it to be used inside of `slot="fixed"`. Changing this behavior would be a breaking change and could also result in incorrect implementations of `ion-refresher` (specifically the iOS pull to refresh animation). 

As such, I have removed the referenced issue for the webkit performance bug. With reviewing that issue, there are a few items that stand out to me:
1. The performance overhead is an issue that Webkit will have to address with ShadowDOM. 
2. The reproduction use case does not use virtual scroll, a feature that could not fully be taken advantage of prior to this active PR. 

I would be happy to test virtual scroll against your reproduction to see if there are noticeable performance improvements that you can taken advantage of in the interim. 

I apologize for committing to handling your use case and the end result not addressing it. We will continue to evaluate improvements we can make that would incrementally improve the performance without causing a breaking change (or hold a larger improvement for v7). 

Thanks for your willingness to test this PR and your continued time in improving Ionic!

amandaesmith3 (approved):

core/src/utils/content/content.utils.spec.ts
```diff
@@ -0,0 +1,189 @@
...
+  });
+
+  it('printIonContentErrorMsg should display "<my-el> must be used inside ion-content or .ion-content-scroll-host."', () => {
+    const consoleErrorMock = jest.spyOn(console, 'error').mockImplementation();
+
+    printIonContentErrorMsg(<any>{
+      tagName: 'MY-EL'
+    });
+
+    expect(consoleErrorMock).toHaveBeenCalledWith('<my-el> must be used inside ion-content or .ion-content-scroll-host.');
```
liamdebeasi, line 184: This is confusing. If I use `<my-el>` inside of `.ion-content-scroll-host`, I would expect the component to work as intended. The reality is that the component needs to be inside of `ion-content`. Whether or not you use `.ion-content-scroll-host` depends on if you are using virtual scroll.

core/src/utils/content/index.ts
```diff
@@ -0,0 +1,98 @@
...
+/**
+ * Selector used for implementations reliant on `<ion-content>` for scroll event changes.
+ *
+ * Developers should use the `.ion-content-scroll-host` selector to target the element emitting
+ * scroll events. With virtual scroll implementations this will be the host element for
+ * the scroll viewport.
+ */
+const ION_CONTENT_SELECTOR = `${ION_CONTENT_ELEMENT_SELECTOR}, ${ION_CONTENT_CLASS_SELECTOR}`;
+
+const isIonContent = (el: Element) => el && el.tagName.toLowerCase() === ION_CONTENT_ELEMENT_SELECTOR;
```
liamdebeasi, line 15: Minor thing, but can we create an `ION_CONTENT_TAG_NAME` const so we do not have to do `toLowerCase` every time?

```js
const ION_CONTENT_TAG_NAME = 'ION-CONTENT';
el && el.tagName === ION_CONTENT_TAG_NAME;
```

liamdebeasi, line 15: Minor thing, but can we create an `ION_CONTENT_TAG_NAME` const so we do not have to do `toLowerCase` every time?

```js
const ION_CONTENT_TAG_NAME = 'ION-CONTENT';

...

el && el.tagName === ION_CONTENT_TAG_NAME
```

core/src/components/header/test/scroll-target/e2e.ts
```diff
@@ -0,0 +1,48 @@
...
+      const collapseHeader = await page.find('ion-header[collapse="fade"]');
+
+      expect(largeHeader.className).not.toContain('header-collapse-condense-inactive');
+      expect(collapseHeader.className).toContain('header-collapse-condense-inactive');
+    });
+
+    describe('when the scroll container has overflow', () => {
+
+      it('should display the collapsed title on scroll', async () => {
+        const largeHeader = await page.find('ion-header[collapse="condense"]');
```
liamdebeasi, line 34: Can we take a screenshot once the large title has collapsed?

liamdebeasi, line 34: Can we take screenshots?

core/src/components/header/test/scroll-target/e2e.ts
```diff
@@ -0,0 +1,48 @@
...
+
+  it('should match existing visual screenshots', async () => {
+    const compare = await page.compareScreenshot();
+    expect(compare).toMatchScreenshot();
+  });
+
+  describe('large title', () => {
+
+    it('should display the large title initially', async () => {
+      const largeHeader = await page.find('ion-header[collapse="condense"]');
```
liamdebeasi, line 24: Can we take screenshots?

liamdebeasi (changes requested):
Please be sure to also update the Virtual Scroll docs:

Vue: https://github.com/ionic-team/ionic-docs/blob/main/docs/vue/virtual-scroll.md#a-note-on-ionic-components
Angular: https://github.com/ionic-team/ionic-docs/blob/main/docs/angular/virtual-scroll.md#a-note-on-ionic-components
React: https://github.com/ionic-team/ionic-docs/blob/main/docs/react/virtual-scroll.md#a-note-on-ionic-components

(In particular, the "A Note on Ionic Components" section)

liamdebeasi (approved):
Good to merge once the other 4 points have been addressed. Great job on this 👍

liamdebeasi:
One more thing: Please make sure you update the commit subject to be more descriptive

sean-perkins:
PR to update the docs for virtual scrolling will be tracked here: https://github.com/ionic-team/ionic-docs/pull/2229

mariusbolik:
Hi @sean-perkins, thank you for the good work! Is there a dev build or any other way to test this feature?

Issue #24209, opened 2021-11-11
feat: datetime disable specific dates


himanshusukhpal:
### Prerequisites

- [X] I have read the [Contributing Guidelines](https://github.com/ionic-team/ionic-framework/blob/main/.github/CONTRIBUTING.md#creating-an-issue).
- [X] I agree to follow the [Code of Conduct](https://ionicframework.com/code-of-conduct).
- [X] I have searched for [existing issues](https://github.com/ionic-team/ionic-framework/issues) that already include this feature request, without success.

### Describe the Feature Request

Ion-Datetime calendar I believe lacks a function to disable holidays/specific dates with a label

### Describe the Use Case

It would be great to have a way to mark specific dates as disabled not for every month

### Describe Preferred Solution

Ion2-calendar can be looked into and a similar daysConfig option can be added which takes an array of objects that have the details to edit a specific date, such as
daysConfig = [
{
  date: new Date(2021-10-05),
  disabled: true,
  cssClass: 'subtitle',
  subTitle: 'Govardhan Pooja'
}
]

![image](https://user-images.githubusercontent.com/10000755/141268498-7a99367f-17ea-4118-8733-c53105aa4801.png)

sean-perkins:
Hello @himanshusukhpal, I am excited to announce that this feature will be available in the upcoming 6.1.0 release :tada:.

PR #24898, opened 2022-03-08
feat(datetime): isDateEnabled to enable/disable specific days 

sean-perkins:
## Pull request checklist

## Pull request type

 

 

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes

Developers do not have the ability to disable specific days, ranges of dates, holidays, weekends, etc.

Issue URL: #24209

## What is the new behavior?

Developers now have full access to each rendered calendar day, allowing them to write custom logic to enable or disable dates. Developers can use this property to disable a specific date, date range, weekends, etc.

- Introduces `isDateEnabled` to `ion-datetime`
- `isDateEnabled` will default to enabling all dates (min/max rules will overrule this behavior)
- `isDateEnabled` accepts an ISO 8601 date string for each rendered calendar day

- [ ] Yes
- [x] No


core/src/components/datetime/datetime.tsx
```diff
@@ -153,6 +153,16 @@ export class Datetime implements ComponentInterface {
    */
   @Prop() readonly = false;
 
+  /**
+   * Returns if an individual date (day) is enabled or disabled.
+   *
+   * Returns the enabled state of a day given an ISO 8601 date string.
+   * By default, all days are enabled. Developers can use this method
+   * to write custom disabled date logic. When working with the JS date
+   * object, it is recommended to use the UTC functions (i.e. `getUTCDay()`).
```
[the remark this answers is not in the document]
sean-perkins: Ended up removing this bit from the documentation here, but if the developer expects October 10th to be disabled, using `.getDay()` with the `Date` object, will disable based on local time instead of universal time. Using `getUTCDay()` will allow them to enforce the day is disabled for all users, regardless of their local time.

core/src/components/datetime/readme.md
```diff
@@ -170,6 +170,60 @@ console.log(formattedString); // Jun 4, 2021
 
 See https://date-fns.org/docs/format for a list of all the valid format tokens.
 
+## Disabling Dates
```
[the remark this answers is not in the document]
sean-perkins: Added per-framework examples. Let me know if it would make sense to rip out the code snippets at this level. The goal was to document a few common use cases. Possibly those should be handled per each framework though. Playground would be ideal here.

core/src/components/datetime/readme.md
```diff
@@ -170,6 +170,60 @@ console.log(formattedString); // Jun 4, 2021
...
+<script>
+  const datetime = document.querySelector('ion-datetime');
+  datetime.isDateEnabled = function(dateIsoString) {
+    // Write your custom logic here
+    return true;
+  }
+</script>
+```
+
+### Disabling a specific date
```
amandaesmith3, line 191: The three below examples are backwards; they'll disable every day *but* the ones listed in the comments.

core/src/components/datetime/test/disable-dates/e2e.ts
```diff
@@ -0,0 +1,166 @@
...
+    describe('when isDateEnabled returns true', () => {
+
+      it('calendar days should be enabled', async () => {
+        await page.$eval('ion-datetime', (el: any) => {
+          el.isDateEnabled = () => true;
+        });
+
+        await page.waitForChanges();
+
+        const disabledDays = await page.findAll('ion-datetime >>> .calendar-day[disabled]:not(.calendar-day-padding)');
```
amandaesmith3, line 26: Could we break this logic out into a util function that returns an array of disabled days (and then check the length when just the number is needed)? That sort of pattern is used a bunch here.

core/src/components/datetime/test/disable-dates/index.html
```diff
@@ -0,0 +1,114 @@
...
+    }
+
+    ion-datetime {
+      box-shadow: 0px 16px 32px rgba(0, 0, 0, 0.25), 0px 8px 16px rgba(0, 0, 0, 0.25);
+      border-radius: 8px;
+    }
+  </style>
+</head>
+
+<body>
```
amandaesmith3, line 36: Can we add a test for a datetime with both a custom `isDateEnabled` and a conflicting `min` and/or` max`, to make sure those take priority?

amandaesmith3 (changes requested):

core/src/components/datetime/readme.md
```diff
@@ -170,6 +170,68 @@ console.log(formattedString); // Jun 4, 2021
...
+### Disabling weekends
+
+```js
+isDateEnabled(dateIsoString) {
+  const date = new Date(dateIsoString);
+  if (date.getUTCDay() === 0 && date.getUTCDay() === 6) {
+    // Disables Saturday and Sunday
+    return true;
+  }
+  return false;
```
amandaesmith3, line 215:
```suggestion
    return false;
  }
  return true;
```

amandaesmith3 (changes requested):
Couple more small tweaks and then this LGTM 👍 Great work on this! I love how elegant the actual component change was.

amandaesmith3 (approved):

core/src/components/datetime/usage/vue.md
```diff
@@ -133,10 +136,20 @@
         return format(parseISO(value), 'MMM dd yyyy');
       };
 
+      const isDateEnabled = (dateIsoString: string) => {
+        const date = new Date(dateIsoString);
+        if (date.getUTCDay() === 1 && date.getUTCMonth() === 0 && date.getUTCFullYear() === 2022) {
```
liamdebeasi, line 141: getUTCDay gives the day of the week. I think you want getUTCDate which gives the calendar day of the month.

liamdebeasi: Non-blocking, but we typically push users to use date-fns for any date parsing. It would be nice to show some date-fns usage here as well.

core/src/components/datetime/usage/vue.md
```diff
@@ -51,6 +51,9 @@
       <ion-button @click="reset()">Reset</ion-button>
     </ion-buttons>
   </ion-datetime>
+
+  <!-- Disable custom days -->
+  <ion-datetime :isDateEnabled="isDateEnabled"></ion-datetime>
```
liamdebeasi, line 56:
```suggestion
  <ion-datetime :is-date-enabled="isDateEnabled"></ion-datetime>
```

core/src/components/datetime/readme.md
```diff
@@ -170,6 +170,68 @@ console.log(formattedString); // Jun 4, 2021
 
 See https://date-fns.org/docs/format for a list of all the valid format tokens.
 
+## Disabling Dates
+
+With the `isDateEnabled` property, developers can customize the `ion-datetime` to disable a specific day, range of dates, weekends or any custom rule using an ISO 8601 date string. 
+
+The `isDateEnabled` property accepts a function returning a boolean, indicating if a date is enabled. The function is called for each rendered calendar day, for the previous, current and next month. Custom implementations should be optimized for performance. This function is called often, so any extra logic should be avoided to reduce and prevent jank.
```
liamdebeasi, line 177:
```suggestion
The `isDateEnabled` property accepts a function returning a boolean, indicating if a date is enabled. The function is called for each rendered calendar day, for the previous, current and next month. Custom implementations should be optimized for performance to avoid jank.
```

We already said how often the function is called, so the part I removed was redundant.

core/src/components/datetime/datetime.tsx
```diff
@@ -1278,6 +1294,21 @@ export class Datetime implements ComponentInterface {
             const referenceParts = { month, day, year };
             const { isActive, isToday, ariaLabel, ariaSelected, disabled } = getCalendarDayState(this.locale, referenceParts, this.activePartsClone, this.todayParts, this.minParts, this.maxParts, this.parsedDayValues);
 
+            let isCalDayDisabled = isCalMonthDisabled || disabled;
+
+            if (!isCalDayDisabled && this.isDateEnabled !== undefined) {
```
liamdebeasi, line 1299: Non-blocking, but doing `const { isDateEnabled } = this` lets you avoid a bit of overhead on line 1306 when accessing `isDateEnabled` again

core/src/components/datetime/datetime.tsx
```diff
@@ -1278,6 +1294,21 @@ export class Datetime implements ComponentInterface {
...
+              try {
+                /**
+                 * The `isDateEnabled` implementation is try-catch wrapped
+                 * to prevent exceptions in the user's function from
+                 * interrupting the calendar rendering.
+                 */
+                isCalDayDisabled = !this.isDateEnabled(convertDataToISO(referenceParts));
+              } catch (e) {
+                console.error('Exception thrown in isDateEnabled function, ignoring', e);
```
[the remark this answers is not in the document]
liamdebeasi: Non-blocking, but we may want to consider adding an error logging util like I did here for warnings: https://github.com/ionic-team/ionic-framework/blob/FW-286/core/src/utils/logging/index.ts#L8

core/src/components/datetime/datetime.tsx
```diff
@@ -153,6 +153,22 @@ export class Datetime implements ComponentInterface {
...
+   * If `true`, the day will be enabled/interactive.
+   * If `false`, the day will be disabled/non-interactive.
+   *
+   * The function accepts an ISO 8601 date string of a given day.
+   * By default, all days are enabled. Developers can use this function
+   * to write custom logic to disable certain days.
+   *
+   * Custom implementations should be optimized for performance. This
+   * function is called often, so any extra logic should be avoided
+   * to reduce and prevent jank.
```
liamdebeasi, line 168:
```suggestion
   * The function is called for each rendered calendar day, for the previous, current and next month.
   * Custom implementations should be optimized for performance to avoid jank.
```

liamdebeasi: Non-blocking, but this matches the docs below

liamdebeasi (commented):
One behavior I don't think we accounted for is how to skip over dates when using the keyboard.

If you focus day 11 on the first datetime in: http://localhost:3333/src/components/datetime/test/disable-dates, it's not possible to use the arrow keys to navigate backwards anymore to reach day 9.

When using the arrow keys, should it skip over the date entirely? Maybe it could highlight it but indicate that it is disabled? We may need to do some additional research here.

sean-perkins:
@liamdebeasi I don't know if we should be doing anything custom with the keyboard navigation here. This aligns with the browser behavior of sibling disabled buttons.

https://codepen.io/sean-perkins/pen/XWVryXw

Users should be able to tab-over disabled buttons, but not use left/right keyboard navigation to access them. With a screen reader/voice over, they would be able to focus that node, hear that it is disabled and focus the next/previous node.

Let me know if you feel different or if we have a native example that operates differently.

PR #24903, opened 2022-03-08
test(refresher): pull to refresh test

sean-perkins:
## Pull request checklist

## Pull request type

 

 

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes
- [x] Other (please describe): Tests

There are no tests for validating that pull-to-refresh completes successfully (after a successful pull and release).

## What is the new behavior?

Added missing tests to validate pull to refresh fires and completes. 

- [ ] Yes
- [x] No

### DIFF
--- a/core/src/components/refresher/test/basic/e2e.ts
+++ b/core/src/components/refresher/test/basic/e2e.ts
@@ -28,4 +28,26 @@ describe('refresher: basic', () => {
     expect(items.length).toBe(60);
   });
 
+  describe('legacy refresher', () => {
+
+    it('should load more items when performing a pull-to-refresh', async () => {
+      const refresherContent = await page.$('ion-refresher-content');
+      refresherContent.evaluate((el: any) => {
+        // Resets the pullingIcon to enable the legacy refresher
+        el.pullingIcon = undefined;
+      });
+
+      await page.waitForChanges();
+
+      const initialItems = await page.findAll('ion-item');
+      expect(initialItems.length).toBe(30);
+
+      await pullToRefresh(page);
+
+      const items = await page.findAll('ion-item');
+      expect(items.length).toBe(60);
+    });
+
+  });
+
 });
--- a/core/src/components/refresher/test/basic/e2e.ts
+++ b/core/src/components/refresher/test/basic/e2e.ts
@@ -1,10 +1,28 @@
+import type { E2EPage } from '@stencil/core/testing';
 import { newE2EPage } from '@stencil/core/testing';
 
-test('refresher: basic', async () => {
-  const page = await newE2EPage({
-    url: '/src/components/refresher/test/basic?ionic:_testing=true'
+import { pullToRefresh } from '../test.utils';
+
+describe('refresher: basic', () => {
+
+  let page: E2EPage;
+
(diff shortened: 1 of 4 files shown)


core/src/components/refresher/test/basic/e2e.ts
```diff
@@ -1,10 +1,28 @@
...
+  it('should match existing visual screenshots', async () => {
+    const compare = await page.compareScreenshot();
+    expect(compare).toMatchScreenshot();
+  });
+
+  it('should load more items when performing a pull-to-refresh', async () => {
+    await pullToRefresh(page);
+
+    const items = await page.findAll('ion-item');
+    expect(items.length).toBe(60);
```
liamdebeasi, line 25: Can we check the number of items before the PTR to verify that it does add new items?

core/src/components/refresher/test/basic/e2e.ts
```diff
@@ -1,10 +1,28 @@
...
+    });
+  });
+
+  it('should match existing visual screenshots', async () => {
+    const compare = await page.compareScreenshot();
+    expect(compare).toMatchScreenshot();
+  });
+
+  it('should load more items when performing a pull-to-refresh', async () => {
+    await pullToRefresh(page);
```
liamdebeasi, line 22: This is a bit tricky because there are 3 different PTR types: legacy, iOS native, and MD native. Which one is this designed to test?

sean-perkins: This should test iOS and MD, unless they are somehow configured through a property instead of the `mode`? How do you render the refresher in the legacy type?

This test emulates a mouse click/drag down from the center of the screen, which should fire the refresher events for each mode the same.

liamdebeasi: Customizing the pullingIcon should trigger legacy refreshers. The iOS native refresher only works on a device with rubber band scrolling, so we probably do not have to worry about it.

As long as we have one test where we customize the pulling icon and one where we do not, that should be fine.

core/src/components/refresher/test/basic/e2e.ts
```diff
@@ -1,10 +1,53 @@
...
+    const items = await page.findAll('ion-item');
+    expect(items.length).toBe(60);
+  });
+
+  describe('legacy refresher', () => {
+
+    it('should load more items when performing a pull-to-refresh', async () => {
+      const refresherContent = await page.$('ion-refresher-content');
+      refresherContent.evaluate((el: any) => {
+        // Resets the pullingIcon to enable the legacy refresher
```
liamdebeasi, line 36: I think you mean native refresher here? When `pullingIcon=undefined` we use the native refresher where possible. When pullingIcon is defined we fallback to legacy.

sean-perkins: Good catch! In the future, I'd like to consider our `basic` test examples as what the drop-in example looks like. So in this case, we would have native refreshers be the base example and then write a test to verify modifications/alt-states.

amandaesmith3 (approved):

liamdebeasi (approved):

Issue #22297, opened 2020-10-12
feat: canLeave event for overlays


EinfachHans:
# Feature Request

**Ionic version:**

[x] **5.x**

**Describe the Feature Request**

I want to add a CanLeave() Event on Overlays (mainly iOS card Style modal, but also for upcoming drawer) which should be somehow triggered while swiping. See attached Video of how native Web.de Mail-App handles this:

<img src="https://user-images.githubusercontent.com/20899975/95760383-40246500-0cab-11eb-9884-66d72f8e0861.gif" width="300">

If nothing was typed in, i can just swipe to close the modal. If i did, the swipe doesn't go to the end and shows a Action Sheet after i released.

**Describe Preferred Solution**

I think this should automatically work if a CanDismiss Callback is set to the Modal Element. This should be able to parse a boolean, or better a Promise<boolean> or even better both.

BerkeAras:
@EinfachHans Good idea, but isn't it possible to something like that with ``ionModalWillDismiss``

EinfachHans:
@BerkeAras it doesn't change the animation and i think i'm not able to "cancel" the dismissing here

BerkeAras:
@EinfachHans event.preventDefault does not work?

EinfachHans:
You don't have a event here and even if you have this won't be async and more a workaround then a solution @BerkeAras

EinfachHans:
@liamdebeasi i can try my best for creating a PR if you give me a few hints:
- Should this automatically happen if a canDismiss is set?
- What's the preferred way to set this method for the modal? In the best case it needs to be handle from within the modal and outside where you created it.

roman-rr:
@EinfachHans 
> You don't have a event here and even if you have this won't be async and more a workaround then a solution @BerkeAras 

Ok I see. Slower drag if dismiss not allowed. And `preventDismiss(true/false)`
Please make new feature request.

liamdebeasi:
Merged https://github.com/ionic-team/ionic-framework/issues/22626 into this issue. I think this is an interesting feature and can see this being useful for modals in general, not just card-style modals.

EinfachHans:
@liamdebeasi this PR is for normal Modals as well 😃

liamdebeasi:
Thanks for the issue. This has been resolved via https://github.com/ionic-team/ionic-framework/pull/24928, and a fix will be available in an upcoming release of Ionic Framework.

PR #24921, opened 2022-03-10
feat(modal): add canDismiss support for sheet modals

liamdebeasi:
## Pull request checklist

## Pull request type

 

 

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes

Part of https://github.com/ionic-team/ionic-framework/issues/22297

## What is the new behavior?

- This adds the base implementation for sheet modals.

- [ ] Yes
- [x] No


sean-perkins (approved):

sean-perkins:
Will need to run lint.fix before merging 👍

core/src/components/modal/gestures/sheet.ts
```diff
@@ -143,8 +162,35 @@ export const createSheetGesture = (
      * compute where the offset of the animation should be
      * relative to where the user dragged.
      */
-    const initialStep = 1 - currentBreakpoint;
-    offset = clamp(0.0001, initialStep + (detail.deltaY / height), 0.9999);
+    const initialBreakpoint = 1 - currentBreakpoint;
+    const secondToLastBreakpoint = 1 - breakpoints[1];
+    const step = initialBreakpoint + (detail.deltaY / height);
+    const isAttempingDismissWithCanDismiss = step >= secondToLastBreakpoint && canDismissBlocksGesture;
```
amandaesmith3, line 168: ~ I tried using a sheet modal with only one breakpoint (0.5) to see if that breaks this due to accessing `breakpoints[1]` directly, but it seemed to work without any issues? I'm not really sure why, but I figured I'd mention that in case you have an "oh shoot" moment, haha.

amandaesmith3 (approved):
I had one thought, but it's not blocking 👍

core/src/components/modal/gestures/sheet.ts
liamdebeasi: Good catch, I should probably add a safe guard for it. 

I think why it doesn't break is:

`const secondToLastBreakpoint = 1 - breakpoints[1]` which equals `NaN`

Then doing `0.5 >= NaN` will return `false`, so `isAttempingDismissWithCanDismiss` is never `true`.

PR #24932, opened 2022-03-12
fix(angular): item styling when control has value

sean-perkins:
## Pull request checklist

## Pull request type

 

 

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

When a control, inside an `ion-item` is displayed within a modal; after opening, closing and re-opening the modal, the `item-has-value` class is not applied to the `ion-item` that wraps the control.

Issue URL: #23809

## What is the new behavior?

- `item-has-value` is manually applied to the `ion-item` from the control value accessor, when applying other Ionic classes

- [ ] Yes
- [x] No

## Other information

This PR does misc clean-up to deprecated `substr` and usage for add/remove classes. 

Closing #24401 in favor of this PR.


amandaesmith3 (approved):

angular/src/directives/control-value-accessors/value-accessor.ts
```diff
@@ -110,13 +110,17 @@ export class ValueAccessor implements ControlValueAccessor, AfterViewInit, OnDes
 
 export const setIonicClasses = (element: ElementRef): void => {
   raf(() => {
-    const input = element.nativeElement as HTMLElement;
+    const input = element.nativeElement as HTMLInputElement;
+    const hasValue = typeof input.value !== 'undefined' && input.value.toString().length > 0;
```
liamdebeasi, line 114: We need a null check here as the `value` prop is nullable: https://github.com/ionic-team/ionic-framework/blob/main/core/src/components/input/input.tsx#L194

liamdebeasi: We should probably also write tests for when the value is `undefined` and `null` so this does not regress

liamdebeasi (changes requested):

angular/src/directives/control-value-accessors/value-accessor.ts
sean-perkins: The internals of `ion-input` don't check for `null`: https://github.com/ionic-team/ionic-framework/blob/main/core/src/components/input/input.tsx#L327-L330

While you can assign `null` as a value, the evaluated value will always be a string (`''` for undefined/null). This aligns with how the HTML input element functions.

Do we still need to cover those cases?

liamdebeasi: From what it looks like `getValue` is only used when parsing the value internally:
Usage 1: https://github.com/ionic-team/ionic-framework/blob/main/core/src/components/input/input.tsx#L236
Usage 2: https://github.com/ionic-team/ionic-framework/blob/main/core/src/components/input/input.tsx#L426
Usage 3: https://github.com/ionic-team/ionic-framework/blob/main/core/src/components/input/input.tsx#L431

None of these usages update the `value` property to be the empty string, and your code is checking the `value` prop, not calling `getValue`.

I think it's worth testing just to be sure. If we convert `null` to the empty string elsewhere such that this is not a concern, then we can proceed without the `null` check. At the moment it is not clear if that is the case.

liamdebeasi (approved):

liamdebeasi:
Prior to merging, please provide the developers on https://github.com/ionic-team/ionic-framework/issues/23809 with a dev build. I'd prefer to have some developer feedback before we ship this.

liamdebeasi:
@sean-perkins We can probably merge this now. We've given developers time to test and this new approach seems safer than the previous approach.

Issue #24967, opened 2022-03-19
bug: Setting the value of `ion-datetime` with presentation time, results in multiple ionChange emissions and incorrect display


CSHaze:
### Prerequisites

- [X] I have read the [Contributing Guidelines](https://github.com/ionic-team/ionic-framework/blob/main/.github/CONTRIBUTING.md#creating-an-issue).
- [X] I agree to follow the [Code of Conduct](https://ionicframework.com/code-of-conduct).
- [X] I have searched for [existing issues](https://github.com/ionic-team/ionic-framework/issues) that already report this problem, without success.

### Ionic Framework Version

- [ ] v4.x
- [ ] v5.x
- [X] v6.x

### Current Behavior

When patching a form value (connected to ion-dateTime) using an asynchronous action, the hours part of the control changes incorrectly.

### Expected Behavior

I would expect the patchValue to change the time control to show the new, correct hour value.

### Steps to Reproduce

Add an ion-DateTime control to the html using a "time" presentation.

Create a dynamic form in the component using a formBuilder.

Connect the form to ion-DateTime control.

Use an asynchronous action to trigger a patchValue to the form Value tied to the ion-DateTime control.

Observe that the  minutes change to the correct value and the hours do not.

### Code Reproduction URL

https://stackblitz.com/edit/ionic6-angular13-pegslx?file=src/app/app.component.ts

### Ionic Info

Ionic:

   Ionic CLI                     : 6.12.4 (...\npm\node_modules\@ionic\cli)   
   Ionic Framework               : @ionic/angular 6.0.11
   @angular-devkit/build-angular : 13.2.5
   @angular-devkit/schematics    : 13.2.5
   @angular/cli                  : 13.2.5
   @ionic/angular-toolkit        : 6.1.0

Capacitor:

   Capacitor CLI   : 3.4.1
   @capacitor/core : 3.4.1

Utility:

   cordova-res : not installed
   native-run  : 1.5.0

System:

   NodeJS : v16.13.0 (C:\Program Files\nodejs\node.exe)
   npm    : 8.4.0
   OS     : Windows 10

sean-perkins:
Thanks for this issue!

After quickly diagnosing, when dynamically changing the value, the `ionChange` is emitting 4 times:

1. `06:02:40`
2. `2022-03-20T06:57:00-04:00`
3. `2022-03-20T00:02:00-04:00`
4. `00:02`

[...]

PR #24968, opened 2022-03-20
fix(datetime): presentation time emits ionChange once 

sean-perkins:
## Pull request checklist

## Pull request type

 

 

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

When modifying the value of an `ion-datetime[presentation="time"]` with the picker wheel visible in the viewport; the datetime component will emit 4 `ionChange` events, resulting in an incorrect value and in Angular, an incorrect display of the picker wheel.

For example, when setting the value to `06:02:40`, the emitted events will be:
```
{ value: '06:02:40' }
{ value: '2022-03-20T06:57:00-04:00' }
{ value: '2022-03-20T00:02:00-04:00' }
{ value: '00:02' }
```

This corresponds to:
- The first emission is the `ion-datetime` watch callback for the value being modified/ionChange.
- The second emission is the `ion-picker-column-internal` hour column detecting a change/modifying the value.
- The third emission is the `ion-picker-column-internal` minute column detecting a change/modifying the value.
- The fourth emission is the `ion-picker-column-internal` am/pm column detecting a change/modifying the value.

Issue URL: #24967

## What is the new behavior?

`ion-datetime[presentation="time"]` will only emit a single `ionChange` event when the value is modified, regardless if the picker wheel is visible in the viewport or not. 

This results in the value correctly reflecting the new datetime value and the display of the component to be correct.

The `ion-picker-column-internal` will not emit `ionChange` events when the value is modified externally. It will only emit change events as a result of selecting a new value or the fallback setting the initial value. 

- [ ] Yes
- [x] No

### DIFF
--- a/core/src/components/datetime/test/presentation/e2e.ts
+++ b/core/src/components/datetime/test/presentation/e2e.ts
@@ -1,4 +1,4 @@
-import { newE2EPage } from '@stencil/core/testing';
+import { newE2EPage, E2EPage } from '@stencil/core/testing';
 
 test('presentation', async () => {
   const page = await newE2EPage({
@@ -12,4 +12,38 @@ test('presentation', async () => {
   for (const screenshotCompare of screenshotCompares) {
     expect(screenshotCompare).toMatchScreenshot();
   }
+});
+
+describe('presentation: time', () => {
+
+  let page: E2EPage;
+
+  beforeEach(async () => {
+    page = await newE2EPage({
+      url: '/src/components/datetime/test/presentation?ionic:_testing=true'
+    });
+  });
+
+  describe('when the time picker is visible in the view', () => {
+
+    it('manually setting the value should emit ionChange once', async () => {
+      const datetime = await page.find('ion-datetime[presentation="time"]');
+      const didChange = await datetime.spyOnEvent('ionChange');
+
+      await page.$eval('ion-datetime[presentation="time"]', (el: any) => {
+        el.scrollIntoView();
+      });
+
+      await page.$eval('ion-datetime[presentation="time"]', (el: any) => {
+        el.value = '06:02:40';
+      });
+
+      await page.waitForChanges();
+
+      expect(didChange).toHaveReceivedEventTimes(1);
+      expect(didChange).toHaveReceivedEventDetail({ value: '06:02:40' });
+    });
+
+  });
+
 });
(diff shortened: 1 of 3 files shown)


core/src/components/picker-column-internal/picker-column-internal.tsx
```diff
@@ -251,6 +254,7 @@ export class PickerColumnInternal implements ComponentInterface {
 
           if (selectedItem.value !== this.value) {
             this.value = selectedItem.value;
+            this.emitIonChange();
```
liamdebeasi, line 257: We should consider having a `setValue` method that is used internally that encapsulated setting `this.value` and emitting the change event. It sounds like any time we change `value` internally we also need to emit `ionChange`?

core/src/components/datetime/test/presentation/e2e.ts
```diff
@@ -13,3 +13,37 @@ test('presentation', async () => {
     expect(screenshotCompare).toMatchScreenshot();
   }
 });
+
+describe('presentation: time', () => {
+
+  let page: E2EPage;
+
+  beforeEach(async () => {
```
liamdebeasi, line 21: Can we also add tests to the picker column? Some ideas:

1. Check that scrolling the picker emits ionChange.
2. Check that setting `value` prop does not emit ionChange.

liamdebeasi, line 21: Can we also add tests to the picker column? Some ideas:

1. Check that scrolling the picker emits ionChange.
2. Check that setting `value` prop does not emit ionChange.

core/src/components/picker-column-internal/test/basic/e2e.ts
```diff
@@ -51,6 +51,28 @@ describe('picker-column-internal', () => {
...
+
+    it('should emit ionChange when the picker is scrolled', async () => {
+      const pickerColumn = await page.find('#default');
+      const ionChangeSpy = await pickerColumn.spyOnEvent('ionChange');
+
+      await page.$eval('#default', (el: any) => {
+        el.scrollTo(0, 300);
+      });
+
+      expect(ionChangeSpy).not.toHaveReceivedEvent();
```
liamdebeasi, line 73:
```suggestion
      expect(ionChangeSpy).toHaveReceivedEvent();
```

liamdebeasi (approved):
Good to merge once test is fixed

core/src/components/picker-column-internal/test/basic/e2e.ts
sean-perkins: Great catch, I also needed to wait for the event:
```ts
await ionChangeSpy.next();

expect(ionChangeSpy).toHaveReceivedEvent();
```

Confirmed this test passed with the changes and failed against main. I'll wait for CI before finally merging.

Issue #24192, opened 2021-11-09
bug: vue 3 vite-plugin-vue-layouts routing duplication


riderx:
### Prerequisites

- [X] I have read the [Contributing Guidelines](https://github.com/ionic-team/ionic-framework/blob/main/.github/CONTRIBUTING.md#creating-an-issue).
- [X] I agree to follow the [Code of Conduct](https://ionicframework.com/code-of-conduct).
- [X] I have searched for [existing issues](https://github.com/ionic-team/ionic-framework/issues) that already report this problem, without success.

### Ionic Framework Version

- [ ] v4.x
- [ ] v5.x
- [X] v6.x

### Current Behavior

To use tab component it's need to use `ionic/vue-router`
but i cannot use `router-view` anymore or i get duplicated route.
<img width="549" alt="Screenshot 2021-11-09 at 17 41 07" src="https://user-images.githubusercontent.com/4084527/140976286-606523c2-2c7e-4f3c-87e8-8853ad5162bb.png">
if i switch totally to `ion-router-outlet` i have css issue.
<img width="709" alt="Screenshot 2021-11-09 at 17 40 59" src="https://user-images.githubusercontent.com/4084527/140976308-50f673fd-bac8-4d1d-a9d5-db555d6c7c1e.png">

### Expected Behavior

Using ionic without ionic/vue-router should be allowed for tabs with `router-view` 
Or 
Allowing to have other route with `router-view`
Or
Ability to disabe host css for `ion-router-outlet`
<img width="323" alt="Screenshot 2021-11-09 at 16 08 30" src="https://user-images.githubusercontent.com/4084527/140976645-4c361b41-856c-4624-83fa-9840eb9cb7b4.png">

### Steps to Reproduce

clone this repo :
https://github.com/riderx/vitesse-ionic/tree/bug
do`pnpm i` 
do `pnpm dev`
go to the new server you should see the duplicated route
To see the second form of but write something in the input and click button `go`
then you will see footer component under hi page.
i found a workaround to make it work with `v-if` condition on `ion-router-outlet` as see there :
https://github.com/riderx/vitesse-ionic/blob/main/src/App.vue

### Code Reproduction URL

https://github.com/riderx/vitesse-ionic/tree/bug

### Ionic Info

Ionic:

   Ionic CLI       : 6.18.0 (/usr/local/lib/node_modules/@ionic/cli)
   Ionic Framework : @ionic/vue 6.0.0-rc.2

Utility:

[...]

PR #24971, opened 2022-03-21
Allow ion-tab to work with router-view 

riderx:
## Pull request checklist

## Pull request type

- [ ] Bugfix
- [X] Feature
- [ ] Build related changes

i don't use ion-vue-router in my project, but i cannot use ion-tab without it, instead of hacking routing it made possible to use it without. 
The only change is there are no animation any more.
To work the 

Issue Number: [24192](https://github.com/ionic-team/ionic-framework/issues/24192)

## What is the new behavior?

- Allow use of ion-tab with basic routing component

- [ ] Yes
- [X] No

### DIFF
--- a/packages/vue/src/components/IonTabs.ts
+++ b/packages/vue/src/components/IonTabs.ts
@@ -16,7 +16,7 @@ export const IonTabs = /*@__PURE__*/ defineComponent({
      * inside of ion-tabs.
      */
     if (slottedContent && slottedContent.length > 0) {
-      routerOutlet = slottedContent.find((child: VNode) => child.type && ((child.type as any).name === 'IonRouterOutlet' || child.type as any).name === 'RouterView');
+      routerOutlet = slottedContent.find((child: VNode) => child.type && ((child.type as any).name === 'IonRouterOutlet' || (child.type as any).name === 'RouterView'));
     }
 
     if (!routerOutlet) {
--- a/packages/vue/src/components/IonTabs.ts
+++ b/packages/vue/src/components/IonTabs.ts
@@ -2,6 +2,7 @@ import { h, defineComponent, VNode } from 'vue';
 
 const WILL_CHANGE = 'ionTabsWillChange';
 const DID_CHANGE = 'ionTabsDidChange';
+const ALLOWED_ROUTER_OUTLET = ['IonRouterOutlet', 'RouterView'];
 
 export const IonTabs = /*@__PURE__*/ defineComponent({
   name: 'IonTabs',
@@ -16,13 +17,16 @@ export const IonTabs = /*@__PURE__*/ defineComponent({
      * inside of ion-tabs.
      */
     if (slottedContent && slottedContent.length > 0) {
-      routerOutlet = slottedContent.find((child: VNode) => child.type && ((child.type as any).name === 'IonRouterOutlet' || (child.type as any).name === 'RouterView'));
+      routerOutlet = slottedContent.find((child: VNode) => child.type && ALLOWED_ROUTER_OUTLET.includes((child.type as any).name));
     }
+
 
     if (!routerOutlet) {
-      throw new Error('IonTabs must contain an IonRouterOutlet or RouterView. See https://ionicframework.com/docs/vue/navigation#working-with-tabs for more information.');
+      throw new Error('IonTabs must contain an IonRouterOutlet at best or RouterView. See https://ionicframework.com/docs/vue/navigation#working-with-tabs for more information.');
     }
-
+    if (!inject('navManager')) { 
+      throw new Error('Your app must use `import { createRouter, createWebHistory } from \'vue-router\'` as router. \n For best transition experience IonTabs must contain an IonRouterOutlet. See https://ionicframework.com/docs/vue/navigation#working-with-tabs for more information.')
+    }
     let childrenToRender = [
       h('div', {
         class: 'tabs-inner',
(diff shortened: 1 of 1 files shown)


riderx:
@liamdebeasi hey sorry to open this again but i was hit again by this issue while building https://github.com/Cap-go/capgo
I found a way to make it work for both worlds.
I Added a warning to prevent ionic team to get support issue, if a user doesn't have transition because they use `RouterView` instead of your outlet.
I hope this is more acceptable solution

sean-perkins:
@riderx can you share what issue exists in the Ionic Vue Router that prevents you from using it? I can let the other team members weigh in here, but we do not want to encourage developers to be able to eject from our implementations to mend other solutions in place.

We have this stance for multiple reasons:
1. We built the routing integrations that exist to offer native-like transitions for web-native applications built with Ionic. Using any other solution in place without those transitions is not something we want to allow, as it degrades from the value of an Ionic application.
2. Using unsupported routing integrations can lead to unknown bugs and gaps in features. By making it "allowed", we invite these issues to exist and cause additional overhead into the maintenance of the project.

riderx:
Hey @sean-perkins thanks for the answer.
You can find my old issue here:
https://github.com/ionic-team/ionic-framework/issues?q=is%3Aissue+author%3A%40me+is%3Aclosed

The main issue who stop me using ionic-router is sub routing.
With the vuejs ionic router we cannot use sub route in page, and that use for layout system in vuejs apps, in desktop website.

You can check the code here https://github.com/Cap-go/capgo
the sub route is used to render markdown blog page.

sean-perkins:
@riderx after discussing with the team further, the problem you are experiencing is a result of how you are autogenerating routes based on the file structure:  https://github.com/Cap-go/capgo/blob/main/src/main.ts#L6

You should not be generating sibling pages in tabs as children. Instead they should be siblings.

For example:
- `/tabs/tab1/child` and `/tabs/tab1` should be siblings. 

There should not be a `/child` route specified as a child of `/tabs/tab1`. 

This is explain in the docs: https://ionicframework.com/docs/vue/navigation#shared-urls-versus-nested-routes

I am going to close out this PR, as this is not a change we are wanting to support. Feel free to open an issue if you run into problems when following that documentation and we can help assist.

Thanks!

riderx:
Understand you feedback, but please understand mine.
i cannot use ionic in my stack because of that i did try many things since the last issue in december, and still cannot do it.
My last option is to totally stop to using ionic and switch to quasar who is really made for vue users.
As it, it's not flexible enough for people who use other packages than just ionic...

sean-perkins:
@riderx I am unfamiliar to Vitesse. Does it prevent you from adjusting the declaration pattern for routes that are registered to Ionic Vue's router?

The issue you describe:
> The main issue who stop me using ionic-router is sub routing.
With the vuejs ionic router we cannot use sub route in page, and that use for layout system in vuejs apps, in desktop website.

Is a supported feature of our router integration. You will just need to adjust the way that routes are declared to take full advantage of it. By refactoring your route definitions from being nested, to siblings, you should be able to have the navigation pattern you desire with tabs, without using `router-view` or Ionic having to open up the routing integration layer.

riderx:
@sean-perkins yes vitesse create the routing automaticly from folder architechture and handle the SSR/SSG system for you for SEO purpose.
So to add ionic i had to disable SSR/SSG who is already a big loss and i had to create my own ionTAB component to have my blog section working because it use sub routing to generate from file in markdown blog articles on my web version of my SaaS.
What i ask is just a flexible way for routing to work in combinaise with other tools. 🥲
as you can see my other issue there for SSR/SSG : https://github.com/ionic-team/ionic-framework/issues/24991#issuecomment-1082484074

riderx:
@sean-perkins i use ionic since years and would love to continue using it!
Learning quasar from 0 is not my goal!
If you look for solution to let this work for desktop, it's wonderful news!

What you propose is perfectly what i expect, allow a bit of flexibility for usage .
Thanks a lot

riderx:
Just saw today the maintainer of vitesse ask about how people do animations between components or page in vue.
Can be interesting to read answers :
https://twitter.com/antfu7/status/1509587065771798528?s=21&t=VWXjNbN9g-DVMRUzT5fA_g
The quasar answer to that, look less magical than ionic one, but also more powerful.
I do really believe of have animation outside of router can be really more flexible.
Maybe ion page csn become the host of that?
Or just ionPageAnimation component around normal router?
This is huge réflexion, share your thoughts with the community can be very helpful :)

sean-perkins:
The animation pattern demonstrated in that tweet is both a very different feature and implementation pattern than what Ionic Framework does for page transitions. Ionic maintains the previous view in the DOM, so that we can animate a swipe-to-go-back experience correctly, without screen jittering from having to render the previous view (it is already rendered, just off screen). This aligns with native app page transitions.

The tweet demonstrates cloning element nodes and then transitioning that cloned node into a new page. You can technically already do this with Ionic or any web solution. It just requires complexities of knowing the start and end position of your transitions and re-attaching event listeners, which is why it is extremely unlikely Ionic would solve that problem for you, as it is an application detail. 

You could this without cloning the node, but it wouldn't be as smooth of an experience, i.e. I did it previously with Ionic v4: https://twitter.com/seanfperkins/status/1136502546577723397.

riderx:
@sean-perkins thanks for the mind-blowing moment, now i understand a lot better!

i didn't understand you modified the Vue router to have kept the view in the DOM, i saw the init function where never call again but couldn't explain why.

This should be explained in this page:

https://ionicframework.com/docs/vue/navigation

That will solve a lot of unnecessary fight with ionic router!
For me, the router was just about transitions, that why it's feel "useless" to overload the whole router for that.
I'm sorry.

Issue #24633, opened 2022-01-24
bug: keys Home, End, ArrowUp, ArrayDown not working in (ion-)textarea in popover


appy-one:
### Prerequisites

- [x] I have read the [Contributing Guidelines](https://github.com/ionic-team/ionic-framework/blob/main/.github/CONTRIBUTING.md#creating-an-issue).
- [X] I agree to follow the [Code of Conduct](https://ionicframework.com/code-of-conduct).
- [X] I have searched for [existing issues](https://github.com/ionic-team/ionic-framework/issues) that already report this problem, without success.

### Ionic Framework Version

- [ ] v4.x
- [ ] v5.x
- [X] v6.x

### Current Behavior

Keys Home, End, ArrowUp, ArrayDown are not working when editing text in a `<textarea>` or `<ion-textarea>` when presented in an popover because `ev.preventDefault()` is called in [overlays.ts:376](https://github.com/ionic-team/ionic-framework/blob/223f36f6adacf8adce47cee4809a60c94a9e0efa/core/src/components/popover/utils.ts#L376) for 'ArrowDown', and other keys below.

### Expected Behavior

Cursor should be able to move in a textarea when pressing those keys.

### Steps to Reproduce

1. create component with a `<textarea>` or `<ion-textarea>`
2. open the component in a popover
3. focus the textarea input and enter some text with multiple lines
4. try moving the cursor with up, down, home and end keys

### Ionic Info

Ionic:

   Ionic CLI                     : 5.4.16 (C:\Users\4ewou\AppData\Roaming\nvm\v14.15.4\node_modules\ionic)
   Ionic Framework               : @ionic/angular 6.0.1
   @angular-devkit/build-angular : 13.0.4
   @angular-devkit/schematics    : 13.0.4
   @angular/cli                  : 13.0.4
   @ionic/angular-toolkit        : 5.0.3

Utility:

   cordova-res : not installed
   native-run  : not installed

System:

   NodeJS : v14.15.4 (C:\Program Files\nodejs\node.exe)
   npm    : 6.14.10
   OS     : Windows 10

appy-one:
For others with this issue, a workaround is adding a `keydown` event handler to the `<textarea>` that stops event propagation: `<textarea (keydown)="$event.stopPropagation()"></textarea>`

sean-perkins:
[...]

PR #24990, opened 2022-03-24
fix(popover): only focus trap ion-item children

sean-perkins:
## Pull request checklist

## Pull request type

 

 

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

`ion-popover` has an internal key event handler for focusing between ion-item children elements. This handler conflicts with elements that have their own keyboard interactions, such as a HTML textarea element. 

Issue URL: #24633

## What is the new behavior?

- `ion-popover` only handles keyboard interactions when the active element is an `ion-item`

- [ ] Yes
- [x] No

## Other information

Dev build: `6.0.15-dev.11649448530.1093d192`


amandaesmith3 (approved):

liamdebeasi (changes requested):
This regresses a behavior in `ion-popover`. In `main`, opening a popover and pressing the down arrow key will focus the first `ion-item`. In this branch, pressing the down arrow key does nothing. You can test in `/src/components/popover/test/nested`.

Maybe we need to change where this `ion-item` check is? (Also can we add a test for this behavior?)

sean-perkins:
@liamdebeasi good catch 👍  I updated the logic so that keyboard navigation will happen for both when the popover host has focus or the `ion-item` within it. 

The previous tests were only verifying that `tab` set focus to the first item. I left that test as-is and added a new test that verifies that `ArrowDown` also focuses the first `ion-item`.

liamdebeasi (approved):
Great job! Can we make a dev build and provide it on the issue thread before merging?

liamdebeasi:
@sean-perkins Doesn't look like we got any feedback from community. I am happy with the PR, and sounds like Amanda is too. Feel free to merge if you are ready to ship this.

PR #25010, opened 2022-03-28
fix(datetime): time picker display matches dynamically set value

sean-perkins:
## Pull request checklist

## Pull request type

 

 

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

If you have a `ion-datetime` with `presentation="time"` and the initial value is in the "PM" and you attempt to update the value to "AM", the value will not update correctly.

This is caused by the working parts not being updated when the value changes. The working parts are used for generating the available hour options, which will include the value being offset by 12, if the value is PM vs. AM. When dynamically setting the value, these options were not being generated correctly, so the resulting set of available hours would not match the current value and our fallback to scroll to the "zero" item would fire, updating the value to "12" PM.

Issue URL: #24967

## What is the new behavior?

When updating the value of the datetime, the time picker will correctly display the value.

- The working parts am/pm representation is updated to be in-sync with the current value. 
- The available hour options are generated based on the current value am/pm

- [ ] Yes
- [x] No

## Other information

Dev build: `6.0.15-dev.11649352526.165b8639` ✅


amandaesmith3:
Would it make sense here to post an updated dev build in the original issue? Wanted to check on that before I review since I noticed an old version of this PR didn't solve things for the OP.

sean-perkins:
> Would it make sense here to post an updated dev build in the original issue? Wanted to check on that before I review since I noticed an old version of this PR didn't solve things for the OP.

I can 👍  the stackblitz example there were using was also pointing to the latest, not the dev build (but the dev build contents have changed since then). 

The Playwright test is checking each condition (AM->AM, AM-PM, PM-AM, PM->PM) x20, so I do feel confident in the behavior more than with our puppeteer tests x1 execution.

sean-perkins:
Dev build verified by issue reporter: https://github.com/ionic-team/ionic-framework/issues/24967#issuecomment-1092327342

core/src/components/datetime/test/set-value/index.html
```diff
@@ -9,19 +9,22 @@
...
+        <title>Datetime - Set Value</title>
+        <meta name="viewport" content="width=device-width, initial-scale=1.0, minimum-scale=1.0">
+        <link href="../../../../../css/ionic.bundle.css" rel="stylesheet">
+        <link href="../../../../../scripts/testing/styles.css" rel="stylesheet">
+        <script src="../../../../../scripts/testing/scripts.js"></script>
+        <script type="module" src="../../../../../dist/ionic/ionic.esm.js"></script>
+        <style>
+          ion-datetime {
+            width: 350px;
+          }
```
amandaesmith3, line 27: Looks like there was some kind of copy-paste goof here?

amandaesmith3 (changes requested):
Not worrying about the cropped screenshots since you have that other PR up to fix those.

amandaesmith3 (approved):

core/src/utils/test/playwright/locator/utils/wait-for-custom-event.ts
```diff
@@ -0,0 +1,27 @@
+import type { Locator } from '@playwright/test';
+
+import type { E2ELocator } from '../../playwright-declarations';
+
+/**
+ * Waits for a custom event to be dispatched on an element handle. Fails after 5000ms without an event emission.
+ */
+export const waitForCustomEvent = (
```
liamdebeasi, line 8: Is this still needed now that event spies exist? The descriptions in the `IonicPage` interface seem to indicate they perform the same functions.

liamdebeasi (approved):

core/src/utils/test/playwright/locator/utils/wait-for-custom-event.ts
sean-perkins: Nope, good catch! I can remove that & clean-up the page interface. Thinking it might be useful if we throw a quick `testing.md` in our playwright folder, for how to use our basic test utilities. Would be a decent reference for us as we leverage those custom function more & a great resource for future contributors.

Issue #17839, opened 2019-03-20
ion-range start and end events


JumBay:
# Feature Request

**Ionic version:**

[x] **4.x**

**Describe the Feature Request**

As discussed here https://github.com/ionic-team/ionic/issues/17299, it would be nice to have a **start even**t triggered when knob is pressed and an **end event** triggered when the knob is released like the behavior of ionFocus and ionBlur in ionic v3.

Thanks

biesbjerg:
I gave it a shot. It's my first pull request-ish, so be gentle ;-)

liamdebeasi:
Hi there,

Thanks for the feature request. How do the `start` and `end` events differ from the `focus` and `blur` events?  The `focus` event should fire when the knob is pressed, and the `blur` is fired when the knob is released.

Thanks!

JumBay:
That was the behavior in ionic 3 but it has changed in v4 see: https://github.com/ionic-team/ionic/issues/17299#issuecomment-458490016

liamdebeasi:
Hi everyone,

I spoke with the team and there are a few things going on here:

1. The `focus`/`blur` behavior in `ion-range` isn't quite right. `focus` should fire when the element comes into focus (tabbing in, clicking, etc). `blur` should fire when the element loses focus (clicking outside of the range).

2. We need a way to capture mousedown/mouseup since essentially what we were capturing in v3. I think the PR that was created attempts to resolve this, so I will work with the PR author on cleaning this up.

Thanks!

biesbjerg:
@liamdebeasi `focus`/`blur` seems to work as advertised (like you would expect any input component to behave). 

I committed another change that also triggers `start`/`end` when changing range using the keyboard. 
The `start`-event will always contain the range value before the knob was moved and the `end`-event will contain the final range value after the knob has been released.

I'm not sure if I'm supposed to squash the commits (or you can do it?) or how - sorry!

liamdebeasi:
Hi @biesbjerg,

I created a CodePen that shows the `ion-range` behavior vs the native `range` input behavior: https://codepen.io/liamdebeasi/pen/eaZyxW. Clicking the track on `ion-range` fires both the focus and blur events, which is not correct.

My `focus`/`blur` comment was in regards to the existing behavior of the `ion-range`, not in regards to your PR. I will post any feedback for that on the PR page. 🙂 

Don't worry about squashing commits, we will squash and merge your PR when it's ready.

biesbjerg:
[...]

PR #25011, opened 2022-03-28
feat(range): add knobMoveStart and knobMoveEnd events

amandaesmith3:
## Pull request checklist

## Pull request type

 

 

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes

The only way to listen for user interaction with an `ion-range` is the `ionChange` event. This doesn't fit all use cases, as it fires many times throughout a single drag gesture (every time the range's value updates). If a dev wants to listen for the user's gesture to finish, debouncing an `ionChange` listener wouldn't work either, as it would delay the drag-end handler by the debounce time.

Issue URL: Resolves https://github.com/ionic-team/ionic-framework/issues/17839

## What is the new behavior?

Adds two new events: `ionKnobMoveStart` and `ionKnobMoveEnd`. These fire at, respectively, the start and end of a touch gesture or mouse drag on the range. Both events include the range's `value` at the time. Keyboard interaction is also handled.

- [ ] Yes
- [x] No


core/src/components/range/range.tsx
```diff
@@ -191,6 +191,16 @@ export class Range implements ComponentInterface {
...
 
+  /**
+   * Emitted when the user starts moving the range knob.
+   */
+  @Event() ionChangeStart!: EventEmitter<RangeChangeStartEventDetail>;
+
+  /**
+   * Emitted when the user finishes moving the range knob.
+   */
+  @Event() ionChangeEnd!: EventEmitter<RangeChangeEndEventDetaill>;
```
amandaesmith3, line 202: I'm not married to these event names. I know they could be confused with hypothetical `ionWillChange` and `ionDidChange` events, which would imply they fire alongside every `ionChange` event instead of once each. I wanted something more specific than just `ionStart` and `ionEnd`, though. Open to suggestions 🤔

core/src/components/range/range-interface.ts
```diff
@@ -8,7 +8,15 @@ export interface RangeChangeEventDetail {
   value: RangeValue;
 }
 
+export interface RangeChangeStartEventDetail {
```
sean-perkins, line 11: ~ Thoughts on re-using the existing `RangeChangeEventDetail`? Unsure of the value of separate interfaces, if the type signature is identical.

core/src/components/range/range.tsx
```diff
@@ -257,7 +267,10 @@ export class Range implements ComponentInterface {
     } else {
       this.ratioB = clamp(0, this.ratioB + step, 1);
     }
+
+    this.ionChangeStart.emit({ value: this.value });
```
sean-perkins, line 271: ~ Consider using a local variable here, since we are accessing the same variable multiple times:
```tsx
const { value } = this;
this.ionChangeStart.emit({ value });
this.updateValue();
this.ionChangeEnd.emit({ value });
```

core/src/components/range/test/basic/e2e.ts
```diff
@@ -17,3 +18,37 @@ test('range:rtl: basic', async () => {
   const compare = await page.compareScreenshot();
   expect(compare).toMatchScreenshot();
 });
+
+test('range: start/end events', async () => {
```
sean-perkins, line 22: ~ Can we add a test case for clicking the range without dragging and making sure that both start and end events fire?

core/src/components/range/range.tsx
```diff
@@ -191,6 +191,16 @@ export class Range implements ComponentInterface {
    */
   @Event() ionBlur!: EventEmitter<void>;
 
+  /**
+   * Emitted when the user starts moving the range knob.
```
sean-perkins, line 195: ~ Should we call out this fires with both keyboard input and dragging the knob?

sean-perkins (commented):
I am also indifferent to the event names. I do not think they are immediately recognizable as to what they are used for, but also cannot think of an alternative name that is more clear than `ionChangeStart` and `ionChangeEnd`.

core/src/components/range/range-interface.ts
amandaesmith3: We'd agreed to use separate event details on the design doc, to avoid potential confusion and make them easier to update if we need new fields later. I'm open to changing it though if y'all don't think it'll look weird down the road 🤷‍♀️

core/src/components/range/range.tsx
amandaesmith3: `updateValue` will change the value, so we need to re-access it rather than using a cached version.

core/src/components/range/range.tsx
amandaesmith3: Good thinking, will update :eyes:

core/src/components/range/range.tsx
sean-perkins: Ahh I missed that, thanks for the clarification!

sean-perkins (approved):
Looks good to me, great work on this 🎉

core/src/components/range/range-interface.ts
```diff
@@ -8,7 +8,15 @@ export interface RangeChangeEventDetail {
...
+  value: RangeValue;
+}
+
+export interface RangeChangeEndEventDetaill {
```
liamdebeasi, line 15:
```suggestion
export interface RangeChangeEndEventDetaill {
```
```suggestion
export interface RangeChangeEndEventDetail {
```
One extra "L"

liamdebeasi, line 15:
```suggestion
export interface RangeChangeEndEventDetail {
```

One extra "L"

liamdebeasi, line 15:
```suggestion
export interface RangeChangeEndEventDetail {
```

One extra "L"

core/src/components/range/range.tsx
```diff
@@ -191,6 +191,18 @@ export class Range implements ComponentInterface {
    */
   @Event() ionBlur!: EventEmitter<void>;
 
+  /**
+   * Emitted when the user starts moving the range knob, whether through
+   * mouse drag, touch gesture, or keyboard interaction.
+   */
+  @Event() ionChangeStart!: EventEmitter<RangeChangeStartEventDetail>;
```
liamdebeasi, line 198: Bikeshedding here, but thoughts on calling this `ionStart` and `ionEnd`? To me `ionChangeStart` implies that `ionChange` will fire after `ionChangeStart`. If you have a pointer on the knob, `ionChangeStart` will fire, but if you simply remove the point and do not move the knob, `ionChange` will not fire.

liamdebeasi, line 198: Bikeshedding here, but thoughts on calling this `ionStart` and `ionEnd`? To me `ionChangeStart` implies that `ionChange` will fire after `ionChangeStart`. If you have a pointer on the knob, `ionChangeStart` will fire, but if you simply remove the point and do not move the knob, `ionChange` will not fire.

core/src/components/range/range-interface.ts
```diff
@@ -8,7 +8,15 @@ export interface RangeChangeEventDetail {
...
+  value: RangeValue;
+}
+
 export interface RangeCustomEvent extends CustomEvent {
-  detail: RangeChangeEventDetail;
+  detail: RangeChangeEventDetail | RangeChangeStartEventDetail | RangeChangeEndEventDetaill;
```
liamdebeasi, line 20: Why are we making separate interfaces if the definitions are all the same?

liamdebeasi, line 20: Why are we making separate interfaces if the definitions are all the same?

liamdebeasi (changes requested):

core/src/components/range/range.tsx
liamdebeasi: Or maybe `ionKnobStart` or `ionKnobEnd` if we wanted something more specific?

core/src/components/range/range-interface.ts
amandaesmith3: We'd agreed to use separate event details on the design doc, to avoid potential confusion and make them easier to update if we need new fields later. I'm open to changing it though if y'all don't think it'll look weird down the road 🤷‍♀️

core/src/components/range/range.tsx
amandaesmith3: Hmm, good point re: changeStart vs change. 🤔 What about `ionKnobMoveStart`? That feels sufficiently descriptive.

liamdebeasi: Yeah that sounds good 👍

core/src/components/range/range-interface.ts
```diff
@@ -8,7 +8,15 @@ export interface RangeChangeEventDetail {
   value: RangeValue;
 }
 
+export interface RangeChangeStartEventDetail {
+  value: RangeValue;
+}
+
+export interface RangeChangeEndEventDetail {
+  value: RangeValue;
+}
```
liamdebeasi, line 17: These need to be documented in `readme.md`. You can add them above/under the `RangeChangeEventDetail` interface.

liamdebeasi (changes requested):

liamdebeasi (commented):
Do we need to call `ensureValueInBounds`? https://github.com/ionic-team/ionic-framework/blob/a4803ec99d395c4eba1e7ebffe426cae4452a641/core/src/components/range/range.tsx#L153-L155

I don't know if the `valueChanged` callback will update the `value` prop with the clamped values.

amandaesmith3:
Good catch, thanks!

PR #25012, opened 2022-03-29
test(playwright): custom mode/rtl preference can now be specified

liamdebeasi:
## Pull request checklist

## Pull request type

 

 

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes
- [x] Other (please describe):  Test Infra

This PR addresses a few points of feedback received during the PR review.

1. Developers can now manually override the mode/RTL preference on a per-test basis. The `getSnapshotSettings` method also works with this customization.
2. Developers can query for the mode/RTL preference via `testInfo.project.metadata`. This is a Playwright feature and has been documented in our internal testing guide

Previously it was discussed that the test name should make note of the mode/RTL file name. However, being able to customize the mode/RTL preference on a per test basis means this is not possible. Customizing the mode/RTL preference is done in the `page.goto` method by passing the preferences as part of the URL. By the time this is done, the test name has already been specified.

A few things though:

1. The screenshot file name notes the mode/RTL preference.
2. The playwright report also notes that mode/RTL preference when it logs the network request.

Open to other ideas, but does not seem to be easily doable if we want to customize the mode/RTL preference.

- [ ] Yes
- [x] No


sean-perkins (approved):
This is perfect 👍  if manually specifying the mode/direction setting, I would expect we are manually specifying that in the test statement/describe anyways.

core/src/utils/test/playwright.ts
```diff
@@ -14,7 +14,19 @@ export const test = base.extend({
      */
     page.goto = (url: string) => {
       const { mode, rtl } = testInfo.project.metadata;
-      const formattedUrl = `${url}?ionic:_testing=true&ionic:mode=${mode}&rtl=${rtl}`;
+
+      const splitUrl = url.split('?');
+      const paramsString = splitUrl[1];
```
amandaesmith3, line 19: Will `url` always have params in it already? (Wondering if directly hitting `splitUrl[1]` with no safeguards will cause problems. Also applies to `getSnapshotSettings`)

liamdebeasi: It will only have params in it if you are explicitly overriding the mode/RTL config. If you index into an out of bounds item in the array with JS, you get `undefined` instead of an error.

The `init` parameter for `URLSearchParams` is optional, so passing in `undefined` will not error either: https://developer.mozilla.org/en-US/docs/Web/API/URLSearchParams/URLSearchParams

amandaesmith3 (approved):

PR #24440, opened 2021-12-17
fix(angular): allow default browser action on CTRL+Click

Juarrow:
## Pull request checklist

Please check if your PR fulfills the following requirements:
- [ ] Tests for the changes have been added (for bug fixes / features)
- [x] Docs have been reviewed and added / updated if needed (for bug fixes / features)
- [ ] Build (`npm run build`) was run locally and any changes were pushed
- [x] Lint (`npm run lint`) has passed locally and any fixes were made for failures

## Pull request type

 

 

Please check the type of change your PR introduces:
- [x] Bugfix
- [ ] Feature
- [ ] Code style update (formatting, renaming)
- [ ] Refactoring (no functional changes, no api changes)
- [ ] Build related changes
- [ ] Documentation content changes
- [ ] Other (please describe): 

## What is the current behavior?

Issue Number: #24413

## What is the new behavior?

- Allow CTRL/Meta + Click to be handled by the browser (e.g. for opening a link in a new tab)

## Does this introduce a breaking change?

- [ ] Yes
- [x] No

Issue #24413, opened 2021-12-15
bug: angular, routerLink prevents opening link in new tab


Juarrow:
### Prerequisites

- [X] I have read the [Contributing Guidelines](https://github.com/ionic-team/ionic-framework/blob/main/.github/CONTRIBUTING.md#creating-an-issue).
- [X] I agree to follow the [Code of Conduct](https://ionicframework.com/code-of-conduct).
- [X] I have searched for [existing issues](https://github.com/ionic-team/ionic-framework/issues) that already report this problem, without success.

### Ionic Framework Version

- [x] v4.x
- [X] v5.x
- [X] v6.x

### Current Behavior

When importing the IonicModule and using ``[routerLink]``,
the following line prevents Ctrl+Click behavior on Desktop browsers.

https://github.com/ionic-team/ionic-framework/blob/732f8e10ce604f1a3e98518ae9c3a4afd7803e9a/angular/src/directives/navigation/router-link-delegate.ts#L56

### Expected Behavior

Ctrl+Click should trigger the default browser behavior (e.g. open the link in a new tab/page/etc).

### Steps to Reproduce

1. Create an anchor element with a ``[routerLink]="[<your link>]"``
2. Ctrl+Click on it

### Code Reproduction URL

[Link to repository](https://github.com/Juarrow/ionic-angular-ctr-click-bug)

On the home page CTRL+Click the link

### Ionic Info

```
Ionic:

   Ionic CLI                     : 6.18.1 (C:\Users\Julian\AppData\Roaming\npm\node_modules\@ionic\cli)
   Ionic Framework               : @ionic/angular 6.0.1
   @angular-devkit/build-angular : 13.1.1
   @angular-devkit/schematics    : 13.0.3
   @angular/cli                  : 13.0.3
   @ionic/angular-toolkit        : 5.0.3

Utility:

   cordova-res : not installed globally
   native-run  : 1.5.0

System:

   NodeJS : v16.13.0 (C:\Program Files\nodejs\node.exe)
   npm    : 8.1.0
   OS     : Windows 10
```

### Additional Information

I'm currently monkey-patching it 🙈 .

Maybe this can be fixed by something like (pseudo-code)
````
if(platform is desktop && (key == CTRL || key == Meta) && button == 0)) {
 return true;
}
````

Juarrow:
@liamdebeasi added the [reproduction repo link](https://github.com/Juarrow/ionic-angular-ctr-click-bug) to the issue.

[...]

PR #25014, opened 2022-03-29
fix(angular): routerLink allows opening in a new tab for link elements

sean-perkins:
## Pull request checklist

## Pull request type

 

 

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

Users are unable to ctrl/CMD + click on link elements (`<a routerLink="/">`) to open a link in a new tab.

Issue URL: #24413

## What is the new behavior?

Removes preventing the default behavior on the click event, to re-add support for ctrl/CMD + clicking on a link element to open the URL in a new tab.

Note: Angular's router link directive [source code can be found here](https://github.com/angular/angular/blob/master/packages/router/src/directives/router_link.ts). There is no behavior that prevents the event, so I am uncertain why we did in the first place.

- [ ] Yes
- [x] No

## Other information

Closing #24440 in favor of this PR (co-authored-by applied to the commit).

Cypress test support for CTRL/CMD + click is pretty tricky/finicky. Opted for not writing tests on this behavior at this time. 

If we would like to support this behavior on our Ionic components in the future, we will likely need to render an `a` tag around the element or swap the inner element to an `a` tag to support the browser default behaviors.

### DIFF
--- a/angular/src/directives/navigation/router-link-delegate.ts
+++ b/angular/src/directives/navigation/router-link-delegate.ts
@@ -5,6 +5,12 @@ import { AnimationBuilder, RouterDirection } from '@ionic/core';
 
 import { NavController } from '../../providers/nav-controller';
 
+/**
+ * Adds support for Ionic routing directions and animations to the base Angular router link directive.
+ *
+ * When the router link is clicked, the directive will assign the direction and
+ * animation so that the routing integration will transition correctly.
+ */
 @Directive({
   selector: '[routerLink]',
 })
@@ -31,19 +37,15 @@ export class RouterLinkDelegateDirective implements OnInit, OnChanges {
     this.updateTargetUrlAndHref();
   }
 
+  @HostListener('click')
+  onClick(): void {
+    this.navCtrl.setDirection(this.routerDirection, undefined, undefined, this.routerAnimation);
+  }
+
   private updateTargetUrlAndHref() {
     if (this.routerLink?.urlTree) {
       const href = this.locationStrategy.prepareExternalUrl(this.router.serializeUrl(this.routerLink.urlTree));
       this.elementRef.nativeElement.href = href;
     }
-  }
-
-  /**
-   * @internal
-   */
-  @HostListener('click', ['$event'])
-  onClick(ev: UIEvent): void {
-    this.navCtrl.setDirection(this.routerDirection, undefined, undefined, this.routerAnimation);
-    ev.preventDefault();
   }
 }


amandaesmith3 (approved):
Gave this a shot with dev build `6.0.14-dev.11648653425.149f5ee0`, works great 👍 I'm not sure why we had that `preventDefault` either haha. The only thing I could think of was maybe preventing hard refreshes in the page transition due to the `href`, but I'm still seeing the nice SPA transition. So yeah, looks good.

liamdebeasi (approved):

angular/src/directives/navigation/router-link-delegate.ts
```diff
@@ -31,19 +37,15 @@ export class RouterLinkDelegateDirective implements OnInit, OnChanges {
     this.updateTargetUrlAndHref();
   }
 
+  @HostListener('click')
```
liamdebeasi, line 40: Do we still need the "@internal" comment?

liamdebeasi: Apologies to the user I just tagged by accident 😂

sean-perkins: I don't believe so. I'm unsure why it was marked as `@internal` in the first place. Not a convention that exists in the Angular code base for a similar implementation: https://github.com/angular/angular/blob/master/packages/router/src/directives/router_link.ts#L238-L239

Also not a method that developers are accessing through a public API.

lincolnthree:
I believe this PR has introduced a regression where *all* `routerLink`s now cause the entire page to reload (e.g. they break the single-page app paradigm). Working on a reproduction in the morning. I can't be sure it was this PR, but upgrading to 6.0.15 introduced the behavior, and rolling back to 6.0.14 has resolved the issue.

Just wanted to give y'all a heads up just in case, since this could affect a bunch of people if they aren't rigorous in testing. More info after some sleep.

liamdebeasi:
Thanks for the heads up. I can reproduce this and will work on a fix.

lincolnthree:
@liamdebeasi Awesome, thank you. I won't bother with a repro then :) You are the best as always.

liamdebeasi:
Mind giving this dev build a test?

```
6.0.15-dev.11649337697.1fd31982
```

Pressing `ion-button` or `ion-item` with a `routerLink` should not cause a page refresh. Pressing an `a` with an `href` should cause a page refresh, and you should also be able to open those links in a new tab with a Ctrl/Cmd + Click.

liamdebeasi:
This issue is resolved in Ionic 6.0.16. Please update your apps to receive a fix:

```
npm install @ionic/angular@6.0.16
```

Issue #24634, opened 2022-01-24
bug: ion-menu containing an ion-accordion scrolls up every time the menu opens


erikramalho:
### Prerequisites

- [X] I have read the [Contributing Guidelines](https://github.com/ionic-team/ionic-framework/blob/main/.github/CONTRIBUTING.md#creating-an-issue).
- [X] I agree to follow the [Code of Conduct](https://ionicframework.com/code-of-conduct).
- [X] I have searched for [existing issues](https://github.com/ionic-team/ionic-framework/issues) that already report this problem, without success.

### Ionic Framework Version

- [ ] v4.x
- [ ] v5.x
- [X] v6.x

### Current Behavior

If we put a ion-accordion inside a ion-menu with a ion-split-pane app, the menu scrolls up every time you close and reopen the ion-menu. The ion-menu scroll up right after the event ionDidOpen finishes.

### Expected Behavior

The ion-menu must keep the scroll postion. This behavior is seen if you put a ion-list + ion-item in a ion-menu.

### Steps to Reproduce

Just create a side menu app (ionic start and side menu) and then place a multiple ion-accordion / ion-accordion-group.  In the link provided, 

1. open the menu, expand the accordion
2. scroll down the menu 
3. click outside the menu to close it
4. reopen the menu: a scroll to top will happen. 

### Code Reproduction URL

https://stackblitz.com/edit/ionic-5-custom-ion-menu-button-jikmqz?file=src/app/tab1/tab1.page.html

### Ionic Info

Ionic CLI : 6.18.1 (/node/v12.22.5/lib/node_modules/@ionic/cli)
Ionic Framework : @ionic/angular 6.0.1
@angular-devkit/build-angular : 12.2.14
@angular-devkit/schematics : 12.2.14
@angular/cli : 12.2.14
@ionic/angular-toolkit : 5.0.1

amandaesmith3:
Thanks for the issue, I'm able to replicate in both Angular and core. The bug triggers if you have at least one accordion in a menu. It doesn't trigger with just a list in the menu, but I saw it using just one accordion group with one accordion. It looks like the accordion's content can be anything; you don't necessarily need an `ion-list` anywhere.

sean-perkins:
Additional context from #24660, this also appears to happen when the `ion-menu` content container has overflow scrolling. You can replicate in the conference app, by adding additional padding to the menu's content container and toggling the visibility of the menu. The menu's scroll position is always snapping back to the top. 

[...]

PR #25044, opened 2022-04-01
fix(menu): preserve scroll position when focusing on open

amandaesmith3:
## Pull request checklist

## Pull request type

 

 

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

When an `ion-menu` is opened, the first (focusable) element inside it is auto-focused for screen readers. This causes the menu's scroll position to jump to the focused element, and only once the opening animation has finished, leading to a jarring experience.

Issue URL: Resolves https://github.com/ionic-team/ionic-framework/issues/24634

## What is the new behavior?

When opening the menu, the host element is focused instead of the first element. This allows focus trapping without affecting the scroll position.

- [ ] Yes
- [x] No

## Other information

An alternative would be to use the `preventScroll` option in `focusFirstDescendant`:
```ts
element.focus({ preventScroll: true });
```
However, this isn't currently supported in Safari: https://caniuse.com/mdn-api_htmlelement_focus_options_preventscroll_parameter

Side note: I elected to write the test in Puppeteer for the sake of time, rather than migrating to Playwright.


core/src/components/menu/menu.tsx
```diff
@@ -396,6 +396,28 @@ export class Menu implements ComponentInterface, MenuI {
     }
 
     this.beforeAnimation(shouldOpen);
+
+    if (shouldOpen) {
+      // focus menu content for screen readers, preserving scroll position
+      if (this.menuInnerEl) {
```
liamdebeasi, line 402: I know Sean had suggested focusing the host of `ion-menu` instead of trying to mess with the scroll position. Any reason why we did not do that here?

liamdebeasi: Another thought: We do have an open issue with focus + menu (FW-293). Perhaps focusing the `ion-menu` host would help resolve that issue as well?

You could do something like the following:

```
// check to see if the activeElement is inside of ion-menu
  // if it is, do nothing and return early
// otherwise focus ion-menu
```

That way you can get the benefits of focus trapping without having to manage the scroll position.

amandaesmith3: I thought it would cause screen reader issues, but I gave it a try and it seems to work okay 🤔 I'll fiddle with it some more in the morning.

amandaesmith3: Tried it with just focusing `this.el` instead of the first descendant (and leaving focus trapping alone) and that seems to work great 😆 No scroll jumping, and it plays nice with screen readers.

core/src/components/menu/menu.tsx
```diff
@@ -619,12 +620,12 @@ export class Menu implements ComponentInterface, MenuI {
...
-        this.focusFirstDescendant();
-      }
-
-      // setup focus trapping
+      /**
+       * Set up focus trapping. Focus the host element to start with
+       * instead of the first descendant to avoid the scroll position
+       * jumping to that element right away.
+       */
+      this.el.focus();
```
liamdebeasi, line 628: Should this check to see if the focused element is already inside of `ion-menu`? If it is, we probably do not want to change the focus.

core/src/components/menu/test/basic/e2e.ts
```diff
@@ -28,13 +28,42 @@ test('menu: focus trap', async () => {
...
+  activeElID = await getActiveElementID(page);
   expect(activeElID).toEqual('start-menu-button');
 
+  // do it again to make sure focus stays inside menu
   await page.keyboard.press('Tab');
   activeElID = await getActiveElementID(page);
   expect(activeElID).toEqual('start-menu-button');
 });
 
+test('menu: preserve scroll position', async () => {
```
liamdebeasi, line 43: I know mandatory Playwright tests do not start until next sprint, but could you write this in Playwright if you have time? Focus management has been tricky across browsers, so it would be great to test this in Firefox, Chrome, and Safari. 

If you don't have time this sprint I can make a ticket for next sprint. Let me know what you would prefer.

core/src/components/menu/menu.tsx
amandaesmith3: This is only when opening the menu, so the focus should be on whatever button etc. triggered that (which won't be inside it).

core/src/components/menu/test/basic/e2e.ts
amandaesmith3: I would go ahead and make the ticket; if I have time I can knock it out, but I'm not confident 🤔

core/src/components/menu/menu.tsx
liamdebeasi: What happens if a user manually shifts focus? (I.e. do we still respect that focus?)

amandaesmith3: Shifts focus to where? If you hit tab after opening the menu, it'll move to the contents and trap your focus, but not sure if that's what you meant.

liamdebeasi: If a user manually focuses an element inside of `ion-menu`, I would expect that the focus is preserved when opening the menu. For example: Someone clicks the menu button and focuses a button inside the menu as it is opening.

core/src/components/menu/menu.tsx
```diff
@@ -619,12 +620,17 @@ export class Menu implements ComponentInterface, MenuI {
...
-      // focus menu content for screen readers
-      if (this.menuInnerEl) {
-        this.focusFirstDescendant();
+      /**
+       * Move focus to the menu to prepare focus trapping, as long as
+       * it isn't already focused. Use the host element instead of the
+       * first descendant to avoid the scroll position jumping around.
+       */
+      const focusedMenu = document.activeElement?.closest('ion-menu');
+      if(!focusedMenu || focusedMenu !== this.el) {
```
liamdebeasi, line 629: Optional: This can be simplified:

```suggestion
      if (focusedMenu !== this.el) {
```

If `focusedMenu` is `null` (possible return value from closest()) then it won't be equal to `this.el` which is an HTML element.

liamdebeasi (approved):
Great job!

PR #25049, opened 2022-04-04
fix(datetime): handle invalid dates when value changes

sean-perkins:
## Pull request checklist

## Pull request type

 

 

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

When assigning an invalid value to a datetime in Firefox, such as:
```
document.querySelector('ion-datetime[presentation="time"]').value = '4:10'; // Valid value would be 04:10
```
In Firefox an exception will be thrown that states:
> TypeError: parseDate(...) is undefined

This exception is not present in Safari or Chrome and isn't helpful or clear to developers. 

## What is the new behavior?

- Adds check for a valid parsable date before trying to assign active parts
- No exception is thrown in Firefox
- Warning is logged to the developer that an invalid date string was provided

- [ ] Yes
- [x] No


core/src/utils/test/playwright.ts
```diff
@@ -106,6 +106,8 @@ export const test = base.extend<CustomFixtures>({
     page.setIonViewport = async () => {
       const currentViewport = page.viewportSize();
 
+      await page.waitForSelector('ion-content');
```
liamdebeasi, line 109: With the other fix I merged into `main`, is this still needed?

liamdebeasi (approved):
Looks good other than the comment. Don't forget to update the commit subject to be more descriptive. Maybe something like "warn when parsing an invalid date value"?

amandaesmith3 (approved):

core/src/utils/test/playwright.ts
sean-perkins: The behavior it was added to resolve, was fixed by your other changes. There is some value of it existing, in that we shouldn't be calling `setIonViewport` on a test template that doesn't use `ion-content`. If we use the `waitForSelector`, an invalid test will fail and help inform that that tag is expected in the test file. I think this is rare though, so I'd be happy to remove for now and re-add if we would find value in the safe guard.

PR #25057, opened 2022-04-04
feat(modal): clicking handle advances to the next breakpoint

sean-perkins:
## Pull request checklist

## Pull request type

 

 

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes

When clicking the handle inside a sheet modal, it does not advance to the next breakpoint.

Issue URL: #24069

## What is the new behavior?

Clicking the handle will advance to the next breakpoint. If the user is at the last breakpoint, the sheet will advance to the first non-zero breakpoint.

While native iOS does have the ability to double tap the handle to dismiss the sheet modal (in certain experiences), we are not replicating that behavior in this feature. I believe that UX could be confusing and disruptive to users.

- [ ] Yes
- [x] No


core/src/components/modal/modal.tsx
```diff
@@ -689,6 +689,26 @@ export class Modal implements ComponentInterface, OverlayInterface {
...
+       * Defaults to the first breakpoint. If the first breakpoint is 0, we select
+       * the second breakpoint to avoid tapping the handle to dismiss the sheet.
+       */
+      let nextBreakpoint = breakpoints[0] || breakpoints[1];
+      if (currentBreakpointIndex !== breakpoints.length - 1) {
+        /**
+         * Since the current breakpoint is not the last breakpoint,
+         * we advance to the next breakpoint.
+         */
+        nextBreakpoint = breakpoints[currentBreakpointIndex + 1];
```
liamdebeasi, line 706: Optional, An alternative way of doing this:

```suggestion

     const allowedBreakpoints = breakpoints.filter(b => b !== 0);
     const currentBreakpointIndex = allowedBreakpoints.indexOf(currentBreakpoint);
     const nextBreakpoint = (currentBreakpointIndex + 1) % allowedBreakpoints.length;
     
     this.setCurrentBreakpoint(nextBreakpoint);
```

The modulo operator ensures we never go out of bounds on the breakpoints array.

core/src/components/modal/modal.tsx
```diff
@@ -756,7 +776,7 @@ export class Modal implements ComponentInterface, OverlayInterface {
           part="content"
           ref={el => this.wrapperEl = el}
         >
-          {showHandle && <div class="modal-handle" part="handle"></div>}
+          {showHandle && <div class="modal-handle" part="handle" onClick={this.onHandleClick}></div>}
```
liamdebeasi, line 779: Since this is not a button, does Axe or other assistive tech not like this?

sean-perkins: Will confirm, I would assume it is an issue. Should be a button based on the revised behavior. Good catch 👍

core/src/components/modal/modal.tsx
sean-perkins: This solution will need to be adapted to:
```ts
const allowedBreakpoints = breakpoints.filter(b => b !== 0);
const currentBreakpointIndex = allowedBreakpoints.indexOf(currentBreakpoint);
const nextBreakpointIndex = (currentBreakpointIndex + 1) % allowedBreakpoints.length + 1;

this.setCurrentBreakpoint(breakpoints[nextBreakpointIndex]);
```

Plugging the original in; `nextBreakpoint` was actually the index position of the breakpoint. After swapping that to `breakpoints[nextBreakpoint]`, it was always evaluating the current breakpoint instead of the next value.

Does the updated snippet look correct to you? Confirmed locally that it loops through the values as expected.

liamdebeasi: That works as well. Another option is to just use `allowedBreakpoints` when calling `setCurrentBreakpoint`. That way you don't have to add `1` to the result every time.

core/src/components/modal/modal.tsx
```diff
@@ -756,7 +771,7 @@ export class Modal implements ComponentInterface, OverlayInterface {
           part="content"
           ref={el => this.wrapperEl = el}
         >
-          {showHandle && <div class="modal-handle" part="handle"></div>}
+          {showHandle && <button class="modal-handle" part="handle" type="button" onClick={this.onHandleClick}></button>}
```
liamdebeasi, line 774:
```suggestion
          {showHandle && <button class="modal-handle" part="handle" onClick={this.onHandleClick}></button>}
```

The type is redundant when we are using `<button>`

core/src/components/modal/modal.scss
```diff
@@ -124,6 +124,8 @@
    */
   transform: translateZ(0);
 
+  border: 0;
```
liamdebeasi, line 127: Should this also have `cursor: pointer` too?

core/src/components/modal/modal.tsx
```diff
@@ -689,6 +689,21 @@ export class Modal implements ComponentInterface, OverlayInterface {
     return this.currentBreakpoint;
   }
 
+  private onHandleClick = () => {
+    const { breakpoints, currentBreakpoint } = this;
```
liamdebeasi, line 693: Tabbing to the handle and pressing Enter multiple times causes the animation to jump.

https://user-images.githubusercontent.com/2721089/161641895-824bebbe-9e22-43fa-821c-6743f0d26c36.mov

core/src/components/modal/modal.tsx
liamdebeasi, line 774: Comparing with native iOS, VoiceOver will announce something like "sheet grabber, button, double tap to expand the sheet". When the sheet is expanded it will say "sheet grabber, button, double tap to collapse the sheet".

Ours just says "web dialog, button" which is not very helpful. We may not be able to get full parity with native, but can we find a way to make the read out more descriptive?

core/src/components/modal/modal.tsx
```diff
@@ -689,6 +689,21 @@ export class Modal implements ComponentInterface, OverlayInterface {
     return this.currentBreakpoint;
   }
 
+  private onHandleClick = () => {
```
amandaesmith3, line 692: The default handle is tiny, so clicking on it is super fiddly. You can drag the whole header, so maybe it would make sense to allow clicking that instead? 🤔 Failing that, maybe some padding or a wrapper element or something around the handle to expand the clickable space?

amandaesmith3 (changes requested):

core/src/components/modal/modal.tsx
sean-perkins: I think padding around the clickable space could make a lot of sense here. I wouldn't advise we add clicking to the header, as it would be different behavior than native.

core/src/components/modal/modal.tsx
sean-perkins: Adjusted this behavior slightly, so that advancing is tied to keyup instead of keydown. Setting the current breakpoint does not currently wait for the previous transition to finish. If you perform an action while a transition is inflight, it will disrupt the transition.

amandaesmith3: Looks like it's still jumping if you mash Enter; I feel like if there's a transition in progress when you click/activate the handle, it should just do nothing.

sean-perkins: @amandaesmith3 to confirm, if the user performs any action (clicking the handle or pressing enter), while the sheet modal is actively transitioning, the expected behavior would be to perform no operation (skip the interaction)?

amandaesmith3: Looking back at the code, `present` and `dismiss` will both wait for the current transition and then go through with the attempted interaction, so we should probably go with that instead here. Thinking on it, if I hit Enter twice quickly, having it only move to the next breakpoint once would feel off. 🤔

core/src/components/modal/modal.tsx
```diff
@@ -719,6 +719,37 @@ export class Modal implements ComponentInterface, OverlayInterface {
...
+    if (ev.key === 'Enter') {
+      this.moveToNextBreakpoint();
+    }
+  };
+
+  private onHandleKeyDown = (ev: KeyboardEvent) => {
+    if (ev.key === 'Enter') {
+      ev.preventDefault();
+    }
+  };
```
liamdebeasi, line 751: What does this do? Typically pressing Enter/Space also dispatches a click event, so `onHandleClick` should be sufficient.

core/src/components/modal/modal.tsx
```diff
@@ -786,7 +817,17 @@ export class Modal implements ComponentInterface, OverlayInterface {
         {mode === 'ios' && <div class="modal-shadow"></div>}
 
         <div role="dialog" class="modal-wrapper ion-overlay-wrapper" part="content" ref={(el) => (this.wrapperEl = el)}>
-          {showHandle && <div class="modal-handle" part="handle"></div>}
+          {showHandle && (
+            <button
+              class="modal-handle"
+              part="handle"
+              aria-label="Activate to adjust the size of the card overlaying the dialog"
```
liamdebeasi, line 824: What does "card overlaying the dialog" mean? The modal itself is a dialog.

core/src/components/modal/modal.tsx
sean-perkins: Waiting for the transition feels like it would conflict with the design document for the API of moving to a breakpoint:
> If the user is in the middle of a gesture when this API is called, the user will be transitioned to the breakpoint value called by the API.

Unless we can cancel the current transition and immediately kick-off the new transition without a glitchy experience. I can do some more discovery on this.

core/src/components/modal/modal.tsx
sean-perkins: Yeah this should probably read:
> Activate to adjust the size of the card overlaying the screen.

It was adapted from the maps example of:
> Card controller half screen, adjust the size of the card overlaying the map.

Thoughts?

liamdebeasi: Yeah I like that

core/src/components/modal/modal.tsx
sean-perkins: Updated the behavior, curious to both of your thoughts/feedback/alternatives. It now won't skip/jitter when you hold the enter key, but it will only performs moving to the next breakpoint when the current animation finishes. Hard to demonstrate in a recording.

amandaesmith3: So when I try it, pressing Enter twice rapidly makes it only move once. It seems like it'll just eat your input if the animation is currently running. My guess is that since the current breakpoint doesn't actually update until after the animation finishes (triggered [here](https://github.com/ionic-team/ionic-framework/blob/main/core/src/components/modal/gestures/sheet.ts#L215)), `nextBreakpoint` is calculated as the same thing for both Enter presses. Which would make the second press not really do anything.

That said, I'm okay with the behavior as-is. Up to you if you want to change it, though if we stick with it as-is, the code should probably be cleaned up to make it more clear.

sean-perkins:
> Nice you tackled this 😊 Maybe it would be good to add a property to be able to disable this behaviour? There are apps that have this and others that don't have this…

@EinfachHans oddly enough just had this exact conversation with the team 😆  we are going to hold on this PR until after 6.1.0 to write up a feature design document that covers the API additions and user behavior. Originally we anticipated this as a bug fix, but after comparing against other app patterns, we noticed that:
1. Showing/hiding the handle is configurable in Swift/iOS
2. Some apps support clicking the handle, others do not

While this is a nice user enhancement, we want to make sure we get it right and offer the best configurations for developers to make use of it for their apps.

Abuturab1234:
Looking back at the code, present and dismiss will both wait for the[ current transition](<a href="https://www.royalfalconevents.com/">text </a>) and then go through with the attempted interaction, so we should probably go with that instead here. Thinking on it, if I hit Enter twice quickly, having it only move to the next breakpoint once would feel off.

sean-perkins:
Closing in favor of: https://github.com/ionic-team/ionic-framework/pull/25540

PR #25071, opened 2022-04-07
fix(angular): button components now routes correctly without reload

liamdebeasi:
## Pull request checklist

## Pull request type

 

 

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

Issue URL: https://github.com/ionic-team/ionic-framework/pull/25014#issuecomment-1091056750

#25014 caused all routerLinks to perform a page reload which interfered with the local history of Ionic's routing integrations. This resulted in `ion-back-button` elements disappear because Ionic thought users could not go back.

This change was originally done to allow `a` elements to open links in new tabs via Ctrl+Click.

## What is the new behavior?

- Reverts the change in #25014
- Excludes `a` elements from Ionic's routerLink delegate so that Ctrl+Click works correctly.

- [ ] Yes
- [x] No

### DIFF
--- a/.github/workflows/actions/build-core/action.yml
+++ b/.github/workflows/actions/build-core/action.yml
@@ -4,6 +4,9 @@ runs:
   using: 'composite'
   steps:
     - uses: actions/checkout@v2
+      with:
+        # Checkout the latest commit in this branch
+        ref: ${{ github.event.pull_request.head.sha }}
     - uses: actions/setup-node@v1
       with:
         node-version: 15.x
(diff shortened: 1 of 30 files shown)


sean-perkins (commented):
Angular does have a router link implementation that targets specifically the `a` and `area` tags: https://github.com/angular/angular/blob/master/packages/router/src/directives/router_link.ts#L280-L281

Should we do a similar approach? Have the previous implementation pattern target `a` and have the updated implementation target the selector within this PR?

I believe with this change clicking the `a` tag without a new tab interaction, will not apply the element specific overrides for router direction & animation.

sean-perkins (approved):

PR #25077, opened 2022-04-07
chore(): expand playwright page viewport by mobile offset

sean-perkins:
## Pull request checklist

## Pull request type

 

 

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes
- [x] Other (please describe): Playwright screenshot

Playwright's config for device emulation has different values for the screen size vs. the viewport size. When attempting to capture fullscreen screenshots for a test, the screenshot image will be cropped by the offset of these two values. 

For example: https://github.com/microsoft/playwright/blob/main/packages/playwright-core/src/server/deviceDescriptorsSource.json#L578-L585

Creates cropped screenshots such as: https://github.com/ionic-team/ionic-framework/blob/65b8639db56cec96fc62c7767700d046013f9224/core/src/components/datetime/test/presentation/datetime.e2e.ts-snapshots/datetime-presentation-diff-ios-ltr-Mobile-Chrome-linux.png

## What is the new behavior?

Using internal options, we can access the device information for the current test. When expanding the viewport to account for the `ion-content` scroll container, we can additionally expand the viewport by the offset of the screen size vs. the viewport size for the mobile device.

The end result is that screenshot images are appropriate sized when capturing.

- [ ] Yes
- [x] No

## Other information

Will run the action to update reference screenshots, so we can see the difference in output.


sean-perkins:
Most images are corrected with this change, mobile linux is still a little strange. We could either buffer by a magic number to account for it or roll with this for a while and continue to evaluate.

liamdebeasi (approved):
Great job! I think it is more important to capture all of the testable area than to have the screenshots be perfectly cropped, so I am fine with merging this. In the event we find a better way to approach this problem we can always re-run the update screenshot job.

amandaesmith3 (changes requested):
The mobile screenshots for `datetime-presentation` are missing the year picker; maybe they just need updating after the fix you added to wait for that selector?

sean-perkins:
Yeah that is the problem I was referring to in stand-up. I think I may need to resize the viewport after the content has loaded, instead of right now we resize when the page is navigated to & the basic Stencil hydration completes. Datetime has some extra behavior that affects the bounding box after initial render.

sean-perkins:
Holding on this solution for a minute to focus on the parent branch (datetime value). This change may be unnecessary by adjusting the test output that screenshots run on. Datetime specific complexities with the intersection observer for adding the `.datetime-ready` class, appear to drive a lot of the problem here. 

By adjusting the test to show 1 presentation at a time, resolves that issue, as the datetime is always intersecting and will render correctly without changing the viewport.

There's another weird (not reproducible on device), mobile safari/playwright issue with datetime that I need to solve first.

sean-perkins:
Closing this PR. I don't think it is necessary at the moment. Will leave branch around for a little while for reference, in case we run into a similar situation with screenshots on another component.

Issue #24456, opened 2021-12-21
bug: Ionic 6 - Ion-Datetime UI should not be solid blue for todays date when no value set


bkarv:
### Prerequisites

- [X] I have read the [Contributing Guidelines](https://github.com/ionic-team/ionic-framework/blob/main/.github/CONTRIBUTING.md#creating-an-issue).
- [X] I agree to follow the [Code of Conduct](https://ionicframework.com/code-of-conduct).
- [X] I have searched for [existing issues](https://github.com/ionic-team/ionic-framework/issues) that already report this problem, without success.

### Ionic Framework Version

- [ ] v4.x
- [ ] v5.x
- [X] v6.x

### Current Behavior

On Ionic 6 and to be honest its a bit harsh to call it a bug. But in short when in popover mode when the user clicks on the text or input date field to open up the new ion-datetime UI. If no value is selected, the date defaults to today's date. This is fine however the circle is blue solid which visually means the user has selected the date already which is incorrect. This could cause confusion. 

<img width="241" alt="Screen Shot 2021-12-21 at 5 33 13 pm" src="https://user-images.githubusercontent.com/10624315/146883652-d032716d-9fd3-4a43-baee-f940ab16c055.png">

### Expected Behavior

If no value set, the default today's date should be a blue outline circle not a solid circle.  See below correct expected behaviour.

<img width="193" alt="Screen Shot 2021-12-21 at 5 33 00 pm" src="https://user-images.githubusercontent.com/10624315/146883334-12f96494-dc4f-4df0-9191-ac93e1f7948f.png">

### Steps to Reproduce

Test new ionic datetime using a ngModel value set to "" or null. I tested in popover mode in Angular 13:

```
<ion-item class="ion-text-wrap" button="true" id="ionic-date-start">
                        <ion-icon name="calendar" slot="start"></ion-icon>
                         <ion-label style="color:#424242">Start Date</ion-label>
                         <ion-text>{{record.start_date|date:'mediumDate'}}</ion-text>

                        <ion-popover trigger="ionic-date-start" show-backdrop="false">
                          <ng-template>
                            <ion-datetime
                              #popoverDatetime
                              presentation="date"
                              [showClearButton]="true"
                             [(ngModel)]="record.start_date"
                            ></ion-datetime>
                          </ng-template>
                        </ion-popover>
                      </ion-item>
```

### Ionic Info

Ionic:

[...]

PR #25151, opened 2022-04-19
fix(datetime): if no default value, don't highlight active day until one is selected

amandaejohnston:
## Pull request checklist

## Pull request type

 

 

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

When an `ion-datetime` is initialized without a value defined, today's date is highlighted with a solid blue circle. This is misleading, as it looks like a day has already been selected.

Issue URL: https://github.com/ionic-team/ionic-framework/issues/24456

## What is the new behavior?

- For datetimes with no initial value, no solid blue circle is rendered at first. Today's date is highlighted with an outline only. Once any day is clicked, the solid circle returns.
- For datetimes with an initial value defined, behavior is unchanged.
- The solid blue circle does _not_ necessarily indicate which day the actual `value` property is set to at that moment. (I.e. it's possible to trigger the solid circle, but still have a nullish `value`.) This is to handle datetimes with confirm/cancel buttons, where the `value` does not change until `confirm` is called. (See: https://ionicframework.com/docs/api/datetime#confirming-selected-values)

- [ ] Yes
- [x] No

## Other information

- We decided not to enforce a non-nullish `value` in order to avoid unwanted `ionChange` events that would fire when the value is automatically switched to today.
- An alternative would be to not default the `activeParts` to today if no initial value is set, but that would require allowing `activeParts` or all its properties to be `null`, which would require a small refactor. I figure the current method is simpler and safer.
- Dev build: `6.1.2-dev.11650392645.12c652f6` (Posted in original issue)


core/src/components/datetime/test/basic/e2e.ts
```diff
@@ -98,6 +98,41 @@ describe('datetime: selecting a day', () => {
 
     expect(newActiveDay.innerText).toEqual('13');
   });
+
+  const testHighlight = async (page) => {
```
sean-perkins, line 102: Any blockers that prevented writing this test in Playwright?

amandaesmith3: ... I just forgot :joy: I'll do that, haha.

core/src/components/datetime/test/basic/datetime.e2e.ts
```diff
@@ -0,0 +1,32 @@
...
+    );
+
+    const classes = await todayBtn.evaluate((el: any) => el.classList.value);
+    expect(classes).toMatch('calendar-day-today');
+    expect(classes).not.toMatch('calendar-day-active');
+
+    await todayBtn.click();
+    await page.waitForChanges();
+
+    const classes2 = await todayBtn.evaluate((el: any) => el.classList.value);
```
liamdebeasi, line 21: You should be able to use the `toHaveClass` method instead: https://playwright.dev/docs/test-assertions#locator-assertions-to-have-class

liamdebeasi: Same thing on line 14

core/src/components/datetime/datetime.tsx
```diff
@@ -1424,7 +1427,7 @@ export class Datetime implements ComponentInterface {
                 class={{
                   'calendar-day-padding': day === null,
                   'calendar-day': true,
-                  'calendar-day-active': isActive,
+                  'calendar-day-active': isActive && highlightActiveParts,
```
liamdebeasi, line 1430: Can we add a comment for why this is needed?

liamdebeasi (approved):
Great job!

sean-perkins (approved):

Issue #20127, opened 2019-12-29
bug: overriding aria attributes is not easy or straightforward


BingSZou:
# Bug Report

**Ionic version:**
[x] **4.x**

**Current behavior:**
Most Accessibility attributes, e.g. Aria-Label, Aria-LabeledBy, Role set on the Ion-Button does not work

**Expected behavior:**
The Accessibility attributes should work on Ion-Button

**Steps to reproduce:**
Create a simple Ion-Button with following attributes: 
```
<ion-button role="link" aria-label="this is my data">
    View my data </ion-button>
```
This sample code should be read out as:  "this is my data link",  but instead it reads out "View my data button"

**Other information:**
Root Cause:  Ion-Button has a shadow root: "button" . The attributes are not copied over to the button element.  So the aria attributes not working.

Note: A workaround is to create a directive with MutationObserver to copy the aria attributes from Ion-Button to its shadow-root button.

VincentCrete:
@BingSZou, can-you show us an example of your workaround? We have the same problem.

pschwindling:
Same problem here, which really is a big issue in our app now.

VittorioEuris:
On Ionic@5.6.7 I see `ion-button` with `slot="icon-only"` apply the right aria-label but this is not updated on changes.
In details I'm using ngx-translate to provide i18n to my app but when user changes language dinamically inside the app every `ion-button` already instantiated doesn't update their value. This is a real problem because of buttons `slot="icon-only"` inside the header.

sean-perkins:
Hello everyone 👋  with #25156, now all non-deprecated `aria-*` and `role` attributes are reflected to the element created within light DOM.

This change applies to:
- `ion-button`
- `ion-back-button`
- `ion-breadcrumb`
- `ion-header`
- `ion-input`
- `ion-textarea` 
- `ion-menu`
- `ion-menu-button`

Content this is slotted is already in the light DOM, so you should be able to already apply aria attributes to those elements directly.

If there is another component that needs this behavior, please let us know in a new issue. Thanks!

PR #25156, opened 2022-04-20
fix(core): inherit aria attributes on host elements

sean-perkins:
## Pull request checklist

## Pull request type

 

 

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

ARIA attributes specified on the host element are not reflected to the cloned element in the light DOM. For example, with `ion-button` we create a `button` and clone it into the light DOM so that screen readers can correctly interact with the element. Currently, we are only supporting `aria-label` to be reflected to the HTML `button`.

Issue URL: Enterprise FW-1138, Enterprise FW-1071, #20127

## What is the new behavior?

All WAI-ARIA attributes, as listed [here](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes) are reflected to the following components:

- `ion-button`
- `ion-back-button`
- `ion-breadcrumb`
- `ion-header`
- `ion-input`
- `ion-textarea` 
- `ion-menu`
- `ion-menu-button`

- [ ] Yes
- [x] No

## Other information

I opted for a manually managed list of aria attributes over regular expression matching, after taking into consideration the execution cost of regex for each web component instance + the number of attributes/properties on an element. 

This approach should be easily applicable to other open issues, to add support for aria attributes.

### DIFF
--- a/core/src/utils/helpers.ts
+++ b/core/src/utils/helpers.ts
@@ -103,6 +103,11 @@ export const inheritAttributes = (el: HTMLElement, attributes: string[] = []) =>
   return attributeObject;
 };
 
+/**
+ * List of available ARIA attributes + `role`.
+ * Removed deprecated attributes.
+ * https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes
+ */
 const ariaAttributes = [
   'role',
   'aria-activedescendant',
@@ -122,13 +127,9 @@ const ariaAttributes = [
   'aria-description',
   'aria-details',
   'aria-disabled',
-  // @deprecated
-  'aria-dropeffect',
   'aria-errormessage',
   'aria-expanded',
   'aria-flowto',
-  // @deprecated
-  'aria-grabbed',
   'aria-haspopup',
   'aria-hidden',
   'aria-invalid',
--- a/core/src/utils/helpers.ts
+++ b/core/src/utils/helpers.ts
@@ -103,6 +103,73 @@ export const inheritAttributes = (el: HTMLElement, attributes: string[] = []) =>
   return attributeObject;
 };
 
+const ariaAttributes = [
+  'role',
+  'aria-activedescendant',
+  'aria-atomic',
+  'aria-autocomplete',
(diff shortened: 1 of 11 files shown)


core/src/utils/helpers.ts
```diff
@@ -103,6 +103,73 @@ export const inheritAttributes = (el: HTMLElement, attributes: string[] = []) =>
...
+  'aria-colindextext',
+  'aria-colspan',
+  'aria-controls',
+  'aria-current',
+  'aria-describedby',
+  'aria-description',
+  'aria-details',
+  'aria-disabled',
+  // @deprecated
+  'aria-dropeffect',
```
liamdebeasi, line 126: Optional: I would prefer not to support deprecated attributes. I see it as us promoting anti-patterns in apps

core/src/utils/helpers.ts
```diff
@@ -103,6 +103,73 @@ export const inheritAttributes = (el: HTMLElement, attributes: string[] = []) =>
   return attributeObject;
 };
 
+const ariaAttributes = [
```
liamdebeasi, line 106: Is there a list where you got all of these from? It might be good to include the source link here if we ever need to update the list.

core/src/utils/helpers.ts
```diff
@@ -103,6 +103,73 @@ export const inheritAttributes = (el: HTMLElement, attributes: string[] = []) =>
...
+  'aria-valuetext',
+];
+
+/**
+ * Returns an array of aria attributes that should be copied from
+ * the shadow host element to a target within the light DOM.
+ * @param el The element that the attributes should be copied from.
+ */
+export const inheritAriaAttributes = (el: HTMLElement) => {
+  return inheritAttributes(el, ariaAttributes);
```
liamdebeasi, line 170: Do we need a test for this? Or do you think this is not needed because most of the implementation is already tested in `inheritAttributes`?

liamdebeasi (approved):

liamdebeasi (approved):

amandaejohnston (approved):

Issue #25126, opened 2022-04-14
bug: ion-select showing multiple alert windows stacked on top of each other


zzjrzz:
### Prerequisites

- [X] I have read the [Contributing Guidelines](https://github.com/ionic-team/ionic-framework/blob/main/.github/CONTRIBUTING.md#creating-an-issue).
- [X] I agree to follow the [Code of Conduct](https://ionicframework.com/code-of-conduct).
- [X] I have searched for [existing issues](https://github.com/ionic-team/ionic-framework/issues) that already report this problem, without success.

### Ionic Framework Version

- [ ] v4.x
- [ ] v5.x
- [X] v6.x

### Current Behavior

Immediately after loading a page, click on `ion-select` component in rapid successions. (spam click)
This will create multiple alert select windows. 

Video of behavior: 
<video src="https://user-images.githubusercontent.com/13732623/164767757-1dabadf3-b3ee-4a3e-bd4a-3cd4769a2f42.mp4"></video>

### Expected Behavior

I expect only one alert select window to appear.
If multiple are presented at once, the user is confused as they have to dismiss multiple windows.

### Steps to Reproduce

Clear all browser caches - load an application for the first time
IMMEDIATELY spam click the component
Happens on: https://ionicframework.com/docs/api/select

### Ionic Info

                          
[WARN] Error loading @capacitor/ios package.json: Error: Cannot find module
       '@capacitor/ios/package'
       
       Require stack:
       -
       C:\Users\AppData\Roaming\nvm\v16.13.2\node_modules\@ionic\cli\lib\projec
t\index.js
       -
       C:\Users\AppData\Roaming\nvm\v16.13.2\node_modules\@ionic\cli\lib\index.
js
       - C:\Users\AppData\Roaming\nvm\v16.13.2\node_modules\@ionic\cli\index.js
       -
       C:\Users\AppData\Roaming\nvm\v16.13.2\node_modules\@ionic\cli\bin\ionic 
[WARN] Error loading @capacitor/android package.json: Error: Cannot find module
       '@capacitor/android/package'

       Require stack:
       -
       C:\Users\AppData\Roaming\nvm\v16.13.2\node_modules\@ionic\cli\lib\projec
t\index.js
       -
       C:\Users\AppData\Roaming\nvm\v16.13.2\node_modules\@ionic\cli\lib\index.
js
       - C:\Users\AppData\Roaming\nvm\v16.13.2\node_modules\@ionic\cli\index.js
       -
[...]

PR #25175, opened 2022-04-22
fix(select): avoid duplicate dialogs and backdrops when clicking

sean-perkins:
## Pull request checklist

## Pull request type

 

 

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

Rapidly clicking an `ion-select` can produce multiple `ion-alert` and backdrops. This is because the internal state flag is only set after the overlay is created (before presenting). You can produce this reliably by manually calling `.click()` multiple times on the same element reference.

Issue URL: #25126

## What is the new behavior?

Sets the flag prior to the overlay being created, this prevents multiple alert/backdrops from being presented.

- [ ] Yes
- [x] No

## Other information

Dev-build: `6.1.3-dev.11650650446.100eff34` 🟨

### DIFF
--- a/core/src/components/select/select.tsx
+++ b/core/src/components/select/select.tsx
@@ -188,8 +188,8 @@ export class Select implements ComponentInterface {
     if (this.disabled || this.isExpanded) {
       return undefined;
     }
-    const overlay = (this.overlay = await this.createOverlay(event));
     this.isExpanded = true;
+    const overlay = (this.overlay = await this.createOverlay(event));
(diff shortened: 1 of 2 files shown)


liamdebeasi (approved):
Prior to merging let's also update the commit subject. Select doesn't only present alert dialogs, so the current message is confusing.

Maybe something like:

`fix(select): avoid duplicate dialogs when clicking`

amandaejohnston (approved):

Issue #25073, opened 2022-04-07
bug: `ion-datetime` selecting a day past the max day value will disable arrow navigation on previous month


alexooi:
### Prerequisites

- [X] I have read the [Contributing Guidelines](https://github.com/ionic-team/ionic-framework/blob/main/.github/CONTRIBUTING.md#creating-an-issue).
- [X] I agree to follow the [Code of Conduct](https://ionicframework.com/code-of-conduct).
- [X] I have searched for [existing issues](https://github.com/ionic-team/ionic-framework/issues) that already report this problem, without success.

### Ionic Framework Version

- [ ] v4.x
- [ ] v5.x
- [X] v6.x

### Current Behavior

In ion-datetime with min and max values set, when a date other than the default date value is selected, the navigation arrow is incorrectly disabled. 

### Expected Behavior

In ion-datetime with min and max values set, when a date other than the default date value is selected, the navigation arrow is not disabled. 

### Steps to Reproduce

https://user-images.githubusercontent.com/9407445/162442230-7430aaf1-a81c-45a3-944c-5514107b63e2.mp4

### Ionic Info

Ionic:

   Ionic CLI                     : 6.16.3 (C:\Users\alexo\AppData\Roaming\npm\node_modules\@ionic\cli)
   Ionic Framework               : @ionic/angular 6.0.14
   @angular-devkit/build-angular : 12.2.16
   @angular-devkit/schematics    : 12.1.1
   @angular/cli                  : 12.1.4
   @ionic/angular-toolkit        : 4.0.0

Capacitor:

   Capacitor CLI      : 2.4.6
   @capacitor/android : 2.5.0
   @capacitor/core    : 2.4.6
   @capacitor/ios     : 2.5.0

Utility:

   cordova-res : 0.15.3
   native-run  : 1.3.0

System:

   NodeJS : v14.15.4 (C:\Program Files\nodejs\node.exe)
   npm    : 6.14.13
   OS     : Windows 10

hardmaster92:
I confirm to have the same problem and it's really weird how and when happens. It appears to me that happens when you bind properties for `min`, `max` and `value` instead of directly passing string attributes.
Here's a stackblitz project where I notice the issue
https://stackblitz.com/edit/ionic6-angular13-hn7ytp

sean-perkins:
@hardmaster92 thanks for the added detail & the reproduction! That is very strange behavior, but reproducible. I'm going to track this specific to Angular for now.

[...]

PR #25182, opened 2022-04-24
fix(datetime): arrow navigation respects min/max values

sean-perkins:
## Pull request checklist

## Pull request type

 

 

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

The value of the `ion-datetime`, when used with `min` or `max`, can incorrectly disable the next/previous navigation. When selecting a date such as "March 20th, 2022", but having a max date of "April 19th, 2022"; the existing comparison logic would disable the next button, since the 1 month offset was April 20th, which was beyond the max value. The same behavior happened for min values.

Issue URL: #25073

## What is the new behavior?

When calculating if the next or previous button should be disabled, it will ignore comparison on the day of the month and only compares if the month & year are within bounds (for next/previous navigation).

- [ ] Yes
- [x] No

## Other information

Dev-build: `6.1.3-dev.11650654723.120c74ee` ✅

Reproduction repo: https://github.com/sean-perkins/ion-datetime-min-max-25073 (uses dev build).


liamdebeasi (changes requested):
Can you write some tests?

core/src/utils/test/playwright/page/utils/set-content.ts
```diff
@@ -0,0 +1,63 @@
...
+ * @param html The HTML content to set on the page.
+ */
+export const setContent = async (page: Page, html: string) => {
+  if (page.isClosed()) {
+    throw new Error('setContent unavailable: page is already closed');
+  }
+
+  if (typeof html !== 'string') {
+    throw new Error('invalid html: must be a string');
+  }
```
amandaejohnston, line 19: Is this needed? TS should catch it for us.

sean-perkins: You're right 👍  I copied from the Stencil Puppeteer implementation, but I don't see any reason/value of this check.

amandaejohnston (approved):

liamdebeasi (approved):

Issue #24998, opened 2022-03-25
bug: datetime, month/year selection with cancel/done buttons has confusing UX


vladimirSnapAddy:
### Prerequisites

- [X] I have read the [Contributing Guidelines](https://github.com/ionic-team/ionic-framework/blob/main/.github/CONTRIBUTING.md#creating-an-issue).
- [X] I agree to follow the [Code of Conduct](https://ionicframework.com/code-of-conduct).
- [X] I have searched for [existing issues](https://github.com/ionic-team/ionic-framework/issues) that already include this feature request, without success.

### Describe the Feature Request

Event should be emitted when user changes view from default to month/year selection opened when clicked on top left button in the datepicker and back to default view.

### Describe the Use Case

I am showing ion-datetime in a modal. When I want to select a date in one of the past years, I must click on the month/date button in top left corner. If I then select wanted month and year, and click on done, modal closes and value is set to previously select date. This behaviour is frustrating, because many users don't expect modal to be closed with previously selected value.

### Describe Preferred Solution

If an event is emitted when these views are toggled, we can handle custom behaviour (show and hide DONE button for example) and prevent users from exiting modal unintentionally. 

### Related Code

```
<ion-button id="open-modal">Open Datetime Modal</ion-button>
<ion-modal trigger="open-modal" [backdropDismiss]="false">
  <ng-template>
    <ion-content>
      <ion-datetime
        size="cover"
        [disabled]="false"
        [max]="'2050-12-31'"
        [(ngModel)]="date"
        (ionChange)="updateValue($event)"
        [showDefaultButtons]="true"
      >
      </ion-datetime>
    </ion-content>
  </ng-template>
</ion-modal>
```

### Additional Information

https://user-images.githubusercontent.com/85163599/160097525-deb50edf-1c29-4f08-9c72-6fe5d7cc979a.mov

sean-perkins:
@vladimirSnapAddy thanks for the feature request!

I think there is a few different areas to dig into here:

1. The "Done" action is confusing to users when changing month/year:

This appears to be reoccurring feedback we are receiving. While the top-left month/year selector does show an arrow indicator that it is expanded, it is unclear to users that the "Done" action is attached to the date picker and not to the expanded view.

[...]

Issue #24962, opened 2022-03-17
bug: datetime confirm method ignores selected month and year when selection dropdown is still open


NilsFrkal:
**Ionic version:**
[] **4.x**
[ ] **5.x**
[ x] **6.x**

[ x] bug report
[ ] feature request

**Current behavior:**

When using the date-time component (presentation="date") the selected year and month gets ignored when passed as element value. Example:

current date: 18/03/2022
selection:
   day: 15
   month: December
   year: 1980

console.log(dateinput.value): 15/03/2022

This behavior can be observed with all versions > @ionic/angular@6.0.1 i.e. 6.0.1 still works, all later versions up to 6.0.12 do not.  

**Expected behavior:**

Selected year and month should be considered i.e. expected output for example above 15/12/1980

**Steps to reproduce:**

1. Update to version > 6.0.1 e.g. latest version @ionic/angular@latest
2. Create date-time component   
3. Select a new date with month and year and pass in ionChange event via element value (see code snippet below)
4. Console.log value

To check working version:

1. npm install @ionic/angular@6.0.1
2. Create date-time component   
3. Select a new date with month and year and pass in ionChange event via element value (see code snippet below)
4. Console.log value

**Related code:**

 <ion-datetime id="XXX" [value]="date" presentation="date"
        (ionChange)="onSelectDate(dateinput.value)" showDefaultButtons="true"
        (ionCancel)="onDatePickerCancel()" #dateinput>
      </ion-datetime>

  onSelectDate (value) {
    console.log(value);
  }

**Ionic info:** 

Ionic:

   Ionic CLI                     : 6.19.0 (D:\Development\Nodist\bin\node_modules\@ionic\cli)  
   Ionic Framework               : @ionic/angular 6.0.12
   @angular-devkit/build-angular : 13.3.0
   @angular-devkit/schematics    : 13.3.0
   @angular/cli                  : 13.3.0
   @ionic/angular-toolkit        : 2.3.3

Capacitor:

   Capacitor CLI      : 3.4.3
   @capacitor/android : 3.4.0
   @capacitor/core    : 3.4.3
   @capacitor/ios     : 3.4.3

Utility:

   cordova-res : 0.15.4
   native-run  : 1.5.0

System:

   NodeJS : v16.13.2 (D:\Development\Nodist\v-x64\16.13.2\node.exe)
   npm    : 8.1.2
   OS     : Windows 10

liamdebeasi:
[...]

PR #25205, opened 2022-04-27
fix(datetime): hide footer when month-year picker is open

amandaejohnston:
## Pull request checklist

## Pull request type

 

 

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

When the month-year dropdown/picker is open, the cancel/done buttons look like they apply to the picker, rather than the whole datetime. This leads to confusing UI where the user might click Done to select a new month/year, but it closes the dialog instead and discards their choice.

Issue URL: https://github.com/ionic-team/ionic-framework/issues/24998, https://github.com/ionic-team/ionic-framework/issues/24962

## What is the new behavior?

When the month/year picker is open, the datetime's footer is hidden. The user must collapse the picker to access the buttons again. They can still cancel their choice by clicking outside the overlay or pressing Escape.

Behavior is unchanged for the `month`, `year`, and `month-year` presentations; the footer will remain visible.

- [ ] Yes
- [x] No

## Other information

Native behavior is inconsistent across platforms:
- MUI has the OK button apply to the whole dialog, as in Ionic's old behavior: https://material-ui-pickers.dev/demo/datepicker
- iOS doesn't typically use overlays with buttons, instead showing the picker inline: https://developer.apple.com/design/human-interface-guidelines/ios/controls/pickers/#date-pickers
- Windows Chrome's web implementation hides buttons when the picker is open, as in this PR's new behavior: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/datetime-local
- iOS's web implementation has the OK button apply to the whole dialog, but keeps the user's updated month/year selection (see above URL)

We decided to go with this PR's behavior across platforms for better consistency and ease of maintenance. It clarifies the UI without adding extra layers to the component's behavior, i.e. changing what the `confirm` method does depending on the datetime's state.

Side note: Technically, `monthYearPickerOpen` is redundant since `showMonthAndYear` would suffice, but I found this easier to reason about. I can tweak things if preferred.


sean-perkins (approved):
Behavior and changes look good to me. 

Question around accessibility. Now that the actions are no longer available, should we add secondary behavior to help screen reader users get back to the previous datetime state or do we expect them to traverse backwards to the expander, to collapse the month/year selection? UX that I am thinking of is `ESC` key to collapse the section, but also know we use that interaction within overlays, so that may not be ideal.

liamdebeasi (approved):
These changes look good. Re: Sean's accessibility question, we could change the month/year item in the top left (the button that says "April 2022") to read "hide year picker" when it is expanded and "show year picker" when it is collapsed. This is what iOS w/ VoiceOver does.

amandaejohnston:
I played around with it some, and while I was able to get it reading "show year picker"/"hide year picker" after the month/year, the usability still isn't great 🤔 I'm going to merge this PR as-is and open a separate issue in a few, since there might be some deeper design discussions around it.

CarstenRuetz:
First of all, thanks for the great work!
@amandaejohnston is there any news around the month/year picker part? 
During user tests, we noticed that for choosing their birthdate, they just don't understand that to close the month/year dropdown, they have to click on the month/year label again...

SaroGFX:
Our users also reported that they would like a Done button on the Month / Year selector. Since the buttons are removed, it created confusion. From an UX point of view I believe what you want is to simply have `Done` to register all the changes. It can either go back to the day calendar, but preferably just registers the selected date, month and year and closes the picker.

PR #25222, opened 2022-05-02
test(playwright): do not check desktop viewports by default

liamdebeasi:
## Pull request checklist

## Pull request type

 

 

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes

Many of the tests in Ionic are exactly the same across mobile/desktop, so explicitly checking desktop on every test is not necessary.

## What is the new behavior?

- Desktop chrome and Desktop safari tests have been removed. (including associated screenshots)
- A special tablet viewport object has been added to make creating tablet-specific tests easier to write.
- Updated certain modal test to make sure we are checking tablet viewports as well.
- I noticed that Playwright's scroll into view behavior for the basic datetime tests was a bit flaky for Firefox, so I added a `setIonViewport`

- [ ] Yes
- [x] No


core/src/components/modal/test/basic/modal.e2e.ts
```diff
@@ -42,6 +42,7 @@ test.describe('modal: focus trapping', () => {
     page,
     browserName,
   }) => {
+    test.skip(browserName === 'firefox', 'Focus is flaky on Firefox');
```
amandaejohnston, line 45: Does this need a tech debt ticket?

core/playwright.config.ts
```diff
@@ -7,22 +7,27 @@ expect.extend(matchers);
 
 const projects = [
   {
-    name: 'chromium',
```
amandaejohnston, line 10: Could you add a note to the internal Notion doc about the change? Might be good to also include a code snippet for adding desktop tests.

core/src/components/modal/test/card/modal.e2e.ts
```diff
@@ -105,4 +105,112 @@ test.describe('card modal', () => {
       expect(content).toHaveJSProperty('scrollY', true);
     });
   });
+  test.describe('card modal: rendering - tablet', () => {
```
amandaejohnston, line 108: Would it be possible to bring these tests into a helper method so we don't have so much repeated code? Ditto for the other tests where tablet versions have been added, but especially this one since it has so many `test` calls and may have more added later.

amandaejohnston (changes requested):

core/playwright.config.ts
liamdebeasi: Added to the "Writing Tests in Core" document under "Tablet Viewport Tests"

core/src/components/modal/test/card/modal.e2e.ts
liamdebeasi: Sure. What parts are you thinking we should move to helper functions?

core/src/utils/test/playwright/viewports/index.ts
```diff
@@ -0,0 +1,4 @@
+export const TabletViewport = {
```
sean-perkins, line 1: This is probably fine for now, but we may want to consider making `Viewports` a constant and differentiating between portrait/landscape for the viewport values; if we find ourselves needing to capture screenshots on common viewports.

i.e.:
```ts
export cont Viewport = {
  tablet: {
    portrait: {
      width: 768,
      height: 900
    }
  }
}
```

sean-perkins (approved):

core/src/utils/test/playwright/viewports/index.ts
liamdebeasi: Good thought. I made that change.

core/src/components/modal/test/card/modal.e2e.ts
liamdebeasi: For `basic/modal.e2e.ts` and `custom/modal.e2e.ts` I created a function that both the mobile and tablet visual tests execute in the body of the `test` callback.

For `card/modal.e2e.ts` I ended up creating a Page Object Model (https://playwright.dev/docs/test-pom) to remove a good chunk of boilerplate.

There's still some duplication, but it's not as excessive as before.

amandaejohnston: Muuuuch better, thank you!

amandaejohnston (approved):

PR #25231, opened 2022-05-03
perf(menu): avoid forced layout when opening menu

liamdebeasi:
## Pull request checklist

## Pull request type

 

 

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

When debugging an Ionic app, I noticed a delay in the opening of a menu on Android. Upon further investigation, I found that we are forcing a Style recalculation followed by a Layout Invalidation: https://github.com/ionic-team/ionic-framework/blob/main/core/src/components/menu/menu.tsx#L412

The `beforeAnimation` method writes classes to the menu element, which will result in a rerender. However, we are requesting `offsetWidth` immediately after that in `loadAnimation`. This means we are asking the browser to re-render and give us the new width in the same frame. This results in a long frame as shown below:

image

A few things to note:

- The Animation frame took 71.12ms. A frame that runs at 60fps should take no more than ~16.67ms.
- The recalculate style and layout bars take up most of this 71.12ms

## What is the new behavior?

- By wrapping the `offsetWidth` in a `requestAnimationFrame` and waiting for that to resolve, we can allow the browser to better schedule these calculations. We still need the width, but since the browser is able to split up the `beforeAnimation` and `loadAnimation` work in separate frames, we can avoid the long frame.
See the new results below:

image

- The frame now takes ~15.07ms 
- There is still a style recalculation, but since it is optimized better it takes significantly less than what it used to take (that anonymous function is the raf in this PR)

- [ ] Yes
- [x] No


liamdebeasi:
Need to investigate the tests a bit more. will re-open when ready.

Issue #25180, opened 2022-04-24
bug: circular, lines-sharp, and lines-sharp-small not getting --spinner-color applied in ion-loading


dexster:
### Prerequisites

- [X] I have read the [Contributing Guidelines](https://github.com/ionic-team/ionic-framework/blob/main/.github/CONTRIBUTING.md#creating-an-issue).
- [X] I agree to follow the [Code of Conduct](https://ionicframework.com/code-of-conduct).
- [X] I have searched for [existing issues](https://github.com/ionic-team/ionic-framework/issues) that already report this problem, without success.

### Ionic Framework Version

- [ ] v4.x
- [ ] v5.x
- [X] v6.x
- [ ] Nightly

### Current Behavior

Specifying the --spinner-color has no effect on a 'circular' spinner. All other spinner shapes work as expected.

### Expected Behavior

Spinner color should respect defined custom color

### Steps to Reproduce

In an Angular 13 project add a spinner.

I am running Angular 13.3 but if the issue is due to the naming I mentioned in additional Info below then I expect this issue is not framework specific 

```
const loader = await this.loadingController.create({
          spinner: 'circular',
          cssClass: 'my-spinner'
        });

```
```
// CSS
.my-spinner {
  --spinner-color: red;
}
```

### Ionic Info

Ionic:

   Ionic CLI : 6.18.2

Utility:

   cordova-res : not installed globally
   native-run  : not installed globally

System:

   NodeJS : v16.14.0
   npm    : 8.3.1
   OS     : macOS Monterey

### Additional Information

Should this line not  be named 'circular' instead of 'circle'?

https://github.com/ionic-team/ionic-framework/blob/a6a772a38d15419ad4a60b6afeb66d91f6736934/core/src/components/spinner/spinner.scss#L122

dexster:
https://stackblitz.com/edit/angular-ivy-etyain?file=src%2Fapp%2Fapp.component.html

sean-perkins:
@dexster the SVG contents for that specific variation uses `fill: currentColor;`.

If you update your example to:

```css
ion-spinner {
  color: red;
}
```

The appropriate styles will apply. Can you confirm the same on your end?

dexster:
@sean-perkins I am using --color currently which is working. Would the resolution then be to update the ion-loading docs for this specific use case or will you be looking at getting this to use --spinner-color?

liamdebeasi:
There are a couple of issues here:

[...]

PR #25259, opened 2022-05-09
test(spinner): migrate tests to playwright

liamdebeasi:
## Pull request checklist

## Pull request type

 

 

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes

This is a prerequisite to https://github.com/ionic-team/ionic-framework/issues/25180

## What is the new behavior?

- Migrated all spinner tests to playwright
- The Transform E2E test has been removed. Upon investigation I found that the test was not testing what we thought it was testing. The animations on the test page were paused at time offset 0, yet the test waited ~1500ms before taking a screenshot. Since the animation was paused, it would not progress past time offset 0. As a result, if the fix it was checking regressed, we would not have caught it. Unfortunately, I was not able to find a stable way to test this. I was able to set the `animation-delay` which lets me track through an animation. But I could not find a stable way to have Playwright wait for the browser to re-render before taking a screenshot.

- [ ] Yes
- [x] No


sean-perkins (approved):
Linux seems to be rendering the circle and bubble SVGs strangely:
- core/src/components/spinner/test/color/spinner.e2e.ts-snapshots/spinner-color-diff-ios-rtl-Mobile-Chrome-linux.png
- core/src/components/spinner/test/color/spinner.e2e.ts-snapshots/spinner-color-diff-md-rtl-Mobile-Chrome-linux.png

I don't think this is an issue, if it consistently captures this way, just interesting that it is the only run target that captures an image with this output.

PR #25264, opened 2022-05-10
fix(css): rtl mixin preserves selectors after :host

crazyserver:
Ie: :host .element-selector when using rtl was translated as :host-context([dir=rtl]) instead of :host-context([dir=rtl]) .element-selector

## Pull request checklist

## Pull request type

 

 

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

With the following CSS:
```css
:host .element-selector {
    @include position(0, 0, null, null);
}
```

When using the rtl mixin so, also add-root-selector it translates the selector as:
```css
:host-context([dir=rtl])
```
So part of it is lost.

Issue URL: #25285

## What is the new behavior?

With the following CSS:
```css
:host .element-selector {
    @include position(0, 0, null, null);
}
```

When using the rtl mixin so, also add-root-selector it translates the selector as:
```css
:host-context([dir=rtl]) .element-selector
```

Maintaining the whole selector.

- [ ] Yes
- [x] No


liamdebeasi:
Thanks for the PR! We require that any fix/feature PRs to be associated with an existing issue prior to merging. Is there an open issue that we can link this to?

Having an issue with steps to reproduce helps us better verify the proposed fix.

crazyserver:
Thanks I've opened: https://github.com/ionic-team/ionic-framework/issues/25285

liamdebeasi:
Thanks! There are several components that make use of the `:host .selector` syntax with RTL utilities. These will also need to be updated to account for this change.

For example this line in datetime probably does not need to position mixin: https://github.com/ionic-team/ionic-framework/blob/main/core/src/components/datetime/datetime.scss#L346

However, the change in this PR breaks its positioning.

-----

We will schedule some time in an upcoming sprint to scrutinize which components need updates and make the appropriate changes.

sean-perkins:
Ignore the large file diff for now (getting things ready for team review/merge into 6.4). Once I sync 6.4 with main the diff will be the single file change.

Issue #25255, opened 2022-05-08
bug: vue, switching tabs causes wrong view to be unmounted


Jannik-KJ:
### Prerequisites

- [X] I have read the [Contributing Guidelines](https://github.com/ionic-team/ionic-framework/blob/main/.github/CONTRIBUTING.md#creating-an-issue).
- [X] I agree to follow the [Code of Conduct](https://ionicframework.com/code-of-conduct).
- [X] I have searched for [existing issues](https://github.com/ionic-team/ionic-framework/issues) that already report this problem, without success.

### Ionic Framework Version

- [ ] v4.x
- [ ] v5.x
- [X] v6.x
- [ ] Nightly

### Current Behavior

App crashes when navigating between tabs and tabs child pages.

Error thrown:

` Uncaught (in promise) TypeError: Cannot set properties of undefined (setting 'mount')
    at Object.unmountLeavingViews (index.esm.js?ba1f:802:1)
    at handlePageTransition (index.esm.js?58b7:1320:1)`

### Expected Behavior

The application can navigate between tabs and child pages of tabsThe application

### Steps to Reproduce

on start page (tab 1) ->
1) Click "go to child page"
2) click TAB 2
3) click TAB 1
4) click TAB 1 again
5) click TAB 2
6) click TAB 1
7) Click "go to child page"
8) click TAB 2
9) click TAB 1
10) click TAB 1 agina

### Code Reproduction URL

https://github.com/Jannik-KJ/ionic-6-router

### Ionic Info

Ionic:

   Ionic CLI       : 6.18.1 (/usr/local/lib/node_modules/@ionic/cli)
   Ionic Framework : @ionic/vue 6.1.3

Capacitor:

   Capacitor CLI      : 3.5.0
   @capacitor/android : not installed
   @capacitor/core    : 3.5.0
   @capacitor/ios     : not installed

Utility:

   cordova-res : not installed globally
   native-run  : 1.5.0

System:

   NodeJS : v16.13.1 (/usr/local/bin/node)
   npm    : 8.1.2
   OS     : macOS Monterey

anttironty:
Hi, this issue is present in our app aswell with Ionic versions over 6.0.8-dev.1644430127.1b9bb92

liamdebeasi:
Thanks for the issue. Can you try the following dev build and let me know if it resolves the issue?

```
[...]

PR #25270, opened 2022-05-10
fix(vue): correct views are now unmounted in tabs

liamdebeasi:
## Pull request checklist

## Pull request type

 

 

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

When using `router.go()` it is possible for users to navigate back over multiple views:

`/page1` --> `/page2` --> `/page3` --> `router.go(-2)`

The above example will bring the user back to `/page1`. Since Ionic renders multiple views in the DOM at the same time, it is important that we unmount any intermediary views. For example, if we did not unmount intermediary views then the view for `/page2` would never get unmounted. This can result in a DOM node leak. Similarly, we need to re-mount intermediary views if you then called `router.go(2)`.

---

For apps that route in a linear manner, this is fine. However, there is a fundamental incompatibility between nested outlets in Ionic (such as tabs) and `router.go()`. The reason is that `router.go()` expects routing to be done in a linear manner (such as the example above). Nested outlets such as tabs are fundamentally non-linear. As a result, `router.go()` typically will not yield the results that developers are expecting.

In the linked issue below, we were trying to anticipate `router.go()` usage which resulted in the wrong view items being unmounted (essentially we tried to unmount an undefined view item).

Issue URL: resolves https://github.com/ionic-team/ionic-framework/issues/25255

## What is the new behavior?

- Apps that do not use linear routing no longer have intermediary views unmounted/mounted. This means that `router.go()` will not be compatible with non-linear applications. In fairness, we never actually supported `router.go()` in non-linear environments. The problem was that when we added `router.go()` support for linear environments, we never put any limits on when it could be used because we were unaware of the tradeoffs at the time of implementing.
- I have also added docs that explain how non-linear routing works as well as the tradeoffs (I.e. no router.go support) https://github.com/ionic-team/ionic-docs/pull/2333

- [ ] Yes
- [x] No

## Other information

Dev build: 6.1.5-dev.11652207609.1ba60afe


packages/vue/test-app/tests/e2e/specs/tabs.js
```diff
@@ -66,7 +66,8 @@ describe('Tabs', () => {
 
     cy.ionPageVisible('tab1');
 
-    cy.ionPageDoesNotExist('tab1childone');
+    // TODO(FW-1420)
+    //cy.ionPageDoesNotExist('tab1childone');
```
liamdebeasi, line 70: This is still an issue which is why I am commenting it out. It was previously working by chance and I mistakenly thought it was fixed so I uncommented this line.

amandaejohnston (approved):

sean-perkins (approved):

Issue #25179, opened 2022-04-23
bug: ion-picker not showing all options on Chrome


Artfloriani:
### Prerequisites

- [X] I have read the [Contributing Guidelines](https://github.com/ionic-team/ionic-framework/blob/main/.github/CONTRIBUTING.md#creating-an-issue).
- [X] I agree to follow the [Code of Conduct](https://ionicframework.com/code-of-conduct).
- [X] I have searched for [existing issues](https://github.com/ionic-team/ionic-framework/issues) that already report this problem, without success.

### Ionic Framework Version

- [X] v4.x
- [X] v5.x
- [X] v6.x
- [ ] Nightly

### Current Behavior

When opening an ion-picker, the 3rd option below the selected value is not shown on Chrome. After scrolling the option renders properly. See the gif below, taken from the [ion-picker docs](https://ionicframework.com/docs/api/picker):

![Screen Recording 2022-04-23 at 14 30 11](https://user-images.githubusercontent.com/6491691/164923293-82380562-4124-4e07-a385-739030d84c4f.gif)

### Expected Behavior

By default, I expected to be able to see 2 options below and above the selected value on ion-picker, without having to scroll.

### Steps to Reproduce

On the latest version of Google Chrome:
1. Open the [Picker Docs](https://ionicframework.com/docs/api/picker);
2. Click on "Show Single Column Picker" on the example; 
3. You should be able to see that not all options are visible before scrolling;

### Ionic Info

```
Ionic:

   Ionic CLI                     : 6.18.1 (/Users/***/.nvm/versions/node/v14.17.5/lib/node_modules/@ionic/cli)
   Ionic Framework               : @ionic/angular 4.11.13
   @angular-devkit/build-angular : 0.801.3
   @angular-devkit/schematics    : 8.3.29
   @angular/cli                  : 8.3.29
   @ionic/angular-toolkit        : 2.3.3

Capacitor:

   Capacitor CLI      : 2.4.7
   @capacitor/android : 2.4.7
   @capacitor/core    : 2.4.7
   @capacitor/ios     : 2.4.7

Cordova:

   Cordova CLI       : not installed
   Cordova Platforms : not available
   Cordova Plugins   : not available

Utility:

   cordova-res : 0.15.4
   native-run  : 1.5.0

System:

   ios-sim : 8.0.2
   NodeJS  : v14.17.5 (/Users/***/.nvm/versions/node/v14.17.5/bin/node)
   npm     : 6.14.13
[...]

PR #25294, opened 2022-05-17
fix(picker-column): columns options render for iOS mode on chrome

sean-perkins:
## Pull request checklist

## Pull request type

 

 

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

When displaying `ion-picker` in iOS mode in Chrome, some of the options in the picker column will not display until the picker is interacted with.

Issue URL: #25179

## What is the new behavior?

- Removes the `contain` style that was affecting Chrome being able to paint the transform correctly
- Adds screenshot tests 

- [ ] Yes
- [x] No


core/src/components/picker-column/test/standalone/picker-column.e2e.ts
```diff
@@ -0,0 +1,19 @@
+import { test } from '@utils/test/playwright';
+
+import { testPickerColumn } from '../test.utils';
+
+test.describe('picker-column', () => {
```
liamdebeasi, line 5: Can we merge the screenshot tests in first so we can better check for visual regressions? I am noticing that removing `contain: content` causes the fonts to render bigger than before in WebKit:


| with `contain: content` | without `contain: content` |
| ------- | ------- |
| IMG_0775 | IMG_0776 |

However, I am not sure if this is another bug or how the browser should be behaving.

liamdebeasi (commented):
Overall I am in favor of this change. The `contain: content` seems like it was added as a premature optimization: https://github.com/ionic-team/ionic-v3/commit/5ad7fd19080b779cb94766f79c697931a09dec18#diff-8c30bebb97021822c7111db7d0aefce9f40295260bfce770f879fede664a1acf

In reality, this should not be needed. These component makes use of 3d transforms, `will-change` (which it probably should not), and other CSS properties that cause the browser to create new stacking contexts. As a result, `contain: content` does not appear to be causing anything new to happen.

Issue #25318, opened 2022-05-19
bug: refresher triggers when scrolling up on custom scroll host


TimGels:
### Prerequisites

- [X] I have read the [Contributing Guidelines](https://github.com/ionic-team/ionic-framework/blob/main/.github/CONTRIBUTING.md#creating-an-issue).
- [X] I agree to follow the [Code of Conduct](https://ionicframework.com/code-of-conduct).
- [X] I have searched for [existing issues](https://github.com/ionic-team/ionic-framework/issues) that already report this problem, without success.

### Ionic Framework Version

- [ ] v4.x
- [ ] v5.x
- [X] v6.x
- [ ] Nightly

### Current Behavior

When using the `<ion-refresher-content>` component in combination with Angular's `<cdk-virtual-scroll-viewport>` component, the refresher gets triggered when scrolling in the list. I also added the `ion-content-scroll-host` class to the `<cdk-virtual-scroll-viewport>` component to no avail.

https://user-images.githubusercontent.com/43609220/169274311-3a40c1d9-b7b6-4c61-a85f-db1ef107b888.mp4

### Expected Behavior

When using the `<ion-refresher-content>` component in combination with Angular's ` <cdk-virtual-scroll-viewport>` component, the refresher should only be triggered when pulling at the start of the list.

### Steps to Reproduce

1. Run the repro Ionic application
2. Put the screen in responsive mode like a phone screen
3. Scroll down
4. Scroll up without immediately releasing the mouse down / finger off the screen

### Code Reproduction URL

https://github.com/TimGels/ionic-virtual-scroll-refresher-repro

### Ionic Info

PS C:\Users\user\projects\ionic-virtual-scroll-refresher-repro> ionic info
[WARN] Error loading @capacitor/ios package.json: Error: Cannot find module '@capacitor/ios/package'

       Require stack:
       - C:\Users\user\AppData\Roaming\npm\node_modules\@ionic\cli\lib\project\index.js
       - C:\Users\user\AppData\Roaming\npm\node_modules\@ionic\cli\lib\index.js
       - C:\Users\user\AppData\Roaming\npm\node_modules\@ionic\cli\index.js
       - C:\Users\user\AppData\Roaming\npm\node_modules\@ionic\cli\bin\ionic
[WARN] Error loading @capacitor/android package.json: Error: Cannot find module '@capacitor/android/package'

       Require stack:
[...]

PR #25335, opened 2022-05-23
fix(refresher): attach scroll listener to custom scroll target

sean-perkins:
## Pull request checklist

## Pull request type

 

 

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [x] Other (please describe): Test migrations

- The `ion-refresher` usage with custom scroll target does not attach to the correct scroll container
- The pull to refresh tests are written in Puppeteer and don't function correctly for the custom scroll target behavior

Issue URL: #25318

## What is the new behavior?

- Attaches the scroll behavior for `ion-refresher` to the specified custom scroll target
- Migrates refresher tests to Playwright
- Adds additional options to `page.goto` to support Playwright API
- Exports the function to extend the page fixture, so it can be used when creating new browser contexts (for video recording within Playwright)

- [ ] Yes
- [x] No

## Other information

This PR also removes the unnecessary screenshot tests (they did not capture anything relevant to `ion-refresher`). A follow-up PR will work to introduce screenshot tests for the native & legacy refresher appearances.

Dev-build: `6.1.7-dev.11653327570.1759e1d5` ✅


core/src/utils/test/playwright/playwright-page.ts
```diff
@@ -31,24 +31,42 @@ type CustomFixtures = {
...
+
+  if (testInfo) {
+    page.goto = (url: string, options) => goToPage(page, url, options, testInfo, originalGoto);
+  }
+  page.setContent = (html: string) => setContent(page, html);
+  page.locator = (selector: string, options?: LocatorOptions) => locator(page, originalLocator, selector, options);
+
+  // Custom Ionic methods
+
+  if (testInfo) {
```
liamdebeasi, line 54: This is confusing. `testInfo` should always be available when running E2E tests in Ionic Core. Additionally, the `E2EPage` interface says that these methods are always available, but that would no longer be the case with this change.

sean-perkins: Agreed, these checks can be removed now. Originally I was trying to extend the `browser` fixture when the `newPage()` from the browser context is called, which didn't readily have access to test info. But, with exporting the function, the calling test suite should always have a page and test info context.

core/src/components/refresher/test/test.utils.ts
```diff
@@ -12,10 +11,28 @@ import { dragElementBy } from '@utils/test';
...
+    return;
+  }
+
+  const startX = boundingBox.x + boundingBox.width / 2;
+  const startY = boundingBox.y + boundingBox.height / 2;
+
+  await page.mouse.move(startX, startY);
+  await page.mouse.down();
+
+  for (let i = 0; i < 400; i += 50) {
```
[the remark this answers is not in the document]
sean-perkins: Thoughts on splitting that change out into a follow-up PR? The refresher ended up requiring a little extra behavior as well (waiting for the component to be attached in the DOM).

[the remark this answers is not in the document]
liamdebeasi: Works for me 👍

core/src/components/refresher/test/test.utils.ts
```diff
@@ -12,10 +11,28 @@ import { dragElementBy } from '@utils/test';
  * @param selector The element selector to center the drag gesture on. Defaults to `body`.
  */
 const pullToRefresh = async (page: E2EPage, selector = 'body') => {
-  const target = (await page.$(selector))!;
+  const target = page.locator(selector);
 
-  await dragElementBy(target, page, 0, 200);
-  const ev = await page.spyOnEvent('ionRefreshComplete', 'document');
+  await page.waitForSelector('ion-refresher.refresher-inactive', { state: 'attached' });
```
liamdebeasi, line 16: Rather than wait for a new class, are you able to do `expect(refresher).not.toHaveClass(/refresher-active/)`?

sean-perkins: I believe this would operate differently than the current behavior. Using an `expect` check here will continue to fail on the few test suites that's threshold of navigating to the test and starting the mouse down behavior, happens prior to the component fully rendering and the gestures being initialized. 

Optionally, I could swap this behavior out with:
```ts
 await page.waitForSelector('ion-refresher.hydrated', { state: 'attached' });
```

since Playwright runs on the lazy bundle. I can swap that out quick and make sure that CI still passes.

sean-perkins: Looks like this approach works. Are you comfortable with this change?

liamdebeasi: yep works for me

liamdebeasi (approved):

PR #25346, opened 2022-05-25
chore(playwright): dragElementByYAxis util

sean-perkins:
## Pull request checklist

## Pull request type

 

 

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes

N/A

## What is the new behavior?

- Splits out the behavior for dragging an element on the Y-axis over a loop, into a separate util that other tests can take advantage of (mainly gesture related tests).

- [ ] Yes
- [x] No

### DIFF
--- a/core/src/utils/test/playwright/drag-element.ts
+++ b/core/src/utils/test/playwright/drag-element.ts
@@ -29,7 +29,42 @@ export const dragElementBy = async (
 
   await page.mouse.move(startX, startY);
   await page.mouse.down();
+
   await page.mouse.move(midX, midY);
   await page.mouse.move(endX, endY);
+  await page.mouse.up();
+};
+
+/**
+ * Drags an element by the given amount of pixels on the Y axis.
+ * @param el The element to drag.
+ * @param page The E2E Page object.
+ * @param dragByY The amount of pixels to drag the element by.
+ * @param startYCoord The Y coordinate to start the drag gesture at. Defaults to the center of the element.
+ */
+export const dragElementByYAxis = async (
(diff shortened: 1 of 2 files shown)


core/src/utils/test/playwright/drag-element.ts
```diff
@@ -29,7 +29,42 @@ export const dragElementBy = async (
...
 };
+
+/**
+ * Drags an element by the given amount of pixels on the Y axis.
+ * @param el The element to drag.
+ * @param page The E2E Page object.
+ * @param dragByY The amount of pixels to drag the element by.
+ * @param startYCoord The Y coordinate to start the drag gesture at. Defaults to the center of the element.
+ */
+export const dragElementByYAxis = async (
```
amandaejohnston, line 45: ~ Would it make sense to allow dragging to and from arbitrary points, rather than just the Y axis? I know that would change the logic up a fair amount, so it's cool with me if you'd rather save that for a future PR.

amandaejohnston (approved):

core/src/utils/test/playwright/drag-element.ts
sean-perkins: I had this thought originally, but the complexity seems high with this situation. Going from two plotted points (20, 100) to (350, 350); how would we best calculate the curve that the mouse would travel between the points? Would it be a straight line and increment evenly over the x-axis and the y-axis? I feel like it could be fairly situation to the UI you are testing over.

amandaejohnston: Very true. 🤔 We can always revisit if the use case comes up.

liamdebeasi (approved):

Issue #25320, opened 2022-05-19
bug: ensure that default month shown is always in bounds


ryaa:
### Prerequisites

- [X] I have read the [Contributing Guidelines](https://github.com/ionic-team/ionic-framework/blob/main/.github/CONTRIBUTING.md#creating-an-issue).
- [X] I agree to follow the [Code of Conduct](https://ionicframework.com/code-of-conduct).
- [X] I have searched for [existing issues](https://github.com/ionic-team/ionic-framework/issues) that already include this feature request, without success.

### Describe the Feature Request

Datetime, if no explicit value provided, always pre-select the current date which, in some use cases, leads to the behavior as described here https://github.com/ionic-team/ionic-framework/issues/25296

### Describe the Use Case

See https://github.com/ionic-team/ionic-framework/issues/25296

### Describe Preferred Solution

Allow to out out and avoid the datetime to preselect any day in the calendar grid, when no explicit value is provided.

### Describe Alternatives

None

### Related Code

See https://github.com/ionic-team/ionic-framework/issues/25296

### Additional Information

See https://github.com/ionic-team/ionic-framework/issues/25296

liamdebeasi:
Thanks for the issue. If we updated `ion-datetime` such that the pre-selected date was guaranteed to always be within bounds, would that be an acceptable solution for your use case?

ryaa:
> Thanks for the issue. If we updated `ion-datetime` such that the pre-selected date was guaranteed to always be within bounds, would that be an acceptable solution for your use case?

yes. can you please also make sure that the time picker is pre-populated taking into consideration the min value (both in the time picker itself and in its popover)?
This should prevent empty and confusion time picker and show valid and correct values there.

Thank you

liamdebeasi:
The team met regarding this issue. Here is our current plan for addressing:

**When no `value` is set:**
- The datetime will render the calendar on an in-bound month and will never start the user out of bounds.
- It will use the following algorithm to determine which month to render:
1. If the minimum month > today’s month, show the minimum month.
2. If the maximum month < today’s month, show the maximum month.
3. Otherwise, show today’s month.


- The time picker will show all possible values until you select a date at which point it will only show the times that are in bounds according to the date you selected.

**When a `value` Is set:**
- The datetime will add a console.warn informing the developer that the current value is out of bounds.
- The datetime will render the calendar on a current in-bound month. It will use the above algorithm to determine which month to render.

----

In neither case will the datetime modify the `value` set.

[...]

PR #25351, opened 2022-05-25
fix(datetime): ensure that default month shown is always in bounds

sean-perkins:
## Pull request checklist

## Pull request type

 

 

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

When the user sets a min/max value that is out of bounds of today, the `ion-datetime` will render the month of today, instead of rendering the user within the bounds of the min/max ranges. 

Also, the time selection is initially clamped by the min/max rules, regardless of a calendar day being selected. 

Issue URL: #25320

## What is the new behavior?

- The default month rendered for `ion-datetime`, when no value is manually set, will be clamped between the min, max and today.
- All time values will be available for selection, regardless of the min/max values, when a value has not been set/selected.

- [ ] Yes
- [x] No

## Other information

This PR does not solve the other half of the problem, of clamping the date when the value is set and clamping the time selection once a value is selected. 

Dev-build: `6.1.9-dev.11654117360.13662197`


core/src/components/datetime/utils/parse.ts
```diff
@@ -106,3 +106,13 @@ export const parseDate = (val: string | undefined | null): any | undefined => {
...
   };
 };
+
+export const clampDate = (date: string, min?: string, max?: string): string => {
+  const parsedDate = new Date(date);
+  if (min && parsedDate < new Date(min)) {
+    return min;
+  } else if (max && parsedDate > new Date(max)) {
+    return max;
+  }
```
liamdebeasi, line 116: We should be using `DateTimeParts` to do the comparison, not JS Date as JS Date takes into account different timezones. I believe there are functions in `comparison.ts` you can use.

core/src/components/datetime/utils/data.ts
```diff
@@ -166,7 +167,7 @@ export const generateTime = (
     processedMinutes = processedMinutes.filter((minute) => minuteValues.includes(minute));
   }
 
-  if (minParts) {
+  if (minParts && currentValue !== undefined) {
```
liamdebeasi, line 170: I don't think this function should be tightly coupled with the state of the `value` property. Maybe we could just not pass in the max/min parts when `value === undefined`? That would result in all possible times being generated.

liamdebeasi (changes requested):
Can we also update the description of what the `reset` method does? That uses `processValue` internally as well.

Specifically `If no value is provided, the internal state will be reset to today.` is missing a mention of clamping the values in bounds.

core/src/components/datetime/test/minmax/datetime.e2e.ts
```diff
@@ -47,4 +47,33 @@ test.describe('datetime: minmax', () => {
     expect(nextButton).toBeDisabled();
     expect(prevButton).toBeEnabled();
   });
+
+  test.describe('when the datetime does not have a value', () => {
+    test('all time values should be available for selection', async ({ page }) => {
+      /**
+       * When the datetime does not have an active value, all time values should
+       * be available for selection.
+       */
```
amandaejohnston, line 56: ~ It may be good to clarify somewhere in here that we're specifically testing when today's date falls outside the min/max.

core/src/components/datetime/utils/parse.ts
```diff
@@ -106,3 +108,20 @@ export const parseDate = (val: string | undefined | null): any | undefined => {
     tzOffset,
   };
 };
+
+export const clampDate = (date: string, min?: string, max?: string): string => {
```
liamdebeasi, line 112: By the time `processValue` is called, we have already parsed the `min` and `max` values, so we should be able to pass in the `DatetimeParts` instead of the strings: https://github.com/ionic-team/ionic-framework/blob/main/core/src/components/datetime/datetime.tsx#L1111-L1113

core/src/components/datetime/utils/parse.ts
```diff
@@ -106,3 +108,20 @@ export const parseDate = (val: string | undefined | null): any | undefined => {
...
+  const dateParts = parseDate(date);
```
liamdebeasi, line 113: You could also optionally parse `date` in `datetime.tsx` so that this function takes in `DatetimeParts` across the board, but I don't have any strong opinions on this. The one benefit I see is consistency of arguments given that they are all dates.

liamdebeasi (changes requested):

liamdebeasi (approved):

Issue #22030, opened 2020-09-04
bug: tap click activates buttons without scrollEvents, but scrollEvents introduces perf issues


zrev2220:
# Bug Report

**Ionic version:**

[ ] **4.x**
[x] **5.x**

**Current behavior:**

When scrolling an `ion-list` with clickable `ion-item`s (`button="true"`), the item the user touches in order to scroll plays its "activated" style or animation (i.e. ripple effect on Android, darkened background on iOS). This happens _even if the user is not "clicking" on that item._

**Expected behavior:**

When scrolling the list, the touched item should not change in appearance (play the ripple effect or change the background). These styles should only activate when the user taps/clicks an item to __select__ it.

**Steps to reproduce:**

1. Create a basic Ionic app (`ionic start`)
2. Add an `ion-list` to one of the pages containing enough `ion-item`s to enable scrolling
3. Scroll the list with the app running in a mobile-ish environment (i.e. emulated touch enabled via devtools in a browser, or run on a physical device)

**Related code:**

StackBlitz: https://stackblitz.com/edit/ionic-list-ripple-bug

Snippet from Angular template:

```html
<ion-list>
  <ion-item *ngFor="let i of range(100)" button="true" detail="false">
    Item #{{i}}
  </ion-item>
</ion-list>
```

> The `range` function is defined as follows:
> 
> ```ts
> range(n) {
>   return Array.from(Array(n).keys());
> }
> ```

**Other information:**

This bug seems similar/identical to that identified in #15752. That issue references 7f38d377 as fixing the problem, but since I can reproduce it in a recent Ionic version, it seems to have reappeared(?).

**Ionic info:**

> _**Note:** I created the above StackBlitz example by forking https://stackblitz.com/edit/ionic-v4-angular-tabs. That codebase does not run as-is via `ionic serve` or `npm start`, and `ionic info` fails with an error:_
> 
> ```
> [WARN] You are not in an Ionic project directory. Project context may be missing.
> ```
> 
> _To at least provide_ some _useful `ionic info` output, I ran `ionic init` to make the directory an Ionic project. However, some Angular packages are still not installed (since StackBlitz magically makes things work), and the example may not be runnable locally without some additional installations and setup work._

```
[WARN] Error loading @ionic/angular-toolkit package.json: Error: Cannot find module '@ionic/angular-toolkit/package'

       Require stack:
       - <path to global node_modules>\@ionic\cli\lib\project\index.js
[...]

PR #25352, opened 2022-05-25
fix(all): ripple effect is no longer added when scrolling

liamdebeasi:
## Pull request checklist

## Pull request type

 

 

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

Issue URL: resolves https://github.com/ionic-team/ionic-framework/issues/22030

The tap click utility in Ionic ensures that the active state is added to interactives on iOS and Android. On iOS this is an active shade color. On Android this is the ripple effect.

To align with native, the active state should never be added when scrolling. For example, if you begin to scroll by tapping on a button, that button should not receive the ripple effect. See the below example on Android 12:

https://user-images.githubusercontent.com/2721089/170359341-42c68778-fb8e-4cf3-9c00-f3446feb0a1b.mp4

Observe that the ripple effect does not appear on items as the user scrolls. However, when tapping/pressing an item when not scrolling shows the ripple effect.

-------

In Ionic, this used to work by listening for the `scroll` event on `ion-content`. However, this behavior unintentionally changed in https://github.com/ionic-team/ionic-framework/commit/fd1b44a40b741088e099f6538dd14caa0dc5540c. The downside of this approach is that it adds quite a bit of main thread work due to the scroll listener.

How this all comes together is on `touchstart`, we create a `setTimeout` (https://github.com/ionic-team/ionic-framework/blob/main/core/src/utils/tap-click.ts#L103) which, when fired, will add the activated state. When we detect scrolling, the `cancelActive` function is invoked (https://github.com/ionic-team/ionic-framework/blob/main/core/src/utils/tap-click.ts#L151) which cancels the `setTimeout`.

## What is the new behavior?

- This PR re-introduces the tap click scrolling fix by utilizing the `pointercancel` event. This event is dispatched when the pointer is used to pan, zoom, or scroll (among other things): https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/pointercancel_event
When tap click was originally built, the PointerEvents API was not supported in all of the browsers Ionic supports. That is no longer true today: https://caniuse.com/mdn-api_pointerevent

When we begin to scroll, the `pointercancel` event is fired which then calls `cancelActive` which prevents the active state from being added.

See a demo below:

|       | `main` | branch |
| --- | ------- | ------- |
| md | <video src="https://user-images.githubusercontent.com/2721089/170360442-079286ad-30f5-4477-9fff-d5576d89408c.mp4"></video> | <video src="https://user-images.githubusercontent.com/2721089/170360785-fef761e8-f2c3-4130-a2a7-c4687da67f2b.mp4"></video> |
| ios | <video src="https://user-images.githubusercontent.com/2721089/170361920-a21f0221-3d60-4479-bbf3-96096f087790.mp4"></video> | <video src="https://user-images.githubusercontent.com/2721089/170362013-513b4980-7259-45b6-bd4a-a55d6c75ef18.mp4"></video> |

------

There is another bug on Android that this change fixes. Currently, when you swipe on an interactive (but don't scroll), the ripple effect is added to the element. However, the browser never dispatches the `click` event, so it looks like you selected the element and nothing happened. With `pointercancel`, this no longer happens.

See a demo below

| `main` | branch |
| ------- | ------- |
| <video src="https://user-images.githubusercontent.com/2721089/170361207-475cb3aa-0461-4aee-9ca8-afbf75d428d8.mp4"></video> | <video src="https://user-images.githubusercontent.com/2721089/170361304-f4e0fd88-223f-4681-9d28-d6c04b73a7e8.mp4"></video> |

Note: Safari does not dispatch `pointercancel` in this scenario. However, this bug existed even when we were using the old scroll events method so I am not sure it's worth fixing at the moment.

- [ ] Yes
- [x] No


liamdebeasi:
A few things: 

1. Let me know if you'd prefer a meeting to review this.
2. Any thoughts on how to automate a test for this? Since the activated state is added asynchronously, it's a bit hard to ensure that the `ion-activated` class is not added without it being flaky.

sean-perkins (approved):
🚀  thanks for the walkthrough!

amandaejohnston (approved):

Issue #16279, opened 2018-11-08
feat: Translate the Time Period (am/pm) in `ion-datetime`


troydcthompson:
# Bug Report

**Ionic Info**
```
@ionic/core@4.0.0-beta.15

Ionic: 

   ionic (Ionic CLI) : 4.2.1 (/Users/troydcthompson/.nvm/versions/node/v8.9.4/lib/node_modules/ionic)

Cordova:

   cordova (Cordova CLI) : 8.0.0
   Cordova Platforms     : none
   Cordova Plugins       : no whitelistedplugins (0 plugins total)

System:

   ios-deploy : 2.0.0
   NodeJS     : v8.9.4 (/Users/troydcthompson/.nvm/versions/node/v8.9.4/bin/node)
   npm        : 6.4.0
   OS         : macOS High Sierra
   Xcode      : Xcode 9.3 Build version 9E145
```

**Describe the Bug**
There isn't a way to translate the am/pm time period in ion-datetime, whereas we can translate the cancelText, doneText, monthShortName, etc.

**Steps to Reproduce**
1. Implement i[on-datetime,](https://beta.ionicframework.com/docs/api/datetime/) 
2. Enter cancelText, doneText, monthShortName variables
3. Try to enter "A" or timePeriod

**Related Code**
Same issue as found on this  [Ionic Forum Question](https://forum.ionicframework.com/t/translate-the-time-period-am-pm-in-ion-datetime/113280)

**Expected Behavior**
Expect to have a variable for the am/pm variables that we can translate

troydcthompson:
?

liamdebeasi:
Thanks for the issue. This has been resolved via https://github.com/ionic-team/ionic-framework/pull/25389, and this feature will be available in an upcoming minor release of Ionic.

PR #25389, opened 2022-06-01
feat(datetime): localize am/pm labels in time picker

sean-perkins:
## Pull request checklist

## Pull request type

 

 

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes

The labels for "am" and "pm" are not localized to the user's current locale. This means in locales that either do not use am/pm in their time selection or have alternative labels, the text is incorrect. 

This also affects the initial display of the time value, when using a presentation mode that opens the time picker in a popover. 

Issue URL: #16279

## What is the new behavior?

- The am/pm labels within the time picker (popover) are localized to the user's locale
- The value displayed within the time picker, when using a presentation mode that presents the time picker in a popover, is localized.
- The width of the popover will respond to grow to account for languages that have am/pm label representations that are longer than the standard width and would cause overflow.
- Locales that format the day period at the beginning of the label, will display the day period as the first column in the time picker
- Migrates the locale datetime tests from puppeteer to playwright.

Example of the change with a locale of: `ta-IN`:

|iOS|MD|
|---|---|
|Screen Shot 2022-06-07 at 1 46 36 PM|Screen Shot 2022-06-07 at 1 46 52 PM|

- [ ] Yes
- [x] No

## Other information

Dev-build: `6.1.8-dev.11654624050.16b62b1b` (outdated)


amandaejohnston (commented):
Some of the ta-IN screenshots show time picker popovers that are too narrow:
- `core/src/components/datetime/test/locale/datetime.e2e.ts-snapshots/datetime-locale-ta-IN-time-diff-md-ltr-Mobile-Chrome-linux.png`
- `core/src/components/datetime/test/locale/datetime.e2e.ts-snapshots/datetime-locale-ta-IN-time-diff-md-ltr-Mobile-Firefox-linux.png`
- `core/src/components/datetime/test/locale/datetime.e2e.ts-snapshots/datetime-locale-ta-IN-time-diff-md-rtl-Mobile-Chrome-linux.png`
- `core/src/components/datetime/test/locale/datetime.e2e.ts-snapshots/datetime-locale-ta-IN-time-diff-md-rtl-Mobile-Firefox-linux.png`

Not sure if that's a screenshot timing issue or an actual bug, though.

Also, don't forget to remove the old e2e.ts file.

core/src/components/datetime/utils/parse.ts
```diff
@@ -106,3 +106,29 @@ export const parseDate = (val: string | undefined | null): any | undefined => {
...
+ * @param locale The locale to format the day period in.
+ * @param value The date string, in ISO format.
+ * @returns The localized day period (am/pm) representation of the given value.
+ */
+export const parseLocalizedAmPm = (locale: string, dateStringIso: string) => {
+  return new Intl.DateTimeFormat(locale, {
+    hour: 'numeric',
+    timeZone: 'UTC',
+  })
+    .formatToParts(new Date(dateStringIso))
```
liamdebeasi, line 132: The exact date string here isn't really important and it seems a bit odd to have to pass in the correct date string every time.

Have you considered making this a function that accepts a locale and either `'am'` or `'pm'` and then returns the formatted value?

Something like this:

```typescript
const formatAMPM = (locale: string, ampm: 'am' | 'pm') => {
  // You could even cache these in the file itself 
  // so you don't have to create the same date objects every invocation
  const date = (ampm === 'am') ? someAMDateObject : somePMDateObject;

  if (supportsDayPeriodInIntlDateTime) {
    /// create formatted label
    return formattedAMPMLabel;
  }

   // fallback for browsers that don't support dayPeriod
  return ampm;
}
```

There are a few benefits to this approach:

1. You can continue to just pass `refParts.ampm` in `getFormattedTime`. 
2. You would no longer have to create throwaway dates in https://github.com/ionic-team/ionic-framework/pull/25389/files#diff-4a407530c60e3cf72bcc11acdd21c4803a94bf47ea81b99e757db1c93d2735b8R1705.
3. You would no longer need an explicit fallback after the code in point 2 because `formatAMPM` handles the fallback.

sean-perkins:
Still working through the popover sizing (specifically on iOS). Will re-ping for reviews once I get that resolved 🤔

liamdebeasi (changes requested):

core/src/components/datetime/datetime.tsx
```diff
@@ -1526,71 +1534,96 @@ export class Datetime implements ComponentInterface {
     ampmItems: PickerColumnItem[],
     use24Hour: boolean
   ) {
+    const isDayPeriodRTL = isLocaleDayPeriodRTL(this.locale);
+    return (
+      <ion-picker-internal>
+        {isDayPeriodRTL && this.renderDayPeriodPickerColumn(ampmItems)}
```
amandaejohnston, line 1540: Should this also check `!use24Hour`?

amandaejohnston (approved):

amandaejohnston (commented):
(Whoops, clicked approve when I meant comment 😆)

core/src/components/datetime/utils/format.ts
```diff
@@ -24,7 +26,11 @@ export const getFormattedTime = (refParts: DatetimeParts, use24Hour: boolean): s
     return `${hour}:${minute}`;
   }
 
-  return `${hour}:${minute} ${getFormattedAMPM(refParts.ampm)}`;
+  return new Intl.DateTimeFormat(locale, {
```
liamdebeasi, line 29: Any reason why we use `Intl.DateTimeFormat` here but not when we have 24 hour time above?

core/src/components/datetime/datetime.tsx
sean-perkins: Yes it should, great catch!

amandaejohnston (approved):

core/src/components/datetime/datetime.tsx
```diff
@@ -1526,71 +1534,96 @@ export class Datetime implements ComponentInterface {
     ampmItems: PickerColumnItem[],
     use24Hour: boolean
   ) {
+    const isDayPeriodRTL = isLocaleDayPeriodRTL(this.locale);
+    return (
+      <ion-picker-internal>
+        {!use24Hour && isDayPeriodRTL && this.renderDayPeriodPickerColumn(ampmItems)}
+        {this.renderHourPickerColumn(hoursItems)}
+        {this.renderMinutePickerColumn(minutesItems)}
+        {!use24Hour && !isDayPeriodRTL && this.renderDayPeriodPickerColumn(ampmItems)}
```
liamdebeasi, line 1543: Any reason why you opted to do this instead of re-ordering with CSS? I believe we re-order with css in then month/year picker, though there may be some keyboard considerations with using `order` instead of physically changing the order of the DOM nodes.

liamdebeasi (approved):
Great job!

core/src/components/datetime/datetime.tsx
sean-perkins: Only reason was because the picker columns are rendered in a popover, but an inline style works great here. Updated the approach!

sean-perkins:
It looks like when we updated picker column internal to make the rendered items buttons, we broke keyboard navigation across the columns (using tab to move across columns). 

Now when tabbing through the interface, you move through each individual button, column by column. 

Thoughts on what we can do here? I know our motivation for swapping it to a button was for screen readers. Would `role="button"` be a bad move here?

liamdebeasi:
Ok PR is up: https://github.com/ionic-team/ionic-framework/pull/25464

Issue #25353, opened 2022-05-25
feat: support Angular 14


lincolnthree:
### Prerequisites

- [X] I have read the [Contributing Guidelines](https://github.com/ionic-team/ionic-framework/blob/main/.github/CONTRIBUTING.md#creating-an-issue).
- [X] I agree to follow the [Code of Conduct](https://ionicframework.com/code-of-conduct).
- [X] I have searched for [existing issues](https://github.com/ionic-team/ionic-framework/issues) that already include this feature request, without success.

### Describe the Feature Request

In Angular 14, there is a breaking change that removes the deprecated `ComponentFactoryResolver`, specifically from router outlets. Since `@ionic/angular` currently uses this to in `ion-router-outlet`, Angular applications using Ionic will stop working once upgraded to version 14. Ionic will need to update to support the next version of Angular (currently in RC).

### Describe the Use Case

Allow users to continue staying current with Angular releases.

### Describe Preferred Solution

Remove references to deprecated resolvers. But whatever you think is best :)

From Angular upgrade guide:
```
Remove the resolver from RouterOutletContract.activateWith function and the resolver from OutletContext class since factory resolvers are no longer needed.
```

### Describe Alternatives

Unknown.

### Additional Information

Link to Angular deprecation info:
https://angular.io/api/core/ComponentFactoryResolver#description 

Angular v14 Upgrade Guide:
https://update.angular.io/?l=3&v=13.0-14.0

Stack trace:

```
E/Capacitor/Console: File: http://localhost/polyfills.js - Line 1349 - Msg: Uncaught Error: Uncaught (in promise): TypeError: resolver.resolveComponentFactory is not a function
    TypeError: resolver.resolveComponentFactory is not a function
        at IonRouterOutlet.activateWith (http://localhost/vendor.js:31027:32)
        at ActivateRoutes.activateRoutes (http://localhost/vendor.js:170835:28)
        at http://localhost/vendor.js:170779:12
        at Array.forEach (<anonymous>)
        at ActivateRoutes.activateChildRoutes (http://localhost/vendor.js:170778:25)
        at ActivateRoutes.activateRoutes (http://localhost/vendor.js:170842:14)
        at http://localhost/vendor.js:170779:12
        at Array.forEach (<anonymous>)
        at ActivateRoutes.activateChildRoutes (http://localhost/vendor.js:170778:25)
        at ActivateRoutes.activateRoutes (http://localhost/vendor.js:170842:14)
[...]

PR #25403, opened 2022-06-03
fix(angular): add support for Angular 14

sean-perkins:
## Pull request checklist

## Pull request type

 

 

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

Applications that are upgrading from Angular 13 to Angular 14 will receive an exception that breaks the entire app experience. 

```
TypeError: resolver.resolveComponentFactory is not a function
```

This is due to the deprecation and removal of `ComponentFactoryResolver` in Angular 14.

Issue URL: #25353

## What is the new behavior?

- Adds backwards compatibility for Angular 13 applications
- Adds compatibility for Angular 14 applications

- [ ] Yes
- [x] No

## Other information

Dev-build: `6.1.9-dev.11654275237.1b595be3`

Angular 14 demo: https://github.com/sean-perkins/angular-14-demo

For Angular 13, you can use the starters and install the dev-build.

Note: Standalone components are not supported in this phase of development. As far as I can tell, they require `EnvironmentInjector` and other essential infrastructure only available in the Angular 14 bundle. More exploration is required to pass the application's injector through to Ionic's module.

### DIFF
--- a/angular/src/directives/navigation/ion-router-outlet.ts
+++ b/angular/src/directives/navigation/ion-router-outlet.ts
@@ -20,7 +20,6 @@ import { componentOnReady } from '@ionic/core';
 import { Observable, BehaviorSubject } from 'rxjs';
 import { distinctUntilChanged, filter, switchMap } from 'rxjs/operators';
 
-
 import { EnvironmentInjector } from '../../di/r3_injector';
 import { AnimationBuilder } from '../../ionic-core';
 import { Config } from '../../providers/config';
--- a/angular/src/directives/navigation/ion-router-outlet.ts
+++ b/angular/src/directives/navigation/ion-router-outlet.ts
@@ -19,12 +19,13 @@ import { OutletContext, Router, ActivatedRoute, ChildrenOutletContexts, PRIMARY_
 import { componentOnReady } from '@ionic/core';
 import { Observable, BehaviorSubject } from 'rxjs';
 import { distinctUntilChanged, filter, switchMap } from 'rxjs/operators';
-import { isComponentFactoryResolver } from '../../util/util';
+
 
 import { EnvironmentInjector } from '../../di/r3_injector';
 import { AnimationBuilder } from '../../ionic-core';
 import { Config } from '../../providers/config';
 import { NavController } from '../../providers/nav-controller';
+import { isComponentFactoryResolver } from '../../util/util';
 
 import { StackController } from './stack-controller';
 import { RouteView, getUrl } from './stack-utils';
--- a/angular/src/directives/navigation/ion-router-outlet.ts
+++ b/angular/src/directives/navigation/ion-router-outlet.ts
@@ -19,6 +19,7 @@ import { OutletContext, Router, ActivatedRoute, ChildrenOutletContexts, PRIMARY_
 import { componentOnReady } from '@ionic/core';
 import { Observable, BehaviorSubject } from 'rxjs';
 import { distinctUntilChanged, filter, switchMap } from 'rxjs/operators';
+import { isComponentFactoryResolver } from '../../util/util';
 
 import { EnvironmentInjector } from '../../di/r3_injector';
 import { AnimationBuilder } from '../../ionic-core';
@@ -407,8 +408,4 @@ class OutletInjector implements Injector {
 
     return this.parent.get(token, notFoundValue);
   }
-}
-
-function isComponentFactoryResolver(item: any): item is ComponentFactoryResolver {
-  return !!item.resolveComponentFactory;
 }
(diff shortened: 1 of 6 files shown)


angular/src/directives/navigation/ion-router-outlet.ts
```diff
@@ -206,7 +207,7 @@ export class IonRouterOutlet implements OnDestroy, OnInit {
     }
   }
 
-  activateWith(activatedRoute: ActivatedRoute, resolver: ComponentFactoryResolver | null): void {
+  activateWith(activatedRoute: ActivatedRoute, resolverOrInjector?: ComponentFactoryResolver | null): void {
```
amandaejohnston, line 210: Should this be `ComponentFactoryResolver | EnvironmentInjector | null`?

angular/src/directives/navigation/ion-router-outlet.ts
```diff
@@ -240,8 +238,33 @@ export class IonRouterOutlet implements OnDestroy, OnInit {
...
 
+      if (resolverOrInjector && isComponentFactoryResolver(resolverOrInjector)) {
+        // Backwards compatibility for Angular 13 and lower
+        const factory = resolverOrInjector.resolveComponentFactory(component);
+        cmpRef = this.activated = this.location.createComponent(factory, this.location.length, injector);
+      } else {
+        /**
+         * Angular 14 and higher.
+         *
+         * When we drop < Angular 14, we can replace the following code with:
```
amandaejohnston, line 250: Could we add a tech debt ticket here too? (Could be the same ticket as the other one)

Also, if/when we don't support Angular 13, do we still need the backwards compatibility code above?

angular/src/providers/angular-delegate.ts
```diff
@@ -96,14 +98,29 @@ export const attachView = (
...
-  const componentRef = location
-    ? location.createComponent(factory, location.length, childInjector)
-    : factory.create(childInjector);
+
+  if (resolverOrInjector && isComponentFactoryResolver(resolverOrInjector)) {
+    // Angular 13 and lower
+    const factory = resolverOrInjector.resolveComponentFactory(component);
+    componentRef = location
+      ? location.createComponent(factory, location.length, childInjector)
+      : factory.create(childInjector);
```
amandaejohnston, line 112: Will we still need this when we drop Angular 13 support? If not, could add the tech debt ticket comment here too.

angular/src/providers/angular-delegate.ts
```diff
@@ -165,3 +182,7 @@ const getProviders = (params: { [key: string]: any }) => {
 const provideNavParamsInjectable = (params: { [key: string]: any }) => {
   return new NavParams(params);
 };
+
+function isComponentFactoryResolver(item: any): item is ComponentFactoryResolver {
+  return !!item.resolveComponentFactory;
+}
```
amandaejohnston, line 188: This function is duplicated in `ion-router-outlet.ts`; would it be possible to make a shared util instead?

amandaejohnston (commented):
Great work on this! Just had a few minor things.

angular/src/providers/angular-delegate.ts
sean-perkins: TBD. The standalone components are in dev preview, so Angular is still reserving the right to completely remove them if they discover a reason why they shouldn't exist. As far as I can tell, all of this is backwards support for the previous model of using the component factory resolver for creating the component to resolve for a route. It is still present in Angular's implementation of `router-outlet`, so we would likely want to maintain it until they remove it, which would be Angular 15 at the earliest.

angular/src/directives/navigation/ion-router-outlet.ts
```diff
@@ -229,9 +234,6 @@ export class IonRouterOutlet implements OnDestroy, OnInit {
       const snapshot = (activatedRoute as any)._futureSnapshot;
       // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
       const component = snapshot.routeConfig!.component as any;
-      resolver = resolver || this.resolver;
```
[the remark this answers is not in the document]
sean-perkins: It looked to be an unnecessary fallback added when stack navigation was added to Ionic Angular: https://github.com/ionic-team/ionic-framework/commit/726938f67cf73d5dcc7cd5c0d2dbbf3b02d91009#diff-676a7dea0a23c6e3559412b3ce504dd48f32062d1756f74aac7f7d72e8d411ac

In Angular's router implementation, the resolver (or injector starting in v14), is always passed through the function: https://github.com/angular/angular/blob/main/packages/router/src/directives/router_outlet.ts#L265-L291. It never uses the instance from where the router outlet is constructed in the DI tree.

liamdebeasi (approved):
Great job!

amandaejohnston (approved):

PR #25422, opened 2022-06-07
feat(datetime): add preferWheel support for date presentation

liamdebeasi:
## Pull request checklist

## Pull request type

 

 

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes

Issue URL: 

## What is the new behavior?

-
-
-

- [ ] Yes
- [ ] No

### DIFF
--- a/core/src/components/datetime/datetime.scss
+++ b/core/src/components/datetime/datetime.scss
@@ -58,21 +58,19 @@
  * the order we need to manually switch
  * the text alignment too.
  */
-:host .datetime-year .order-month-first .month-column {
-  order: 1;
+:host .wheel-order-year-first .day-column {
+ order: 3;
+
+ text-align: end;
 }
 
-:host .datetime-year .order-month-first .year-column {
-  order: 2;
-}
-
-:host .datetime-year .order-year-first .month-column {
+:host .wheel-order-year-first .month-column {
   order: 2;
 
   text-align: end;
 }
 
-:host .datetime-year .order-year-first .year-column {
+:host .wheel-order-year-first .year-column {
   order: 1;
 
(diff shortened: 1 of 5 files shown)


liamdebeasi:
Outstanding question: Say you have August 31st selected. You then change the month to June, which only has 30 days. What should happen to your previous "31st" selection? Currently, it resets you to the 1st, but not sure that's the correct behavior here. Maybe it should reset you to the nearest item?

sean-perkins (commented):
Ran locally and the primary behavior looks great, nice job!

Is this layout shifting something we need to address?

<video src="https://user-images.githubusercontent.com/13732623/172682426-e8344162-abaa-42ed-bea6-4e041eddc657.mp4"></video>

I believe it is happening because the max range (October) doesn't render double digits for the days column, so the width is different when changing between the available months.

sean-perkins:
I think it makes sense in the `ion-datetime` stylesheet with the selector you used. Putting that style within `ion-picker-column-internal` could be potentially confusing, especially if that component becomes the public API for developers. The min-width is more of a side effect of the dynamic nature of the data we are rendering for the date picker.

amandaejohnston:
> Outstanding question: Say you have August 31st selected. You then change the month to June, which only has 30 days. What should happen to your previous "31st" selection? Currently, it resets you to the 1st, but not sure that's the correct behavior here. Maybe it should reset you to the nearest item?

I think resetting to the nearest item would feel better -- if you're on the 31st, you probably intend to pick something late in the month anyway -- although I recall there being issues when we tried to implement that in the context of interactions with min/max. 🤔

amandaejohnston (approved):

PR #25424, opened 2022-06-07
fix(modal): status bar color now correct with sheet modal

liamdebeasi:
## Pull request checklist

## Pull request type

 

 

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

Issue URL: resolves https://github.com/ionic-team/ionic-framework/issues/20501

## What is the new behavior?

- When the card modal is open, the status bar will change to white text on a dark background. This also applies when dragging the card modal down then up.
- When the card modal is closed, the status bar will revert to whatever setting it was on before. This also applies when dragging the card modal down.

Note: This is a Capacitor 3 only feature. Feature detection was added so errors do not occur in Cordova/Capacitor 2.

Since this functionality is dependent on a native plugin I was not able to write a test for this. `6.1.7-dev.11653592708.12aae268` is a dev build for testing.

- [ ] Yes
- [x] No

### DIFF
--- a/core/src/components/modal/gestures/swipe-to-close.ts
+++ b/core/src/components/modal/gestures/swipe-to-close.ts
@@ -226,11 +226,11 @@ export const createSwipeToCloseGesture = (el: HTMLIonModalElement, animation: An
     if (clampedStep >= DISMISS_THRESHOLD && lastStep < DISMISS_THRESHOLD) {
       setStatusBarDefault();
 
-    /**
-     * However, if we swipe back up, then the
-     * status bar style should be set to have light
-     * text on a dark background.
-     */
+      /**
+       * However, if we swipe back up, then the
+       * status bar style should be set to have light
+       * text on a dark background.
+       */
     } else if (clampedStep < DISMISS_THRESHOLD && lastStep >= DISMISS_THRESHOLD) {
       setStatusBarDark();
     }
--- a/core/src/components/modal/gestures/swipe-to-close.ts
+++ b/core/src/components/modal/gestures/swipe-to-close.ts
@@ -4,7 +4,7 @@ import { isIonContent, findClosestIonContent } from '../../../utils/content';
 import type { GestureDetail } from '../../../utils/gesture';
 import { createGesture } from '../../../utils/gesture';
 import { clamp, getElementRoot } from '../../../utils/helpers';
-import { setStatusBarDark, setStatusBarDefault } from '../utils';
+import { setCardStatusBarDark, setCardStatusBarDefault } from '../utils';
 
 import { calculateSpringStep, handleCanDismiss } from './utils';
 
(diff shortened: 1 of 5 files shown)


amandaejohnston (approved):

sean-perkins (approved):

michielcrabbe96:
When you close the card modal, the status bar always reverts to the default style (not to the previous setting). So when the Status bar style is manually set to dark (but the default is the light style) it doesn't revert back to it on card modal close. Is this supposed to happen?

Nonetheless, this feature is an awesome addition!!

liamdebeasi:
Yes, this is intentional according to the Capacitor Docs for the `Default` status bar style: https://capacitorjs.com/docs/apis/status-bar#style. 

At the moment Capacitor does not track what the previous style was. If this is a behavior you need, I recommend opening a feature request on https://github.com/ionic-team/capacitor-plugins. 

As a workaround, you could wait for the `ionModalDidDismiss` event before manually setting the status bar style to `Dark`.

michielcrabbe96:
Okay, thank you for the explanation and tip.

EinfachHans:
@liamdebeasi The StatusBar Plugin has a [`getInfo`](https://github.com/ionic-team/ionic-framework/pull/25424#issuecomment-1156464642) function, can't this be used to fetch the previous setting before?

damida80:
> Yes, this is intentional according to the Capacitor Docs for the `Default` status bar style: https://capacitorjs.com/docs/apis/status-bar#style.
> 
> At the moment Capacitor does not track what the previous style was. If this is a behavior you need, I recommend opening a feature request on https://github.com/ionic-team/capacitor-plugins.
> 
> As a workaround, you could wait for the `ionModalDidDismiss` event before manually setting the status bar style to `Dark`.

Thanks for providing this. But I see and I have two problems here:
- If you have nested modals (you open another modal in the modal) and close the 2nd modal, the Statusbar is changed to Default (black text) and therefore to the wrong color.
- The workaround works of course, thanks for providing this, but you have to check everywhere, whether you are closing the modal on another modal. Additionally you get an ugly flickering, because the StatusBar changes back and forth then.
- If a user sets a darkmode preference manually within the App independently from the OS settings, the StatusBar style is also wrong after dismissing the modal.

torgnywalin:
For our project this fix can introduce a bug thats worst then missing the statusbar when showing a card modal.
For us there are two scenarios:
 - App supports light / dark mode and user can select from (light, dark, follow system)
 - App is forced to be in either light or dark, regardless of users device settings.

If we set the correct status bar at app startup, this will be overriden after user opens and closes first card modal.
This will happen if: 
 - user selects a theme in the app thats different from os theme.
 - our app has a dark header even when in light mode

Suggestions: 
 - Option for this integration with Capacitor to be disabled (I dont think Ionic components should make calls to capacitor plugins without it being really clear to the developer.) (Could be default enabled to be backward compatible)
 - Option for for what statusbar style to revert back to (if integration with capacitor is enabled). (Could be default to `default` to be backward compatible)

When working i ionic-vue having a global setting where one could send in a ref for what statusbar style to revert to would have been a great option.

liamdebeasi:
@torgnywalin Can you file an issue?

torgnywalin:
@liamdebeasi Thanks for quick reply. I tried to describe in this issue
https://github.com/ionic-team/ionic-framework/issues/26173

Issue #20524, opened 2020-02-18
feat: Optional I/O or On/Off label for ion-toggle


wilk-polarny:
**Ionic version:**
[x] **4.x**

**Describe the Feature Request**
I would like to have optional labels on `<ion-toggle>` elements - in iOS, you are still able to enable labels within the accessibility settings. We would like to honor these settings and to have a way to enable labels for the respective platforms (iOS/Android).

**Describe Preferred Solution**
How? An optional parameter/directive would be nice.

It should look like this:
![image](https://user-images.githubusercontent.com/819238/74693298-8a704780-51eb-11ea-8052-98c73b78c3f5.png)
(Source: https://osxdaily.com/2014/04/15/ios-settings-switch-on-off-labels/)

As far as I know, there are also Android equivalents with on/off.

**Describe Alternatives**
I considered a DIY solution, but I'm not currently not sure how to achieve that effect without having to fork the Ionic project.

**Related Code**
```html
<ion-toggle label="false"></ion-toggle>
<ion-toggle label="true"></ion-toggle>

```

**Additional Context**
None

tawfiek:
I am working on this feature

brandyscarney:
Thanks for the issue! Here's an example of how it could be done now using iOS: https://codepen.io/brandyscarney/pen/YzyxeXa

```css
.ios.toggle-labels::before {
  position: absolute;

  top: 11px;
  right: 7px;

  content: "";

  width: 8px;
  height: 8px;

  border: solid 1px rgba(0, 0, 0, 0.23);

  border-radius: 50%;
}

.ios.toggle-labels::after {
  position: absolute;

  top: 12px;
  left: 11px;
  
  content: "";

  width: 10px;
  height: 10px;

  border-left: 1px solid #fff;
}
```

I like the idea of adding this for accessibility reasons, but we will need to think through how this should be handled in Material Design for Android. For example, if we add a `label` property to toggle, should it add the words "on/off" in `md` mode? If so, should the text for these be customizable? I would think it would need to be in order to support other languages. I'll need to discuss this with the team to determine how the API should look if we decide to add it as a feature, but the code above should get you very close to a native look for iOS.

liamdebeasi:
Thanks for the issue. This has been resolved via https://github.com/ionic-team/ionic-framework/pull/25459 and the feature will be available in an upcoming release of Ionic.

PR #25459, opened 2022-06-13
feat(toggle): on/off icons for toggle

sean-perkins:
## Pull request checklist

## Pull request type

 

 

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes

The Toggle component does not support the accessibility indications for the toggle checked state.

Issue URL: #20524

## What is the new behavior?

Introduces the `switchLabelsEnabled` property to `ion-toggle`, that developers can use to enable the icon indicators for the toggle's checked state.

For Material Design, the checkbox icon will display when the toggle is checked. When the toggle is unchecked, the minus icon will display.

For iOS, the "I" icon will display when the toggle is checked. When the toggle is unchecked, the "O" icon will display. iOS renders both icons in the DOM, in order to accomplish the pulling effect when the toggle changes state. 

- [ ] Yes
- [x] No

## Other information

Dev-build: `6.1.10-dev.11655149274.1a5ec2d4`


core/src/components/toggle/toggle.tsx
```diff
@@ -204,7 +224,15 @@ export class Toggle implements ComponentInterface {
       >
         <div class="toggle-icon" part="track">
           <div class="toggle-icon-wrapper">
-            <div class="toggle-inner" part="handle" />
+            {switchLabelsEnabled && mode === 'ios' && (
+              <ion-icon class="toggle-switch-icon checked" icon={this.getSwitchLabelIcon(true)}></ion-icon>
+            )}
+            <div class="toggle-inner" part="handle">
+              {switchLabelsEnabled && mode === 'md' && this.renderOnOffSwitchLabels()}
```
liamdebeasi, line 231: Optional: We can pass in the mode to both `getSwitchLabelIcon` and `renderOnOffSwitchLabels` so we do not need to re-call it again each time the function is called.

liamdebeasi, line 231: Also any reason why this is extracted to a function but the iOS icons are written inline?

core/src/components/toggle/toggle.tsx
```diff
@@ -63,6 +64,11 @@ export class Toggle implements ComponentInterface {
    */
   @Prop() value?: string | null = 'on';
 
+  /**
+   * Enables the on/off accessibility switch labels within the toggle.
+   */
+  @Prop() switchLabelsEnabled = false;
```
liamdebeasi, line 70: APIs should be present tense according to our style guide.

Maybe something like `enableLabels`, `enableOnOffLabels`, or `enableSwitchLabels`? I am partial to "enableOnOffLabels" as there is no confusion as to whether or not this prop impacts `ion-label`.

liamdebeasi: What do you think about defaulting this to `undefined`? If Capacitor adds the ability to automatically detect the I/O preference in the device settings, we could support that functionality without breaking changes as developers would be able to opt-out of this functionality by explicitly setting `false`. If we default to `false`, then having the feature be enabled by default is harder to do without breaking changes.

I know the design doc states it should default to `false`, so feel free to disregard. Just might make life a bit easier down the road if we support that new functionality.

liamdebeasi (changes requested):
The "-" does not show up in dark mode: 
image

Also the checkmark text appears black. We should probably add some screenshot tests for dark mode.

liamdebeasi (commented):
Also the I/O label supposed to appear, not slide in on iOS

| native | ionic |
| ------ | ----- |
| <video src="https://user-images.githubusercontent.com/2721089/173444653-83815949-7e9a-477b-844e-821d31dea3cf.MP4"></video> | <video src="https://user-images.githubusercontent.com/2721089/173444693-e6c6cf1d-843a-4f8c-a269-086b0cf57269.MP4"></video> |

----

Somewhat related, but when dragging the toggle the "O" label overlaps the handle

| native | ionic |
| ------ | ----- |
| <video src="https://user-images.githubusercontent.com/2721089/173444959-664ed21f-0cf5-4d5d-b9c8-c00599e02111.MP4"></video> | <video src="https://user-images.githubusercontent.com/2721089/173444991-cbafd9ca-6db8-4b80-bf76-cf723f3aad9f.MP4"></video> |

core/src/components/toggle/toggle.tsx
sean-perkins: I think defaulting to undefined makes sense here, I'll update the design doc.

sean-perkins:
Pushed an update to fix the colors for iOS to align with the spec. This does cause an issue with `ion-toggle` when `color="dark"` and the toggle is checked:

Screen Shot 2022-06-13 at 5 26 13 PM

Unsure if we should concern ourself with overriding the text color for this, since it isn't within the iOS spec.

I'll need to do some more discovery into rendering of the icons. Originally when just having the icon flip on iOS (like it does for MD), it made the animation of the knob look choppy. It looks like when the knob is active, I need to set the opacity of the icons to zero (and make sure the icons have a transition for opacity).

liamdebeasi:
When rendering the on icon over an element with an Ionic color, it might be easiest to use the contrast (ex: `var(--ion-color-primary-contrast)` color. That should work for both light and dark mode.

For the off text, you might be able to use a step color? The color would change slightly across light vs. dark mode, but I don't know if it really matters (as long as the text is accessible). A benefit of using a stepped color is it is easy for developers to customize on a global level. Full customization is out of scope for this task, but using a stepped color would at least let developers customize the text color without much hassle. (For example, if a developers customized the track color, our default color may not be accessible when used on the new track color)

amandaejohnston (changes requested):

amandaejohnston (approved):
Nice work!

core/src/components/toggle/toggle.ios.vars.scss
```diff
@@ -75,3 +75,9 @@ $toggle-ios-item-end-padding-bottom:   5px !default;
 
 /// @prop - Padding start of the toggle positioned on the end in an item
 $toggle-ios-item-end-padding-start:    $item-ios-padding-start !default;
+
+/// @prop - The text color of the on/off labels
+$toggle-ios-on-off-label-color: #b0b0b0 !default;
```
liamdebeasi, line 80: The color on light mode is not accessible:

Screen Shot 2022-06-15 at 2 09 01 PM

Dark mode seems fine though 

Screen Shot 2022-06-15 at 2 10 09 PM

core/src/components/toggle/toggle.md.vars.scss
```diff
@@ -99,3 +99,9 @@ $toggle-md-item-end-padding-bottom:               12px !default;
 
 /// @prop - Padding start of the toggle positioned on the end in an item
 $toggle-md-item-end-padding-start:                $item-md-padding-start !default;
+
+/// @prop - The text color of the on/off labels
+$toggle-md-on-off-label-color: #000 !default;
```
liamdebeasi, line 104: These colors seem good -- we will likely need to expose a way to customize these colors in the future though

sean-perkins:
The change from the `sizes` test is from this change:
```css
.toggle-icon-wrapper {
  align-items: center;
}
```

|Before|After|
|---|---|
|Screen Shot 2022-06-15 at 4 08 25 PM| Screen Shot 2022-06-15 at 4 08 06 PM|

I can see the argument for the before, since the shadow is contained within the toggle. I'll explore alternative patterns to vertically align the on/off label icons without using flex box.

sean-perkins:
Ok, I think this is finally ready for another review. Was getting burned by trying to nest our SASS mixins:
```scss
@include ltr() {
  @include position(x, x, x, x);
}
```
Instead of just using the positioning values directly and disabling stylelint.

core/src/components/toggle/toggle.md.vars.scss
liamdebeasi: How does the color get set for the "O" on iOS mode?

This definitely some scope creep, so feel free to ignore this for now... But what if we had a global css var for this?

For example, tab bar has a default background that can be customized with `--ion-tab-bar-background`: https://github.com/ionic-team/ionic-framework/blob/1b1b1a3800c4d044b4a3e7418f534e9271770ec6/core/src/themes/ionic.theme.default.md.scss#L17

Reason why I bring it up is iOS does change the on/off text color when adding high contrast mode.

core/src/components/toggle/toggle.tsx
```diff
@@ -178,8 +184,29 @@ export class Toggle implements ComponentInterface {
...
+  };
+
+  private renderOnOffSwitchLabels(mode: Mode, checked: boolean) {
+    const icon = this.getSwitchLabelIcon(mode, checked);
+
+    return (
+      <ion-icon
+        class={{
+          'toggle-switch-icon': true,
+          'toggle-switch-icon--checked': checked,
```
liamdebeasi, line 201: Minor thing, but this convention does not appear to be used anywhere else in the framework. Can we make this decision as a team before adopting this syntax?

liamdebeasi (commented):
The "O" doesn't seem to be centered

image

sean-perkins:
Screen Shot 2022-07-05 at 3 30 48 PM

The "O" label appears to be vertically centered. On iOS it is not horizontal aligned (it appears to be offset further to the left than the right). The screenshot is at 150% zoom.

Your screenshot appears to draw the SVG asset thicker 🤔

liamdebeasi (commented):
Another thought: It might be better to do some more design work around this CSS Variable idea we've been playing with. I'd be fine merging this feature in for now, and then we can figure out the CSS Variable stuff next sprint. Does that work for you? I think this feature is great and pretty much done, so I don't think it makes sense to hold this up on the CSS Variable detail.

sean-perkins:
Yeah I can revert the CSS variable changes and leave them static for now.

liamdebeasi (approved):
Much better. Great job!

Issue #25483, opened 2022-06-15
bug: Item with multiple inputs including datetime renders incorrectly in RTL mode


sean-perkins:
### Prerequisites

- [X] I have read the [Contributing Guidelines](https://github.com/ionic-team/ionic-framework/blob/main/.github/CONTRIBUTING.md#creating-an-issue).
- [X] I agree to follow the [Code of Conduct](https://ionicframework.com/code-of-conduct).
- [X] I have searched for [existing issues](https://github.com/ionic-team/ionic-framework/issues) that already report this problem, without success.

### Ionic Framework Version

- [ ] v4.x
- [ ] v5.x
- [X] v6.x
- [ ] Nightly

### Current Behavior

Only in RTL mode, when rendering an `ion-item` with multiple inputs, such as a checkbox slotted to the start and end and a datetime in the primary content; will clip the datetime in RTL mode only.

![Screen Shot 2022-06-15 at 7 18 44 PM](https://user-images.githubusercontent.com/13732623/173957900-9e62c1a5-dc59-4012-957b-9d9212fa6c1d.png)

### Expected Behavior

Datetime should not be clipped in RTL mode.

![Screen Shot 2022-06-15 at 7 19 22 PM](https://user-images.githubusercontent.com/13732623/173957963-686589d6-18dc-43cb-abeb-09feb2bf36db.png)

### Steps to Reproduce

1. Open reproduction app
2. Observe: `ion-datetime` is clipped
3. Change `dir="rtl"` to `dir="ltr"`
4. Wait for Stackblitz to reload
5. Observe: Datetime is not clipped

Ignore datetime months changing rapidly within stackblitz with this bug. Due to the width of the datetime being clipped, the intersection observer can fire many times.

### Code Reproduction URL

https://stackblitz.com/edit/ztbrct?file=index.html,package.json

### Ionic Info

N/A

sean-perkins:
Resolved with: https://github.com/ionic-team/ionic-framework/pull/25484

Will be available in the next release of Ionic.

PR #25479, opened 2022-06-15
test(item): migrate tests to playwright

sean-perkins:
## Pull request checklist

Please check if your PR fulfills the following requirements:
- [x] Tests for the changes have been added (for bug fixes / features)
- [ ] Docs have been reviewed and added / updated if needed (for bug fixes / features)
  - Some docs updates need to be made in the `ionic-docs` repo, in a separate PR. See the [contributing guide](https://github.com/ionic-team/ionic-framework/blob/main/.github/CONTRIBUTING.md#modifying-documentation) for details.
- [x] Build (`npm run build`) was run locally and any changes were pushed
- [x] Lint (`npm run lint`) has passed locally and any fixes were made for failures

## Pull request type

 

 

Please check the type of change your PR introduces:
- [ ] Bugfix
- [ ] Feature
- [ ] Code style update (formatting, renaming)
- [ ] Refactoring (no functional changes, no api changes)
- [ ] Build related changes
- [ ] Documentation content changes
- [x] Other (please describe): Test migrations

## What is the current behavior?

Item tests are written in Puppeteer. 

Issue URL:  N/A

## What is the new behavior?

Migrates tests to Playwright

- Removes flaky exception fallback test for item counter
- Introduces new option to `setIonViewport` that allows resizing the horizontal size of the viewport to match the scroll contents (for niche tests like the form test for item). 

## Does this introduce a breaking change?

- [ ] Yes
- [x] No

PR #25484, opened 2022-06-15
fix(item): multiple input appearance when using datetime

sean-perkins:
## Pull request checklist

## Pull request type

 

 

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

Discovered when migrating item tests: https://github.com/ionic-team/ionic-framework/pull/25479

Only in Google Chrome, when using `ion-datetime` with multiple inputs in an `ion-item` in RTL mode; the datetime will become clipped.

Issue URL: #25483

## What is the new behavior?

- Explicitly sets the top/left position styles instead of using the RTL utility that incorrectly sets styles on `host-context`.

- [ ] Yes
- [x] No

## Other information

You can confirm this change locally with the existing `item/test/inputs` suite in RTL mode: http://localhost:3333/src/components/item/test/inputs?rtl=true


sean-perkins:
Sounds great! Thanks for pointing me to where it's being appended in our code. At one point I was able to narrow it to:
```css
:host .calendar-day:after {
 @include position(50%, null, null, 50%);
}
```
But didn't consider that the mixin may have been incorrectly applying the styles. Happy to look more into it if other tasks require your time 👍

liamdebeasi:
Turns out we don't even need the RTL support in https://github.com/ionic-team/ionic-framework/blob/main/core/src/components/datetime/datetime.scss#L346. The position+transform styles in the block are fixes for an iOS 13 bug. I tested it out and just doing `top: 50%` and `left: 50%` seem to work fine.

We can probably remove the `@include` and add the top/left for now and add a tech debt ticket to remove the top/left/transform styles when iOS 13 support is dropped.

We should also look into getting https://github.com/ionic-team/ionic-framework/pull/25264 ready for review. However, there are likely more instances of this issue that we have not seen yet. When fixing this it might be good to get a list of affected components. We can probably get a solid list by searching for `@include` in the code, and then filtering out any selectors that either a) only have `:host` or b) do not use `:host` in the selector.

core/src/components/datetime/datetime.scss
```diff
@@ -341,9 +341,12 @@
    * Explicit position values are required here
    * as pseudo element positioning is incorrect
    * in older implementations of css grid.
+   *
+   * TODO: FW-1720: Remove top/left styles when depreciating iOS 13 support
```
liamdebeasi, line 345:
```suggestion
   * TODO: FW-1720: Remove top/left styles when deprecating iOS 13 support
```

core/src/components/datetime/datetime.scss
```diff
@@ -341,9 +341,12 @@
...
    */
-
-  @include position(50%, null, null, 50%);
+  top: 50%;
```
liamdebeasi, line 347:
```suggestion
  /* stylelint-disable-next-line property-disallowed-list */
  top: 50%;
```

liamdebeasi (approved):

PR #25490, opened 2022-06-16
fix(vue): adding class to IonPage no longer hides component

liamdebeasi:
## Pull request checklist

## Pull request type

 

 

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

Issue URL: Internal ticket

Dynamically adding a class to `IonPage` causes the component to re-render. This means that `IonPage` components used inside of `IonRouterOutlet` would get the `.ion-page-invisible` class re-added, even if it was removed: https://github.com/ionic-team/ionic-framework/blob/main/packages/vue/src/components/IonPage.ts#L13

This resulted in the `IonPage` component hiding when a custom class is added. Also, the custom class showed up twice because we re-added it in `existingClasses`: https://github.com/ionic-team/ionic-framework/blob/main/packages/vue/src/components/IonPage.ts#L20

## What is the new behavior?

- Removed `existingClasses`, so the custom class only shows up once.
- Tweaked how we set `ion-page-invisible`. This class is now added in `IonRouterOutlet` (as it is only needed when using routing) when the `IonPage` component is mounted initially.
- The private `isInOutlet` property on `IonPage` was used to determine when to add `ion-page-invisible`. As this class  is now added by `IonRouterOutlet`, the `isInOutlet` property is no longer needed and has been removed.

- [ ] Yes
- [x] No

### DIFF
--- a/packages/vue/src/components/IonPage.ts
+++ b/packages/vue/src/components/IonPage.ts
@@ -9,7 +9,7 @@ export const IonPage = /*@__PURE__*/ defineComponent({
   mounted() {
     this.$props.registerIonPage(this.$refs.ionPage);
   },
-  setup(props, { attrs, slots }) {
+  setup(_, { attrs, slots }) {
     return () => {
       return h(
         'div',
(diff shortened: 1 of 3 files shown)


sean-perkins (approved):
This looks good to me.

PR #25510, opened 2022-06-21
test(img): update screenshot capture 

sean-perkins:
## Pull request checklist

## Pull request type

 

 

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes
- [x] Other (please describe): Tests

Screenshots are clipped and `ion-img` renders inconsistently when capturing compare screenshots.

Issue URL: N/A

## What is the new behavior?

- Waits for the images to load prior to capturing the screenshot
- Adjusts styling around `ion-img` within the test to prevent clipping
- Error state is now captured in screenshot output 

- [ ] Yes
- [x] No


core/src/components/img/test/basic/img.e2e.ts
```diff
@@ -5,8 +5,12 @@ test.describe('img: basic', () => {
   test('should not have visual regressions', async ({ page }) => {
     await page.goto('/src/components/img/test/basic');
 
+    const ionImgDidLoad = await page.spyOnEvent('ionImgDidLoad');
+
     await page.setIonViewport();
 
+    await ionImgDidLoad.next();
```
liamdebeasi, line 12: There are multiple ion-img elements on this page, and this would only wait for the first occurrence of the event. It might be better to simplify the E2E test to only have on `ion-img` (or you could wait for each of the events to fire).

sean-perkins: I believe the history around why the test template is set-up the way it is, is to check the behavior of intersection observers, to determine when the image should be loaded.

With the first `ion-img` being initially visible, the load event actually fires before the spy can be created. So doing something like:

```ts
const ionImgDidLoad = await page.spyOnEvent('ionImgDidLoad');

await ionImgDidLoad.next();
await page.setIonViewport();
```

times out, since the event is not re-emitted.

However, when doing:

```ts
const ionImgDidLoad = await page.spyOnEvent('ionImgDidLoad');

await page.setIonViewport();
await ionImgDidLoad.next();
```

The viewport is resized, all of the `ion-img` elements are now intersected and the second `ion-img` fires the load event, resolving the spy.

I can update test to check that the first image is loaded, before setting up the spy, if we feel that gives better coverage:
```ts
const ionImgDidLoad = await page.spyOnEvent('ionImgDidLoad');

const img = await page.locator('.img-part img');
await img.evaluate((el: HTMLImageElement) => el.complete);

await page.setIonViewport();

await ionImgDidLoad.next();
```

Confirmed locally this still passes the test suite.

Issue #24881, opened 2022-03-04
bug: datetime, setting value out of bounds prevents user from navigating in bounds 


Yura13:
### Prerequisites

- [X] I have read the [Contributing Guidelines](https://github.com/ionic-team/ionic-framework/blob/main/.github/CONTRIBUTING.md#creating-an-issue).
- [X] I agree to follow the [Code of Conduct](https://ionicframework.com/code-of-conduct).
- [X] I have searched for [existing issues](https://github.com/ionic-team/ionic-framework/issues) that already report this problem, without success.

### Ionic Framework Version

- [ ] v4.x
- [ ] v5.x
- [X] v6.x

### Current Behavior

![datepicker-min-max-attributes](https://user-images.githubusercontent.com/17030431/156815324-48e1e344-85c3-45b7-a177-768d373b662a.gif)

### Expected Behavior

I should have a possibility to navigate to the available months.

### Steps to Reproduce

I am trying to edit the saved date but can't navigate to the next month.
Saved date is 2021-12-12
Min date is 2022-03-04
Max date is 2022-04-03

### Ionic Info

Ionic:

   Ionic CLI                     : 6.18.1 (C:\Users\Xiaomi\AppData\Roaming\nvm\v16.13.2\node_modules\@ionic\cli)
   Ionic Framework               : @ionic/angular 6.0.10
   @angular-devkit/build-angular : 13.1.3
   @angular-devkit/schematics    : 13.1.3
   @angular/cli                  : 13.1.3
   @ionic/angular-toolkit        : 5.0.3

Cordova:

   Cordova CLI       : 10.0.0 (cordova-lib@10.1.0)
   Cordova Platforms : android 9.1.0, browser 6.0.0
   Cordova Plugins   : cordova-plugin-ionic-keyboard 2.2.0, cordova-plugin-ionic-webview 5.0.0, cordova-plugin-ionic 5.5.1, (and 28 other plugins)

Utility:

   cordova-res : not installed globally
   native-run  : 1.5.0

System:

   Android SDK Tools : 26.1.1 (C:\Android\android-sdk)
   NodeJS            : v16.13.2 (C:\Program Files\nodejs\node.exe)
   npm               : 8.1.2
   OS                : Windows 10

liamdebeasi:
Thanks for the issue. The current expectation is that developers must provide a valid date to the `value` property. Given the `max` and `min` values you provided, `2021-12-12` is not a valid value as it is out of bounds.

[...]

PR #25513, opened 2022-06-21
fix(datetime): add dev warnings when setting out of bounds value

amandaejohnston:
## Pull request checklist

## Pull request type

 

 

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

Originally, assigning a `value` to `ion-datetime` that's sufficiently earlier than the `min` or later than the `max` would lead to a disabled month being displayed, and the user being unable to navigate in either direction. This was mostly resolved here https://github.com/ionic-team/ionic-framework/pull/25351 since both assigning a value and leaving it to the default go through the `processValue` function. The dev experience just needs some polish.

Issue URL: https://github.com/ionic-team/ionic-framework/issues/24881

## What is the new behavior?

Console warnings for out-of-bounds values are now printed when either assigning an initial `value`, or updating it programmatically.

- [ ] Yes
- [x] No

## Other information

Tests for console warnings have been flaky in the past, so I didn't include one. I did add a test to ensure an in-bounds month is displayed, though, since the existing tests only cover datetimes with no initial value.

### DIFF
--- a/core/src/components/datetime/test/minmax/datetime.e2e.ts
+++ b/core/src/components/datetime/test/minmax/datetime.e2e.ts
@@ -109,7 +109,7 @@ test.describe('datetime: minmax', () => {
     });
   });
 
-  test.only('setting value outside bounds should show in-bounds month', async ({ page }) => {
+  test('setting value outside bounds should show in-bounds month', async ({ page }) => {
     await page.setContent(`
       <ion-datetime min="2021-06-01" max="2021-06-30" value="2021-05-01"></ion-datetime>
     `);
--- a/core/src/components/datetime/test/minmax/datetime.e2e.ts
+++ b/core/src/components/datetime/test/minmax/datetime.e2e.ts
@@ -1,4 +1,5 @@
 import { expect } from '@playwright/test';
+import type { E2EPage } from '@utils/test/playwright';
 import { test } from '@utils/test/playwright';
 
 test.describe('datetime: minmax', () => {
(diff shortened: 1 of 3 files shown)


sean-perkins (approved):
Tested locally and it does exactly what it needs to do. Great work 👍 I agree that adding tests for the contents of the warning message has been flaky and does not introduce value here.

~ Would there be any value to the developer if the warning message included the values back to them? For example:
> [Ionic Warning]: The value (2021-05-01) passed to ion-datetime is earlier than the min (2021-06-01) or later than the max (2021-10).

or:

> [Ionic Warning]: The value passed to ion-datetime is earlier than the min or later than the max. { min: "2021-06-01", value: "2021-05-01", max: "2021-06-30" }

core/src/components/datetime/test/minmax/datetime.e2e.ts
```diff
@@ -105,4 +108,15 @@ test.describe('datetime: minmax', () => {
       expect(await minutes.count()).toBe(60);
     });
   });
+
+  test('setting value outside bounds should show in-bounds month', async ({ page }) => {
```
liamdebeasi, line 112: Can we add tests for when only the `min` + `value` is defined and another for when only the `max` + `value` is defined?

liamdebeasi (approved):
Good to go once the comment below is addressed

amandaejohnston:
I like that multiline formatting 👀 I'll pull it into a separate util too.

amandaejohnston:
I went with logging the min/max/value as stringified objects, so we don't have to worry about date formatting across locales.

Issue #24674, opened 2022-01-28
feat: ion-datetime with multiple date selection


Saqib92:
### Prerequisites

- [X] I have read the [Contributing Guidelines](https://github.com/ionic-team/ionic-framework/blob/main/.github/CONTRIBUTING.md#creating-an-issue).
- [X] I agree to follow the [Code of Conduct](https://ionicframework.com/code-of-conduct).
- [X] I have searched for [existing issues](https://github.com/ionic-team/ionic-framework/issues) that already include this feature request, without success.

### Describe the Feature Request

There should be an option to select multiple dates or Date range to select Multiple dates. 

### Describe the Use Case

Use case might be a travel related application. like if someone wants to reserve hotel room he can select a range of multiple dates. 

### Describe Preferred Solution

New option to pass in properties. `[range]="true"` and for multiple different dates : `[multiple]="true"` and default `false` for both. 

### Describe Alternatives

https://github.com/hsuanxyz/ion2-calendar

### Related Code

`<ion-datetime [range]="true"></ion-datetime>`

### Additional Information
<img width="400" alt="calendar-1" src="https://user-images.githubusercontent.com/7135885/151526011-6b3e6b80-da1c-4dc8-bf00-375464059617.png">

Can take range example from here:
https://github.com/hsuanxyz/ion2-calendar

amandaesmith3:
Thanks for the feature request. Selecting a range of dates is a duplicate of https://github.com/ionic-team/ionic-framework/issues/23572, but we can use this to track selecting multiple, separate dates. I discussed with the team and we agree this feature would be a good fit for Ionic 👍

liamdebeasi:
Thanks for the issue. This has been resolved via https://github.com/ionic-team/ionic-framework/pull/25514, and this feature should be available in an upcoming minor release of Ionic.

PR #25514, opened 2022-06-21
feat(datetime): add multiple date selection

amandaejohnston:
## Pull request checklist

## Pull request type

 

 

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes

Only one date can be selected at a time.

Issue URL: Resolves https://github.com/ionic-team/ionic-framework/issues/24674

## What is the new behavior?

New `multiple` prop added, allowing for multiple date selection when `presentation="date"`. See internal design doc for full description of functionality.

- [ ] Yes
- [x] No

## Other information

Dev build: `6.1.14-dev.11657720710.1df41c89`
PR for playground: https://github.com/ionic-team/ionic-docs/pull/2442


packages/vue/src/vue-component-lib/utils.ts
```diff
@@ -1,7 +1,7 @@
 import { VNode, defineComponent, getCurrentInstance, h, inject, ref, Ref } from 'vue';
 
-export interface InputProps {
```
sean-perkins, line 3: You will need to install dependencies for `core` to be up-to-date with the latest patch from output targets. Then re-run `npm run build`. This change shouldn't add `extends Object` back in.

core/src/components/datetime/test/multiple/datetime.e2e.ts
```diff
@@ -0,0 +1,159 @@
+import { expect } from '@playwright/test';
+import type { E2EPage } from '@utils/test/playwright';
+import { test } from '@utils/test/playwright';
+
+const setup = async (page: E2EPage, datetimeID: string, shouldNavigate = true) => {
+  if (shouldNavigate) await page.goto('/src/components/datetime/test/multiple/');
+  const datetime = page.locator(`#${datetimeID}`);
+  await datetime.scrollIntoViewIfNeeded();
+  await page.waitForSelector(`#${datetimeID}.datetime-ready`);
+  return page.locator(`#${datetimeID}`);
```
sean-perkins, line 10: ~ Can this return `datetime`?

core/src/components/datetime/datetime.tsx
```diff
@@ -308,57 +308,78 @@ export class Datetime implements ComponentInterface {
...
   /**
    * Update the datetime value when the value changes
    */
   @Watch('value')
   protected valueChanged() {
+    const { value, minParts, maxParts, workingParts, multiple } = this;
+
     if (this.hasValue()) {
+      if (!multiple && Array.isArray(value)) {
+        this.value = value[0];
```
sean-perkins, line 332: Is this a safe guard for if someone passes an array to `value` when `multiple` is not true? 

This causes a weird state inconsistency between the developer's state (`["someString"]`) and the value of the datetime element `"someString"`). This also does not appear to emit `ionChange`, so the developer is unable to get their state in-sync with this.

If this is a safe guard, I think we should throw a warning instead.

core/src/components/datetime/datetime.tsx
```diff
@@ -308,57 +308,78 @@ export class Datetime implements ComponentInterface {
...
       } else {
-        printIonWarning(`Unable to parse date string: ${this.value}. Please provide a valid ISO 8601 datetime string.`);
+        printIonWarning(`Unable to parse date string: ${value}. Please provide a valid ISO 8601 datetime string.`);
       }
     }
 
     this.emitStyle();
     this.ionChange.emit({
-      value: this.value,
+      value: value,
```
sean-perkins, line 382: ~ Can use shorthand syntax:
```ts
this.ionChange.emit({ value });
```

core/src/components/datetime/datetime.tsx
```diff
@@ -459,27 +480,41 @@ export class Datetime implements ComponentInterface {
...
-       * Important: Take the timezone offset based on
-       * the date that is currently selected, otherwise
-       * there can be 1 hr difference when dealing w/ DST
-       */
-      const date = new Date(convertDataToISO(this.activeParts));
-      this.activeParts.tzOffset = date.getTimezoneOffset() * -1;
+    if (highlightActiveParts || !isCalendarPicker) {
+      const activePartsIsArray = Array.isArray(activeParts);
+      if (activePartsIsArray && activeParts.length === 0) {
+        this.value = undefined;
```
sean-perkins, line 495: What is the reasoning for setting value to undefined here?

core/src/components/datetime/datetime.tsx
```diff
@@ -528,10 +563,49 @@ export class Datetime implements ComponentInterface {
...
+      /**
+       * We read from activePartsClone here because valueChanged() only updates that,
+       * so it's the more reliable source of truth. If we read from activeParts, then
+       * if you click July 1, manually set the value to July 2, and then click July 3,
+       * the new value would be [July 1, July 3], ignoring the value set.
+       *
+       * We can then pass the new value to activeParts (rather than activePartsClone)
+       * since the clone will be updated automatically by activePartsChanged().
+       */
+      const activePartsArray = Array.isArray(activePartsClone) ? activePartsClone : [activePartsClone];
```
sean-perkins, line 579: ~ We repeat this similar manipulation in multiple places. Would it make sense to extract this behavior in a util that all the implementations share?

```ts
const activePartsArray = valueToArray(activePartsClone);
```

```ts
const valueToArray = (data: string | string[]) => Array.isArray(data) ? data : [data];
```

core/src/components/datetime/datetime.tsx
```diff
@@ -1386,8 +1488,9 @@ export class Datetime implements ComponentInterface {
             ...findPart,
           });
 
+          // wheel pickers don't support multiple selection, so assume this won't be an array
           this.setActiveParts({
-            ...this.activeParts,
+            ...(this.activeParts as DatetimeParts),
```
sean-perkins, line 1493: ~ Typescript should be able to automatically narrow the type if we add checks that validate that `activeParts` is not an array. That would be my preference over type casting, which doesn't add functional safety if somehow it is set to an array in the implementation.

core/src/components/datetime/datetime.tsx
```diff
@@ -1459,7 +1566,7 @@ export class Datetime implements ComponentInterface {
         class="day-column"
         color={this.color}
         items={days}
-        value={workingParts.day || this.todayParts.day}
+        value={'' + (workingParts.day || this.todayParts.day)}
```
sean-perkins, line 1569: What is the reasoning behind this change? Did you receive an error that a number could not be assigned to `value`?

core/src/components/datetime/datetime.tsx
```diff
@@ -1415,7 +1518,11 @@ export class Datetime implements ComponentInterface {
 
     if (isDateEnabled) {
       days = days.map((dayObject) => {
-        const referenceParts = { month: workingParts.month, day: dayObject.value, year: workingParts.year };
+        const referenceParts = {
+          month: workingParts.month,
+          day: dayObject.value,
+          year: workingParts.year,
+        } as DatetimeParts;
```
sean-perkins, line 1525: Do we cast after the value since `referenceParts` doesn't satisfy the full type signature?
i.e.: 
```ts
const referenceParts: DatetimeParts = {};
// vs.
const referenceParts = {} as DatetimeParts;
```

core/src/components/datetime/datetime.tsx
```diff
@@ -1474,7 +1581,7 @@ export class Datetime implements ComponentInterface {
           });
 
           this.setActiveParts({
-            ...this.activeParts,
+            ...(this.activeParts as DatetimeParts),
```
sean-perkins, line 1584: Similar thoughts here around type narrowing. If at some point a feature sets `activeParts` to an array by accident, type checking will not catch the error since we are type casting.

core/src/components/datetime/datetime.tsx
```diff
@@ -2056,7 +2169,8 @@ export class Datetime implements ComponentInterface {
     const hasWheelVariant =
       (presentation === 'date' || presentation === 'date-time' || presentation === 'time-date') && preferWheel;
 
-    renderHiddenInput(true, el, name, value, disabled);
+    const formattedValue = Array.isArray(value) ? JSON.stringify(value) : value;
```
sean-perkins, line 2172: ~ Comparing against Select, the implementation abstracts this behavior to a `parseValue` function. Also interestingly it uses `.join(',')` instead of `JSON.stringify`, I could not find reasoning behind this.

It may make sense to make the implementations use similar/same language, so that when later working with these components they have the same implementation style.

sean-perkins (changes requested):

core/src/components/datetime/datetime.tsx
amandaejohnston: If no dates are selected, the value is set to `undefined` per the design doc: https://www.notion.so/ionic/ion-datetime-multiple-date-selection-63c95d300d2a42a79d956b3c4e3244b1#e80ea660462541c2a2e8a51a1ab157b9 The intention was to keep in parity with a single-selection datetime that has nothing selected due to not being inited with a default value.

core/src/components/datetime/datetime.tsx
amandaejohnston: I could, though it's only done here and one other time in `state.ts`; I'm not sure the extra layer of logic is worth the minimal code length difference 🤔

core/src/components/datetime/datetime.tsx
amandaejohnston: On closer inspection, the actual reason for the error I was getting is that `DatetimeParts.day` can be `null`, but value is `string | number | undefined` (but not `null`). I'll finagle it a different way 👀

core/src/components/datetime/datetime.tsx
amandaejohnston: Huh, I'm not sure why we use `join` instead of `stringify` there but it would be good to have the two match up. 🤔 Otherwise, `datetime` will have the `[]` brackets around the values, while `select` doesn't. Good catch!

amandaejohnston: Done 👍 I used `formatValue` instead of `parseValue` since `ion-datetime` tends to use `parse` language to signify manipulating date parts specifically.

core/src/components/datetime/datetime.tsx
sean-perkins: Makes sense, thanks for the explanation 👍

core/src/components/datetime/datetime.tsx
sean-perkins: Ahh I see now. If you haven't selected a date and you click to confirm, the value should be emitted as `undefined`, instead of trying to convert the active parts into an ISO date string. 

This makes sense, I missed that part originally 👍

core/src/components/datetime/datetime.tsx
```diff
@@ -308,58 +308,77 @@ export class Datetime implements ComponentInterface {
    */
   @Prop() firstDayOfWeek = 0;
 
+  /**
+   * If `true`, multiple dates can be selected at once. Only
+   * applies to `presentation="date"` and `preferWheel="false"`.
+   */
+  @Prop() multiple = false;
```
sean-perkins, line 315: ~ Noticed that 1/2 of our implementations have a default value and the other set the value to `false`. Unsure if any action is needed here, but we should probably land on consistency in the next major.

For example:
- AccordionGroup, Input - multiple is unset
- Select - multiple defaults to false 
 
This does affect the type signatures that are generated, but I am unsure of the impact of the field being marked as optional vs. required from the devs perspective.

sean-perkins (commented):
General questions around the experience:

After deselecting a date that was previous selected, the day remains active, if you hover away to a different day:
Screen Shot 2022-07-15 at 12 27 20 PM

I compared against Mobiscroll and there's remains active until you hover off the date cell. It does feel strange that it remains active until clicking off.

Now that we can select multiple days in the same month, the active ring has the ability to collide with a selected cell (visually). Should we adjust the design at all?
Screen Shot 2022-07-15 at 12 30 49 PM

The functional behavior looks great though 🎉

amandaejohnston:
Hmm... the faded circle around focused days shows you where you are when using keyboard navigation, so changing how that behaves may get confusing 🤔 Mobiscroll shows an entirely different circle (a gray outline) when using the keyboard. I'm okay with everything how it is, since it balances form and function in my eyes, but I could be persuaded otherwise.

core/src/components/datetime/test/multiple/datetime.e2e.ts
```diff
@@ -0,0 +1,159 @@
...
+  });
+
+  test('clicking unselected days should select them', async ({ page }) => {
+    const datetime = await setup(page, 'singleDefaultValue');
+    const june1Button = datetime.locator('[data-month="6"][data-day="1"]');
+    const june2Button = datetime.locator('[data-month="6"][data-day="2"]');
+    const june3Button = datetime.locator('[data-month="6"][data-day="3"]');
+
+    await june2Button.click();
+    await expect(datetime).toHaveJSProperty('value', ['2022-06-01', '2022-06-02']);
```
liamdebeasi, line 47: Optional: Do we need to do `page.waitForChanges()` here? Or perhaps we could wait for the `ionChange` event to be emitted and check the payload? I worry that this may be flaky if the component takes a long time to re-render on CI.

core/src/components/datetime/datetime.tsx
```diff
@@ -308,58 +308,77 @@ export class Datetime implements ComponentInterface {
...
+
   /**
    * The value of the datetime as a valid ISO 8601 datetime string.
+   * Will be an array of strings if `multiple="true"`.
```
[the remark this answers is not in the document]
amandaejohnston: Hmm... more the `ionChange` piece, though it also applies to the `value` you can access via e.g. `datetimeEl.value`. I think saying "should" instead of "will" gets closer to what I meant.

core/src/components/datetime/utils/parse.ts
```diff
@@ -55,7 +55,16 @@ export const getPartsFromCalendarDay = (el: HTMLElement): DatetimeParts => {
  * We do not use the JS Date object here because
  * it adjusts the date for the current timezone.
  */
-export const parseDate = (val: string | undefined | null): any | undefined => {
+export function parseDate(val: string): DatetimeParts;
+export function parseDate(val: string[]): DatetimeParts[];
+export function parseDate(val: undefined | null): undefined;
+export function parseDate(val: string | string[]): DatetimeParts | DatetimeParts[];
+export function parseDate(val: string | string[] | undefined | null): DatetimeParts | DatetimeParts[] | undefined;
+export function parseDate(val: string | string[] | undefined | null): DatetimeParts | DatetimeParts[] | undefined {
```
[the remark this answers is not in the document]
amandaejohnston: These are overload signatures: https://www.typescriptlang.org/docs/handbook/2/functions.html#function-overloads In this case, it's a way of communicating to Typescript the connection between the param types and return type. For example, if you call `parseDate` with a single `string`, TS knows the return type will always be a single `DatetimeParts`. This avoids needing extra type assertions on the caller's side.

core/src/components/datetime/test/multiple/datetime.e2e.ts
```diff
@@ -0,0 +1,159 @@
...
+
+    await expect(datetime).toHaveJSProperty('value', undefined);
+  });
+
+  test('setting value programmatically should update active days', async ({ page }) => {
+    const datetime = await setup(page, 'singleDefaultValue');
+    const june1Button = datetime.locator('[data-month="6"][data-day="1"]');
+    const june2Button = datetime.locator('[data-month="6"][data-day="2"]');
+    const june3Button = datetime.locator('[data-month="6"][data-day="3"]');
+    const june4Button = datetime.locator('[data-month="6"][data-day="4"]');
```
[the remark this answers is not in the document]
amandaejohnston: Good thinking 👀

core/src/components/datetime/datetime.tsx
```diff
@@ -1452,14 +1560,14 @@ export class Datetime implements ComponentInterface {
...
-    const { workingParts } = this;
+    const { activeParts, workingParts } = this;
 
     return (
       <ion-picker-column-internal
         class="day-column"
         color={this.color}
         items={days}
-        value={workingParts.day || this.todayParts.day}
+        value={(workingParts.day || this.todayParts.day) ?? undefined}
```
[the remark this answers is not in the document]
amandaejohnston: If you mean the `??` operator, that's this: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Nullish_coalescing_operator

`DatetimeParts.day` has a type of `number | null`, while `ion-picker-column-internal`'s `value` prop is `string | number | undefined` and _doesn't_ allow for `null`. This causes a type error since `null` can't be assigned to `value`. The coalescing operator ensures that if `workingParts.day` and `this.todayParts.day` are both `null`, we pass in `undefined` instead, since the coalescing operator returns the right-hand side.

core/src/components/datetime/utils/state.ts
```diff
@@ -89,13 +89,15 @@ export const isDayDisabled = (
...
   refParts: DatetimeParts,
-  activeParts: DatetimeParts,
+  activeParts: DatetimeParts | DatetimeParts[],
   todayParts: DatetimeParts,
   minParts?: DatetimeParts,
   maxParts?: DatetimeParts,
   dayValues?: number[]
 ) => {
-  const isActive = isSameDay(refParts, activeParts);
+  const activePartsArray = Array.isArray(activeParts) ? activeParts : [activeParts];
```
liamdebeasi, line 98: Why do we not take the recursive approach here like we do in https://github.com/ionic-team/ionic-framework/pull/25514/files/6d336632b022f40c9bd38629bc5e990247f46c06#diff-bc71320a199f4545788c57d1cb7c033c228a9ed44195afa3848d6b88f0a963abR64-R66?

core/src/components/datetime/utils/manipulation.ts
```diff
@@ -10,7 +10,14 @@ const fourDigit = (val: number | undefined): string => {
   return ('000' + (val !== undefined ? Math.abs(val) : '0')).slice(-4);
 };
 
-export const convertDataToISO = (data: any): string => {
+export function convertDataToISO(data: DatetimeParts): string;
+export function convertDataToISO(data: DatetimeParts[]): string[];
+export function convertDataToISO(data: DatetimeParts | DatetimeParts[]): string | string[];
```
liamdebeasi, line 15: Is this line needed since the main function implementation has the same param/return types?

core/src/components/datetime/utils/parse.ts
```diff
@@ -55,7 +55,16 @@ export const getPartsFromCalendarDay = (el: HTMLElement): DatetimeParts => {
  * We do not use the JS Date object here because
  * it adjusts the date for the current timezone.
  */
-export const parseDate = (val: string | undefined | null): any | undefined => {
+export function parseDate(val: string): DatetimeParts;
+export function parseDate(val: string[]): DatetimeParts[];
+export function parseDate(val: undefined | null): undefined;
+export function parseDate(val: string | string[]): DatetimeParts | DatetimeParts[];
+export function parseDate(val: string | string[] | undefined | null): DatetimeParts | DatetimeParts[] | undefined;
```
liamdebeasi, line 62: Is this line needed since the main function implementation has the same param/return types?

core/src/components/datetime/utils/state.ts
```diff
@@ -89,13 +89,15 @@ export const isDayDisabled = (
 export const getCalendarDayState = (
   locale: string,
   refParts: DatetimeParts,
-  activeParts: DatetimeParts,
+  activeParts: DatetimeParts | DatetimeParts[],
```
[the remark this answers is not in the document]
liamdebeasi: I'm fine with this approach as long as we are consistent with how we handle single values vs arrays

core/src/components/datetime/datetime.tsx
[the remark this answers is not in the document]
liamdebeasi: Oh right I forgot `day` could be `null`. Ok sounds good.

core/src/components/datetime/utils/manipulation.ts
amandaejohnston: Yep, same idea as with `parseDate`.

core/src/components/datetime/utils/state.ts
liamdebeasi: Can we add some comments here to clarify what this code is doing? It wasn't immediately clear when I reviewed, but your explanation makes sense.

core/src/components/datetime/utils/comparison.ts
```diff
@@ -38,13 +38,28 @@ export const isAfter = (baseParts: DatetimeParts, compareParts: DatetimeParts) =
...
+
+  if (Array.isArray(value)) {
+    for (const val of value) {
+      if (isOutOfBounds(val)) {
+        printWarning();
+        break;
+      }
+    }
+  } else if (isOutOfBounds(value)) {
+    printWarning();
```
liamdebeasi, line 63: Can we change this so that we do not have diverging code paths? Maybe we can do something similar to what `getCalendarDayState` does:

```js
const activePartsArray = Array.isArray(activeParts) ? activeParts : [activeParts];
```

liamdebeasi (approved):
Great job -- this is good to go once the below comments are addressed.

sean-perkins (approved):

PR #25522, opened 2022-06-22
test(angular): add timeout to avoid flaky value accessor test

sean-perkins:
## Pull request checklist

## Pull request type

 

 

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

Occasionally the Angular E2E test suite will fail when validating if setting a value changes the form status of the control. 

Upon debugging this failing test, it appears to be the result of `ion-select`. After assigning the value, the control continues to have the `ion-invalid` class, which means that the value accessor never received the `ionChange` event.

Theory is that the value is set by Cypress before the component fully loads. By removing the safe guard, the event is still able to fire prior to the `componentDidLoad` resolving.

Issue URL: N/A

## What is the new behavior?

- Adds TODO statements for code that will be removed if the `ionChange` RFC is approved
- Adds timeout to the e2e test to reduce the flakiness of the value accessor test

- [ ] Yes
- [x] No


amandaejohnston:
Would it be possible to have the test wait for the component to fully load instead?

liamdebeasi:
Do you think the issue that what the `didInit` tries to avoid will be resolved with our proposed `ionChange` revisions?

sean-perkins:
Yes, I think the checks were to work around the value accessors setting the value before the web component `defineCustomElement` finishes. 

With our proposed changes in the RFC, setting the value from the value accessor should not emit `ionChange`.

liamdebeasi:
Can we keep the `didInit` patch for now and work around it in our test somehow? Then we can add a TODO to remove the `didInit` when the `ionChange` updates are merged (we can probably track these changes in the main ionChange ticket)

liamdebeasi (approved):
Thanks for taking care of this. Don't forget to change the commit subject when merging so this doesn't get logged as a bug.

sean-perkins:
Thanks for the reminder; updated title + description, will wait for Amanda's review before merge 👍

amandaejohnston (approved):

Issue #25147, opened 2022-04-19
bug: focus trapping should be disabled for sheet modals


EinfachHans:
### Prerequisites

- [X] I have read the [Contributing Guidelines](https://github.com/ionic-team/ionic-framework/blob/main/.github/CONTRIBUTING.md#creating-an-issue).
- [X] I agree to follow the [Code of Conduct](https://ionicframework.com/code-of-conduct).
- [X] I have searched for [existing issues](https://github.com/ionic-team/ionic-framework/issues) that already report this problem, without success.

### Ionic Framework Version

- [ ] v4.x
- [ ] v5.x
- [X] v6.x
- [ ] Nightly

### Current Behavior

Ok this one is strange 😃 I experienced a bug, that i was able to reproduce very easy.

When tapping a Toast the Ion Input gets focused (and keyboard opens on mobile).

### Expected Behavior

To not happen 😃 

### Steps to Reproduce

See the linked repo. This repo was created for another issue, but can be used here as well. Steps:

1. Run App
2. Click on filled button "Bieten" -> Sheet moves to upper breakpoint and reveals the input
3. Click again on the button --> Toast gets displayed
4. Click on the Toast --> **Input gets focuses**

### Code Reproduction URL

https://github.com/EinfachHans/ionic-modal-breakpoints-example

### Ionic Info

Ionic:

   Ionic CLI                     : 6.19.0 (/Users/hans/.nvm/versions/node/v16.13.1/lib/node_modules/@ionic/cli)
   Ionic Framework               : @ionic/angular 6.1.0
   @angular-devkit/build-angular : 13.2.6
   @angular-devkit/schematics    : 13.2.6
   @angular/cli                  : 13.2.6
   @ionic/angular-toolkit        : 6.1.0

Capacitor:

   Capacitor CLI      : 3.4.3
   @capacitor/android : not installed
   @capacitor/core    : 3.4.3
   @capacitor/ios     : not installed

Utility:

   cordova-res (update available: 0.15.4) : 0.15.1
   native-run                             : 1.5.0

System:

   NodeJS : v16.13.1 (/Users/hans/.nvm/versions/node/v16.13.1/bin/node)
   npm    : 8.4.0
   OS     : macOS Monterey

liamdebeasi:
Thanks for the issue. Focus trapping is enabled with sheet modals. When you focus the toast, focus is moved back inside of the sheet.

Focus trapping should probably be disabled for sheet modals since it can overlay interactive content.

PR #25540, opened 2022-06-27
feat(modal): clicking handle advances to the next breakpoint

sean-perkins:
## Pull request checklist

## Pull request type

 

 

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes

Clicking the sheet modal handles not advance to the next available breakpoint.

Issue URL: #24069

## What is the new behavior?

- Introduces `handleBehavior`, which controls if clicking the sheet modal handle advances to the next available breakpoint.
- `handleBehavior` defaults to `none`, meaning that existing sheet modals will need to opt-in to this behavior.
- Clicking the handle multiple times will ignore additional presses until the modal moves to the next breakpoint
- While a breakpoint is transitioning, clicking the backdrop will not dismiss the modal.
- When `handleBehavior` is `cycle`, the handle element will be focusable and include aria-labels informing "Activate to adjust the size of the dialog overlaying the screen".

- [ ] Yes
- [x] No


core/src/components/modal/modal.tsx
```diff
@@ -64,6 +65,8 @@ export class Modal implements ComponentInterface, OverlayInterface {
   private sortedBreakpoints?: number[];
   private keyboardOpenCallback?: () => void;
   private moveSheetToBreakpoint?: (options: MoveSheetToBreakpointOptions) => void;
+  // Whether or not the modal is moving to a new breakpoint
+  private breakpointMoving?: boolean;
```
amandaejohnston, line 69: Would it make sense to leverage `currentTransition` for this instead? This would take some adjustments (having `moveSheetToBreakpoint` return a Promise, for example), but it would mean one less variable to worry about and allow other transition functions (such as `present` and `dismiss`) to avoid running into the breakpoint move animation without extra work. It may also let you get rid of `onAnimationFinish`, since I think you could resolve the promise in `moveSheetToBreakpoint` in the `onFinish` callback. Something like that anyway.

core/src/components/modal/modal.tsx
```diff
@@ -820,7 +881,20 @@ export class Modal implements ComponentInterface, OverlayInterface {
         {mode === 'ios' && <div class="modal-shadow"></div>}
 
         <div role="dialog" class="modal-wrapper ion-overlay-wrapper" part="content" ref={(el) => (this.wrapperEl = el)}>
-          {showHandle && <div class="modal-handle" part="handle"></div>}
+          {showHandle && (
+            <button
+              class={{
+                'modal-handle': true,
+                'modal-handle--clickable': isHandleCycle,
```
liamdebeasi, line 888: Shouldn't the handle always be clickable? Users can double tap here to drag and adjust the size of the sheet manually:

https://user-images.githubusercontent.com/2721089/176693672-56bc2053-2b86-41f8-bc11-90629233190e.mp4

I know you can drag the entire header, but the handle was added as a cue for how to expand/collapse the sheet. Even if the handle cycle is 'none' you should still be able to click and drag on it.

core/src/components/modal/modal.tsx
```diff
@@ -820,7 +881,20 @@ export class Modal implements ComponentInterface, OverlayInterface {
...
+              }}
+              // Prevents the handle receiving focus when it is not clickable
```
liamdebeasi, line 890:
```suggestion
              // Prevents the handle from receiving keyboard focus when it does not cycle
```

core/src/components/modal/modal.tsx
```diff
@@ -761,7 +773,55 @@ export class Modal implements ComponentInterface, OverlayInterface {
...
 
+  private async moveToNextBreakpoint() {
+    const { breakpoints, currentBreakpoint } = this;
+
+    if (!breakpoints || currentBreakpoint == null) {
+      /**
+       * If the modal does not have breakpoints and/or the current
+       * breakpoint is not set, we can't move to the next breakpoint.
+       */
+      return false;
```
liamdebeasi, line 784: We don't seem to use the return value anywhere. Was this added in case we do need the results in the future?

core/src/components/modal/modal.tsx
sean-perkins: This class name may be misleading. Before this feature the handle did not have a `cursor: pointer` styling. This class conditionally applies that styling if clicking the handle advances to the next breakpoint. 

Should we show the cursor pointer style always or rename this class to better explain it is applying cursor styles?

core/src/components/modal/modal.tsx
sean-perkins: Correct. The intention was that it allows us to make decisions in the future based on if the operation was successful or not. I could be convinced that this is not necessary since the method is not part of a public API that developers can rely on.

liamdebeasi: probably fine to keep it 👍

core/src/components/modal/modal.tsx
```diff
@@ -820,7 +881,20 @@ export class Modal implements ComponentInterface, OverlayInterface {
...
+          {showHandle && (
+            <button
+              class={{
+                'modal-handle': true,
+                'modal-handle--clickable': isHandleCycle,
+              }}
+              // Prevents the handle receiving focus when it is not clickable
+              tabIndex={!isHandleCycle ? -1 : 0}
+              aria-label={isHandleCycle ? 'Activate to adjust the size of the dialog overlaying the screen' : null}
+              aria-controls={isHandleCycle ? modalId : null}
```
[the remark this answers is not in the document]
sean-perkins: You are correct, thanks for the call out!

core/src/components/modal/gestures/sheet.ts
```diff
@@ -354,6 +354,10 @@ export const createSheetGesture = (
     } else if (!shouldRemainOpen) {
       onDismiss();
     }
+
+    await animation.play();
```
liamdebeasi, line 358: Why we do we call `play` here? The `progressEnd` call should automatically play the rest of the animation.

core/src/components/modal/modal.tsx
```diff
@@ -820,7 +880,16 @@ export class Modal implements ComponentInterface, OverlayInterface {
         {mode === 'ios' && <div class="modal-shadow"></div>}
 
         <div role="dialog" class="modal-wrapper ion-overlay-wrapper" part="content" ref={(el) => (this.wrapperEl = el)}>
-          {showHandle && <div class="modal-handle" part="handle"></div>}
+          {showHandle && (
+            <button
+              class="modal-handle"
+              // Prevents the handle from receiving keyboard focus when it does not cycle
+              tabIndex={!isHandleCycle ? -1 : 0}
```
liamdebeasi, line 887: I think it's `tabindex` but I also am not sure if the casing makes a difference here

liamdebeasi, line 887: Can we add a test to check that the handle can be focused by the keyboard when the handle behavior is `cycle`?

core/src/components/modal/modal.tsx
```diff
@@ -820,7 +880,16 @@ export class Modal implements ComponentInterface, OverlayInterface {
...
+              aria-label={isHandleCycle ? 'Activate to adjust the size of the dialog overlaying the screen' : null}
+              onClick={this.onHandleClick}
```
liamdebeasi, line 889: We should remove the event listener if `!isHandleCycle`. I believe some screen readers will indicate that clicking the handle is supposed to do something if there's a click listener.

liamdebeasi (changes requested):

core/src/components/modal/modal.tsx
sean-perkins: It appears that both `tabindex` and `tabIndex` casing is valid. We use both examples in multiple components. Do we have a preferred casing pattern here?

core/src/components/modal/gestures/sheet.ts
sean-perkins: By calling `animation.play()`, `moveSheetToBreakpoint` will not resolve until the animation completes. This allows me to set and clear the `sheetTransition` state depending on if an active sheet transition is occurring. I followed the implementation pattern of how modal present/dismiss transitions work. 

|Using `animation.play()`|Without|
|---|---|
|<video src="https://user-images.githubusercontent.com/13732623/176975585-69b9c791-269f-4bb8-9b19-0f21b369128f.mp4"></video>|<video src="https://user-images.githubusercontent.com/13732623/176975630-e29ffcfd-b32f-42c4-9189-ea47b673ade0.mp4"></video>|

core/src/components/modal/modal.tsx
liamdebeasi: No preference, especially if we are not being consistent already. Though it would be good to pick one and stick with it for the future.

core/src/components/modal/gestures/sheet.ts
liamdebeasi: In this case you're going to have the animation play twice. You probably won't notice anything since `play()` is called immediately after `progressEnd`, but you may get a duplicate animation on older/slower devices. 

You should be able to get the same result you need by returning a promise that resolves in the `onFinish` callback.

core/src/components/modal/modal.tsx
```diff
@@ -56,14 +57,15 @@ export class Modal implements ComponentInterface, OverlayInterface {
...
+  private sheetTransition?: Promise<any>;
   private destroyTriggerInteraction?: () => void;
   private isSheetModal = false;
   private currentBreakpoint?: number;
   private wrapperEl?: HTMLElement;
   private backdropEl?: HTMLIonBackdropElement;
   private sortedBreakpoints?: number[];
   private keyboardOpenCallback?: () => void;
-  private moveSheetToBreakpoint?: (options: MoveSheetToBreakpointOptions) => void;
+  private moveSheetToBreakpoint?: (options: MoveSheetToBreakpointOptions) => Promise<any>;
```
liamdebeasi, line 68: Can we add a type here?

core/src/components/modal/modal.tsx
```diff
@@ -140,6 +142,13 @@ export class Modal implements ComponentInterface, OverlayInterface {
    */
   @Prop() handle?: boolean;
 
+  /**
+   * The interaction behavior for the sheet modal when the handle is pressed.
+   * Defaults to `"none"`, which  means the modal will not change size or position when the handle is pressed.
+   * Set to `"cycle"` to let the modal cycle between available breakpoints when pressed.
```
liamdebeasi, line 148: Can we add a note to state that the behavior does not apply if `handle` is `false` or when the sheet modal is not used?

core/src/components/modal/modal.tsx
```diff
@@ -833,7 +893,16 @@ export class Modal implements ComponentInterface, OverlayInterface {
         {mode === 'ios' && <div class="modal-shadow"></div>}
 
         <div role="dialog" class="modal-wrapper ion-overlay-wrapper" part="content" ref={(el) => (this.wrapperEl = el)}>
-          {showHandle && <div class="modal-handle" part="handle"></div>}
+          {showHandle && (
+            <button
+              class="modal-handle"
+              // Prevents the handle from receiving keyboard focus when it does not cycle
+              tabIndex={!isHandleCycle ? -1 : 0}
+              aria-label={isHandleCycle ? 'Activate to adjust the size of the dialog overlaying the screen' : null}
```
liamdebeasi, line 901: Even when handleCycle is `none`, the handle can still be used to adjust the size of the dialog. Should we keep the aria-label here? Or maybe have a different aria label?

liamdebeasi (approved):
Great job! Just a few small things below.

core/src/components/modal/modal.tsx
```diff
@@ -146,6 +146,9 @@ export class Modal implements ComponentInterface, OverlayInterface {
    * The interaction behavior for the sheet modal when the handle is pressed.
    * Defaults to `"none"`, which  means the modal will not change size or position when the handle is pressed.
    * Set to `"cycle"` to let the modal cycle between available breakpoints when pressed.
+   *
+   * Handle behavior is only available when the `handle` property is set to `true` and
```
liamdebeasi, line 150: The `handle` property defaults to `undefined` and is enabled by default, so "Handle behavior is only available when the `handle` property is set to `true`" is not accurate.

core/src/components/modal/gestures/sheet.ts
```diff
@@ -294,66 +294,69 @@ export const createSheetGesture = (
...
+
+                  gesture.enable(true);
+                });
+              } else {
                 gesture.enable(true);
-              });
-            } else {
-              gesture.enable(true);
+              }
+              resolve();
```
amandaejohnston, line 341: ~ Since this `resolve` is outside the `raf` called on L310, it could be hit before all the logic in there is done. I'm admittedly not sure of the practical implications there, but it may be worth calling it at the end of the `raf`, and after L339 to cover that case as well.

core/src/components/modal/test/sheet/modal.e2e.ts
```diff
@@ -195,3 +195,111 @@ test.describe('sheet modal: setting the breakpoint', () => {
     expect(updatedBreakpoint).toBe(0.5);
   });
 });
+
+test.describe('sheet modal: handleBehavior cycle: focus trap', () => {
+  test.beforeEach(async ({ page, browserName }, testInfo) => {
+    test.skip(testInfo.project.metadata.rtl === true, 'This does not test LTR vs. RTL layout.');
+    test.skip(browserName === 'firefox', 'Firefox incorrectly allows keyboard focus to move to ion-content');
+    test.skip(browserName === 'webkit', 'Safari incorrectly allows keyboard focus to move to the slotted content');
```
amandaejohnston, line 203: Do we need tech debt tickets for either of these? The `webkit` one almost just sounds like a bug, but there may be context I'm missing.

liamdebeasi: I'm not sure there is much we can do other than file issues with browser vendors. This the same issue as https://github.com/ionic-team/ionic-framework/blob/main/core/src/components/modal/test/basic/modal.e2e.ts#L8.

The problem isn't that focus trapping isn't working, it's that focus is not being moved inside of the container to where we expect it to be. We could test in the negative (I.e. check that body is not focused), but I also worry we may miss some edge cases with that.

liamdebeasi: Alternatively, we could check if the focused element exists inside of the container. That would be a bit of custom code, but it could be more reliable.

amandaejohnston: If it's the same issue, why does the basic test work in Webkit but this one doesn't? Just a different HTML layout? 🤔

core/src/components/modal/gestures/sheet.ts
liamdebeasi: +1 we can probably call it in the if/else block.

sean-perkins: I may need to refactor this approach altogether. Once returning a promise wrapping the `onFinish` callback, it causes the test suite for setting the breakpoint to zero and checking if the event fires, to fail.

core/src/components/modal/test/sheet/modal.e2e.ts
liamdebeasi: On second thought... do we even need this test? The focus trapping tests are handled in https://github.com/ionic-team/ionic-framework/blob/main/core/src/components/modal/test/basic/modal.e2e.ts#L5. Focus trapping in sheet modals should not even be enabled https://github.com/ionic-team/ionic-framework/issues/25147

liamdebeasi: Sounds good. Not sure what would cause the inconsistency. FWIW Chrome + Firefox did focus the handle when pressing Tab with the sheet open. Safari moved focus to the Dismiss Modal button, but pressing Shift + Tab did focus the handle. Maybe a bug with our focus trap util?

amandaejohnston (approved):

PR #25559, opened 2022-06-29
fix(datetime): switching months in wheel picker now selected nearest neighbor

liamdebeasi:
## Pull request checklist

## Pull request type

 

 

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

Issue URL: Internal ticket.

In https://github.com/ionic-team/ionic-framework/pull/25422#issuecomment-1149172489 it was discovered that switching months in the new `preferWheel` date picker would sometimes reset the selected day to 1.

This was happening when switching between months with different maximum days. For example, selecting May 31st then switching the month to April would cause the 1st to be selected. This is because there is no April 31st.

Under the hood, we were filtering the 31st out of the items passed into the picker column, and the picker column did not know what to do.

## What is the new behavior?

- When the `items` prop changes, the picker column will check to see if the selected item is still in the `items` array.
- If it is, then it does nothing.
- If the item is no longer available, it performs the following steps:

1. Finds the location of the active item in the previous `items` array. If the index is not found, it returns early.
2. If the index is found, it steps backwards through the current set of `items`, starting at the last known position of the selected item.
3. For each item, it checks to see if a) the item is defined and b) the item is enabled. It breaks out of the loop if it finds an item that meets both criteria.
4. If an item is found that meets the criteria, the `value` prop is updated and the nearest item is selected.

| `feature-6.2` | branch |
| -------------- | ------ |
| <video src="https://user-images.githubusercontent.com/2721089/176504204-d26033a1-e4c7-4d00-89ac-d60278db1a66.mov"></video> | <video src="https://user-images.githubusercontent.com/2721089/176504242-79f2589f-42a7-4e92-ade9-c0e19ce4068c.mov"></video> |

- [ ] Yes
- [x] No


sean-perkins:
In your after example, when going from "March 29" to February, why does it activate "February 27" instead of the 28th?

liamdebeasi:
Good question! I should have simplified the example. There is an `isDateEnabled` callback that disables all even dates, so the 28th is never a selectable option. https://github.com/ionic-team/ionic-framework/blob/feature-6.2/core/src/components/datetime/test/prefer-wheel/index.html#L48-L53

amandaejohnston (approved):

sean-perkins (approved):
Thanks for the added explanation, this looks good 👍

Issue #25404, opened 2022-06-03
feat: Angular 14 Standalone Components


sean-perkins:
### Prerequisites

- [X] I have read the [Contributing Guidelines](https://github.com/ionic-team/ionic-framework/blob/main/.github/CONTRIBUTING.md#creating-an-issue).
- [X] I agree to follow the [Code of Conduct](https://ionicframework.com/code-of-conduct).
- [X] I have searched for [existing issues](https://github.com/ionic-team/ionic-framework/issues) that already include this feature request, without success.

### Describe the Feature Request

Angular 14 introduces the ability to use standalone components, without having to declare them on an `NgModule`. This feature is used for rendering components within a parent component, as well as for lazy loading a standalone component on a route, instead of using a routing module. 

Standalone components encompass all declarations (pipes, directives and components). 

As of Angular 14, standalone components are a **dev preview** feature. They are also not backwards compatible with older versions of Angular. This means for Ionic to add support for standalone components, Ionic would have to remove support for Angular 13 and lower. 

If you are wanting to _experiment_ with an Angular 14 project and Ionic using standalone components, the following dev-build is available:

```
6.1.7-dev.11654279011.1b971e3e
```

For an example Ionic application on Angular 14, you can clone: https://github.com/sean-perkins/angular-14-demo. Remember to install the available dev-build.

<img width="1440" alt="Screen Shot 2022-06-03 at 2 02 40 PM" src="https://user-images.githubusercontent.com/13732623/171921026-d7b5f180-9528-4529-b2ed-cbea5c118053.png">

### Describe the Use Case

Ionic developers should be able to use the following code:
```ts
{
   path: 'standalone',
   loadComponent: () => import('./standalone.component').then(m => m.StandaloneComponent)
 }
```

To create a route of `/standalone` that renders the `StandaloneComponent` inside of the `ion-router-outlet`.

ntorrey:
Any idea when standalone components will be fully supported?

mackelito:
When will we be able to use standalone components? is there a rough ETA?

sean-perkins:
@mackelito the experimental API support will be available in the next minor release (6.3.0).

This issue is now resolved, since the PR has been merged into the feature branch: https://github.com/ionic-team/ionic-framework/pull/25589

hakimio:
@sean-perkins can you also incorporate PR #25502 in 6.3? I think it's been more than enough time for people to try it out.

sean-perkins:
[...]

PR #25589, opened 2022-07-07
feat(angular): use standalone components with routing

sean-perkins:
## Pull request checklist

## Pull request type

 

 

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes

Standalone components do not work within Ionic's routing implementation.

Issue URL: #25404

## What is the new behavior?

- Introduces an `environmentInjector` property to `ion-router-outlet` and `ion-tabs` to enable developers to make use of standalone components with routing in Angular 14.
- Developer warning to guide developers trying to use standalone components with routing without our required syntax.

Screen Shot 2022-07-07 at 3 13 28 PM

- [ ] Yes
- [x] No

## Other information

Dev-build: `6.1.14-dev.11657224247.185f8192`

In your Angular 14 application, you will need to expose the new `EnvironmentInjector` to Ionic routing components (`ion-router-outlet` and `ion-tabs`).

For example:

```ts
import { EnvironmentInjector } from '@angular/core';

@Component({
  selector: 'app-root',
  template: `<ion-router-outlet [environmentInjector]="environmentInjector"></ion-router-outlet>`
})
export class AppComponent {
  constructor(public environmentInjector: EnvironmentInjector) {}
}
```


sean-perkins:
TODO: Need to rebase 6.3 against `main`, to get the updated multi-framework version testing. Then need to write a specific test in the Angular 14 test app that verifies standalone components with routing.

angular/src/di/r3_injector.ts
```diff
@@ -30,5 +30,5 @@ export abstract class EnvironmentInjector implements Injector {
   /**
    * @internal
    */
-  abstract onDestroy(callback: () => void): void;
+  abstract onDestroy?(callback: () => void): void;
```
amandaejohnston, line 33: Why was this change needed?

angular/src/directives/navigation/ion-router-outlet.ts
```diff
@@ -55,6 +56,16 @@ export class IonRouterOutlet implements OnDestroy, OnInit {
 
   tabsPrefix: string | undefined;
 
+  /**
+   * @experimental
+   *
+   * The `EnvironmentInjector` provider instance from the parent component.
+   * Required for using standalone components with `ion-router-outlet`.
+   *
+   * Will be deprecated and removed when Angular 13 support is dropped.
```
amandaejohnston, line 65: Do we need a tech debt ticket for this? (Ditto for `ion-tabs`)

angular/src/di/r3_injector.ts
sean-perkins: This was the align with the type signature of the `EnvironmentInjector` type from `@angular/core`. We are maintaining our own "version" of the types, since we cannot update our library dependencies to target Angular 14. When I originally ported this, I forgot to make the `onDestroy` callback optional.

Without this change, developers would receive a typescript language server warning in their IDE that the type of their `environmentInjector` does not match the type signature of what we are expecting to be passed in.

angular/src/directives/navigation/ion-router-outlet.ts
sean-perkins: I'm indifferent since this is an experimental API that could be ripped out completely if either Angular doesn't proceed with standalone as part of the official public API or if don't think it is a good architectural fit with Ionic. 

I think once we consider it public API, having tech debt to maintain it could make sense. 

But we also don't have many experimental features, so if we want to follow our existing process like everything else, I'm happy to create tickets 👍

amandaejohnston (approved):

liamdebeasi (approved):
Changes look good. Are these test app config changes intentional or do we just need to resync `feature-6.3` with `main`?

sean-perkins:
The changes to the test apps are intentional. Removed the non-applicable README, added a missing tsconfig and created a script to sync only the e2e content changes (faster than syncing all of the node modules and app files when trying to modify a cypress test file in the base app).

The ng14 test app changes are required to validate the standalone routing behavior works.

PR #24399, opened 2021-12-14
feat: range, add reference point to start range slider at #24348

sachingarg05:
Added property barActiveStart. bar-active is shown between `barActiveStart` and knob for single knob range-bar. By default `barActiveStart` is set to `min` (Minimum integer value of the range).

## Pull request checklist

Please check if your PR fulfills the following requirements:
- [x] Tests for the changes have been added (for bug fixes / features)
- [x] Docs have been reviewed and added / updated if needed (for bug fixes / features)
- [x] Build (`npm run build`) was run locally and any changes were pushed
- [x] Lint (`npm run lint`) has passed locally and any fixes were made for failures

## Pull request type

 

 

Please check the type of change your PR introduces:
- [ ] Bugfix
- [x] Feature
- [ ] Code style update (formatting, renaming)
- [ ] Refactoring (no functional changes, no api changes)
- [ ] Build related changes
- [ ] Documentation content changes
- [ ] Other (please describe): 

## What is the new behavior?
Details were discussed in issue Number: #24348

## Does this introduce a breaking change?

- [ ] Yes
- [x] No

## Other information

md:
image
ios:
image

Issue #24348, opened 2021-12-08
feat: range, add reference point to start range slider at


sachingarg05:
### Prerequisites

- [X] I have read the [Contributing Guidelines](https://github.com/ionic-team/ionic-framework/blob/main/.github/CONTRIBUTING.md#creating-an-issue).
- [X] I agree to follow the [Code of Conduct](https://ionicframework.com/code-of-conduct).
- [X] I have searched for [existing issues](https://github.com/ionic-team/ionic-framework/issues) that already include this feature request, without success.

### Describe the Feature Request

ion-range shows an active bar which starts from extreme left (0 value) to current knob location. 

Very often ion-range needs to have its floor value (or base value) in the center or somewhere else. In these cases, it is required for the active bar to show between the floor value and current knob value, instead of always starting from 0 (extreme left).

### Describe the Use Case

A slider to reduce or increase brightness of an image.

### Describe Preferred Solution

Adding a 'floor' value attribute for ion-range, and then updating the left/right calculation of active bar to start from 'floor' value to current knob value (instead of always starting from 0 to current knob value)

### Related Code

/main/core/src/components/range/range.tsx, line 424
```
    const barStart = `${ratioLower * 100}%`;
    const barEnd = `${100 - ratioUpper * 100}%`;

```

These calculations will need to be updated to start from a floor value (or base value) to current knob value, instead of always starting from 0 (ratioLower is always zero for single knob ion-range).

liamdebeasi:
Thanks for the issue. Can you provide a code sample of the issue you are running into to help me better understand your use case?

sachingarg05:
Consider this ion-range with min and max of -100 and 100
```
<ion-range color="dark" [min]="-100" [max]="100" [value]="0" [snaps]="10" [step]="10" mode="md" pin="true">
```
![image](https://user-images.githubusercontent.com/4737179/145359734-0ba12974-8f42-4782-b4eb-f776b8a81362.png)

Side note: Maybe the 'tick' mark at 'floor value' location can also be slightly 'heavier' than other tick marks.

liamdebeasi:
Thanks! That makes sense. Do you happen to have examples of other applications/libraries that offer this behavior?

sachingarg05:
ngx-slider has this option, by setting property 'showSelectionBarFromValue'

On their demo page, check the demo for "Slider with visible selection bar from specified value":

https://angular-slider.github.io/ngx-slider/demos#selection-bar-from-value-slider

[...]

PR #25598, opened 2022-07-08
feat(range): add reference point for start position of range slider

sean-perkins:
## Pull request checklist

## Pull request type

 

 

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes

Developers are unable to specify a start position of the range bar. This prevents developers from having experiences where the start position may be fixed at "0", but have a min value of less than zero and a maximum value higher than zero. 

For example, this is not possible:

Screen Shot 2022-07-08 at 3 41 47 PM

Issue URL: #24348

## What is the new behavior?

- Introduces `barActiveStart` property to enable developers to set the position the bar will begin from in the range slider (only available in single knob)

- [ ] Yes
- [x] No

## Other information

This PR is in favor of #24399, will give co-authored-by credit when merging.


amandaejohnston (approved):

sean-perkins:
Not sure why the accordion screenshot has a diff... going to add a task for that in our backlog and see if another run of updating the reference screenshots resolves it.

core/src/components/range/range.tsx
```diff
@@ -510,7 +549,7 @@ export class Range implements ComponentInterface {
 
         const tick: any = {
           ratio,
-          active: ratio >= ratioLower && ratio <= ratioUpper,
+          active: ratio >= Math.min(ratioLower, ratioUpper) && ratio <= Math.max(ratioLower, ratioUpper),
```
liamdebeasi, line 552: Can we add some comments to explain what this checks?

core/src/components/range/range.tsx
```diff
@@ -142,6 +143,29 @@ export class Range implements ComponentInterface {
    */
   @Prop() ticks = true;
 
+  /**
+   * The start position of the range active bar, only available with a single knob (dualKnobs="false").
+   * Valid values are between the `min` and `max` values.
```
liamdebeasi, line 148:
```suggestion
   * Valid values are between the `min` and `max` values.
```

Not quite sure how best to word this, but it would be nice to state that it is inclusive of `max` and `min`.

core/src/components/range/range.tsx
```diff
@@ -142,6 +143,29 @@ export class Range implements ComponentInterface {
...
+   */
+  @Prop() barActiveStart?: number;
+  @Watch('barActiveStart')
```
liamdebeasi, line 151: Bikeshedding: We call this the "active bar" in the description. Should we call the prop `activeBarStart` instead?

liamdebeasi (changes requested):
According to the design doc, "If [the floor value] falls outside [max and min], a console warning will be logged for the developer and floor will fall back to the min value specified."

The warning is there, but the floor does not fall back to the min value.

Example: 

```html
<ion-range
  class="ion-no-padding"
  min="-100"
  max="100"
  snaps="10"
  step="10"
  pin="true"
  bar-active-start="1000"
  value="0"
>
</ion-range>
```

core/src/components/range/range.tsx
sean-perkins: I like it 👍 I'll rework the API.

liamdebeasi (approved):
Great job! Good to go once my remaining comment has been addressed.

PR #25649, opened 2022-07-15
feat(datetime-button): add support for opening datetimes in overlays with button

liamdebeasi:
## Pull request checklist

## Pull request type

 

 

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

Issue URL: Internal ticket

## What is the new behavior?

- This feature adds support for ensuring that overlays are correctly sized when opening modal/popover from a datetime button.

- [ ] Yes
- [x] No

## Other information

Note: The modal + datetime tests will not work until https://github.com/ionic-team/ionic-framework/pull/25648 is merged in


core/src/components/datetime/datetime.tsx
```diff
@@ -1080,6 +1086,10 @@ export class Datetime implements ComponentInterface {
     }
 
     this.initializeListeners();
+
+    requestAnimationFrame(() => {
```
liamdebeasi, line 1090: Personally I don't love the idea of adding a private event just to capture when a component re-renders, but I was not able to think of any other solutions.

Example of the issue:

1. Datetime in a popover starts with `presentation="date"` (calendar grid).
2. User clicks the time portion of the datetime button.
3. Datetime updates `presentation` to `"time"`.
4. `popoverEl.present()` is called. 
5. At this point the datetime component has not fully re-rendered to the time wheel. As a result, the following code gets the dimensions of the calendar grid, not the wheel picker: https://github.com/ionic-team/ionic-framework/blob/b7afcb0f0c36d84f3b4d65844df28e6293bc1ea5/core/src/components/popover/utils.ts#L58
6. The datetime re-renders and is displayed as a time wheel, but the popover position is not correct.

I did look into doing something similar to this: https://github.com/ionic-team/ionic-framework/blob/5a606de14b088316905bf3ece478ae8ab5ac7211/core/src/components/datetime/test/utils/month-did-change-event.js
But that did not seem to work either. I listened for when the `datetime-presentation-time` class was added, but that did not mean that the component was visually updated yet. Open to other ideas!

liamdebeasi:
One behavior that does not look great with the `fit-content` usage is when you switch between the grid and the wheel picker in `presentation="date"`:

https://user-images.githubusercontent.com/2721089/179299565-4d1ce981-a8bc-4582-92be-2e15efef169c.mov

Thoughts on how best to account for this? Maybe set a min-width on the datetime when the calendar grid is being used and `size="fixed"`?

core/src/components/datetime-button/datetime-button.tsx
```diff
@@ -222,15 +250,19 @@ export class DatetimeButton implements ComponentInterface {
      * the datetime is opened.
      */
     this.selectedButton = 'date';
+
+    this.setOverlaySize(ev, needsPresentationChange, this.dateTargetEl);
   };
 
-  private handleTimeClick = () => {
+  private handleTimeClick = async (ev: Event) => {
```
sean-perkins, line 257: ~ Does this need the `async` modifier?

core/src/components/datetime/datetime.tsx
sean-perkins: When using a mutation observer, did you additionally use a `requestAnimationFrame` in the callback handler for the observer? 

I'm indifferent to this approach. I see the argument against internal event APIs like this. If a mutation observer cannot suffice, I would say this is appropriate for now, until we can investigate other patterns.

core/src/components/datetime-button/datetime-button.tsx
liamdebeasi: Nope, I will remove

liamdebeasi:
When using the mutation observer I needed like three `raf`s in order for it to work on my Mac. However, when throttling the CPU I sometimes needed more. There's probably a better way to solve this, but I can't quite think of it at the moment.

sean-perkins:
I'm good with the internal event pattern then 👍 

For the width changes in the grid, I would recommend we set a min width/styles to make the month/year selection match the same width as the date picker. Having the layout not shift when toggling between would be ideal here.

amandaejohnston:
I'm also okay with the `ionRender` event pattern for now; my immediate thought was also to use a `raf` in the mutation observer callback, but if that wasn't enough, what we've got now should be okay.

core/src/components/datetime/datetime.scss
```diff
@@ -32,24 +32,37 @@ ion-picker-column-internal {
...
+/**
+ * When using the datetime grid interface,
+ * we want the component to keep the same
+ * width even when the month/year picker is open.
+ * If we do not set the width here then
+ * the component will cause a layout shift
+ * when the month/year picker opens.
+ */
+:host(.datetime-size-fixed.datetime-grid) {
+  width: #{$datetime-fixed-width};
```
amandaejohnston, line 63: Many of the screenshot updates that came from this change have blanked out calendars (ex. `core/src/components/datetime/test/color/datetime.e2e.ts-snapshots/datetime-color-custom-dark-color-md-ltr-Mobile-Firefox-linux.png`). We may need an `await page.waitForSelector('.datetime-ready')` somewhere in the test(s), or some other tweak to the `ionRender` logic.

amandaejohnston (changes requested):

liamdebeasi:
I found a better way to fix the sizing of the datetime in a less intrusive way. (See latest commits)

sean-perkins:
Observing that the lower bounds for the min-width is causing clipping in the overlay test (with popover):

Screen Shot 2022-07-20 at 2 14 55 PM

Also, likely unrelated to any of these changes, but there is some odd rendering happening at the bottom of the picker columns in iOS emulation:

Screen Shot 2022-07-20 at 2 15 22 PM

It looks like the top few pixels of text of the columns are rendering above the border that separates the picker column and the actions.

liamdebeasi:
Good catch. The usage in one of the tests was incorrect, resulting in the size/positioning being wrong.

----

> It looks like the top few pixels of text of the columns are rendering above the border that separates the picker column and the actions.

I can repro in `main` using the popover example in http://localhost:3333/src/components/datetime/test/basic?ionic:mode=ios. Oddly enough I can only reproduce that on desktop Safari -- mobile Safari is fine.

amandaejohnston (approved):

sean-perkins (approved):

Issue #25590, opened 2022-07-07
bug:  Conditionally rendering ionModal in react - Failed to execute 'removeChild' on 'Node': The node to be removed is not a child of this node.


corysmc:
### Prerequisites

- [X] I have read the [Contributing Guidelines](https://github.com/ionic-team/ionic-framework/blob/main/.github/CONTRIBUTING.md#creating-an-issue).
- [X] I agree to follow the [Code of Conduct](https://ionicframework.com/code-of-conduct).
- [X] I have searched for [existing issues](https://github.com/ionic-team/ionic-framework/issues) that already report this problem, without success.

### Ionic Framework Version

- [ ] v4.x
- [ ] v5.x
- [X] v6.x
- [ ] Nightly

### Current Behavior

Conditionally rendering an IonModal, breaks when dismissing when using ionic v6.

Observed behavior: white/blank screen.

Error in console:
react-dom.development.js:10301 Uncaught DOMException: Failed to execute 'removeChild' on 'Node': The node to be removed is not a child of this node.

### Expected Behavior

In an ideal scenario - the ionic modal could be conditionally rendered (without using the isOpen prop) this is a common pattern in react applications.

### Steps to Reproduce

1. Conditionally render an IonModal

```tsx
...
const [showIonModal, setShowIonModal] = useState(false);
...

return (
<div>
  <IonButton onClick={() => setShowIonModal(true)}>
    Show Modal
  </IonButton>
  {showIonModal && (
    <IonModal isOpen={true} onDidDismiss={() => setShowIonModal(false)}>
      <IonHeader>
        <IonToolbar>
          <IonTitle>Content</IonTitle>
        </IonToolbar>
      </IonHeader>
      <IonContent fullscreen>
        Ionic Modal Content here!
        <IonButton color="danger" onClick={() => setShowIonModal(false)}>
          Hide Modal (will break)
        </IonButton>
      </IonContent>
    </IonModal>
  )}
</div>
)
...
```

2. Dismiss the modal (so that IonModal is pulled from the dom)
3. Observe the error.

### Code Reproduction URL

https://github.com/corysmc/react-overlay-bug

### Ionic Info

Ionic:

   Ionic CLI : 6.15.0
   Ionic Framework : @ionic/react 6.1.13

Utility:

   cordova-res : not installed
   native-run  : 1.6.0

System:

   NodeJS : v14.16.0
   npm    : 6.14.11
   OS     : macOS Big Sur

### Additional Information

Related:
[...]

PR #25663, opened 2022-07-20
fix(react): Inline overlays can be conditionally rendered (outdated)

sean-perkins:
## Pull request checklist

## Pull request type

 

 

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

Inline overlays (modal, popover) were never configured with a framework delegate. This resulted in us removing DOM nodes through manual manipulation, instead of relying on React to clear the React component and associated DOM node.

This resulted in two problems:
1. Developers could not conditionally render an inline modal/popover.
2. Dynamic sibling nodes rendered before an IonModal/IonPopover that were mutated while the overlay was dismissing, would result in the same exception from 1. 

Issue URL: #25590, #25775

## What is the new behavior?

- Inline overlays are configured with a framework delegate implementation that teleports the inline overlay to the root of the app and back to it's original DOM location (when dismissed)
- `IonModal` and `IonPopover` can be conditionally rendered.
- `IonModal` and `IonPopover` will not throw exceptions when sibling content is dynamically rendered before its node.
- Overlays are unmounted when history push/pop/replace occurs
- Overlays are moved to the top `z-index` when they are presented
  - ActionSheet and Alert internal elements assigned an id based on the `z-index` are statically computed

- [ ] Yes
- [x] No

## Other information

Open to feedback on naming/folder structure for Github specific issue test-cases. 

Dev-build: `6.2.8-dev.11663187107.1a1cbb3f`


sean-perkins:
Going to update this PR to account for the changes we identified to resolve this: #25775. Will re-add everyone for review once that work is ready.

packages/react/src/components/createInlineOverlayComponent.tsx
```diff
@@ -99,7 +114,45 @@ export const createInlineOverlayComponent = <PropType, ElementType>(
...
-      attachProps(node, this.props, prevProps);
+
+      const newProps = {
+        ...this.props,
+        delegate: this.delegate,
+      };
+
+      attachProps(
+        node,
+        /**
```
sean-perkins, line 125: Need to remove this comment block.

packages/react/src/components/createInlineOverlayComponent.tsx
```diff
@@ -99,7 +114,36 @@ export const createInlineOverlayComponent = <PropType, ElementType>(
...
+        this.isDismissing = true;
+      }
+
+      return true;
+    }
+
+    componentWillUnmount() {
+      if (this.ref.current) {
+        this.ref.current.dismiss();
+      }
```
amandaejohnston, line 146: While playing around in a small test app, I noticed some errors when doing replace or push operations that I think are due to the timing of this dismiss call.

Repo/branch: https://github.com/amandaejohnston/sandbox-ionic-react/tree/FW-1889
Screencast: https://user-images.githubusercontent.com/90629384/194619453-2bab84cd-bfa6-421f-a9dd-ea68dec8b36c.mp4

My guess is this is happening because the `dismiss` event is firing and/or being picked up by the `Home` component *after* the modal has been removed from the DOM. This would cause the `onDidDismiss` handler for the modal to try and set the `isOpen` state on the modal when it's unmounted, throwing the error. Although, the fact that I don't get the error when doing a goBack operation may throw a wrench in that. 🤔 

Let me know if this is an implementation issue on my end, of course! The code is mostly copy-pasted from the test app page in this PR but maybe something went screwy in the process.

packages/react/test-app/cypress/integration/issues/gh-25590.spec.ts
```diff
@@ -0,0 +1,41 @@
+describe('IonModal', () => {
```
amandaejohnston, line 1: Could we name this file something more descriptive, maybe `conditional-render.spec.ts` or something? I would rather us not have to dig through the code and/or click through to the GH issue to figure out the context.

core/src/components/action-sheet/action-sheet.tsx
```diff
@@ -239,22 +243,21 @@ export class ActionSheet implements ComponentInterface, OverlayInterface {
...
     return (
       <Host
         role="dialog"
         aria-modal="true"
-        aria-labelledby={header !== undefined ? headerID : null}
+        aria-labelledby={header !== undefined ? headerId : null}
         tabindex="-1"
         {...(htmlAttributes as any)}
         style={{
-          zIndex: `${20000 + this.overlayIndex}`,
```
liamdebeasi, line 257: Are these changes required for this fix to land? If not, can we move them to a separate PR? Same goes for the changes in alert.tsx.

core/src/components/toast/toast.tsx
```diff
@@ -177,7 +177,7 @@ export class Toast implements ComponentInterface, OverlayInterface {
     if (this.durationTimeout) {
       clearTimeout(this.durationTimeout);
     }
-    return dismiss(this, data, role, 'toastLeave', iosLeaveAnimation, mdLeaveAnimation, this.position);
+    return dismiss(this, data, role, 'toastLeave', iosLeaveAnimation, mdLeaveAnimation, true, this.position);
```
liamdebeasi, line 180: Why does toast need the `true` param but others like alert/action-sheet/etc do not?

core/src/components/modal/modal.tsx
```diff
@@ -705,17 +705,26 @@ export class Modal implements ComponentInterface, OverlayInterface {
...
+    const { delegate, inline } = this.getDelegate();
+
+    this.currentTransition = dismiss(
+      this,
+      data,
+      role,
+      'modalLeave',
+      iosLeaveAnimation,
+      mdLeaveAnimation,
+      inline === false || delegate === undefined,
```
liamdebeasi, line 717: Can we extract this to a variable that better defines what it is checking? My understanding is we want to only remove the overlay if the overlay is presented via a controller, right?

packages/react/src/components/createInlineOverlayComponent.tsx
```diff
@@ -123,23 +167,48 @@ export const createInlineOverlayComponent = <PropType, ElementType>(
...
-          ref: this.wrapperRef,
-          style: {
-            display: 'flex',
-            flexDirection: 'column',
-            height: '100%'
-          }
-        }, children) :
-        null
+      const keepContentsMounted =
+        this.props.trigger !== undefined || this.props.keepContentsMounted;
```
liamdebeasi, line 180: Why do we check for a trigger with `keepContentsMounted`?

packages/react/src/components/createInlineOverlayComponent.tsx
```diff
@@ -123,23 +167,48 @@ export const createInlineOverlayComponent = <PropType, ElementType>(
...
+
+      const mountOverlay =
+        this.state.isOpen || this.props.isOpen || this.isDismissing || keepContentsMounted;
```
liamdebeasi, line 183: Why do we need to check both `this.state.isOpen` and `this.props.isOpen`?

packages/react/src/components/createInlineOverlayComponent.tsx
```diff
@@ -99,7 +114,36 @@ export const createInlineOverlayComponent = <PropType, ElementType>(
...
     componentDidUpdate(prevProps: IonicReactInternalProps<PropType>) {
       const node = this.ref.current! as HTMLElement;
-      attachProps(node, this.props, prevProps);
+
+      const newProps = {
+        ...this.props,
+        delegate: this.delegate,
+      };
+
+      attachProps(
```
liamdebeasi, line 123: Why do we need to attach the props here? Shouldn't the render function pass the updated props via `newProps`?

core/src/components/toast/toast.tsx
sean-perkins: The last parameter of the `dismiss` utility is optional. Alert/ActionSheet/etc. do not use the last parameter to specify additional options, toast does. 

The new property introduced to `dismiss` has a default value of `true` that matches the base behavior on `main` (removing the overlay element). 

Since Toast specifies a value for the last optional parameter, is has to specify a value for the parameter before it. Alert/ActionSheet/etc. do not, so they do not need to specify the parameter, as it will use the default value.

core/src/components/modal/modal.tsx
sean-perkins: Correct, the overlay should only be removed if presented via a controller.

Should it be something like:
```ts
const hasController = inline === false || delegate === undefined;
```

or explicit to the removal of the overlay element?

```ts
const removeOverlay = inline === false || delegate === undefined;
```

packages/react/src/components/createInlineOverlayComponent.tsx
sean-perkins: I will need to dig back into this decision. It was deliberate to solve an issue with the overlay not mounting when using the `isOpen` property directly. I will see if I can reproduce the issue that influenced it.

sean-perkins:
Closing this PR in favor of https://github.com/ionic-team/ionic-framework/pull/26111. I will keep the branch around for reference if we decide to move to React Portals in a future major version.

PR #25669, opened 2022-07-21
fix(reorder-group): synchronize reordering with style update

### LINKED ISSUE #21182: bug: IonReorder items animate twice in Firefox
# Bug Report

**Ionic version:**
**5.x**

**Current behavior:**

With IonReorder, when a user drags an item to a new location the "displaced" items use a translateY animation to reposition themselves. When the dragged item is released, the displaced items animate a second time, even though they are already positioned correctly. The user experience is that some items "jump" unexpectedly.

**Expected behavior:**

The translateY animation should only run only once, with no "jump" when the reorder completes.

**Steps to reproduce:**

The easiest way to reproduce is to view the [official ion-reorder example](https://ionicframework.com/docs/api/reorder/). Click the "Toggle" text to enable reordering. Next, slowly drag "Item 1" to the location of "Item 4". Notice how the other items translate to their new locations. Then release "Item 1" in its final location and notice how the other items animate again.

dan-rp1:
closes #21182

- [X] Bugfix

In the function completeSync(), in the reorder-group component:

When used in Firefox, the most recently set value of the translateY css property is applied to the selectedItemEl variable and rendered visually before the cleared transform property is rendered. This creates the appearance of a "double" animation as the element jumps out of place and slides to its final location.

Issue URL:  https://github.com/ionic-team/ionic-framework/issues/21182

## What is the new behavior?

The reorder operation and the update to the style property are moved to the next turn of the event loop by setTimeout (with default of 0 delay), which causes Firefox to render them in the expected way (no jumping of the selected element).

- [ ] Yes
- [X] No

### DIFF
--- a/core/src/components/reorder-group/reorder-group.tsx
+++ b/core/src/components/reorder-group/reorder-group.tsx
@@ -226,13 +226,16 @@ export class ReorderGroup implements ComponentInterface {
         this.el.insertBefore(selectedItemEl, ref);
       }
 
-      if (Array.isArray(listOrReorder)) {
-        listOrReorder = reorderArray(listOrReorder, fromIndex, toIndex);
-      }
+      // setTimeout forces the style update to synchronize with the reordering.
+      setTimeout(() => {
+        if (Array.isArray(listOrReorder)) {
+          listOrReorder = reorderArray(listOrReorder, fromIndex, toIndex);
+        }
 
-      for (let i = 0; i < len; i++) {
-        children[i].style['transform'] = '';
-      }
+        for (let i = 0; i < len; i++) {
+          children[i].style['transform'] = '';
+        }
+      });
 
       selectedItemEl.style.transition = '';
       selectedItemEl.classList.remove(ITEM_REORDER_SELECTED);


sean-perkins:
Thanks for this PR @dan-rp1. Couple items that will need to be addressed so that our team can consider it:
1. We avoid using `setTimeout` as a mechanism for delaying rendering (unless as a feature designed to operate after a specific duration). Can you try with our `raf` utility instead and see if waiting for the next available frame resolves the issue instead?
2. The original issue does not have a reproduction app to test against. We will need a reproduction app to confirm these changes function as expected.
3. Due to the nature of this bug, a test covering the expected behavior would be good to add here. 

Let me know if you have any questions!

liamdebeasi:
Thank you very much for the PR! We appreciate your contribution to Ionic Framework. I developed another solution that uses a `requestAnimationFrame` rather than a `setTimeout` which should help make sure that the DOM manipulation and the transform update happen in the same frame. I am going to close this PR in favor of https://github.com/ionic-team/ionic-framework/pull/26326, but I will be sure to give you co-author credit if the PR is merged. Let me know if you have any questions.

dan-rp1:
Thank you! Glad I could be involved.

PR #25497, opened 2022-06-17
chore(lerna): downgrade lerna for now

liamdebeasi:
## Pull request checklist

Please check if your PR fulfills the following requirements:
- [ ] Tests for the changes have been added (for bug fixes / features)
- [ ] Docs have been reviewed and added / updated if needed (for bug fixes / features)
  - Some docs updates need to be made in the `ionic-docs` repo, in a separate PR. See the [contributing guide](https://github.com/ionic-team/ionic-framework/blob/main/.github/CONTRIBUTING.md#modifying-documentation) for details.
- [ ] Build (`npm run build`) was run locally and any changes were pushed
- [ ] Lint (`npm run lint`) has passed locally and any fixes were made for failures

## Pull request type

 

 

Please check the type of change your PR introduces:
- [ ] Bugfix
- [ ] Feature
- [ ] Code style update (formatting, renaming)
- [ ] Refactoring (no functional changes, no api changes)
- [ ] Build related changes
- [ ] Documentation content changes
- [ ] Other (please describe): 

## What is the current behavior?

According to https://github.com/lerna/lerna/releases/tag/v5.0.0, we should be fine to upgrade to lerna v5. Our CI is not impacted by the Node version breaking changes, and we _should_ be fine with the `@npmcli/run-script` changes (we make use of `prepublishOnly`, but that appears to still be part of the API).

## What is the new behavior?

- There appear to be intermittent failures with the Ionic Angular Server package after these changes: https://github.com/ionic-team/ionic-framework/runs/6937044588?check_suite_focus=true

The error log is pretty minimal, so until I can dig more into this I am going to revert the Lerna change for now.

At first glance, it appears that Lerna 5 is running multiple builds in parallel, which is not what we need. Since Ionic angular server depends on Ionic angular, we need Ionic angular to build first. More investigation is needed.

We are possibly running into this bug: https://github.com/lerna/lerna/issues/3146

## Does this introduce a breaking change?

- [ ] Yes
- [x] No

PR #25682, opened 2022-07-27
chore(): update lerna to 5.3.0

JamesHenry:
Hi Folks 👋 

I'm one of the new maintainers of Lerna (via Nrwl).

As I noted on https://github.com/ionic-team/ionic-framework/pull/25497, I'm sorry that you had to initially downgrade from v5 back to v4.

I used Ionic's repo to troubleshoot a subtle difference between some of the implementation details we upgraded as part of v4->v5. Lerna uses a lot of packages produced by the npm team and in this case the package concerned with running lifecycle scripts needed to swapped out entirely after npm reimplemented it from scratch.

I added a missing piece of logic needed for the new package and released 5.3.0 of lerna just now, which addresses the issues you experienced before.

I even verified your full publishing flow locally by using a verdaccio instance instead of the real npm registry.

Please let me know if you have any questions or concerns! 🙏


liamdebeasi:
Thank you very much! Apologies for not reaching out sooner. I saw your comment on my previous PR, but it slipped my mind. I'll give this a test and let you know how it goes.

JamesHenry:
P.S. 

It is by no means a requirement, but one huge benefit you can opt into after this change goes in is to get lerna to delegate task running to Nx (instead of `p-map` and `p-queue` like it has done historically).

All you need to do over and above this PR is set `"useNx": true` in your `lerna.json` and everything will still work as it did before: `lerna run` will just get faster, have smart computation caching and allow for distributed task execution:

- https://lerna.js.org/docs/features/run-tasks
- https://lerna.js.org/docs/features/cache-tasks

Over and above that, the Nx Console VSCode extension is then able to analyze your lerna workspace and give you visualizations of how your projects relate to one another:

image

Again, all you need to do to get these things is set `"useNx": true` in your `lerna.json` (after merging this PR)

https://github.com/redwoodjs/redwood is an example of a popular OSS repo that is already leveraging these benefits with their lerna setup.

JamesHenry:
Awesome, thanks for the fast review and testing @liamdebeasi! Feel free to reach out if you have any queries or concerns post-merge! @mhartington can always reach me on slack too 🙂

liamdebeasi:
Merged. Thank you again!

Issue #18593, opened 2019-06-21
bug: NavController.pop() doesn't work when leave Tabs layout.


evgeniyefimov:
# Bug Report

**Ionic version:**

- [x] **4.5.0**
- [x] v5.x
- [x] v6.x

**Current behavior:**

When I call navCtrl.pop from a nested outlet which is not in a tab, it redirects back to the right tab which has some nav stack, but then navCtrl.pop doesn't work. With ion-back-button it works fine. 

**Expected behavior:**

NavContoller.pop() should be able to navigate to the correct page in nav stack as ion-back-button does.

**Steps to reproduce:**

https://github.com/sean-perkins/gh-18593-nav-pop

1. Start on: http://localhost:4200/tabs/tab1/item1
2. Click "Go to Item 2 Page"
3. Click "Go to Global Page"
4. Click "Go to Page 2 Programmatically"
5. Click "Go to Page 1 Programmatically"
6. Observe: Tab does not pop to the previous view
7. Click `ion-back-button` (< Item 2 Page)
8. Observe: Tab pops to the previous view

Expected:
- Clicking "Go to Page 1 Programmatically" should pop to the previous view in the tab

TrulsLoug:
Same issue here

NicolasVargas:
Any news about this issue ?

fhouenou:
Any update on this?

sean-perkins:
I've updated this issue with an up-to-date reproduction of the issue. 

An alternative pattern to avoid this issue is to use `IonRouterOutlet` as a provider instead of `NavController`.

For example, in the `Item2Page1` component:

```ts
@Component({
  selector: 'app-item21',
  templateUrl: 'item2.page.html',
})
export class Item2Page {
  constructor(@Optional() private ionRouterOutlet: IonRouterOutlet) { }

  async goBack() {
    return this.ionRouterOutlet.pop();
  }
}
```

Results in the "Go to Page 1 Programmatically" button popping to the correct view. 

My assessment of the issue is that `NavController` is not finding the correct top outlet when performing a pop operation. I will need to create a debug build to see why the `ion-router-outlet` implementation is not assigning the correct top outlet.

sean-perkins:
Ok the problem is here: https://github.com/ionic-team/ionic-framework/blob/main/angular/src/directives/navigation/ion-router-outlet.ts#L311-L315

[...]

PR #25690, opened 2022-07-28
fix(angular): nav controller can pop views after leaving tabs outlet

sean-perkins:
## Pull request checklist

## Pull request type

 

 

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

The Angular routing implementation had a race condition when re-entering a tabs outlet. The router outlet would be activated with both the application root `ion-router-outlet` as well as the tab's router outlet. 

Setting the active view waits for the entering view to finish transitioning. Due to how we deterministically set the top outlet, this would always cause the root router outlet being assigned as the top outlet, even though the tabs router outlet was technically the top outlet. 

This results in the `navController.pop()` attempting to pop a view from the top outlet (the root outlet), which does not have views in the stack to remove. 

Issue URL: #18593

## What is the new behavior?

- The top outlet is set synchronously, to avoid a race condition of the top outlet being set incorrectly (when dealing with nested outlets).

- [ ] Yes
- [x] No

## Other information

Dev-build: `6.2.5-dev.11661886076.141059a8`


angular/src/directives/navigation/ion-router-outlet.ts
```diff
@@ -308,10 +308,10 @@ export class IonRouterOutlet implements OnDestroy, OnInit {
     }
 
     this.activatedView = enteringView;
-    this.stackCtrl.setActive(enteringView).then((data) => {
-      this.navCtrl.setTopOutlet(this);
+
+    this.navCtrl.setTopOutlet(this, this.stackCtrl.setActive(enteringView)).subscribe((stackEvent) => {
```
sean-perkins, line 312: ~ An alternative pattern here that may be more clear, is having an API like:
```ts
this.navCtrl.setTopOutlet(this, enteringView)
```

Then providing the stack controller to the nav controller, so that it can call the `setActive` method. This would cut down on passing around a promise, which is an uncommon pattern. 

I will explore this more.

sean-perkins: Opted against this. The `stackCtrl` isn't an Angular provider, it's a constructed class local to each router outlet. Trying to pass the instance through to the `setTopOutlet` implementation led to failing tests and inconsistent results.

sean-perkins:
Ok @liamdebeasi @amandaejohnston this is ready for review. I'm creating a new dev-build, but based on the age of the issue I do not believe we will get testers. We have added test cases to account for expected behavior in the test apps.

amandaejohnston (approved):

liamdebeasi (approved):

Issue #25507, opened 2022-06-21
bug: react 18, page hides on hot reload


philjones88:
### Prerequisites

- [X] I have read the [Contributing Guidelines](https://github.com/ionic-team/ionic-framework/blob/main/.github/CONTRIBUTING.md#creating-an-issue).
- [X] I agree to follow the [Code of Conduct](https://ionicframework.com/code-of-conduct).
- [X] I have searched for [existing issues](https://github.com/ionic-team/ionic-framework/issues) that already report this problem, without success.

### Ionic Framework Version

- [ ] v4.x
- [ ] v5.x
- [X] v6.x
- [ ] Nightly

### Current Behavior

The page works fine but when you save a file and the react server reloads part of the UI the react router outlet (`IonRouterOutlet`) goes blank.

### Expected Behavior

The UI to be updated and not blank

### Steps to Reproduce

I put together this code sandbox.

Simply change say `hello` to `hello!` and save the file, the page goes blank and you see `UNMOUNTED` in the console but no `MOUNTED`

This also happens locally with the CRA dev server.

https://codesandbox.io/s/ionic-router-issue-test-ovee4z?file=/src/app.tsx

i.e. this is the state after a save:

<img width="625" alt="image" src="https://user-images.githubusercontent.com/677167/174802355-cc7cd1f0-3ddf-4f8c-a270-bb516e4a4d8d.png">

You then have to physically refresh the browser page to get the UI to work again.

### Code Reproduction URL

https://codesandbox.io/s/ionic-router-issue-test-ovee4z?file=/src/app.tsx

### Ionic Info

Can't do this on a sandbox, but locally with the same issue:

```
➜  broadway git:(main) ✗ ionic info
[WARN] Error loading @capacitor/android package.json: Error: Cannot find module '@capacitor/android/package'

       Require stack:
       - /Users/philjones/.config/yarn/global/node_modules/@ionic/cli/lib/project/index.js
       - /Users/philjones/.config/yarn/global/node_modules/@ionic/cli/lib/index.js
       - /Users/philjones/.config/yarn/global/node_modules/@ionic/cli/index.js
       - /Users/philjones/.config/yarn/global/node_modules/@ionic/cli/bin/ionic

Ionic:

   Ionic CLI       : 6.18.1 (/Users/philjones/.config/yarn/global/node_modules/@ionic/cli)
   Ionic Framework : @ionic/react 6.1.10

Capacitor:

   Capacitor CLI      : 3.6.0
[...]

PR #25703, opened 2022-08-02
fix(react): outlet will not clear in react 18 with hot reload

sean-perkins:
## Pull request checklist

## Pull request type

 

 

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

In React 18, components are mounted/unmounted in dev build to check for side effects. The `StackManager` implementation has logic to clear the outlet when it is unmounted. Unfortunately this implementation of clearing the outlet has an inner `setTimeout` that will resolve 500ms after it is invoked. 

This results in:
- Page Mounted
- Page Unmounted
  - clearOutlet / setTimeout
- Page Remounted
  - setTimeout resolves and clears the page
- Page is blank

This is only a problem in React 18 & in development mode. Noticeable with hot reloading, as any changes will clear the view. 

Issue URL: #25507

## What is the new behavior?

- Returns the timeout to the implementation layer so that we can cancel the operation when the component is re-mounted in dev-mode (created tech debt task in backlog to investigate removing timeout in React Router v6 work)
- React 18 and hot reload will no longer clear the page/outlet when the page remounts

- [ ] Yes
- [x] No

## Other information

Dev-build: `6.2.1-dev.11659463817.1943a7f4` ✅  (need to install for both `@ionic/react` and `@ionic/react-router`)


amandaejohnston (approved):

packages/react-router/src/ReactRouter/StackManager.tsx
```diff
@@ -197,11 +216,16 @@ export class StackManager extends React.PureComponent<StackManagerProps, StackMa
     const canStart = () => {
       const config = getConfig();
       const swipeEnabled = config && config.get('swipeBackEnabled', routerOutlet.mode === 'ios');
-      if (!swipeEnabled) { return false; }
+      if (!swipeEnabled) {
+        return false;
+      }
```
liamdebeasi, line 221: Are these changes from prettier or from your built in IDE? If they are from the IDE, could we remove them?

liamdebeasi (approved):

packages/react-router/src/ReactRouter/StackManager.tsx
sean-perkins: These changes are from prettier. The extension detects the local configuration: https://github.com/ionic-team/ionic-framework/blob/main/packages/react-router/.prettierrc.json and is formatting based off that.

Unfortunately the React packages do not actually run prettier as part of the lint step at all.

liamdebeasi: Oh right! Never mind 👍

PR #25713, opened 2022-08-03
test(radio-group): migrate tests to playwright

sean-perkins:
## Pull request checklist

## Pull request type

 

 

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes
- [x] Other (please describe): 

N/A

## What is the new behavior?

- Migrate tests to Playwright
- Removes unnecessary interactivity from basic test 

- [ ] Yes
- [x] No


core/src/components/radio-group/test/basic/radio-group.e2e.ts
```diff
@@ -0,0 +1,110 @@
...
+import { test } from '@utils/test/playwright';
+import type { E2EPage } from '@utils/test/playwright';
+
+test.describe('radio-group: basic', () => {
+  test('should not have visual regressions', async ({ page }) => {
+    await page.goto(`/src/components/radio-group/test/basic`);
+
+    await page.setIonViewport();
+
+    expect(await page.screenshot()).toMatchSnapshot(`radio-group-diff-${page.getSnapshotSettings()}.png`);
```
liamdebeasi, line 12: Do we need to screenshot the entire page, or can we screenshot just `ion-radio-group`?

core/src/components/radio-group/test/form/radio-group.e2e.ts
```diff
@@ -0,0 +1,31 @@
...
+  test('selecting a disabled option should not update the value', async ({ page }) => {
+    const value = page.locator('#value');
+    const disabledRadio = page.locator('ion-radio[value="george"]');
+
+    await expect(value).toHaveText('');
+    await expect(disabledRadio).toHaveAttribute('disabled', '');
+
+    await disabledRadio.click({ force: true });
+
+    await expect(value).toHaveText('');
```
liamdebeasi, line 29: I don't think this would reliably catch a regression because we don't wait for Stencil to update the state of the component. We may need to do `await page.waitForChanges()` after we click.

liamdebeasi (approved):
Good on my end after the 2 comments below are addressed

core/src/components/radio-group/test/form/radio-group.e2e.ts
```diff
@@ -0,0 +1,31 @@
+import { expect } from '@playwright/test';
+import { test } from '@utils/test/playwright';
+
+test.describe('radio-group: form', () => {
```
amandaejohnston, line 4: Would it make sense to use `toHaveReceivedEventDetail` instead of checking the `#value` text? We could cut out the middleman that way.

core/src/components/radio-group/test/form/radio-group.e2e.ts
```diff
@@ -0,0 +1,34 @@
...
+
+    await expect(radioGroup).toHaveAttribute('value', 'biff');
+
+    await griffRadio.click();
+    await page.waitForChanges();
+
+    await expect(ionChange).toHaveReceivedEventDetail({ value: 'griff' });
+  });
+
+  test('selecting a disabled option should not update the value', async ({ page }) => {
```
amandaejohnston, line 22: May want to use `toHaveAttribute`/`toHaveReceivedEventDetail` on this test too? Otherwise this LGTM 👍

amandaejohnston (approved):

core/src/components/radio-group/test/form/radio-group.e2e.ts
sean-perkins: The reason I opted against this for this test case is:
1. When the radio is disabled, `ionChange` will not be emitted, so we cannot check for the spy to have received an event.
2. When the value is changed, the reflected value on the `ion-radio-group` is not updated. Checking `toHaveAttribute` would capture no change from interaction. 

I am unsure if 2 is a bug or not. It currently is the same behavior as `ion-input` (the reflected value doesn't match user interaction unless set manually/from a framework) & consistent with how the built-in input works.

Issue #25447, opened 2022-06-10
bug: Overlays dismiss() method can cause getTop() to return dismissed instance


andreaslarssen:
### Prerequisites

- [X] I have read the [Contributing Guidelines](https://github.com/ionic-team/ionic-framework/blob/main/.github/CONTRIBUTING.md#creating-an-issue).
- [X] I agree to follow the [Code of Conduct](https://ionicframework.com/code-of-conduct).
- [X] I have searched for [existing issues](https://github.com/ionic-team/ionic-framework/issues) that already report this problem, without success.

### Ionic Framework Version

- [ ] v4.x
- [ ] v5.x
- [X] v6.x
- [ ] Nightly

### Current Behavior

I want to show a loading screen while my (Angular) app navigates. I solve this by adding and removing a loading screen (via LoadingController) on navigation start and navigation end (as well as cancel and error).

It usually works, _but_: 

I have a parent route w/ children routes. The parent route redirects the user to a specific child (the first) each time it's loaded. That means, if the user refreshes the browser when standing on another child than the first, the parent route loads first and redirects them to the first child. Standard stuff. Let me just say that there's no indication that this is what's causing the problem, just consider it context. 

This is the service method that should close the loading indicator(s):

```
private closeLoadingIndicators(): void {
    this.loadingController.getTop().then((loadingIndicator: HTMLIonLoadingElement | undefined) => {
      console.log('loadingIndicator:', loadingIndicator);
      console.log('undefined:', loadingIndicator === undefined);
      console.log('typeof:', typeof loadingIndicator);
      console.log('present:', loadingIndicator?.present);
      console.log('Removing loading screen');

      if (loadingIndicator) {
        loadingIndicator.dismiss().then(() => {
          this.closeLoadingIndicators();
        });
      }
    })
  }
```

The idea is that the DOM can have several loading indicator present, and as long as it has, remove them one by one.
The output from the `console.log()`s are as following:

![Screenshot from 2022-06-10 13-47-49](https://user-images.githubusercontent.com/3169836/173058274-eca01802-5ddb-4b00-b60c-98a68df56b9b.png)

As you can see, the 'loadingIndicator' is nothing(?), but the type seems to be `object`.
Even though the instance seems to be nothing, the if check renders true, and the app re-runs the method, causing an endless loop.

[...]

PR #25718, opened 2022-08-04
fix(overlays): dismissed overlays will resolve in the next frame

sean-perkins:
## Pull request checklist

## Pull request type

 

 

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

With a very specific configuration of presenting and dismissing overlays, occasionally `getTop()` will return the overlay that was just dismissed. When used with recursion (where you are trying to dismiss all overlays until there is no more), this creates problems/infinite loops. 

Issue URL: #25447

## What is the new behavior?

- The `dismiss()` API will resolve in the next frame, after the element has completed the transition and been removed from the DOM.
- This improves the consistency of `getTop()` to return the active visible, top overlay. 

- [ ] Yes
- [x] No

## Other information

You can verify this behavior in this reproduction app: https://github.com/sean-perkins/gh-25447-overlay-get-top it is a fork of the original reproduction, with safe guards to avoid an infinite loop locking up the browser tab. 

The `main` branch is on 6.2.0 and will highlight the problem in console (the infinite loop occurs, but bails with an error).

You can install the dev-build and see this behavior does not happen and the loading indicator properly dismisses. 

Dev-build: `6.2.2-dev.11659644639.1b018cb5`

Writing a test case around this problem was unsuccessful (without copying all the bolierplate from the reproduction app into our test app). I do not believe the value of covering this test condition is worth maintaining all the boilerplate.


amandaejohnston (approved):

core/src/utils/overlays.ts
```diff
@@ -515,7 +515,14 @@ export const dismiss = async (
   }
 
   overlay.el.remove();
-  return true;
+
+  /**
+   * Waits 1 frame after the element has been removed from the DOM to resolve.
+   * Otherwise getTop() may return the dismissed overlay as the topmost overlay.
+   */
+  return new Promise((resolve) => {
```
liamdebeasi, line 523: Can you clarify why this is needed? When the overlay is dismissed, we add the `overlay-hidden` class to the host element: https://github.com/ionic-team/ionic-framework/pull/25718/files#diff-e6fe3d2c2375aaff02644f8767f36a3c49825d333f9764644871b0541a95b296R511

The `getOverlay` function, which is called by `getTop` filters out hidden overlays: https://github.com/ionic-team/ionic-framework/blob/f9ab9b54ddb5a3004673e4aaa9cb62fd8e97ba07/core/src/utils/overlays.ts#L350

It does this by excluding overlays that have the `overlay-hidden` class: https://github.com/ionic-team/ionic-framework/blob/f9ab9b54ddb5a3004673e4aaa9cb62fd8e97ba07/core/src/utils/overlays.ts#L105

If we add the `overlay-hidden` class synchronously, why does `getTop` still return a visible overlay?

sean-perkins: The issue behind this change is triggered by trying to present and dismiss overlays during a SPA navigation event. On navigation start the loading overlay is displayed and on navigation end it is dismissed. For some reason only in Angular, dismissing an overlay in the navigation end recursively through `getTop()`, resolves the dismissed overlay (the classlist has not been mutated at the point in time it is resolved). 

Resolving dismiss into the next frame, prevents this side effect. 

Most of the implementation for the reproduction is here: https://github.com/sean-perkins/gh-25447-overlay-get-top/blob/main/src/app/navigation.service.ts

If we decided against this change, the implementer would need to `requestAnimationFrame` the dismiss method to have the proper overlay resolve from `getTop()`.

liamdebeasi: (Apologies in advance for the essay 😅)

Thanks for the additional context, that was really helpful. I tested using the dev build, but I can still reproduce the infinite loop issue. I used your sample app to test.

------

I took a closer look at why this issue is happening and it looks like `closeLoadingIndicators` is getting called while another `closeLoadingIndicators` operation is in progress. So when navigation ends on `/app/child-2` `closeLoadingIndicators` gets called which begins to dismiss the open loading overlay. During this time `overlay.presented` is set to false:

https://github.com/ionic-team/ionic-framework/blob/abb56d22b4a81d1bc34c689de4ef7218e7503b20/core/src/utils/overlays.ts#L484

However, when the nested router outlet's `ngOnInit` fires, it redirects the app to `/app/child-1`. When the navigation ends on `/app/child-1` `closeLoadingIndicators` gets called again which begins to dismiss the same loading overlay.

Since `overlay.presented` is already `false`, this function returns early: https://github.com/ionic-team/ionic-framework/blob/abb56d22b4a81d1bc34c689de4ef7218e7503b20/core/src/utils/overlays.ts#L478

As a result, `closeLoadingIndicators` is called again and the problem repeats until the loop limit is reached.

I think this is why the `requestAnimationFrame` implemented in the dev's app works. It delays the next `dismiss` call just enough until the first `dismiss` call can fully resolve.

------

If you add logs immediately before and after `loadingIndicator.dismiss()` in `navigation.service.ts` you should see that the first dismiss operation does not resolve until the loop limit is reached. You should see logs that look something like this:

```
begin dismiss <-- first dismiss operation start
begin dismiss
end dismiss
begin dismiss
end dismiss

...

begin dismiss
end dismiss
end dismiss <-- first dismiss operation end
```

-----

At what point is an overlay considered no longer active? (When the dismiss operation begins or when it ends) We set `overlay.presented = false` right when the dismiss operation begins which makes me think that an overlay should no longer be considered active at that point too.

I wonder if we could add a `dismissing` class to the overlay or similar thing right before the async overlay animation is played:

https://github.com/ionic-team/ionic-framework/blob/abb56d22b4a81d1bc34c689de4ef7218e7503b20/core/src/utils/overlays.ts#L499

From there, we could exclude overlays with `.dismissing` from the `getTop` method. Thoughts?

Issue #17560, opened 2019-02-21
Footer toolbar should not have safe-area-inset-bottom with tab-bar active


Spinnenzunge:
# Bug Report
Footer toolbar adds safe-area-inset-bottom padding although it was already added for ion-tab-bar on iPhone X

**Ionic version:**

[x] **4.x**

**Current behavior:**
Padding in footer toolbar added beneath buttons, that should not be added with tab-bar visible.
![image from ios](https://user-images.githubusercontent.com/11189008/53170190-ca541d00-35df-11e9-91f9-2ee1cb90a5f8.png)

**Expected behavior:**
No safe-area-inset-bottom padding on toolbar in ion-footer if there is ion-tab-bar on bottom already.

**Steps to reproduce:**
Create ionic 4 tab starter and add ion-footer with ion-toolbar inside.

**Other information:**
Bug was already opened for Ionic 3 once: https://github.com/ionic-team/ionic/issues/13615

**Ionic info:**

```
Ionic:

   ionic (Ionic CLI)             : 4.10.3 (/usr/local/lib/node_modules/ionic)
   Ionic Framework               : @ionic/angular 4.0.2
   @angular-devkit/build-angular : 0.12.4
   @angular-devkit/schematics    : 7.2.4
   @angular/cli                  : 7.2.4
   @ionic/angular-toolkit        : 1.4.0

Cordova:

   cordova (Cordova CLI) : 7.1.0
   Cordova Platforms     : ios 5.0.0
   Cordova Plugins       : cordova-plugin-ionic-keyboard 2.1.3, cordova-plugin-ionic-webview 3.1.2, (and 4 other plugins)

System:

   Android SDK Tools : 26.1.1 (/Users/pascalgraf/Library/Android/sdk)
   ios-deploy        : 1.9.4
   ios-sim           : 7.0.0
   NodeJS            : v10.14.1 (/usr/local/bin/node)
   npm               : 6.7.0
   OS                : macOS Mojave
   Xcode             : Xcode 10.1 Build version 10B61
```

Spinnenzunge:
Could be fixed with following line of code
```
ion-tabs {
  ion-footer ion-toolbar:last-child {
    padding-bottom: 0;
  }
}
```

sebaferreras:
Something similar happens when opening the keyboard... The native keyboard already adds the `safe-area-inset-bottom` so there's no need to add it to the `ion-footer` as well:

[...]

PR #23193, opened 2021-04-15
fix(footer): Remove Toolbar padding if in Tabs (bottom) or Keyboard opened

EinfachHans:
closes #17560

## Pull request checklist

Please check if your PR fulfills the following requirements:
- [X] Tests for the changes have been added (for bug fixes / features)
- [ ] Docs have been reviewed and added / updated if needed (for bug fixes / features)
- [X] Build (`npm run build`) was run locally and any changes were pushed
- [X] Lint (`npm run lint`) has passed locally and any fixes were made for failures

## Pull request type

 

 

Please check the type of change your PR introduces:
- [X] Bugfix
- [ ] Feature
- [ ] Code style update (formatting, renaming)
- [ ] Refactoring (no functional changes, no api changes)
- [ ] Build related changes
- [ ] Documentation content changes
- [ ] Other (please describe): 

## What is the current behavior?

Currently the Ion Footer Elements always adds `padding-bottom: var(--ion-safe-area-bottom, 0);` to the last Ion Toolbar.

Issue Number: #17560

## What is the new behavior?

The Style is not applied if:
- The Keyboard is open ([See comment](https://github.com/ionic-team/ionic-framework/issues/17560#issuecomment-573059275))
- The Footer is in a `<ion-tabs>` Element with an at the bottom placed `<ion-tab-bar>`

## Does this introduce a breaking change?

- [ ] Yes
- [X] No

PR #25746, opened 2022-08-10
fix(footer): remove toolbar bottom padding if near bottom slot tabs or keyboard is open

amandaejohnston:
## Pull request checklist

## Pull request type

 

 

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

Currently, an `ion-toolbar` inside an `ion-footer` receives bottom padding according to the safe area, via [this CSS](https://github.com/ionic-team/ionic-framework/blob/ab65e9a7b51c3a3f8c59962d3e1faff1564ab801/core/src/components/footer/footer.scss#L17-L19). However, this padding becomes unnecessary if there is also an `ion-tab-bar` on the bottom of the screen, as it also has its own safe area bottom padding.

extra padding

This situation also applies when you have the phone’s keyboard open, with or without an `ion-tab-bar`.

Issue URL: Resolves https://github.com/ionic-team/ionic-framework/issues/17560

## What is the new behavior?

Extra bottom padding removed if there is a `slot="bottom"` tab bar nearby, or if the keyboard is open. (The keyboard logic is the same as in `ion-tab-bar`.)

This PR also pulls the `keyboardVisible` logic shared between `ion-footer` and `ion-tab-bar` into a separate `createKeyboardController` function.

- [ ] Yes
- [x] No

## Other information

This PR is adapted from https://github.com/ionic-team/ionic-framework/pull/23193. Co-author credit will be given when merging.

I left off a test for the keyboard being open since I don't believe that's possible in Playwright 🤔


core/src/components/footer/footer.tsx
```diff
@@ -46,6 +51,30 @@ export class Footer implements ComponentInterface {
...
   }
 
+  connectedCallback() {
+    if (typeof (window as any) !== 'undefined') {
+      this.keyboardWillShowHandler = () => {
+        this.keyboardVisible = true;
+      };
+
+      this.keyboardWillHideHandler = () => {
+        setTimeout(() => (this.keyboardVisible = false), 50);
```
sean-perkins, line 61: Is there any way we can avoid a `setTimeout` here? Would request animation frame wrapping this callback work instead?

If not, can we document why the set timeout is used and why the value of `50` was selected? 

Edit: This also looks like maybe copy & paste from the tab bar. Can they share this behavior and also can we investigate what drove a the setTimeout & duration?

core/src/components/footer/footer.tsx
```diff
@@ -20,6 +20,11 @@ export class Footer implements ComponentInterface {
   private scrollEl?: HTMLElement;
   private contentScrollCallback: any;
 
+  private keyboardWillShowHandler?: () => void;
+  private keyboardWillHideHandler?: () => void;
+
+  @State() keyboardVisible = false;
```
sean-perkins, line 26: ~ Can this state variable be `private`?

core/src/components/footer/footer.tsx
```diff
@@ -46,6 +51,30 @@ export class Footer implements ComponentInterface {
     this.checkCollapsibleFooter();
   }
 
+  connectedCallback() {
+    if (typeof (window as any) !== 'undefined') {
```
sean-perkins, line 55: Can we avoid casting to `any` here? I know there are a few places in core where we have a similar check out casting; although those may be `.ts` files instead of `.tsx`. 

Edit: This looks like maybe the result of copy & paste from the tab bar component. I do not think the type cast is necessary here (or in tab bar).

core/src/components/footer/footer.tsx
```diff
@@ -94,6 +123,9 @@ export class Footer implements ComponentInterface {
   render() {
     const { translucent, collapse } = this;
     const mode = getIonMode(this);
+    const tabs = this.el.closest('ion-tabs');
+    const tabBar = tabs ? tabs.querySelector('ion-tab-bar') : null;
```
sean-perkins, line 127: Can the tab bar selector be simplified to:
```ts
const tabBar = tabs?.querySelector('ion-tab-bar');
```

or do you need to the type signature to be `HTMLElement | null` instead of `HTMLElement | null | undefined`?

sean-perkins (changes requested):

sean-perkins:
One possible approach to sharing the logic across both implementations would be through a controller, similar to how we handle modal, popover, menu, etc.

```ts
const createKeyboardController = () => {
  let keyboardWillShowHandler: (() => void) | undefined;
  let keyboardWillHideHandler: (() => void) | undefined;
  let keyboardVisible: boolean;

  const init = (el?: HTMLElement) => {
    keyboardWillShowHandler = () => {
      if (el && el.getAttribute('slot') !== 'top') {
        keyboardVisible = true;
      }
    };

    keyboardWillHideHandler = () => {
      setTimeout(() => (keyboardVisible = false), 50);
    };

    window.addEventListener('keyboardWillShow', keyboardWillShowHandler!);
    window.addEventListener('keyboardWillHide', keyboardWillHideHandler!);
  }

  const destroy = () => {
    if (typeof (window as any) !== 'undefined') {
      window.removeEventListener('keyboardWillShow', keyboardWillShowHandler!);
      window.removeEventListener('keyboardWillHide', keyboardWillHideHandler!);

      keyboardWillShowHandler = keyboardWillHideHandler = undefined;
    }
  }

  const isKeyboardVisible = () => keyboardVisible;

  return { init, destroy, isKeyboardVisible };
}
```

```ts

private keyboardCtrl: any; // TODO type

connectedCallback() {
  this.keyboardCtrl = createKeyboardController();
  this.keyboardCtrl.init(this.el);
}

disconnectedCallback() {
  if (this.keyboardCtrl) {
    this.keyboardCtrl.destroy();
  }
}
```

In your implementation you need to trigger a re-render when the keyboard visibility changes. For this, you may need to pass in a callback that can trigger your state variable to re-render.

core/src/components/footer/footer.tsx
liamdebeasi: This is an SSR check to make sure that we do not access `window` when it is not available. However, it would be better to use https://github.com/ionic-team/ionic-framework/blob/main/core/src/utils/window/index.ts so that we do not need to keep doing this check everywhere.

core/src/utils/keyboard/keyboard-controller.ts
```diff
@@ -0,0 +1,45 @@
+import { win } from '@utils/window';
```
liamdebeasi, line 1: We can't use relative imports in non-test code until https://github.com/ionic-team/stencil/issues/2815 is resolved.

core/src/components/footer/footer.tsx
```diff
@@ -1,5 +1,7 @@
 import type { ComponentInterface } from '@stencil/core';
-import { Component, Element, Host, Prop, h } from '@stencil/core';
+import { Component, Element, Host, Prop, State, h } from '@stencil/core';
+import type { KeyboardController } from '@utils/keyboard/keyboard-controller';
+import { createKeyboardController } from '@utils/keyboard/keyboard-controller';
```
liamdebeasi, line 4: We can't use relative imports in non-test code until https://github.com/ionic-team/stencil/issues/2815 is resolved.

core/src/components/footer/test/with-tabs/footer.e2e.ts
```diff
@@ -0,0 +1,11 @@
+import { expect } from '@playwright/test';
+import { test } from '@utils/test/playwright';
+
+test.describe('footer: with tabs', () => {
+  test('should not have extra padding when near a tab bar', async ({ page }) => {
+    await page.goto('/src/components/footer/test/with-tabs');
```
liamdebeasi, line 6: Can we skip the RTL tests?

core/src/components/tab-bar/tab-bar.tsx
```diff
@@ -59,43 +60,32 @@ export class TabBar implements ComponentInterface {
...
+    });
 
-      this.keyboardWillHideHandler = () => {
-        setTimeout(() => (this.keyboardVisible = false), 50);
-      };
-
-      window.addEventListener('keyboardWillShow', this.keyboardWillShowHandler!);
-      window.addEventListener('keyboardWillHide', this.keyboardWillHideHandler!);
-    }
+    this.keyboardCtrl.init();
```
liamdebeasi, line 67: What's the reasoning for having an explicit init call rather than having the controller auto-init?

core/src/components/footer/footer.tsx
```diff
@@ -94,6 +113,9 @@ export class Footer implements ComponentInterface {
   render() {
     const { translucent, collapse } = this;
     const mode = getIonMode(this);
+    const tabs = this.el.closest('ion-tabs');
+    const tabBar = tabs?.querySelector('ion-tab-bar');
```
liamdebeasi, line 117: Do we need to query for this every render, or are we able to grab it in `connectedCallback`?

core/src/components/tab-bar/tab-bar.tsx
amandaejohnston: Hm, I suppose the separate init isn't needed if the controller is created at the same time 🤔 I'll make it auto-init.

core/src/components/footer/footer.tsx
amandaejohnston: Querying on every render seems safer since it lets us react to a tab bar being added dynamically later, but I don't feel that strongly about it if you'd rather take the better performance.

core/src/utils/keyboard/keyboard-controller.ts
```diff
@@ -0,0 +1,46 @@
+import { win } from '../window';
+
+/**
+ * Creates a controller that tracks and reacts to opening or closing the keyboard.
+ *
+ * @internal
+ * @param keyboardChangeCallback A function to call when the keyboard opens or closes.
+ */
+export const createKeyboardController = (keyboardChangeCallback?: (keyboardOpen: boolean) => void) => {
```
liamdebeasi, line 9: Can we write some spec tests for this? We can use a jest function mock to check if the callback is being called.

core/src/components/footer/test/with-tabs/footer.e2e.ts
```diff
@@ -0,0 +1,13 @@
...
+import { test } from '@utils/test/playwright';
+
+test.describe('footer: with tabs', () => {
+  test('should not have extra padding when near a tab bar', async ({ page }, testInfo) => {
+    test.skip(testInfo.project.metadata.rtl === true, 'This does not test LTR vs. RTL layout.');
+
+    await page.goto('/src/components/footer/test/with-tabs');
+
+    const footer = page.locator('[tab="tab-one"] ion-footer');
+    expect(await footer.screenshot()).toMatchSnapshot(`footer-with-tabs-${page.getSnapshotSettings()}.png`);
```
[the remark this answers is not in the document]
amandaejohnston: The screenshot test seems safer to me since we're actually checking for whether there's extra space present.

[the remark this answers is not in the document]
liamdebeasi: That works. Can we at least skip the RTL screenshots though?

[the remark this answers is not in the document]
amandaejohnston: Yup, already done 👍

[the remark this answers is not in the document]
liamdebeasi: Oof I can't read 🤦 looks like you already did that.  thank you!

liamdebeasi (approved):
Great job! This is good to go once tests for the keyboard controller are written.

core/src/utils/keyboard/keyboard-controller.ts
sean-perkins, line 9: Can we update the return type of this function so that it uses the custom `KeyboardController` type?

sean-perkins (changes requested):

sean-perkins (approved):

Issue #25495, opened 2022-06-16
bug: 


WillooWisp:
### Prerequisites

- [X] I have read the [Contributing Guidelines](https://github.com/ionic-team/ionic-framework/blob/main/.github/CONTRIBUTING.md#creating-an-issue).
- [X] I agree to follow the [Code of Conduct](https://ionicframework.com/code-of-conduct).
- [X] I have searched for [existing issues](https://github.com/ionic-team/ionic-framework/issues) that already report this problem, without success.

### Ionic Framework Version

- [ ] v4.x
- [ ] v5.x
- [X] v6.x
- [ ] Nightly

### Current Behavior

Refresher only works, gets visible, on active page upon reload.

### Expected Behavior

Expects refresher to be visible on every page upon refresh, even if not active page upon refresh.

### Steps to Reproduce

Use the tabs sample. Add a cdk virtual scroll list with refresher to each single tab page. Refresher only works on one of the pages, the one being active upon refresh in browser.

### Ionic Info

Ionic:

   Ionic CLI                     : 6.19.0 (npm\node_modules\@ionic\cli)
   Ionic Framework               : @ionic/angular 6.1.9
   @angular-devkit/build-angular : 14.0.1
   @angular-devkit/schematics    : 13.3.5
   @angular/cli                  : 14.0.1
   @ionic/angular-toolkit        : 6.1.0

Capacitor:

   Capacitor CLI      : 3.5.1
   @capacitor/android : 3.5.1
   @capacitor/core    : 3.5.1
   @capacitor/ios     : 3.5.1

Utility:

   cordova-res : 0.15.2
   native-run  : 1.5.0

System:

   NodeJS : v14.17.5 (C:\Program Files\nodejs\node.exe)
   npm    : 7.22.0
   OS     : Windows 10

### Additional Information

This issue https://github.com/ionic-team/ionic-framework/issues/25318 solved the problem where the refresher got visible too often, not sure if this bug fix actually causes this new problem.

TimGels:
I also noticed this. For me the refresher has this issue in iOS mode where the refresher isnt visible on the second (or any other) page I try to use the refresher. It does fire the refresh event but the animation is not visible. On md mode the refresher appears to be working for me on all my pages.

WillooWisp:
Here is a sample reproducing the problem, https://stackblitz.com/edit/ionic6-angular13-bat2vm.
[...]

PR #25750, opened 2022-08-11
fix(refresher): refresher is visible with multiple custom scroll targets

sean-perkins:
## Pull request checklist

## Pull request type

 

 

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

When using refresher with multiple instances of CDK virtual scroll and custom scroll target definitions, the pull to refresh behavior will trigger, but the refresher content will not be visible. 

This was due to the background content element resolving as `null`, instead of the `ion-content`'s background content element. This behavior seems to be derived from the refresher being slotted into the light DOM and the parent element not being fully rendered when the connected callback fires for the refresher. 

Issue URL: #25495

## What is the new behavior?

- Refresher will wait for the parent `ion-content` element to be ready, before querying the scroll and background content element. 
- Introduces an internal API to `ion-content` for querying the background content element 

- [ ] Yes
- [x] No

## Other information

Dev-build: `6.2.2-dev.11660183625.1a3e60c7`.


core/src/components/refresher/refresher.tsx
```diff
@@ -443,34 +443,28 @@ export class Refresher implements ComponentInterface {
...
-     * custom scroll target host.
-     *
-     * This makes it so that implementers do not need to re-create the background content
-     * element and styles.
+     * Query the background content element from the host ion-content element directly.
      */
-    this.backgroundContentEl = getElementRoot(contentEl ?? customScrollTarget).querySelector(
-      '#background-content'
-    ) as HTMLElement;
+    this.backgroundContentEl = await contentEl.getBackgroundElement();
```
liamdebeasi, line 467: Do we still need to account for custom scroll targets that create their own `#background-content` element?

sean-perkins: Nope. Technically they never could based on how I originally implemented it. We have a check above this that requires an `ion-content` element or prints an error and returns early. The check for `contentEl ?? customScrollTarget` would always resolve the `ion-content`. 

In hindsight, this is how we expected it to behave. We did not want developers ejecting from `ion-content` and also needing to provide the background content element. It just got lost in translation while trying to explore the alternative bug that overlapped with much of that work.

liamdebeasi (approved):

amandaejohnston (approved):

PR #25758, opened 2022-08-12
test(playwright): add new utilities for skipping tests

liamdebeasi:
## Pull request checklist

## Pull request type

 

 

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes

Issue URL:  Internal ticket

There are many instances where we need to skip tests. For example, a particular test may only need to be run in a certain browser or on a certain mode. The current syntax, while functional, contains more boilerplate than is necessary. As a result, there is quite a bit of duplicate conditionals across tests.

## What is the new behavior?

- Introduces a new `skip` fixture on the `test` function. This fixture can be used to skip tests based on text direction, mode, or browser. See the readme in this PR for examples/docs.

- [ ] Yes
- [x] No


liamdebeasi:
Unresolved question: Do we want to require a reason with this syntax? Playwright does not require a reason when doing `test.skip`.

amandaejohnston:
I think it'd be good if there were default/fallback reasons (with the param in the `skip` function being optional), since we use something like "This does not test RTL-specific logic" and "This feature is only available in x mode" for the majority of cases.

liamdebeasi:
Updated, thanks!

amandaejohnston (approved):
~ We can probably update most of the existing skips to use the default values, but this LGTM as-is. Thanks for doing this!

sean-perkins (approved):

Issue #25797, opened 2022-08-22
bug: react 18 with createRoot causes duplicate page transitions


liamdebeasi:
### Prerequisites

- [X] I have read the [Contributing Guidelines](https://github.com/ionic-team/ionic-framework/blob/main/.github/CONTRIBUTING.md#creating-an-issue).
- [X] I agree to follow the [Code of Conduct](https://ionicframework.com/code-of-conduct).
- [X] I have searched for [existing issues](https://github.com/ionic-team/ionic-framework/issues) that already report this problem, without success.

### Ionic Framework Version

- [ ] v4.x
- [ ] v5.x
- [X] v6.x
- [ ] Nightly

### Current Behavior

When using Ionic React + React 18 + the new createRoot API, duplicate page transitions occur. This is due to `IonPage` being unmounted and remounted by react. The `registerIonPage` function calls for a transition every time a page is registered: https://github.com/ionic-team/ionic-framework/blob/bd715a52562f1f175d4bb6ea2dbfdd67a3e91db1/packages/react-router/src/ReactRouter/StackManager.tsx#L212

However, it was written with the intent that page elements would be registered at most once.

### Expected Behavior

I expect there to be only 1 page transition.

### Steps to Reproduce

1. Clone repo.
2. Run `npm install`.
3. Run `npm run start`.
4. Click one of the list items in the app. Observe that the page transition occurs twice.

### Code Reproduction URL

https://github.com/liamdebeasi/react-18-transition

### Ionic Info
```
Ionic:

   Ionic CLI       : 6.20.1 (/Users/liamdebeasi/Ionic/ionic-cli/packages/@ionic/cli)
   Ionic Framework : @ionic/react 6.2.3

Capacitor:

   Capacitor CLI      : 4.1.0
   @capacitor/android : not installed
   @capacitor/core    : 4.1.0
   @capacitor/ios     : not installed

Utility:

   cordova-res : not installed globally
   native-run  : 1.6.0

System:

   NodeJS : v16.11.1 (/Users/liamdebeasi/.nvm/versions/node/v16.11.1/bin/node)
   npm    : 8.0.0
   OS     : macOS Monterey
```

liamdebeasi:
Resolved via https://github.com/ionic-team/ionic-framework/pull/25798

PR #25798, opened 2022-08-22
fix(react): duplicate page transitions do not happen on react 18

liamdebeasi:
## Pull request checklist

## Pull request type

 

 

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

Issue URL: resolves https://github.com/ionic-team/ionic-framework/issues/25797

## What is the new behavior?

- If we are mounting an IonPage element on a view that already has a reference to the element, do not fire a new page transition. This ensures that only the initial IonPage registration causes a transition.

- [ ] Yes
- [x] No

## Other information

Dev build: `6.2.4-dev.11661268305.10458734`

Manually verified that this is fixed on React 18 and does not regress the list starter app on React 17.

### DIFF
--- a/packages/react-router/src/ReactRouter/StackManager.tsx
+++ b/packages/react-router/src/ReactRouter/StackManager.tsx
@@ -206,8 +206,18 @@ export class StackManager extends React.PureComponent<StackManagerProps, StackMa
   registerIonPage(page: HTMLElement, routeInfo: RouteInfo) {
     const foundView = this.context.findViewItemByRouteInfo(routeInfo, this.id);
     if (foundView) {
+      const oldPageElement = foundView.ionPageElement;
       foundView.ionPageElement = page;
       foundView.ionRoute = true;
+
+      /**
+       * React 18 will unmount and remount IonPage
+       * elements in development mode when using createRoot.
+       * This can cause duplicate page transitions to occur.
+       */
+      if (oldPageElement === page) {
+        return;
+      }
     }
     this.handlePageTransition(routeInfo);
   }
--- a/packages/react/src/routing/OutletPageManager.tsx
+++ b/packages/react/src/routing/OutletPageManager.tsx
@@ -18,16 +18,26 @@ export class OutletPageManager extends React.Component<OutletPageManagerProps> {
   ionLifeCycleContext!: React.ContextType<typeof IonLifeCycleContext>;
   context!: React.ContextType<typeof StackContext>;
   ionRouterOutlet: HTMLIonRouterOutletElement | undefined;
+  outletIsReady: boolean;
 
   constructor(props: OutletPageManagerProps) {
     super(props);
+
+    this.outletIsReady = false;
   }
 
   componentDidMount() {
     if (this.ionRouterOutlet) {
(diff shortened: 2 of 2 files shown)


packages/react-router/src/ReactRouter/StackManager.tsx
```diff
@@ -206,8 +206,18 @@ export class StackManager extends React.PureComponent<StackManagerProps, StackMa
   registerIonPage(page: HTMLElement, routeInfo: RouteInfo) {
```
sean-perkins, line 206: Is there any way to move the checks here up into the calling components, so that `registerIonPage` is not invoked multiple times?

The React 18 change was designed to help developers migrate problematic life cycle method implementations that produce side effects, as they prepare for larger changes in React 19 to how components are attached/re-attached in the DOM (for optimization benefits). I believe the intention is that the dev-mode behavior will become the default production behavior in one of the upcoming majors React 19/20. 

My only concern here with fixing in this API method, is that technically the problematic pattern still exists in the components:
- https://github.com/ionic-team/ionic-framework/blob/main/packages/react/src/routing/PageManager.tsx#L33
- https://github.com/ionic-team/ionic-framework/blob/main/packages/react/src/routing/OutletPageManager.tsx#L29

while the behavior has obviously been patched in this specific method. Future updates to `PageManager`/`OutletPageManager`, would not have the immediate awareness that additional method calls in `componentDidMount` will lead to problematic behavior. 

If there isn't a tangible way to move this up into the component life cycle calls, then I would be in favor of this change as-is.

liamdebeasi: How would you recommend I change the code? I tried a couple alternatives already:
1. Moving the logic to `constructor` will not work as the `this.ionPageElementRef.current` may not be available at that point. 
2. I could add a "init" flag on each component so that subsequent `componentDidMount` calls do not re-fire `registerIonPage`, but that seems a bit hacky.

sean-perkins: Great question... not entirely sure myself. I've seen the `init` flag being used in a few places before, but agree it is a bit hacky. 

I agree with your point that `registerIonPage` should be idempotent & works great for `PageManager`. The only downside with the current implementation is within `OutletPageManager`. Since `componentOnReady` will be invoked multiple times; it will end up calling `requestAnimationFrame` and tying up the repaint. 

I could see the justification in that component alone, to have a flag set for when the component is "ready" and checking if that flag has already been set before invoking that call again. This would prevent unnecessary `raf` calls.

Thoughts?

liamdebeasi: That should be fine. I made the changes in https://github.com/ionic-team/ionic-framework/pull/25798/commits/0458734308c8722fc9e9e48f3089b8eda1ea5bb6.

liamdebeasi: Ok tested with React 17 and React 18 and works fine 👍

sean-perkins (approved):
Changes look good to me 👍

Issue #23108, opened 2021-03-26
bug: ion-modal has no aria-label attribute


Ksan8:
# Bug Report

**Ionic version:**
[ ] **4.x**
[x] **5.x**

**Current behavior:**
In an Ionic app that uses `ion-modal`, the modal wrapper (with CSS class of the same name) has an attribute of `role="dialog"`, but no `aria-label` or `title`. This causes the app to fail accessibility checks, such as [axe browser extension](https://www.deque.com/axe/browser-extensions/).

The issue axe finds is:

- ARIA dialog and alertdialog nodes must have an accessible name

**Expected behavior:**
The modal wrapper should have either an `aria-label` attribute or a `title`, to go with the role attribute.

**Steps to reproduce:**

1. Clone [ionicAngular](https://github.com/Ksan8/ionicAngular) (a minimal Ionic Angular app)
2. Install the [axe browser extension](https://www.deque.com/axe/browser-extensions/) if not already installed
3. Run `ionic serve` from the app root to open the app in the browser
4. Click on "Show Modal"
4. Open DevTools and click on the axe tab
5. Choose "Scan all of my page"
6. Notice the issue listed above appears in the axe results

**Related code:**

Sample Ionic Angular app: [https://github.com/Ksan8/ionicAngular](https://github.com/Ksan8/ionicAngular)

**Other information:**

This is considered an accessibility issue with serious impact. More info on this specific rule can be found on [Deque University's documentation for axe](https://dequeuniversity.com/rules/axe/4.1/aria-dialog-name?application=AxeChrome).

**Ionic info:**

```
Ionic:

   Ionic CLI                     : 6.13.1 (/Users/keziawineberg/.nvm/versions/node/v12.20.0/lib/node_modules/@ionic/cli)
   Ionic Framework               : @ionic/angular 5.5.2
   @angular-devkit/build-angular : 0.1000.8
   @angular-devkit/schematics    : 10.0.8
   @angular/cli                  : 10.0.8
   @ionic/angular-toolkit        : 2.3.3

Capacitor:

   Capacitor CLI   : 2.4.6
   @capacitor/core : 2.4.6

Utility:

   cordova-res : not installed
   native-run  : not installed

System:

   NodeJS : v12.20.0 (/Users/keziawineberg/.nvm/versions/node/v12.20.0/bin/node)
   npm    : 7.5.4
   OS     : macOS Big Sur
```

liamdebeasi:
[...]

PR #25801, opened 2022-08-22
fix(modal): set default aria-label

liamdebeasi:
## Pull request checklist

## Pull request type

 

 

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

Issue URL: resolves https://github.com/ionic-team/ionic-framework/issues/23108

Dialog elements require an `aria-label` which `ion-modal` did not have. Additionally, the `role="dialog"` was set on an inner element while `aria-modal="true"` was set on the host. This caused an issue with voiceover for macOS where the modal was announced as a "group" instead of a "dialog"

## What is the new behavior?

- Moves `role="dialog"` from `.modal-wrapper` to the Host.
- Adds a default `aria-label`.

There's probably an argument to be made that all of these attributes should be encapsulated inside of the shadow DOM on `.modal-wrapper` (same goes for the other overlays), but I think that should go through a design process first. This PR focuses on resolving the basic issues with modal.

- [ ] Yes
- [x] No

### DIFF
--- a/core/src/components/modal/modal.tsx
+++ b/core/src/components/modal/modal.tsx
@@ -18,7 +18,8 @@ import type {
 } from '../../interface';
 import { findIonContent, printIonContentErrorMsg } from '../../utils/content';
 import { CoreDelegate, attachComponent, detachComponent } from '../../utils/framework-delegate';
-import { raf } from '../../utils/helpers';
+import { raf, inheritAttributes } from '../../utils/helpers';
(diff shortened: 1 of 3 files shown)


liamdebeasi:
Adding a default label may cause more harm than good as it simply masks any errors from tools like Axe. Setting `aria-label="modal"` provides nothing of value. Instead, we should encourage developers to set useful labels based on the content they are showing in `ion-modal`.

PR #25805, opened 2022-08-23
fix(alert): use aria-labelledby and aria-describedby instead of aria-label

amandaejohnston:
## Pull request checklist

## Pull request type

 

 

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes
- [x] Other (please describe): Accessibility fixes

In https://github.com/ionic-team/ionic-framework/pull/25800, we added a default `aria-label` to `ion-alert` in order to resolve an Axe violation. However, there are two problems with this resolution:

1. The default `aria-label` of `'Alert'` isn't any better than having no label at all, as it doesn't give any useful information. If anything, it can lead developers into a false sense of security due to the lack of Axe errors.
2. The [ARIA spec](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/Alertdialog_Role#description) mentions that the `alertdialog` role must also have an `aria-describedby` attribute.

Issue URL: N/A

## What is the new behavior?

- `aria-label` replaced with `aria-labelledby` and `aria-describedby`. The former points to the `header` (or the `subHeader` if no header is defined), and the latter points to the `message`.
- If the required properties aren't set for either attribute (for example, no `message`), the attribute is left unset.

- [ ] Yes
- [x] No

## Other information

A separate PR will be put up to document this change and show best practices with accessibility.


core/src/components/alert/alert.tsx
```diff
@@ -578,20 +578,26 @@ export class Alert implements ComponentInterface, OverlayInterface {
...
     const mode = getIonMode(this);
     const hdrId = `alert-${overlayIndex}-hdr`;
     const subHdrId = `alert-${overlayIndex}-sub-hdr`;
     const msgId = `alert-${overlayIndex}-msg`;
     const role = this.inputs.length > 0 || this.buttons.length > 0 ? 'alertdialog' : 'alert';
-    const defaultAriaLabel = header || subHeader || 'Alert';
+
+    /**
+     * If the header is defined, use that. Otherwise, fall back to the subHeader.
+     * If neither is defined, don't set aria-labelled by.
```
liamdebeasi, line 590:
```suggestion
     * If neither is defined, don't set aria-labelledby.
```

liamdebeasi (approved):

sean-perkins (approved):

Issue #25668, opened 2022-07-21
feat: datetime multiple select needs a label on how many dates are selected


liamdebeasi:
### Prerequisites

- [X] I have read the [Contributing Guidelines](https://github.com/ionic-team/ionic-framework/blob/main/.github/CONTRIBUTING.md#creating-an-issue).
- [X] I agree to follow the [Code of Conduct](https://ionicframework.com/code-of-conduct).
- [X] I have searched for [existing issues](https://github.com/ionic-team/ionic-framework/issues) that already report this problem, without success.

### Ionic Framework Version

- [ ] v4.x
- [ ] v5.x
- [X] v6.x
- [ ] Nightly

### Current Behavior

In Ionic 6.2 we introduced a feature for `ion-datetime` that lets users select multiple dates (with `presentation="date"`). As part of these changes, we found that the header in MD mode was confusing as it only showed one date:
![image](https://user-images.githubusercontent.com/2721089/180315784-57389923-1854-4751-ab0c-2896ae022a64.png).

We considered other options such as:

1. Showing all selected dates.
2. Showing the current selected dates and then a count for the remaining n-1 (I.e. Thu, Jul 21 and 4 other dates)
3. Showing the number of dates selected (I.e. 5 dates)

Additional considerations such as how localization will work for this feature made us decide to not move forward with changes to the header text until we could do more research and collect feedback from the community. As a result, the selected date will not appear on the MD datetime when `multiple="true"`. This temporary restriction also applies to the new `ion-datetime-button`.

We plan to prioritize addressing this, and we would also appreciate feedback from the community regarding this behavior.

amandaejohnston:
This has been resolved via https://github.com/ionic-team/ionic-framework/pull/25817 and the update will be available in the upcoming release of Ionic 6.3 👍 See the PR description for details on what we decided.

PR #25817, opened 2022-08-24
feat(datetime): add header text to multiple selection; improve header consistency between modes

amandaejohnston:
## Pull request checklist

## Pull request type

 

 

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes

- When `multiple="true"` and the header is enabled (e.g. when `showDefaultTitle="true"`), the part of the datetime's header that shows what date you have selected (hereafter referred to as just "the header") is not rendered.
- The header doesn't render in `ios` mode at all, whether `multiple` is enabled or not.

Issue URL: resolves https://github.com/ionic-team/ionic-framework/issues/25668

| `md` mode | `ios` mode |
| ------- | ------- |
| md mode | ios mode |

## What is the new behavior?

- The header now shows if `multiple="true"`. If 0 or more than 1 dates are selected, the header shows the number, rather than the exact selected dates. (The behavior when exactly 1 date is selected is unchanged.)
- Added a new prop `titleSelectedDatesFormatter` which can be used to customize the display of the header for localization purposes.
  - Note that this function is not called when exactly 1 date is selected, since that part is already localized automatically. Otherwise, devs would need to replicate the existing localization in their callback to handle this case.
- The selected date(s) will also now show in the header in `ios` mode. The text is identical between the two modes in all cases.
  - This also applies to single-selection datetimes.

| `md` mode | `ios` mode |
| ------- | ------- |
| md mode | ios mode |

- [ ] Yes
- [x] No


core/src/components/datetime/datetime.ios.scss
```diff
@@ -20,12 +20,16 @@
...
 }
 
 :host .datetime-header .datetime-title {
   color: var(--title-color);
+}
 
-  font-size: 14px;
+:host .datetime-header .datetime-selected-date {
+  @include margin(10px, null, null, null);
 }
```
amandaejohnston, line 33: This styling was fairly arbitrary on my part; open to suggestions. I just re-used the same font size and picked a margin that looked good with the existing design.

core/src/components/datetime/datetime.tsx
```diff
@@ -313,6 +314,14 @@ export class Datetime implements ComponentInterface {
    */
   @Prop() firstDayOfWeek = 0;
 
+  /**
+   * A callback used to format the header text that shows how many
+   * dates are selected. Only used if there are 0 or more than 1
+   * selected (i.e. unused for exactly 1). By default, the header
+   * text is set to "datesSelected days".
+   */
+  @Prop() selectedDatesLabelFormatter?: SelectedDatesLabelFormatter;
```
amandaejohnston, line 323: I'm not married to this name if anyone wants to suggest alternatives.

sean-perkins: Perhaps `headerTextFormatter` or `headerSelectedDatesFormatter`?

core/src/components/datetime/datetime-interface.ts
```diff
@@ -23,3 +23,5 @@ export interface DatetimeParts {
 }
 
 export type DatetimePresentation = 'date-time' | 'time-date' | 'date' | 'time' | 'month' | 'year' | 'month-year';
+
+export type SelectedDatesLabelFormatter = (datesSelected: number) => string;
```
sean-perkins, line 27: ~ Should we tweak the naming of `dateSelected` to represent a number? e.g.: `selectedDateCount`, `numberOfDates`, `count`, etc. 

With the current naming it may be mistaken as the array of dates, until they interact with the type.

core/src/components/datetime/datetime.tsx
```diff
@@ -2071,20 +2079,39 @@ export class Datetime implements ComponentInterface {
...
+  private renderCalendarViewHeader() {
     const hasSlottedTitle = this.el.querySelector('[slot="title"]') !== null;
     if (!hasSlottedTitle && !this.showDefaultTitle) {
       return;
     }
 
+    const { activeParts, selectedDatesLabelFormatter } = this;
+    const isArray = Array.isArray(activeParts);
+
+    let headerText;
```
sean-perkins, line 2091:
```suggestion
    let headerText: string;
```

core/src/components/datetime/datetime.tsx
```diff
@@ -2071,20 +2079,39 @@ export class Datetime implements ComponentInterface {
...
+    if (isArray && activeParts.length !== 1) {
+      const defaultText = `${activeParts.length} days`;
```
sean-perkins, line 2093: Can we just the `headerText` here, so we don't need to have additional assignments for exceptions/else checks?

e.g.:
```ts
headerText = `${activeParts.length} days`;
if (selectedDatesLabelFormatter !== undefined) {
  try {
    headerText = selectedDatesLabelFormatter(activeParts.length);
  } catch (e) {
    printIonError('Exception in provided `selectedDatesLabelFormatter`: ', e);
  }
}
```

core/src/components/datetime/datetime.tsx
```diff
@@ -2071,20 +2079,39 @@ export class Datetime implements ComponentInterface {
...
+
     return (
       <div class="datetime-header">
         <div class="datetime-title">
           <slot name="title">Select Date</slot>
         </div>
-        {mode === 'md' && !this.multiple && (
-          <div class="datetime-selected-date">{getMonthAndDay(this.locale, this.activeParts as DatetimeParts)}</div>
-        )}
+        <div class="datetime-selected-date">{headerText}</div>
```
sean-perkins, line 2114: I didn't think of this aspect during the design phase, but should we allow `selectedDatesLabelFormatter` to return undefined to disable the rendering of this space?

Then this area could be conditionally rendered:
```jsx
{headerText && (<div class="datetime-selected-date">{headerText}</div>}
```

sean-perkins (commented):
Since this feature changes the behavior of single selection datetime, should we also have similar checks as the multiple tests, to verify that text is rendering correctly (outside of screenshot tests)?

core/src/components/datetime/datetime.tsx
amandaejohnston: Hmm, maybe `titleSelectedDatesFormatter` to match up with the `showDefaultTitle` prop?

amandaejohnston:
> Since this feature changes the behavior of single selection datetime, should we also have similar checks as the multiple tests, to verify that text is rendering correctly (outside of screenshot tests)?

I would say the screenshot tests cover that, since the header text can only show the one selected date -- unlike when `multiple="true"`, where there are a few different cases to account for. I could be convinced otherwise though.

core/src/components/datetime/datetime.tsx
amandaejohnston: I don't see why not 👀

core/src/components/datetime/datetime.tsx
sean-perkins: That works for me 👍 I like aligning context with `showDefaultTitle`.

liamdebeasi (commented):
Is integrating `datetime-button` with these changes part of this work, or would you recommend we do separate design work for that?

core/src/components/datetime/datetime.tsx
```diff
@@ -2071,20 +2082,36 @@ export class Datetime implements ComponentInterface {
...
     ];
   }
-  private renderCalendarViewHeader(mode: Mode) {
+  private renderCalendarViewHeader() {
     const hasSlottedTitle = this.el.querySelector('[slot="title"]') !== null;
     if (!hasSlottedTitle && !this.showDefaultTitle) {
       return;
     }
 
+    const { activeParts, titleSelectedDatesFormatter: titleSelectedDatesFormatter } = this;
```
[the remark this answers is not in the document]
amandaejohnston: This is outdated; I fixed that in https://github.com/ionic-team/ionic-framework/pull/25817/commits/7d2c6a8e76ebd5d0179564dcd7d4e7d998e8a042. No idea how it got there 😆

amandaejohnston:
I figure updating `datetime-button` is a separate task.

liamdebeasi (approved):
Great job! The property name is a bit wordy, but I also can't think of a shorter name that is still descriptive.

PR #25827, opened 2022-08-25
fix(range): indication when range knob is focused

sean-perkins:
## Pull request checklist

## Pull request type

 

 

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

When the range knob is hovered, focused or active, users do not receive any visual feedback.

Issue URL: Internal

## What is the new behavior?

**Android (md)**

Implements the pattern for hover, focus and active statement as defined in this graphic:

- When the range knob is focused, the knob indicator will display a highlight surrounding the knob (themed to the background color of the knob).
- When the range knob is focused and the pin is enabled, the pin will display on focus.

**iOS**
- When the range knob is focused and the pin is enabled, the pin will display on focus.

- [ ] Yes
- [x] No


core/src/components/range/test/a11y/range.e2e.ts
```diff
@@ -0,0 +1,65 @@
...
+    const tabKey = browserName === 'webkit' ? 'Alt+Tab' : 'Tab';
+
+    const range = page.locator('ion-range');
+    const rangeHandle = page.locator('ion-range .range-knob-handle');
+
+    await page.keyboard.press(tabKey);
+    await page.waitForChanges();
+
+    await rangeHandle.hover();
+    await page.waitForChanges();
```
amandaejohnston, line 27: Do we need to focus the range first to check the hover styling? I would guess we'd want to just do `rangeHandle.hover()` so we only screenshot that state.

core/src/components/range/test/a11y/range.e2e.ts
```diff
@@ -0,0 +1,65 @@
...
+
+    expect(await range.screenshot()).toMatchSnapshot(`range-hover-${page.getSnapshotSettings()}.png`);
+
+    await page.keyboard.press(tabKey);
+    await page.waitForChanges();
+
+    expect(await range.screenshot()).toMatchSnapshot(`range-focus-${page.getSnapshotSettings()}.png`);
+
+    await rangeHandle.click();
```
amandaejohnston, line 36: Since the active styling only shows when you're actively dragging the handle, should this be a mousedown instead?

core/src/components/range/test/a11y/range.e2e.ts
sean-perkins: Good call, this was definitely the result of copy & pasting some of the test contents around 👍 hover should be checked first (without focus), then with focus, then finally when active/mouse down.

amandaejohnston (approved):

liamdebeasi (changes requested):

core/src/components/range/range.md.scss
```diff
@@ -63,6 +64,28 @@
...
+    /* stylelint-disable property-disallowed-list */
+    left: -9px;
+    /* stylelint-enable property-disallowed-list */
+
+    width: 36px;
+    height: 36px;
+
+    transform: scale(0.1);
+
+    transition: 0.3s cubic-bezier(0.25, 0.8, 0.5, 1);
```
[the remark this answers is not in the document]
sean-perkins: The `cubic-bezier` value comes from Vuetify: https://github.com/vuetifyjs/vuetify/blob/aeb2100dc055190f6a3ebdcaee7627f3e13dd0bf/packages/vuetify/src/components/VSlider/VSlider.ts#L134

The live examples shows the timing duration of `0.3s` if you inspect the knob: https://vuetifyjs.com/en/components/sliders/

If there is a better way to calculate these values or a different example to refer to, I'm all for it 👍

core/src/components/range/range.md.scss
```diff
@@ -63,6 +64,28 @@
...
+    position: absolute;
+    top: -9px;
+    /* stylelint-disable property-disallowed-list */
+    left: -9px;
+    /* stylelint-enable property-disallowed-list */
+
+    width: 36px;
+    height: 36px;
+
+    transform: scale(0.1);
```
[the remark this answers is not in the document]
sean-perkins: This also comes from Vuetify, but the intention is that when the indicator begins to scale to full size, it will start at a scale value closer to the existing knob size. This creates a more "smooth" growth of indicator behind the knob.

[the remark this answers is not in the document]
liamdebeasi: When not make the focus indicator the same size as the knob? Then you can have it transform from `scale(1)` to `scale(>1)`? This is what the web implementation for Material Design components do: https://material-components.github.io/material-components-web-catalog/#/component/slider. That way you don't have to guess at an approximate size.

core/src/components/range/range.md.scss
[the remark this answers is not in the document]
liamdebeasi: That's probably fine. I took a look at what other libraries do:

| library | curve | duration |
| ------ | ------ | --------- |
| Angular Material | `cubic-bezier(0.25, 0.8, 0.25 ,1)` | `400ms` |
| Material Design components for the web | `cubic-bezier(0, 0, 0.58, 1)` | `267ms` |
| MUI | `cubic-bezier(0.4, 0, 0.2, 1)` | `150ms` |

sean-perkins:
I need to update the screenshot tests to capture the individual states with just the class assigned vs. Playwright performing focus/hover/etc.

Will re-ping once that is ready.

core/src/components/range/range.md.scss
```diff
@@ -130,11 +157,40 @@
...
 
-.range-knob-pressed .range-pin {
+.range-knob-pressed .range-pin,
+.range-knob-handle.ion-focused .range-pin {
   @include transform(translate3d(0, -24px, 0), scale(1));
 }
 
-:host(:not(.range-has-pin)) .range-knob-pressed .range-knob {
+@media (any-hover: hover) {
+  .range-knob-handle:hover .range-knob:before {
```
sean-perkins, line 166: I can combine these selectors.

core/src/components/range/range.md.scss
```diff
@@ -130,11 +157,38 @@
...
+    transform: scale(2);
+    
+    opacity: 0.04;
```
liamdebeasi, line 169: Thoughts on making the hover opacity the same as the focus opacity? I can't really see `0.04` 
image

The main distinction between hover + focus would be the scale of the outline/glow.

core/src/components/range/range.md.scss
```diff
@@ -63,6 +64,32 @@
...
+
+    transform: scale(1);
+
+    // Value from Material Design components for the web
+    transition: 0.267s cubic-bezier(0, 0, 0.58, 1);
+
+    background: var(--knob-background);
+
+    content: "";
+    opacity: 0.3;
```
liamdebeasi, line 90: Thoughts on starting the opacity at whatever the hover + focus opacity is? Otherwise you get a strange fade effect when hovering over the range (range is enlarged + animation is slowed down for emphasis)

https://user-images.githubusercontent.com/2721089/189433236-13527017-1c5a-4364-b51f-fedc1c8e281c.mov

Alternatively, this issue may less noticeable if the hover opacity is the same as the focus opacity.

liamdebeasi (approved):
Comments below are non-blocking. Great job on this!

core/src/components/range/range.md.scss
sean-perkins: I'll reach out to Ben Monday on this. The opacity values selected were calculated by taking the color samples from the MD spec and applying an opacity that matched the new background color for each indicator state.

Even in the spec screenshot, it isn't shown on a white background, which makes me think the hover opacity almost being not visible could be deliberate.

liamdebeasi:
Thinking about this more, this is a new behavior so we should target `feature-6.3` instead of `main`. Also, is there a way to customize this focused state independent of the `color` prop? Seems like you can do `::part(knob)::before`, right?

core/src/components/range/range.md.vars.scss
```diff
@@ -19,20 +19,14 @@ $range-md-hit-width:                         42px !default;
...
-$range-md-bar-height:                        2px !default;
+$range-md-bar-height:                        4px !default;
 
 /// @prop - Background of the range bar
 $range-md-bar-background-color:              $background-color-step-250 !default;
 
 /// @prop - Background of the active range bar
 $range-md-bar-active-background-color:       current-color(base) !default;
 
-/// @prop - Width of the range knob
```
sean-perkins, line 30: These variables are unused in the implementation. When the `--knob-size` CSS variable was introduced these were not removed.

core/src/components/range/range.md.scss
```diff
@@ -63,6 +64,32 @@
...
   z-index: 2;
+
+  &::before {
+    @include border-radius(50%);
+    @include position(null, null, null, 0);
+
+    @include rtl() {
+      /* stylelint-disable-next-line property-disallowed-list */
+      left: unset;
+    }
```
liamdebeasi, line 75: Why is the RTL mixin needed?

liamdebeasi (approved):
One lingering question, otherwise everything else is good to go. Nice job!

core/src/components/range/range.md.scss
sean-perkins: These are outdated styles from previous changes 👍 removed them & confirmed they do not impact the display of the indicator.

Issue #25833, opened 2022-08-26
bug: passing max only causes ion-datetime is out of bounds


RRGT19:
### Prerequisites

- [X] I have read the [Contributing Guidelines](https://github.com/ionic-team/ionic-framework/blob/main/.github/CONTRIBUTING.md#creating-an-issue).
- [X] I agree to follow the [Code of Conduct](https://ionicframework.com/code-of-conduct).
- [X] I have searched for [existing issues](https://github.com/ionic-team/ionic-framework/issues) that already report this problem, without success.

### Ionic Framework Version

- [ ] v4.x
- [ ] v5.x
- [X] v6.x
- [ ] Nightly

### Current Behavior

I have a field to select the date of birth. The rule is to start showing dates 10 years before today. In the meanwhile, the field should be empty until you select a date.

![example image](https://i.ibb.co/mqHP3fn/Screen-Shot-2022-08-26-at-8-54-27-AM.png)

When setting the `max` to `2012-08-26T12:47:01.977Z` or to `2012`, I receive this warning:

```
The value provided to ion-datetime is out of bounds.

Min: undefined
Max: {"month":8,"day":26,"year":2012,"hour":12,"minute":47}
Value: {"year":2022,"month":8,"day":26,"hour":8,"minute":56,"tzOffset":0}
```

I understand the warning and I see that the **Value** (2022) is greater than the **Max** (2012). It seems that the component is picking up the today Date automatically.

Trying to set the field's value with the same `max` date that I have provided, the warning disappears.

![example image](https://i.ibb.co/SfTC0nD/Screen-Shot-2022-08-26-at-8-58-26-AM.png)

This can be confusing because the reality is that the user did not select a date yet. Also, when the user see that date, he can think that was automatically pulled from its account settings, which is false.

### Expected Behavior

- The component should not take `today` date automatically.
- If no Value exists, should render the dates taken into account min/max settings.

### Steps to Reproduce

```
<ion-modal trigger="open-modal">
  <ng-template>
    <ion-content>
      <ion-datetime
        #datetime
        size="cover"
        [(ngModel)]="value"
        max="2012-08-26T12:47:01.977Z"
        [preferWheel]="true"
      ></ion-datetime>
    </ion-content>
  </ng-template>
</ion-modal>
```

### Ionic Info

```
Ionic:

   Ionic CLI                     : 6.20.1
   Ionic Framework               : @ionic/angular 6.2.4
[...]

PR #25836, opened 2022-08-26
fix(datetime): passing min/max no longer errors on missing date information

liamdebeasi:
## Pull request checklist

## Pull request type

 

 

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

Issue URL: N/A

When fixing https://github.com/ionic-team/ionic-framework/issues/25833, the author reported another bug where they would get `RangeError: Invalid time value at getDayColumnData` when setting `max="2012"`.

The problem is that when we clamp the date to use inside of `ion-datetime`, we simply return the user provided min/max datetime parts: https://github.com/ionic-team/ionic-framework/blob/2cba460ef682b0f570e02be3a486b3db04f77b24/core/src/components/datetime/utils/parse.ts#L120

This means that if developers pass `min="2012"`, then the `minParts` object will have no information for month, day, hour, or minute.

## What is the new behavior?

- Updated the min/max process logic to fill in any missing information.
- The min parts will fill in data for January 1st at 00:00. User specified data will override as available. For example, if you passed `min="2012-10"` you would get a min parts object for October 1, 2012 at 00:00.
- The max parts will fill in data for December 31st at 23:59. User specified data will override as available. For example, if you passed `max="2022-05`, you would get a max parts object for May 31, 2022 at 23:59.

- [ ] Yes
- [x] No


amandaejohnston (approved):

core/src/components/datetime/datetime.tsx
```diff
@@ -1723,6 +1718,7 @@ export class Datetime implements ComponentInterface {
 
     return (
       <ion-picker-column-internal
+        class="hour-column"
```
sean-perkins, line 1721: Is there a reason we introduced class names for: `hour-column`, `minute-column` and `day-period-column`?

liamdebeasi: I added those for some old tests, but those tests are no longer used. I will remove them.

sean-perkins (approved):

Issue #3523, opened 2015-04-14
bug: Can't place <input> in <ion-list>


victorwoo:
**Type**: <span ionic-type>bug</span>

**Platform**: <span ionic-platform>all</span>  

<span ionic-description>I need to use some feature of ion-list directive, for example ion-option-button.
But any approach of putting input in ion-list doesn't looks right.
Please have a look at this codepen:
http://codepen.io/victorwoo/pen/Eaqymr/</span>

<span is-issue-template></span>

mhartington:
Something like this can be done with just a few lines of css.
Note though, its ends up becoming a bit tough to swipe the item and reveal the option buttons, so it may not be great ux. 

http://codepen.io/mhartington/pen/myNWeM

PR #25838, opened 2022-08-26
refactor(datetime): use key to preserve calendar body ref node

### LINKED ISSUE #3253: bug: iOSswipe cause white screen with actionsheet back
**Type**: <span ionic-type>bug</span>

**Platform**: <span ionic-platform>ios</span> <span ionic-platform-version>8</span>  <span ionic-webview>webview</span>

<span ionic-description>After using actionsheet to go to the new state, using the new "iOS swipe to go back" will cause white screen.</span>

<span is-issue-template></span>

alicewriteswrongs:
This fixes an issue with the Datetime component where in certain
circumstances a `ref` on an element rendered by the datetime component
would end up erroneously set to `null`, necessitating a yucky
`document.querySelector` based workaround.

The issue arises because of how Stencil's virtual DOM implementation
handles reconciling the children of old and new nodes, and in particular
how it deals with situations in which the order of children changes but
some of the nodes are conserved.

The way this was affecting the datetime component was that in certain
circumstances a method which returns a `<div>` with a `ref` would change
position in the children of another node. In particular, this can happen
if the `presentation` prop changes, causing the return value of the
`renderDatetime` function to change in turn. The salient changes to
the return value of _that_ function which would change look like
going from:

```ts
[
  this.renderCalendar(mode),
  this.renderTime()
]
```

to

```ts
[
  this.renderTime(),
  this.renderCalendar(mode)
]
```

`renderCalendar` in turn calls `renderCalendarBody` which returns the
`<div>` with the offending `ref` on it.

Anyhow, in reconciling these changed children when re-rendering the
component things can go wrong. Stencil's algorithm in this case does a
"best effort" at identifying nodes which should be kept between
re-renders, but it does not do an exhaustive check _unless nodes have a
key attr_. So, more or less, what was happening was that in cases where
the `<div>` with the `ref` on it was going to be included in the next
rendered output from the component Stencil would nonetheless remove the
'old' node and create a new identical one for the next render. Then
there is also a race condition where the new node would be created
before the old one is removed, causing the `ref` to be briefly set to
the new value (when the new VNode is created) and then reset to `null`
(when the old one was destroyed).

The fix is to set the `key` attribute on the `<div>` which is the
parent of the `<div>` with the `ref`. This allows Stencil to do an
exhaustive check when reconciling old and new children in order to
not needlessly throw away the old VDom node, and gets rid of the
behavior where the `ref` would end up set to `null` erroneously.
It's possible (but I think unlikely) that this change will result
in slightly less GC pressure, but I did no testing of that.

closes ionic-team/stencil#3253
FW-901

## Pull request checklist

## Pull request type

 

 

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

There is a bug in Stencil which this fixes (https://github.com/ionic-team/stencil/issues/3253).

Basically, there was an issue where switching the `presentation` prop on the Datetime component would erroneously set a `ref` stored on the component to `null`. This fixes that by setting a `key` attr on the relevant div so that Stencil's VDom child reconciliation / update algorithm can do a better job of finding nodes that are the same between renders. 

Issue #3523

- [ ] Yes
- [x] No

### DIFF
--- a/core/src/components/datetime/datetime.tsx
+++ b/core/src/components/datetime/datetime.tsx
@@ -629,18 +629,8 @@ export class Datetime implements ComponentInterface {
     return presentation === 'date' || presentation === 'date-time' || presentation === 'time-date';
   }
 
-  /**
-   * Stencil sometimes sets calendarBodyRef to null on rerender, even though
-   * the element is present. Query for it manually as a fallback.
-   *
-   * TODO(FW-901) Remove when issue is resolved: https://github.com/ionic-team/stencil/issues/3253
-   */
-  private getCalendarBodyEl = () => {
-    return this.calendarBodyRef || this.el.shadowRoot?.querySelector('.calendar-body');
-  };
-
   private initializeKeyboardListeners = () => {
-    const calendarBodyRef = this.getCalendarBodyEl();
+    const calendarBodyRef = this.calendarBodyRef;
     if (!calendarBodyRef) {
       return;
     }
@@ -818,7 +808,7 @@ export class Datetime implements ComponentInterface {
   };
 
   private initializeCalendarListener = () => {
-    const calendarBodyRef = this.getCalendarBodyEl();
+    const calendarBodyRef = this.calendarBodyRef;
     if (!calendarBodyRef) {
       return;
     }
@@ -1241,7 +1231,7 @@ export class Datetime implements ComponentInterface {
   };
 
   private nextMonth = () => {
-    const calendarBodyRef = this.getCalendarBodyEl();
+    const calendarBodyRef = this.calendarBodyRef;
     if (!calendarBodyRef) {
       return;
     }
@@ -1261,7 +1251,7 @@ export class Datetime implements ComponentInterface {
   };
 
   private prevMonth = () => {
-    const calendarBodyRef = this.getCalendarBodyEl();
+    const calendarBodyRef = this.calendarBodyRef;
     if (!calendarBodyRef) {
       return;
     }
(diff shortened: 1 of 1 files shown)


alicewriteswrongs (commented):
just pointing one thing out

amandaejohnston (commented):
Amazing 🤯 Testing this fix out, things look great on my end. I had a couple questions:

1. Your explanation mentions that the `key` attribute should be set on the same element as the `ref`, but here it's actually set on the ref's parent (`.datetime-calendar` vs `.calendar-body`). Was that intentional?
2. Is there a general guide floating around anywhere for when to use the `key` attribute, especially as it pertains to diagnosing and fixing this sort of issue? (Or will the VDom guide you're working on address that?)

alicewriteswrongs:
To your questions:

1. I think my wording was a little fuzzy there. The place to set the `key`
attr is on whatever element is going to be changing position from one
render to another within the children of its parent, so I think on
`.datetime-calendar` is the right spot.

2. I don’t think using the key attr is documented anywhere at present. As
far as I can tell it basically works the same way the key attr does in
React in similar situations, allowing the Vdom implementation to detect
children which should be retained across renders in more situations. I
believe that the key attr checks have been in Stencil for a long time but
we’re never documented, possibly they were just copied over when the vdom
implementation was done. I am going to look into both adding documentation
to explain when to use it and try to see if we could print a compile- or
run-time warning in Stencil when the lack of a key attr could lead to this
type of problem.

amandaejohnston:
Perfect, thank you 👍 I'm gonna add the team back on as a reviewer but I think just one more pair of eyes as a sanity check would be good, whoever gets to it first.

liamdebeasi (approved):
Thank you!!

amandaejohnston (approved):

amandaejohnston:
Merged; thanks again!

Issue #20061, opened 2019-12-11
bug: The ionChange event of ion-input is triggered the second time the page is entered


C-racker:
# Bug Report

**Ionic version:**

[x] **4.x**

**Current behavior:**

`<ion-input [(ngModel)]="tests" (ionChange)="test($event)"></ion-input>`

```
ngOnInit(
  this.tests = '123';
)

test($event){
 console.log('$event :', $event);
}
```

This is triggered the second time the page is accessed

**Expected behavior:**

If the user is not actively typing, no change event needs to be triggered

mhartington:
👋  Hi there! Are you able to provide a small demo repository to debug this? I'm not able to replicate this locally.

C-racker:
> 👋 Hi there! Are you able to provide a small demo repository to debug this? I'm not able to replicate this locally.

https://stackblitz.com/edit/ionic-v4-angular-tabs-2ynupd

hscheuerle:
In the example, your button's method goToPage pushes the test1 component onto the router history. When you click the ion-back-button, you navigate back in the stack which pops the router history and destroys the component. You are seeing multiple onInit calls because the component is destroying and initializing correctly.

sithwarrior:
I changed the example, just to answer the question from @hscheuerle . 
The Exact same thing happens, if we dont use the goToPage, but routerLink instead, and even if we omit using the init, to set a value, IonChanges still fires on 2nd load of the page.

@mhartington I actually think this might be stencil related? When debugging locally using Ionic 5, and the ionChange on the Ion Toggle component, I see that the @Watch seems to be firing before ComponentWillLoad, the 2nd time I visit the page.

According to the docs, this should not be happening?
`Watch Decorator
When a user updates a property, Watch will fire what ever method it's attached to and pass that method the new value of the prop along with the old value. Watch is useful for validating props or handling side effects. Watch decorator does not fire when a component initially loads.`
[https://stenciljs.com/docs/reactive-data#watch-decorator](https://stenciljs.com/docs/reactive-data#watch-decorator)

sithwarrior:
This is still an issue, I just wanted to update that i fixed my problem, by using reactive forms in Angular, instead of the ionChange method, as a work-around.

mread1208:
We were able to get around this issue by using `(ngModelChange)=""` Instead of using `(ionChange)=""`

McTribals:
@mread1208, muchas gracias me ayudo bastante tu aporte.

[...]

PR #25858, opened 2022-08-30
feat(input): ionChange will only emit from user committed changes 

### LINKED ISSUE #20106: bug: changing value of component in ionChange results in an infinite loop
# Bug Report

**Ionic version:**
[x] **4.7.11**

**Current behavior:**
When the value of an Ionic input element is changed from OUTSIDE of the element, the element's onIonChange is triggered, thereby duplicating the change event.

**Expected behavior:**
As with every other React input, the onIonChange should only be triggered when the value was change from the element itself.

**Steps to reproduce:**
See console in included stackblitz. 
1. Both the IonInput and regular input are controlled via the same value. 
2. When changing it from the IonInput, only the onIonChange is triggered. 
3. When changing it from the regular input, the input's onChange is called AS WELL AS the onIonChange from the IonInput.

**Related code:**
https://stackblitz.com/edit/ionic-v4-11-7-controlled-inputs

**Other information:**
This same bug is happening with other Ionic inputs as well. I haven't tested all of them, but it is happening with at least IonToggle and IonDateTime.

This is bad because unless each input has its own separate controlled state, doing something like clearing a state object could trigger 10+ onIonChange events, leading to race conditions and other annoyances.

**Ionic info:**
```
Ionic:

   Ionic CLI       : 5.4.13 (/Users/evan/.config/yarn/global/node_modules/ionic)
   Ionic Framework : @ionic/react 4.11.7

Capacitor:

   Capacitor CLI   : 1.4.0
   @capacitor/core : 1.4.0

Utility:

   cordova-res : not installed
   native-run  : not installed

System:

   NodeJS : v13.1.0 (/usr/local/Cellar/node/13.1.0/bin/node)
   npm    : 6.12.1
   OS     : macOS Catalina
```

sean-perkins:
## Pull request checklist

## Pull request type

 

 

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

`ionChange` is emitted from both internal and external value changes of `ion-input`.

**Angular**
- `ion-input` form control value updates when `ionChange` is emitted (`ngModel`, `formControl`, `formControlName`).

Issue URL: Resolves #20106, #20061

## What is the new behavior?

- `ionChange` will only be emitted from user committed changes to `ion-input`:
  -  User types within the control and the control is blurred/loses focus
  -  User clicks the clear action in an input with an existing value
- `ionChange` **will not** be emitted from externally assigning a value to `ion-input`

**Angular**
- `ion-input` form control value updates when `ionInput` is emitted (`ngModel`, `formControl`, `formControlName`).

- [x] Yes
- [ ] No

Developers that need the immediate value of the input as the user types should migrate their implementations to using `ionInput` instead.

Developers that assign a value directly to `ion-input` should not expect `ionChange` to be emitted. In these cases, developers should execute expected side-effect behavior when assigning the value. 

## Other information

This is the first component being resolved of this issue. As others are migrated, we will likely have a "Form controls" section in the `BREAKING` doc, instead of copy/pasting for each affected component. 

Once all of the components of a specific value accessor are migrated, the temporary value accessors will be removed and the legacy class name will be the same API. 

Dev-build: `6.2.4-dev.11661977223.1d42c906`


angular/src/directives/control-value-accessors/numeric-value-accessor.ts
```diff
@@ -18,13 +18,13 @@ export class NumericValueAccessorDirective extends ValueAccessor {
     super(injector, el);
   }
 
-  @HostListener('ionChange', ['$event.target'])
-  _handleIonChange(el: any): void {
+  @HostListener('ionInput', ['$event.target'])
+  handleInputEvent(el: HTMLIonInputElement): void {
     this.handleChangeEvent(el, el.value);
   }
```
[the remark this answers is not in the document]
sean-perkins: Is the name `handleChangeEvent` the confusing piece here? 

That function is responsible for performing the `onChange` callback for Angular's control value accessor as well as syncing the Ionic "status" classes (`.ion-invalid`, etc.). With us no longer listening to `ionChange`, I can see where that name is problematic. I want to avoid renaming it to `handleInputEvent` or similar, since currently only `ion-input` is updated to the revised approach.

Perhaps a rename like this would work?

```ts
@HostListener('ionInput', ['$event.target'])
handleInputEvent(el: HTMLIonInputElement): void {
  this.commitValueChange(el, el.value);
}
```

I don't think we are using the base class that comes from [`@stencil/angular-output-targets`](https://github.com/ionic-team/stencil-ds-output-targets/blob/main/packages/angular-output-target/resources/control-value-accessors/value-accessor.ts) so we have full control to refactor as necessary here.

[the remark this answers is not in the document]
amandaejohnston: That works for me 👍

angular/src/directives/control-value-accessors/numeric-value-accessor.ts
```diff
@@ -18,13 +18,13 @@ export class NumericValueAccessorDirective extends ValueAccessor {
     super(injector, el);
   }
 
-  @HostListener('ionChange', ['$event.target'])
-  _handleIonChange(el: any): void {
-    this.handleChangeEvent(el, el.value);
+  @HostListener('ionInput', ['$event.target'])
+  handleInputEvent(el: HTMLIonInputElement): void {
+    this.commitValueChange(el, el.value);
```
liamdebeasi, line 23: The naming of `commitValueChange` seems odd here. In the core implementation, we consider "committing the value" to be when `change` fires. In this file, we consider committing to be when `input` (and therefore `ionInput`) fires.

I'm assuming this is so that form validation is re-run whenever `ionInput` fires? If so, maybe a better name would be `handleInputEvent`?

angular/src/directives/control-value-accessors/text-value-accessor.ts
```diff
@@ -21,6 +21,27 @@ export class TextValueAccessorDirective extends ValueAccessor {
...
+  selector: 'ion-input:not([type=number])',
+  providers: [
+    {
+      provide: NG_VALUE_ACCESSOR,
+      useExisting: InputValueAccessorDirective,
+      multi: true,
+    },
+  ],
+})
+export class InputValueAccessorDirective extends ValueAccessor {
```
liamdebeasi, line 38: Why do we exclude `type="number"`?

angular/src/directives/control-value-accessors/numeric-value-accessor.ts
sean-perkins: The only reason I would suggest not renaming this function to `handleInputEvent` at this stage, is this base class is shared across all the value accessors. Only `ion-input` is updated to use `ionInput` instead of `ionChange`. All other value accessors currently use `ionChange`. 

Does it make sense to update that API now, knowing they will all eventually use `ionInput` or similar or should that happen once the final component is migrated and all components use the input event?

angular/src/directives/control-value-accessors/text-value-accessor.ts
sean-perkins: `ion-input[type="number"]` has a separate directive [implementation](https://github.com/ionic-team/ionic-framework/blob/4c13e81679ea79f3a99f7268f6fa0e35251baf43/angular/src/directives/control-value-accessors/numeric-value-accessor.ts#L7). The numeric value accessor parses the value to a float/numeric value.

angular/src/directives/control-value-accessors/numeric-value-accessor.ts
liamdebeasi: As long as we update this to be consistent before the feature launches, that should be fine I think

angular/src/directives/control-value-accessors/text-value-accessor.ts
liamdebeasi: Oh that's right. Thanks!

angular/src/directives/control-value-accessors/numeric-value-accessor.ts
```diff
@@ -18,13 +18,13 @@ export class NumericValueAccessorDirective extends ValueAccessor {
...
   }
 
   registerOnChange(fn: (_: number | null) => void): void {
-    super.registerOnChange((value) => {
+    super.registerOnChange((value: string) => {
```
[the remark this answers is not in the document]
sean-perkins: This is a refactor to make the types accurate to the implementation (currently it inherits type `any`). Below we are asserting if the value is an empty string and passing the result to `parseFloat`. 

https://github.com/ionic-team/ionic-framework/blob/4c13e81679ea79f3a99f7268f6fa0e35251baf43/angular/src/directives/control-value-accessors/numeric-value-accessor.ts#L28

`parseFloat` accepts a parameter as a string, so `value` must be a string in this context.

core/src/components/input/input.tsx
```diff
@@ -393,6 +407,12 @@ export class Input implements ComponentInterface {
 
     this.value = '';
 
+    /**
+     * Clearing the text input is considered a user committed change
+     * and should emit a value change event.
+     */
+    this.emitValueChange();
```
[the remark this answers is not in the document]
sean-perkins: Yeah it should, I'll get that updated. I'll also update the clear action to emit `ionInput` when the clear action is performed: https://codepen.io/sean-perkins-ionic/pen/YzLzPLB

sean-perkins:
General note surrounding the value accessors. I compared against using `input` with `ngModel` in an isolated Angular application. In that example, `ngModelChange` is emitted whenever the `input` event was raised (as the user types). This also updates the control's value when used in a form. 

To match this pattern with the change in behavior to `ionChange`, we are migrating to listening to `ionInput` instead to offer the same experience.

angular/src/directives/control-value-accessors/numeric-value-accessor.ts
sean-perkins: Thinking about this more, not all value accessors will bind to `ionInput` (that event is only available on a subset of our components). 

The fact that this wraps Angular's `onChange` method implementation for value accessors, should we call it: `handleChange` or `handleValueChange` instead?

core/src/components/input/input.tsx
```diff
@@ -407,11 +407,10 @@ export class Input implements ComponentInterface {
...
     this.value = '';
 
-    /**
-     * Clearing the text input is considered a user committed change
-     * and should emit a value change event.
-     */
-    this.emitValueChange();
+    this.ionInput.emit({
+      ...ev,
+      data: this.value,
```
liamdebeasi, line 423: I'm comparing this with the native behavior in Chrome, and noticed that it does not pass `data` when pressing the "X" button. It only seems to pass `data` when typing. Should we align the behavior here? One thing to note is that the native event is of type `Event`, not `InputEvent`.

CodePen: https://codepen.io/liamdebeasi/pen/oNdXJrp

core/src/components/input/input.tsx
```diff
@@ -338,11 +355,25 @@ export class Input implements ComponentInterface {
...
+  private onChange = () => {
+    this.emitValueChange();
+  };
+
   private onBlur = (ev: FocusEvent) => {
     this.hasFocus = false;
     this.focusChanged();
     this.emitStyle();
 
+    if (this.value != null && this.previousValue !== this.value) {
```
liamdebeasi, line 367: The logic here is not correct. This will emit `ionChange` even if the user never commits a value.

Steps to repro:

1. Type "123" but do not click outside the input.
2. Click the "clear" button.
3. Blur the input. Observe that "ionChange" is emitted.


This is a bit tricky since there does not seem to be native browser behavior we are able to leverage for this. This also means that the `data` payload will probably be empty when `ionChange` is emitted. Perhaps this is why we decided not to emit the native event and instead emit a custom payload?

core/src/components/input/input.tsx
```diff
@@ -310,6 +314,19 @@ export class Input implements ComponentInterface {
...
+  /**
+   * Emits an `ionChange` event.
+   *
+   * This API should be called for user committed changes.
+   * This API should not be used for external value changes.
+   */
+  private emitValueChange() {
+    const { value } = this;
+    const newValue = value == null ? value : value.toString();
+    this.ionChange.emit({ value: newValue });
```
liamdebeasi, line 326: I know this would be a breaking change, but would it be simpler to pass along the native `Event` payload here instead of adding this logic? Would align the API closer to `ionInput` too. Not sure why this behavior was added in the first place though, so worth checking that out too.

core/src/components/input/input.tsx
```diff
@@ -387,6 +418,11 @@ export class Input implements ComponentInterface {
 
     this.value = '';
 
+    this.ionInput.emit({
+      ...ev,
+      data: this.value,
+    } as InputEvent);
```
liamdebeasi, line 424: The underlying event is not actually an `InputEvent`. We should probably update the event type information instead of typecasting this. It looks like native can emit `InputEvent | Event`.

core/src/components/input/input.tsx
sean-perkins: I like the idea of simplifying this and to align against the `change` event 1:1. I'll do a little discovery as to why this exists. To my knowledge, this payload hasn't been modified in many years, so the likelihood of us needing additional custom properties that requires an object seems low.

core/src/components/input/input.tsx
sean-perkins: Ok - I think I have this resolved now. Instead of tracking the `previousValue` against the last emitted `ionChange` state, we really only care about the change in the input's value from the time it was focused vs. when it is blurred. I've updated the implementation to track the value difference between these states and only emit when this happens as a result of a clear action.

core/src/components/input/input.tsx
```diff
@@ -324,7 +324,7 @@ export class Input implements ComponentInterface {
   private emitValueChange() {
     const { value } = this;
     const newValue = value == null ? value : value.toString();
-    this.ionChange.emit({ value: newValue });
+    this.ionChange.emit(newValue);
```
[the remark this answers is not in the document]
sean-perkins: I can totally rework this change, I misunderstood your comment here: https://github.com/ionic-team/ionic-framework/pull/25858#discussion_r963855711. You were suggesting passing along the event from the input's change event emission, not aligning with the change event.

liamdebeasi (approved):
Great job!

amandaejohnston (approved):

BREAKING.md
```diff
@@ -50,6 +51,12 @@ This section details the desktop browser, JavaScript framework, and mobile platf
 
 <h2 id="version-7x-components">Components</h2>
 
+<h4 id="version-7x-input">Input</h4>
+
+`ionChange` will no longer be emitted when the `value` of `ion-input` is modified externally. `ionChange` will only be emitted from user committed changes, such as typing in the input and the input losing focus or from clicking the clear action within the input.
```
liamdebeasi, line 56:
```suggestion
`ionChange` is no longer emitted when the `value` of `ion-input` is modified externally. `ionChange` is only emitted from user committed changes, such as typing in the input and the input losing focus or from clicking the clear action within the input.
```

Minor nit, but we use present tense in the rest of the breaking changes guide for v7.

Issue #25872, opened 2022-09-01
bug: iOS range knob appearance uses incorrect box shadow value


sean-perkins:
### Prerequisites

- [X] I have read the [Contributing Guidelines](https://github.com/ionic-team/ionic-framework/blob/main/.github/CONTRIBUTING.md#creating-an-issue).
- [X] I agree to follow the [Code of Conduct](https://ionicframework.com/code-of-conduct).
- [X] I have searched for [existing issues](https://github.com/ionic-team/ionic-framework/issues) that already report this problem, without success.

### Ionic Framework Version

- [ ] v4.x
- [ ] v5.x
- [X] v6.x
- [ ] Nightly

### Current Behavior

The knob displayed in `ion-range` incorrectly implements the iOS design interface.

Here is a zoomed-in comparison of the knob appearance for iOS (top) vs. Ionic (bottom). 

![Screen Shot 2022-09-01 at 4 55 40 PM](https://user-images.githubusercontent.com/13732623/188011645-a4e7fe1a-fd76-4ec0-8e8b-6764d2a93e01.png)

### Expected Behavior

The `ion-range` design should match the iOS design interface when `mode` is `ios`. 

The updated value should be:
```scss
$range-ios-knob-box-shadow: 0px 0.5px 4px rgb(0 0 0 / 12%), 0px 6px 13px rgb(0 0 0 / 12%) !default;
```

### Steps to Reproduce

1. Render an `ion-range`
2. View on an iPhone/iOS
3. Observe the knob has an incorrect box shadow compared to the iOS native control

### Ionic Info

N/A

### Additional Information

The iOS range knob appearance hasn't been updated in 5+ years. Likely in an iOS update since then this design changed. This issue is based on the current iOS version (15).

sean-perkins:
This has been resolved and will be available in the next major release of Ionic (v7): https://github.com/ionic-team/ionic-framework/pull/25873

PR #25873, opened 2022-09-01
fix(range): range matches iOS design specification

sean-perkins:
## Pull request checklist

## Pull request type

 

 

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

The range knob design does not match iOS 15. 

Issue URL: #25872

## What is the new behavior?

- `ion-range` is updated to match the iOS design spec:
  - Box shadow has been updated to `0px 0.5px 4px rgba(0, 0, 0, 0.12), 0px 6px 13px rgba(0, 0, 0, 0.12)`
  - The bar height has been increased to `4px` (from `2px`)
  - The bar now includes a border radius of `2px`
  - The knob width and height has been decreased to `26px` (from `4px`)
  - Bar background has been updated to a solid color (this avoid layering semi-transparent elements like ticks over the bar and creating different colors).
- `ion-range` with "tick" marks (steps) has been updated to match the iOS design spec

- [ ] Yes
- [x] No


amandaejohnston:
What do older versions of iOS look like? This raises the question of which version we want to target; Ionic v6 supports back to iOS v13, and v7 will support iOS v14.

sean-perkins:
iOS 14 looks the same as this PR:

Screen Shot 2022-09-02 at 2 06 22 PM

iOS 13 I am unable to find a reference for. 

I believe our framework support is less about the iOS design and more about keeping up to date with Apple's guidelines of supporting the last 2 iOS versions (e.g. When iOS 16 releases, we support iOS 15 and iOS 14).

liamdebeasi:
Here is iOS 13-15 (left to right);

image

Can we double check the box shadow value with Ben? Maybe it's just a trick of the eye, but the box shadow in this PR looks slightly off when compared to native.

amandaejohnston (approved):

sean-perkins:
For sure! Pinged him but expecting a response after the long weekend.

liamdebeasi (commented):
Actually, I think this may be a breaking change since it updates the default value of a public CSS Variable: https://github.com/ionic-team/ionic-framework/blob/main/core/src/components/range/range.ios.scss#L10

Thoughts on merging this into `feature-7.0`?

sean-perkins:
@liamdebeasi I think merging into `feature-7.0` makes sense (once we have validated box shadow values) 👍

sean-perkins:
Observed one thing I need to fix: 
- The bar border radius is incorrect when the bar starts at the 0 position. Need to get the border-radius to apply to `.range-bar`

sean-perkins:
Ok, this should be ready for another review.

Figma design for reference: https://www.figma.com/file/hhknVX5thi5fd6PZ843b5J/Range-Component?node-id=0%3A1

The step/tick design updates are inferred from the iOS settings UI: https://user-images.githubusercontent.com/13732623/188678676-c6b1c5e7-3da1-49c2-a67d-2965c242bb98.jpg

core/src/components/range/range.ios.scss
```diff
@@ -12,7 +12,7 @@
   --bar-height: #{$range-ios-bar-height};
   --bar-background: #{$range-ios-bar-background-color};
   --bar-background-active: #{ion-color(primary, base)};
-  --bar-border-radius: 0;
+  --bar-border-radius: #{$range-ios-bar-border-radius};
```
liamdebeasi, line 15: This is a breaking change, so we need to note it in `BREAKING.md`.

core/src/components/range/range.ios.vars.scss
```diff
@@ -4,76 +4,79 @@
...
 /// @prop - Background of the active range bar
-$range-ios-bar-active-background-color:       current-color(base) !default;
+$range-ios-bar-active-background-color: current-color(base) !default;
+
+/// @prop - Border radius of the range bar
+$range-ios-bar-border-radius: 2px !default;
 
 /// @prop - Width of the range knob
-$range-ios-knob-width:                        28px !default;
+$range-ios-knob-width: 26px !default;
```
liamdebeasi, line 34: The `--knob-size` variable defaults to `28px`. Does that need to be updated too?

core/src/components/range/range.ios.vars.scss
```diff
@@ -4,76 +4,79 @@
...
 
 /// @prop - Height of the range knob
-$range-ios-knob-height:                       $range-ios-knob-width !default;
+$range-ios-knob-height: $range-ios-knob-width !default;
 
 /// @prop - Box shadow of the range knob
-$range-ios-knob-box-shadow:                   0 3px 1px rgba(0, 0, 0, .1), 0 4px 8px rgba(0, 0, 0, .13), 0 0 0 1px rgba(0, 0, 0, .02) !default;
+$range-ios-knob-box-shadow: 0px 0.5px 4px rgba(0, 0, 0, 0.12), 0px 6px 13px rgba(0, 0, 0, 0.12) !default;
```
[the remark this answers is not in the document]
sean-perkins: Thoughts on this way of documenting the updates to both the SASS variables and CSS variables? https://github.com/ionic-team/ionic-framework/blob/3b90d03d963d24d8691149aeb604225ecc32b3f1/BREAKING.md#range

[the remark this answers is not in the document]
liamdebeasi: sure that works

core/src/components/range/range.ios.vars.scss
```diff
@@ -4,76 +4,79 @@
...
-$range-ios-hit-height:                        $range-ios-slider-height !default;
+$range-ios-hit-height: $range-ios-slider-height !default;
 
 /// @prop - Height of the range bar
-$range-ios-bar-height:                        2px !default;
+$range-ios-bar-height: 4px !default;
 
 /// @prop - Background of the range bar
-$range-ios-bar-background-color:              rgba(var(--ion-text-color-rgb, 0, 0, 0), .1) !default;
+$range-ios-bar-background-color: #e4e4e6 !default;
```
liamdebeasi, line 25: I believe we used `--ion-text-color-rgb` so that background could be adjusted globally/adjusted for dark mode. My understanding is this fixes a bug where the background was no solid. Maybe a step color is more appropriate here?

liamdebeasi (approved):
Great job!

PR #25905, opened 2022-09-09
fix(item): show the highlight on iOS when --highlight-height is set

brandyscarney:
## Pull request checklist

## Pull request type

 

 

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

While documenting item and adding a playground for customizing the input highlight, I noticed that `--highlight-height` being set was no longer showing the highlight in `ios` mode. Here is the documentation that explains how this should work: https://ionicframework.com/docs/api/item#input-highlight

It looks like the code for this to work was removed by this commit: https://github.com/ionic-team/ionic-framework/commit/5ca2ce91971408218d7bdc52509ce61a6ebb46aa#diff-c74ddc9584918db0c15a8ba1db280abfe14e7f7632b84e1dfbf74ce86315c6c3L378

Issue URL: I could not find any open issues for this 

## What is the new behavior?

- Adds the removed code back to `item.scss` to make it work with iOS again
- Adds a test in `item/test/highlight` that applies focus to all of the items while showing the different states
- Updates the `item/test/fill` test to add a row of items with focus to see the behavior better
- Adds playwright for both of the above tests

- [ ] Yes
- [x] No

### Item: Fill Before & After

In this test you can see that the iOS highlight does not show until `--highlight-height` is set for the last item.

| | Before | After |
| ----- | ----- | ----- |
| `ios`| localhost_3333_src_components_item_test_fill_ionic_mode=ios (1) | localhost_3333_src_components_item_test_fill_ionic_mode=ios |
| `md` | localhost_3333_src_components_item_test_fill (1) | localhost_3333_src_components_item_test_fill |

### Item: Highlight Before & After

For this test I have set the `--highlight-height` for all custom items, so all iOS items in the last 2 sections should show a highlight.

| | Before | After |
| ----- | ----- | ----- |
| `ios`| localhost_3333_src_components_item_test_highlight_ionic_mode=ios | localhost_3333_src_components_item_test_highlight_ionic_mode=ios |
| `md` | localhost_3333_src_components_item_test_highlight | localhost_3333_src_components_item_test_highlight |

### Item: Form Before & After

I did not change this test at all, it should be the same before & after my change.

| | Before | After |
| ----- | ----- | ----- |
| `ios`| localhost_3333_src_components_item_test_form_ionic_mode=ios | localhost_3333_src_components_item_test_form_ionic_mode=ios |
| `md` | localhost_3333_src_components_item_test_form | localhost_3333_src_components_item_test_form |


core/src/components/item/test/highlight/item.e2e.ts
```diff
@@ -0,0 +1,12 @@
+import { expect } from '@playwright/test';
+import { test } from '@utils/test/playwright';
+
+test.describe('item: highlight', () => {
+  test('should not have visual regressions', async ({ page }) => {
+    await page.goto(`/src/components/item/test/highlight`);
+
+    await page.setIonViewport({ resizeViewportWidth: true });
```
sean-perkins, line 8: Question for this configuration used in both tests - were you observing horizontal clipping of the screenshot viewport? 

Based on the screenshot captures, I think this can likely be removed if not.

```ts
await page.setIonViewport();
```

The `resizeViewportWidth` option was introduced for 1 item test we have, where the rendered contents are larger than the device width (some legacy test that renders on like a landscape tablet viewport width).

brandyscarney: Oh no I just copied from another test, but also I saw that `setIonViewport` was used to take a full height screenshot so I thought that would make it work. I don't think it did though ha.

amandaejohnston: The `resizeViewportWidth` option can also likely be removed from the `fill` test. `setIonViewport` by itself will handle resizing the height 👍

amandaejohnston (approved):
LGTM aside from the `resizeViewportWidth` piece 👍

sean-perkins (approved):

Issue #21194, opened 2020-05-02
feat: Add `form` property to IonButton to submit from outside form


TamoMaez:
# Bug Report

**Ionic version:**
**4.x**

**Current behavior:**
IonButton does not trigger submission of the form. You need this when your submit button is outside the form itself.
```
<form id="some-form">....</form>
<ion-button form="some-form">...</ion-button>
```

**Expected behavior:**
Should trigger the submit event of the form like so: https://www.w3schools.com/tags/att_button_form.asp

**Steps to reproduce:**

Add the code below:
```
<form id="some-form" (ngSubmit)="submit('Triggered')"></form>
<ion-button form="some-form">Trigger submit</ion-button>
```

Tap on the button and nothing will happen

**Related code:**

https://stackblitz.com/edit/ionic-v4-angular-tabs-qx5u9v?file=src%2Fapp%2Ftab1%2Ftab1.page.html

**Workaround:**
Hide native button element and trigger it by reference
```
...<form ... id="transaction-form">...</form>
...
<ion-footer>
    <ion-toolbar button (click)="submitBtn.click()">
        <ion-title>Some title</ion-title>
        <button hidden #submitBtn type="submit" form="transaction-form"></button>
    </ion-toolbar>
</ion-footer>
```

indraraj26:
if you place ion-button with type submit inside form it should trigger automatically. let me know if this doesn't work for you

liamdebeasi:
@indraraj26 Is correct. You need `type="submit"` on the `ion-button` otherwise it will not submit the form.

TamoMaez:
> @indraraj26 Is correct. You need `type="submit"` on the `ion-button` otherwise it will not submit the form.

Hi @indraraj26, adding `type="submit" ` doesn't fix the problem https://stackblitz.com/edit/ionic-v4-angular-tabs-qx5u9v?file=src%2Fapp%2Ftab1%2Ftab1.page.html

When you are outside the form, adding the form attribute of the ion-button (which exists for this purpose) doesn't trigger submission. Please reopen

indraraj26:
Hi @liamdebeasi , 

He is just assigned the id of form to button, Native button is triggering the ngSubmit but not ion-button which is wrapper around button.

```
<form id="some-form" (ngSubmit)="submit()"></form>
    <ion-button type="submit" form="some-form">Trigger form submit (doesn't work)</ion-button>
    <hr />
[...]

PR #25913, opened 2022-09-09
feat(button): submit from outside of form

postnerd:
## Pull request checklist

## Pull request type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes

Buttons with type "submit" will only work from inside forms.

Issue URL: https://github.com/ionic-team/ionic-framework/issues/21194

## What is the new behavior?
Now you can add a "form" prop to every <ion-button> that will be used, if the button is placed outside of a form. This can be helpful, if the button is added to a header/footer toolbar but should be responsible for a specific form.

-
-
-

- [ ] Yes
- [x] No

## Other information

Not sure, if the docs have to be updated. If so, I'm happy to do so. Maybe you could guide me to the right place.


core/src/components/button/button.tsx
```diff
@@ -123,6 +123,11 @@ export class Button implements ComponentInterface, AnchorInterface, ButtonInterf
    */
   @Prop() type: 'submit' | 'reset' | 'button' = 'button';
 
+  /**
+   * To specify the form to be submitted if the button is placed outside of that form.
```
sean-perkins, line 127:
```suggestion
   * The HTML form element or form element id. Used to submit a form when the button is not a child of the form.
```

sean-perkins (changes requested):
Small suggestion on the documentation (will require a new `npm run build` for the generated outputs). 

We will also need Playwright tests for this feature.

For example, in `core/src/components/button/test/form-reference` create a new `button.e2e.ts` file.

You will need to test 3 conditions:
1. Form with an ID
2. Form as a reference
3. No form (querying the closest form). 

For these, setting the HTML mark-up per test is probably easiest. 

I did not run these tests locally, but something like this:
```ts
test.describe('button: form', () => {

  test('should submit the form by id', async({ page }) => {
    await page.setContent(`
      <form id="myForm"></form>
      <ion-button form="myForm" type="submit">Submit</ion-button>
    `);

    const submitEvent = await page.spyOnEvent('submit');
    
    await page.click('ion-button');

    expect(submitEvent).toHaveReceivedEvent();
  });

  test('should submit the form by reference', async({ page }) => {
    await page.setContent(`
      <form></form>
      <ion-button type="submit">Submit</ion-button>
      <script>
        const form = document.querySelector('form');
        const button = document.querySelector('ion-button');

        button.form = form;
      </script>
    `);

    const submitEvent = await page.spyOnEvent('submit');
    
    await page.click('ion-button');

    expect(submitEvent).toHaveReceivedEvent();
  });

  test('should submit the closest form', async({ page }) => {
    await page.setContent(`
    <form>
      <ion-button type="submit">Submit</ion-button>
    </form>
  `);

  const submitEvent = await page.spyOnEvent('submit');
  
  await page.click('ion-button');

  expect(submitEvent).toHaveReceivedEvent();
  });
});
```

Let me know if you have questions.

postnerd:
@sean-perkins Thanks for the support. But I can't get the tests to run.

I've tried all possible commands for testing `npm run test.screenshot` or `npm run test` or `npm run run test.e2e` but only the plain _e2e.ts_ files will be executed not the _button.e2e.ts_ files.

`npm test src/components/button/test/basic/button.e2e.ts ` will run into the following error message `Pattern: src/components/button/test/basic/button.e2e.ts - 0 matches`.

liamdebeasi:
@postnerd We are in the middle of a transition between two E2E testing systems. The `npm run test.e2e` command is for the old system. You should do `npx playwright test src/components/button/test/basic` to run the new Playwright tests.

postnerd (commented):
Added the required tests.

Please let me know if there is anything else to do.

(Messed up this one commit when adding the tests with a wrong commit message.) Since this is such a small PR I could just create another branch and PR with just one commit if you want me to.)

sean-perkins:
The commit message is ok here. We will end up squashing the commits anyways, so only the topmost description (the PR title) will matter. 

I approved CI to run to verify the tests. I don't see any other items requiring changes on my end. 

If CI passes, I'll add the team and we can discuss the release target for this feature.

postnerd:
@sean-perkins Is there anything  to do from my side because of the failed tests?

sean-perkins:
@postnerd merge the latest change from `main` into this branch, run `npm run build` and commit any changes (if any).

postnerd:
Merged again (after @liamdebeasi already merged) because the branch was again one commit behind. 

`npm run build` didn't bring anything up related to this pull request.

Only one change to _angular/src/directives/proxies-list.ts_ that I've ignored before, because it looks like an unrelated style fix (a strange one, adding a line at the beginning of the file ...).

Screenshot 2022-09-19 at 21 49 31
Screenshot 2022-09-19 at 21 49 39

sean-perkins:
Yeah, you can ignore that diff. Running `npm run prettier` from the `/angular` folder will overwrite that change and reformat the file with the current state. 

I re-triggered CI to run.

sean-perkins:
@postnerd update on our end: The team has drafted and approved a design document around this feature. The only item that is outstanding as a result, is we want to introduce a developer warning when the form cannot be found, when using the `form` syntax. 

I can snag this scope & can commit it directly to your fork. 

We will target this feature for 6.3 after I make that change & then the team can do a final review.

sean-perkins (approved):

amandaejohnston (approved):
LGTM once the comment is addressed 👍

core/src/components/button/button.tsx
```diff
@@ -160,21 +166,69 @@ export class Button implements ComponentInterface, AnchorInterface, ButtonInterf
...
+         * the button to submit, but it was not found.
+         */
+        if (typeof form === 'string') {
+          printIonWarning(
+            `Form with selector: "#${form}" could not be found. Verify that the selector is correct and the form is rendered in the DOM.`,
+            el
+          );
+        } else {
+          printIonWarning(
+            `The provided "form" element could not be found. Verify that the form is rendered in the DOM.`,
```
liamdebeasi, line 210: This warning is a bit confusing because it implies that the form element is not in the DOM at all. However, it could be the case that the element is but is not an HTMLFormElement.

core/src/components/button/button.tsx
```diff
@@ -160,21 +166,69 @@ export class Button implements ComponentInterface, AnchorInterface, ButtonInterf
...
+        }
+        return;
+      }
+
+      if (!formEl) {
+        /**
+         * If the form element is not set, the button may be inside
+         * of a form element. Query the closest form element to the button.
+         */
+        formEl = el.closest('form');
```
liamdebeasi, line 222: Why was this logic placed outside `findForm`? I would have expected it to be included in that function as the fallback.

core/src/components/button/test/form-reference/button.e2e.ts
```diff
@@ -0,0 +1,93 @@
...
+
+      page.on('console', (msg) => {
+        logs.push(msg.text());
+      });
+
+      await page.click('ion-button');
+
+      expect(logs.length).toBe(1);
+      expect(logs[0]).toContain(
+        '[Ionic Warning]: Form with selector: "#missingForm" could not be found. Verify that the selector is correct and the form is rendered in the DOM.'
```
liamdebeasi, line 64: This check seems brittle. The moment we change the contents of the warning, this test will fail. Are you trying to check that a warning is logged or are you trying to check what the contents of the warning are?

core/src/components/button/test/form-reference/index.html
```diff
@@ -0,0 +1,51 @@
...
+
+  <body>
+    <ion-app>
+      <ion-header>
+        <ion-toolbar>
+          <ion-title>Outside button - Form Submit</ion-title>
+        </ion-toolbar>
+      </ion-header>
+
+      <ion-content class="ion-padding ion-text-center" id="content" no-bounce>
```
liamdebeasi, line 25:
```suggestion
      <ion-content class="ion-padding ion-text-center" id="content">
```

liamdebeasi (changes requested):

core/src/components/button/button.tsx
sean-perkins: This was done for this check here: https://github.com/ionic-team/ionic-framework/pull/25913/files#diff-8b0e185d2b3a4ceee4ec690a5f8ff2992ef3720594562f34bf2cae17f05254ffR198

If the user provides a `form` element, but the `findForm()` function returned the fallback of querying the closest form, there could be instances in applications where the button is attaching to the wrong form element, that the user is not expecting. 

If the `findForm()` function only returns the result of querying the provided `form` or returning the element directly (if `form` is an `HTMLFormElement`), then we can check if that is not set and warn appropriately.

core/src/components/button/test/form-reference/button.e2e.ts
sean-perkins: Both. Although I could just check for `[Ionic Warning]` if we are concerned with the message contents being updated and needing to update the test over time.

liamdebeasi: I'm fine keeping this as is for now. Long term it may be good to have a warning/error ID like "MISSING_FORM_ELEMENT" that we could check, but that needs design work first and should not block this PR.

core/src/components/button/button.tsx
sean-perkins: Updated the warning message. Let me know if it is more clear or requires additional adjustment.

liamdebeasi (approved):

sean-perkins:
Great work here @postnerd 👍 thanks for the contribution!

PR #25922, opened 2022-09-12
feat(accordion): ionChange will only emit from user committed changes

liamdebeasi:
## Pull request checklist

## Pull request type

 

 

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes

Issue URL: N/A Internal ticket

## What is the new behavior?

- The one exception is when an app passes an array of values to a single-select accordion. When that happens, we set the `value` to be the first item in the array and re-emit `ionChange`.

- [x] Yes
- [ ] No

`ionChange` will no longer be emitted when the `value` of `ion-accordion-group` is modified externally. `ionChange` will only be emitted from user committed changes, such as clicking or tapping the accordion header.

### DIFF
--- a/angular/src/directives/proxies.ts
+++ b/angular/src/directives/proxies.ts
@@ -31,7 +31,10 @@ export class IonAccordion {
 import type { AccordionGroupChangeEventDetail as IAccordionGroupAccordionGroupChangeEventDetail } from '@ionic/core';
 export declare interface IonAccordionGroup extends Components.IonAccordionGroup {
   /**
-   * Emitted when the value property has changed. 
+   * Emitted when the value property has changed
+as a result of a user action such as a click.
+This event will not emit when programmatically setting
+the value property. 
    */
   ionChange: EventEmitter<CustomEvent<IAccordionGroupAccordionGroupChangeEventDetail>>;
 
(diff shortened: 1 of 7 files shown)


angular/src/directives/proxies.ts
```diff
@@ -31,7 +31,10 @@ export class IonAccordion {
 import type { AccordionGroupChangeEventDetail as IAccordionGroupAccordionGroupChangeEventDetail } from '@ionic/core';
 export declare interface IonAccordionGroup extends Components.IonAccordionGroup {
   /**
-   * Emitted when the value property has changed. 
+   * Emitted when the value property has changed
```
liamdebeasi, line 34: This was auto-generated. The formatting looks weird, but should be fine?

sean-perkins: Yeah this should be fine. Intellisense still formats the text correctly. Just a quirk of the output target generator (we would have to split on line breaks and generate per line - which is possible, but the raw output works as is for now).

sean-perkins (approved):
Good work here 👍

amandaejohnston:
Holding on review until tests are passing; seeing a bunch of random failures 🤔 Not sure what's going on there.

liamdebeasi:
Good to review now. There were some issues with installing dependencies for Playwright. Possibly related to https://www.githubstatus.com/incidents/8blgv351d71j.

amandaejohnston (approved):

PR #25934, opened 2022-09-13
feat(segment): ionChange will only emit from user committed changes

sean-perkins:
## Pull request checklist

## Pull request type

 

 

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes

`ionChange` is emitted on `ion-segment` when the value is programmatically set. 

Issue URL: Internal

## What is the new behavior?

- `ionChange` is only emitted for user committed actions (selecting an `ion-segment-button`, dragging to a new active segment button, using keyboard navigation to activate an `ion-segment-button`). 

- [x] Yes
- [ ] No

## Other information

Dev-build: `6.2.8-dev.11663602676.1e458a39`

### DIFF
--- a/BREAKING.md
+++ b/BREAKING.md
@@ -64,9 +64,11 @@ This section details the desktop browser, JavaScript framework, and mobile platf
 
 <h4 id="version-7x-input">Input</h4>
 
-`ionChange` is no longer emitted when the `value` of `ion-input` is modified externally. `ionChange` is only emitted from user committed changes, such as typing in the input and the input losing focus or from clicking the clear action within the input.
+- `ionChange` is no longer emitted when the `value` of `ion-input` is modified externally. `ionChange` is only emitted from user committed changes, such as typing in the input and the input losing focus or from clicking the clear action within the input.
 
-If your application requires immediate feedback based on the user typing actively in the input, consider migrating your event listeners to using `ionInput` instead.
+  - If your application requires immediate feedback based on the user typing actively in the input, consider migrating your event listeners to using `ionInput` instead.
+
+- The `debounce` property has been updated to control the timing in milliseconds to delay the event emission of the `ionInput` event after each keystroke. Previously it would delay the event emission of `ionChange`.
 
 <h4 id="version-7x-overlays">Overlays</h4>
 
--- a/core/src/components.d.ts
+++ b/core/src/components.d.ts
@@ -1068,7 +1068,7 @@ export namespace Components {
          */
         "color"?: Color;
         /**
-          * Set the amount of time, in milliseconds, to wait to trigger the `ionChange` event after each keystroke. This also impacts form bindings such as `ngModel` or `v-model`.
+          * Set the amount of time, in milliseconds, to wait to trigger the `ionInput` event after each keystroke.
          */
         "debounce": number;
         /**
@@ -4857,7 +4857,7 @@ declare namespace LocalJSX {
          */
         "color"?: Color;
         /**
-          * Set the amount of time, in milliseconds, to wait to trigger the `ionChange` event after each keystroke. This also impacts form bindings such as `ngModel` or `v-model`.
+          * Set the amount of time, in milliseconds, to wait to trigger the `ionInput` event after each keystroke.
          */
         "debounce"?: number;
         /**
--- a/core/src/components/input/input.tsx
+++ b/core/src/components/input/input.tsx
@@ -92,13 +92,13 @@ export class Input implements ComponentInterface {
   @Prop() clearOnEdit?: boolean;
 
   /**
-   * Set the amount of time, in milliseconds, to wait to trigger the `ionChange` event after each keystroke. This also impacts form bindings such as `ngModel` or `v-model`.
+   * Set the amount of time, in milliseconds, to wait to trigger the `ionInput` event after each keystroke.
    */
   @Prop() debounce = 0;
 
   @Watch('debounce')
   protected debounceChanged() {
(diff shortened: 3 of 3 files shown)


core/src/components/segment/segment.tsx
```diff
@@ -181,11 +177,21 @@ export class Segment implements ComponentInterface {
 
     const value = this.valueAfterGesture;
     if (value !== undefined) {
-      this.ionChange.emit({ value });
```
sean-perkins, line 184: If you compare the changes side-by-side, you might think `ionChange` is no longer emitted here. While I did remove this event emission, the earlier function calls execute `setNextIndex` → `checkButton` → `ionChange.emit`.

It was removed at this point as it was redundant and causing an extra event emission (once `ionChange.emit` was introduced to `checkButton`).

BREAKING.md
```diff
@@ -84,6 +85,11 @@ iOS:
 |`$range-ios-knob-box-shadow`|`0 3px 1px rgba(0, 0, 0, .1), 0 4px 8px rgba(0, 0, 0, .13), 0 0 0 1px rgba(0, 0, 0, .02)`|`0px 0.5px 4px rgba(0, 0, 0, 0.12), 0px 6px 13px rgba(0, 0, 0, 0.12)`|
 |`$range-ios-knob-width`|`28px`|`26px`|
 
+<h4 id="version-7x-slides">Segment</h4>
```
farisfaisalthena, line 88: the `id` should be `version-7x-segment` instead of `version-7x-slides` right?

sean-perkins: Correct - great catch!

core/src/components/segment/segment.tsx
amandaejohnston: I'm not seeing `ionChange` fire when dragging the segment (on Windows Chrome), only when clicking.

core/src/components/segment/segment.tsx
```diff
@@ -76,15 +75,13 @@ export class Segment implements ComponentInterface {
...
 
   @Watch('value')
   protected valueChanged(value: string | undefined, oldValue: string | undefined | null) {
     this.ionSelect.emit({ value });
-    if (oldValue !== '' || this.didInit) {
-      if (!this.activated) {
-        this.ionChange.emit({ value });
-      } else {
+    if (oldValue !== '') {
+      if (this.activated) {
```
amandaejohnston, line 84:
```suggestion
    if (oldValue !== '' && this.activated) {
```

amandaejohnston (changes requested):

sean-perkins:
I need to find a better way of testing the pointer-up behavior. I'll re-ping for reviews once I figure that out.

core/src/components/segment/test/segment-events.e2e.ts
```diff
@@ -0,0 +1,159 @@
...
+
+        await firstButton.hover();
+        await page.mouse.up();
+
+        expect(ionChangeSpy).toHaveReceivedEventTimes(0);
+      });
+    });
+  });
+
+  test.describe('should not emit', () => {
```
amandaejohnston, line 121: Optional: Might be good to make this more descriptive, something like "should not emit on button click." The descriptions here are very similar to the ones for dragging above, so it's a bit confusing at a glance.

amandaejohnston (approved):
Working great now, nice work on this!

core/src/components/segment/segment.tsx
```diff
@@ -76,18 +75,15 @@ export class Segment implements ComponentInterface {
   /**
    * the value of the segment.
    */
-  @Prop({ mutable: true }) value?: string | null;
```
liamdebeasi, line 79: Why did we remove `null`?

liamdebeasi (changes requested):
Tabbing and selecting a button with "Enter" does not emit `ionChange` on a segment with no initial value.

core/src/components/segment/segment.tsx
sean-perkins: After reviewing the internals of the `ion-segment` component, I could not find a reason why the `value` prop accepts `null` - as nothing from segment appears to assign a value of `null`. The change was made as part of this commit: https://github.com/ionic-team/ionic-framework/commit/ba2230510e95b31d8a6a6ab103ca09888f8237b5#diff-7e01db7218b2b21c70a8e558819e7cdb4f6b6220459509b3bd798371a6555277

There were also a few inconsistencies with the types:
-  `ionChange` emits a type of `string | undefined`, but the value type is `string | undefined | null`. 
-  The types for `newValue` and `previousValue` for the watch callback: https://github.com/ionic-team/ionic-framework/blob/main/core/src/components/segment/segment.tsx#L82

This resulted in type errors when trying to emit the `value` against `ionChange`:

e.g.:
```
const { value } = this;
this.ionChange.emit({ value });
```

> Type 'string | null | undefined' is not assignable to type 'string | undefined'.
  Type 'null' is not assignable to type 'string | undefined'.

Unless type narrowing or type casting (like the previous implementation did with the `@Watch` callback function. 

I'm open to feedback on this though - if we know a reason that `null` should be an accepted value type.

liamdebeasi (approved):
Great job!

core/src/components/segment/segment.tsx
sean-perkins: After doing more research into this, I think the original intent of `null` was to offer developers a way to clear the checked item in the segment. 

I will update the breaking change docs to reflect this, but assigning a value of `''` would be preferred in this use case.

PR #25953, opened 2022-09-16
feat(textarea): ionChange will only emit from user committed changes

sean-perkins:
## Pull request checklist

## Pull request type

 

 

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes

Textarea emits `ionChange` each time the `value` is modified internally (e.g. user typing) or externally (e.g. value being set programmatically).

Issue URL: Internal

## What is the new behavior?

- `ionChange` is only emitted when the inner `textarea` change event is fired.
- `debounce` applies to the `ionInput` event instead of `ionChange`
-

- [x] Yes
- [ ] No


angular/src/directives/control-value-accessors/text-value-accessor.ts
```diff
@@ -35,6 +35,7 @@ export class TextValueAccessorDirective extends ValueAccessor {
     },
   ],
 })
+// TODO rename this value accessor to `TextValueAccessorDirective` when search-bar is updated
```
amandaejohnston, line 38: How come? Wouldn't we just get rid of the `TextValueAccessorDirective` class/directive altogether and be done?

core/src/components/textarea/textarea.tsx
```diff
@@ -159,16 +159,23 @@ export class Textarea implements ComponentInterface {
...
   /**
-   * Emitted when the input value has changed.
+   * The `ionChange` event is fired for `<ion-textarea>` elements when the user
+   * modifies the element's value. Unlike the `ionInput` event, the `ionChange`
+   * event is not necessarily fired for each alteration to an element's value.
+   *
+   * Depending on the way the users interacts with the element, the `ionChange`
+   * event fires at a different moment:
+   * - When the element loses focus after its value has changed : for elements
+   * where the user's interaction is typing.
```
amandaejohnston, line 172: Could you rephrase this somehow? The formatting looks a bit odd and I'm not sure I understand what it's saying 🤔

amandaejohnston (changes requested):

angular/src/directives/control-value-accessors/text-value-accessor.ts
sean-perkins: Removing `TextValueAccessorDirective` would introduce a breaking change to the Angular package, which we don't need to make. The "new" value accessor only exists in the span of migrating the controls to using `ionInput` instead of `ionChange`, then it can assume the existing name.

core/src/components/textarea/textarea.tsx
sean-perkins: Do you have an example of what would be more preferred? This was inspired from the Mozilla docs around the `change` event: https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/change_event

With `ion-textarea` not having extra conditions for when `ionChange` fires however, I could see possibly something like:
> The `ionChange` event is fired when the element loses focus after its value has changed.

Thoughts?

amandaejohnston: Yeah, that sounds good to me. The current description is strange because it mentions the event firing at different moments, but then only one condition is listed. Said condition also seems to include elements besides `textarea`, which isn't needed since this is explicitly the `ion-textarea` docs.

amandaejohnston (approved):

core/src/components/textarea/textarea.tsx
```diff
@@ -252,6 +257,18 @@ export class Textarea implements ComponentInterface {
...
 
+  /**
+   * Emits an `ionChange` event.
+   *
+   * This API should be called for user committed changes.
+   * This API should not be used for external value changes.
+   */
+  private emitValueChange() {
+    const { value } = this;
+    const newValue = value == null ? value : value.toString();
```
liamdebeasi, line 268: Can we either make this an explicit null + undefined check or add a comment that notes the `==` checks both null and undefined?

liamdebeasi (changes requested):
Given `<ion-textarea clear-on-edit="true" value="123"></ion-textarea>`, `ionChange` does not emit when pressing backspace (After blurring). 

`ionInput` does not fire either, but that seems to be the correct behavior since Ionic is clearing the input, not the user. The current textarea does this too. (might be worth mentioning in the docs)

core/src/components/textarea/textarea.tsx
```diff
@@ -298,14 +315,18 @@ export class Textarea implements ComponentInterface {
     return this.value || '';
   }
 
-  private onInput = (ev: Event) => {
-    if (this.nativeInput) {
-      this.value = this.nativeInput.value;
+  private onInput = (ev: InputEvent | Event) => {
```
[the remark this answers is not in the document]
sean-perkins: It can be removed 👍 was misinterpreted from the spec:
> For [<textarea>](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/textarea) and [<input>](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input) elements that accept text input (type=text, type=tel, etc.), the interface is [InputEvent](https://developer.mozilla.org/en-US/docs/Web/API/InputEvent); for others, the interface is [Event](https://developer.mozilla.org/en-US/docs/Web/API/Event).

Good catch.

core/src/components/textarea/textarea.tsx
sean-perkins: This change was made to align with [`ion-input`](https://github.com/ionic-team/ionic-framework/blob/main/core/src/components/input/input.tsx#L257). Do we have a preference on the approach - I can update both input and textarea to match.

Personally, I feel that using the loose equality operator for checking for `null`/`undefined` is a fairly common practice; so I would opt for the comment if we feel it needs calling out.

liamdebeasi: Yeah a comment is probably fine. 👍

core/src/components/textarea/textarea.tsx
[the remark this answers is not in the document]
sean-perkins: Update... actually it is sorta required 🙃 Stencil does not assign the correct types for textarea vs. general elements, so the type is `Event`. 

I am going to update the implementation to:
```ts
private onInput = (ev: Event) => {
  const input = ev.target as HTMLTextAreaElement | null;
  if (input) {
     this.value = input.value || '';
  }
  this.ionInput.emit(ev as InputEvent);
};
```

since we have to type cast anyways.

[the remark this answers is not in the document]
liamdebeasi: That works 👍  Can we add a comment about the Stencil issue?

sean-perkins:
@liamdebeasi couple of call outs from the last changes + feedback to your questions.

> Given `<ion-textarea clear-on-edit="true" value="123"></ion-textarea>`, `ionChange` does not emit when pressing backspace (After blurring).

This has been fixed. This appears to be a bug in the `clearOnEdit` implementation. It does not clear the textarea if the element has an initial value, unless it is focused, blurred, focused and then modified. I have modified this to clear on edit when the control has a value. 

> Should `ionInput` fire because the user pressed backspace?

I do not believe so. We aren't stopping event propagation here. The default behavior with a textarea without a value, when pressing backspace, is to not emit an `input` event: https://codepen.io/sean-perkins-ionic/pen/gOzxGGQ. With the input value being cleared on edit, we would have to manually emit an `ionInput` event.

---

Other misc call outs:
- `clearOnEdit` was mutable, even though its value was never internally modified (removed the mutable flag)
- Property documentation kept referring to `type` being specified, but it has no effect on the `ion-textarea` implementation - removed this incorrect documentation.

core/src/components/textarea/textarea.tsx
```diff
@@ -278,6 +302,8 @@ export class Textarea implements ComponentInterface {
       this.value = '';
     }
 
+    this.ionInput.emit(null);
```
liamdebeasi, line 305: Minor thing, but does `ion-input` emit undefined here? https://github.com/ionic-team/ionic-framework/blob/feature-7.0/core/src/components/input/input.tsx#L446

When I tested in my browser, the `detail` payload showed up as `null`.

liamdebeasi (approved):
Great job!

core/src/components/textarea/textarea.tsx
sean-perkins: So in `ion-input`, it can because of this: https://github.com/ionic-team/ionic-framework/blob/feature-7.0/core/src/components/input/input.tsx#L418

I can open up a separate PR to fix that behavior based on our internal discussions of `InputEvent | null` for now.

I will say this pattern doesn't feel intuitive and our thoughts around: 
```
{ value: string, event?: InputEvent | Event }
``` 
appear advantageous before beta.

liamdebeasi: Thanks for the extra info. I don't think this is a blocker right now, but I'll prioritize the research around the `{ value: string, event?: InputEvent | Event }` in the next sprint.

Issue #17055, opened 2019-01-11
ion-input clearOnEdit not working


RobotBroer:
# Bug Report

**Ionic version:**

[x] **4.x**

**Current behavior:**

ion-item clearOnEdit property has no effect. 

**Expected behavior:**

If property clearOnEdit is set then on selecting the input the current value of the input should be replaced by the new value being inserted. 

**Steps to reproduce:**

1. Set Property clearOnEdit on ion-input 
2. Set a value on the input
3. Set a new value on the input 

**Related code:**

```
insert short code snippets here
```

**Other information:**

**Ionic info:** 

```
Ionic:

   ionic (Ionic CLI)             : 4.6.0
   Ionic Framework               : @ionic/angular 4.0.0-rc.0
   @angular-devkit/build-angular : 0.11.4
   @angular-devkit/schematics    : 7.1.4
   @angular/cli                  : 7.1.4
   @ionic/angular-toolkit        : 1.2.2

Cordova:

   cordova (Cordova CLI) : 8.0.0
   Cordova Platforms     : none
   Cordova Plugins       : no whitelisted plugins (0 plugins total)

System:

   Android SDK Tools : 26.1.1 (/Users/christo/Library/Android/sdk)
   ios-deploy        : 1.9.2
   NodeJS            : v11.6.0 (/Users/christo/.nvm/versions/node/v9.10.1/lib/node_modules/node/bin/node)
   npm               : 6.5.0
   OS                : macOS Mojave
   Xcode             : Xcode 10.1 Build version 10B61

```

paulstelzer:
You are right (can be easily check on the docs simulator -> https://beta.ionicframework.com/docs/api/input/ )

PR #26005, opened 2022-09-23
fix(input): clearOnEdit clears input when user initially types

sean-perkins:
## Pull request checklist

## Pull request type

 

 

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

The `ion-input` field does not initially clear when `clearOnEdit` is enabled, with a default value, when the control is first focused and modified. 

`ionInput` can emit an detail of `undefined`.

Issue URL: Internal

## What is the new behavior?

- `ion-input` with `clearOnEdit` enabled will clear the input when the control receives focus and is modified
  - It will not clear if the user presses "Enter" to submit a form.
  - It will not clear for subsequent input while the control has focus (e.g. does not clear on each keypress).
 - `ionInput` correctly emits the parent event as the detail.

- [ ] Yes
- [x] No

### DIFF
--- a/core/src/components/input/input.tsx
+++ b/core/src/components/input/input.tsx
@@ -28,20 +28,19 @@ import { createColorClasses } from '../../utils/theme';
 export class Input implements ComponentInterface {
   private nativeInput?: HTMLInputElement;
   private inputId = `ion-input-${inputIds++}`;
-  private didBlurAfterEdit = false;
   private inheritedAttributes: Attributes = {};
   private isComposing = false;
   /**
-   * If `true`, the user cleared the input by pressing the clear icon,
-   * within the session of the input being focused.
+   * `true` if the input was cleared as a result of the user typing
+   * with `clearOnEdit` enabled.
    *
-   * This property is reset to `false` when the input is blurred.
+   * Resets when the input loses focus.
    */
-  private inputCleared = false;
+  private didInputClearOnEdit = false;
   /**
    * The value of the input when the input is focused.
    */
-  private focusedValue: string | number | null | undefined;
+  private focusedValue?: string | number | null;
 
   @State() hasFocus = false;
 
@@ -346,6 +345,8 @@ export class Input implements ComponentInterface {
     const { value } = this;
     // Checks for both null and undefined values
     const newValue = value == null ? value : value.toString();
+    // Emitting a value change should update the internal state for tracking the focused value
+    this.focusedValue = newValue;
     this.ionChange.emit({ value: newValue });
   }
 
@@ -383,20 +384,17 @@ export class Input implements ComponentInterface {
 
   private onBlur = (ev: FocusEvent) => {
     this.hasFocus = false;
-    this.focusChanged();
     this.emitStyle();
 
-    if (this.inputCleared) {
-      if (this.focusedValue !== this.value) {
-        /**
-         * Emits the `ionChange` event when the input value
-         * is different than the value when the input was focused.
-         */
-        this.emitValueChange();
-      }
-      this.focusedValue = undefined;
-      this.inputCleared = false;
+    if (this.focusedValue !== this.value) {
+      /**
+       * Emits the `ionChange` event when the input value
+       * is different than the value when the input was focused.
+       */
+      this.emitValueChange();
     }
+
+    this.didInputClearOnEdit = false;
 
     this.ionBlur.emit(ev);
   };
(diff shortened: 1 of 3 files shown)


core/src/components/input/input.tsx
```diff
@@ -432,37 +434,17 @@ export class Input implements ComponentInterface {
     this.isComposing = false;
   };
 
-  private clearTextInput = (ev?: Event) => {
+  private onClearButtonClick = (ev?: Event) => {
```
sean-perkins, line 437: Renamed this function as it is the click handler for the button and should not be re-used directly by other parts of the implementation. 

Instead if inner behavior needs to be shared, there should be another private function that both implementations call.

core/src/components/input/input.tsx
```diff
@@ -432,37 +434,17 @@ export class Input implements ComponentInterface {
...
-    this.inputCleared = true;
-
     this.ionInput.emit(ev);
-
-    /**
-     * This is needed for clearOnEdit
-     * Otherwise the value will not be cleared
-     * if user is inside the input
-     */
-    if (this.nativeInput) {
```
sean-perkins, line 454: This change was made 3 years ago. I am unable to replicate the original bug: https://github.com/ionic-team/ionic-framework/issues/17055 with this removed.

core/src/components/input/input.tsx
```diff
@@ -383,47 +384,48 @@ export class Input implements ComponentInterface {
...
+    }
+    /**
+     * Clear the input if the control has not been previously cleared during focus.
+     * Do not clear if the user hitting enter to submit a form.
+     */
+    if (!this.didInputClearOnEdit && this.hasValue() && ev.key !== 'Enter') {
+      this.value = '';
+      this.didInputClearOnEdit = true;
+    }
+    this.ionInput.emit(ev);
```
sean-perkins, line 426: This should emit `null` to be consistent with Textarea.

core/src/components/input/input.tsx
```diff
@@ -383,47 +385,48 @@ export class Input implements ComponentInterface {
...
+    }
+    /**
+     * Clear the input if the control has not been previously cleared during focus.
+     * Do not clear if the user hitting enter to submit a form.
+     */
+    if (!this.didInputClearOnEdit && this.hasValue() && ev.key !== 'Enter') {
+      this.value = '';
+    }
+    this.didInputClearOnEdit = true;
+    this.ionInput.emit(null);
```
amandaejohnston, line 427: This event doesn't seem to be getting emitted, but I'm not sure why it's needed anyway? In theory, wouldn't this cause duplicate `ionInput` emissions on clear -- one for the clear itself, and one for the key press that triggered the clear?

Recording of me typing while logging `ionInput` events (note only one event firing when input is cleared): https://user-images.githubusercontent.com/90629384/193068697-468f0cd0-c5c7-463b-89e9-c441f6a026ac.mp4

Edit: On closer examination, I _am_ seeing an `ionInput` event with `null` fire when pressing keys like Ctrl and Shift, which I'm not sure is intended.

core/src/components/input/input.tsx
```diff
@@ -383,47 +385,48 @@ export class Input implements ComponentInterface {
...
 
+  private checkClearOnEdit(ev: KeyboardEvent) {
+    if (!this.shouldClearOnEdit()) {
+      return;
+    }
+    /**
+     * Clear the input if the control has not been previously cleared during focus.
+     * Do not clear if the user hitting enter to submit a form.
+     */
+    if (!this.didInputClearOnEdit && this.hasValue() && ev.key !== 'Enter') {
```
amandaejohnston, line 423: This causes the value to be cleared when pressing keys like Ctrl and Shift that don't actually "type" anything. Not sure whether that's desired behavior.

core/src/components/input/test/basic/input.e2e.ts
```diff
@@ -201,3 +201,74 @@ test.describe('input: clear button', () => {
...
+
+    const input = page.locator('ion-input');
+
+    await input.click();
+    await input.type('hello world');
+
+    expect(await input.evaluate((el: HTMLIonInputElement) => el.value)).toBe('hello world');
+  });
+
+  test('should emit ionChange once on blur', async ({ page }, testInfo) => {
```
amandaejohnston, line 243: I feel like this test should be in the `input-events.e2e.ts` file instead, since that's where the `ionInput` tests for `clearOnEdit` are located.

sean-perkins: I opted for locating this with the existing `clearOnEdit` tests, since the bug and behavior was specific to that feature being enabled.

sean-perkins:
Ok, this PR is updated to align with the conversations we had on the textarea `clearOnEdit` PR: https://github.com/ionic-team/ionic-framework/pull/26006

amandaejohnston (approved):

liamdebeasi (approved):

Issue #25964, opened 2022-09-18
bug: ion-datetime date and time pickers do not render correctly on (some) devices with maximum font size set


ppolce:
### Prerequisites

- [X] I have read the [Contributing Guidelines](https://github.com/ionic-team/ionic-framework/blob/main/.github/CONTRIBUTING.md#creating-an-issue).
- [X] I agree to follow the [Code of Conduct](https://ionicframework.com/code-of-conduct).
- [X] I have searched for [existing issues](https://github.com/ionic-team/ionic-framework/issues) that already report this problem, without success.

### Ionic Framework Version

- [ ] v4.x
- [X] v5.x
- [X] v6.x
- [ ] Nightly

### Current Behavior

Date and Time pickers do not render correctly on Samsung Galaxy S21 when the font size is set to largest. 
Labels overlap and make the text unreadable.

NOTE: It only seem to affect **some** android phones (e.g. Galaxy S21).

### Expected Behavior

Rows should not overlap. Text should be visible at any font size setting.

### Steps to Reproduce

1. ionic start choose Angular / Tabs application
2. edit tab1.page.html, comment out (or remove) the `<app-explorer-container>` line
3. add `<ion-datetime></ion-datetime>`
4. add android `npm i @capacitor/android`
5. build a testapp.apk
6. Install the apk on an android device that allows for very large fonts (in my case, Samsung **Galaxy S21**).
7. Open settings and choose the **largest** font size available

Run the app. A page with the current month's calendar appears. 
Click on the date picker, you’ll see that the months overlap:

![image](https://user-images.githubusercontent.com/3269489/190904308-85abcd73-1768-480e-b3ca-8568f46f7c55.png)

and

![image](https://user-images.githubusercontent.com/3269489/190904423-40625e59-de8e-41cf-88d9-3c020e3e5658.png)

### Ionic Info

```

Ionic:

   Ionic CLI                     : 6.20.1 (/home/develop/.local/lib/node_modules/@ionic/cli)
   Ionic Framework               : @ionic/angular 5.9.4
   @angular-devkit/build-angular : 13.3.9
   @angular-devkit/schematics    : 13.3.9
   @angular/cli                  : 13.3.9
   @ionic/angular-toolkit        : 6.1.0

Capacitor:

   Capacitor CLI      : 4.2.0
   @capacitor/android : 4.2.0
   @capacitor/core    : 4.2.0
   @capacitor/ios     : 4.2.0

Utility:

[...]

PR #26017, opened 2022-09-26
fix(datetime): wheel picker font resizing works on android

liamdebeasi:
## Pull request checklist

## Pull request type

 

 

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

Issue URL: resolves https://github.com/ionic-team/ionic-framework/issues/25964

This bug uncovered a few issues:

1. The `.picker-item` CSS in `picker-column-internal` assumed the container height would always be `34px`. When Android's font resize feature is used, the larger text is clipped because it overflows out of the `34px` tall container.
2. The `centerPickerItemInView` logic in `picker-column-internal` used a magic number for centering that also relied on each item being `34px` tall. When this changed, the centering logic broke causing the wrong item to be selected.
3. When fixing problem 2, it was discovered that the `.picker-item-empty` elements can have a different height that the non empty picker elements. The enabled items were `<button>` which defaulted to one font-family where the disabled items were `<div>` which defaulted to a different font-family. This resulted in different elements heights.

## What is the new behavior?

- Removed the `34px` assumption from the `.picker-item` declaration. The container now scales with the font size.
- Refactored the `centerPickerItemInView` logic by determining how much an element needs to be scrolled in order to reach the midpoint of the picker column container.
- Converted `.picker-item-empty` elements to `<button>`. Also added `disabled="true"` and `aria-hidden="true"` so they cannot receive focus or be announced by screen readers.

One design note: This `centerPickerItemInView` approach assumes that all buttons in the picker column are the same height. For now this is fine because long text truncates instead of wraps. If we expose the picker items as shadow parts in the future we may need to revisit this approach.

- [ ] Yes
- [x] No


liamdebeasi:
The screenshots are failing due to font family diffs. This is fixed by https://github.com/ionic-team/ionic-framework/pull/26020.

core/src/components/picker-column-internal/picker-column-internal.tsx
```diff
@@ -228,10 +228,53 @@ export class PickerColumnInternal implements ComponentInterface {
...
+      const targetOffsetFromTop = (numberButton + 3) * targetBox.height;
+
+      /**
+       * Compute the scroll position necessary
+       * for the button to be centered within the view.
+       * Convert the value to an integer to avoid half
+       * pixel scrolls which can be inconsistent across
+       * browsers.
+       */
+      const top = Math.round(targetOffsetFromTop - midpoint);
```
amandaejohnston, line 274: https://user-images.githubusercontent.com/90629384/192820135-e2c14721-88fb-4aed-a157-eea30811f826.mp4 I'm seeing a weird issue where the time picker on the basic datetime opens with the wheels in the wrong positions, and then things jump around when I make the first click.

amandaejohnston (changes requested):

liamdebeasi:
Going to move this back to a draft until I can figure out a better way to handle scrolling.

liamdebeasi:
Closing for now until I can loop back to this.

PR #26026, opened 2022-09-27
feat(searchbar): ionChange will only emit from user committed changes

amandaejohnston:
## Pull request checklist

## Pull request type

 

 

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes

Searchbar emits `ionChange` each time the value is modified internally (e.g. user typing) or externally (e.g. value being set programmatically).

Issue URL: Internal ticket

## What is the new behavior?

- Searchbar only emits `ionChange` when the native input's `change` event fires, or when the input is blurred after clicking the clear or cancel buttons.
- The `debounce` prop applies to the `ionInput` event instead of `ionChange`.
- Searchbar's change event type changed to include `null` as a possibility, to match the type of `ion-searchbar`'s `value` prop.
- `debounce` now defaults to `undefined` instead of 250. If `debounce` is not defined, the event emitter is not wrapped in a `setTimeout`.

- [x] Yes
- [ ] No


BREAKING.md
```diff
@@ -22,6 +22,7 @@ This is a comprehensive list of the breaking changes introduced in the major ver
   - [Segment](#version-7x-segment)
   - [Slides](#version-7x-slides)
   - [Textarea](#version-7x-textarea)
+  - [Searchbar](#version-7x-searchbar)
```
sean-perkins, line 25: Can we move this to be alphabetical in the list?

core/src/components/searchbar/searchbar.tsx
```diff
@@ -240,45 +250,86 @@ export class Searchbar implements ComponentInterface {
...
+             * fire. Manually revert focusedValue so onBlur can
+             * compare against what was in the box before the clear.
+             */
+            this.focusedValue = value;
+          }
         }
-      }
-    }, 16 * 4);
+
+        resolve('cleared');
```
sean-perkins, line 302: Can this resolve `void` instead of a string?

e.g.:

```ts
return new Promise<void>((resolve) => {
  // existing logic
  resolve();
});
```

core/src/components/searchbar/searchbar.tsx
```diff
@@ -240,45 +250,86 @@ export class Searchbar implements ComponentInterface {
...
-        if (shouldFocus && !this.focused) {
-          this.setFocus();
+    return new Promise((resolve) => {
+      // setTimeout() fixes https://github.com/ionic-team/ionic/issues/7527
+      // wait for 4 frames
+      setTimeout(() => {
+        const value = this.getValue();
+        if (value !== '') {
+          this.value = '';
+          this.ionInput.emit();
```
sean-perkins, line 281: Should this emit `null`? `ion-textarea` emits `null` with `clearOnEdit` and `ion-input` will after this change: https://github.com/ionic-team/ionic-framework/pull/26005

sean-perkins (changes requested):
Unsure if this is a bug or not (same behavior as `ion-input`).

If a `ion-searchbar` has a value "123", you focus the searchbar and clear the contents with the clear button. With the searchbar still having focus, you type "123". You blur the searchbar. Should `ionChange` be emitted in this case since the value has not really changed? Currently we are emitting a change event even when the value hasn't really changed since it last emitted.

BREAKING.md
liamdebeasi, line 25: Nit: Can we alphabetize this? (as well as the content below)

core/src/components/searchbar/searchbar.tsx
```diff
@@ -304,13 +359,18 @@ export class Searchbar implements ComponentInterface {
     this.focused = false;
     this.ionBlur.emit();
     this.positionElements();
+
+    if (this.focusedValue !== this.value) {
+      this.emitValueChange();
+    }
```
liamdebeasi, line 365: Why do we not clear `this.focusedValue`? Input does this: https://github.com/ionic-team/ionic-framework/blob/feature-7.0/core/src/components/input/input.tsx#L397

core/src/components/searchbar/searchbar.tsx
```diff
@@ -240,45 +250,86 @@ export class Searchbar implements ComponentInterface {
...
       ev.stopPropagation();
     }
     this.ionCancel.emit();
-    this.onClearInput();
+
+    // get cached values before clearing the input
+    const value = this.getValue();
+    const focused = this.focused;
+
+    await this.onClearInput();
```
liamdebeasi, line 323: I'm noticing that the `ionChange` event fires before the `ionInput` event here.

amandaejohnston:
> Unsure if this is a bug or not (same behavior as `ion-input`).
> 
> If a `ion-searchbar` has a value "123", you focus the searchbar and clear the contents with the clear button. With the searchbar still having focus, you type "123". You blur the searchbar. Should `ionChange` be emitted in this case since the value has not really changed? Currently we are emitting a change event even when the value hasn't really changed since it last emitted.

That particular firing of the event is coming from the native `input`'s `change` listener, so I'm inclined to keep it as-is since it seems like we're matching native behavior 🤔

core/src/components/searchbar/searchbar.tsx
amandaejohnston: I'm not 100% sure it's needed but I'll add it in to account for any edge cases I'm missing 👀

core/src/components/searchbar/searchbar.tsx
amandaejohnston: Looks like not explicitly passing an event to `emit` already causes it to emit as `null`, but I can make it explicit 👍

core/src/components/searchbar/searchbar.tsx
amandaejohnston: Admittedly I can't figure out why 🤔 The `ionInput.emit` call is definitely made before `ionChange.emit`, but the actual event doesn't get picked up by listeners until after. Maybe something with how Stencil is emitting the event(s)? Do you think this could cause problems?

sean-perkins:
> That particular firing of the event is coming from the native input's change listener, so I'm inclined to keep it as-is since it seems like we're matching native behavior 🤔

This is because we are clearing the input's value when the clear button is clicked and then manually focusing it again. 

e.g.:
- SearchBar value is `"hello world"`.
- SearchBar clear button is clicked, focus is moved to the button.
- SearchBar and internal input value is `""`, input is blurred.
- SearchBar manually focuses the input 4-ish frames later.
- Internal input value is still `""` and it will now start accounting for changes in the value since it was last focused 
- User types "hello world"
- User blurs the input
- Internal input compares value from when it was last focused `""` to current value `"hello world"`, value is considered different and change is emitted.
- In the session of the component however, the value was initially "hello world" and ended with "hello world".

I compared against Material's Input however and this behavior appears to be consistent with their clear implementation; so I am good with being aware of it, but not addressing it until someone reports a use case of when this is problematic.

core/src/components/searchbar/searchbar.tsx
liamdebeasi: The problem is that the debounce function is not being removed when `debounce="0"`. So in this case, the `ionInput` event is being emitted inside of `setTimeout(() => { ... }, 0);`. This will cause it to be emitted after the `ionChange` event, which gets emitted synchronously.

I'd argue that this is a bug in how `debounce` was originally implemented. Due to how `ionChange` used to function, the team probably never considered this case.

liamdebeasi: Discussed offline: We are going to have `debounce` default to `undefined` and remove the debounce effect in that case. A debounce of `0` will remain as a `setTimeout`.

amandaejohnston:
Debounce behavior has been updated 👍

core/src/components/searchbar/searchbar.tsx
```diff
@@ -76,11 +77,12 @@ export class Searchbar implements ComponentInterface {
...
    * Set the amount of time, in milliseconds, to wait to trigger the `ionInput` event after each keystroke.
    */
-  @Prop() debounce = 250;
+  @Prop() debounce?: number;
 
   @Watch('debounce')
   protected debounceChanged() {
-    this.ionInput = debounceEvent(this.ionInput, this.debounce);
+    const { ionInput, debounce, unmoddedIonInput } = this;
+    this.ionInput = debounce === undefined ? unmoddedIonInput : debounceEvent(ionInput, debounce);
```
sean-perkins, line 85: Is there a challenge we ran into that led to a private variable to track the event vs.:
```ts
const { debounce } = this;
if (debounce !== undefined) {
   this.ionInput = debounceEvent(this.ionInput, debounce);
}
```

amandaejohnston: Doing it that way would cause `ionInput` to stay debounced if you dynamically changed `debounce` from some number to `undefined`.

sean-perkins: Ahh that makes sense, can we add a note about that somewhere?

Also nit: We use `original` to prefix variables in [a few places](https://github.com/ionic-team/ionic-framework/search?q=original), should we follow that naming strategy here? e.g.: `originalIonInput`

amandaejohnston: Done 👍

sean-perkins (approved):

liamdebeasi (approved):

Issue #25885, opened 2022-09-06
bug: item and card buttons should inherit aria-label attribute


DeyBo:
### Prerequisites

- [X] I have read the [Contributing Guidelines](https://github.com/ionic-team/ionic-framework/blob/main/.github/CONTRIBUTING.md#creating-an-issue).
- [X] I agree to follow the [Code of Conduct](https://ionicframework.com/code-of-conduct).
- [X] I have searched for [existing issues](https://github.com/ionic-team/ionic-framework/issues) that already report this problem, without success.

### Ionic Framework Version

- [ ] v4.x
- [ ] v5.x
- [X] v6.x
- [ ] Nightly

### Current Behavior

Hi !

I saw that there are already some components that reflect _aria-label_ property on their native element (i.e. ion-button).
In my case, I use _ion-card_ and _ion-item_. Sometimes, I set _button_ to true for these, but my custom _aria-label_ is not extended to native button.

### Expected Behavior

I wanted to know if it is possible to use _aria-label_ with another way ?
Or maybe make it possible to extend _aria-label_ to native button when _ion-item_ and _ion-card_ has _button_ to true ?

### Steps to Reproduce

To see it, you can use _ion-item_ or _ion-card_ with _button_ property to true. Set a custom _aria-label_ to your _ion-item_ or _ion-card_ and check accessibility name.

### Ionic Info

Ionic:

   Ionic CLI                     : 5.4.16 (/usr/local/lib/node_modules/ionic)
   Ionic Framework               : @ionic/angular 6.0.9
   @angular-devkit/build-angular : 13.3.9
   @angular-devkit/schematics    : 12.2.15
   @angular/cli                  : 13.1.4
   @ionic/angular-toolkit        : 5.0.3

Capacitor:

   Capacitor CLI   : 2.4.7
   @capacitor/core : 2.4.7

Utility:

   cordova-res : not installed
   native-run  : not installed

System:

   NodeJS : v16.15.1 (/usr/local/bin/node)
   npm    : 8.12.1
   OS     : macOS Monterey

### Additional Information

Thanks and sorry for my medium english !

sean-perkins:
Hello @DeyBo! Can you try with this dev-build and let me know if you run into any issues?

```shell
npm install @ionic/angular@6.2.9-dev.11664315083.10f0210f
```

In this build any `aria-label` specified on the host of `ion-item` or `ion-card` when using `button`, will be reflected to the inner button.

[...]

PR #26028, opened 2022-09-27
fix(item,card): aria-label is reflected to the inner button

sean-perkins:
## Pull request checklist

## Pull request type

 

 

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

`aria-label` is not reflected to the focusable element (button) when using `ion-item` or `ion-card`.

Issue URL: #25885

## What is the new behavior?

- If an `aria-label` is set on an `ion-card` using `button`, it will be reflected to the inner button.
- If an `aria-label` is set on an `ion-item` using `button`, it will be reflected to the inner button.

- [ ] Yes
- [x] No

## Other information

Opted against writing a spec test for `ion-item`, since it is rendered x2 on load, causing issues with the spec test.

Created a basic card template, since no example exists currently for quick preview/manipulation.

### DIFF
--- a/core/src/components/card/card.tsx
+++ b/core/src/components/card/card.tsx
@@ -1,9 +1,11 @@
 import type { ComponentInterface } from '@stencil/core';
-import { Component, Host, Prop, h } from '@stencil/core';
+import { Element, Component, Host, Prop, h } from '@stencil/core';
 
 import { getIonMode } from '../../global/ionic-global';
 import type { AnimationBuilder, Color, Mode, RouterDirection } from '../../interface';
 import type { AnchorInterface, ButtonInterface } from '../../utils/element-interface';
+import type { Attributes } from '../../utils/helpers';
+import { inheritAriaAttributes } from '../../utils/helpers';
 import { createColorClasses, openURL } from '../../utils/theme';
 
 /**
@@ -20,6 +22,9 @@ import { createColorClasses, openURL } from '../../utils/theme';
   shadow: true,
 })
 export class Card implements ComponentInterface, AnchorInterface, ButtonInterface {
+  private inheritedAttributes: Attributes = {};
+
+  @Element() el!: HTMLElement;
   /**
    * The color to use from your application's color palette.
    * Default options are: `"primary"`, `"secondary"`, `"tertiary"`, `"success"`, `"warning"`, `"danger"`, `"light"`, `"medium"`, and `"dark"`.
@@ -81,6 +86,10 @@ export class Card implements ComponentInterface, AnchorInterface, ButtonInterfac
    */
   @Prop() target: string | undefined;
 
+  componentWillLoad() {
+    this.inheritedAttributes = inheritAriaAttributes(this.el);
+  }
+
   private isClickable(): boolean {
     return this.href !== undefined || this.button;
   }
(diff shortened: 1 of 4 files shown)


core/src/utils/helpers.ts
```diff
@@ -167,9 +167,14 @@ const ariaAttributes = [
  * Returns an array of aria attributes that should be copied from
  * the shadow host element to a target within the light DOM.
  * @param el The element that the attributes should be copied from.
+ * @param ignoreList The list of aria-attributes to ignore reflecting and removing from the host. Use this in instances where we manually specify aria attributes on the `<Host>` element.
  */
-export const inheritAriaAttributes = (el: HTMLElement) => {
-  return inheritAttributes(el, ariaAttributes);
+export const inheritAriaAttributes = (el: HTMLElement, ignoreList?: string[]) => {
```
amandaejohnston, line 172: Did you mean to use this instead of `inheritAttributes` when updating `card` and `item`?

sean-perkins: Using `inheritAttributes` is intentional. We only want to reflect very specific attributes to the inner `<button>` element inside of `ion-item` and `ion-card`. 

`inheritAriaAttributes` will reflect all aria attributes, including the ones that are manually specified on the `<Host>` element. When I discovered the potential issues with `inheritAriaAttributes`, I posted it in our internal [Slack channel](https://ionicteam.slack.com/archives/C02U5GFFXED/p1664466945191339) about it + added a secondary parameter for future use cases that run into the `<Host>` issue.

amandaejohnston (approved):

core/src/components/card/card.tsx
```diff
@@ -91,11 +100,12 @@ export class Card implements ComponentInterface, AnchorInterface, ButtonInterfac
...
       return [<slot></slot>];
     }
-    const { href, routerAnimation, routerDirection } = this;
+    const { href, routerAnimation, routerDirection, inheritedAttributes } = this;
     const TagType = clickable ? (href === undefined ? 'button' : 'a') : ('div' as any);
+    const { ['aria-label']: ariaLabel } = inheritedAttributes;
     const attrs =
       TagType === 'button'
-        ? { type: this.type }
+        ? { type: this.type, ['aria-label']: ariaLabel }
```
liamdebeasi, line 108: Wouldn't we want to do the following instead?

```js
<TagType
  {...attrs}
  class="card-native"
  part="native"
  disabled={this.disabled}
  onClick={(ev: Event) => openURL(href, ev, routerDirection, routerAnimation)}
  {...inheritedAttributes}
>
```

The current code would only apply the custom label if `TagType === 'button'` and would miss when `TagType === 'a'`.

core/src/components/item/item.tsx
```diff
@@ -359,15 +362,18 @@ export class Item implements ComponentInterface, AnchorInterface, ButtonInterfac
...
     const mode = getIonMode(this);
     const clickable = this.isClickable();
     const canActivate = this.canActivate();
     const TagType = clickable ? (href === undefined ? 'button' : 'a') : ('div' as any);
+    const { ['aria-label']: ariaLabel } = inheritedAttributes;
+
     const attrs =
       TagType === 'button'
-        ? { type: this.type }
+        ? { type: this.type, ['aria-label']: ariaLabel }
```
liamdebeasi, line 376: Same question as with `ion-card`

core/src/utils/helpers.ts
liamdebeasi: Can we move this change to a separate PR if it's not needed to resolve the item/card bug?

core/src/components/card/card.tsx
sean-perkins: Applying the `aria-label` regardless of the `TagType` makes sense to me, I can make that change. I was cautious to append all values of `inheritedAttributes`, since we could potentially reflect attributes to different target elements in the component's template. 

For `ion-card`, I think this wouldn't matter, as the template is relatively simple and flat. 

`ion-item` does have more complexity in its template. 

I can reflect all of them for now and try to document in a way that alerts devs to this possible scenario.

liamdebeasi (approved):

Issue #18617, opened 2019-06-25
bug: segment-button height (ios)


simonhaenisch:
# Bug Report

**Ionic version:**
[x] **4.5**

**Current behavior:**

I built a pagination component using a toolbar, some buttons and a segment. Since upgrading to `@ionic/core >= 4.5`, the segment buttons are too high, because a variable called `--min-height` on the native button inside it changed from `auto` to `56px`.

![image](https://user-images.githubusercontent.com/3410759/60108531-043d5780-9769-11e9-9d59-b65ed913c562.png)

**Expected behavior:**

![image](https://user-images.githubusercontent.com/3410759/60108479-f25bb480-9768-11e9-882a-8aef75ea3482.png)

**Steps to reproduce/Related code:**

https://codepen.io/simonhaenisch/pen/Kjvarm?editors=1000

**Other information:**

BTW there's some really weird behavior when putting the `ion-margin` class on the `ion-segment`.

**Ionic info:**

See Codepen.

simonhaenisch:
Oh, just figured out that it's fixed if I also put `mode="ios"` on the `ion-toolbar` (which wasn't necessary in version 4.4). Is that required?

brandyscarney:
Thanks for the issue! I believe the min-height was added to match the spec but it looks a bit too tall now so I will look into this: https://material.io/components/tabs/#specs

liamdebeasi:
Hi everyone,

Does this issue still reproduce on the latest version of Ionic 6?

simonhaenisch:
Not actively using Ionic myself anymore but there's a codepen link so you should be able to tell within a couple minutes...

simonhaenisch:
<img width="782" alt="image" src="https://user-images.githubusercontent.com/3410759/192964187-1c1fad44-89af-4a9a-8c75-e61837ad2e0b.png">

Looks like it's even worse with Ionic 6 😲

liamdebeasi:
Thanks for the additional information. The following code seems to be the culprit: https://github.com/ionic-team/ionic-framework/blob/f3b5a3f6fcddd70475fc2c6023327dcb411d1ea8/core/src/components/toolbar/toolbar.md.scss#L34-L36

This code only applies to the MD toolbar, but it expects that the slotted segment is also running in MD mode. Your code sample has the `ion-segment` running in iOS mode, so this `min-height` code should not apply.

liamdebeasi:
[...]

PR #17537, opened 2019-02-19
fix(toolbar): update md toolbar button spacing and segment padding

brandyscarney:
### Short description of what this resolves:
Updates the toolbar buttons to follow the spec and modifies the padding for segment in a toolbar

**Fixes**: #16950 #14444

### What's done in this PR:

**This applies to Material Design styles only**

- Removes the padding from the main toolbar so that we can individually style the components inside of it
- Adds a `has-icon-only` class to button, this is used to switch between `unbounded` and `bounded` ripples on buttons in a toolbar. If the button is clear and only has an icon, we use the unbounded "circular" ripple effect, otherwise still use the bounded one. This matches the MD spec, without making the other buttons look weird.
- Using the class above, style the button differently to match the MD spec
- Updates the back button and menu button to use the proper size / icon size
- Removes the opacity on an activated back button, it should use the ripple
- Moves the margin to the slots in a toolbar by grabbing the "first" and "last" slot and applying a class
- Makes the segment in a toolbar use the min height from `--min-height` on the toolbar
- Updates the back button so that it matches the [MD spec](https://material.io/design/navigation/understanding-navigation.html#forward-navigation).
- Updates the header box shadow to use the old v3 datauri per Ben

### Previews

*Ionic on the left, Material Design on the right*

Screen Shot 2019-05-15 at 5 47 06 PM

*Comparison between MD & Ionic Back Button*

*Comparison between MD & Ionic Other Buttons*

*Updated Ripple Effect*

PR #26042, opened 2022-09-29
fix(toolbar): MD height only applies to MD segment

liamdebeasi:
⚠️ Give co-author credit to Brandy when merging.

## Pull request checklist

## Pull request type

 

 

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

Issue URL: resolves https://github.com/ionic-team/ionic-framework/issues/18617

This relevant only applies to the MD toolbar, but it expects that the slotted segment is also running in MD mode. When using the segment in iOS mode, the min height is applied too.

This fix was done to align with the MD spec for segments in a toolbar: https://github.com/ionic-team/ionic-framework/pull/17537

## What is the new behavior?

- The min-height now only applied to MD segments.

- [ ] Yes
- [x] No


brandyscarney:
Could we just apply `min-height` to all slotted `ion-segment` in `toolbar.scss` and set the `ios` one to something like `auto`?

liamdebeasi:
What's the benefit of doing it that way?

brandyscarney:
That way we don't have a CSS variable that only works in `md` mode.

liamdebeasi:
The `--min-height` CSS Variable already works in both iOS and MD mode, so this PR does not change that functionality. See https://github.com/ionic-team/ionic-framework/blob/b43f17fc4af1c79ea340204e26ef31598092cacf/core/src/components/toolbar/toolbar.scss#L69

amandaejohnston (approved):

liamdebeasi:
Brandy proposed an alternative offline that I adapted. Updated the implementation and added a test.

amandaejohnston (approved):

brandyscarney (approved):
This looks good. Should we add a screenshot that shows it inheriting the md height though?

liamdebeasi:
We have screenshots for the inheritance already (example: https://github.com/ionic-team/ionic-framework/blob/main/core/src/components/segment/test/toolbar/segment.e2e.ts-snapshots/segment-toolbar-md-rtl-Mobile-Safari-linux.png)

brandyscarney:
@liamdebeasi Sorry I meant one where we set `--min-height` on toolbar to another value and verify it only changes the md one, but if you think that one is enough then this is fine with me.

liamdebeasi:
Yeah I think the current screenshot should be fine since it has `--min-height` explicitly set to `56px`. I'm not sure that having a screenshot where `--min-height: 40px` or something would show us anything new.

PR #26066, opened 2022-10-03
feat(select): ionChange will only emit from user committed changes

liamdebeasi:
⚠️ Be sure to set commit message to the breaking change note when merging 

## Pull request checklist

## Pull request type

 

 

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes

Issue URL: N/A internal ticket

## What is the new behavior?

- [ ] Yes
- [x] No

### DIFF
--- a/BREAKING.md
+++ b/BREAKING.md
@@ -21,6 +21,7 @@ This is a comprehensive list of the breaking changes introduced in the major ver
   - [Overlays](#version-7x-overlays)
   - [Range](#version-7x-range)
   - [Segment](#version-7x-segment)
+  - [Segment](#version-7x-select)
   - [Slides](#version-7x-slides)
   - [Textarea](#version-7x-textarea)
   - [Virtual Scroll](#version-7x-virtual-scroll)
@@ -108,6 +109,10 @@ iOS:
 
 - The type signature of `value` supports `string | undefined`. Previously the type signature was `string | null | undefined`.
   - Developers needing to clear the checked segment item should assign a value of `''` instead of `null`.
+  
+<h4 id="version-7x-select">Select</h4>
+
+- `ionChange` is no longer emitted when the `value` of `ion-select` is modified externally. `ionChange` is only emitted from user committed changes, such confirming a selected option in the select's overlay.
 
 <h4 id="version-7x-slides">Slides</h4>
 
--- a/BREAKING.md
+++ b/BREAKING.md
@@ -22,7 +22,7 @@ This is a comprehensive list of the breaking changes introduced in the major ver
   - [Range](#version-7x-range)
   - [Searchbar](#version-7x-searchbar)
   - [Segment](#version-7x-segment)
-  - [Segment](#version-7x-select)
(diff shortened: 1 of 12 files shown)


angular/test/base/e2e/src/modal.spec.ts
```diff
@@ -102,18 +102,23 @@ describe('when in a modal', () => {
...
   it('should not render ion-item item-has-value class when control value is undefined', () => {
-    cy.get('[formControlName="select"]').invoke('attr', 'value', undefined);
+    cy.get('#set-to-undefined').click();
     cy.get('#inputWithFloatingLabel').should('not.have.class', 'item-has-value');
   });
 
   it('should not render ion-item item-has-value class when control value is null', () => {
-    cy.get('[formControlName="select"]').invoke('attr', 'value', null);
+    cy.get('#set-to-null').click();
     cy.get('#inputWithFloatingLabel').should('not.have.class', 'item-has-value');
```
liamdebeasi, line 122: Sean, do we still need these tests? I updated to have them set the FormControl value instead of setting the `ion-select` value.

core/src/components/select/test/basic/select.e2e.ts
```diff
@@ -95,3 +95,98 @@ test.describe('select: basic', () => {
     });
   });
 });
+
+test.describe('select: ionChange', () => {
+  test.beforeEach(({ skip }) => {
+    skip.rtl();
```
amandaejohnston, line 101: Do we need both modes or can we skip one of those too?

amandaejohnston (approved):
LGTM once the comments are addressed 👍

BREAKING.md
```diff
@@ -119,6 +120,10 @@ iOS:
 
 - The type signature of `value` supports `string | undefined`. Previously the type signature was `string | null | undefined`.
   - Developers needing to clear the checked segment item should assign a value of `''` instead of `null`.
+  
+<h4 id="version-7x-select">Select</h4>
+
+- `ionChange` is no longer emitted when the `value` of `ion-select` is modified externally. `ionChange` is only emitted from user committed changes, such confirming a selected option in the select's overlay.
```
amandaejohnston, line 126:
```suggestion
- `ionChange` is no longer emitted when the `value` of `ion-select` is modified externally. `ionChange` is only emitted from user committed changes, such as confirming a selected option in the select's overlay.
```
Just noticed this 👀

angular/test/base/e2e/src/modal.spec.ts
sean-perkins: We still need some sort of test coverage on the behavior of reflecting the `.item-has-value` class (from our value accessors). The original bug was that styles were not being assigned correctly for floating labels and `ion-select`, with the timing of the `emitStyle` vs. the value accessor. 

It was tested in a modal based on the reproduction, but we could probably just have a standalone Angular test for this in the future.

sean-perkins (approved):

PR #26067, opened 2022-10-03
feat(loading): isOpen presents and dismisses loading overlay

sean-perkins:
## Pull request checklist

## Pull request type

 

 

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes

`ion-loading` cannot be used inline. 

Issue URL: Internal

## What is the new behavior?

- `ion-loading` supports `isOpen` property to present and dismiss the loading indicator
- Framework delegate will skip cloning element to a parent when there are no CSS classes to append to the parent 

- [ ] Yes
- [x] No

### DIFF
--- a/core/src/components.d.ts
+++ b/core/src/components.d.ts
@@ -1377,6 +1377,7 @@ export namespace Components {
           * Additional classes to apply for custom CSS. If multiple classes are provided they should be separated by spaces.
          */
         "cssClass"?: string | string[];
+        "delegate"?: FrameworkDelegate;
         /**
           * Dismiss the loading overlay after it has been presented.
           * @param data Any data to emit in the dismiss events.
@@ -1391,10 +1392,15 @@ export namespace Components {
           * Animation to use when the loading indicator is presented.
          */
         "enterAnimation"?: AnimationBuilder;
+        "hasController": boolean;
         /**
           * Additional attributes to pass to the loader.
          */
         "htmlAttributes"?: LoadingAttributes;
+        /**
+          * If `true`, the loading indicator will open. If `false`, the loading indicator will close. Use this if you need finer grained control over presentation, otherwise just use the loadingController or the `trigger` property. Note: `isOpen` will not automatically be set back to `false` when the loading indicator dismisses. You will need to do that in your code.
+         */
+        "isOpen": boolean;
         /**
           * If `true`, the keyboard will be automatically dismissed when the overlay is presented.
          */
@@ -5163,6 +5169,7 @@ declare namespace LocalJSX {
           * Additional classes to apply for custom CSS. If multiple classes are provided they should be separated by spaces.
          */
         "cssClass"?: string | string[];
+        "delegate"?: FrameworkDelegate;
         /**
           * Number of milliseconds to wait before dismissing the loading indicator.
          */
@@ -5171,10 +5178,15 @@ declare namespace LocalJSX {
           * Animation to use when the loading indicator is presented.
          */
         "enterAnimation"?: AnimationBuilder;
+        "hasController"?: boolean;
         /**
           * Additional attributes to pass to the loader.
          */
         "htmlAttributes"?: LoadingAttributes;
+        /**
+          * If `true`, the loading indicator will open. If `false`, the loading indicator will close. Use this if you need finer grained control over presentation, otherwise just use the loadingController or the `trigger` property. Note: `isOpen` will not automatically be set back to `false` when the loading indicator dismisses. You will need to do that in your code.
+         */
+        "isOpen"?: boolean;
         /**
           * If `true`, the keyboard will be automatically dismissed when the overlay is presented.
          */
--- a/core/src/components.d.ts
+++ b/core/src/components.d.ts
@@ -5204,6 +5204,14 @@ declare namespace LocalJSX {
          */
         "mode"?: "ios" | "md";
         /**
+          * Emitted after the loading indicator has dismissed. Shorthand for ionLoadingDidDismiss.
+         */
+        "onDidDismiss"?: (event: IonLoadingCustomEvent<OverlayEventDetail>) => void;
+        /**
+          * Emitted after the loading indicator has presented. Shorthand for ionLoadingWillDismiss.
+         */
+        "onDidPresent"?: (event: IonLoadingCustomEvent<void>) => void;
+        /**
           * Emitted after the loading has dismissed.
          */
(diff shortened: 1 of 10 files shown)


core/src/components.d.ts
```diff
@@ -1391,10 +1392,15 @@ export namespace Components {
           * Animation to use when the loading indicator is presented.
          */
         "enterAnimation"?: AnimationBuilder;
+        "hasController": boolean;
         /**
           * Additional attributes to pass to the loader.
          */
         "htmlAttributes"?: LoadingAttributes;
+        /**
+          * If `true`, the loading indicator will open. If `false`, the loading indicator will close. Use this if you need finer grained control over presentation, otherwise just use the loadingController or the `trigger` property. Note: `isOpen` will not automatically be set back to `false` when the loading indicator dismisses. You will need to do that in your code.
```
sean-perkins, line 1401: `trigger` is not supported as of this pull request, but I opted it for aligning the documentation with `ion-modal` now vs. risking forgetting to update this later. When `FW-2334` lands in `feature-7.0`, this will be accurate.

core/src/utils/framework-delegate.ts
```diff
@@ -86,7 +86,7 @@ export const CoreDelegate = () => {
       BaseComponent.appendChild(el);
 
       await new Promise((resolve) => componentOnReady(el, resolve));
-    } else if (BaseComponent.children.length > 0) {
+    } else if (BaseComponent.children.length > 0 && cssClasses.length > 0) {
```
sean-perkins, line 89: This change is intentional. Without it, the logic of cloning to a parent element causes exceptions and broken behavior when the loading component is used inline. 

The logic here is only intended to reflect specified CSS classes to a host element. When there is no CSS classes, we can skip that behavior.

core/src/components/loading/loading.tsx
```diff
@@ -185,6 +258,39 @@ export class Loading implements ComponentInterface, OverlayInterface {
...
+  /**
+   * Determines whether or not an overlay
+   * is being used inline or via a controller/JS
+   * and returns the correct delegate.
+   * By default, subsequent calls to getDelegate
+   * will use a cached version of the delegate.
+   * This is useful for calling dismiss after
+   * present so that the correct delegate is given.
+   */
+  private getDelegate(force = false) {
```
sean-perkins, line 270: ~ With this logic being 1:1 to modal, thoughts on abstracting this into an `overlays/utils` or similar, so implementations would call:

```ts
const { delegate } = this.getDelegate(this);
```

It would assume that `inline`, `hasController`, `workingDelegate`, `coreDelegate` and `delegate` are all members of the class.

core/src/components/loading/test/isOpen/loading.e2e.ts
```diff
@@ -0,0 +1,29 @@
+import { test } from '@utils/test/playwright';
+
+test.describe('loading: isOpen', () => {
+  test.beforeEach(async ({ page, skip }) => {
+    skip.rtl('isOpen does not behave differently in RTL');
```
amandaejohnston, line 5: Do we need both modes here? If not, could we skip one of them too?

sean-perkins: I think we can skip one here as well 👍 I need to remember that testing across modes is only valuable for design differences or if we are explicitly checking for mode and doing something different in the implementation.

core/src/components/loading/loading.tsx
sean-perkins: Updated: https://github.com/ionic-team/ionic-framework/pull/26067/files#diff-e6fe3d2c2375aaff02644f8767f36a3c49825d333f9764644871b0541a95b296R617-R675

Let me know your thoughts 👍

amandaejohnston (approved):

core/src/components/loading/loading.tsx
```diff
@@ -158,11 +253,22 @@ export class Loading implements ComponentInterface, OverlayInterface {
...
+    const dismissed = await this.currentTransition;
+
+    if (dismissed) {
+      const { delegate } = this.delegateController.getDelegate();
+      if (delegate) {
+        await detachComponent(delegate, this.usersElement);
+      }
+    }
+
+    return dismissed;
```
liamdebeasi, line 271: I do not understand the purpose of this code. `this.usersElement` should only be defined in components that accept user-defined children such as in `ion-modal` and `ion-popover`. `ion-loading` does not accept user-defined children, so this should not be needed.

sean-perkins: I can refactor this. The inline overlays require that `attachViewToDom`/`removeViewFromDom` is called when presenting/dismissing; otherwise the overlay cannot be re-presented after being dismissed.

The `detachComponent` utility handled this behavior, but also deals with the user's provided component, which isn't needed here.

core/src/utils/framework-delegate.ts
liamdebeasi: I agree that we should skip this, but I don't think this is the correct check as it will break modals/popovers without css classes. This if-block is designed to ensure that user-specified components passed into `ion-modal` and `ion-popover` have a single single parent element.

Example:

```html
<ion-modal>
  <my-cmp></my-cmp>
  <my-cmp></my-cmp>
</ion-modal>
```

would become

```html
<ion-modal>
  <div class="ion-page">
    <my-cmp></my-cmp>
    <my-cmp></my-cmp>
  </div>
</ion-modal>
```

For `ion-loading` this is not needed because there is no user-specified components. Instead, this code was incorrectly moving the internals of `ion-loading`.

Perhaps we could add add some kind of `hasUserDefinedComponent` check? It could check to see if `BaseComponent.children.length > 0` and if the component is a modal or popover.

liamdebeasi (changes requested):

core/src/utils/overlays.ts
```diff
@@ -669,7 +669,30 @@ export const createDelegateController = (ref: {
...
+  /**
+   * Attaches a component in the DOM. Teleports the component
+   * to the root of the app.
+   * @param component The component to optionally construct and append to the element.
+   */
+  const attachViewToDom = async (component?: any) => {
+    const { delegate } = getDelegate(true);
+    if (delegate) {
+      await delegate.attachViewToDom(ref.el, component);
+    }
```
liamdebeasi, line 681: Should this throw an error if there's no delegate? We have something similar for modal/popover delegates: https://github.com/ionic-team/ionic-framework/blob/main/core/src/utils/framework-delegate.ts#L17

liamdebeasi (approved):

Issue #25776, opened 2022-08-17
bug: Setting ion-datetime value does not update selected day


aktivdigital-frontend:
### Prerequisites

- [X] I have read the [Contributing Guidelines](https://github.com/ionic-team/ionic-framework/blob/main/.github/CONTRIBUTING.md#creating-an-issue).
- [X] I agree to follow the [Code of Conduct](https://ionicframework.com/code-of-conduct).
- [X] I have searched for [existing issues](https://github.com/ionic-team/ionic-framework/issues) that already report this problem, without success.

### Ionic Framework Version

- [ ] v4.x
- [ ] v5.x
- [X] v6.x
- [X] Nightly

### Current Behavior

When setting the value of an `ion-datetime` in a `setTimeout`, the display date updates, but after opening the datetime, the selected day does not display.

### Expected Behavior

After setting the value of a `ion-datetime`, the selected day should also update.

### Steps to Reproduce

1. Have an `ion-datetime` (example uses `ion-datetime-button`)
2. Update the value after a few seconds
3. Open the date picker
4. Observe: The new value does not show as the selected day
5. Expected: The new value would display as the highlighted/selected day

### Code Reproduction URL

https://stackblitz.com/edit/angular-ramjpp?file=src%2Fapp%2Fexample.component.html,src%2Fapp%2Fexample.component.ts

### Ionic Info

N/A

sean-perkins:
@aktivdigital-frontend thanks for reporting this issue. I am seeing the same behavior as well. 

The datetime has internal logic that requires the date be active & that we should highlight the day:
https://github.com/ionic-team/ionic-framework/blob/436a8ce50847d26f8fb63ce64ea85be6f5c0bc8e/core/src/components/datetime/datetime.tsx#L1927

`highlightActiveParts` is unfortunately never set when the value is dynamically changed, so the day will never show as selected unless the user manually taps the day again.

liamdebeasi:
@aktivdigital-frontend Can you try the following dev build and let me know if it resolves the issue?

```
npm install @ionic/angular@6.2.10-dev.11664900496.1d961808
```

Note: You may need to install this in a local Ionic app. StackBlitz sometimes has issues installing dev builds of Ionic.

liamdebeasi:
[...]

PR #26070, opened 2022-10-04
fix(datetime): setting date async updates calendar grid

liamdebeasi:
## Pull request checklist

## Pull request type

 

 

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

Issue URL: resolves https://github.com/ionic-team/ionic-framework/issues/25776

Setting `value` programmatically when no value was initially selected was not highlighting the correct day because of a defect in how we update `highlightActiveParts`.

`activeParts` was originally designed to default to today's date when no `value` was set. This had the side effect of today's date being highlighted, indicating that it was selected, even if `value === undefined`.

This was confusing, and we changed the behavior in https://github.com/ionic-team/ionic-framework/pull/25151 so that today's date was outlined when it was not explicitly selected. However, we opted not to change the underlying behavior of `activeParts` defaulting to today. This behavior is what is ultimately responsible for the reported issue. (In other words, yes `highlightActiveParts` is preventing the datetime from working as expected, but `highlightActiveParts` exists because `activeParts` is not working correctly)

`activeParts` was designed to represent dates that had been selected by the user _or_ today's date. Since defaulting to today's date is a behavior we are actively trying to avoid with `highlightActiveParts`, `activeParts` should no longer default to today's date.

## What is the new behavior?

- `activeParts` now defaults to `[]` instead of today's date
-  I added a helper utility so that elements of datetime can still have today's date as a fallback.
- Remove highlightActiveParts logic since it is no longer needed.

- [ ] Yes
- [x] No


core/src/components/datetime/utils/parse.ts
```diff
@@ -115,6 +115,7 @@ export function parseDate(val: string | string[] | undefined | null): DatetimePa
     hour: parse[4],
     minute: parse[5],
     tzOffset,
+    ampm: parse[4] < 12 ? 'am' : 'pm',
```
liamdebeasi, line 118: `parseDate` now includes the `ampm` information which is required for this fix.

 Do we still need this code and its implementation? https://github.com/ionic-team/ionic-framework/blob/02756c5fe408fc9957c25a1be008d5b8b708953d/core/src/components/datetime/datetime.tsx#L1210

sean-perkins: `parseDate` will be called on the value(s), but will not be called on the `minParts` or `maxParts`, which can be returned from `clampDate`. 

If the value is before the min or greater than the max, I believe it would result in `ampm` being `undefined` if the `parseAmPm` was removed here.

liamdebeasi: Ah good point. I suppose we could also update `parseMinParts` and `parseMaxParts` to set the `ampm` value (https://github.com/ionic-team/ionic-framework/blob/02756c5fe408fc9957c25a1be008d5b8b708953d/core/src/components/datetime/utils/parse.ts#L185), but that seems out of scope for this PR, so I will leave `parseAmPm` alone for now.

sean-perkins (approved):

amandaejohnston (approved):

PR #26073, opened 2022-10-04
feat(input, textarea): change default debounce to undefined

amandaejohnston:
## Pull request checklist

## Pull request type

 

 

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes

`ion-input` and `ion-textarea` have default `debounce` values of `0`. This causes the `ionInput` emission to be wrapped in a `setTimeout` of 0, adding an unnecessary delay.

Issue URL: N/A

## What is the new behavior?

- input and textarea now have a default `debounce` of `undefined`, and when `debounce` is undefined, no `setTimeout` is used. (This brings the components in line with the changes to searchbar here: https://github.com/ionic-team/ionic-framework/pull/26026)
- searchbar's `originalIonInput` was tweaked slightly for better type safety.

- [x] Yes
- [ ] No

For `ion-input` and `ion-textarea`, the `debounce` property's default value has changed from 0 to `undefined`. If `debounce` is undefined, the `ionInput` event will fire immediately.


BREAKING.md
```diff
@@ -76,6 +76,8 @@ This section details the desktop browser, JavaScript framework, and mobile platf
 
 - The `debounce` property has been updated to control the timing in milliseconds to delay the event emission of the `ionInput` event after each keystroke. Previously it would delay the event emission of `ionChange`.
 
+- The `debounce` property's default value has changed from 0 to `undefined`. If `debounce` is undefined, the `ionInput` event will fire immediately.
```
sean-perkins, line 79:
```suggestion
- The `debounce` property's default value has changed from `0` to `undefined`. If `debounce` is undefined, the `ionInput` event will fire immediately.
```

BREAKING.md
```diff
@@ -138,6 +140,8 @@ Developers using these components will need to migrate to using Swiper.js direct
...
```
sean-perkins, line 143:
```suggestion
- The `debounce` property's default value has changed from `0` to `undefined`. If `debounce` is undefined, the `ionInput` event will fire immediately.
```

sean-perkins (approved):
Optional suggestions to help make the previous value standout in the breaking changes. Thanks for doing this!

core/src/components/input/input.tsx
```diff
@@ -293,6 +301,7 @@ export class Input implements ComponentInterface {
   }
 
   componentDidLoad() {
+    this.originalIonInput = this.ionInput;
```
sean-perkins, line 304: Came across this when working on the `ion-range` v7 work, should this be in `connectedCallback` instead of waiting for the component to load?

liamdebeasi (approved):

core/src/components/input/input.tsx
amandaejohnston: What were you running into on range?

sean-perkins: I did not run into a specific issue, but was viewing the change top-down in `ion-range` and thinking about the potential test cases that exist.

Waiting until after `render` for storing a reference to the original event, seems strange. If for nothing other than logical organization, I recommend we move this behavior up into `connectedCallback` so that it has no risk of the event being re-assigned before we are able to cache it:

https://stenciljs.com/docs/component-lifecycle

amandaejohnston: I played around with this some. The practical difference is that `connectedCallback` fires again if the component is removed and re-added to the DOM, and it turns out this can cause problems if the component's `debounce` is defined on reconnect. This causes `originalIonInput` to be set to a debounced version of the event emitter. So if `debounce` is later set to `undefined`, the event stays debounced. Screencast: https://user-images.githubusercontent.com/90629384/195137814-4215a298-4e37-47a7-8148-3c07c185ac4e.mp4

This is obviously an extremely specific set of circumstances, but it may be indicative of other similar issues, so I'd rather keep the `originalIonInput` init where it is since the benefit of moving it seems to be mainly semantic. (You may want to adjust `ion-range` though 👀)

sean-perkins: Sounds good, thanks for doing the extra discovery here!

PR #26081, opened 2022-10-05
feat(loading): trigger presents loading overlay

sean-perkins:
## Pull request checklist

## Pull request type

 

 

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes

`ion-loading` cannot be used with a `trigger` element.

Issue URL: Internal

## What is the new behavior?

- `ion-loading` supports a `trigger` prop to specify the id of the element, that when clicked, will present the loading overlay.
- Abstracted base trigger interaction behavior into `createTriggerController()` utility

- [ ] Yes
- [x] No

## Other information

Builds on top of https://github.com/ionic-team/ionic-framework/pull/26067


core/src/utils/overlays.ts
```diff
@@ -612,3 +612,45 @@ export const safeCall = (handler: any, arg?: any) => {
...
+    if (destroyTriggerInteraction) {
+      destroyTriggerInteraction();
+    }
+
+    const triggerEl = trigger !== undefined ? document.getElementById(trigger) : null;
+    if (!triggerEl) {
+      return;
+    }
+
+    const configureTriggerInteraction = (trigEl: HTMLElement, overlayEl: HTMLIonOverlayElement) => {
```
amandaejohnston, line 639: The duplicate function name between this and the outer `configureTriggerInteraction` is a little confusing; could we add comments if this is intentional to work some magic, or else rename one of them?

core/src/utils/overlays.ts
```diff
@@ -612,3 +612,45 @@ export const safeCall = (handler: any, arg?: any) => {
...
+ * Constructs a trigger interaction for an overlay.
+ * Presents an overlay when the trigger is clicked.
+ *
+ * Usage:
+ * ```ts
+ * triggerController = createTriggerController();
+ * triggerController.configureTriggerInteraction(el, trigger);
+ * ```
+ */
+export const createTriggerController = () => {
```
amandaejohnston, line 626: I'm guessing other components could be adjusted to use this as well? (Maybe better for a separate PR?)

sean-perkins: Correct. I think the only exception is `ion-popover`, since it has additional trigger actions/logic that wouldn't make sense being abstracted into this. 

But `ion-modal` could be refactored to this, if we agree with the implementation pattern, and the remaining inline overlays could benefit from this.

core/src/utils/overlays.ts
sean-perkins: Agreed this is confusing. I renamed the public implementation to `initializeClickListener`.

Usage:
```ts
triggerController = createTriggerController();
triggerController.initializeClickListener(el, trigger);
```

amandaejohnston (approved):

core/src/components/loading/loading.tsx
```diff
@@ -138,6 +139,19 @@ export class Loading implements ComponentInterface, OverlayInterface {
...
+  /**
+   * An ID corresponding to the trigger element that
+   * causes the loading indicator to open when clicked.
+   */
+  @Prop() trigger: string | undefined;
+  @Watch('trigger')
+  triggerChanged() {
+    const { trigger, el, triggerController } = this;
+    if (trigger) {
+      triggerController.initializeClickListener(el, trigger);
```
liamdebeasi, line 151: I know we do not explicitly destroy the listener when unmounting the component in modal/popover, but I think we should. The `click` listener likely will not be GC'd when the overlay is removed because the click listener is on the button, not the overlay. Perhaps we can create in `connectedCallback` and destroy in `disconnectedCallback`?

Here's an example of what I mean: https://codepen.io/liamdebeasi/pen/XWqyaWq

Observe that there is a click listener on `ion-button` that was placed there by the modal. (You can remove the `trigger` prop from the modal to see this listener go away). Call `remove()` on the `ion-modal` element. Then manually run garbage collection by clicking the trashcan button in the "Memory" tab in Chrome dev tools. The click listener on `ion-button` should still be there.

liamdebeasi (changes requested):

core/src/components/loading/loading.tsx
sean-perkins: Should I create a new bug to track `ion-modal` and `ion-popover`? Do we want to fix those as part of 6.x or include this change only with the 7.x changes? 

Instead of adding the event listener in the `connectedCallback`, I think we should do that in `componentDidLoad`. Otherwise the overlay could try to present from a click event on the trigger element, before the overlay has actually rendered.

liamdebeasi: Separate ticket for 6.x is fine

liamdebeasi: Only concern about `componentDidLoad` is how do you re-attach the listener if the node was removed and then re-added to the DOM?

core/src/utils/overlays.ts
```diff
@@ -673,3 +673,60 @@ export const createDelegateController = (ref: {
...
+   */
+  const addClickListener = (el: HTMLIonOverlayElement, trigger: string): void => {
+    removeClickListener();
+
+    const triggerEl = trigger !== undefined ? document.getElementById(trigger) : null;
+    if (!triggerEl) {
+      return;
+    }
+
+    const configureTriggerInteraction = (trigEl: HTMLElement, overlayEl: HTMLIonOverlayElement) => {
```
liamdebeasi, line 714: Nit: We use the full names elsewhere (i.e. `overlayEl` not `overEl`), so we should probably call this `triggerEl`

core/src/components/loading/test/trigger/index.html
```diff
@@ -0,0 +1,58 @@
...
+
+      <ion-content class="ion-padding">
+        <div class="grid">
+          <div class="grid-item">
+            <h2>Default</h2>
+            <ion-button id="default">Open Loading</ion-button>
+          </div>
+        </div>
+
+        <ion-loading trigger="default" message="Hello world"></ion-loading>
```
liamdebeasi, line 54: Trying to present the loading again after it dismisses does not work. This is because the `ion-loading` is being moved, so `disconnectedCallback` is fired which destroys the trigger listener. Since the listener is never re-created on `connectedCallback`, trying to open the loading component again does not work.

core/src/components/loading/loading.tsx
sean-perkins: Update: `connectedCallback` is the correct call here, since the overlay can be re-added to the DOM.

liamdebeasi (approved):
Great job!

Issue #24452, opened 2021-12-20
bug: `ion-datetime` ngModel usage emits ionChange event without value change every time after initial load


hatsantos:
### Prerequisites

- [X] I have read the [Contributing Guidelines](https://github.com/ionic-team/ionic-framework/blob/main/.github/CONTRIBUTING.md#creating-an-issue).
- [X] I agree to follow the [Code of Conduct](https://ionicframework.com/code-of-conduct).
- [X] I have searched for [existing issues](https://github.com/ionic-team/ionic-framework/issues) that already report this problem, without success.

### Ionic Framework Version

- [ ] v4.x
- [ ] v5.x
- [X] v6.x

### Current Behavior

I'm migrating an app from Ionic 5 to Ionic 6.

When using the ion-datetime component, the first time the component is used it works as aspected, the component does not dispatch the ionChange event when is open, it dispatch the event only after the first selection... but after a value is selected, if a click is done on the component (second time) a ionChange is dispatched without any change or selection.

### Expected Behavior

The even should only be dispatched if a change occurs, for example, a different day is selected.

### Steps to Reproduce

1 - Add a button/input to invoke the ion-datetime component in a modal with the default buttons on the ion-datetime component;
2 - Setup a ngModel and a ionChange event on the ion-datetime component;
3 - Do not set any value to the ngModel;
4 - Click on the component to invoke the ion-datetime modal;
5 - No ionChange event happens (that seems good);
5 - Selected a value and confirm the selected value;
6 - Value is updated and a ionChange event is dispatched;
7 - Click on the component to invoke the ion-datetime modal;
8 - An ionChange event happens, with out any selection, change other user interaction (new day selected);

And example can be installed from here: https://github.com/hatsantos/Ionic6DateTimeIonChangeBug

I added console logs and a loading screen with a 2 seconds delay when the ionChange is dispatched.

### Code Reproduction URL

https://github.com/hatsantos/Ionic6DateTimeIonChangeBug

### Ionic Info

Ionic:

   Ionic CLI                     : 6.18.1 (C:\Users\hsantos\AppData\Roaming\nvm\v14.18.2\node_modules\@ionic\cli)
   Ionic Framework               : @ionic/angular 6.0.1
   @angular-devkit/build-angular : 13.1.2
   @angular-devkit/schematics    : 13.1.2
   @angular/cli                  : 13.1.2
   @ionic/angular-toolkit        : 5.0.3

Capacitor:

[...]

Issue #20873, opened 2020-03-25
bug: ion-datetime triggers ionChange on load


jeydsamy:
# Bug Report

**Ionic version:**

[x] **5.x**

**Current behavior:**

ion-datetime is triggering ionChange on page load

**Expected behavior:**

ion-datetime should trigger ionChange only after user changes the date/time

**Steps to reproduce:**

A sample application via GitHub
https://github.com/jeydsamy/ionic-datetime-sample
1. Clone project from git, npm install and run with ionic serve
2. Navigate between tabs, console will print "Date Changed!" even when it is not changed.

jeydsamy:
@liamdebeasi Can I please get an answer on this?

liamdebeasi:
Thanks for the issue. I was able to reproduce this behavior. Seems similar to https://github.com/ionic-team/ionic/issues/19204.

jeya-duraisamy:
Yes, it is similar to #19204. Thanks for looking into it @liamdebeasi

jeya-duraisamy:
@liamdebeasi Sorry if it is too soon, any idea when this might be fixed?

sean-perkins:
Hello everyone, we have an open RFC that proposes a change to `ionChange` that would help alleviate this issue. 

Please take a moment to review and let us know your thoughts: https://github.com/ionic-team/ionic-framework/discussions/25532

liamdebeasi:
Thanks for the issue. This has been resolved via https://github.com/ionic-team/ionic-framework/pull/26083, and a fix will be available in an upcoming release of Ionic Framework. Please note that this fix requires a breaking change, so it is being targeted for the Ionic 7.0 release. Please feel free to continue testing the dev build, and let me know if you have any questions.

PR #26083, opened 2022-10-06
feat(datetime): ionChange will only emit from user committed changes

liamdebeasi:
⚠️ Be sure to set commit message to the breaking change note when merging 

## Pull request checklist

## Pull request type

 

 

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes

Issue URL: resolves https://github.com/ionic-team/ionic-framework/issues/20873, resolves https://github.com/ionic-team/ionic-framework/issues/24452

## What is the new behavior?

- [x] Yes
- [ ] No


core/src/components/datetime/datetime.tsx
```diff
@@ -1176,7 +1189,7 @@ export class Datetime implements ComponentInterface {
 
     const { minParts, maxParts, multiple } = this;
     if (!multiple && Array.isArray(value)) {
-      this.value = value[0];
+      this.setValue(value[0]);
```
amandaejohnston, line 1192: `processValue` only gets called when the datetime has a default value set (or when resetting to said default via `reset()`) so I don't think we'd want to fire `ionChange` here?

If `ionChange` should fire when resetting, we could manually fire it in `reset`, though we'd also need to remove the `ionChange` fire in `clearButtonClick` below since that calls `reset`.

liamdebeasi: Not sure I follow. In the line you highlighted, the `value` would change from something like`['2022-01-01', '2022-01-02']` to `'2022-01-01'` which is the same process that the accordion group follows: https://github.com/ionic-team/ionic-framework/blob/b052d3b2622b795cde102591f0191338e15b14a0/core/src/components/accordion-group/accordion-group.tsx#L86

amandaejohnston: The difference is that this line is from a function that's only called when initing the datetime with a defined default value, or resetting to that default value. So if you had something like this (Vue syntax):
```
<ion-datetime :multiple="false" :value="['2022-01-01', '2022-01-02', '2022-01-03']"></ion-datetime>
```
... that would cause `ionChange` to fire immediately on load when the `value` is updated from an array to a single value (due to `multiple` being `false`). So, I believe we should just set `this.value` directly instead of going through `setValue`, to avoid the `ionChange` fire. I'm cool with keeping it as-is if that's desired behavior; just checking since it struck me as a little weird.

liamdebeasi: Discussed offline. We are going to change datetime to no longer adjust the value when used incorrectly. Instead we will warn/error in the console.

amandaejohnston (approved):

sean-perkins (approved):

PR #26084, opened 2022-10-06
fix(card-header): reverse order of title and subtitle on ios

brandyscarney:
## Pull request checklist

## Pull request type

 

 

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

Card header displays the title & subtitle in the order they are placed in the DOM

Issue URL: N/A 

This was brought up while implementing the card documentation here: https://github.com/ionic-team/ionic-docs/pull/2584#discussion_r984583915

## What is the new behavior?

Card header changes the display to flex and reverses the order for the title & subtitle on `ios` mode. 

- [ ] Yes
- [ ] No
- [X] Maybe

Unknown. This may be considered a breaking change if users are adding any other elements into the `ion-card-header` or if they are already changing the order to account for iOS. If they are leaving it as title -> subtitle then this would be a fix to get it closer to native. 

Looking through the [card api documentation](https://ionicframework.com/docs/api/card) it looks like we were showing them usage with the order reversed, likely for iOS but this would've been wrong for Material Design.

The [kitchen sink demo](https://ionicframework.com/docs/) ([source code](https://github.com/ionic-team/docs-demo/blob/main/src/components/card/card.tsx#L22-L25)) also showed subtitle before title, making it wrong for Material Design.

## Other information

I believe this needs some discussion before implementing it.


sean-perkins (commented):
Personally, I think we should treat this as a breaking change and target for v7. While it doesn't change public APIs, it does change the default style representation for iOS. 

We have been considering other misc. spec alignment/design clean-up tasks as v7 as well, for similar reasons. 

All this being said - I agree with the changes. I do think we need new screenshot tests captured for this change though.

amandaejohnston:
I agree with this change, and also that we should target v7 instead. Holding on review until tests pass though; you may need to run the screenshot update job.

liamdebeasi (approved):
Ditto to making this a v7 change.

amandaejohnston (approved):

sean-perkins (approved):

PR #26089, opened 2022-10-07
feat(range): ionChange will only emit from user committed changes

sean-perkins:
## Pull request checklist

## Pull request type

 

 

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes

`ion-range` will emit an `ionChange` event when the value is programmatically assigned.

Issue URL: Internal

## What is the new behavior?

- `ion-range` will only emit `ionChange` from user interaction, such as dragging and releasing the knob or using keyboard navigation to change the value.
- The `debounce` property defaults to `undefined` instead of `0`. 
- Isolated form control test for individual component vs. whole suite (simplifies tests moving forward and avoids conflicts with cypress interacting with element).
- Introduced `ionInput` event on `ion-range`, that is emitted whenever the internal value changes (such as dragging the range knob). 

- [x] Yes
- [ ] No

- `ionChange` is no longer emitted when the `value` of `ion-range` is modified externally. `ionChange` is only emitted from user committed changes, such as dragging and releasing the range knob or selecting a new value with the keyboard arrows.
  - If your application requires immediate feedback based on the user actively dragging the range knob, consider migrating your event listeners to using `ionInput` instead.

- The `debounce` property's value value has changed from `0` to `undefined`. If `debounce` is undefined, the `ionInput` event will fire immediately.


core/src/components/range/range.tsx
```diff
@@ -73,11 +74,16 @@ export class Range implements ComponentInterface {
...
 
   @Watch('debounce')
   protected debounceChanged() {
-    this.ionChange = debounceEvent(this.ionChange, this.debounce);
+    const { ionChange, debounce, originalIonChange } = this;
+    /**
+     * If debounce is undefined, we have to manually revert the ionChange emitter in case
+     * debounce used to be set to a number. Otherwise, the event would stay debounced.
+     */
+    this.ionChange = debounce === undefined ? originalIonChange ?? ionChange : debounceEvent(ionChange, debounce);
```
sean-perkins, line 86: The Range component does not have an `ionInput` event to migrate the debounce implementation to. Should we still offer `debounce` here or alternatively, should we have an `ionInput` event that emits for each value change as the knob is dragged? 

The current `ionChange` emission pattern aligns with `<input type="range" />`, where the `change` event is not emitted until the knob is released or from a keypress.

core/src/components/range/range.tsx
```diff
@@ -382,6 +403,7 @@ export class Range implements ComponentInterface {
     this.update(detail.currentX);
     this.pressedKnob = undefined;
 
+    this.emitValueChange();
```
liamdebeasi, line 406: Why do we emit the value change here instead of in `updateValue`? The `this.update` call above should cause `updateValue` to get called.

amandaejohnston:
I'm seeing `Maximum call stack size exceeded` errors when dragging either knob on a dual knobs range, specifically the one in the basic test. (Doesn't happen in the `feature-7.0` branch.)

core/src/components/range/range.tsx
liamdebeasi: I think we should have an `ionInput` event otherwise developers may see this change as a regression since there currently is no way to know when the value changes while dragging.

liamdebeasi (commented):
Also ditto what Amanda said re: dual knobs

core/src/components/range/range.tsx
sean-perkins: The browser implementation of range does not emit `change` for each move, only on release (or through keyboard navigation). By emitting in `updateValue`, `ionChange` would be emitted for each value change, even before release. 

This was done so that `ionChange` would be only emitted as a result of releasing the knob.

core/src/components/range/test/range-events.e2e.ts
```diff
@@ -0,0 +1,132 @@
...
+      await rangeHandle.hover();
+      await page.mouse.down();
+      await page.mouse.move(boundingBox!.x + 100, boundingBox!.y);
+
+      await page.mouse.up();
+
+      await ionChangeSpy.next();
+
+      expect(ionChangeSpy).toHaveReceivedEvent();
+      expect(ionChangeSpy).toHaveReceivedEventTimes(1);
```
amandaejohnston, line 52:
```suggestion
      expect(ionChangeSpy).toHaveReceivedEventTimes(1);
```
I believe the first call is redundant.

core/src/components/range/test/range-events.e2e.ts
```diff
@@ -0,0 +1,132 @@
...
+
+      const boundingBox = await rangeHandle.boundingBox();
+
+      await rangeHandle.hover();
+      await page.mouse.down();
+      await page.mouse.move(boundingBox!.x + 100, boundingBox!.y);
+
+      await ionInputSpy.next();
+
+      expect(ionInputSpy).toHaveReceivedEvent();
```
amandaejohnston, line 100: Do we want to check that `ionInput` was emitted multiple times throughout the drag? Unsure if `toHaveReceivedEventTimes` would be flaky or something.

angular/test/base/src/app/inputs/inputs.component.ts
```diff
@@ -6,12 +6,11 @@ import { Component } from '@angular/core';
...
 
-  datetime = '1994-03-15';
-  input = 'some text';
+  datetime? = '1994-03-15';
+  input? = 'some text';
   checkbox = true;
   toggle = true;
-  select = 'nes';
-  range = 10;
+  select? = 'nes';
```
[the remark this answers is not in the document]
sean-perkins: The `?` updates the type signature of the field to allow for `undefined`. In this test file, some of the fields are explicitly set to undefined when resetting the form state. These changes are to resolve type errors that are being caught by VSCode/typescript.

core/src/components/range/test/range-events.e2e.ts
sean-perkins: I opted against checking for a specific number of emissions with similar concerns of flaky tests. Since the mouse pointer is not released, this does cover the expected behavior of the `ionInput` event being emitted as the value is being updated (and before `ionChange` is emitted).

core/src/components/range/range.tsx
```diff
@@ -185,14 +190,10 @@ export class Range implements ComponentInterface {
    */
   @Prop({ mutable: true }) value: RangeValue = 0;
   @Watch('value')
-  protected valueChanged(value: RangeValue) {
+  protected valueChanged() {
     if (!this.noUpdate) {
       this.updateRatio();
     }
-
-    value = this.ensureValueInBounds(value);
```
liamdebeasi, line 193: Does this change mean the `ensureValueInBounds` will no longer be called if the developer passes a bad value? If so, we should call that out in the breaking changes guide.

liamdebeasi (approved):

Issue #25894, opened 2022-09-07
bug: Using 2 ion-datetimes linked by min/max causes unexpected re-renders


Chronos77:
### Prerequisites

- [X] I have read the [Contributing Guidelines](https://github.com/ionic-team/ionic-framework/blob/main/.github/CONTRIBUTING.md#creating-an-issue).
- [X] I agree to follow the [Code of Conduct](https://ionicframework.com/code-of-conduct).
- [X] I have searched for [existing issues](https://github.com/ionic-team/ionic-framework/issues) that already report this problem, without success.

### Ionic Framework Version

- [ ] v4.x
- [ ] v5.x
- [X] v6.x
- [ ] Nightly

### Current Behavior

the bug appears when you have two ion-datetimes (with presentation="date-time") linked (by min and max) and you try to put an end date lower than the start date.

For more precision: on the first ion-datetime(startDate) have a max value and the second(endDate) have min date(startDate) and max date (current date)

The app freezes completely

### Expected Behavior

normally it should put us the day before with the corresponding time and don't have a freeze

### Steps to Reproduce

URL to reproduce : https://stackblitz.com/edit/angular-ivy-zh9eog?file=src/app/app.component.ts

- have 2 ion-datetime with 3 days between and with presentation="date-time"
- on the first (startDate) select day - 1 (in the example you can click on 20) and add 1 hours (04:00)
- on the other (endDate) select 20 
- the app freezes completely

### Code Reproduction URL

https://stackblitz.com/edit/angular-ivy-zh9eog?file=src/app/app.component.ts

### Ionic Info

Ionic CLI                     : 6.20.1 (/home/ssa-dw/.nvm/versions/node/v14.15.5/lib/node_modules/@ionic/cli)
   Ionic Framework               : @ionic/angular 6.2.5
   @angular-devkit/build-angular : 13.2.3
   @angular-devkit/schematics    : 13.3.7
   @angular/cli                  : 13.2.6
   @ionic/angular-toolkit        : 6.1.0

Capacitor:

   Capacitor CLI      : 3.5.1
   @capacitor/android : 3.5.1
   @capacitor/core    : 3.5.1
   @capacitor/ios     : 3.5.1

Cordova:

   Cordova CLI       : 10.0.0
   Cordova Platforms : not available
   Cordova Plugins   : not available

Utility:

   cordova-res                          : not installed globally
   native-run (update available: 1.7.0) : 1.6.0

System:

[...]

Issue #25708, opened 2022-08-03
bug: date picker jumps to max date when navigating to next year 


daniel-zero:
### Prerequisites

- [X] I have read the [Contributing Guidelines](https://github.com/ionic-team/ionic-framework/blob/main/.github/CONTRIBUTING.md#creating-an-issue).
- [X] I agree to follow the [Code of Conduct](https://ionicframework.com/code-of-conduct).
- [X] I have searched for [existing issues](https://github.com/ionic-team/ionic-framework/issues) that already report this problem, without success.

### Ionic Framework Version

- [ ] v4.x
- [ ] v5.x
- [X] v6.x
- [ ] Nightly

### Current Behavior

When the `max` date property is used and the user moves from December 2022 to January 2023 the datepicker will switch to the given max date instead of moving to January

### Expected Behavior

The datepicker should switch to January

### Steps to Reproduce

```html
  <ion-datetime
    locale="de-DE"
    presentation="date"
    [min]="'2022-08-03'"
    [max]="'2023-07-29'"
    showDefaultButtons="true"
    firstDayOfWeek="1"
  ></ion-datetime>
```

### Ionic Info

Ionic:

   Ionic CLI                     : 6.20.1 (/Users/danielsogl/.nvm/versions/node/v16.16.0/lib/node_modules/@ionic/cli)
   Ionic Framework               : @ionic/angular 6.2.0
   @angular-devkit/build-angular : 14.1.0
   @angular-devkit/schematics    : 14.1.0
   @angular/cli                  : 14.1.0
   @ionic/angular-toolkit        : 6.1.0

Capacitor:

   Capacitor CLI      : 4.0.1
   @capacitor/android : not installed
   @capacitor/core    : 4.0.1
   @capacitor/ios     : not installed

Utility:

   cordova-res : not installed globally
   native-run  : 1.6.0

System:

   NodeJS : v16.16.0 (/Users/danielsogl/.nvm/versions/node/v16.16.0/bin/node)
   npm    : 8.15.1
   OS     : macOS

daniel-zero:
One thing to note, there looks like there is an Angular 14 issue. Our app with the latest Ionic version and Angular 13 has no issues.

liamdebeasi:
Thanks for the issue. Can you provide a minimal reproduction of the issue?

liamdebeasi:
[...]

PR #26125, opened 2022-10-13
fix(datetime): values are adjusted to be in bounds

liamdebeasi:
## Pull request checklist

## Pull request type

 

 

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

Issue URL: resolves https://github.com/ionic-team/ionic-framework/issues/25894 resolves https://github.com/ionic-team/ionic-framework/issues/25708

The wheel picker component that datetime uses has a built-in recovery system when `value` is no longer represented in `items`. The picker will scroll to the nearest item and update the value.

This was needed by `ion-datetime` when switching between min or max values such that the selected time become invalid. (i.e. selecting the min date that has a min time of 09:00 even though the previously selected time is 08:00).

This caused multiple render issues and unexpected value updates.

## What is the new behavior?

- Moves the adjustment logic out of the picker component and into datetime. Now the value will be adjusted as the state is being updated (whereas the picker was sometimes updating the state immediately after the state was already updated).
Note: Ultimately, the team would like to move away from this automatic value adjustment and align more closely with the native `datetime-local` input behavior. However, that is a breaking change so we opted to fixing the existing behavior for now while we can evaluate alternatives.

- [ ] Yes
- [x] No


core/src/components/datetime/utils/manipulation.ts
```diff
@@ -361,5 +366,55 @@ export const validateParts = (parts: DatetimeParts): DatetimeParts => {
...
+    if (partsCopy.hour !== undefined && minParts.hour !== undefined && partsCopy.hour <= minParts.hour) {
+      partsCopy.hour = minParts.hour;
+
+      /**
+       * If the minute is out of bounds,
+       * set it to the min minute.
+       */
+      if (partsCopy.minute !== undefined && minParts.minute !== undefined && partsCopy.minute < minParts.minute) {
+        partsCopy.minute = minParts.minute;
+      }
```
amandaejohnston, line 390: Playing around with this made me notice something. Say you have a datetime with a `min` of 5/1 at 8AM. You select a value of 5/3 at 4:30AM. You then click on 5/1. With the current logic, because the hour and minute are handled separately, the time is auto-adjusted to 8:30AM.

This feels strange to me; from a user perspective, I would expect the time to go to the min (8AM), since that's the closest possible time to my selection. Maybe the times should be compared as a whole instead? (This would also apply to the `max` logic below.)

liamdebeasi: Good point 👍  I changed this in https://github.com/ionic-team/ionic-framework/pull/26125/commits/0f7db03f203dcafd0bf9539c4741522ccea0ca76.

amandaejohnston (approved):

sean-perkins (approved):
Good work 👍

Issue #25463, opened 2022-06-14
bug: ion-header fade/condense usage is incorrect with Angular CDK Virtual Scroll


mtnair:
### Prerequisites

- [X] I have read the [Contributing Guidelines](https://github.com/ionic-team/ionic-framework/blob/main/.github/CONTRIBUTING.md#creating-an-issue).
- [X] I agree to follow the [Code of Conduct](https://ionicframework.com/code-of-conduct).
- [X] I have searched for [existing issues](https://github.com/ionic-team/ionic-framework/issues) that already report this problem, without success.

### Ionic Framework Version

- [ ] v4.x
- [ ] v5.x
- [X] v6.x
- [ ] Nightly

### Current Behavior

When using the `<ion-header>` component in combination with Angular's `<cdk-virtual-scroll-viewport>` component, to fade/condense the header when scrolling in the list, it does not work. I also added the `ion-content-scroll-host` class to the `<cdk-virtual-scroll-viewport>` component to no avail.

I created a demo with the header effect inspired by the workaround proposed in the feature request #25157:
https://stackblitz.com/edit/ionic6-angular13-rsezrm?file=src%2Fapp%2Fapp.component.ts

If you set `virtualScroll` in the app.component.ts to "false", you can see the expected behavior without a virtual scroll.

Another problem in the current behavior: When setting the height of the `<cdk-virtual-scroll-viewport>` element to 100%, the space used by the `<ion-header>` inside `<ion-content>` causes to overflow the virtual scroll component, so the last element in the list is invisible.

### Expected Behavior

When using the `<ion-header>` component in combination with Angular's `<cdk-virtual-scroll-viewport>` component, the header should fade and condense like it's described in the official documentation: https://ionicframework.com/docs/api/header#usage-with-virtual-scroll

### Steps to Reproduce

1. Run the Stackblitz Demo application: https://ionic6-angular13-rsezrm.stackblitz.io/ OR clone the repo: https://github.com/mtnair/ionic6-virtual-scroll-header-repro
2. Scroll down
3. See the header behavior with the virtual scroll component
4. Change the param `virtualScroll` in `app.component.ts` to "false"
5. Scroll down
6. See the expected behavior with an `<ion-list>` without the virtual scroll

### Code Reproduction URL

https://github.com/mtnair/ionic6-virtual-scroll-header-repro

### Ionic Info

$ ionic info

Ionic:

[...]

PR #26152, opened 2022-10-18
fix(header): fade/condense works with custom scroll target

sean-perkins:
## Pull request checklist

## Pull request type

 

 

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

When using the `ion-header` with a custom scroll target (virtual scroll), the header does not translate out of view and transition between the large title to the condensed display. 

Also, our test template was showing a usage pattern that did not align with the documented approach. This led to confusion with how to correctly use `ion-header` with virtual scroll.

Issue URL: #25463

## What is the new behavior?

- `ion-header` is translated off the view when attaching to a custom scroll target.
- The condense and fade features attach to the correct content and scroll container.

Since the header is being translated off the viewport and the scroll container of the nearest `ion-content` is not rendered, developers are responsible for creating a container that hides overflow, if they do not want the header to show under the condensed header.

e.g.:

```css
#container {
  height: 100%;
  overflow: hidden;
}
```

```html
<ion-content scroll-y="false">
  <div id="container">
    <ion-header></ion-header>
    <virtual-scroll class="ion-content-scroll-host"></virtual-scroll>
  </div>
</div>
```

- [x] Yes
- [ ] No

Before this change, developers were forced to have the `ion-header` inside of the virtual scroll container:

```html
<ion-content>
  <virtual-scroll class="ion-content-scroll-host">
    <ion-header></ion-header>
    <!-- items -->
  </virtual-scroll>
</ion-content>
```

This is incorrect and was not the original intention of the feature. After this change, the `ion-header` can correctly sit outside of the virtual scroll container.

```html
<ion-content>
  <ion-header></ion-header>
  <virtual-scroll class="ion-content-scroll-host">
    <!-- items -->
  </virtual-scroll>
</ion-content>
```

## Other information

Dev-build: `6.3.3-dev.11666118895.1f9c70e6`


amandaejohnston:
We should probably target this for v7 (and update BREAKING.md); I know it's technically not right as-is, but it's not broken, just not using the intended design.

core/src/components/header/header.tsx
```diff
@@ -85,7 +85,7 @@ export class Header implements ComponentInterface {
 
     if (hasCondense) {
       const pageEl = this.el.closest('ion-app,ion-page,.ion-page,page-inner');
-      const contentEl = pageEl ? findIonContent(pageEl) : null;
+      const contentEl = pageEl ? pageEl.querySelector('ion-content') : null;
```
amandaejohnston, line 88: Why was this change made? Isn't this reverting the custom scroll target behavior for `condense`?

sean-perkins: This change was made to have the intersection observer attach the root container as the `ion-content` instead of the virtual scroll target.

e.g.: https://github.com/ionic-team/ionic-framework/blob/23359198231f8b205f2d8578191252f3a411554b/core/src/components/header/header.tsx#L145-L189

The condense feature works by subscribing to different thresholds when the toolbar intersects the root element. 

We are still querying the custom scroll target's scroll container in the implementation: https://github.com/ionic-team/ionic-framework/blob/23359198231f8b205f2d8578191252f3a411554b/core/src/components/header/header.tsx#L154

This is different than the fade implementation, because the fade feature only requires listening to the `scroll` event from our scroll container. 

Prior to this change, if `ion-header` is outside of the scroll container, e.g.:
```html
<ion-content>
  <ion-header></ion-header>
  <div class="ion-content-scroll-host"></div>
</ion-content>
```

The intersection observer would attach to the `.ion-content-scroll-host`. `ion-header` would never intersect this container, since it sits outside of it. This made the condense feature never work with our expected implementation pattern.

sean-perkins:
I noted this as a breaking change since it requires a different implementation to work correctly after this change. So developers that currently have it inside of their `.ion-content-scroll-host` will have to change their code over a patch release, for the feature to work correctly.

amandaejohnston (approved):
Hmm... I'll let others chime in on whether this should target v7 or not, I could go either way on it 🤔

core/src/components/header/header.utils.ts
```diff
@@ -58,6 +58,16 @@ export const handleContentScroll = (scrollEl: HTMLElement, scrollHeaderIndex: He
...
+    /**
+     * Custom scroll targets are used with virtual scroll. We need to move the
+     * header off the viewport when the scroll container is scrolled, so that
+     * the intersection observer for handling the fade and condensed header fires.
+     */
+    const isCustomScrollTarget = scrollEl.classList.contains('ion-content-scroll-host');
+    if (isCustomScrollTarget) {
+      scrollHeaderIndex.el.style.transform = `translateY(${-1 * scrollTop}px)`;
+    }
+
```
liamdebeasi, line 70: This seems a bit risky. Why not have the `ion-header` with the large title inside of `.ion-content-scroll-host`?

Issue #26117, opened 2022-10-13
bug: DatetimeButton with modal results in nested ion-page elements


loopezz:
### Prerequisites

- [X] I have read the [Contributing Guidelines](https://github.com/ionic-team/ionic-framework/blob/main/.github/CONTRIBUTING.md#creating-an-issue).
- [X] I agree to follow the [Code of Conduct](https://ionicframework.com/code-of-conduct).
- [X] I have searched for [existing issues](https://github.com/ionic-team/ionic-framework/issues) that already report this problem, without success.

### Ionic Framework Version

- [ ] v4.x
- [ ] v5.x
- [X] v6.x
- [ ] Nightly

### Current Behavior

Using the ion-datetime-button in combination with inline ion-modal creates infinite nested ion-page elements after opening and closing it multiple times.

### Expected Behavior

The ion-page element must not be rendered or only rendered once.

### Steps to Reproduce

1. https://ionicframework.com/docs/api/datetime-button#basic-usage
2. Open the browser's inspector
3. Inspect 'ion-modal' in the StackBlitz example
4. Verify there is no ion-page element rendered
5. Toggle the datetime selector multiple times
6. Inspect 'ion-modal' in the StackBlitz example again

Before:
<img width="378" alt="Screenshot 2022-10-13 at 10 55 09" src="https://user-images.githubusercontent.com/11312418/195552029-ddf7aceb-6de0-4f00-990b-1e8db2a4ce66.png">

After:
<img width="381" alt="Screenshot 2022-10-13 at 10 55 47" src="https://user-images.githubusercontent.com/11312418/195552155-504766d8-8520-4986-8c03-4b7f1b3795e9.png">

### Ionic Info

Example code from the Ionic docs

sean-perkins:
@loopezz thanks for reporting this issue! I am able to reproduce the same. 

A fix that I have proposed for v7 appears to resolve this issue as well. Would you mind testing with this dev-build and let me know if you experience any issues?

```bash
6.3.3-dev.11666207854.1e96b6a8
```

I've split out that specific change into a new pull request so that we could resolve this in v6.

loopezz:
@sean-perkins I'm not able to reproduce it anymore. Thanks for the fix. 👍

sean-perkins:
This issue has been resolved and will be available in both the `nightly` build as well as the next patch release `6.3.5`.

Thanks again for reporting this!

PR #26154, opened 2022-10-19
fix(overlays): presenting an overlay does not create nested elements

sean-perkins:
## Pull request checklist

## Pull request type

 

 

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

When overlays are presented without a user defined component, the overlay will continue to create elements at the root of the overlay and append the contents of the teleported component inside of it. This results in scenarios where in implementations like `ion-datetime-button` with an `ion-modal`, the `ion-modal` will create a nested DOM tree of `.ion-page` elements.

Issue URL: #26117

## What is the new behavior?

- Inline overlays do not create nested root nodes from the delegate
- DatetimeButton with a modal does not create nested `.ion-page` elements

- [ ] Yes
- [x] No

## Other information

Dev-build: `6.3.3-dev.11666207854.1e96b6a8`

This is the same fix that we landed on for this feature: https://github.com/ionic-team/ionic-framework/pull/26153, but since that is targeted for v7, it would be great to resolve this specific bug for v6.


sean-perkins:
Should we be reflecting the `presentation` instead of evaluating the property value, like we are doing in this flaky test: https://github.com/ionic-team/ionic-framework/blob/main/core/src/components/datetime-button/test/basic/datetime-button.e2e.ts#L23-L29?

amandaejohnston (approved):

zakariahmusic:
Z world

core/src/components/datetime-button/test/overlays/datetime-button.e2e.ts
```diff
@@ -159,4 +159,24 @@ test.describe('datetime-button: modal', () => {
     await ionModalDidPresent.next();
     expect(datetime).toBeVisible();
   });
+  test('datetime should be the first child of the modal', async ({ page }, testInfo) => {
```
liamdebeasi, line 162: Can we move this test to the `modal` directory? The bug is modal-specific not datetime-specific.

core/src/utils/framework-delegate.ts
```diff
@@ -86,7 +88,7 @@ export const CoreDelegate = () => {
       BaseComponent.appendChild(el);
 
       await new Promise((resolve) => componentOnReady(el, resolve));
-    } else if (BaseComponent.children.length > 0) {
+    } else if (hasUserDefinedComponent && BaseComponent.children.length > 0) {
```
liamdebeasi, line 91: This regresses inline modals in vanilla JS. The `.ion-page` class is no longer added when presenting a modal in http://localhost:3333/src/components/modal/test/inline. (We should probably add a test to check for this)

liamdebeasi (changes requested):

core/src/components/modal/test/inline/modal.e2e.ts
```diff
@@ -21,4 +21,44 @@ test.describe('modal: inline', () => {
...
+    expect(await modal.evaluate((el: HTMLIonModalElement) => el.firstElementChild!.className)).toBe('ion-page');
+
+    await modal.evaluate((el: HTMLIonModalElement) => el.dismiss());
+    await ionModalDidDismiss.next();
+
+    await page.locator('#date-button').click();
+    await ionModalDidPresent.next();
+
+    // Verifies that presenting the overlay again does not create a new host element
+    expect(await modal.evaluate((el: HTMLIonModalElement) => el.firstElementChild!.className)).toBe('ion-page');
```
liamdebeasi, line 59: Are we able to do `/ion-page/`? If so, we should do that otherwise this check will fail if more classes are added. Same thing with the below line. (This is similar to what `toHaveClass` recommends: https://playwright.dev/docs/test-assertions#locator-assertions-to-have-class)

liamdebeasi (approved):
Great job!

PR #26193, opened 2022-10-31
2591 template

liamdebeasi:
## Pull request checklist

## Pull request type

 

 

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes

Issue URL: N/A internal ticket

This is one PR of several for the FW-2591 feature

## What is the new behavior?

- Moved legacy input tests to the `legacy` directory.
- Added a single `basic` test file. These tests will be filled out as functionality is added.
- The legacy input template will be rendered if there is no `label` property usage **and** not `aria-label` usage.
- When the legacy template is rendered, a deprecation warning will be logged with information on how to migrate.
Note: Deprecation messages regarding using `ion-input` inside of `ion-item` or using the shape/fill props on `ion-item` will be handled in a separate PR.

- [ ] Yes
- [ ] No


amandaejohnston (approved):

core/src/utils/forms/form-controller.ts
```diff
@@ -0,0 +1,31 @@
...
+  legacyControl = !hasLabelProp && !hasAriaLabelAttribute;
+
+  const hasLegacyControl = () => {
+    return legacyControl;
+  };
+
+  return { hasLegacyControl };
+};
+
+export type LegacyFormController = {
```
sean-perkins, line 29: I _think_ you could use `ReturnType` here, so you wouldn't need to manually maintain the type.

e.g.:
```ts
export type LegacyFormController = ReturnType<typeof createLegacyFormController>;
```

core/src/utils/forms/form-controller.ts
```diff
@@ -0,0 +1,31 @@
...
+  const controlEl: HTMLElement = el;
+  let legacyControl = true;
+
+  /**
+   * Detect if developers are using the legacy form control syntax
+   * so a deprecation warning is logged. This warning can be disabled
+   * by either using the new `label` property or setting `aria-label`
+   * on the control.
+   */
+  const hasLabelProp = (controlEl as any).label !== undefined;
```
sean-perkins, line 17: Instead of casting to `any` here, should we instead add the additional types we are expecting from our custom element in the `controlEl` declaration? 

e.g.:
```ts
const controlEl: HTMLElement & { label?: string } = el;
```

Alternatively, is there an existing type that already captures this vs. using `HTMLElement`?

core/src/utils/forms/form-controller.ts
liamdebeasi: This gave me a `Type alias 'LegacyFormController' circularly references itself.` error

sean-perkins: Ahh, this is because the function would then control the type, so having this:
```ts
export const createLegacyFormController = (el: HTMLIonInputElement): LegacyFormController => {
```

creates the circular reference, vs:

```ts
export const createLegacyFormController = (el: HTMLIonInputElement) => {
```

I'm not wedded to this change. Just thinking about how to avoid retyping the implementation manually.

sean-perkins (approved):

core/src/utils/forms/form-controller.ts
liamdebeasi: I'll keep the code as-is for now, but if you think of an alternative let me know.

PR #26223, opened 2022-11-03
feat(radio-group): ionChange will only emit from user committed changes

sean-perkins:
## Pull request checklist

## Pull request type

 

 

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes

`ion-radio-group` will emit an `ionChange` event when the value is programmatically assigned.

Issue URL: Internal

## What is the new behavior?

- `ion-radio-group` will only emit ionChange from user interaction, such as a radio child element being tapped or clicked.
- `ion-radio-group` will emit the event that originated the change event firing, if available, in the `ionChange` event detail.

- [x] Yes
- [ ] No

`ionChange` is no longer emitted when the `value` of `ion-radio-group` is modified externally. `ionChange` is only emitted from user committed changes, such as clicking or tapping an `ion-radio` in the group.


core/src/components/radio-group/radio-group.tsx
```diff
@@ -38,8 +38,6 @@ export class RadioGroup implements ComponentInterface {
   @Watch('value')
   valueChanged(value: any | undefined) {
     this.setRadioTabindex(value);
-
-    this.ionChange.emit({ value });
```
amandaejohnston, line 42: I noticed that if you set the `value` of the group programmatically, the selected radio no longer updates. On closer inspection, `ion-radio` is listening for the `ionChange` event in order to handle this: https://github.com/ionic-team/ionic-framework/blob/FW-2033/core/src/components/radio/radio.tsx#L101 We may need to make a separate `ionValueChange` event or something, similar to what `ion-datetime` did.

core/src/components/radio-group/radio-group.tsx
```diff
@@ -139,13 +150,15 @@ export class RadioGroup implements ComponentInterface {
...
           this.value = next.value;
+          this.emitValueChange(ev);
         }
       }
 
       // Update the radio group value when a user presses the
       // space bar on top of a selected radio
       if (['Space'].includes(ev.code)) {
         this.value = this.allowEmptySelection && this.value !== undefined ? undefined : current.value;
+        this.emitValueChange(ev);
```
amandaejohnston, line 161: This causes `ionChange` to fire if you press Space when highlighting the radio that's already selected (when `allowEmptySelection="false"`), thus not changing the value.

core/src/components/radio-group/test/radio-group-events.e2e.ts
```diff
@@ -0,0 +1,104 @@
...
+      </ion-radio-group>
+    `);
+
+    const radioGroup = page.locator('ion-radio-group');
+    const ionChangeSpy = await page.spyOnEvent('ionChange');
+
+    await radioGroup.evaluate((el: HTMLIonRadioGroupElement) => (el.value = '2'));
+
+    expect(ionChangeSpy).toHaveReceivedEventTimes(0);
+    expect(await radioGroup.evaluate((el: HTMLIonRadioGroupElement) => el.value)).toBe('2');
```
amandaejohnston, line 81: With my other comment about programmatic value updates in mind, maybe this should directly check whether the `'2'` radio is selected instead?

amandaejohnston (changes requested):

core/src/components/radio-group/radio-group.tsx
```diff
@@ -51,6 +50,13 @@ export class RadioGroup implements ComponentInterface {
     this.setRadioTabindex(this.value);
   }
 
+  private setRadioChecked = (value: any | undefined) => {
+    const radios = this.getRadios();
+    for (const radio of radios) {
```
liamdebeasi, line 55: If we implemented the `ionValueChange` event that amanda suggested in https://github.com/ionic-team/ionic-framework/pull/26223/files#r1014326964, this should't be needed. We could do something like this in `ion-radio` instead: https://github.com/ionic-team/ionic-framework/blob/feature-7.0/core/src/components/accordion/accordion.tsx#L88

sean-perkins: Yeah I originally spec'd out using `ionValueChange` in the spike, but have been exploring the strategy of adding one-to-many event listeners (`ion-accordion`/`ion-accordion-group`) vs. an API to set the state. 

Do you know of where adding event listeners would perform better vs. a private API method? It seems that adding many listeners (and through our monkey-patched ael for Angular) would be higher memory consumption/bloat.

liamdebeasi: I'm not aware of any benchmarks that we have done in the past regarding this, though my guess would be the performance difference is small. 

Could we keep the architecture consistent for now and then switch in another PR if we find that there's better performance using `setRadioChecked`? That way we can make sure that all components using this architecture get the performance improvements.

It's also worth checking to see if anyone has reported performance issues when checking/unchecking values for components that use the current architecture.

sean-perkins: I'm good with that. I can create a tech debt ticket to investigate this more. 

Looks like one immediate difference is not using event listeners results in the DOM manipulation in the same frame, where using event listeners introduces a delay. This requires us to use things like `page.waitForChanges()` in our tests around it. Probably not noticeable to a user, but small optimizations we can consider in the future.

core/src/components/radio-group/test/radio-group-events.e2e.ts
sean-perkins: Added an explicit test so that this doesn't regress in the future: https://github.com/ionic-team/ionic-framework/pull/26223/files#diff-76824f2e5c0253e7042a5be9519cf1ad96f80c89e4bac7ccd0b59965d0237047

core/src/components/radio-group/radio-group.tsx
liamdebeasi: Ah yeah that's a good point -- the listeners would be async. DX for us would also be a good thing to consider in the tech debt ticket too

core/src/components/radio-group/radio-group.tsx
```diff
@@ -38,15 +38,21 @@ export class RadioGroup implements ComponentInterface {
...
 
   /**
    * Emitted when the value has changed.
    */
   @Event() ionChange!: EventEmitter<RadioGroupChangeEventDetail>;
 
+  /**
+   * Emitted when the `value` property has changed.
+   * This is used to ensure that `ion-radio` can respond
+   * to any value property changes from the group.
```
sean-perkins, line 52: I need to add `@internal` to this.

amandaejohnston (approved):

liamdebeasi (approved):
Great job!

Issue #24716, opened 2022-02-04
bug: inline popover positioned wrong with auto width


EinfachHans:
### Prerequisites

- [X] I have read the [Contributing Guidelines](https://github.com/ionic-team/ionic-framework/blob/main/.github/CONTRIBUTING.md#creating-an-issue).
- [X] I agree to follow the [Code of Conduct](https://ionicframework.com/code-of-conduct).
- [X] I have searched for [existing issues](https://github.com/ionic-team/ionic-framework/issues) that already report this problem, without success.

### Ionic Framework Version

- [ ] v4.x
- [ ] v5.x
- [X] v6.x

### Current Behavior

I have an inline Popover with either `--width: fit-content` or `--width: auto`. This causes the positioning to be wrong and the popover is out of the screen.

### Expected Behavior

The Positioning should be the same if inline or not. Also the arrow is not displayed 8for both), why is that?

### Steps to Reproduce

See linked Repo. The Popovers have the same content but only the inline one is positioned wrong

### Code Reproduction URL

https://github.com/EinfachHans/ionic-popover-issue

### Ionic Info

Ionic:

   Ionic CLI                     : 6.18.1 (/Users/hanskrywalsky/.nvm/versions/node/v14.17.0/lib/node_modules/@ionic/cli)
   Ionic Framework               : @ionic/angular 6.0.5
   @angular-devkit/build-angular : 13.0.4
   @angular-devkit/schematics    : 13.0.4
   @angular/cli                  : 13.0.4
   @ionic/angular-toolkit        : 5.0.3

Capacitor:

   Capacitor CLI      : 3.4.0
   @capacitor/android : not installed
   @capacitor/core    : 3.4.0
   @capacitor/ios     : not installed

Utility:

   cordova-res (update available: 0.15.4) : 0.15.1
   native-run                             : 1.5.0

System:

   NodeJS : v14.17.0 (/Users/hanskrywalsky/.nvm/versions/node/v14.17.0/bin/node)
   npm    : 6.14.13
   OS     : macOS Monterey

sean-perkins:
@EinfachHans thanks for reporting this issue!

I'm seeing the same issues as you, in your example. The `ion-popover` with inline usage, doesn't accurately calculate the positioning. It's possible that the width for `fit-content` cannot be evaluated until after the DOM node is rendered, so we may need to adjust how/when we parse that information. This affects both the `left` positioning, but also the transform-origin position for auto placement. 

[...]

PR #26230, opened 2022-11-04
fix(popover): inline popover positioning with fit-content or auto width

sean-perkins:
## Pull request checklist

## Pull request type

 

 

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

Inline popovers with a width of `auto` or `fit-content` are incorrectly positioned when the popover is presented. 

This bug would apply to _any_ enter animation for overlays that relies on the dimensions of the content inside of the overlay, to position the overlay during the transition.

Issue URL: #24716

## What is the new behavior?

- Inline popovers with a width of `auto` or `fit-content` are positioned correctly when presented.
- `AnimationBuilder` can be resolved asynchronously, to handle animations that require additional async operations

- [ ] Yes
- [x] No

### DIFF
--- a/angular/src/directives/overlays/popover.ts
+++ b/angular/src/directives/overlays/popover.ts
@@ -108,7 +108,7 @@ export class IonPopover {
   constructor(c: ChangeDetectorRef, r: ElementRef, protected z: NgZone) {
     this.el = r.nativeElement;
 
-    this.el.addEventListener('willPresent', () => {
+    this.el.addEventListener('ionMount', () => {
       this.isCmpOpen = true;
       c.detectChanges();
     });
--- a/core/src/components.d.ts
+++ b/core/src/components.d.ts
@@ -5908,7 +5908,7 @@ declare namespace LocalJSX {
          */
         "onDidPresent"?: (event: IonPopoverCustomEvent<void>) => void;
         /**
-          * Emitted before the popover has presented, but after the component has been mounted in the DOM.
+          * Emitted before the popover has presented, but after the component has been mounted in the DOM. This event exists for ion-popover to resolve an issue with the popover and the lazy build, that the transition is unable to get the correct dimensions of the popover with auto sizing. This is not required for other overlays, since the existing overlay transitions are not effected by auto sizing content.
          */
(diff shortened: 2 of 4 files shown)


amandaejohnston (approved):

liamdebeasi (commented):
I tested this on a Samsung Galaxy S3 and the changes worked correctly in the sample application. I did not note any major performance degredation.

However, this is an API change so we need to target a feature branch. Alternatively, is there a way we can pull this logic into `ion-popover` and `ion-modal`?

Something like the following:

```js
async present() {

  /**
   * Check to see if the .ion-page element has not been added
   * You could also check to see if this component is inline
   */
  if (doesNotHaveIonPageElement) {
    /**
     * Fire the ionMount event
     * The framework wrappers will need to listen for ionMount instead of willPresent
     */
    this.ionMount.emit();
    
    /**
     * Wait one raf
     * Alternatively, can use the delegateController to detect when
     * the JS Framework has mounted the inner content? If not, then the raf
     * is probably fine.
     */
    await waitOneFrame();
  }

  // regular present logic
  present();
}
```

This should have the same functionality as your current approach but without any feature changes to the animation library, letting us get this fix into a patch instead of a minor. This code also has a fast track for controller and `keepContentsMounted` popovers which lets it skip the `raf`.

sean-perkins:
With moving this logic to the component-layer, it would introduce an additional `requestAnimationFrame` for custom animations (unless we imposed extra checks to determine if the transition is the built-in one or custom). 

Personally, I would rather target this change for `6.4.0`, knowing that it helps solve animation limitations when the transition is dependent on reading the dimensions of the animating element. There is also a viable workaround in the interim, of using `keepContentsMounted` to avoid this bug, so I do not think we need to fast track a fix.

Thoughts?

liamdebeasi (approved):
I'm fine with either option, but I think it's strange to change the public API of all overlay components in addition to page transitions in order to fix a bug only currently found in `ion-popover`.

If developers are creating a custom popover animation they are still going to need to read the dimensions of the component in order to position the popover correctly, and the alternative implementation fixes the issue for custom animations too. With the original fix, developers would need to update their transition to make use of a `requestAnimationFrame` and a Promise.

Issue #26185, opened 2022-10-27
bug: [React] IonModal is not removed from DOM when parent component unmounts


puopg:
### Prerequisites

- [X] I have read the [Contributing Guidelines](https://github.com/ionic-team/ionic-framework/blob/main/.github/CONTRIBUTING.md#creating-an-issue).
- [X] I agree to follow the [Code of Conduct](https://ionicframework.com/code-of-conduct).
- [X] I have searched for [existing issues](https://github.com/ionic-team/ionic-framework/issues) that already report this problem, without success.

### Ionic Framework Version

- [ ] v4.x
- [ ] v5.x
- [X] v6.x
- [ ] Nightly

### Current Behavior

When a component rendering an IonModal is unmounted by React, the ion modal DOM node is not removed.

### Expected Behavior

Expected similar/same behavior as Ionic v5, where when a React component unmounts, the corresponding components being rendered by that component are also cleaned up. Even if there is no animation, I believe the DOM should be cleaned up

### Steps to Reproduce

1. Have a component (Call it `MyComponent`) that renders an IonModal
2. Conditionally mount that component somewhere.
3. Toggle the modal to present by changing `isOpen` to true.
4. Now, without toggling the modal's `isOpen` prop, simply toggle the mounting condition so React unmounts `MyComponent`.
5. The Modal is not removed from the DOM

### Code Reproduction URL

https://codesandbox.io/s/ionic-6-3-3-modal-unmount-bug-0etb2k?file=/src/pages/Home.tsx
_No response_

### Ionic Info

Ionic React: 6.3.3

### Additional Information

Link to codesandox example:  https://codesandbox.io/s/ionic-6-3-3-modal-unmount-bug-0etb2k?file=/src/pages/Home.tsx

sean-perkins:
@puopg thanks for reporting this issue! 

I believe this may overlap with #25775, but we can track independently in the mean time. 

Now that #25590 has been resolved, I can port over the exploration fix for dismissing any presented overlay when it's react component is unmounted.

puopg:
Ah I see, for some reason I thought https://github.com/ionic-team/ionic-framework/issues/25775 was resolved with your fix and just not marked! Thanks for looking into this!

sean-perkins:
Hello @puopg can you test with this dev-build and let me know if you run into any issues?

```bash
[...]

Issue #25775, opened 2022-08-17
bug: conditionally rendered modal via route doesn't close when navigating away


corysmc:
### Prerequisites

- [X] I have read the [Contributing Guidelines](https://github.com/ionic-team/ionic-framework/blob/main/.github/CONTRIBUTING.md#creating-an-issue).
- [X] I agree to follow the [Code of Conduct](https://ionicframework.com/code-of-conduct).
- [X] I have searched for [existing issues](https://github.com/ionic-team/ionic-framework/issues) that already report this problem, without success.

### Ionic Framework Version

- [ ] v4.x
- [ ] v5.x
- [X] v6.x
- [X] Nightly

### Current Behavior

After updating to ionic 6, Conditionally rendered ion-modals are not dismissing after changing routes. There's a fix in route for an error that happened when the modal was removed from the dom https://github.com/ionic-team/ionic-framework/issues/25590 - but now the modal won't dismiss when navigating away, and the animations aren't happening - and the modal won't dismiss after changing routes.

A route based conditionally rendered IonModal worked in ionic v5 - and I've been told this is a common pattern in a react app, so I would expect this to work out of the box.

## Video:
Here's a video of the simple demo I put together based off of a [fork](https://github.com/corysmc/ionic-react-conference-app) from the ionic-react-conference app
https://user-images.githubusercontent.com/6452188/185195963-06ad49bf-0533-41ba-9f6e-0cb01326eff9.mov

### Expected Behavior

When conditionally rendering an IonModal in React I would expect to be able to show/hide an ion modal based on a matched route. Ideally it would also animate in/out when mounting and unmounting (see video for hack workaround to get this working)

### Steps to Reproduce

1. Install the dev build that fixes the original issue "Failed to execute 'removeChild' on 'Node': The node to be removed is not a child of this node."
2. Add a Modal component that has prop `isOpen={true}`
3. Add a Switch & Route to the component
4. Inside the Modal Component change the route
5. Route to main component ("/support"), click on "New Support Ticket", then click one of the back buttons
6. ⚠️ Notice the modal doesn't close (see video above)

**Step 1**
`yarn add @ionic/react@6.2.3-dev.11660337759.18ea0f7e` and `yarn add @ionic/react-router@6.2.3-dev.11660337759.18ea0f7e` or `npm i @ionic/react@6.2.3-dev.11660337759.18ea0f7e` and `npm i @ionic/react-router@6.2.3-dev.11660337759.18ea0f7e`

[...]

PR #24569, opened 2022-01-12
fix(react): prevent errors when dismissing inline popover after containing element is removed

amandaesmith3:
## Pull request checklist

Please check if your PR fulfills the following requirements:
- [x] Tests for the changes have been added (for bug fixes / features)
- [x] Docs have been reviewed and added / updated if needed (for bug fixes / features)
- [x] Build (`npm run build`) was run locally and any changes were pushed
- [x] Lint (`npm run lint`) has passed locally and any fixes were made for failures

## Pull request type

 

 

Please check the type of change your PR introduces:
- [x] Bugfix
- [ ] Feature
- [ ] Code style update (formatting, renaming)
- [ ] Refactoring (no functional changes, no api changes)
- [ ] Build related changes
- [ ] Documentation content changes
- [ ] Other (please describe): 

## What is the current behavior?

Issue Number: Resolves https://github.com/ionic-team/ionic-framework/issues/24391

Relevant comment: https://github.com/ionic-team/ionic-framework/issues/24391#issuecomment-996883255

In React, if you have an inline popover and delete the element containing it while the popover is open, dismissing the popover causes the app to crash.

## What is the new behavior?

- Check added to make sure the containing element still exists before attempting to move the popover back to it.

## Does this introduce a breaking change?

- [ ] Yes
- [x] No

## Other information

Repo for testing: https://github.com/amandaesmith3/sandbox-ionic-react/tree/FW-434
Dev build: `6.0.3-dev.1642000204.608cbbc`

To reproduce: click the "..." in any item and click "Remove." The popover will be auto-dismissed due to `dismissOnSelect={true}`, triggering the error.

PR #26245, opened 2022-11-08
fix(react): inline overlays dismiss when parent component unmounts

sean-perkins:
## Pull request checklist

## Pull request type

 

 

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

When navigation events occur (push, pop, replace), presented inline overlays are not unmounted from the DOM. Additionally, when a parent component is unmounted, presented overlays are not unmounted. 

Issue URL: #25775, #26185

## What is the new behavior?

- When a parent component is unmounted, any presented overlay within the context of that component will additionally be unmounted.
  - This operation **skips** the dismiss life cycle and does not perform the dismiss transition, emit the `willDismiss` or `didDismiss` events.
- Navigation actions, such as push, pop or replace will unmounted presented overlays.
  - This operation **skips** the dismiss life cycle and does not perform the dismiss transition, emit the `willDismiss` or `didDismiss` events.

- [ ] Yes
- [x] No

## Other information

Dev-build: `6.5.4-dev.11676062329.117ad0d6` (updated 02/10/2023)


packages/react/src/components/createInlineOverlayComponent.tsx
```diff
@@ -105,6 +105,12 @@ export const createInlineOverlayComponent = <PropType, ElementType>(
       attachProps(node, this.props, prevProps);
     }
 
+    componentWillUnmount() {
+      if (this.ref.current) {
+        this.ref.current.dismiss();
```
amandaejohnston, line 110: I'm getting a console error about changing state on an unmounted component when using the Replace or Push operations. I'm guessing `dismiss` is trying to update things after the modal has already been unmounted; maybe it needs to be `await`ed? 🤔 

Test repo/branch I'm using (need to install the dev build): https://github.com/amandaejohnston/sandbox-ionic-react/tree/FW-1889
Screencast: https://user-images.githubusercontent.com/90629384/200869178-3e9a8130-c1b9-42be-84b7-d870bd518887.mp4

amandaejohnston (changes requested):

packages/react/src/components/createInlineOverlayComponent.tsx
liamdebeasi: `componentWillUnmount` is synchronous, so `await` won't do anything here. We might need to add logic here to only update the state if the component is not actively unmounting: https://github.com/ionic-team/ionic-framework/blob/1ea0cd29d93eb29561c016c215e79f182e07ce8b/packages/react/src/components/createInlineOverlayComponent.tsx#L66-L99 (I'm guessing this is where the state update is coming from, but it could be somewhere else too)

sean-perkins: This issue occurs because the `onDidDismiss` fires and the `setState` in the app implementation updates after the component unmounts.

This resolves it in implementations: https://github.com/ionic-team/ionic-framework/pull/26245/files#diff-6c9571ce83a153612a7dcb0a9ced4fb6eaf6f3a613c9e7d4c614136311be6173R14

For a Framework fix, we could look at not handling `onDidDismiss` if the component has already unmounted.

amandaejohnston: If it's possible to handle this in the framework, I would rather go that route, since otherwise this could be considered a breaking change 🤔

sean-perkins:
Updated dev-build: `6.3.6-dev.11668106431.125dbfa9`

amandaejohnston (approved):

packages/react/src/components/createInlineOverlayComponent.tsx
```diff
@@ -107,13 +114,38 @@ export const createInlineOverlayComponent = <PropType, ElementType>(
           this.setState({ isOpen: false });
         }
 
-        this.props.onDidDismiss && this.props.onDidDismiss(evt);
+        const didComponentUnmount = this.ref.current === null;
```
liamdebeasi, line 117: Optional: Would it be better to remove the `didDismiss` and `willPresent` listeners in `componentWillUnmount` instead?

packages/react/src/components/createInlineOverlayComponent.tsx
```diff
@@ -107,13 +114,38 @@ export const createInlineOverlayComponent = <PropType, ElementType>(
...
+      const propsToPass = Object.keys(this.props).reduce((acc, name) => {
+        if (!excludedOverlayEvents.includes(name)) {
+          /**
+           * Maintains all props except for the event handlers
+           * that are manually handled in the componentDidMount.
+           */
+          (acc as any)[name] = (this.props as any)[name];
+        }
+        return acc;
+      }, {});
```
liamdebeasi, line 141: Why is this needed?

packages/react/src/components/createInlineOverlayComponent.tsx
```diff
@@ -107,13 +114,38 @@ export const createInlineOverlayComponent = <PropType, ElementType>(
...
+      attachProps(node, propsToPass, prevProps);
+    }
+
+    componentWillUnmount() {
+      if (this.ref.current) {
+        this.ref.current.dismiss();
+      }
```
liamdebeasi, line 148: If React handles the rendering of the Web Component, won't the `ion-modal` element be removed automatically?

sean-perkins: React will only remove the DOM node if the overlay is dismissed. When the modal is presented, we are manually moving the node to the root and React is unable to know to remove it when the React component unmounts. This check/code moves the node back in time for React to remove it.

packages/react/src/components/createInlineOverlayComponent.tsx
sean-perkins: Investigated this path, results in the previous error:
> react-dom.development.js:67 Warning: Can't perform a React state update on an unmounted component. This is a no-op, but it indicates a memory leak in your application. To fix, cancel all subscriptions and asynchronous tasks in a useEffect cleanup function.

If I remove the `didComponentUnmount` check even when removing the event listeners in `componentWillUnmount`.

packages/react/src/components/createInlineOverlayComponent.tsx
sean-perkins: This change prevents `onDidDismiss` being invoked after the component is unmounted, resulting in the React error:
> Can't perform a React state update on an unmounted component.

There are multiple conditions that cause this, from either us trying to handle the method in the event listener after the component has unmounted (`didComponentUnmount` prevents this) or from the `attachProps` util (which also adds an event listener and can execute after the component has unmounted).

sean-perkins:
Thinking about this approach a little more, if we do not like the filtering out of events to pass to `attachProps`. We could probably come up with a `detachProps` utility that pulls the cached events from the `node.__events` event store and remove the event listener when the component is unmounted.

packages/react/src/components/createInlineOverlayComponent.tsx
```diff
@@ -107,13 +114,38 @@ export const createInlineOverlayComponent = <PropType, ElementType>(
...
-        this.props.onDidDismiss && this.props.onDidDismiss(evt);
+        const didComponentUnmount = this.ref.current === null;
+
+        if (didComponentUnmount === false) {
+          /**
+           * If the component is unmounted, we do not
+           * want to call the onDidDismiss handler to
+           * prevent a React state update on an unmounted component.
+           */
+          this.props.onDidDismiss && this.props.onDidDismiss(evt);
```
liamdebeasi, line 125: Can we add a comment that explains why we don't remove the event listener in the on unmount method?

packages/react/src/components/createInlineOverlayComponent.tsx
```diff
@@ -107,13 +114,38 @@ export const createInlineOverlayComponent = <PropType, ElementType>(
...
+        }
       });
     }
 
     componentDidUpdate(prevProps: IonicReactInternalProps<PropType>) {
       const node = this.ref.current! as HTMLElement;
-      attachProps(node, this.props, prevProps);
+      const propsToPass = Object.keys(this.props).reduce((acc, name) => {
```
liamdebeasi, line 132: Optional: It might be better to iterate through the object keys and build a new object as that will only interate through the props once. The proposed approach will iterate through the props twice: One iteration when calling `Object.keys` and then another iteration when calling `reduce`.

puopg:
Just checking in on this fix, could the team provide an update what is its status?

patdx:
I just tried out the dev build `6.3.6-dev.11668106431.125dbfa9` above, it also gives better behavior for React Error Boundary:

```
const BuggyComponent = () => {
  throw new Error("Failed to render!");
};

const App = () => {
  return (
    <ErrorBoundary>
      <IonModal>
        <BuggyComponent />
      </IonModal>
    </ErrorBoundary>
  );
};
```

In the current prod build I just discovered that code like above freezes the (inactive) modal over the whole screen.

puopg:
@sean-perkins Bump?

sean-perkins:
Open PRs are considered active work. There are a number of factors that effects our prioritization of different work items. 

This specific PR addresses behavior that has not been designed or documented by Ionic and is important that we spent time to thoroughly designing the expected behaviors internally, so we do not need to perform a breaking change to correct that behavior in the future. 

This PR has been updated to remove incorrect assumptions I had made in the first implementation pass, such as trying to perform the dismiss transition for the overlay when the parent component unmounts or a router action is performed. In these scenarios, we should be treating the overlay lifecycle as:

```mermaid
graph TD
A[Overlay]--> Mounted
Mounted --> Presented
Presented --> Unmounted
```

In this diagram we show that the "Dismiss" lifecycle does not occur. This means that a dismiss transition is not performed and the respective events `willDismiss` and `didDismiss` are not emitted. This is intentional to avoid memory leaks and unexpected scenarios where callback events are firing as a result of side-effects. 

Here is an updated dev-build to test with: `6.5.4-dev.11676056919.1a0af097`.

packages/react/src/components/createInlineOverlayComponent.tsx
```diff
@@ -172,6 +151,46 @@ export const createInlineOverlayComponent = <PropType, ElementType>(
     static get displayName() {
       return displayName;
     }
+
+    private handleIonMount = () => {
+      /**
+       * Mount the inner component when the
+       * overlay is about to open.
+       *
+       * For ion-popover, this is when `ionMount` is emitted.
```
liamdebeasi, line 160: What does only the popover emit this? Related to popover positioning?

packages/react/src/components/createInlineOverlayComponent.tsx
```diff
@@ -54,65 +55,43 @@ export const createInlineOverlayComponent = <PropType, ElementType>(
...
-        }
-
-        this.props.onDidDismiss && this.props.onDidDismiss(evt);
-      });
-    }
-
-    componentDidUpdate(prevProps: IonicReactInternalProps<PropType>) {
-      const node = this.ref.current! as HTMLElement;
-      attachProps(node, this.props, prevProps);
+        node.removeEventListener('didDismiss', this.handleDidDismiss);
```
liamdebeasi, line 91: Why do we only remove the `didDismiss` event?

packages/react/src/components/createInlineOverlayComponent.tsx
```diff
@@ -54,65 +55,43 @@ export const createInlineOverlayComponent = <PropType, ElementType>(
...
+        node.remove();
```
liamdebeasi, line 92: Why do we need to manually remove the DOM node? Shouldn't React handle this for us?

packages/react/src/components/createInlineOverlayComponent.tsx
sean-perkins: Correct, we ran into an issue with popover implementations with auto widths: https://github.com/ionic-team/ionic-framework/pull/26274.

packages/react/src/components/createInlineOverlayComponent.tsx
sean-perkins: The `createInlineOverlayComponent` binds two internal event listeners for `didDismiss` and `willPresent`. Both implementations have an internal state update for:

```ts
this.setState({ isOpen: false }); // `true` for willPresent
```

At this point in the lifecycle of the component the component is unmounting. `willPresent` cannot fire. However, `didDismiss` can fire and cause any user-implemented callback handlers for `onDidDismiss` to run after the React component has unmounted. If this occurs, React will throw an exception in the console about a memory leak. 

This code disconnects the internal event listener manually, before the element can dispatch `didDismiss`.

packages/react/src/components/createInlineOverlayComponent.tsx
sean-perkins: At this point in the execution/logic, the overlay is presented and relocated to the root of the DOM. It is outside of the context of React's VDOM to unmount. 

We could try teleporting the node back to have React also remove it, but we would need to also repaint the overlay as dismissed, otherwise it would cause visual overlapping issues with moving a presented overlay back into the nested DOM tree. 

I opted for `node.remove()` since it cleans up the presented overlay and doesn't require any style recalculations or visual changes. We unfortunately are dealing with a limited execution window with `componentWillUnmount`, so we cannot do a blocking task like `await node.dismiss()`.

packages/react/src/components/createInlineOverlayComponent.tsx
```diff
@@ -172,6 +151,46 @@ export const createInlineOverlayComponent = <PropType, ElementType>(
...
+      this.props.onWillPresent && this.props.onWillPresent(evt);
+    };
+
+    private handleDidDismiss = (evt: any) => {
+      const wrapper = this.wrapperRef.current;
+      const el = this.ref.current;
+
+      /**
+       * This component might be unmounted already, if the containing
+       * element was removed while the popover was still open. (For
```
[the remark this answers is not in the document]
sean-perkins: This comment was just relocated in this commit, prior it came from: https://github.com/ionic-team/ionic-framework/pull/24569

@amandaejohnston do you remember if this is specific to popover or applies to all overlays?

packages/react/src/components/createInlineOverlayComponent.tsx
```diff
@@ -54,65 +55,43 @@ export const createInlineOverlayComponent = <PropType, ElementType>(
...
-        const wrapper = this.wrapperRef.current;
-        const el = this.ref.current;
-
+      if (node && this.state.isOpen) {
         /**
-         * This component might be unmounted already, if the containing
-         * element was removed while the popover was still open. (For
-         * example, if an item contains an inline popover with a button
-         * that removes the item.)
+         * Detach the local event listener than performs the state updates,
```
amandaejohnston, line 86:
```suggestion
         * Detach the local event listener that performs the state updates,
```

packages/react/src/components/createInlineOverlayComponent.tsx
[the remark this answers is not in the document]
amandaejohnston: In theory this should apply to all overlays; my comment was overly specific 😆

packages/react/src/components/createInlineOverlayComponent.tsx
amandaejohnston: Would `detachProps` handle removing that listener? If so, would it make sense to instead call `detachProps` before `node.remove` and then remove this line?

sean-perkins: `detachProps` only removes event listeners added by `attachProps` (because we track the events on the element node on a key called `__events`). This specific event listener is manually added within the React class and isn't removed as a result of `detachProps`.

packages/react/src/components/createInlineOverlayComponent.tsx
liamdebeasi: oh that's right. thanks!

PR #26252, opened 2022-11-09
fix(input): floating label works correctly with long text and notch

liamdebeasi:
## Pull request checklist

## Pull request type

 

 

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

During testing, I noticed that my implemented of the notch in the MD outline was incorrect, causing the gap to be much wider than necessary when using longer labels.

I also noticed that long label texts do not flow correctly in the container.

## What is the new behavior?

- Long labels are now correctly truncated so they do not break the input layout
- Revamped how we create the notch in the MD outline style so the gap is correctly sized.
- Added tests

- [ ] Yes
- [x] No


core/src/components/input/input.vars.scss
```diff
@@ -1 +1,4 @@
 @import "../../themes/ionic.globals";
+
+/// @prop - How much to scale the floating label by
+$input-floating-label-scale:           0.75 !default;
```
sean-perkins, line 4: ~ We discussed in one of our alignments we were going to do away with manually maintaining the spacing in SASS files. 

Should this be:
```
$input-floating-label-scale: 0.75 !default;
```

liamdebeasi: Good catch, fixed

sean-perkins (approved):

core/src/components/input/input.scss
```diff
@@ -384,6 +377,16 @@
...
 
+  /**
+   * Label text should not extend
+   * beyond the bounds of the input.
+   * However, we do not set the max
+   * width to 100% because then
+   * only the label would show and users
+   * would not be able to see what they are typing.
+   */
+  max-width: 200px;
```
amandaejohnston, line 388: Does native do this? I feel like this should be an implementation detail; if the app has a really long label, that's on the dev. This feels like `fixed` label behavior sneaking in where it might not be wanted.

core/src/components/input/input.tsx
```diff
@@ -590,10 +589,14 @@ export class Input implements ComponentInterface {
       return [
         <div class="input-outline-container">
           <div class="input-outline-start"></div>
-          <div class="input-outline-notch">{needsNotch && this.renderLabel()}</div>
+          <div class="input-outline-notch">
```
amandaejohnston, line 592: Why do we now render the notch for all label placements, rather than just the ones that need it?

Also, if this is definitely needed, the comment above here should be updated since it still mentions only rendering the notch for certain label placements.

core/src/components/input/test/label-placement/input.e2e.ts
```diff
@@ -93,4 +121,34 @@ test.describe('input: label placement floating', () => {
       `input-focused-placement-floating-no-value-${page.getSnapshotSettings()}.png`
     );
   });
+  test('long label should truncate', async ({ page }) => {
+    await page.setContent(`
+      <ion-input label="Email Email Email Email Email Email Email Email Email Email Email Email" value="example@ionic.io" label-placement="floating"></ion-input>
+    `);
+    const input = page.locator('ion-input');
+    expect(await input.screenshot()).toMatchSnapshot(
+      `input-placement-floating-long-label-${page.getSnapshotSettings()}.png`
```
amandaejohnston, line 130: I noticed that only some of these screenshots show the label as truncated; looks like it's a matter of ios vs md? Is the behavior supposed to be inconsistent like that?

core/src/components/input/input.tsx
liamdebeasi: The way the outline works with `fill="outline"` is that it creates 3 elements: The start, the notch (middle), and the end. Each piece has an outline which gives the appearance of a single continuous outline. I previously rendered the label inside of the notch so that the "cut out" with stacked/floating labels appears, but that caused the notch to be the wrong size.

The middle piece has always been needed with `fill="outline"` for all labels, but creating the cut out by removing the top border is only used by stacked/floating labels.

The 3 pieces are only rendered if `fill==="outline"` on MD mode (see `hasOutlineFill` above)

core/src/components/input/test/label-placement/input.e2e.ts
liamdebeasi: Whoops I had some styles only applied to iOS mode. I fixed this in https://github.com/ionic-team/ionic-framework/pull/26252/commits/9ee347212134307ff01a5e73de1ad1298ab39d0b and will update screenshots now.

core/src/components/input/input.scss
amandaejohnston: Aha, works for me 👍

PR #26260, opened 2022-11-10
chore(ci): run against stencil evaluation build

rwaskiewicz:
## Pull request checklist

## Pull request type

 

 

- [ ] Bugfix
- [ ] Feature
- [x] Build related changes

Today, there isn't a programmatic test that what's in the `HEAD` of
Stencil's primary branch works in the Ionic Framework. We manually
test this to verify changes on the Stencil team. While we can't eliminate
manual testing altogether, it would be a nice to have to have some
programmatic checks.

Issue URL:  N/A

## What is the new behavior?

this pr introduces the ability for the ionic framework to run it's
continuous integration (ci) suite using the latest commits in the
[stencil repository](https://github.com/ionic-team/stencil). the motivation
behind this pr is the ability for both the ionic framework and stencil
teams to have better insight as to whether a new version of stencil
will work correctly in the ionic framework. the intent here is that any
regressions can be caught and rectified by the appropriate team prior
to a release of either piece of software.

the stencil repository currently does not have the ability to publish
a "developer" release, or any other type of release that might otherwise
be considered for a "canary" or "nightly" release. the stencil team is
working towards this goal. however, we believe both teams can benefit
from this type of testing in the mean time. as such, we temporarily build
stencil from source and install the generated tarball in ci. this allows us
to begin to work through the operational issues related to having this
new integration without being reliant on a new publishing pipeline.

the job is introduced as a cron job, that runs in the early hours of the
morning. it closely resembles this repo's build step that runs on every
pr that's opened - with the difference of pulling down the stencil repo,
building the project, and installing the tarball

- [ ] Yes
- [x] No

## Testing Info

https://github.com/ionic-team/ionic-framework/pull/26260/commits/7093db3c0a57a178109e29571b88eabd03205768 is a special commit that pretends that this workflow is meant to run on new pull requests. That's not going to be the case, but allows us to test the action and verify that its green/works. The latest run of it can be [found here](https://github.com/ionic-team/ionic-framework/actions/runs/3471357595/jobs/5800880475).

I am going to revert that commit, which should be the `HEAD` of this PR and makes this a cron job.
### A Note on Naming + Conventions Used

Generally speaking, I tried to follow naming conventions/folder structure that were already in the repo & aligned with how GitHub Actions work (with regard to their conventions for action naming). That said, anything from a naming perspective/structure is up for comment/change. Just let me know

### Additional/Future Improvements

The following are changes that I would like to make in future/additional PRs, and not include in this one. Again, up for further discussion:

1. Using Stencil Nightly

    Once Stencil is publishing a nightly release, the steps where we pull down the Stencil repo & build the project from source can be removed. The installation can then be sourced from the NPM registry, rather than a locally built tarball.

2. Notifications

    One or both teams ought to be notified when this workflow fails. To be discussed between the teams, but can happen offline.

3. Expansion

   This workflow should be expanded to have the ability to run not only against one artifact, but many. e.g. We should be able to run with the v2.x.nightly, and v3.x.beta


.github/workflows/actions/build-core-stencil-eval/action.yml
```diff
@@ -0,0 +1,50 @@
...
+      run: npm ci && npm run build && npm pack
+      working-directory: ./stencil-repo-tmp
+      shell: bash
+    - name: Move the Stencil Build Artifact
+      # there isn't a great way to get the output of `npm pack`, just grab the first and hope for the best
+      run: mv $(ls ./stencil-repo-tmp/*.tgz | head -1) ./core/stencil-eval.tgz
+      shell: bash
+    - name: Install Stencil Eval
+      working-directory: ./core
+      run: npm i ./stencil-eval.tgz
```
liamdebeasi, line 32: Are you able to do `npm install ./stencil-repo-tmp/stencil-eval.tgz` instead to avoid the extra `mv` step?

rwaskiewicz: We cannot, as the result of `npm pack` is something like `stencil-core-2.19.3.tgz`, where the version is dictated by `package.json#version`. That field will be changing over time, so we don't know it's value (the `mv` is also acting as a rename).

There _are_ ways of finding the version from `package.json` with `grep | head | awk`, but IMO that overcomplicated the pipeline a little bit (although not by much). Since we intend on getting rid of building Stencil from source (including this `mv`), I decided to keep it as is.

rwaskiewicz: Actually...I'd prefer to keep it as is. That way we can more safely rimraf the entire Stencil directory we created

liamdebeasi (approved):

PR #26265, opened 2022-11-10
refactor(ci): do not cache node modules

liamdebeasi:
## Pull request checklist

## Pull request type

 

 

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes

Caching `node_modules` on CI is an anti-pattern: https://github.com/actions/cache/blob/main/examples.md#node---npm

The longstanding recommendation for Ionic is to not install `ionicons` as a dependency in your app and instead rely on the version of `ionicons` that ships with Ionic. However, this is a bad practice as it results in ghost dependencies and issues with Yarn. As a result, we cached `node_modules` so that the correct installation of `ionicons` could be referenced when running CI jobs and running the E2E test apps.

## What is the new behavior?

- Ionicons is now a direct dependency of each Ionic package that needs it
- Ionicons is now a direct dependency of each E2E test app
- GitHub Action scripts no longer cache any `node_modules`.
- Ionic Angular no longer references the local `@ionic/core` when building. Developers that need to reference the local copy of `@ionic/core` should use `npm run sync` from the `angular` directory.
- React Router test app now uses Ionicons v6.

Note: As a result, certain jobs now re-install `node_modules` such as the `test-core-lint` step. These are non-blocking tasks so it should not impact overall CI run performance.

- [ ] Yes
- [x] No


.github/workflows/actions/test-core-lint/action.yml
```diff
@@ -6,14 +6,12 @@ runs:
...
-      uses: actions/cache@v3
-      env:
-        cache-name: core-node-modules
-      with:
-        path: ./core/node_modules
-        key: ${{ runner.os }}-build-${{ env.cache-name }}-${{ hashFiles('./core/package-lock.json') }}-v2
+    - name: Install Dependencies
+      # Do not create a new package-lock file
+      # as it will cause the git diff check to fail.
+      run: npm install --no-save
```
sean-perkins, line 12: Can we use `npm ci` here or is there a reason we need to install the latest deps?

packages/react-router/test-app/package.json
```diff
@@ -15,7 +15,7 @@
     "@types/react-dom": "^16.9.4",
     "@types/react-router": "^5.1.8",
     "@types/react-router-dom": "^5.1.7",
-    "ionicons": "^5.4.0",
+    "ionicons": "^5.5.4",
```
sean-perkins, line 18: Can get a tech debt ticket for this? The test app should probably be on v6 for everything.

.github/workflows/actions/test-core-lint/action.yml
liamdebeasi: Nope, we can use `npm ci`.

packages/react-router/test-app/package.json
liamdebeasi: We aren't affected by the Ionicons v6 breaking changes, so I was able to update this in: https://github.com/ionic-team/ionic-framework/pull/26265/commits/84cbcae0b694e24751f66f3fbdc5da182fe4b0cf

rwaskiewicz (approved):
Thanks for doing this!

angular/scripts/sync.sh
```diff
@@ -0,0 +1,5 @@
+# Copy core dist
```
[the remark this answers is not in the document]
rwaskiewicz: @liamdebeasi FYI - this could probably be propagated to the other `sync.sh` scripts. I only added one here as an example in case you didn't want to do it (I didn't want to spam you with the same suggestion 🤣)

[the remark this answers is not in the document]
liamdebeasi: Thx! Fixed in https://github.com/ionic-team/ionic-framework/pull/26265/commits/efad7e0d5b122fa79c1343ff5163b8149f7796fb

sean-perkins (approved):

PR #26266, opened 2022-11-10
chore(many): add tech debt tickets and remove unnecessary todos

amandaejohnston:
## Pull request checklist

## Pull request type

 

 

- [ ] Bugfix
- [ ] Feature
- [x] Code style update (formatting, renaming)
- [ ] Build related changes

## What is the new behavior?

- Removes a bunch of `TODO`s that were added a long time ago, don't seem necessary anymore, and/or have no context. Some are also likely to be handled in other tickets, such as notes to add types when we have FW-2690.
- Adds tech debt tickets to `TODO`s that were missing them and seem useful to keep around.

Let me know if any of these should be kept and if you have context I could add when creating a ticket.

- [ ] Yes
- [x] No


packages/vue-router/src/router.ts
```diff
@@ -104,7 +103,6 @@ export const createIonRouter = (opts: IonicVueRouterOptions, router: Router) =>
   });
 
   const handleNavigateBack = (defaultHref?: string, routerAnimation?: AnimationBuilder) => {
-    // todo grab default back button href from config
```
liamdebeasi, line 107: I noticed that Angular and React both get `backButtonDefaultHref` from the config (https://github.com/ionic-team/ionic-framework/blob/0a8a958cba02994ea0fefa265b17edddffe62e98/angular/src/directives/navigation/ion-back-button.ts#L30). However, it looks like we update `defaultHref` on `ion-back-button` with `backButtonDefaultHref`: https://github.com/ionic-team/ionic-framework/blob/0a8a958cba02994ea0fefa265b17edddffe62e98/core/src/components/back-button/back-button.tsx#L76

Are you able to check if this is needed in Ionic Vue? You can set `backButtonDefaultHref` in the list starter app and make sure that the back button shows up on the list page when no `default-href` is set. Just want to make sure we aren't missing something important here. Maybe not needed since we render the actual Web Component?


Somewhat related: Looks like `backButtonDefaultHref` is not documented: https://ionicframework.com/docs/angular/config

liamdebeasi (approved):
One question otherwise this looks good. Thanks for doing this!

packages/vue-router/src/router.ts
amandaejohnston: Vue does indeed operate as expected with `backButtonDefaultHref` set 👍 The back-button stays visible, even after refreshing the page.

I'll see about getting `backButtonDefaultHref` added to the docs in a separate PR.

liamdebeasi: Thanks for checking!

packages/vue/src/components/IonRouterOutlet.ts
```diff
@@ -198,7 +196,7 @@ export const IonRouterOutlet = /*@__PURE__*/ defineComponent({
     const transition = (
       enteringEl: HTMLElement,
       leavingEl: HTMLElement,
-      direction: any, // TODO types
+      direction: any,
```
sean-perkins, line 199: If we are removing the todo statement for types, should we add the type here or create a new ticket to track?

packages/vue-router/src/types.ts
```diff
@@ -63,7 +63,7 @@ export interface ViewItem {
   outletId: number;
   matchedRoute: RouteLocationMatched;
   ionPageElement?: HTMLElement;
-  vueComponent: any; // todo
+  vueComponent: any;
```
sean-perkins, line 66: Similar question here regarding removing the todo for something with an `any` type.

packages/vue/src/components/IonRouterOutlet.ts
amandaejohnston: I've got FW-2690 in this sprint, which is to revisit `any` types across the codebase. I figure I can handle it then.

sean-perkins (approved):

Issue #25337, opened 2022-05-23
bug: popover alignment property renders incorrect location in frameworks


amandaejohnston:
### Prerequisites

- [X] I have read the [Contributing Guidelines](https://github.com/ionic-team/ionic-framework/blob/main/.github/CONTRIBUTING.md#creating-an-issue).
- [X] I agree to follow the [Code of Conduct](https://ionicframework.com/code-of-conduct).
- [X] I have searched for [existing issues](https://github.com/ionic-team/ionic-framework/issues) that already report this problem, without success.

### Ionic Framework Version

- [ ] v4.x
- [ ] v5.x
- [X] v6.x
- [ ] Nightly

### Current Behavior

When using various combinations of `side` and `alignment` on an inline `ion-popover`, the popover renders at the incorrect location when opened. This occurs in all three framework integrations (Angular/React/Vue), but *not* in core.

The linked repo has buttons for all combinations of `side` and `alignment`; here are the specific issues I noticed:

- `side="top"` shows the popover too low, overlapping the button. This occurs with all values of `alignment`, including when it isn't set at all.
![side top, alignment start](https://i.gyazo.com/334311600aca643d5d29b38e1f19186e.png)
- For `side`s of `right`, `left`, `start`, and `end`, and an `alignment` of `center`, the popover is rendered a pixel or two lower than the top edge of the button, when it should be vertically centered.
![side left, alignment center](https://i.gyazo.com/7a63d7eddb21bad14054f3646933cea8.png)
- For the same `side`s as the above, and an `alignment` of `end`, the popover is also rendered a few pixels lower than the button's top edge (slightly lower than with `alignment="center"`), when it should be raised higher so the bottom edges align.
![side left, alignment end](https://i.gyazo.com/aae73fd5c94e88cdd2bf41d3d2e5a0b3.png)

### Expected Behavior

Here are corresponding screenshots from the `position` test in core, which shows the correct behavior:

![side top, alignment start](https://i.gyazo.com/ce60599f3c46b15df5a9d01446225910.png)
![side left, alignment center](https://i.gyazo.com/e46891049da934763970087264432f4c.png)
![side left, alignment end](https://i.gyazo.com/5180e0641073e778048a0ac8e357820b.png)

### Steps to Reproduce

1. Clone repo (note the branch), `npm install`, `ionic serve`
2. Click buttons corresponding to described side/alignment values (headers list side, then alignment)

### Code Reproduction URL

[...]

PR #26306, opened 2022-11-16
fix(popover): popover positions correctly on all frameworks

sean-perkins:
## Pull request checklist

## Pull request type

 

 

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

In framework implementations, the popover will incorrectly position when using an alignment property such as `top` or `bottom`. The delay in timing between when the popover contents are mounted in the DOM vs. when the transition starts is too low, resulting in the popover content's bounding box height evaluating to `0` and the popover positioning incorrectly.

Issue URL: #25337

## What is the new behavior?

- React and Vue framework implementations use the new `ionMount` internal event when the popover is presented instead of `willPresent`. This allows the framework to mount the popover content earlier in the lifecycle of an overlay being presented.
- Uses two request animation frame loops to provide enough time for the framework to mount the popover contents. In Angular/React a single request animation frame appears to be enough time, but for Vue, two is required. Opted for two to be consistent. 

- [ ] Yes
- [x] No

## Other information

Dev-build: `6.3.8-dev.11668643678.159ccab8`


packages/react/src/components/createInlineOverlayComponent.tsx
```diff
@@ -63,7 +63,7 @@ export const createInlineOverlayComponent = <PropType, ElementType>(
        * cause the event handlers to be
        * destroyed and re-created.
        */
-      this.ref.current?.addEventListener('willPresent', (evt: any) => {
+      this.ref.current?.addEventListener('ionMount', (evt: any) => {
```
sean-perkins, line 66: I need to apply this change to only popover (for both React and Vue).

packages/react/src/components/createInlineOverlayComponent.tsx
```diff
@@ -55,6 +55,17 @@ export const createInlineOverlayComponent = <PropType, ElementType>(
...
 
+      /**
+       * Mount the inner component when the
+       * overlay is about to open.
+       *
+       * For ion-popover, this is when `ionMount` is emitted.
+       * For other overlays, this is when `willPresent` is emitted.
+       */
+      this.ref.current?.addEventListener('ionMount', () => {
+        this.setState({ isOpen: true });
```
amandaejohnston, line 66: Does this also need to call `onWillPresent` as in the `willPresent` listener below?

sean-perkins: No, when `willPresent` is emitted from the overlay, it will invoke that handler. The introduction of `ionMount` does not remove `willPresent` being emitted. 

The objective of `ionMount` is to mount the overlay contents sooner in the present lifecycle, so that transitions dependent on the bounding box are able to calculate dimensions correctly. All of the existing plumbing remains the same.

amandaejohnston: Makes sense, thank you!

core/src/components/popover/popover.tsx
```diff
@@ -464,40 +459,48 @@ export class Popover implements ComponentInterface, PopoverInterface {
...
+       * This allows the framework implementations enough time to mount
+       * the popover contents, so the bounding box is set when the popover
+       * transition starts.
+       *
+       * On Angular and React, a single raf is enough time, but for Vue
+       * we need to wait two rafs. As a result we are using two rafs for
+       * all frameworks to ensure the popover is presented correctly.
+       */
+      raf(() => {
+        raf(async () => {
```
liamdebeasi, line 476: Why do we need to wait 2 frames just in Ionic Vue?

sean-perkins: Short answer, I'm not sure. I tested with a single request animation frame delay and compared against each and only Vue did not render the inner contents in time for the transition. Bumping to two consistently rendered correctly. I'm not familiar to the internals of how Vue handles rendering. 

My intention with this bug fix is looking at a target for a patch release, but realizing we should be able to rip most/all of this away in a minor or major release, if we change either or both:
1. The animation library to support asynchronous behavior.
2. Add support to overlays to accurately track when content is mounted before running the transition.

liamdebeasi (approved):
I'm fine with the changes since they resolve some critical popover functionality. Can we add a tech debt ticket to research alternative implementations that do not rely on dual rafs?

PR #26316, opened 2022-11-17
feat(alert): add ability to use alert inline

amandaejohnston:
## Pull request checklist

## Pull request type

 

 

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes

Issue URL: 

## What is the new behavior?

-
-
-

- [ ] Yes
- [ ] No

### DIFF
--- a/core/src/components/alert/alert.scss
+++ b/core/src/components/alert/alert.scss
@@ -52,6 +52,10 @@
   align-items: flex-start;
 }
 
+:host ion-backdrop {
+  z-index: 0;
+}
+
 .alert-wrapper {
   display: flex;
 
--- a/core/api.txt
+++ b/core/api.txt
@@ -80,20 +80,26 @@ ion-alert,prop,enterAnimation,((baseEl: any, opts?: any) => Animation) | undefin
 ion-alert,prop,header,string | undefined,undefined,false,false
 ion-alert,prop,htmlAttributes,undefined | { [key: string]: any; },undefined,false,false
 ion-alert,prop,inputs,AlertInput[],[],false,false
+ion-alert,prop,isOpen,boolean,false,false,false
 ion-alert,prop,keyboardClose,boolean,true,false,false
 ion-alert,prop,leaveAnimation,((baseEl: any, opts?: any) => Animation) | undefined,undefined,false,false
 ion-alert,prop,message,IonicSafeString | string | undefined,undefined,false,false
 ion-alert,prop,mode,"ios" | "md",undefined,false,false
 ion-alert,prop,subHeader,string | undefined,undefined,false,false
 ion-alert,prop,translucent,boolean,false,false,false
+ion-alert,prop,trigger,string | undefined,undefined,false,false
 ion-alert,method,dismiss,dismiss(data?: any, role?: string) => Promise<boolean>
 ion-alert,method,onDidDismiss,onDidDismiss<T = any>() => Promise<OverlayEventDetail<T>>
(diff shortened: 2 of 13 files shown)


core/src/components/alert/alert.scss
```diff
@@ -52,6 +52,10 @@
   align-items: flex-start;
 }
 
+:host ion-backdrop {
+  z-index: 0;
+}
```
amandaejohnston, line 57: This was added because I was running into an issue where inline alerts would show the backdrop on top of the overlay. I figure the backdrop's default `z-index` of 2 isn't needed since the overlay wrapper around it gets its own massive z-index anyway.

core/src/components/alert/alert.scss
```diff
@@ -54,6 +54,7 @@
 
 .alert-wrapper {
   display: flex;
+  position: relative;
```
amandaejohnston, line 57: This was added to resolve an issue where inline alerts show underneath the backdrop, due to the added `ion-delegate-host` wrapper messing up the stacking.

core/src/components/alert/test/a11y/alert.e2e.ts
```diff
@@ -41,7 +41,12 @@ test.describe('alert: a11y', () => {
 
     await didPresent.next();
 
-    const results = await new AxeBuilder({ page }).analyze();
+    /**
+     * alert overlays the entire screen, so
+     * Axe will be unable to verify color contrast
+     * on elements under the backdrop.
+     */
+    const results = await new AxeBuilder({ page }).disableRules('color-contrast').analyze();
```
amandaejohnston, line 49: Adding `position: relative` to the alert seems to have caused the color contrast test to start failing and/or exacerbated an existing issue, but I can't see any difference visually, so it seems to be an Axe detection issue rather than an actual contrast problem.

core/src/components/alert/alert.scss
liamdebeasi: Hmm this looks like a bug. `ion-delegate-host` should only be added to `ion-modal` and `ion-popover` since they accept user components.

liamdebeasi (approved):
The `position: relative` change is concerning, but it looks like this is due to the `ion-delegate-host` bug. Once that is resolved we should be able to remove `position: relative`. Looks great otherwise!

core/src/components/alert/test/a11y/alert.e2e.ts
liamdebeasi: Since there are layered elements with opacity (such as the backdrop), the `position: relative` with `z-index: 10` likely creates a new stacking context which can cause Axe to be unable to verify the color contract.

core/src/components/alert/alert.scss
sean-perkins: `ion-delegate-host` doesn't have any style responsibility. It was just used to track the existing wrapper element that was created, so we didn't create another one. e.g. During the PR at one point I was setting a private flag vs. adding a class. We landed on a class just for simplicity and to avoid someone accidentally setting that flag. 

I would need more information to what caused this style being introduced.

liamdebeasi (approved):
Great job!

amandaejohnston:
After talking on Slack, I've patched the `ion-delegate-host` element to only be added for modals and popovers, and reverted the patches to the alert CSS and a11y test. I also added a dev build to the bottom of the PR description. 👍

sean-perkins (approved):

PR #26441, opened 2022-12-07
feat(range): add modern markup and label

liamdebeasi:
## Pull request checklist

## Pull request type

 

 

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes

The new range markup does not support labels.

## What is the new behavior?

- Added support for labelPlacement
- Fixed padding/margin issues when using the modern markup in an ion-item
- Fixed disabled state rendering with modern markup
- Fixed `emitStyle` so it now only renders with the legacy markup
- Added modern and legacy range templates
- Lots of tests! In particular, I broke the single `basic` screenshot test out into several screenshot tests. You can find this in the `basic`, `color`, `custom`, and `states` directories.
- Added TODOs for when we remove the legacy markup.

A few things: 

- Of the 1600 lines, only ~400 lines is actual new  behavior (with the rest being tests), but let me know if you'd like the tests added in a separate PR.
- The a11y test skips the "color-contrast" check. This is due to a confirmed bug in Axe where it does not account for <slot> elements: https://github.com/ionic-team/ionic-framework/pull/26441/files#diff-48b749fa43ddb0e43e20780d88109fc96020ad00414f77b9d631f20ecef60f75R19
- There is a lingering iOS bug with the knob: The edge of the knob should never extend past the edge of the track on iOS, but it does in Ionic. This can cause spacing between the range and the label text to look a bit odd. I did not address it in this PR since it is a bug in `main` too. However, we should prioritize a fix for this because the bug is a bit more noticeable now that the other ion-item alignment issues have been resolved.

- [ ] Yes
- [x] No


amandaejohnston (approved):

core/src/components/range/range.tsx
```diff
@@ -665,79 +741,61 @@ Developers can dismiss this warning by removing their usage of the "legacy" prop
       }
     }
 
-    renderHiddenInput(true, el, this.name, JSON.stringify(this.getValue()), disabled);
+    let labelledBy;
```
sean-perkins, line 744:
```suggestion
    let labelledBy: string;
```
A variable without a type will inherently be `any`.

core/src/components/range/range.scss
```diff
@@ -172,3 +173,140 @@
...
+  /**
+   * The margin between the label and
+   * the input should be on the end
+   * when the label sits at the start.
+   */
+  @include margin(0, var(--margin), 0, 0);
+}
+
+/**
+ * Label is on the left of the input in LTR and
```
sean-perkins, line 303: There are a few instances of comments referring to the "input" here instead or range. Should we update those as well?

sean-perkins (approved):

PR #26471, opened 2022-12-12
feat(checkbox): fix rendering in ion-item with new syntax

amandaejohnston:
## Pull request checklist

## Pull request type

 

 

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

## What is the new behavior?

- `ionStyle` event now only emits with legacy checkbox to avoid styling issues when inside an item.
- Tests added for rendering a checkbox with the new syntax in an `ion-item`.

- [ ] Yes
- [x] No


liamdebeasi (approved):
Changes look good. Will the ripple effect be added in a separate PR?

amandaejohnston:
Aha, didn't notice that had broken 👀 I've added it in here.

core/src/components/checkbox/checkbox.tsx
```diff
@@ -245,6 +259,7 @@ export class Checkbox implements ComponentInterface {
             onBlur={() => this.onBlur()}
             ref={(focusEl) => (this.focusEl = focusEl)}
           />
+          {mode === 'md' && inItem && <ion-ripple-effect></ion-ripple-effect>}
```
liamdebeasi, line 262: IMO we should always render this in MD mode that way developers can use the ripple effect outside of `ion-item` if they want to

sean-perkins (approved):
This looks good to me. The failing test is flaky, which I've ran into as well:
>   1) [Mobile Safari] › src/utils/animation/test/multiple/animation.e2e.ts:16:3 › animation: multiple › should resolve grouped animations using css animations 

    Error: expect(received).toEqual(expected) // deep equality

    Expected: "AnimationCSubAFinished"
    Received: "AnimationCFinished"

We may want to just `test.skip` if re-running the action doesn't pass.

liamdebeasi (changes requested):
I'm noticing that the item's hover/focused/active effects are also applying: 
image (notice a lighter shade of gray on the left where the item is)

You likely need to update the `emitStyle` function to only run on a legacy checkbox

liamdebeasi (changes requested):
Also the checkbox does not take up the container's height in an item, so it looks misaligned: 
image


for context, this is what I did with toggle: https://github.com/ionic-team/ionic-framework/blob/16404ab12f4ef9095e57fbf31d7036ab7799b9f4/core/src/components/toggle/toggle.scss#L41-L44

edit: Discussed offline. I cannot reproduce this anymore.

core/src/components/checkbox/checkbox.tsx
amandaejohnston: Checkboxes don't currently have a ripple effect with the legacy syntax, so I'm hesitant to change that. The [MD spec](https://m2.material.io/develop/web/components/input-controls/checkboxes) does show some different-looking hover/active/focus states, with the circle wrapping around the checkbox, but we don't have any of that in `main` either so making the change feels out of scope here.

liamdebeasi (changes requested):
It looks like the hover/focus/activated effects are only applying on `ion-item` and not `ion-checkbox`.

It looks like you need to update the `hasCover` function in `ion-item` to only consider `ion-checkbox.legacy-checkbox`

liamdebeasi:
Oh I think the same thing needs to be done for `setMultipleInputs` too (i.e. check for `ion-checkbox.legacy-checkbox`)

amandaejohnston:
Removing the new CSS properties seems to have basically fixed all the item-related issues, so I was able to remove most of the new code here. Even the ripple effect is now properly handled by `ion-item` instead.

liamdebeasi (commented):
Clicking the far left of the ion-item triggers the ripple effect but does not check the checkbox


https://user-images.githubusercontent.com/2721089/208174770-8a833604-6031-4ea1-a930-caf64767e9b4.mov


The rest of the changes look good though 👍

liamdebeasi (approved):
We discussed handling the ion-item margin issue separately as this behavior already exists in the latest version of Ionic 6.

sean-perkins (approved):

PR #26499, opened 2022-12-16
feat(radio): support label, placement, justify

sean-perkins:
## Pull request checklist

## Pull request type

 

 

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes

The new standalone radio does not support labels, label placement and control justification.

Issue URL: Internal

## What is the new behavior?

- Adds base rendering of the radio
  - Render function is abstracted to share common rendering across legacy and modern radios
- Added support for rendering a label with `ion-radio`
- Added support for control the label placement relative to the control with the `labelPlacement` property
- Added support for packing the label and control in a line with the `justify` property
- Added tests to check label, label placement, justification, color prop, states and usage in `ion-item`

- [ ] Yes
- [x] No

## Other information

Focus/Active states are handled here: https://github.com/ionic-team/ionic-framework/pull/26527


core/src/components/radio/radio.tsx
```diff
@@ -141,6 +143,12 @@ export class Radio implements ComponentInterface {
 
   componentWillLoad() {
     this.emitStyle();
```
liamdebeasi, line 145: The contents of `emitStyle` should only fire with the legacy control

liamdebeasi (commented):
The radios in items have additional top/bottom margin that likely should only apply to the legacy radio: 

https://github.com/ionic-team/ionic-framework/blob/main/core/src/components/radio/radio.ios.scss#L73-L81

https://github.com/ionic-team/ionic-framework/blob/main/core/src/components/radio/radio.md.scss#L104-L112

Similar to https://github.com/ionic-team/ionic-framework/pull/26471, clicking the left margin of the item does not check the radio: https://user-images.githubusercontent.com/2721089/208175642-99b858bc-1498-47bb-bf55-c7ef60ed9f88.mov

core/src/components/radio/radio.tsx
sean-perkins: Should this apply to `ion-toggle` as well or does `ion-toggle` behave directly for style emissions than `ion-checkbox`/`ion-radio`?

liamdebeasi: Ah good catch - toggle should have that logic too. I'll make a PR for that

sean-perkins: Addressed the extra margin + `emitStyle`.

Still investigating the left margin not selecting the radio. This appears the legacy format as well + looks to apply to the modern format of `ion-toggle` as well.

sean-perkins:
This should be ready for another review. This does not account for the `20px` → `16px` margin that we are tracking.

core/src/components/radio/radio.scss
```diff
@@ -59,3 +61,131 @@ input {
...
+  flex-direction: row;
+}
+
+:host(.radio-label-placement-start) .label-text-wrapper {
+  /**
+   * The margin between the label and
+   * the radio should be on the end
+   * when the label sits at the start.
+   */
+  @include margin(0, 8px, 0, 0);
```
amandaejohnston, line 146: The label is still visible even if it's empty, leading to an extra bit of white space. (This is especially obvious in the new `a11y` test, which has extra space to the left of the second radio.) You'll want to hide the label element if a label isn't provided; toggle and checkbox use a `label-text-wrapper-hidden` class you could copy: https://github.com/ionic-team/ionic-framework/blob/8f573238909292088969e679a17dc1ff3be97539/core/src/components/checkbox/checkbox.scss#L102-L109

core/src/components/radio/test/label-placement/radio.e2e.ts
```diff
@@ -0,0 +1,110 @@
...
+
+/**
+ * By default ion-radio only takes up
+ * as much space as it needs. Justification is
+ * used for when the radio takes up the full
+ * line (such as in an ion-item). As a result,
+ * we set the width of the radio so we can
+ * see the justification results.
+ */
+test.describe('radio: label', () => {
```
amandaejohnston, line 12: Could we have these radios `checked` by default? Since an unchecked radio isn't visible on `ios` mode, those screenshots aren't super useful for showing the positioning.

core/src/components/radio/test/states/radio.e2e.ts
```diff
@@ -0,0 +1,39 @@
+import { expect } from '@playwright/test';
+import { test } from '@utils/test/playwright';
+
+test.describe('radio: states', () => {
+  test('should render disabled radio correctly', async ({ page, skip }) => {
```
amandaejohnston, line 5: This is another one that would be good to have checked by default to show `ios` mode better.

amandaejohnston (changes requested):

amandaejohnston (approved):

core/src/components/radio/radio.ios.scss
```diff
@@ -70,13 +72,28 @@
...
-:host(.in-item[slot="start"]) {
+:host(.in-item.legacy-radio[slot="start"]) {
   @include margin($radio-ios-item-start-margin-top, $radio-ios-item-start-margin-end, $radio-ios-item-start-margin-bottom, $radio-ios-item-start-margin-start);
 }
+
+// Radio Wrapper
+// ----------------------------------------------------------------
+
+.radio-wrapper {
+  min-height: 42px;
```
liamdebeasi, line 90: Can we move this to the `:host` and have .radio-wrapper inherit it? This was developers can customize it by targeting `ion-radio` in their CSS.

Note: I fixed this in input as well: https://github.com/ionic-team/ionic-framework/commit/0cd6e918e83ff4c2cdc41e1e9da05fefe6ca5c84

core/src/components/radio/test/item/index.html
```diff
@@ -0,0 +1,142 @@
...
+      </ion-header>
+
+      <ion-content id="content" class="ion-padding">
+        <h1>Placement Start</h1>
+        <div class="grid">
+          <div class="grid-item">
+            <h2>Justify Start</h2>
+            <ion-list>
+              <ion-item>
+                <ion-radio label-placement="start" justify="start">Enable Notifications</ion-radio>
```
liamdebeasi, line 56: Noticing a couple things here:

1. There is no `cursor: pointer`
2. There is no focus indicator.

Should these examples all use `ion-radio-group` too?

core/src/components/radio/test/a11y/index.html
```diff
@@ -0,0 +1,22 @@
...
+    <script nomodule src="../../../../../dist/ionic/ionic.js"></script>
+    <script type="module" src="../../../../../dist/ionic/ionic.esm.js"></script>
+  </head>
+
+  <body>
+    <main>
+      <h1>Radio - a11y</h1>
+
+      <ion-radio>my label</ion-radio><br />
+      <ion-radio aria-label="my aria label"></ion-radio><br />
```
liamdebeasi, line 19: Can we add a radio inside of a radio group? Also let's add a radio + radio group inside of a list

liamdebeasi (changes requested):

core/src/components/radio/test/item/index.html
sean-perkins: Updated to add the cursor styles and to render examples in `ion-radio-group` (for both the template and tests). 

I'd prefer to knock out the focus/active/hover indicator styles for radio in a separate work than label placement, just to not group too much in a single PR for later git bisect (unless you are thinking of something else). I'll track that in a new branch with additional screenshots for those individual states. 

Just waiting on new screenshots to generate for the radio usage in a list 👍

sean-perkins:
Looks like I need to tweak the styling a bit more in a list/item. Radio is unique in that it renders inside of `ion-radio-group`. The radio group does not currently have any styles, but it should likely fill the container space.

core/src/components/radio-group/radio-group.scss
```diff
@@ -0,0 +1,3 @@
+ion-radio-group {
+  flex: 1;
+}
```
liamdebeasi, line 3: What does this do?

core/src/components/radio/radio.scss
```diff
@@ -59,3 +67,142 @@ input {
...
+
+  flex-grow: 1;
+
+  align-items: center;
+
+  height: 100%;
+
+  min-height: inherit;
+  
+  transition: background-color 15ms linear;
```
liamdebeasi, line 90: This was added for MD inputs/textareas. Do we need this for the radio?

core/src/components/radio-group/radio-group.scss
sean-perkins: This causes the radio group to fill the remaining space of the container, alternatively we could use `width: 100%` here, it will have the same visual effect in this instance.

e.g.:

Without filling remaining space:

localhost_3333_src_components_radio_test_item (1)

Filling remaining space:

localhost_3333_src_components_radio_test_item

This ties to this comment: https://github.com/ionic-team/ionic-framework/pull/26499#issuecomment-1360393586.

core/src/components/radio/radio.scss
sean-perkins: Good callout, I don't think so. The radio wrapper doesn't have a background color directly. I'll remove and validate 👍

sean-perkins:
Note: I still need to add the `ion-focusable` class to the implementation so that the previous focus styling applies with the implementation. Then the other PR: https://github.com/ionic-team/ionic-framework/pull/26527 isn't needed (can be slated for 7.1 or minimally after a design doc).

core/src/components/radio/radio.tsx
```diff
@@ -214,18 +214,21 @@ export class Radio implements ComponentInterface {
...
-          'in-item': hostContext('ion-item', el),
-          interactive: true,
+          'in-item': inItem,
           'radio-checked': checked,
           'radio-disabled': disabled,
           [`radio-justify-${justify}`]: true,
           [`radio-label-placement-${labelPlacement}`]: true,
+          // Focus and active styling should not apply when the radio is in an item
+          'ion-activatable': !inItem,
+          'ion-focusable': !inItem,
```
liamdebeasi, line 231: Why?

liamdebeasi (approved):
One question, but otherwise looks great.

core/src/components/radio/radio.tsx
sean-perkins: This disabled the focus and active styling for the radio, when it is rendered inside of an item. Without it, the focused styles display, in addition to the item's focus styles:

localhost_3333_src_components_radio_test_item(iPhone 12 Pro)

This is inconsistent with how a radio behaves within an item on `main`:

CleanShot 2023-01-09 at 12 13 22

We do however, want focus styles when the radio is rendered by itself and then focused.

Issue #26542, opened 2022-12-28
bug: Presenting a modal with a leave animation immediately after dismissing clears the presented modal's content


thor9n:
### Prerequisites

- [x] I have read the [Contributing Guidelines](https://github.com/ionic-team/ionic-framework/blob/main/.github/CONTRIBUTING.md#creating-an-issue).
- [X] I agree to follow the [Code of Conduct](https://ionicframework.com/code-of-conduct).
- [X] I have searched for [existing issues](https://github.com/ionic-team/ionic-framework/issues) that already report this problem, without success.

### Ionic Framework Version

- [ ] v4.x
- [ ] v5.x
- [X] v6.x
- [ ] Nightly

### Current Behavior

**Behaviour**
- Calling `modalController.dismiss(undefined, undefined, "my-modal-id")` in an existing modal, and then immediately opening a new modal, causes the new modal's content to be deleted.
- Everything works as expected if `leaveAnimation` are disabled when creating the modal, but obviously this shouldn't be needed.

**Links**
[Live example on Codesandbox](https://codesandbox.io/s/ionic-modal-dismiss-bug-7wmotx)
[Github issue explaining the three dismiss() arguments](https://github.com/ionic-team/ionic-framework/issues/21173#issuecomment-622058594)

**Background/Use-case**
When my users creates a new playlist they are redirected to that list's page, and immediately a Sheet Modal with the option to add songs are shown.

### Expected Behavior

The following steps should display the last modal's content, no matter what animation settings are applied to the first modal.

```
1. User opens Modal #1
2. User clicks button → Redirects to New Page
3. New Page opens Modal #2
4. Content does NOT disappear
```

### Steps to Reproduce

Link to Codesandbox showing the bug:
https://codesandbox.io/s/ionic-modal-dismiss-bug-7wmotx

_(don't ask me why the routes are placed above eachother when redirecting, but the example works anyways)_

### Code Reproduction URL

https://codesandbox.io/s/ionic-modal-dismiss-bug-7wmotx

### Ionic Info

```
   Ionic CLI       : 6.20.4 (/usr/local/lib/node_modules/@ionic/cli)
   Ionic Framework : @ionic/vue 6.3.8

Capacitor:

   Capacitor CLI      : 4.5.0
   @capacitor/android : 4.5.0
   @capacitor/core    : 4.5.0
   @capacitor/ios     : 4.5.0

Utility:

   cordova-res : not installed globally
   native-run  : 1.7.1

System:

[...]

PR #26543, opened 2022-12-28
fix(vue): unmount teleported user component

sean-perkins:
## Pull request checklist

## Pull request type

 

 

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

Presenting a modal immediately after dismissing a modal, can incorrectly clear the contents of the presented modal.

Issue URL: #26542

## What is the new behavior?

- Presenting a modal immediately after dismissing a modal does not clear the contents of the presented modal.
- Teleported components are removed using `splice` instead of `filter`.

- [ ] Yes
- [x] No

## Other information

Dev-build: `6.4.2-dev.11672257788.1a6cf37c`

### DIFF
--- a/.github/ionic-issue-bot.yml
+++ b/.github/ionic-issue-bot.yml
@@ -1,15 +1,10 @@
 triage:
-  label: "holiday triage"
+  label: triage
   removeLabelWhenProjectAssigned: true
   dryRun: false
 
 comment:
   labels:
-    - label: "holiday triage"
-      message: >
-        Thanks for the issue! This issue has been labeled as `holiday triage`. With the winter holidays quickly approaching, much of the Ionic Team will soon be taking time off. During this time, issue triaging and PR review will be delayed until the team begins to return. After this period, we will work to ensure that all new issues are properly triaged and that new PRs are reviewed.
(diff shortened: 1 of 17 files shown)


liamdebeasi (commented):
I don't get why this would fix the issue since these two code snippets should yield the same results:

```js
const toRemove = '3';
let dataArray = ['1', '2', '3', '4', '5'];
dataArray = dataArray.filter(i => i !== toRemove);

console.log(dataArray);
// ['1', '2', '4', '5']
```

```js
const toRemove = '3';
let dataArray = ['1', '2', '3', '4', '5'];

const index = dataArray.findIndex(i => i === toRemove);
dataArray.splice(index, 1);

console.log(dataArray);
// ['1', '2', '4', '5']
```

sean-perkins:
There shouldn't be a difference in the implementations with static data, but for some reason there is a tangible difference in the outcome with our instance, where `userComponents.value` is being manipulated by both the new modal being presented and the previous modal being dismissed; resulting in the incorrect modal being filtered out. 

I've ran the same test at least 30 times while locally debugging and it is consistent with `findIndex` unmounting the correct view and `filter` removing the presented view.

Is this some weird Vue quirk with shallow refs and rendering?

liamdebeasi:
`filter` returns a new array whereas `splice` updated the array in-place. Maybe that causes some difference for Vue?

liamdebeasi (approved):
Looks good, but can we add a comment that explains why we need to use `splice` here?

PR #26564, opened 2023-01-04
chore(item): modify slot deprecation notice to avoid JSDoc formatting issue

alicewriteswrongs:
This changes the deprecation notice for a few `@slot` tags on the `ion-item` component. Previously the lines in question looked something like this:

```ts
/**
 * @slot a slot that does something @deprecated this is going away!
 */
export class Item { ...
```

The intention behind formatting the JSDoc this way was to mark the `@slot` itself as deprecated, to advice users that it will be removed in the future.

Unfortunately, JSDoc tooling doesn't interpret the `@deprecated` tag as referring to the `@slot` tag here, but instead as one of the tags applied to the `Item` class. If we wanted to use `@deprecated` to mark only a slot as deprecated then we'd need a way to write a JSDoc just for that slot itself.

Since we don't want to mark the whole class as deprecated, as a workaround, we can change to using `**DEPRECATED**` instead, which will appear prominently in generated documentation but will not confuse tools which are JSDoc-aware.

## Pull request checklist

## Pull request type

 

 

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes

Issue URL: 

## What is the new behavior?

-
-
-

- [ ] Yes
- [x] No

### DIFF
--- a/core/src/components/item/item.tsx
+++ b/core/src/components/item/item.tsx
@@ -19,8 +19,8 @@ import type { CounterFormatter } from './item-interface';
  * @slot - Content is placed between the named slots if provided without a slot.
  * @slot start - Content is placed to the left of the item text in LTR, and to the right in RTL.
  * @slot end - Content is placed to the right of the item text in LTR, and to the left in RTL.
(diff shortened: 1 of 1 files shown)


liamdebeasi (approved):
Thank you!

PR #26504, opened 2022-12-16
refactor(select): add base template rendering

liamdebeasi:
## Pull request checklist

Please check if your PR fulfills the following requirements:
- [ ] Tests for the changes have been added (for bug fixes / features)
- [ ] Docs have been reviewed and added / updated if needed (for bug fixes / features)
  - Some docs updates need to be made in the `ionic-docs` repo, in a separate PR. See the [contributing guide](https://github.com/ionic-team/ionic-framework/blob/main/.github/CONTRIBUTING.md#modifying-documentation) for details.
- [ ] Build (`npm run build`) was run locally and any changes were pushed
- [ ] Lint (`npm run lint`) has passed locally and any fixes were made for failures

## Pull request type

 

 

Please check the type of change your PR introduces:
- [ ] Bugfix
- [ ] Feature
- [ ] Code style update (formatting, renaming)
- [ ] Refactoring (no functional changes, no api changes)
- [ ] Build related changes
- [ ] Documentation content changes
- [ ] Other (please describe): 

## What is the current behavior?

Select does not support rendering components using the new syntax

## What is the new behavior?

- aria-label is now inherited from the host to the inner button component
- Added label rendering markup based on `ion-input`'s new markup
- Added classes for justification, ltr/rtl, and label placement

## Does this introduce a breaking change?

- [ ] Yes
- [x] No

PR #26572, opened 2023-01-05
feat(select): component can be used outside ion-item

liamdebeasi:
## Pull request checklist

## Pull request type

 

 

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes

Issue URL: Internal ticket

## What is the new behavior?

- `ion-select` can now be used outside of `ion-item` with a new property to specify the label.
This PR is a combination of the following:

https://github.com/ionic-team/ionic-framework/pull/26470
https://github.com/ionic-team/ionic-framework/pull/26502
https://github.com/ionic-team/ionic-framework/pull/26504
https://github.com/ionic-team/ionic-framework/pull/26505
https://github.com/ionic-team/ionic-framework/pull/26506
https://github.com/ionic-team/ionic-framework/pull/26570

- [ ] Yes
- [ ] No


core/src/components/select/test/fill/index.html
```diff
@@ -0,0 +1,151 @@
...
+          <div class="grid-item">
+            <h2>Shaped Solid</h2>
+            <ion-select shape="round" fill="solid" label="Fruit" value="apple">
+              <ion-select-option value="apple">Apple</ion-select-option>
+            </ion-select>
+          </div>
+
+          <div class="grid-item">
+            <h2>Solid, Floating Label</h2>
+            <ion-select fill="solid" label="Fruit" label-placement="floating" value="apple"></ion-select>
```
sean-perkins, line 75: ~ Is there a reason that some of the examples have options and others do not? I know the test does not use this template, but it might be helpful if all the examples render at least 1 option for a consistent display (unless we are wanting to demonstrate a state of no options).

Adding options would also allow manual verification of the visual state after selecting a value.

sean-perkins (commented):
Should we have examples/tests for the different interface appearances of `ion-select` and the modern rendering? 

The legacy tests show examples with `interface="popover"`.

core/src/components/select/test/fill/index.html
liamdebeasi: I think that was an oversight on my part. I pushed a commit that ensures all selects have values.

liamdebeasi:
> Should we have examples/tests for the different interface appearances of ion-select and the modern rendering?
> The legacy tests show examples with interface="popover".

I added tests. Will re-run screenshot jobs shortly.

amandaejohnston:
I think it should still appear on the end side, since that's where the actual select appears to be. Otherwise it can easily look like the popover is just opening under empty space, which is strange. IMO behavior should match with how the legacy select acts when placed within an item, since that's what it looks like, even when the new syntax isn't used with an item as well.

liamdebeasi:
Ok I fixed a handful of things:

1. Noticed an issue where justification wasn't working. This broke as a result of some code I missed when resolving a merge conflict.
2. Fixed an issue Amanda noted where the popover is not aligned correctly. Should now be aligned relative to the placeholder/icon.
3. Fixed an issue where the popover does not take up the full width when using outline/solid styles
4. Fixed an issue where the chevron was rotating on open on iOS. This was due to some code that slipped through when fixing a merge conflict.

Updating screenshots now and will re-request another review when that is done.

amandaejohnston (approved):

sean-perkins (approved):

PR #26575, opened 2023-01-06
fix(segment): nested interactives are not rendered

liamdebeasi:
## Pull request checklist

## Pull request type

 

 

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

Issue URL: Internal ticket

Axe test warns of nested interactive in `ion-segment-button`. This is happening because `ion-segment-button` has `role="tab"` on the host element but then renders a `<button>` in its Shadow DOM. Axe thinks that both the host and the nested button are interactive.

This was added in https://github.com/ionic-team/ionic-framework/pull/23145 before Axe warned about nested interactive (so this was never caught). The original fix was architected this way due to a bug in Chrome on macOS where it does not read the total tab count correctly due to the `role="tab"` elements being in the Shadow DOM.
However, the workaround that avoided this issue introduced the nested interactive issue.

I filed https://bugs.chromium.org/p/chromium/issues/detail?id=1405462 with the Chrome team. I think it is better to avoid the nested interactive issue at the expense of surfacing this Chrome bug. The nested interactive issue is an actual bug that we introduced and impacts all platforms that use assistive tech. The incorrect count bug is a Chrome bug and is limited to Chrome on macOS. The surface area of the Chrome bug is smaller than the nested interactive bug.

## What is the new behavior?

- aria-selected and role states on `ion-segment-button` have been moved to the native `<button>` element inside of the Shadow DOM.
- Segment now calls the internal `setFocus` method on `ion-segment-button` so that the native button element is focused instead of the host element.
- `ariaLabel` on `ion-segment-button` is now inherited to the native `<button>` element.

Notable Changes:

1. Axe will no longer warn about "nested interactives"
2. VoiceOver with Chrome on macOS will now read "tab `n` of 1"  even if there is more than 1 tab in the tab group.

Not impacted by this change:

- TalkBack on Android
- VoiceOver on iOS
- VoiceOver with Safari on macOS
- NVDA on Windows

- [ ] Yes
- [x] No

### DIFF
--- a/core/src/components.d.ts
+++ b/core/src/components.d.ts
@@ -2483,6 +2483,7 @@ export namespace Components {
           * The mode determines which platform styles to use.
          */
         "mode"?: "ios" | "md";
+        "setFocus": () => Promise<void>;
         /**
           * The type of the button.
          */
--- a/core/src/components/segment-button/segment-button.tsx
+++ b/core/src/components/segment-button/segment-button.tsx
@@ -1,5 +1,5 @@
 import type { ComponentInterface } from '@stencil/core';
-import { Component, Element, Host, Prop, State, forceUpdate, h } from '@stencil/core';
+import { Component, Element, Host, Prop, Method, State, forceUpdate, h } from '@stencil/core';
 
 import { getIonMode } from '../../global/ionic-global';
 import type { SegmentButtonLayout } from '../../interface';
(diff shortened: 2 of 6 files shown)


sean-perkins (approved):
Regressing Chrome macOS in favor of the larger accessibility improvement for other targets, makes complete sense to me.

liamdebeasi:
Thanks! Also I added https://github.com/ionic-team/ionic-framework/pull/26575/commits/bb57d37e8830eae81eea99f59317c86864c51c5f which inherits `ariaLabel` to the native button inside of `ion-segment-button` and added a test to a11y (missed this during my initial dev)

sean-perkins (approved):

amandaejohnston (approved):

Issue #20819, opened 2020-03-19
bug: ion-list attribute "lines" not working when attribute "inset" is true


DwieDima:
# Bug Report

**Ionic version:**

[x] **5.0.5**

**Current behavior:**

When you create an `ion-list` with `ion-items` and set the attribute `inset="true"` on `ion-list`, the attribute `lines` wont have any effect on `ion-item`. Instead you'll always see full lines.

**Expected behavior:**

attribute `lines` on `ion-list` should work when attribute `inset="true"` is set. 

**Steps to reproduce:**

copy paste this snippet to any page:

```html
 <ion-list inset="true" lines="full">
    <ion-item detail="true">
      <ion-label>test</ion-label>
    </ion-item>
    <ion-item detail="true">
      <ion-label>test</ion-label>
    </ion-item>
    <ion-item detail="true">
      <ion-label>test</ion-label>
    </ion-item>
    <ion-item detail="true">
      <ion-label>test</ion-label>
    </ion-item>
    <ion-item detail="true">
      <ion-label>test</ion-label>
    </ion-item>
    <ion-item detail="true">
      <ion-label>test</ion-label>
    </ion-item>
    <ion-item detail="true">
      <ion-label>test</ion-label>
    </ion-item>
  </ion-list>
```
- change lines attribute to `lines="none"`, `lines="inset"` --> no effect.
- removing lines attribute will remain in `lines="full"`.
  - Isn't the default behaviour `lines="inset"` ?

**Other information:**

**Ionic info:**

```
Ionic:

   Ionic CLI                     : 6.2.2 (/usr/local/lib/node_modules/@ionic/cli)
   Ionic Framework               : @ionic/angular 5.1.0-dev.202003171608.9d9a02f
   @angular-devkit/build-angular : 0.900.5
   @angular-devkit/schematics    : 9.0.5
   @angular/cli                  : 9.0.5
   @ionic/angular-toolkit        : 2.2.0

Capacitor:

   Capacitor CLI   : 1.5.0
   @capacitor/core : 1.5.0

Cordova:

   Cordova CLI       : 9.0.0 (cordova-lib@9.0.1)
   Cordova Platforms : none
   Cordova Plugins   : no whitelisted plugins (1 plugins total)

Utility:

   cordova-res : not installed
   native-run  : not installed

System:

   ios-deploy : 1.10.0
   ios-sim    : 8.0.2
   NodeJS     : v12.16.0 (/usr/local/bin/node)
[...]

PR #26586, opened 2023-01-10
fix(list): inset lists render correctly

liamdebeasi:
## Pull request checklist

## Pull request type

 

 

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

Issue URL: resolves https://github.com/ionic-team/ionic-framework/issues/20819

**iOS**

During a design review, Ben noted that the border width on inset lists in iOS was too thick. The list component was overriding the item border width to be 1px (item sets the width to 0.55px on iOS)

Upon further investigation, it was noted that the `lines` property was never being respected when `inset="true"`. As a result, it was impossible to get inset item lines using the inset list feature. Lines should be inset by default on iOS.

**MD**

During a design review, it was noticed that the padding on inset lists that contain interactive ion-items was incorrect. Fixing this issue revealed another bug where full lines were not being removed from items when `lines="inset"` was set on `ion-list`.

## What is the new behavior?

Architecture Note:

I decided to change the architecture of how list applies these styles for a couple reasons:

1. It was buggy (as noted above)
2. It was hard to reason about. IMO `.list-md-lines-inset .item, .list-md .item-lines-inset` is not the most "readable" selector.
3. It reimplemented styles already found in `ion-item`. For example: https://github.com/ionic-team/ionic-framework/pull/26586/files#diff-7dc0bcee862e87a0457c625c8ac1487390b4d3d9a39c667e01485537255c1cb2L67-L83
These lines made it so developers could set the `lines` property on `ion-item` and have that particular instance override whatever line `ion-list` was trying to set. However, it did this by duplicating styles that were already set in the `ion-item` stylesheets.

This is the new architecture:

1. A `item-lines-default` class is set on `ion-item` when no `lines` property is set.
2. `ion-list` targets `ion-item` with the `item-lines-default` class when setting the lines.
3. If a developers sets `lines` on `ion-item`, then the `item-lines-default` class is removed and the styles in 2 no longer apply. This allows developers to override the list behavior on a per-item basis.

**iOS**

- Inset lists no longer override the border on item

| `main` | branch | native |
| - | - | - |
| IMG_1775 | IMG_1776 | IMG_1768 |

Note: This was previously considered an Ionic 7.0 task. However, I would like to consider this a non-breaking change since the `lines` property on `ion-list` is not functional. Additionally, the lines should always have been 0.55px instead of 1px.

**MD**

- List no longer removes padding on items with interactives

Fixing the padding issue reveals the lines bug.

| `main` | branch |
| - | - |
| IMG_1781 | IMG_1782 |

In the "branch" screenshot, the first item in the list has both an inset and a full line. This is also happening in `main`, but the lack of padding causes this issue to not be visible.

- The full line is removed when `lines="inset"` is set on `ion-list`

| pre-fix | branch |
| - | - |
| IMG_1782 | IMG_1783 |

This likely has been an issue for several years, but we never caught it since we did not have test coverage for this use case. Tests have been added.

- [ ] Yes
- [x] No

This does not change the specificity of the selector. Prior to this change we were doing `.list-ios-lines-full .item`. Now we are doing `.list-ios-lines-full .item-lines-default`. As a result, any app-specific CSS overrides should still work.


core/src/components/list/list.md.scss
```diff
@@ -68,9 +62,22 @@
...
 
-.list-md-lines-inset .item,
+/**
+ * We remove the full lines below, but developers
+ * should still be able to override the lines on
+ * a per-item basis. As a result, we this selector
+ * to exclude a) items with full lines, b) items with
+ * no lines, and c) the last item in an inset list.
+ */
+.list-md-lines-inset .item:not(.item-lines-full):not(.item-lines-none):not(:last-child),
```
liamdebeasi, line 72: This gnarly selector is so you can still override lines on a per-item basis. Any other ideas for how to do this? I know you can do complex selectors in a single `:not`  but that is not available in all browser versions we support.

liamdebeasi:
Going to loop back to this, but I think there is a better way to architect this.

amandaejohnston (approved):

Issue #26553, opened 2023-01-02
bug: Refresher overwrites virtual scroll overflow styling


piotr-cz:
### Prerequisites

- [X] I have read the [Contributing Guidelines](https://github.com/ionic-team/ionic-framework/blob/main/.github/CONTRIBUTING.md#creating-an-issue).
- [X] I agree to follow the [Code of Conduct](https://ionicframework.com/code-of-conduct).
- [X] I have searched for [existing issues](https://github.com/ionic-team/ionic-framework/issues) that already report this problem, without success.

### Ionic Framework Version

- [ ] v4.x
- [ ] v5.x
- [X] v6.x
- [ ] Nightly

### Current Behavior

The `ion-refresher` prevents contents of Virtual Scroll to be scrolled.

### Expected Behavior

The `ion-refresher` doesn't prevent contents of Virtual Scroll to be scrolled.

### Steps to Reproduce

1. Install `@ionic/react` and `@ionic/core` in version 6.2.2 (correct behaviour)
   ```console
   npm install --save-exact @ionic/react@6.2.2 @ionic/core@6.2.2
   ```
1. Create list with Virtual Scroll as in Ionic Docs > [Virtual Scroll](https://ionicframework.com/docs/react/virtual-scroll)
1. Add Refresher component above the list as in Ionic Docs > Refresher > ion-refresher > [Usage with Virtual Scroll](https://ionicframework.com/docs/api/refresher#usage-with-virtual-scroll)
1. Make sure it's possible to scroll contents and pull to refresh
3. Install `@ionic/react` and `@ionic/core` in version 6.2.4 or higher (bug introduced)
   ```console
   npm install --save-exact @ionic/react^@6.4.1 @ionic/core@^6.4.1
   ```
2. See that it's not possible to scroll contents

### Code Reproduction URL

https://stackblitz.com/edit/angular-4uocfn (Forked from docs > ion-refresher > [Advanced Usage](https://ionicframework.com/docs/api/refresher#advanced-usage))

### Ionic Info

Ionic:

   Ionic CLI       : 6.20.6 (C:\Users\xxxx\AppData\Roaming\npm\node_modules\@ionic\cli)
   Ionic Framework : @ionic/react 6.4.1

Capacitor:

   Capacitor CLI      : 3.7.0
   @capacitor/android : 3.7.0
   @capacitor/core    : 3.7.0
   @capacitor/ios     : 3.7.0

Utility:

   cordova-res : not installed globally
[...]

PR #26613, opened 2023-01-17
fix(refresher): prevent clearing virtual scroll overflow styling

sean-perkins:
## Pull request checklist

## Pull request type

 

 

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

`ion-refresher` will overwrite 3rd-party overflow styles assigned by virtual scroll libraries (e.g. Virtuoso). 

Issue URL: #26553

## What is the new behavior?

- `ion-refresher` will only assign `hidden` overflow styling when overflow is visible.
- `ion-refresher` will restore inline overflow styles set on the scroll element.

- [ ] Yes
- [x] No

## Other information

Dev-build: `6.4.3-dev.11674237446.1a3aa9dc`

Updated reproduction: https://stackblitz.com/edit/angular-4uocfn-n1qkbm?file=src%2Fmain.tsx,package.json

Also the local scroll target test template has been updated to leverage the underlying behavior that originally causes this bug.


core/src/components/refresher/refresher.tsx
```diff
@@ -729,7 +729,15 @@ export class Refresher implements ComponentInterface {
...
         scrollStyle.transitionDuration = backgroundStyle.transitionDuration = duration;
         scrollStyle.transitionDelay = backgroundStyle.transitionDelay = delay;
-        scrollStyle.overflow = overflowVisible ? 'hidden' : '';
+        /**
+         * The goal is to avoid overriding the settings of certain virtual scroll
+         * libraries that already set the overflow style on the scroll container.
+         */
+        if (overflowVisible) {
+          scrollStyle.overflow = 'hidden';
+        } else if (scrollStyle.overflow === 'hidden') {
```
liamdebeasi, line 738: I don't understand how this fixes the issue. I tried the reproduction you provided in https://github.com/ionic-team/ionic-framework/issues/26553#issuecomment-1385754385 with iOS mode, and the scroller is still disabled after completing the pull to refresh action.

My understanding of the issue is that we are overriding an `overflow` style that is set inline on the scroller (as opposed to in a stylesheet). If that's the case, we likely need to keep track of the inline style before we call `setCss` so we can restore the correct style.

core/src/components/refresher/refresher.tsx
```diff
@@ -729,11 +731,35 @@ export class Refresher implements ComponentInterface {
...
+    }
+  }
+
+  private restoreOverflowStyle() {
+    if (this.overflowStyles !== undefined && this.scrollEl !== undefined) {
+      const { overflow, overflowX, overflowY } = this.overflowStyles;
+      this.scrollEl.style.overflow = overflow;
+      this.scrollEl.style.overflowX = overflowX;
+      this.scrollEl.style.overflowY = overflowY;
+    }
```
liamdebeasi, line 760: Do we need to clear `this.overflowStyles` when done?

core/src/components/refresher/refresher.tsx
```diff
@@ -729,11 +731,35 @@ export class Refresher implements ComponentInterface {
         scrollStyle.transform = backgroundStyle.transform = y > 0 ? `translateY(${y}px) translateZ(0px)` : '';
         scrollStyle.transitionDuration = backgroundStyle.transitionDuration = duration;
         scrollStyle.transitionDelay = backgroundStyle.transitionDelay = delay;
-        scrollStyle.overflow = overflowVisible ? 'hidden' : '';
```
liamdebeasi, line 732: One thing I noticed with the MD native refresher is that it sets `--overflow: hidden` (assumes `ion-content`) so the overflow is never set to `hidden` on virtual scroll targets. Not something we need to fix in this PR, but should probably track this.

liamdebeasi (approved):
A couple comments, but otherwise looks good. Great job!

core/src/components/refresher/refresher.tsx
sean-perkins: I did some initial testing and clearing the overflow styles does not impact the experience, so I updated the implementation to clear the styles after assigning them.

PR #26635, opened 2023-01-18
chore(vue): add eslint and prettier

liamdebeasi:
## Pull request checklist

## Pull request type

 

 

- [ ] Bugfix
- [ ] Feature
- [x] Build related changes

Since we are migrating `@ionic/react` and `@ionic/react-router` to ESLint + Prettier, I wanted to take this opportunity to migrate `@ionic/vue` as well.

## What is the new behavior?

- Add ESLint to `@ionic/vue` and uses the Ionic ESLint config: https://github.com/ionic-team/eslint-config
- Adds Prettier and installs the Ionic Prettier config: https://github.com/ionic-team/prettier-config/
- Updates the CI scripts to avoid the `--legacy-peer-deps` flag and prefer `npm ci` instead of `npm install`

Note: For ESLint I only used rules that likely would not cause any functional changes. We've been burned before with rules such as `prefer-optional-chain` causing functional changes. I also added several ignores to quirks in Ionic React such as how we use the spread operator to remove keys from the remaining object (but then never use the keys)

- [ ] Yes
- [x] No


packages/vue/.eslintignore
```diff
@@ -0,0 +1,2 @@
+vue-component-lib/
+src/components/Overlays.ts
```
amandaejohnston, line 2: Why was `Overlays.ts` ignored? (Ditto for its inclusion in `.prettierignore`)

liamdebeasi: Overlays.ts is autogenerated

amandaejohnston (approved):

PR #26646, opened 2023-01-19
feat(base-components): create separate mode and platform terms

liamdebeasi:
## Pull request checklist

## Pull request type

 

 

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes

Ionic's "mode" option currently controls both the visuals and functionalities of components. Setting `<ion-modal mode="ios">` loads the iOS stylesheet for modal and enables the use of the card modal, a feature only available on the iOS platform. However, it is currently impossible to clear and iOS styles while retaining the functionality.

For example, if developers wanted to apply heavy customizations to how the card modal looks, they would first need to manually disable each iOS style they did not want before applying their custom styles. This is tedious and results in developers having to fight against the platform styles in order to get an app that does not look like a "cookie-cutter" iOS or MD app.

Base Components solves this problem by creating a new API that only controls the visuals of an application. This means that developers can retain iOS-specific behaviors such as swipe to go back and the card modal while customizing the look of their application.

## What is the new behavior?

**Note** There should be no functional changes as a result of this PR. Components are still unable to opt-in to use base components. This PR focuses on the architecture/bookkeeping required to ensure there is a clear separation between mode visuals and platform functionality.

However, you can test this manually by doing the following:

1. Add `base: '[component].scss'` to the `styleUrls` object for a component
2. Use the `baseComponents` global config or the `use-base` attribute to enable base components.
This functionality will be hooked up in a future PR.

This PR does a few things: 

**getIonBehavior**
Internally, we separate the visuals from the functionality by creating a new `getIonBehavior` function. While `getIonMode` tells developers which stylesheet is loaded for a component, `getIonBehavior` tells developers which platform the component is tailoring its capabilities for.

Example: `getIonMode(this) === 'base' && getIonBehavior(this) === 'ios'` means that a component has no visual opinionation but functions as it would on an iOS device.

**getIonStylesheet** (formerly `getIonMode`)

`getIonMode` has traditionally been used to detect the visual and functional style that an app is running in. `getIonMode`'s behavior has been preserved for backwards compatibility. Developers can still use `mode` (either via the global config or via the property on a component) to set both the visuals and the functionalities at once. Components that use `getIonMode` to render the platform class on the host still do this. However, this class may sometimes be `'base'` instead of `'ios'` or `'md'`. This allows developers to target instances of components that are using base components.

To better show the separation between visuals and functionality, `getIonMode` has been renamed to `getIonStylesheet`. This function is used to determine the stylesheet that Stencil has loaded.

Example:

A developer can write the following:
```html
<!-- Assume mode="ios" globally -->
<ion-button id="first" use-base="true">Button</ion-button>
<ion-button id="second">Button</ion-button>
```

Which would render out to
```html
<ion-button id="first" use-base="true" class="base ...">Button</ion-button>
<ion-button id="second" class="ios ...">Button</ion-button>
```

From here, developers can target just the button instance that uses base components:

```css
ion-button.base {
  --background: green; // only the button using base components receives a green background
}
```

Developers who opt an entire component in (using the global config) do not need to target this `.base` class. Instead they can target just `ion-button` since every instance of the button would use base components in this scenario.

**API**

The API has been marked as experimental. Base components will remain in developer preview for now so we can collect feedback from the community. This also means we can make breaking changes to this feature outside of the typical major release cycle.

**setMode**

The `setMode` function is used by Stencil to load the correct stylesheet (see `styleUrls` in `ion-button` for an example). If `setMode` returns `'ios'` then the `'ios'` stylesheet would be loaded for that instance of the component.

This function has been updated to return `'base'` if an instance of a component is using base components. This will cause Stencil to only load the base/structural stylesheet for that instance of a component.

We also set the `platform` and `useBase` properties so we can correctly track the platform capabilities of each component (see `getIonPlatform` above).

**Component Consumption**

Several components have been updated to migrate their internal usage of `getIonMode` to `getIonPlatform`. I only modified components where the code was trying to determine the capabilities of a component. The `getIonMode` usage to determine the visuals of a component (or anything related to public APIs for styling) have remained untouched.

- [ ] Yes
- [x] No

### DIFF
--- a/core/src/components/content/content.tsx
+++ b/core/src/components/content/content.tsx
@@ -125,8 +125,8 @@ export class Content implements ComponentInterface {
 
   private shouldForceOverscroll() {
     const { forceOverscroll } = this;
-    const mode = getIonPlatform(this);
-    return forceOverscroll === undefined ? mode === 'ios' && isPlatform('ios') : forceOverscroll;
+    const platform = getIonPlatform(this);
+    return forceOverscroll === undefined ? platform === 'ios' && isPlatform('ios') : forceOverscroll;
   }
 
   private resize() {
--- a/core/src/components/content/content.tsx
+++ b/core/src/components/content/content.tsx
@@ -1,7 +1,7 @@
 import type { ComponentInterface, EventEmitter } from '@stencil/core';
 import { Component, Element, Event, Host, Listen, Method, Prop, forceUpdate, h, readTask } from '@stencil/core';
 
-import { getIonMode } from '../../global/ionic-global';
+import { getIonMode, getIonPlatform } from '../../global/ionic-global';
 import type { Color, ScrollBaseDetail, ScrollDetail } from '../../interface';
 import { componentOnReady } from '../../utils/helpers';
 import { isPlatform } from '../../utils/platform';
@@ -125,7 +125,7 @@ export class Content implements ComponentInterface {
 
   private shouldForceOverscroll() {
     const { forceOverscroll } = this;
-    const mode = getIonMode(this);
+    const mode = getIonPlatform(this);
     return forceOverscroll === undefined ? mode === 'ios' && isPlatform('ios') : forceOverscroll;
   }
 
--- a/core/src/components/content/content.tsx
+++ b/core/src/components/content/content.tsx
@@ -1,7 +1,7 @@
 import type { ComponentInterface, EventEmitter } from '@stencil/core';
 import { Component, Element, Event, Host, Listen, Method, Prop, forceUpdate, h, readTask } from '@stencil/core';
 
-import { getIonMode, getIonPlatform } from '../../global/ionic-global';
+import { getIonStylesheet, getIonPlatform } from '../../global/ionic-global';
 import type { Color, ScrollBaseDetail, ScrollDetail } from '../../interface';
 import { componentOnReady } from '../../utils/helpers';
 import { isPlatform } from '../../utils/platform';
@@ -324,7 +324,7 @@ export class Content implements ComponentInterface {
   render() {
     const { isMainContent, scrollX, scrollY, el } = this;
     const rtl = isRTL(el) ? 'rtl' : 'ltr';
-    const mode = getIonMode(this);
+    const mode = getIonStylesheet(this);
     const forceOverscroll = this.shouldForceOverscroll();
     const transitionShadow = mode === 'ios';
     const TagType = isMainContent ? 'main' : ('div' as any);
--- a/core/src/components/content/content.tsx
+++ b/core/src/components/content/content.tsx
@@ -1,7 +1,7 @@
 import type { ComponentInterface, EventEmitter } from '@stencil/core';
 import { Component, Element, Event, Host, Listen, Method, Prop, forceUpdate, h, readTask } from '@stencil/core';
 
-import { getIonStylesheet, getIonPlatform } from '../../global/ionic-global';
+import { getIonStylesheet, getIonBehavior } from '../../global/ionic-global';
 import type { Color, ScrollBaseDetail, ScrollDetail } from '../../interface';
 import { componentOnReady } from '../../utils/helpers';
 import { isPlatform } from '../../utils/platform';
@@ -125,7 +125,7 @@ export class Content implements ComponentInterface {
 
   private shouldForceOverscroll() {
     const { forceOverscroll } = this;
-    const platform = getIonPlatform(this);
+    const platform = getIonBehavior(this);
     return forceOverscroll === undefined ? platform === 'ios' && isPlatform('ios') : forceOverscroll;
   }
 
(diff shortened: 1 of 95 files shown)


core/src/components/header/header.tsx
```diff
@@ -71,7 +71,7 @@ export class Header implements ComponentInterface {
   }
 
   private async checkCollapsibleHeader() {
-    const mode = getIonMode(this);
+    const mode = getIonPlatform(this);
```
sean-perkins, line 74: We renamed `mode` → `platform` in similar changes. Should we do the same here?

core/src/components/content/content.tsx
```diff
@@ -125,7 +125,7 @@ export class Content implements ComponentInterface {
 
   private shouldForceOverscroll() {
     const { forceOverscroll } = this;
-    const mode = getIonMode(this);
+    const mode = getIonPlatform(this);
```
sean-perkins, line 128: We renamed `mode` → `platform` in similar changes. Should we do the same here?

sean-perkins (approved):
Changes look good to me. Called out a few small inconsistencies, but those could be refactored later.

core/src/components/refresher-content/refresher-content.tsx
```diff
@@ -52,18 +52,18 @@ export class RefresherContent implements ComponentInterface {
 
   componentWillLoad() {
     if (this.pullingIcon === undefined) {
-      const mode = getIonMode(this);
+      const platform = getIonPlatform(this);
       const overflowRefresher = (this.el.style as any).webkitOverflowScrolling !== undefined ? 'lines' : arrowDown;
       this.pullingIcon = config.get(
         'refreshingIcon',
-        mode === 'ios' && isPlatform('mobile') ? config.get('spinner', overflowRefresher) : 'circular'
+        platform === 'ios' && isPlatform('mobile') ? config.get('spinner', overflowRefresher) : 'circular'
```
amandaejohnston, line 59: Optional: I wonder if the `isPlatform` util should be renamed? It looks confusing to use it and `getIonPlatform` at the same time like this, since the two are actually referring to different things.

core/src/components/router-outlet/route-outlet.tsx
```diff
@@ -41,7 +41,7 @@ export class RouterOutlet implements ComponentInterface, NavOutlet {
   /**
    * The mode determines which platform styles to use.
    */
-  @Prop({ mutable: true }) mode = getIonMode(this);
+  @Prop({ mutable: true }) mode = getIonPlatform(this);
```
amandaejohnston, line 44: Does setting the mode on a `router-outlet` also affect any components rendered as part of the routing? If so, this doesn't seem correct since we'd be mixing the behaviors of `mode` and `platform`. For example, if the `platform` is globally set to `ios` and the `mode` globally set to `base`, this would result in child components using `ios` styling as well as behavior, instead of base styling.

liamdebeasi: Yes, `mode` is inherited to child components.

`mode` controls the look and the feel of an application. `baseComponents`/`useBase` overrides the look that is set by default with `mode`. In other words, it takes priority over whatever is set using `mode` even if you set the `mode` directly on a component.

For example, `<ion-button mode="md" use-base="true">` would remove any MD styles, even though the mode is explicitly being set here. Doing it this way gives the developers control over both the look and the feel without being able to mix ineligible values. For example, the developer could never apply the iOS look to a component running with MD capabilities.

The same is true for `ion-router-outlet` here. `getIonPlatform` will default to whatever is set in the global `mode` (`'ios'` or `'md'`). However, if you have `use-base="true"` set on a child component, that component will not load ios/md stylesheets.

core/src/components/refresher-content/refresher-content.tsx
liamdebeasi: I'd prefer not to name `isPlatform` since that is a public API. What if we changed `getIonPlatform` to `getIonCapability`? 

IMO our internal usage of the term "mode" is a bit loaded now. It might make sense to do the following:

`getIonMode` --> `getIonStyle`: This returns the visual style that a component is running in.
`getIonPlatform` --> `getIonCapability`: This returns the platform that a component is running in.

liamdebeasi: Actually `getIonStyle` may collide with that internal `ionStyle` event we use in some components. Maybe `getIonMode` --> `getIonVisual` or `getIonStylesheet`?

amandaejohnston: `getIonStylesheet` sounds great 👍 For `getIonPlatform`, maybe something like `getIonBehaviors`? I feel like that better captures how we're cherry-picking certain pieces of functionality. Super bikeshedding though, `getIonCapability` also works for me.

core/src/components/router-outlet/route-outlet.tsx
amandaejohnston: Ahh, got it, I'd gotten myself turned around with how the API works. Thanks for explaining 👍

amandaejohnston (approved):
LGTM aside from the function name stuff which I don't see as blocking.

PR #26663, opened 2023-01-23
chore(deps): update to stencil v3

liamdebeasi:
## Pull request checklist

## Pull request type

 

 

- [ ] Bugfix
- [ ] Feature
- [x] Build related changes

Ionic Framework is using Stencil 2

## What is the new behavior?

- Updates core to using Stencil v3
- Disabled sourceMaps until the related stencil bug can be resolved. (We are not generating sourcemaps in Ionic v6, so there should be no behavior change)
- Updated sync scripts to account for dependencies (see https://github.com/ionic-team/ionic-framework/pull/26686)
- Refactor so internal-only interfaces are imported from the files directly instead of the `interfaces.d.ts` file
- Updates `interfaces.d.ts` file so only public interfaces are explicitly exported
- Updated the output target dependencies to account for the Stencil v3 dep

- [ ] Yes
- [x] No

Note: I discovered that we had a dozen or so private interfaces exported as a result of how the `interfaces.d.ts` file is structured in `main`. This PR makes it so those interfaces are no longer exported. However, I am not considering this a breaking change since these are private APIs and should not be used in external apps.


core/src/components/datetime/datetime-interface.ts
```diff
@@ -1,7 +1,3 @@
-export interface DatetimeOptions {
```
liamdebeasi, line 1: This is not used anywhere

core/src/components/spinner/spinner-interface.ts
```diff
@@ -1,5 +1,3 @@
-export { SpinnerTypes } from './spinner-configs';
```
liamdebeasi, line 1: SpinnerTypes is already exported by Stencil

core/stencil.config.ts
```diff
@@ -9,6 +9,7 @@ import { apiSpecGenerator } from './scripts/api-spec-generator';
 
 export const config: Config = {
   autoprefixCss: true,
+  sourceMap: false,
```
liamdebeasi, line 12: There is a bug in stencil that causes sourcemap generation to fail in Ionic. The stencil team is aware of this and they are going to fix it.

core/src/components/button/button.tsx
```diff
@@ -2,12 +2,15 @@ import type { ComponentInterface, EventEmitter } from '@stencil/core';
...
-import type { AnimationBuilder, Color, RouterDirection } from '../../interface';
+import type { AnimationBuilder, Color } from '../../interface';
 import type { AnchorInterface, ButtonInterface } from '../../utils/element-interface';
 import type { Attributes } from '../../utils/helpers';
 import { inheritAriaAttributes, hasShadowDom } from '../../utils/helpers';
 import { printIonWarning } from '../../utils/logging';
 import { createColorClasses, hostContext, openURL } from '../../utils/theme';
+import type { RouterDirection } from '../router/utils/interface';
+
+0;
```
amandaejohnston, line 13: I'm guessing this `0` slipped in by mistake? 😆

amandaejohnston (approved):
Great work! Confirmed that I'm still able to build and run core on my Windows machine too 👍

core/src/components/button/button.tsx
liamdebeasi: 👀 yes

angular/src/directives/angular-component-lib/utils.ts
```diff
@@ -4,46 +4,40 @@ import { fromEvent } from 'rxjs';
 
 export const proxyInputs = (Cmp: any, inputs: string[]) => {
   const Prototype = Cmp.prototype;
-  inputs.forEach(item => {
+  inputs.forEach((item) => {
```
rwaskiewicz, line 7: Double checking....should these format changes be in this PR? I didn't see a callout about it in the summary || any prettier changes in `package.json`

liamdebeasi: Good catch -- I added this to the original PR. I updated the output target deps, and it looks like there were some formatting changes: https://github.com/ionic-team/stencil-ds-output-targets/blob/f20a7780bf6416052e338fd31bd1a813477f9792/packages/angular-output-target/angular-component-lib/utils.ts#L7

rwaskiewicz (approved):

sean-perkins (approved):
These changes look good to me. I had the same question as Ryan, regarding the `@utils` alias import paths for the animations file.

liamdebeasi:
Good catch. I reverted those changes in https://github.com/ionic-team/ionic-framework/pull/26663/commits/45c4ff49c9c85b9d85ae74a99fd07cbe0d7d75ac. Also sorry to the person I unintentionally tagged in my commit message 😅.

core/src/utils/menu-controller/animations/base.ts
```diff
@@ -1,5 +1,5 @@
-import type { Animation } from '../../../interface';
-import { createAnimation } from '../../animation/animation';
+import { createAnimation } from '@utils/animation/animation';
```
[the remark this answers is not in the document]
liamdebeasi: Good catch - reverted that https://github.com/ionic-team/ionic-framework/pull/26663/commits/45c4ff49c9c85b9d85ae74a99fd07cbe0d7d75ac

PR #26670, opened 2023-01-24
fix(select): modern component takes up full line

liamdebeasi:
## Pull request checklist

## Pull request type

 

 

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

While working on the documentation I discovered that the select was missing `width: 100%` which allows it to take up the full line. This was due to old code that only applied with `.in-item`. 

## What is the new behavior?

- Updated select to align with the new `ion-input` code to apply `width: 100%` and remove the `.in-item` code. (The height: 100% should not be needed when inside of ion-item)
- I also added TODOs to remove the legacy syntax

- [ ] Yes
- [x] No

### DIFF
--- a/core/src/components/select/select.ios.scss
+++ b/core/src/components/select/select.ios.scss
@@ -4,6 +4,7 @@
 // iOS Select
 // --------------------------------------------------
 
+// TODO FW-3194 - Remove this
 :host(.legacy-select) {
   --padding-top: #{$select-ios-padding-top};
   --padding-end: #{$select-ios-padding-end};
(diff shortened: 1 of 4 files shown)


sean-perkins (approved):
This looks good, toggle appears to have a similar implementation: https://github.com/ionic-team/ionic-framework/blob/feature-7.0/core/src/components/toggle/toggle.scss#L41-L44 does it also need this change?

liamdebeasi:
Possibly? Toggle, Checkbox, and Radio currently only take up as much space as they need. I could be convinced to keep that behavior consistent. Thoughts?

sean-perkins:
Good point - let's leave it as-is for now. `ion-select` makes practical sense to be the full line. For toggle/checkbox/radio I don't think that is always the case.

Issue #26643, opened 2023-01-19
bug: ion-select with interface="popover" does not open popover


martinmitteregger:
### Prerequisites

- [X] I have read the [Contributing Guidelines](https://github.com/ionic-team/ionic-framework/blob/main/.github/CONTRIBUTING.md#creating-an-issue).
- [X] I agree to follow the [Code of Conduct](https://ionicframework.com/code-of-conduct).
- [X] I have searched for [existing issues](https://github.com/ionic-team/ionic-framework/issues) that already report this problem, without success.

### Ionic Framework Version

- [ ] v4.x
- [ ] v5.x
- [X] v6.x
- [ ] Nightly

### Current Behavior

After clicking onto an `ion-select` with `interface="popover"` the backdrop appears but no popover is shown.
The console shows the error or rather warning "_Found a 'popover' attribute with an invalid value._" 

![current_behavior](https://user-images.githubusercontent.com/15977182/213466996-cb65207f-2df1-4734-b877-377e63485c8b.png)

### Expected Behavior

After cclicking onto an `ion-select` with `interface="popover"` the backdrop AND the popover itself should be shown.

### Steps to Reproduce

Just run the stackblitz example from the [docs](https://ionicframework.com/docs/api/select#popover).
After opening the stackblitz example the docs show the same incorrect behavior!

![image](https://user-images.githubusercontent.com/15977182/213467819-b788a39b-250a-4844-b511-06d9bbe403ac.png)

### Ionic Info

I can't run `ionic info` in the stackblitz project but as I have the same error in my project here's my output

Ionic:
   Ionic CLI                     : 6.20.1
   Ionic Framework               : @ionic/angular 6.5.0
   @angular-devkit/build-angular : 15.1.1
   @angular-devkit/schematics    : 15.1.1
   @angular/cli                  : 15.1.1
   @ionic/angular-toolkit        : 7.0.0

Capacitor:
   Capacitor CLI      : 4.6.1
   @capacitor/android : 4.6.1
   @capacitor/core    : 4.6.1
   @capacitor/ios     : 4.6.1

Utility:
   cordova-res : 0.15.3
   native-run  : 1.7.1

System:
   NodeJS : v16.14.2
   npm    : 8.7.0
   OS     : Windows 10

### Additional Information

After removing `interface="popover"` it works as expected (but not as popover)

liamdebeasi:
[...]

PR #26694, opened 2023-01-27
fix(vue): cache attached view reference

sean-perkins:
## Pull request checklist

## Pull request type

 

 

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

Developers using Vue can run into an error when using an `ion-select` that presents a popover of options. The Vue delegate was incorrectly assuming that a component instance would always be passed as the argument to `attachViewToDom`.

This is not the case for `ion-select`, where a string of the tag name is used: https://github.com/ionic-team/ionic-framework/blob/main/core/src/components/select/select.tsx#L414

Issue URL: https://github.com/ionic-team/ionic-framework/issues/26643#issuecomment-1406611223
resolves https://github.com/ionic-team/ionic-framework/issues/26695

## What is the new behavior?

- Caches the reference to the created element, so the `WeakMap` always receives an object to that can be used for looking up and unmounting an overlay.
- Refactored the property to better signal that it can be an "object" (`any`) or a string. Note: Because any will overwrite the types, this does nothing from a typings perspective. This is just a visual signal to a developer. 

- [ ] Yes
- [x] No

## Other information

Repro: https://github.com/acoldfox/ionic-select-issue (you will need to install the dev-build to verify)

Dev-build: `6.5.2-dev.11674852369.1c9fe737`

### DIFF
--- a/packages/vue/src/framework-delegate.ts
+++ b/packages/vue/src/framework-delegate.ts
@@ -17,7 +17,7 @@ export const VueDelegate = (
   // TODO(FW-2969): types
   const attachViewToDom = (
     parentElement: HTMLElement,
-    component: any,
+    componentOrTagName: any | string,
     componentProps: any = {},
     classes?: string[]
   ) => {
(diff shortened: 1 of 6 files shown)


liamdebeasi (changes requested):
We need a test for this. Also why does this issue not impact Ionic React?

liamdebeasi:
Also be sure to note that this fixes https://github.com/ionic-team/ionic-framework/issues/26695 when merging

liamdebeasi (approved):
Good to go once we have a test. Thanks for taking care of this!

sean-perkins:
Ok - test case has been added.

Also confirmed against the original root issue: https://github.com/raymer/ionic-modal-bug that the expected behavior still occurs after this change.

PR #26713, opened 2023-02-01
chore(ci): enable fail-fast for matrix jobs

liamdebeasi:
Screenshot matrix jobs had "fail-fast: false". This enabled each test runner to provide its `playwright-report` directory as an artifact so team members can view test failures. At the time, all tests ran even if previous tests failed.

In https://github.com/ionic-team/ionic-framework/pull/26447, we updated the Playwright config to have tests fail fast within a single test runner. However, other test runners continue to run. This is not helpful in the case of failing tests because the artifacts for each test runner will not be reported until all test runners are done.

This PR changes the matrix strategy to enable fail-fast (which is the default) so in-progress/queued test runners fail if a previous test runner fails.

## Pull request checklist

## Pull request type

 

 

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes

Issue URL: 

## What is the new behavior?

-
-
-

- [ ] Yes
- [ ] No


amandaejohnston:
Wouldn't this mean that if one test fails, we may not see the results for the others even if there are other failures? I.e. we'd have to run everything multiple times to catch all the failing tests.

liamdebeasi:
Yes, if there are multiple failures you will potentially not see every error the first time. However, this is true today since individual test runners fail fast. There are a few reasons why I think the proposed behavior is desirable:

1. Developers should be running tests locally before pushing their changes. (i.e. bad tests that repro locally should be caught before pushing to the repo)
2. Test runners run the same test multiple times (different configs: ios, md, ltr, rtl, etc) so the same failure can pop up on multiple test runners. Waiting for those subsequent failures does not give us additional information.
3. GitHub does not let us access the artifacts for each test runner in a matrix until all test runners are done (at least as far as I can tell). This means that even though a test runner fails fast, you still need to wait for all the runners to complete to download the `playwright-report` directory.


I'm trying to find ways to get CI feedback to team members quickly, so I'm open to other ideas too. I suppose where this could get tricky is when you have multiple distinct unintentional regressions. However, my experience has been that item 2 above happens more frequently than multiple distinct unintentional regressions.

amandaejohnston (approved):
Makes sense to me. Plus we can always revert the change if it causes problems.

liamdebeasi:
Adding Sean as a reviewer so I can get buy-in from the entire team.

sean-perkins (approved):
I'm in favor of trying out this change. In many instances I find the same test failing across multiple jobs, so tying up CI for the remainder of the run is not beneficial.

liamdebeasi:
Blocked on https://github.com/ionic-team/ionic-framework/pull/26731. This resolves a handful of the more common flaky tests, so would like to get this in first.

Issue #16289, opened 2018-11-10
bug: size attribute of ion-button ignored when placed in ion-buttons


michaelcamper:
# Bug Report

**Ionic Info**
Run `ionic info` from a terminal/cmd prompt and paste the output below.

```
Ionic:

   ionic (Ionic CLI)             : 4.3.1 (/usr/local/lib/node_modules/ionic)
   Ionic Framework               : @ionic/angular 4.0.0-beta.15
   @angular-devkit/build-angular : 0.8.7
   @angular-devkit/schematics    : 0.8.7
   @angular/cli                  : 6.2.7
   @ionic/angular-toolkit        : 1.1.0

Cordova:

   cordova (Cordova CLI) : 8.1.1 (cordova-lib@8.1.0)
   Cordova Platforms     : not available
   Cordova Plugins       : not available

System:

   ios-deploy : 2.0.0
   NodeJS     : v8.12.0 (/usr/local/bin/node)
   npm        : 6.4.1
   OS         : macOS
   Xcode      : Xcode 10.1 Build version 10B61
```

**Describe the Bug**
The `size` attribute of an `ion-button` is ignored if it's placed inside `ion-buttons`

**Steps to Reproduce**
1. Place `ion-button` inside `ion-buttons`
2. Set `size` to `large`
3. Button size is unchanged

**Related Code**
```
    <ion-toolbar>

      <ion-buttons slot="start">

        <ion-button fill="clear" size="large" slot="icon-only">
          <ion-icon name="list"></ion-icon>
        </ion-button>

        <ion-button fill="clear" slot="icon-only">
          <ion-icon name="list"></ion-icon>
        </ion-button>
      </ion-buttons>

    </ion-toolbar>

    <ion-button fill="clear" size="large" slot="icon-only">
      <ion-icon name="list"></ion-icon>
    </ion-button>

    <ion-button fill="clear" slot="icon-only">
      <ion-icon name="list"></ion-icon>
    </ion-button>

```

**Expected Behavior**
Button should be large

**Additional Context**
![localhost_8100_home iphone 6_7_8 2](https://user-images.githubusercontent.com/39185098/48300657-be859480-e4e1-11e8-8c4a-815548837f5f.png)

liamdebeasi:
Thanks for the issue. This has been resolved via https://github.com/ionic-team/ionic-framework/pull/26726, and a fix will be available in an upcoming release of Ionic Framework.

PR #26726, opened 2023-02-03
fix(button): size and strong are respected in ion-buttons

liamdebeasi:
## Pull request checklist

## Pull request type

 

 

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

Issue URL: resolves https://github.com/ionic-team/ionic-framework/issues/16289

The `size` and `strong` properties on `ion-button` are not respected when placed inside `ion-buttons`. This is because Ionic manually sets the font-size/font-weight on all slotted buttons: 

https://github.com/ionic-team/ionic-framework/blob/9d6ec2925cb2314d5379b864aef467b34afbb318/core/src/components/buttons/buttons.ios.scss#L16-L17
https://github.com/ionic-team/ionic-framework/blob/9d6ec2925cb2314d5379b864aef467b34afbb318/core/src/components/buttons/buttons.md.scss#L18-L19

This was likely done to ensure buttons in a toolbar have consistent container sizes (since we manually set the height as well). However, this breaks the `size`/`strong` functionality on `ion-button`.

However, default size iOS buttons have a slightly increased font size when used in a toolbar. This difference needs to be preserved while allowing for `size` functionality:

| button in toolbar ("Cancel") | button outside of toolbar ("Continue") |
| - | - |
| IMG_1835 | IMG_1836 |

## What is the new behavior?

- Removed the `font-size` and `font-weight` declarations in `ion-buttons` in MD mode. These values are used  inside the `ion-button` stylesheets, so this change should not impact buttons that do not make use of `size`/`strong`. Default size /weight MD buttons have the same font size and weight regardless of usage.
- Removed the `font-size` and `font-weight` declaration in `ion-buttons` in iOS mode. Instead, I added a `.in-buttons` selector to the iOS button stylesheet that applies the increased font size/weight. This approach allows the `size` and `strong` properties to override the default font size and font weight, respectively. The alternative was to use `:not` to create a specific selector that excluded larger/smaller `size` values, but I felt that would make customizing the button hard.
- I also refactored the `size` and `strong` tests while I was in this code. These screenshots had quite a bit of whitespace and tested configurations that did not test the actual functionality.

I opted to keep the `height` property here because this allows the font sizes to increase while keeping the container sizes consistent. If we removed the `height`, then the container sizes would be different which I think is incorrect:

| with height | without height |
| - | - |
| image |image |

- [ ] Yes
- [x] No


core/src/components/button/test/size/button.e2e.ts
```diff
@@ -2,11 +2,53 @@ import { expect } from '@playwright/test';
...
+  test('should render large buttons', async ({ page }) => {
+    await page.setContent(`
+      <ion-button size="large" fill="solid">Large Button</ion-button>
+    `);
+
+    const wrapper = page.locator('ion-button');
+
+    expect(await wrapper.screenshot()).toMatchSnapshot(`button-size-large-${page.getSnapshotSettings()}.png`);
+  });
+  test.describe('in ion-button', () => {
```
amandaejohnston, line 26:
```suggestion
  test.describe('in ion-buttons', () => {
```

core/src/components/button/test/size/button.e2e.ts
```diff
@@ -2,11 +2,53 @@ import { expect } from '@playwright/test';
...
+      const wrapper = page.locator('ion-button');
+
+      expect(await wrapper.screenshot()).toMatchSnapshot(
+        `button-size-small-in-buttons-${page.getSnapshotSettings()}.png`
+      );
+    });
+    test('should render large button', async ({ page }) => {
+      await page.setContent(`
+        <ion-buttons>
+          <ion-button size="large" fill="solid">Small Button</ion-button>
```
amandaejohnston, line 43:
```suggestion
          <ion-button size="large" fill="solid">Large Button</ion-button>
```

amandaejohnston (approved):
LGTM aside from a few typos.

PR #26717, opened 2023-02-01
fix(datetime): add aria-labels to interactive buttons

sean-perkins:
## Pull request checklist

Please check if your PR fulfills the following requirements:
- [x] Tests for the changes have been added (for bug fixes / features)
- [ ] Docs have been reviewed and added / updated if needed (for bug fixes / features)
  - Some docs updates need to be made in the `ionic-docs` repo, in a separate PR. See the [contributing guide](https://github.com/ionic-team/ionic-framework/blob/main/.github/CONTRIBUTING.md#modifying-documentation) for details.
- [x] Build (`npm run build`) was run locally and any changes were pushed
- [x] Lint (`npm run lint`) has passed locally and any fixes were made for failures

## Pull request type

 

 

Please check the type of change your PR introduces:
- [x] Bugfix
- [ ] Feature
- [ ] Code style update (formatting, renaming)
- [ ] Refactoring (no functional changes, no api changes)
- [ ] Build related changes
- [ ] Documentation content changes
- [ ] Other (please describe): 

## What is the current behavior?

Screen readers do not announce to show/hide the year picker when selecting the month/year toggle button. 

Issue URL: Internal

## What is the new behavior?

- Aria label has been added to the month/year toggle of the calendar for announcing "Show/Hide year picker" along with the current month/year
  - This aligns with how screen readers announce the iOS date picker
- Aria labels have been capitalized to align with other labels
- Screen readers announce show/hide year picker along with the current month/year

## Does this introduce a breaking change?

- [ ] Yes
- [x] No

PR #26744, opened 2023-02-07
refactor(datetime): button for calendar month/year toggle

sean-perkins:
## Pull request checklist

## Pull request type

 

 

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes

`ion-datetime` uses the `ion-item` component for rendering the button for the calendar month/year toggle. 

Split from PR: #26717

## What is the new behavior?

- Replaces the usage of `ion-item` with a `button`

There is an expected minor visual diff from these changes, as the `ion-button` has different placement styles for the slot than the `ion-item` usage had.

- [ ] Yes
- [x] No


amandaejohnston:
Why was this change made? Was there an issue caused by using `ion-item` here?

sean-perkins:
`ion-item` is a list item, intended to be used as part of `ion-list`. Using it as a button inside of the datetime header was not the appropriate usage of the component. It would be the equivalent of using:
```html
<li>
  <button>Text</button>
</li>
```

This has a variety of benefits:
- Decouples `ion-datetime` from `ion-item` 
- Removes a component layer from the datetime header
- Reduces the JS cost of the component 
- Reduces the stylesheet cost of the component (applications only using `ion-datetime` do not eat all the extra styles of `ion-item`).

amandaejohnston (approved):

sean-perkins:
Had some offline discussion with Amanda, going to experiment with using `ion-button` with the clear fill here. It has a lot of the design responsibility already baked in and may be able to get us the same outcome with just CSS variable overrides.

liamdebeasi:
Will this impact the dynamic aria-label behavior we want to implement in https://github.com/ionic-team/ionic-framework/pull/26717? The inherited aria attributes are not reactive: https://github.com/ionic-team/ionic-framework/blob/main/core/src/components/button/button.tsx#L150

sean-perkins:
> Will this impact the dynamic aria-label behavior we want to implement in #26717? The inherited aria attributes are not reactive: https://github.com/ionic-team/ionic-framework/blob/main/core/src/components/button/button.tsx#L150

I've confirmed even with `ion-item`, if we use a `key` attribute on the rendered interactive element, where the key is the aria-label, it will force a new instance of the component to be rendered with the correct `aria-label` applied. This would be "tech debt" until we can collaborate with the Stencil team on adding support for allowing reserved properties to be used as `@Prop()` or `@Watch()` decorated fields.

sean-perkins:
I need to rework the PR title and description, but the contents of the PR have been updated to use `ion-button` here. This allows us to add a hover style to the iOS display of the month/year toggle and accomplishes the same visual design as `ion-item`, without having to port a bunch of duplicate styles across the stylesheets. 

I'm going to re-request everyone for a review, since the contents are completely different. Appreciate any testing/verification for the display, active/focus states, etc.

core/src/components/datetime/datetime.scss
```diff
@@ -423,6 +422,6 @@ ion-picker-column-internal {
 
 // Year Picker
 // -----------------------------------
-:host(.show-month-and-year) .calendar-action-buttons ion-item {
-  --color: #{current-color(base)};
+:host(.show-month-and-year) .calendar-action-buttons .calendar-month-year-button {
```
liamdebeasi, line 425: Looks like there still is a `--color` variable on `ion-button`: https://ionicframework.com/docs/api/button#css-custom-properties-1

Should we keep that usage?

core/src/components/datetime/datetime.tsx
```diff
@@ -1890,17 +1890,17 @@ export class Datetime implements ComponentInterface {
...
-                {getMonthAndYear(this.locale, this.workingParts)}
-                <ion-icon
-                  aria-hidden="true"
-                  icon={this.showMonthAndYear ? expandedIcon : collapsedIcon}
-                  lazy={false}
-                  flipRtl={true}
-                ></ion-icon>
-              </ion-label>
-            </ion-item>
+            <ion-button class="calendar-month-year-button" fill="clear" onClick={() => this.toggleMonthAndYearView()}>
```
liamdebeasi, line 1893: The button is no longer aligned with the next/prev buttons:

Left is `ion-button` right is `ion-item`:

image

liamdebeasi (changes requested):

core/src/components/datetime/datetime.scss
sean-perkins: The `text-inherit` mixin resets the `color` style property to `inherit`. I'll revisit what style inherits are actually needed for this element and then move away from the mixin so we can use the `--color` CSS variable.

sean-perkins:
Closing in favor of: https://github.com/ionic-team/ionic-framework/pull/26848, I'm going to keep the branch around to keep making progress on the refactor, but this won't land in the current sprint.

Issue #25832, opened 2022-08-26
feat: Dot indicator for individual calendar day in Datetime


sean-perkins:
### Prerequisites

- [X] I have read the [Contributing Guidelines](https://github.com/ionic-team/ionic-framework/blob/main/.github/CONTRIBUTING.md#creating-an-issue).
- [X] I agree to follow the [Code of Conduct](https://ionicframework.com/code-of-conduct).
- [X] I have searched for [existing issues](https://github.com/ionic-team/ionic-framework/issues) that already include this feature request, without success.

### Describe the Feature Request

The datetime component should include options to optionally render a dot indicator for specific calendar days. To align with iOS design, at most one dot indicator can display at a time for a single calendar day. Displaying a dot should only be available for `presentation="date"` and `presentation="date-time"`. 

Dot indicators are not theme-able in the iOS interface, but I think it makes sense to allow developers to customize the appearance (background color).

 

### Describe the Use Case

The iOS design for the calendar grid commonly shows a dot indicator on a calendar day to indicate either an event or additional information is available for a specific day.

<img src="https://user-images.githubusercontent.com/13732623/186808479-e2afb878-3f94-40bd-b6c4-3943f09391f5.jpeg" width="300px" />
 
Other libraries such as V-Calendar support this API: https://vcalendar.io/attributes.html#dots

### Describe Preferred Solution

Datetime should accept a function that allows developers to control when a dot indicator is displayed.

```html
<ion-datetime></ion-datetime>

<script>
  const datetime = document.querySelector('ion-datetime');
  datetime.displayDotForDate = (dateIsoString) => {
    const date = new Date(dateIsoString);
    // check for a specific date
    return true;
  }
</script>

```

The default behavior would be to not render a dot for a calendar day.

Developers can customize the appearance of the dot through CSS variables:
```css
ion-datetime {
  --dot-background: blue;
}
```

NeluQi:
any news?

sean-perkins:
Hello everyone 👋 

The team would love to hear your use cases for this feature. While the original request shows the iOS Events interface, we'd like to better understand what features you would build in your applications if Ionic added dot indicators.

Would you need additional APIs, such as knowing when a month changed to load more dots? Would you expect this to be part of `ion-datetime` or a new component (`ion-calendar`)? Do you expect it to visually match iOS or just need the ability to render a dot in the calendar day and don't really care if it is visually identical to iOS?

Feedback is appreciated, thanks!

[...]

PR #26383, opened 2022-11-30
chore(many): remove redundant :host selector from styles

sean-perkins:
## Pull request checklist

Please check if your PR fulfills the following requirements:
- [ ] Tests for the changes have been added (for bug fixes / features)
- [ ] Docs have been reviewed and added / updated if needed (for bug fixes / features)
  - Some docs updates need to be made in the `ionic-docs` repo, in a separate PR. See the [contributing guide](https://github.com/ionic-team/ionic-framework/blob/main/.github/CONTRIBUTING.md#modifying-documentation) for details.
- [x] Build (`npm run build`) was run locally and any changes were pushed
- [x] Lint (`npm run lint`) has passed locally and any fixes were made for failures

## Pull request type

 

 

Please check the type of change your PR introduces:
- [ ] Bugfix
- [ ] Feature
- [ ] Code style update (formatting, renaming)
- [x] Refactoring (no functional changes, no api changes)
- [ ] Build related changes
- [ ] Documentation content changes
- [ ] Other (please describe): 

## What is the current behavior?

Components using the shadow DOM do not need to specify `:host` when targeting a child selector. `ion-datetime`, `ion-datetime-button`, `ion-picker-internal` and `ion-picker-column-internal` have redundant usage of `:host .element-selector`.

Issue URL: Relates to #25285

## What is the new behavior?

- Removes the `:host` selector in usages that target an explicit child selector in `ion-datetime`, `ion-datetime-button`, `ion-picker-internal` and `ion-picker-column-internal`

## Does this introduce a breaking change?

- [ ] Yes
- [x] No

## Other information

This PR blocks #25264

Issue #26361, opened 2022-11-25
feat: display customized notes for dates in component "<IonDatetime />", e.g. holiday


creatxrgithub:
### Prerequisites

- [X] I have read the [Contributing Guidelines](https://github.com/ionic-team/ionic-framework/blob/main/.github/CONTRIBUTING.md#creating-an-issue).
- [X] I agree to follow the [Code of Conduct](https://ionicframework.com/code-of-conduct).
- [X] I have searched for [existing issues](https://github.com/ionic-team/ionic-framework/issues) that already include this feature request, without success.

### Describe the Feature Request

I wish that I can add some notes to the component "<IonDatetime />".

e.g. specify a holiday and display it in the calendar by a function

### Describe the Use Case

to display holiday or lunar date in calendar

### Describe Preferred Solution

by setting a function or a array or a key-value (date-notes) object in a attribute of "<IonDatetime />"

### Related Code

```
let notes = { "2022-11-25": "初二", "2022-11-26": "one's birthday"}
<IonDatetime notes={notes} />
```
![IMG_20221125_142305](https://user-images.githubusercontent.com/27437178/204044668-6ce6e50b-df04-43fd-b268-af5fe6ef6b33.jpg)

sean-perkins:
Hello @creatxrgithub, thank you for this feature request!

The individual date cell is a pretty small space to render additional UI within. We have a feature to track rendering a dot indicator in each cell, that developers could use for indicating additional information (https://github.com/ionic-team/ionic-framework/issues/25832).

Does that feature request solve your core need or are you explicitly looking to render text within the date cells?

creatxrgithub:
It's similar, but not really.
I wish to display some notes at the "dot" position.
His code:
```
<ion-datetime></ion-datetime>

<script>
  const datetime = document.querySelector('ion-datetime');
  datetime.displayDotForDate = (dateIsoString) => {
    const date = new Date(dateIsoString);
    // check for a specific date
    return true;
  }
</script>
```
could be:
```
<ion-datetime></ion-datetime>

<script>
  const datetime = document.querySelector('ion-datetime');
  datetime.displayDotForDate = (dateIsoString) => {
    const date = new Date(dateIsoString);
    // check for a specific date
    return dotNotes; //or empty string ""
  }
</script>
```
or like this:
```
<IonDatetime displayDotForDate={ (dateIsoString) => { return dotNotes || "" }}/>
```
thanks.



[...]

PR #26775, opened 2023-02-09
feat(datetime): add ability to specify custom colors for specific dates

amandaejohnston:
## Pull request checklist

## Pull request type

 

 

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes

Developers have no way of highlighting specific days on the calendar picker with custom styles.

Issue URL: Resolves https://github.com/ionic-team/ionic-framework/issues/25832

## What is the new behavior?

Adds a new property, `highlightedDates`, which allows certain dates in the calendar view to be styled with custom text and/or background colors. This can be in the form of either an array of specific dates and their colors, or a callback that takes an ISO string and returns the colors to use.

As a side effect, all styles around focused/selected/"today" dates have been consolidated into the new `.calendar-day-highlight` element. A new element was necessary in order to programmatically apply background colors, since inline styles can't be used with pseudo-elements.

Note that this feature does not include the ability to show details about events associated with highlighted dates, such as birthdays or scheduled meetings. This is intentional as `ion-datetime` is only intended to concern itself with date selection, and any additional context should be an application-level detail provided in a separate UI.

Playgrounds/docs updates are made in a separate PR: https://github.com/ionic-team/ionic-docs/pull/2769

- [ ] Yes
- [x] No


core/src/components/datetime/datetime.scss
```diff
@@ -357,6 +357,11 @@ ion-picker-column-internal {
 }
 
 :host .calendar-day:after {
+  content: " ";
+}
+
+:host .calendar-day:after,
```
sean-perkins, line 363: ~ Nit: Can we remove the `:host` style selectors from these additions? We have an open PR/task to remove it from the existing implementation (https://github.com/ionic-team/ionic-framework/pull/26383), but there are some odd difs that need to be worked through in that PR. 

`:host` shouldn't be necessary since we are encapsulated within the shadow dom here.

core/src/components/datetime/datetime.tsx
```diff
@@ -2001,6 +2029,37 @@ export class Datetime implements ComponentInterface {
...
+                /**
+                 * Wrap in a try-catch to prevent exceptions in the user's function
+                 * from interrupting the calendar's rendering.
+                 */
+                try {
+                  dateStyle = highlightedDates(dateIsoString);
+                } catch (e) {
+                  printIonError(
+                    'Exception thrown from provided `highlightedDates` callback. Please check your function and try again.',
+                    e
```
sean-perkins, line 2057: Can we additionally log the host element reference with this exception, so they can quickly find the `ion-datetime` instance this came from, if they have multiple rendered on the view?

core/src/components/datetime/datetime.tsx
```diff
@@ -11,6 +11,9 @@ import type {
   Mode,
   StyleEventDetail,
   TitleSelectedDatesFormatter,
+  DatetimeHighlight,
```
sean-perkins, line 14: ~ It might be worth updating these imports to come directly from the `datetime-interface` file located within the same folder vs. the barrel file. Otherwise we will need to manually fix this when syncing `main` with `feature-7.0`. Stencil v3 requires that we import from relative paths instead of the type declaration file that we share with consumers.

sean-perkins (commented):
Small recommended changes, but this feature looks great testing locally. Nice work!

amandaejohnston:
Docs PR created: https://github.com/ionic-team/ionic-docs/pull/2769

core/src/components/datetime/datetime.tsx
```diff
@@ -2001,6 +2029,37 @@ export class Datetime implements ComponentInterface {
...
 
+            let dateStyle: DatetimeHighlightStyle | undefined = undefined;
+
+            /**
+             * Custom highlight styles should not override the style for selected dates,
+             * nor apply to "filler days" at the start of the grid.
+             */
+            if (highlightedDates !== undefined && !isActive && day !== null) {
+              if (Array.isArray(highlightedDates)) {
+                const matchingHighlight = highlightedDates.find((hd) => hd.date === dateIsoString);
```
liamdebeasi, line 2040: We might want to consider removing time values from the ISO 8601 string. For example, this is a valid input for the `date` key: `2023-02-01T16:30`. However, we do a strict equality with the ISO 8601 string for just the date, so the example date I passed would never match anything.

liamdebeasi (changes requested):

core/src/components/datetime/datetime.tsx
```diff
@@ -2050,6 +2114,14 @@ export class Datetime implements ComponentInterface {
...
                 }}
               >
+                {dateStyle && (
+                  <div
+                    class="calendar-day-event-highlight"
+                    style={{
+                      backgroundColor: dateStyle.backgroundColor,
+                    }}
+                  ></div>
+                )}
```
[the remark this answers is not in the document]
amandaejohnston: Using both the highlight and the pseudo-element allows the existing focus styles to work in tandem with the highlight with no additional effort, as well as the outline from the "today" style on `md` mode.

core/src/components/datetime/datetime.scss
```diff
@@ -357,6 +357,11 @@ ion-picker-column-internal {
...
+:host .calendar-day-event-highlight {
```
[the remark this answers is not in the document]
amandaejohnston: Do you mean change the color of the focus style to match the highlight's BG color? I'm hesitant to do that because then the focus highlight would no longer always match the color of a selected date. (And the design doc says that the highlight shouldn't override the selected date styles.) Having the focus highlight change color as you navigate around with the keyboard would also feel strange IMO.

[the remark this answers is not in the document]
liamdebeasi: I'm referring to dates that have the highlight styles applied: 
image

In MD, you can focus a non-selected button. However, it is using the default color palette so the blue focus state ends up mixing with the red background that was specified.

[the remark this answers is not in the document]
liamdebeasi: As opposed to something like this: 
image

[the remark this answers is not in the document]
liamdebeasi: (Notice how the red color in the first image is slightly mixed with the blue)

core/src/components/datetime/datetime.tsx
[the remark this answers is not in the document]
amandaejohnston: On second thought, it looks like your idea would be best. I think I'll still need an `:after` on the highlight itself to make the MD focus styles work, but most everything else can be moved to the highlight.

core/src/components/datetime/datetime.scss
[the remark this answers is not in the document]
amandaejohnston: I've updated it so that the focus state matches the background color of the date (if one is specified) using an opacity of 0.2, on both MD and iOS. I don't love it, though, for a couple reasons:
1. If you use a lighter background color for the highlight, the focus state is hard to see.
2. The color of your "cursor" changes as you navigate around using the keyboard, which feels strange IMO.

We could solve item 1 by letting the developer set the opacity on a per-highlight basis, but that feels overly complicated. I would rather either keep the old behavior (where the focus state is always that light blue, or whatever the base color is set to, and overlays the date), or maybe have the focus color sit behind the highlight BG instead so you just get a light blue ring around it, rather than the colors mixing together. Have a look and let me know what you think.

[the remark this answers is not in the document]
liamdebeasi: I like having the focus ring sit behind the background. Honestly, I don't know why it did not already do that. Maybe an old bug that we're noticing?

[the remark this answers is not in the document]
amandaejohnston: The (default) focus ring and selected date circle are the same color, so before, there wouldn't be any visual difference if it were in front of or behind the background, since the selected date style is 100% opacity anyway.

Either way, I've updated it again; all styles are now contained within the `calendar-day-highlight` element, and the focus style always uses the default light blue, showing as a ring around the highlight in MD mode instead of overlaying it.

Note that in iOS, I added an `!important` to the focus/active background color to overwrite the highlight background (which is inline). The only way I can think of to get around this (while avoiding the focus and highlight colors blending together) would be to use an :after with 100% opacity that covers up the highlight, but then we'd need to plug in a separate color that looks like the default blue at 20% opacity, which means you wouldn't be able to customize it easily. 🤔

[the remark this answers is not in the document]
liamdebeasi: I like having the focus ring behind the element so we can avoid the blending. Thanks for fixing that.

I'm fine having the focus state default to the component's color palette. This wasn't a behavior that was scoped, so I think it would be fine to keep it as you have it and see what the developer feedback is.

core/src/components/datetime/datetime.tsx
```diff
@@ -322,6 +324,17 @@ export class Datetime implements ComponentInterface {
...
+  /**
+   * Used to apply custom text and background colors to specific dates.
+   *
+   * Can be either an array of objects containing ISO strings and colors,
+   * or a callback that receives an ISO string and returns the colors.
+   *
+   * Only applies to the `date`, `date-time`, and `time-date` presentations,
+   * with `preferWheel="false"`.
+   */
+  @Prop() highlightedDates?: DatetimeHighlight[] | DatetimeHighlightCallback;
```
liamdebeasi, line 336: Our API conventions guide states that APIs should be present tense. Could we change this to `highlightDates` or something similar?

core/src/components/datetime/datetime-interface.ts
```diff
@@ -25,3 +25,17 @@ export interface DatetimeParts {
 export type DatetimePresentation = 'date-time' | 'time-date' | 'date' | 'time' | 'month' | 'year' | 'month-year';
 
 export type TitleSelectedDatesFormatter = (selectedDates: string[]) => string;
+
+export type DatetimeHighlightStyle =
+  | {
+      color: string;
```
liamdebeasi, line 31: Thoughts on naming this `textColor`? A couple reasons:

1. Consistency with the `backgroundColor` key
2. I worry that devs may confuse this key with the `color` feature that can be used as a property i.e. `ion-datetime color="tertiary"`.

core/src/components/datetime/datetime.scss
```diff
@@ -383,10 +382,6 @@ ion-picker-column-internal {
   z-index: -1;
 }
 
-.calendar-day-highlight:after {
```
liamdebeasi: Since we are getting rid of the pseudo element, do we still need FW-1720 and the associated hack/workaround above?

liamdebeasi (changes requested):

core/src/components/datetime/datetime.tsx
amandaejohnston: `highlighted` is used as an adjective here, so it's still present tense, plus `highlightDates` sounds like it's the function that actually does the highlighting, like `highlight` as a verb. 🤔 If I'm still missing the point, maybe `datesWithHighlight`...? Or just `highlights`? I also don't mind `notableDates` from the bikeshedding Slack thread, although that might necessitate renaming the types too. (cc @sean-perkins in case you had any thoughts)

core/src/components/datetime/datetime-interface.ts
amandaejohnston: I went with `color` to match the CSS property, but you make a good point re: confusion with the `color` prop on the datetime as a whole. I'll rename this 👍

core/src/components/datetime/datetime.tsx
liamdebeasi: Ah right "present tense" isn't the correct term here. However, our API conventions state that we should avoid this usage. For example, we try to do things like `onPresent` instead of `onPresented` to make things easier to reason about for non-native English speakers. This also aligns with how the browser APIs are named (for example `change` instead of `changed` for `<input />`)

amandaejohnston: Oh I know, I'll go with `dateHighlights` 👀

liamdebeasi: Though thinking about this more... `onPresent` and `change` are events whereas this is a property name. Also we have `isDateEnabled` so maybe I'm overthinking this 😂 . `highlightedDates` is probably fine.

core/src/components/datetime/datetime-interface.ts
amandaejohnston: Done; playgrounds update coming soon.

core/src/components/datetime/datetime.scss
amandaejohnston: Good catch! Removed 👍

liamdebeasi (approved):
Great job!

core/src/components/datetime/datetime.ios.scss
```diff
@@ -92,14 +92,15 @@
   font-size: 20px;
 }
 
-:host .calendar-day:after {
+.calendar-day:focus .calendar-day-highlight,
+.calendar-day.calendar-day-active .calendar-day-highlight {
+  // !important is needed to overwrite custom highlight background, which is inline.
+  /* stylelint-disable-next-line declaration-no-important */
+  background: current-color(base) !important;
```
sean-perkins, line 99: Can you explain why `!important` is needed here?

`current-color(base)` evaluates to: `var(--ion-color-base)`, which has a value of: `var(--ion-color-primary, #3880ff)!important`.

So this style is evaluating as:

```css
background: var(--ion-color-primary, #3880ff)!important !important;
```

I removed the `!important` from your stylesheet and tested against iOS mode, but do not see a visual difference.

sean-perkins (approved):

seatechdev:
Just looked at the new release.  It is amazing!  Thanks for adding this so quickly.

creatxrgithub:
where is its full docs? thanks.

it seems that it has only two properties "textColor" and "backgroundColor". 
```
{
          date: '2023-01-10',
          textColor: '#09721b',
          backgroundColor: '#c8e5d0',
},
```
has it a property to attach text? so that i could do : 
https://github.com/ionic-team/ionic-framework/issues/26361
or 
https://github.com/ionic-team/ionic-framework/issues/25832

ChanakaWeerasinghe:
@averyjohnston @seatechdev @creatxrgithub @sean-perkins @liamdebeasi 

```
its docs: https://ionicframework.com/docs/api/datetime#highlighting-specific-dates

it seems that it has only two properties "textColor" and "backgroundColor".

{
          date: '2023-01-10',
          textColor: '#09721b',
          backgroundColor: '#c8e5d0',
},
is it possible to attach text? so that i could do :
https://github.com/ionic-team/ionic-framework/issues/26361
or
https://github.com/ionic-team/ionic-framework/issues/25832
```

gyoung-tf:
Sorry to come back to this PR after literally years, but I'm interested in understanding why this PR is considered to resolve https://github.com/ionic-team/ionic-framework/issues/25832. Showing dots underneath each date is I think a pretty standardised way to indicate the existence (and sometimes number) of events on each date (used by both iOS and Android calendars), and in my opinion changing the colours of the dates doesn't convey that same information (at least not 'intuitively' i.e in a way that users are already familiar with ). I understand that the philosophy of ion-datetime is more focused on date selection rather than conveying information about the dates, but I think that the existence of events is important information to convey to the user as part of their date selection.

liamdebeasi:
I don't work for Ionic anymore, but I worked on this feature during the planning stages. We wanted a clear separation between date selection (what Ionic has now) and calendar selection (think of `calendar.google.com`). While seemingly similar, a date picker and a calendar picker have diverging feature sets, and we didn't think it would be good to have those two use cases combined into a single component. Additionally, the work required to have a standalone `ion-calendar` component was more than we were willing to take on at the time given the size of the team and the other areas of the project that required our attention.

We saw a date picker as focusing on generic dates (I.e. June 13, 2025) and a calendar picker focusing more on the _context_ of the date (e.g. "I have a dentist appointment on June 13, 2025"). However, Ionic devs wanted some level of context awareness, so this feature was a compromise where users were able to provide context without `ion-datetime` having to be concerned with the specifics. We recognized that it might not be a great fit for every Ionic dev, but in the years since this was released it has proven to be a useful feature for most Ionic devs. Hope this clarifies things!

seatechdev:
I think that I had originally requested this feature i.e. ability to color code dates, because I didn't want to use a third party calendar component.  Later on I determined that I didn't really need an out of box calendar component from Ionic.  Instead, I was able to use the current date time component and modify the presentation layer with ng-if statements.  
Screenshot 2025-06-13 at 10 44 45 AM

PR #23474, opened 2021-06-21
feat(select): allow multiple selections in popover interface, ui updates

brandyscarney:
## Features / Improvements
- Adds support for `multiple` with the `"popover"` interface
- Properly add styles for the label, highlight, dropdown arrow on focus / present
- Removes the ripple effect for outline selects
- Update styles for the popover itself to take up the full width of the item, remove borders, hide the radio buttons themselves, use correct colors for hovering / focusing items
- First radio is focused when presented

resolves https://github.com/ionic-team/ionic-framework/issues/23657
resolves https://github.com/ionic-team/ionic-framework/issues/15500
resolves https://github.com/ionic-team/ionic-framework/issues/12310

| `master` | `branch` |
| ---| ---|
| master | branch |
| master-open | branch-open |

PR #26796, opened 2023-02-14
fix(select): emit single ionChange event for popover option selection

### LINKED ISSUE #26789: bug: ion-select popover triggers duplicate ionChange event
**Describe the bug**
I recognized a weird behavior on ion-select on "popover" interface. The ionChange event gets triggered doubles when i clear the selection.
Same code with different interface like "action-sheet" works as expected tho.

**To reproduce**
1. take the example code from here https://ionicframework.com/docs/api/select#popover
2. change it to the following code (https://stackblitz.com/edit/angular-xobls8?file=src%2Fcomponents%2FExample.vue)
```
<template>
  <div @click="clear()">clear fruits</div>
  <ion-list>
    <ion-item>
      <ion-select
        @ionChange="
          (event) => {
            test(event.detail.value);
          }
        "
        ref="selectEl"
        interface="popover"
        placeholder="Select fruit"
      >
        <IonSelectOption value="apples">Apples</IonSelectOption>
        <IonSelectOption value="oranges">Oranges</IonSelectOption>
        <IonSelectOption value="bananas">Bananas</IonSelectOption>
      </ion-select>
    </ion-item>
  </ion-list>
  <div v-for="fruit in fruits" :key="fruit">{{ fruit }}</div>
</template>

<script>
import { IonList, IonItem, IonSelect, IonSelectOption } from '@ionic/vue';
import { defineComponent, ref } from 'vue';

export default defineComponent({
  components: { IonList, IonItem, IonSelect, IonSelectOption },
  setup() {
    const selectEl = ref();
    let fruits = ref([]);
    const test = (value) => {
      if (value) fruits.value.push(value);
      selectEl.value.$el.value = false;
    };
    const clear = () => {
      fruits = [];
    };
    return {
      selectEl,
      fruits,
      test,
      clear,
    };
  },
});
</script>
```

3. add items to "fruits" array by selecting a fruit on the ionSelect
5. change the interface to "action-sheet" to see the expected behavior

**Expected behavior**
interface "popover" should behave exactly like other interface options when ionChange is triggered

**Browser and OS (please complete the following information)**
 - OS: Windows
 - Browser: Chrome

sean-perkins:
## Pull request checklist

## Pull request type

 

 

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

- When using `ion-select` with `interface="popover"`, selecting an option will result in a duplicate emission of the `ionChange` event. 
- Selecting an existing selected option with the keyboard, does not dismiss the parent popover.

The duplicate emission issue spawns from: https://github.com/ionic-team/ionic-framework/pull/23474

The root problem was that we had both a click handler on the radio and a `@Listen` decorator for `ionChange`, both invoking the callback handler for the popover options. When a radio is selected, it would fire the click handler directly attached to the element and then the `ion-radio-group` would fire an `ionChange` event, additionally executing the callback handler (again).

Issue URL: Resolves #26789 (this issue is resolved as a result, but the problem is existent with a basic implementation in v7 beta). 

## What is the new behavior?

- `<ion-select interface="popover">` emits a single `ionChange` event when an option is selected from single selection
- Resolves `any` types within the `SelectPopover` class
- Selecting an existing selected option with the keyboard (space or enter key) will dismiss the popover (aligning with click behavior and browser behavior with `select`)
- Filled in UX tests for when the popover should dismiss from interaction within `ion-select-popover`
- Updates the legacy form syntax warning on `ion-checkbox` to include information for developers to use the `legacy` prop if they require the legacy form template

- [ ] Yes
- [x] No

## Other information

Dev-build: `7.0.0-dev.11676402953.118c3b36`


sean-perkins:
Not a huge fan of the PR title, open to ideas.

liamdebeasi (commented):
Can you add some more context as to why the proposed fix works?

core/src/components/select/test/basic/select.e2e.ts
```diff
@@ -153,7 +155,39 @@ test.describe('select: ionChange', () => {
     await radioButtons.nth(0).click();
 
     await ionChange.next();
-    expect(ionChange).toHaveReceivedEventDetail({ value: 'apple', event: { isTrusted: true } });
+    expect(ionChange).toHaveReceivedEventDetail({ value: 'apple' });
+    expect(ionChange).toHaveReceivedEventTimes(1);
+  });
+
+  test('should fire ionChange when confirming multiple values from a popover', async ({ page }) => {
```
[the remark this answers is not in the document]
sean-perkins: I can fill this information back into the main PR description, but this problem effects single selection for `ion-select`, regressing here: https://github.com/ionic-team/ionic-framework/pull/23474

The root problem was that we had both a click handler on the radio and a `@Listen` decorator for `ionChange`, both invoking the callback handler for the popover options. When a radio is selected, it would fire the click handler directly attached to the element and then the `ion-radio-group` would fire an `ionChange` event, additionally executing the callback handler (again).

This proposed fix works, because it maintains the click handler on the radio and moves the listener behavior directly onto the click handler for the checkbox (multiple selection). The `setChecked` implementation is only used/necessary for multiple selection with the checkboxes.

The tests you are commenting on are within a `test.describe` block, that skips execution for RTL and iOS. 

```ts
test.describe('select: ionChange', () => {
  test.beforeEach(({ skip }) => {
    skip.rtl();
    skip.mode('ios', 'ionChange has a consistent behavior across modes');
  });

  ...
});
```

One thing that I did not think to test originally that I will run through, is making sure keyboard interactions behave the same as well for selections. If not, there may be some rework to use the `ionChange` event instead of a click handler.

core/src/components/checkbox/checkbox.tsx
```diff
@@ -289,7 +289,9 @@ Example: <ion-checkbox>Label</ion-checkbox>
 
 For checkboxes that do not have a visible label, developers should use "aria-label" so screen readers can announce the purpose of the checkbox.
 
-For checkboxes that do not render the label immediately next to the checkbox, developers may continue to use "ion-label" but must manually associate the label with the checkbox by using "aria-labelledby".`,
+For checkboxes that do not render the label immediately next to the checkbox, developers may continue to use "ion-label" but must manually associate the label with the checkbox by using "aria-labelledby".
+
+Developers can use the "legacy" property to continue using the legacy form markup. This property will be removed in an upcoming major release of Ionic where this form control will use the modern form markup.`,
```
sean-perkins, line 294: If we are good with the wording here, I can follow-up with another PR to apply across all form controls for v7.

liamdebeasi: Looks good 👍

core/src/components/select-popover/test/basic/select-popover.e2e.ts
```diff
@@ -0,0 +1,63 @@
...
+      await expect(selectPopoverPage.popover).not.toBeVisible();
+    });
+
+    test('pressing Space on a selected option should dismiss the popover', async ({ browserName }) => {
+      test.skip(browserName === 'firefox', 'Same behavior as https://ionic-cloud.atlassian.net/browse/FW-2979');
+
+      await selectPopoverPage.updateOptions([
+        { value: 'apple', text: 'Apple', disabled: false, checked: true },
+        { value: 'banana', text: 'Banana', disabled: false, checked: false },
+      ]);
```
liamdebeasi, line 56: Is there a way we can start the test with these options pre-configured in the `<script>` tag? One thing that is slowing our tests down on CI is the amount of "setup work" we do before we begin our tests. This operation likely runs pretty quickly on your computer, but it can take ~1-2 seconds on CI (which adds up pretty quickly if many tests are doing this)

core/src/components/select-popover/select-popover.tsx
```diff
@@ -135,11 +137,26 @@ export class SelectPopover implements ComponentInterface {
...
+              onKeyUp={(ev) => {
+                if (ev.key === 'Enter' || ev.key === ' ') {
+                  /**
+                   * Selecting a radio option with keyboard navigation,
+                   * either through the Enter or Space keys, should
+                   * dismiss the popover.
+                   */
+                  this.dismissParentPopover();
+                }
+              }}
```
liamdebeasi, line 158: Why not dismiss on the `onIonChange` callback?

liamdebeasi (changes requested):

core/src/components/select-popover/select-popover.tsx
liamdebeasi: Nvm answered my own question: Selecting a previously selected value will not cause `ionChange` to fire

core/src/components/select-popover/select-popover.tsx
```diff
@@ -135,11 +137,26 @@ export class SelectPopover implements ComponentInterface {
...
           <ion-item class={getClassMap(option.cssClass)}>
             <ion-label>{option.text}</ion-label>
-            <ion-radio value={option.value} disabled={option.disabled} onClick={(ev) => this.rbClick(ev)}></ion-radio>
+            <ion-radio
+              value={option.value}
+              disabled={option.disabled}
+              legacy={true}
+              onClick={() => this.dismissParentPopover()}
+              onKeyUp={(ev) => {
+                if (ev.key === 'Enter' || ev.key === ' ') {
```
liamdebeasi, line 150: RadioGroup only selects if the user presses `Space`. Should we align with that behavior? Also, any reason why we check `key` versus `code` in the radio group?

sean-perkins: Good call on aligning the interactions. Updated to only handle the `Space` key for now. 

Personally, I think the radio group implementation is probably the incorrect usage for `key` vs. `code`, while it technically accomplishes the same end result. `code` is used in environments where the physical location of the key matters, such as a game that is controlled by `WASD` keys.

The MDN docs somewhat touch on this: 
- https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/code
- https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/key

`key` seems to be the default when that behavior is not required and the value of the key pressed is important. This can also include scenarios where modifier keys are important (using `Shift`, etc.).

liamdebeasi: Ok makes sense. Can you make a tech debt ticket to migrate radio group from `code` to `key`?

sean-perkins: Created FW-3502 to track and linked this discussion in the ticket 👍

liamdebeasi (approved):
Great job!

Issue #26829, opened 2023-02-20
bug: ion-select firing logging warning/deprecation message to console even when condition not met


lincolnthree:
### Prerequisites

- [X] I have read the [Contributing Guidelines](https://github.com/ionic-team/ionic-framework/blob/main/.github/CONTRIBUTING.md#creating-an-issue).
- [X] I agree to follow the [Code of Conduct](https://ionicframework.com/code-of-conduct).
- [X] I have searched for [existing issues](https://github.com/ionic-team/ionic-framework/issues) that already report this problem, without success.

### Ionic Framework Version

- [ ] v4.x
- [ ] v5.x
- [ ] v6.x
- [X] v7.x
- [ ] Nightly

### Current Behavior

ion-select appears to be logging warning messages about using ion-label even when there is no such ion-label element in the DOM. In fact, even a completely unconfigured ion-select component will cause the message to be printed:

![image](https://user-images.githubusercontent.com/362329/220176255-a4d6b8e5-06e5-4b8f-9020-071401ad43e4.png)

### Expected Behavior

Warning & deprecation messages should only be logged when criteria for those messages are met. (Unless you've decided you want to do this for some reason, in which case I defer to your judgement.)

### Steps to Reproduce

<ion-select></ion-select>

### Ionic Info

$ ionic info

Ionic:

   Ionic CLI                     : 6.20.3 (/Users/lincoln/.nvm/versions/node/v16.14.0/lib/node_modules/@ionic/cli)
   Ionic Framework               : @ionic/angular 7.0.0-dev.11676644846.156e8507
   @angular-devkit/build-angular : 15.1.6
   @angular-devkit/schematics    : 15.1.6
   @angular/cli                  : 15.1.6
   @ionic/angular-toolkit        : 8.0.0

Capacitor:

   Capacitor CLI      : 4.6.3
   @capacitor/android : 4.6.3
   @capacitor/core    : 4.6.3
   @capacitor/ios     : 4.6.3

Utility:

   cordova-res : not installed globally
   native-run  : 1.7.1

System:

   NodeJS : v16.14.0 (/Users/lincoln/.nvm/versions/node/v16.14.0/bin/node)
   npm    : 8.3.1
   OS     : macOS Monterey

lincolnthree:
I am also seeing this occur for `ion-range` and `ion-toggle`.

liamdebeasi:
[...]

PR #26834, opened 2023-02-21
refactor(many): clarify deprecation warning

liamdebeasi:
## Pull request checklist

## Pull request type

 

 

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes

Developers have noted that the form control deprecation warning is confusing: https://github.com/ionic-team/ionic-framework/issues/26829

This is because we assume devs always use `ion-label`. However, devs could use `<ion-select>` without a label and they would still get the deprecation warning. The deprecation warning should be logged, but the reason we give for why it is logged is confusing since it assumes the presence of `ion-label`. 

Additionally, several of the shadow dom components noted you could use `aria-labelledby` with an `ion-label`. This is incorrect as `aria-labelledby` does not work across shadow dom boundaries. This was likely a copy and paste mistake from `ion-input` and `ion-textarea` which do not use shadow dom.

Finally, I noticed that many of the components never mentioned the `legacy` prop in the deprecation warning.

## What is the new behavior?

- Removes the `ion-label` assumption and states that devs must now pass the label to the component via a prop/slot (depending on the component) or an aria-label
- Added examples for passing the label using aria-label
- Only checkbox had a note about using the `legacy` prop, so I added that comment to the remaining form control components too
- Remove the aria-labelledby wording from the shadow dom components

- [ ] Yes
- [x] No

### DIFF
--- a/core/src/components/checkbox/checkbox.tsx
+++ b/core/src/components/checkbox/checkbox.tsx
@@ -283,12 +283,10 @@ export class Checkbox implements ComponentInterface {
   private renderLegacyCheckbox() {
     if (!this.hasLoggedDeprecationWarning) {
       printIonWarning(
-        `Using ion-checkbox with an ion-label has been deprecated. To migrate, remove the ion-label and pass your label directly into ion-checkbox instead.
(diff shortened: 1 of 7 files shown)


sean-perkins (approved):
Thanks for adding the legacy prop message to the remaining controls 👍 

The message is much more clear now.

PR #26836, opened 2023-02-21
fix(form): shadow components using aria-labelledby do not use modern syntax

liamdebeasi:
## Pull request checklist

## Pull request type

 

 

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

Issue URL: resolves https://github.com/ionic-team/ionic-framework/issues/26829

Using `aria-labelledby` opts components into the modern syntax, but `aria-labelledby` is only usable with ion-input and ion-textarea since they don't use the shadow dom.

## What is the new behavior?

- The form controller only considers `aria-labelledby` with input or textarea components

- [ ] Yes
- [x] No


amandaejohnston (approved):

core/src/utils/forms/form-controller.ts
```diff
@@ -22,7 +22,10 @@ export const createLegacyFormController = (el: HTMLLegacyFormControlElement): Le
        * in the light DOM.
        */
       const hasLabelProp = (controlEl as any).label !== undefined || hasLabelSlot(controlEl);
-      const hasAriaLabelAttribute = controlEl.hasAttribute('aria-label') || controlEl.hasAttribute('aria-labelledby');
+      const hasAriaLabelAttribute =
+        controlEl.hasAttribute('aria-label') ||
+        // Shadow DOM form controls cannot use aria-labelledby
+        (controlEl.hasAttribute('aria-labelledby') && ALLOWS_ARIA_LABELLEDBY_COMPONENTS.includes(controlEl.tagName));
```
sean-perkins, line 28: I'm fine with this approach, but if this is specific to the shadow DOM, is there an alternate way to check that (e.g.: `controlEl.shadowRoot === undefined`) so we don't need to manage hardcoding selectors?

liamdebeasi: I think it depends on when `hasLegacyControl` is used. I _believe_ `connectedCallback` can be called before a component is upgraded to use the Shadow DOM, so if used there then the component won't have its shadow root connected.

liamdebeasi: Thanks! Fixed in https://github.com/ionic-team/ionic-framework/pull/26836/commits/d86839f8bff372bf9bb5d655333553b3acd195ac

sean-perkins (approved):

PR #26848, opened 2023-02-22
fix(datetime): aria-label for show/hide year picker

sean-perkins:
## Pull request checklist

## Pull request type

 

 

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

Screen readers do not announce to show/hide the year picker when selecting the month/year toggle button.

Issue URL: Internal

## What is the new behavior?

- Aria label has been added to the month/year toggle of the calendar for announcing "Show/Hide year picker" along with the current month/year
  - This aligns with how screen readers announce the iOS date picker
- Aria labels have been capitalized to align with other labels
- Screen readers announce show/hide year picker along with the current month/year

- [ ] Yes
- [x] No

## Other information

This PR is in favor of #26717 and #26744. While refactoring from `ion-item` to `ion-button` has value, the work effort to align the styles to match the previous experience will not fit into this sprint.


liamdebeasi (commented):
There is a flicker when the month/year picker toggles possibly due to the `key` usage: 

https://user-images.githubusercontent.com/2721089/220756325-7bf2a7af-4ed6-4fa2-9d33-477476b7be15.mov

sean-perkins:
Discussed offline, we have some limitations when it comes to re-rendering just the part of the DOM that the `aria-label` is attached to. I added a tech debt ticket to dig in more, but landed on manually setting the attribute on the element.

core/src/components/datetime/datetime.tsx
```diff
@@ -1890,7 +1891,31 @@ export class Datetime implements ComponentInterface {
...
+                /**
+                 * TODO: FW-3547
+                 *
+                 * Currently there is not a way to set the aria-label on the inner button
+                 * on the `ion-item` and have it be reactive to changes. This is a workaround
+                 * until we either refactor `ion-item` to a button or Stencil adds a way to
+                 * have reactive props for built-in properties, such as `aria-label`.
+                 */
+                if (this.monthYearToggleItemRef) {
+                  const btn = this.monthYearToggleItemRef.shadowRoot?.querySelector('.item-native');
```
liamdebeasi, line 1911: Can we make this a const to avoid the additional access?

```js
const { monthYearToggleItemRef } = this;

if (monthYearToggleItemRef) {
  const btn = monthYearToggleItemRef.shadowRoot?.querySelector('.item-native');
...
}
```

liamdebeasi (approved):

core/src/components/datetime/test/a11y/datetime.spec.ts
```diff
@@ -0,0 +1,52 @@
...
+describe('datetime', () => {
+  beforeEach(() => {
+    const mockIntersectionObserver = jest.fn();
+    mockIntersectionObserver.mockReturnValue({
+      observe: () => null,
+      unobserve: () => null,
+      disconnect: () => null,
+    });
+    global.IntersectionObserver = mockIntersectionObserver;
+  });
```
amandaejohnston, line 16: Why is this needed?

core/src/components/datetime/test/a11y/datetime.spec.ts
```diff
@@ -0,0 +1,52 @@
+import type { SpecPage } from '@stencil/core/testing';
+import { newSpecPage } from '@stencil/core/testing';
```
amandaejohnston, line 2: Was this written as a Stencil spec test vs a Playwright test as a workaround for something?

core/src/components/datetime/test/a11y/datetime.spec.ts
sean-perkins: The `ion-datetime` implementation uses an`IntersectionObserver` but Jest does not have an implementation for it. You have to mock certain web APIs, such as `IntersectionObserver`, otherwise the spec test will throw an exception.

core/src/components/datetime/test/a11y/datetime.spec.ts
sean-perkins: This was intentionally written as a unit test instead of an integration test. Similar tests for other components are spec tests as well (e.g.: [`ion-card`](https://github.com/ionic-team/ionic-framework/blob/main/core/src/components/card/test/aria.spec.ts)). A spec test is more efficient to use here, since we can validate the behavior with JSDOM vs. setting up the entire browser instance.

amandaejohnston (approved):

Issue #26833, opened 2023-02-21
bug: chrome, complex selector causing elements to not get garbage collected


joewoodhouse:
### Prerequisites

- [X] I have read the [Contributing Guidelines](https://github.com/ionic-team/ionic-framework/blob/main/.github/CONTRIBUTING.md#creating-an-issue).
- [X] I agree to follow the [Code of Conduct](https://ionicframework.com/code-of-conduct).
- [X] I have searched for [existing issues](https://github.com/ionic-team/ionic-framework/issues) that already report this problem, without success.

### Ionic Framework Version

- [ ] v4.x
- [ ] v5.x
- [X] v6.x
- [ ] v7.x
- [ ] Nightly

### Current Behavior

When a controller modal is dismissed, it's contents are detached from the DOM but are unable to me garbage collected. This represents a resource leak, and repeated opening/closing of a modal like this will eventually lead to the app being killed by the operating system.

### Expected Behavior

When a modal is dismissed, all resources associated it should be freed.

### Steps to Reproduce

1. Go to the documentation page for controller modals: 
https://ionicframework.com/docs/usage/v6/modal/controller/demo.html?ionic:mode=ios
2. Open Chrome Devtools "Performance Monitor" tab
3. Take a Heap Snapshot (Memory > Record button)
4. Repeatedly open and close the modal
5. Watch the Performance Monitor tab's "DOM Nodes" number increase indefinitely
6. Force manual GC (trash icon in the Memory tab) - note the DOM nodes do not reduce
7. Take another Heap Snapshot and compare. You will see 100s of Detached Elements

A video of me doing these steps:
https://drive.google.com/file/d/1ciQxC1xSa_aiyrA2OfEIJCjXHJH_sBhx/view?usp=sharing

### Code Reproduction URL

https://ionicframework.com/docs/usage/v6/modal/controller/demo.html?ionic:mode=ios

### Ionic Info

Ionic:

   Ionic CLI : 6.20.6 (/Users/joewoodhouse/.nvm/versions/node/v16.14.2/lib/node_modules/@ionic/cli)

Capacitor:

   Capacitor CLI      : 4.6.2
   @capacitor/android : 4.6.3
   @capacitor/core    : 4.6.3
   @capacitor/ios     : 4.6.3

Utility:

   cordova-res : not installed globally
   native-run  : 1.7.1

System:

[...]

PR #26856, opened 2023-02-24
fix(modal): keyboard listener removed on dismiss

liamdebeasi:
## Pull request checklist

## Pull request type

 

 

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

Part of https://github.com/ionic-team/ionic-framework/issues/26833

The keyboard listener is added after the transition completes. This means that if you present and dismiss a modal quickly, the dismiss method will run before the keyboard listener is added. This is because the dismiss method only waits for the transition to finish before proceeding. As a result, the keyboard listener is never removed.

## What is the new behavior?

- Updated the present method to create the keyboard listener before the transition runs, that way it can always be removed in `dismiss`. Note that the callback already checks to see if the gesture is configured so we can safely add this listener before the gestures are fully configured.

- [ ] Yes
- [x] No

### DIFF
--- a/core/src/components/modal/modal.tsx
+++ b/core/src/components/modal/modal.tsx
@@ -333,7 +333,7 @@ export class Modal implements ComponentInterface, OverlayInterface {
     if (this.gesture) {
       this.gesture.enable(enable);
     } else if (enable) {
-      await this.initSwipeToClose();
+      this.initSwipeToClose();
     }
   }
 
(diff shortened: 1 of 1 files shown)


core/src/components/modal/modal.tsx
```diff
@@ -333,7 +333,7 @@ export class Modal implements ComponentInterface, OverlayInterface {
     if (this.gesture) {
       this.gesture.enable(enable);
     } else if (enable) {
-      await this.initSwipeToClose();
+      this.initSwipeToClose();
```
liamdebeasi, line 336: This method is not async, so it does not need await

amandaejohnston (approved):

Issue #26697, opened 2023-01-29
bug: can't place multiple form controls in one ion-item when using new syntax


rdlabo:
### Prerequisites

- [X] I have read the [Contributing Guidelines](https://github.com/ionic-team/ionic-framework/blob/main/.github/CONTRIBUTING.md#creating-an-issue).
- [X] I agree to follow the [Code of Conduct](https://ionicframework.com/code-of-conduct).
- [X] I have searched for [existing issues](https://github.com/ionic-team/ionic-framework/issues) that already report this problem, without success.

### Ionic Framework Version

- [ ] v4.x
- [ ] v5.x
- [ ] v6.x
- [X] v7.x
- [ ] Nightly

### Current Behavior

ion-select[slot=end] require attribute of `label` or `area-label`.
But if added, break main label.

#### 1. console.log: `Using ion-select with an ion-label has been deprecated.` , but it displays as expected.

```html
<ion-item>
    <ion-input
            label="Price"
            labelPlacement="fixed"
            value="100.0"
            type="number"
    ></ion-input>
    <ion-select
            slot="end"
            placeholder="UNIT"
            value="USD"
    >
      <ion-select-option value="USD">USD</ion-select-option>
      <ion-select-option value="JPY">JPY</ion-select-option>
    </ion-select>
</ion-item>
```

![スクリーンショット 2023-01-29 午後8 07 28](https://user-images.githubusercontent.com/9690024/215322074-91fd37bf-72c3-4dce-b5a2-b080040579fb.png)

#### 2. If add label to `ion-select`, ion-input is removed:

![スクリーンショット 2023-01-29 午後8 08 49](https://user-images.githubusercontent.com/9690024/215322129-6641d9e8-12b0-4e11-9f22-d0cc21a868d3.png)

### Expected Behavior

ion-select[slot=end] doesn't require attribute of `label` or `area-label`, or else.

### Steps to Reproduce

```
<!-- as expect -->
<ion-item>
    <ion-input
            label="Price"
            labelPlacement="fixed"
            value="100.0"
            type="number"
    ></ion-input>
    <ion-select
            slot="end"
            placeholder="UNIT"
            value="USD"
    >
      <ion-select-option value="USD">USD</ion-select-option>
      <ion-select-option value="JPY">JPY</ion-select-option>
    </ion-select>
  </ion-item>


<!-- removed ion-input -->
  <ion-item>
    <ion-input
            label="Price"
            labelPlacement="fixed"
[...]

PR #26897, opened 2023-03-02
fix(item): prevent slotted form controls from taking whole width of item

amandaejohnston:
## Pull request checklist

## Pull request type

 

 

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

When any of the form controls (using the new syntax) are placed in the `start` or `end` slots of `ion-item`, the control takes up the entire width of the item, pushing other things out. This makes it appear as though other components in the item have disappeared.

Issue URL: Resolves https://github.com/ionic-team/ionic-framework/issues/26697

## What is the new behavior?

Width of the form controls changed to `auto` in this circumstance. The developer can further tweak things by setting their own custom `width` value in their CSS.

While the screenshot test shows the inputs and textareas overlapping on mobile screen sizes, I intentionally avoided adding any extra opinionation to the width of the form controls. In a real app, devs would either not use inputs in both slots at once like this, or manually adjust the width to their needs so everything fits.

- [ ] Yes
- [x] No

## Other information

Side note: I'm seeing some screenshot changes for the `disabled` item test on Firefox -- namely the toggle in the "Checkbox + Toggle" row has shifted to the side -- but it looks like the original version was incorrect since I'm seeing the current placement on the `feature-7.0` branch in Chrome.

I also updated the `enable-on-off-labels` test for `ion-toggle`; all the toggles were placed in the `end` slot, likely as a holdover when migrating them to the new syntax, and the branch change caused them to no longer span the entire width of the page, leading to screenshot diffs. I see `slot="end"` as incorrect usage rather than this being an actual regression, since there isn't anything else in the items. There are still some diffs coming from that, but I don't see the differences, so I think those are fine.

For the fix itself, I initially tried something like this in `ion-item`'s styling instead:
```
::slotted(ion-checkbox:not(.legacy-checkbox)),
::slotted(ion-radio:not(.legacy-radio)),
::slotted(ion-toggle:not(.legacy-toggle)),
::slotted(ion-range:not(.legacy-range)),
::slotted(ion-input:not(.legacy-input)),
::slotted(ion-textarea:not(.legacy-textarea)),
::slotted(ion-select:not(.legacy-select)) {
  width: auto;
}
```
Setting aside the work to make that selector less brittle (unifying all controls under the same `.form-control` and `.legacy` classes to avoid needing to list out each component), this didn't work for `ion-input` and `ion-textarea` because they have `:host { width: 100% }` which was overriding my styles. Theoretically we could use that fix for all the other components, but that would break them if any others get updated to have their `width` set on the host (which is highly likely as we move away from the legacy syntax).

I also tried wrapping `ion-item`'s slots in wrapper elements and setting `display: flex` on those, but this led to small visual regressions elsewhere (mainly I noticed a height difference when looking at slotted icons).


sean-perkins (approved):
FWIW the diff on the `enableOnOffLabels` is ~ 2px on the right of selected toggles (the pink dots), but I'm also not able to perceive that difference when zooming on the screenshots.

textcompare-image-diff

I think that "visual change" is warranted for the proposed fix here.

Issue #26939, opened 2023-03-09
bug: ion-input new v7 errorText displayed right away


cyrilletuzi:
### Prerequisites

- [X] I have read the [Contributing Guidelines](https://github.com/ionic-team/ionic-framework/blob/main/.github/CONTRIBUTING.md#creating-an-issue).
- [X] I agree to follow the [Code of Conduct](https://ionicframework.com/code-of-conduct).
- [X] I have searched for [existing issues](https://github.com/ionic-team/ionic-framework/issues) that already report this problem, without success.

### Ionic Framework Version

v7.x

### Current Behavior

v7 introduced `errorText` property to `<ion-input>`. But if the input has the `required` attribute, which is one of the most basic and common control, the `errorText` will be displayed right away when the form is displayed. This is problematic for several reason:
- it's bad UX (we should not annoy the user right away, only once the user touched the input and moved to another one)
- it's inconsistent with the visual behavior: on first display, the CSS is normal, and it turns red only if the input is invalid *and* touched and the user has moved to another input, as expected
- it ruins all the new simpler `<ion-input>` behavior introduced in v7, as it forces to add back a `<ion-item>` with `<ion-note slot="error" *ngIf="emailControl.invalid && emailControl.dirty">` (assuming it's still possible, which is not even sure, it doesn't work in v7 in my application anymore, probably because you can't mix the 2 systems)

### Expected Behavior

`<ion-input>` `errorText` should behave like CSS, ie. be displayed only when the input is invalid *and* touched and moved away to another input.

### Steps to Reproduce

1. Add a `<ion-input type="text" label="text" name="hello" required errorText="Error" [(ngModel)]="hello">`

(sorry I tried to do a stackblitz but I couldn't make it work with ionic v7 and I have no more time to create a full github repo)

### Ionic Info

`@ionic/angular` v7.0.0-rc.1
`@angular/*` v15.2.2x
node v18.15.0
npm v9.6.0

cyrilletuzi:
(Another solution would be to introduce a new property, like `errorTextTrigger` with values like `touched`, `dirty`...).

liamdebeasi:
Thanks for the report. Here is a dev build if you are interested in testing the proposed fix:

```
npm install @ionic/angular@7.0.0-dev.11678373650.1975657a
```

liamdebeasi:
[...]

Issue #21643, opened 2020-06-29
bug: form validation should not show red highlight until user has modified the field


Marius-Romanus:
# Bug Report

**Ionic version:**

[ ] **4.x**
[x] **5.x**

**Current behavior:**

In reactive Angular forms, the ion-items within an ion-input, the highlights are marked in red when they have not yet been touched.

**Expected behavior:**

They should be blue until they are not touched, even though the validation of required giving an error.

**Steps to reproduce:**

Create a reactive form group and add the required validation

**Related code:**

```
import { ReactiveFormsModule } from '@angular/forms';
imports: [
    ReactiveFormsModule,
    ...
],
```

```
formGroup: FormGroup;

constructor(public formBuilder: FormBuilder) {
    this.formGroup = this.formBuilder.group({
      text: [
        "",
        [Validators.required]
      ]
    });
}
```

```
<form [formGroup]="formGroup">
	<ion-item>
		<ion-label floating>Text</ion-label>
		<ion-input type="text" formControlName="text"></ion-input>
	</ion-item>
</form>
```

**Ionic info:**

```
Ionic:

   Ionic CLI                     : 6.10.1
   Ionic Framework               : @ionic/angular 5.2.2
   @angular-devkit/build-angular : 0.1000.0
   @angular-devkit/schematics    : 8.3.25
   @angular/cli                  : 10.0.0
   @ionic/angular-toolkit        : 2.2.0

Capacitor:

   Capacitor CLI   : 2.2.0
   @capacitor/core : 2.2.0

Utility:

   cordova-res : not installed
   native-run  : not installed

System:

   NodeJS : v14.3.0
   npm    : 6.14.5
   OS     : Windows 10
```

Marius-Romanus:
Code reproduction in tab 1 of StackBlitz.

https://stackblitz.com/edit/ionic-v4-angular-tabs-qni8lg?file=src%2Fapp%2Ftab1%2Ftab1.page.ts

Regards

liamdebeasi:
Thanks for the follow up. Can you please provide some steps to reproduce the issue? The form appears to be functioning correctly when I check the StackBlitz.

Marius-Romanus:
The strict mode behavior is correct, but it is not correct for a user as far as UX is concerned, since the user has not yet written anything to have it automatically red-marked.
To be correct in my opinion, it should not be marked in red until it is touched or dirty.
Sorry if my English is not entirely correct. Regards

[...]

PR #26940, opened 2023-03-09
fix(input, textarea): show error state after touch

liamdebeasi:
## Pull request checklist

## Pull request type

 

 

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

Issue URL: resolves https://github.com/ionic-team/ionic-framework/issues/26939, resolves https://github.com/ionic-team/ionic-framework/issues/21643

There are a couple issues with the form validation highlighting:

1. The valid highlighting was always showing on a valid input, even when the input was blurred.
2. The invalid error text was always showing on an invalid input, even when the input was untouched (touched = the user has focused and subsequently blurred the input)

Form validation highlighting is used to indicate that the user did something right/wrong, so having these states always show can be confusing.

## What is the new behavior?

I updated the behavior as per the following table:

| Validity State | Touched State | Focused State | Validity Highlighting Shown |
| - | - | - | - |
| Valid | Untouched | Blurred | ⚫️ None |
| Valid | Touched | Blurred | ⚫️ None |
| Valid | Untouched | Focused | 🟢 Valid |
| Valid | Touched | Focused | 🟢 Valid |
| Invalid | Untouched | Blurred | ⚫️ None |
| Invalid | Touched | Blurred |  🔴 Invalid |
| Invalid | Untouched | Focused | ⚫️ None |
| Invalid | Touched | Focused | 🔴 Invalid |

Reasoning for invalid state changes: An invalid state indicates the user entered malformed/incorrect data. As a result, having the invalid state show before the user has even interacted with the input can be confusing for the user. We show the input regardless of focus state once it has been touched. This is because users may need to correct data in order to submit a form. Showing the invalid highlight on both focus and blur lets users quickly identify which form controls need fixing.

Reasoning for the valid changes: A valid state indicates that the user entered correct data. As a result, we should give them that positive feedback as soon as their data is valid. We also show the valid state regardless of touched state because some forms may pre-populate data. Users may not necessarily want to change that pre-populated data if it is valid. Additionally, the user does not need to always know that an input is valid which is why we only show it on focus. This behavior also aligns with the behavior found in Ionic 6.

- I found and removed duplicate code in the input and textarea stylesheets
- I updated the input and textarea tests to check the correct valid/invalid conditions
- I added validity state tests for select

- [ ] Yes
- [x] No

## Other information

Dev build: 7.0.0-dev.11678373650.1975657a


core/src/components/textarea/textarea.scss
```diff
@@ -286,17 +286,6 @@
...
-// ----------------------------------------------------------------
-
-:host(.ion-touched.ion-invalid) {
-  --highlight-color: var(--highlight-color-invalid);
-}
-
-:host(.ion-touched.ion-valid) {
-  --highlight-color: var(--highlight-color-valid);
-}
-
```
liamdebeasi, line 299: This was duplicate code

core/src/components/input/input.scss
```diff
@@ -398,17 +414,6 @@
...
-// ----------------------------------------------------------------
-
-:host(.ion-touched.ion-invalid) {
-  --highlight-color: var(--highlight-color-invalid);
-}
-
-:host(.ion-touched.ion-valid) {
-  --highlight-color: var(--highlight-color-valid);
-}
-
```
liamdebeasi, line 411: This was duplicate code

sean-perkins (approved):
Changes look good. 

The test templates (e.g. test/bottom-content/index.html) are a little confusing to reason about right now, unless you are aware of the extra classes that need to be appended to the host to make the error slot visible and/or the styling. Should we add a minimal implementation here with a button that adds the required classes to help with manually testing?

liamdebeasi:
Good catch! I updated those templates so `ion-touched` is added and the error styles show up.

PR #26942, opened 2023-03-09
feat(slides): add migration styles for Swiper Element

averyjohnston:
## Pull request checklist

## Pull request type

 

 

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes

Our current stylesheet for migrating from `ion-slides` to Swiper ([ionic-swiper.scss](https://github.com/ionic-team/ionic-framework/blob/swiper-element-css/core/src/css/ionic-swiper.scss)) is not compatible with the new [Swiper Element](https://swiperjs.com/element). Swiper Element is required for using Swiper 9 with Angular, with React and Vue set to follow in the near future.

Issue URL: N/A

## What is the new behavior?

New stylesheet added that carries over all the same styles in a way that's compatible with Swiper Element. Notably, it uses the shadow DOM, so some styles were converted to CSS props from Swiper's own API.

I elected to make a separate stylesheet for the sake of keeping things clean, but I'm open to merging the two for ease of developer experience if desired.

I was unable to figure out what [these styles](https://github.com/ionic-team/ionic-framework/blob/swiper-element-css/core/src/css/ionic-swiper.scss#L69-L75) apply to -- the [Zoom feature](https://swiperjs.com/swiper-api#zoom) uses different classes -- so I left them off the new stylesheet, but kept them in the old one just in case.

I'm targeting this towards `main` because it's sort of a bug fix, in that our current `ionic-swiper.scss` sheet doesn't work for Swiper Element. I'm open to switching to 6.x or 7.0, but this still needs to go out before the upcoming Swiper migration guides can be published.

Dev build: `6.6.2-dev.11678390792.160fa5e6`
Repo/branch I've been using to test (which currently uses the above dev build and new stylesheet): https://github.com/amandaejohnston/sandbox-ionic-angular/tree/swiper-migration-testing

- [ ] Yes
- [x] No

Usage with Swiper's Angular/React/Vue components is unaffected. The existing `ionic-swiper.scss` still works with Swiper 9 as long as you aren't using Swiper Element.

### DIFF
--- a/core/src/css/ionic-swiper-element.scss
+++ b/core/src/css/ionic-swiper-element.scss
@@ -9,7 +9,7 @@ swiper-container {
   --swiper-pagination-progressbar-bg-color: #{$slides-ios-progress-bar-background};
   // note: active progressbar color also uses --swiper-pagination-color
   --swiper-scrollbar-bg-color: #{$slides-ios-scroll-bar-background};
-  --swiper-scrollbar-drag-bg-color: #{$slides-ios-progress-bar-background-active};
+  --swiper-scrollbar-drag-bg-color: #{$slides-ios-scroll-bar-drag-background};
 }
 
 swiper-slide {


core/src/css/ionic-swiper-element.scss
```diff
@@ -0,0 +1,38 @@
+@import "../components/slides/slides.ios.vars.scss";
```
liamdebeasi, line 1: Threading for discussion:

How would you feel about not shipping this in Ionic but instead keeping it as an optional code snippet on the migration guide?

A couple reasons:

1. One less "public API" we need to maintain, and we can remove the other Swiper stylesheet in a future major release.
2. If Swiper 10 comes out and makes a breaking change to styling, we'll need to scramble to update this stylesheet and do it in a way such that both Swiper 9 and 10 are supported. At that point, we'll be back to having to maintain some level of Swiper support.

-----

Another thought: We could also provide a migration table for the CSS variables like we do for other Swiper APIs (example: https://ionicframework.com/docs/angular/slides#events)

This table would show the Ionic CSS variable and the equivalent Swiper CSS variable. That way developers who are customizing Swiper can have a similar style result without having to use Ionic-created variables.

sean-perkins: Makes sense to me to have the CSS available in the docs instead of us managing and shipping another stylesheet 👍

amandaejohnston:
Closing this out per discussion in https://github.com/ionic-team/ionic-framework/pull/26942#discussion_r1131606441. I'll document these styles directly in the Angular migration guide instead.

PR #27001, opened 2023-03-22
fix(select): update iOS icon

sean-perkins:
## Pull request checklist

## Pull request type

 

 

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

The icon displayed for `ion-select` in iOS mode is incorrect based on iOS 16.

Issue URL: Internal

## What is the new behavior?

__Note__: Screenshot is outdated.

- Updates the icon for `ion-select` in `ios` mode to render the new `chevron-expand` icon available in the dev-build of ionicons 7.1
- Moves the icon width styles specific to md mode to the md stylesheet
- Fixes a bug with the modern `ion-select` form syntax, where when used in an `ion-item` with the `color` attribute applied, the icon of the select was not correctly inheriting the color for appropriate color contrast. 

- [ ] Yes
- [x] No


brandyscarney (changes requested):
Okay so I did a comparison of our select icon vs native iOS and here are my findings:

- Our icon is a lot lighter than native
- Our icon is a bit smaller than native (but it is also designed more narrow)
 
select-icon-ios

I used the Settings app to get the native example. It is under Settings -> Focus -> Do Not Disturb -> People

We can update the color to match the light theme by changing it to `--ion-color-step-600` which would be `$text-color-step-400`. This will be off for dark theme but we don't have a way to adjust this. Users can always change the value of `--ion-color-step-600` for their dark theme. I am using what is generated by our [stepped color generator](https://ionicframework.com/docs/theming/themes#stepped-color-generator) and using `#000` as the background color and `#fff` as the text color to get these hex values. I attempted to see if we could use a `rgb` value with `--ion-text-color` instead but this had similar results.

The size of our icon is more narrow but it is also a bit shorter. Adjusting the `width` and `height` to `18px` gives about the same distance between the top and bottom chevrons as native. 

Let me know if you need me to explain anything in this image better!

sean-perkins:
Ok, this is ready for another review. Here is a summary of what has changed/where we are at:

The iOS colors fail AAA color contrast, so we have selected a color that passes and falls within our color variables. To better align the iOS icon to the iOS spec, we have adjusted the width and height to `18px`. 

While working on this change, I identified that the icon color had regressed when in an `ion-item` with `color` specified, in the modern form control syntax. The legacy class for `in-item-color` has been applied to the modern template so that colors are inherited correctly for better color contrast. 

The select primary text (not to be confused with the placeholder text color) is still not in spec with iOS. I am going to create a spike for further investigation into this (FW-3850). Currently the color is `inherit` and developers can customize as they desire. The Settings and Calendar app on iOS have similar implementations of the select interface, but the colors are different between the two. 

The goal of this PR is to update the icon and improve the color contrast issues that currently exist in v6/7.0.0-rc.4.

brandyscarney (approved):

PR #27014, opened 2023-03-23
fix(item): use thumbnail's size when present

### LINKED ISSUE #22935: bug: thumbnail not being properly sized when slotted inside item
# Bug Report

**Ionic version:**
[x] **5.x**

**Current behavior:**

The ionic docs says : 

> Thumbnails can be used by themselves or inside of any element. If placed inside of an ion-item, the thumbnail will resize to fit the parent component

But when a ion-thumbnail is inside a ion-item and slotted, it doesn't fit the ion-item, instead it forces its size to 56px which makes the ion-item height 56px.

image

If we want to make the thumbnail smaller we can use the --size attribute, but it is overridden by the code above when slotted.

I add the size attribute : 
image

The ion-thumbnail size attribute is overridden by the slotted css : 

image

**Expected behavior:**

The size attribute shouldn't be overridden ?
Why force the height/width to 56px when slotted ?

**Steps to reproduce:**

Go to : https://ionicframework.com/docs/api/thumbnail 

* With your browser, inspect a thumbnail in the example in the top right corner
* Set a --size attribute to a value, for example 20px
* The thumbnail doesn't change its size

thetaPC:
## Pull request checklist

## Pull request type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

The `--size` from `ion-thumbnail` is being overwritten when component is placed inside `ion-item` by `ion-item`'s slotted styling.

Screenshot 2023-03-23 at 11 37 44 AM

This contradicts the behavior described in the `ion-thumbnail` [documentation](https://ionicframework.com/docs/api/thumbnail).

> Thumbnails can be used by themselves or inside of any element. If placed inside of an ion-item, the thumbnail will resize to fit the parent component.

Issue URL: resolves #22935

## What is the new behavior?

- `ion-item` will set a default value for `ion-thumbnail`'s `--size` when present

Passes a default size to thumbnail when slotted.

- [ ] Yes
- [x] No

## Other information

N/A


liamdebeasi (changes requested):
The changes look great! Could we add a test so this does not regress?

Maybe something like this:

```javascript
test('size should be customizable in item', async ({ page, skip }) => {
  skip.rtl();

  await page.setContent(`
    <ion-item>
      <ion-thumbnail style="--size: 20px">
        <img src="/src/components/thumbnail/test/thumbnail.svg" />
      </ion-thumbnail>
    </ion-item>
  `);
  const item = page.locator('ion-item');
  await expect(item).toHaveScreenshot(`item-thumbnail-size-${page.getSnapshotSettings()}.png`);
});
```

We will want to test iOS and MD mode since we are updating the code for both modes. However, we can skip the LTR vs RTL tests. The focus of this test should be ensuring that developers can use the `--size` API on `ion-thumbnail` to customize its size even when used inside of `ion-item`.

core/src/components/item/item.ios.scss
```diff
@@ -154,20 +154,24 @@
...
+// iOS Item Thumbnail
+// --------------------------------------------------
+
 ::slotted(ion-thumbnail) {
-  width: $item-ios-thumbnail-width;
-  height: $item-ios-thumbnail-height;
+  --size: #{$item-ios-thumbnail-size};
 }
 
+// iOS Design Item Avatar/Thumbnail
```
brandyscarney, line 172:
```suggestion
// iOS Item Avatar/Thumbnail
```

brandyscarney (changes requested):

brandyscarney (approved):
Looks great!

PR #27024, opened 2023-03-24
fix(item-divider): add margin to end slots

thetaPC:
## Pull request checklist

## Pull request type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

End slots are lacking margin of 2px when in RTL.

Issue URL: Part of #17012  

## What is the new behavior?

- `item-md-end-slot-margin-end` = 2px

End slots will have a margin-end of 2px regardless of direction.

### Screenshots

Screen Shot 2023-03-24 at 13 48 25
Screen Shot 2023-03-24 at 13 48 17

- [ ] Yes
- [x] No

## Other information
N/A


liamdebeasi (commented):
I don't think this is the correct way to solve this bug. According to the report, this issue only impacts content slotted using `ion-item-divider`. However, the proposed fix impacts anything in `ion-item` even if `ion-item-divider` is not being used.

If you look at the computed CSS in the original issue reproduction, you should see the following output in Chrome:

Group 1

The margin, denoted by `margin-inline-end` inside of the green highlight, adds the correct margin of 2px on the left side when the content is in RTL. (margin-inline-end docs: https://developer.mozilla.org/en-US/docs/Web/CSS/margin-inline-end). These are logical properties, so they use "start" and "end" terminology instead of "left" and "right".

If you look at the red highlight you should see the `margin-left` is being unset. So the problem here is the margin is being set correctly initially, but something else is overriding it.

------

The source for the `margin-left: unset` is coming from `ion-item-divider`. You can find this by clicking the "<style>" across from the `margin-left: unset`. This appears to be the source of the issue: https://github.com/ionic-team/ionic-framework/blob/f68644065c227b73fe49330b030e2cfdad7b603a/core/src/components/item-divider/item-divider.md.scss#L29

The `margin-horizontal` mixin is a custom mixin we wrote to make it easy to use logical properties for RTL support. The source for that is here: https://github.com/ionic-team/ionic-framework/blob/f68644065c227b73fe49330b030e2cfdad7b603a/core/src/themes/ionic.mixins.scss#L304-L306

It also relies on a more generic `property-horizontal` mixin that we use for logical margins, padding, and borders: https://github.com/ionic-team/ionic-framework/blob/f68644065c227b73fe49330b030e2cfdad7b603a/core/src/themes/ionic.mixins.scss#L239-L266

In particular, it looks like it sets margin-left/margin-right for browsers that do not support these logical properties and then uses `unset` inside of the `@supports` block for browsers that do support logical properties.

-----

The investigation here ended up being a little more complex than I originally expected. I also have some thoughts on how we can fix this, so let me know when you have some time and we can sync over Zoom.

thetaPC:
> I don't think this is the correct way to solve this bug. According to the report, this issue only impacts content slotted using `ion-item-divider`. However, the proposed fix impacts anything in `ion-item` even if `ion-item-divider` is not being used.
> 
> If you look at the computed CSS in the original issue reproduction, you should see the following output in Chrome:
> 
> Group 1
> 
> The margin, denoted by `margin-inline-end` inside of the green highlight, adds the correct margin of 2px on the left side when the content is in RTL. (margin-inline-end docs: https://developer.mozilla.org/en-US/docs/Web/CSS/margin-inline-end). These are logical properties, so they use "start" and "end" terminology instead of "left" and "right".
> 
> If you look at the red highlight you should see the `margin-left` is being unset. So the problem here is the margin is being set correctly initially, but something else is overriding it.
> 
> The source for the `margin-left: unset` is coming from `ion-item-divider`. You can find this by clicking the "<style>" across from the `margin-left: unset`. This appears to be the source of the issue:
> 
> https://github.com/ionic-team/ionic-framework/blob/f68644065c227b73fe49330b030e2cfdad7b603a/core/src/components/item-divider/item-divider.md.scss#L29
> 
> The `margin-horizontal` mixin is a custom mixin we wrote to make it easy to use logical properties for RTL support. The source for that is here:
> 
> https://github.com/ionic-team/ionic-framework/blob/f68644065c227b73fe49330b030e2cfdad7b603a/core/src/themes/ionic.mixins.scss#L304-L306
> 
> It also relies on a more generic `property-horizontal` mixin that we use for logical margins, padding, and borders:
> 
> https://github.com/ionic-team/ionic-framework/blob/f68644065c227b73fe49330b030e2cfdad7b603a/core/src/themes/ionic.mixins.scss#L239-L266
> 
> In particular, it looks like it sets margin-left/margin-right for browsers that do not support these logical properties and then uses `unset` inside of the `@supports` block for browsers that do support logical properties.
> 
> The investigation here ended up being a little more complex than I originally expected. I also have some thoughts on how we can fix this, so let me know when you have some time and we can sync over Zoom.

Let me know if we should still sync up.

liamdebeasi (commented):
We're getting much closer I think!

What I was thinking instead is to simplify the `property-horizontal` mixin:

```scss
@mixin property-horizontal($prop, $start, $end: $start) {
  @if $start == 0 and $end == 0 {
    #{$prop}-left: $start;
    #{$prop}-right: $end;
  } @else {
    -webkit-#{$prop}-start: $start;
    #{$prop}-inline-start: $start;
    -webkit-#{$prop}-end: $end;
    #{$prop}-inline-end: $end;
  }
}
```

The root problem here is we are unsetting the margin inside of this mixin when we should not be. While the code you have works, it does not address the root issue.

We use `property-horizontal` for margin-inline, padding-inline, and border-inline. Here is what the browser support looks like:

margin-inline-start/end: https://caniuse.com/?search=margin-inline-end 

padding-inline-start/end: https://caniuse.com/?search=padding-inline-end 

border-inline-start/end: https://caniuse.com/?search=border-inline-end

The only thing we need to be concerned about is border-inline-* which has support starting in Chrome 69. (Ionic v6 supports Chrome 60+). If we moved forward with the mixin change, we could target this fix for v7 instead and get the following benefits:

1. Resolve the underlying issue and prevent this from popping up in other places
2. Reduce the amount of code this mixin generates, which leads to a smaller bundle size.

thetaPC:
> 

This makes sense. Plus it makes it easier to trace back styles. I'll go ahead and swap to this solution.

thetaPC:
> We're getting much closer I think!
> 
> What I was thinking instead is to simplify the `property-horizontal` mixin:
> 
> ```css-scss
> @mixin property-horizontal($prop, $start, $end: $start) {
>   @if $start == 0 and $end == 0 {
>     #{$prop}-left: $start;
>     #{$prop}-right: $end;
>   } @else {
>     -webkit-#{$prop}-start: $start;
>     #{$prop}-inline-start: $start;
>     -webkit-#{$prop}-end: $end;
>     #{$prop}-inline-end: $end;
>   }
> }
> ```
> 
> The root problem here is we are unsetting the margin inside of this mixin when we should not be. While the code you have works, it does not address the root issue.
> 
> We use `property-horizontal` for margin-inline, padding-inline, and border-inline. Here is what the browser support looks like:
> 
> margin-inline-start/end: https://caniuse.com/?search=margin-inline-end
> 
> padding-inline-start/end: https://caniuse.com/?search=padding-inline-end
> 
> border-inline-start/end: https://caniuse.com/?search=border-inline-end
> 
> The only thing we need to be concerned about is border-inline-* which has support starting in Chrome 69. (Ionic v6 supports Chrome 60+). If we moved forward with the mixin change, we could target this fix for v7 instead and get the following benefits:
> 
>     1. Resolve the underlying issue and prevent this from popping up in other places
> 
>     2. Reduce the amount of code this mixin generates, which leads to a smaller bundle size.

This makes sense. Plus it makes it easier to trace back styles. I'll go ahead and swap to this solution.

liamdebeasi (commented):
The changes look good! A couple final things:

1. We need to target the `feature-7.0` branch instead of `main` now because these changes are dependent on the updated browser support Ionic 7 has.
2. We either need to update existing screenshots or we need to add a test for this.

brandyscarney:
Just noting here that tests for this might conflict with [this PR](https://github.com/ionic-team/ionic-framework/pull/27019) updating the `item-divider` tests. We might want to get one done before the other or keep the tests in separate directories.

thetaPC:
> Just noting here that tests for this might conflict with [this PR](https://github.com/ionic-team/ionic-framework/pull/27019) updating the `item-divider` tests. We might want to get one done before the other or keep the tests in separate directories.

@brandyscarney This PR has been moved to ionic-team/ionic-framework#27042.

I'll keep an eye out on both PRs. If their [PR](ionic-team/ionic-framework/pull/27019) is merged in first, I'll accommodate. Otherwise, I'll let them know that mine was merged first.

thetaPC:
Closing since this ionic-team/ionic-framework#27042 will be used to address the issue instead. The reason for the new PR is due to easily swapping to a different base.

PR #27040, opened 2023-03-28
bug(breadcrumbs): color attribute shows on DOM for Vue

thetaPC:
## Pull request checklist

Please check if your PR fulfills the following requirements:
- [x] Tests for the changes have been added (for bug fixes / features)
- [ ] Docs have been reviewed and added / updated if needed (for bug fixes / features)
  - Some docs updates need to be made in the `ionic-docs` repo, in a separate PR. See the [contributing guide](https://github.com/ionic-team/ionic-framework/blob/main/.github/CONTRIBUTING.md#modifying-documentation) for details.
- [x] Build (`npm run build`) was run locally and any changes were pushed
- [x] Lint (`npm run lint`) has passed locally and any fixes were made for failures

## Pull request type

Please check the type of change your PR introduces:
- [x] Bugfix
- [ ] Feature
- [ ] Code style update (formatting, renaming)
- [ ] Refactoring (no functional changes, no api changes)
- [ ] Build related changes
- [ ] Documentation content changes
- [ ] Other (please describe): 

## What is the current behavior?

Color variant is passed by `breadcrumbs`. However, `breadcrumb` children are not using the variant when using Vue. Notice that the `breadcrumb` components do not reflect the red color from `danger`.

Screenshot 2023-03-28 at 11 50 53 AM

Screenshot 2023-03-28 at 11 51 04 AM

Issue URL: resolves #25446

## What is the new behavior?

- `color` prop uses `reflect: true` to verify that it will be rendered in the DOM as an HTML attribute

## Does this introduce a breaking change?

- [ ] Yes
- [x] No

## Other information

- Demo was added to the `packages/vue` under the `Breadcrumbs` view.
- Test was added to verify that the color attribute is in the DOM.

Screenshot 2023-03-28 at 11 45 27 AM

PR #27045, opened 2023-03-29
chore: sync script deletes existing tgz packages

sean-perkins:
## Pull request checklist

## Pull request type

 

 

- [ ] Bugfix
- [ ] Feature
- [x] Build related changes

When testing changes locally, you can have an existing test app that you sync the package contents to. If you have done this across different versions of Ionic, it can install the wrong .tgz file instead of the local changes. 

Experienced here: https://github.com/ionic-team/ionic-framework/pull/27040#pullrequestreview-1362136995

Issue URL: N/A

## What is the new behavior?

- Sync script deletes all .tgz files local to the directory before locally packing and installing the contents of the parent local packages

- [ ] Yes
- [x] No


packages/vue/scripts/sync.sh
```diff
@@ -2,6 +2,9 @@
 
 set -e
 
+# Delete old packages
+rm -rf *.tgz
```
liamdebeasi, line 6: Will this work on Windows?

sean-perkins: Unfortunately no. I'm not familiar with Windows bash commands, do you know if something like this would work?

```
if [[ "$(expr substr $(uname -s) 1 6)" == "CYGWIN" ]]; then
  del /s *.tgz
else
  rm -rf *.tgz
fi
```

liamdebeasi: I've used [rimraf](https://www.npmjs.com/package/rimraf) before and it's worked pretty well. Maybe we could do that and have `rimraf` as a project-level dependency?

amandaejohnston: `rm -rf` works just fine 👀 I just have to do it in a bash terminal (which is already the case with the rest of the script).

amandaejohnston (approved):

packages/vue/scripts/sync.sh
liamdebeasi: Perfect! In that case, let's just stick with `rm`.

liamdebeasi (approved):
Nit: Do we even need the recursive flag here since we're only deleting archives and not directories?

Otherwise this looks good. Thanks for taking care of this!

sean-perkins:
True, the recursive flag isn't needed here. I'll update that before merging.

PR #27125, opened 2023-04-05
fix(select): text does not overlap icon

liamdebeasi:

Issue URL: 

## What is the new behavior?

-
-
-

- [ ] Yes
- [ ] No

### DIFF
--- a/core/src/components/select/select.ios.vars.scss
+++ b/core/src/components/select/select.ios.vars.scss
@@ -19,6 +19,9 @@ $select-ios-padding-start:          $item-ios-padding-start !default;
 /// @prop - Color of the select icon
 $select-ios-icon-color:             $text-color-step-650 !default;
 
+/// @prop - Size of the select icon
+$select-ios-icon-size: 18px !default;
+
 /// @prop - Color of the select placeholder
 $select-ios-placeholder-color:      $select-ios-icon-color !default;
 
--- a/core/src/components/select/select.md.scss
+++ b/core/src/components/select/select.md.scss
@@ -25,7 +25,7 @@
 }
 
 .select-icon {
-  width: 13px;
+  width: $select-md-icon-size;
 
   transition: transform .15s cubic-bezier(.4, 0, .2, 1);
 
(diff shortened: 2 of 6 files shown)


core/src/components/select/select.ios.vars.scss
```diff
@@ -19,6 +19,9 @@ $select-ios-padding-start:          $item-ios-padding-start !default;
 /// @prop - Color of the select icon
 $select-ios-icon-color:             $text-color-step-650 !default;
 
+/// @prop - Size of the select icon
+$select-ios-icon-size: 18px !default;
```
brandyscarney, line 23:
```suggestion
$select-ios-icon-size:              18px !default;
```

Can we line up the variables, double check that my above suggestion looks aligned in the source file

core/src/components/select/select.md.vars.scss
```diff
@@ -19,6 +19,9 @@ $select-md-padding-start:       $item-md-padding-start !default;
 /// @prop - Color of the select icon
 $select-md-icon-color:          $text-color-step-600 !default;
 
+/// @prop - Size of the select icon
+$select-md-icon-size: 13px !default;
```
brandyscarney, line 23:
```suggestion
$select-md-icon-size:           13px !default;
```

Same as ios, aligning the variables. May need to adjust all to align with `$select-md-floating-label-padding`

core/src/components/select/select.vars.scss
```diff
@@ -2,3 +2,5 @@
 
 /// @prop - How much to scale the floating label by
 $select-floating-label-scale: 0.75 !default;
+
+$select-icon-margin-start: 4px !default;
```
brandyscarney, line 6:
```suggestion
/// @prop - Margin start of the select icon
$select-icon-margin-start:    4px !default;
```

core/src/components/select/select.md.scss
```diff
@@ -142,3 +142,11 @@
 :host(.select-shape-round) {
   --border-radius: 16px;
 }
+
+// Select Native Wrapper
+// ----------------------------------------------------------------
+
+:host(.select-label-placement-stacked) .native-wrapper,
+:host(.select-label-placement-floating) .native-wrapper {
+  width: calc(100% - $select-md-icon-size - $select-icon-margin-start);
```
brandyscarney, line 151: What was the reasoning behind calculating the width instead of using `padding-right` or `margin-right`?

brandyscarney (changes requested):

core/src/components/select/select.ios.vars.scss
liamdebeasi: We did away with this convention last year

brandyscarney (approved):

Issue #27130, opened 2023-04-06
bug: IonItem --background-activated is too dim on iOS


eugr:
### Prerequisites

- [X] I have read the [Contributing Guidelines](https://github.com/ionic-team/ionic-framework/blob/main/.github/CONTRIBUTING.md#creating-an-issue).
- [X] I agree to follow the [Code of Conduct](https://ionicframework.com/code-of-conduct).
- [X] I have searched for [existing issues](https://github.com/ionic-team/ionic-framework/issues) that already report this problem, without success.

### Ionic Framework Version

v7.x

### Current Behavior

On iOS, when you tap on a list item, it doesn't highlight because the default highlight color (step-150) in combination with activated background opacity results in the same black color as default background.

I've fixed this by defining the following style in variables.css:

```css
  .ios ion-item {
    --background-activated: var(--ion-color-step-500); /* because otherwise opacity is too low */
  }
```

However, it would be great if it was fixed on the framework level.

### Expected Behavior

Visual feedback is perceivable by a regular person with normal eyesight.

### Steps to Reproduce

1. Run any app that displays clickable list items on iOS simulator or real device with dark mode activated
2. Click on the item and observe lack of visual feedback

### Ionic Info

Ionic:

   Ionic CLI       : 7.0.1 (/opt/homebrew/lib/node_modules/@ionic/cli)
   Ionic Framework : @ionic/react 7.0.1

Capacitor:

   Capacitor CLI      : 4.7.3
   @capacitor/android : 4.7.3
   @capacitor/core    : 4.7.3
   @capacitor/ios     : 4.7.3

Utility:

   cordova-res : not installed globally
   native-run  : 1.7.2

System:

   NodeJS : v19.7.0 (/opt/homebrew/Cellar/node/19.7.0/bin/node)
   npm    : 9.5.0
   OS     : macOS Unknown

liamdebeasi:
Thanks for the report. I can reproduce this behavior. It looks like the activated color of the item is always `#000` even in dark mode. It should probably change to be `#fff` in dark mode so the item gets lighter when tapped (since the default background itself is `#000` on dark mode)

liamdebeasi:
Here is a dev build if you would like to give it a shot in your app: `7.0.2-dev.11680787869.125aa3a3`

eugr:
Just checked, seems to be working fine!

liamdebeasi:
[...]

PR #27134, opened 2023-04-06
fix(item): ios active state has correct contrast

liamdebeasi:

Issue URL: resolves https://github.com/ionic-team/ionic-framework/issues/27130

The active/focus background colors on ion-item for iOS are always #000. This allows for the correct contrast on light mode. This is a problem on dark mode because the item background color defaults to #000 as well.

## What is the new behavior?

- Updates the default active/focus colors to be the text color. This ensures that the active/focus colors are always the opposite of what the default item background color is.

- [ ] Yes
- [x] No

Note: We are updating the default values of CSS Variables, but our implementation of the spec is wrong and creates color contrast issues.


core/src/components/item/test/buttons/index.html
```diff
@@ -188,6 +347,11 @@ <h3>Avatar Start</h3>
     </style>
 
     <script>
+      const params = new URL(window.location).searchParams;
+      const dark = params.get('dark');
+      if (dark !== null) {
+        document.body.classList.add('dark');
+      }
```
liamdebeasi, line 354: Once we move to the generators + with Brandy's proposed dark mode testing improvement this should be no longer needed.

brandyscarney: Can we make a tech debt ticket for this dependent on FW-1774?

core/src/components/item/item.ios.scss
```diff
@@ -11,8 +11,18 @@
...
-  --background-activated: #000;
-  --background-focused: #000;
+  /**
+   * The active/focus states need
+   * to be the opposite of the background color
+   * so there is enough contrast. The background
+   * defaults to $background-color, so we
+   * default to $text-color which will give us
+   * the opposite color.
+   * i.e. is $background-color is #000
```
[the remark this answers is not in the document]
brandyscarney: I am not sure if this comment makes total sense. The `$background-color` and `$text-color` variables don't change, it will only change when the user sets `--ion-background-color` or `--ion-text-color`.

brandyscarney (commented):
Looks good!

brandyscarney:
Here's the before & after of this change:

Frame 1

brandyscarney (approved):

darrylnoakes:
Will these sort of fixes be backported to 6.x? Or should I apply them manually?

liamdebeasi:
This fix will not be backported to 6.x. We typically only backport fixes that were in progress before Ionic 7 was released.

PR #27142, opened 2023-04-07
docs(input,searchbar): setFocus usage within overlays

sean-perkins:

The `setFocus` method does not mention extra implementation details required when working within an overlay. 

Issue URL: N/A

## What is the new behavior?

- Adds additional detail about how to use `setFocus()` when inside an overlay
- Aligns the `ion-searchbar` documentation to match the `ion-input` documentation

- [ ] Yes
- [x] No

### DIFF
--- a/core/src/components.d.ts
+++ b/core/src/components.d.ts
@@ -1270,7 +1270,7 @@ export namespace Components {
          */
         "required": boolean;
         /**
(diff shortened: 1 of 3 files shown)


amandaejohnston (changes requested):
Could we specify the specific event/lifecycle hook that should be used? Otherwise people have to go digging for that.

amandaejohnston (approved):

PR #27154, opened 2023-04-10
chore(angular): test apps use modern label syntax

sean-perkins:

The Angular test apps use the legacy label syntax for the e2e test templates.

Issue URL: N/A

## What is the new behavior?

- Updates the test templates for the Angular test apps to use the modern syntax available in Ionic v7.

- [ ] Yes
- [x] No

### DIFF
--- a/angular/test/base/src/app/searchbar/searchbar.component.html
+++ b/angular/test/base/src/app/searchbar/searchbar.component.html
@@ -2,7 +2,8 @@
   <form [formGroup]="form">
     <ion-list>
       <ion-item>
-        <ion-searchbar label="Searchbar" formControlName="searchbar"></ion-searchbar>
+        <ion-label>Searchbar</ion-label>
+        <ion-searchbar formControlName="searchbar"></ion-searchbar>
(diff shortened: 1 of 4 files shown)


angular/test/base/src/app/searchbar/searchbar.component.html
```diff
@@ -2,8 +2,7 @@
   <form [formGroup]="form">
     <ion-list>
       <ion-item>
-        <ion-label>Searchbar</ion-label>
-        <ion-searchbar formControlName="searchbar"></ion-searchbar>
+        <ion-searchbar label="Searchbar" formControlName="searchbar"></ion-searchbar>
```
liamdebeasi, line 5: Searchbar doesn't have a label prop

liamdebeasi, line 5: Searchbar doesn't have a label prop

liamdebeasi (changes requested):

liamdebeasi (approved):

PR #27155, opened 2023-04-10
fix(segment): segment disables segment buttons created asyncronously

amandaejohnston:

Issue URL: 

## What is the new behavior?

-
-
-

- [ ] Yes
- [ ] No

### DIFF
--- a/core/src/components/segment/test/basic/segment.e2e.ts
+++ b/core/src/components/segment/test/basic/segment.e2e.ts
@@ -58,4 +58,31 @@ test.describe('segment: basic', () => {
       await expect(segment).toHaveScreenshot(`segment-color-${page.getSnapshotSettings()}.png`);
     });
   });
+
+  test.describe('segment: behavior', () => {
+    test.only('segment buttons should be disabled when added to disabled segment asyncronously', async ({ page, skip }) => {
+      skip.rtl();
+      skip.mode('ios');
+
+      test.info().annotations.push({
+        type: 'issue',
+        description: 'https://github.com/ionic-team/ionic-framework/issues/25396',
+      });
+
+      await page.setContent(`
+        <ion-segment disabled="true"></ion-segment>
+        <script>
+          const segment = document.querySelector('ion-segment');
+          setTimeout(() => {
+            segment.insertAdjacentHTML('beforeend', '<ion-segment-button>Segment</ion-segment-button>');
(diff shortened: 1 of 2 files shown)


core/src/components/segment/test/basic/segment.e2e.ts
```diff
@@ -58,4 +58,31 @@ test.describe('segment: basic', () => {
...
+        description: 'https://github.com/ionic-team/ionic-framework/issues/25396',
+      });
+
+      await page.setContent(`
+        <ion-segment disabled="true"></ion-segment>
+        <script>
+          const segment = document.querySelector('ion-segment');
+          setTimeout(() => {
+            segment.insertAdjacentHTML('beforeend', '<ion-segment-button>Segment</ion-segment-button>');
+          }, 500); // shorter times don't trigger the issue consistently enough
```
liamdebeasi, line 78: Could we have this logic controlled by the test rather than set on a timeout?

liamdebeasi (changes requested):

core/src/components/segment/test/basic/segment.e2e.ts
liamdebeasi: Rather than wait for an arbitrary timeout, the test itself should append the `ion-segment-button`. The problem with the timeout is it's not always going to be called right at 500ms, so the test can potentially be flaky.

Since we are testing component functionality, it might be better to make this a unit test instead:

```js
import { Segment } from '../../segment';
import { SegmentButton } from '../../../segment-button/segment-button';

it('should render my component', async () => {
    components: [Segment, SegmentButton],
    html: `<ion-segment disabled="true"></ion-segment>`,
  });

  const segment = page.body.querySelector('ion-segment');
  segment.innerHTML = `<ion-segment-button>Segment Button</ion-segment-button>`;
  await page.waitForChanges();

  const segmentButton = page.body.querySelector('ion-segment-button');
  expect(segmentButton.disabled).toBe(true);
});
```

This should also run much faster than an E2E test.

liamdebeasi: I was able to get the bug to trigger using the spec test (and by removing your fix). Stencil should wait for the DOM to be ready before `newSpecPage` resolves, so by the time you append the segment button the parent segment should have initialized already.

liamdebeasi (approved):

Issue #27085, opened 2023-04-02
bug: v7, cannot pass dynamic label to form controls


judsonmusic:
### Prerequisites

- [X] I have read the [Contributing Guidelines](https://github.com/ionic-team/ionic-framework/blob/main/.github/CONTRIBUTING.md#creating-an-issue).
- [X] I agree to follow the [Code of Conduct](https://ionicframework.com/code-of-conduct).
- [X] I have searched for [existing issues](https://github.com/ionic-team/ionic-framework/issues) that already report this problem, without success.

### Ionic Framework Version

v7.x

### Current Behavior

Upgraded to IONIC 7 RC a couple of day ago. I got all of the warnings about the ion label being deprected and needing to replace with label="something" labelPlacement="stacked" etc...

I have a reusable component in which i am passing the label via attribute like so:
` <app-frequency-select
          controlName="medfrequency"
          label="When will this be taken?"
          [required]="true"></app-frequency-select>`

Inside the component you see i'm trying to pass an input to customize the value of the select.
HTML: 
```
<ion-item>
  <ion-select
    required
    [label]="label"
    labelPlacement="stacked"
    #frequencySelect
    [formControlName]="controlName"
    placeholder="Please Select"
  >
    <ion-select-option [value]="f.frequencyid" *ngFor="let f of frequencies">{{ f.frequency }}</ion-select-option>
  </ion-select>
</ion-item>
```

component.ts:

```
@Component({
  selector: 'app-frequency-select',
  standalone: true,
  templateUrl: './frequency-select.component.html',
  styleUrls: ['./frequency-select.component.scss'],
  imports: [CommonModule, IonicModule, FormsModule, ReactiveFormsModule, RequiredDirective, SelectLabelFixDirective],
  viewProviders: [
    {
      provide: ControlContainer,
      useFactory: (container: ControlContainer) => container,
      deps: [[new SkipSelf(), ControlContainer]]
    }
  ]
})
export class FrequencySelectComponent {
  @ViewChild('frequencySelect') fs: IonSelect;
  @Input() label: string;
  @Input() required: boolean = false;
  @Input() controlName: string;
  frequencies: any[] = [];
  disabled: boolean;

  constructor(private _referenceService: ReferenceService) {}

  async ngOnInit() {
    await this.buildReferences();
    console.log('The label for FS:', this.label); //this is coming through just fine..
  }
  async buildReferences() {
    const t = setTimeout(async () => {
[...]

PR #27156, opened 2023-04-10
fix(many): dynamic label support for modern form controls

sean-perkins:

Developers that are using Ionic v7 are experiencing an issue where implementations that are intended to use the modern control syntax will render with the legacy syntax and a warning will be displayed.

The issue is most easily reproduced by not assigning a label to the control and then asynchronously assigning a label after a duration. 

Angular example:
```html
<ion-item>
  <ion-input [label]="label"></ion-input>
</ion-item>
```

```ts
@Component({ ... })
export class MyComponent {
  @Input() label?: string; // initially unset

  ngOnInit() {
    setTimeout(() => this.label = 'Hello world', 500);
  }
}
```

Issue URL: #27085

## What is the new behavior?

- Form controls will attempt to render modern syntax by default and will fallback to the legacy syntax if an `ion-item` is found within the closest `ion-item` (containing the control).
- Modern form syntax supports dynamically set labels

- [ ] Yes
- [x] No

While this change does not immediately impose a breaking change, developers with complex conditional rendering of labels next to controls or dynamically constructing forms should validate that their UI is rendering correctly after this change (only if your application requires the legacy syntax). It would be recommended to apply `legacy="true"` to these form controls in your template, if this scenario applies to you. 

## Other information

Dev-build: `7.0.2-dev.11681149441.1ec983b5`

Notes:

Working on tests.


core/src/components/input/test/item/input.e2e.ts
```diff
@@ -36,4 +36,35 @@ test.describe('input: item', () => {
...
+      <ion-button onclick="setLabel()">Set label</ion-button>
+
+      <script>
+        function setLabel() {
+          const input = document.querySelector('ion-input');
+          input.label = 'New label';
+        }
+      </script>
+      `
+    );
```
liamdebeasi, line 58: Any particular reason why you did not do the following instead?

```js
await input.evaluate((el: HTMLIonInputElement) => el.value = 'New label');
```

liamdebeasi (changes requested):
There are quite a few tests using the legacy syntax that should be using the modern syntax. Is this intentional? My understanding of https://github.com/ionic-team/ionic-framework/issues/27199 is this only impacts ranges used inside of `ion-item`, but these impacted ranges are not being used in an item.

core/src/components/input/test/item/input.e2e.ts
sean-perkins: I _think_ you meant `el.label`, but yes this can be simplified to use Playwright's `evaluate` method to change the label vs. having the test template have a script for it. I'll update that.

The only tests that should be using the legacy syntax are the range tests (for now). This is the same behavior as `main` today (these screenshot tests are capturing the legacy output). While migrating the range tests I ran into the clipping bug: #27199, which I decided against committing the buggy screenshots. I can split that PR out if that would help with this review.

liamdebeasi: Yep, I meant el.label`. Good catch!

liamdebeasi (approved):

PR #27160, opened 2023-04-10
refactor(many): use utils import

thetaPC:

Does not use the alias for utils import.

Issue URL: N/A

## What is the new behavior?

- Uses the alias, `@utils/*`, for the imports

- [ ] Yes
- [x] No

## Other information

N/A


liamdebeasi (commented):
Can we create a dev build and test this in a Stencil project? I'm fairly certain the fix for the bug that prevented us from doing this originally already shipped in Stencil, but we've been bitten by this more than once so would like to do an extra check before we merge.

edit: Would be good to try it with a Stencil unit test in that same project too.

liamdebeasi:
A few questions:

1. What is the output when running `npm ls @ionic/core`?
2. What is the output when running `npm ls @stencil/core`?

thetaPC:
> 2\. npm ls @stencil/core

Here's a screenshot. Let me know if you prefer a text version.
Screenshot 2023-04-11 at 11 05 01 AM

thetaPC:
> Are you able to isolate the files where the warnings are coming from?

I was able to find the file that was causing the issues. I've reverted the paths. No console errors were found when installing the latest dev build (`@ionic/core@7.0.2-dev.11681240978.1b9c00ca`). Also `npm run test` runs without issues now.

The issues were coming from:
1. awaits: [await example](https://github.com/ionic-team/ionic-framework/blob/main/core/src/components/item-sliding/item-sliding.tsx#L77), [then example](https://github.com/ionic-team/ionic-framework/blob/main/core/src/components/app/app.tsx#L22)
2. [types example](https://github.com/ionic-team/ionic-framework/blob/main/core/src/components/action-sheet/action-sheet.tsx#L19)
3. [constants example](https://github.com/ionic-team/ionic-framework/blob/main/core/src/components/modal/modal.tsx#L19)

amandaejohnston:
Since this is a pretty sweeping change, it may be good to target this to a minor release instead, in case of surprise hidden consequences. We've had a few times where the PR shouldn't have any functional changes, but some edge case was missed that caused regressions, just due to the difficulty of testing this many file updates. Shipping the update in a minor release makes it less likely that devs will be caught off-guard by any regressions, since patch releases are supposed to be pretty safe.

If that's agreeable, we would just need to create a `feature-7.1` branch off of `main` (it doesn't look like it exists yet) and use that as the base branch for this PR instead of `main`.

thetaPC:
> Since this is a pretty sweeping change, it may be good to target this to a minor release instead, in case of surprise hidden consequences. We've had a few times where the PR shouldn't have any functional changes, but some edge case was missed that caused regressions, just due to the difficulty of testing this many file updates. Shipping the update in a minor release makes it less likely that devs will be caught off-guard by any regressions, since patch releases are supposed to be pretty safe.
> 
> If that's agreeable, we would just need to create a `feature-7.1` branch off of `main` (it doesn't look like it exists yet) and use that as the base branch for this PR instead of `main`.

I'll go ahead and do that, thanks!

thetaPC:
Merged into [feature-7.1](https://github.com/ionic-team/ionic-framework/tree/feature-7.1) branch.

Issue #27159, opened 2023-04-10
bug: v7 `ion-list` + `ion-radio` with new label behavior has a smaller interactive zone than in v6


Julien-Marcou:
### Prerequisites

- [X] I have read the [Contributing Guidelines](https://github.com/ionic-team/ionic-framework/blob/main/.github/CONTRIBUTING.md#creating-an-issue).
- [X] I agree to follow the [Code of Conduct](https://ionicframework.com/code-of-conduct).
- [X] I have searched for [existing issues](https://github.com/ionic-team/ionic-framework/issues) that already report this problem, without success.

### Ionic Framework Version

v7.0.1

### Current Behavior

Inside an `<ion-list>` When using the new `label` slot of the `<ion-radio>` instead of an `<ion-label>` + an `<ion-radio>`, the interactive (clickable) zone of the `<ion-radio>` is smaller than with Ionic v6.

When you hover & click an item of the list, you still get the "hover" + "ripple" effects, but the radio button does not get selected unless you click the correct part of it.

In v6:

```html
<ion-list>
  <ion-radio-group>
    <ion-item>
      <ion-label>Label 1</ion-label>
      <ion-radio value="1"></ion-radio>
    </ion-item>
    <ion-item>
      <ion-label>Label 2</ion-label>
      <ion-radio value="2"></ion-radio>
    </ion-item>
  </ion-radio-group>
</ion-list>
```

In v7:

```html
<ion-list>
  <ion-radio-group>
    <ion-item>
      <ion-radio value="1">Label 1</ion-radio>
    </ion-item>
    <ion-item>
      <ion-radio value="2">Label 2</ion-radio>
    </ion-item>
  </ion-radio-group>
</ion-list>
```

### Expected Behavior

You should be able to click anywhere in the `<ion-item>` and it should select the radio button (like it did in Ionic v6).

(Or the `<ion-radio>` should take the entire size of the `<ion-item>`).

### Steps to Reproduce

1. Create a `<ion-list>` with `<ion-item>` & `<ion-radio>` inside of it
2. Hover an `<ion-item>` until you get the "hover" effect, but not the "pointer" cursor
3. Click the `<ion-item>`
4. The ripple effect is triggered, but the `<ion-radio>` does not get selected

---

Here is what I get In Ionic v7:

https://user-images.githubusercontent.com/4369075/230986500-073dfbc9-8eab-4b66-bd77-807ead8a2bb7.mp4

---

Here is what I had in Ionic v6:

[...]

PR #27168, opened 2023-04-11
fix(radio): takes up full height in item

liamdebeasi:

Issue URL: resolves https://github.com/ionic-team/ionic-framework/issues/27159

## What is the new behavior?

- Radio elements take up the full height when used inside of an item. This aligns with what checkbox and toggle do:

Checkbox: https://github.com/ionic-team/ionic-framework/blob/cb8f07c5530ffc222580a3e3bae8dc85f62c73e5/core/src/components/checkbox/checkbox.scss#L42
Toggle: https://github.com/ionic-team/ionic-framework/blob/cb8f07c5530ffc222580a3e3bae8dc85f62c73e5/core/src/components/toggle/toggle.scss#L43

- [ ] Yes
- [x] No

## Other information

Dev build: `7.0.2-dev.11681223603.14da7831`

### DIFF
--- a/core/src/components/radio/radio.scss
+++ b/core/src/components/radio/radio.scss
@@ -72,6 +72,7 @@ input {
 
 :host(.in-item:not(.legacy-radio)) {
   width: 100%;
+  height: 100%;
 }
 
 :host(.in-item[slot="start"]:not(.legacy-radio)),


mapsandapps (approved):

brandyscarney:
While this fixes the height, the width is still not 100% which doesn't allow selecting on the far left:

liamdebeasi:
The start and end padding is being applied by the item. FW-3064 (in this current sprint) is dedicated to find a solution for that.

liamdebeasi:
I've filed a separate issue to track the start/end margin bug: https://github.com/ionic-team/ionic-framework/issues/27169

Going to merge this since the user has verified the fix: https://github.com/ionic-team/ionic-framework/issues/27159#issuecomment-1503547502

PR #27182, opened 2023-04-12
refactor(input): migrate composition events to JSX bindings

brandyscarney:
Updates the `compositionstart` and `compositionend` events on input to use JSX bindings since issue https://github.com/ionic-team/stencil/issues/3235 was resolved in Stencil v3.


liamdebeasi (approved):
Looks good. Both events are being bound to `ion-input`: 
image

PR #27185, opened 2023-04-12
fix(datetime): clamp date between min and max when using month picker

### LINKED ISSUE #27027: bug: ion-datetime does not pin current date to min/max when switching months
### Prerequisites

### Ionic Framework Version

v6.x

### Current Behavior

`ion-datetime` does not pin the date to the min/max when switching months with the month picker. If the day of the month is out of range when switching months, a warning is logged to the console and the current date is out of range.

### Expected Behavior

The current date should be pinned to the min/max, as it is when using `ion-datetime-button`.

### Steps to Reproduce

1. Open the reproduction.
2. Open the console.
3. Select April 1.
4. Open the month picker and switch to March. A warning is logged to the console and the current date is March 1, which is out of range.
5. Select March 30.
6. Open the month picker and switch to May. A warning is logged to the console and the current date is May 30, which is out of range.

### Code Reproduction URL

https://stackblitz.com/edit/angular-7s6sah?file=src/components/Example.vue

### Ionic Info

In Stackblitz:

Ionic:

   Ionic CLI : 6.20.9

Utility:

System:

   NodeJS : v16.14.2
   npm    : 7.17.0
   OS     : Linux 5.0

### Additional Information

Since `ion-datetime-button` does pin the date to the min/max, it seems that `ion-datetime` should do the same when switching months.

Originally reported in #26904.

amandaejohnston:

When using the dropdown month picker, the value is updated when the month is changed, to follow native behavior. However, because only the month is updated (the day remains the same), it's possible for the newly chosen date to fall outside the min/max bounds of the datetime.

For example, if you have a datetime with `min="2021-01-15" value="2021-02-01"`, then use the month picker to switch to January, the new value will be `2021-01-01` which is earlier than the `min`.

Issue URL: Resolves #27027

## What is the new behavior?

When updating the `activeParts` in any scenario, the date to set is now clamped between the max and min.

- [ ] Yes
- [x] No

### DIFF
--- a/core/src/components/datetime/datetime.tsx
+++ b/core/src/components/datetime/datetime.tsx
@@ -1679,7 +1679,7 @@ export class Datetime implements ComponentInterface {
       return [];
     }
 
-    const { minParts, maxParts, workingParts } = this;
+    const { workingParts } = this;
 
     const activePart = this.getActivePartsWithFallback();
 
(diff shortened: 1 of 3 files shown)


liamdebeasi (commented):
Would it make sense to clamp the date in `setActiveParts` instead? That would ensure that all dates set are guaranteed to be valid.

amandaejohnston:
Looks like that works!

liamdebeasi (approved):

sean-perkins (approved):

madalinporojan2010:
It still reproduces,
ionic v7.7.4
https://angular-e3blvt.stackblitz.io

amandaejohnston:
@madalinporojan2010 Please create a new issue with details: https://github.com/ionic-team/ionic-framework/issues/new/choose

PR #27186, opened 2023-04-12
test(item): use modern form syntax

sean-perkins:

There are tests for `ion-item` that use the legacy syntax for form controls, without being in the `/legacy` test folder. 

Issue URL: N/A

## What is the new behavior?

- Updates the tests to use the modern form syntax
- Removes the `fill` tests, which are already tested in the `inputs` test folder
- Removes invalid control UI from test templates (i.e. rendering a segment with other controls in a single item).

- [ ] Yes
- [x] No


sean-perkins:
I re-added those tests, but moved them into a `/legacy` folder, since they deal with the legacy control appearance.

liamdebeasi (approved):

PR #27188, opened 2023-04-13
 fix(range): range knob does not overflow in item

sean-perkins:

The range knob will clip when displayed without a label inside of `ion-item`.

Additionally, there are legacy range tests that do not use the correct syntax.

Issue URL: #27199

## What is the new behavior?

- Range knob does not clip when inside of `ion-item` with the modern form syntax
- Legacy range tests specify `legacy="true"`
- Modern range tests use modern form control syntax

- [ ] Yes
- [x] No


sean-perkins:
The updated syntax/screenshots appear to have caught a visual bug with the knob positioning/clipping when used without a label. I'll circle back to confirming that + opening a bug report if that is the case.

thetaPC (changes requested):
I didn't review the tests since I expect them to change after resolving the conflicts.

sean-perkins:
Ok, I dug into this further and I am no longer concerned with that screenshot diff. The reason for clipping is that the bounding box does not contain the knob overflow:

CleanShot 2023-06-23 at 11 02 09

So when playwright captures the element screenshot, it results in the clipped appearance. In an application, this should not occur (unless they have a container with `overflow: hidden` directly around the `ion-range`. 

Once CI passes, I think this is ready for review to see if the team is in favor of the approach for solving this bug. Open to other ideas/suggestions as well 👍

core/src/components/range/test/label/range.e2e.ts-snapshots/range-no-items-no-label-ios-ltr-Mobile-Chrome-linux.png
thetaPC, line 1: Could we wrap the range in a container through the test file? The container could have padding. If we take a screenshot of the container, it should show the entire knob. Would that work?

thetaPC (commented):
Everything looks good. One thing that I noticed is that the focused blue circle is getting clipped when there isn't an icon when moving the knob either far right or far left.

Is it possible to have the focused circle to render on top?

https://github.com/ionic-team/ionic-framework/assets/13530427/525f530f-44ac-49ea-9796-c54a6594738a

sean-perkins:
@thetaPC here's an approach we could take:

https://github.com/ionic-team/ionic-framework/assets/13732623/9d958180-aedb-4f8c-9a50-606b7cc8b636

It requires bumping the default padding from `14px` to `18px` on MD. This does cause the usage without a start icon or label to be padded an additional 4px to offset for the focus ring.

Thoughts? I think if we do this, we should target for a minor release since we are adjusting default values in pursuit of correcting visual issues.

thetaPC:
> @thetaPC here's an approach we could take:
> Kapture.2023-07-12.at.15.39.11.mp4
> 
> It requires bumping the default padding from `14px` to `18px` on MD. This does cause the usage without a start icon or label to be padded an additional 4px to offset for the focus ring.
> 
> Thoughts? I think if we do this, we should target for a minor release since we are adjusting default values in pursuit of correcting visual issues.

2 pixels feels like a lot. I do agree that if we make this change then it should be a minor.

I'm leaning towards starting a with the rest of the team for their thoughts. Otherwise, your solution works but wondering if it could lead to issues later.

sean-perkins:
Closing this PR in favor of https://github.com/ionic-team/ionic-framework/pull/28199

PR #27198, opened 2023-04-13
fix(toast): screen readers announce content

### LINKED ISSUE #25866: bug(a11y): assertive toast not announced
### Prerequisites

### Ionic Framework Version

- [ ] v4.x
- [ ] v5.x
- [X] v6.x
- [ ] Nightly

### Current Behavior

The toast message in not announced by screenreaders even with aria-live='assertive'.

Tested with:
- Ionic 6.2.4
- NVDA 2022.2.2 on Chrome and Firefox
- JAWS 2022.2207.25 ILM on Chrome and Firefox

### Expected Behavior

According to [ARIA live regions spec](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Live_Regions):

> aria-live="assertive" should only be used for time-sensitive/critical notifications that absolutely require the user's immediate attention. Generally, a change to an assertive live region will interrupt any announcement a screen reader is currently making. As such, it can be extremely annoying and disruptive and should only be used sparingly.

So I expect assertive toast message to be announced when toast is shown.

### Steps to Reproduce

- Checkout v.6.2.4
- npm install in ./core
- npm start
- go to http://localhost:3333/src/components/toast/test/a11y using Chrome/Firefox
- launch NVDA/JAWS
- click on 'Present Assertive Toast'
- no announcement

I could provide a demo app using @ionic/angular, but I need to test the toast behaviour with JAWS (with NVDA on Chrome/Firefox is not working).

### Code Reproduction URL

_No response_

### Ionic Info

Ionic:                        
                              
   Ionic CLI : 5.4.16         
                              
Utility:                      
                              
   cordova-res : not installed
   native-run  : not installed
                              
System:                       
                              
   NodeJS : v14.17.1          
   npm    : 6.14.13           
   OS     : Linux 4.4         

### Additional Information

_No response_

liamdebeasi:
Issue URL: resolves #25866

---------

Docs PR: https://github.com/ionic-team/ionic-docs/pull/2914

 

 

NVDA is not announcing toasts on present.

## What is the new behavior?

- Toast has a "status" role and "polite" announcement.
- We also revisited the intended behavior of toasts to better align with the Material Design v2 spec: https://m2.material.io/components/snackbars/web#accessibility

- [ ] Yes
- [x] No

## Other information

Dev build: 7.0.3-dev.11681482468.19d7784f

### DIFF
--- a/core/src/components/toast/toast.tsx
+++ b/core/src/components/toast/toast.tsx
@@ -55,6 +55,11 @@ export class Toast implements ComponentInterface, OverlayInterface {
 
   presented = false;
 
+  /**
+   * When `true`, content inside of .toast-content
+   * will have aria-hidden elements removed causing
+   * screen readers to announce the remaining content.
+   */
   @State() revealContentToScreenReader = false;
 
   @Element() el!: HTMLIonToastElement;
@@ -413,6 +418,11 @@ export class Toast implements ComponentInterface, OverlayInterface {
     );
   }
 
+  /**
+   * Render the `message` property.
+   * @param key - A key to give the element a stable identity. This is used to improve compatibility with screen readers.
+   * @param ariaHidden - If "true" then content will be hidden from screen readers.
+   */
   private renderToastMessage(key: string, ariaHidden: 'true' | null = null) {
     const { customHTMLEnabled, message } = this;
     if (customHTMLEnabled) {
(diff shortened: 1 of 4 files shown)


core/src/components/toast/toast.tsx
```diff
@@ -470,13 +508,42 @@ export class Toast implements ComponentInterface, OverlayInterface {
...
+                animation.
+
+                The "key" attribute is used here to force Stencil to render
+                new nodes and not re-use nodes. Otherwise, NVDA would not
+                detect any changes to the live region.
+
+                The "old" content is hidden using aria-hidden otherwise
+                VoiceOver will announce the toast content twice when presenting.
+              */}
+              {!revealContentToScreenReader && this.header !== undefined && this.renderHeader('oldHeader', 'true')}
```
sean-perkins, line 541: `this.header` and `this.message` are used a few times, should we deconstruct the variable reference above or is there a reason we need to access it directly in each check?

liamdebeasi: Nope, we can deconstruct these. Fixed in https://github.com/ionic-team/ionic-framework/pull/27198/commits/ab37f25d8bb9997c90b7026c1109795ebcccaf3f

Issue #27086, opened 2023-04-02
bug: Label not aligned properly between ion input and ion select when in card component


judsonmusic:
### Prerequisites

- [X] I have read the [Contributing Guidelines](https://github.com/ionic-team/ionic-framework/blob/main/.github/CONTRIBUTING.md#creating-an-issue).
- [X] I agree to follow the [Code of Conduct](https://ionicframework.com/code-of-conduct).
- [X] I have searched for [existing issues](https://github.com/ionic-team/ionic-framework/issues) that already report this problem, without success.

### Ionic Framework Version

v7.x

### Current Behavior

Notice in the screen shot when every i put inputs inside of an ion card side by side the select label is pushed down by a few pixels.
<img width="779" alt="image" src="https://user-images.githubusercontent.com/2080617/229339149-cc9002ee-873c-4ffc-ab4b-e5c1f133f12a.png">

Code:
```
 <form>
    <ion-card #cardmedtype>
      <ion-card-header color="primary">
        <ion-card-title><strong>Choose a Medication Type</strong></ion-card-title>
      </ion-card-header>
      <ion-card-content>
        <ion-grid>
          <ion-row>
            <ion-col>
              <ion-item>
                <ion-input label="input" labelPlacement="stacked"></ion-input>
              </ion-item>
            </ion-col>
            <ion-col>
              <ion-item>
                <ion-select label="select" labelPlacement="stacked"></ion-select>
              </ion-item>
            </ion-col>
          </ion-row>
        </ion-grid>
      </ion-card-content>
    </ion-card>
  </form>
```

However removing it from the ion card solves the issue. 

<img width="1129" alt="image" src="https://user-images.githubusercontent.com/2080617/229339241-e921f1aa-1903-4e3a-a19e-cd1ed1f3a703.png">

Code:
```
  <form>
    <ion-grid>
      <ion-row>
        <ion-col>
          <ion-item>
            <ion-input label="input" labelPlacement="stacked"></ion-input>
          </ion-item>
        </ion-col>
        <ion-col>
          <ion-item>
            <ion-select label="select" labelPlacement="stacked"></ion-select>
          </ion-item>
        </ion-col>
      </ion-row>
    </ion-grid>
  </form>
```

### Expected Behavior

I should be able to see the labels aligned on the same horizontal axis regardless of whether or not they are in a card.

### Steps to Reproduce

Create ion input with label and ion select with label side by side using ion-col (grid)
You will see the label are aligned horizontally.
[...]

PR #27202, opened 2023-04-13
fix(select): adjust label alignment when in a card

mapsandapps:

When an `ion-select` is inside a card, the label is slightly lower than it should be, making it out of alignment with the label of an `ion-input`.

Issue URL: https://github.com/ionic-team/ionic-framework/issues/27086

## What is the new behavior?

- When inside a card, the labels of `ion-input` and `ion-select` are in line with each other. 

- [ ] Yes
- [x] No

## Other information

This bug was occurring because the label of ion-select is in the shadow DOM. [As recommended](https://github.com/ionic-team/ionic-framework/issues/27086#issuecomment-1494427311), this fix was also added to ion-input and ion-textarea in case those labels are also later moved to the shadow DOM.

Before:
Screenshot 2023-04-13 at 4 35 44 PM

After:
Screenshot 2023-04-13 at 4 36 02 PM

### DIFF
--- a/core/src/components/input/input.scss
+++ b/core/src/components/input/input.scss
@@ -299,6 +299,8 @@
   transition: background-color 15ms linear;
 
   background: var(--background);
+
+  line-height: normal;
 }
 
 // Input Native Wrapper
--- a/core/src/components/select/select.scss
+++ b/core/src/components/select/select.scss
@@ -213,6 +213,8 @@ button {
 
   background: var(--background);
 
+  line-height: normal;
+
   cursor: inherit;
 
   box-sizing: border-box;
--- a/core/src/components/textarea/textarea.scss
+++ b/core/src/components/textarea/textarea.scss
@@ -260,6 +260,8 @@
   transition: background-color 15ms linear;
 
   background: var(--background);
+
+  line-height: normal;
 }
 
 // Textarea Native Wrapper


sean-perkins (commented):
Can we add screenshot tests for the select being rendered in a card, so that it doesn't regress in the future?

core/src/components/card/test/basic/card.e2e.ts
```diff
@@ -155,5 +155,12 @@ test.describe('card: basic', () => {
       const card = page.locator('ion-card');
       await expect(card).toHaveScreenshot(`card-no-content-or-header-${page.getSnapshotSettings()}.png`);
     });
+    test('should not have visual regressions with form elements inside card', async ({ page }) => {
+      await page.goto(`/src/components/card/test/form`);
+
+      await page.setIonViewport();
+
+      await expect(page).toHaveScreenshot(`card-form-${page.getSnapshotSettings()}.png`);
+    });
```
liamdebeasi, line 164: Can we take a screenshot of just the card instead of the entire page? Also, this test should probably be split into input, textarea, and select tests in the respective component directories.

liamdebeasi (changes requested):

core/src/components/card/test/basic/card.e2e.ts
liamdebeasi: Also we don't need Mode or RTL tests here, so we can skip them.

core/src/components/input/test/card/input.e2e.ts
```diff
@@ -0,0 +1,22 @@
...
+
+test.describe('input: card', () => {
+  test('should render correctly in card', async ({ page, skip }) => {
+    skip.rtl();
+    skip.mode('ios');
+
+    await page.setContent(`
+      <ion-card>
+        <ion-card-content>
+          <ion-item style="border: 1px solid grey; margin: 16px">
```
sean-perkins, line 12: Is the border for a visual reference for the label alignment? If so, what is the `margin: 16px` for?

~ Nit: Can we add `lines="none"` to the `ion-item` if we are going to draw a border around the container? Currently there are two bottom border lines.

liamdebeasi (approved):
+1 to Sean's comments, but otherwise looks good.

PR #27203, opened 2023-04-13
fix(item-sliding): options display on rtl

### LINKED ISSUE #25285: bug: :host selector on element combined with rtl mixins does not create the proper selector
### Prerequisites

### Ionic Framework Version

- [ ] v4.x
- [X] v5.x
- [X] v6.x
- [X] Nightly

### Current Behavior

With the following CSS:

```css
:host .element-selector {
    @include position(0, 0, null, null);
}
```

When using the `rtl` mixin so, also `add-root-selector` it translates the selector as:

```css
:host-context([dir=rtl])
```

So part of it is lost.

### Expected Behavior

With the following CSS:

```css
:host .element-selector {
@include position(0, 0, null, null);
}
```

When using the rtl mixin so, also add-root-selector it translates the selector as:

```css
:host-context([dir=rtl]) .element-selector
```

Maintaining the whole selector.

### Steps to Reproduce

1. Create a blank template
2. Add a div.test-element to the HTML
3. Add the following code to the related CSS:
```css
:host .test-element {
    @include position(0, 0, null, null);
}
```
4. Run the app, inspect code of the created div and check css is not included on the element.
5. You can also check the resulting selector is not correct on the generated css.

### Code Reproduction URL

_No response_

### Ionic Info

Ionic:

   Ionic CLI                     : 6.19.0 (/Users/pau/.nvm/versions/node/v16.14.2/lib/node_modules/@ionic/cli)
   Ionic Framework               : @ionic/angular 5.9.2
   @angular-devkit/build-angular : 0.1000.8
   @angular-devkit/schematics    : 10.0.8
   @angular/cli                  : 10.0.8
   @ionic/angular-toolkit        : 2.3.3

Cordova:

   Cordova CLI       : 11.0.0
   Cordova Platforms : android 10.1.1, ios 6.2.0
   Cordova Plugins   : cordova-plugin-ionic-keyboard 2.2.0, (and 22 other plugins)

Utility:

   cordova-res (update available: 0.15.4) : 0.15.3
   native-run (update available: 1.5.0)   : 1.4.0

System:

   ios-sim : 8.0.2
   NodeJS  : v16.14.2 (/Users/pau/.nvm/versions/node/v16.14.2/bin/node)
   npm     : 8.5.0
   OS      : macOS Monterey
   Xcode   : Xcode 13.3 Build version 13E113

### Additional Information

_No response_

### LINKED ISSUE #26103: bug: item-sliding, options don't display correctly in RTL mode on Firefox or Safari
### Prerequisites

### Ionic Framework Version

- [ ] v4.x
- [ ] v5.x
- [X] v6.x
- [ ] Nightly

### Current Behavior

In core, a basic `ion-item-sliding` will not show its options properly when opened in Firefox or Safari in RTL mode. Interestingly, using the `expandable` prop causes the options to display somewhat when dragging far enough, but the text is positioned incorrectly.

Screencast from Firefox: https://user-images.githubusercontent.com/90629384/195179136-fda916dd-14ad-4b30-a5f4-3e801ac9f2c9.mp4

### Expected Behavior

Screencast from Chrome: https://user-images.githubusercontent.com/90629384/195179307-1a31d302-7764-46c4-89e9-406097630444.mp4

### Steps to Reproduce

Host core locally and open http://localhost:3333/src/components/item-sliding/test/basic?rtl=true in Firefox or Safari, then drag any item open.

### Code Reproduction URL

http://localhost:3333/src/components/item-sliding/test/basic?rtl=true

### Ionic Info

```

Ionic:

   Ionic CLI : 6.18.1

Utility:

   native-run  : 1.4.1

System:

   NodeJS : v16.13.0
   npm    : 8.1.0
   OS     : Windows 10
```

### Additional Information

_No response_

thetaPC:
Issue URL: resolves #26103, resolves #25285

---------

 

 

Options in `item-sliding` will not display when using RTL with Firefox and Safari.

## What is the new behavior?

Issue was coming from `:host-context`. Firefox would keep [removing the entire](https://www.w3.org/TR/selectors-3/#grouping) compiled style when using this unsupported style. This would led to the RTL styles to not being applied to the component.

- Split the CSS group from `add-root-selector()`
- Added comments to make it easier to navigate through `add-root-selector()`
- Added `:dir()`

- [ ] Yes
- [x] No

## Other information

Co-authored-by: brandyscarney <brandyscarney@users.noreply.github.com>

- Updating `add-root-selector()` would also fix another [issue](https://github.com/ionic-team/ionic-framework/issues/25285) unintentionally


core/src/components/popover/popover.tsx
```diff
@@ -663,10 +664,11 @@ export class Popover implements ComponentInterface, PopoverInterface {
 
   render() {
     const mode = getIonMode(this);
-    const { onLifecycle, popoverId, parentPopover, dismissOnSelect, side, arrow, htmlAttributes } = this;
+    const { onLifecycle, popoverId, parentPopover, dismissOnSelect, side, arrow, htmlAttributes, el } = this;
     const desktop = isPlatform('desktop');
     const enableArrow = arrow && !parentPopover;
-
+    const rtl = isRTL(el) ? 'rtl' : 'ltr';
+    console.log(`popover-side-${rtl}`);
```
liamdebeasi, line 671:
```suggestion
```

core/src/components/alert/alert.ios.scss
```diff
@@ -286,13 +286,13 @@
   overflow: hidden;
 }
 
-.alert-button:first-child {
+:host .alert-button:first-child {
```
liamdebeasi, line 289: Why was this added?

core/src/components/popover/popover.ios.scss
```diff
@@ -75,17 +75,13 @@
   transform: rotate(180deg);
 }
 
-:host(.popover-side-start) .popover-arrow {
-  @include rtl() {
-    transform: rotate(-90deg);
-  }
+:host(.popover-side-start.popover-ltr) .popover-arrow,
```
liamdebeasi, line 78: Checking my understanding: We removed the `@include rtl` mixin usage in favor of explicit RTL/LTR classes to fix support for firefox/safari, correct?

core/src/themes/ionic.functions.string.scss
```diff
@@ -110,41 +94,74 @@
...
+          // .fixed -> [dir=rtl] :host(.fixed)
+          $updated-element: str-replace($updated-element, $updated-element, "#{$addHostSelector} :host(#{$updated-element})");
+
+          // Add the final selector after all transformations:
+          // :host(.fixed) -> [dir=rtl] :host(.fixed)
+          $new-element: append($new-element, $updated-element, space);
+        } @else {
+          // Add back any selectors that followed the host after transforming the
+          // first selector:
+          // :host(.fixed) ::slotted(ion-icon) -> [dir=rtl] :host(.fixed) ::slotted(ion-icon)
```
liamdebeasi, line 122: I'm not sure I understand the output comment here. `[dir=rtl] :host` won't match anything: https://codepen.io/liamdebeasi/pen/ExdgJzg

core/src/components/alert/alert.ios.scss
thetaPC: Without `host-context` then it leads to compiling the wrong css. Notice the lack of space between `[dir=rtl]` and `.sc-ion-alert-ios`

```css
[dir=rtl].sc-ion-alert-ios .alert-button.sc-ion-alert-ios:last-child {
 border-right:0.55px solid rgba(var(--ion-text-color-rgb, 0, 0, 0), 0.2)
}
```

By adding `:host`, the compiled css has the space:

```css
[dir=rtl] .sc-ion-alert-ios-h .alert-button.sc-ion-alert-ios:last-child {
 border-right:0.55px solid rgba(var(--ion-text-color-rgb, 0, 0, 0), 0.2)
}
```

This only happens when the component is [scoped](https://github.com/ionic-team/ionic-framework/blob/main/core/src/components/alert/alert.tsx#L43). 

If the component is not scoped or a shadow then it would have worked without `:host`:

```css
.alert-button:first-child {
  @include rtl() {
    ...
  }
}
```

core/src/components/popover/popover.ios.scss
thetaPC: Yes and no. When I removed `host-context` then the arrow was no longer displaying correctly for Chrome. Since popover is [using the shadow DOM](https://github.com/ionic-team/ionic-framework/pull/27203/files#diff-002321860b0f9ec9fd960b5274d8a90c3d7be3246b03be0510e154562ea104c3R161-R162), it doesn't know what direction the document it is so classes was my solution. This ended fixing it for Firefox and Safari.

core/src/themes/ionic.functions.string.scss
thetaPC: This outcome originated from [input outline](https://github.com/ionic-team/ionic-framework/blob/main/core/src/components/input/input.md.outline.scss#L177).

Without `dir` in the front then it wouldn't compile the correct styling which is the following:

```css
[dir=rtl] :host(.input-fill-outline) .input-outline-start {
...
}
```

liamdebeasi: So this only applies to non-shadow DOM components, correct?

thetaPC: That is correct. The new changes would cause `add-root-selector` to not work with scoped or shadow DOM components. The main reason that `host-context` was removed is because when Firefox sees it then it completely removes it along with the attached list.

```css
:host(.popover-side-end) .popover-arrow, :host-contenxt(...)... {
...
}
```

Firefox wouldn't even apply the example above. It sees the `host-context` and ignores the entire style (this is the cause of the issue on item-sliding).

Another approach to solving the issue is to determine if the browser is Chrome then add `host-context` to the list. Thoughts?

brandyscarney:
So I used the following input SCSS against the functions on both `main` and `FW-2608`:

```scss
// shadow element
:host {
  @include rtl() {
    color: blue;
  }
}

// shadow element targeting class
:host(.fixed) {
  @include rtl() {
    color: red;
  }
}

// shadow element targeting class and child icon
:host(.icon-hide) ::slotted(ion-icon) {
  @include rtl() {
    color: blue;
  }
}

// shadow element targeting class and child icon
:host(.icon-left) ::slotted(ion-icon) {
  @include rtl() {
    color: blue;
  }
}

// shadow element targeting child class
.shadow {
  @include rtl() {
    color: green;
  }
}

// non-shadow element
.not-shadow {
  @include rtl() {
    color: pink;
  }
}

:host .element-selector {
  @include position(0, 0, null, null);
}
```

On `main` the resulting selectors are separated by commas, so I split those out to their own rules and then compared that output with the output from this PR in order to make the diff easier to compare:

Screenshot 2023-04-28 at 1 41 50 PM
Screenshot 2023-04-28 at 1 41 56 PM

The diff looks good as all of the selectors that shouldn't have changed have not (besides splitting them up instead of having them comma separated which is intentional). The new selectors seem to have the desired changes.

I checked the `item-sliding` tests in Safari and Firefox and they seem to be working correctly in RTL.  

I think once the screenshots are passing this should be good to go.

core/src/themes/ionic.mixins.scss
```diff
@@ -193,8 +193,22 @@
...
+  // will be ignored.
+  // @link https://www.w3.org/TR/selectors-3/#grouping
+  @each $selector in $selectors {
+    @at-root #{$selector} {
+      @content;
+    }
+  }
 
-  @at-root #{add-root-selector($root, "[dir=rtl]")} {
+  @at-root #{$dirSelector} {
```
brandyscarney, line 211: The browser support for [:dir](https://caniuse.com/css-dir-pseudo) is pretty limited. Did we decide to add this since Chrome has `:host-context`? Maybe we should add a comment on why this is here.

thetaPC: It's more for a catch in case `:host-context()` doesn't work but the browser does support `:dir()`. I've added a comment as requested.

core/src/themes/ionic.mixins.scss
```diff
@@ -193,23 +193,33 @@
...
   }
 
-  @at-root #{$dirSelector} {
-    @content;
+  // Adding :dir() in case the browser doesn't support `:host-context()` and does support `:dir()`.
+  // `:host-context()` is added:
+  // - through the `add-root-selector()` function.
+  // - first so that it takes precedence over `:dir()`.
+  // For example, Firefox doesn't support `:host-context()`, but does support `:dir()`.
+  // MDN says that `:dir()` is supported by Safari, but it doesn't seem to work.
```
liamdebeasi, line 217: This feature was only supported starting in iOS 16.4: https://webkit.org/blog/13966/webkit-features-in-safari-16-4/

This may be why it did not work. Can we update the comment?

liamdebeasi: Actually question regarding this: Is `:dir` required for this fix to work? Last time I reviewed the change the `:dir` usage was not present.

brandyscarney (approved):

core/src/themes/ionic.mixins.scss
liamdebeasi: Did some digging and the answer is yes. In that case, can we clarify the comment as per above?

liamdebeasi: So Firefox/Safari will still work as expected if we remove `:dir`?

thetaPC: Yes, the only time that `:dir` shines is with a nested popover. The arrow doesn't rotate correctly on Safari and Firefox b/c they don't handle `:host-context`. With `:dir`, the arrow renders correctly on Firefox.
Honestly, with no major changes, the addition of `:dir` is a nice to have as of now.

thetaPC:
> I think change is looking really good so far!
> 
> One thing I am noticing is the bundle size can increase in some places because we are splitting up selectors: image
> 
> Is is possible to optimize the code a bit so that we can use multiple selectors for a single code block?
> 
> So in the screenshot above the preferred output would be:
> 
> ```css
> [dir=rtl] .range-knob-handle,
> :host-context([dir=rtl]) .range-knob-handle,
> .range-knob-handle:dir(rtl) {
>  ...
> }
> ```
> 
> I realize this task is taking longer than we originally pointed, so if you'd like to come back to it next sprint when we can re-point it that's fine too.

We can't group them back in that sense because the entire styling would be ignored on Firefox and Safari as soon as they detect `:hosts-context`. I did manage to group them up this way:

```css
:host-context([dir=rtl]) .range-knob-handle,
:host-context([dir=rtl]) .... {
 ...
}

[dir=rtl] .range-knob-handle,
[dir=rtl] ... {
 ...
}
```

All host-context will be grouped. Any vanilla selectors will be grouped. As for `:dir()`, those have also been grouped. But will be based on `@supports selector`.

core/src/themes/ionic.mixins.scss
liamdebeasi: Sounds good. Let's keep it as is since more browsers are starting to adopt `:dir`.

thetaPC: Awesome, then it's ready for you to review.

liamdebeasi (approved):
The screenshot test threshold was reduced from 0.2 to 0.1, so it got more sensitive. As a result, I updated your PR to pull in the latest `main` and re-ran the update screenshot job. We were able to capture the fix that you noted with popover arrows in RTL in https://github.com/ionic-team/ionic-framework/pull/27203/commits/5786f9204618b4028170fc094db9460ec3949866

Good to merge once CI passes. Great work!

PR #27207, opened 2023-04-14
fix(modal): set default text color

### LINKED ISSUE #26060: bug: modal text color is set incorrectly when app and system dark mode settings aren't the same
### Prerequisites

### Ionic Framework Version

- [ ] v4.x
- [ ] v5.x
- [X] v6.x
- [ ] Nightly

### Current Behavior

Hi!,

The CCS has strange behavior when you play with the dark and light theme of ionic and the operating system.

**dark ionic, dark windows OK**
dark-ionic-dark-windows-ok

**dark ionic, light windows KO**
dark-ionic-light-windows-ko1
dark-ionic-light-windows-ko2

**light ionic, light windows OK**
light-ionic-light-windows-ok

**light ionic, dark windows KO**
light-ionic-dark-windows-ko1
light-ionic-dark-windows-ko2

### Expected Behavior

It is expected that the elements be seen, and also pass the contrast color and have a good visibility.

### Steps to Reproduce

Create a new project, follow the dark theme guide: https://ionicframework.com/docs/theming/dark-mode#ionic-dark-theme. Remove the media query to use only the body.dark and change the theme both in ionic and in the operating system.

### Code Reproduction URL

https://github.com/Marius-Romanus/ionic-bug-css

### Ionic Info

@ionic/angular 6.2.9

### Additional Information

_No response_

thetaPC:

Occurs only when using modal with datetime: text does not update appropriately based on light or dark mode. For example, when using Ionic dark theme, the calendar days does not switch to white.

Issue URL: resolves #26060 

## What is the new behavior?

- Text (days and time) changes to the text color set via light (default) or dark mode. The color change applies to days and time that are not selected: not the current day.

- [ ] Yes
- [x] No

## Other information

Recommendation:
- Test both `datetime` and `datetime-button`

The bug originates when modal's content is not wrapped by `ion-content` and because the modal renders outside of `<main />`. Text color is passed to components properly inside `main` because `main` is rendered inside another `ion-content`. `ion-content` provides the correct text-color based on light or dark mode. 

Example of text color working for modal when wrapped by `ion-content`:

```html
<ion-modal trigger="open-modal">
    ...
    <div>Hello world</div>
  </ion-content>
</ion-modal>
```

However, there are times that the modal does not require this wrap:

```html
<ion-modal>
  <ion-datetime id="datetime"></ion-datetime>
</ion-modal>
```

The same bug would have occurred on `popover` if it wasn't for the fact that the component provides the [text color](https://github.com/ionic-team/ionic-framework/blob/main/core/src/components/popover/popover.scss#L42).


liamdebeasi (changes requested):
Can we write a test for this?

sean-perkins (commented):
Changes look good! +1 to what Liam commented. The datetime basic test template already has an option for enabling dark mode and presenting the datetime in a modal. You may be able to use that test template to have a test that:
- Enables dark mode
- Presents the datetime in a modal
- Captures a screenshot

core/src/components/datetime/test/basic/index.html
```diff
@@ -394,7 +394,7 @@ <h2>Modal - Custom</h2>
           // create component to open
           const element = document.createElement('div');
           element.innerHTML = `
-            <ion-datetime show-default-buttons="true" color="danger">
+            <ion-datetime id="datetime-modal" show-default-buttons="true" color="danger">
```
sean-perkins, line 397: Add `value="2023-01-01"` to the `ion-datetime` and that should help create stable screenshots for your test (you will need to re-generate the reference screenshots with the Github Action after committing the change).

Currently it is always loading the "latest date/time" when the modal is presented, which will produce a failed test for any time in the future.

thetaPC: Thank you!!! Once again, screenshots making it difficult.

amandaejohnston:
So, the issue here occurs with more than just `ion-datetime`. If you put a random `ion-text` in the modal, for example, that also exhibits the incorrect text color when switching to dark mode. While I hadn't been able to isolate the exact cause when triaging the issue, I theorized about it here: https://github.com/ionic-team/ionic-framework/issues/26060#issuecomment-1287307163 Maybe a similar fix to what you have here should be applied to `ion-modal` instead? (I do think the test you have here covers our bases either way.)

thetaPC:
> So, the issue here occurs with more than just `ion-datetime`. If you put a random `ion-text` in the modal, for example, that also exhibits the incorrect text color when switching to dark mode. While I hadn't been able to isolate the exact cause when triaging the issue, I theorized about it here: [#26060 (comment)](https://github.com/ionic-team/ionic-framework/issues/26060#issuecomment-1287307163) Maybe a similar fix to what you have here should be applied to `ion-modal` instead? (I do think the new test covers our bases either way.)

Understood, I'll give modal a peek. Thank you!

core/src/components/datetime/test/basic/datetime.e2e.ts
```diff
@@ -55,7 +55,7 @@ test.describe('datetime: selecting a day', () => {
   });
 
   test('should not highlight a day until one is selected, with default-buttons', async ({ page }) => {
-    await testHighlight(page, 'custom-datetime');
+    await testHighlight(page, 'inline-datetime-no-value');
```
amandaejohnston, line 58:
```suggestion
    await testHighlight(page, 'custom-datetime');
```
I'm assuming this was edited by mistake; this is meant to check a datetime with `show-default-buttons` set.

core/src/components/datetime/test/basic/datetime.e2e.ts
```diff
@@ -107,7 +107,7 @@ test.describe('datetime: confirm date', () => {
   test('should not update value if Done was clicked without selecting a day first', async ({ page }) => {
     await page.goto('/src/components/datetime/test/basic');
 
-    const datetime = page.locator('#custom-datetime');
+    const datetime = page.locator('#inline-datetime-no-value');
```
amandaejohnston, line 110:
```suggestion
    const datetime = page.locator('#custom-datetime');
```

amandaejohnston (approved):
LGTM once the comments are addressed 👍

core/src/components/datetime/test/basic/datetime.e2e.ts
```diff
@@ -408,3 +408,30 @@ test('datetime: md highlight should not clip at start or end of month', async ({
 
   await expect(datetime).toHaveScreenshot(`date-highlight-end-of-month-${page.getSnapshotSettings()}.png`);
 });
+
+test.describe('datetime: dark mode', () => {
```
liamdebeasi, line 412: This is a fix for `ion-modal`, so this test should be in the `ion-modal` tests instead. Additionally, we should be able to pass text into the modal instead of a datetime.

liamdebeasi (changes requested):
The fix itself works well. Great job! A few comments on the tests.

core/src/components/datetime-button/test/basic/datetime-button.e2e.ts
```diff
@@ -183,3 +183,29 @@ test.describe('datetime-button: wheel', () => {
     await expect(page.locator('#time-button')).not.toBeVisible();
   });
 });
+
+test.describe('datetime-button: dark mode', () => {
```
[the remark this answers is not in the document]
thetaPC: Originally the fix was setting text color directly to [datetime](https://github.com/ionic-team/ionic-framework/pull/27207/commits/ab687f7176341767c1397e63cbaac1f1375849dc#diff-83857955657186aba3b158e5561a577b94c5b9b8e74aff61fc97c78f4b77343cR344). I noticed that the time picker rendered through datetime-button was still displaying the wrong text color so I [updated](https://github.com/ionic-team/ionic-framework/pull/27207/commits/ab687f7176341767c1397e63cbaac1f1375849dc#diff-83857955657186aba3b158e5561a577b94c5b9b8e74aff61fc97c78f4b77343cR406) the code again. I figured it was best to create screenshots for both of them. But with the latest change, it won't be needed due to your [suggestion](https://github.com/ionic-team/ionic-framework/pull/27207#discussion_r1173892310).

core/src/components/datetime/test/basic/datetime.e2e.ts
thetaPC: It wasn't by mistake. I ended up passing a value to `custom-datetime` in order to solve the issue with the changing time that we saw via Zoom. Since these tests ([test 1 ](https://github.com/ionic-team/ionic-framework/blob/main/core/src/components/datetime/test/basic/datetime.e2e.ts#L57)and [test 2](https://github.com/ionic-team/ionic-framework/blob/main/core/src/components/datetime/test/basic/datetime.e2e.ts#L53)) required a datetime that had no value then I ended up switching. However, I'll be reverting this file and moving the test to modal as [suggested](https://github.com/ionic-team/ionic-framework/pull/27207#discussion_r1173892310).

core/src/components/modal/test/dark-mode/model.e2e.ts
```diff
@@ -0,0 +1,29 @@
...
+
+    const button = page.locator('#basic-modal');
+    await button.click();
+
+    await ionModalDidPresent.next();
+
+    const modal = await page.locator('ion-modal');
+    await expect(modal).toHaveClass(/show-modal/);
+
+    await page.setIonViewport();
```
liamdebeasi, line 25:
```suggestion
```

This is not needed since there is no content that scrolls offscreen.

core/src/components/modal/test/dark-mode/model.e2e.ts
```diff
@@ -0,0 +1,29 @@
...
+
+    const ionModalDidPresent = await page.spyOnEvent('ionModalDidPresent');
+
+    const button = page.locator('#basic-modal');
+    await button.click();
+
+    await ionModalDidPresent.next();
+
+    const modal = await page.locator('ion-modal');
+    await expect(modal).toHaveClass(/show-modal/);
```
liamdebeasi, line 23:
```suggestion
```

`ionModalDidPresent` fires when the modal is visible and has completed any transition, so this check should not be needed.

core/src/components/modal/test/dark-mode/index.html
liamdebeasi, line 1: Can we use an inline modal instead to reduce the test boilerplate?

```html
<ion-modal trigger="basic-modal">
  <h1>Content of doom</h1>
  <div>Here's some more content</div>
  <ion-button class="dismiss">Dismiss Modal</ion-button>
</ion-modal>
```

This would allow you to remove all of the `<script>` content. You can place this code as a sibling of the `#basic-modal` element.

core/src/components/modal/test/dark-mode/index.html
```diff
@@ -0,0 +1,260 @@
+<!DOCTYPE html>
+<html lang="en" dir="ltr">
+  <head>
+    <meta charset="UTF-8" />
+    <title>Modal - Basic</title>
```
liamdebeasi, line 5:
```suggestion
    <title>Modal - Dark Mode</title>
```

core/src/components/modal/test/dark-mode/index.html
```diff
@@ -0,0 +1,260 @@
...
+      }
+    </style>
+  </head>
+
+  <body class="dark">
+    <ion-app>
+      <div class="ion-page">
+        <ion-header>
+          <ion-toolbar>
+            <ion-title>Modal - Basic</ion-title>
```
liamdebeasi, line 204:
```suggestion
            <ion-title>Modal - Dark Mode</ion-title>
```

liamdebeasi (changes requested):

liamdebeasi (approved):
Great work!

core/src/components/modal/test/dark-mode/index.html
```diff
@@ -0,0 +1,233 @@
...
+          </ion-toolbar>
+        </ion-header>
+
+        <ion-content class="ion-padding">
+          <div class="grid">
+            <div class="grid-item">
+              <h2>Outside "ion-content"</h2>
+              <ion-button id="basic-modal">Present modal</ion-button>
+              <ion-modal trigger="basic-modal">
+                <h1>Content of doom</h1>
```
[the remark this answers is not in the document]
liamdebeasi: We've had that verbiage in another test for a while, so it was likely copy and pasted: https://github.com/ionic-team/ionic-framework/blob/0ac451998c5567172515b51be1bc1b4ff13af611/core/src/components/modal/test/basic/index.html#L70

PR #27209, opened 2023-04-17
chore: add angular 16 test infrastructure 

sean-perkins:

Issue URL: Internal

## What is the new behavior?

- Adds Angular 16 test app
- Introduces version test to validate which major version of Angular is being tested

- [ ] Yes
- [x] No


angular/test/apps/ng16/e2e/src/angular-version.spec.ts
```diff
@@ -0,0 +1,5 @@
+it("should be on Angular 16", () => {
```
sean-perkins, line 1: Need to add the Angular version test for both ng14 and ng15 apps.

.github/workflows/actions/test-angular-e2e/action.yml
```diff
@@ -29,7 +29,7 @@ runs:
       shell: bash
       working-directory: ./angular/test
     - name: Install Dependencies
-      run: npm install
+      run: npm install --legacy-peer-deps
```
sean-perkins, line 32: Need to create a ticket to track swapping this once stable versions of v16 dependencies land. Alternatively we could continue to allow legacy peer dependencies for the test apps. The only advantage I see to not allowing it currently, is if some package we are using is incompatible with the version of Angular we are testing with. Our tests _should_ catch this scenario however. Unfortunately it seems any time we try to add the next major version test app before official release, we have to change this.

angular/test/apps/ng16/src/app/app.component.html
```diff
@@ -0,0 +1,3 @@
+<ion-app>
```
sean-perkins, line 1: With ng14 being the minimal supported version and no longer having the `[environmentInjector]` optional syntax, I can do away with the duplicate `AppComponent` for each individual test app and move this to the base app.

angular/test/apps/ng16/tsconfig.json
```diff
@@ -0,0 +1,35 @@
...
+  "compileOnSave": false,
+  "compilerOptions": {
+    "outDir": "./dist/out-tsc",
+    "forceConsistentCasingInFileNames": true,
+    "strict": true,
+    "noImplicitOverride": true,
+    "noPropertyAccessFromIndexSignature": true,
+    "noImplicitReturns": true,
+    "noFallthroughCasesInSwitch": true,
+    "sourceMap": false,
```
sean-perkins, line 11: I had to disable source maps for Cypress to work (would throw an exception when opening cypress). This clearly isn't an issue in CI for the other test apps. I should re-enable and see if it is just a local Cypress issue (perhaps and outdated dmg).

.github/workflows/actions/test-angular-e2e/action.yml
sean-perkins: Created FW-4092 to track.

amandaejohnston (approved):

PR #27222, opened 2023-04-17
feat(segment, segment-button): update segment value property to accept numbers

### LINKED ISSUE #27221: feat: Update Segment component "value" property to accept numbers
### Prerequisites

This pull request updates the value property of the segment component to accept numbers in addition to strings. Previously, the property was defined as a string, but this update extends it to be a union of string and number. This change will enable users to input numerical values into the segment component and increase its versatility.

### Describe the Use Case

The updated value property of the segment component will provide users with more flexibility in their data input. Prior to this change, the segment component only accepted string values, which could be limiting in certain use cases. For instance, if a user wanted to display numerical data in a segmented format, they would have to manually convert the numbers to strings. However, with the updated value property, users can now input numerical values directly into the component. This change will improve the usability and efficiency of the segment component for users who work with numerical data.

The preferred solution for this feature request is to update the value property of the segment component to accept a union of string and number types. This change should be made in a backwards-compatible manner, meaning that existing code that uses the segment component with a string value will continue to function as before. The documentation for the segment component should also be updated to reflect this change, and any examples should demonstrate the use of numerical values in addition to strings. By implementing this solution, users of the Ionic Framework will have increased flexibility in their data input and improved usability of the segment component.

### Describe Alternatives

One alternative solution to this feature request would be to create a separate component specifically for numerical data input. However, this would lead to unnecessary complexity and duplication of code. Another alternative would be to continue to only accept string values in the segment component, but provide a separate utility function to convert numerical values to strings. While this approach would work, it would add an extra step for users and could potentially lead to errors. Overall, the preferred solution of updating the value property of the segment component to accept a union of string and number types is the simplest and most straightforward solution.

### Related Code

Here is an example of the updated value property of the segment component with a numerical value in React:
```typescript
import { useState } from 'react';
import { IonSegment, IonSegmentButton } from '@ionic/react';

function Example() {
  const [selectedValue, setSelectedValue] = useState<string | number>('1');

  const handleValueChange = (event) => {
    setSelectedValue(event.detail.value);
  };

  return (
    <IonSegment value={selectedValue} onIonChange={handleValueChange}>
      <IonSegmentButton value="1">Option 1</IonSegmentButton>
      <IonSegmentButton value="2">Option 2</IonSegmentButton>
      <IonSegmentButton value="3">Option 3</IonSegmentButton>
      <IonSegmentButton value={4}>Option 4</IonSegmentButton>
      <IonSegmentButton value={5}>Option 5</IonSegmentButton>
    </IonSegment>
  );
}
```
In this example, the value property of each segment button can now accept numerical values in addition to strings. This will allow for more versatile data input in the segment component, particularly for users who work with numerical data. The useState hook is used to manage the selected value state, and the onIonChange event handler is used to update the state when the value changes.

### Additional Information

_No response_

Morritz:
Issue number: resolves #27221

---------

The value property of the segment component in Ionic Framework currently only accepts string values.

## What is the new behavior?

This pull request updates the "value" property of the segment component to accept a union of string and number types. This allows for more versatile data input in the segment component, particularly for users who work with numerical data.

- [ ] Yes
- [X] No

## Other information

N/A

### DIFF
--- a/core/src/components/segment/segment-interface.ts
+++ b/core/src/components/segment/segment-interface.ts
@@ -1,7 +1,9 @@
 export type SegmentButtonLayout = 'icon-top' | 'icon-start' | 'icon-end' | 'icon-bottom' | 'icon-hide' | 'label-hide';
 
+export type SegmentValue = string | number;
+
 export interface SegmentChangeEventDetail {
-  value?: string;
+  value?: SegmentValue;
 }
 
 export interface SegmentCustomEvent extends CustomEvent {
--- a/angular/src/index.ts
+++ b/angular/src/index.ts
@@ -117,6 +117,7 @@ export {
   SearchbarInputEventDetail,
   SegmentChangeEventDetail,
   SegmentCustomEvent,
+  SegmentValue,
   SelectChangeEventDetail,
   SelectCustomEvent,
   TabsCustomEvent,
--- a/core/api.txt
+++ b/core/api.txt
@@ -1184,7 +1184,7 @@ ion-segment,prop,mode,"ios" | "md",undefined,false,false
 ion-segment,prop,scrollable,boolean,false,false,false
 ion-segment,prop,selectOnFocus,boolean,false,false,false
 ion-segment,prop,swipeGesture,boolean,true,false,false
-ion-segment,prop,value,string | undefined,undefined,false,false
+ion-segment,prop,value,number | string | undefined,undefined,false,false
 ion-segment,event,ionChange,SegmentChangeEventDetail,true
 ion-segment,css-prop,--background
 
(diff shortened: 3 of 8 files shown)


JakubCzura:
So interesting feature. It will be good if they merge it.

liamdebeasi:
Thanks for the request and PR! I think this would be a good feature to add to Ionic. The next step for us is to create a design document of how the feature will work/be implemented. This design process allows us to a) discuss the feature so there is consensus and b) ensure that the API is well designed.

I have created the design document internally, and the team will be looking at it soon. If the design document gets approved, then someone will review your PR and potentially request changes. Let me know if you have questions.

core/src/components/segment/segment-interface.ts
```diff
@@ -1,7 +1,9 @@
 export type SegmentButtonLayout = 'icon-top' | 'icon-start' | 'icon-end' | 'icon-bottom' | 'icon-hide' | 'label-hide';
 
+export type SegmentValue = string | number;
```
liamdebeasi, line 3: The `SegmentValue` interface needs to be exported from `packages/vue/src/index.ts`, `angular/src/index.ts`, and `packages/react/src/components/index.ts` too.

liamdebeasi (changes requested):
The team has completed our design document, so we are looking to move forward with this PR. I left a requested change below.

One other thing is since this is a feature, we'll need to merge into the `feature-7.1` branch instead of `main`. Let me know if you have any questions!

sean-perkins:
Rebased the branch against `feature-7.1`, this is ready for new reviews.

liamdebeasi (approved):
Thank you!

amandaejohnston (approved):

PR #27226, opened 2023-04-18
fix(input): string values are emitted

liamdebeasi:
Issue number: #

---------

 

 

## What is the new behavior?

-
-
-

- [ ] Yes
- [ ] No

### DIFF
--- a/core/src/components/input/input.tsx
+++ b/core/src/components/input/input.tsx
@@ -424,7 +424,11 @@ export class Input implements ComponentInterface {
    */
   private emitInputChange(event?: Event) {
     const { value } = this;
-    this.ionInput.emit({ value, event });
+
+    // Checks for both null and undefined values
+    const newValue = value == null ? value : value.toString();
+
+    this.ionInput.emit({ value: newValue, event });
   }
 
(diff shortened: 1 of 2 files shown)


core/src/components/input/input.tsx
```diff
@@ -424,7 +424,11 @@ export class Input implements ComponentInterface {
    */
   private emitInputChange(event?: Event) {
     const { value } = this;
-    this.ionInput.emit({ value, event });
+
+    // Checks for both null and undefined values
+    const newValue = value == null ? value : value.toString();
```
amandaejohnston, line 429: The type interface only allows `string | null`, no `undefined`, so should we return `null` instead of `value` on the left half? Or should the type be updated?

liamdebeasi: Not sure I understand. The interface for `InputInputEventDetail` has `value?: string | null` where the `?` indicates the value can be `undefined`.

amandaejohnston: Oh, you're right, I missed that 🤪

amandaejohnston (approved):

PR #27227, opened 2023-04-18
fix(menu): export menu interfaces

sean-perkins:
Issue number: N/A

---------

 

 

When updating to Stencil v3 ([commit](https://github.com/ionic-team/ionic-framework/commit/1a8bd6d8c6a5227f803250722442199cbfa3510d#diff-83006355bd18aced9f88fd3b27ff6a40527c713b314145c2a34ad54b6361fa87)), we removed many manual exports that are automatically exported by Stencil. This change included removing the `MenuI` type from being exported. This type is not automatically exported since it is not directly associated to a `@Prop`.

This causes implementers that previously had a typed value against `MenuI` in v6 for registering menus, to have to import from: 
```ts
import { MenuI } from "@ionic/core/dist/types/components/menu/menu-interface";
```

## What is the new behavior?

- Re-exports the `MenuI` interface

- [ ] Yes
- [x] No


liamdebeasi:
Is `MenuControllerI` supposed to be exported too?

sean-perkins:
It wouldn't hurt, it is only referenced from `MenuI`. I'll make that change 👍

PR #27230, opened 2023-04-18
refactor(): update directories to use kebab case

liamdebeasi:
Issue number: N/A

---------

 

 

The team decided on some new processes for adding tests. One of those processes is that test directories should use kebab case.

## What is the new behavior?

- Updates test directories to use kebab case

- [ ] Yes
- [x] No


thetaPC (approved):
LGTM

PR #27246, opened 2023-04-19
test(action-sheet): migrate to generator

liamdebeasi:
Issue number: N/A

---------

This PR migrates the action sheet tests to use the generator infrastructure.


core/src/components/action-sheet/test/a11y/action-sheet.e2e.ts
```diff
@@ -0,0 +1,62 @@
...
+       * action-sheet overlays the entire screen, so
+       * Axe will be unable to verify color contrast
+       * on elements under the backdrop.
+       */
+      const results = await new AxeBuilder({ page }).disableRules('color-contrast').analyze();
+      expect(results.violations).toEqual([]);
+    });
+
+    test('should have aria-labelledby when header is set', async ({ page }) => {
+      await page.goto(`/src/components/action-sheet/test/a11y`, config);
```
amandaejohnston, line 45: The `goto` is already in this block's `beforeEach`, so it can be removed from this and the two tests below.

amandaejohnston (approved):
LGTM after the comment is addressed 👍

core/src/components/action-sheet/test/a11y/action-sheet.e2e.ts
liamdebeasi: Fixed in https://github.com/ionic-team/ionic-framework/pull/27246/commits/6574ae1f97dadced85f58b126af513f941d0e6c8

PR #27247, opened 2023-04-19
test(alert): migrate to generator

liamdebeasi:
Issue number: N/A

---------

This PR migrates the alert tests to use the generator infrastructure.

In https://github.com/ionic-team/ionic-framework/pull/27247/commits/91ba3a1290e4ff42cd106d27209d6ea20afb8fd3 I migrated the basic test to use [Playwright Fixtures](https://playwright.dev/docs/test-fixtures#creating-a-fixture) instead since it is a) a best practice and b) easier to add the generator support to.


core/src/components/alert/test/a11y/alert.e2e.ts
```diff
@@ -0,0 +1,65 @@
...
+  const ariaDescribedBy = await alert.getAttribute('aria-describedby');
+
+  expect(ariaLabelledBy).toBe(expectedAriaLabelledBy);
+  expect(ariaDescribedBy).toBe(expectedAriaDescribedBy);
+};
+
+test.describe('alert: a11y', () => {
+  configs({ directions: ['ltr'] }).forEach(({ config, title }) => {
+    test(title('should not have accessibility violations when header and message are defined'), async ({ page }) => {
+      await page.goto(`/src/components/alert/test/a11y`, config);
```
amandaejohnston, line 34: Can these `goto`s be moved to a `beforeEach`? Or is that not a pattern we're encouraging? Ditto for the other test files, or at least the ones that have more than one test.

core/src/components/alert/test/basic/alert.e2e.ts
```diff
@@ -0,0 +1,176 @@
...
+
+      const alertFixture = new AlertFixture(page, screenshot);
+
+      await alertFixture.open('#checkbox');
+      await alertFixture.screenshot('checkbox');
+    });
+  });
+});
+
+class AlertFixture {
```
amandaejohnston, line 137: Many of the other tests share this functionality. Would it make sense to pull the fixture into a separate file and import it?

core/src/components/alert/test/a11y/alert.e2e.ts
liamdebeasi: No, we can't use `beforeEach` here to house the `goto` calls. Placing `beforeEach` inside of the `configs` callback will cause multiple `beforeEach` blocks to be generated.

liamdebeasi:
Going to rethink this a bit and see if there's a way we can continue using `beforeEach`

core/src/components/alert/test/a11y/alert.e2e.ts
liamdebeasi: Ok so I we moved the `configs` outside of the `test.describe` block we get a few benefits:

1. We can keep using `test.beforeEach` to do `page.goto`
2. We can eliminate the `title()` function usage on everything except for the `test.describe` block

amandaejohnston (approved):
We should consider simplifying this and other overlay tests in separate PR(s) to pull open/dismiss/screenshot behavior into a fixture all tests can share 👀

PR #27248, opened 2023-04-19
test(docs): add documentation on custom Ionic functions

liamdebeasi:
Issue number: N/A

---------

This PR adds documentation for the following:

1. A step-by-step guide for migrating to generators.
2. Documentation on the custom tooling we built on top of Playwright
3. A best practice guide. Note: This was originally in the Ionic internal Notion, but I moved it here because it is of value to community contributors.

I also added a few new best practices that were not on the Notion guide:

- Components must have a `basic` test with an `index.html` file
- Slow tests should be broken up across multiple files for CI performance
- Rendering and functionality tests should be broken up across multiple test.describe blocks or files


core/src/utils/test/playwright/docs/api.md
```diff
@@ -0,0 +1,458 @@
...
+
+configs().forEach(({ config, title }) => {
+  test.describe(title('my test block'), () => {
+    test('my custom test', ({ page }) => {
+      await page.goto('path/to/file', config);
+    });
+  });
+});
+```
+
```
thetaPC, line 38: For consistency, add the following: `Deprecated: Use a [generator](#generators) instead.` This would let each section of DEPRECATED to have this.

Or I recommend to just add a note that any deprecation should use generators instead.

thetaPC (changes requested):

core/src/utils/test/playwright/docs/api.md
liamdebeasi: Each section has a note below it to use a generator:

> Deprecated: Use a [generator](#generators) instead.


However, I'll take this as a sign that the note is hard to find and will make this more discoverable.

liamdebeasi: I moved the deprecation note to be the first line: https://github.com/ionic-team/ionic-framework/pull/27248/commits/3d587bf5e672d61f673d5a5c202e0939c2afd6b5

thetaPC (approved):
LGTM

core/src/utils/test/playwright/docs/api.md
```diff
@@ -0,0 +1,459 @@
...
+</details>
+
+## `page` fixture
+
+The [page fixture](https://playwright.dev/docs/test-fixtures) has been extended to provide additional methods:
+
+| Method | Description |
+| - | - |
+| `goto` | The [page.goto](https://playwright.dev/docs/api/class-page#page-goto) method extended to support a config from a [generator](#generator) and to automatically wait for Stencil components to initialize. |
+| `setContent` | The [page.setContent](https://playwright.dev/docs/api/class-page#page-set-content) method extended to support a config from a [generator](#generator) and to automatically wait for Stencil components to initialize. |
```
amandaejohnston, line 106:
```suggestion
| `setContent` | The [page.setContent](https://playwright.dev/docs/api/class-page#page-set-content) method extended to support a config from a [generator](#generators) and to automatically wait for Stencil components to initialize. |
```

amandaejohnston (approved):

PR #27255, opened 2023-04-20
fix(toggle): swipe gesture applies to knob

### LINKED ISSUE #27254: bug: `ion-toggle` and rendered label blocks vertical gestures/touch-scrolling
### Prerequisites

### Ionic Framework Version

v7.x

### Current Behavior

With the new 'modern form syntax' `ion-toggle` now renders or contains its label as a child element. This means that its gesture config is now taking over the entire width of the parent container, and blocks touch/scroll gestures on not just the toggle icon/button, but on the entire label/row. This can lead to some very confusing user interactions.

See video:

https://user-images.githubusercontent.com/362329/233424827-c6d2ed5a-8492-4736-b211-47a5ca038eea.mov

Note, I have not tried it, but I suspect that other elements such as `ion-range`, which also have gestures, may also suffer from this same issue.

### Expected Behavior

The toggle's label should not be a gesture capturing element.

### Steps to Reproduce

Checkout repro url and scroll as seen in video.

### Code Reproduction URL

https://github.com/lincolnthree/ionic-bugs/tree/issue-27254

### Ionic Info

Ionic:

   Ionic CLI                     : 6.20.3 (/Users/lincoln/.nvm/versions/node/v16.14.0/lib/node_modules/@ionic/cli)
   @angular-devkit/build-angular : 15.1.6
   @angular-devkit/schematics    : 15.1.6
   @angular/cli                  : 15.1.6

Capacitor:

   Capacitor CLI      : 4.6.3
   @capacitor/core    : 4.6.3

Utility:

   native-run (update available: 1.7.2) : 1.7.1

System:

   NodeJS : v16.14.0 (/Users/lincoln/.nvm/versions/node/v16.14.0/bin/node)
   npm    : 8.3.1
   OS     : macOS

### Additional Information

_No response_

liamdebeasi:
Issue number: resolves #27254

---------

 

 

The swipe gesture is currently applied to the entire `ion-toggle` element. This was fine for the legacy syntax, but with the modern syntax it means users can swipe on the label text which is not correct.

## What is the new behavior?

- The toggle now creates the gesture on the `.toggle-icon` element which is the container for the knob for both modern and legacy syntaxes.
- Moved `touch-action: none` to the host of the legacy toggle. This was preventing scrolling from happening on the modern toggle. I double checked with native iOS and you can scroll when a pointer moves over a toggle.

The structure of this fix was designed to match what `ion-range` does: https://github.com/ionic-team/ionic-framework/blob/a8749929e01b07043631fbc8c522d39cbc3ae798/core/src/components/range/range.tsx#L282-L296

| Modern | Legacy |
| - | - |
| <video src="https://user-images.githubusercontent.com/2721089/233431240-11f0c94f-d86b-4975-afd5-e534262a6f16.mov"></video> | <video src="https://user-images.githubusercontent.com/2721089/233431275-6c6f7fef-6cc0-4adc-8915-6fd5c3795ade.mov"></video> |

- [ ] Yes
- [x] No


core/src/components/toggle/toggle.scss
```diff
@@ -240,8 +240,6 @@ input {
 
   background: var(--track-background);
 
-  pointer-events: none;
```
liamdebeasi, line 243: This is removed from the `.toggle-icon` element so users can swipe on the element.

sean-perkins (approved):

liamdebeasi:
I added another part of the fix so re-requesting reviews:

> Moved touch-action: none to the host of the legacy toggle. This was preventing scrolling from happening on the modern toggle. I double checked with native iOS and you can scroll when a pointer moves over a toggle.

See https://github.com/ionic-team/ionic-framework/pull/27255/commits/656526f98fce157189c918751bb8cf1129efc3b2

lincolnthree:
Dev build works great in my testing.

sean-perkins (approved):
Nice work 👍

PR #27262, opened 2023-04-21
test(radio): enable keyboard tests

sean-perkins:
Issue number: Internal

---------

 

 

`Tab` keyboard navigation tests are disabled for `ion-radio` because they were extremely flaky in CI.

## What is the new behavior?

- Re-enables `ion-radio` `Tab` keyboard navigation tests
- Ports a slimmed down version of Wordpress' playwright utilities for using modifier keys with `Tab` keyboard navigation on environments that do not support "natural tab navigation"

- [ ] Yes
- [x] No

## Other information

This utility should likely be used in all places where we are using the `Alt+Tab` logic for webkit-based browsers today. If/when the changes are agreed upon, I can follow-up with PR that adds this behavior to those tests.

### DIFF
--- a/core/src/utils/test/press-keys.ts
+++ b/core/src/utils/test/press-keys.ts
@@ -1,5 +1,17 @@
 import type { Browser, BrowserContext, Page } from '@playwright/test';
 
+/**
+ * The purpose of this utility is to provide a way to press keys in a way that
+ * is consistent across browsers and platforms. Playwright does not automatically
+ * normalize key presses, so we need to do it ourselves.
+ *
+ * In certain environments, such as Webkit on macOS, the browser will not focus
+ * the correct element in the DOM when the tab key is pressed.
+ * This utility will detect if the browser has natural tab navigation and
+ * will use the appropriate key combination to simulate a tab press.
+ * The utility will normalize key presses for other combinations as well.
+ */
+
 const SHIFT = 'shift';
 const CTRL = 'ctrl';
 const ALT = 'alt';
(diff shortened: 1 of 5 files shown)


core/src/utils/test/press-keys.ts
liamdebeasi, line 1: Can you document the purpose of this file and why it works as a comment?

liamdebeasi (changes requested):

core/src/utils/test/press-keys.ts
```diff
@@ -0,0 +1,140 @@
...
+ * await pressKeys('Shift+Tab');
+ * ```
+ */
+export async function pressKeys(this: PageUtils, key: string, { times, ...pressOptions }: Options = {}) {
+  const hasNaturalTabNavigation = await getHasNaturalTabNavigation(this.page);
+  const keys = key.split('+').flatMap((keyCode) => {
+    if (Object.prototype.hasOwnProperty.call(modifiers, keyCode)) {
+      return modifiers[keyCode as keyof typeof modifiers](isAppleOS).map((modifier) =>
+        modifier === CTRL ? 'Control' : capitalCase(modifier)
+      );
```
liamdebeasi, line 112: Can we document what this is doing?

core/src/utils/test/press-keys.ts
```diff
@@ -0,0 +1,140 @@
...
+  });
+  const normalizedKeys = keys.join('+');
+  const command = () => this.page.keyboard.press(normalizedKeys);
+
+  times = times ?? 1;
+  for (let i = 0; i < times; i += 1) {
+    await command();
+
+    if (times > 1 && pressOptions.delay !== undefined) {
+      await this.page.waitForTimeout(pressOptions.delay);
```
liamdebeasi, line 130: Can we document what this is doing?

liamdebeasi (changes requested):
Overall looks good, just would like to document some parts of the util code

core/src/utils/test/press-keys.ts
sean-perkins: Let me know if the updated documentation makes sense 👍

liamdebeasi (approved):
Great find! This should help a great deal

PR #27263, opened 2023-04-22
chore(ci): enable npm package provenance

ltm:
See https://github.blog/2023-04-19-introducing-npm-package-provenance/

### DIFF
--- a/.github/workflows/actions/publish-npm/action.yml
+++ b/.github/workflows/actions/publish-npm/action.yml
@@ -22,6 +22,9 @@ runs:
     - uses: actions/setup-node@v3
       with:
         node-version: 16.x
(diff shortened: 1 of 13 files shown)


.github/workflows/actions/publish-npm/action.yml
```diff
@@ -22,6 +22,9 @@ runs:
     - uses: actions/setup-node@v3
       with:
         node-version: 16.x
+    - name: Install latest npm
+      run: npm install -g npm@latest
```
ltm, line 26: Node 16 includes npm 8.19.4, but package provenance requires npm 9.5. Alternatively we could install Node 18.

liamdebeasi (approved):
Thank you!

PR #27267, opened 2023-04-24
test(many): do not await page.locator

liamdebeasi:
Issue number: N/A

---------

 

 

`page.locator` is synchronous, but we were `await`ing the calls: https://playwright.dev/docs/api/class-page#page-locator

We were also doing `page.locator(...).click()` when we can just do `page.click([selector])` directly,

## What is the new behavior?

- Removes `await` usage from `page.locator`
- Removes `page.locator().click()` usage in favor of `page.click()`

- [ ] Yes
- [x] No

### DIFF
--- a/core/src/components/action-sheet/test/basic/action-sheet.e2e.ts
+++ b/core/src/components/action-sheet/test/basic/action-sheet.e2e.ts
@@ -85,7 +85,7 @@ test.describe('action sheet: basic', () => {
       const ionAlertDidPresent = await page.spyOnEvent('ionAlertDidPresent');
       await actionSheetFixture.open('#alertFromActionSheet');
 
-      await page.locator('#open-alert').click();
+      await page.click('#open-alert');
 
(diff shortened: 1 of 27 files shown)


brandyscarney:
Is there a reason these were skipped:

- https://github.com/ionic-team/ionic-framework/blob/test-locator-sync/core/src/components/datetime/test/presentation/datetime.e2e.ts#L18 (and others)
- https://github.com/ionic-team/ionic-framework/blob/test-locator-sync/core/src/components/menu/test/basic/menu.e2e.ts#L59 (and others)
- https://github.com/ionic-team/ionic-framework/blob/test-locator-sync/core/src/components/picker-column-internal/test/basic/picker-column-internal.e2e.ts#L25 (and others)

liamdebeasi:
`evaluate` and `selectOption` are asynchronous, so they need the `await`.

brandyscarney (approved):

PR #27269, opened 2023-04-24
docs(tabs): document methods only available for vanilla

mapsandapps:
When using Angular, React, or Vue, the use of tabs is coupled to the framework's routing. These methods are only available when using tabs with vanilla JavaScript.

Issue number: N/A

---------

 

 

## What is the new behavior?

- Adds more context to the description of the methods on tabs

- [ ] Yes
- [x] No

### DIFF
--- a/core/src/components.d.ts
+++ b/core/src/components.d.ts
@@ -2888,16 +2888,16 @@ export namespace Components {
     interface IonTabs {
         "getRouteId": () => Promise<RouteID | undefined>;
         /**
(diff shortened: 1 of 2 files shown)


brandyscarney:
Do we want to specifically mention Stencil here? I know we have been trying to discourage developers from using Stencil for mobile apps.

mapsandapps:
@brandyscarney Sure, that makes sense. I used that verbiage since that's what we're saying [here](https://ionicframework.com/docs/api/tab). Should I make a PR to the docs repo to update that too?

sean-perkins:
Yeah I think we should simplify the language on the docs. Cut out Stencil, the documentation pertains to any JS environment that isn't one of our official framework wrapped packages.

sean-perkins (approved):

Issue #27273, opened 2023-04-25
bug: focused background on breadcrumb is set incorrectly for md


tetkosimi:
### Prerequisites

- [X] I have read the [Contributing Guidelines](https://github.com/ionic-team/ionic-framework/blob/main/.github/CONTRIBUTING.md#creating-an-issue).
- [X] I agree to follow the [Code of Conduct](https://ionicframework.com/code-of-conduct).
- [X] I have searched for [existing issues](https://github.com/ionic-team/ionic-framework/issues) that already report this problem, without success.

### Ionic Framework Version

v6.x, v7.x

### Current Behavior

When breadcrumb is focused in md mode, background is not set.

![2023-04-25 16_53_57-ion-breadcrumbs _ Ionic Documentation](https://user-images.githubusercontent.com/37016280/234317388-1dd44a72-f7cb-4e46-a9f9-e9c8059b7f72.png)

### Expected Behavior

A slight background should have applied on breadcrumb when focused.

![2023-04-25 16_54_31-ion-breadcrumbs _ Ionic Documentation](https://user-images.githubusercontent.com/37016280/234317947-52033877-a609-40c3-b2f0-381688432f59.png)

### Steps to Reproduce

1.  Put focus on any breadcrumb in md mode.

### Ionic Info

@ionic/angular 7.0.3

PR #27274, opened 2023-04-25
fix(breadcumb): set background when focused on md

tetkosimi:
Issue number: #27273

---------

 

 

When breadcrumb is focused in md mode, background is not set.

## What is the new behavior?
A slight background should have applied on breadcrumb when focused.

-
-
-

- [ ] Yes
- [ x] No


liamdebeasi (approved):

liamdebeasi:
Verified locally that the fix works. Going to merge this when ready. Thank you for the PR!

PR #27293, opened 2023-04-26
Button generator

liamdebeasi:
Issue number: #

---------

 

 

## What is the new behavior?

-
-
-

- [ ] Yes
- [ ] No


core/src/components/button/test/form-reference/button.e2e.ts
```diff
@@ -0,0 +1,107 @@
+import { expect } from '@playwright/test';
+import { configs, test } from '@utils/test/playwright';
+
+configs({ directions: ['ltr'], modes: ['ios'] }).forEach(({ title, config }) => {
```
liamdebeasi, line 4: This used to test all configs, but it's not needed so I have it only check LTR + iOS mode since it's just checking behavior

amandaejohnston (approved):

PR #27298, opened 2023-04-27
chore(ci): fix stencil nightly job dependencies

rwaskiewicz:
Issue number: N/A

---------

 

 

The Stencil nightly job fails with the following error:

```
The workflow is not valid. .github/workflows/stencil-nightly.yml (Line: 101, Col: 13): Job 'test-core-screenshot-legacy' depends on unknown job 'build-core'. 

.github/workflows/stencil-nightly.yml (Line: 122, Col: 12): Job 'verify-screenshots-legacy' depends on job 'test-core-screenshot-legacy' which creates a cycle in the dependency graph.
```

This started to fail last night. I believe this is a result of https://github.com/ionic-team/ionic-framework/pull/27228, specifically this line https://github.com/ionic-team/ionic-framework/commit/4fe8de7df71b0d9fb8d618aca5a428129d96a04b#diff-2f087e8fac034d51c50fc9350e10ceb4034298c16dd0d4f414f79d88ebb71aa0R101

## What is the new behavior?

this commit updates the job dependency hierarchy for test-core-screenshot-legacy from `build-core` to
`build-core-with-stencil-nightly`. the former is not a valid job name for the `stencil-nightly` workflow, likely just a copy/paste error

- [ ] Yes
- [x] No

## Other information

To test this, I was able to kick off a version of this job with this branch: https://github.com/ionic-team/ionic-framework/actions/runs/4819613134 and saw the new/legacy screenshot tests running:
Screenshot 2023-04-27 at 8 29 14 AM

### DIFF
--- a/.github/workflows/stencil-nightly.yml
+++ b/.github/workflows/stencil-nightly.yml
@@ -98,7 +98,7 @@ jobs:
         # to be the length of the shard array.
         shard: [1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15, 16, 17, 18, 19, 20]
         totalShards: [20]
-    needs: [build-core]
+    needs: [build-core-with-stencil-nightly]
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v3


.github/workflows/stencil-nightly.yml
```diff
@@ -98,7 +98,7 @@ jobs:
         # to be the length of the shard array.
         shard: [1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15, 16, 17, 18, 19, 20]
         totalShards: [20]
-    needs: [build-core]
+    needs: [build-core-with-stencil-nightly]
```
rwaskiewicz, line 101: Alternatively, we could rename the job itself to `build-core`. The original intent was that when looking at the 'main' CI workflow for Framework, I found myself getting confused (was I looking at `main.yml`, or `stencil-nightly.yml`?). This served as an easier way to differentiate b/w the two. LMK if you'd prefer to make the `s/build-core-with-stencil-nightly/build-core/` change

liamdebeasi (approved):

PR #27304, opened 2023-04-27
test(datetime): migrate to generators

liamdebeasi:
Issue number: N/A

---------

 

 

Datetime tests use the legacy syntax

## What is the new behavior?

- Datetime tests use the generator syntax

**Note to reviewers:**

I recommend reviewing this commit by commit because datetime has a lot of tests. A few commits I'd like to call out:

https://github.com/ionic-team/ionic-framework/pull/27304/commits/77d9ab0ff24d23aafad053ed08a7ba8be3669ac7

- I removed several screenshots that were not being tested anymore. It seems the actual test was modified/removed in the past, but the screenshot ground truths remained.

https://github.com/ionic-team/ionic-framework/pull/27304/commits/ed45c7fa057024bdbf79a41b2e3ab6a822e2685f

- This test checks that the datetime can properly expand in width based on the `cover` property. As a result, no RTL tests are needed because the behavior does not vary across directions. I narrowed the config of this test to only check LTR and removed extraneous screenshots.

https://github.com/ionic-team/ionic-framework/pull/27304/commits/96c08410f21d1edca02c7975e9c3e9287b845815

- This test only checks that the correct text content is used within the component. As a result, I narrowed the config to only check LTR direction and removed extraneous screenshots.

- [ ] Yes
- [x] No


thetaPC (approved):
LGTM

liamdebeasi:
Going to request another set of eyes since this is a bigger change

amandaejohnston (approved):

PR #27306, opened 2023-04-27
fix(many): form components do not take up full width in slot

### LINKED ISSUE #27305: bug: modern select takes up full width in toolbar slot
### Prerequisites

### Ionic Framework Version

v7.x

### Current Behavior

The modern select takes up 100% of the toolbar width when used in a slot. This is happening because our `width: auto` code is too narrow and only works with `ion-item`: https://github.com/ionic-team/ionic-framework/blob/87b2aedc94b65977076ee9a2c4c62b7134955c85/core/src/components/select/select.scss#L98-L99

We should probably remove `.in-item`.

### Expected Behavior

I expect select to not take up 100% of the toolbar width when in a slot.

### Steps to Reproduce

1. Open https://codepen.io/liamdebeasi/pen/KKGqXMo
2. Observe that the modern select takes up 100% of the width.

### Code Reproduction URL

_No response_

### Ionic Info

N/A

### Additional Information

_No response_

liamdebeasi:
Issue number: resolves #27305

---------

 

 

Checkbox, radio, range, select, and toggle take up 100% of its width even when in a start/end slot.

## What is the new behavior?

- Checkbox, radio, range, select, and toggle do not take up 100% of its width even when in a start/end slot.
. The "no slot" behavior is unchanged and aligns with the legacy select.

- [ ] Yes
- [x] No

### DIFF
--- a/core/src/components/select/test/label/index.html
+++ b/core/src/components/select/test/label/index.html
@@ -33,10 +33,6 @@
           padding: 0;
         }
       }
-
-      ion-select {
-        width: 100%;
-      }
     </style>
   </head>
 
@@ -49,6 +45,20 @@
       </ion-header>
 
       <ion-content id="content" class="ion-padding">
+
+        <ion-select
+          label="Favorite Fruit"
+          placeholder="Select a fruit"
+          label-placement="start"
+          justify="start"
+        ></ion-select>
+
+        <ion-label>abc123</ion-label>
+        <ion-select
+          placeholder="Select a fruit"
+          legacy="true"
+        ></ion-select>
+
         <h1>Placement Start</h1>
         <div class="grid">
           <div class="grid-item">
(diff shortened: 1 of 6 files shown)


core/src/components/select/test/label/index.html
```diff
@@ -33,10 +33,6 @@
           padding: 0;
         }
       }
-
-      ion-select {
-        width: 100%;
-      }
```
liamdebeasi, line 39: This was redundant, so I removed it.

amandaejohnston (approved):
It might be worth making this change to the other form controls, which also have similar `width: auto` pieces when slotted in an item.

liamdebeasi:
Good call, I updated this in https://github.com/ionic-team/ionic-framework/pull/27306/commits/316f91e0d9b9294e92248c45259e08b107dd5651

amandaejohnston (approved):
The comments all still say "select" despite the different components, but LGTM once that's adjusted.

brandyscarney (approved):

PR #27307, opened 2023-04-27
test(datetime-button): migrate to generators

liamdebeasi:
Issue number: N/A

---------

 

 

Datetime button tests use the legacy syntax

## What is the new behavior?

- Datetime button tests use the generator syntax

- [ ] Yes
- [x] No


core/src/components/datetime-button/test/overlays/datetime-button.e2e.ts
```diff
@@ -0,0 +1,180 @@
+import { expect } from '@playwright/test';
+import type { Locator } from '@playwright/test';
+import { configs, test } from '@utils/test/playwright';
+import type { EventSpy } from '@utils/test/playwright';
+
+/**
+ * This behavior does not vary across directions.
+ */
+configs({ directions: ['ltr'] }).forEach(({ title, screenshot, config }) => {
```
liamdebeasi, line 9: Tests were previously running for all configs, but they did not need to be. As a result, I limited the tests to only run in LTR direction

thetaPC (approved):
LGTM

PR #27321, opened 2023-04-28
test(input): migrate to generators

liamdebeasi:
Issue number: N/A

---------

 

 

Input tests use legacy syntax

## What is the new behavior?

- Input tests use modern syntax

- [ ] Yes
- [x] No


liamdebeasi:
Going to wait for at least 2 reviews before merging since this is another big change.

thetaPC (approved):
LGTM

core/src/components/input/test/item/input.e2e.ts
```diff
@@ -0,0 +1,71 @@
...
+            ></ion-input>
+          </ion-item>
+        </ion-list>
+      `,
+        config
+      );
+      const list = page.locator('ion-list');
+      expect(await list.screenshot()).toMatchSnapshot(screenshot(`input-inset-list-no-fill`));
+    });
+    test('input renders as modern when label is set asynchronously', async ({ page }, testInfo) => {
```
amandaejohnston, line 46: Optional: Looks like this one could be pulled into a separate block and only run on LTR/one mode.

core/src/components/input/test/label-placement/input.e2e.ts
```diff
@@ -0,0 +1,192 @@
+import { expect } from '@playwright/test';
+import { configs, test } from '@utils/test/playwright';
+
+configs().forEach(({ title, screenshot, config }) => {
```
amandaejohnston, line 4: Optional: Could the identical `configs()` loops be merged into one here?

amandaejohnston (approved):

core/src/components/input/test/item/input.e2e.ts
liamdebeasi: I ended up making this a spec test. Good catch!

PR #27325, opened 2023-04-28
test(label): migrate to generators

liamdebeasi:
Issue number: N/A

---------

 

 

Label sliding tests use legacy syntax

## What is the new behavior?

- Label sliding tests use modern syntax

- [ ] Yes
- [x] No


thetaPC (approved):
LGTM

PR #27334, opened 2023-04-28
chore(): add missing screenshot function

liamdebeasi:
Issue number: N/A

---------

 

 

item sliding tests were missing a `screenshot` function

## What is the new behavior?

- Added the function
- Also removed stray `config` params

This was not caught because these tests are currently skipped.

- [ ] Yes
- [x] No


thetaPC (approved):
LGTM

PR #27339, opened 2023-04-30
feat(input): component api for input masking

sean-perkins:
Issue number: Internal

---------

 

 

N/A

## What is the new behavior?

- Adds `mask`, `maskVisibility` and `maskPlaceholder` properties to `ion-input` for input masking

- [ ] Yes
- [x] No


sean-perkins:
This is one of many pull requests for adding input masking to `ion-input`. As functionality becomes available for user testing, I will call that out in the pull request. 

I will do my best to break down the work effort behind this initiative. There are going to be some complex chunks of work along the journey 🙂 

All individual PRs should merge back into `FW-2573` before the feature is eventually merged into a 7.1 feature branch.

core/src/components/input/test/input.spec.ts
```diff
@@ -7,8 +8,32 @@ it('should inherit attributes', async () => {
...
   });
 
-  const nativeEl = page.body.querySelector('ion-input input');
+  const nativeEl = page.body.querySelector('ion-input input')!;
   expect(nativeEl.getAttribute('title')).toBe('my title');
   expect(nativeEl.getAttribute('tabindex')).toBe('-1');
   expect(nativeEl.getAttribute('data-form-type')).toBe('password');
 });
+
+describe('input masking', () => {
```
amandaejohnston, line 17: Are these tests necessary? We're just checking if Stencil passed the correct default value here. I can't imagine the defaults changing by accident/without being caught in review.

sean-perkins: I don't know if any tests are necessary, but it helps avoid the potential for human error in both development and review and allows projects to iterate quicker with extra validation that existing behavior is maintained.

core/src/components/input/input.tsx
```diff
@@ -272,6 +273,29 @@ export class Input implements ComponentInterface {
    */
   @Prop({ mutable: true }) value?: string | number | null = '';
 
+  /**
+   * Mask format. Can be either a string or array of characters and regular expressions.
+   *
+   * More complex masks can be defined as an array of regular expressions and
+   * constant characters. Each index of the mask array, defines the mask for
```
liamdebeasi, line 280:
```suggestion
   * constant characters. Each index of the mask array defines the mask for
```

core/src/components/input/input.tsx
```diff
@@ -272,6 +273,29 @@ export class Input implements ComponentInterface {
...
+  /**
+   * The visibility of the mask placeholder. With `always`, the placeholder will be
+   * visible even when the control does not have focus. With `focus`, the placeholder
+   * will only be visible when the control has focus. With `never`, the placeholder will
+   * never be visibly displayed.
+   */
+  @Prop() maskVisibility?: MaskVisibility = 'always';
+
+  /**
+   * Character or string to cover unfilled parts of the mask. Default character is `_`.
```
liamdebeasi, line 294:
```suggestion
   * Character or string to cover unfilled parts of the mask. The default character is `_`.
```

core/src/components/input/input.tsx
```diff
@@ -272,6 +273,29 @@ export class Input implements ComponentInterface {
    */
   @Prop({ mutable: true }) value?: string | number | null = '';
 
+  /**
+   * Mask format. Can be either a string or array of characters and regular expressions.
```
liamdebeasi, line 277: What's the difference between the "string" and "character" terminology?

core/src/components/input/input.tsx
```diff
@@ -272,6 +273,29 @@ export class Input implements ComponentInterface {
...
+   *
+   * More complex masks can be defined as an array of regular expressions and
+   * constant characters. Each index of the mask array, defines the mask for
+   * the corresponding index in the input’s string.
+   */
+  @Prop() mask?: MaskFormat;
```
liamdebeasi, line 283: We should also document how certain characters define the allowed inputs

core/src/components/input/input.tsx
sean-perkins: I believe this is answered based on our Slack thread + design doc adjustment proposal, but let me know if that isn't the case. 

Once I have approval on the design doc adjustment, I'll update the type and description for this.

core/src/components/input/input.tsx
```diff
@@ -272,6 +273,36 @@ export class Input implements ComponentInterface {
...
+  /**
+   * The predefined format of the user's input. For example, you can set a mask
+   * that only accepts digits, or you can configure a more complex pattern like
+   * a phone number or credit card number.
+   *
+   * The mask supports two formats:
+   * 1. A valid [regular expression pattern](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Regular_Expressions)
+   * 2. An array containing regular expression and fixed character patterns
+   *
+   * Each string in the mask array format is a fixed character in the mask.
```
liamdebeasi, line 285: When I first read this I understood it to mean each entry in the array is a fixed character. This would contradict the above line which states the array can accept a regex. However, I now understand that it means string types.

Perhaps we can just say "The fixed characters in the mask cannot be erased or replaced by the user"

core/src/components/input/test/input.spec.ts
liamdebeasi: I don't mind keeping them, but I will note that any change to the default value of props will cause `api.txt` to be updated so we should be able to catch accidental changes to the default value during review.

liamdebeasi (approved):

sean-perkins:
Sweet - in the case of the `api.txt` I'll remove the test for now and can re-evaluate as the feature is developed.

PR #27346, opened 2023-05-01
feat(input): mask controller set-up

sean-perkins:
Issue number: N/A

---------

 

 

N/A

## What is the new behavior?

Introduces a `MaskController` class to handle the implementation details of input masking. The implementation is stubbed out to guide future PRs that will implement the underlying details.

The design is implemented based on [maskito](https://github.com/Tinkoff/maskito). Ionic Framework supports newer versions of Firefox than this library targets, so we can diverge on the implementation and access modern APIs in portions of the implementation. 

- [ ] Yes
- [x] No

## Other information

In terms of testing, you could run the mask test template and verify that different `console.debug` messages are logging when interacting with the control.


sean-perkins:
There is additional diff in this PR, until https://github.com/ionic-team/ionic-framework/pull/27348 is merged and the parent branch is rebased.

core/src/utils/input-masking/utils/format.spec.ts
```diff
@@ -0,0 +1,18 @@
+import { formatMask } from './format';
+
+describe('formatMask', () => {
+  it('should convert a string to a RegExp', () => {
+    const mask = formatMask('0');
+    expect(mask).toBeInstanceOf(RegExp);
```
amandaejohnston, line 6: Could we make this check for the exact regex to be safe?

core/src/utils/input-masking/utils/get-not-empty-selection.spec.ts
```diff
@@ -0,0 +1,73 @@
...
+    it('should not change everything when start value is 0', () => {
+      const elementStateStub: ElementState = {
+        value: 'testValue',
+        selection: [0, 4],
+      };
+
+      expect(getNotEmptySelection(elementStateStub, false)).toEqual([0, 4]);
+    });
+
+    it('should decrease by one start value', () => {
```
amandaejohnston, line 35: This and the `'should decrease by one start position value'` test above seem to be testing the same thing.

core/src/utils/input-masking/mask-controller.ts
```diff
@@ -0,0 +1,158 @@
...
+      console.debug('keydown', event);
+    });
+
+    if (isBeforeInputEventSupported(element)) {
+      this.eventListener.listen('beforeinput', (event) => {
+        console.debug('beforeinput', event);
+      });
+    } else {
+      /**
+       * TODO: Remove once Firefox 87+ is the minimum supported version
```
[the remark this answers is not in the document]
sean-perkins: I'll make sure tech debt tickets are created when `FW-2573` is opened up against `feature-7.1` 👍 I'll validate all `TODO` statements are addressed and/or a ticket is attached.

core/src/utils/input-masking/utils/format.spec.ts
sean-perkins: Now that our API signature is consistent (`RegExp` instead of `string`), `formatMask` isn't needed so I deleted it and the respective tests 🎉

PR #27352, opened 2023-05-02
test(picker, picker-column): migrate to generators

liamdebeasi:
Issue number: N/A

---------

 

 

Picker and picker column tests use legacy syntax

## What is the new behavior?

- Picker and picker column tests use generator syntax

- [ ] Yes
- [x] No


thetaPC (approved):
LGTM

PR #27359, opened 2023-05-03
fix(picker-column): dynamically change options

### LINKED ISSUE #21763: bug: Error occurs when dynamically changing picker column options
# Bug Report

**Ionic version:**

[ ] **4.x**
[x] **5.x**

**Current behavior:**

An error occurs when dynamically changing picker column options and the picker UI is not updated as required.

**Expected behavior:**

picker UI must be updated accordingly and no errors must occur.

**Steps to reproduce:**

1) open the link to stackblitz (see below)
2) click/tap OPEN PICKER button
3) change the value in the first column
Note that there is an error shown in console "Error: Cannot read property 'duration' of undefined" and values in the second picker column are not updated.

**Related code:**
See https://stackblitz.com/edit/ionic-angular-v5-y1p2sa?file=src%2Fapp%2Fapp.component.ts

**Other information:**
The problem is that the picker component does not correctly handle the use case when the number of options in a column changed. It loops through the options that were before the update and failed because the number of options has been changed/decreased - see below

Screenshot 2020-07-16 at 11 33 15

Screenshot 2020-07-16 at 11 33 45

**Ionic info:**
```
Ionic:

   Ionic CLI                     : 6.9.2 (/Users/alexryltsov/.nvm/versions/node/v12.16.3/lib/node_modules/@ionic/cli)
   Ionic Framework               : @ionic/angular 5.1.1
   @angular-devkit/build-angular : 0.901.7
   @angular-devkit/schematics    : 9.1.7
   @angular/cli                  : 9.1.7

Capacitor:

   Capacitor CLI   : 2.1.1
   @capacitor/core : 2.1.1

Utility:

   cordova-res : not installed
   native-run  : not installed

System:

   NodeJS : v12.16.3 (/Users/alexryltsov/.nvm/versions/node/v12.16.3/bin/node)
   npm    : 6.14.5
   OS     : macOS Catalina
```

thetaPC:
Issue number: resolves #21763

---------

 

 

Picker options do not render correctly when dynamically changed. It throws an error when the original options' length is no longer the same as the updated options' length. This is due to `refresh()` being called with `this.optsEl` having a stale list of children. The list doesn't get updated until the upcoming render.

## What is the new behavior?

The column will call `refresh()` when it detects the columns has changed. The call needs to be done after `this.optsEl` gets updated with the correct number of children.

`componentShouldUpdate()` will check if options has changed -> re-renders with the updated options -> `componentDidUpdate()` calls `refresh()` based on `componentShouldUpdate()`

The standalone test page has been updated to include a way to test this fix.

- [ ] Yes
- [x] No

## Other information

It might be beneficial to review the [component lifecycle](https://stenciljs.com/docs/component-lifecycle).

Co-authored-by: liamdebeasi <liamdebeasi@users.noreply.github.com>


core/src/components/picker-column/picker-column.tsx
```diff
@@ -33,6 +33,7 @@ export class PickerColumnCmp implements ComponentInterface {
   private rafId?: ReturnType<typeof requestAnimationFrame>;
   private tmrId?: ReturnType<typeof setTimeout>;
   private noAnimate = true;
+  private shouldAnimteOnRender = true;
```
amandaejohnston, line 36:
```suggestion
  private shouldAnimateOnRender = true;
```

core/src/components/picker-column/test/standalone/index.html
```diff
@@ -31,7 +34,7 @@
         ['Tomato', 'Avocado', 'Onion', 'Potato', 'Artichoke'],
       ];
 
-      async function openPicker(numColumns = 1, numOptions = 5, columnOptions = defaultColumnOptions) {
+      async function openPicker(numColumns = 1, numOptions = 5, columnOptions = defaultColumnOptions, dynamic = false) {
```
amandaejohnston, line 37: It might be tough to find this new test in the future since `standalone` is just for checking that basic functionality works without `ion-app`. Could we either move this to picker's `basic` test or create a new one entirely, and then add an E2E test to verify the behavior?

core/src/components/picker-column/picker-column.tsx
```diff
@@ -87,10 +88,35 @@ export class PickerColumnCmp implements ComponentInterface {
...
+
+  componentDidUpdate(): void | Promise<void> {
+    if (!this.shouldAnimteOnRender) {
+      const colEl = this.optsEl;
+      if (colEl) {
+        // DOM READ
+        // We perfom a DOM read over a rendered item, this needs to happen after the the column options have changed
+        this.optHeight = colEl.firstElementChild ? colEl.firstElementChild.clientHeight : 0;
+      }
+      this.refresh(true);
```
amandaejohnston, line 116: This is duplicated across here and `componentDidLoad` now; can we abstract it out?

amandaejohnston (changes requested):

core/src/components/picker-column/picker-column.tsx
```diff
@@ -87,10 +88,35 @@ export class PickerColumnCmp implements ComponentInterface {
...
+    // The change is most likely triggered by an outside source
+    // For example, when adding or removing an option dynamically, etc.
+    // If the dynamic option was added then it needs to be added to the DOM
+    // without animation
+    // Otherwise the animation will be render the option
+    // at the top of the column and then it will fall down to its place
+    if (propName === 'col' && JSON.stringify(newVal.options) !== JSON.stringify(oldVal.options)) {
+      this.shouldAnimteOnRender = false;
+    }
+  }
```
liamdebeasi, line 106: Can we set `this.shouldAnimateOnRender` in the `@Watch` callback? That callback runs before the component re-renders and will only fire when the `col` data changes.

core/src/components/picker-column/picker-column.tsx
```diff
@@ -159,7 +185,7 @@ export class PickerColumnCmp implements ComponentInterface {
       }
 
       // Update transition duration
-      if (this.noAnimate) {
+      if (this.noAnimate || !this.shouldAnimteOnRender) {
```
liamdebeasi, line 188: Why is this needed?

core/src/components/picker-column/test/standalone/index.html
liamdebeasi: +1 to this. My preference would be to create a separate E2E test so we can avoid screenshot diffs.

liamdebeasi (changes requested):

core/src/components/picker-column/picker-column.tsx
thetaPC: The `col` data changes can also be triggered when a user toggles through the picker. Is there a way to compare `oldVal` and `newVal` in `@Watch`? I didn't see an example on the Stencil docs.

core/src/components/picker-column/picker-column.tsx
thetaPC: If a user ends up adding an option and has the animation on, then the option will render on top of the first option then slide down to the bottom. With animation turned off, it will automatically render at the bottom.

core/src/components/picker-column/picker-column.tsx
thetaPC: I had my function order wrong. You're right that we don't need to compare. It'll be changed in the next commit.

core/src/components/picker-column/test/columns/index.html
amandaejohnston, line 1: Nitpick: Could we call this something like `dynamic-content` or `dynamic-items` so it's more obvious what it's testing?

core/src/components/picker-column/test/columns/index.html
```diff
@@ -0,0 +1,171 @@
...
+      const multiColumnOptions = [
+        ['Minified', 'Responsive', 'Full Stack', 'Mobile First', 'Serverless'],
+        ['Tomato', 'Avocado', 'Onion', 'Potato', 'Artichoke'],
+      ];
+
+      async function removeOption(numColumns = 1, numOptions = 5, columnOptions = defaultColumnOptions) {
+        const picker = await pickerController.create({
+          columns: this.getColumns(numColumns, numOptions, columnOptions),
+          buttons: this.getButtons(),
+          animated: false,
```
amandaejohnston, line 67: Why the `animated: false`? I would rather leave as many options at the default as possible, especially since we're changing the `noAnimate` behavior so it would be good to be able to check that the animations still work at a glance.

core/src/components/picker-column/picker-column.tsx
```diff
@@ -74,21 +75,27 @@ export class PickerColumnCmp implements ComponentInterface {
       onEnd: (ev) => this.onEnd(ev),
     });
     this.gesture.enable();
+    // Options have not been initialized yet
+    // Animation must be disabled through the `noAnimate` flag
+    // Otherwise, the options will will render
+    // at the top of the column and transition down
     this.tmrId = setTimeout(() => {
-      this.noAnimate = false;
```
amandaejohnston, line 78: Why was this removed?

core/src/components/picker-column/picker-column.tsx
```diff
@@ -74,21 +75,27 @@ export class PickerColumnCmp implements ComponentInterface {
...
-      // We perfom a DOM read over a rendered item, this needs to happen after the first render
-      this.optHeight = colEl.firstElementChild ? colEl.firstElementChild.clientHeight : 0;
-    }
+    this.onDomChange();
+  }
 
-    this.refresh();
+  componentDidUpdate(): void | Promise<void> {
+    // Options have changed since last update.
+    if (this.noAnimate) {
```
amandaejohnston, line 96: I wonder if it would be better to use a new, separate variable like `colsChanged` to track this, and set both when the cols change? I worry that if we have something else set `noAnimate` in the future, it could lead to unwanted extra `refresh` calls. The variable name doesn't indicate that it's tied to that behavior.

amandaejohnston (changes requested):

core/src/components/picker-column/picker-column.tsx
thetaPC: I ended up moving it to the `update()` but I'll be reverting the `noAnimate` back and using another flag as requested.

core/src/components/picker-column/test/dynamic-options/index.html
```diff
@@ -0,0 +1,170 @@
...
+      const multiColumnOptions = [
+        ['Minified', 'Responsive', 'Full Stack', 'Mobile First', 'Serverless'],
+        ['Tomato', 'Avocado', 'Onion', 'Potato', 'Artichoke'],
+      ];
+
+      async function removeOption(numColumns = 1, numOptions = 5, columnOptions = defaultColumnOptions) {
+        const picker = await pickerController.create({
+          columns: this.getColumns(numColumns, numOptions, columnOptions),
+          buttons: this.getButtons(),
+          animated: false,
```
amandaejohnston, line 67:
```suggestion
```

amandaejohnston (approved):
LGTM aside from one last tweak 👍 Thank you!

core/src/components/picker-column/picker-column.tsx
```diff
@@ -159,7 +178,7 @@ export class PickerColumnCmp implements ComponentInterface {
       }
 
       // Update transition duration
-      if (this.noAnimate) {
+      if (this.noAnimate || this.didColChange) {
```
liamdebeasi, line 181: Why is this flag needed here?

core/src/components/picker-column/picker-column.tsx
```diff
@@ -185,6 +204,8 @@ export class PickerColumnCmp implements ComponentInterface {
         button.classList.remove(PICKER_OPT_SELECTED);
       }
     }
+
+    this.didColChange = false;
```
liamdebeasi, line 208: Given that `didColChange` is only needed for `componentDidUpdate`, would it make more sense to set this to `false` in `componentDidUpdate`?

core/src/components/picker-column/picker-column.tsx
liamdebeasi, line 181: `this.update` accepts a `duration` parameter that lets us customize the duration of the animation. I think this code is going to be hard to maintain now that there are 3 different ways to skip the animation. Rather than check `didColChange` to skip the animation could we make use of the `duration` parameter instead?

You could update `refresh` to accept an `animated` parameter. From there, you'd update the `refresh` body to have the following code:

```js
this.update(y, animated === false ? 0 : TRANSITION_DURATION, true);
```

And then you'd update the `this.refresh` call in `onDomChange` to pass in `this.refresh(forceRefresh, false)` where `false` is the `animated` paramter.

core/src/components/picker-column/test/dynamic-options/index.html
thetaPC: My bad! I thought I took it out last time.

core/src/components/picker-column/picker-column.tsx
```diff
@@ -132,6 +141,16 @@ export class PickerColumnCmp implements ComponentInterface {
...
     const children = this.optsEl.children;
+
+    // Options might have been changed
+    // ex: adding or removing
+    // `this.optsEl` children will not match the new `this.col` order
+    // until after `componentDidUpdate()`
+    // Skip this update cycle
+    if (children.length !== col.options.length) {
+      return;
+    }
```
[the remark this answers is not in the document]
thetaPC: This happen when using multiple columns:
1. Recommended: Add a `console.log()` inside the if statement
2. Navigate to the [test page](http://localhost:3333/src/components/picker-column/test/dynamic-options)
3. Click the "Remove option" under the multiple columns
4. Select the last option ("Artichoke") on the second/right column
5. Select the second option ("Responsive") on the first/left column

Without this if statement then the app will throw this error: `TypeError: opt is undefined`

core/src/components/picker-column/test/dynamic-options/picker-column.e2e.ts
```diff
@@ -0,0 +1,99 @@
...
+        await page.waitForChanges();
+
+        const pickerOpts = column.locator('.picker-opts');
+        const optionsCount = await pickerOpts.locator('.picker-opt').count();
+
+        expect(optionsCount).toBe(6);
+      });
+    });
+
+    test.describe('multiple columns', () => {
```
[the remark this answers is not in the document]
thetaPC: Verified that they're not needed.

core/src/components/picker-column/picker-column.tsx
```diff
@@ -356,9 +379,20 @@ export class PickerColumnCmp implements ComponentInterface {
     const selectedIndex = clamp(min, this.col.selectedIndex ?? 0, max);
     if (this.col.prevSelected !== selectedIndex || forceRefresh) {
       const y = selectedIndex * this.optHeight * -1;
+      const duration = animated === false ? 0 : TRANSITION_DURATION;
```
liamdebeasi, line 382:
```suggestion
      const duration = animated ? TRANSITION_DURATION : 0;
```

We try to write positive conditions so they are easy to understand in the event we need to negate them. For example "not animated" is easier to understand than "not not animated".

This one is my bad since I gave you this code snippet, and I should have written it in the positive to begin with.

core/src/components/picker-column/picker-column.tsx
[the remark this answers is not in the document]
liamdebeasi: This is happening because we are still calling `this.refresh` from the `colChanged` watch callback. If we only want to run `this.refresh` once the component has re-rendered, then we probably should not be calling `this.refresh` in `colChanged` anymore. Removing that means we can also remove the above ^ check.

core/src/components/picker-column/test/dynamic-options/picker-column.e2e.ts
```diff
@@ -0,0 +1,72 @@
...
+import { configs, test } from '@utils/test/playwright';
+
+configs({ modes: ['md'], directions: ['ltr'] }).forEach(({ title, config }) => {
+  test.describe(title('picker-column'), () => {
+    test.beforeEach(async ({ page }) => {
+      await page.goto('/src/components/picker-column/test/dynamic-options', config);
+    });
+
+    test.describe('add option', () => {
+      test('should have the style attribute', async ({ page }) => {
```
liamdebeasi, line 11: This seems like it would be better as a unit test since we are testing component functionality. Something like this should work:

```js
import { h } from '@stencil/core';

import { PickerColumnCmp } from '../../picker-column';

describe('dynamic options', () => {
  it('should add an option', async () => {
    const defaultOptions = [
      { text: 'Dog', value: 'dog' },
      { text: 'Cat', value: 'cat' }
    ];

      components: [PickerColumnCmp],
      template: () => <ion-picker-column col={{ options: defaultOptions }}></ion-picker-column>,
    });

    const pickerCol = page.body.querySelector('ion-picker-column');

    pickerCol.col = {
      options: [...defaultOptions, { text: 'Carrot', value: 'carrot' } ]
    }

    await page.waitForChanges();

    const pickerOpt = pickerCol.querySelector('.picker-opt:nth(2)');
    expect(pickerOpt.getAttribute('style')).toContain('transform');
  });
});
```

It sounds like we want to verify that the newly added option gets the correct transform value when the component re-renders (which it was not getting before this patch). Using the spec test would let us get rid of the HTML and E2E files in `dynamic-options`.

liamdebeasi (changes requested):

core/src/components/picker-column/test/dynamic-options/picker-column.e2e.ts
```diff
@@ -0,0 +1,72 @@
...
+        const column = page.locator('ion-picker-column');
+
+        const middleOption = column.locator('.picker-opt').nth(2);
+        await middleOption.click();
+
+        const lastOption = column.locator('.picker-opt').last();
+        await lastOption.click();
+
+        await page.waitForChanges();
+        expect(errors.length).toBe(0);
```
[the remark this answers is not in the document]
thetaPC: I'll remove this since the edge case doesn't occur after removing `this.refresh` from the `colChanged`.

core/src/components/picker-column/picker-column.tsx
```diff
@@ -74,21 +80,32 @@ export class PickerColumnCmp implements ComponentInterface {
...
-      // DOM READ
-      // We perfom a DOM read over a rendered item, this needs to happen after the first render
-      this.optHeight = colEl.firstElementChild ? colEl.firstElementChild.clientHeight : 0;
-    }
+    this.onDomChange();
+  }
 
-    this.refresh();
+  componentDidUpdate() {
+    // Options may have changed since last update.
```
liamdebeasi, line 101:
```suggestion
    // Options have changed since last update.
```

If `colDidChange` is `true` then the columns definitely changed since the last render

core/src/components/picker-column/picker-column.tsx
```diff
@@ -74,21 +80,32 @@ export class PickerColumnCmp implements ComponentInterface {
...
+    if (this.colDidChange) {
+      // Animation must be disabled through the `onDomChange` parameter.
+      // Otherwise, the recently added options will will render
```
liamdebeasi, line 104:
```suggestion
      // Otherwise, the recently added options will render
```

core/src/components/picker-column/picker-column.tsx
```diff
@@ -74,21 +80,32 @@ export class PickerColumnCmp implements ComponentInterface {
       onEnd: (ev) => this.onEnd(ev),
     });
     this.gesture.enable();
+    // Options have not been initialized yet
+    // Animation must be disabled through the `noAnimate` flag
+    // Otherwise, the options will will render
```
liamdebeasi, line 85:
```suggestion
    // Otherwise, the options will render
```

core/src/components/picker-column/picker-column.tsx
```diff
@@ -33,6 +33,12 @@ export class PickerColumnCmp implements ComponentInterface {
   private rafId?: ReturnType<typeof requestAnimationFrame>;
   private tmrId?: ReturnType<typeof setTimeout>;
   private noAnimate = true;
+  // `colDidChange` is a flag that gets set when the column is changed
+  // dynamically. When this flag is set, the column will refresh
+  // to incorporate the new column data.
```
liamdebeasi, line 38:
```suggestion
  // after the component re-renders to incorporate the new column data.
  // This is necessary because `this.refresh` queries for the option elements,
  // so it needs to wait for the latest elements to be available in the DOM.
```

core/src/components/picker-column/picker-column.tsx
```diff
@@ -132,6 +149,7 @@ export class PickerColumnCmp implements ComponentInterface {
     const scaleStr = `scale(${this.scaleFactor})`;
 
     const children = this.optsEl.children;
+
```
liamdebeasi, line 152:
```suggestion
```

core/src/components/picker-column/picker-column.tsx
```diff
@@ -185,6 +203,7 @@ export class PickerColumnCmp implements ComponentInterface {
         button.classList.remove(PICKER_OPT_SELECTED);
       }
     }
+
```
liamdebeasi, line 206:
```suggestion
```

liamdebeasi (approved):
A few small comment/style things, but otherwise this is good to go. Great job on this!

PR #27373, opened 2023-05-03
test(searchbar): migrate to generators

liamdebeasi:
Issue number: N/A

---------

Searchbar tests are using legacy syntax

## What is the new behavior?

- Searchbar tests are using generator syntax

https://github.com/ionic-team/ionic-framework/pull/27373/commits/343458dc57905357b7da948c81b98aad106a27e9

The searchbar event tests do not vary across modes, so I removed those additional checks.

- [ ] Yes
- [x] No


thetaPC (approved):
LGTM

PR #27374, opened 2023-05-03
test(segment): migrate to generators

liamdebeasi:
Issue number: N/A

---------

Segment tests are using legacy syntax

## What is the new behavior?

- Segment tests are using generator syntax

https://github.com/ionic-team/ionic-framework/pull/27374/commits/3f05ad247fbb206aa8491bb4a61422b19cd9996c

The segment events tests do not vary across modes, so I removed the extra checks.

- [ ] Yes
- [x] No


thetaPC (approved):
LGTM

PR #27375, opened 2023-05-03
fix(range): round value to same number of decimal places as props to avoid floating point rounding errors

### LINKED ISSUE #21968: ion-range's ionChange event emitter emits fractional numbers unexpectedly
**Ionic version:**
[ ] **4.x**
[x] **5.3.1**

**Current behavior:**
Given the following code:

```
    <ion-range [min]="option.min" [max]="option.max"
      [step]="option.step" (ionChange)="log($event); option.value = $event.target.value">
    </ion-range>
```

```
export class AppComponent {

  option: = {
          type: "range",
          label: "Distance",
          name: "distance",
          step: 0.05,
          min: 0,
          max: 1,
          value: 0.1
  }

  log(event) {
    console.log(event.target.value)
  }

}
```
We can observe fractional numbers (0.35000000000000003) assigned as input value.

**Expected behavior:**
The value should not be fractional number.

**Steps to reproduce:**
https://stackblitz.com/edit/ionic-angular-v5-ic1yf2?file=src/app/app.component.ts

Or just create ion-range with steps below 1 (e.g. 0.05 as a step)

I guess this relates to Javascript's specificity (quirk)?:
Numbers in JavaScript are "double-precision 64-bit format IEEE 754 values", according to the spec...

So if we would use standard HTML input type range:

```
<input type="range" [min]="option.min" [max]="option.max"
      [step]="option.step" onchange="console.log(event.target.value)"/>
```
Then there is no fractional value.

Not sure what is the best to address this but at the least we could remind of the 0.1 + 0.2 = 0.30000000000000004 issue in Javascript and that developer has to "toFixed" this situation?

amandaejohnston:
Issue number: Resolves #21968

---------

When using fractional values for `min`, `max`, or `step`, it is possible for floating point rounding errors to cause unexpected values to be emitted. For example, `step="0.05" min="0.1" max="1"` emits a value of `0.150000000004` after moving one step.

## What is the new behavior?

Values are rounded to the max number of decimal places between the three props. Note that it isn't mathematically possible to arrive at a value with more decimal places than the props*, since addition (i.e. starting at `min` and adding multiples of `step`) can't increase the precision of a number.

\* Unless the `value` is set manually, but in that case, `ion-range` currently snaps to a multiple of `step` as soon as the slider is moved, resuming normal behavior.

- [ ] Yes
- [x] No


core/src/components/range/test/basic/range.e2e.ts
liamdebeasi, line 1: Do you think there's value in extracting `getDecimalPlaces` to a util function and writing spec tests that test that? My understanding of this bug is it's less about the interaction and more about the underlying logic that caused this bug.

amandaejohnston: I pulled the functionality out in https://github.com/ionic-team/ionic-framework/pull/27375/commits/3537810e01b967f72a264642aabeb478335b599a. Let me know if you think this is too much abstraction; I couldn't think of another use case for this util off the top of my head, so I made the function pretty specific, but I could also break it up into smaller functions and move more of the logic (probably the `Math.max` piece) back into `range.tsx`.

core/src/components/range/range.tsx
```diff
@@ -902,10 +903,14 @@ const renderKnob = (
...
   let value = (max - min) * ratio;
+
   if (step > 0) {
+    // round to nearest multiple of step, then add min
     value = Math.round(value / step) * step + min;
   }
-  return clamp(min, value, max);
+  const clampedValue = clamp(min, value, max);
+
+  return roundToMaxDecimalPlaces(clampedValue, min, max, step);
```
liamdebeasi, line 913: Out of curiosity, why did you opt for this approach instead of multiplying by a scaled `step` (and then dividing by the scale amount)?

For example: 

```typescript
let value = (max - min) * ratio;

const SCALE = 100;
const SCALED_STEP = step * SCALE;

if (step > 0) {
  // round to nearest multiple of step, then add min
  value = Math.round(value / step) * SCALED_STEP + min;
}

return clamp(min, value / SCALE, max);
```

Scaling the value ensures that you deal with integers to avoid JavaScript's floating point precision funkiness. As long as you scale down by the same amount you scaled up, you should get a nice neat decimal as a multiple of `step`.

amandaejohnston: Because it wouldn't work for all numbers. I'd rather not set an arbitrary limit on how high or low the props can be, nor the precision they can have. For example, a `SCALE` of `10000` sets the bounds within 0.0001 < n < (MAX_INT / 10000), and you can't have more than 4 decimal places -- 5.12345 * 10000 = 51234.5.

liamdebeasi: Ah yeah good point. Though I think the current approach may have some accuracy issues too since `toFixed` can sometimes yield funky results.

```html
<ion-range
  min="0.1"
  max="0.54321876549876532"
  value="0.2"
  step="0.001"
>
  <span slot="label">Label</span>
</ion-range>
```

Video sample:

https://user-images.githubusercontent.com/2721089/236343077-c6a0f1b4-cbb2-4ab4-a686-30bdb775fab0.mov

Is there a defined level of precision we support with `ion-range`?

amandaejohnston: What's happening there is you've got a `max` with the same number of decimal places as the result with the rounding errors, and `toFixed` is rounding to as many as the `max`, so the error isn't trimmed. I would say that's okay, since we wouldn't be able to tell the difference between a rounding error result and something legitimate in a case like that. (While in your case, the only legitimate value with that precision is the actual `max`, if the step were 0.00000000000000001 instead, the two would definitely be indistinguishable.)

I think it's okay that this means we implicitly only support up to a certain level of precision, because the problem is with floating point math itself, not Ionic. If people are using precision levels _that high_, they're going to run into rounding error issues like this all _over_ the place. 🤭 (Or in other words, JS itself only supports up to a certain level of precision anyway, unless you start using third party data types meant for this sort of thing. And I'd say supporting those is out of scope for this fix 😆)

liamdebeasi: Sounds good to me! Could we add a little bit of explanation about these limitations in the code so other devs are aware?

PR #27380, opened 2023-05-03
chore(angular): test app uses v16 dependencies

sean-perkins:
Issue number:  FW-4092

---------

## What is the new behavior?

- Updates the Angular 16 test app to the official dependencies
- Removes the legacy peer dependency install flag (needed during the RC cycle)

- [ ] Yes
- [x] No


sean-perkins:
Yeah good call - updated the PR to remove the legacy peer deps install flag.

liamdebeasi (approved):
thanks!

PR #27381, opened 2023-05-03
test(thumbnail): migrate to generators

liamdebeasi:
Issue number: N/A

---------

Thumbnail tests are using legacy syntax

## What is the new behavior?

- Thumbnail tests are using modern syntax

- [ ] Yes
- [x] No


thetaPC (approved):
LGTM

PR #27397, opened 2023-05-04
Modal generator perf

liamdebeasi:
Issue number: #

---------

## What is the new behavior?

-
-
-

- [ ] Yes
- [ ] No


thetaPC (approved):
LGTM

core/src/components/modal/test/can-dismiss/modal.e2e.ts
```diff
@@ -0,0 +1,29 @@
+import { expect } from '@playwright/test';
+import { configs, test, dragElementBy } from '@utils/test/playwright';
+
+/**
+ * This behavior does not vary across modes/directions
+ */
+configs({ modes: ['ios'], directions: ['ltr'] }).forEach(({ title, config }) => {
```
amandaejohnston, line 7: It looks like [this test](https://github.com/ionic-team/ionic-framework/blob/c67a0f2fb76e6789bc8598f68d4f145dc06dd38c/core/src/components/modal/test/can-dismiss/modal.e2e.ts#L64) was lost in the shuffle. Was it removed on purpose? (Or am I just missing it?)

liamdebeasi: I think I may have removed it because it was redundant. For example, the `should dismiss when canDismiss is Promise<true>` test checks how a `canDismiss` callback with a Promise that eventually returns `true` works.

The action sheet test is functionality the same thing since it waits for the user to click on the action sheet. I'm open to adding it back, but it just seemed like we were really testing action sheet behavior at that point.

amandaejohnston: Makes sense 👍

Issue #8700, opened 2016-10-14
bug: long text with max width not truncating in button


alan-agius4:
#### Short description of the problem:

Long text in buttons is not being truncated. I am using `full` and `block` buttons

![image](https://cloud.githubusercontent.com/assets/17563226/19384889/28eb0bc8-920c-11e6-8a36-25250f8890b8.png)

While the button does have the ellipsis, as show below;

``` sass
.button {
    text-align: center;
    text-overflow: ellipsis;
    white-space: nowrap;
}
```

This doesn't work, as the `.button-inner` class has  `display: flex` 
#### What behaviour are you expecting?

Display ellipsis.

![image](https://cloud.githubusercontent.com/assets/17563226/19384896/326da1f6-920c-11e6-9733-925ed7f3bd69.png)

This can be done by setting the display of  `.button-inner` to `inline`. However, I don't really know the reason for it's display being flex.

**Which Ionic Version?** 2 RC.0

ncapito:
I think it's there to help with centering?   Not sure how you all would want to fix it, but  the only way I have been able to fix something like this is to give it another inner child. e.g.

 ``` 

 <button ion-button full block>
   <span class='overflow-me'>button[ion-button][block] with other text</span>
  </button>

```

I then put a class on `.overflow-me` to do the ellipsis.

```
   .overflow-me{
    overflow: hidden;
    text-overflow: ellipsis;
   }
```

brandyscarney:
This is actually pretty complex since buttons are used in so many components. Adding an inner div to buttons would cause issues elsewhere, and changing the display from flex to inline causes issues with buttons containing icons and anchor links styled to look like buttons. I am going to remove this from the milestone for now since the code above by @ncapito works, but I'll revisit this issue.

mapsandapps:
Thank you for submitting this issue.

Starting in [Ionic Framework 7.2](https://github.com/ionic-team/ionic-framework/releases/tag/v7.2.0), we encouraged developers to [opt in](https://github.com/ionic-team/ionic-framework/pull/27547) to text wrapping in buttons by setting `ion-text-wrap`. This is available now if you are on Ionic 7.2+.

As a result of the changes merged in #28682, text wrapping of buttons will be the default behavior starting in Ionic version 8.

PR #27402, opened 2023-05-04
fix(button): ellipsis text in a button when it is too long to wrap

brandyscarney:
Issue number: #8700

---------

The text in a button is slotted in the default slot as a text node. Text nodes cannot be targeted using `::slotted(*)`. The default behavior when text is really long in a button is to align it in the center and overflow it on both sides (but the overflow is not visible). Due to this we are unable to target the slotted text without adding another wrapper element and targeting that or having the user add their own wrapper class and style that.

## What is the new behavior?

- Adds a wrapper around the default slot and styles the overflow on this wrapper
- Adds flex-shrink to slotted `start` and `end` elements so icons do not disappear from the text growing
- Adds tests for long text in a button

- [ ] Yes
- [x] No

## Other information

When the original issue was created we were using the `ion-button` component in other components (see [toast](https://github.com/ionic-team/ionic-framework/blob/8671efe78b46806ff9070c71fead403cfd2da525/packages/core/src/components/toast/toast.tsx#L152)). As of this PR I can only see it being used by [`ion-datetime`](https://github.com/ionic-team/ionic-framework/blob/154aa1dbf609f2cd4b4c24c787df11236a0a767a/core/src/components/datetime/datetime.tsx#L1396). I think this should have minimal impact on other components.

### DIFF
--- a/core/src/components/button/button.scss
+++ b/core/src/components/button/button.scss
@@ -209,6 +209,15 @@
   z-index: 1;
 }
 
+.button-text {
+  text-overflow: ellipsis;
+  overflow: hidden;
+}
+
+::slotted([slot=start]),
+::slotted([slot=end]) {
+  flex-shrink: 0;
+}
 
 // Button Icons
 // --------------------------------------------------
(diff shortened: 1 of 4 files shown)


brandyscarney:
Wrong ticket number

PR #27406, opened 2023-05-05
fix(button): ellipsis text in a button when it is too long to wrap

### LINKED ISSUE #8700: bug: long text with max width not truncating in button

brandyscarney:
Issue number: resolves #8700

---------

The default behavior when text is really long in a button is to horizontally align it in the center and overflow it on both sides (but the overflow is not visible so the text is cut off at the beginning and end). The text in a button is slotted in the default slot as a text node. Text nodes cannot be targeted to style with CSS using `::slotted(*)`. Due to this we are unable to style the button text in order to add an ellipsis without adding another wrapper element and targeting that or having the user add their own wrapper class and style that.

## What is the new behavior?
Updates the button text so that it will add an ellipsis when it is so long that it overflows by doing the following:
- Adds a wrapper around the default slot and styles the overflow on this wrapper
- Adds `flex-shrink: 0` to the slotted `start` and `end` elements so that icons (or any slotted elements) do not disappear when the text grows
- Adds tests for long text in a button
- Changes the `line-height` on a button from `1` to `normal` due to the bottom of characters being cut off (`y`, `p`, etc):
  
- Adds `display: flex` to the slotted `ion-icon`. This was done because when an `ion-icon` is slotted in the default slot (aka not using `icon-only`, `start` or `end` on the `ion-icon`) it is no longer vertically aligned, but if I set `display: flex` on the button text wrapper it removes the ellipsis from the text

- [ ] Yes
- [x] No

## Other information

When the original issue was created we were using the `ion-button` component in other components (see [toast](https://github.com/ionic-team/ionic-framework/blob/8671efe78b46806ff9070c71fead403cfd2da525/packages/core/src/components/toast/toast.tsx#L152)). As of this PR I can only see it being used by [`ion-datetime`](https://github.com/ionic-team/ionic-framework/blob/154aa1dbf609f2cd4b4c24c787df11236a0a767a/core/src/components/datetime/datetime.tsx#L1396). I think this should have minimal impact on other components.


brandyscarney:
I did some investigation into the screenshot diffs for the overlays (alert, action sheet, loading, etc). The text in some of the buttons gets shifted vertically by a pixel, but not all of them. This is only showing up for Safari, and I can't reproduce it locally. I think this is an issue with our screenshot testing. Below is a video toggling between the two photos in [this diff](https://github.com/ionic-team/ionic-framework/pull/27406/files#diff-8e83d2900cd4229a7d247b98bab79bedaa30615a5c0bef94bd1dd6bcb8cc8c9f). Pay attention to the 2nd button and the buttons below the overlay.

https://github.com/ionic-team/ionic-framework/assets/6577830/e5da7027-7532-4f32-b3e8-014fa09823fb

This seems to be the case for all of the screenshot diffs where the only browser that is reporting a difference is Safari. For example, just like in the overlay tests, only some of the text is shifted in the [button diffs](https://github.com/ionic-team/ionic-framework/pull/27406/files#diff-b14c740dc0ea25032f96c0867d3cdc912e2a55bbce7e569fe2345b26d5b39fe6).

core/src/components/button/button.tsx
```diff
@@ -313,7 +313,9 @@ export class Button implements ComponentInterface, AnchorInterface, ButtonInterf
           <span class="button-inner">
             <slot name="icon-only"></slot>
             <slot name="start"></slot>
-            <slot></slot>
+            <span class="button-text">
```
brandyscarney, line 316: I'm open to suggestions on the naming of this class. Since it can technically hold things other than text, maybe it should be something more vague.

thetaPC: What about `button-content`?

thetaPC:
> 

Were you able to resolve this?

thetaPC (commented):
The new screenshots for wrap do not show ellipsis. I also couldn't view it on the test page. Is there something that I'm missing?

brandyscarney:
Ahh I implemented a fix for icons that are in the normal slot which broke the ellipsis. Will look into.

core/src/components/item/test/buttons/item.e2e.ts-snapshots/item-buttons-dark-diff-ios-ltr-Mobile-Safari-linux.png
thetaPC, line 1: Shouldn't the right icon button be left alone? It doesn't overflow so I would assume that it has enough room to display everything without the need of ellipsis?

core/src/components/popover/test/nested/popover.e2e.ts-snapshots/popover-nested-ios-rtl-Mobile-Firefox-linux.png
thetaPC, line 1: These screenshots shouldn't have the arrow flipped. I assume this is part of the issues you've been dealing with screenshots?

core/src/components/button/test/wrap/index.html
```diff
@@ -37,6 +37,14 @@ <h2>Default</h2>
           <ion-icon slot="end" name="star"></ion-icon>
         </ion-button>
 
+        <h2>Icons only</h2>
```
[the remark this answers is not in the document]
brandyscarney: Oh yeah I was using that for testing because this fix broke the icons inside of a button. I should probably just remove the entire file.

core/src/components/item/test/buttons/item.e2e.ts-snapshots/item-buttons-dark-diff-ios-ltr-Mobile-Safari-linux.png
brandyscarney: I think this is another issue with our screenshots. 😩 I'm not able to reproduce this locally on the branch.

core/src/components/popover/test/nested/popover.e2e.ts-snapshots/popover-nested-ios-rtl-Mobile-Firefox-linux.png
brandyscarney: Yeah this seems like another screenshot issue. I don't modify this code at all.

core/src/components/item/test/buttons/item.e2e.ts-snapshots/item-buttons-dark-diff-ios-ltr-Mobile-Safari-linux.png
liamdebeasi: Screenshots are working correctly here. I can reproduce the behavior locally with Safari 16.4 
image

However, I _cannot_ reproduce the behavior with Safari 16.2 (Playwright uses 16.4 I believe)

brandyscarney: Oh thanks! I am on 16.3 and was not able to reproduce it.

colabottles:
I've been working with Ionic for a few months now and had been going through old issues and found #8700 which led me here. I have a strong objection, as do other accessibility experts, about using truncated text on interactive elements. Accessibility-wise, it is a concern. https://yatil.net/blog/text-overflow-ellipsis-harmful

I would object to implementing this and suggest hyphenating words that are long or button text that overflows the button.

This could cause issues not only with screen readers but also has the potential to cause [Reflow](https://www.w3.org/WAI/WCAG21/Understanding/reflow.html) and break [Text Resize](https://www.w3.org/WAI/WCAG21/Understanding/resize-text.html) and also for users that need wider text spacing as well, this could fail [Text Spacing](https://www.w3.org/WAI/WCAG21/Understanding/text-spacing.html).

liamdebeasi:
Thanks for the feedback, @colabottles. I'll talk this over with the team and see if there's a good resolution we can come to.

brandyscarney:
After discussing with the team, we are going to move forward with wrapping the text to the next line when the button text overflows, instead of adding an ellipsis, per the comment by @colabottles. 

I am going to close this PR In favor of https://github.com/ionic-team/ionic-framework/pull/27547

liamdebeasi:
Hey @colabottles,

Wanted to provide an update on this. As noted in https://github.com/ionic-team/ionic-framework/pull/27406#issuecomment-1592024745, we decided to move forward with having text in `ion-button` wrap to the next line instead of truncating.

During development we identified some instances where this new behavior may cause unexpected layout changes in developer applications. In particular, buttons inside of list headers may wrap instead of displaying on the same line even if the text was never cut off. As a result, we have decided to enable this behavior in Ionic 8.0, so developers can make any necessary changes at their own pace. We wanted to avoid developers being surprised with potential breakages when updating their installed version of Ionic.

However, the changes Brandy made in https://github.com/ionic-team/ionic-framework/pull/27547 make it so developers will be able to opt-in to the text wrapping behavior early once Ionic 7.2 ships (releasing tomorrow if all goes well). Developers will be able to either add the `ion-text-wrap` class to `ion-button`, or set `white-space: normal` on `ion-button` in their CSS.

-----

Once Ionic 7.2 ships we will also launch an accessibility guide on the `ion-button` component documentation. This guide will go over why developers should have their text wrap instead of truncate, and it will show them how to opt-in to this new wrapping behavior early.

You can preview the documentation here: https://ionic-docs-git-fw-1808-a11y-ionic1.vercel.app/docs/api/button#accessibility (Note that the playground example may look a little funky since it is still pulling in Ionic 7.1. This should be fixed once Ionic 7.2 ships).

Thanks again for all your help, we really appreciate it!

PR #27408, opened 2023-05-05
feat(range): add label prop

amandaejohnston:
Issue number: N/A

---------

Labels on `ion-range` can only be set via the `label` slot, which is cumbersome when only plain text is needed.

## What is the new behavior?

Label prop added. If both the prop and slot are used, the prop will take priority.

- [ ] Yes
- [x] No

### DIFF
--- a/core/src/components/range/range.tsx
+++ b/core/src/components/range/range.tsx
@@ -97,6 +97,13 @@ export class Range implements ComponentInterface {
   @Prop() name = this.rangeId;
 
   /**
+   * The text to display as the control's label. Use this over the `label` slot if
+   * you only need plain text. If both this and the slot are used, this property
+   * takes priority.
+   */
+  @Prop() label?: string;
+
+  /**
    * Show two knobs.
    */
   @Prop() dualKnobs = false;
@@ -602,7 +609,7 @@ Developers can dismiss this warning by removing their usage of the "legacy" prop
   }
 
   private renderRange() {
-    const { disabled, el, rangeId, pin, pressedKnob, labelPlacement } = this;
+    const { disabled, el, rangeId, pin, pressedKnob, labelPlacement, label } = this;
 
     const mode = getIonMode(this);
 
@@ -629,7 +636,7 @@ Developers can dismiss this warning by removing their usage of the "legacy" prop
               'label-text-wrapper-hidden': !this.hasLabel,
             }}
           >
-            <slot name="label"></slot>
+            {label !== undefined ? <div class="label-text">{label}</div> : <slot name="label"></slot>}
           </div>
           <div class="native-wrapper">
             <slot name="start"></slot>
@@ -642,7 +649,7 @@ Developers can dismiss this warning by removing their usage of the "legacy" prop
   }
 
   private get hasLabel() {
-    return this.el.querySelector('[slot="label"]') !== null;
+    return this.el.querySelector('[slot="label"]') !== null || this.label !== undefined;
   }
 
   private renderRangeSlider() {
--- a/core/src/components/range/test/label/range.e2e.ts
+++ b/core/src/components/range/test/label/range.e2e.ts
@@ -125,6 +125,32 @@ configs().forEach(({ title, screenshot, config }) => {
         expect(await range.screenshot()).toMatchSnapshot(screenshot(`range-items-fixed`));
       });
     });
+
+    test.describe('range: label prop', () => {
+      test('should render label in the start placement', async ({ page }) => {
+        await page.setContent(`<ion-range label-placement="start" label="Volume"></ion-range>`, config);
+
+        const range = page.locator('ion-range');
+
+        expect(await range.screenshot()).toMatchSnapshot(screenshot(`range-label-prop-start`));
+      });
+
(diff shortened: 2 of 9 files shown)


core/src/components/range/test/label/range.e2e.ts
```diff
@@ -125,6 +125,32 @@ configs().forEach(({ title, screenshot, config }) => {
         expect(await range.screenshot()).toMatchSnapshot(screenshot(`range-items-fixed`));
       });
     });
+
+    test.describe('range: label prop', () => {
```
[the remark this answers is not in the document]
amandaejohnston: I intentionally left all the configs in because the prop needed some styling separate from the slot, so I figured it would be good to verify no visual regressions in all cases, but I could go either way on it, honestly.

core/src/components/range/range.tsx
```diff
@@ -642,7 +649,7 @@ Developers can dismiss this warning by removing their usage of the "legacy" prop
   }
 
   private get hasLabel() {
-    return this.el.querySelector('[slot="label"]') !== null;
+    return this.el.querySelector('[slot="label"]') !== null || this.label !== undefined;
```
sean-perkins, line 652: Can we check for the `label` being defined first? It will logically be less expensive than querying the DOM.

core/src/components/range/range.tsx
```diff
@@ -96,6 +96,13 @@ export class Range implements ComponentInterface {
    */
   @Prop() name = this.rangeId;
 
+  /**
+   * The text to display as the control's label. Use this over the `label` slot if
+   * you only need plain text. If both this and the slot are used, this property
```
sean-perkins, line 101: ~ Suggestion for tweaking the verbiage, but I am also fine with the changes as-is:

> The plain text to display as the control's label. The `label` property will take priority over the `label` slot if both are used.

sean-perkins (changes requested):
Can we add an additional test for when using both a `label` property and `label` slot? This can likely be a unit test, just to query for the expected DOM node.

sean-perkins (changes requested):
Can we add an additional test for when using both a `label` property and `label` slot? This can likely be a unit test, just to query for the expected DOM node.

core/src/components/range/range.tsx
```diff
@@ -629,7 +636,7 @@ Developers can dismiss this warning by removing their usage of the "legacy" prop
               'label-text-wrapper-hidden': !this.hasLabel,
             }}
           >
-            <slot name="label"></slot>
+            {label !== undefined ? <div class="label-text">{label}</div> : <slot name="label"></slot>}
```
liamdebeasi, line 639: Should we be conditionally rendering the `<slot>` or should we be conditionally hiding/showing the slot?

liamdebeasi (changes requested):
Can we add a usage of the `label` prop to the `a11y` file so we can run axe checks?

liamdebeasi (changes requested):
Can we add a usage of the `label` prop to the `a11y` file so we can run axe checks?

core/src/components/range/range.tsx
amandaejohnston: I kept the "plain text" bit as-is since I like the current more explicit wording, but I like the priority part better in yours. 👍

core/src/components/range/range.tsx
amandaejohnston: I went with conditional rendering for performance reasons; there's little reason to render the slot content if it's just going to be hidden anyway. Do you think it should be hidden instead?

mapsandapps (approved):

sean-perkins (approved):

core/src/components/range/range.tsx
liamdebeasi: Let's keep it the way you have it. We should be able to easily change the behavior to hide the slot if the current approach causes problems.

liamdebeasi (commented):
Is there a docs PR associated with this?

amandaejohnston:
Docs PR created: https://github.com/ionic-team/ionic-docs/pull/2955

core/src/components/range/range.tsx
liamdebeasi, line 639: Nit: Why do we not render `.label-text` when using the slot?

In other words, why not do `<div class="label-text">{ label === undefined ? <slot name="label"></slot> : label }</div>`

amandaejohnston: For consistency with other components. Components that use a prop wrap it in `.label-text` (e.g. [input](https://github.com/ionic-team/ionic-framework/blob/FW-3669/core/src/components/input/input.tsx#L587)) while those that use a slot don't (e.g. [checkbox](https://github.com/ionic-team/ionic-framework/blob/FW-3669/core/src/components/checkbox/checkbox.tsx#L268)). I can add the wrapper for both cases here if desired, though.

liamdebeasi: You can leave it as is if you'd like. Main reason I ask is I'm going to need to do something similar with my `ion-select` label slot work, so would be good to be consistent. I'll make sure my PR follows what you're doing here.

PR #27417, opened 2023-05-08
fix(footer, tab-bar): wait for resize before re-showing

### LINKED ISSUE #25990: bug: Tab bar is unhidden too quickly after keyboard dismiss
### Prerequisites

### Ionic Framework Version

- [ ] v4.x
- [ ] v5.x
- [X] v6.x
- [ ] Nightly

### Current Behavior

When the keyboard is dismissed the tab bar temporarily flashes on the screen as it is unhidden before the webview has expanded back to the full size of the viewport.

https://user-images.githubusercontent.com/76167627/191752633-e4068d12-a205-421b-9005-ba5194948679.mp4

### Expected Behavior

The tab bar should not be unhidden until the web view has expanded back to its full size so that the tab bar does not flash in the center of the screen.

### Steps to Reproduce

Use the default capacitor keyboard config:
```
// capacitor-config.ts
Keyboard: {
    resize: KeyboardResize.Native,
    resizeOnFullScreen: true,
},
```
This happens on any page with an `IonTabBar`.

### Code Reproduction URL

_No response_

### Ionic Info

Ionic:

   Ionic CLI       : 6.20.1 (/usr/local/lib/node_modules/@ionic/cli)
   Ionic Framework : @ionic/react 6.2.8

Capacitor:

   Capacitor CLI      : 4.3.0
   @capacitor/core    : 4.3.0
   @capacitor/ios     : 4.3.0

Utility:

   native-run  : 1.7.0

System:

   NodeJS : v16.14.2 (/usr/local/bin/node)
   npm    : 8.7.0
   OS     : macOS Monterey

### Additional Information

_No response_

liamdebeasi:
Issue number: resolves #25990

---------

The tab bar and footer are being shown too soon after the keyboard begins to hide. This is happening because the webview resizes _after_ the keyboard begins to dismiss. As a result, it is possible for the tab bar and footer to briefly appear on the top of the keyboard in environments where the webview resizes.

## What is the new behavior?

- The tab bar and footer wait until after the webview has resized before showing again

| before | after |
| - | - |
| <video src="https://user-images.githubusercontent.com/2721089/236905066-42ac17a5-a5bf-458b-9c62-005fcce05e20.MP4"></video> | <video src="https://user-images.githubusercontent.com/2721089/236905185-d2f539d1-6d93-4385-b1cb-24dd7aa06393.MP4"></video> |

This code works by adding an optional parameter to the keyboard controller callback called `waitForResize`. When defined, code within Ionic can wait for the webview to resize as a result of the keyboard opening or closing. Tab bar and footer wait for this `waitForResize` promise to resolve before re-showing the relevant elements.

This `waitForResize` parameter is only only defined when all of the following are two:

**1. The webview resize mode is known and is _not_ "None".**

If the webview resize mode is unknown then either the Keyboard plugin is not installed (in which case the tab bar/footer are never hidden in the first place) or the app is being deployed in a browser/PWA environment (in which case the web content typically does not resize). If the webview resize mode is "None" then that means the keyboard plugin is installed, but the webview is configured to never resize when the keyboard opens/closes. As a result, there is no need to wait for the webview to resize.

**2. The webview has previously resized.**

If the keyboard is closed _before_ the opening keyboard animation completes then it is possible for the webview to never resize. In this case, the webview is at full height and the tab bar/footer can immediately be re-shown.

------

Under the hood, we use a [ResizeObserver](https://developer.mozilla.org/en-US/docs/Web/API/ResizeObserver) to listen for when the web content resizes. Which element we listen on depends on the resize mode set in the developer's Capacitor app. We determine this in the `getResizeContainer` function.

From there, we wait for the ResizeObserver callback, then wait 1 more frame so the promise resolves _after_ the resize has finished.

- [ ] Yes
- [x] No

## Other information

Dev build: `7.0.6-dev.11683905366.13943af0`


core/src/utils/keyboard/keyboard-controller.ts
```diff
@@ -1,33 +1,98 @@
-import { win } from '../window';
+import { raf } from '@utils/helpers';
+import { win } from '@utils/window';
+
+import { KeyboardResize, Keyboard } from '../native/keyboard';
```
liamdebeasi, line 4: This gave me a strange build error when I tried to do `@utils/native/keyboard`:

```
[ ERROR ]  Rollup: Unresolved Import
           Could not resolve '../native/keyboa' from ./src/utils/keyboard/keyboard-controller.ts
```

thetaPC: Was this resolved? I wasn't able to reproduce any build errors.

liamdebeasi: I'm still getting the issue locally when doing `npm run start`:

```
[ ERROR ]  Rollup: Unresolved Import
           Could not resolve '../native/keyboa' from
           ./src/utils/keyboard/keyboard-controller.ts
```

core/src/utils/keyboard/keyboard-controller.ts
```diff
@@ -1,33 +1,175 @@
...
   let keyboardVisible: boolean;
+  /**
+   * This lets us determine if the webview content
+   * has resized as a result of the keyboard.
+   */
+  let initialResizeContainerHeight: number;
+
+  const init = async () => {
+    const resizeOptions = await Keyboard.getResizeMode();
+    const resizeMode = resizeOptions == undefined ? undefined : resizeOptions.mode;
```
thetaPC, line 64: Change `==` to `===`

thetaPC (changes requested):

core/src/utils/keyboard/keyboard-controller.ts
thetaPC: I'm also not able to reproduce any issues when running `npm run start`. Does it happen as soon as the command is entered or does it show when on a specific page?

amandaejohnston: I'm getting the error as soon as I run `npm start`, or at least a bit later once the build has finished. I did also try `rm ./dist` and `npm install` in case I had some build cache or oudated Stencil issues, but the error's still occuring. Odd 🤔

amandaejohnston (approved):
LGTM once Maria's comment is addressed 👍

core/src/utils/keyboard/keyboard-controller.ts
liamdebeasi: I can ask the Stencil team about this during our sync today, but I don't think it's worth holding up this PR over.

thetaPC: Super weird, I'll be curious to know the outcome.

thetaPC (approved):
LGTM

core/src/utils/keyboard/keyboard-controller.ts
liamdebeasi: Spoke with Ryan, and he thinks the compiler may be expecting to see a `ts` extension since only 2 characters are being cut off (i.e. the length of the file extension). Tanner is working on a similar bug, and Ryan had me try dev build with Tanner's proposed fix. However, the proposed fix did not change the behavior here.

rwaskiewicz: For posterity, the fix for this is tied to https://github.com/ionic-team/stencil/pull/4481

rwaskiewicz: Fix landed/released in Stencil v3.4.1 - https://github.com/ionic-team/stencil/releases/tag/v3.4.1

Issue #27268, opened 2023-04-24
bug: Radio group tab navigation is incorrect for modern form control syntax


sean-perkins:
### Prerequisites

- [X] I have read the [Contributing Guidelines](https://github.com/ionic-team/ionic-framework/blob/main/.github/CONTRIBUTING.md#creating-an-issue).
- [X] I agree to follow the [Code of Conduct](https://ionicframework.com/code-of-conduct).
- [X] I have searched for [existing issues](https://github.com/ionic-team/ionic-framework/issues) that already report this problem, without success.

### Ionic Framework Version

v7.x

### Current Behavior

In v7 the tab navigation for `ion-radio-group` is incorrect with the modern form syntax.

Pressing `Tab` will focus the next available `ion-radio` in the group. Once reaching the last radio, it will advance to focus the next available radio in the nearest `ion-radio-group`.

|Legacy|Modern|
|---|---|
|<video src="https://user-images.githubusercontent.com/13732623/234109075-d688d7ab-dfc3-4d7d-bc03-d89ff89ca35a.mp4"></video>|<video src="https://user-images.githubusercontent.com/13732623/234109274-9e2bf07f-5997-4407-bc3a-815e5e64a180.mp4"></video>

### Expected Behavior

Pressing `Tab` should focus the next available option in the nearest `ion-radio-group`. It should not advance to the next `ion-radio` in the current group. The `up/down` arrow keys should be responsible for that behavior.

### Steps to Reproduce

1. Render multiple radio groups with the modern form syntax
2. Press `Tab` (repeatedly)
3. Observe: The next `ion-radio` is focused in the group
4. Observe: The focus will change between the radio groups after reaching the last radio in the group.

---

If using the reproduction, run the app and you can click the "Legacy Syntax" or "Modern Syntax" item to experiment with the different syntax and `Tab` key behavior.

### Code Reproduction URL

https://github.com/sean-perkins/ionic-radio-group-tabs-bug

### Ionic Info

N/A

thetaPC:
Thank you for the issue! The fix has been released in [v7.0.11](https://github.com/ionic-team/ionic-framework/releases/tag/v7.0.11).

PR #27420, opened 2023-05-08
fix(select-popover): migrate inner checkboxes/radios to new syntax

amandaejohnston:
Issue number: N/A

---------

`ion-select` with `interface="popover"` triggers console warnings when opened, due to the radios and checkboxes inside still using the legacy form control syntax.

## What is the new behavior?

Controls migrated. Some extra styling was also added to maintain the same margins as before, since the new syntax has a slightly different look in that regard.

- [ ] Yes
- [x] No

Well, sort of. This PR will cause selects to run into https://github.com/ionic-team/ionic-framework/issues/27268 and https://github.com/ionic-team/ionic-framework/issues/27438.

### DIFF
--- a/core/src/components/select-popover/select-popover.ios.scss
+++ b/core/src/components/select-popover/select-popover.ios.scss
@@ -1,13 +1,6 @@
 @import "./select-popover";
 @import "./select-popover.ios.vars";
 
-// TODO: remove and make ion-items lines="full" instead once
-// https://github.com/ionic-team/ionic-framework/issues/26955 is fixed
-ion-list ion-item {
-  --inner-border-width: 0;
-  border-bottom: 0.55px var(--border-style) var(--border-color);
-}
-
 ion-list ion-checkbox::part(container) {
   @include margin(2px, 8px, 0, 0);
 }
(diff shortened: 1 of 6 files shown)


core/src/components/select-popover/select-popover.ios.scss
```diff
@@ -1,3 +1,17 @@
 @import "./select-popover";
 @import "./select-popover.ios.vars";
 
+// TODO: remove and make ion-items lines="full" instead once
+// https://github.com/ionic-team/ionic-framework/issues/26955 is fixed
+ion-list ion-item {
+  --inner-border-width: 0;
+  border-bottom: 0.55px var(--border-style) var(--border-color);
+}
```
amandaejohnston, line 9: If this is acceptable, I can create a tech debt ticket.

amandaejohnston:
If the changes are acceptable, I can add tech debt tickets to the new `TODO` and test skips.

liamdebeasi:
I'm going to close this until the blocking tasks are resolved. Feel free to re-open once the blockers have been addressed.

PR #27424, opened 2023-05-08
fix(spinner): allow resizing of dots, bubbles, and circles

### LINKED ISSUE #18115: bug: increasing size of ion-spinner results in blurry animation on iOS
# Bug Report

**Ionic version:**
[x] **4.2**

**Current behavior:**
I am using the `LoadingController` in `@ionic/angular` to show a loading animation with spinner:

```
    this.loadingController.create({
      cssClass: 'loading-container',
      spinner: 'circles',
      translucent: true
    })
```

It's quite small on the screen, so I'm trying to increase the size of it using CSS.

Sadly there don't appear to be any good ways to increase the size of `<ion-spinner>` with Ionic v4.

Using CSS `transform: scale(5);` *will* scale it up. However, because of the CSS animations, it ends up looking blurry on iOS. (android is fine though)

From what I understand, this is a long standing bug in mobile Safari, that occurs when scale is applied to a CSS animation. (if you google for this, you'll get a tonne of links)

Note that in Ionic v1, sizing an `<ion-spinner>` wasn't an issue, since it was possible to size it by just doing this:
```
      .spinner svg {
        width: 15rem;
        height: 15rem;
      }
```
Which avoids the `transform: scale` Safari bug.

Unfortunately this method doesn't work in Ionic v4, and it results in the `<ion-spinner>` looking very odd if tried.

End result is that there's no good way to increase the size of `<ion-spinner>`, and I'm kind of forced to use the `transform: scale` approach and just have it look all blurry on iOS.

**Expected behavior:**
It should be possible to size a `<ion-spinner>` without making it blurry

**Steps to reproduce:**
As above

**Related code:**
As above

**Other information:**
None

**Ionic info:**
```
Ionic:

   ionic (Ionic CLI)             : 4.12.0 (/Users/asgeo1/.node/lib/node_modules/ionic)
   Ionic Framework               : @ionic/angular 4.2.0
   @angular-devkit/build-angular : 0.12.4
   @angular-devkit/schematics    : 7.2.4
   @angular/cli                  : 7.2.4
   @ionic/angular-toolkit        : 1.4.0

Cordova:

   Cordova Platforms     : android 7.1.4, ios 4.5.5
   Cordova Plugins       : cordova-plugin-ionic-keyboard 2.1.3, cordova-plugin-ionic-webview 3.1.2, (and 20 other plugins)

System:

   Android SDK Tools : 26.1.1 (/Users/asgeo1/Development/Android/sdk)
   ios-deploy        : 1.9.4
   ios-sim           : 5.0.6
   NodeJS            : v10.10.0 (/usr/local/bin/node)
   npm               : 6.8.0
   OS                : macOS Mojave
   Xcode             : Xcode 10.2.1 Build version 10E1001
```

mapsandapps:
Issue number: Resolves #18115

---------

Resizing the `ion-spinner` by setting the width & height using CSS works for most spinner variants, but not for `dots`, `bubbles`, and `circles`. The spacing between the circles on these spinner variants is not scaling with the spinner.

## What is the new behavior?

- `ion-spinner` may now be resized by setting the width & height using CSS for all variants.

- [ ] Yes
- [x] No

### Before

Screenshot 2023-05-08 at 5 17 59 PM

### After

Screenshot 2023-05-08 at 5 18 45 PM


sean-perkins (approved):
This looks fantastic, great work!

I updated the PR description to include `Resolve #` so that the ticket is automatically closed when the PR is merged.

core/src/components/spinner/test/resize/spinner.e2e.ts
```diff
@@ -0,0 +1,20 @@
+import { expect } from '@playwright/test';
+import { configs, test } from '@utils/test/playwright';
+
+/**
+ * This behavior does not vary across directions.
+ */
+configs({ directions: ['ltr'] }).forEach(({ title, screenshot, config }) => {
```
liamdebeasi, line 7: Spinner has no per-mode styles, so we don't need both iOS and MD screenshots.

liamdebeasi: Yeah you can, though spinner does have per-mode _behavior_, so we should probably have a spec test that checks that the default spinner is set correctly on each mode: https://github.com/ionic-team/ionic-framework/blob/eba4c7f6e629dedf68de24a512e92b7ce4c43154/core/src/components/spinner/spinner.tsx#L42-L48

We might be able to check that the correct `spinner-*` class is set. 

I'd also be fine if we made a ticket and updated the non-resize tests as a separate task.

core/src/components/spinner/test/resize/spinner.e2e.ts
```diff
@@ -0,0 +1,33 @@
...
+ */
+configs({ directions: ['ltr'], modes: ['md'] }).forEach(({ title, screenshot, config }) => {
+  test.describe(title('spinner: resize'), () => {
+    test.beforeEach(async ({ page }) => {
+      await page.setViewportSize({ width: 320, height: 340 });
+    });
+    test('should not have visual regressions', async ({ page }) => {
+      await page.setContent(
+        `
+        <ion-content id="content">
```
liamdebeasi, line 15: It doesn't look like the spinners are being resized at all. Is this intentional?

mapsandapps: Oops! Nope, I hadn't had a chance to look at the screenshots yet from CI, and I can't run this test file locally for some reason (but others are fine) 🤔

core/src/components/spinner/test/resize/spinner.e2e.ts
```diff
@@ -0,0 +1,33 @@
...
+          <ion-spinner name="dots"></ion-spinner>
+          <ion-spinner name="bubbles"></ion-spinner>
+          <ion-spinner name="circles"></ion-spinner>
+          <ion-spinner name="crescent"></ion-spinner>
+        </ion-content>
+      `,
+        config
+      );
+
+      await expect(page.locator('#content')).toHaveScreenshot(screenshot(`spinner-resize-diff`));
```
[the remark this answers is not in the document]
mapsandapps: thanks for the tips!

PR #27426, opened 2023-05-08
test(modal): remove redundant tests

liamdebeasi:
Issue number: N/A

---------

There were several redundant tests I missed during the initial generator pass.

## What is the new behavior?

https://github.com/ionic-team/ionic-framework/pull/27426/commits/3044c2dc4b184a6689cc69ef93f083a7ef93c722

- Sheet rendering does not vary across directions, so I removed those additional checks. Note: Modal _does_ have some RTL-specific styles but those are already captured in the basic screenshots.
- The backdrop interaction behavior does not vary across modes, so I removed those additional checks.

https://github.com/ionic-team/ionic-framework/pull/27426/commits/f5ff834196591b02f9a89872298d4b6e906564e1

- Similar to above, the card rendering does not vary across directions.

https://github.com/ionic-team/ionic-framework/pull/27426/commits/aba6c5c2a22e43e521d47a42868feaa32fccf0d5

- During my generator pass, I had broke the `card/modal.e2e.ts` tests into `card/modal-card.e2e.ts` and `card/modal-sheet.e2e.ts`, but I forgot to delete the original `card/modal.e2e.ts` file, so we had redundant tests running.

https://github.com/ionic-team/ionic-framework/pull/27426/commits/ae762d190e49d449fbbfd5320430575658e1327d

- [ ] Yes
- [x] No


thetaPC (approved):
LGTM

PR #27427, opened 2023-05-09
feat(input): input masking insertion - on hold, do not review

sean-perkins:
Issue number: Internal

---------

N/A

## What is the new behavior?

- Introduces the ability to type into a mask through insertion (entering new characters consecutively) to conform to a defined mask expression. 
  - This PR __does not__ support deletion/backspace

In order to support a developer friendly syntax with string-based attributes:
```html
<ion-input mask="^\d{0,4}$"></ion-input>
```

We are supporting a "string" value and converting it into a regular expression with `new RegExp(mask)`. `el.getAttribute` does not support resolving an attribute with an assigned `RegExp` instance. As a result, having a property in Stencil that is type of `RegExp`, cannot accept a string value and will resolve `undefined`. 

### Testing

You can open `/src/components/input/test/mask` to interact and experiment with the mask behavior. Fixed characters (such as `(`, `+`, etc.) accept any key as an input. For example, typing `a` in the US Phone Number mask will pre-fill `+1 ` portion of the mask, but not accept the `a` input key. This behavior we may adjust as the mask placeholder behavior is introduced on top of the maskito implementation style. Contrasting with [react-input-mask](http://sanniassin.github.io/react-input-mask/demo.html) where fixed characters are part of the mask placeholder and do not advance the cursor selection when entering an unaccepted mask value on a fixed character. I don't have strong opinions at this time as both accomplish "masking" the input. 

- [ ] Yes
- [x] No

### DIFF
--- a/core/src/utils/input-masking/types/mask-interface.ts
+++ b/core/src/utils/input-masking/types/mask-interface.ts
@@ -43,8 +43,17 @@ export type MaskPostprocessor = (elementState: ElementState, initialElementState
 
 export interface MaskOptions {
   readonly mask: MaskExpression | ((elementState: ElementState) => MaskExpression);
+  /**
+   * @internal
+   */
   readonly preprocessor?: MaskPreprocessor;
+  /**
+   * @internal
+   */
   readonly postprocessor?: MaskPostprocessor;
+  /**
+   * @internal
+   */
   readonly overwriteMode?: 'replace' | 'shift' | ((elementState: ElementState) => 'replace' | 'shift');
 }
 
--- a/core/api.txt
+++ b/core/api.txt
@@ -553,7 +553,7 @@ ion-input,prop,inputmode,"decimal" | "email" | "none" | "numeric" | "search" | "
 ion-input,prop,label,string | undefined,undefined,false,false
 ion-input,prop,labelPlacement,"end" | "fixed" | "floating" | "stacked" | "start",'start',false,false
 ion-input,prop,legacy,boolean | undefined,undefined,false,false
-ion-input,prop,mask,(string | RegExp)[] | RegExp | undefined,undefined,false,false
+ion-input,prop,mask,(string | RegExp)[] | RegExp | string | undefined,undefined,false,false
 ion-input,prop,maskPlaceholder,null | string | undefined,'_',false,false
 ion-input,prop,maskVisibility,"always" | "focus" | "never" | undefined,'always',false,false
 ion-input,prop,max,number | string | undefined,undefined,false,false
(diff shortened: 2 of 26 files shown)


core/src/utils/input-masking/types/mask-interface.ts
```diff
@@ -43,8 +43,17 @@ export type MaskPostprocessor = (elementState: ElementState, initialElementState
 
 export interface MaskOptions {
   readonly mask: MaskExpression | ((elementState: ElementState) => MaskExpression);
+  /**
```
sean-perkins, line 46: The `preprocessor`, `postprocessor` and `overwriteMode` are considered internal APIs at this time. They are public API of `maskito`, but we did not agree/design out these options with input masking in our design document. 

I opted for leaving the internal API surface because:
1. It is very minimal (the functions act as noop, returning the existing mask value/selection).
2. These APIs offer fairly complex mask functionality that we likely want to discuss exposing in future iterations. 

As our mask implementation is refined (taking the best from existing solutions + applying our API requirements), this API surface may be removed if it ends up being unneeded or introducing undesired complexity.

amandaejohnston (approved):
I really like how the masking behavior feels right now, great work! In a later PR, we should document when to use an array of characters vs. a regular regex. (Looks like the former is better for a fixed length, while the latter is better for filtering out certain types of characters.)

While playing with this, I noticed a few things that make sense, but seem like they'll be common pitfalls that we could also document and warn against:
- If you do something like `^\d{5}$` (5 digits), you can't actually type anything, only paste in a 5-digit number. You have to do `^\d{0,5}$` instead.
- If you do `\d{0,5}` (without the `^` and `$`), you can type anything you want, since even `"abc"` matches that.
- `\d{1,5}` (which seems like a logical fix to the above) makes it so you can still type anything you want, as long as you start with a number or paste something with a number in it.

sean-perkins:
Spoke with Liam + Alex from the Maskito project, going to hit pause on this PR until after Monday of next week (when I'm back). Maskito might allow us to support masking without directly bundling it or copying their implementation. I'm going to do some discovery work on Monday to validate our original requirements against this. If it works, it would be similar to Swiper.js, where we have implementation guides, but do not need to maintain the implementation ourselves. There may be some opportunity for us to contribute back to Maskito however, with additional consideration for touch-based devices. 

I think they may also be open to us contributing back documentation PRs to the code (similar to what I have done here), to help make the code easier to reason about for someone new to input masking.

sean-perkins:
Closing this pull request based on our revised direction.

Issue #27415, opened 2023-05-08
bug: Ionic 7 React - Ionic classes are overwritten when changing class attribute on component


DevonStern:
### Prerequisites

- [X] I have read the [Contributing Guidelines](https://github.com/ionic-team/ionic-framework/blob/main/.github/CONTRIBUTING.md#creating-an-issue).
- [X] I agree to follow the [Code of Conduct](https://ionicframework.com/code-of-conduct).
- [X] I have searched for [existing issues](https://github.com/ionic-team/ionic-framework/issues) that already report this problem, without success.

### Ionic Framework Version

v7.x

### Current Behavior

When changing the value of the class attribute on an Ionic component in Ionic 7 React, the Ionic-specific classes for that component are removed.

For example, in a blank Ionic 7 React project, I added an IonButton component to the Home page and set the class attribute based on the value of a state variable from the useState hook. I then set up the IonButton to toggle the value of the state variable when clicked.

Before toggling the state, inspecting the ion-button element shows the following classes:

> md button button-solid ion-activatable ion-focusable

After toggling the state for the first time and every time thereafter, the ion-button element shows only the class I applied (or no class if I applied an empty string). All of the classes applied by Ionic are gone.

### Expected Behavior

As happened in Ionic 6, the classes applied by Ionic should be retained even after changing the value of the class attribute. So any classes that I apply through the class attribute should be in addition to the Ionic-specific classes, even after changing the value of the class attribute.

### Steps to Reproduce

1. In an Ionic 7 React app, add an Ionic component to the page (for example, IonButton)
2. Add a useState hook that defaults to false
3. Set up the button onClick handler to toggle the value of the state variable using the setter
4. Depending on the value of the state variable, set the class attribute on the Ionic component (could be set to any string in either case, though at least one should not be an empty string)

### Code Reproduction URL

https://github.com/DevonStern/ionic-bug-ionic-classes-overwritten

### Ionic Info

Ionic:

   Ionic CLI       : 6.19.0 (C:\Users\DevonStern\AppData\Roaming\npm\node_modules\@ionic\cli)
   Ionic Framework : @ionic/react 7.0.5

Capacitor:

   Capacitor CLI      : 5.0.1
   @capacitor/android : not installed
   @capacitor/core    : 5.0.1
   @capacitor/ios     : not installed

Utility:

   cordova-res : 0.15.4
   native-run  : 1.7.2

System:

[...]

PR #27432, opened 2023-05-09
fix(react): remove incorrect class key from IonicReactProps

liamdebeasi:
Issue number: N/A

---------

As noted in https://github.com/ionic-team/ionic-framework/issues/27415, `IonicReactProps` still has `class` as an exposed attribute on Ionic React components. React components should use the `className` property instead. It looks like this was left over code from the Ionic React beta that never got removed.

`class` was originally added in https://github.com/ionic-team/ionic-framework/commit/c79e74b91fdfd931f7ae082cfc374b2e55ee6d06#diff-065219c56a8c5dfc32564fac317b8cc9c03f1fbcb7f9cc2de0c5581970f21b51

And `className` was added in https://github.com/ionic-team/ionic-framework/commit/ab0f92e01f892c1c5f0db6e1e8467fe553e48bdd#diff-065219c56a8c5dfc32564fac317b8cc9c03f1fbcb7f9cc2de0c5581970f21b51

## What is the new behavior?

- Removes the `class` key

- [ ] Yes
- [x] No

### DIFF
--- a/packages/react/src/components/IonicReactProps.ts
+++ b/packages/react/src/components/IonicReactProps.ts
@@ -1,5 +1,4 @@
 export interface IonicReactProps {
-  class?: string;
   className?: string;
   style?: { [key: string]: any };
 }


packages/react/src/components/IonicReactProps.ts
```diff
@@ -1,5 +1,4 @@
 export interface IonicReactProps {
-  class?: string;
```
liamdebeasi, line 2: Threading for discussion:

It's possible that there are other developers relying on this `class`. However, using this will cause all Ionic classes to be wiped, so I can't imagine that the experience is very good. 

Reviewers: Do you have any concerns about potential breakages?

From my perspective, React devs should be using `className`, so this `class` never should have been exposed.

sean-perkins: I don't have any concerns with this breaking change (in a patch). `className` is the standard for all React apps/libraries. `class` is confusing and leads to a degraded experience.

sean-perkins (approved):

PR #27433, opened 2023-05-09
test(picker): clean up tests

liamdebeasi:
Issue number: N/A

---------

There are several redundant tests I missed during my generator pass.

## What is the new behavior?

https://github.com/ionic-team/ionic-framework/pull/27433/commits/feb1f6c1388c234d194d31bb7ef99c82c8642878

https://github.com/ionic-team/ionic-framework/pull/27433/commits/f14b15becac3c644cba25d9490f094bb92b760e8

- Removed the "standalone" screenshot in favor of just checking that the overlay is visible. Rendering of the contents of the overlay are not dependent on `ion-app` being present.

- [ ] Yes
- [x] No


thetaPC (approved):
LGTM

PR #27435, opened 2023-05-09
test(toast): clean up tests

liamdebeasi:
Issue number: N/A

---------

There are several redundant tests I missed during my generator pass.

## What is the new behavior?

https://github.com/ionic-team/ionic-framework/pull/27435/commits/ec529cfe1292e3a09f071f98a0220e3ba4cfeada

https://github.com/ionic-team/ionic-framework/pull/27435/commits/3771d8e5b2e2d92f6b1a1eb87977f0706893d699

- Removed the "standalone" screenshot in favor of just checking that the overlay is visible. Rendering of the contents of the overlay are not dependent on `ion-app` being present.

- [ ] Yes
- [x] No


thetaPC (approved):
LGTM

PR #27440, opened 2023-05-09
fix(content): prevent forceUpdate in SSR

### LINKED ISSUE #27411: bug: ion-content fullscreen is not compatible with pre-rendering
### Prerequisites

### Ionic Framework Version

v7.x

### Current Behavior

SSR not working after upgrading to angular 16, even in ionic starter app. After starting dev server and opening the app, the page loads until the server throws out of memory. 
With a production build server throws "NotYetImplemented" error and returns only original index.html file without rendered page.

### Expected Behavior

SSR should work with angular 16.

### Steps to Reproduce

1. Start SSR dev server (npm run dev:ssr)
2. Visit app located on localhost:4200
3. Loading continues until the server throws out of memory 
 OR
1. Build app
2. Run npm run serve:ssr 
3. Visit app with disabled javascript

### Code Reproduction URL

https://github.com/WDrzewiecki/ionic-ssr

### Ionic Info

Ionic:

   Ionic CLI                     : 7.1.1 (/usr/local/lib/node_modules/@ionic/cli)
   Ionic Framework               : @ionic/angular 7.0.5
   @angular-devkit/build-angular : 16.0.0
   @angular-devkit/schematics    : 16.0.0
   @angular/cli                  : 16.0.0

Capacitor:

   Capacitor CLI      : 5.0.1
   @capacitor/core    : 5.0.1

Utility:

   cordova-res : 0.15.4
   native-run  : 1.7.2

System:

   NodeJS : v18.16.0 (/usr/local/bin/node)
   npm    : 9.5.1
   OS     : macOS Unknown

### Additional Information

_No response_

sean-perkins:
Issue number: Resolves #27411, https://github.com/ionic-team/stencil/issues/2429, https://github.com/ionic-team/stencil/issues/4076

---------

Rendering `<ion-content fullscreen="true">` in an Angular Universal project will result in a javascript heap exception and the browser tab timing out.

`forceUpdate` is not a compatible API with pre-rendering and results in calling itself indefinitely.

## What is the new behavior?

- Updates the fullscreen implementation of `ion-content` to only call `forceUpdate` and related functionality when running in a browser environment. 
- `<ion-content fullscreen="true">` is compatible with Angular Universal

- [ ] Yes
- [x] No

## Other information

Dev-build: `7.0.6-dev.11683653232.1ddc5840` ✅

### DIFF
--- a/core/src/components/content/content.tsx
+++ b/core/src/components/content/content.tsx
@@ -172,6 +172,10 @@ export class Content implements ComponentInterface {
   }
 
   private resize() {
+    /**
+     * Only force update if the component is rendered in a browser context.
+     * Using `forceUpdate` in a server context with pre-rendering can lead to an infinite loop.
+     */
     if (Build.isBrowser) {
       if (this.fullscreen) {
         readTask(() => this.readDimensions());
@@ -190,9 +194,7 @@ export class Content implements ComponentInterface {
     if (dirty) {
       this.cTop = top;
       this.cBottom = bottom;
-      if (Build.isBrowser) {
-        forceUpdate(this);
-      }
+      forceUpdate(this);
     }
   }
 
--- a/core/src/components/content/content.tsx
+++ b/core/src/components/content/content.tsx
@@ -1,5 +1,5 @@
 import type { ComponentInterface, EventEmitter } from '@stencil/core';
-import { Component, Element, Event, Host, Listen, Method, Prop, forceUpdate, h, readTask } from '@stencil/core';
+import { Build, Component, Element, Event, Host, Listen, Method, Prop, forceUpdate, h, readTask } from '@stencil/core';
 
 import { getIonMode } from '../../global/ionic-global';
 import type { Color } from '../../interface';
@@ -172,11 +172,13 @@ export class Content implements ComponentInterface {
   }
 
   private resize() {
-    if (this.fullscreen) {
-      readTask(() => this.readDimensions());
-    } else if (this.cTop !== 0 || this.cBottom !== 0) {
-      this.cTop = this.cBottom = 0;
-      forceUpdate(this);
+    if (Build.isBrowser) {
+      if (this.fullscreen) {
+        readTask(() => this.readDimensions());
+      } else if (this.cTop !== 0 || this.cBottom !== 0) {
+        this.cTop = this.cBottom = 0;
+        forceUpdate(this);
+      }
     }
   }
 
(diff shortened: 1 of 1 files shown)


core/src/components/content/content.tsx
```diff
@@ -188,7 +190,9 @@ export class Content implements ComponentInterface {
     if (dirty) {
       this.cTop = top;
       this.cBottom = bottom;
-      forceUpdate(this);
+      if (Build.isBrowser) {
```
liamdebeasi, line 193: `readDimensions` is only called by `resize` which already has the `Build.isBrowser` check. Do we need another `Build.isBrowser` check here?

core/src/components/content/content.tsx
```diff
@@ -172,11 +172,13 @@ export class Content implements ComponentInterface {
   }
 
   private resize() {
-    if (this.fullscreen) {
-      readTask(() => this.readDimensions());
-    } else if (this.cTop !== 0 || this.cBottom !== 0) {
-      this.cTop = this.cBottom = 0;
-      forceUpdate(this);
+    if (Build.isBrowser) {
```
liamdebeasi, line 175: Can we add a comment as to why this is needed?

liamdebeasi, line 175: And so I understand, this is to work around a stencil bug correct?

sean-perkins: This isn't a workaround per-say. The reported issues attached are resolved by this change.  

I had some discussions with Tanner around if `forceUpdate` should internally account for this. It checks for build conditions to see if it updatable: https://github.com/ionic-team/stencil/blob/main/src/runtime/update-component.ts#L328-L342, but should likely also check if it is in a browser context. At this time, we don't have enough historic data to confidently determine if `forceUpdate` should only execute in the browser or not. He is leading that conversation with the team and we will re-sync on the outcomes (if we can remove this in a future update of Stencil).

The problem that was occurring is that the pre-rendering logic would attempt to mock `ion-content` DOM tree, but would call `forceUpdate` as part of the first render, resulting in an infinite loop.

liamdebeasi: Thanks! Can we include this context in the comment too?

liamdebeasi: Also if there's an open stencil issue it would be good to link to that in the comment

liamdebeasi (approved):
Good to go once comments are addressed. thanks for the quick turnaround!

core/src/components/content/content.tsx
sean-perkins: Added some extra context + created a spike in JIRA for Stencil to track (and referenced it in the code comments).

PR #27447, opened 2023-05-10
fix(many): form controls labels have increased margin

### LINKED ISSUE #27129: bug: form controls do not have enough margin
### Prerequisites

### Ionic Framework Version

v7.x

### Current Behavior

Consider the following snippet:
```
<ion-content>
  <ion-radio-group>
    <!-- pre-migration version -->
    <ion-item>
      <ion-label>
        <div translate>xx</div>
      </ion-label>
      <ion-radio slot="start"></ion-radio>
    </ion-item>
    <ion-item>
      <ion-label>
        <div translate>xx</div>
      </ion-label>
      <ion-radio slot="start"></ion-radio>
    </ion-item>

    <!-- post-migration version with ion-item -->
    <ion-item>
      <ion-radio slot="start" label-placement="end" translate>
        <div translate>xx</div>
      </ion-radio>
    </ion-item>
    <ion-item>
      <ion-radio slot="start" label-placement="end" translate>
        <div translate>xx</div>
      </ion-radio>
    </ion-item>

    <!-- post-migration version without ion-item -->
    <ion-radio slot="start" label-placement="end" translate>
      <div translate>xx</div>
    </ion-radio>
    <ion-radio slot="start" label-placement="end" translate>
      <div translate>xx</div>
    </ion-radio>
  </ion-radio-group>
</ion-content>
```

The first block is the 6.x version of the radio buttons. For the second one, the label was moved into the body of the radio as the migration guide suggested, but the ion-item element was kept. For the third one, the ion-item element was dropped as well.

This snippet renders the following:
image

The third block I don't want to go too much details into, I assume the `<ion-item>` plays an important role in rendering the widget the way I want it to look. I just wnat to note that despite what the documentation says, it cannot just be removed without breaking a lot of things.

More important are the functional aspects.
- The red highlight indicates the click area. For the old version, the radio button can be selected when clicking anywhere in the whole row. The new version, however, requires precise clicking either on the text or the circle. Also important that when I hover over the line both with the older and the newer version, the whole row receives the light gray highlight, which can be confusing (I expected the presence of the highlight to indicate that I can click anywhere to select the radio button).
- With the old version, the grey item separator line reaches under the label, while with the new version it does not. If the label is longer (like it was in my original), the difference is even more noticeable, and will look awful when the consecutive lines contain labels of different width:
image
- The label is too close to the radio button, which looks bad both when there is just a small amount of text, and when there is a large amount of text. This is something I was able to fix with some custom css, but I think it should be looking nicer out of the box.
- When there is a lot of text in the label part, it fills out the whole row, there is no top and bottom padding, which looks awful. This I could not fix.

### Expected Behavior

Summarized from the above section:
- whole row should be clickable
- gray separator line should be visible under the label
- space between radio button and label text should be bigger
- padding above and below the label should be present

### Steps to Reproduce

1. Pull in the snippet above in any component.
2. Check the output.

### Code Reproduction URL

_No response_

### Ionic Info

[WARN] You are not in an Ionic project directory. Project context may be missing. // this is incorrect btw

Ionic:

   Ionic CLI : 7.0.1

Utility:

System:

   NodeJS : v19.8.1
   npm    : 9.5.1
   OS     : Windows 10

NX workspace. What you are interested in is probably the following:
    "@ionic/angular": "7.0.1",
    "@ionic/core": "7.0.1",

    "@angular-devkit/build-angular": "15.2.5",
    "@angular-devkit/core": "15.2.5",
    "@angular-devkit/schematics": "15.2.5",
    "@angular/cli": "15.2.5",

    "@capacitor/core": "4.7.3",
    "@capacitor/android": "4.7.3",
    "@capacitor/cli": "4.7.3",
    "@capacitor/ios": "4.7.3",

### Additional Information

_No response_

liamdebeasi:
Issue number: resolves #27129

---------

Material Design has 16px of margin between the form control and the label, but we have 8px: https://m2.material.io/components/selection-controls#usage

## What is the new behavior?

- Updates default margin from 8px to 16px for checkbox, input, radio, range, select, textarea, and toggle.

Note: This should only apply to labels that are on the same line as the form control. In other words, they do not apply to stacked/floating labels.

- [ ] Yes
- [x] No


core/src/components/range/range.ios.scss
```diff
@@ -28,11 +27,11 @@
 }
 
 ::slotted([slot="start"]) {
-  @include margin(0, var(--margin), 0, 0);
+  @include margin(0, 16px, 0, 0);
 }
 
 ::slotted([slot="end"]) {
-  @include margin(0, 0, 0, var(--margin));
```
liamdebeasi, line 35: I incorrectly assumed that the margins in the slots should be the same as the margin between the control and the label, so I reverted the `--margin` usage.

amandaejohnston (approved):
Full disclosure that I didn't look through all the screenshots; Github seems to have somehow gotten even worse at rendering large PRs like this 😅 In the future it may be useful to make the changes per component and run the screenshot job each time, so that the screenshot commits are broken up into smaller chunks.

PR #27449, opened 2023-05-10
test(input): merge render spec files

thetaPC:
Issue number: N/A

---------

Input had two spec files for the same behavior. The only difference was the file type: `.ts` and `.tsx`

## What is the new behavior?

- Merged the spec test into one file

- [ ] Yes
- [x] No

## Other information

N/A


liamdebeasi (approved):
Thank you!

PR #27457, opened 2023-05-11
feat(segment): display segment as a grid and add an ellipsis to overflowing text in a segment button 

### LINKED ISSUE #16532: bug: segment button does not truncate long text correctly
**Ionic Info**

Ionic-angular v4-beta.17

**Describe the Bug**
Ionic-Angular v4-beta.17 doesn't handle correctly long text in `ion-segment-button`. It seems that no max width applies and therefore that no ellipsis would be displayed.

**Steps to Reproduce**
1. Git clone https://github.com/peterpeterparker/ionic-segment-texxt
2. cd ionic-segment-texxt
3. nom install
4. ionic serve

**Related Code**
https://github.com/peterpeterparker/ionic-segment-texxt

**Expected Behavior**
As previously, I would expect that long text would be trimmed to the max width of the `ion-segment-button` and display with `ellipsis`

**Additional Context**

capture d ecran 2018-11-30 a 08 34 30

brandyscarney:
Issue number: resolves #16532

---------

Text that is too long to fit in a segment button does not ellipsis, instead it centers all of the text and cuts it off at the beginning and end of the text.

## What is the new behavior?
Text that is wider than the segment button will now add an ellipsis and cut the text off instead of horizontally centering the text and overflowing the button (while being cut off visually).

While researching how to fix this issue I discovered that the button text was not properly overflowing when it should due to a limitation of flex. I was able to mock segments using divs and see that certain buttons were adding an ellipsis when there was room to grow. This is due to a combination of using `flex-basis: 0` on the segment buttons and `width: auto` on the segment inside of a toolbar. This [blog](https://css-tricks.com/equal-columns-with-flexbox-its-more-complicated-than-you-might-think) sums it up well, but it is not something I could work around with segment set as `display: flex`. When I changed the mocked segment to use `display: grid`, it allowed the text to properly grow, while overflowing and adding an ellipsis when it couldn't grow. This can be seen in my [Codepen example](https://codepen.io/brandyscarney/pen/poOpbWE?editors=1100).

As a result, I made the following updates:

- Changed the `ion-segment` to [`display: grid`](https://github.com/ionic-team/ionic-framework/pull/27457/files#diff-dedcf5921daa49880ebae649e04d4f488a6b965c885a7bb1fdf29a5f1b3d501fR14) (`display: inline-grid` could not be used because the highlight was not properly aligned to the bottom of a toolbar for `md` mode) 
- Moves the `max-width` for Material Design segment buttons to the parent segment by using [`grid-auto-columns: minmax(auto, $segment-button-md-max-width);`](https://github.com/ionic-team/ionic-framework/pull/27457/files#diff-8df7c6681b616fdc975b068e3d31282cc4997222e786db6365ebcef3bccbb6d3R10). This is necessary for the buttons to properly center inside of the grid when they all hit the max-width (360px at the moment). 
  - The Material Design segment buttons will now take up equal widths. This matches the [MD2 spec for fixed segments](https://m2.material.io/components/tabs#fixed-tabs). 
- Sets [`grid-row: 1`](https://github.com/ionic-team/ionic-framework/pull/27457/files#diff-761a18ae6f41275e0eb63e9710045cafd5b221721ef1dad1d46d562e50404615R75) on the host segment button. This tells the segment buttons to stay on the same row. 
- Sets [`max-width: 100%`](https://github.com/ionic-team/ionic-framework/pull/27457/files#diff-761a18ae6f41275e0eb63e9710045cafd5b221721ef1dad1d46d562e50404615R214) and [`overflow: hidden`](https://github.com/ionic-team/ionic-framework/pull/27457/files#diff-761a18ae6f41275e0eb63e9710045cafd5b221721ef1dad1d46d562e50404615R222) on the native button. This allows the text itself to ellipsis.
- Added tests for segment wrapping & went through all existing tests to add missing `ion-label` elements in the segment buttons

- [ ] Yes
- [x] No

### High-level changes between main and this branch

| `main` | `FW-3401` |
| ---| ---|
| main-ios | branch-ios |
| main-md | branch-md |


core/src/components/segment-button/segment-button.ios.scss
```diff
@@ -22,9 +22,9 @@
   --indicator-transform: none;
   --transition: #{$segment-button-ios-transition};
   --padding-top: 0;
-  --padding-end: 13px;
+  --padding-end: 7px;
   --padding-bottom: 0;
-  --padding-start: 13px;
+  --padding-start: 7px;
```
amandaejohnston, line 27: Does changing default values for CSS variables count as a breaking change? 🤔

core/src/components/segment/test/basic/segment.e2e.ts-snapshots/segment-color-ios-ltr-Mobile-Chrome-linux.png
amandaejohnston, line 1: It looks like the text is no longer vertically centered inside the buttons in `ios` mode; it's been shifted up slightly. `md` mode looks good though.

core/src/components/segment/test/wrap/segment.e2e.ts
```diff
@@ -0,0 +1,149 @@
+import { expect } from '@playwright/test';
+import { configs, test } from '@utils/test/playwright';
+
+configs().forEach(({ title, screenshot, config }) => {
+  test.describe(title('segment: wrap'), () => {
+    test('should not wrap short text', async ({ page }) => {
```
amandaejohnston, line 6: I don't think the short text tests (both this and the toolbar one) are needed; the other segment tests already cover that.

core/src/components/segment/test/wrap/segment.e2e.ts
```diff
@@ -0,0 +1,149 @@
+import { expect } from '@playwright/test';
+import { configs, test } from '@utils/test/playwright';
+
+configs().forEach(({ title, screenshot, config }) => {
```
amandaejohnston, line 4: Do these tests need RTL? If we're just checking whether the text wraps or not, I don't believe LTR/RTL affects that.

amandaejohnston (changes requested):

core/src/components/segment-button/segment-button.ios.scss
brandyscarney: In the past we haven't considered it to be a breaking change since we update the colors for the spec all of the time. In this case it would probably be more of a breaking change for them if we don't update these. The reason for the update is because the text is now completely surrounded by the padding where it wasn't before, see this image:

image

core/src/components/segment/segment.scss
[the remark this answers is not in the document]
brandyscarney: So I've been looking into this and it appears that _a fix_ would be to change the styles on `ion-segment` for the grid columns:

```diff
:host {
-   grid-auto-columns: 1fr;
+   grid-auto-columns: minmax(auto, $segment-button-md-max-width);
}
```

This would solve the problem of them not being centered, but then we would be moving the style that sets the `max-width` from the `ion-segment-button` to the `ion-segment` which could be unexpected for users.

I could remove the `max-width` entirely which would make the buttons take up the entire width, but the spec does mention [360px should be the maximum width](https://m2.material.io/components/tabs#specs) and I am not sure if this would be a breaking change.

There is mention of centered tabs here as well but they don't seem to be hitting the max-width:

Screenshot 2023-05-16 at 10 32 02 AM

I can look into this a bit more to see if I can find a better solution for the centering issue.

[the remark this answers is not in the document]
amandaejohnston: Ah okay, I missed the fixed tabs idea. Sounds like `main` is the correct styling in that case. While the `minmax` style does look like it works perfectly, I agree that moving the `max-width` responsibility to `ion-segment` could cause problems 🤔 Since segment-button is currently setting `max-width` on the host, devs might be overwriting it, and those styles would break.

One idea I had was to introduce a CSS variable on `ion-segment` that sets the max width of the buttons, since that would at least allow devs to adjust it without having to overwrite the whole `grid-auto-columns` rule, although that doesn't avoid the breaking change issue I mentioned. Hmm...

core/src/components/segment/test/wrap/segment.e2e.ts
amandaejohnston: I would say it's still not needed since we're just testing browser behavior at that point.

brandyscarney: Removed: https://github.com/ionic-team/ionic-framework/pull/27457/commits/b5c0e337d2144a24d9b231fd0cad8ad6229a5b69

core/src/components/segment/segment.scss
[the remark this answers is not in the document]
brandyscarney: This sent me down a rabbit hole trying to make it work with grid, and then I went back and looked at why it didn't work with flex in the first place and [I might have found a solution with flex](https://codepen.io/brandyscarney/pen/MWPPprj?editors=1100). 😂 I think I am going to look into this more tomorrow, because that would save us introducing a breaking change. I just have to see if this fix would work inside of segment like it does in my mocked up segment.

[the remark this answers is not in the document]
brandyscarney: Okay so we had some further discussion about this in [this Slack thread](https://ionicteam.slack.com/archives/C02U5GFFXED/p1684512718177379) and we decided the best solution was to add the following CSS to the Material Design segment:

```css
:host {
  grid-auto-columns: minmax(auto, $segment-button-md-max-width);
}
``` 

This will remove it from being set on the segment button so we are going to add documentation on this and target this for the next feature release instead.

core/src/components/segment-button/segment-button.ios.scss
brandyscarney: I am actually going to revert this padding change because this doesn't actually get us closer to native, it is the opposite.

core/src/components/segment/test/basic/segment.e2e.ts-snapshots/segment-color-ios-ltr-Mobile-Chrome-linux.png
brandyscarney: This is another case where I am unable to reproduce this locally. Below is a screenshot from this branch at 50% on top of a screenshot from main:

ios-segment-diff

brandyscarney:
Changed the base branch to `feature-7.1`, it needs to be updated from main

brandyscarney:
Separated the screenshots out to this PR: https://github.com/ionic-team/ionic-framework/pull/27531/files

Made this PR against feature-7.1. Requesting a re-review on this please!

amandaejohnston (approved):
Great work!

liamdebeasi (commented):
Only noticing one (possibly) unintentional change (See below). Otherwise this is looking really great.

core/src/components/segment/test/basic/index.html
```diff
@@ -23,53 +23,91 @@
 
         <ion-toolbar>
           <ion-segment class="event-tester" value="Free">
-            <ion-segment-button value="Paid"> PaidPaidPaid </ion-segment-button>
```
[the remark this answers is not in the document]
brandyscarney: Ah nope. That was a result of my fix for scrollable segments. 😂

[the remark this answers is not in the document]
brandyscarney: Should be fixed by https://github.com/ionic-team/ionic-framework/pull/27457/commits/d86401b76eb654d89aa9e28eebeee64fa080bc13

liamdebeasi (approved):
Great job!

PR #27458, opened 2023-05-11
fix(picker-column): prevSelected is set to the correct value

### LINKED ISSUE #21764: bug: picker column prevSelect is always the same as selectedIndex
# Bug Report

**Ionic version:**

[ ] **4.x**
[x] **5.x**

**Current behavior:**

ionPickerColChange picker event handler receives change details and its property prevSelected value is always the same as selectedIndex prop value;

**Expected behavior:**

I assume that prevSelected should reflect the previously selected index and selectedIndex should reflect the currently selected index and those should be different if there has been a changed made.

**Steps to reproduce:**
1) open the link to stackblitz (see below)
2) click/tap OPEN PICKER button
3) change the value in the first column
Note that in the console the prevSelected shown is the same as selectedIndex

**Related code:**
See https://stackblitz.com/edit/ionic-angular-v5-dbgcak?file=src%2Fapp%2Fapp.component.ts

**Ionic info:**
```
Ionic:

   Ionic CLI                     : 6.9.2 (/Users/alexryltsov/.nvm/versions/node/v12.16.3/lib/node_modules/@ionic/cli)
   Ionic Framework               : @ionic/angular 5.1.1
   @angular-devkit/build-angular : 0.901.7
   @angular-devkit/schematics    : 9.1.7
   @angular/cli                  : 9.1.7

Capacitor:

   Capacitor CLI   : 2.1.1
   @capacitor/core : 2.1.1

Utility:

   cordova-res : not installed
   native-run  : not installed

System:

   NodeJS : v12.16.3 (/Users/alexryltsov/.nvm/versions/node/v12.16.3/bin/node)
   npm    : 6.14.5
   OS     : macOS Catalina
```

thetaPC:
Issue number: resolves #21764 

---------

The `prevSelected` returns the same value as `selectedIndex`.

When a user has selected the first option then the second:
- `prevSelected` returns `1`
- `selectedIndex` returns `1`

## What is the new behavior?

The `prevSelected` returns the index of the last selected option.

When a user has selected the first option then the second:
- `prevSelected` returns `0`
- `selectedIndex` returns `1`

- [ ] Yes
- [x] No

## Other information

Read the comment below regarding `toHaveReceivedEventDetail`.


core/src/components/picker-column/test/standalone/picker-column.e2e.ts
```diff
@@ -16,5 +18,29 @@ configs({ modes: ['ios'], directions: ['ltr'] }).forEach(({ title, config }) =>
...
+      await page.click('#single-column-button');
+
+      await ionPickerDidPresent.next();
+
+      secondOption.click();
+
+      await ionPickerColChangeEvent.next();
+
+      expect(ionPickerColChangeEvent.events[0].detail.prevSelected).toBe(0);
+      expect(ionPickerColChangeEvent.events[0].detail.selectedIndex).toBe(1);
```
thetaPC, line 43: Decided to not use `toHaveReceivedEventDetail` because it returns the entire `col` object. The test only needs to test the `prevSelected` and `selectedIndex` value.

core/src/components/picker-column/test/standalone/picker-column.e2e.ts
```diff
@@ -16,5 +18,29 @@ configs({ modes: ['ios'], directions: ['ltr'] }).forEach(({ title, config }) =>
...
+      const ionPickerColChangeEvent = await page.spyOnEvent('ionPickerColChange');
+
+      const column = page.locator('ion-picker-column');
+      const secondOption = column.locator('.picker-opt').nth(1);
+
+      await page.click('#single-column-button');
+
+      await ionPickerDidPresent.next();
+
+      secondOption.click();
```
liamdebeasi, line 38:
```suggestion
      await secondOption.click();
```

`click` is asynchronous

liamdebeasi (approved):
Great job! One test change below, but otherwise the code looks good.

Also small nitpick: Can we change "returns" to "is set to" in the PR title? IMO "returns" implies `prevSelected` is a function.

PR #27459, opened 2023-05-11
fix(input, select, textarea): display --border-radius correctly on start side with outline and non-floating label

### LINKED ISSUE #27116: bug: --border-radius custom css style is not applied to left hand corners of controls
### Prerequisites

### Ionic Framework Version

v7.x

### Current Behavior

Setting `--border-radius:30px;` in vue.js scoped css  does not apply the full radius to the left hand side of ion-input, ion-select, and ion-textarea. Resulting in the following shapes (respectively):
image

### Expected Behavior

Setting `--border-radius:30px;` in the css should render an input similar to this image:
image

### Steps to Reproduce

Start a new blank ionic-vue project, and replace the content of `src/views/HomePage.vue` with the following content:
```
<template>
  <ion-page>
    <ion-content :fullscreen="true">
      <ion-grid style="max-width: 500px;">
        <ion-row>
          <ion-col style="padding: 20px;">
            <ion-input aria-label="" class="custom-input" fill="outline"></ion-input> 
          </ion-col>
        </ion-row>
        <ion-row>
          <ion-col style="padding: 20px;">
            <ion-select class="custom-select" aria-label="" fill="outline">
              <ion-select-option v-for="[value, text] in choices" :value="value" :key="value">{{ text }}</ion-select-option>
            </ion-select> 
          </ion-col>
        </ion-row>
        <ion-row>
          <ion-col style="padding: 20px;">
            <ion-textarea class="custom-textarea" aria-label="" fill="outline"></ion-textarea> 
          </ion-col>
        </ion-row>
      </ion-grid>
    </ion-content>
  </ion-page>
</template>

<script setup lang="ts">
import { IonPage, IonContent, IonInput, IonGrid, IonRow, IonCol, IonSelect, IonSelectOption, IonTextarea } from '@ionic/vue';

const choices = new Map([
  ['1', 'Choice 1'],
  ['2', 'Choice 2'],
  ['3', 'Choice 3'],
  ['4', 'Choice 4'],
]);

</script>

<style scoped>
  ion-input.custom-input{
    --border-radius: 30px; /* does not apply to top left and bottom left corners */
    --border-width: 2px;
  }

  ion-select.custom-select{
    --border-radius: 30px; /* does not apply to top left and bottom left corners */
    --border-width: 2px;
  }

  ion-textarea.custom-textarea{
    --border-radius: 30px; /* does not apply to top left and bottom left corners */
    --border-width: 2px;
  }

</style>
```
Run the Created project.

### Code Reproduction URL

https://github.com/mp035/ionic_2023_04_05

### Ionic Info

[```
WARN] Error loading @capacitor/ios package.json: Error: Cannot find module '@capacitor/ios/package'
       
       Require stack:
       - /home/mark/.nvm/versions/node/v16.19.0/lib/node_modules/@ionic/cli/lib/project/index.js
       - /home/mark/.nvm/versions/node/v16.19.0/lib/node_modules/@ionic/cli/lib/index.js
       - /home/mark/.nvm/versions/node/v16.19.0/lib/node_modules/@ionic/cli/index.js
       - /home/mark/.nvm/versions/node/v16.19.0/lib/node_modules/@ionic/cli/bin/ionic
[WARN] Error loading @capacitor/android package.json: Error: Cannot find module '@capacitor/android/package'
       
       Require stack:
       - /home/mark/.nvm/versions/node/v16.19.0/lib/node_modules/@ionic/cli/lib/project/index.js
       - /home/mark/.nvm/versions/node/v16.19.0/lib/node_modules/@ionic/cli/lib/index.js
       - /home/mark/.nvm/versions/node/v16.19.0/lib/node_modules/@ionic/cli/index.js
       - /home/mark/.nvm/versions/node/v16.19.0/lib/node_modules/@ionic/cli/bin/ionic

Ionic:

   Ionic CLI       : 7.0.0 (/home/mark/.nvm/versions/node/v16.19.0/lib/node_modules/@ionic/cli)
   Ionic Framework : @ionic/vue 7.0.0

Capacitor:

   Capacitor CLI      : 4.7.3
   @capacitor/core    : 4.7.3

Utility:

   native-run  : 1.7.2

System:

   NodeJS : v16.19.0 (/home/mark/.nvm/versions/node/v16.19.0/bin/node)
   npm    : 8.19.3
   OS     : Linux 5.10

```

### Additional Information

_No response_

### PR DESCRIPTION
**Pull request by amandaejohnston**
Issue number: resolves #27116

---------

With `ion-input`, `ion-select`, or `ion-textarea` using `fill="outline"`, applying a large value for the `--border-radius` CSS prop causes the radius to render differently on the left and right sides:

uneven radii

This is due to the structure of the outline container, which is used to render the empty space in the border when a `stacked` or `floating` label floats upwards. The container has three pieces, which combine to form a seamless-looking border:

container pieces

The issue is that the browser will only render a radius that fits within the element's dimensions, and the piece on the left side (or right on RTL) is not wide enough.

## What is the new behavior?

The left-side piece was made wider, but only for non-stacked/floating labels. This is because doing so pushes the middle piece -- the notch spacer -- off to the side, which breaks the visuals when the label floats upwards:

broken outline

If devs wish to use a large border radius with a stacked/floating label, they should either increase `--padding-start` to be at least as high as the `--border-radius`, or use `shape="round"` (which will change both the padding and border radius automatically). This has been documented in the `--border-radius` CSS variable description, although I can move it elsewhere if desired.

- [ ] Yes
- [x] No

## Other information

This update triggered a few other screenshot differences. This is due to the browser rendering a tiny gap in the border where the start and middle pieces of the outline container meet:

gap in border

This gap has moved to the middle of the control. I consider this acceptable since the gap is hardly perceptible in the first place; you really need to zoom in to see it.

### CODE REVIEW

File: core/src/components/input/input.md.outline.scss [hunk 1]
```diff
@@ -189,6 +192,19 @@
...
+/**
+ * If a large --border-radius is applied through CSS, the
+ * .input-outline-start element will not be wide enough with
+ * default padding, causing the left and right sides of the
+ * border to show different radii. Increasing the width of
+ * this element fixes the issue, but this can't be done with
+ * stacked/floating labels because it pushes the notch spacer
+ * to the side, breaking the visuals when the label floats up.
+ */
+:host(.input-fill-outline:not(.input-label-placement-stacked):not(.input-label-placement-floating)) .input-outline-start {
```
liamdebeasi, line 204: Out of curiosity, why was having developers use `--padding-start` not used as a solution? For example, developers customizing border radius already have to do this:

```css
ion-input.custom-border-radius-input {
  --border-radius: 32px;
}
```

Since they are already in the styles, it doesn't seem like much additional effort to do the following:

```css
ion-input.custom-border-radius-input {
  --border-radius: 32px;
  --padding-start: 32px;
}
```

The benefit of this is that it keeps the solution consistent regardless of `fill`. The proposed PR would make it so developers would need to understand implementation details of `fill` in order to know if they need `--padding-start` to make the border radius work as expected.

Additionally, very specific selectors like the one here can make it challenging for users to override since they need to beat the specificity of this selector. Components like `ion-input` use scoped styles which already presents some challenges for customizing re: specificity: https://ionicframework.com/docs/api/input#css-custom-properties

Demo: https://codepen.io/liamdebeasi/pen/abRKywR

------

I took a look at other libraries that implement the notch effect, and the results are similar to the defect ours has.

Material Design Components for the Web: https://material-components.github.io/material-components-web-catalog/#/component/text-field

image

MUI: https://mui.com/material-ui/react-text-field/

image

MUI avoids part of this issue by using a [fieldset](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/fieldset), but the padding of the floating label is still incorrect. Notice how it gets cut off at an awkward angle: 
image

amandaejohnston: I'm okay with just having people do that, honestly. I figured it we can fix it for some cases, it couldn't hurt, but you make valid points. I'd like to keep these commits in case we need the fix in the future; does closing this PR and making a new one with the fix added to the `--border-radius` descriptions sound good? (Or do you think it's better placed somewhere else?)

liamdebeasi: Yeah I think that makes sense 👍

amandaejohnston:
Closing this per the discussion in https://github.com/ionic-team/ionic-framework/pull/27459#discussion_r1192355815. Will open a new PR soon.

Issue #16294, opened 2018-11-11
Datetime flips HH:mm to mm:HH in RTL


Newbie012:
# Bug Report

**Ionic Info**
Run `ionic info` from a terminal/cmd prompt and paste the output below.

```
Ionic:

   ionic (Ionic CLI)             : 4.3.1 (C:\Users\Eliya\AppData\Roaming\npm\node_modules\ionic)
   Ionic Framework               : @ionic/angular 4.0.0-beta.15
   @angular-devkit/build-angular : 0.8.7
   @angular-devkit/schematics    : 0.8.7
   @angular/cli                  : 6.2.7
   @ionic/angular-toolkit        : 1.1.0

Cordova:

   cordova (Cordova CLI) : 8.1.2 (cordova-lib@8.1.1)
   Cordova Platforms     : browser 5.0.4
   Cordova Plugins       : cordova-plugin-ionic-keyboard 2.1.3, cordova-plugin-ionic-webview 2.2.1, (and 5 other plugins)

System:

   Android SDK Tools : 26.1.1 (E:\android-sdk)
   NodeJS            : v10.11.0 (C:\Program Files\nodejs\node.exe)
   npm               : 6.4.1
   OS                : Windows 10
```

**Describe the Bug**
Using:

```
<html dir="rtl">
...
<ion-datetime display-format="HH:mm"></ion-datetime>
```

Will flip the numbers from HH:mm to mm:HH. This shouldn't be happen even on RTL languages.

**Steps to Reproduce**
1. add `[dir="rtl"]` to `<html>`
2. add `<ion-datetime display-format="HH:mm"></ion-datetime>`
4. See occurence

**Related Code**
Showed above

**Expected Behavior**
HH:mm should stick as HH:mm even on RTL.

**Additional Context**

![image](https://user-images.githubusercontent.com/10504365/48317455-3eebe880-e5fb-11e8-9170-6d0e42db7a86.png)

paulstelzer:
Thanks for your issue! I saw you created different issues for RTL issues. There are many more issues related to RTL, so it will take some time to fix all!

mohamedsharaf:
I agree date default reading is ltr 

But i think for global make right to left as the default for both date and time and who need ltr add direction to the element to force ltr

PR #27460, opened 2023-05-11
fix(picker-column): correct RTL direction

### LINKED ISSUE #21205: bug: Ionic 5 ion-picker-column position internationalization when dir="rtl"
# Bug Report

**Ionic version:** 5

**Current behavior:** 
ion-picker-column days' component is expicitly set to float on the right (start for ltr languages)
and years' on the left (end of ltr languages)

**Expected behavior:**
when document's dir="rtl" days' columns should be on the right/start and years' on the left/end

**Steps to reproduce:**
go to https://ionicframework.com/docs/api/picker
set the example's iframe's document's html-element's dir attribute to `rtl`

**Related code:**
ion-picker-column is absolutely positioned with class names that use "right" and "left" that use `justify-content: flex-end;` and `justify-content: flex-start;` respectively.

**Other information:**

**Ionic info:**

not sure exactly, certainly 5+, it's the ionic docs site's https://ionicframework.com/docs/api/picker you can know that.

thetaPC:
Issue number: resolves #21205 

---------

`picker-column` keeps using LTR display when using RTL.

LTR:
`Column 1   Column 2`

RTL:
`Column 1   Column 2`

## What is the new behavior?

`picker-column` correctly displays for RTL.

LTR:
`Column 1   Column 2`

RTL:
`Column 2   Column 1`

- [ ] Yes
- [x] No

## Other information

This PR is already being tested through the `picker -> basic` and `picker -> value` screenshots. Due to those tests, a new test is not needed to verify the fix.

The current fix is only changing the `picker` and `picker-column`. The `datetime` and `datetime-button` do not change, leaving the format up to the users. It wouldn't change the functionality that was provided through PR ionic-team/ionic-framework#17018.


core/src/components/picker-column/test/standalone/picker-column.e2e.ts
```diff
@@ -18,3 +18,27 @@ configs({ modes: ['ios'], directions: ['ltr'] }).forEach(({ title, config }) =>
     });
   });
 });
+
+/**
+ * ion-picker-column does not have mode-specific styling for document direction
+ */
+configs({ modes: ['ios'] }).forEach(({ title, config, screenshot }) => {
+  test.describe(title('picker-column: document direction'), () => {
```
liamdebeasi, line 26: It looks like we have RTL screenshots in `picker/test/basic` (the screenshots that got updated). What does this test do differently?

thetaPC: Not really, I was on the fence of whether it would be beneficial. I figured that it would be rejected during a review if that was the case. I'll remove it.

liamdebeasi (approved):
Looks good to me. I'm going to have Brandy take a look too in case she knows of any "gotchas" from when `direction: ltr` was originally applied.

brandyscarney:
Adding `dir: ltr` was intentionally done as part of our main RTL issue https://github.com/ionic-team/ionic-framework/issues/17012 

The specific issue related to this is https://github.com/ionic-team/ionic-framework/issues/16294 and the PR was https://github.com/ionic-team/ionic-framework/pull/17018

We might need to do some more investigation into this as this was [feedback from our RTL users](https://github.com/ionic-team/ionic-framework/issues/16294#issuecomment-452715300). Maybe we should make sure we cover their use cases or have documentation for how to achieve it before making this change.

thetaPC:
> Adding `dir: ltr` was intentionally done as part of our main RTL issue #17012
> 
> The specific issue related to this is #16294 and the PR was #17018
> 
> We might need to do some more investigation into this as this was [feedback from our RTL users](https://github.com/ionic-team/ionic-framework/issues/16294#issuecomment-452715300). Maybe we should make sure we cover their use cases or have documentation for how to achieve it before making this change.

The current fix is only changing the `picker` and `picker-column`. The `datetime` and `datetime-button` do not change, leaving the format up to the users. It wouldn't change the functionality that was provided through PR ionic-team/ionic-framework#17018.

PR #27468, opened 2023-05-12
3532 base

liamdebeasi:
Issue number: #

---------

## What is the new behavior?

-
-
-

- [ ] Yes
- [ ] No


core/src/components/select/select.tsx
```diff
@@ -649,6 +656,30 @@ export class Select implements ComponentInterface {
...
+    const { el, label } = this;
+
+    if (label !== undefined) {
+      return label;
+    }
+
+    const labelSlot = el.querySelector('[slot="label"]');
+
+    if (labelSlot !== null) {
+      return labelSlot.textContent;
```
amandaejohnston, line 677: Passing this as-is to the alert header gives odd results when using a more complicated label structure. For example, opening this:
```html
<ion-select placeholder="Select">
  <div slot="label">
    <div>Select a Fruit</div>
    <ion-note>This will be remembered.</ion-note>
  </div>
  <ion-select-option value="apple">Apple</ion-select-option>
</ion-select>
```
Leads to this:

odd header

While I can't think of a good way to have Ionic solve this automatically, I noticed you can overwrite it manually using `interfaceOptions`, and that still works:

```html
<ion-select placeholder="Select" id="weird-label">
  <div slot="label">
    <div>Select a Fruit</div>
    <ion-note>This will be remembered.</ion-note>
  </div>
  <ion-select-option value="apple">Apple</ion-select-option>
</ion-select>

<script>
  const select = document.querySelector('#weird-label');
  select.interfaceOptions = {
    header: 'Select a Fruit'
  };
</script>
```

Maybe we'll just want to document this pattern and leave it as-is?

core/src/components/select/test/label/select.e2e.ts
amandaejohnston, line 1: Could we also add a quick screenshot test to ensure the label slot renders correctly? I don't think we need to check all the combinations of label placements, but since it has its own styling, a check with the default options would be useful.

liamdebeasi: Do you think we need a screenshot, or would a `toBeVisible` assertion suffice?

amandaejohnston: I would prefer a screenshot to make sure the basic styling is applied correctly, since we had to edit one of the selectors to do so.

liamdebeasi: Added in https://github.com/ionic-team/ionic-framework/pull/27468/commits/3d1eda86d6b6958f248f55419ce88e837645c3f3. Screenshots are updating now.

amandaejohnston (approved):
Thank you!

PR #24510, opened 2022-01-05
fix(react): scrolling to bottom of modal contents

sean-perkins:
## Pull request checklist

Please check if your PR fulfills the following requirements:
- [ ] Tests for the changes have been added (for bug fixes / features)
- [ ] Docs have been reviewed and added / updated if needed (for bug fixes / features)
- [x] Build (`npm run build`) was run locally and any changes were pushed
- [x] Lint (`npm run lint`) has passed locally and any fixes were made for failures

## Pull request type

 

 

Please check the type of change your PR introduces:
- [x] Bugfix
- [ ] Feature
- [ ] Code style update (formatting, renaming)
- [ ] Refactoring (no functional changes, no api changes)
- [ ] Build related changes
- [ ] Documentation content changes
- [ ] Other (please describe): 

## What is the current behavior?

Users are unable to scroll to the bottom of modals that have overflow.

Issue Number: #24478

## What is the new behavior?

The wrapper for inline overlays will fill its height based on the contents.

## Does this introduce a breaking change?

- [ ] Yes
- [x] No

PR #27481, opened 2023-05-15
fix(react, vue): inline modals apply ion-page class

### LINKED ISSUE #27470: bug: List in inline modal last item is cut off
### Prerequisites

### Ionic Framework Version

v7.x

### Current Behavior

I have attempted to create a modal with an `ion-list` inside it in Vue 3.X (latest), it is supposed to show all the way to the bottom, but the content of the modal does not scroll to the last item, it is instead cut off. I have seen the other issues that were supposedly resolved, but perhaps the bug has resurfaced.

### Expected Behavior

Modal scrolls the content

### Steps to Reproduce

1. Create a modal with a toolbar
2. Create enough list items to overflow the parent
3. Experience bug

### Code Reproduction URL

https://stackblitz.com/edit/aawkvs?file=src/components/Example.vue

### Ionic Info

Ionic:

   Ionic CLI       : 7.0.1 (/usr/local/lib/node_modules/@ionic/cli)
   Ionic Framework : @ionic/vue 7.0.6

Capacitor:

   Capacitor CLI      : 5.0.3
   @capacitor/android : 5.0.3
   @capacitor/core    : 5.0.3
   @capacitor/ios     : 5.0.3

Utility:

   native-run  : 1.7.2

System:

   NodeJS : v20.1.0 (/usr/local/Cellar/node/20.1.0/bin/node)
   npm    : 9.6.4
   OS     : macOS Unknown

### Additional Information

Code reproduction is based on the example online from docs with additional elements copy + pasted

liamdebeasi:
Issue number: resolves #27470

---------

Passing multiple elements in to an inline modal causes `.ion-page` to not get set. This causes content to get pushed off the bottom of the modal equal to the height of the header. React has some special CSS that prevents this: https://github.com/ionic-team/ionic-framework/blob/eb2772c0ce623de70cce42b5d36e86cdbba7bafb/packages/react/src/components/createInlineOverlayComponent.tsx#L137-L140

However, I think this should be delegated to `.ion-page` instead so the behavior is consistent across frameworks. For example, Angular uses `.ion-page`: https://github.com/ionic-team/ionic-framework/blob/eb2772c0ce623de70cce42b5d36e86cdbba7bafb/angular/src/directives/overlays/modal.ts#L82 

## What is the new behavior?

- Inline overlays in Ionic React and Ionic Vue wrap child content in `.ion-delegate-host.ion-page`.
- Removed the custom flex styles from Ionic React as `.ion-page` has its own  styles.

- [ ] Yes
- [x] No

## Other information

Revised Design Doc: https://github.com/ionic-team/ionic-framework-design-documents/pull/84


packages/react/src/components/createInlineOverlayComponent.tsx
```diff
@@ -132,14 +144,8 @@ export const createInlineOverlayComponent = <PropType, ElementType>(
             ? createElement(
                 'div',
                 {
-                  id: 'ion-react-wrapper',
                   ref: this.wrapperRef,
```
liamdebeasi, line 147: All overlays in React still need a wrapper because of how we handle teleporting in react

packages/react/src/components/createInlineOverlayComponent.tsx
```diff
@@ -132,14 +144,8 @@ export const createInlineOverlayComponent = <PropType, ElementType>(
...
-                  className: 'ion-delegate-host',
-                  style: {
-                    display: 'flex',
-                    flexDirection: 'column',
-                    height: '100%',
```
liamdebeasi, line 141: These styles were only needed for `ion-modal` and are now accomplished by the `.ion-page` class. See https://github.com/ionic-team/ionic-framework/pull/24510

packages/vue/src/vue-component-lib/overlays.ts
```diff
@@ -162,10 +162,22 @@ export const defineOverlayContainer = <Props extends object>(name: string, defin
           }
         }
 
+        /**
+         * `ion-modal` needs a wrapper element so content
+         * takes up the full size of the parent modal.
+         */
+        const renderChildren = () => {
+          if (name === 'ion-modal') {
```
sean-perkins, line 170: ~ I'm not personally a huge fan of coupling component specific checks in the shared utility for all overlays. 

Would you be open to refactor this to be an option of `defineOverlayContainer` and update the manual wrappers to set the option: https://github.com/ionic-team/ionic-framework/blob/a1e3e4ef04226fdd661689ce6f0e8064d8ed28d2/packages/vue/src/components/Overlays.ts#L30?

liamdebeasi: I'm open to changing the approach, though I don't follow the proposed alternative. Can you clarify?

sean-perkins: For sure!

Updating the implementation to:

```ts
export const defineOverlayContainer = <Props extends object>(name: string, defineCustomElement: () => void, componentProps: string[] = [], controller?: any, delegateHost: boolean) => {
  ...
  const renderChildren = () => {
    if(delegateHost) {
      return h('div', { className: 'ion-delegate-host ion-page' }, slots)
    }
    return slots;
  }
}
```

Making the usage for the overlay wrappers:
```ts
export const IonModal = /*@__PURE__*/ defineOverlayContainer<JSX.IonModal>('ion-modal', defineIonModalCustomElement, ['props'], true);
    
```

This may be an advantageous time to refactor `defineOverlayContainer` to take an object of configurations, vs. many different arguments for configuration. 

e.g.:
```ts
defineOverlayContainer<JSX.IonModal>({
  tag: 'ion-modal',
  defineCustomElement: defineIonModalCustomElement,
  props: [],
  hasDelegateHost: true
});
```

but that could also be a separate discussion/work-effort, personally many arguments can be hard to reason about.

The whole goal here of introducing a new argument, is that the responsibility of rendering the `ion-delegate-host ion-page` container is independent of the tag name and can be enabled for any overlay that needs it.

liamdebeasi: Ok I made the changes. Lmk your thoughts.

amandaejohnston (approved):

sean-perkins (approved):
Looks good to me, great work!

PR #27483, opened 2023-05-15
fix(select): slotted label content works with outline notch

liamdebeasi:
Issue number: N/A

---------

The label slot content does not work with the notch behavior for `fill="outline"`. The notch cutout does not match the dimensions of the floating/stacked label.

## What is the new behavior?

- `ion-select` now computes the notch size manually if slotted label content is used.

Stencil re-renders the component when slotted content is added, removed, or modified, so we call `setNotchWidth` in `componentDidRender`. This ensures that the method is run if `labelPlacement` is changed too.

`setNotchWidth` has a couple optimizations built in to ensure we are not doing unnecessary computations:

1. If no label exists _or_ the label prop is in use then the function returns early. The notch for the label prop is computed based on its intrinsic size because we can render that text content in multiple places.
2. If the slotted label has a width of 0 but has an offsetParent then we return early to avoid adding an IntersectionObserver on content that is not hidden.

In the event that the select is hidden, then we use an IntersectionObserver to be notified of when the select is visible again. At that point, we will attempt to set the notch width again.

- [ ] Yes
- [x] No


core/src/components/select/select.tsx
```diff
@@ -665,13 +668,13 @@ export class Select implements ComponentInterface {
...
   private get labelText() {
-    const { el, label } = this;
+    const { label } = this;
 
     if (label !== undefined) {
       return label;
     }
 
-    const labelSlot = el.querySelector('[slot="label"]');
+    const labelSlot = this.labelSlot;
```
sean-perkins, line 677: ~ Should we match the code style above and do something similar to:

```ts
const { labelSlot } = this;
```

core/src/components/select/select.tsx
```diff
@@ -740,14 +743,140 @@ export class Select implements ComponentInterface {
...
   }
 
+  componentDidRender() {
+    /**
+     * Run this the frame after
+     * the browser has re-painted the select.
+     * Otherwise, the label element may have a width
+     * of 0 and the IntersectionObserver will be used.
+     */
+    raf(() => {
```
sean-perkins, line 753: Are there any performance considerations and/or optimizations we should handle here?

`componentDidRender` will call after each `render()` of the select, which can occur many times over the lifespan of the `ion-select` in the view. Each time that occurs we are invoking a `requestAnimationFrame` callback. 

Maybe a simpler way to ask this question - is there any conditions where we can opt out of calling  the `raf`/`setNotchWidth`? Such as if they are using legacy syntax, not using a visible label, etc. I understand the limitations with accessing the width in the current frame.

liamdebeasi: `setNotchWidth` returns early if neither the label prop nor the label slot is used. This causes `setNotchWidth` to not run on legacy selects and selects without a visible label. It also returns early if `this.label` is defined, causing it to only run on selects with a label slot.

We potentially need to re-check the slot width if the `labelPlacement` or `label` props change or if the slotted content changes. For example, if the `label` prop becomes defined then we want to remove the explicit `width` CSS property. Additionally, if the `labelPlacement` prop changes from `start` to `stacked` then we need to compute the notch width.

liamdebeasi: Oh I guess I could move the `raf` into `setNotchWidth` so it is subject to the same early return checks. Is that what you are referring to?

core/src/components/select/select.tsx
amandaejohnston, line 1: Unsure how much of this you want to address in this PR and how much is best saved for another, but I'm noticing the behavior is weird with a multi-line label and `label-placement="floating"`:
1. The placement of the label is too low when the select is closed, as well as open when not using `fill`.
2. When the select is opened, all lines of the label are displayed, while the design doc notes that only the first should be shown to avoid text overlapping. (I'm okay reverting this requirement if you can't find a way to make the animation smooth, but we'll want to document that floating labels of more than two lines should be avoided.)

Screencast and markup below (added to the `fill` test):

https://github.com/ionic-team/ionic-framework/assets/90629384/52420da2-7bbb-4fac-a09f-06f295403a52

```html
<div class="grid-item">
  <ion-select placeholder="Select one" fill="outline" label-placement="floating">
    <div slot="label">
      <div>Main label</div>
      <ion-note>Sub-label</ion-note>
    </div>
    <ion-select-option value="apple">Apple</ion-select-option>
  </ion-select>
</div>

<div class="grid-item">
  <ion-select placeholder="Select one" label-placement="floating">
    <div slot="label">
      <div>Main label</div>
      <ion-note>Sub-label</ion-note>
    </div>
    <ion-select-option value="apple">Apple</ion-select-option>
  </ion-select>
</div>

<div class="grid-item">
  <ion-select placeholder="Select one" fill="outline">
    <div slot="label">
      <div>Main label</div>
      <ion-note>Sub-label</ion-note>
    </div>
    <ion-select-option value="apple">Apple</ion-select-option>
  </ion-select>
</div>
```

core/src/components/select/test/fill/select.e2e.ts
```diff
@@ -167,3 +168,60 @@ configs({ modes: ['md'] }).forEach(({ title, screenshot, config }) => {
...
+      expect(await select.screenshot()).toMatchSnapshot(screenshot(`select-fill-outline-slotted-label`));
+    });
+    test('should render the notch correctly with a slotted label after the select was originally hidden', async ({
+      page,
+    }) => {
+      await page.setContent(
+        `
+        <style>
+          .custom-label {
+            font-size: 30px;
```
amandaejohnston, line 204: Setting this style on a floating label causes the label to render too low when the select is closed:

weird placement

Not sure if this is best addressed in another PR, though.

core/src/components/select/select.tsx
liamdebeasi: I don't think multi-line labels are appropriate for the floating/stacked. It's also worth noting that this behavior is consistent with the v6/legacy behavior too: https://codepen.io/liamdebeasi/pen/vYVVyYw

However, this is probably something we should align on in a design doc.

liamdebeasi: > When the select is opened, all lines of the label are displayed, while the design doc notes that only the first should be shown to avoid text overlapping. (I'm okay reverting this requirement if you can't find a way to make the animation smooth, but we'll want to document that floating labels of more than two lines should be avoided.)

This is going to be hard to do because the label could be made up of any number of elements. If it was a single element we could use [line-clamp](https://css-tricks.com/almanac/properties/l/line-clamp/), but we can't guarantee that only 1 element will be used.

core/src/components/select/test/fill/select.e2e.ts
liamdebeasi: This is another tricky one. The problem here is we are translating based on %, which will be based on the height of the label element. If the label element grows in height so will the translation.

MDC "fixes" this by giving the label a fixed height. However, this causes the text to get cut off at larger font sizes. This also causes the label to no longer be centered in the container as the select increases in height, though I'm not 100% if that's by design.

We could also make the label height 100%, but then we're going to need to translate by seemingly arbitrary values that may also be wrong as the select changes in height.

Any thoughts on how best to address? I'll keep testing it. It's worth noting that Material Design doesn't seem to do a lot of customizations on floating/stacked labels beyond colors and such. Perhaps this is by design?

amandaejohnston: I was able to get pretty close by changing the following styles on `.label-text-wrapper` (screenshots linked):
- [In the default styles](https://i.gyazo.com/e762efeb8f26a2701a89ea0779e548e5.png), replace the current `transform` with `top: 50%` and `transform: translateY(-50%) scale(1)`. The top: 50/translateY(-50) is a trick I've used to vertically center absolutely positioned items.
- [When floating](https://i.gyazo.com/749fc04809eb7a7f68f6ca9c743b2fea.png), add `top: auto`.

The animation doesn't work right, but maybe there's a good way to fix it? An alternative could be to not do `position: absolute` at all and use flex styles to vertically center the label instead, maybe only when the label isn't floating if the position is necessary to make the float work.

core/src/components/select/select.tsx
liamdebeasi: Updated https://github.com/ionic-team/ionic-framework-design-documents/pull/87

core/src/components/select/test/fill/select.e2e.ts
liamdebeasi: During refinement we discussed that this is the same problem as https://ionic-cloud.atlassian.net/browse/FW-4083, so we are going to handle this fix in a separate ticket.

amandaejohnston (approved):

core/src/components/select/select.tsx
liamdebeasi: Fixed in https://github.com/ionic-team/ionic-framework/pull/27483/commits/13a6547db14ec4d78c1b59c8f000014f5e0ebecb

core/src/components/select/select.tsx
sean-perkins: Yes, I was curious if the checks inside `setNotchWidth` are dependent on being called within `requestAnimationFrame` or can we opt to do the `raf` only after the early returns?

liamdebeasi: Good call. I optimized this in https://github.com/ionic-team/ionic-framework/pull/27483/commits/714f8c470505360fdbbefe6c2611aa6fef6e1e82.

(I removed the extra console.log in https://github.com/ionic-team/ionic-framework/pull/27483/commits/2339727c8e8871e5074a8036fe9265be505e75ef)

sean-perkins (approved):

PR #27507, opened 2023-05-18
test(select): clean up basic test template

amandaejohnston:
Issue number: N/A

---------

The `basic` test for `ion-select` contains a lot of content that is either duplicated within the test, duplicated by other tests, or simply not used/needed. Additionally, some select options (notably gender and hair color) are not inclusive.

## What is the new behavior?

Test template cleaned up, and options replaced with less personal ones as needed. Legacy test HTML was intentionally left alone since we'll be removing that in the near future anyway, though I did have to tweak a legacy E2E test since it was visiting the non-legacy page.

- [ ] Yes
- [x] No

### DIFF
--- a/core/src/components/select/test/basic/index.html
+++ b/core/src/components/select/test/basic/index.html
@@ -22,52 +22,65 @@
         </ion-toolbar>
       </ion-header>
 
-      <ion-content class="outer-content test-content">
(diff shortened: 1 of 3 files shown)


thetaPC (approved):
LGTM

PR #23146, opened 2021-04-05
fix(toggle): prevent click event on toggle from double firing

willmartindev:
## Pull request checklist

Please check if your PR fulfills the following requirements:
- [x] Tests for the changes have been added (for bug fixes / features)
- [x] Docs have been reviewed and added / updated if needed (for bug fixes / features)
- [x] Build (`npm run build`) was run locally and any changes were pushed
- [x] Lint (`npm run lint`) has passed locally and any fixes were made for failures

## Pull request type

 

 

Please check the type of change your PR introduces:
- [x] Bugfix
- [ ] Feature
- [ ] Code style update (formatting, renaming)
- [ ] Refactoring (no functional changes, no api changes)
- [ ] Build related changes
- [ ] Documentation content changes
- [ ] Other (please describe): 

## What is the current behavior?

Issue Number: resolves #23041 

## Does this introduce a breaking change?

- [ ] Yes
- [x] No

PR #27508, opened 2023-05-18
fix(checkbox, radio): click event is triggered once

### LINKED ISSUE #27506: bug: ion-checkbox fires onclick twice
### Prerequisites

### Ionic Framework Version

v7.x

### Current Behavior

If an ion-checkbox has an onClick listener - it will fire onClick twice

### Expected Behavior

Should only fire onClick once when you click the checkbox.

### Steps to Reproduce

1. Use an IonCheckbox with an onClick
2. notice that the onClick is called twice when you click once

### Code Reproduction URL

https://stackblitz.com/edit/8t9ymz?file=src%2Fmain.tsx

### Ionic Info

See stackblitz repro above - not using ionic CLI

### Additional Information

Maybe related - not sure: https://github.com/ionic-team/ionic-framework/issues/27492

sean-perkins:
Issue number: Resolves #27506

---------

Modern form controls (`ion-radio`, `ion-checkbox`) that associate a label to an input result in an additional click event, when the label is clicked.

e.g.:

```jsx
<ion-checkbox onClick={() => console.log('clicked')}>
  #shadow-root
  <label>
    <input type="checkbox" />
  </label>
</ion-checkbox>
```

When the host is clicked, the label is the click target. When a label is associated to an input, it will trigger a click event on the control/input. This causes the event to bubble and a duplicate click event to fire.

## What is the new behavior?

- Prevent the event bubbling from the inner control when the label is clicked for `ion-checkbox` and `ion-radio`
- Click events are emitted a single time for `ion-checkbox` and `ion-radio`

- [ ] Yes
- [x] No

## Other information

Dev-build: `7.0.8-dev.11684439870.1ade4229`


core/src/components/checkbox/checkbox.tsx
```diff
@@ -204,6 +204,14 @@ export class Checkbox implements ComponentInterface {
     this.ionBlur.emit();
   };
 
+  private onClick = (ev: MouseEvent) => {
+    // The modern control syntax renders a label as a parent element
+    // to the checkbox, which means when the label is clicked, the
+    // checkbox receives this event. To prevent a duplicate click event,
+    // we need to stop the event from bubbling.
+    ev.stopPropagation();
```
liamdebeasi, line 212: Does `pointer-events: none` on the native element work here too? We fixed a similar bug in toggle a couple years ago: https://github.com/ionic-team/ionic-framework/pull/23146

I'd prefer a CSS-only approach rather than adding even more JS.

sean-perkins: `pointer-events: none` on the inner `input` control continues to result in duplicate clicks. The problem is the label invokes a click event on the `input` and does not take into consideration the pointer events.

liamdebeasi (commented):
I'm fine with the changes, but do we know why this does not happen with `ion-toggle`?

sean-perkins:
Removed the tests that I originally introduced. This issue does not reproduce outside of a browser environment (an emulated environment does not result in 2x events). 

This issue is not present in `ion-toggle` because of this line: https://github.com/ionic-team/ionic-framework/blob/main/core/src/components/toggle/toggle.tsx#L255. Removing it results in 2x click events.

brandyscarney:
@liamdebeasi Was this PR supposed to be closed too?

liamdebeasi:
Whoops yes. Thanks!

PR #27514, opened 2023-05-19
fix(core): handle uncaught native keyboard exceptions

### LINKED ISSUE #27503: bug: Ionic call Keyboard.getResizeMode() on Android throws uncaught error
### Prerequisites

### Ionic Framework Version

v7.x

### Current Behavior

when app start in android the ```Keyboard.getResizeMode()``` call from one of the ionic global component that it catched and show  runtime error . I read docs of capacitor keyboard and we know that getResizeMode() method is for iOS only . why in android called on app init automatically ? 

This is my console error : 

```result Keyboard.getResizeMode (#71292250)```
```
{
    "message": "not implemented",
    "code": "UNIMPLEMENTED"
}
```

Also this is screen shots of android version

**it’s make my app unuseable**  with vue-cli in vite version the runtime error not show but console throw excetiopn still fire

Screenshot 2023-05-18 at 4 16 12 PM

And my android Emulator

Screenshot 2023-05-18 at 4 16 40 PM

### Expected Behavior

I think if ionic use  the ```Keyboard.getResizeMode()``` in any components it should check platform first for ignore calling on android and call in iOS only

### Steps to Reproduce

1. Check the global ionic components that using Keyboard.getResizeMode()
2. make sure the platform checked first and call Keyboard.getResizeMode() only in iOS
3. Maybe Catche exception in ```capacitor/keyboard``` plugin  for ignore runtime error even Keyboard.getResizeMode() called in any platform by 3rd Party plugins or components thats not official .

### Code Reproduction URL

GitHub starter tab app created with  latest ionic cli  and ionic vue and capacitor 5 ( But the bug is still with cap 4  too  and I think its fire from ionic )

https://github.com/farshidpg/ionic-starter/tree/main

**Starter app ionic  info** 

```
Ionic:

   Ionic CLI       : 7.1.1
   Ionic Framework : @ionic/vue 7.0.7

Capacitor:

   Capacitor CLI      : 5.0.3
   @capacitor/android : 5.0.3
   @capacitor/core    : 5.0.3

Utility:

   cordova-res : 0.15.4
   native-run  : 1.7.2

System:

   NodeJS : v19.6.1
   npm    : 9.5.0
   OS     : macOS Unknown

```

the problem still get error on app init 

### Ionic Info

``` Ionic:

   Ionic CLI       : 7.1.1
   Ionic Framework : @ionic/vue 7.0.7

Capacitor:

   Capacitor CLI      : 4.8.0
   @capacitor/android : 4.8.0
   @capacitor/core    : 4.8.0
   @capacitor/ios     : 4.8.0

Utility:

   cordova-res : 0.15.4
   native-run  : 1.7.2

System:

   NodeJS : v19.6.1
   npm    : 9.5.0
   OS     : macOS



Android SDK : 33
```

### Additional Information

Capacitor keyboard plugin information ```getresizemode``` method

https://capacitorjs.com/docs/v4/apis/keyboard#getresizemode

I check my project and I know  I don't use  ```Keyboard.getResizeMode()``` method manually , I'm sure it's called globally

sean-perkins:
Issue number: Resolves #27503

---------

Ionic Framework wraps the implementation around Capacitor's Keyboard plugin API, to provide additional functionality and behavior "automatically" in Ionic, when the plugin is installed.

Certain methods such as `getResizeMode()` are only available for certain platforms and will cause an error when running in the unsupported platform. 

Ionic Framework does not check to see if that platform is active before calling potentially unsupported methods, which leads to an exception for scenarios such as this - calling `getResizeMode()` on Android. 

## What is the new behavior?

- Handles the uncaught exception by returning `undefined` if the plugin method is unavailable.
- Developers do not receive an uncaught exception on Android when using the Capacitor Keyboard plugin with Ionic Framework

- [ ] Yes
- [x] No

## Other information

Dev-build: `7.0.8-dev.11684444351.1b1ab142` (outdated)

### DIFF
--- a/core/src/utils/native/keyboard.ts
+++ b/core/src/utils/native/keyboard.ts
@@ -20,7 +20,7 @@ export const Keyboard = {
   },
   getResizeMode(): Promise<KeyboardResizeOptions | undefined> {
     const engine = this.getEngine();
-    if (!engine || !isPlatform('ios') || !engine.getResizeMode) {
+    if (!isPlatform('ios') || !engine?.getResizeMode) {
       // getResizeMode is only available on iOS
       // see: https://ionicframework.com/docs/native/keyboard#getresizemode
       return Promise.resolve(undefined);
--- a/core/src/utils/native/keyboard.ts
+++ b/core/src/utils/native/keyboard.ts
@@ -1,3 +1,5 @@
+import { isPlatform } from '@utils/platform';
+
 import { win } from '../browser';
 
 // Interfaces source: https://capacitorjs.com/docs/apis/keyboard#interfaces
@@ -18,7 +20,9 @@ export const Keyboard = {
   },
   getResizeMode(): Promise<KeyboardResizeOptions | undefined> {
     const engine = this.getEngine();
-    if (!engine || !engine.getResizeMode) {
+    if (!engine || !isPlatform('ios') || !engine.getResizeMode) {
+      // getResizeMode is only available on iOS
+      // see: https://ionicframework.com/docs/native/keyboard#getresizemode
       return Promise.resolve(undefined);
     }
 
--- a/core/src/utils/native/keyboard.ts
+++ b/core/src/utils/native/keyboard.ts
@@ -1,5 +1,7 @@
 import { win } from '../browser';
 
+import type { NativePluginError } from './native-interface';
+
 // Interfaces source: https://capacitorjs.com/docs/apis/keyboard#interfaces
 export interface KeyboardResizeOptions {
   mode: KeyboardResize;
@@ -21,7 +23,7 @@ export const Keyboard = {
     if (!engine?.getResizeMode) {
       return Promise.resolve(undefined);
     }
-    return engine.getResizeMode().catch((e: any) => {
+    return engine.getResizeMode().catch((e: NativePluginError) => {
       if (e.code === 'UNIMPLEMENTED') {
         // If the native implementation is not available
         // we treat it the same as if the plugin is not available.
--- a/core/src/utils/native/keyboard.ts
+++ b/core/src/utils/native/keyboard.ts
@@ -1,5 +1,3 @@
-import { isPlatform } from '@utils/platform';
-
 import { win } from '../browser';
 
 // Interfaces source: https://capacitorjs.com/docs/apis/keyboard#interfaces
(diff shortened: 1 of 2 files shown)


thetaPC (approved):
LGTM

sean-perkins:
@liamdebeasi I'm going to add you as an additional reviewer since you wrote a lot of the code around this. Any concerns to this approach?

core/src/utils/native/keyboard.ts
```diff
@@ -18,7 +20,9 @@ export const Keyboard = {
   },
   getResizeMode(): Promise<KeyboardResizeOptions | undefined> {
     const engine = this.getEngine();
-    if (!engine || !engine.getResizeMode) {
+    if (!isPlatform('ios') || !engine?.getResizeMode) {
```
liamdebeasi, line 23: Are we able to check the `UNIMPLEMENTED` value from Capacitor instead of checking the platform?

So something like this:

```js
try {
  return engine.getResizeMode();
} catch(e) {
  if (e.code === 'UNIMPLEMENTED') {
    return Promise.resolve(undefined); 
  }
  
  // throw the error unchanged if not an unimplemented error
  throw e;
}
```

My concern with the proposed approach is we are coupling our code to implementation details of the Capacitor plugin. Additionally, `isPlatform` relies on user agent sniffing which can be spoofed, so that check may not always be correct. Checking the unimplemented state lets us rely on Capacitor to tell us about functionality.

sean-perkins: Ooh that's brilliant, I didn't think of that approach!

Yes, we can do something very similar to this. I ended up using a `.catch` on the original promise, so when it is chained in the implementation, it will resolve `undefined`. It'll swallow the error in the client runtime, but still indicate the unimplemented method invocation in the native bridge logs. 

This is a slightly different net result that the original implementation, which doesn't log the `console.error` for the unimplemented method invocation (because it isn't executed), but solves the original problem of the uncaught error. 

CleanShot 2023-05-23 at 23 14 12@2x

core/src/utils/native/keyboard.ts
```diff
@@ -18,10 +18,16 @@ export const Keyboard = {
   },
   getResizeMode(): Promise<KeyboardResizeOptions | undefined> {
     const engine = this.getEngine();
-    if (!engine || !engine.getResizeMode) {
+    if (!engine?.getResizeMode) {
       return Promise.resolve(undefined);
     }
-
-    return engine.getResizeMode();
+    return engine.getResizeMode().catch((e: any) => {
```
liamdebeasi, line 24: Can we make an interface for this? Could be something as simple as

```typescript
interface NativeError {
  code: string;
  message: string;
}
```

PR #27515, opened 2023-05-19
chore(many): clean up legacy control syntax in tests

amandaejohnston:
Issue number: N/A

---------

Many tests are using legacy form control syntax for secondary things like controls used as random children or to toggle extra functionality like dark mode.

## What is the new behavior?

Form controls in non-legacy tests converted to modern syntax as needed to clear console warnings. All screenshot updates represent only changes to the test pages themselves, _not_ component functionality.

Other changes made:
- `item/test/bottom`: removed entirely since it had no E2E file and was duplicated by various input tests
- `item/test/inputs`: removed buttons to change label placement since they weren't being used

- [ ] Yes
- [x] No


core/src/components/item/test/counter/index.html
liamdebeasi, line 1: Should we keep these tests until the legacy behavior has been removed? This at least lets us verify the old behavior has not regressed.

amandaejohnston: Sounds good, I'll re-add them.

liamdebeasi (approved):
Thank you!

core/src/components/item/test/bottom/index.html
brandyscarney, line 1: Should this be moved to legacy instead?

brandyscarney (commented):
Do we have some coverage for the legacy tests being removed?

core/src/components/item/test/bottom/index.html
amandaejohnston: > Do we have some coverage for the legacy tests being removed?

Looking through it again, these are the only ones where I've removed test functionality that doesn't have an existing legacy counterpart:

- `item/a11y`
- `radio-group/basic`
- `radio-group/form`
- `radio-group/search`

I can add tests for these to the relevant `legacy` directories if desired. As for `item/bottom`, I removed that one in the interest of clean-up since it didn't have an E2E file, and thus it seemed unlikely to be needed, but I can add it back in under the legacy folder too if you like.

amandaejohnston:
On second thought, I'm going to close this for now and recreate it with less aggressive changes later. I need to approach this ensuring that I don't remove any legacy testing functionality by mistake.

Issue #27516, opened 2023-05-19
 Incorrect Unit Test Results for AM/PM Indications


jon-pruessmeier:
### Prerequisites

- [X] I have read the [Contributing Guidelines](https://github.com/ionic-team/ionic-framework/blob/main/.github/CONTRIBUTING.md#creating-an-issue).
- [X] I agree to follow the [Code of Conduct](https://ionicframework.com/code-of-conduct).
- [X] I have searched for [existing issues](https://github.com/ionic-team/ionic-framework/issues) that already report this problem, without success.

### Ionic Framework Version

v7.x

### Current Behavior

The unit tests for AM/PM indications are failing due to the use of regular spaces instead of slim spaces (U+202F). The assertions in the unit tests compare the expected values, which contain regular spaces, with the received values that include slim spaces. This mismatch leads to assertion failures and inaccurate test results.

### Expected Behavior

The unit tests for AM/PM indications should use slim spaces (U+202F) for proper formatting. The expected values in the unit tests should be updated to include slim spaces instead of regular spaces. This will ensure accurate and consistent test results.

### Steps to Reproduce

1. Run the unit tests for AM/PM indications using the command npm run test.spec.

### Ionic Info

Happens in the Ionic Directory itself

### Additional Information

It is important to use slim spaces (U+202F) for AM/PM indications to ensure proper localization and formatting of time values. The use of regular spaces in the unit tests leads to inconsistencies and test failures. Please update the unit tests to reflect the correct usage of slim spaces for AM/PM indications.

liamdebeasi:
Hey there,

I replied to your PR in https://github.com/ionic-team/ionic-framework/pull/27517#issuecomment-1557166488, but I'm not able to reproduce the original issue. Could you provide some clarification around the errors you are receiving and the environment you are testing this on?

liamdebeasi:
Thanks! I am able to reproduce this on Node 19 but not on earlier versions. Let me do some digging and see if I can identify what is causing this.

liamdebeasi:
@jon-pruessmeier So it looks like this was a bug in the v8 engine that has since been fixed: https://bugs.chromium.org/p/v8/issues/detail?id=13494

I tried Node 19.9 and the error does not reproduce there, so I believe this has been fixed. Can you give Node 19.9 a try too?

jon-pruessmeier:
Hey @liamdebeasi 

[...]

PR #27517, opened 2023-05-19
fix(datetime, tests): using slim spaces in unit tests of datetime

jon-pruessmeier:
Issue number: #27516

---------

The unit tests for AM/PM indications are failing due to the use of regular spaces instead of slim spaces (U+202F). The assertions in the unit tests compare the expected values, which contain regular spaces, with the received values that include slim spaces. This mismatch leads to assertion failures and inaccurate test results.

## What is the new behavior?

I changed the occurences of all spaces with AM/PM to slim spaces. Now the tests are accepted.

- [ ] Yes
- [X] No

## Other information

Furthermore I left a comment in the respective file with an info about the space-topic so that other contributors have this topic in their mind if they adjust the file.

### DIFF
--- a/core/src/components/datetime/test/format.spec.ts
+++ b/core/src/components/datetime/test/format.spec.ts
@@ -8,6 +8,12 @@ import {
   getLocalizedTime,
 } from '../utils/format';
 
+/**
+ * IMPORTANT: AM/PM indications should use slim spaces (U+202F) for proper formatting.
+ * Please make sure to replace regular spaces with slim spaces when representing AM/PM.
+ * Slim spaces (U+202F) are used to ensure correct localization and formatting of time values.
+ */
+
 describe('generateDayAriaLabel()', () => {
   it('should return Wednesday, May 12', () => {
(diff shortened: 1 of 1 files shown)


liamdebeasi (commented):
Thanks for the PR! I'm not sure this is the correct solution as the tests are now failing on CI. What environment were you testing this in originally?

jon-pruessmeier:
> Thanks for the PR! I'm not sure this is the correct solution as the tests are now failing on CI. What environment were you testing this in originally?

Thanks for your feedback! Interesting, that those tests are failing now in CI. 

My system that I'm using is MacOS 13.1 with M1-Chip and I'm using Nodejs 19.4.0 and NPM 9.2.0
Maybe it is better when I replace the spaces with a regex that uses both versions of spaces? So any difficulties in regard to this pattern are catched?

liamdebeasi:
I am testing on macOS 13.1 with an M1 Pro chip on `main`, and the tests are passing as expected. Are you getting an error when you run the tests on `main`?

PR #27523, opened 2023-05-20
refactor(accordion): remove duplicate selector

cesarkohl:
Hey guys, I've found a minor issue.

There is a duplicate selector for the same rules. The lines `:24` and `:25` are the same.


liamdebeasi (approved):
Thank you!

PR #27529, opened 2023-05-22
feat(datetime): add shadow parts and CSS variables for styling wheel pickers

### LINKED ISSUE #25945: feat: ion-datetime wheel appearance
### Prerequisites

Hello team currently I had been working with the ion-datetime and have a lot of amazing things. It will be very useful I we can modify the styling of the ion-picker-internal, because is not meet the default font defined at global styles, we can change the Theme colors and background but no the size of the font, font-weight, font-size, font-family.

### Describe the Use Case

This feature will help to customize and meet the UI look & feel of each individual personalization.

_No response_

### Describe Alternatives

_No response_

### Related Code

`<ion-datetime
                        presentation="date"
                        [preferWheel]="true"
                    >
                    </ion-datetime>`

on sytles.scss
ion-datetime {
    color: #{$navy};
    font-weight: 700;
    --background: white;
    --ion-color-base: #{$navy} !important;
    --ion-color-step-150: #{$light-teal};
}

The result:
image

The expected result:
image

### Additional Information

This element should be customizable or have ::part element to get into it

- picker-item
- picker-item-active
- <div class="picker-before"></div>
- <div class="picker-after"></div>
- <div class="picker-highlight"></div>

image

amandaejohnston:
Issue number: resolves #25945

---------

Datetime wheel pickers cannot be styled.

## What is the new behavior?

Adds styling APIs in accordance with the Wheel Pickers and Time Picker sections of [this design doc](https://github.com/ionic-team/ionic-framework-design-documents/blob/main/projects/ionic-framework/components/datetime/datetime-styling.md).

Shadow parts added:
- `wheel-item`
- `wheel-item active`
- `time-button`
- `time-button active`

CSS properties added:
- `--wheel-highlight-background`
- `--wheel-fade-background-rgb`

- [ ] Yes
- [x] No

## Other information

Dev build: `7.0.7-dev.11685554390.10c2ca9b`
Docs PR: https://github.com/ionic-team/ionic-docs/pull/2982


core/src/components/datetime/datetime.tsx
```diff
@@ -74,6 +74,10 @@ import {
  * @slot title - The title of the datetime.
  * @slot buttons - The buttons in the datetime.
  * @slot time-label - The label for the time selector in the datetime.
+ * 
+ * @part picker-item - The individual items when using a wheel style layout, or in the
+ * month/year picker when using a grid style layout.
+ * @part picker-item active - The currently selected picker-item.
```
amandaejohnston, line 80: I couldn't just do `@part active` here since that part may apply to other elements too (such as selected dates) so I went with this since it's how the code will look (e.g. `::part(picker-item active)`).

I'm open to changing the active part name to `picker-item-active` instead. The [W3C guidelines](https://drafts.csswg.org/css-shadow-parts/#example-72ff019b) imply that state-based shadow parts should be non-specific (i.e. it should be `part="picker-item active"` and not `part="picker-item picker-item-active"`) but doing it that way does lead to somewhat awkward documentation.

core/src/components/datetime/test/custom/index.html
```diff
@@ -0,0 +1,83 @@
...
+        grid-template-columns: 1fr;
+        padding: 0;
+      }
+    }
+
+    /*
+      The second selectors that target ion-picker(-column)-internal
+      directly are for styling the time picker. This is currently
+      undocumented usage.
+    */
```
amandaejohnston, line 41: I added a note to update this in the design doc for making the internal picker public: https://ionic-cloud.atlassian.net/browse/FW-4401

core/src/components/datetime/test/custom/index.html
```diff
@@ -0,0 +1,89 @@
...
+          grid-template-columns: 1fr;
+          padding: 0;
+        }
+      }
+
+      /*
+      The second selectors that target ion-picker(-column)-internal
+      directly are for styling the time picker. This is currently
+      undocumented usage.
+    */
```
amandaejohnston, line 40: I added a note to update this in the design doc for making the internal picker public: https://ionic-cloud.atlassian.net/browse/FW-4401

core/src/components/datetime/test/custom/datetime.e2e.ts-snapshots/datetime-custom-time-button-active-md-ltr-Mobile-Firefox-linux.png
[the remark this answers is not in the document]
amandaejohnston: The shaded part is the popover slightly overlapping the button:

time picker popover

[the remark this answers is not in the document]
liamdebeasi: Oh makes sense. Thanks!

core/src/components/picker-internal/picker-internal.scss
```diff
@@ -4,6 +4,9 @@
 // --------------------------------------------------
 
 :host {
+  --picker-highlight-background: inherit;
+  --picker-fade-background: inherit;
```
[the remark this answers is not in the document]
amandaejohnston: Huh, guess not. Will remove 👀

core/src/components/picker-internal/picker-internal.ios.scss
```diff
@@ -2,13 +2,13 @@
 @import "../../themes/ionic.globals.ios";
 
 :host .picker-before {
-  background: linear-gradient(to bottom, var(--background, var(--ion-background-color, #fff)) 20%, rgba(var(--background-rgb, var(--ion-background-color-rgb, 255, 255, 255)), 0.8) 100%);
+  background: linear-gradient(to bottom, var(--picker-fade-background, var(--background, var(--ion-background-color, #fff))) 20%, rgba(var(--background-rgb, var(--ion-background-color-rgb, 255, 255, 255)), 0.8) 100%);
```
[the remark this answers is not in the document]
amandaejohnston: The problem with that is that then the gradient always goes from the highlight color, _to_ the highlight color, when it should actually be from the highlight color to the background. That way, you can have a highlight color that's different from the background, and it still blends with the rest of the page properly. I called out in the design doc that the end color of the gradient should not be editable for similar reasons.

What do you think about instead requiring that `--background-rgb` also be defined, if you're going to set `--background`? The `rgb` prop is [already documented](https://ionicframework.com/docs/api/datetime#css-custom-properties); maybe we could add a note to the upcoming styling playground that if you define one, you should define both.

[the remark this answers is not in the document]
amandaejohnston: Actually, on second thought, maybe you're right? Generally, _both_ colors of the gradient should actually be the same as the background -- that way it looks like the other options are fading into the BG. But people have asked for ways to change the color of the gradient, which implies they want it to be different from the BG... or maybe they just don't realize the BG and gradient colors are tied together, and thus they can already control it? 🤔 Hmm. Let me know your thoughts.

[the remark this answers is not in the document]
amandaejohnston: Sounds good. I'll give your suggestion a go 👀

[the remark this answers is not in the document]
amandaejohnston: Updated in https://github.com/ionic-team/ionic-framework/pull/27529/commits/ae8174f576734497a956caa51f1eadb3688ff436, and screenshots regenerated 👍

core/src/components/picker-internal/picker-internal.md.scss
```diff
@@ -1,10 +1,13 @@
 @import "./picker-internal.scss";
 @import "../../themes/ionic.globals.md";
 
+$fallback: var(--background-rgb, var(--ion-background-color-rgb, 255, 255, 255));
+$background: var(--picker-fade-background-rgb, #{$fallback});
```
liamdebeasi, line 5: Can we move these vars to a `picker-internal.vars.scss` file and have `picker-internal.{ios, md}.scss` pull from it?

core/src/components/datetime/test/custom/datetime.e2e.ts-snapshots/datetime-custom-time-button-active-ios-ltr-Mobile-Chrome-linux.png
sean-perkins, line 1: Is the darker pink/purple intended to cover the entire button?

core/src/components/picker-internal/picker-internal.vars.scss
```diff
@@ -0,0 +1,2 @@
+$picker-fade-background-fallback: var(--background-rgb, var(--ion-background-color-rgb, 255, 255, 255));
+$picker-fade-background: var(--picker-fade-background-rgb, #{$picker-fade-background-fallback});
```
sean-perkins, line 2:
```suggestion
$picker-fade-background: var(--picker-fade-background-rgb, $picker-fade-background-fallback);
```

I do not believe we need to escape the variable here.

core/src/components/datetime/test/custom/datetime.e2e.ts-snapshots/datetime-custom-time-button-active-ios-ltr-Mobile-Chrome-linux.png
amandaejohnston: See: https://github.com/ionic-team/ionic-framework/pull/27529#discussion_r1202829237

core/api.txt
```diff
@@ -424,7 +424,13 @@ ion-datetime,event,ionChange,DatetimeChangeEventDetail,true
 ion-datetime,event,ionFocus,void,true
 ion-datetime,css-prop,--background
 ion-datetime,css-prop,--background-rgb
+ion-datetime,css-prop,--picker-fade-background-rgb
+ion-datetime,css-prop,--picker-highlight-background
 ion-datetime,css-prop,--title-color
+ion-datetime,part,picker-item
+ion-datetime,part,picker-item active
```
liamdebeasi, line 431: Bikeshed: Can we use `wheel` terminology instead of `picker` to be consistent with the `preferWheel` property?

amandaejohnston: Works for me 👍

sean-perkins (approved):

liamdebeasi (approved):
Great job!

Issue #27438, opened 2023-05-09
bug: radio cannot be manually focused in Firefox using modern form control syntax


amandaejohnston:
### Prerequisites

- [X] I have read the [Contributing Guidelines](https://github.com/ionic-team/ionic-framework/blob/main/.github/CONTRIBUTING.md#creating-an-issue).
- [X] I agree to follow the [Code of Conduct](https://ionicframework.com/code-of-conduct).
- [X] I have searched for [existing issues](https://github.com/ionic-team/ionic-framework/issues) that already report this problem, without success.

### Ionic Framework Version

v7.x

### Current Behavior

When calling `focus()` on an `ion-radio` element using the modern syntax in Firefox, focus does not actually move to the radio. Screencast below. Things work as expected with the legacy syntax, where `document.activeElement` returns the `ion-radio`.

https://github.com/ionic-team/ionic-framework/assets/90629384/c74b78c4-9f4e-4dba-9e6e-24e029f8b21b

### Expected Behavior

`document.activeElement` should return the radio in the modern syntax.

### Steps to Reproduce

1. Host core locally and visit http://localhost:3333/src/components/radio/test/basic in Firefox.
2. Call `focus()` on the `ion-radio`.
3. Check `document.activeElement` and note that it returns the `body` tag.

### Code Reproduction URL

http://localhost:3333/src/components/radio/test/basic

### Ionic Info

```
[WARN] You are not in an Ionic project directory. Project context may be missing.

Ionic:

   Ionic CLI : 7.0.1

Utility:

   cordova-res : not installed globally
   native-run  : 1.4.1

System:

   NodeJS : v16.19.1
   npm    : 8.19.3
   OS     : Windows 10
```

### Additional Information

This causes problems in `select-popover` when using the modern syntax for the radios used to render each option. When opening the popover, focus does not move to the first radio as it does in other browsers and instead remains on the `ion-popover`.

PR #27530, opened 2023-05-22
fix(radio): radio with modern syntax is keyboard navigable

### LINKED ISSUE #27268: bug: Radio group tab navigation is incorrect for modern form control syntax

thetaPC:
Issue number: resolves #27268 

---------

When tabbing on a radio group (modern syntax), it focuses on the next radio option inside the radio group. It replicates the behavior of the up/down keys.

## What is the new behavior?

- When tabbing on a radio group (modern syntax), it focuses on the next radio option of the next radio group.

A spike ticket has been created to further investigate web accessibility and browser compatibility.

- [ ] Yes
- [x] No

## Other information

N/A


core/src/components/radio/radio.tsx
```diff
@@ -24,7 +24,9 @@ import { createColorClasses, hostContext } from '../../utils/theme';
     ios: 'radio.ios.scss',
     md: 'radio.md.scss',
   },
-  shadow: true,
+  shadow: {
+    delegatesFocus: true,
```
liamdebeasi, line 28: Threading:

This behavior does not work on Safari. Tabbing while focused on the first radio group moves focus to the `body`.


Also pressing the "Down" arrow selects the same radio.

liamdebeasi: Also, does tabbing between radio groups work in Safari on `main` with the legacy syntax? If not, then I think a separate ticket to fix that would be fine since we aren't regressing behavior.

thetaPC: It does work. However, `<ion-radio>.focus()` [still wouldn't work](https://github.com/ionic-team/ionic-framework/issues/27438). `delegatesFocus` was helping to fix that.
-
Tabbing on radio groups work on `main`. But it stops working if `delegatesFocus` is introduced.

liamdebeasi: The PR description links to https://github.com/ionic-team/ionic-framework/issues/27268, but it sounds like we are also trying to fix https://github.com/ionic-team/ionic-framework/issues/27438. I think we are going to have the most luck with this PR if we focus on fixing one bug at a time.

Updating the internal `setFocus` method to focus the native `input` element should fix #27268 since the radio group calls `setFocus`. It sounds like the other issue is something that is specific to Firefox and may require additional investigation.

core/src/components/radio/radio.tsx
```diff
@@ -135,7 +135,7 @@ export class Radio implements ComponentInterface {
     ev.stopPropagation();
     ev.preventDefault();
 
-    this.el.focus();
+    this.nativeInput.focus();
```
liamdebeasi, line 138: It looks like this causes some issues with the legacy syntax just in Safari. Doing a Shift+Tab to get back to the previous radio group requires pressing Shift+Tab twice.
```suggestion
    const element = this.legacyFormController.hasLegacyControl() ? this.el : this.nativeInput;
    element.focus();
```

Let's keep focusing `this.el` only when in the legacy syntax.

liamdebeasi (approved):
Good to go once my other comment has been addressed.

Can you make the commit title more descriptive too? Instead of "tab to the next radio group" maybe we say "fix(radio): radio with modern syntax is keyboard navigable"?

Great job!

PR #27545, opened 2023-05-24
feat(select): add label slot

### LINKED ISSUE #26838: feat: support `slot=label` syntax on v7+ `ion-select` component (like `ion-range`)
### Prerequisites

Currently, it is not possible to provide customized labels to `ion-select`, rather, only the `[label]="my label"` attribute is supported. `ion-select` should be updated to support complex labels - bringing it in line with other components like `ion-toggle`.

### Describe the Use Case

The new form syntax doesn't allow customization of `ion-select` labels. Making something like this impossible without using legacy `<ion-label>` code. We need the ability to add custom HTML / components to our labels. Plain text is not enough for our use cases:

image

Pass through non-`ion-select-option` elements directly to the `slot=label` slot of `ion-segment`:

Use `slot=label` to provide custom label elements, like `ion-range`:

```
  <ion-select placeholder="Choose one">
    <div slot=label>View mode</div>
    <ion-note slot=label>Choose how decks are presented visually when viewing or editing.</ion-note>

    <ion-select-option value=foo>foo</ion-select-option>
  </ion-select>
```

Additionally, I'd propose that ALL input components be updated to support `slot=label` as appropriate. It's an extra tripping point for developers to have to remember which of these components use `slot=label` and which accept label elements directly as children.

Passing elements as children and having them show up in label slot without requiring the `slot=label` attribute might also create additional migration scenarios in the future, should components start to use the Shadow DOM, or be re-worked to use slots in the future.

If everything required `slot=label` today, I think it would be more consistent for developers and more future-proof across the board.

### Describe Alternatives

_No response_

### Related Code

_No response_

### Additional Information

_No response_

liamdebeasi:
Issue number: resolves #26838

---------

Developers are unable to pass custom HTML as a label to `ion-select`.

## What is the new behavior?

- Adds a label slot to `ion-select` so developers can pass custom HTML for the label.

- [ ] Yes
- [x] No

## Other information

Docs PR: https://github.com/ionic-team/ionic-docs/pull/2971


core/src/components/select/select.tsx
```diff
@@ -576,7 +586,7 @@ export class Select implements ComponentInterface {
       label = this.getLabel();
       labelText = label ? label.textContent : null;
     } else {
-      labelText = this.label;
+      labelText = this.labelText;
```
thetaPC, line 589: Shouldn't this be `this.labelText()`?

core/src/components/select/select.tsx
```diff
@@ -566,7 +576,7 @@ export class Select implements ComponentInterface {
      * TODO FW-3194
      * Remove legacyFormController logic.
      * Remove label and labelText vars
-     * Pass `this.label` instead of `labelText`
+     * Pass `this.labelText` instead of `labelText`
```
thetaPC, line 579: Shouldn't this be `this.labelText()`?

core/src/components/select/select.tsx
```diff
@@ -904,10 +1091,10 @@ Developers can use the "legacy" property to continue using the legacy form marku
   }
 
   private get ariaLabel() {
-    const { placeholder, label, el, inputId, inheritedAttributes } = this;
+    const { placeholder, el, inputId, inheritedAttributes } = this;
     const displayValue = this.getText();
     const { labelText } = getAriaLabel(el, inputId);
-    const definedLabel = label ?? inheritedAttributes['aria-label'] ?? labelText;
+    const definedLabel = this.labelText ?? inheritedAttributes['aria-label'] ?? labelText;
```
thetaPC, line 1097: Shouldn't this be `this.labelText()`?

core/src/components/select/select.tsx
liamdebeasi: No, `this.labelText` is a getter which binds the property to a function. When you look up `this.labelText` the function is automatically called: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/get

thetaPC (approved):
LGTM

core/src/components/select/select.tsx
```diff
@@ -32,9 +33,12 @@ import type { SelectChangeEventDetail, SelectInterface, SelectCompareFn } from '
 /**
  * @virtualProp {"ios" | "md"} mode - The mode determines which platform styles to use.
  *
+ * @slot label - label - The label text to associate with the select. Use the "labelPlacement" property to control where the label is placed relative to the select. Use this if you need to render a label with custom HTML.
```
amandaejohnston, line 36:
```suggestion
 * @slot label - The label text to associate with the select. Use the "labelPlacement" property to control where the label is placed relative to the select. Use this if you need to render a label with custom HTML.
```

core/src/components/select/select.tsx
```diff
@@ -696,17 +730,170 @@ export class Select implements ComponentInterface {
...
+       * do not need to estimate the notch width.
+       */
+      !this.hasLabel ||
+      /**
+       * If the label property is being used
+       * then we can render the label text inside
+       * of the notch and let the browser
+       * determine the notch size for us.
+       */
+      this.label !== undefined
```
amandaejohnston, line 796: Optional: As far as the `label` checks go, `!this.hasLabel || this.label !== undefined` is equivalent to `this.label === undefined || this.label !== undefined`, which is always true, so you could combine these checks into `this.labelSlot === null`. I can understand wanting to black-box the `hasLabel` logic in case it changes in the future, though, so feel free to leave this as-is.

amandaejohnston: Ah, true! Sure, that works for me.

liamdebeasi: updated in https://github.com/ionic-team/ionic-framework/pull/27545/commits/ccc3b601d5f1305131496d3e02f6e43e0f4938de

amandaejohnston (approved):

PR #27559, opened 2023-05-25
fix(angular): remove invalid z-index style

sean-perkins:
Issue number: N/A

---------

The Angular implementation of `ion-tabs` uses SASS styles in the inline usage for styles:

```ts
styles: [`
  :host {
    z-index: $z-index-page-container;
  }
`]
```

However Angular is expecting CSS mark-up by default. This results in an invalid value being assigned to the `z-index`.

image (10)

## What is the new behavior?

- Removes the invalid style from the tabs stylesheet
- Removes unnecessary leading whitespace from the template

- [ ] Yes
- [x] No


liamdebeasi (approved):
Thank you!

PR #27560, opened 2023-05-26
fix(header,footer): resolve CSP violations with box shadow

sean-perkins:
Issue number: N/A

---------

The `ion-header` and `ion-footer` use a base64 encoded image for a box shadow instead of using the CSS box-shadow property directly. The use of the background image creates CSP violations. The historic reasoning of using an image instead of box shadow was to improve scroll performance. Browsers and devices have improved a lot since that was implemented (5 years ago). 

## What is the new behavior?

- Updates the usage of `ion-header` and `ion-footer` to use a box shadow. The value comes from Material's web implementation: https://material-components.github.io/material-components-web-catalog/#/component/top-app-bar

- [ ] Yes
- [x] No


sean-perkins:
There are so many screenshot tests that include the header, so this PR is naturally huge with file diffs. 

The implementation changes are tracked in this commit: [56e616f](https://github.com/ionic-team/ionic-framework/pull/27560/commits/56e616f4f81eb4f3dbd511a8f14ffaf90ba3ddf4)

Let me know if there are any question around verification/testing.

amandaejohnston (approved):
I was initially skeptical of the new shadow value since it looked too dark and too tall to me on the header, but if I check the HTML on [the actual demos](https://material-components.github.io/material-components-web-catalog/#/component/top-app-bar?variant=shortCollapsed), this is indeed the exact shadow used. It looks better when the header/footer has a dark background like MD's screenshots, such as in [this one](https://github.com/ionic-team/ionic-framework/blob/03f640a25b1dba2d29a0a2b8e3ae5612fad08d54/core/src/components/app/test/safe-area/app.e2e.ts-snapshots/app-toast-diff-md-ltr-Mobile-Chrome-linux.png).

liamdebeasi (approved):
This change looks good. I tested this on a Samsung Galaxy S3 with Chrome 81 and observed similar FPS between `main` and this branch. I did observe a slight increase in memory consumption when using the box shadow, but it did not seem to impact device performance.

PR #27570, opened 2023-05-30
fix(select): prevent click event from firing twice

mapsandapps:
Issue number: N/A

---------

The click event was firing twice.

## What is the new behavior?

- The click event fires once.

- [ ] Yes
- [x] No

## Other information

You can use the test file to see that the event now fires once, and before this change it fired twice. I couldn't figure out how to write an automated test for this that fails before the change and passes after the change. (Perhaps because of how Jest "clicks" elements? Not sure.)

Also, you can use the repro repo in the Jira ticket and update it to `"@ionic/angular": "7.0.10-dev.11685472954.170be0cc",` and see that the issue is fixed.

### DIFF
--- a/core/src/components/select/select.scss
+++ b/core/src/components/select/select.scss
@@ -224,6 +224,8 @@ button {
   cursor: inherit;
 
   box-sizing: border-box;
+
+  pointer-events: none;
 }
 
 .select-wrapper .select-placeholder {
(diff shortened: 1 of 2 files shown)


liamdebeasi:
FYI Sean is fixing a similar issue in https://github.com/ionic-team/ionic-framework/pull/27508. You two may want to link up and see if there is any knowledge/code that can be shared.

PR #27574, opened 2023-05-30
test(segment): undo skip for keyboard navigable

thetaPC:
Issue number: #

---------

## What is the new behavior?

-
-
-

- [ ] Yes
- [ ] No

### DIFF
--- a/core/src/components/segment/test/a11y/segment.e2e.ts
+++ b/core/src/components/segment/test/a11y/segment.e2e.ts
@@ -12,11 +12,10 @@ configs().forEach(({ title, config }) => {
     });
 
     // TODO FW-3710
-    test.skip('segment buttons should be keyboard navigable', async ({ page, browserName, skip }, testInfo) => {
+    test('segment buttons should be keyboard navigable', async ({ page, pageUtils, skip }) => {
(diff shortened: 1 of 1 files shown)


sean-perkins (approved):
I'm ok with these changes as-is. From your experience with working on this, do you think the feedback in the linked Playwright issue applies to our component as well?

Specifically:
> This may happen if the (js) event handlers are added after the element appears in the dom. This is an issue that would affect a human user too so should be considered a bug. Right way to fix it would be to make sure the element cannot be focused until it has all the event handlers ready.

thetaPC:
> I'm ok with these changes as-is. From your experience with working on this, do you think the feedback in the linked Playwright issue applies to our component as well?
> 
> Specifically:
> 
> > This may happen if the (js) event handlers are added after the element appears in the dom. This is an issue that would affect a human user too so should be considered a bug. Right way to fix it would be to make sure the element cannot be focused until it has all the event handlers ready.

I was able to narrow down the issue and it's not related to the linked issue. I'll be pushing updates in a bit.

sean-perkins (approved):
Great work debugging this flaky test 🎉

PR #27582, opened 2023-06-01
fix(angular): tabs supports conditional slot bindings

### LINKED ISSUE #19484: bug: ion-tab-bar slot not working with Angular binding
# Bug Report

**Ionic version:**

[x] **4.5.0**

**Current behavior:**

When assigning a variable to position the tab bar to top or bottom, the tab bar does not position correctly

**Expected behavior:**

For example, setting the slot position to top from a variable in a component, the tab bar should be on the top.

**Steps to reproduce:**

<ion-tab-bar [slot]="tabSlotPosition"> in the template and tabSlotPosition set to 'top' in the component.

**Related code:**

**Other information:**

If I set the value hard coded in the code, it works normally. For example:
<ion-tab-bar slot="top">

**Ionic info:**

Ionic:

   Ionic CLI                     : 5.4.1 (/Users/jeremie/.nvm/versions/node/v11.1.0/lib/node_modules/ionic)
   Ionic Framework               : @ionic/angular 4.5.0
   @angular-devkit/build-angular : 0.13.9
   @angular-devkit/schematics    : 7.3.9
   @angular/cli                  : 7.3.9
   @ionic/angular-toolkit        : 1.5.1

Cordova:

   Cordova Platforms : android 8.1.0
   Cordova Plugins   : cordova-plugin-ionic-keyboard 2.1.3, cordova-plugin-ionic-webview 4.1.1, (and 14 other plugins)

Utility:

   cordova-res : 0.3.0 (update available: 0.6.0)
   native-run  : 0.2.6 (update available: 0.2.8)

System:

   Android SDK Tools : 26.1.1 (/Users/jeremie/Library/Android/sdk)
   ios-deploy        : 1.9.4
   ios-sim           : 8.0.2
   NodeJS            : v11.1.0 (/Users/jeremie/.nvm/versions/node/v11.1.0/bin/node)
   npm               : 6.10.3
   OS                : macOS Mojave
   Xcode             : Xcode 11.0 Build version 11A420a

sean-perkins:
Issue number: Resolves #19484

---------

The Angular implementation of `ion-tabs` does not support conditional binding the `slot` property on `ion-tab-bar` or assigning a variable as the slot. 

For example, this usage is invalid:

```html
<ion-tabs>
  <ion-tab-bar [slot]="slot"></ion-tab-bar>
</ion-tabs>
```

This occurs because `ng-content` only supports static content projection. It is not intended for scenarios where the content can be relocated or conditionally rendered.

An example of static content projection would be:
```html
<ion-tabs>
  <ion-tab-bar slot="top"></ion-tab-bar>
</ion-tabs>
```

## What is the new behavior?

- `ion-tabs` supports conditional slot bindings or a slot that is bound to a variable in Angular.

The revised implementation relocates the tab bar in the `ion-tabs` template, based on it's current slot attribute. The implementation checks the tab bar slot whenever the content changes. 

- [ ] Yes
- [x] No

## Other information

Dev-build: `7.0.11-dev.11685631370.18980633`


angular/src/directives/navigation/ion-tabs.ts
```diff
@@ -137,4 +164,43 @@ export class IonTabs {
...
+        this.relocateTabBar();
+      }
+    });
+  }
+
+  /**
+   * Relocates the tab bar to the new slot position.
+   */
+  private relocateTabBar(): void {
+    const tabBar = (this.tabBar as any).el as HTMLElement;
```
sean-perkins, line 190: We have options here regarding the `any` type (and the `any` on L175). Our automatic generated proxies have the `el` property as protected. To make the types valid, we would need to disable automatically generating the proxy and manually wrap it with a `public` field member. 

I don't think the effort of maintaining the proxy is worth the type efficiency here.

sean-perkins:
Thoughts on targeting this to `feature-7.1` instead of `main`? While CI passes and I experimented locally, I have a small concern that someone is doing some weird content projection/rendering that _may_ not work as intended with this logic. Targeting a minor would minimize this risk.

thetaPC:
> Thoughts on targeting this to `feature-7.1` instead of `main`? While CI passes and I experimented locally, I have a small concern that someone is doing some weird content projection/rendering that _may_ not work as intended with this logic. Targeting a minor would minimize this risk.

I agree to target this to `feature-7.1`. There might be more edge cases that might happen down the line.

sean-perkins:
Sounds good 👍 I'll get `feature-7.1` up-to-date with `main` and then rebase so the diff is only the few files.

angular/src/directives/navigation/ion-tabs.ts
thetaPC: That makes sense. It might be a good idea to include a comment for the reasoning to keep it as type `any`.

thetaPC (approved):
LGTM, tested the build on the tabs starter.

PR #27602, opened 2023-06-06
fix(refresher): scroll styles are reset when using non-native refresher

### LINKED ISSUE #27601: bug: non-native refresher breaks scrolling
### Prerequisites

### Ionic Framework Version

v7.x

### Current Behavior

When using a non-native refresher, moving the view (refresher) slightly down, back to the top of the view (deltaY <= 0) and down again in one move (one tap and hold) breaks the scrolling of the view. See attached video for a better understanding.

https://github.com/ionic-team/ionic-framework/assets/86344176/eae73145-8d3a-4b58-9621-9760874466ee

### Expected Behavior

View scrolling does not break when using the refresher.

### Steps to Reproduce

1. Checkout the sample application: [https://github.com/mrahn24/ionic-bug-examples -> Branch: bug/r](https://github.com/mrahn24/ionic-bug-examples/tree/bug/refresher-breaks-scrolling) and run it
2. Tap and hold to use the refresher
3. Move down (do not move over pullMax)
4. Move up to the top of the view (deltaY <= 0)
5. Move down (do not move over pullMax)
6. Release tap and hold
7. Try to scroll down -> Scroll should be broken

### Code Reproduction URL

https://github.com/mrahn24/ionic-bug-examples/tree/bug/refresher-breaks-scrolling

### Ionic Info

Ionic:

   Ionic CLI                     : 6.20.1 (/Users/mrahn/.config/yarn/global/node_modules/@ionic/cli)
   @angular-devkit/build-angular : 16.0.4
   @angular-devkit/schematics    : 15.2.8
   @angular/cli                  : 16.0.4

Capacitor:

   Capacitor CLI      : 5.0.4
   @capacitor/core    : 5.0.4

Utility:

   cordova-res (update available: 0.15.4) : 0.15.1
   native-run                             : 1.7.2

System:

   NodeJS : v18.15.0 (/Users/mrahn/.n/bin/node)
   npm    : 9.5.0
   OS     : macOS

### Additional Information

Bug introduced with #26613

mrahn24:
Issue number: resolves #27601 

---------

The current behavior restores overflow styles while moving (within the setCSS function).

## What is the new behavior?
Overflow styles are restored when refresher gesture ends.

- [ ] Yes
- [x] No

Honestly, I don't know exactly. From code perspective I would say 'Yes', but I can't get the impact of the change.

Ionic Team edit: There are no changes to the public API, and this is fixing a behavior that used to work so there are no breaking changes.

### DIFF
--- a/core/src/components/refresher/refresher.tsx
+++ b/core/src/components/refresher/refresher.tsx
@@ -686,12 +686,19 @@ export class Refresher implements ComponentInterface {
     if (this.state === RefresherState.Ready) {
       // they pulled down far enough, so it's ready to refresh
       this.beginRefresh();
-    } else if ([RefresherState.Pulling, RefresherState.Inactive].includes(this.state)) {
+    } else if (this.state === RefresherState.Pulling) {
       // they were pulling down, but didn't pull down far enough
       // set the content back to it's original location
       // and close the refresher
       // set that the refresh is actively cancelling
       this.cancel();
+    } else if (this.state === RefresherState.Inactive) {
+      /**
+       * The pull to refresh gesture was aborted
+       * so we should restore any overflow styles
+       * that have been modified.
+       */
+      this.restoreOverflowStyle();
     }
   }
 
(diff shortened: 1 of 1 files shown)


liamdebeasi (commented):
I made some changes to your PR. The original fix worked, but the `restoreOverflowStyle` was  happening before the `setTimeout` callback fired in `close()`. This prevented some inline transition styles from being cleared properly. It did not cause any functional issues that I was aware of, but we should still make sure we are cleaning up styles when we no longer need them.

The revised fix causes the styles to be cleared once the setTimeout callback fires. I also added a case for when the pull gesture is reset.

sean-perkins:
I'm observing this behavior testing on device, specifically if you cancel the gesture, the first attempt to scroll prevents the scroll and activates the item. However, attempting to scroll after does work as expected. 

Is this expected?

https://github.com/ionic-team/ionic-framework/assets/13732623/0fb6d57a-6bce-45fc-8bea-430000e47f16

sean-perkins (approved):
We chatted in Slack, the behavior I observed is from: https://github.com/ionic-team/ionic-framework/blob/9d459bdc88f70d0094fd646c63d677573ab162e6/core/src/components/refresher/refresher.tsx#L715-L720 and is existing behavior that is independent of these changes. 

Great work both of you!

PR #27618, opened 2023-06-07
feat(datetime): add part for month/year button

liamdebeasi:
Issue number: #

---------

## What is the new behavior?

-
-
-

- [ ] Yes
- [ ] No

### DIFF
--- a/core/src/components/datetime/datetime.md.scss
+++ b/core/src/components/datetime/datetime.md.scss
@@ -36,7 +36,7 @@
 
 :host .calendar-action-buttons ion-item,
 :host .calendar-action-buttons ion-button {
-  color: #{$text-color-step-350};
+  --color: #{$text-color-step-350};
 }
 
 // Calendar / Header / Days of Week
--- a/core/src/components/datetime/datetime.tsx
+++ b/core/src/components/datetime/datetime.tsx
@@ -82,6 +82,9 @@ import {
  * @part time-button - The button that opens the time picker when using a grid style
  * layout with `presentation="date-time"` or `"time-date"`.
  * @part time-button active - The time picker button when the picker is open.
+ *
+ * @part month-year-button - The button that opens the month/year picker when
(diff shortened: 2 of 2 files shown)


core/src/components/datetime/datetime.md.scss
```diff
@@ -36,7 +36,7 @@
 
 :host .calendar-action-buttons ion-item,
 :host .calendar-action-buttons ion-button {
-  color: #{$text-color-step-350};
+  --color: #{$text-color-step-350};
```
liamdebeasi, line 39: Since we are exposing the `ion-item` instance we should use the public API for setting the color otherwise developers will run into issues changing the color using `--color`.

liamdebeasi: Turns out this also fixes a bug where the expanded state of the month/year toggle never changed color! This is why there are screenshot diffs here.

Example: https://ionicframework.com/docs/api/datetime#basic-usage

Switch to MD mode and click the month/year button. Observe that the color of the button only changes to blue when you hover over it. Repeat the same steps on iOS mode. Observe that the color of the button correctly changes to blue regardless of hover state when expanded.

sean-perkins (approved):

amandaejohnston (approved):
LGTM aside from a small description change 👍

core/src/components/datetime/datetime.tsx
```diff
@@ -82,6 +82,9 @@ import {
  * @part time-button - The button that opens the time picker when using a grid style
  * layout with `presentation="date-time"` or `"time-date"`.
  * @part time-button active - The time picker button when the picker is open.
+ *
+ * @part month-year-button - The button that opens the month/year picker when
+ * using a grid style layout with `presentation="date-time"` or `"time-date"`.
```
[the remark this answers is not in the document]
liamdebeasi: Good call -- fixed in https://github.com/ionic-team/ionic-framework/pull/27618/commits/0ae7407e6e34e6551e1fcfce46d077f788d1a236

PR #27634, opened 2023-06-12
chore(global): remove unused context references

rwaskiewicz:
Issue number: N/A

---------

Ionic Framework references a deprecated object,`Context`, exposed by Stencil

## What is the new behavior?

starting with stencil v4, the deprecated `Context` object will no longer be exposed by stencil. this change was introduced in https://github.com/ionic-team/stencil/pull/4437, and will be present in the first v4 prerelease following v4.0.0-beta.2. in anticipation for this change, we seek to remove references to `Context` early.

- [ ] Yes
- [x] No - To the best of my knowledge

## Other information

ATM, the Stencil v4 nightly build is passing, as it's grabbing `@stencil/core@4.0.0-beta.2`. The change in which we remove the `Context` object will occur in the _next_ pre-release. I'm removing `Context` here to try to get a jump on things. 

The current iteration of the code that I'm deleting was added in https://github.com/ionic-team/ionic-framework/commit/c415bbe1d712a609ac565ef42640bcb16bea02c8#diff-ce62e75f0c31a76aac491f13a64e9c7771a6cbae8ca6635541164b69f0479bf1

### DIFF
--- a/core/src/global/ionic-global.ts
+++ b/core/src/global/ionic-global.ts
@@ -7,8 +7,6 @@ import { config, configFromSession, configFromURL, saveConfig } from './config';
 
 // TODO(FW-2832): types
 
-declare const Context: any;
-
 let defaultMode: Mode;
 
 export const getIonMode = (ref?: any): Mode => {
@@ -22,7 +20,6 @@ export const initialize = (userConfig: IonicConfig = {}) => {
 
   const doc = window.document;
   const win = window;
-  Context.config = config;
   const Ionic = ((win as any).Ionic = (win as any).Ionic || {});
 
   const platformHelpers: any = {};


core/src/global/ionic-global.ts
```diff
@@ -22,7 +20,6 @@ export const initialize = (userConfig: IonicConfig = {}) => {
 
   const doc = window.document;
   const win = window;
-  Context.config = config;
```
liamdebeasi, line 25: In https://github.com/ionic-team/ionic-framework/commit/c415bbe1d712a609ac565ef42640bcb16bea02c8#diff-ce62e75f0c31a76aac491f13a64e9c7771a6cbae8ca6635541164b69f0479bf1, I also see references to [DomController](https://github.com/ionic-team/ionic-framework/blob/1f0c8049a339e3a77c468ddba243041d08ead0be/angular/src/providers/dom-controller.ts#L30). This util also has references to some `Ionic.queue` variable.

Is the queue related to this context change at all?

liamdebeasi: Originally added in https://github.com/ionic-team/ionic-framework/commit/6a31f3960a96f094eabfd350a9437e12acacf6df. However, I don't really know how `ionic.queue` gets set.

sean-perkins: It looks like `Ionic.queue` is the result of [`getContext`](https://github.com/ionic-team/stencil/blob/87fb34a2534a94fdf5627da814cd301a2ed9b661/src/runtime/context.ts#L19-L28) when `context` is `queue`. 

I am seeing only a reference to it being set in the compiled code for screenshot compares on Stencil's repo:
```ts
(this.queue = o(this, "queue"));
```

Source: https://github.com/ionic-team/stencil/blob/main/screenshot/compare/build/p-227a1e18.entry.js (hasn't been updated in 3 years). 

A feature removed at some point perhaps?

liamdebeasi: Hmm so it looks like this will break DomController then. Is there an alternative queue implementation that Stencil implements?

rwaskiewicz: I think @sean-perkins is right - this was likely functionality that was removed (and is only findable in the Stencil repo due to checked in build artifacts). 

I don't know quite enough about Ionic's architecture to really understand how/when this service (`DomController`) gets used. In the test subdir,  I see it gets [injected into the ctor of the `ProvidersComponent`,](https://github.com/ionic-team/ionic-framework/blob/fe39a2ab9b609827a1d67ec5632181d1991d12e3/angular/test/base/src/app/providers/providers.component.ts#L36) but that seems to only act as a [sentinel to tell if the component is loaded](https://github.com/ionic-team/ionic-framework/blob/fe39a2ab9b609827a1d67ec5632181d1991d12e3/angular/test/base/src/app/providers/providers.component.ts#L43-L45). I also seeing it getting [exported here](https://github.com/ionic-team/ionic-framework/blob/fe39a2ab9b609827a1d67ec5632181d1991d12e3/angular/src/index.ts#L33), but I'm not sure why.

That is all to say, I'm not fully sure how `DomController` gets used here, which makes it tricky to figure out for sure if this is actually affecting the behavior or not.

Are there any docs (internal/external) you could point me to to give me a better sense of it all?

sean-perkins: Where is `DomController` used? I see it in the old diff @liamdebeasi linked, but I don't actually see a reason why it exists. It looks like a dead Angular provider.

IMO I think the recommended changes here make sense, unless I'm missing where `DomController` is used.

liamdebeasi: It's an old API from the v3 days that got ported over for v4: https://github.com/ionic-team/ionic-v3/blob/cf35d5eb7f38d088e7c2588e42738228c39e9964/src/platform/dom-controller.ts#L57. The goal of the DomController was to provide a way of manipulating the DOM while avoiding layout thrashing. It's a public API that was never documented. However, it has been used in some community posts (https://ionic.io/blog/building-interactive-ionic-apps-with-gestures-and-animations).

Am I correct in my understanding that the underlying `queue` functionality has already been removed, meaning DomController is currently just using raf?

rwaskiewicz: > Am I correct in my understanding that the underlying queue functionality has already been removed, meaning DomController is currently just using raf?

As far as I understand it (based on this conversation), yes. I don't see any assignments to `Ionic.queue` or any calls to Stencil's `getContext` that (Sean linked) in the Framework codebase

liamdebeasi: Ok cool. In that case I'm fine with the proposed change. Sean and I should discuss whether or not to deprecate the DomController since it's really a wrapper around raf at this point.

liamdebeasi (approved):

PR #27636, opened 2023-06-12
feat(input): add workaround for dynamic slot content

liamdebeasi:
Issue number: N/A

---------

Slotted content added async does not cause the component to re-render.

## What is the new behavior?

- Adding slotted content async causes the component to re-render
- Changes within the slotted content will also cause the component to re-render so the notch is updated

- [ ] Yes
- [x] No


core/src/utils/forms/slot-mutation-controller.ts
amandaejohnston, line 1: Nitpick: `forms` doesn't seem like the best directory for this since it's more generic. Right in `utils` might be better, so it would sit alongside `watch-options.ts` which is similar.

amandaejohnston (approved):

core/src/utils/forms/slot-mutation-controller.ts
liamdebeasi: fixed in https://github.com/ionic-team/ionic-framework/pull/27636/commits/b66bd8dfc634377bb65cbde3d66a306a994fd2ba

core/src/components/input/test/slot/index.html
```diff
@@ -91,8 +91,49 @@ <h2>Outline / Floating</h2>
               <div slot="label">Email <span class="required">*</span></div>
             </ion-input>
           </div>
+
+          <div class="grid-item">
+            <h2>Outline / Floating / Async</h2>
+            <ion-input id="solid-async" label-placement="floating" fill="outline" value="hi@ionic.io"></ion-input>
```
brandyscarney, line 97: If I update the slotted content to the really long label, then remove it, the notch does not re-calculate, should it?
Screenshot 2023-06-14 at 1 14 19 PM

When it starts out with no label the notch is much smaller:
Screenshot 2023-06-14 at 1 12 48 PM

liamdebeasi: Removing the slotted content was fixed in https://github.com/ionic-team/ionic-framework/pull/27636/commits/db1d32e33e74175447814bc8a242f4fa0ee71fa0

The small notch with no label is a bug in `main` too, but I fixed it in https://github.com/ionic-team/ionic-framework/pull/27636/commits/7d2ea9dca4985efa0c343cefbdf4fa2b6a988b75

brandyscarney: I was actually going to comment that the small notch showing without a label was a bug too but I looked and Material Design has it too. 😂 I do think it's better to just hide it though so this is an improvement.

liamdebeasi: Actually the small notch thing also impacts textarea and select, so I'm going to fix that in `main` instead.

brandyscarney (approved):
Looks great!

core/src/components/input/test/slot/index.html
brandyscarney: Makes sense to me. I'm fine with leaving it separate.

liamdebeasi: Textarea fix: https://github.com/ionic-team/ionic-framework/pull/27647/commits/178cdebb8fa0525554ddbc193219cb93434b2e81i

Select fix: https://github.com/ionic-team/ionic-framework/pull/27649

Issue #18495, opened 2019-06-09
bug: ion-textarea wrap="off" is not working 


scrpgil:
Bug Report

I can not do English. This sentence uses Google translation.

**Ionic version:**
[x] **4.4.2**

**Current behavior:**

Line feed will occur even if the attribute of `wrap="off"` is added to ion-textarea.
The reason is that `white-space: pre-wrap` is set in src/components/textarea/textarea.scss.

https://github.com/ionic-team/ionic/blob/10de1da9480896c58bedde06ee022c1456d68354/core/src/components/textarea/textarea.scss#L92

Also, even if this `white-space: pre-wrap` is deleted, `white-space: inherit` is set in ionic.mixins.scss, so the line feed is always Will occur.

https://github.com/ionic-team/ionic/blob/10de1da9480896c58bedde06ee022c1456d68354/core/src/themes/ionic.mixins.scss#L32

If you have `wrap="off"` attribute to solve this problem, it would be better to add CSS to set `white-space: nowrap`.

I am ready for a pull request to solve this problem.
https://github.com/scrpgil/ionic/pull/1

**Expected behavior:**

If `wrap="off"` the text is not broken.

**Steps to reproduce:**

1. Add the `wrap="off"` attribute to ion-textarea.

2. Enter text in ion-textarea that you do not want to break. For example, like ASCII art.

3. Make sure that the line breaks and the display is broken.

line wrap(Current behavior)
![Screen Shot 2019-06-10 at 3 25 10](https://user-images.githubusercontent.com/26167770/59162759-9f132080-8b31-11e9-8948-0f45b8315476.png)

It should be displayed as below.

line no wrap(Expected behavior:)
![Screen Shot 2019-06-10 at 3 24 59](https://user-images.githubusercontent.com/26167770/59162765-c4079380-8b31-11e9-84b4-2f3d7b22b1b3.png)

**Related code:**

```
<ion-textarea wrap="off"></ion-textarea>
```

not working text wrap.

**Other information:**

**Ionic info:**

```
Ionic:

   ionic (Ionic CLI)             : 4.12.0 (/usr/local/lib/node_modules/ionic)
   Ionic Framework               : @ionic/angular 4.4.2
   @angular-devkit/build-angular : 0.13.8
   @angular-devkit/schematics    : 7.3.8
   @angular/cli                  : 7.3.8
   @ionic/angular-toolkit        : 1.5.1

System:

   NodeJS : v10.15.3 (/usr/local/bin/node)
[...]

PR #27647, opened 2023-06-14
feat(textarea): add label slot

liamdebeasi:
Issue number: N/A

---------

Textarea only accepts plaintext labels.

## What is the new behavior?

- Textarea can now accept custom HTML labels via a slot

A few implementation notes:

- Support for dynamically adding/removing slotted content will be added in a separate PR.
- Support for the notch outline will be added in a separate PR.
- The existing start/end placement label overflow tests were made redundant with the new tests I added. As a result, I removed the old tests.

- [ ] Yes
- [x] No


core/src/components/textarea/textarea.scss
```diff
@@ -67,13 +67,27 @@
...
+   * textarea defaults to break-spaces which
+   * can cause extra whitespace to appear
+   * when using newlines before a slotted
+   * label element. To avoid this, we use
+   * "collapse" which is the default value
+   * for an <input />
+   * If textarea is converted to Shadow DOM
+   * then this may not be needed.
+   */
+  white-space-collapse: collapse;
```
liamdebeasi, line 81: `<input />` defaults to `collapse` whereas textarea defaults to `preserve` which is why this change was only needed on ion-textarea.

core/src/components/textarea/textarea.scss
```diff
@@ -74,6 +72,8 @@
   flex: 1;
 
   background: var(--background);
+
+  white-space: pre-wrap;
```
liamdebeasi, line 76: This change is necessary because the old code was causing whitespace to appear in the textarea if you added newlines when slotted in a label:

```html
<ion-textarea label-placement="start" value="hi@ionic.io">
  <div slot="label">Email <span class="required">*</span></div>
</ion-textarea>
```

This code was added as a result of https://github.com/ionic-team/ionic-framework/issues/18495 to make it so you can set `white-space` on `ion-textarea` and have it inherit to the native textarea. This is not a pattern we should recommend using the modern form controls as it will impact both the textarea and the label, which is not always what developers want. Instead we should encourage developers to target the label or the textarea directly when changing white space.

liamdebeasi: You'll also see a few screenshot diffs because this is impacting textareas in `main`: https://github.com/ionic-team/ionic-framework/pull/27647/commits/cf85fa4a4164d96d3408cdb7f27488ff5d773cf2. The buggy behavior is not common right now but it will likely become more common once label slots become available.

core/src/components/textarea/textarea.tsx
```diff
@@ -530,17 +536,37 @@ Developers can use the "legacy" property to continue using the legacy form marku
...
     return (
-      <div class="label-text-wrapper">
-        <div class="label-text">{this.label}</div>
+      <div
+        class={{
+          'label-text-wrapper': true,
+          'label-text-wrapper-hidden': !this.hasLabel,
+        }}
+      >
+        {label === undefined ? <slot name="label"></slot> : <div class="label-text">{label}</div>}
```
thetaPC, line 547: The screen reader (Voice Over) is reading the label three time when the textarea is using a slot and twice for when it's using a label prop. This isn't related to the ticket but it would be good to create a spike ticket for further investigation.

thetaPC (approved):
LGTM

core/src/components/textarea/textarea.tsx
liamdebeasi: How are you testing this? I tested this with VoiceOver on macOS (Safari and Chrome) and iOS 16+17 and the screen reader announces the slotted label once.

liamdebeasi: Ah I can reproduce this on Firefox. However, I can reproduce this outside of Ionic too so I think this may be a platform bug (either with Firefox or VoiceOver). The following reads "Hello World, edit text, Hello World, group":

```html
<label>
  Hello World
  <textarea></textarea>
</label>
```

However, the following reads "Hello World, edit text, main":

```html
<label for="text">Hello World</label>
<textarea id="text"></textarea>
```

liamdebeasi: Filed a bug: https://bugzilla.mozilla.org/show_bug.cgi?id=1838502

thetaPC: Interesting indeed. I agree that it's an external bug. Then I vote to create a spike to figure out where and submit a bug ticket to them if there isn't one.

thetaPC: I can verify that this is also happening with `ion-radio`. Works great for Chrome but not Firefox.

core/src/components/textarea/test/label-placement/textarea.e2e.ts
```diff
@@ -234,3 +211,32 @@ configs().forEach(({ title, screenshot, config }) => {
     });
   });
 });
+
+configs({ modes: ['md'], directions: ['ltr'] }).forEach(({ title, screenshot, config }) => {
+  test.describe(title('textarea: label overflow'), () => {
+    test('label property should be truncated with ellipses', async ({ page }) => {
```
brandyscarney, line 217:
```suggestion
    test('label property should be truncated with an ellipsis', async ({ page }) => {
```

Nitpick, feel free to ignore but singular is ellipsis

core/src/components/textarea/test/label-placement/textarea.e2e.ts
```diff
@@ -234,3 +211,32 @@ configs().forEach(({ title, screenshot, config }) => {
...
+        `
+            <ion-textarea label="Label Label Label Label Label" placeholder="Text Input"></ion-textarea>
+          `,
+        config
+      );
+
+      const textarea = page.locator('ion-textarea');
+      expect(await textarea.screenshot()).toMatchSnapshot(screenshot(`textarea-label-truncate`));
+    });
+    test('label slot should be truncated with ellipses', async ({ page }) => {
```
brandyscarney, line 228:
```suggestion
    test('label slot should be truncated with an ellipsis', async ({ page }) => {
```

Nitpick, feel free to ignore but singular is ellipsis

core/src/components.d.ts
```diff
@@ -2921,6 +2921,9 @@ export namespace Components {
          */
         "mode"?: "ios" | "md";
     }
+    /**
+     * @experimental The label text to associate with the textarea. Use the `labelPlacement` property to control where the label is placed relative to the textarea. Use this if you need to render a label with custom HTML.
```
brandyscarney, line 2925: Why is this experimental tag showing up on top of textarea but it doesn't mention the actual `label` slot or property?

core/src/components/textarea/textarea.scss
brandyscarney: This makes sense to me.

core/src/components/textarea/textarea.tsx
```diff
@@ -15,6 +15,8 @@ import type { TextareaChangeEventDetail, TextareaInputEventDetail } from './text
 
 /**
  * @virtualProp {"ios" | "md"} mode - The mode determines which platform styles to use.
+ *
+ * @slot label - @experimental The label text to associate with the textarea. Use the `labelPlacement` property to control where the label is placed relative to the textarea. Use this if you need to render a label with custom HTML.
```
brandyscarney, line 19: Is `@experimental` in the right place here? It shows up weird in the component file

brandyscarney (approved):
Looks good besides the notch outline calculation which will be done in a separate PR. 👍

core/src/components.d.ts
liamdebeasi: Not sure. Looks like this is also happening for the input slot too. Let me check with the Stencil team to see if I'm doing something wrong.

liamdebeasi: Fixed in https://github.com/ionic-team/ionic-framework/pull/27647/commits/de54350bb6540b246df3207bb16422461dc59004

core/src/components/textarea/textarea.tsx
liamdebeasi: See https://github.com/ionic-team/ionic-framework/pull/27647#discussion_r1230172878

PR #27648, opened 2023-06-14
feat(select): add props to customize toggle icons

### LINKED ISSUE #17248: feat: Support custom icons for `ion-select`

amandaejohnston:
Issue number: resolves #17248

---------

## What is the new behavior?

-
-
-

- [ ] Yes
- [ ] No

### DIFF
--- a/core/src/components/select/select.tsx
+++ b/core/src/components/select/select.tsx
@@ -184,6 +184,19 @@ export class Select implements ComponentInterface {
   @Prop() selectedText?: string | null;
 
   /**
+   * The toggle icon to use. Defaults to `chevronExpand` for `ios` mode,
+   * or `caretDownSharp` for `md` mode.
+   */
+  @Prop() toggleIcon?: string;
+
+  /**
+   * The toggle icon to show when the select is open. If defined, the icon
+   * rotation behavior in `md` mode will be disabled. If undefined, `toggleIcon`
+   * will be used for when the select is both open and closed.
+   */
+  @Prop() toggleIconWhenOpen?: string
+
+  /**
    * The shape of the select. If "round" it will have an increased border radius.
    */
   @Prop() shape?: 'round';
@@ -929,7 +942,7 @@ export class Select implements ComponentInterface {
   }
 
   private renderSelect() {
-    const { disabled, el, isExpanded, labelPlacement, justify, placeholder, fill, shape, name, value } = this;
+    const { disabled, el, isExpanded, toggleIconWhenOpen, labelPlacement, justify, placeholder, fill, shape, name, value } = this;
     const mode = getIonMode(this);
     const hasFloatingOrStackedLabel = labelPlacement === 'floating' || labelPlacement === 'stacked';
     const justifyEnabled = !hasFloatingOrStackedLabel;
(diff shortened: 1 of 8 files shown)


core/src/components/select/select.tsx
```diff
@@ -1083,7 +1110,16 @@ Developers can use the "legacy" property to continue using the legacy form marku
    */
   private renderSelectIcon() {
     const mode = getIonMode(this);
-    const icon = mode === 'ios' ? chevronExpand : caretDownSharp;
+    const { isExpanded, toggleIcon, toggleIconWhenOpen } = this;
+    let icon;
```
sean-perkins, line 1114: This will be an `any` type, can we explicitly set a type for it?

core/src/components/select/select.tsx
```diff
@@ -183,6 +183,19 @@ export class Select implements ComponentInterface {
...
+   * or `caretDownSharp` for `md` mode.
+   */
+  @Prop() toggleIcon?: string;
+
+  /**
+   * The toggle icon to show when the select is open. If defined, the icon
+   * rotation behavior in `md` mode will be disabled. If undefined, `toggleIcon`
+   * will be used for when the select is both open and closed.
+   */
+  @Prop() toggleIconWhenOpen?: string;
```
liamdebeasi, line 197: Thoughts on making `toggleIcon` and `toggleIconWhenOpened` named `collapsedIcon` and `expandedIcon`, respectively? `ion-refresher` uses `pullingIcon` and `refreshingSpinner`, and datetime uses `collapsedIcon` and `expandedIcon` internally. It seems like we tend to change the name entirely when the icon is different.

liamdebeasi (commented):
Functionality works great. Nice job! Left a thread for discussing the prop names.

core/src/components/select/select.tsx
amandaejohnston: There are a couple reasons I went with the current names:
1. `ion-accordion` uses `toggleIcon` for basically the same thing.
2. `toggleIcon` can also be used for when the select is open/expanded, if `toggleIconWhenOpen` isn't specified.

I'm open to your suggestions though, if you feel it's still worth it.

liamdebeasi: cc @brandyscarney any thoughts on this? You're good with naming things.

Maybe we have `toggleIcon` and `expandedIcon`?

amandaejohnston: We talked in Slack and are going with `toggleIcon` and `expandedIcon` 👍

liamdebeasi (approved):
Good to go once the CI errors are resolved. Nice work!

sean-perkins (approved):

PR #27649, opened 2023-06-14
fix(select): hide notch cutout if no visible label provided

liamdebeasi:
Issue number: N/A

---------

Brandy discovered a bug in input, textarea, and select where the notch cutout remains visible even when no visible label is provided. I fixed this for input in https://github.com/ionic-team/ionic-framework/pull/27636 and for textarea in https://github.com/ionic-team/ionic-framework/pull/27647.

```html
<ion-select fill="outline" label-placement="stacked" value="select" aria-label="my select"></ion-select>
```

image

## What is the new behavior?

- The notch cut out is hidden if no visible label is provided

- [ ] Yes
- [x] No

### DIFF
--- a/core/src/components/select/select.scss
+++ b/core/src/components/select/select.scss
@@ -316,7 +316,8 @@ button {
  * then the element should be hidden otherwise
  * there will be additional margins added.
  */
-.label-text-wrapper-hidden {
+.label-text-wrapper-hidden,
+.select-outline-notch-hidden {
   display: none;
 }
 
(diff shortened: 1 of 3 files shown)


brandyscarney (approved):
Looks good but I think your example in the description is wrong 🙂 

localhost_3333_src_components_select_test_fill

mapsandapps (approved):
Works well. The notch cutout is still shown if you provide an empty string to `label`. I'm not sure if that's expected or not.

liamdebeasi:
> Looks good but I think your example in the description is wrong 🙂

Oops I used the textarea by mistake. Updated!

> Works well. The notch cutout is still shown if you provide an empty string to label. I'm not sure if that's expected or not.

I think that's expected. We determine if a control has no label based on if the `label` prop is `undefined` or is `[slot="label"]` is `null` when doing a querySelector.

PR #27641, opened 2023-06-13
feat(datetime): add parts for calendar day, active, and today

brandyscarney:
Issue number: resolves #25340

---------

## What is the current behavior?
It is not possible to style the calendar days in a datetime since it uses shadow DOM and there are not exposed parts or CSS variables for the calendar days.

## What is the new behavior?
- Exposes the following parts for a calendar day: `calendar-day`, `today`, and `active`
- Combines the `calendar-day-highlight` element with the `calendar-day` element so developers don't have to know to style two different elements & we don't have to expose them as separate parts
- Updates the `color` e2e test to include an example of styling days using the newly exposed CSS parts

## Does this introduce a breaking change?

- [ ] Yes
- [x] No

## Other information

This PR has a number of screenshot diffs. I have gone through all of them and believe they are either improvements to datetime or inconsequential changes. 

1) Material Design height changes - several screenshots show the height of the datetime for `md` increasing, [this comment](https://github.com/ionic-team/ionic-framework/pull/27641/files#r1300532834) explains why this is a good change. (_I did not comment on every screenshot with a height increase, but you can see the pixel change at the bottom to verify it was caused by this_).
2) iOS height changes - several screenshots show a shift in the height of the datetime for `ios`. This is caused by the browsers rendering fractional pixels differently. [This comment](https://github.com/ionic-team/ionic-framework/pull/27641/files/64fa7cbbbef3a47d3f6113bcd0c1454c53050c5c#r1301870771) explains why this is a desired change. (_I did not comment on every screenshot with a height increase, but you can see the pixel change at the bottom to verify it was caused by this_).
3) Material Design circle height increase - The calendar day highlight has increased in height for non-active days. I believe this is a good change because I don't see why they would be smaller, but I am open to discussion if there are some guidelines I missed here. [This comment](https://github.com/ionic-team/ionic-framework/pull/27641/files/64fa7cbbbef3a47d3f6113bcd0c1454c53050c5c#r1301788277) shows the difference in height.
4) Numbers shifting - many, many screenshots show a horizontal shift in 1-3 of the numbers. I added comments to (I believe) all of the screenshots where this is happening because it is quite difficult to see the difference on some. [This comment](https://github.com/ionic-team/ionic-framework/pull/27641/files/64fa7cbbbef3a47d3f6113bcd0c1454c53050c5c#r1297680604) shows a video of the shift.

PR #27651, opened 2023-06-14
chore(): add updated snapshots

brandyscarney:
Adds screenshots for https://github.com/ionic-team/ionic-framework/pull/27641


liamdebeasi:
Have you considered placing all of the circle styles on `.calendar-day-highlight` instead of removing it entirely? I think part of the problem is each button is a column in the grid. The grid columns expand as the datetime expands in width, but the visible circle shouldn't change in size.

I made a sample on https://github.com/ionic-team/ionic-framework/tree/1593-alt. I was able to keep the dimensions for the MD datetime pretty similar. iOS is close, but the button height is wrong (I didn't have time to figure that part out, but I will if I have time this week).

PR #27655, opened 2023-06-15
fix(keyboard): keyboard events emit correctly when Capacitor is available but the Keyboard plugin is not

### LINKED ISSUE #27654: ionKeyboardDidShow and ionKeyboardDidHide browser keyboard events stopped working
https://github.com/ionic-team/ionic-framework/blob/d3232dcc007738fae851660bd818c8056b48b7d2/core/src/utils/keyboard/keyboard.ts#LL34C2-L34C2

Keyboard.getEngine() returns false and never undefined, hence the else block will never run and visualViewport.onresize is never registered.

liamdebeasi:
Issue number: resolves #27654

---------

When Capacitor and the Keyboard plugin are available, `Keyboard.getEngine()` returns an object: https://github.com/ionic-team/ionic-framework/blob/d3232dcc007738fae851660bd818c8056b48b7d2/core/src/utils/native/keyboard.ts#L19

When Capacitor is _not_ available this method returns `undefined`.

However, when Capacitor is available but the Keyboard plugin is not, `Keyboard.getEngine()` returns `false`.

In https://github.com/ionic-team/ionic-framework/pull/27569 I fixed a bug where the wrong listeners were being used to emit keyboard lifecycle events. However, I did not know that method can return `false`, and our tests only account for the `undefined` or defined edge cases. As a result, if you are using Capacitor in an app that is deployed to the browser then the visual viewport Ionic keyboard events never fire.

## What is the new behavior?

- Visual viewport Ionic keyboard events fire in browser

- [ ] Yes
- [x] No

### DIFF
--- a/core/src/utils/keyboard/keyboard.ts
+++ b/core/src/utils/keyboard/keyboard.ts
@@ -31,7 +31,7 @@ export const startKeyboardAssist = (win: Window) => {
    * how it resizes such that the Visual Viewport API is not
    * reliable here.
    */
-  if (nativeEngine !== undefined) {
+  if (!!nativeEngine) {
     startNativeListeners(win);
   } else {
     if (!(win as any).visualViewport) {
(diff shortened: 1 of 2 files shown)


amandaejohnston (approved):

thetaPC (approved):
LGTM

PR #27656, opened 2023-06-15
chore(ci): use local version of lerna

liamdebeasi:
Issue number: N/A

---------

[GitHub Actions comes with Lerna 7 preinstalled](https://github.com/actions/runner-images/blob/main/images/linux/Ubuntu2204-Readme.md#project-management) which has a breaking change for `lerna bootstrap`. This is causing Dev builds and releases to intermittently fail despite our `package.json` wanting Lerna 5.

## What is the new behavior?

- Ensure that the local version of Lerna is used instead of the global Lerna. Long term we should look to update to Lerna 7 and migrate off `lerna bootstrap`.

- [ ] Yes
- [x] No

### DIFF
--- a/.github/workflows/actions/publish-npm/action.yml
+++ b/.github/workflows/actions/publish-npm/action.yml
@@ -23,7 +23,7 @@ runs:
       with:
         node-version: 16.x
     - name: Install Dependencies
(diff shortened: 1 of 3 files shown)


liamdebeasi:
Thanks for reaching out! I DM'ed you on Twitter.

PR #27661, opened 2023-06-15
feat(textarea): add workaround for dynamic slot content

liamdebeasi:
Issue number: N/A

---------

Slotted content added async does not cause the component to re-render.

## What is the new behavior?

- Adding slotted content async causes the component to re-render
- Changes within the slotted content will also cause the component to re-render so the notch is updated

- [ ] Yes
- [x] No


core/src/components/textarea/test/fill/textarea.e2e.ts-snapshots/textarea-fill-outline-hidden-slotted-label-md-ltr-Mobile-Chrome-linux.png
amandaejohnston, line 1: Looks like the branch might need a rebase? I'm guessing these screenshots shouldn't be in here.

core/src/components/textarea/test/slot/index.html
```diff
@@ -91,8 +91,49 @@ <h2>Outline / Floating</h2>
...
+
+          <div class="grid-item">
+            <h2>Outline / Floating / Async</h2>
+            <ion-textarea id="solid-async" label-placement="floating" fill="outline" value="hi@ionic.io"></ion-textarea>
+          </div>
         </div>
+
+        <ion-button onclick="addSlot()">Add Slotted Content</ion-button>
+        <ion-button onclick="updateSlot()">Update Slotted Content</ion-button>
+        <ion-button onclick="removeSlot()">Remove Slotted Content</ion-button>
```
amandaejohnston, line 103: The notch is still present when the slot is removed:

slot removed

amandaejohnston (changes requested):

core/src/components/textarea/test/slot/index.html
liamdebeasi: Ah yeah same issue as your screenshot comment -- I think I need to rebase since I fixed a few bugs related to this in the input PR.

liamdebeasi:
Grabbed latest changes from feature-7.1, so both issues should be resolved.

PR #27665, opened 2023-06-16
Add alt implementation

liamdebeasi:
Issue number: #

---------

## What is the new behavior?

-
-
-

- [ ] Yes
- [ ] No


core/src/components/datetime/datetime.tsx
```diff
@@ -2139,12 +2139,16 @@ export class Datetime implements ComponentInterface {
...
                   style={{
                     backgroundColor: dateStyle?.backgroundColor,
+                    color: dateStyle?.textColor,
                   }}
-                ></div>
-                {text}
+                >
+                  {text}
+                </div>
+                <span aria-hidden="true" style={{ opacity: '0' }}>{text}</span>
```
liamdebeasi, line 2151: hacky, but this ensures the grid rows keep the same height as they have in `main`

PR #27676, opened 2023-06-20
fix(nav): root component is mounted with root params

### LINKED ISSUE #27146: bug: Upgrade to ionic 7 cause bugs when passing data to modal
### Prerequisites

### Ionic Framework Version

v7.x

### Current Behavior

After I upgrade to ionic 7, 

modal will work normally only the first time I open any modal.

If I close and then open modal again, the modal with any data passed through componentProps will not be received on the target component and cause bugs.

### Expected Behavior

ionic 7 should not break the existing behavior of modal.

### Steps to Reproduce

1.  create functionality to open modal with a component and pass some data to it.
2. in that component, reference the member of that data
3. run and open modal
4. close the modal, then open again
5. bug

### Code Reproduction URL

_No response_

### Ionic Info

Ionic:

   Ionic CLI                     : 7.0.1 (C:\Users\User\AppData\Roaming\npm\node_modules\@ionic\cli)
   Ionic Framework               : @ionic/angular 7.0.1
   @angular-devkit/build-angular : 14.2.11
   @angular-devkit/schematics    : 10.1.7
   @angular/cli                  : 14.2.11

Capacitor:

   Capacitor CLI      : 2.5.0
   @capacitor/android : 2.5.0
   @capacitor/core    : 2.5.0
   @capacitor/ios     : 2.5.0

Utility:

   native-run (update available: 1.7.2) : 1.4.0

System:

   NodeJS : v16.18.0 (C:\Program Files\nodejs\node.exe)
   npm    : 8.19.2
   OS     : Windows 10

### Additional Information

_No response_

sean-perkins:
Issue number: Resolves #27146

---------

Mounting the same root component twice in `ion-nav` will result in the `rootParams` not being assigned in the same frame as when the `attributesChangedCallback` fires for setting the root component: https://github.com/ionic-team/ionic-framework/blob/main/core/src/components/nav/nav.tsx#L77-L87

## What is the new behavior?

- Wraps the `setRoot` method call in a `requestAnimationFrame` callback that resolves after the `rootParams` have been assigned on the element (if any are specified).

- [ ] Yes
- [x] No

## Other information

Dev-build: `7.0.15-dev.11687289605.183f41ea`

You can install against the reproduction app in the linked issue to verify the behavior before and after. Before, the app will throw an exception when presenting the modal the second time. After, the app will not throw an exception when presenting the modal the second time.

### DIFF
--- a/core/src/components/nav/nav.tsx
+++ b/core/src/components/nav/nav.tsx
@@ -1,11 +1,12 @@
 import type { EventEmitter } from '@stencil/core';
 import { Build, Component, Element, Event, Method, Prop, Watch, h } from '@stencil/core';
+import { printIonWarning } from '@utils/logging';
 
 import { config } from '../../global/config';
 import { getIonMode } from '../../global/ionic-global';
 import type { Animation, AnimationBuilder, ComponentProps, FrameworkDelegate, Gesture } from '../../interface';
 import { getTimeGivenProgression } from '../../utils/animation/cubic-bezier';
-import { assert } from '../../utils/helpers';
+import { assert, raf } from '../../utils/helpers';
 import type { TransitionOptions } from '../../utils/transition';
 import { lifecycle, setPageHidden, transition } from '../../utils/transition';
 import type { NavOutlet, RouteID, RouteWrite, RouterDirection } from '../router/utils/interface';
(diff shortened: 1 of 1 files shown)


core/src/components/nav/nav.tsx
```diff
@@ -77,12 +78,19 @@ export class Nav implements NavOutlet {
...
-      } else if (isDev) {
-        console.warn('<ion-nav> does not support a root attribute when using ion-router.');
-      }
+
+    if (this.root === undefined) {
+      return;
+    }
+
+    if (!this.useRouter) {
+      raf(() => {
```
liamdebeasi, line 87: Why does using an raf work?

liamdebeasi (approved):
Tested and this works well. Nice job!

PR #27678, opened 2023-06-20
test(alert): remove unneeded disabled Axe rule

amandaejohnston:
Issue number: N/A

---------

One of the Alert a11y tests skips the `color-contrast` rule when checking Axe. This was done in https://github.com/ionic-team/ionic-framework/pull/27351 because the test started failing due to `color-contrast` behavior changes on Axe's end: https://github.com/dequelabs/axe-core/releases/tag/v4.7.0

## What is the new behavior?

Skip removed. It looks like the test failure was due to a problem on Axe's side exclusive to v4.7.0, as I'm not able to reproduce with Axe Core 4.7.1 (using Chrome's dev tools) or Framework's currently used 4.7.3 (by running the a11y test itself).

- [ ] Yes
- [x] No

### DIFF
--- a/core/src/components/alert/alert.ios.vars.scss
+++ b/core/src/components/alert/alert.ios.vars.scss
@@ -15,7 +15,7 @@ $alert-ios-max-width:                                   270px !default;
 $alert-ios-border-radius:                               13px !default;
 
 /// @prop - Background color of the alert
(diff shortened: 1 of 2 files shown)


amandaejohnston:
Closing this since the test is still failing in CI. Will add notes and further steps to the original ticket.

PR #27695, opened 2023-06-23
test(item-sliding): update visual regressions

thetaPC:
Issue number: -

---------

I forgot to address the tests when working on FW-2608 to address issue #26103 

## What is the new behavior?

- Updated the visual regressions tests to verify that the item-sliding displays correctly on LTR/RTL. 
- Updated the item sliding tests to use the generators

- [ ] Yes
- [x] No

## Other information

The tests for item-sliding are disabled due to gesture flakiness. This will be addressed in another ticket. I recommend testing locally by adding the `.only` to all tests called `should not have visual regressions` and running `npm run test.e2e item-sliding`.


mapsandapps (approved):
looks good when running locally 👍🏻

Issue #24583, opened 2022-01-16
bug: scrolling issues with bottom sheet modal


davideas:
### Prerequisites

- [X] I have read the [Contributing Guidelines](https://github.com/ionic-team/ionic-framework/blob/main/.github/CONTRIBUTING.md#creating-an-issue).
- [X] I agree to follow the [Code of Conduct](https://ionicframework.com/code-of-conduct).
- [X] I have searched for [existing issues](https://github.com/ionic-team/ionic-framework/issues) that already include this feature request, without success.

### Describe the Feature Request

Possibility to scroll content if this exceeds screen size only when bottom sheet is fully expanded and touched the max breakpoint.
Then, when finishing to scroll the content down and there's nothng to scroll anymore, the bottom sheet should be dragged again at lower breakpoint.

### Describe the Use Case

Having longer content in bottom sheet, content that exceeds the screen height is not visible because container not scrollable.
It's even a real problem for smaller screens.

amandaesmith3:
Thanks for the issue. Could you replicate your use case in a starter app and post the link here? This sounds like it may just be a bug, so I want to make sure I'm following.

davideas:
I created a quick sample project here: https://ionic-angular-action-sheet-g1wwzx.stackblitz.io
But it should be used in **mobile view** with any browser not desktop view! In desktop view, scrolling works, it is usable nicely with mouse wheel too, but the problem of course is on mobile view.
There, the touch creates issues when content is longer than the screen.

I saw a css property on `bottom-sheet-modal` element in ion-page class: `overflow: hidden;`
Well even modifying it or adding overflow to an internal div of the bottom sheet when touching the component it drag and scroll at same moment. That's why I suggest to activate `overflow-y: auto` only when max breakpont is reached.
When at lower breakpoints the touch should only drag the bottom sheet up and down without interferring with the scrolling of the content.

Could you please verify how we can scroll this content?

amandaesmith3:
Thanks! I see what you mean now. After investigation, it looks like the behavior you want exists on `ion-content` -- once the sheet modal is at its max breakpoint, any `ion-content` components inside get the `scroll-y` class, allowing scrolling. Could you try replacing the `div.modal-content` with `ion-content` and let me know if that looks how you would expect?

Either way, our documentation around this should be improved, since there isn't any example HTML for `ModalPage` in the usage examples (https://ionicframework.com/docs/api/modal#usage).

matthewgoodman13:
[...]

PR #27696, opened 2023-06-24
fix(modalSheet): change gesture to allow draging when content scrolle…

Captainfl4me:
Change gesture to allow draging when user overscroll

Issue number: #24583

---------

As mentionned in [this comment](https://github.com/ionic-team/ionic-framework/issues/24583#issuecomment-1020197730) on the issue [#24583](https://github.com/ionic-team/ionic-framework/issues/24583). Once you can scroll, dragging down on anywhere but the handle when the content is scrolled to the top doesn't do anything, when it should move the modal.

## What is the new behavior?
- When their is a ion-content tag inside a modal and the modal is fully opened (max breakpoint) ion-content is scrollable and scrolling at the top now drag the modal instead of doing nothing.
- Ion-content only scroll when you drag the content and not the ion-toolbar at the top of the modal

- [ ] Yes
- [x] No

## Other information

Apart from scrolling and dragging modal everything stay unchanged.

### DIFF
--- a/core/src/components/modal/gestures/sheet.ts
+++ b/core/src/components/modal/gestures/sheet.ts
@@ -41,7 +41,6 @@ export const createSheetGesture = (
   backdropBreakpoint: number,
   animation: Animation,
   breakpoints: number[] = [],
-  getCurrentBreakpoint: () => number,
   onDismiss: () => void,
   onBreakpointChange: (breakpoint: number) => void
 ) => {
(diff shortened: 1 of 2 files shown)


thetaPC:
Thank you for the PR! 

This PR will be closed since the [issue](https://github.com/ionic-team/ionic-framework/issues/24583) was [resolved](https://github.com/ionic-team/ionic-framework/pull/29312).

PR #27719, opened 2023-06-29
refactor(angular): move to packages directory

liamdebeasi:
Issue number: N/A

---------

The `angular` directory sits at the root of the project instead of in `packages` with all the other JS Framework integrations. This does not cause any functional issues with Ionic, but it is confusing since integrations are not in a consistent place.

## What is the new behavior?

- Moves the `angular` directory to `packages/angular`

Note: Most files should remain unchanged. The only files I changed are the files that had direct paths to the old `angular` directory:

1. Removes the `angular` path in `lerna.json`. This is now covered by `packages/*`
2. Updated the angular file path in `.gitignore`
3. Updates the path to the angular package in `stencil.config.ts` for the Angular Output Targets
4. Updates some of Angular's sync scripts to correctly get the core stylesheets as well as the core package.
5. Updates the test app sync script to correctly sync core and angular-server

~I'm not entirely sure why GitHub thinks https://github.com/ionic-team/ionic-framework/pull/27719/files#diff-f5bba7e7c7c75426e2b9c89868310cb03890493b4efe0252adf8d12cc8398962 is a new file since it exists in `main` here: https://github.com/ionic-team/ionic-framework/blob/1f06be4a31965f2a949b4866a585aee6af0af29d/angular/test/base/scripts/build-ionic.sh~ Fixed in https://github.com/ionic-team/ionic-framework/pull/27719/commits/6e7fc49827778799f3907e1e7178e9b79b4c4c00

- [ ] Yes
- [x] No

## Other information

Dev build: `7.1.2-dev.11688052109.13454f5c`


liamdebeasi:
Oh looks like I didn't actually move `angular/test/base/scripts`, I just copied it which is why `build-ionic.sh` was showing up as new. That file and the other noted files should be updated in https://github.com/ionic-team/ionic-framework/pull/27719/commits/6e7fc49827778799f3907e1e7178e9b79b4c4c00.

liamdebeasi:
Great catch! Opened a PR to update the Stencil Site: https://github.com/ionic-team/stencil-site/pull/1169

sean-perkins (approved):
Tested locally and things look good 👍

PR #27732, opened 2023-06-30
refactor(select): use border mixin

liamdebeasi:
Issue number: N/A

---------

The outline select uses `border-left` and `border-right` properties to handle LTR vs RTL borders. However, our border mixin takes care of this by using logical border properties.

## What is the new behavior?

- Updates the outline select to use the border mixin.

Note 1: There should be no visual changes as a result of this. This is simpler way of doing what we are already doing.
Note 2: There do exist logical border radius properties (for the other explicit LTR vs RTL work we do), but Ionic 7 supports browsers that do not support these properties yet. (I created FW-4661 to track this)

- [ ] Yes
- [x] No

### DIFF
--- a/core/src/components/select/select.md.outline.scss
+++ b/core/src/components/select/select.md.outline.scss
@@ -196,16 +196,16 @@
   pointer-events: none;
 }
 
+:host(.select-fill-outline) .select-outline-start {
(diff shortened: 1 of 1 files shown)


thetaPC (approved):
LGTM

PR #27745, opened 2023-07-05
feat: add dynamic type variables

liamdebeasi:
Issue number: N/A

---------

Developers have no way of opting in to using Dynamic Type on iOS.

## What is the new behavior?

- The typography styles sets `font` to `--ion-dynamic-type`. This variable is unset initially to not cause unexpected breakages in applications. Developers must opt-in to using Dynamic Type with the following CSS:

```css
html {
}
```

This will be enabled by default in an upcoming major release of Ionic.

- Updated the testing stylesheet to opt-in to using Dynamic Type to make it easier for team members to test Dynamic Type behavior when testing on iOS devices.

- [ ] Yes
- [x] No


core/src/css/typography.scss
```diff
@@ -32,6 +32,16 @@ html {
...
 
+/**
+ * Dynamic Type is an iOS-only feature, so
+ * this should only be enabled on iOS devices.
+ */
+@supports (-webkit-touch-callout: none) {
+  html {
+    font: var(--ion-dynamic-type);
+  }
+}
```
liamdebeasi, line 43: macOS now has Dynamic Type support. However, the base font size on macOS is 13px. While this aligns with the rest of the macOS design system, it does not align with Ionic components which are designed for mobile. As a result, enabling this on macOS would cause fonts to appear too small on desktop. As a result, the dynamic type font is only applied on iOS. `-webkit-touch-callout` detects WebKit + iOS support.

sean-perkins (approved):

brandyscarney (approved):
This looks good! I was running into an issue because I tried changing the `h1` font-size in [typography.scss](https://github.com/ionic-team/ionic-framework/blob/4906d02e08c0809146f95c32a8c25f7b13c0a234/core/src/css/typography.scss#L14) to use `rem` units but it needed a full re-build to see the changes. 

I tested it out using the `src/components/text/test/basic` test.

PR #27747, opened 2023-07-05
feat(action-sheet): support dynamic type

liamdebeasi:
Issue number: N/A

---------

Action Sheet does not support Dynamic Type on iOS

## What is the new behavior?

- Convert iOS font sizes to rem and added a `max` function to ensure that text does not go smaller than the set values
- Convert MD font sizes to rem
- Update buttons to use `min-height` instead of `height`. I also needed to change `contain` from `strict` to `content` because `strict` caused the button to ignore children when setting the dimensions.
- Added `max-height: 10dvh` so the container does not get clipped by the URL toolbar in Safari. This has technically always been an issue, but it is more easily visible now that the container can grow in size. 100vh is calculated assuming the iOS toolbar is collapsed, which is not always correct. dvh (The "d" stands for "Dynamic") changes based on if the toolbar is collapsed or not.

| 100vh | 100dvh |
| - | - |
| IMG_0006 | IMG_0007 |

- [ ] Yes
- [x] No

## Other information

The min font size for iOS was not in the original design doc as I missed this initially. I updated the design doc to account for this: https://github.com/ionic-team/ionic-framework-design-documents/pull/114

Demo Video: https://github.com/ionic-team/ionic-framework/assets/2721089/d08c8a81-c11f-4a1d-8635-09f7f91a849a


core/src/components/action-sheet/action-sheet.ios.vars.scss
```diff
@@ -97,19 +97,19 @@ $action-sheet-ios-sub-title-padding-start:                        $action-sheet-
 $action-sheet-ios-button-height:                                  56px !default;
 
 /// @prop - Padding of the action sheet button
-$action-sheet-ios-button-padding:                                 18px !default;
+$action-sheet-ios-button-padding:                                 14px !default;
```
liamdebeasi, line 100: Now that we are using `min-height` instead of `height`, the default height was being set to `64px` due to this padding. Our old `height: 56px` masked the problem. You can see it in the following screenshot where the icon extends beyond the boundary of the button height (excluding the padding)
IMG_0008

core/src/components/action-sheet/action-sheet.md.vars.scss
```diff
@@ -70,10 +70,10 @@ $action-sheet-md-button-height:                         52px !default;
 $action-sheet-md-button-text-color:                     $text-color-step-150 !default;
 
 /// @prop - Font size of the action sheet button
-$action-sheet-md-button-font-size:                      16px !default;
+$action-sheet-md-button-font-size:                      1rem !default;
 
 /// @prop - Padding top of the action sheet button
-$action-sheet-md-button-padding-top:                    0 !default;
+$action-sheet-md-button-padding-top:                    12px !default;
```
liamdebeasi, line 76: I added padding here because the buttons looked too cramped when the text scaled up. This should have no impact on buttons at the default text scale

| without padding | with padding |
| - | - |
| IMG_0012 | IMG_0011 |

brandyscarney (approved):
Looks good!

amandaejohnston (approved):

PR #27750, opened 2023-07-05
feat(back-button): support dynamic type

liamdebeasi:
Issue number: #

---------

## What is the new behavior?

-
-
-

- [ ] Yes
- [ ] No

### DIFF
--- a/core/src/components/back-button/back-button.ios.scss
+++ b/core/src/components/back-button/back-button.ios.scss
@@ -13,10 +13,21 @@
   --color: #{$back-button-ios-color};
   --icon-margin-end: 1px;
   --icon-margin-start: -4px;
-  --icon-font-size: 1.6em;
+  /**
+   * This value used to be 1.6em which would evaluate
+   * to 1.6em * 17px (the font size of the button) = 27.2px.
+   * As a result, we set that to the minimum font size.
+   */
+  --icon-font-size: clamp(27.2px, 1.6rem, 36px);
   --min-height: 32px;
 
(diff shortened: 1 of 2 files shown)


core/src/components/back-button/back-button.ios.scss
```diff
@@ -13,10 +13,21 @@
   --color: #{$back-button-ios-color};
   --icon-margin-end: 1px;
   --icon-margin-start: -4px;
-  --icon-font-size: 1.6em;
+  /**
+   * This value used to be 1.6em which would evaluate
+   * to 1.6em * 17px (the font size of the button) = 27.2px.
+   * As a result, we set that to the minimum font size.
+   */
+  --icon-font-size: clamp(27.2px, 1.6rem, 36px);
```
brandyscarney, line 21: Do we have a tech debt ticket to fix these weird font sizes?

brandyscarney (commented):
Should the `size="large"` text be allowed to scale larger or should it also hit the max size?

| `100%` | `135%` |
| --| --|
| IMG_2385 | IMG_2384 |

core/src/components/back-button/back-button.ios.scss
liamdebeasi: Brandy and I discussed offline and we kept 1.6em because the icon should be sized relative to the text, and the text already has a) dynamic type support and b) font size clamping

brandyscarney (approved):
Great work! I re-ran the failing screenshot job, seems like a flaky test.

PR #27755, opened 2023-07-06
refactor: use capacitor types for native plugins

sean-perkins:
Issue number: Internal

---------

Ionic currently detects and uses Capacitor APIs for different plugins (haptics, status bar and keyboard). This implementation does not have type safety and can result in unexpected behaviors. 

## What is the new behavior?

- Adds `@capacitor/core`, `@capacitor/keyboard`, `@capacitor/haptics` and `@capacitor/status-bar` as dev dependencies. These should _only_ be used with `import type { }`. 
- Refactors the plugin usages to be typed against the plugin packages, while using a duplicate enum when needing a value. This allows us to not bundle the capacitor plugins with Ionic Framework.
- Introduces a `getCapacitor()` function for interacting with the `window.Capacitor` object through a typed object. 

**How does it work?**

The idea is we want the type safety from the Capacitor packages, without directly bundling that source code within Ionic Framework. This means we use the Capacitor deps where a type is needed, but clone any enums where a value is referenced. If a Capacitor dep changes the supported values, Typescript will fail to compile and that will signal to use to update our enum values to match any changes. 

- [ ] Yes
- [x] No

## Other information

Dev-build: `7.1.2-dev.11688696027.1c4d4ad1`

Tested against a demo app for some of the core behavior: https://github.com/sean-perkins/capacitor-ionic-plugins-demo

### DIFF
--- a/core/package.json
+++ b/core/package.json
@@ -37,6 +37,10 @@
   },
   "devDependencies": {
     "@axe-core/playwright": "^4.7.3",
+    "@capacitor/core": "^5.1.1",
+    "@capacitor/haptics": "^5.0.5",
+    "@capacitor/keyboard": "^5.0.5",
+    "@capacitor/status-bar": "^5.0.5",
     "@ionic/eslint-config": "^0.3.0",
     "@ionic/prettier-config": "^2.0.0",
     "@jest/core": "^27.5.1",
--- a/core/src/utils/native/capacitor.ts
+++ b/core/src/utils/native/capacitor.ts
@@ -1,13 +1,13 @@
-import type { CapacitorInstance } from '@capacitor/core/types/definitions-internal';
+import type { CapacitorGlobal } from '@capacitor/core';
 
-let capacitor: CapacitorInstance | undefined;
+let capacitor: CapacitorGlobal | undefined;
 
 export const getCapacitor = () => {
   if (capacitor !== undefined) {
     return capacitor;
   }
   if (window !== undefined) {
-    return (window as any)?.Capacitor as CapacitorInstance;
+    return (window as any)?.Capacitor as CapacitorGlobal;
   }
   return undefined;
 };
--- a/core/src/utils/native/capacitor.ts
+++ b/core/src/utils/native/capacitor.ts
@@ -1,12 +1,7 @@
 import type { CapacitorGlobal } from '@capacitor/core';
 import { win } from '@utils/browser';
 
-let capacitor: CapacitorGlobal | undefined;
-
 export const getCapacitor = () => {
-  if (capacitor !== undefined) {
-    return capacitor;
-  }
   if (win !== undefined) {
     return (win as any).Capacitor as CapacitorGlobal;
   }
--- a/core/src/utils/native/capacitor.ts
+++ b/core/src/utils/native/capacitor.ts
@@ -1,3 +1,3 @@
-import type { CapacitorGlobal } from '@capacitor/core';
+import type { CapacitorInstance } from '@capacitor/core/types/definitions-internal'
 
-export const capacitor = (window as any).capacitor as CapacitorGlobal;
+export const capacitor = (window as any)?.Capacitor as CapacitorInstance;
--- a/core/src/utils/native/capacitor.ts
+++ b/core/src/utils/native/capacitor.ts
@@ -1,0 +1,3 @@
+import type { CapacitorGlobal } from '@capacitor/core';
+
+export const capacitor = (window as any).capacitor as CapacitorGlobal;
--- a/core/src/utils/native/capacitor.ts
+++ b/core/src/utils/native/capacitor.ts
@@ -1,4 +1,5 @@
 import type { CapacitorGlobal } from '@capacitor/core';
+import { win } from '@utils/browser';
 
 let capacitor: CapacitorGlobal | undefined;
 
@@ -6,8 +7,8 @@ export const getCapacitor = () => {
   if (capacitor !== undefined) {
     return capacitor;
   }
-  if (window !== undefined) {
-    return (window as any)?.Capacitor as CapacitorGlobal;
+  if (win !== undefined) {
+    return (win as any).Capacitor as CapacitorGlobal;
   }
   return undefined;
 };
--- a/core/src/utils/native/capacitor.ts
+++ b/core/src/utils/native/capacitor.ts
@@ -1,3 +1,13 @@
-import type { CapacitorInstance } from '@capacitor/core/types/definitions-internal'
+import type { CapacitorInstance } from '@capacitor/core/types/definitions-internal';
 
-export const capacitor = (window as any)?.Capacitor as CapacitorInstance;
+let capacitor: CapacitorInstance | undefined;
+
+export const getCapacitor = () => {
+  if (capacitor !== undefined) {
+    return capacitor;
+  }
+  if (window !== undefined) {
+    return (window as any)?.Capacitor as CapacitorInstance;
+  }
+  return undefined;
+};
--- a/core/src/utils/native/haptic.ts
+++ b/core/src/utils/native/haptic.ts
@@ -1,19 +1,78 @@
-// Main types for this API
+import type { HapticsPlugin, NotificationType as CapacitorNotificationType, ImpactStyle as CapacitorImpactStyle } from '@capacitor/haptics';
+
+import { capacitor } from './capacitor';
+
+export enum ImpactStyle {
+  /**
+   * A collision between large, heavy user interface elements
+   *
+   * @since 1.0.0
+   */
+  Heavy = "HEAVY",
+  /**
+   * A collision between moderately sized user interface elements
+   *
+   * @since 1.0.0
+   */
+  Medium = "MEDIUM",
+  /**
+   * A collision between small, light user interface elements
+   *
+   * @since 1.0.0
+   */
+  Light = "LIGHT"
+}
+
 interface HapticImpactOptions {
-  style: 'light' | 'medium' | 'heavy';
+  style: CapacitorImpactStyle;
+}
+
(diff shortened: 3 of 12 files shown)


core/package.json
```diff
@@ -37,6 +37,10 @@
   },
   "devDependencies": {
     "@axe-core/playwright": "^4.7.3",
+    "@capacitor/core": "^5.1.1",
+    "@capacitor/haptics": "^5.0.5",
+    "@capacitor/keyboard": "^5.0.5",
+    "@capacitor/status-bar": "^5.0.5",
```
liamdebeasi, line 43: I think it would be good to have dependabot update these deps so we can be aware of any changes to the types

core/src/utils/native/capacitor.ts
```diff
@@ -0,0 +1,13 @@
+import type { CapacitorInstance } from '@capacitor/core/types/definitions-internal';
```
liamdebeasi, line 1: Is it safe to rely on internal definitions?

sean-perkins: Ideally we could use `CapacitorGlobal` (the public type definition that Capacitor exports). There is two problems that exist with it and our implementation however:

1. We make use of "private" APIs: https://github.com/ionic-team/ionic-framework/blob/main/core/src/utils/native/status-bar.ts#L24
2. `window.Capacitor.Plugins` is considered deprecated on the public interface, but is treated as stable on the internal type definition. It doesn't seem that Capacitor has intentions of removing that property in the future.

I think there is risk in both directions:
- Using the internal type does mean they could move or change the type. This would result in a build error when updating the dependency. It should not result in shipped regressions. 
- Using the public type, we would need to disable type checking for the areas of the implementation that specifically use private APIs (just status bar right now). This puts us into a similar scenario as exists on main, where that API could be removed or changed and we would not know about it until someone reported an issue.

liamdebeasi: If we used `CapacitorGlobal` we could use `CapacitorGlobal.isPluginAvailable` it sounds like: https://github.com/ionic-team/capacitor/blob/06fc5f1d819ffaffb57423a3f459449b0ec08c0f/core/src/definitions.ts#L31C3-L31C20

This would let us use public definitions for all but [the status bar](https://github.com/ionic-team/ionic-framework/blob/096d9cc931dfe3a1b4b39d6652112ef209be7147/core/src/utils/native/status-bar.ts#L24). As for that line, we can continue to use `win.Capacitor.PluginHeaders` and then make a tech debt ticket to officially drop support for older Capacitor versions in Ionic 8 or something similar.

sean-perkins: Sounds like a good compromise 👍 I'll create a ticket to track. We will effectively be dropping support for the status bar appearance automatically updating back to the default style when using the card modal on Capacitor 2 with Ionic v8.

core/src/utils/native/haptic.ts
```diff
@@ -1,24 +1,90 @@
...
 
 const HapticEngine = {
-  getEngine() {
-    const win = window as any;
-    return win.TapticEngine || (win.Capacitor?.isPluginAvailable('Haptics') && win.Capacitor.Plugins.Haptics);
+  getEngine(): HapticsPlugin | undefined {
+    const tapticEngine = (window as any).TapticEngine;
+    if (tapticEngine) {
+      // Cordova
+      return tapticEngine;
```
liamdebeasi, line 70: can we cast this as `HapticsPlugin` too?

core/src/utils/native/haptic.ts
```diff
@@ -28,37 +94,34 @@ const HapticEngine = {
...
-    if (win.Capacitor?.getPlatform() === 'web') {
+    if (capacitor?.getPlatform() === 'web') {
       return typeof navigator !== 'undefined' && navigator.vibrate !== undefined;
     }
 
     return true;
   },
   isCordova() {
-    return !!(window as any).TapticEngine;
+    return (window as any).TapticEngine !== undefined;
```
liamdebeasi, line 104: Are we confident that `TapticEngine` is undefined when unset? (as opposed to null or `''`). Is this documented somewhere?

core/src/utils/native/capacitor.ts
```diff
@@ -0,0 +1,13 @@
+import type { CapacitorGlobal } from '@capacitor/core';
+
+let capacitor: CapacitorGlobal | undefined;
+
+export const getCapacitor = () => {
+  if (capacitor !== undefined) {
+    return capacitor;
+  }
+  if (window !== undefined) {
```
liamdebeasi, line 9: We should use the `win` util otherwise this will throw in SSR: https://github.com/ionic-team/ionic-framework/blob/82c568b8c8e1e9934e1928452aa5216619290e7b/core/src/utils/browser/index.ts#L23

core/src/utils/native/capacitor.ts
```diff
@@ -0,0 +1,13 @@
+import type { CapacitorGlobal } from '@capacitor/core';
+
+let capacitor: CapacitorGlobal | undefined;
```
liamdebeasi, line 3: If we are querying `window.Capacitor` do we even need to cache the value here?

core/src/utils/native/haptic.ts
```diff
@@ -28,37 +94,34 @@ const HapticEngine = {
...
+    return getCapacitor() !== undefined;
   },
   impact(options: HapticImpactOptions) {
     const engine = this.getEngine();
     if (!engine) {
       return;
     }
-    const style = this.isCapacitor() ? options.style.toUpperCase() : options.style;
-    engine.impact({ style });
+    engine.impact({ style: options.style });
```
liamdebeasi, line 114: The Cordova plugin uses all lowercase keywords, so we might need to do `toLowerCase()` for compatibility: https://github.com/EddyVerbruggen/cordova-plugin-taptic-engine

We should also add a tech debt ticket to remove that support since the other native integrations we used only support Capacitor.

core/src/utils/native/haptic.ts
```diff
@@ -68,7 +131,7 @@ const HapticEngine = {
     if (this.isCapacitor()) {
       engine.selectionStart();
     } else {
-      engine.gestureSelectionStart();
+      (engine as unknown as TapticEngine).gestureSelectionStart();
```
liamdebeasi, line 134: Why do we need to cast as `unknown` first?

core/src/utils/native/haptic.ts
sean-perkins: When `getEngine` is used, it will be of type `HapticsPlugin` (or `undefined`), regardless if it is from the taptic engine or the Capacitor haptics plugin. Manually casting won't change the behavior since we have an explicit return type on the method.

core/src/utils/native/haptic.ts
sean-perkins: When converting between an explicit type signature to another type signature that doesn't overlap, you need to cast to `unknown` which is the equivalent of "I know what I am doing" to disable Typescript errors.

In this case, the `engine` is Cordova's taptic engine, but the return type of `getEngine` is `HapticsEngine`. This strategy was used to get explicit type checking for the Cordova usage.

core/src/utils/native/capacitor.ts
sean-perkins: I still think this feedback is useful and we should use the `win` util to avoid any unexpected issues in the future. Addressed here: https://github.com/ionic-team/ionic-framework/pull/27755/commits/07db5fea3dca832c952aa8e1a747dd6e51f3be60

core/src/utils/native/capacitor.ts
sean-perkins: Removed here: https://github.com/ionic-team/ionic-framework/pull/27755/commits/ad7c02c2d5e5e91e5cb957b213c0b166c34fc6b9 was probably a premature optimization.

core/src/utils/native/haptic.ts
sean-perkins: Re-added the fallback behavior here: https://github.com/ionic-team/ionic-framework/pull/27755/commits/a2339cbfd8afdc7b1e6cb4afb451a89a9335cc9d

Also created a ticket to track removing the legacy Cordova integrations for the haptics plugin.

core/package.json
sean-perkins: https://github.com/ionic-team/ionic-framework/pull/27755/commits/263604cd991e8a69492de0b2e5a964ec2cb641ac

core/src/utils/native/keyboard.ts
```diff
@@ -1,30 +1,60 @@
...
+  /**
+   * Neither the app nor the Web View are resized.
+   *
+   * @since 1.0.0
+   */
   None = 'none',
 }
 
+// Interfaces source: https://capacitorjs.com/docs/apis/keyboard#interfaces
+export interface KeyboardResizeOptions {
```
liamdebeasi, line 38: Are we able to import this from Capacitor too?

core/src/utils/native/haptic.ts
liamdebeasi: Thanks! Will this no longer be needed once Cordova support for the haptic plugin is dropped?

liamdebeasi (commented):
I'm going to test the dev build on my end too, but some final comments on the code.

liamdebeasi (approved):
Tested locally and it works great! Tested with the datetime wheel picker for haptics as well as when Capacitor (but not the plugins) are available. I left a couple comments, but this is good to merge once those are addressed.

core/src/utils/native/haptic.ts
sean-perkins: Correct! We can remove this fallback and the `TapticEngine` type once the Cordova support is removed.

PR #27759, opened 2023-07-06
feat(breadcrumb): support dynamic type

liamdebeasi:
Issue number: N/A

---------

Breadcrumb does not support dynamic type

## What is the new behavior?

- Convert fonts to rem
- Add maximum font size for iOS to prioritize main content. Note that 22px was chosen to match the [back button max font size](https://github.com/ionic-team/ionic-framework/pull/27750) since breadcrumbs can be used in a toolbar.

- [ ] Yes
- [x] No

## Other information

Demo Video: https://github.com/ionic-team/ionic-framework/assets/2721089/3e59160b-e18e-4f96-975d-f9dc7f8a49d4

### DIFF
--- a/core/src/components/breadcrumb/breadcrumb.ios.scss
+++ b/core/src/components/breadcrumb/breadcrumb.ios.scss
@@ -14,6 +14,8 @@
   /**
    * Main content should be prioritized
    * on iOS, so we set max font size for breadcrumbs.
+   * Breadcrumbs can be placed in the content too, so
+   * we add a min font size to keep the text legible.
    */
   font-size: clamp(16px, #{$breadcrumb-font-size}, 22px);
 }
--- a/core/src/components/breadcrumb/breadcrumb.ios.scss
+++ b/core/src/components/breadcrumb/breadcrumb.ios.scss
@@ -10,6 +10,12 @@
   --color-hover: #{$breadcrumb-ios-color-active};
   --color-focused: var(--color-active);
   --background-focused: #{$breadcrumb-ios-background-focused};
+
+  /**
+   * Main content should be prioritized
+   * on iOS, so we set max font size for breadcrumbs.
+   */
+  font-size: min(#{$breadcrumb-font-size}, 22px);
 }
 
 :host(.breadcrumb-active) {
@@ -58,7 +64,7 @@
 ::slotted(ion-icon) {
   color: $breadcrumb-ios-icon-color;
 
-  font-size: 18px;
+  font-size: min(1.125rem, 22px);
 }
 
 ::slotted(ion-icon[slot="start"]) {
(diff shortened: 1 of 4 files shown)


core/src/components/breadcrumb/breadcrumb.vars.scss
```diff
@@ -7,7 +7,7 @@
 $breadcrumb-font-weight:         400 !default;
 
 /// @prop - Font size of the breadcrumb
-$breadcrumb-font-size:           16px !default;
+$breadcrumb-font-size:           1rem !default;
```
brandyscarney, line 10: Do we also want to support a minimum font-size in case this value changes in the future or do we want to deal with it then? We could just copy the values back-button uses.

liamdebeasi: Good question! I don't have any strong feelings. iOS doesn't really have this UI pattern, so we can make a call. Any preference?

brandyscarney: I didn't test this out but does the `body` text have a minimum font size? Maybe we should match what that does? Or we could just make the minimum font size what it currently is (16px). I don't know what the most relatable component would be here.

liamdebeasi: Yeah good thought -- the components on iOS all have different minimum font sizes, so not sure that would work for all components. For example, chip has a minimum font size of 11px or 13px or something similar: https://developer.apple.com/design/human-interface-guidelines/typography#Specifications

iOS also has some built-in minimum font sizes. It looks like the "Body" font size does not go smaller than 14px on iOS.

brandyscarney: Oh yeah I was thinking maybe just for breadcrumbs we match the body minimum since it could be placed in the body too.

liamdebeasi: Fixed in https://github.com/ionic-team/ionic-framework/pull/27759/commits/7227630495891d1f0a7cf826e8323c072043021e

core/src/components/breadcrumb/breadcrumb.ios.scss
```diff
@@ -10,6 +10,12 @@
   --color-hover: #{$breadcrumb-ios-color-active};
   --color-focused: var(--color-active);
   --background-focused: #{$breadcrumb-ios-background-focused};
+
+  /**
+   * Main content should be prioritized
+   * on iOS, so we set max font size for breadcrumbs.
```
brandyscarney, line 16: Can we update this comment too to mention the min font size now?

liamdebeasi: Addressed in https://github.com/ionic-team/ionic-framework/pull/27759/commits/11c4ff9ca9d364d6f04c38bc8196b3e468a003a4

brandyscarney (approved):

sean-perkins (approved):

Issue #27762, opened 2023-07-07
bug: stencil does not transpile tsconfig aliases on 3.4.1


mikaelkaron:
### Prerequisites

- [X] I have read the [Contributing Guidelines](https://github.com/ionic-team/ionic-framework/blob/main/.github/CONTRIBUTING.md#creating-an-issue).
- [X] I agree to follow the [Code of Conduct](https://ionicframework.com/code-of-conduct).
- [X] I have searched for [existing issues](https://github.com/ionic-team/ionic-framework/issues) that already report this problem, without success.

### Ionic Framework Version

v7.x

### Current Behavior

when upgrading to `7.1.2` I get a bunch of errors similar to this one during build

```
[ WARN  ]  Bundling Warning UNRESOLVED_IMPORT
           '@utils/helpers' is imported by ../../node_modules/@ionic/core/dist/collection/components/input/input.js,
           but could not be resolved – treating it as an external dependency
```

### Expected Behavior

no errors to display

### Steps to Reproduce

- install `@stencil/core@3.4.1`
- install `@ionic/core@7.1.2`
- import `@ionic/core` in global  `app.ts`
- `stencil build`

### Ionic Info

n/a (not using ionic CLI)

### Additional Information

this is a new regression in `7.1.2`, downgrading to `7.1.1` fixes the problem

liamdebeasi:
Thanks for the report. It looks like there was a Stencil regression that caused `@utils/helpers` to not be transpiled to the correct relative path on build. I'll get in touch with the Stencil team to see if we can get this fixed on their end.

liamdebeasi:
I am going to downgrade Stencil to 3.4.0 for now until the issue can be resolved on Stencil's side. Here is a dev build with the fix: `7.1.3-dev.11688736787.10e452c1`

liamdebeasi:
The PR has been merged, and the next release of Ionic will have this fix. In the meantime, you can either downgrade to 7.1.1 or use the dev build I provided in https://github.com/ionic-team/ionic-framework/issues/27762#issuecomment-1625432351. Thanks for reporting this!

PR #27767, opened 2023-07-07
chore: re-land stencil 3.4.1

liamdebeasi:
Issue number: N/A

---------

In https://github.com/ionic-team/ionic-framework/issues/27762 it was discovered that Stencil 3.4.1 caused aliased paths in `dist/collection` to no longer be transpiled on build. As part of https://github.com/ionic-team/stencil/pull/4501 the Stencil team fixed a bug which caused paths in dist-collection to (correctly) no longer be transpiled unless `transformAliasedImportPathsInCollection` is set to `true`. This flag is `false` by default, so upon upgrade our the paths in dist-collection stopped being transpiled.

## What is the new behavior?

- Adds `transformAliasedImportPathsInCollection: true` to the `dist` output target in the stencil config. (dist-collection is part of dist in this case)
- Re-upgrades Ionic to Stencil 3.4.1

- [ ] Yes
- [x] No

## Other information

Dev Build: `7.1.3-dev.11688751672.167a6cdd`

Reviewers can manually verify the built changes by looking in `core/dist/collection/components/` and then looking at a JS file for a component that uses aliased paths (such as datetime-button)

### DIFF
--- a/core/package-lock.json
+++ b/core/package-lock.json
@@ -9,7 +9,7 @@
       "version": "7.1.2",
       "license": "MIT",
       "dependencies": {
-        "@stencil/core": "^3.4.0",
+        "@stencil/core": "^3.4.1",
         "ionicons": "7.1.0",
         "tslib": "^2.1.0"
       },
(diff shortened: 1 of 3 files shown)


liamdebeasi:
https://github.com/ionic-team/ionic-framework/blob/2085025644f075e63d04bece56eca4f2beeadbb6/core/src/utils/slot-mutation-controller.ts#L2 still is not being transpiled correctly. It gets transpiled to `"helpers"` instead of `"./helpers"`.

liamdebeasi:
Closing until https://github.com/ionic-team/ionic-framework/pull/27767#issuecomment-1625746820 is resolved in Stencil.

Issue #27498, opened 2023-05-17
bug: slotted labels in radio, checkbox, and toggle are missing margin


DavidStrausz:
### Prerequisites

- [X] I have read the [Contributing Guidelines](https://github.com/ionic-team/ionic-framework/blob/main/.github/CONTRIBUTING.md#creating-an-issue).
- [X] I agree to follow the [Code of Conduct](https://ionicframework.com/code-of-conduct).
- [X] I have searched for [existing issues](https://github.com/ionic-team/ionic-framework/issues) that already report this problem, without success.

### Ionic Framework Version

v7.x

### Current Behavior

When using the modern syntax for inputs (e.g. `ion-radio`, `ion-checkbox`) the nice spacing inside an `ion-item` is not added anymore (as compared to the "legacy" syntax). This is an issue especially with long labels (`ion-text-wrap`).

| ios | md |
|--------|--------|
| <img width="190" alt="ion-item-spacing" src="https://github.com/ionic-team/ionic-framework/assets/7592254/9cbe81f2-39be-4f6c-8b61-6c5988e8a41c"> | <img width="200" alt="ion-item-spacing-md" src="https://github.com/ionic-team/ionic-framework/assets/7592254/2333d9f7-d893-418b-bca9-4d898e3aba99"> |

### Expected Behavior

I'd expect it to look the same as in previous versions of Ionic.

### Steps to Reproduce

1. Run reproduction
2. Check differences in `ion-item` spacing in legacy and modern syntax

### Code Reproduction URL

https://github.com/DavidStrausz/modern-input

### Ionic Info

```
Ionic:

   Ionic CLI                     : 7.1.1 (C:\Users\dstrauss\AppData\Roaming\npm\node_modules\@ionic\cli)
   Ionic Framework               : @ionic/angular 7.0.7
   @angular-devkit/build-angular : 15.2.8
   @angular-devkit/schematics    : 15.2.8
   @angular/cli                  : 15.2.8
   @ionic/angular-toolkit        : 9.0.0

Capacitor:

   Capacitor CLI      : 5.0.3
   @capacitor/android : not installed
   @capacitor/core    : 5.0.3
   @capacitor/ios     : not installed

Utility:

   cordova-res : 0.15.4
   native-run  : 1.7.2

System:

   NodeJS : v18.12.1 (C:\Program Files\nodejs\node.exe)
   npm    : 9.4.1
   OS     : Windows 10
```

### Additional Information

[...]

PR #27771, opened 2023-07-10
fix(radio): add top and bottom margins when in ion-item

amandaejohnston:
Issue number: Part of #27498

---------

When `ion-radio` is used within `ion-item`, top and bottom margins are not added to the label, while they are present when using the legacy syntax. We didn't catch this because the issue is only visible when using a label that breaks onto more than one line; otherwise, the height of the item exceeds what the margins would've added.

## What is the new behavior?

Margins added. Values were taken from `ion-item` styles [here](https://github.com/ionic-team/ionic-framework/blob/096d9cc931dfe3a1b4b39d6652112ef209be7147/core/src/components/item/item.ios.scss#L203-L206) and [here](https://github.com/ionic-team/ionic-framework/blob/096d9cc931dfe3a1b4b39d6652112ef209be7147/core/src/components/item/item.md.scss#L298-L301), which no longer get applied because the `ion-label` is nested in an additional shadow component. Note that left/right margins are already included in the modern syntax labels.

- [ ] Yes
- [x] No


core/src/components/radio/radio.scss
```diff
@@ -175,6 +181,10 @@ input {
   @include margin(0, $form-control-label-margin, 0, 0);
 }
 
+:host(.radio-label-placement-start.in-item) .label-text-wrapper {
+  @include margin($radio-item-label-margin-top, $form-control-label-margin, $radio-item-label-margin-bottom, 0);
```
liamdebeasi, line 185: It might be easier to do this:

```scss
:host(.radio-label-placement-end) .label-text-wrapper {
  @include margin(null, null, null, $form-control-label-margin);
}

:host(.in-item:not(.legacy-radio)) .label-text-wrapper {
  @include margin($radio-item-label-margin-top, null, $radio-item-label-margin-bottom, null);
}
```

When you use `null` the margin values do not get set (rather than getting explicitly set to `0`). This approach lets you add the top and bottom margins without re-defining the start/end margins for each label placement.

amandaejohnston: Ooh, good tip. I think I can actually remove the extra `.in-item` rules, since this lets me avoid overwriting the new `margin-top`/`margin-bottom` further up.

core/src/components/radio/radio.scss
```diff
@@ -122,6 +123,11 @@ input {
   overflow: hidden;
 }
 
+:host(.in-item:not(.legacy-radio)) .label-text-wrapper {
+  margin-top: $radio-item-label-margin-top;
+  margin-bottom: $radio-item-label-margin-bottom;
```
liamdebeasi, line 128:
```suggestion
  @include margin($radio-item-label-margin-top, null, $radio-item-label-margin-bottom, null);
```
Optional: Does the same thing, but now the code is consistent with other margin usage

liamdebeasi (approved):
One optional change, but otherwise good to go. Great job!

PR #27783, opened 2023-07-12
fix(checkbox): add top and bottom margins when in ion-item

amandaejohnston:
Issue number: Part of #27498

---------

When `ion-checkbox` is used within `ion-item`, top and bottom margins are not added to the label, while they are present when using the legacy syntax. We didn't catch this because the issue is only visible when using a label that breaks onto more than one line; otherwise, the height of the item exceeds what the margins would've added.

## What is the new behavior?

Margins added. Values were taken from `ion-item` styles [here](https://github.com/ionic-team/ionic-framework/blob/096d9cc931dfe3a1b4b39d6652112ef209be7147/core/src/components/item/item.ios.scss#L203-L206) and [here](https://github.com/ionic-team/ionic-framework/blob/096d9cc931dfe3a1b4b39d6652112ef209be7147/core/src/components/item/item.md.scss#L298-L301), which no longer get applied because the `ion-label` is nested in an additional shadow component. Note that left/right margins are already included in the modern syntax labels.

- [ ] Yes
- [x] No


sean-perkins (approved):
Changes look good 👍 can we update the PR description to include the same verbiage as the linked PR, but about `ion-checkbox`?

PR #27789, opened 2023-07-12
chore(toggle): rename item test to use correct component name

amandaejohnston:
Issue number: #

---------

## What is the new behavior?

-
-
-

- [ ] Yes
- [ ] No


sean-perkins (approved):
> Unsure why Github detects some of these as removals/additions; there should be no changes aside from renames.

When renaming the file, did you additional use `git mv` for the snapshot files to relocate them to the new file directory? `label.e2e.ts-snapshots` -> `toggle.e2e.ts-snapshots`

If not, that is likely why the deletions/additions are being tracked. 

I don't think it is a big deal either way.

liamdebeasi:
You may need to use `-f` as well since these files are gitignored to avoid accidentally overwriting them.

amandaejohnston:
Aha, `git mv -f` did the trick 👍

PR #27790, opened 2023-07-12
fix(button): submit form when pressing enter key

### LINKED ISSUE #19368: bug: IonButton type="submit" does not submit form when enter pressed in inputs
# Bug Report

**Ionic version:**
[x] @ionic/react 4.9.0-rc.2

**Current behavior:**
`IonButtons` created with `type="submit"` do not submit the forms when a user presses the "enter" key in an input field.

You can see that the generated Shadow DOM input does NOT contain the `type="submit"` attribute.

IonButton Submit

**Expected behavior:**
Standard HTML `button`s created with `type="submit"` will automatically attempt to submit the form when the user presses "enter" in an input field.  For accessibility, `IonButtons` inside a form should do the same.

**Steps to reproduce:**
In the form below, press the "enter" key and observe that the form does not attempt to submit.
**Related code:**

```
<form onSubmit={handleSubmit1}>
  <IonItem>
    <IonLabel>Ionic Input</IonLabel>
    <IonInput
      name="ionicInput1"
      id="ionicInput1"
      placeholder="Press enter in this field"
    />
  </IonItem>
  <IonItem>
    <label for="genericInput1">Generic Input</label>
    <input
      type="text"
      name="genericInput1"
      id="genericInput1"
      placeholder="Press enter in this field"
    />
  </IonItem>
  <IonButton type="submit">Submit</IonButton>
</form>
```

Working sample that demonstrate a regular `<button type="submit">Submit</button>` does properly submit the form and that an `IonButton type="submit"/>` does not:
- https://codesandbox.io/s/ionic-react-rc2-inputs-submit-button-doesnt-work-issue-19368-pzdtu

**Other information:**
None

**Ionic info:**

```
Ionic:

   Ionic CLI       : 5.2.5 (/Users/jn/.nvm/versions/node/v10.16.3/lib/node_modules/ionic)
   Ionic Framework : @ionic/react 4.9.0-rc.2

Utility:

   cordova-res : not installed
   native-run  : not installed

System:

   NodeJS : v10.16.3 (/Users/jn/.nvm/versions/node/v10.16.3/bin/node)
   npm    : 6.9.0
   OS     : macOS Mojave
```

thetaPC:
Issue number: resolves #19368 

---------

The form submits when clicking on the `ion-button`. However, users cannot submit the form when focused on a form element and pressing the `enter` button. This does not follow the behavior of the native button on a form.

## What is the new behavior?

- Form now submits when a form element is focused and the `enter` button is pressed.
  - It also submits regardless of the amount of form elements present.
- Form will not submit if the `ion-button` is disabled.

- [ ] Yes
- [x] No

## Other information

N/A

### DIFF
--- a/core/src/components/button/button.tsx
+++ b/core/src/components/button/button.tsx
@@ -1,5 +1,5 @@
 import type { ComponentInterface, EventEmitter } from '@stencil/core';
-import { Component, Element, Event, Host, Prop, h } from '@stencil/core';
+import { Component, Element, Event, Host, Prop, Watch, h } from '@stencil/core';
 import type { AnchorInterface, ButtonInterface } from '@utils/element-interface';
 import type { Attributes } from '@utils/helpers';
 import { inheritAriaAttributes, hasShadowDom } from '@utils/helpers';
@@ -32,6 +32,8 @@ export class Button implements ComponentInterface, AnchorInterface, ButtonInterf
   private inItem = false;
   private inListHeader = false;
   private inToolbar = false;
+  private formButtonEl: HTMLButtonElement | null = null;
+  private formEl: HTMLFormElement | null = null;
   private inheritedAttributes: Attributes = {};
 
   @Element() el!: HTMLElement;
@@ -52,6 +54,13 @@ export class Button implements ComponentInterface, AnchorInterface, ButtonInterf
    * If `true`, the user cannot interact with the button.
    */
   @Prop({ reflect: true }) disabled = false;
+  @Watch('disabled')
+  async disabledChanged() {
+    const { disabled } = this;
+    if (this.formButtonEl) {
+      this.formButtonEl.disabled = disabled;
+    }
+  }
 
   /**
    * Set to `"block"` for a full-width button or to `"full"` for a full-width button
@@ -144,6 +153,22 @@ export class Button implements ComponentInterface, AnchorInterface, ButtonInterf
    */
   @Event() ionBlur!: EventEmitter<void>;
 
+  connectedCallback(): void {
+    // Allow form to be submitted through `ion-button`
+    if (this.type !== 'button' && hasShadowDom(this.el)) {
+      this.formEl = this.findForm();
+      if (this.formEl) {
+        // Create a hidden native button inside of the form
+        this.formButtonEl = document.createElement('button');
+        this.formButtonEl.type = this.type;
+        this.formButtonEl.style.display = 'none';
+        // Only submit if the button is not disabled.
+        this.formButtonEl.disabled = this.disabled;
+        this.formEl.appendChild(this.formButtonEl);
+      }
+    }
+  }
+
   componentWillLoad() {
     this.inToolbar = !!this.el.closest('ion-buttons');
     this.inListHeader = !!this.el.closest('ion-list-header');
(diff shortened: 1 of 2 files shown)


core/src/components/button/button.tsx
```diff
@@ -52,6 +54,13 @@ export class Button implements ComponentInterface, AnchorInterface, ButtonInterf
    * If `true`, the user cannot interact with the button.
    */
   @Prop({ reflect: true }) disabled = false;
+  @Watch('disabled')
+  async disabledChanged() {
```
sean-perkins, line 58: We can remove `async` here.

core/src/components/button/button.tsx
```diff
@@ -182,57 +207,48 @@ export class Button implements ComponentInterface, AnchorInterface, ButtonInterf
...
       }
     }
-    return null;
+    if (form !== undefined) {
+      /**
+       * The developer specified a form selector for
+       * the button to submit, but it was not found.
+       */
+      if (typeof form === 'string') {
+        printIonWarning(
```
sean-perkins, line 216: I don't believe this warning prints under the conditions it specifies. Let me know if I misread it, this is what I am observing:

If the `form` is a string, but the matched element is not an `HTMLFormElement`, this warning would log.

This statement would not return early, but would match the current conditionals for this warning: https://github.com/ionic-team/ionic-framework/blob/85058cc1cf9fab4ec979224cb6a5e1437a664b58/core/src/components/button/button.tsx#L204-L209

I think instead we want to check here: https://github.com/ionic-team/ionic-framework/blob/85058cc1cf9fab4ec979224cb6a5e1437a664b58/core/src/components/button/button.tsx#L205 if `el` is `null`/`undefined` log this warning. 

I think then the message for the `form` element being invalid can be the default behavior if neither of these conditions returns: https://github.com/ionic-team/ionic-framework/blob/85058cc1cf9fab4ec979224cb6a5e1437a664b58/core/src/components/button/button.tsx#L201-L209

thetaPC: I updated the code and added comments.

core/src/components/button/button.tsx
```diff
@@ -177,62 +202,58 @@ export class Button implements ComponentInterface, AnchorInterface, ButtonInterf
...
       const el = document.getElementById(form);
       if (el instanceof HTMLFormElement) {
         return el;
       }
+      /**
+       * The developer specified a string for the form attribute, but the
+       * form element with that id does not exist in the DOM.
+       */
+      printIonWarning(
+        `Form with selector: "#${form}" could not be found. Verify that the id is correct and the form is rendered in the DOM.`,
```
sean-perkins, line 215: `el` could be either defined and not an instance of `HTMLFormElement` or not found in the DOM at this point. 

I'm thinking something like this, thoughts?

```ts
private findForm(): HTMLFormElement | null {
  const { form } = this;
  if (form instanceof HTMLFormElement) {
    return form;
  }
  if (typeof form === 'string') {
    // Check if the string provided is a form id.
    const el = document.getElementById(form);
    if (el) {
      if (el instanceof HTMLFormElement) {
        return el;
      } else {
        // warn about not an instance of HTMLFormElement
        return null;
      }
    } else {
      // warn about form id not found in document
      return null;
    }
  }
  /**
   * If the form element is not set, the button may be inside
   * of a form element. Query the closest form element to the button.
   */
  return this.el.closest('form');
}
```

thetaPC: I like this. It provides more in-depth warnings. Just one question, are you proposing to not include the following [edge case](https://github.com/ionic-team/ionic-framework/pull/27790/files/83d04bd4ef7991c39095e0d0da1af529cbcc86e2#diff-8b0e185d2b3a4ceee4ec690a5f8ff2992ef3720594562f34bf2cae17f05254ffR220-R232)?

sean-perkins: Let's leave it in. I checked against the browser behavior here: https://codepen.io/sean-perkins/pen/rNQJozW

If you pass an "invalid" `form` value it does not submit the closest form.

joshangehr:
This is awesome, thanks so much!

Issue #23560, opened 2021-07-01
bug: Datetime label change should sync with transition


willmartian:
# Bug Report

**Ionic version:**

[ ] **4.x**
[ ] **5.x**
[x] **6.x**

**Current behavior:**
When changing months by clicking the next/previous buttons, the component waits for the month change animation to finish before updating the state of the month and year label.

In the attached video, observe how navigating with keyboard updates the month and year label instantly. Compare this to the second half of the video, where the month is changed by click.

https://user-images.githubusercontent.com/17113462/124149453-98c8a780-da5e-11eb-9b17-6ee6bee4c912.mov

**Expected behavior:**
I would expect the month and year label to change _with_ the month content, not after. 

This is in line with other specs and implementations: 
- https://material.io/components/date-pickers#desktop-pickers
- https://vuetifyjs.com/en/components/date-pickers/#usage

**Steps to reproduce:**

Open a `ion-datetime` component and navigate by click.

**Ionic info:**

Ionic:

   Ionic CLI : 6.13.1

Utility:

   cordova-res : 0.15.3
   native-run  : not installed

System:

   NodeJS : v14.17.1
   npm    : 6.14.13
   OS     : macOS Big Sur

  Firefox, Desktop

liamdebeasi:
Thanks for the report. The team does not have plans to change this behavior in Datetime at the moment, so I am going to close this.

PR #27806, opened 2023-07-14
fix(datetime): scroll to newly selected date when value changes

### LINKED ISSUE #26391: bug: Setting ion-datetime value programatically does not update the grid
### Prerequisites

### Ionic Framework Version

- [ ] v4.x
- [ ] v5.x
- [X] v6.x
- [ ] Nightly

### Current Behavior

If you change the value of an ion-datetime programatically, and then open the ion-datetime, the calendar grid still displays the month/year of the old date.

E.g. if you programatically change the date from today to May 5, 2022, and then open the ion-datetime, the calendar grid will show today's month instead of May 2022 (if you do then navigate to May 2022, you will see that May 5 is correctly selected, but it should have opened there in the first place)

### Expected Behavior

Upon opening the ion-datetime, I would expect the calendar grid to initially display the month/year of the selected date.

### Steps to Reproduce

1. Start with an ion-datetime, which defaults to today's date
2. After 4 seconds, programatically update the date to May 5, 2022
3. Open the date picker
The date picker shows the current month, instead of showing May 2022
(if you navigate back to May 2022, you will see that May 5 is highlighted as expected, but the date picker should have opened in that month to start with)

### Code Reproduction URL

https://stackblitz.com/edit/angular-ramjpp-zz9vub?file=package.json

### Ionic Info

Stackblitz

### Additional Information

Prior to 6.3.2, the selected date was not updated either. This was reported in https://github.com/ionic-team/ionic-framework/issues/25776, and I have used that as a template for this issue. Although the selected date does now get updated, the calendar grid does not open to the correct month, hence this new issue.

amandaejohnston:
Issue number: Resolves #26391

---------

When updating the `value` programmatically on an `ion-datetime` after it has already been created:
- With grid style: The selected date visually updates, but the calendar does not scroll to the newly selected month.
- With wheel style: The selected date does not visually update, i.e. the wheels do not move to show the newly selected date.

## What is the new behavior?

- Grid style datetimes now scroll to the selected date using the same animation as when clicking the next/prev month buttons.
  - This animation mirrors the behavior in both MUI and native iOS. See the [design doc](https://github.com/ionic-team/ionic-framework-design-documents/blob/main/projects/ionic-framework/components/datetime/0003-datetime-async-value.md) for more information and screen recordings.
  - The animation will not occur if the month/year did not change, or when the datetime is hidden.
- Wheel style datetimes now visually update to the selected date. No animation occurs, also mirroring native.
- The `parseDate` util has also had its type signatures updated to account for returning `undefined` when the date string is improperly formatted. This was missed when the util was refactored to support multiple date selection.

- [ ] Yes
- [x] No

## Other information

- Docs PR: https://github.com/ionic-team/ionic-docs/pull/3053
- While this can technically be considered a bug fix, we are merging it into a feature branch for safety; it's a fairly significant change to how datetime behaves, and may interfere with custom logic when updating a datetime's value async.
- Jumping to the newly selected value is handled by replacing everything [here](https://github.com/ionic-team/ionic-framework/pull/27806/files#diff-4a407530c60e3cf72bcc11acdd21c4803a94bf47ea81b99e757db1c93d2735b8L364-L407) with `processValue()`. This covers both wheel and grid datetimes.
- `activePartsClone` as a whole was also removed. It was added in https://github.com/ionic-team/ionic-framework/pull/24244 to enable changing `activeParts` without triggering a rerender (and thus jumping to the new value) but since we now want to do that jump, the clone is no longer needed.
- The animation code might be tricky to follow, so I recorded going through it: https://github.com/ionic-team/ionic-framework/assets/90629384/1afa5762-f493-441a-b662-f0429f2d86a7


core/src/components/datetime/datetime.tsx
```diff
@@ -1262,6 +1236,58 @@ export class Datetime implements ComponentInterface {
...
+      this.forceRenderDate = { month, year, day };
+
+      /**
+       * Flag that we've started scrolling to the forced date.
+       * The resolve function will be called by the datetime's
+       * scroll listener when it's done updating everything.
+       * This is a replacement for making prev/nextMonth async,
+       * since the logic we're waiting on is in a listener.
+       */
+      const forceDateScrollingPromise: Promise<void> = new Promise((resolve) => {
```
sean-perkins, line 1264:
```suggestion
      const forceDateScrollingPromise = new Promise<void>((resolve) => {
```

sean-perkins (commented):
I'm still in progress of manually testing. Dropping some early suggestions for changes.

core/src/components/datetime/datetime.tsx
```diff
@@ -1262,6 +1236,58 @@ export class Datetime implements ComponentInterface {
...
     }
+
+    /**
+     * Only animate if:
+     * 1. We're using grid style (wheel style pickers should just jump to new value)
+     * 2. The month and/or year actually changed (otherwise there's nothing to animate to)
+     * 3. The datetime is visible (prevents animation when in collapsed datetime-button, for example)
+     */
+    const didChangeMonth = month !== workingParts.month || year !== workingParts.year;
+    const elIsVisible = el.offsetParent !== null;
```
liamdebeasi, line 1247: The view does not update if the month/year picker is open on the grid style

core/src/components/datetime/utils/parse.ts
```diff
@@ -85,6 +87,7 @@ export function parseDate(val: string | string[] | undefined | null): DatetimePa
 
   if (parse === null) {
     // wasn't able to parse the ISO datetime
+    printIonWarning(`Unable to parse date string: ${val}. Please provide a valid ISO 8601 datetime string.`);
```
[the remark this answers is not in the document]
sean-perkins: One thing I am observing with assigning an invalid value, is this message doesn't occur.

e.g.:

Assigning a value of `2021-04-20T14:23:00.000z` (lowercase `z` instead of `Z`)

Results in this error:
> Cannot destructure property 'month' of 'targetValue' as it is undefined.

core/src/components/datetime/datetime.tsx
```diff
@@ -1262,6 +1236,58 @@ export class Datetime implements ComponentInterface {
...
+      targetMonthIsEarlier ? this.prevMonth() : this.nextMonth();
+      await forceDateScrollingPromise;
+      this.resolveForceDateScrolling = undefined;
+      this.forceRenderDate = null;
+    } else {
+      /**
+       * We only need to do this if we didn't just animate to a new month,
+       * since prevMonth/nextMonth call setWorkingParts for us.
+       */
+      this.setWorkingParts({
```
liamdebeasi, line 1282: Not really sure where to put this, so I'll put it here:

When the time popover is open and the value updates, pressing "Next" seems to jump over multiple months:

https://github.com/ionic-team/ionic-framework/assets/2721089/5172aa05-04d3-49e3-81b8-c7ff38751154

core/src/components/datetime/datetime.tsx
```diff
@@ -1262,6 +1236,58 @@ export class Datetime implements ComponentInterface {
...
+       */
+      const forceDateScrollingPromise: Promise<void> = new Promise((resolve) => {
+        this.resolveForceDateScrolling = resolve;
+      });
+
+      /**
+       * Animate smoothly to the forced month. This will also update
+       * workingParts and correct the surrounding months for us.
+       */
+      const targetMonthIsEarlier = isBefore(targetValue, workingParts);
```
liamdebeasi, line 1272: Threading:

My understanding of the original issue is that asynchronously setting the datetime on load does not update the view. So if the datetime is going to be set either on load or shortly after load, will users even see this scrolling animation? For example, if a datetime in a modal has its value set as the modal opens, the scrolling would either happen on screen or as the modal transitions in.

core/src/components/datetime/datetime.tsx
sean-perkins: Similar(?) problem, if you have the time picker open on that test and change the value dynamically to several months ahead, the next button is not actionable.

In the recording I show that with the time picker collapsed, the button works. After refreshing and opening the time picker, it does not. 

https://github.com/ionic-team/ionic-framework/assets/13732623/a309db00-93d7-4662-bcfd-cb009c462df5

core/src/components/datetime/datetime.tsx
amandaejohnston: The idea is mainly to mirror native behavior, which animates on both MUI and native iOS: https://github.com/ionic-team/ionic-framework-design-documents/blob/main/projects/ionic-framework/components/datetime/0003-datetime-async-value.md#references

core/src/components/datetime/datetime.tsx
```diff
@@ -912,6 +867,29 @@ export class Datetime implements ComponentInterface {
...
+        const dataMonth = firstDayEl?.getAttribute('data-month');
+        const dataYear = firstDayEl?.getAttribute('data-year');
+        if (
+          forceRenderDate !== null &&
+          dataMonth &&
+          dataYear &&
+          parseInt(dataMonth) === forceRenderDate.month &&
+          parseInt(dataYear) === forceRenderDate.year
+        ) {
+          return { month: forceRenderDate.month, year: forceRenderDate.year, day: forceRenderDate.day };
```
liamdebeasi, line 890: What should datetime do when being set to an out of bounds value? Right now it scrolls as far as it can:

```html
<ion-datetime max="2023"></ion-datetime>
```

```js
const datetime = document.querySelector('ion-datetime');

setTimeout(() => {
  datetime.value = "2024-01-01T16:30"
}, 2000);
```

However, this brings me to December 2023 which is confusing since I set the datetime to January 2024.

core/src/components/datetime/datetime.tsx
```diff
@@ -912,6 +867,29 @@ export class Datetime implements ComponentInterface {
...
+         * automatically animated directly to it.
+         */
+        const { forceRenderDate } = this;
+        const firstDayEl = month.querySelector('.calendar-day');
+        const dataMonth = firstDayEl?.getAttribute('data-month');
+        const dataYear = firstDayEl?.getAttribute('data-year');
+        if (
+          forceRenderDate !== null &&
+          dataMonth &&
+          dataYear &&
```
liamdebeasi, line 886: If the year picker is open while the date changes, selecting a new time brings you back to the month you were last on:


https://github.com/ionic-team/ionic-framework/assets/2721089/c5ef0e66-fbd1-4f69-aa27-e361d62868e4

core/src/components/datetime/datetime.tsx
amandaejohnston: @liamdebeasi Before I get too deep in bug fixes here, I wanted to check whether you still had any concerns with moving forward with the animation. I can understand wanting to avoid making datetime even more complex, and deciding to scrap the animation part would make fixing things up here easier, but there are a few reasons why I still feel we should go through with it:

1. The animation matches native behavior, as mentioned above.
2. The datetime won't always be hidden when the value is updated async. While that will definitely be our recommendation (see https://github.com/ionic-team/ionic-docs/pull/3053), we should avoid glitchy-looking behavior when it isn't the dev's desired pattern. For example, an app could have a dropdown with appointments, and the datetime's value updates to reflect which one you've selected.
3. The animation is something we already decided on in the design doc, and I would rather commit to that than constantly be rethinking our decisions. This isn't a case where we missed something that dramatically changes the scope; we already recognized that the animation will be difficult, and decided to proceed anyway.

liamdebeasi: I agree we shouldn't omit the animation just because it is challenging. But I'd also like us to keep in mind that the primary objective is to update the UI, with the animation being a secondary goal. This behavior is similar to https://github.com/ionic-team/ionic-framework/issues/23560 where it _is_ closer to the spec, but it's not critical functionality.

I'm not advocating for removing this, but I do want to make sure we are being cautious. If you recall the old IntersectionObserver implementation in datetime, that consumed a lot of our time before we eventually abandoned that approach in favor of a scroll listener.

core/src/components/datetime/datetime.tsx
amandaejohnston: Fixed in https://github.com/ionic-team/ionic-framework/pull/27806/commits/7a3f92751372589c94d3cab63c915c64c4d60c77.

core/src/components/datetime/datetime.tsx
amandaejohnston: Fixed by https://github.com/ionic-team/ionic-framework/pull/27806/commits/7a3f92751372589c94d3cab63c915c64c4d60c77.

core/src/components/datetime/utils/parse.ts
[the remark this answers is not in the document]
amandaejohnston: Well, the message does occur, but the case isn't being handled once we get back to datetime.tsx 🤔 

console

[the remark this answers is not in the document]
amandaejohnston: Fixed in https://github.com/ionic-team/ionic-framework/pull/27806/commits/3dd1732a9ce2628dec03ae3c33ee9f68004a4e66 and https://github.com/ionic-team/ionic-framework/pull/27806/commits/607f43cb6695f428b9ba413e790d6da402e63eb7, and then cleaned up in https://github.com/ionic-team/ionic-framework/pull/27806/commits/c57c6febf4f7f4f69db64ce615b95896afeb205e, which also updates `parseDate`'s type signatures to account for this error case.

core/src/components/datetime/datetime.tsx
```diff
@@ -1290,17 +1326,19 @@ export class Datetime implements ComponentInterface {
...
     const yearValues = (this.parsedYearValues = convertToArrayOfNumbers(this.yearValues));
     const dayValues = (this.parsedDayValues = convertToArrayOfNumbers(this.dayValues));
 
-    const todayParts = (this.todayParts = parseDate(getToday()));
+    const todayParts = (this.todayParts = parseDate(getToday())!);
     this.defaultParts = getClosestValidDate(todayParts, monthValues, dayValues, yearValues, hourValues, minuteValues);
-    this.processValue(this.value);
+
+    this.processMinParts();
+    this.processMaxParts();
```
amandaejohnston, line 1339: I moved these further down to avoid an error where if a `min` or `max` is provided without a year, these functions would error out. This was because `parseMinParts` and `parseMaxParts` fall back to the current year ([link](https://github.com/ionic-team/ionic-framework/blob/f3679c132dc801c1e1aa55c28f0d94927e118f53/core/src/components/datetime/utils/parse.ts#L217)), which is passed in via `todayParts` grabbed from the datetime. Previously, `todayParts` was being set _after_ these functions were called, so it was still `undefined`.

I'm unsure why this wasn't causing problems before, but it was causing the minmax test to crash.

core/src/components/datetime/datetime.tsx
```diff
@@ -1196,14 +1178,22 @@ export class Datetime implements ComponentInterface {
     });
   }
 
-  private processValue = (value?: string | string[] | null) => {
-    const hasValue = value !== null && value !== undefined;
+  private processValue = async (value?: string | string[] | null, animate = false) => {
+    const hasValue = value !== null && value !== undefined && (!Array.isArray(value) || value.length > 0);
```
amandaejohnston, line 1182: This change avoids a crash when manually setting a multiple selection datetime's value to `[]`, by treating an empty array the same as setting the value to `undefined` (i.e. clearing the selected dates). This is done by the `multiple` test if the fixture's `goto` is called without passing in a default value.

I believe we weren't seeing the test crash before because `processValue` wasn't used when updating the value async, although I believe there was still a bug present, just with stricter circumstances.

core/src/components/datetime/datetime.tsx
amandaejohnston: Both fixed in https://github.com/ionic-team/ionic-framework/pull/27806/commits/692a270e9a458efc4263aa0c2ddf0a65328583e1.

core/src/components/datetime/utils/data.ts
```diff
@@ -254,7 +254,22 @@ export const generateTime = (
  * Given DatetimeParts, generate the previous,
  * current, and and next months.
  */
-export const generateMonths = (refParts: DatetimeParts): DatetimeParts[] => {
+export const generateMonths = (refParts: DatetimeParts, forcedDate: DatetimeParts | null = null): DatetimeParts[] => {
```
liamdebeasi, line 257: This might be better as `forcedDate?: DatetimeParts`. We call `generateMonths(workingParts)` in `renderCombinedDatePickerColumn`, so `forcedDate` is technically `undefined` here.

core/src/components/datetime/utils/data.ts
```diff
@@ -254,7 +254,22 @@ export const generateTime = (
...
+  /**
+   * If we're forcing a month to appear, and it's different from the current month,
+   * ensure it appears by replacing the next or previous month as appropriate.
+   */
+  if (forcedDate !== null && (refParts.month !== forcedDate.month || refParts.year !== forcedDate.year)) {
+    const current = { month: refParts.month, year: refParts.year, day: refParts.day };
```
liamdebeasi, line 263: Nit: You could pull `current` outside of the `if` block and use it in both return statements.

core/src/components/datetime/utils/data.ts
```diff
@@ -254,7 +254,22 @@ export const generateTime = (
  * Given DatetimeParts, generate the previous,
  * current, and and next months.
  */
-export const generateMonths = (refParts: DatetimeParts): DatetimeParts[] => {
+export const generateMonths = (refParts: DatetimeParts, forcedDate: DatetimeParts | null = null): DatetimeParts[] => {
+  /**
+   * If we're forcing a month to appear, and it's different from the current month,
+   * ensure it appears by replacing the next or previous month as appropriate.
+   */
+  if (forcedDate !== null && (refParts.month !== forcedDate.month || refParts.year !== forcedDate.year)) {
```
liamdebeasi, line 262: Is there ever a time when `refParts` could be the same as `forcedDate`? In `processValue` we check for `didChangeMonth` which should cover this.

core/src/components/datetime/datetime.tsx
```diff
@@ -1196,14 +1196,22 @@ export class Datetime implements ComponentInterface {
     });
   }
 
-  private processValue = (value?: string | string[] | null) => {
-    const hasValue = value !== null && value !== undefined;
+  private processValue = async (value?: string | string[] | null, animate = false) => {
```
liamdebeasi, line 1199: Is the `animate` param needed here?

It sounds like we only want to animate when the value is change async. In which case, it should probably animate when the `reset` method is called too.

The only other time `processValue` is called is on `componentWillLoad` which is before the component is rendered. At that point, `elIsVisible` will be `false` anyways.

core/src/components/datetime/datetime.tsx
```diff
@@ -1262,9 +1262,63 @@ export class Datetime implements ComponentInterface {
...
+    const elIsVisible = el.offsetParent !== null;
+    const { isGridStyle, showMonthAndYear } = this;
+    if (animate && isGridStyle && didChangeMonth && elIsVisible && !showMonthAndYear) {
+      /**
+       * Tell other render functions that we need to force the
+       * target month to appear in place of the actual next/prev month.
+       * Because this is a State variable, a rerender will be triggered
+       * automatically, updating the rendered months.
+       */
+      this.forceRenderDate = { month, year, day };
```
liamdebeasi, line 1283: Any reason why we used the individual pieces here when we use `targetValue` below? Wondering if we should just use `targetValue` here for consistency.

core/src/components/datetime/datetime.tsx
```diff
@@ -841,17 +841,35 @@ export class Datetime implements ComponentInterface {
...
+         * The topmost of these will be elements inside of the active month.
+         *
+         * We filter out popovers to exclude the time picker, if present,
+         * since it sits on top of the entire screen. This handles the
+         * time picker being open when the datetime's value is changed
+         * programmatically, for example.
+         */
+        const elementsAtCenter = root
+          .elementsFromPoint(box.x + box.width / 2, box.y + box.height / 2)
+          .filter((el) => el.tagName !== 'ION-POPOVER');
```
liamdebeasi, line 854: This is technically a bug in `main`, right?

core/src/components/datetime/datetime.tsx
```diff
@@ -912,6 +885,29 @@ export class Datetime implements ComponentInterface {
...
+         * is mostly for safety; in theory, if there's a forced month,
+         * that means a new value was manually set, so we should have
+         * automatically animated directly to it.
+         */
+        const { forceRenderDate } = this;
+        const firstDayEl = month.querySelector('.calendar-day');
+        const dataMonth = firstDayEl?.getAttribute('data-month');
+        const dataYear = firstDayEl?.getAttribute('data-year');
+        if (
+          forceRenderDate !== null &&
```
liamdebeasi, line 902: Is there a scenario where we could return the forced month even if it hasn't scrolled into view? Wondering if we'd be fine to just return the forced date if `forceRenderDate !== null`

liamdebeasi (changes requested):
The scrolling is working well! My focus now is mainly on cleaning up the code, but if I find any other behavior bugs I'll be sure to note them. Great work so far!

core/src/components/datetime/utils/data.ts
amandaejohnston: Yes, `generateMonths` gets called again after the scroll to the forced date occurs but before `forceRenderDate` is reset due to Stencil's render cycle. Having the check in place here ensures that the forced month is only rendered as long as it needs to be.

core/src/components/datetime/datetime.tsx
amandaejohnston: Good point 🤔 I'll remove it.

core/src/components/datetime/datetime.tsx
amandaejohnston: Good catch; I think this was leftover from an earlier draft where I was only passing in the month and year.

core/src/components/datetime/datetime.tsx
amandaejohnston: Sort of, but it's very slight. The issue arises when you try to scroll months with the time picker open. In main, this is only possible by manually calling `.click()` on the next/prev month buttons, which are in the shadow DOM, so you'd have to really be going out of your way.

core/src/components/datetime/datetime.tsx
amandaejohnston: As mentioned in the comments above this, the extra check is out of an abundance of caution. I was nervous that we might see bugs if `forceRenderDate` is ever set outside the specific circumstance of scrolling to it immediately. I can remove the extra checks if you don't feel this is useful at the moment, though.

core/src/components/datetime/datetime.tsx
liamdebeasi: Makes sense! The bug seems more noticeable with this change, so probably good to keep the fixes together.

core/src/components/datetime/datetime.tsx
```diff
@@ -1262,9 +1262,63 @@ export class Datetime implements ComponentInterface {
...
+        this.resolveForceDateScrolling = resolve;
+      });
+
+      /**
+       * Animate smoothly to the forced month. This will also update
+       * workingParts and correct the surrounding months for us.
+       */
+      const targetMonthIsBefore = isBefore(targetValue, workingParts);
+      targetMonthIsBefore ? this.prevMonth() : this.nextMonth();
+      await forceDateScrollingPromise;
```
liamdebeasi, line 1302: So I'm wondering if this should be pulled out into its own function. I'm not sure it makes sense to have `processValue` return a promise now. The purpose of that function is to grab `this.value` and process it into `DatetimeParts` (or warn if it's an invalid date). I don't think it should necessarily wait for any scrolling operations to finish since that's not required to process an ISO-8601 date.

The benefit of moving it to its own function is `processValue` doesn't need to be async anymore, but the new function cal still wait for scrolling to finish.

core/src/components/datetime-button/datetime-button.tsx
```diff
@@ -206,6 +206,10 @@ export class DatetimeButton implements ComponentInterface {
      */
     const parsedDatetimes = parseDate(parsedValues.length > 0 ? parsedValues : [getToday()]);
 
+    if (!parsedDatetimes) {
```
liamdebeasi, line 209: Just so I understand: `parseDate` has always been able to return `undefined`, we just never accounted for it, right?

core/src/components/datetime/datetime.tsx
```diff
@@ -141,6 +137,16 @@ export class Datetime implements ComponentInterface {
   @State() isPresented = false;
   @State() isTimePopoverOpen = false;
 
+  /**
+   * When non-null, will force the datetime to render the month
```
liamdebeasi, line 141:
```suggestion
   * When defined, will force the datetime to render the month
```

`forceRenderDate` doesn't use null anymore

core/src/components/datetime/datetime.tsx
```diff
@@ -886,17 +841,35 @@ export class Datetime implements ComponentInterface {
...
-         * This will be an element inside of the active month.
+         * Get the elements that are in the center of the calendar body.
+         * The topmost of these will be elements inside of the active month.
+         *
+         * We filter out popovers to exclude the time picker, if present,
+         * since it sits on top of the entire screen. This handles the
+         * time picker being open when the datetime's value is changed
+         * programmatically, for example.
+         */
+        const elementsAtCenter = root
```
liamdebeasi, line 852: FYI This might be impacted by https://github.com/ionic-team/ionic-framework/pull/27917. I found a way to accomplish this without relying on `elementFromPoint` which seemed to have a lot of edge cases.

core/src/components/datetime/datetime.tsx
liamdebeasi: I think it's worth removing for now. `forceRenderDate` (as it stands right now) should only be used as a tool to scroll to the new month when updating the `value` async. If that remains true, then we can assume that month/year has changed.

Might be worth adding a comment to `forceRenderDate` surrounding its intended usage if one does not already exist.

core/src/components/datetime/datetime.tsx
liamdebeasi: Discussed offline, but the problem here was the `elIsVisible` was `true` in `componentWillLoad` because it checked for `ion-datetime` visibility, not the calendar grid visibility. This caused the component to wait for scrolling to finish before the calendar body was visible, resulting in the scrolling to never happen.

Updating the check to `const elIsVisible = el.classList.contains('datetime-ready'); ` should fix the issue and allow us to remove the extra `animated` check.

core/src/components/datetime-button/datetime-button.tsx
amandaejohnston: Right. When setting up the type signature overloads, I missed that `undefined` gets returned if the date wasn't parsed correctly; at the time I thought this was only for `undefined`/`null` being passed in.

core/src/components/datetime/datetime.tsx
amandaejohnston: Ah, good callout. I think your change might avoid the popover issue completely 👀

core/src/components/datetime/datetime.tsx
amandaejohnston: Good call, I'll get this rearranged 👍

core/src/components/datetime-button/datetime-button.tsx
liamdebeasi: Awesome, thanks!

core/src/components/datetime/datetime.tsx
amandaejohnston: Got your changes merged in 👍

core/src/components/datetime/test/set-value/datetime.e2e.ts
```diff
@@ -39,29 +41,26 @@ configs({ modes: ['ios'], directions: ['ltr'] }).forEach(({ title, config }) =>
...
-      await monthYearButton.click();
-      await page.waitForChanges();
-
-      // Select October 2021
-      // The year will automatically switch to 2021 when selecting 10
-      await monthColumn.locator('.picker-item[data-value="10"]').click();
-      await ionChange.next();
-
-      // Close month/year picker
-      await monthYearButton.click();
```
amandaejohnston, line 58: These parts aren't needed anymore because changing the value already jumps to the new month.

core/src/components/datetime/datetime.tsx
```diff
@@ -1259,6 +1233,66 @@ export class Datetime implements ComponentInterface {
...
+     * Only animate if:
+     * 1. We're using grid style (wheel style pickers should just jump to new value)
+     * 2. The month and/or year actually changed (otherwise there's nothing to animate to)
+     * 3. The calendar body is visible (prevents animation when in collapsed datetime-button, for example)
+     * 4. The month/year picker is not open (since you wouldn't see the animation anyway)
+     */
+    const didChangeMonth = month !== workingParts.month || year !== workingParts.year;
+    const bodyIsVisible = el.classList.contains('datetime-ready');
+    const { isGridStyle, showMonthAndYear } = this;
+    if (isGridStyle && didChangeMonth && bodyIsVisible && !showMonthAndYear) {
```
liamdebeasi, line 1247: If you set datetime to a time it still scrolls to another month and then abruptly resets to the selected month.

Example:

```html
<ion-datetime value="2022-02-22T16:30:00"></ion-datetime>

<script>
setTimeout(() => {
  datetime.value = '16:30';
}, 2000);
</script>
```

liamdebeasi (commented):
Great work!

core/src/components/datetime/datetime.tsx
amandaejohnston: I fixed the wonky scrolling in https://github.com/ionic-team/ionic-framework/pull/27806/commits/fbf2f9c4c526fb0b6047e977309bbba7b2440952. The datetime now errors out if you do this, but it looks like this case (setting the value to just a time when the `presentation` is something other than just `time`) was never supported; I'm able to trigger the same error on `main` after messing with it a bit. (Screencast below.) I'm thinking it would be good to add a warning for this improper usage in a separate ticket. Thoughts?

https://github.com/ionic-team/ionic-framework/assets/90629384/0bac133b-0804-4574-aee3-0f9666535e25

core/src/components/datetime/datetime.tsx
```diff
@@ -1259,6 +1233,67 @@ export class Datetime implements ComponentInterface {
...
+
+    /**
+     * Only animate if:
+     * 1. We're using grid style (wheel style pickers should just jump to new value)
+     * 2. The month and/or year actually changed, and both are defined (otherwise there's nothing to animate to)
+     * 3. The calendar body is visible (prevents animation when in collapsed datetime-button, for example)
+     * 4. The month/year picker is not open (since you wouldn't see the animation anyway)
+     */
+    const didChangeMonth = month !== workingParts.month || year !== workingParts.year;
+    const monthYearDefined = month !== undefined && year !== undefined;
```
liamdebeasi, line 1245: Nit: I think these could be combined since we can only really consider a month/year to have changed if the month/year is defined in the first place
```suggestion
    const didChangeMonth = (month !== undefined && month !== workingParts.month) || (year !== undefined && year !== workingParts.year);
```

liamdebeasi:
Also thoughts on getting a dev build and getting the community to test?

core/src/components/datetime-button/datetime-button.tsx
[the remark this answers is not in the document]
amandaejohnston: I'm actually seeing it update; maybe I'm just not doing it quickly enough? (Screencast below.) Even if there is a specific timing that borks it, though, I'm not too worried as long as the highlighted date is correct once you do scroll to it. Seems like a small edge case that could be addressed separately.

https://github.com/ionic-team/ionic-framework/assets/90629384/1c4c16d6-e26d-40a1-bbce-9c42a31f6b00

amandaejohnston:
> Also thoughts on getting a dev build and getting the community to test?

Dev build posted in the original issue 👍

core/src/components/datetime-button/datetime-button.tsx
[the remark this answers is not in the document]
liamdebeasi: I think I had outdated code. However, another bug exists:

Switching to the month/year picker as the animation starts seems to break date selection altogether.

https://github.com/ionic-team/ionic-framework/assets/2721089/8b34875d-e525-4322-9849-8a8fce1e36f8

Tested on Chrome 116. I added the following code to `datetime/test/display/index.html` to get the datetime to update async:

```js
setTimeout(() => {
  datetime.value = '2023-10-23T16:30'
}, 2000);
```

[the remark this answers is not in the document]
amandaejohnston: I'm able to trigger that on `main` as well, using the next/prev month buttons. The animation goes through the same methods, so it's the same bug. Thoughts on addressing it separately?

https://github.com/ionic-team/ionic-framework/assets/90629384/a3265d2c-017c-40c3-8c15-9e34fd1b00be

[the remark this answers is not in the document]
liamdebeasi: Yeah that works for me 👍

liamdebeasi (approved):
Great work! 

Let's do a couple things before merging:

1. Rebase this off `feature-7.4` since we want to get this into a feature release.
2. Wait a couple more days to give devs time to provide feedback re: the dev build. If there's no response by Wednesday I think it's fine to merge. (That way we give them a full week to test)

amandaejohnston:
I'm waiting on @sean-perkins's review anyway 👀 Definitely think this is worth getting an extra pair of eyes on.

core/src/components/datetime/datetime.tsx
```diff
@@ -2401,7 +2437,7 @@ export class Datetime implements ComponentInterface {
             [`datetime-presentation-${presentation}`]: true,
             [`datetime-size-${size}`]: true,
             [`datetime-prefer-wheel`]: hasWheelVariant,
-            [`datetime-grid`]: hasGrid,
+            [`datetime-grid`]: this.isGridStyle,
```
sean-perkins, line 2440: ~ nit: Can we refactor this to `isGridStyle` to match the code style (destructure from `this`)?

core/src/components/datetime/datetime.tsx
amandaejohnston: Ticket created: https://ionic-cloud.atlassian.net/browse/FW-4996

playaz87:
Any ETA on when this fix will be released? This bug is currently blocking us from releasing an app.

liamdebeasi:
As noted in https://github.com/ionic-team/ionic-framework/issues/26391#issuecomment-1690471982, this fix will be available in an upcoming minor release of Ionic.

m1nka:
Was this feature ever released? We are experiencing the same problem. Thank you!

m1nka:
Ok, I found a workaround using [these](https://github.com/ionic-team/ionic-framework/issues/26391#issuecomment-1518131281) instructions. Using `@ViewChild('mydatepicker') mydatepicker: IonDateTime;` and the `this.mydatepicker.reset();` works fine.

I think it would make a lot of sense to reset automatically in case popover/modal is closed / reopened.

liamdebeasi:
> Was this feature ever released? We are experiencing the same problem. Thank you!

This feature was released in Ionic 7.4.0: https://ionic.io/blog/announcing-ionic-7-4

PR #27808, opened 2023-07-14
test(many): gestures flakiness

thetaPC:
Issue number: #

---------

## What is the new behavior?

-
-
-

- [ ] Yes
- [ ] No

### DIFF
--- a/core/src/utils/test/playwright/drag-element.ts
+++ b/core/src/utils/test/playwright/drag-element.ts
@@ -1,3 +1,11 @@
+/**
+ * The drag gesture will not operate as expected when the element is dragged outside of the viewport because the Mouse class does not fire events outside of the viewport.
+ *
+ * For example, if the mouse is moved outside of the viewport, then the `mouseup` event will not fire.
+ *
+ * See https://playwright.dev/docs/api/class-mouse#mouse-move for more information.
+ */
+
 import type { ElementHandle, Locator } from '@playwright/test';
 
 import type { E2EPage } from './';
@@ -21,13 +29,21 @@ export const dragElementBy = async (
   const startX = startXCoord === undefined ? boundingBox.x + boundingBox.width / 2 : startXCoord;
   const startY = startYCoord === undefined ? boundingBox.y + boundingBox.height / 2 : startYCoord;
 
-  const endX = startX + dragByX;
-  const endY = startY + dragByY;
+  const viewport = page.viewportSize();
+  if (viewport === null) {
+    throw new Error(
+      'Cannot get viewport size. See https://playwright.dev/docs/api/class-page#page-viewport-size for more information'
+    );
+  }
 
-  await page.mouse.move(startX, startY);
+  const endX = calculateEndX(startX, dragByX, viewport.width);
+  const endY = calculateEndY(startY, dragByY, viewport.height);
+
+  await page.mouse.move(startX, startY, { steps: 20 });
   await page.mouse.down();
 
-  await page.mouse.move(endX, endY, { steps: 10 });
+  await page.mouse.move(endX, endY, { steps: 20 });
+  await page.waitForChanges();
   await page.mouse.up();
 };
 
@@ -55,12 +71,53 @@ export const dragElementByYAxis = async (
   const startX = boundingBox.x + boundingBox.width / 2;
   const startY = startYCoord === undefined ? boundingBox.y + boundingBox.height / 2 : startYCoord;
 
+  const viewport = page.viewportSize();
+  if (viewport === null) {
+    throw new Error(
+      'Cannot get viewport size. See https://playwright.dev/docs/api/class-page#page-viewport-size for more information'
+    );
+  }
+
+  const endY = calculateEndY(startY, dragByY, viewport.height);
+
   await page.mouse.move(startX, startY);
   await page.mouse.down();
 
-  for (let i = 0; i < dragByY; i += 20) {
-    await page.mouse.move(startX, startY + i);
-  }
+  await page.mouse.move(startX, endY, { steps: 10 });
 
   await page.mouse.up();
+};
+
+const calculateEndX = (startX: number, dragByX: number, viewportWidth: number) => {
+  let endX = startX + dragByX;
+  // The element is being dragged past the right of the viewport.
+  if (endX > viewportWidth) {
+    // The x coordinate is set to 5 pixels to the left of the right of the viewport to avoid the mouseup event not firing.
+    endX = viewportWidth - 5;
+  }
+
+  // The element is being dragged past the left of the viewport.
+  if (endX < 0) {
+    // The x coordinate is set to 5 pixels to the right of the left of the viewport to avoid the mouseup event not firing.
+    endX = 5;
+  }
+
+  return endX;
+};
+
+const calculateEndY = (startY: number, dragByY: number, viewportHeight: number) => {
+  let endY = startY + dragByY;
+  // The element is being dragged past the bottom of the viewport.
+  if (endY > viewportHeight) {
+    // The y coordinate is set to 5 pixels above the bottom of the viewport to avoid the mouseup event not firing.
+    endY = viewportHeight - 5;
+  }
+
+  // The element is being dragged past the top of the viewport.
+  if (endY < 0) {
+    // The y coordinate is set to 5 pixels below the top of the viewport to avoid the mouseup event not firing.
+    endY = 5;
+  }
+
+  return endY;
 };
--- a/core/src/components/item-sliding/test/basic/item-sliding.e2e.ts
+++ b/core/src/components/item-sliding/test/basic/item-sliding.e2e.ts
@@ -25,12 +25,7 @@ configs({ modes: ['md'] }).forEach(({ title, screenshot, config }) => {
 configs({ modes: ['md'], directions: ['ltr'] }).forEach(({ title, screenshot, config }) => {
   test.describe(title('item-sliding: basic'), () => {
     // mouse gesture is flaky on CI, skip for now
-    test.fixme('should open when swiped', async ({ page, skip }) => {
-      skip.browser(
-        (browserName: string) => browserName !== 'chromium',
-        'dragElementBy is flaky outside of Chrome browsers.'
-      );
-
+    test('should open when swiped', async ({ page }) => {
       await page.goto(`/src/components/item-sliding/test/basic`, config);
       const item = page.locator('#item2');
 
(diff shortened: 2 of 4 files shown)


core/src/components/modal/test/fixtures.ts
```diff
@@ -16,7 +16,7 @@ export class CardModalPage {
     this.ionModalDidDismiss = await page.spyOnEvent('ionModalDidDismiss');
   }
   async openModalByTrigger(selector: string) {
-    await this.page.click(selector);
+    await this.page.locator(selector).click();
```
liamdebeasi, line 19: Why was this change made?

core/src/components/modal/test/card/modal-tablet.e2e.ts
```diff
@@ -61,12 +61,12 @@ configs({ modes: ['ios'], directions: ['ltr'] }).forEach(({ title, config }) =>
...
         const modal = await cardModalPage.openModalByTrigger('#card');
 
-        const content = (await page.$('ion-modal ion-content'))!;
+        const content = page.locator('ion-modal ion-content');
         await content.evaluate((el: HTMLIonContentElement) => el.scrollToBottom(0));
 
         await cardModalPage.swipeToCloseModal('ion-modal ion-content', false);
 
-        await content.waitForElementState('stable');
+        await content.waitFor();
```
liamdebeasi, line 69: Why was this change made?

core/src/components/range/test/range-events.e2e.ts
```diff
@@ -10,15 +10,28 @@ configs({ modes: ['ios'], directions: ['ltr'] }).forEach(({ title, config }) =>
...
+         * will be 99 instead of 100.
+         */
+        await page.setContent(
+          `
+          <div style="padding: 0 20px">
+            <ion-range value="20"></ion-range>
+          </div>
+        `,
+          config
+        );
```
liamdebeasi, line 27: If this is skipped do we still need to make these changes?

core/src/utils/test/playwright/drag-element.ts
```diff
@@ -55,12 +67,76 @@ export const dragElementByYAxis = async (
...
-  for (let i = 0; i < dragByY; i += 20) {
-    await page.mouse.move(startX, startY + i);
-  }
+  // Drag the element.
+  await moveElement(page, startX, startY, 0, dragByY);
 
   await page.mouse.up();
 };
+
+const calculateEndX = (startX: number, dragByX: number, viewportWidth: number) => {
```
liamdebeasi, line 80: I wonder if this should error instead of trying to adjust the values in-bounds. My concern is that if we adjust the values such that certain gestures no longer work as expected, this may be hard to debug if team members don't know their values are being adjusted for them.

liamdebeasi: Alternatively, is there info we can expose that lets us set safe values without having to set explicit pixel values? For example, I might want to have a gesture that pulls all the way down to the bottom of the viewport. I don't care what the exact pixel coordinate is, as long as it is the bottom of the viewport.

core/src/components/modal/test/fixtures.ts
thetaPC: I switched it since the Playwright documents [discourage](https://playwright.dev/docs/api/class-page#page-click) the use of `page.click` and suggest `locator.click` instead.

core/src/components/modal/test/card/modal-tablet.e2e.ts
thetaPC: Playwright [discourages](https://playwright.dev/docs/api/class-elementhandle) the use of `ElementHandle`. `waitForElementState` is based on that class. It recommends to use `Locators` instead.

core/src/components/range/test/range-events.e2e.ts
thetaPC: Technically no. But I figured since these changes will be implemented when the flakiness is solved then we won't need to add it later.

core/src/utils/test/playwright/drag-element.ts
thetaPC: How about `dragByX: number | undefined`? If undefined then we would drag all the way to the bottom. The same would go for `calculateEndY`. If we go this route, then the parameter `dragByX` from `dragElementBy` should update to also handle undefined.

core/src/components/modal/test/card/modal-tablet.e2e.ts
liamdebeasi: Ok makes sense 👍

core/src/components/range/test/range-events.e2e.ts
```diff
@@ -10,15 +10,28 @@ configs({ modes: ['ios'], directions: ['ltr'] }).forEach(({ title, config }) =>
...
 
         const rangeStart = await page.spyOnEvent('ionKnobMoveStart');
         const rangeEnd = await page.spyOnEvent('ionKnobMoveEnd');
 
         const rangeEl = page.locator('ion-range');
 
-        await dragElementBy(rangeEl, page, testInfo.project.metadata.rtl ? -300 : 300, 0);
+        await dragElementBy(rangeEl, page, 300, 0);
```
amandaejohnston, line 34: Why was this changed? I didn't see RTL support added to `dragElementBy` at a glance, but maybe I missed something.

core/src/utils/test/playwright/drag-element.ts
```diff
@@ -21,14 +30,17 @@ export const dragElementBy = async (
   const startX = startXCoord === undefined ? boundingBox.x + boundingBox.width / 2 : startXCoord;
   const startY = startYCoord === undefined ? boundingBox.y + boundingBox.height / 2 : startYCoord;
 
-  const endX = startX + dragByX;
-  const endY = startY + dragByY;
-
+  // Navigate the start position.
```
amandaejohnston, line 33:
```suggestion
  // Navigate to the start position.
```

core/src/utils/test/playwright/drag-element.ts
amandaejohnston: I assume using a Direction would drag to the far edge of the element's bounding box?

Also the literal up/down language is a little confusing when applied to the X axis, but admittedly I can't think of a better way of capturing it 😆 My immediate thought is having `dragElementBy` accept `from` and `to` params that can be compass directions (e.g. `from='center'`, `to='northeast'`) but that might be overkill.

core/src/components/modal/test/fixtures.ts
thetaPC: Plus it did help lessen the flakiness when switched over. The same for `waitForElementState`.

core/src/components/range/test/range-events.e2e.ts
thetaPC: I removed it since the config is only testing `LTR`. Additionally the behavior doesn't change based on mode so I didn't add a new test for `RTL`.

core/src/utils/test/playwright/drag-element.ts
liamdebeasi: It might be good to do some design work on this. I do think there's a nice solution in here somewhere, I just can't think of it at the moment. How do we feel about erroring/warning when the coordinate is out of bounds for now and letting the test implementer make the adjustment themselves? That way it's clear to the implementer what is happening rather than the coordinates being adjusted without them knowing.

core/src/components/range/test/range-events.e2e.ts
amandaejohnston: Ah, yup there's the part I missed 😆 Thanks!

core/src/utils/test/playwright/drag-element.ts
thetaPC: Understood, I'll update the code to error for now and create a design doc for a better solution. Once the design doc gets approved, I'll make a new PR to implement the solution.

core/src/utils/test/playwright/drag-element.ts
```diff
@@ -21,14 +30,17 @@ export const dragElementBy = async (
...
   await page.mouse.down();
 
-  await page.mouse.move(endX, endY, { steps: 10 });
-  await page.mouse.up();
+  // Drag the element.
+  await moveElement(page, startX, startY, dragByX, dragByY);
+
+  if (releaseDrag) {
+    await page.mouse.up();
+  }
```
liamdebeasi, line 43: Why do we conditionally release the drag gesture? I'd expect the gesture to always be released when it's done.

core/src/utils/test/playwright/drag-element.ts
```diff
@@ -55,12 +67,83 @@ export const dragElementByYAxis = async (
...
+
+    await page.mouse.move(middleX, middleY);
+
+    // Safari needs to wait for a repaint to occur before moving the mouse again.
+    if (browser === 'webkit' && i % 2 === 0) {
+      // Repainting every 2 steps is enough to keep the drag gesture smooth.
+      // Anything past 4 steps will cause the drag gesture to be flaky.
+      await page.evaluate(() => new Promise(requestAnimationFrame));
+    }
+  }
```
liamdebeasi, line 148: Non-blocking, but something to consider: Playwright optionally fires intermediary steps with `page.mouse.move`. Have we tried using this instead of manually generating these steps? https://playwright.dev/docs/api/class-mouse#mouse-move

core/src/utils/test/playwright/drag-element.ts
thetaPC: This is prep work for when range tests are enabled. One of the range tests is to check that [scroll cannot occur while the knob is swiped](https://github.com/ionic-team/ionic-framework/pull/27808/files#diff-4cc6a4738b72ca53b59df31fc34e2e29e5f510dde3ff52dad323c1b7bedf59d1R80). The test needs to do the check while the knob is being held down and moving.

core/src/utils/test/playwright/drag-element.ts
liamdebeasi: Thanks for checking!

core/src/components/range/test/range-events.e2e.ts
```diff
@@ -10,15 +10,28 @@ configs({ modes: ['ios'], directions: ['ltr'] }).forEach(({ title, config }) =>
       /**
        * The mouse events are flaky on CI
        */
-      test.fixme('should emit start/end events', async ({ page }, testInfo) => {
-        await page.setContent(`<ion-range value="20"></ion-range>`, config);
+      test.fixme('should emit start/end events', async ({ page }) => {
```
sean-perkins, line 13: I know this is already merged, but just a general question if `test.fixme` should be removed as part of these changes?

thetaPC: Range is still flaky so the `fixme` was left in until a fix is implemented.

PR #27810, opened 2023-07-15
chore(angular): improve types of `platform.ts`

donaldxdonald:
Issue number: N/A

---------

The `backButton` and `keyboardDidShow` properties on the `Platform` service do not have type signatures.

## What is the new behavior?

- `backButton` is typed to the `BackButtonEmitter`
- `keyboardDidShow` is typed to `Subject<KeyboardEventDetail>`

- [ ] Yes
- [x] No


sean-perkins (commented):
Thanks for this PR!

Can you share what some of the existing challenges or motivations are for these changes?

I'm all for addressing some of the usages of `any` and having stricter types, but globally augmenting the `Window` interface isn't really something I think we should be doing here.

donaldxdonald:
@sean-perkins Thanks for you reply. At first, I just wanted to enhance the type of `keyboardDidShow`. When I noticed that the types related to 'window' were also `any`, I thought it would be convenient to make the change. However, I realize now that it was a bit radical and unnecessary. The related modifications have been reverted.

thetaPC (approved):
LGTM

sean-perkins (approved):

PR #27816, opened 2023-07-18
fix(datetime-button): render correct text when passing partial date values

### LINKED ISSUE #27797: bug: date value is not finite in DateTimeFormat format() when presentation='year' on ios device
### Prerequisites

### Ionic Framework Version

v7.x

### Current Behavior

We try to show a ion-datetime with presentation="year" in ion-modal. And we got this bug on ios device. On android. it worked well.

When using ion-datetime with presentation='year'. I got error: "date value is not finite in DateTimeFormat format()"
image

I see it happen in getNormalizedDate funtion when we try to new Date() with refParts that have some undefined propertise
image

On chrome browser, it work well:
image

Result: we dont see the year number.

### Expected Behavior

No more error

### Steps to Reproduce

- Implement ion-datetime with presentation='year'
- value = 2023;
`<ion-datetime presentation="year"
                size="cover"
                width="100%"
                [min]="2019"
                [max]="2030"
                [(value)]="value"
                id="datetime"></ion-datetime>`
- View it on ios device

### Code Reproduction URL

_No response_

### Ionic Info

Ionic:

   Ionic CLI                     : 7.1.1 (/usr/local/lib/node_modules/@ionic/cli)
   Ionic Framework               : @ionic/angular 7.1.2
   @angular-devkit/build-angular : 15.2.9
   @angular-devkit/schematics    : 15.2.9
   @angular/cli                  : 15.2.9

Cordova:

   Cordova CLI       : 12.0.0 (cordova-lib@12.0.1)
   Cordova Platforms : android 12.0.0, ios 6.3.0
   Cordova Plugins   : cordova-plugin-ionic-keyboard 2.2.0, cordova-plugin-ionic-webview 5.0.0, (and 33 other plugins)

Utility:

   cordova-res : 0.15.4

System:

   ios-sim : 8.0.2
   NodeJS  : v18.16.0 (/usr/local/bin/node)
   npm     : 9.5.1
   OS      : macOS Unknown
   Xcode   : Xcode 14.3 Build version 14E222b

### Additional Information

_No response_

liamdebeasi:
Issue number: resolves #27797

---------

Datetime Button passes a parsed value to one of the many text formatting utilities we have, such as `getMonthAndYear`. However, developers can pass partial date values such as `2022` or `2022-04` (April 2022). According to https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date#date_time_string_format, these are still valid date strings.

However, the `parseDate` utility does not add fallback values. So passing `2022` will cause the `day` and `month` fields to be `undefined`. This means that `getNormalizedDate` passes `'//2022'` to the `Date` constructor.

Some browsers, such as Chrome, will automatically account for the stray slashes and still return a valid date. Other browsers, such as Safari, do not do this and will either return "Invalid Date" or throw an error.

## What is the new behavior?

- Date normalizing utility now has fallback values so we always pass in a valid date. In the example above, `getNormalizedDate` will now pass `'1/1/2022'` instead of `'//2022'` to the `Date` constructor.
- Refactored other utils that use `new Date` to make use of `getNormalizedDate` since they are also impacted.

Note: I added an E2E test instead of a spec test because I want to test cross-browser behavior to ensure consistency.

- [ ] Yes
- [x] No

### DIFF
--- a/core/src/components/datetime/utils/format.ts
+++ b/core/src/components/datetime/utils/format.ts
@@ -201,7 +201,7 @@ export const getNormalizedDate = (refParts: DatetimeParts) => {
    * account for this and still return a valid date. However,
    * this is not a consistent behavior across all browsers.
    */
-  return new Date(`${refParts.month}/${refParts.day}/${refParts.year}${timeString} GMT+0000`);
+  return new Date(`${refParts.month ?? 1}/${refParts.day ?? 1}/${refParts.year ?? 2023}${timeString} GMT+0000`);
 };
 
 /**
--- a/core/src/components/datetime/utils/format.ts
+++ b/core/src/components/datetime/utils/format.ts
@@ -113,7 +113,7 @@ export const generateDayAriaLabel = (locale: string, today: boolean, refParts: D
   /**
    * MM/DD/YYYY will return midnight in the user's timezone.
    */
-  const date = new Date(`${refParts.month}/${refParts.day}/${refParts.year} GMT+0000`);
+  const date = getNormalizedDate(refParts);
 
   const labelString = new Intl.DateTimeFormat(locale, {
     weekday: 'long',
@@ -134,7 +134,7 @@ export const generateDayAriaLabel = (locale: string, today: boolean, refParts: D
  * Used for the header in MD mode.
  */
 export const getMonthAndDay = (locale: string, refParts: DatetimeParts) => {
-  const date = new Date(`${refParts.month}/${refParts.day}/${refParts.year} GMT+0000`);
+  const date = getNormalizedDate(refParts);
   return new Intl.DateTimeFormat(locale, { weekday: 'short', month: 'short', day: 'numeric', timeZone: 'UTC' }).format(
     date
   );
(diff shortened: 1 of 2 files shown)


core/src/components/datetime/utils/format.ts
```diff
@@ -183,11 +183,25 @@ export const getYear = (locale: string, refParts: DatetimeParts) => {
...
+   * We use / notation here for the date
+   * so we do not need to do extra work and pad values with zeroes.
+   * Values such as YYYY-MM are still valid, so
+   * we add fallback values so we still get
+   * a valid date otherwise we will pass in a string
+   * like "//2023". Some browsers, such as Chrome, will
+   * account for this and still return a valid date. However,
+   * this is not a consistent behavior across all browsers.
+   */
+  return new Date(`${refParts.month ?? 1}/${refParts.day ?? 1}/${refParts.year ?? 2023}${timeString} GMT+0000`);
```
sean-perkins, line 204: Should we fallback to the current year, instead of hardcoding 2023?

e.g.:
```ts
refParts.year ?? new Date().getFullYear()
```

liamdebeasi: We could, though we also default to 1/1/2023 here: https://github.com/ionic-team/ionic-framework/blob/89ed213a582877f45f91a77f3cc3905715114384/core/src/components/datetime/utils/format.ts#L58

Any preference?

amandaejohnston (approved):

core/src/components/datetime/utils/format.ts
liamdebeasi: > Presumably, if I'm passing in a value without the year, I'm using the month presentation, in which case I don't care about the year since it'll never be displayed.

Yep exactly this. If any of those values are `undefined` then they aren't actually used in the rendering of the component -- we just need them to get a valid date

sean-perkins: I'm ok with this as-is then 👍

PR #27829, opened 2023-07-19
fix(item-sliding): buttons are not interactive on close

### LINKED ISSUE #22722: bug: Ion-item-sliding option can be clicked for a moment after item is closed
'ion-item-sliding' 'option' is clickable even when slide is closed. 

for example in the following is user click on same location where 'Lorem' button was displayed, after closing the slide, it will trigger 'unread' function. 
```
 <ion-item-sliding>
    <ion-item>
      <ion-label>Item Options</ion-label>
      <ion-icon name="trash" (click)="ipsum()"></ion-icon>
    </ion-item>
    <ion-item-options side="end">
      <ion-item-option (click)="unread(item)">Lorem</ion-item-option>
    </ion-item-options>
  </ion-item-sliding>
```

Problem is that after sliding is closed it takes bit time to remove 'slide-active' classes. In this mean time when user click on item on same function, it get clicked on item option even item option is not visible.

**Ionic version:**
ionic --version:  6.12.1

**Expected behavior:**
After item sliding is closed the options should not be clickable. This should happen fast.

**Steps to reproduce:**
 From above code, Just slide to see 'Lorem' option.
 Now close the slide and click on same position where lorem option was displayed. (or on trash icon)
    - it will trigger 'unread' function.

liamdebeasi:
Issue number: resolves #22722

---------

Item Sliding Options are not disabled until after a 600ms timeout. This timeout exists to allow the close transition to complete. Without the timeout, the item sliding options disappear without a transition. I explored waiting for the `transitionend` event instead of the `setTimeout`, but the bug persisted.

## What is the new behavior?

- If an item sliding is closing then we add a class to the host. This class disables pointer events on all `ion-item-options` children so the buttons cannot be accidentally tapped while closing. This class is then removed after the 600ms timeout.

- [ ] Yes
- [x] No

### DIFF
--- a/core/src/components/item-sliding/item-sliding.scss
+++ b/core/src/components/item-sliding/item-sliding.scss
@@ -29,6 +29,9 @@ ion-item-sliding .item {
   will-change: transform;
 }
 
+.item-sliding-closing ion-item-options {
+  pointer-events: none;
+}
 
 .item-sliding-active-swipe-end .item-options-end .item-option-expandable {
   @include multi-dir() {
--- a/core/src/components/item-sliding/item-sliding.tsx
+++ b/core/src/components/item-sliding/item-sliding.tsx
@@ -407,6 +407,9 @@ export class ItemSliding implements ComponentInterface {
     if (!this.item) {
       return;
     }
+
+    const { el } = this;
+
     const style = this.item.style;
     this.openAmount = openAmount;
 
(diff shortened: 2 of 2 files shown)


brandyscarney (approved):
I tested this out by slowing the animation to 10% so I could click the item options and it seems to work great!

PR #27838, opened 2023-07-20
feat(reorder): support dynamic type

liamdebeasi:
Issue number: N/A

---------

Reorder does not support dynamic type

## What is the new behavior?

- Convert fonts to rem
- Also added logic so that slotted icons can participate in Dynamic Type

- [ ] Yes
- [x] No

## Other information

Note that our reorder icons are bigger than the native iOS reorder icons. As a result, when scaling up to large sizes the reorder icon may look particularly large when compared with native. 

Screenshot 2023-07-20 at 3 50 19 PM

The native reorder icons also seem to use a different scale than what we have access to. For example, I could match the largest font size using ~1.3rem, but at 85% text scaling the icon was way too small.

The native reorder has neither a min nor a max font size from what I could tell. Not sure what the right call is on this. At the max font size, the reorder takes up quite a bit of space such that the text in items is truncated. However, the text probably should not be truncated in the first place.

Demo: https://github.com/ionic-team/ionic-framework/assets/2721089/b473d3ca-89ea-4010-a9cb-898b86ae7f03


core/src/components/reorder/reorder.scss
```diff
@@ -17,6 +17,12 @@
 
 .reorder-icon {
   display: block;
+}
 
-  font-size: 22px;
```
liamdebeasi, line 21: Each mode stylesheet overrides this, so this code never did anything

core/src/components/reorder/reorder.scss
```diff
@@ -17,6 +17,12 @@
...
+/**
+ * This ensures that custom reorder
+ * icons resize with Dynamic Type.
+ */
+::slotted(ion-icon) {
+  font-size: 1rem;
```
liamdebeasi, line 27: Slotted icons currently inherit a font size of 16px

mapsandapps (commented):
It looks good, but I do have a question:

If the user passes in text to the slot, e.g.

`<ion-reorder slot="end">drag me</ion-reorder>`

it's not using rem. I didn't see this functionality (putting text in the reorder instead of an icon) in the docs, so I'm guessing it's not actually supported, but it might be something to consider?

liamdebeasi:
Good catch! So when you pass text in it's going to inherit the font size from `ion-item` which does not support Dynamic Type yet. So once we add support in `ion-item`, this should work as expected.

mapsandapps (approved):
looks good 👍🏻

PR #27845, opened 2023-07-24
feat(card): support dynamic type

thetaPC:
Issue number: N/A

---------

The card component does not support dynamic type.

## What is the new behavior?

- Convert fonts to rem

- [ ] Yes
- [x] No

## Other information

Dev build: 7.0.15-dev.11690215123.18be9175

Make sure to enable dynamic type by adding the following to your app:
```css
html {
}
```

https://github.com/ionic-team/ionic-framework/assets/13530427/4eef3130-052a-4567-b6aa-d7e1d34e259b


liamdebeasi (changes requested):
Can we add a screenshot test of card, card subtitle, card title, card content all being used with the text scaled up? Here's an example you can follow: https://github.com/ionic-team/ionic-framework/pull/27848/files#diff-2796ab1a929613b79f68ba352b11a68994e442070db3a7bbdbd7a1c0ded56c47R9-R26

(Note: previously the team discussed using a linter to ensure rem usage, but as I got more into development I realized some of these components are a bit more in-depth than I initially thought, so to be safe we're going to add a screenshot test)

PR #27848, opened 2023-07-24
feat(datetime-button): add dynamic type support 

brandyscarney:
Issue number: N/A

---------

The datetime button does not support dynamic type.

## What is the new behavior?
Adds support for dynamic type by converting `font-size` to use `rem` units and adding an ellipsis to text when the size is too large.

- [ ] Yes
- [x] No

## Other information

Dynamic type can be enabled with the following CSS:

```css
html {
}
```


core/src/components/datetime-button/test/a11y/datetime-button.e2e.ts
```diff
@@ -0,0 +1,47 @@
+import { expect } from '@playwright/test';
+import { configs, test } from '@utils/test/playwright';
+
+/**
+ * The tested behavior does not vary across directions
+ */
+configs({ directions: ['ltr'] }).forEach(({ title, screenshot, config }) => {
```
liamdebeasi, line 7: datetime-button doesn't have per-mode styles, so we can skip one of the modes
```suggestion
configs({ directions: ['ltr'], modes: ['ios'] }).forEach(({ title, screenshot, config }) => {
```

liamdebeasi (changes requested):

core/src/components/datetime-button/test/a11y/datetime-button.e2e.ts
liamdebeasi: Also if you merge this you'll need to manually delete the redundant generated screenshots

core/src/components/datetime-button/datetime-button.scss
```diff
@@ -18,6 +18,10 @@
 
   position: relative;
 
+  flex: 0 1 auto;
+
+  min-width: 0;
```
liamdebeasi, line 23: What does this do?

liamdebeasi (commented):
thank you!

core/src/components/datetime-button/datetime-button.scss
brandyscarney: Actually I had this in the wrong place. It is necessary for when an `ion-datetime-button` is inside of an `ion-item` (so all of the `ion-datetime/basic` tests). It should be on the host, I updated it here: https://github.com/ionic-team/ionic-framework/pull/27848/commits/e52c28bcefe93370e9e81dd498b428d2161697d6

Also added a test here: https://github.com/ionic-team/ionic-framework/pull/27848/commits/8ea5fcd1c1b380d7e7b50e236c810a26fb5f3639

Here's how it looks without and with the `min-width: 0`:

| without `min-width: 0` | with `min-width: 0` |
| ---| ---|
| image 9 | image 11 | 

I'm running screenshots now so this should be updated soon.

core/src/components/datetime-button/test/a11y/datetime-button.e2e.ts
```diff
@@ -0,0 +1,69 @@
+import { expect } from '@playwright/test';
+import { configs, test } from '@utils/test/playwright';
+
+/**
+ * The tested behavior does not vary across directions/modes
+ */
+configs({ directions: ['ltr'], modes: ['ios'] }).forEach(({ title, screenshot, config }) => {
+  test.describe(title('loading: font scaling'), () => {
```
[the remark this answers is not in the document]
brandyscarney: Ah thanks, good catch! I swear I fixed this but then I must've copied it again haha.

thetaPC (approved):
LGTM

liamdebeasi (approved):
Great job!


edit: one comment below, but not really a bug

core/src/components/datetime-button/datetime-button.scss
```diff
@@ -6,9 +6,13 @@
 :host {
   display: flex;
 
+  flex: 0 1 auto;
```
liamdebeasi, line 9: Just noticed that this line doesn't seem to do anything

brandyscarney: You're right, [that is the initial value](https://developer.mozilla.org/en-US/docs/Web/CSS/flex#values)

I removed it here: https://github.com/ionic-team/ionic-framework/pull/27848/commits/924601f0d0b4f9b81da9a10604aa312cf665d6aa

PR #27853, opened 2023-07-25
fix(item-options): use correct safe area padding

liamdebeasi:
Issue number: Internal

---------

Item sliding options in the "start" slot use the "left" safe area padding on the end edge. This causes the padding to be added in the wrong place.

During development I also discovered that the RTL padding was wrong for both start and end options.

## What is the new behavior?

LTR:

- `side="start"` options set "left" safe area padding on left edge of first child
- `side="end"` options set "right" safe area padding on right edge of last child

RTL:

- `side="start"` options set "right" safe area padding on right edge of first child
- `side="end"` options set "left" safe area padding on the left edge of the last child

- [ ] Yes
- [x] No

## Other information

| side prop/text direction | `main` | `branch` |
| - | - | - |
| start/LTR | start-ltr | start-ltr |
| end/LTR | end-ltr | end-ltr |
| start/RTL | start-rtl | start-rtl |
| end/RTL | end-rtl | end-rtl |

### DIFF
--- a/core/src/components/item-options/item-options.scss
+++ b/core/src/components/item-options/item-options.scss
@@ -56,20 +56,14 @@ ion-item-options {
   }
 }
 
-[dir="ltr"] .item-options-start ion-item-option:first-child {
+[dir="ltr"] .item-options-start ion-item-option:first-child,
+[dir="rtl"] .item-options-end ion-item-option:first-child {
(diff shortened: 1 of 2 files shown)


thetaPC (approved):
LGTM

PR #27854, opened 2023-07-25
feat(chip): add dynamic type support

mapsandapps:
Issue number: N/A

---------

The chip component does not support dynamic type.

## What is the new behavior?

The chip component supports dynamic type, has a dynamic height, has min and max font sizes in `ios` mode, and has a screenshot test.

Small md mode:
Screenshot 2023-07-25 at 3 10 24 PM

Medium md mode:
Screenshot 2023-07-25 at 3 10 15 PM

Large md mode:
Screenshot 2023-07-25 at 3 10 06 PM

 

- [ ] Yes
- [x] No


core/src/components/chip/chip.scss
```diff
@@ -8,24 +8,23 @@
   --background: #{rgba($text-color-rgb, .12)};
   --color: #{rgba($text-color-rgb, .87)};
 
-  @include border-radius(16px);
+  @include border-radius(calc(16 / 14 * 1em));
```
mapsandapps, line 11: With this code, the corners aren't fully rounded if the text is wrapping.

Should we instead do e.g. a hack of 9999px to make sure all four corners round equally and the short corners are fully rounded? I didn't see anywhere in the code that we're currently doing that, so I wasn't sure what we want to do here.

with current code:
Screenshot 2023-07-25 at 4 15 18 PM

with 9999px:
Screenshot 2023-07-25 at 4 10 38 PM

core/src/components/chip/chip.ios.vars.scss
```diff
@@ -0,0 +1,7 @@
+@import "../../themes/ionic.globals.ios";
+
+// iOS Chip
+// --------------------------------------------------
+
+/// @prop - Font size of the chip
+$chip-ios-font-size: clamp(13px, 0.875rem, 22px);
```
mapsandapps, line 7: I wasn't sure where to get the min & max values from. I did some quick searching and couldn't find any iOS spec for chip font sizes.

mapsandapps:
gonna reopen

PR #27855, opened 2023-07-25
feat(toast): add htmlAttributes property for passing attributes to buttons  

brandyscarney:
Issue number: N/A

---------

Buttons containing only icons are not accessible as there is no way to pass an `aria-label` attribute (or any other html attribute). 

## What is the new behavior?
- Adds the `htmlAttributes` property on the `ToastButton` interface 
- Passes the `htmlAttributes` to the buttons
- Adds a test to verify `aria-label` and `aria-labelled-by` are passed to the button

- [ ] Yes
- [x] No

### DIFF
--- a/core/src/components/toast/test/a11y/toast.e2e.ts
+++ b/core/src/components/toast/test/a11y/toast.e2e.ts
@@ -39,7 +39,9 @@ configs({ directions: ['ltr'] }).forEach(({ title, config }) => {
       expect(results.violations).toEqual([]);
     });
 
-    test('should have aria-labelledby and aria-label added to the button when htmlAttributes is set', async ({ page }) => {
+    test('should have aria-labelledby and aria-label added to the button when htmlAttributes is set', async ({
+      page,
+    }) => {
       const ionToastDidPresent = await page.spyOnEvent('ionToastDidPresent');
 
       await page.click('#aria-label-toast-trigger');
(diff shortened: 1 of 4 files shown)


sean-perkins (approved):

core/src/components/toast/test/a11y/toast.e2e.ts
```diff
@@ -38,5 +38,26 @@ configs({ directions: ['ltr'] }).forEach(({ title, config }) => {
...
+    }) => {
+      const didPresent = await page.spyOnEvent('ionToastDidPresent');
+
+      await page.click('#aria-label-toast-trigger');
+      await didPresent.next();
+
+      const toastButton = page.locator('#aria-label-toast .toast-button');
+
+      /**
+       * expect().toHaveAttribute() can't check for a null value, so grab and check
```
liamdebeasi, line 53: Same as https://github.com/ionic-team/ionic-framework/pull/27862/files#r1276705971:

Why can't we use the `toHaveAttribute` API? We aren't checking for null values, so that limitation should not apply here.

brandyscarney: Fixed by https://github.com/ionic-team/ionic-framework/pull/27855/commits/353d2b9bf2f847b3be5ac078b3eb05cea52c2117

thetaPC (approved):
LGTM

liamdebeasi (approved):

PR #27863, opened 2023-07-26
feat(action-sheet): add htmlAttributes property for passing attributes to buttons

brandyscarney:
Issue number: N/A

---------

Buttons containing only icons are not accessible as there is no way to pass an `aria-label` attribute (or any other html attribute). 

## What is the new behavior?
- Adds the `htmlAttributes` property on the `ActionSheetButton` interface 
- Passes the `htmlAttributes` to the buttons (both the buttons array and the cancelButton)
- Adds two tests to verify `aria-label` and `aria-labelled-by` are passed to a button with and without the cancel role - this was done because action sheet breaks these buttons up when rendering

- [ ] Yes
- [x] No

### DIFF
--- a/core/src/components/action-sheet/action-sheet.tsx
+++ b/core/src/components/action-sheet/action-sheet.tsx
@@ -386,7 +386,13 @@ export class ActionSheet implements ComponentInterface, OverlayInterface {
                 </div>
               )}
               {buttons.map((b) => (
-                <button type="button" id={b.id} class={buttonClass(b)} onClick={() => this.buttonClick(b)}>
+                <button
+                  {...b.htmlAttributes}
+                  type="button"
+                  id={b.id}
+                  class={buttonClass(b)}
+                  onClick={() => this.buttonClick(b)}
+                >
                   <span class="action-sheet-button-inner">
                     {b.icon && <ion-icon icon={b.icon} aria-hidden="true" lazy={false} class="action-sheet-icon" />}
                     {b.text}
@@ -398,7 +404,12 @@ export class ActionSheet implements ComponentInterface, OverlayInterface {
 
             {cancelButton && (
               <div class="action-sheet-group action-sheet-group-cancel">
-                <button type="button" class={buttonClass(cancelButton)} onClick={() => this.buttonClick(cancelButton)}>
+                <button
+                  {...cancelButton.htmlAttributes}
+                  type="button"
+                  class={buttonClass(cancelButton)}
+                  onClick={() => this.buttonClick(cancelButton)}
+                >
                   <span class="action-sheet-button-inner">
                     {cancelButton.icon && (
                       <ion-icon icon={cancelButton.icon} aria-hidden="true" lazy={false} class="action-sheet-icon" />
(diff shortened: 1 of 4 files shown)


sean-perkins (approved):
Good work 👍

core/src/components/action-sheet/action-sheet.tsx
```diff
@@ -386,7 +386,13 @@ export class ActionSheet implements ComponentInterface, OverlayInterface {
                 </div>
               )}
               {buttons.map((b) => (
-                <button type="button" id={b.id} class={buttonClass(b)} onClick={() => this.buttonClick(b)}>
+                <button
+                  {...b.htmlAttributes}
```
brandyscarney, line 390: The design document called for this to be added like the following:

```
  {...(b.htmlAttributes as any)}
```

I couldn't see any reason that it needs to be cast as `any` though - no failing linters or builds. Anyone know if this is needed for some other reason?

thetaPC: I'm not sure. I even checked if Visual Studio would give an error of sorts but nothing came up.

core/src/components/action-sheet/test/a11y/action-sheet.e2e.ts
```diff
@@ -10,16 +10,43 @@ const testAria = async (page: E2EPage, buttonID: string, expectedAriaLabelledBy:
...
+
+  await didPresent.next();
+
+  const actionSheetButton = page.locator('ion-action-sheet .action-sheet-button');
+
+  /**
+   * expect().toHaveAttribute() can't check for a null value, so grab and check
+   * the value manually instead.
+   */
+  const ariaLabelledBy = await actionSheetButton.getAttribute('aria-labelledby');
```
liamdebeasi, line 43: Same as https://github.com/ionic-team/ionic-framework/pull/27862/files#r1276705971: 
Why can't we use the `toHaveAttribute` API? We aren't checking for null values, so that limitation should not apply here.

brandyscarney: Fixed by https://github.com/ionic-team/ionic-framework/pull/27863/commits/54945521542d7af73695fbfda290aae579b10c85

core/src/components/action-sheet/test/a11y/index.html
```diff
@@ -63,6 +67,39 @@ <h1>Action Sheet - A11y</h1>
...
+        openActionSheet({
+          header: 'Header',
+          subHeader: 'Subtitle',
+          buttons: [
+            {
+              text: 'Cancel',
+              role: 'cancel',
+              htmlAttributes: {
+                'aria-label': 'cancel button',
+                'aria-labelledby': 'cancel-label',
```
liamdebeasi, line 97: I don't understand the `aria-labelledby` usage here. Is there an element with `#cancel-label`?

brandyscarney: No I was just adding a test to make sure it passes more than one property

liamdebeasi (approved):

PR #27867, opened 2023-07-26
test(angular): update test app to account for standalone components

liamdebeasi:
Issue number: N/A

---------

We do not have a way of testing Ionic UI standalone components isolated from the regular lazy loaded bundle

## What is the new behavior?

In order to test standalone components isolated from the lazy loaded components, we need to and environment where the lazy loaded components never get initialized. As a result, I made the following changes:

1. Created two routes: `/lazy` and `/standalone`. Any `/lazy/*` routes will have `IonicModule` initialized at the root: https://github.com/ionic-team/ionic-framework/pull/27867/files#diff-174033d040c9d2279d515398002826936441e7892b50ecde6a815fb719c48b9eR9-R18. Any `/standalone/*` routes will **not** have `IonicModule` initialized.
2. Updated any URL references to ensure the `/lazy` prefix is added. For example, `/accordions` is now `/lazy/accordions`.
3. Created an `app-lazy` component. This renders IonApp and IonRouterOutlet using the lazy components. All lazy routes are rendered as children of this component.
4. Created an `app-standalone` component. This renders a regular `router-outlet` for now as a placeholder (since we don't have any Ionic UI components as standalone at the moment). But the idea here is that team members will be able to create routes here to test standalone UI components.

- [ ] Yes
- [x] No


packages/angular/test/base/src/app/test/test.component.ts
```diff
@@ -0,0 +1,12 @@
+import { Component } from '@angular/core';
+
+@Component({
+  selector: 'app-test',
+  templateUrl: './test.component.html',
+  standalone: true
+})
+export class TestComponent {
+  constructor() {
+    console.log('This should be undefined:', customElements.get('ion-app'));
```
liamdebeasi, line 10: This is a placeholder for now to verify that the IonicModule is not being defined

packages/angular/test/base/src/app/app.routes.ts
```diff
@@ -1,82 +1,6 @@
...
-        path: 'page2',
-        component: NestedOutletPage2Component
-      }
-    ]
-  },
-  {
-    path: 'form-controls/range',
-    loadChildren: () => import('./form-controls/range/range.module').then(m => m.RangeModule)
-  }
+  { path: 'lazy', loadChildren: () => import('./app-lazy/app.module').then(m => m.AppModule) },
```
sean-perkins, line 4: ~ Should we add a default match path to redirect to a known route? Otherwise when serving the app you need to know to go to `/lazy/something` or `/standalone/something`?

e.g.:
```
{
  path: '',
  pathMatch: 'full',
  redirectTo: '/lazy'
}
```

sean-perkins (approved):

packages/angular/test/base/src/app/app.routes.ts
liamdebeasi: Good call. I updated this in https://github.com/ionic-team/ionic-framework/pull/27867/commits/85cac110653bbb6170e5f93143f6f6338a368266

PR #27868, opened 2023-07-26
feat(chip): add dynamic type support

mapsandapps:
Issue number: N/A

---------

The chip component does not support dynamic type.

## What is the new behavior?

The chip component supports dynamic type, has a dynamic height, has min and max font sizes in `ios` mode, and has a screenshot test.

### md mode

Small md mode:
md-small

Medium md mode:
md-medium

Large md mode:
md-large

 
### ios mode
Small ios mode:
ios-small

Medium ios mode:
ios-medium

Large ios mode:
ios-large

- [ ] Yes
- [x] No

## Other information

This also fixes an existing bug that the contents of a chip could get smooshed if the chip doesn't have enough horizontal space for its contents.

### Before
rStll

### After
Screenshot 2023-08-02 at 5 00 39 PM


core/src/components/chip/chip.scss
```diff
@@ -8,24 +8,23 @@
   --background: #{rgba($text-color-rgb, .12)};
   --color: #{rgba($text-color-rgb, .87)};
 
-  @include border-radius(16px);
+  @include border-radius(calc(16 / 14 * 1em));
```
mapsandapps, line 11: With this code, the corners aren't fully rounded if the text is wrapping.

Should we instead do e.g. a hack of 9999px to make sure all four corners round equally and the short corners are fully rounded? I didn't see anywhere in the code that we're currently doing that, so I wasn't sure what we want to do here.

with current code:
Screenshot 2023-07-25 at 4 15 18 PM

with 9999px:
Screenshot 2023-07-25 at 4 10 38 PM

core/src/components/chip/chip.ios.vars.scss
```diff
@@ -0,0 +1,7 @@
+@import "../../themes/ionic.globals.ios";
+
+// iOS Chip
+// --------------------------------------------------
+
+/// @prop - Font size of the chip
+$chip-ios-font-size: clamp(13px, 0.875rem, 22px);
```
mapsandapps, line 7: I wasn't sure where to get the min & max values from. I did some quick searching and couldn't find any iOS spec for chip font sizes.

core/src/components/chip/chip.scss
thetaPC: Why not use `1rem` instead? `@include border-radius(1rem);`

liamdebeasi: The border radius values should not change as the text scales -- we only want the text to change.

mapsandapps: if i set it to `1rem`, the ios ones look like this at small sizes

Screenshot 2023-07-27 at 9 36 43 AM

i assume we want the corners to be fully rounded, not like this.

this is happening because of the requirements in the ticket:

1. the ios font size has a max & min value
2. height is relative to that font size (and therefore also clamped)

therefore, `1rem` can be much smaller than `1em`, as seen here. so the border radius needs to be in `em` not in `rem` to adjust accordingly.

do we want to
a) change it to `1rem` or
b) have me add a comment?

mapsandapps: if we set it explicitly to 16px, it looks like this at large font sizes:

Screenshot 2023-07-27 at 10 40 54 AM

core/src/components/chip/chip.ios.vars.scss
thetaPC: This looks good to me.

core/src/components/chip/chip.scss
```diff
@@ -19,13 +19,12 @@
 
   align-items: center;
 
-  height: 32px;
+  min-height: calc(32 / 14 * 1em);
```
thetaPC, line 22: What happens if the `32px` is left in?

mapsandapps: If we leave a min-height of 32px, then of course chips at small font sizes are now large. I suppose this might actually be what we want, but the ticket has a requirement of dynamic heights, so I assumed that meant in both directions. (But perhaps not.)

mapsandapps: Another option is that we could remove the height & min-height entirely and they'll just scale on their own. I think the only chips that look weird in this situation are chips that only have an icon, no label. I'm not sure if those are even really supported.

thetaPC: Good point, why not use `2rem`? It would scale properly while keeping the default of 32px.

liamdebeasi: > but the ticket has a requirement of dynamic heights, so I assumed that meant in both directions. (But perhaps not.)

"dynamic height" means that the height can adjust according to the size of the text. In this case, `min-height: 32px` is what we want. This will retain the correct dimensions with the default text size but still allow the chip to scale up if the font size increases.

core/src/components/chip/chip.ios.scss
```diff
@@ -0,0 +1,26 @@
...
+::slotted(ion-avatar) {
+  width: calc($chip-ios-font-size * 24 / 14);
+  height: calc($chip-ios-font-size * 24 / 14);
+}
+
+// Chip Icon
+// ---------------------------------------------
+
+::slotted(ion-icon) {
+  font-size: calc($chip-ios-font-size * 20 / 14);
```
liamdebeasi, line 25: What do these 24 and 14 values represent?

mapsandapps: this is essentially providing a min & max size to the icon & avatar (for ios mode only). 
24 is the avatar's original width & height; 20 is the icon's original font size, and 14 is the font-size of the chip label when it's not being scaled. so here we're providing it a max & min value when the chip label font size has one.

i suppose i can convert these to say

```
::slotted(ion-avatar) {
  width: clamp(22px, 1.5rem, 38px);
  height: clamp(22px, 1.5rem, 38px);
}

::slotted(ion-icon) {
  font-size: clamp(18px, 1.25rem, 32px);
}
```

which does approximately the same thing. but these numbers feel more "magical" and less precise to me than the numbers i had before, which were just the original sizes.

mapsandapps: also, i couldn't find anywhere in the design doc or anywhere else where we should get the max & min ios font sizes from. so my earlier code had just two "made up" numbers, whereas this version has 6 of them.

core/src/components/chip/chip.scss
```diff
@@ -104,13 +101,14 @@
...
 }
 
-
 // Chip Avatar
 // ---------------------------------------------
 
 ::slotted(ion-avatar) {
-  width: 24px;
-  height: 24px;
+  flex-shrink: 0;
```
liamdebeasi, line 108: Why was this added?

core/src/components/chip/chip.scss
```diff
@@ -67,33 +65,32 @@
...
+  background: rgba($text-color-rgb, 0.08);
 }
 
-
 // Chip Icon
 // ---------------------------------------------
 
 ::slotted(ion-icon) {
-  font-size: 20px;
+  font-size: 1.25em;
```
liamdebeasi, line 89: This isn't quite the same as what 20px was doing. `1.25em` is going to use the computed font-size of `ion-chip` as the basis. Given a root font size of `16px`, the `ion-chip` in your PR will have a computed font size of `14px`. 

1.25 * 14 = 17.5px which means `ion-icon` will be smaller than in `main`. This is likely why you are seeing changed screenshots. (Same goes for the proposed `ion-icon` change).

In this case, you probably want to stick with `rem` instead of `em`.

mapsandapps:
oof, i accidentally pulled `main` (again) and then had to force-push to fix it. sorry y'all 😕

core/src/components/chip/chip.scss
mapsandapps: This was part of the fix for that (pre-existing) bug that I discovered while working this ticket. I added the info for that to the PR description. Without this line, the avatars look like this if the chip width is narrower than its contents:

Screenshot 2023-08-02 at 5 02 35 PM

core/src/components/chip/chip.scss
liamdebeasi: Can't we use `clamp` to enforce a min/max?

i..e `clamp(13px, 1.25rem, 22px)` I made the max/min values up, but that's the general idea.

mapsandapps: yes, we can do that. but if we do, we have to separate back out the ios & md styles for these. (because the md ones don't have a max or min.) using em seemed simpler.

liamdebeasi: hmm so maybe I'm not fully understanding what you are trying to accomplish with getting the icons/avatars to scale.

My understanding of the design document is both the text and the icons should scale according to the root font size. On iOS, both the text and the icons should have max/min font sizes.

The reason why I originally commented was the proposed change is causing visual regressions which is going to cause layout changes in developer applications.

Could you clarify what you are trying to accomplish? I think I'll be able to give helpful feedback once I understand this a bit better.

mapsandapps: option 1: how it is in this PR right now. 

note: this should not change anything visually from how it was, at regular font sizes. in my browser, the icons and avatars are precisely the same size they were before this PR. i'm not sure what's up with the screenshots. i don't visually see a difference in the screenshots either.

```scss
// ios only
:host {
  font-size: clamp(13px, 0.875rem, 22px);
}

// md only
:host {
  font-size: 0.875rem;
}

// ios & md
::slotted(ion-icon) {
  font-size: calc(20em / 14);
}
::slotted(ion-avatar) {
  width: calc(24em / 14);
  height: calc(24em / 14);
}
```

option 2

```scss
// ios only
:host {
  font-size: clamp(13px, 0.875rem, 22px);
}
::slotted(ion-icon) {
  // we will have to remember to change the 19 and/or 31 here if we change 13 and/or 22 above... and remember how to calculate them
  // alternatively, we can make 13px and 22px variables above and utilize those here, but that makes it harder to read
  font-size: clamp(19px, 1.25rem, 31px);
}
::slotted(ion-avatar) {
  // we will have to remember to change the 22 and/or 38 here if we change 13 and/or 22 above... and remember how to calculate them
  width: clamp(22px, 1.5rem, 38px);
  height: clamp(22px, 1.5rem, 38px);
}

// md only
:host {
  font-size: 0.875rem;
}
::slotted(ion-icon) {
  font-size: 1.25rem;
}
::slotted(ion-avatar) {
  width: 1.5rem;
  height: 1.5rem;
}
```

i prefer option 1, because of the reasons in the comments of option 2 and because option 1 is simpler and more similar between ios & md.

liamdebeasi: Ok I think I understand now. You want to size the icons relative to the text initially to avoid visual diffs, but you also want it to scale according to the parent font size (i.e. the size of the text) and account for the max/min font sizes.

I think the confusing part for me is it's not immediately clear where the "14" comes from since the text font size is a rem not a pixel value anymore.

I think we could use some sass variables to make this a bit easier to understand.

Some rough code, but this might work:

```scss
$chip-font-size: 14;
$avatar-font-size: math.div(24em, $chip-font-size);

::slotted(ion-avatar) {
  flex-shrink: 0;

  width: $avatar-font-size;
  height: $avatar-font-size;
}
```

This way it's clear that the 14 is referring to the original font size. The other benefit is we can do these calculations on build and not at runtime (which is what `calc` was doing). I can think on this more tomorrow to see if there are other ways to do this.

thetaPC: As someone who was also confused on the "14", I really like the idea of variables that Liam is proposing. So +1 on that.

mapsandapps: @liamdebeasi @thetaPC this does make sense to me. i updated the PR

thetaPC (approved):
LGTM

core/src/components/chip/test/a11y/chip.e2e.ts
```diff
@@ -0,0 +1,67 @@
+import { expect } from '@playwright/test';
+import { configs, test } from '@utils/test/playwright';
+
+configs({ directions: ['ltr'], modes: ['ios', 'md'] }).forEach(({ title, screenshot, config }) => {
```
liamdebeasi, line 4:
```suggestion
configs({ directions: ['ltr'] }).forEach(({ title, screenshot, config }) => {
```

Our test infra defaults to testing both md and ios

core/src/components/chip/chip.scss
```diff
@@ -19,13 +20,12 @@
 
   align-items: center;
 
-  height: 32px;
+  min-height: 32px;
```
liamdebeasi, line 23: Ok so I figured out why there are screenshot diffs:

The height of the `ion-label` is 17px. This caused the intrinsic height of `ion-chip` to be 17px. Add in the 14px for padding (7px on top 7px on bottom) and 2px for border (1px on top 1px on bottom) and you have 33px.

We previously had `height: 32px` which forced the height of the chip to be `16px` since 16 + 14 + 2 = 32.

We should be able to correct this by reducing the padding from 7px to 6px.

liamdebeasi:
I _think_ the changed screenshots are left over from the first update screenshots job, but the diff threshold is so small that Playwright isn't picking it up. I reverted them in https://github.com/ionic-team/ionic-framework/pull/27868/commits/5db6f1cf1db9e0060584ca523c27972b1e8748c3 to see if they are needed. However, if they are indeed needed I'll revert the commit

PR #27870, opened 2023-07-27
fix(input, textarea): input does not block label

liamdebeasi:
Issue number: #

---------

## What is the new behavior?

-
-
-

- [ ] Yes
- [ ] No

### DIFF
--- a/core/src/components/input/input.md.outline.scss
+++ b/core/src/components/input/input.md.outline.scss
@@ -82,8 +82,6 @@
 :host(.input-fill-outline) .label-text-wrapper,
 :host(.input-fill-outline) .label-text-wrapper {
   position: relative;
-
-  z-index: 1;
 }
 
 /**
--- a/core/src/components/input/input.scss
+++ b/core/src/components/input/input.scss
@@ -581,6 +581,8 @@
   @include transform-origin(start, top);
 
   max-width: 100%;
+
+  z-index: 1;
 }
 
 /**
--- a/core/src/components/textarea/textarea.md.outline.scss
+++ b/core/src/components/textarea/textarea.md.outline.scss
@@ -82,8 +82,6 @@
 :host(.textarea-fill-outline) .label-text-wrapper,
 :host(.textarea-fill-outline) .label-text-wrapper {
   position: relative;
-
-  z-index: 1;
 }
 
 /**
--- a/core/src/components/textarea/textarea.scss
+++ b/core/src/components/textarea/textarea.scss
@@ -629,6 +629,8 @@
   @include padding(0px);
 
   max-width: 100%;
+
+  z-index: 1;
 }
 
 /**


liamdebeasi:
The textarea screenshot that changed looks to be due to a browser quirk in Firefox:

image

Changing the z-index resulted in a 1px diff at the edge of the input. I think this change is fine given that it appears to be a browser rendering quirk.

packages/angular/test/README.md
```diff
@@ -65,6 +65,14 @@ If you want to add a version-specific change, add the change inside of the appro
 
 If you need to add E2E tests that are only run on a specific version of the JS Framework, replicate the `VersionTest` component on each partial application. This ensures that tests for framework version X do not get run for framework version Y.
 
+### Testing Lazy Loaded Ionic Components
```
sean-perkins, line 68: This change should be reverted. It probably just got tracked between switching branches.

sean-perkins (changes requested):

packages/angular/test/README.md
liamdebeasi: Well that explains why the docs didn't show up the first time I thought I pushed to https://github.com/ionic-team/ionic-framework/pull/27867 🤦 

Fixed in https://github.com/ionic-team/ionic-framework/pull/27870/commits/0713c14368fecdd904a756e9878b06e04217fe6f

mapsandapps (approved):
Looks good. I confirmed that the fix works (and I saw the bug submitter confirmed as well), and it looks like everything else works like before. 👍🏻

core/src/components/input/test/label-placement/input.e2e.ts
```diff
@@ -213,4 +210,23 @@ configs({ modes: ['md'], directions: ['ltr'] }).forEach(({ title, screenshot, co
       expect(await input.screenshot()).toMatchSnapshot(screenshot(`input-async-label`));
     });
   });
+  test.describe(title('input: floating/stacked label layering'), () => {
+    test('label should not be covered by text field', async ({ page }) => {
```
thetaPC, line 214: Let's add a `testInfo.annotations` to track the reason behind this test.

core/src/components/textarea/test/label-placement/textarea.e2e.ts
```diff
@@ -302,4 +302,23 @@ configs({ modes: ['md'], directions: ['ltr'] }).forEach(({ title, screenshot, co
       await expect(page.locator('.container')).toHaveScreenshot(screenshot(`textarea-multi-line-sizing`));
     });
   });
+  test.describe(title('textarea: floating/stacked label layering'), () => {
+    test('label should not be covered by text field', async ({ page }) => {
```
thetaPC, line 306: Let's add a `testInfo.annotations` to track the reason behind this test.

thetaPC (changes requested):

core/src/components/textarea/test/label-placement/textarea.e2e.ts
liamdebeasi: Fixed in https://github.com/ionic-team/ionic-framework/pull/27870/commits/9ffc1d933c2d295ed832c23162f0b741679e903e

core/src/components/input/test/label-placement/input.e2e.ts
liamdebeasi: Fixed in https://github.com/ionic-team/ionic-framework/pull/27870/commits/9ffc1d933c2d295ed832c23162f0b741679e903e

thetaPC (approved):
LGTM

PR #27871, opened 2023-07-27
feat(buttons): add dynamic type support

amandaejohnston:
Issue number: N/A

---------

`ion-buttons` uses static font sizes.

## What is the new behavior?

iOS icon font sizes converted to em, to mirror MD. I stuck with em rather than rem so that the icons' sizes match their containing button, avoiding us need to control min/max font sizes for both the button text and the icon.

I'm unsure if anything else needs doing here; it seems like most of the work will be in button and toolbar. Similarly, a screenshot test at this point would be premature since button and toolbar need updating first.

- [ ] Yes
- [x] No


brandyscarney:
I'm going to close this as I've remade the branch off of FW-4422 to use some of those values! 🙂

PR #27872, opened 2023-07-27
fix(datetime-button): buttons wrap instead of truncate

liamdebeasi:
Issue number: N/A

---------

During the design phase of Dynamic Type, I mistakenly wrote the datetime button requirements based off an outdated version of iOS.

On iOS 15, the text inside of the buttons for the native datetime button get truncated with ellipsis.

On iOS 16, the buttons themselves wrap to the next line and the text does not get truncated.

| ios 15 | ios 16 |
| - | - | 
| IMG_8802 | IMG_0067 |

## What is the new behavior?

- Datetime buttons now wrap to the next line instead of truncating the text inside of it

- [ ] Yes
- [x] No

## Other information

https://github.com/ionic-team/ionic-framework-design-documents/pull/131


core/src/components/datetime-button/datetime-button.scss
```diff
@@ -6,17 +6,17 @@
...
   justify-content: center;
-
-  min-width: 0;
 }
 
 :host button {
   @include border-radius(8px);
   @include padding(6px, 12px, 6px, 12px);
-  @include margin(0px, 2px, 0px, 2px);
+  @include margin(2px);
```
liamdebeasi, line 19: This ensures there is some space between the date and time buttons if they wrap to the next line

core/src/components/datetime-button/test/a11y/datetime-button.e2e.ts
```diff
@@ -41,10 +41,10 @@ configs({ directions: ['ltr'], modes: ['ios'] }).forEach(({ title, screenshot, c
 
       const datetimeButton = page.locator('ion-datetime-button');
 
-      await expect(datetimeButton).toHaveScreenshot(screenshot(`datetime-button-scale-truncate`));
+      await expect(datetimeButton).toHaveScreenshot(screenshot(`datetime-button-scale-wrap`));
     });
 
-    test('should truncate text on larger font sizes inside of an item', async ({ page }) => {
+    test('should wrap text on larger font sizes inside of an item', async ({ page }) => {
```
brandyscarney, line 47: Optional: We might be able to remove this test if the behavior doesn't need any special CSS rules to work. I added it because the item specifically needed that `min-width: 0` added in order to truncate.

liamdebeasi: Oh ok good to know! I'll remove it then

liamdebeasi: Removed in https://github.com/ionic-team/ionic-framework/pull/27872/commits/bac8632bc70dc424c4b08e875efe3f692a92f7d3

brandyscarney (approved):

amandaejohnston (approved):

PR #27876, opened 2023-07-27
feat(angular): add standalone nav

thetaPC:
Issue number: internal

---------

Nav does not have Angular standalone component support.

## What is the new behavior?

- Added nav as a standalone component
- Added a test

- [ ] Yes
- [x] No

## Other information

N/A


sean-perkins (approved):
One non-blocking comment.

PR #27883, opened 2023-07-28
feat(angular): add standalone popover

liamdebeasi:
Issue number: N/A

---------

Popover cannot be used as a standalone component

## What is the new behavior?

- Added popover as a standalone component
- Added a test
- Lazy and standalone popover now extend from the same base implementation

- [ ] Yes
- [x] No


packages/angular/common/src/utils.ts
```diff
@@ -0,0 +1,59 @@
+// TODO: Is there a way we can grab this from angular-component-lib instead?
```
liamdebeasi, line 1: So the main problem here is these functions are originally defined in `src/angular-component-lib` which is treated as a separate module than what we are in currently. Former is in `@ionic/angular` latter is in `@ionic/angular/common`.

I could export these functions from `@ionic/angular`, but then they'd be public API which seems strange. I could also have these files be generated in the common directory, but then that means the autogenerated lazy loaded components will be added to the common directory too.

Given that these functions do not often change, the best approach I have right now is to copy it over. Open to other ideas!

liamdebeasi:
I'm experimenting with a simpler way of doing this. Going to move back to a draft for now.

thetaPC (approved):
LGTM

packages/angular/common/src/overlays/popover.ts
```diff
@@ -0,0 +1,121 @@
...
+  /**
+   * Emitted after the popover has dismissed.
+   */
+  ionPopoverWillDismiss: EventEmitter<CustomEvent>;
+  /**
+   * Emitted after the popover has dismissed.
+   */
+  ionPopoverDidDismiss: EventEmitter<CustomEvent>;
+  /**
+   * Emitted after the popover has presented. Shorthand for ionPopoverWillDismiss.
```
sean-perkins, line 32:
```suggestion
   * Emitted after the popover has presented. Shorthand for ionPopoverDidPresent.
```

sean-perkins (changes requested):
Not related to your changes directly, but we should fix this documentation mistake while we are here.

sean-perkins (approved):

PR #27887, opened 2023-07-28
feat(button): support dynamic type

brandyscarney:
Issue number: N/A

---------

The button does not support dynamic type.

## What is the new behavior?
Adds support for dynamic type by converting `font-size` to use `rem` units and setting a `max-size` which matches native iOS.

- [ ] Yes
- [x] No

## Other information

Dynamic type can be enabled with the following CSS:

```css
html {
}
```


core/src/components/button/button.ios.vars.scss
```diff
@@ -34,7 +34,7 @@ $button-ios-height:                                   3.1em !default;
 $button-ios-border-radius:                            14px !default;
 
 /// @prop - Font size of the button text
-$button-ios-font-size:                                16px !default;
+$button-ios-font-size:                                min(1rem, 22px) !default;
```
brandyscarney, line 37: I couldn't find a native example of a solid button to compare against but I did find a clear one and since that was 1px larger before the conversion I just decreased the max size by `1px` compared to clear:

```scss
$button-ios-clear-font-size:                          min(1.0625rem, 23px) !default;
```

core/src/components/button/button.ios.vars.scss
```diff
@@ -108,7 +108,7 @@ $button-ios-small-height:                             2.1em !default;
 $button-ios-small-border-radius:                      6px !default;
 
 /// @prop - Font size of the small button
-$button-ios-small-font-size:                          13px !default;
+$button-ios-small-font-size:                          0.8125rem !default;
```
brandyscarney, line 111: Should I set the max size here to `4px` less than clear?

```scss
$button-ios-small-font-size:                          min(0.8125rem, 19px) !default;
```

core/src/components/button/button.ios.vars.scss
```diff
@@ -83,7 +83,7 @@ $button-ios-large-height:                             3.1em !default;
 $button-ios-large-border-radius:                      16px !default;
 
 /// @prop - Font size of the large button
-$button-ios-large-font-size:                          20px !default;
+$button-ios-large-font-size:                          1.25rem !default;
```
brandyscarney, line 86: Should I set the max size here to `3px` more than clear?

```scss
$button-ios-small-font-size:                          min(0.8125rem, 26px) !default;
```

brandyscarney, line 86:
```suggestion
$button-ios-large-font-size:                          min(1.25rem, 26px) !default;
```

Should I set the max size here to `3px` more than clear?

core/src/components/button/button.ios.vars.scss
brandyscarney, line 111:
```suggestion
$button-ios-small-font-size:                          min(0.8125rem, 19px) !default;
```

Should I set the max size here to 4px less than clear?

core/src/components/button/button.ios.vars.scss
brandyscarney: Thank you! I haven't even been testing going up to 310%...I need to take another look at these values.

core/src/components/button/button.ios.vars.scss
```diff
@@ -34,7 +34,7 @@ $button-ios-height:                                   3.1em !default;
 $button-ios-border-radius:                            14px !default;
 
 /// @prop - Font size of the button text
-$button-ios-font-size:                                16px !default;
+$button-ios-font-size:                                min(1rem, 48px) !default;
```
brandyscarney, line 37: I set the maximums to the value at 310% on `ios`. The actual value of the `font-size` at 310% is in between 46 and 47px but I thought it would be nicer to use a value 3x the normal one.

core/src/components/button/button.ios.vars.scss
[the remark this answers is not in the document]
brandyscarney: I updated the max to use the values at `310%` like we discussed. This is now even more of an issue if you have really long buttons. I think it should be fine though because we have the docs that recommend [updating the `white-space` property to `"normal"`](https://ionicframework.com/docs/api/button#overflowing-text-content) and the `min-height` change went out in `7.2`.

core/src/components/button/button.ios.vars.scss
```diff
@@ -151,13 +151,13 @@ $button-ios-outline-background-color-focused:         ion-color(primary, base, $
 // --------------------------------------------------
 
 /// @prop - Font size of the clear button
-$button-ios-clear-font-size:                          17px !default;
+$button-ios-clear-font-size:                          min(1.0625rem, 51px) !default;
```
brandyscarney, line 154: Same as with the solid, the actual value of the `font-size` at `310%` is closer to `52px` but I thought 3x would be nicer.

core/src/components/button/button.ios.vars.scss
```diff
@@ -83,7 +83,7 @@ $button-ios-large-height:                             3.1em !default;
 $button-ios-large-border-radius:                      16px !default;
 
 /// @prop - Font size of the large button
-$button-ios-large-font-size:                          20px !default;
+$button-ios-large-font-size:                          min(1.25rem, 60px) !default;
```
brandyscarney, line 86: I used 3x the value of the `font-size` at `100%` to calculate the `max`. In this case `20px * 3`.

core/src/components/button/button.ios.vars.scss
```diff
@@ -108,7 +108,7 @@ $button-ios-small-height:                             2.1em !default;
 $button-ios-small-border-radius:                      6px !default;
 
 /// @prop - Font size of the small button
-$button-ios-small-font-size:                          13px !default;
+$button-ios-small-font-size:                          min(0.8125rem, 39px) !default;
```
brandyscarney, line 111: I used 3x the value of the `font-size` at `100%` to calculate the `max`. In this case `13px * 3`.

core/src/components/button/button.ios.vars.scss
liamdebeasi: Do you think it's worth documenting how we derived these values somewhere?

liamdebeasi (approved):
Looks good, great work! One outstanding question.

core/src/components/button/button.ios.vars.scss
brandyscarney: Yeah probably. If you agree with the approach I could add comments above the values.

liamdebeasi: yeah I think doing the 3x scaling makes sense

Issue #27877, opened 2023-07-28
bug: IonPopover size "auto" does not match content 


Facj:
### Prerequisites

- [X] I have read the [Contributing Guidelines](https://github.com/ionic-team/ionic-framework/blob/main/.github/CONTRIBUTING.md#creating-an-issue).
- [X] I agree to follow the [Code of Conduct](https://ionicframework.com/code-of-conduct).
- [X] I have searched for [existing issues](https://github.com/ionic-team/ionic-framework/issues) that already report this problem, without success.

### Ionic Framework Version

v7.x

### Current Behavior

[According to the documentation](https://ionicframework.com/docs/api/popover#css-custom-properties), IonPopover width will match popover content if type "auto" is selected in size attribute.
However, it forces width to 250px, the default size described in [popover-md-width](https://github.com/ionic-team/ionic-framework/blob/main/core/src/components/popover/popover.md.vars.scss) and does not adapt to wider content.

### Expected Behavior

IonPopover width should adapt to content width when attribute size = "auto".

### Steps to Reproduce

I have created a minimal app with three different pages that can be accessed using buttons in Home page. All of them include IonPopover or IonSelect with popover interface. Popover width does not match wide content in any of them.

Page1 includes a popover that can be opened using a button.
Page 2 include a form with IonSelect fields with popover interface.
Page 3 is the same as Page 2 but the form in nested inside a more complex page content, that matchs closely my real app.

### Code Reproduction URL

https://github.com/Facj/ionpopover-issue

### Ionic Info

$ ionic info

Ionic:

   Ionic CLI       : 6.6.0 (C:\Users\fcastro.NORVENTO\AppData\Roaming\npm\node_modules\@ionic\cli)
   Ionic Framework : @ionic/react 7.2.1

Utility:

   cordova-res : not installed
   native-run  : 1.7.2

System:

   NodeJS : v16.15.1 (C:\Program Files\nodejs\node.exe)
   npm    : 9.4.2
   OS     : Windows 10

### Additional Information

[...]

PR #27894, opened 2023-07-31
docs(popover): clarify size property

liamdebeasi:
Issue number: N/A

---------

See: https://github.com/ionic-team/ionic-framework/issues/27877

The current description does not accurately describe what `size="auto"` does. 

## What is the new behavior?

- Description clarifies that the width of the popover is set based on platform defaults.

- [ ] Yes
- [x] No

### DIFF
--- a/core/src/components.d.ts
+++ b/core/src/components.d.ts
@@ -2176,7 +2176,7 @@ export namespace Components {
          */
         "side": PositionSide;
         /**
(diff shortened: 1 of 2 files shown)


amandaejohnston (changes requested):
"Mode defaults" doesn't really mean anything; the default could be to expand based on popover content, fill the whole screen, etc. Maybe something like "the popover width will be set to a static default value" would work better?

PR #27896, opened 2023-07-31
feat(buttons): support dynamic type

brandyscarney:
Issue number: N/A

---------

Toolbar buttons do not support dynamic type.

## What is the new behavior?
- Adds support for dynamic type by converting `font-size` to use `rem` units. 
- Updates the button width/height for icon only buttons in `md` to use `rem` units to scale the icon larger
- Adds test for scaling the font which checks default buttons, clear buttons (because they use a different `font-size`), buttons with an icon, and a button containing only an icon

- [ ] Yes
- [x] No

## Other information

| 80% | 100% | 135% |
| ---| ---| ---|
| ios 80 | ios 100 | ios 135 |
| md 80 | md 100 | md 135 |


core/src/components/button/button.ios.scss
```diff
@@ -34,7 +34,8 @@
  * can add the different font size/weight in a toolbar
  * but still let "size"/"strong" override it.
  */
-:host(.in-buttons) {
+:host(.in-buttons),
+:host(.in-buttons.button-clear) {
```
brandyscarney, line 38: This selector is necessary because the `.button-clear` class comes after this and overrides this class. Alternatively we could move this down below the button fills, while leaving it above the size/strong selectors.

core/src/components/buttons/buttons.ios.scss
```diff
@@ -81,7 +81,7 @@
   @include margin(0);
   @include margin-horizontal(null, .3em);
 
-  font-size: 24px;
+  font-size: 1.4em;
```
brandyscarney, line 84: This value is calculated off of `17px` which is the default `font-size` here, not `16px`

core/src/components/buttons/buttons.ios.scss
```diff
@@ -90,7 +90,7 @@
   @include margin(0);
   @include margin-horizontal(.4em, null);
 
-  font-size: 24px;
+  font-size: 1.4em;
```
brandyscarney, line 93: This value is calculated off of `17px` which is the default `font-size` here, not `16px`

core/src/components/buttons/buttons.ios.scss
```diff
@@ -99,7 +99,7 @@
   @include padding(0);
   @include margin(0);
 
-  font-size: 28px;
+  font-size: 1.65em;
```
brandyscarney, line 102: This value is calculated off of `17px` which is the default `font-size` here, not `16px`

core/src/components/button/test/strong/button.e2e.ts-snapshots/button-clear-strong-in-buttons-ios-ltr-Mobile-Chrome-linux.png
brandyscarney, line 1: This is an undesired effect, but it is caused by the selector I added [here](https://github.com/ionic-team/ionic-framework/pull/27896/files#r1279588081). I will wait for a response to that, but I think we should just move the selector down instead of adding a `:not(.button-strong)` to it.

core/src/components/toolbar/toolbar.ios.vars.scss
```diff
@@ -32,7 +32,7 @@ $toolbar-ios-padding-bottom:                        $toolbar-ios-padding-top !de
 $toolbar-ios-padding-start:                         $toolbar-ios-padding-end !default;
 
 /// @prop - Font size of the toolbar button
-$toolbar-ios-button-font-size:                      17px !default;
+$toolbar-ios-button-font-size:                      clamp(17px, 1.0625rem, 21px) !default;
```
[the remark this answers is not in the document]
liamdebeasi: Good catch! Mind updating the design doc when you get a sec? https://github.com/ionic-team/ionic-framework-design-documents/blob/main/projects/ionic-framework/accessibility/dynamic-type/component-changes.md#buttons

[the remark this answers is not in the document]
brandyscarney: Absolutely! I wanted to make sure this was a good change first haha.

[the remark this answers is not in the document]
brandyscarney: Done: https://github.com/ionic-team/ionic-framework-design-documents/pull/134

core/src/components/toolbar/test/basic/toolbar.e2e.ts-snapshots/toolbar-basic-ios-ltr-Mobile-Chrome-linux.png
brandyscarney, line 1: The difference here is a fraction of a pixel. It is caused by the conversion of `ion-icon` to `em` units.

main

```html
<ion-icon slot="end">

24px width/height
9.6px margin
78.88px wide
```

---

branch

```html
<ion-icon slot="end">

23.97px width/height
9.588px margin
78.84px wide
```

core/src/components/buttons/buttons.ios.scss
brandyscarney, line 1: The conversions for `ion-icon`'s `font-size` are calculated off of the base value of `17px`, not `16px`:

`24px` = `24 / 17` = `1.412em`
`28px` = `28 / 17` = `1.647em`

liamdebeasi (commented):
FYI I think you need to sync with FW-4146 since it's showing the button changes that were already merged

brandyscarney:
I synced it with `FW-4146` but it's still showing them as changes. Maybe I need to rebase this.

liamdebeasi (approved):
Great job!

PR #27897, opened 2023-07-31
fix(datetime): add animation to month/year picker arrow for clarity

liamdebeasi:
Issue number: N/A

---------

As part of [ISSUE] we received feedback that it was not immediately clear how to return to the main grid interface after opening the month/year picker. Native iOS adds an animation to the arrow as an indication that the button toggles the visibility of the month/year picker. However, Ionic's datetime does not have an animation.

## What is the new behavior?

-
-
-

- [ ] Yes
- [x] No

### DIFF
--- a/core/src/components/datetime/datetime.tsx
+++ b/core/src/components/datetime/datetime.tsx
@@ -1917,7 +1917,7 @@ export class Datetime implements ComponentInterface {
    */
 
   private renderCalendarHeader(mode: Mode) {
-    const collapsedIcon = mode === 'ios' ? chevronForward : caretDownSharp;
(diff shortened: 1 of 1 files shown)


core/src/components/datetime/datetime.scss
```diff
@@ -243,7 +243,9 @@ ion-picker-column-internal {
 }
 
 :host .calendar-action-buttons ion-item ion-icon {
-  @include padding(0, 0, 0, 4px);
+  @include padding(4px);
```
liamdebeasi, line 246: This change was needed otherwise the arrow would be shifted down 4px on iOS mode

liamdebeasi:
Tests are now failing for some reason, and I don't have the bandwidth to look into that right now.

PR #27898, opened 2023-07-31
fix(alert): radio and checkbox labels wrap to next line

### LINKED ISSUE #17269: ion-select with long text options doesn't render properly on ios and md
# Bug Report

**Ionic version:**
4.0.0

**Current behavior:**
Long ion-select-option won't be displayed properly in the pop-up compared to ionic 3

**Expected behavior:**
Long ion-select-option should properly size the item and allow the user to see all the text

**Steps to reproduce:**

1. Create ionic 4 project
2. Add <ion-select> with 2 options: 1 with very long text and 1 with short text
3. Open the ion select. The first element should display only 30-40 characters and hyphens

ios example
image

md example
image

if you do the same with an ionic 3 project, this should show properly both of the lines:

ios example
image

md example
image

**Other information:**

Setting `white-space: normal;` on the alert-radio-label does help a bit to show longer text but still it is maximized to 2 lines.
**Ionic info:** 

```
Ionic:
   ionic (Ionic CLI)             : 4.9.0 (C:\Users\Cactus\AppData\Roaming\npm\node_modules\ionic)
   Ionic Framework               : @ionic/angular 4.0.0
   @angular-devkit/build-angular : 0.12.3
   @angular-devkit/schematics    : 7.2.3
   @angular/cli                  : 7.2.3
   @ionic/angular-toolkit        : 1.2.2

Cordova:

   Cordova Platforms     : none
   Cordova Plugins       : no whitelisted plugins (0 plugins total)

System:

   NodeJS : v8.12.0 (C:\Program Files\nodejs\node.exe)
   npm    : 6.4.1
   OS     : Windows 10
```

liamdebeasi:
Issue number: resolves #17269

---------

Long radio and checkbox labels truncate with ellipsis instead of wrapping to the next line. This makes the label hard to understand because users cannot read the entire label. 

## What is the new behavior?

- Labels now wrap to the next line instead of truncating.

Note: Some of the screenshots may show wider alerts than before. This is the correct behavior. When rendering text, the browser will try to render as much text on a single line as it can. Since our alerts can expand in width, the alert wrapper will expand in width (up to the max width) to accommodate this text. Once the the wrapper is at the max width, text will then wrap to the next line.

For example, you may notice the MD alert increasing to 280px in width: https://github.com/ionic-team/ionic-framework/blob/b8553d89f84970db452646ad7a88e2fc0fd88657/core/src/components/alert/alert.md.vars.scss#L11.

There should be no visual diff for alerts with text that would not normally wrap to the next line.

This was not happening before because we did not allow text to wrap to the next line.

- [ ] Yes
- [x] No


core/src/components/alert/alert.ios.scss
```diff
@@ -123,7 +123,7 @@
 }
 
 .alert-tappable {
-  height: $alert-ios-tappable-height;
+  min-height: $alert-ios-tappable-height;
```
liamdebeasi, line 126: The ticket notes that we should set `height: auto`, but I found that created some undesired diffs on radios/checkboxes with short text. I found that `min-height` gets us to the same desired result without the additional diffs for short text.

core/src/components/alert/alert.scss
```diff
@@ -152,6 +152,8 @@
 
   width: 100%;
   height: 100%;
+
+  min-height: inherit;
```
liamdebeasi, line 156: Without this the button inner may have visual diffs since the height is no longer fixed at 48px

core/src/components/alert/alert.scss
```diff
@@ -184,7 +186,7 @@
   text-align: start;
   appearance: none;
 
-  contain: strict;
+  contain: content;
```
liamdebeasi, line 189: Without this, each radio/checkbox will still be sized at 48px. This change allows the height of the element to increase as the text wraps.

core/src/components/alert/test/basic/index.html
```diff
@@ -228,7 +228,8 @@
           inputs: [
             {
               type: 'radio',
-              label: 'Radio 1',
+              label:
```
brandyscarney, line 231: Could we do this in a separate test or move the disabled radio so that it doesn't fall off the screenshot? Maybe since `Radio 6` already has long text that wraps, we could move that and the disabled one up?

```javascript
{
  type: 'radio',
  label: 'Radio 1',
  value: 'value1',
},
{
  type: 'radio',
  label: 'Radio 2',
  value: 'value2',
  disabled: true,
},
{
  type: 'radio',
  label: 'Radio 3 Radio 3 Radio 3 Radio 3 Radio 3 Radio 3 Radio 3 Radio 3 Radio 3 Radio 3',
  value: 'value3',
  checked: true,
},
{
  type: 'radio',
  label: 'Radio 4',
  value: 'value4',
},
{
  type: 'radio',
  label: 'Radio 5',
  value: 'value5',
},
{
  type: 'radio',
  label: 'Radio 6',
  value: 'value6',
},
```

core/src/components/alert/test/basic/index.html
```diff
@@ -284,7 +285,8 @@
           inputs: [
             {
               type: 'checkbox',
-              label: 'Checkbox 1',
+              label:
```
brandyscarney, line 288: Could we do this in a separate test or move the disabled checkbox so that it doesn't fall off the screenshot? Maybe since `Checkbox 6` already has long text that wraps, we could move that and the disabled one up?

```javascript
{
  type: 'checkbox',
  label: 'Checkbox 1',
  value: 'value1',
  checked: true,
},

{
  type: 'checkbox',
  label: 'Checkbox 2',
  value: 'value2',
  disabled: true,
},

{
  type: 'checkbox',
  label: 'Checkbox 3 Checkbox 3 Checkbox 3 Checkbox 3 Checkbox 3 Checkbox 3 Checkbox 3 Checkbox 3 Checkbox 3 Checkbox 3',
  value: 'value3',
},

{
  type: 'checkbox',
  label: 'Checkbox 4',
  value: 'value4',
},

{
  type: 'checkbox',
  label: 'Checkbox 5',
  value: 'value5',
},

{
  type: 'checkbox',
  label: 'Checkbox 6',
  value: 'value6',
},
```

brandyscarney (commented):
Code looks good. I am wondering if we shouldn't separate out the wrapping example. Do we have an existing screenshot for checkbox/radio in an alert without wrapping?

core/src/components/alert/test/basic/index.html
liamdebeasi: Fixed in https://github.com/ionic-team/ionic-framework/pull/27898/commits/fb3d0b621d0c429650ed64a80612a6f971e1eff1. Screenshots are updating now.

brandyscarney (approved):
Looks good to me! The changes to `min-height` and `contain` make sense.

Issue #22895, opened 2021-02-08
bug: ionic vue, swipe back transition gets stuck when swiping quickly


jakubkoje:
# Bug Report

**Ionic version:**
[ ] **4.x**
[x] **5.x**

**Current behavior:**
Sometimes, when swiping back fast, the view gets stuck in left half and prevents navigating to the page I was navigating back from.

**Expected behavior:**
I should be able to swipe with any speed and not cause the application to be stuck on left side of the screen.

**Steps to reproduce:**
```
1. npm install
2. ionic serve
3. Click the button to go to the subpage
4. Swipe back
5. Repeat until the screen gets stuck
```
**Related code:**

https://github.com/jakubkoje/ionic-back-issue

**Other information:**
No errors in console when it happens or when I try to navigate away while the screen is halfway. 

Example video:
https://imgur.com/g0Sy7Zm

**Ionic info:**

```
Ionic:

   Ionic CLI       : 6.12.4 (/home/jakub/.nvm/versions/node/v14.15.4/lib/node_modules/@ionic/cli)
   Ionic Framework : @ionic/vue 5.5.4

Capacitor:

   Capacitor CLI   : 2.4.4
   @capacitor/core : 2.4.4

Utility:

   cordova-res : not installed
   native-run  : not installed

System:

   NodeJS : v14.15.4 (/home/jakub/.nvm/versions/node/v14.15.4/bin/node)
   npm    : 7.5.2
   OS     : Linux 4.19
```

liamdebeasi:
Can you try the following dev build and let me know if it resolves the issue?

```
npm install @ionic/vue@5.6.0-dev.202102082152.e5c304b @ionic/vue-router@5.6.0-dev.202102082152.e5c304b
```

jakubkoje:
Now it seems like instead of being stuck in between, it is stuck on the subpage, not allowing me to navigate back trough swipe or back button.

liamdebeasi:
Thanks, this dev build should fix the issue.

```
npm install @ionic/vue@5.6.0-dev.202102082226.d47cbf7 @ionic/vue-router@5.6.0-dev.202102082226.d47cbf7
```

jakubkoje:
It still gets stuck sometimes. It has to be really slight and fast swipe, so the animation doesn't start rendering, but the element styles change.

liamdebeasi:
>  so the animation doesn't start rendering, but the styles of the components change.

[...]

PR #27904, opened 2023-08-01
fix(nav): improve reliability of swipe back gesture when quickly swiping back

### LINKED ISSUE #27893: bug: nav, swipe back transition gets stuck with swiping quickly
### Prerequisites

### Ionic Framework Version

v7.x

### Current Behavior

Currently, when using `ion-nav` and having at least two transitions between two screens, there is a chance that it will not be possible to go back from the nested page. this behaviour is totally random

### Expected Behavior

All transitions between pages should work correctly, regardless of the nesting of pages and the frequency of transitions between them.

### Steps to Reproduce

1. Set up a basic example using `iov-nav`, `ion-link`, `ion-content` with a root page and the nested one
2. Navigate to the nested page
3. Swipe back
4. Repeat 2 and 3 steps until the bug appears

### Code Reproduction URL

https://replit.com/@RamilMusaiev/React-TypeScript-1#src/App.tsx

### Ionic Info

Since I was integrating ionic on top of the existing project the output is corresponding

```
ionic info

Ionic:

   Ionic CLI : 7.1.1

Utility:

System:

   NodeJS : v18.16.0
   npm    : 9.5.1
   OS     : macOS Unknown
```

### Additional Information

Here is the video:

https://github.com/ionic-team/ionic-framework/assets/5896973/05be8a67-1c80-4cb9-8714-a2c6dc81806f

As you can see, at some point the subpage has the CSS property `pointer-events: none;` even though I was right on the subpage. When I tried to override this property in the browser, the page became responsive again, but the swipe back gesture didn't work.

liamdebeasi:
Issue number: resolves #27893

---------

This is another instance of https://github.com/ionic-team/ionic-framework/issues/22895. The `progressCallback` function is fires asynchronously, so it's possible for the gesture start and end callbacks to run before the animation is ever set in `progressCallback`. When this happens, the animation gets locked up.

I previously fixed this in https://github.com/ionic-team/ionic-framework/pull/23527 for `ion-router-outlet`, but I did not fix it for `ion-nav`.

## What is the new behavior?

- If the gesture has ended by the time `progressCallback` fires, reset the animation to the beginning so it does not get locked up.

- [ ] Yes
- [x] No

## Other information

Dev build: `7.2.2-dev.11690896715.12338339`

### DIFF
--- a/core/src/components/nav/nav.tsx
+++ b/core/src/components/nav/nav.tsx
@@ -31,7 +31,7 @@ import { VIEW_STATE_ATTACHED, VIEW_STATE_DESTROYED, VIEW_STATE_NEW, convertToVie
 export class Nav implements NavOutlet {
   private transInstr: TransitionInstruction[] = [];
   private sbAni?: Animation;
-  private animationEnabled = true;
+  private gestureOrAnimationInProgress = false;
   private useRouter = false;
   private isTransitioning = false;
   private destroyed = false;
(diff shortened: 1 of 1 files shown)


thetaPC (approved):
LGTM, also [verified](https://github.com/ionic-team/ionic-framework/issues/27893#issuecomment-1660600392) by OP.

amandaejohnston (approved):

PR #27909, opened 2023-08-01
refactor(angular): move remaining directives to common module

liamdebeasi:
Issue number: #

---------

See https://github.com/ionic-team/ionic-framework/pull/27899. There are many directives that have not been migrated to the common module.

## What is the new behavior?

Below is a breakdown of each directive that moved and what was changed. Note that `@ionic/angular` provides new directives that extend from the base directives rather than simply importing the base directives directly. This is an Angular restriction as the directives need to be part of the library that is being bundled.

**Router Link Delegates**

- This directive was moved to the common module
- `@ionic/angular` provides new directives that extend from these base directives

**Nav Delegate**

- This directive was moved to the common module
- `@ionic/angular` provides new directives that extend from this base directive

**Back Button**

- This directive was moved to the common module
- `@ionic/angular` provides new directives that extend from this base directive
- We pass the router outlet in to the `@ionic/angular` back button so it gets the correct instance of the router outlet (otherwise it will be undefined)

**Router Outlet**

- This directive was moved to the common module
- `@ionic/angular` provides new directives that extend from this base directive
- I also migrated the Stack Controller, Stack Utils, and Nav Controller to the common module. These are all used by the router outlet, so I had to migrate them all at once.

**Tabs**

- This directive was moved to the common module
- `@ionic/angular` provides new directives that extend from this base directive
- The outlet, tabBar, and tabBars variables need to be re-defined for each child class since we need to reference the correct generated component. See: https://github.com/ionic-team/ionic-framework/pull/27909/files#diff-796a35731da718886679a12f39d9cf8ca5c8a3a009f3e836fda3bf7a14c9b570R22-R29 and https://github.com/ionic-team/ionic-framework/pull/27909/files#diff-0f095f0b33ae4aa88ab6edfcf921737b858bc4bfcf8652b44bc1d6dafd87918bR46-R49. Wasn't sure if this was the best way to do this, so open to other ideas!

- [ ] Yes
- [x] No


thetaPC (approved):
LGTM. I recommend getting a second approval in case there's another opinion on how to handle the outlet, tabBar, and tabBars variables.

sean-perkins (approved):

PR #27917, opened 2023-08-02
fix(datetime): changing months work if partially visible

liamdebeasi:
Issue number: #

---------

## What is the new behavior?

-
-
-

- [ ] Yes
- [ ] No


core/src/components/datetime/test/minmax/datetime.e2e.ts
```diff
@@ -301,21 +301,20 @@ configs({ directions: ['ltr'], modes: ['ios'] }).forEach(({ title, config }) =>
...
+            presentation="month-year"
           ></ion-datetime>
         `,
           config
         );
 
         const datetime = page.locator('ion-datetime');
-        const monthYearToggle = page.locator('ion-datetime .calendar-month-year');
         const monthColumnItems = page.locator('ion-datetime .month-column .picker-item:not(.picker-item-empty)');
+        const ionChange = await page.spyOnEvent('ionChange');
```
liamdebeasi, line 312: This test started to flaky after my changes. This test touches the wheel picker which does not use any of the changed code. I noticed that the test was not waiting for `.datetime-ready` and it was not waiting for the `ionChange` event, so it's possible that my change made datetime slightly faster such that this test is now consistently failing.

core/src/components/datetime/test/set-value/datetime.e2e.ts
```diff
@@ -49,11 +50,9 @@ configs({ modes: ['ios'], directions: ['ltr'] }).forEach(({ title, config }) =>
       await page.waitForChanges();
 
       // Select October 2021
+      // The year will automatically switch to 2021 when selecting 10
       await monthColumn.locator('.picker-item[data-value="10"]').click();
-      await page.waitForChanges();
-
-      await yearColumn.locator('.picker-item[data-value="2021"]').click();
```
liamdebeasi, line 55: This ended up being redundant because clicking "October" automatically switched the year to "2021"

core/src/components/datetime/test/set-value/datetime.e2e.ts
```diff
@@ -49,11 +50,9 @@ configs({ modes: ['ios'], directions: ['ltr'] }).forEach(({ title, config }) =>
...
-      await page.waitForChanges();
+      await ionChange.next();
```
liamdebeasi, line 55: This test also started to fail only in Firefox. The problem was that tapping the "October" button  started to scroll, but we closed the month/year picker before the scroll finished so the value was never updated. By waiting for `ionChange` we can ensure the value is updated.

sean-perkins (approved):

core/src/components/datetime/test/minmax/datetime.e2e.ts
```diff
@@ -301,21 +301,20 @@ configs({ directions: ['ltr'], modes: ['ios'] }).forEach(({ title, config }) =>
           <ion-datetime
             min="2022-01-15"
             value="2022-02-01"
-            presentation="date"
+            presentation="month-year"
```
amandaejohnston, line 304: Was this change intentional?

amandaejohnston (approved):
Nice work, this should make things way more efficient. Just had one quick question about a test but otherwise LGTM.

core/src/components/datetime/test/minmax/datetime.e2e.ts
liamdebeasi: Yup! So this test was checking the month/year picker to make sure it updates the value prop correctly. Previously it was using the grid style layout, clicking the month/year toggle button, then changing the values. This simplifies it by just using the month-year layout to begin with.

amandaejohnston: Ahh, clever 👍

Et3rnal:
Will this fix make it to version 6?
Is there a work around the issue?
I'm only experiencing the issue when using Edge!
I'm ionic/angular `6.1.11`

Thanks

liamdebeasi:
No, this fix is only for Ionic 7. We are not actively developing Ionic 6 anymore.

PR #27921, opened 2023-08-03
feat(toast): add shadow part for cancel button

### LINKED ISSUE #27920: feat: toast, add shadow part to customize cancel button
### Prerequisites

I would like to independently modify the styles of the button with the role "cancel" without affecting the other buttons of the toasts.

### Describe the Use Case

Display the button with the role "cancel" in a different color than the others or modify its position.

To solve this problem, I would like the buttons with the role "cancel" to have a CSS Shadow Part called "button-cancel."

### Describe Alternatives

_No response_

### Related Code

_No response_

### Additional Information

_No response_

luisbytes:
Issue number: resolves #27920

---------

It is not possible to style the buttons with the cancel role in a toast since it uses shadow DOM and there are no exposed parts or CSS variables for the cancel buttons.

## What is the new behavior?

- Exposes the following part of a toast: `button with rol cancel`
- Update the e2e toast test to check if the `button-cancel` part is exposed
-

- [ ] Yes
- [x] No

Change the css shadow part for all buttons added to a toast with the cancel role from 'button' to 'button-cancel'.


liamdebeasi:
Thanks for the PR! We appreciate you contributing to Ionic. I discussed this with the team, and we think this is a good feature to add. Our next step is to do a design document to make sure we are implementing this feature in a way that addresses the original use case.

Once that is done, someone from the team will review this PR and get it ready to merge. Let me know if you have any questions.

core/src/components/toast/toast.tsx
```diff
@@ -568,5 +575,9 @@ const buttonClass = (button: ToastButton): CssClassMap => {
   };
 };
 
+const buttonPart = (button: ToastButton): string => {
+  return isCancel(button.role) ? 'button-cancel' : 'button';
```
liamdebeasi, line 579:
```suggestion
  return isCancel(button.role) ? 'button cancel' : 'button';
```

For parts with state/modifiers, we should add it as multiple parts. This way the cancel button still matches the `button` part in addition to `button cancel`.

core/src/components/toast/toast.tsx
```diff
@@ -34,6 +34,7 @@ import type { ToastButton, ToastPosition, ToastLayout } from './toast-interface'
  * @virtualProp {"ios" | "md"} mode - The mode determines which platform styles to use.
  *
  * @part button - Any button element that is displayed inside of the toast.
+ * @part button-cancel - Any button element with role "cancel" that is displayed inside of the toast.
```
liamdebeasi, line 37: Same as my other comment: Let's call this `button cancel`
```suggestion
 * @part button cancel - Any button element with role "cancel" that is displayed inside of the toast.
```

core/src/components/toast/test/buttons/toast.e2e.ts
```diff
@@ -0,0 +1,16 @@
...
+configs().forEach(({ title, config }) => {
+  test.describe(title('toast: cancel button'), () => {
+    test('should render the cancel button with part button-cancel', async ({ page }) => {
+      await page.goto('/src/components/toast/test/buttons', config);
+
+      const toast = page.locator('#closeArray');
+      await toast.click();
+
+      const buttonCancel = page.locator('.toast-button-cancel');
+      await expect(buttonCancel).toHaveAttribute('part', 'button-cancel');
```
liamdebeasi, line 13: Can we make this a spec test? E2E tests fire up an entire browser instance which is slow and not necessary for the kind of validation we are trying to do.

Here's an example of a spec test in Ionic: https://github.com/ionic-team/ionic-framework/blob/1cf1eca00239f3e98854466116e42f9a2e7b4590/core/src/components/select/test/select.spec.tsx#L2

liamdebeasi (changes requested):
Overall this looks good! A few small changes.

liamdebeasi (changes requested):
Can you also rebase this off `feature-7.3`? This is the next minor release we are targeting.

luisbytes:
@liamdebeasi Hello, I'm ready with the changes you suggested. I failed to do the rebase.

liamdebeasi:
Merged. Thank you very much!

luisbytes:
@liamdebeasi Thank you, I hope to contribute more to the community.

PR #27924, opened 2023-08-03
fix(modal): setCurrentBreakpoint respects animated prop

liamdebeasi:
Issue number: #

---------

## What is the new behavior?

-
-
-

- [ ] Yes
- [ ] No


mapsandapps (approved):
Looks good. I tested it out, and it animates when it should and doesn't when it shouldn't 👍🏻

PR #27946, opened 2023-08-07
Fw 4437

brandyscarney:
Issue number: #

---------

## What is the new behavior?

-
-
-

- [ ] Yes
- [ ] No

### DIFF
--- a/core/src/components/item/item.md.vars.scss
+++ b/core/src/components/item/item.md.vars.scss
@@ -188,7 +188,7 @@ $item-md-label-slot-end-margin-start:       $item-md-label-slot-end-margin-end !
 // --------------------------------------------------
 
 /// @prop - Font size of a note in the start/end slot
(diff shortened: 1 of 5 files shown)


core/src/components/item/item.md.vars.scss
```diff
@@ -188,7 +188,7 @@ $item-md-label-slot-end-margin-start:       $item-md-label-slot-end-margin-end !
 // --------------------------------------------------
 
 /// @prop - Font size of a note in the start/end slot
-$item-md-note-slot-font-size:               11px !default;
+$item-md-note-slot-font-size:               0.6875rem !default;
```
brandyscarney, line 191: Material Design doesn't call for a minimum font size, but it is really small at 80% inside of an item. Should we give it one?

mapsandapps (approved):
looks good 👍🏻

PR #27947, opened 2023-08-07
feat(item-divider): support dynamic type

brandyscarney:
Issue number: N/A

---------

Item dividers do not support dynamic type.

## What is the new behavior?
- Adds test for scaling the font which checks an item divider at 310%

- [ ] Yes
- [x] No

### Native

I used the Contacts app for comparison

| `80%` | `100%` | `135%` | `310%` | 
| ---| ---| ---| ---|
| ios-80 | ios-100 | ios-135 | ios-310 |

### Ionic Framework

| `80%` | `100%` | `135%` | `310%` | 
| ---| ---| ---| ---|
| ios-80 | ios-100 | ios-135 | ios-310 |
| md-80 | md-100 | md-135 | md-310 |


core/src/components/item-divider/item-divider.ios.vars.scss
```diff
@@ -8,7 +8,7 @@
 $item-divider-ios-min-height:             28px !default;
 
 /// @prop - Font size of the item
-$item-divider-ios-font-size:              17px !default;
+$item-divider-ios-font-size:              1.0625rem !default;
```
brandyscarney, line 11: The default value this is set at (`17px`) is actually wrong when compared against native. It should be `15px` (or `0.9375rem`) at 100%. Something to keep in mind if you are comparing this against native.

Updating the default value is outside the scope of this feature & we have an open ticket to investigate changes for iOS 17 so we could update the font size as part of that.

amandaejohnston (approved):

PR #27961, opened 2023-08-09
test(playwright): enable github reporter, test retries

liamdebeasi:
Issue number: N/A

---------

The team would like to explore solutions for being informed of flaky tests in a way that is not disruptive to our workflow. Currently, flaky tests fail immediately which means we have to re-run them every time. We'd like flaky tests to be automatically retried but also reported to us so we can address them in a separate PR.

## What is the new behavior?

- Enables the Playwright GitHub reporter. This will report about flaky tests on the PR if applicable as well as in the CI results.
- Enables test retries. Tests will be retried up to 2 times before failing.
- Disables reporting slow tests in the GitHub reporter. Some of our tests require gesture interaction which are inherently slow but otherwise working as intended. We don't necessarily need to know about these right now.
- Disables "maxFailures". Tests that can fail at most once are never detected as flaky since they are never retried. As a result, we need to disable this in order to have flaky tests be reported to us.

- [ ] Yes
- [x] No

### DIFF
--- a/core/playwright.config.ts
+++ b/core/playwright.config.ts
@@ -69,7 +69,10 @@ const config: PlaywrightTestConfig = {
   /* Opt out of parallel tests on CI. */
   workers: process.env.CI ? 1 : undefined,
   /* Reporter to use. See https://playwright.dev/docs/test-reporters */
-  reporter: 'html',
+  reporter: [
+    ['html'],
+    ['github']
+  ],
   /* Shared settings for all the projects below. See https://playwright.dev/docs/api/class-testoptions. */
   use: {
     /* Maximum time each action such as `click()` can take. Defaults to 0 (no limit). */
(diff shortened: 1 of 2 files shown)


sean-perkins (approved):

core/playwright.config.ts
```diff
@@ -63,13 +63,16 @@ const config: PlaywrightTestConfig = {
   /* Fail the build on CI if you accidentally left test.only in the source code. */
   forbidOnly: !!process.env.CI,
   /* Fail fast on CI */
```
mapsandapps, line 65: i think this comment needs to be removed?

mapsandapps (changes requested):

core/playwright.config.ts
liamdebeasi: Good catch! Addressed in https://github.com/ionic-team/ionic-framework/pull/27973

PR #27962, opened 2023-08-09
feat(label): support dynamic type

brandyscarney:
Issue number: N/A

---------

Labels do not support dynamic type.

## What is the new behavior?
  - I did **not** add `font-size: 1rem` to `ion-label` because doing this breaks the label font sizes inside of other components such as items, list header, and chip (unless you override it with `font-size: inherit` inside of those), when this CSS is not needed for label to dynamically size
- Adds test for scaling the font which checks the default label, label with wrapping text, stacked label and a label containing headings & a paragraph at 310%

I also made changes to `item-divider` in this PR because it used a shared variable with `label` to change the paragraph `font-size`:
- Removed heading styles that were *not* used because they were overridden by [typography](https://github.com/ionic-team/ionic-framework/blob/main/core/src/css/typography.scss#L53-L79). 
- Adds test for headings inside of an `ion-label` in an `ion-item-divider`

- [ ] Yes
- [x] No


core/src/components/label/label.md.scss
```diff
@@ -175,9 +175,9 @@
 ::slotted(*) p {
   @include margin(0, 0, 2px);
 
-  font-size: 14px;
+  font-size: dynamic-font(14px);
 
-  line-height: 20px;
+  line-height: dynamic-font(20px);
```
brandyscarney, line 180: Should this still use the Sass function? It does read a bit weird calling dynamic-_font_.

mapsandapps (approved):
looks good to me. nice cleanup of unused styles 👍🏻

PR #27970, opened 2023-08-10
chore(angular): generate standalone component wrappers

sean-perkins:
Issue number: N/A

---------

Ionic Framework currently only generates the lazy/hydrated Angular component wrappers for the web components.

## What is the new behavior?

- Generates Angular standalone component wrappers for the web components, using the CE build.
  - Adds manual component wrapper for `ion-icon` with the CE build.
  - Migrates the `ion-back-button` and `ion-nav` to be manual component wrappers
- Continues to generate the lazy/hydrated Angular component wrappers.
- Refactors "NavDelegate" etc. language to `IonNav` to avoid exporting as and simplify navigating the code.

- [ ] Yes
- [x] No

## Other information

Related output targets PR: https://github.com/ionic-team/stencil-ds-output-targets/pull/367


core/stencil.config.ts
```diff
@@ -7,6 +7,52 @@ import { vueOutputTarget } from '@stencil/vue-output-target';
...
+      directivesProxyFile: '../packages/angular/standalone/src/directives/proxies.ts',
+      excludeComponents: [
+        ...excludeComponents,
+        /**
+         * IonIcon is a special case because it does not come
+         * from the `@ionic/core` package, so generating proxies that
+         * are reliant on the CE build will reference the wrong
+         * import location.
+         */
+        'ion-icon'
```
liamdebeasi, line 49: Do we need to exclude other components like IonNav and IonBackButton?

sean-perkins: I think we can exclude them, but we should update the existing manual proxies to be `@Component` vs. `@Directive`, detach from Angular's change detection and assign a default template of `<ng-content>` to be consistent with the implementation of those proxies on `main` with the lazy build.

liamdebeasi: That works for me 👍

sean-perkins:
Ok, this is ready for another review. Component wrappers should be generating correctly. `IonNav`, `IonIcon` and `IonBackButton` are all manual component wrappers. 

I've updated the test app to render the standalone tests within the generated `ion-app` standalone component wrapper.

liamdebeasi (approved):

PR #27976, opened 2023-08-10
fix(datetime): reset and cancel updates state correctly

### LINKED ISSUE #27975: bug: Datetime active day does not clear when pressing cancel and re-opening
### Describe the Bug

On pressing default Cancel Button in ion-datetime with ion-datetime-button and opening the ion-datetime again the date that was canceled is preselected.

### Expected Behavior

on pressing the cancel button and opening the ion-datetime again should select the date that wasn't the one that was selected on cancel

### Steps to Reproduce
```
  <ion-datetime-button datetime="datetime"></ion-datetime-button>

  <ion-modal [keepContentsMounted]="true">
    <ng-template>
      <ion-datetime id="datetime"[showDefaultButtons]="true"></ion-datetime>
    </ng-template>
  </ion-modal>
```

  use this and click the button and select a date and then press the cancel button 
  now click on the button again, the date selected will be one selected when we pressed cancel not the one that is in ion-daretime-button

Reproduction: https://stackblitz.com/edit/angular-h94fr8?file=src%2Fapp%2Fexample.component.html

### Screenshots

_No response_

### Operating System

macOS

### Browser

Chrome 

### Version

115.0.5790.170

### Additional Information

_No response_

liamdebeasi:
Issue number: resolves #27975

---------

There are a few problems:

1. The "cancel" method never updates the internal state of the datetime. This means if you select June 6, 2023 and click cancel, that date is still selected even though the value prop has not been updated.
2. The "reset" method forcibly resets the internal state to today's date even if the value prop has been set.
3. The button shown by `showClearButton` sets the state of `ion-datetime` to an `undefined` value. This means that the `reset` method and the clear button do two different behaviors.

## What is the new behavior?

- The "cancel" method now calls "reset" to reset the internal state of the component.
- The "reset" method will reset the internal state to the value prop if the value prop is set and no param is passed to `reset`. If no value was passed and no value is set then it will continue to reset to today.
- The clear button now just calls the "reset" method. This is a slight behavior change from what it does now, but I'd argue that what it is doing now is not correct.
- Added test coverage for the cancel and reset methods

As part of this change, the behavior for `showClearButton` needed to be updated to align with the `reset` method. Curren

- [ ] Yes
- [x] No

## Other information

Dev build: `7.2.4-dev.11691705608.182f8ca8`

### DIFF
--- a/core/src/components/datetime/test/basic/datetime.e2e.ts
+++ b/core/src/components/datetime/test/basic/datetime.e2e.ts
@@ -384,40 +384,6 @@ configs({ modes: ['ios'], directions: ['ltr'] }).forEach(({ title, config }) => 
  * This behavior does not differ across
  * modes/directions.
  */
-configs({ modes: ['ios'], directions: ['ltr'] }).forEach(({ title, config }) => {
-  test.describe(title('datetime: clear button'), () => {
-    test('should clear the active calendar day', async ({ page }, testInfo) => {
-      testInfo.annotations.push({
-        type: 'issue',
-        description: 'https://github.com/ionic-team/ionic-framework/issues/26258',
-      });
-
-      await page.setContent(
-        `
-        <ion-datetime value="2022-11-10" show-clear-button="true"></ion-datetime>
-      `,
-        config
-      );
-
-      await page.waitForSelector('.datetime-ready');
-
-      const selectedDay = page.locator('ion-datetime .calendar-day-active');
-
-      await expect(selectedDay).toHaveText('10');
-
-      await page.click('ion-datetime #clear-button');
-
-      await page.waitForChanges();
-
-      await expect(selectedDay).toHaveCount(0);
-    });
-  });
-});
-
-/**
- * This behavior does not differ across
- * modes/directions.
- */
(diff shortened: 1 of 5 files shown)


core/src/components/datetime/test/basic/datetime.e2e.ts
```diff
@@ -380,40 +380,6 @@ configs({ modes: ['ios'], directions: ['ltr'] }).forEach(({ title, config }) =>
   });
 });
 
-/**
- * This behavior does not differ across
- * modes/directions.
- */
-configs({ modes: ['ios'], directions: ['ltr'] }).forEach(({ title, config }) => {
-  test.describe(title('datetime: clear button'), () => {
```
liamdebeasi, line 388: This is already tested in the "reset" test I added, so I removed this.

liamdebeasi:
Fixing this requires a few other changes. I think we should get team consensus on the how we'd like to proceed. Going to mark as draft for now while I make design changes.

liamdebeasi:
Closing for now until design work can be completed.

PR #27989, opened 2023-08-14
feat(item): support dynamic type

brandyscarney:
Issue number: N/A

---------

Items do not support dynamic type.

## What is the new behavior?
- Converts icon font sizes to `em` units based on their parent element font size
- Adds test for scaling the font which checks an item with only a label, an item with slotted icons, an item with a detail icon, an item with a counter, an item with helper/error text, and an item with buttons at 310%
  - I considered adding a test for an item with headings, but this is covered by the [label PR](https://github.com/ionic-team/ionic-framework/pull/27962) so unless we remove the label from an item one day we don't need the tests in item

I also made changes to `item-divider` in this PR because it used a shared variable with `item` to set the slotted icon `font-size`.

- [ ] Yes
- [x] No


core/src/components/reorder-group/test/a11y/reorder-group.e2e.ts-snapshots/reorder-group-scale-ios-ltr-Mobile-Chrome-linux.png
brandyscarney, line 1: This update is due to the `reorder-group` test using an `ion-item` with an `ion-label`. We could remove the label to avoid updates to this test with future item changes, or leave as is.

core/src/components/item-divider/item-divider.md.scss
```diff
@@ -44,7 +44,10 @@
 ::slotted(ion-icon) {
   color: $item-md-icon-slot-color;
 
-  font-size: $item-md-icon-slot-font-size;
+  // The font size should be 24px which is
+  // calculated in relation to the host's font
+  // size of 14px. e.g. 24 / 14 = 1.714
+  font-size: 1.71em;
```
brandyscarney, line 50: We can't use the `$item-md-icon-slot-font-size` variable here anymore because of the following:

- The value of `$item-md-icon-slot-font-size` before converting to `em` units is `24px`
- **Item** uses a `font-size` of `16px` which converts the value to `1.5em`
- **Item divider** uses a `font-size` of `14px` which, when this value is set to `1.5em`, makes the pixel value equal to 21px
- Therefore, in order to avoid visual diffs we need to specifically set it to `1.71em` for item divider so the converted value is equal to `24px` still

thetaPC: Why not use?:

```scss
$item-md-icon-slot-font-size: 24px;
$item-divider-md-font-size-base: 14px;

font-size: math.div($item-md-icon-slot-font-size, $item-divider-md-font-size-base) * 1em;
```

This would let us keep the variables and make it easy for us to understand the chosen values without the need of comments.

thetaPC, line 50: Why was `1.71em` used instead of `1.714em`? Is using `1.71em` causing the screenshots to be updated?

core/src/components/reorder-group/test/a11y/reorder-group.e2e.ts-snapshots/reorder-group-scale-ios-ltr-Mobile-Chrome-linux.png
thetaPC: It might be best to split them into one with label and one without. This would cover both edge cases.

core/src/components/item-divider/item-divider.md.scss
brandyscarney: No legitimate reason. I just have been rounded them to 2 decimal places.

brandyscarney: Updated in https://github.com/ionic-team/ionic-framework/pull/27989/commits/9cb4f2739103706f12449dc450c36dc5f8e26db3

brandyscarney: Updated in https://github.com/ionic-team/ionic-framework/pull/27989/commits/9cb4f2739103706f12449dc450c36dc5f8e26db3

core/src/components/item-sliding/test/a11y/item-sliding.e2e.ts-snapshots/item-sliding-scale-md-ltr-Mobile-Firefox-linux.png
brandyscarney, line 1: This diff is intentional, the item label increase causes the item to be taller which also makes the item options taller.

core/src/components/item-sliding/test/a11y/item-sliding.e2e.ts-snapshots/item-sliding-scale-ios-ltr-Mobile-Chrome-linux.png
brandyscarney, line 1: The diffs on all of these `item-sliding-scale` tests is intentional, the item label increase causes the item to be taller which also makes the item options taller.

core/src/components/reorder-group/test/a11y/reorder-group.e2e.ts-snapshots/reorder-group-scale-ios-ltr-Mobile-Chrome-linux.png
brandyscarney, line 1: The diffs on all of these `reorder-group-scale` tests is expected, the reorder group is using an item and now the label also scales with the reorder icons.

brandyscarney:
I have addressed all feedback & made updates. Screenshots have been updated. I added comments explaining why there are diffs in the 2 tests I did not change. This is ready for another review.

core/src/components/item/item.md.vars.scss
```diff
@@ -165,7 +165,7 @@ $item-md-icon-end-slot-margin-end:          null !default;
 $item-md-icon-slot-color:                   rgba($text-color-rgb, 0.54) !default;
 
 /// @prop - Font size of an icon in the start/end slot
-$item-md-icon-slot-font-size:                  24px !default;
+$item-md-icon-slot-font-size:               24px !default;
```
liamdebeasi, line 168: why does this not scale?

core/src/components/item/test/a11y/item.e2e.ts-snapshots/item-buttons-scale-md-ltr-Mobile-Chrome-linux.png
thetaPC, line 1: The `md` versions are showing the large buttons covering the text. I would expect them to behave like iOS, text will truncate and not be hidden behind the button.

core/src/components/item-divider/item-divider.md.vars.scss
```diff
@@ -14,7 +14,7 @@ $item-divider-md-color:              $text-color-step-600 !default;
 $item-divider-md-background:         $background-color !default;
 
 /// @prop - Font size for the divider
-$item-divider-md-font-size:          0.875rem !default;
+$item-divider-md-font-size:          14px !default;
```
thetaPC, line 17: This seems to be reverting your [work](https://github.com/ionic-team/ionic-framework/pull/27947/files#diff-6bd03c0d8ac5b5a7f71d9d20ffe7069d6606a73701153130d282938ad21dd3aaR17) on `item-divider`. Is this intentional?

brandyscarney: It is still using `dynamic-font` I just didn't put it in the variable since the font size in px is reused: https://github.com/ionic-team/ionic-framework/pull/27989/files#diff-40f8af9d7728f2c5784d79c2a2ec084752d277e28b30a31c6ab66d65cc206ce7L17

thetaPC: That makes sense. 🤦‍♀️

core/src/components/item/test/a11y/item.e2e.ts-snapshots/item-buttons-scale-md-ltr-Mobile-Chrome-linux.png
thetaPC: @brandyscarney thoughts?

brandyscarney: Oh sorry I must've missed this! The text is technically truncating there is just not enough room to show more of it. 

| `100%` | `135%` | `150%` |
| ---| ---| ---|
|  |  |  |

What _should_ be happening is the item content should be wrapping to the next line so the button is on a line by itself and the label is fully visible. There was a note on the ticket to make the text wrap but we decided against doing it because it would be considered a breaking change to make the text wrap by default. I do have a ticket for [preventing the label from collapsing when it contains a note](https://ionic-cloud.atlassian.net/browse/FW-4905), which in theory should fix this, but I made a separate ticket to make sure we [wrap the text and also check for buttons in an item](https://ionic-cloud.atlassian.net/browse/FW-5092). Lmk if that is sufficient!

thetaPC: Thank you for the reminder!!

thetaPC (approved):
LGTM

PR #28005, opened 2023-08-16
fix(input, textarea): clearOnEdit does not clear when pressing Tab

### LINKED ISSUE #27746: bug: clearOnEdit should not clear input when tabbing to next input
### Prerequisites

### Ionic Framework Version

v7.x

### Current Behavior

The clearOnEdit option of ion-input is supposed to clear the content of the ion-input when a new character is entered

If filling a form with the keyboard, if you tab into an input which has contents and then you tab into another input, the input clears, which i think is not what's expected?

### Expected Behavior

The input should not be erased if the input "entered" is the tab key..

### Steps to Reproduce

1. Go to https://ionicframework.com/docs/api/input#clear-options
2. Activate the text input which has clearOnEdit activated
3. tab change into the following input
4. Shift-Tab into the previous input, which results in the input with clearOnEdit option activated and it has already text
5. Tab again to change into the next input
6. ... the input clears.. 

### Code Reproduction URL

https://ionicframework.com/docs/api/input#clear-options

### Ionic Info

Latest.. Test based on ionic Docs 

### Additional Information

_No response_

liamdebeasi:
Issue number: resolves #27746

---------

Pressing the Tab key when focused on an input/textarea with `clearOnEdit` clears the text field and then moves focus to the next focusable element.

## What is the new behavior?

- Pressing the Tab key does not clear the text field even when clearOnEdit is enabled.
- Added test coverage
- I also noticed that input did not have an `index.html` file in the basic directory, so I added that.

- [ ] Yes
- [x] No

## Other information

Dev build: `7.3.1-dev.11692202566.13cd16c4`

### DIFF
--- a/core/src/components/input/input.tsx
+++ b/core/src/components/input/input.tsx
@@ -532,7 +532,7 @@ export class Input implements ComponentInterface {
      * Clear the input if the control has not been previously cleared during focus.
      * Do not clear if the user hitting enter to submit a form.
      */
-    if (!this.didInputClearOnEdit && this.hasValue() && ev.key !== 'Enter') {
+    if (!this.didInputClearOnEdit && this.hasValue() && ev.key !== 'Enter' && ev.key !== 'Tab') {
       this.value = '';
       this.emitInputChange(ev);
     }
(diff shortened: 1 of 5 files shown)


brandyscarney (approved):
Looks good! Nice fix and addition of tests. 👍

PR #28006, opened 2023-08-16
feat(range): support dynamic type

liamdebeasi:
Issue number: N/A

---------

Range does not support dynamic type on iOS

## What is the new behavior?

- Range text now scales with Dynamic Type

 A few notes:

## Icon Sizes

The native iOS spec does not have icons scale with the range, so the icons in `ion-range` do not scale either.

## Pin Sizing

The proportions of the pin on MD mode will change as text scales. [Example](https://github.com/ionic-team/ionic-framework/blob/d8767a90a609e1a84376176ccbe3143a2b460939/core/src/components/range/test/a11y/range.e2e.ts-snapshots/range-scale-md-ltr-Mobile-Chrome-linux.png)

I intentionally did not have the tip of the pin scale since we are scaling according to the text size, and the tip of the pin does not have any text in it. I also thought that having the tip grow significantly would look strange since it would increase the overall size of the component without adding a lot of value.

## Text Alignment within Pin

The screenshot diffs here appear to be correct. The text within the pin in `main` is aligned with neither the circle in the pin nor the pin itself. As a result of the changes here, the text is now aligned within the circle (which is the container that has the text).

**Alignment with Circle**

| `main` | branch |
| - | - |
| main-circle | branch-circle |

**Alignment with Pin**

| `main` | branch |
| - | - |
| main-pin | branch-pin |

- [ ] Yes
- [x] No


core/src/components/range/range.md.scss
```diff
@@ -54,7 +59,7 @@
 }
 
 :host(.range-has-pin) {
-  @include padding($range-md-padding-vertical + $range-md-pin-font-size + $range-md-pin-padding-vertical, null, null, null);
+  @include padding($range-md-pin-dimension, null, null, null);
```
liamdebeasi, line 62: The existing padding did not account for the new pin size which is why I changed it here

core/src/components/range/range.md.scss
```diff
@@ -16,19 +16,24 @@
...
+:host(.legacy-range) ::slotted([slot="label"]) {
   font-size: initial;
 }
 
+:host(:not(.legacy-range)) ::slotted(:not(ion-icon)[slot="start"]),
+:host(:not(.legacy-range)) ::slotted(:not(ion-icon)[slot="end"]),
+:host(:not(.legacy-range)) .native-wrapper {
+  font-size: $range-md-pin-font-size;
+}
+
```
liamdebeasi, line 31: This is pretty gnarly, but I basically wanted to apply the smaller font size to the pin and text in the start/end slots (excluding icons). Any ideas for how I can do this better?

core/src/components/range/test/a11y/range.e2e.ts-snapshots/range-scale-ios-ltr-Mobile-Chrome-linux.png
brandyscarney, line 1: I updated the reference screenshots since the checks were failing and the range is kind of hidden by the label now, we might want to update these tests to remove the label or something: https://github.com/ionic-team/ionic-framework/pull/28006/commits/a25fd15c31833ce672900c2527d32ce8913f541f

core/src/components/range/range.md.scss
brandyscarney: You could target the icon after but then we would be always overriding icon size:

```
:host(:not(.legacy-range)) ::slotted(*),
:host(:not(.legacy-range)) .native-wrapper {
  font-size: $range-md-pin-font-size;
}

:host(:not(.legacy-range)) ::slotted(ion-icon[slot]) {
  font-size: 24px;
}
```

On that note - are the icons supposed to respond to dynamic type?

liamdebeasi: I opted not to since native doesn't seem to do it: 
IMG_2579

brandyscarney: Oh good call. Is there a native example with text?

liamdebeasi: IMG_2581

This is the closest I could find

brandyscarney: Thanks for that example! So I did some comparisons and it looks like it should be using the following:

```
font-size: max(14px, 17px); 
```

which is roughly:

```
font-size: dynamic-font-min(0.82, 17px);
```

However, since it is currently using `16px` we probably don't want to update it, so I think it's fine as is.

brandyscarney: As for the original question, I think it's fine unless you want to condense it like:

```
:host(:not(.legacy-range)) ::slotted(:not(ion-icon)[slot]),
:host(:not(.legacy-range)) .native-wrapper {
  font-size: $range-md-pin-font-size;
}
```

brandyscarney (approved):
Looks great and the pin scales nicely!

PR #28009, opened 2023-08-16
feat(angular): add enhanced icon support with addIcons

liamdebeasi:
Issue number: N/A

---------

Developers have no easy way of passing icon data to the `ion-icon` component. The icon component also has no support for switching icons based on mode.

## What is the new behavior?

- Update test apps to use dev build of ionicons with new addIcons feature
- Added tests

- [ ] Yes
- [x] No


packages/angular/test/base/src/app/standalone/icon/icon.component.ts
```diff
@@ -0,0 +1,27 @@
+import { Component } from '@angular/core';
+import { IonIcon } from '@ionic/angular/standalone';
+import { logoIonic, logoIonitron, logoApple, logoAndroid } from 'ionicons/icons';
+import { addIcons } from 'ionicons';
+
+addIcons({ logoIonic, logoIonitron, logoApple, logoAndroid });
```
sean-perkins, line 6: Played with this locally, everything looks/works good 👍 

It seems to work consistently if you have this outside the class or within the constructor. Does not work (too late in the execution cycle, after `connectedCallback` fires) if you try to add icons in the `ngOnInit` lifecycle.

sean-perkins (approved):

packages/angular/test/base/src/app/standalone/icon/icon.component.ts
liamdebeasi: Good catch! We'll want to note this in the docs

liamdebeasi:
Going to merge, but I'll make sure to review the test app once the new version of Ionicons ships.

PR #28021, opened 2023-08-17
feat(checkbox, radio, toggle, range): stacked labels for form controls

### LINKED ISSUE #27229: feat: stacked labels for form controls
### Prerequisites

Controls should be able to work with external labels. I.e. for any control sibling `<label for="control-id">Text</label>` might be attached and pressing on this label should apply focus on the control.

### Describe the Use Case

Right now in v.7.0.1 I can't attach labels to the `ion-checkbox` elements to implement the following layout:
image

with this code:
```
<div class="week-container" formGroupName="weekDays">
  <label
    *ngFor="let day of weekDays; let i = index"
    for="day{{ i }}"
    class="label"
    >{{ day }}</label
  >
  <ion-checkbox
    *ngFor="let control of dayArray.controls; let i = index"
    formControlName="{{ i }}"
    id="day{{ i }}"
  ></ion-checkbox>
</div>
```

_No response_

### Describe Alternatives

_No response_

### Related Code

_No response_

### Additional Information

_No response_

thetaPC:
Issue number: resolves #27229 

---------

Checkbox, radio, toggle, and range do not allow labels to be stacked.

## What is the new behavior?

- Checkbox can stack the label on top of the control and adjust the alignment.
   - New prop `align` was added with `center` as the default[^1]
   - New `label-placement` value was added
   - Snapshots were added
- Radio can stack the label on top of the control and adjust the alignment.
   - New prop `align` was added with `center` as the default[^1]
   - New `label-placement` value was added
   - Snapshots were added
- Toggle can stack the label on top of the control and adjust the alignment.
   - New prop `align` was added with `center` as the default[^1]
   - New `label-placement` value was added
   - Snapshots were added
- Range can stack the label on top of the control and adjust the alignment.
   - New `label-placement` value was added
   - Alignment cannot be changed. The label will be on the left side when LTR and right when RTL.
   - Snapshots were added, padding was added to the container to prevent the knob from being cut off

[^1]: `center` is being used because `align-items="center"` was the original style prior to the implementation. Otherwise, the label will not line up correctly on default. For example, the checkbox for iOS will shift the label to the top left corner when `align` defaulted to `start`.[^2] This shift happens because the input is larger than the label, making the alignment obvious. But this is not correct, the label must be centered vertically as the default to prevent breaking changes to the current visualizations.
[^2]: Screenshot 2023-08-17 at 4 51 36 PM 

- [ ] Yes
- [x] No

## Other information

- Recommendation: Use the file filter during the PR review to filter out images. It'll make it easier to review code.
- Associated [PR](https://github.com/ionic-team/ionic-docs/pull/3097) to update `ion-docs`


core/src/components/checkbox/test/label/checkbox.e2e.ts
```diff
@@ -138,5 +138,31 @@ configs().forEach(({ title, screenshot, config }) => {
         expect(await checkbox.screenshot()).toMatchSnapshot(screenshot(`checkbox-label-fixed-justify-space-between`));
       });
     });
+
+    test.describe('checkbox: stacked placement', () => {
+      test('should render a start alignment with label in the stacked position', async ({ page }) => {
```
sean-perkins, line 143: ~ The test descriptions are a bit hard to decipher/read. 

Thoughts on rewording to:
> should align the label to the start of the container in the stacked position

This would apply to the center align test as well.

core/src/components/radio/test/label-placement/radio.e2e.ts
```diff
@@ -137,5 +137,35 @@ configs().forEach(({ title, screenshot, config }) => {
         expect(await radio.screenshot()).toMatchSnapshot(screenshot(`radio-label-fixed-justify-space-between`));
       });
     });
+
+    test.describe('radio: stacked placement', () => {
+      test('should render a start alignment with label in the stacked position', async ({ page }) => {
```
sean-perkins, line 142: ~ Similar feedback here as checkbox for test descriptions.

core/src/components/range/test/label/range.e2e.ts
```diff
@@ -124,6 +140,22 @@ configs().forEach(({ title, screenshot, config }) => {
 
         expect(await range.screenshot()).toMatchSnapshot(screenshot(`range-items-fixed`));
       });
+      test('should render label in the stacked placement', async ({ page }) => {
```
sean-perkins, line 143: ~ Thoughts on rewording to:

> should render label above the range slider

core/src/components/toggle/toggle.tsx
```diff
@@ -105,7 +106,7 @@ export class Toggle implements ComponentInterface {
   @Prop() legacy?: boolean;
 
   /**
-   * How to pack the label and toggle within a line.
+   * How to pack the label and toggle within a line`.
```
sean-perkins, line 109:
```suggestion
   * How to pack the label and toggle within a line.
```

core/src/components/range/range.scss
```diff
@@ -308,3 +308,24 @@
...
+/**
+ * Label is on top of the range.
+ */
+:host(.range-label-placement-stacked) .range-wrapper {
+  flex-direction: column;
+
+  align-items: normal;
+}
+
+:host(.range-label-placement-stacked) .label-text-wrapper {
```
liamdebeasi, line 324: The pin on a range with a stacked label currently overlaps the text. We add padding to the host `.range-has-pin` is present, but we should probably add that as bottom margin to the label so the pin never obscures the label.

At the very least, you'll need to add bottom margin to the `.label-text-wrapper` element in each mode stylesheet. You can use the existing padding values that we set on the host when the `.range-has-pin` class is present in each mode stylesheet.

We then need to remove the host padding since we are setting it as margin on the label wrapper instead. There are a couple ways you can do this:

1. Set `@include padding(0px)` in the base stylesheet when `.range-label-placement-stacked` is present.
2. Update the existing `.range-has-pin` selectors to exclude `.range-label-placement-stacked` using `:not`.

We'll also want to capture a screenshot for this.

core/src/components/checkbox/checkbox.scss
```diff
@@ -248,6 +256,25 @@ input {
   max-width: 200px;
 }
 
+// Label Placement - Stacked
+// ----------------------------------------------------------------
+
+/**
+ * Label is on top of the checkbox.
+ */
+:host(.checkbox-label-placement-stacked) .checkbox-wrapper {
```
liamdebeasi, line 265: The following applies to checkbox, radio, and toggle (range is not impacted).

When used in an item, there is not enough bottom margin. It looks like we used to account for this on `.label-text-wrapper` which works fine when the label and control are on the same line, but it looks a little smushed when using stacked layout:

image

You'll want to apply margin on the top of `.label-text-wrapper` as well as margin on the bottom of `.native-wrapper`. So something like this:

```scss
:host(.in-item.toggle-label-placement-stacked) .label-text-wrapper {
  @include margin($toggle-item-label-margin-top, null, $form-control-label-margin, null);
}

:host(.in-item.toggle-label-placement-stacked) .native-wrapper {
  @include margin(null, null, $toggle-item-label-margin-bottom, null);
}
```

We'll want screenshot tests for this too.

liamdebeasi (changes requested):
This is looking really good! I found two edge cases we'll want to account for.

thetaPC:
Closing this PR in favor of another [PR](https://github.com/ionic-team/ionic-framework/pull/28075)

PR #28034, opened 2023-08-21
docs(picker): describe how to set the initial value of a picker column

dtarnawsky:
There is no description of how to set the value of a picker.

## What is the new behavior?
Readers can find this description in the documentation.

- [ ] Yes
- [x] No


liamdebeasi (approved):

liamdebeasi:
Thanks!

PR #28043, opened 2023-08-22
feat(searchbar): support dynamic type

liamdebeasi:
Issue number: N/A

---------

- Searchbar does not support Dynamic Type

## What is the new behavior?

- Searchbar supports Dynamic Type
- Added dynamic type verification test
- I also found a bug where the text on the iOS searchbar was always accounting for the clear button, even when the clear button was hidden. I fixed this and added a test.

- [ ] Yes
- [x] No


core/src/components/searchbar/searchbar.md.vars.scss
```diff
@@ -31,7 +31,7 @@ $searchbar-md-cancel-button-font-size:           1.6em !default;
 $searchbar-md-input-search-icon-color:           $text-color-step-400 !default;
 
 /// @prop - Size of the searchbar input search icon
-$searchbar-md-input-search-icon-size:            20px !default;
+$searchbar-md-input-search-icon-size:            dynamic-font(21px) !default;
```
amandaejohnston, line 34: This size was bumped up from 20px to 21px; was that intentional?

amandaejohnston (approved):

PR #28044, opened 2023-08-22
chore(infra): clean build check will log recommended steps

sean-perkins:
Issue number: N/A

---------

When the CI process detects a diff between the output in CI after running `npm run build` versus the checked out branch, it will log the diff and exit. It isn't very clear to developers what they need to do or why this job step exists.

## What is the new behavior?

- Logs a prettier error message to the developer to let them know the steps they can take when a diff is detected in CI

CleanShot 2023-08-22 at 15 05 03

Example run: https://github.com/ionic-team/ionic-framework/actions/runs/5942940488/job/16117074372?pr=28044

- [ ] Yes
- [x] No

### DIFF
--- a/.github/workflows/actions/test-core-clean-build/action.yml
+++ b/.github/workflows/actions/test-core-clean-build/action.yml
@@ -15,9 +15,12 @@ runs:
     - name: Check Diff
       run: |
         git diff --exit-code || {
-          echo -e "\033[1;31mWarning: There is a diff between the build output from main and your branch.\033[0m"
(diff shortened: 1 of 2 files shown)


mapsandapps (approved):
Super helpful, thank you! I've run into this before, and it took me some time to figure out what was wrong. This would have eliminated that. 👏🏻

thetaPC (approved):
LGTM, love that it's easy to read and understand.

brandyscarney (approved):

PR #28045, opened 2023-08-22
feat(typography): support dynamic type

brandyscarney:
Issue number: N/A

---------

The styles for native headings do not support dynamic type.

## What is the new behavior?
- Adds test for scaling the font in `core/src/css/test/a11y`
  - I thought this made more sense than putting it in a component folder like `text` since `core/src/css` is where the styles exist, but maybe we don't need the additional `a11y` folder since we could just keep all related tests in one file
- Updates the `sass` version to `^1.33.0`

- [ ] Yes
- [x] No


core/package.json
```diff
@@ -66,7 +66,7 @@
     "jest-cli": "^27.5.1",
     "prettier": "^2.6.1",
     "rollup": "^2.26.4",
-    "sass": "^1.26.10",
+    "sass": "^1.33.0",
```
brandyscarney, line 69: Prior to upgrading `sass` I was getting this error when building with the `dynamic-font` function:

Screenshot 2023-08-22 at 4 04 52 PM

I found this [Stack Overflow thread](https://stackoverflow.com/questions/67857768/sass-math-div-function-is-undefined-in-angular-app-but-is-defined-in-angular-li) which mentions upgrading to `1.33.0` or higher and that fixed the problem so this change is required for the build to work.

PR #28049, opened 2023-08-23
 fix(menu): emit ionMenuChange when re-mounted

### LINKED ISSUE #28030: bug: missing ion-menu-button on 7.3.0
### Prerequisites

### Ionic Framework Version

v7.x

### Current Behavior

`ion-menu-button` not showing anymore

### Expected Behavior

`ion-menu-button` show display

### Steps to Reproduce

upgrade from `7.2.4` to `7.3.0`

### Code Reproduction URL

_No response_

### Ionic Info

```
Ionic:

   Ionic Framework               : @ionic/angular 7.3.0
   @angular-devkit/build-angular : 15.2.9
   @angular-devkit/schematics    : 15.2.9
   @angular/cli                  : 15.2.9

Capacitor:

   Capacitor CLI      : 5.2.3
   @capacitor/android : 5.2.3
   @capacitor/core    : 5.2.3
   @capacitor/ios     : 5.2.3

Utility:

   cordova-res : 0.15.4
   native-run  : 1.7.2

System:

   NodeJS : v16.14.0
   npm    : 8.3.1
   OS     : macOS Unknown
```

### Additional Information

_No response_

liamdebeasi:
Issue number: resolves #28030

---------

`ion-menu` registers itself with the menu controller when `connectedCallback` fires and then unregisters itself when `disconnectedCallback` fires. When the menu was removed from the DOM, `disconnectedCallback` was not always being fired due to https://github.com/ionic-team/stencil/issues/4070.

`ion-menu-button` checks to see if it should be visible by grabbing the current menu in https://github.com/ionic-team/ionic-framework/blob/314055cf7a66a58e4e88c22e6e755fadd494ed70/core/src/components/menu-button/menu-button.tsx#L74.

Since `disconnectedCallback` was not being fired, `ion-menu-button` would still find the menu even when it was no longer in the DOM. In this case, the menu was not being unregistered due to `disconnectedCallback` not firing.

When the linked Stencil bug was resolved in Stencil 4.0.3, the menu button started to disappear. This is happening because `disconnectedCallback` is now correctly called when the menu is removed from the DOM. Since `disconnectedCallback` is called, the menu is un-registered from the menu controller, and the menu button can no longer find it.

However, this revealed a long-standing bug where re-adding the menu would not fire `ionMenuChange` again. As a result, the menu button remained hidden.

## What is the new behavior?

- The menu now fires `ionMenuChange` on `connectedCallback` as long as `componentDidLoad` has already been run.

- [ ] Yes
- [x] No

## Other information

Dev build: `7.3.2-dev.11692803611.15c1bc87`

### DIFF
--- a/core/src/components/menu-button/test/async/index.html
+++ b/core/src/components/menu-button/test/async/index.html
@@ -29,7 +29,7 @@
       <div class="ion-page" id="main">
         <ion-header>
           <ion-toolbar>
-            <ion-title>Menu - Basic</ion-title>
+            <ion-title>Menu - Async</ion-title>
             <ion-buttons slot="start" id="buttons-container"></ion-buttons>
           </ion-toolbar>
         </ion-header>
--- a/core/src/components/menu-button/test/async/index.html
+++ b/core/src/components/menu-button/test/async/index.html
@@ -29,12 +29,6 @@
       </div>
 
       <div class="ion-page" id="main">
-        <ion-header>
-          <ion-toolbar>
-            <ion-title>Menu - Basic</ion-title>
-            <ion-buttons slot="start" id="buttons-container"></ion-buttons>
-          </ion-toolbar>
-        </ion-header>
         <ion-content class="ion-padding">
           Main Content
           <button onclick="trigger()" id="trigger">Add Menu To DOM</button>
--- a/core/src/components/menu-button/test/async/index.html
+++ b/core/src/components/menu-button/test/async/index.html
@@ -1,0 +1,70 @@
+<!DOCTYPE html>
+<html lang="en" dir="ltr">
+  <head>
+    <meta charset="UTF-8" />
+    <title>Menu - Async</title>
+    <meta
+      name="viewport"
+      content="width=device-width, initial-scale=1.0, minimum-scale=1.0, maximum-scale=1.0, user-scalable=no, viewport-fit=cover"
+    />
+    <link href="../../../../../css/ionic.bundle.css" rel="stylesheet" />
+    <link href="../../../../../scripts/testing/styles.css" rel="stylesheet" />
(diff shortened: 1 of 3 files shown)


core/src/components/menu-button/test/async/index.html
```diff
@@ -0,0 +1,64 @@
...
+      <div id="menu-container">
+        <ion-menu content-id="main">
+          <ion-content class="ion-padding"> Menu Content </ion-content>
+        </ion-menu>
+      </div>
+
+      <div class="ion-page" id="main">
+        <ion-header>
+          <ion-toolbar>
+            <ion-title>Menu - Basic</ion-title>
```
sean-perkins, line 32:
```suggestion
            <ion-title>Menu - Async</ion-title>
```

sean-perkins (approved):

core/src/components/menu-button/test/async/index.html
```diff
@@ -0,0 +1,64 @@
...
+      <div class="ion-page" id="main">
+        <ion-header>
+          <ion-toolbar>
+            <ion-title>Menu - Async</ion-title>
+            <ion-buttons slot="start" id="buttons-container"></ion-buttons>
+          </ion-toolbar>
+        </ion-header>
+        <ion-content class="ion-padding">
+          Main Content
+          <button onclick="trigger()" id="trigger">Add Menu To DOM</button>
```
amandaejohnston, line 38: If I click the button multiple times, the menu is still present in the DOM but the menu button disappears; is that what's supposed to happen? 🤔

liamdebeasi: It looks like connectedCallback fires then disconnectedCallback, so it could be a Stencil bug. I was able to reproduce this behavior even after I removed the fix.

amandaejohnston (approved):

PR #28050, opened 2023-08-23
feat(item-option): support dynamic type

brandyscarney:
Issue number: N/A

---------

The styles for item option do not support dynamic type.

## What is the new behavior?
- Adds test for scaling the font in `item-sliding`
  - I put the test under `item-sliding` because that is where the sliding item utils are, and we have historically kept all of the tests there since an item option should be used with an `ion-item-sliding` element. 

- [ ] Yes
- [x] No

### Native

I used the Reminders app for comparison. The Reminders app uses different maximum sizes for item options in the same sliding item. We are going to follow the "Delete" button. 

| `80%` | `100%` | `135%` | `310%` | 
| ---| ---| ---| ---|
|  |  |  |  |
|  |  |  |  | 

### Ionic Framework

| `80%` | `100%` | `135%` | `310%` | 
| ---| ---| ---| ---|
| ios 80% | ios 100% | ios 135% | ios 310% |
| MD 80% | MD 100% | MD 135% | MD 310% |


core/src/components/item-options/item-options.scss
```diff
@@ -33,7 +33,7 @@ ion-item-options {
 
   height: 100%;
 
-  font-size: 14px;
+  font-size: dynamic-font(14px);
```
brandyscarney, line 36: This shouldn't actually be used anywhere since it is overridden by `item-option` styles, but I didn't want to remove it because I don't know if anyone is using an `ion-item-options` element without `ion-item-option` children and relying on this `font-size` to exist.

core/src/components/item-sliding/test/a11y/item-sliding.e2e.ts
```diff
@@ -0,0 +1,37 @@
...
+        </style>
+
+        <ion-item-sliding id="fontScaling">
+          <ion-item>
+            <ion-label>Item Sliding</ion-label>
+          </ion-item>
+          <ion-item-options>
+            <ion-item-option color="danger">
+              <ion-icon slot="start" name="trash"></ion-icon>
+              Delete
```
brandyscarney, line 26: I combined the icon + text to one test, but can split them if desired.

core/src/components/item-option/item-option.ios.vars.scss
```diff
@@ -4,7 +4,7 @@
 // --------------------------------------------------
 
 /// @prop - Font size of the item option button
-$item-option-button-ios-font-size:          16px !default;
+$item-option-button-ios-font-size:          dynamic-font-clamp(1, 16px, 1.6) !default;
```
brandyscarney, line 7: The iOS native item option uses a `font-size` of `15px` so I used that base to calculate how much the `min` and `max` font size should be. If we update the `font-size` in the future, the min and max ratios do not need to change.

core/src/components/item-options/item-options.scss
brandyscarney: Created FW-5020 to investigate removing this.

mapsandapps (approved):
looks good to me 👍🏻

brandyscarney:
Updated the max size to match the `Delete` button in the Reminders app instead.

Issue #27200, opened 2023-04-13
bug: v7 ion-loading isOpen property sometimes stuck and ignored on fast change


dobripet:
### Prerequisites

- [X] I have read the [Contributing Guidelines](https://github.com/ionic-team/ionic-framework/blob/main/.github/CONTRIBUTING.md#creating-an-issue).
- [X] I agree to follow the [Code of Conduct](https://ionicframework.com/code-of-conduct).
- [X] I have searched for [existing issues](https://github.com/ionic-team/ionic-framework/issues) that already report this problem, without success.

### Ionic Framework Version

v7.x

### Current Behavior

Hello, 
When using Angular ion-loading component I noticed that when there is fast change in isOpen value sometimes the loader gets stuck in wrong state.
See example screen 

![image](https://user-images.githubusercontent.com/11678095/231816462-dea28bec-1467-452e-b59f-9ff898706545.png)

Thank you for looking into that.

### Expected Behavior

Loader is shown when isOpen is true.
Loader is hidden when isOpen is false.

### Steps to Reproduce

Forked docs example 

https://stackblitz.com/edit/angular-lwgtrv?file=src/app/example.component.ts

click few times and it should happen.

### Ionic Info

Forked docs example on Stackblitz, not sure how to do this there...

mapsandapps:
Thank you for the bug report and reproduction!

gianbasagre:
This is an issue for us as well. It would be good to get this fixed as the inline method is the recommended way from the official docs.

jvanderberg:
We are getting this with Ionic React as well. It periodically causes our tests to fail because the popup is modal and blocks access to the page behind, which has actually loaded. It's very intermittent.

ryan-ju:
We're facing the same issue when using `IonLoading`, where parent unmount doesn't clear the loading modal.

Looking at the [source code](https://github.com/ionic-team/ionic-framework/blob/721fc6fd9a21c9c03bb21d290e9b1bbd703e43e3/packages/react/src/components/createInlineOverlayComponent.tsx#L69), `IonLoading` internally uses a class component.  I guess the bug is around the bridging code between functional and class components.

Ferdysd96:
I had the same problem with the `isOpen` attribute, the solution I did to solve this problem is the following I hope it helps someone !!

```
import { Injectable } from '@angular/core';
import { LoadingController, LoadingOptions } from '@ionic/angular';


@Injectable({
    providedIn: 'root',
})
[...]

PR #28056, opened 2023-08-24
fix(overlays): prevent overlays from getting stuck open

mapsandapps:
Issue number: #27200 

---------

## Background info

Previous reading:
* https://github.com/ionic-team/ionic-framework/issues/27200
* https://ionic-cloud.atlassian.net/browse/FW-4374
* https://ionic-cloud.atlassian.net/browse/FW-4053

NOTE: This PR contains a different solution than recommended in the tickets. This draft PR is so we can decide if we're ok with this solution instead of the options proposed previously.

## The bug

A bug occurs when you click twice quickly to open an overlay with a small timeout. In some cases, the overlay will present, dismiss, present, then not dismiss the second time, getting stuck open. The bug occurs in action sheet, alert, loading, picker, and toast, but not in modal or popover. You can reproduce manually this by grabbing the test HTML included in this PR and putting it in a branch that doesn't include a fix.

## A potential solution

All the overlays had a check in `present()` for if it was dismissing while presenting. Two of the overlays - modal & popover - weren't experiencing the bug. Those two components also had a similar check in the `dismiss()` method (for if it was presenting while dismissing). The solution presented in this PR is to add this check in the `dismiss()` method of all overlays.

[Maria's notes](https://ionic-cloud.atlassian.net/browse/FW-4374?focusedCommentId=20881) said that we would need to leave this check in `present()` even if we implement another solution. Based on that logic, I think we would also need this check in `dismiss()` even if we did the more fancy lock controller or one of the other proposed solutions. If that assumption is true, I believe this is the simplest solution.

## Misc

Side note: I was unable to find a way to test this bug that consistently failed or was flaky (while the bug was present). I did make manual tests so you can reproduce that the bug is happening without the code changes and not happening after them. I'm not sure if we need to leave those test snippets in the HTML files. They will be useful for debugging for this PR, but not helpful in screenshots, etc. perhaps I should remove them before merging.

---------

When an overlay has a short timeout, clicking twice to trigger opening it can result in it getting stuck open. This bug is present in all overlays except modal and popover.

## What is the new behavior?

- When an overlay with a short timeout is triggered twice quickly, it will open-close-open-close.
- The behavior is the same for all overlay components

- [ ] Yes
- [x] No

### DIFF
--- a/core/src/components/action-sheet/action-sheet.tsx
+++ b/core/src/components/action-sheet/action-sheet.tsx
@@ -230,6 +230,18 @@ export class ActionSheet implements ComponentInterface, OverlayInterface {
    */
   @Method()
   async dismiss(data?: any, role?: string): Promise<boolean> {
+    /**
+     * When using an inline action sheet
+     * and presenting an action sheet it is possible to
+     * quickly dismiss the action sheet while it is
+     * presenting. We need to await any current
+     * transition to allow the present to finish
+     * before dismissing again.
+     */
+    if (this.currentTransition !== undefined) {
+      await this.currentTransition;
+    }
+
     this.currentTransition = dismiss(this, data, role, 'actionSheetLeave', iosLeaveAnimation, mdLeaveAnimation);
     const dismissed = await this.currentTransition;
 
--- a/core/src/components/action-sheet/test/is-open/index.html
+++ b/core/src/components/action-sheet/test/is-open/index.html
@@ -14,7 +14,7 @@
     <style>
       .grid {
         display: grid;
-        grid-template-columns: repeat(2, 1fr);
+        grid-template-columns: repeat(3, 1fr);
         grid-row-gap: 20px;
         grid-column-gap: 20px;
       }
@@ -52,6 +52,10 @@
           <div class="grid-item">
             <h2>Open, then close after 500ms</h2>
             <ion-button id="timeout" onclick="openActionSheet(500)">Open ActionSheet</ion-button>
+          </div>
+          <div class="grid-item">
+            <h2>Open, then close after 50ms</h2>
+            <ion-button id="immediate" onclick="openActionSheet(50)">Open ActionSheet</ion-button>
           </div>
         </div>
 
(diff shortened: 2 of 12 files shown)


core/src/components/action-sheet/action-sheet.tsx
```diff
@@ -230,6 +230,18 @@ export class ActionSheet implements ComponentInterface, OverlayInterface {
...
   async dismiss(data?: any, role?: string): Promise<boolean> {
+    /**
+     * When using an inline action sheet
+     * and presenting an action sheet it is possible to
+     * quickly dismiss the action sheet while it is
+     * presenting. We need to await any current
+     * transition to allow the present to finish
+     * before dismissing again.
+     */
+    if (this.currentTransition !== undefined) {
```
sean-perkins, line 241: Did we try to abstract this shared behavior across overlays to the [`dismiss`](https://github.com/ionic-team/ionic-framework/blob/main/core/src/utils/overlays.ts#L515) utility?

Since we are passing an instance of the class into the utility, we can access all the shared members we expect across implementations. 

Note: We likely would need to add `currentTransition` as a property to `OverlayInterface`.

Then, it might be possible to do this behavior once, in the utility, and all overlays would have the same consistent behavior. 

e.g.: Something such as:

```ts
if (overlay.currentTransition !== undefined) {
  await overlay.currentTransition;
}
// existing dismiss logic
```

core/src/components/action-sheet/action-sheet.tsx
```diff
@@ -230,6 +230,18 @@ export class ActionSheet implements ComponentInterface, OverlayInterface {
...
+      await this.currentTransition;
```
liamdebeasi, line 242: I'm not sure if I'm testing this correctly, but the fix does not appear to work.

In the `action-sheet/test/trigger` test I added the following to the script. I expect the action sheet to remain open, but it disappear after this code executes.

```js
setTimeout(() => {
  actionSheet.isOpen = true;
  setTimeout(() => {
    actionSheet.isOpen = false;
    setTimeout(() => {
      actionSheet.isOpen = true;
    }, 50);
  }, 50);
}, 2000);
```

https://github.com/ionic-team/ionic-framework/assets/2721089/e6fe9c16-8299-4c05-bc9b-4cc7c249c275

mapsandapps:
As noted in Liam's comment, the simple fix I had proposed here didn't work for more complex reproductions.

Closing in favor of #28069

PR #28065, opened 2023-08-27
Update value-accessor.ts

JustwellSolets:
Optimized getClasses method

Issue number: #

---------

In the given scenario, a variable is utilized to store classes. The process involves employing a basic "for" loop with an indexing variable "i". Within the loop, an "if" condition is employed, and if the condition evaluates to true, the corresponding class is stored. Finally, the processed classes are returned.

## What is the new behavior?
To achieve the desired outcome, JavaScript's native "filter" feature is employed to filter out the relevant classes, followed by using "map" to make modifications according to the intended results. Finally, the processed classes are returned.

-
-
-

- [ ] Yes
- [*] No

### DIFF
--- a/packages/angular/src/directives/control-value-accessors/value-accessor.ts
+++ b/packages/angular/src/directives/control-value-accessors/value-accessor.ts
@@ -128,15 +128,7 @@ export const setIonicClasses = (element: ElementRef): void => {
 };
 
 const getClasses = (element: HTMLElement) => {
-  const classList = element.classList;
-  const classes = [];
-  for (let i = 0; i < classList.length; i++) {
-    const item = classList.item(i);
-    if (item !== null && startsWith(item, 'ng-')) {
-      classes.push(`ion-${item.substring(3)}`);
-    }
-  }
-  return classes;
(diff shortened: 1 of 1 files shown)


sean-perkins:
Hello @JustwellSolets is there a bug or issue you are experiencing? If so, can you please link the appropriate issue?

JustwellSolets:
There is no issue assigned with it. I was just exploring the code and suddenly I found that.

sean-perkins:
I appreciate the contribution, but changing code should be the result of a desired outcome. Otherwise it is difficult to reason why a refactor or change in functionality occurred. 

This approach moves to using [`startsWith`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/startsWith) vs. the [local utility](https://github.com/ionic-team/ionic-framework/blob/756bbcfa3f521a82e733d89ed6196f9fc060f4f5/packages/angular/src/directives/control-value-accessors/value-accessor.ts#L140-L142) and leaves an unused function as a result. Additionally, there is not a measurable performance tradeoff from either approach. As a result, I'm going to close this PR for now.

PR #28068, opened 2023-08-28
feat(textarea): support dynamic type

mapsandapps:
Issue number: N/A

---------

ion-textarea does not support dynamic type

## What is the new behavior?

- ion-textarea does support dynamic type

- [ ] Yes
- [x] No

## Other information

Mostly, dynamic type for textarea was already supported. Only the bottom-content had a static size. All other font sizes were all already relative. 

Note: Textareas do not currently look correct when they are in an item, because #27989 has not yet merged. Once that PR is merged, the font sizes within textareas will be relative to the item sizes, and both will scale appropriately. If preferred, I can base this PR off branch FW-4432 instead.


liamdebeasi (approved):
Great job!

PR #28069, opened 2023-08-28
fix(overlays): prevent overlays from getting stuck open

### LINKED ISSUE #27200: bug: v7 ion-loading isOpen property sometimes stuck and ignored on fast change

mapsandapps:
Issue number: resolves #27200 

---------

A bug occurs when you click twice quickly to open an overlay with a small timeout. In some cases, the overlay will present, dismiss, present, then not dismiss the second time, getting stuck open. You can reproduce manually this by grabbing the test HTML included in this PR and putting it in a branch that doesn't include a fix.

## What is the new behavior?

- When an overlay with a short timeout is triggered twice quickly, it will open-close-open-close.
- The behavior is the same for all overlay components

- [ ] Yes
- [x] No

## Other information

Relevant links:
* https://github.com/ionic-team/ionic-framework/issues/27200
* https://ionic-cloud.atlassian.net/browse/FW-4374
* https://ionic-cloud.atlassian.net/browse/FW-4053

I'm not sure how to write an automated test for this bug due to the short timeout required.

You can manually test the fix in [this Stackblitz](https://stackblitz.com/edit/g1kjci?file=package.json) by changing the Ionic version between 7.3.1 and 7.3.2-dev.11693262117.17edbf6d


core/src/utils/overlays.ts
```diff
@@ -743,6 +743,37 @@ export const createDelegateController = (ref: {
   };
 };
 
+/**
+ * Creates a lock controller.
+ */
+export const createLockController = () => {
```
liamdebeasi, line 749: Can we rename this to `src/utils/lock-controller.ts` and then create a tech debt ticket to update `ion-router-outlet` to use this? (be sure to add a TODO comment inside of the router outlet implementation).

The code is the same, so would be an easy win for us internally.

core/src/utils/overlays.ts
```diff
@@ -743,6 +743,37 @@ export const createDelegateController = (ref: {
   };
 };
 
+/**
+ * Creates a lock controller.
```
liamdebeasi, line 747: Can we add more comments as to what this controller does and why it is needed?

core/src/components/modal/modal.tsx
```diff
@@ -426,19 +427,9 @@ export class Modal implements ComponentInterface, OverlayInterface {
       return;
     }
 
-    const { presentingElement, el } = this;
+    const unlock = await this.lockController.lock();
```
liamdebeasi, line 430: For modal and popover we'll want to acquire the lock before we do the `this.presented` check otherwise the modal will just return early if you try to present, dismiss, then present quickly.

core/src/components/popover/popover.tsx
```diff
@@ -434,17 +435,7 @@ export class Popover implements ComponentInterface, PopoverInterface {
...
-     * and dismissing a popover it is possible to
-     * quickly present the popover while it is
-     * dismissing. We need to await any current
-     * transition to allow the dismiss to finish
-     * before presenting again.
-     */
-    if (this.currentTransition !== undefined) {
-      await this.currentTransition;
-    }
+    const unlock = await this.lockController.lock();
```
liamdebeasi, line 438: For modal and popover we'll want to acquire the lock before we do the `this.presented` check otherwise the modal will just return early if you try to present, dismiss, then present quickly.

core/src/utils/overlays.ts
liamdebeasi: Also you can add "needs_refinement" to the ticket so we take care of it in an upcoming sprint

core/src/components/modal/modal.tsx
```diff
@@ -422,24 +423,14 @@ export class Modal implements ComponentInterface, OverlayInterface {
    */
   @Method()
   async present(): Promise<void> {
+    const unlock = await this.lockController.lock();
+
     if (this.presented) {
       return;
```
liamdebeasi, line 429:
```suggestion
      unlock();
      return;
```

We need to release the lock otherwise the overlay will get deadlocked

core/src/components/modal/modal.tsx
```diff
@@ -650,6 +639,8 @@ export class Modal implements ComponentInterface, OverlayInterface {
       return false;
     }
 
+    const unlock = await this.lockController.lock();
```
liamdebeasi, line 642: Can we add a comment on why this lock is before the `checkCanDismiss` call?

liamdebeasi: We also need to release the lock inside of that `checkCanDismiss` if block

liamdebeasi (changes requested):
I found a couple bugs, but otherwise this looks good!

core/src/components/modal/modal.tsx
mapsandapps: I guess I'm not actually sure why it's before the `checkCanDismiss` call? It seems like we would need the lock to be claimed before any `await`s, for timing reasons. But beyond that, I'm not sure.

liamdebeasi: Pretty much! The `canDismiss` check is async (since devs can pass a promise), so the dismiss transition can _potentially_ run but we won't know until that canDismiss callback resolves, so we claim the lock just in case.

If we claimed the lock _after_ that `checkCanDismiss` call and then opened, closed, and then re-opened the modal then the modal would not re-open.

liamdebeasi: Here's an example you can try:

```html
<script>
  const modal = document.querySelector('ion-modal');
  modal.canDismiss = () => {
    return new Promise((resolve) => {
      setTimeout(() => {
        resolve(true);
      }, 2000);
    })
  }

  setTimeout(() => {
    modal.isOpen = true;
    setTimeout(() => {
      modal.isOpen = false;
      setTimeout(() => {
        modal.isOpen = true;
      }, 50);
    }, 50);
  }, 1000);
</script>
```

| lock is claimed before `checkCanDismiss` | lock is claimed after `checkCanDismiss` | 
| - | - |
| <video src="https://github.com/ionic-team/ionic-framework/assets/2721089/eb623850-b6a4-4ffe-8f78-5fb95a0c1f0c"></video> | <video src="https://github.com/ionic-team/ionic-framework/assets/2721089/0fa2a8dd-d982-4334-ad5b-20a3771c02fa"></video> |

liamdebeasi (approved):
Great work on this!

core/src/utils/lock-controller.ts
```diff
@@ -0,0 +1,35 @@
+/**
+ * Creates a lock controller.
+ *
+ * Claiming a lock means that nothing else can aquire the lock until it is released.
```
liamdebeasi, line 4:
```suggestion
 * Claiming a lock means that nothing else can acquire the lock until it is released.
```

liamdebeasi (commented):
sorry one last thing - found a typo :)

adamduren:
I am still seeing this with:
```
{
  "@ionic/react": "^7.5.7",
  "react": "^18.0.0",
  "react-dom": "^18.0.0",
}
```

Please let me know how I can help.

adamduren:
Here's a video of a test I created using a fresh react project made with `ionic start`. The test shows 10 iterations with a dismiss delay of 20ms in which the bug is reproduced 0% of the time. Changing this to 10ms however shows it can be reproduced 20% of the time. It gets worse of course if there's no delay.

This may seem like an edge case and it could be argued that `IonLoading` is not appropriate here but we have some async tasks that are sometimes very quick <10ms but sometimes they are much longer than which necessitates the use of `IonLoading`.  

https://github.com/ionic-team/ionic-framework/assets/581097/6c10d497-d9cc-4256-8047-d9ba36807f5b

hussam-001:
> I am still seeing this with:
> 
> ```
> {
>   "@ionic/react": "^7.5.7",
>   "react": "^18.0.0",
>   "react-dom": "^18.0.0",
> }
> ```
> 
> Please let me know how I can help.

here is a workaround for this

```tsx
import { IonLoading } from "@ionic/react";
import { useEffect, useRef } from "react";

const LoadingIndicator = ({
  isOpen,
  ...props
}: React.ComponentProps<typeof IonLoading>) => {
  const ref = useRef<HTMLIonLoadingElement>(null);

  useEffect(() => {
    ref.current?.[isOpen ? "present" : "dismiss"]();
  }, [isOpen]);

  if (!isOpen) return null;
  return <IonLoading ref={ref} {...props} />;
};

export default LoadingIndicator;
```

----------

```tsx
<LoadingIndicator isOpen={isLoading} />
```

DunhamGitHub:
Still seing this bug...

```
{
  "@ionic/react": "^8.1.1",
  "react": "^18.3.1",
  "react-dom": "^18.3.1",
}
```

image

PR #28073, opened 2023-08-29
feat(toolbar): support dynamic type

liamdebeasi:
Issue number: N/A

---------

Most of the work to support dynamic type has been handled in individual components such as `ion-title`, `ion-button`, and `ion-back-button`. However, I noticed two issues:

1. Searchbars in toolbars do not scale with dynamic type.
2. Toolbar has code that targets an improper usage of `ion-back-button`. The following code set the height of the `ion-back-button` when used in a toolbar with a searchbar: https://github.com/ionic-team/ionic-framework/blob/01fc9b45116f7ad6ddc56c7fb1535dec798c2b3a/core/src/components/toolbar/toolbar.ios.scss#L51-L53

However, this code should never apply because `ion-back-button` should be slotted inside of `ion-buttons` not `ion-back-button`. Slotting `ion-back-button` directly inside of `ion-toolbar` causes the back button and searchbar to appear on different lines.

## What is the new behavior?

- Updated searchbar to scale in a toolbar
- Added test for new searchbar behavior
-  Removed code targeting improper usage

- [ ] Yes
- [x] No


mapsandapps (approved):
It took me a while to figure out that none of the weird things I'm seeing with toolbars are actually related to the toolbar styles. It looks like this PR has all that's needed for the toolbar itself 👍🏻

PR #28075, opened 2023-08-29
feat(checkbox, radio, toggle, range): stacked labels for form controls

### LINKED ISSUE #27229: feat: stacked labels for form controls

thetaPC:
Issue number: resolves #27229 

---------

Checkbox, radio, toggle, and range do not allow labels to be stacked.

## What is the new behavior?

- Checkbox can stack the label on top of the control and adjust the alignment.
   - New prop `align` was added with `center` as the default[^1]
   - New `label-placement` value was added
   - Snapshots were added
- Radio can stack the label on top of the control and adjust the alignment.
   - New prop `align` was added with `center` as the default[^1]
   - New `label-placement` value was added
   - Snapshots were added
- Toggle can stack the label on top of the control and adjust the alignment.
   - New prop `align` was added with `center` as the default[^1]
   - New `label-placement` value was added
   - Snapshots were added
- Range can stack the label on top of the control and adjust the alignment.
   - New `label-placement` value was added
   - Alignment cannot be changed. The label will be on the left side when LTR and right when RTL.
   - Snapshots were added, padding was added to the container to prevent the knob from being cut off

[^1]: `center` is being used because `align-items="center"` was the original style prior to the implementation. Otherwise, the label will not line up correctly on default. For example, the checkbox for iOS will shift the label to the top left corner when `align` defaulted to `start`.[^2] This shift happens because the input is larger than the label, making the alignment obvious. But this is not correct, the label must be centered vertically as the default to prevent breaking changes to the current visualizations.
[^2]: Screenshot 2023-08-17 at 4 51 36 PM 

- [ ] Yes
- [x] No

## Other information

[Original PR](https://github.com/ionic-team/ionic-framework/pull/28021)

- Recommendation: Use the file filter during the PR review to filter out images. It'll make it easier to review code.
- Associated [PR](https://github.com/ionic-team/ionic-docs/pull/3097) to update `ion-docs`


core/src/components/toggle/toggle.tsx
```diff
@@ -115,6 +116,13 @@ export class Toggle implements ComponentInterface {
    */
   @Prop() justify: 'start' | 'end' | 'space-between' = 'space-between';
 
+  /**
+   * How to pack the label and control along the cross axis.
+   * `"start"`: The label and control will appear at the top of the container.
+   * `"center"`: The label and control will appear at the center of the container.
+   */
+  @Prop() align: 'start' | 'center' = 'center';
```
liamdebeasi, line 124: We're going to need to name this something other than `align` because it collides with the global deprecated `align` attribute: https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes. The `align` attribute is going to cause the text inside of the label container to be centered.

Maybe we can call this `alignment`? Open to other ideas from the team.

liamdebeasi: You'll also notice that long labels are no longer centered within their container. We're currently discussing the best path forward for this on Slack.

core/src/components/toggle/toggle.tsx
```diff
@@ -115,6 +116,13 @@ export class Toggle implements ComponentInterface {
    */
   @Prop() justify: 'start' | 'end' | 'space-between' = 'space-between';
 
+  /**
+   * How to pack the label and control along the cross axis.
```
liamdebeasi, line 120: Can we update this comment (and ditto for other components) that this only applies when the label and control are on separate lines such as when using `labelPlacement="stacked"`?

core/src/components/toggle/toggle.tsx
```diff
@@ -90,8 +90,9 @@ export class Toggle implements ComponentInterface {
    * `"start"`: The label will appear to the left of the toggle in LTR and to the right in RTL.
    * `"end"`: The label will appear to the right of the toggle in LTR and to the left in RTL.
    * `"fixed"`: The label has the same behavior as `"start"` except it also has a fixed width. Long text will be truncated with ellipses ("...").
+   * `"stacked"`: The label will appear above the toggle regardless of the direction. The alignment of the label can be controlled with the `align` property.
```
liamdebeasi, line 93: Don't forget to update this to reference "alignment" instead of "align" (or whatever name we decide on)

liamdebeasi (approved):
The implementation of this is really good! We'll need to choose a different property name than `align` since it conflicts with a global deprecated attribute.

Feel free to add me for another review, but once my comments are addressed I think this is good to go.

core/src/components/toggle/toggle.tsx
thetaPC: `alignment` works for me. And based on the Slack thread, no changes will be needed for centering.

core/src/components/checkbox/checkbox.tsx
```diff
@@ -95,6 +96,13 @@ export class Checkbox implements ComponentInterface {
    */
   @Prop() justify: 'start' | 'end' | 'space-between' = 'space-between';
 
+  /**
+   * How to control the alignment of the checkbox and label on the cross axis.
+   * `"start"`: The label and control will appear at the top of the container.
```
amandaejohnston, line 101: "Top of the container" is a little misleading since it's only true for non-stacked label placements. Maybe something like "start of the cross axis"? Or "left side of the cross axis on LTR, or right side on RTL"? (Ditto for the other components.)

amandaejohnston (approved):
LGTM once my comment is addressed 👍

core/src/components/checkbox/checkbox.scss
```diff
@@ -248,6 +267,24 @@ input {
   max-width: 200px;
 }
 
+// Label Placement - Stacked
+// ----------------------------------------------------------------
+
+  /**
+  * Label is on top of the checkbox.
+  */
+  :host(.checkbox-label-placement-stacked) .checkbox-wrapper {
```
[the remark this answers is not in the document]
thetaPC: Thanks! Lint doesn't seem to catch it.

core/src/components/range/range.scss
```diff
@@ -308,3 +308,24 @@
...
+// Range Label Placement - Stacked
+// ----------------------------------------------------------------
+
+/**
+ * Label is on top of the range.
+ */
+:host(.range-label-placement-stacked) .range-wrapper {
+  flex-direction: column;
+
+  align-items: normal;
```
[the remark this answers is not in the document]
thetaPC: I went with `normal` because it's the default value for `align-items`. I'll switch it over to prevent any changes based on layout mode.

liamdebeasi (approved):

PR #28086, opened 2023-08-30
test(toggle): migrate to toHaveScreenshot

liamdebeasi:
Issue number: N/A

---------

Several tests for this component are still using Playwright's old `toMatchSnapshot` assertion. It's now recommended to use the newer `toHaveScreenshot` assertion. This new assertion reduces the size of each screenshot and brings anti-flake improvements such as disabling animations by default.

We previously migrated most of our codebase to use `toHaveScreenshot`, but it looks like we missed the tests that were written during the development of Ionic 7 in a separate branch off `main`.

## What is the new behavior?

- Migrate component tests to use `toHaveScreenshot`.

Note: There should be no layout changes to any of the screenshots. The only difference between the old and new screenshots should be image and file size.

- [ ] Yes
- [x] No


thetaPC (approved):
LGTM

PR #28091, opened 2023-08-30
chore: add lint rule to block toMatchSnapshot usage

liamdebeasi:
Issue number: N/A

---------

In addition to migrating away from `toMatchSnapshot` as found in PRs such as https://github.com/ionic-team/ionic-framework/pull/28083, I think it would be valuable to have a lint rule to prevent developers from accidentally using this API in the future.

## What is the new behavior?

- Adds a lint rule which errors if `toMatchSnapshot` is being used in `*.e2e.ts` files.

- [ ] Yes
- [x] No

## Other information

~Note: This PR's lint step will continue to fail until the remaining `toHaveScreenshot` migration PRs have been merged. Do note merge this PR until that has been completed.~ All set!


core/custom-rules/no-playwright-to-match-snapshot-assertion.js
```diff
@@ -0,0 +1,29 @@
+module.exports = {
+  meta: {
+    messages: {
+      noPlaywrightToMatchSnapshotAssertion: '"toHaveSnapshot" assertions should be used in favor of "toMatchSnapshot". "toHaveSnapshot" brings file size reductions and anti-flake behaviors such as disabling animations by default.',
```
brandyscarney, line 4:
```suggestion
      noPlaywrightToMatchSnapshotAssertion: '"toHaveScreenshot" assertions should be used in favor of "toMatchSnapshot". "toHaveScreenshot" brings file size reductions and anti-flake behaviors such as disabling animations by default.',
```

brandyscarney (commented):
How do I get this to fail locally? I tried running a test with it and `npm run lint` and neither showed me the error.

liamdebeasi:
@brandyscarney In `src/components/accordion/test/basic/accordion.e2e.ts`:

```diff
configs().forEach(({ config, screenshot, title }) => {
  test.describe(title('accordion: basic'), () => {
    test('should not have visual regressions', async ({ page }) => {
      await page.goto(`/src/components/accordion/test/basic`, config);

      await page.setIonViewport();
-    await expect(page).toHaveScreenshot(screenshot('accordion-basic'));
+    expect(await page.screenshot()).toMatchSnapshot(screenshot('accordion-basic'));
    });
  });
});
```

Then run `npm run lint`.

brandyscarney:
Oh okay I tried changing it from this:

```diff
configs().forEach(({ config, screenshot, title }) => {
  test.describe(title('accordion: basic'), () => {
    test('should not have visual regressions', async ({ page }) => {
      await page.goto(`/src/components/accordion/test/basic`, config);

      await page.setIonViewport();
-    await expect(page).toHaveScreenshot(screenshot('accordion-basic'));
+    await expect(page).toMatchSnapshot(screenshot('accordion-basic'));
    });
  });
});
```

and it didn't fail. I realize now the syntax is wrong but should it fail, or is this okay?

liamdebeasi:
Oh! Did you run `npm install` first to install the new lint rules?

brandyscarney:
Sorry I should've specified that it _did_ fail when I used your code example, but not mine, because mine looks to be invalid syntax.

liamdebeasi:
Ohh I see. Yeah it won't catch that, though the test itself will fail to run since it needs `page.screenshot()` so in that case you won't be able to accidentally use it.

core/playwright.config.ts
```diff
@@ -55,9 +55,6 @@ const config: PlaywrightTestConfig = {
     timeout: 5000,
     toHaveScreenshot: {
       threshold: 0.1
-    },
-    toMatchSnapshot: {
-      threshold: 0.1
```
liamdebeasi, line 60: Since this API isn't used anymore we can safely remove the config

brandyscarney (approved):

PR #28092, opened 2023-08-30
test(datetime): remove duplicate test

liamdebeasi:
Issue number: N/A

---------

That particular test is flaky. When going to try and fix the flakiness, I realized that this behavior is already tested by another test. This test presents the `ion-picker-internal` components which are also tested in https://github.com/ionic-team/ionic-framework/blob/7c2b6aed053c47c64e80867991c57903b717e597/core/src/components/datetime/test/custom/datetime.e2e.ts#L10. These show the same components which have the same APIs, so this test shouldn't even be needed.

## What is the new behavior?

- Removed the duplicate test

- [ ] Yes
- [x] No


thetaPC (approved):
LGTM

PR #28093, opened 2023-08-30
feat(angular): add standalone tabs

amandaejohnston:
Issue number: N/A

---------

Tabs cannot be used as a standalone component.

## What is the new behavior?

- Added tabs as a standalone component.
- Added a quick test. I included the event checking from the lazy test in the HTML as a smoke check, but didn't bring over the Cypress tests for them since I noticed the other standalone tests have been quick, stripped down affairs. I'm assuming I would just be duplicating effort. Let me know if I should bring more tests over from the lazy version, or even get rid of the event logging from the standalone HTML.

- [ ] Yes
- [x] No

### DIFF
--- a/packages/angular/test/base/e2e/src/standalone/tabs.spec.ts
+++ b/packages/angular/test/base/e2e/src/standalone/tabs.spec.ts
@@ -3,9 +3,14 @@ describe('Tabs', () => {
     cy.visit('/standalone/tabs');
   });
 
-  it('should render the active tab', () => {
-    cy.get('#tab-button-tab-one').click();
+  it('should redirect to the default tab', () => {
     cy.get('app-tab-one').should('be.visible');
     cy.contains('Tab 1');
+  });
+
+  it('should render new content when switching tabs', () => {
+    cy.get('#tab-button-tab-two').click();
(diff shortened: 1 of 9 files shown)


packages/angular/test/base/e2e/src/standalone/tabs.spec.ts
```diff
@@ -0,0 +1,11 @@
+describe('Tabs', () => {
+  beforeEach(() => {
+    cy.visit('/standalone/tabs');
+  });
+
+  it('should render the active tab', () => {
+    cy.get('#tab-button-tab-one').click();
```
sean-perkins, line 7: We should check that the tab is not visible prior to clicking the button for better regression testing around tab activation.

sean-perkins: Actually.. I was thinking on this more. We should probably have a default route in Angular to redirect to the first tab by default. Loading an empty tabs component isn't a valid use case we should test against. 

Instead we should default to the first tab and then click the `tab-two` button and validate the tab switches.

sean-perkins (approved):
Great work!

PR #28100, opened 2023-08-31
fix(nav-link): prevent navigation when clicking disabled button

### LINKED ISSUE #26541: feat: Add disa
### Prerequisites

It would be beneficial to have a `disabled`-state on `ion-nav-link`, like the one that `ion-button` has.

### Describe the Use Case

With a `disabled`-state, we could conditionally disable an `ion-nav-link`, that includes an `ion-button`. Currently, this causes some problems:
```
<ion-nav-link [component]="someComponent" [componentProps]="{ someProps }">
  <ion-button [disabled]="condition">
    <div>Continue</div>
  </ion-button>
</ion-nav-link>
```
Even when `ion-button` is disabled, `ion-nav-link` triggers.

_No response_

### Describe Alternatives

```
<ion-nav-link [disabled]="condition" [component]="someComponent" [componentProps]="{ someProps }">
  <ion-button>
    <div>Continue</div>
  </ion-button>
</ion-nav-link>
```

### Related Code

_No response_

### Additional Information

_No response_

sean-perkins:
Issue number: Resolves #26541

---------

When a disabled `ion-button` is clicked within `ion-nav-link`, the navigation stack will navigate. 

## What is the new behavior?

- When a disabled`ion-button` is clicked within `ion-nav-link`, it will not navigate.

- [ ] Yes
- [x] No

### DIFF
--- a/core/src/components/nav-link/nav-link.tsx
+++ b/core/src/components/nav-link/nav-link.tsx
@@ -34,7 +34,14 @@ export class NavLink implements ComponentInterface {
   @Prop() routerAnimation?: AnimationBuilder;
 
   private onClick = () => {
-    return navLink(this.el, this.routerDirection, this.component, this.componentProps, this.routerAnimation);
+    const { el } = this;
+    const button = el.querySelector('button,ion-button,a');
+    if (button?.hasAttribute('disabled')) {
+      // Do not navigate if the button is disabled
+      return;
+    }
+
+    return navLink(el, this.routerDirection, this.component, this.componentProps, this.routerAnimation);
   };
 
   render() {
--- a/core/src/components/nav-link/nav-link.tsx
+++ b/core/src/components/nav-link/nav-link.tsx
@@ -35,8 +35,8 @@ export class NavLink implements ComponentInterface {
 
   private onClick = () => {
     const { el } = this;
-    const button = el.querySelector('button,ion-button,a');
-    if (button?.hasAttribute('disabled')) {
+    const button = el.querySelector<HTMLIonButtonElement>('ion-button');
+    if (button?.disabled) {
       // Do not navigate if the button is disabled
       return;
     }
--- a/core/src/components/nav-link/test/nav-link.spec.tsx
+++ b/core/src/components/nav-link/test/nav-link.spec.tsx
@@ -1,0 +1,32 @@
+import { h } from '@stencil/core';
+import { newSpecPage } from '@stencil/core/testing';
+
+import { Button } from '../../button/button';
+import { Nav } from '../../nav/nav';
+import { NavLink } from '../nav-link';
+
+describe('NavLink', () => {
+  it('should not navigate when button is disabled', async () => {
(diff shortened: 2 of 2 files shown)


core/src/components/nav-link/nav-link.tsx
```diff
@@ -34,7 +34,14 @@ export class NavLink implements ComponentInterface {
   @Prop() routerAnimation?: AnimationBuilder;
 
   private onClick = () => {
-    return navLink(this.el, this.routerDirection, this.component, this.componentProps, this.routerAnimation);
+    const { el } = this;
+    const button = el.querySelector<HTMLIonButtonElement>('ion-button');
```
liamdebeasi, line 38: I don't think this is the correct solution since this bug could happen with any component, not just `ion-button`. For example, the following would still reproduce the issue:

```html
<ion-nav-link>
  <button disabled>Hello</button>
</ion-nav-link>
```

We could potentially look into adding a `disabled` property on `ion-nav-link`, though maybe devs could conditionally set the `component` prop based on some application state.

I also checked Vue Router, and disabling a child button does not prevent `<router-link>` from functioning.

sean-perkins: This example:
```html
<ion-nav-link>
  <button disabled>Hello</button>
</ion-nav-link>
```

Strangely enough does not reproduce the issue. The click event is not propagated to the `ion-nav-link` when clicking a disabled HTML button. This same behavior doesn't occur with a disabled button in the shadow DOM though.

https://github.com/ionic-team/ionic-framework/assets/13732623/61aa3e0c-b8bf-4696-b87f-da4a529e709e

> I also checked Vue Router, and disabling a child button does not prevent <router-link> from functioning.

Would that suggest we shouldn't handle accounting for disabled interactive buttons from `ion-nav-link`?

liamdebeasi: The issue doesn't reproduce because you're clicking where `ion-nav-link` overlaps with `button`. The two don't perfectly overlap, so if click on that area then the issue will reproduce:


https://github.com/ionic-team/ionic-framework/assets/2721089/d966b625-6ed4-4b44-84a7-25155a6a91df

liamdebeasi: In this case, developers could probably conditionally set the `component` property. If they want `ion-nav-link` to do nothing then they can pass `undefined` to `component`

liamdebeasi: We'll want to test this though -- I think keeping the `onClick` handler around will still cause screen readers to read this as "clickable", so we might want to set the `onClick` listener only if `component` is set

sean-perkins:
Will circle back to addressing these changes/exploring alternative approaches after other sprint work. This was part of R&D investigation.

liamdebeasi:
Since this was part of R&D I am going to close this for now. Feel free to re-open when this is ready for another review!

PR #28101, opened 2023-08-31
chore: disable codeflow for pull requests

sean-perkins:
Issue number: N/A

---------

Team isn't using the integration. 

## What is the new behavior?

- Disables the Codeflow integration on pull requests

- [ ] Yes
- [x] No

### DIFF
--- a/.stackblitz/codeflow.json
+++ b/.stackblitz/codeflow.json
@@ -4,7 +4,7 @@
       "enabled": false
     },
     "pullRequests": {
-      "enabled": true
+      "enabled": false
     }
   }
 }


.stackblitz/codeflow.json
```diff
@@ -4,7 +4,7 @@
       "enabled": false
```
liamdebeasi, line 4: If we're going to disable everything, should we remove the file entirely?

liamdebeasi (approved):
One non blocking question

thetaPC (approved):
LGTM

.stackblitz/codeflow.json
sean-perkins: Their documentation mentions disabling the bot via the configuration file or through admin permissions for the organization: https://developer.stackblitz.com/codeflow/integrating-codeflowapp-bot#disabling-codeflowapp-bot

My thoughts were disabling it via the configuration and then follow-up with deleting the config file to make sure it is in fact disabled for the repository. 

I believe Codeflow is closed sourced, so I'm unable to see how it detects the enabled status (if it reads from `main`, stashes to an external service, etc.).

liamdebeasi (approved):

PR #28103, opened 2023-09-01
chore(ci): use node v18 for github actions

sean-perkins:
Issue number: N/A

---------

`action/setup-node@v3` shipped a "breaking" change that requires node v18. This breaks our dev build release process. 

https://github.com/ionic-team/ionic-framework/actions/runs/6044495970/job/16403194005

## What is the new behavior?

- Updates the publish action to use node v18

- [ ] Yes
- [x] No

## Other information

Here is a successful run with these changes: https://github.com/ionic-team/ionic-framework/actions/runs/6044521110

### DIFF
--- a/.github/workflows/actions/publish-npm/action.yml
+++ b/.github/workflows/actions/publish-npm/action.yml
@@ -21,7 +21,7 @@ runs:
   steps:
     - uses: actions/setup-node@v3
       with:
-        node-version: 16.x
+        node-version: 18.x
     # Provenance requires npm 9.5.0+
     - name: Install latest npm
       run: npm install -g npm@latest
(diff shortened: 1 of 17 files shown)


.github/workflows/actions/publish-npm/action.yml
```diff
@@ -21,7 +21,7 @@ runs:
   steps:
     - uses: actions/setup-node@v3
       with:
-        node-version: 16.x
+        node-version: 18.x
```
liamdebeasi, line 24: Should we update everything to node 18 while we're at it?

liamdebeasi (approved):
Might want to consider bumping everything to node 18, but otherwise this looks good.

PR #28109, opened 2023-09-01
fix(react): overlay content is shown with hook 

### LINKED ISSUE #28102: bug: IonModal content doesn't render in some cases
### Prerequisites

### Ionic Framework Version

v7.x

### Current Behavior

When navigating between modals alongside with router changed next modal content is not rendered. 
Additionally `useEffect` unmount callback is not called on when modal is closed.

### Expected Behavior

Modal content should be always rendered correctly. 
`useEffect` unmount callback should be called after modal is dismissed. 

### Steps to Reproduce

It's very hard to reproduce. 
1. Open modal. 
2. Open another modal from the previous one (nested modal)
3. Close both modals, open another modal at the root level and navigate to another router in parallel.
This is a minimal STR that I found. 

In the example repo:
1. On the home tab click "Open root modal"
2. In the root modal click "Open deep modal"
3. Click the red button "Close and go to result" 
4. See a blank modal, but expect to see "modal result" content. Also 'unmount' callback is not called. 
At the same time after clicking on the blue button "Close and go to result" in the root modal the "modal result" content is rendered correctly. 
When clicking on the "CLOSE" button in the "deep" modal then `useEffect` unmount callback is called. 

### Code Reproduction URL

https://github.com/legendar/ion-input-label-issue/

### Ionic Info

$ ionic info

       Require stack:
       - /home/***/.config/yarn/global/node_modules/@ionic/cli/lib/project/index.js
       - /home/***/.config/yarn/global/node_modules/@ionic/cli/lib/index.js
       - /home/***/.config/yarn/global/node_modules/@ionic/cli/index.js
       - /home/***/.config/yarn/global/node_modules/@ionic/cli/bin/ionic

       Require stack:
       - /home/***/.config/yarn/global/node_modules/@ionic/cli/lib/project/index.js
       - /home/***/.config/yarn/global/node_modules/@ionic/cli/lib/index.js
       - /home/***/.config/yarn/global/node_modules/@ionic/cli/index.js
       - /home/***/.config/yarn/global/node_modules/@ionic/cli/bin/ionic

Ionic:

   Ionic CLI       : 7.1.1 (/home/***/.config/yarn/global/node_modules/@ionic/cli)
   Ionic Framework : @ionic/react 7.3.2

Capacitor:

   Capacitor CLI      : 5.2.1
   @capacitor/core    : 5.2.1

Utility:

   native-run  : 1.7.2

System:

   NodeJS : v16.15.0 (/home/***/.nvm/versions/node/v16.15.0/bin/node)
   npm    : 8.5.5
   OS     : Linux 5.10

### Additional Information

This is a video with an issue: https://www.loom.com/share/ab28492d01ca43ab9357ccd0c843c79e

liamdebeasi:
Issue number: resolves #28102

---------

When one modal is added and another modal is removed, the modal that is removed does not account for the newly added modal when updating the overlay context in React. As a result, the inner contents of the newly added modal is not mounted.

We originally tried to fix this in https://github.com/ionic-team/ionic-framework/pull/24553, but the fix was not complete. While storing the latest information in a React ref was correct, the way we updated the ref was done in a way such that data was still stale.

In particular, the `overlaysRef` is updated whenever `IonOverlayManager` is re-rendered. State updates are batched, so updating the state twice in quick succession does not necessarily result in 2 separate renders.

## What is the new behavior?

- We need to make sure the ref is updated synchronously before any render so that `addOverlay` and `removeOverlay` always have access to the latest data.
- Added a test

- [ ] Yes
- [x] No

## Other information

Dev build: `7.3.3-dev.11693592339.18e000af`


packages/react/src/components/IonOverlayManager.tsx
```diff
@@ -51,12 +50,50 @@ export const IonOverlayManager: React.FC<IonOverlayManagerProps> = ({ onAddOverl
...
+    /**
+     * In order for this function to use the latest data
+     * we need to update the ref synchronously.
+     * However, updating a ref does not cause a re-render
+     * which is why we still update the state.
+     *
+     * Note that updating the ref in the body
+     * of IonOverlayManager is not sufficient
+     * because that relies on overlaysRef being
+     * updated as part of React's render loop. states updates
```
amandaejohnston, line 63:
```suggestion
     * updated as part of React's render loop. State updates
```

packages/react/src/components/IonOverlayManager.tsx
```diff
@@ -51,12 +50,50 @@ export const IonOverlayManager: React.FC<IonOverlayManagerProps> = ({ onAddOverl
...
```
amandaejohnston, line 88:
```suggestion
     * updated as part of React's render loop. State updates
```

amandaejohnston (approved):

PR #28110, opened 2023-09-01
fix(modal): swipe to dismiss resets status bar style

liamdebeasi:
Issue number: N/A

---------

When swiping to dismiss the card modal, the status bar style is reset too late. Since we are using 1 animation for the card modal, the dismiss animation is played _then_ the `dismiss` method is called (which resets the status bar style). This means the status bar style is wrong for the duration of the dismiss animation.

This does not impact dragging to close the modal such that the status bar changes mid-gesture or calling the `dismiss` method directly -- only quickly swiping to dismiss.

## What is the new behavior?

-
-
-

- [ ] Yes
- [x] No


thetaPC (approved):
LGTM

PR #28123, opened 2023-09-05
fix(menu): remove app dir from safe area padding

thetaPC:
Issue number: internal

---------

The `--ion-safe-area-left` and `--ion-safe-area-right` variables in `ion-menu` are being set as if they use the app's direction. It's been determined that safe area is not logical and uses the device's direction. The current implementation is adding padding in the wrong sides for `ion-toolbar` and `ion-content` within a `ion-menu`.

Additionally, `ion-menu` does not use the entire screen so the safe area only needs to be applied to the side that is touching the device screen.

## What is the new behavior?

- Set the `--ion-safe-area-left` and `--ion-safe-area-right` variables to the correct values based on the device's direction.
- Padding is only added to the side that is not in the safe area.
- `ion-toolbar` is adding `--ion-safe-area-left` and `--ion-safe-area-right` based on the device's direction.
- `ion-toolbar` can now inherit the correct values from `--ion-safe-area-left` and `--ion-safe-area-right`.
- `ion-content` can now inherit the correct values from `--ion-safe-area-left` and `--ion-safe-area-right`.

- [ ] Yes
- [x] No

## Other information

Dev build: 7.3.4-dev.11694015543.18bc484f


core/src/components/menu/menu.scss
```diff
@@ -59,15 +59,72 @@
...
 :host(.menu-side-start) .menu-inner {
+  /**
+    * Menu does not cover the whole screen so we need to set the safe area for the
+    * side that touches the screen edge only. Since safe area is not logical, it
+    * needs to be applied to the correct side depending on the language direction.
+    * Otherwise, the content will have less space on both sides.
+    *
+    * LTR:
+    * The left side of the menu touches the screen edge. Padding is applied to the
+    * left side of the menu. No changes are needed since the value has been set.
```
amandaejohnston, line 70: This phrasing feels strange to me, and I'm not totally clear on "no changes are needed since the value has been set" means. I think you were going for something like this?
> The left side of the menu touches the screen edge. The safe area has already been set elsewhere [specify where?], so extra padding does not need to be applied here.

Ditto for the corresponding comment on `side-end`.

amandaejohnston (commented):
This looks good, just had a quick comment on... well, the comments. 😆

amandaejohnston (approved):
Thank you!

PR #28125, opened 2023-09-06
Sp/angular standalone value accessors

sean-perkins:
Issue number: N/A

---------

Ionic standalone form controls cannot participate in Angular forms.

## What is the new behavior?

- Ionic form controls can participate in Angular forms by importing the standalone component

- [ ] Yes
- [x] No


packages/angular/test/base/src/app/standalone/value-accessors/value-accessor-test/value-accessor-test.component.ts
```diff
@@ -0,0 +1,25 @@
...
+  selector: 'app-value-accessor-test',
+  templateUrl: 'value-accessor-test.component.html',
+  standalone: true,
+  imports: [
+    ReactiveFormsModule,
+    FormsModule,
+    JsonPipe,
+    KeyValuePipe,
+    /**
+     * NgFor directive is not available until Angular 15.
```
thetaPC, line 15: Let's add a TODO to make the switch at a later time.

packages/angular/src/directives/control-value-accessors/radio-value-accessor.ts
```diff
@@ -19,9 +18,12 @@ export class RadioValueAccessorDirective extends ValueAccessor {
     super(injector, el);
   }
 
-  // TODO(FW-2827): type (HTMLIonRadioElement and HTMLElement are both missing `checked`)
   @HostListener('ionSelect', ['$event.target'])
   _handleIonSelect(el: any): void {
+    /**
+     * The `el` type is any to access the `checked` state property
+     * that is not exposed on the type interface.
```
thetaPC, line 25: Why is it not exposed? Are there plans to expose it?

thetaPC (changes requested):

liamdebeasi (approved):

packages/angular/src/directives/control-value-accessors/radio-value-accessor.ts
sean-perkins: Sorry just seeing this feedback! This decision was made before the current team in the framework wrappers repository. 

If I was to guess, it was `protected` since the element reference was only ever used internal to the framework wrapper class. They likely did not plan for what happens when extending the framework wrapper to append additionally functionality that requires the element reference.

We could absolutely change this, we would just need to do a design document to de-risk the assumptions.

sean-perkins: I deleted my original reply, if you received the alert for it. I should have looked at the referenced file for this comment.

The `checked` property on `ion-radio` is an internal API that we don't want to expose to developers. 

I'm not sure why historically they used a `@State` property for this vs. doing:
```ts
/**
 * @internal 
 */
@Prop() checked = false;
```

We could create a tech debt ticket to investigate this more if you think it would be helpful!

thetaPC (approved):
LGTM

packages/angular/src/directives/control-value-accessors/radio-value-accessor.ts
thetaPC: It could be worth looking into. Maybe start by asking the team in case anyone remembers.

PR #28129, opened 2023-09-06
feat(datetime): support dynamic type

liamdebeasi:
Issue number: N/A

---------

Datetime does not support dynamic type

## What is the new behavior?

- Convert fonts to rem units
- Add tests

There are two special behaviors found on the native iOS datetime worth noting:

1. The days of the week labels change to single characters at 160% scaling. MD always renders the days of the week as single characters so this behavior is not needed there.
2. The calendar day indicator appears as a rounded square instead of a circle so the text does not spill out of the element. MD's indicators are larger such that the text does not spill out of the circle on larger font sizes.

https://github.com/ionic-team/ionic-framework/assets/2721089/314b0f58-4780-4848-8cab-b107bdb15ba1

I implemented both by taking advantage of the [`mod()` function in CSS](https://developer.mozilla.org/en-US/docs/Web/CSS/mod). Safari shipped support for this, Firefox has support for this behind a flag, and Chrome is [working on adding support](https://bugs.chromium.org/p/chromium/issues/detail?id=1407473&q=css%20mod%28%29&can=2).

These features enhance the Dynamic Type experience, but I don't think they are necessary behaviors in order for Dynamic Type to work. As a result, I am treating these are progressive enhancement features: As browsers add support for `mod()` these behaviors will be automatically enabled. We did this with smooth scrolling inside of `ion-datetime`, and it worked well.

This also means that there may be some screenshot diffs as support is added and Playwright gets updated in the future. Example screenshots:

1. Safari (with support for these behaviors): https://github.com/ionic-team/ionic-framework/blob/1b6ae4a50c2d12936025df45cde099de40d3d276/core/src/components/datetime/test/a11y/datetime.e2e.ts-snapshots/datetime-scale-ios-ltr-Mobile-Safari-linux.png
2. Firefox (without support): https://github.com/ionic-team/ionic-framework/blob/1b6ae4a50c2d12936025df45cde099de40d3d276/core/src/components/datetime/test/a11y/datetime.e2e.ts-snapshots/datetime-scale-ios-ltr-Mobile-Firefox-linux.png
3. Chrome (without support): https://github.com/ionic-team/ionic-framework/blob/1b6ae4a50c2d12936025df45cde099de40d3d276/core/src/components/datetime/test/a11y/datetime.e2e.ts-snapshots/datetime-scale-ios-ltr-Mobile-Chrome-linux.png

(Note: I increased the width of the datetime in these screenshots, but without the noted behaviors the UI looks a bit cramped on smaller datetime sizes)

- [ ] Yes
- [x] No


core/src/components/datetime/datetime.ios.scss
```diff
@@ -72,6 +72,118 @@
...
+
+  .calendar-day-highlight {
+    /**
+     * When the computed font size is 24px
+     * we want the calendar day circles to appear
+     * as rounded rectangle so the text does not
+     * spill outside of the circle (the circle
+     * itself does not change size).
+     */
+    $fontSizeBreakpoint: 24px;
```
amandaejohnston, line 128: Would it make sense to reuse `$datetime-dynamic-font-breakpoint` instead? If not, could you rename one or both of the variables to make the difference more clear?

core/src/components/datetime/datetime.ios.scss
```diff
@@ -72,6 +72,118 @@
...
+     *
+     * Note that when the font size is 24px
+     * then the remainder will be 0, so $width
+     * will still be zero and the min value
+     * will apply, setting the width of each
+     * element to 20px.
+     */
+    $width: calc($remainder * 10);
+    width: clamp(20px, #{$width}, 100%);
+    height: $datetime-dynamic-font-breakpoint;
```
amandaejohnston, line 108: I'm having trouble wrapping my head around how this works 😆 The comments talk about the what, but not so much the why:
- Why do we need to scale the remainder?
- Where did the `20px` in the clamp come from?
- Why is the height set to the font breakpoint value?

core/src/components/datetime/datetime.ios.scss
liamdebeasi: Whoops, yes I can do that. I had moved that code out of `datetime.scss` into `datetime.ios.scss` but I forgot that variable even existed. Fixed in https://github.com/ionic-team/ionic-framework/pull/28129/commits/792a59255a6bd5916b08ba4c970f393440d07fb5

core/src/components/datetime/datetime.ios.scss
liamdebeasi: I clarified my comments in https://github.com/ionic-team/ionic-framework/pull/28129/commits/9710cdc016beae6fa5e90ab6bee642245d713776. Lmk if I can improve them

core/src/components/datetime/datetime.ios.scss
```diff
@@ -72,6 +72,120 @@
...
+     * will be 0.
+     */
+    $max-font-size: min(1rem, $datetime-dynamic-font-breakpoint);
+    $remainder: mod($max-font-size, $datetime-dynamic-font-breakpoint);
+
+    /**
+     * From here, we scale the remainder to get
+     * a width value larger than 20px. The purpose of
+     * scaling the remainder is to get a border radius
+     * large enough such that the element resembles a circle.
```
amandaejohnston, line 95: I think this is in the wrong place? This part is calculating the font size of the weekday labels, not the border radius. Maybe I'm misunderstanding?

liamdebeasi: You are correct - that comment is in the wrong place. Fixed in You are correct - that comment is in the wrong place. Fixed in https://github.com/ionic-team/ionic-framework/pull/28129/commits/67b2c9c1fe4c2fd1a298da109cf876dddf49d356

core/src/components/datetime/datetime.ios.scss
```diff
@@ -72,6 +72,116 @@
...
+     * unchanged (i.e. 16 % 24 = 16). If the
+     * font size is 24px then the remainder
+     * will be 0.
+     */
+    $max-font-size: min(1rem, $datetime-dynamic-font-breakpoint);
+    $remainder: mod($max-font-size, $datetime-dynamic-font-breakpoint);
+
+    /**
+     * From here, we scale the remainder to get
+     * a width value larger than 20px.
```
amandaejohnston, line 93: Why do we need to do this? I'm confused about why we can't skip the scaling step and clamp the remainder directly; that would ensure that font sizes < 20px would get a minimum width, and very large font sizes wouldn't bleed out of the container since the width would get clamped at 100%.

I'm especially wondering about font sizes that would lead to a clamped width in-between 20px and 100%. For example, a global font size/remainder of 8px would lead to a `$width` of 80px, which could reasonably fit on wide/landscape displays but seems way too wide for such a small font.

core/src/components/datetime/datetime.ios.scss
```diff
@@ -72,6 +72,116 @@
...
+     * rectangle appearance.
+     * For font sizes smaller than 24px
+     * the border radius will give a circle appearance.
+     */
+    @include border-radius(max(8px, $radius));
+  }
+}
+
+/**
+ * iOS <17 has a bug where the result of mod() is
```
sean-perkins, line 165: I'm not observing the day of the week truncating to a single letter when increasing the font size on iOS 16.6.

Should it? I assumed it should based on the [browser compatibility](https://developer.mozilla.org/en-US/docs/Web/CSS/mod#browser_compatibility). But I also see you called out a bug.

liamdebeasi: No it should not. As per the comment, iOS <17 has a bug with `mod()` values so this feature is limited to iOS 17+ for now.

core/src/components/datetime/datetime.ios.scss
liamdebeasi: The `20px` exists so that when the font size is >=24px the days of the week still appear. Since the `$remainder` at that point will be 0, doing 0 * 10 still gives you 0, so 20px exists to ensure that the first letter of each day of week label shows up.

We need to scale the values for font sizes <24px because otherwise the label width won't be large enough. For example if the font size is 23px, then the `$remainder` will be 23px. However, a font size of 23px looks something like this:

image

core/src/components/datetime/datetime.ios.scss
```diff
@@ -72,6 +72,116 @@
...
+ * used incorrectly if used inside of another CSS function such
+ * as calc(). The first set of @support checks accounts for iOS <17.
+ * We check -webkit-named-image to ensure that non-WebKit browsers are ignored.
+ * The second set of @support checks account for all other browsers that
+ * do not support mod() yet.
+ */
+@supports ((border-radius: mod(1px, 1px)) and (background: -webkit-named-image(apple-pay-logo-black)) and (not (contain-intrinsic-size: none))) or (not (border-radius: mod(1px, 1px))) {
```
amandaejohnston, line 172: My desktop Mac running Safari 16.6 (I just updated it to the latest) trips this query and always shows the full days; should it?

core/src/components/datetime/datetime.ios.scss
amandaejohnston: Okay, after sleeping on this and playing around with it some more, I think I've got a handle on it now. I was missing a few things:
- The days of the week render at 0.75rem, not 1rem, so a global font size of 24px actually renders these labels at 18px, which is small enough to fit in a 20px width
- The 100% max width kicks in a lot sooner than I was imagining due to the days of the week being in a grid layout
- The days of the week have a max font size, so it's okay that all global font sizes > 24px always result in a width of 20px since the max fits within that, without cutting off parts of the first character

Up to you how much of that you want to include in the comments 😄 Most of it can be sussed out through context, I just had to play around with it to put the pieces together.

core/src/components/datetime/datetime.ios.scss
liamdebeasi: Safari <17 should always show the full days due to a bug in WebKit's implementation of `mod` so I disabled this behavior for browsers like Safari 16.6

core/src/components/datetime/datetime.ios.scss
liamdebeasi: Thanks! I'll work on getting this reflected in the comments

amandaejohnston (approved):
Thank you for getting the comments updated 😄

sean-perkins (approved):

PR #28133, opened 2023-09-06
refactor(item-options): remove unnecessary font-size styles

brandyscarney:
Issue number: N/A

---------

Item options defines a font-size that is unnecessary because `ion-item-option` also defines it. 

## What is the new behavior?
Remove the font-size from `ion-item-options`.

- [ ] Yes
- [x] No


sean-perkins (approved):

thetaPC (approved):
LGTM

PR #28135, opened 2023-09-06
chore: sync with main

thetaPC:
Issue number: #

---------

## What is the new behavior?

-
-
-

- [ ] Yes
- [ ] No


packages/angular/common/src/index.ts
```diff
@@ -25,7 +25,7 @@ export { IonPopover } from './overlays/popover';
 export { IonModal } from './overlays/modal';
 
 export { IonRouterOutlet, provideComponentInputBinding } from './directives/navigation/router-outlet';
-export type { StackEvent } from './directives/navigation/stack-utils';
+export type { StackWillChangeEvent } from './directives/navigation/stack-utils';
```
liamdebeasi, line 28: This doesn't seem to be utilized outside of the common package, so we probably won't need this.

liamdebeasi (commented):
One small change, otherwise looks good.

liamdebeasi:
Don't forget we'll want to regular merge this

liamdebeasi (approved):

PR #28142, opened 2023-09-07
fix: add fallback for when not enabled

liamdebeasi:
Issue number: N/A

---------

During testing I realized that some components on iOS will uses the wrong font family when Dynamic Type is not enabled.

## What is the new behavior?

- Added a fallback to use `var(--ion-font-family)`

| without fallback (notice the tab labels) | with fallback |
| - | - |
| IMG_0129 | IMG_0131 |

- [ ] Yes
- [x] No


core/src/css/typography.scss
```diff
@@ -38,6 +38,10 @@ html {
  */
 @supports (-webkit-touch-callout: none) {
   html {
+    /**
+     * Fallback if Dynamic Type is not enabled.
+     */
+    font: 16px var(--ion-font-family);
```
liamdebeasi, line 44: I could also change the `font-family` declaration above to use `font: 16px var(--ion-font-family)` instead. Any preference?

brandyscarney (approved):

sean-perkins (approved):

PR #28143, opened 2023-09-07
fix(header): do not apply safe area padding to collapsible large title toolbar

liamdebeasi:
Issue number: N/A

---------

During testing I discovered that the following removal caused safe area to be applied to the first toolbar in a header that has the collapsible large title: https://github.com/ionic-team/ionic-framework/commit/a23c1222b958b58e51591ceb416d601cd5f9838b#diff-cfac3c4c4ee6be17fd1c5d458d277c31bdb6d95b3246152acff5b5fa78010ae3L54

Since the large title is in the content, no safe area padding should be applied.

## What is the new behavior?

- Explicitly apply `padding-top: 0` to the first toolbar in a header that has the collapsible large title
- Updated the test so we can catch this in the future

| before | after |
| - | - |
| IMG_0132 | IMG_0133 |

- [ ] Yes
- [x] No


core/src/components/header/test/condense/header.e2e.ts-snapshots/header-condense-large-title-collapsed-diff-ios-ltr-Mobile-Chrome-linux.png
liamdebeasi, line 1: This is correct since this is screenshotting the default header, not the header that has the collapsible large title

mapsandapps (approved):

PR #28144, opened 2023-09-07
fix(searchbar): searchbar in collapsible header has correct height

liamdebeasi:
Issue number: N/A

---------

When testing Dynamic Type I discovered that the searchbar has the wrong height when placed inside of a toolbar inside of a collapsible header.

## What is the new behavior?

- Removed the additional CSS. This is no longer needed because `ion-searchbar` has `min-height: 52px`

| `FW-4146` | branch |
| - | - |
| IMG_0148 | IMG_0147 |

- [ ] Yes
- [x] No

### DIFF
--- a/core/src/components/header/header.ios.scss
+++ b/core/src/components/header/header.ios.scss
@@ -69,8 +69,6 @@
 }
 
 .header-collapse-condense ion-toolbar ion-searchbar {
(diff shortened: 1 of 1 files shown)


thetaPC (approved):
LGTM

PR #28146, opened 2023-09-07
fix(item): elements should wrap and label should not collapse when text size increases

brandyscarney:
This will resolve the following internal tickets: 
- [FW-4905](https://ionic-cloud.atlassian.net/browse/FW-4905): bug: label should not collapse in an item or item divider with a note
- [FW-5092](https://ionic-cloud.atlassian.net/browse/FW-5092): bug: label should wrap instead of truncating inside of an item

---------

There are a few problems with how labels inside of items behave. Labels inside of items do not wrap their text, instead they truncate with an ellipsis. The label will continue to show an ellipsis while shrinking if there is wide content next to it, until eventually it completely disappears. This is especially problematic for dynamic type, when the font size can be increased up to `310%` on `ios`. Below are some screenshots of how Ionic behaves when increasing the font size with an `ion-label` and `ion-note` in an `ion-item`. You'll notice that the `ion-label` (`Wi-Fi`) completely disappears at `310%`.

| `100%` | `145%` | `310%` |
| ---| ---| ---|
| ios ionic 100% | ios ionic 145% | ios ionic 310% |

Here is how the same item looks on native iOS at `310%`:

ios native 310%

## What is the new behavior?
A lot of changes were made in order to support wrapping components inside of an item. Since a lot of the code in `ion-label` and `ion-item` is used throughout multiple components, several changes were wrapped into this PR.
 
1) [Item scss](https://github.com/ionic-team/ionic-framework/pull/28146/files#diff-c74ddc9584918db0c15a8ba1db280abfe14e7f7632b84e1dfbf74ce86315c6c3) was modified to allow wrapping of its containers. Comments have been added to explain the new additions.
2) [Label scss](https://github.com/ionic-team/ionic-framework/pull/28146/files#diff-4a1156704dbf45b0dad273b6909b190ca45e4380aa7378ba88d0dd7d48d7d473) has been updated to allow wrapping (removing the `white-space: nowrap`) and allow it to grow.
3) A `legacy` class was added to item (`.item-legacy`) by all of the components that have `legacy` as a property. This class was targeted in the CSS to undo the wrapping changes for legacy items, otherwise it breaks a lot of tests: https://github.com/ionic-team/ionic-framework/pull/28146/commits/6ec8bf419c12612145a1e94569ec6336a7704e54
  a) Tests with legacy inputs were updated to include the `legacy` attribute. Without it **the UI will break for several examples**: https://github.com/ionic-team/ionic-framework/pull/28146/commits/a8a34ce18bb312d75633955de248c9132025e9dc
4) The `item/test/text` test was removed. It was almost exactly a duplicate of `item/test/basic`: https://github.com/ionic-team/ionic-framework/pull/28146/commits/eb79794fd7b0f0bbedf5ee9779ade4f01385fd29
5) The Item inside of Datetime was styled to set the `width` to `auto`, undoing where I set the width to `min-content` elsewhere, in order to prevent it from wrapping all of the time. I also added a comment linking to a JIRA ticket to remove the Item because of the issues it causes by being used: https://github.com/ionic-team/ionic-framework/pull/28146/commits/88106d847ab0a305788ee2328e496ef90ad1e541
6) Added some CSS to allow the select to shrink when inside of an item, otherwise it always wraps to the next line even when it can shrink and fit: https://github.com/ionic-team/ionic-framework/pull/28146/commits/6804aea1047d47d882e592f22cd524e3fcd10abf
7) Adds a test to `a11y` for Note to ensure that the Label next to a Note doesn't disappear: https://github.com/ionic-team/ionic-framework/pull/28146/commits/1f23ffc3568d83612c2b35e26e4d567f3bd6d1a2
8) Some tests were updated to remove the `ion-text-wrap` and/or add `ion-text-nowrap` to prevent screenshot diffs. [For example](https://github.com/ionic-team/ionic-framework/pull/28146/files#diff-328e125a8637f935958e921b0922050a5035999a992ab14278541e3faf3164c9)
9) Passes the `skeleton-text` class up to item (as `item-skeleton-text`) in order to add `overflow: hidden` to the label for skeleton text wrappers: https://github.com/ionic-team/ionic-framework/pull/28146/commits/077fb08f139920409c001e688e9c214ffe72cd52 - this also adds `overflow: hidden` for legacy labels to keep the ellipsis but the modern items do not need it 

- [ ] Yes
- [x] No

## Other information
Screenshots for this branch are separated into the following PR: https://github.com/ionic-team/ionic-framework/pull/28278

[FW-4905]: https://ionic-cloud.atlassian.net/browse/FW-4905?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[FW-5092]: https://ionic-cloud.atlassian.net/browse/FW-5092?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[FW-5138]: https://ionic-cloud.atlassian.net/browse/FW-5138?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


sean-perkins (commented):
The implementation changes look great. 

Should we have any screenshot test coverage for text wrapping? I feel like your comments in the code callout potential areas we would not want regressions and could be accounted for with a screenshot.

core/src/components/label/label.scss
```diff
@@ -51,6 +64,15 @@
 }
 
 
+// TODO(FW-3547): this can be removed if we
+// refactor datetime to not use an ion-item
+:host-context(.item-in-datetime) {
```
liamdebeasi, line 69: Would it make sense to keep this all in the datetime CSS? IIRC `:host-context` isn't supported in Safari/Firefox, so this fix won't work there.

Doing the following in the datetime CSS could work:

```css
.item-in-datetime ion-label {
  width: auto;
}
```

core/src/components/label/label.scss
```diff
@@ -19,18 +19,31 @@
...
+// TODO(FW-5289): remove
+:host-context(.item-legacy) {
+  white-space: nowrap;
+}
+
+// TODO(FW-5289): move to :host-context(.item)
+// Shouldn't need :not(.item-input) as this was
+// only needed because of the specificity with
+// :not(.item-legacy)
+:host-context(.item:not(.item-input):not(.item-legacy)) {
```
liamdebeasi, line 37: How is this going to work in Safari/Firefox? They don't support `:host-context` IIRC

liamdebeasi (commented):
A few questions on the implementation, but I didn't see any glaring issues in the code/screenshots

core/src/components/label/label.scss
brandyscarney: So I went and refreshed my memory on this because we use `:host-context` in a ton of places and there's no way we would've implemented that without it working in Safari/Firefox for those components. For example, it is used in the following components:

- [buttons](https://github.com/ionic-team/ionic-framework/blob/d669fbefd29788c5fb8fd8650279401ea311a206/core/src/components/buttons/buttons.ios.scss#L32)
- [input](https://github.com/ionic-team/ionic-framework/blob/d669fbefd29788c5fb8fd8650279401ea311a206/core/src/components/input/input.scss#L86)
- [label](https://github.com/ionic-team/ionic-framework/blob/d669fbefd29788c5fb8fd8650279401ea311a206/core/src/components/label/label.scss#L7)
- [searchbar](https://github.com/ionic-team/ionic-framework/blob/d669fbefd29788c5fb8fd8650279401ea311a206/core/src/components/searchbar/searchbar.ios.scss#L168)
- [select](https://github.com/ionic-team/ionic-framework/blob/d669fbefd29788c5fb8fd8650279401ea311a206/core/src/components/select/select.md.scss#L120)
- [textarea](https://github.com/ionic-team/ionic-framework/blob/d669fbefd29788c5fb8fd8650279401ea311a206/core/src/components/textarea/textarea.scss#L114)

In some of these cases (buttons, searchbar) it is specifically used in the `ios` mode. 

All of these are scoped components, except for select. I created a new Stencil project and made two components. One is scoped, and the other is shadow. Both have the following CSS:

```css
:host {
  color: purple;
}

:host-context(.item) {
  color: red;
}
```

When I ran the project in each browser these were the results:

host-context

So it seems that `:host-context` only works in all browsers when the component is scoped, which I think has always been the case and is the reason we use it so many places. 

In this particular case though we have full control over the item and the label so we can target it directly. I have updated the code with your recommendation: https://github.com/ionic-team/ionic-framework/pull/28146/files#diff-83857955657186aba3b158e5561a577b94c5b9b8e74aff61fc97c78f4b77343c

core/src/components/label/label.scss
brandyscarney: See my comment here: https://github.com/ionic-team/ionic-framework/pull/28146#discussion_r1357307512

core/src/components/label/label.scss
liamdebeasi: Thank you!

core/src/components/skeleton-text/skeleton-text.tsx
```diff
@@ -18,6 +19,24 @@ export class SkeletonText implements ComponentInterface {
    */
   @Prop() animated = false;
 
+  /**
+   * Emitted when the styles change.
+   * @internal
+   */
```
liamdebeasi, line 25: Why is this needed now?

brandyscarney: An `ion-label` containing `ion-skeleton-text` needs to have `overflow: hidden`, otherwise it collapses the margins between the elements:

| `overflow: visible` | `overflow: hidden` |
| ---| ---|
| overflow: visible | overflow: hidden |

When I added the `overflow: hidden` to all labels in an item, it resulted in this weird wrapping behavior:

| | |
| ---| ---|
| item overflow hidden | checkbox overflow hidden |

I attempted to solve this for `skeleton-text` by adding `display: inline-block` when it is inside of an item and label: https://github.com/ionic-team/ionic-framework/pull/28289/commits/6fdccc4cdf1f157907651b7e199935d0a2228d61 https://github.com/ionic-team/ionic-framework/pull/28289/commits/a00f4f5318b1a7c3a1b17baa7568cb05b32f5198

But that resulted in the thumbnail and avatar alignment being off. Ultimately I decided that passing a class up to item to let it know it contained a `skeleton-text` and style the label that way was the best solution. In addition I kept `overflow: hidden` for legacy items because otherwise the text doesn't ellipsis: 

| `overflow: visible` | `overflow: hidden` |
| ---| ---|
| overflow: visible | overflow: hidden |

I tried to add comments explaining why specific CSS was added, but lmk if you think there are places I can further clarify!

liamdebeasi: Thanks! Can we add a comment in the JSX file too?

liamdebeasi: Thanks! Can we add a comment in the JSX file too?

liamdebeasi (approved):
One comment, but good to go once that's addressed.

liamdebeasi:
I've updated the [release notes](https://github.com/ionic-team/ionic-framework/releases/tag/v7.6.0), but I recommend allowing text to wrap by default. The previous behavior was broken from an accessibility perspective, so forcing text to not wrap can degrade the experience for many users. We don't consider this a breaking change because the previous experience was broken for many users.

tgangso:
> I've updated the [release notes](https://github.com/ionic-team/ionic-framework/releases/tag/v7.6.0), but I recommend allowing text to wrap by default. The previous behavior was broken from an accessibility perspective, so forcing text to not wrap can degrade the experience for many users. We don't consider this a breaking change because the previous experience was broken for many users.

Specifically in places where I also had an ion-input with the label, a label that was previously displaying fine, was now splitting into multiple lines even when there is plenty of room for the label.

SaroGFX:
I built a lot of custom CSS upon the ion-item for our specific needs, and this minor update was definitely breaking a lot for us 😅 Maybe I shouldn't have used the ion-items in these custom UI's

liamdebeasi:
You can set `.ion-text-nowrap` on elements you do not want to wrap: https://ionicframework.com/docs/api/item#basic-usage

SaroGFX:
> You can set `.ion-text-nowrap` on elements you do not want to wrap: https://ionicframework.com/docs/api/item#basic-usage

This utility does nothing to undo the changes to flex-wrap, flex-grow and shrink properties. It looks like I can't overwrite them either, because these properties look like they are not exposed. Is there any other workaround for me except rebuilding everything without using ion-item or wrapping it in a container?

tgangso:
I also had some elements where we used ion-item as a base container for custom elements that I had to rewrite. Still do not agree this is not a breaking change.

liamdebeasi:
Try setting `width: 0` on the `ion-label` inside of the item. One of the primary changes was updating this element to `flex: 1 0 auto`: https://github.com/ionic-team/ionic-framework/blob/516b84475e5d78060f35fa2c4821efc712536353/core/src/components/item/item.scss#L451

Previously the flex-basis was `0` which defined the initial minimum size of the container. From there, the flex-grow value was `1` which meant the container could grow beyond 0.

The `flex-basis` was changed to `auto` which means the initial container size is now based on the `width` property, so setting `width` should allow you to retain the old flex-basis behavior.

Of course, if you get down to smaller widths such as 100px, the elements may still wrap/get truncated. However, I'd argue that not wrapping content at that small of a width makes the content hard to understand since most text will be truncated.

Here's a demo: https://codepen.io/liamdebeasi/pen/mdvZdjg

liamdebeasi:
Hi everyone, 

The team is having discussions into revising part of the text wrapping behavior. In particular, we are considering not having the end/start slot content wrap to the next line. In other words, only the text in the default slot would wrap within its own container.

We are having this discussion based on some new information we've found related to how iOS handles wrapping. We're tracking this in https://github.com/ionic-team/ionic-framework/issues/28769.

kyleabens:
Is there anyway we can set the item-legacy class globally on all instances of ion-item and ion-item-divider without having to apply it to each one individually?  Maybe add a property to IonicConfig for this so we can say:

```
IonicModule.forRoot({
      mode: 'md',
      itemLegacy: true
})
```

This recent change of wrapping ion-label by default does in a way break the UI of my app.

PR #28161, opened 2023-09-13
fix(range): add correct margin in item

liamdebeasi:
Issue number: N/A

---------

As part of https://github.com/ionic-team/ionic-framework/commit/e6c7bb60e7e61c965f45e2bf3e3bd16f5125ad56, we added margin to the stacked label when in an item so the label doesn't run up against the divider line for an item above it. However, we did not add this same margin for range.

You can see this in the screenshot comparisons:

| Checkbox | Range |
| - | - |
| image | image |

Notice how the checkbox in an item screenshot has top/bottom margin, but the range in an item does not.

## What is the new behavior?

- Adds margin to the top of the label and the bottom of the native wrapper when a range is in an item.

- [ ] Yes
- [x] No


core/src/components/range/range.vars.scss
```diff
@@ -0,0 +1,5 @@
+/// @prop - Top margin of range's label when in an item
+$range-item-label-margin-top: 10px !default;
+
+/// @prop - Bottom margin of range's label when in an item
+$range-item-label-margin-bottom: 10px !default;
```
thetaPC, line 5: I know that this is the default but it does seems a bit too much when comparing it to the other labels.
Screenshot 2023-09-13 at 9 46 02 AM

Should we aim to keep it visually similar or stick with the default value?

liamdebeasi: Good point. I'll try removing the bottom margin and see how that looks.

liamdebeasi: Ok how does https://github.com/ionic-team/ionic-framework/pull/28161/commits/db20bd1b8e705144b6d4db0e8c0c98ad4a328d26 look?

mapsandapps (approved):
Looks good to me after the bottom margin adjustment 👍🏻

core/src/components/range/range.vars.scss
thetaPC: That looks better. I can't tell if my eyes are playing tricks but it seems that the stacked spacing is less compared to a non-stacked label.
Screenshot 2023-09-13 at 10 35 49 AM

Using `3px` seems closer. I'm not a stickler on this though.

liamdebeasi: When using a non-stacked label the label-text-wrapper element adds 10px of bottom margin which is why there's a difference. The same difference exists with the other form controls too, so I don't think that's something that should be handled in this PR.

thetaPC (approved):
LGTM

PR #28163, opened 2023-09-13
fix(many): add correct scale to stacked labels

liamdebeasi:
Issue number: N/A

---------

The checkbox, radio, range, and toggle stacked labels do not scale which makes them look a bit out of place when used in a list with input/textarea/select stacked labels.

Notice that the stacked Radio label is much larger than the Input label when using the modern syntax. The legacy syntax is consistent where both the Input and Radio stacked labels are scaled.

image

## What is the new behavior?

- Added logic to checkbox, radio, range, and toggle so that the labels scale to 0.75
-  Added the appropriate transform origin for each alignment value
- Refactored input/textarea/select to use the new global sass variable
- Added max-width logic to the stacked labels so long labels don't run off the page

- [ ] Yes
- [x] No


thetaPC (approved):
LGTM

PR #28167, opened 2023-09-13
 fix(vue): respect keepContentsMounted if passed as attribute

### LINKED ISSUE #28165: bug: Ionic Vue - `keep-contents-mounted` issue with `vue/prefer-true-attribute-shorthand` listing rule
### Prerequisites

### Ionic Framework Version

v7.x

### Current Behavior

The vue linter is reporting an issue on the `<IonModal>` component when it's used with the `keep-contents-mounted` props. The linter wants to replace `:keep-contents-mounted="true"` with `keep-contents-mounted`, because it is preferred as a shortand. 
But when the `:keep-contents-mounted="true"` is replaced by `keep-contents-mounted`, the IonModal wrapping the IonDateTime doesn't show up, and so does the `<IonDatetimeButton>`. The error is as follow: `[Ionic Error]: No ion-datetime instance found for ID 'durationIonDatetime'. `.

### Expected Behavior

The modal and the input should show up whether the props has the syntax: `:keep-contents-mounted="true"` but also with the shortand syntax : `keep-contents-mounted`.

### Steps to Reproduce

1. Setup a <IonDatetimeButton slot="end" datetime="testing" /> inside an Ionic App
2. Add the following in the template:
``` vue
<IonModal v-if="canShowDuration" keep-contents-mounted>
  <IonDatetime
    id="testing"
    presentation="time"
    :value="duration"
    @ion-change="onInput"
  />
</IonModal>
```
3. It shouldn't work, but with the props as `:keep-contents-mounted="true"`, it should display the input and show the modal on input.

### Code Reproduction URL

_No response_

### Ionic Info

Ionic:

   Ionic CLI       : 7.1.1 (/Users/guillaumetorreilles/.asdf/installs/nodejs/20.6.1/lib/node_modules/@ionic/cli)
   Ionic Framework : @ionic/vue 7.3.3

Capacitor:

   Capacitor CLI      : 5.3.0
   @capacitor/android : 5.3.0
   @capacitor/core    : 5.3.0
   @capacitor/ios     : 5.3.0

Utility:

   native-run  : 1.7.2

System:

   NodeJS : v20.6.1 (/Users/guillaumetorreilles/.asdf/installs/nodejs/20.6.1/bin/node)
   npm    : 9.8.1
   OS     : macOS Unknown

### Additional Information

_No response_

liamdebeasi:
Issue number: resolves #28165

---------

The overlay implementation in Vue only checks for truthy `keepContentsMounted` values. When setting this prop as an attribute, the value of it is `''` which is falsy. As a result, content does not get mounted.

One of Vue's ESLint rules states that this should be supported: https://eslint.vuejs.org/rules/prefer-true-attribute-shorthand.html

Part of the issue may also be that Vue does not know the type of this property and so it assume "any":

> The shorthand form is not always equivalent! If a prop accepts multiple types, but Boolean is not the first one, a shorthand prop won't pass true.

## What is the new behavior?

- The overlay wrapper now checks for `''` values. If `keepContentsMounted === ''` then the inner contents will be mounted because this means the prop is being set using the attribute shorthand.

- [ ] Yes
- [x] No

## Other information

Dev build: `7.3.5-dev.11694621267.1e5f63c2`

### DIFF
--- a/packages/vue/src/vue-component-lib/overlays.ts
+++ b/packages/vue/src/vue-component-lib/overlays.ts
@@ -177,7 +177,14 @@ export const defineOverlayContainer = <Props extends object>(name: string, defin
         return h(
           name,
           { ...restOfProps, ref: elementRef },
-          (isOpen.value || restOfProps.keepContentsMounted) ? renderChildren() : undefined
+
+          /**
+           * When binding keepContentsMounted as an attribute
+           * i.e. <ion-modal keep-contents-mounted></ion-modal>
+           * the value of the prop will be the empty string which is
+           * why we still call renderChildren() in that case.
+           */
+          (isOpen.value || restOfProps.keepContentsMounted || restOfProps.keepContentsMounted === '') ? renderChildren() : undefined
         )
       }
     });
--- a/packages/vue/test/base/src/views/OverlaysKeepContentsMounted.vue
+++ b/packages/vue/test/base/src/views/OverlaysKeepContentsMounted.vue
@@ -16,6 +16,8 @@
 
       <ion-button id="open-auto-mount-modal-attribute">Open Auto Mount Modal (Attribute)</ion-button>
 
+      <ion-button id="open-auto-mount-modal-attribute-false">Open Auto Mount Modal (Attribute False)</ion-button>
+
       <br /><br />
 
       <ion-modal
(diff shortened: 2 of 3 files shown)


thetaPC (approved):
LGTM

PR #28177, opened 2023-09-14
test(menu): safe area and proper var reset

thetaPC:
Issue number: N/A

---------

- There are no tests for menu when safe area is applied.
- The safe area variables on menu weren't being reset properly to allow easy local customization.

Currently, some of the variables are being set to `env()`. This is the same structure that is being used in core. However, this doesn't prevents users from mocking the safe area when using `--ion-safe-area-left: 50px` on `html`. It makes it hard to create tests to validate that padding is being applied to the safe area.

## What is the new behavior?

- Tests have been added.
- The safe area variables on menu are now being reset to use the values from `:root`.

The variables are being `unset` in order for the variables to [default to parent styles](https://stackoverflow.com/a/69491310/5374225). Since core styles has already declared the variables, then developers can use `--ion-safe-area-left: 50px` on `html`.

- [ ] Yes
- [x] No

## Other information

N/A


core/src/components/menu/menu.scss
```diff
@@ -83,12 +83,12 @@
       * Otherwise, the content will have less space on both sides.
       *
       * RTL:
-      * The right side of the menu touches the screen edge. The safe area padding has
-      * already been set in the core styles, so there's no need to set it again.
+      * The right side of the menu touches the screen edge.
+      * The right side needs to revert back to the default value that was set in the core styles, so we unset it here.
```
amandaejohnston, line 87: Can you add details to this comment explaining why we need to do this? Ditto for the other comment below.

core/src/components/menu/test/safe-area/menu.e2e.ts-snapshots/menu-end-safe-area-left-notch-md-rtl-Mobile-Chrome-linux.png
amandaejohnston, line 1: This doesn't look right; the padding is on the wrong side of the menu and isn't applying to the header 🤔

core/src/components/menu/test/safe-area/menu.e2e.ts
```diff
@@ -0,0 +1,117 @@
...
+
+        await page.evaluate(() => {
+          const style = document.createElement('style');
+          style.innerHTML = `
+            :root {
+              --ion-safe-area-left: 50px;
+              --ion-safe-area-right: 10px;
+            }
+          `;
+          document.head.appendChild(style);
```
amandaejohnston, line 25: I know this new `<style>` will take priority since it's lower on the page, but it seems strange to append a new one when safe area styles already exist on the page we're visiting. Could we query for the existing style tag and update its innerHTML instead?

core/src/components/menu/test/safe-area/menu.e2e.ts
```diff
@@ -0,0 +1,117 @@
...
+
+        await page.click('#open-start');
+        await ionDidOpen.next();
+
+        await expect(page).toHaveScreenshot(screenshot(`menu-start-safe-area-left-notch`));
+
+        await page.locator('[menu-id="start-menu"]').evaluate(async (el: HTMLIonMenuElement) => {
+          await el.close();
+        });
+        await ionDidClose.next();
```
amandaejohnston, line 36: Do we need to close the menus? It adds extra time to the tests, and I think the page will be refreshed at the start of each new test anyway due to the `beforeEach`.

core/src/components/menu/test/safe-area/menu.e2e.ts-snapshots/menu-end-safe-area-right-notch-md-ltr-Mobile-Chrome-linux.png
amandaejohnston, line 1: Could we add some content to the menu that reaches across the whole thing, maybe wrapping to new lines, to better showcase the padding being added in cases like this? At the moment, you can't really tell anything is different.

core/src/components/menu/test/safe-area/menu.e2e.ts-snapshots/menu-start-safe-area-left-notch-md-rtl-Mobile-Firefox-linux.png
amandaejohnston, line 1: The header text is slightly offset from the item here; should it be?

core/src/components/menu/test/safe-area/menu.e2e.ts-snapshots/menu-start-safe-area-right-notch-md-rtl-Mobile-Chrome-linux.png
amandaejohnston, line 1: Looks like the right notch padding only affected the header here 🤔

amandaejohnston (changes requested):

core/src/components/menu/test/safe-area/menu.e2e.ts-snapshots/menu-end-safe-area-left-notch-md-rtl-Mobile-Chrome-linux.png
thetaPC: It's being applied. Padding is added to the left side of the menu and no padding on the right side side. The issue is that the title isn't long enough to highlight the padding (I wish there was a way to capture padding highlighting). Here's what it will look like.

Screenshot 2023-09-15 at 8 29 39 AM

core/src/components/menu/test/safe-area/menu.e2e.ts-snapshots/menu-start-safe-area-left-notch-md-rtl-Mobile-Firefox-linux.png
thetaPC: It shouldn't. It's a known issue that will be addressed in a different PR.

core/src/components/menu/test/safe-area/menu.e2e.ts-snapshots/menu-start-safe-area-right-notch-md-rtl-Mobile-Chrome-linux.png
thetaPC: This is the same issue that you [spotted](https://github.com/ionic-team/ionic-framework/pull/28177#discussion_r1327361324) with the offset. It's a known issue that will be fixed in a different PR. `ion-item` isn't using the correct safe area padding.

core/src/components/menu/test/safe-area/menu.e2e.ts-snapshots/menu-end-safe-area-left-notch-md-rtl-Mobile-Chrome-linux.png
amandaejohnston, line 1: It looks like the padding was applied on two different sides here -- the left side for the header (correct) but the right side for the item (incorrect). Is that related to the known issue mentioned in https://github.com/ionic-team/ionic-framework/pull/28177#discussion_r1327507716?

core/src/components/menu/test/safe-area/menu.e2e.ts-snapshots/menu-start-safe-area-left-notch-md-rtl-Mobile-Firefox-linux.png
amandaejohnston: Do we have a ticket for that yet?

core/src/components/menu/test/safe-area/menu.e2e.ts-snapshots/menu-end-safe-area-left-notch-md-rtl-Mobile-Chrome-linux.png
thetaPC: Screenshot 2023-09-15 at 2 37 14 PM

I'm not seeing padding on the right side on the `ion-toolbar`. The padding that you're seeing is from `ion-title`. That is the expected padding regardless of the app or device direction. Is this what you're referring to?

core/src/components/menu/test/safe-area/menu.e2e.ts-snapshots/menu-start-safe-area-left-notch-md-rtl-Mobile-Firefox-linux.png
thetaPC: Yup, I created it today (FW-5172).

core/src/components/menu/test/safe-area/menu.e2e.ts-snapshots/menu-end-safe-area-left-notch-md-rtl-Mobile-Chrome-linux.png
amandaejohnston: The right side on the item, not the toolbar. It looks like the header and content are being pushed in two different directions:

image

amandaejohnston: Actually, looking at the ticket for the issue with `ion-item`, it sounds like this is expected. 👍

amandaejohnston (approved):

core/src/components/menu/test/safe-area/menu.e2e.ts-snapshots/menu-end-safe-area-left-notch-md-rtl-Mobile-Chrome-linux.png
thetaPC: Yup. The `ion-item` isn't applying the safe area padding correctly so it looks like that. Ideally, it will line up with the title after the fix.

PR #28184, opened 2023-09-15
feat(input): support dynamic type

amandaejohnston:
Issue number: N/A

---------

`ion-input` does not support dynamic type.

## What is the new behavior?

`ion-input` supports dynamic type. Most of the sizing was already inherited; only the helper text needed updating.

- [ ] Yes
- [x] No


sean-perkins (approved):
Tested it out on device with a variety of our test templates and everything is looking good 👍 Love that it was a very straightforward change!

PR #28209, opened 2023-09-21
refactor(many): convert dynamic type rem values to Sass functions

brandyscarney:
Converts some of the `rem` units for dynamic font scaling to use the `dynamic-font` functions that didn't exist when these were implemented. 

There were a few other components that still use the old functions `clamp`, `min`, `max` (chip, datetime) but I did not implement those so I don't know the purpose of leaving them. I can migrate if desired.


core/src/components/list-header/list-header.ios.vars.scss
```diff
@@ -8,7 +8,7 @@
 $list-header-ios-padding-start:       $item-ios-padding-start !default;
 
 /// @prop - Font size of the list header
-$list-header-ios-font-size:           min(1.375rem, 56px) !default;
+$list-header-ios-font-size:           dynamic-font-max(22px, 2.55) !default;
```
brandyscarney, line 11: `22px * 2.55 = 56.1px` which may be causing the weird screenshot diffs from the max being `56px` previously

core/src/components/toolbar/toolbar.ios.vars.scss
```diff
@@ -34,7 +34,7 @@ $toolbar-ios-padding-start:                         $toolbar-ios-padding-end !de
 /// @prop - Font size of the toolbar button
 /// The minimum and maximum font sizes for a toolbar button are
 /// 100% and 135%, respectively, of their default font size
-$toolbar-ios-button-font-size:                      clamp(17px, 1.0625rem, 21px) !default;
+$toolbar-ios-button-font-size:                      dynamic-font-clamp(1, 17px, 1.24) !default;
```
brandyscarney, line 37: `17px * 1.24 = 21.08px` which may be causing the weird screenshot diffs from the max being `21px` previously

liamdebeasi (approved):

amandaejohnston (approved):

PR #28224, opened 2023-09-23
feat(safe-area): safe area mixin

thetaPC:
Issue number: N/A

---------

The safe area defines the area within a view that isn't covered by a navigation bar, tab bar, toolbar, or other views a view controller might provide.

The safe areas are reliant on the device's direction, not the app's. This leads to using styles like `padding-right` and `padding-left`. However, this can cause issues with the app's direction (LTR/RTL). The team needs to add the correct styles, which can lead to either a bigger block of code or forgetting about the fact that it's based on device direction.

## What is the new behavior?

I added a new mixin that will take accept two values: `start` and `end`. These values are the additional values that the developer wants to add to either `--ion-safe-area-left` or `--ion-safe-area-right`. For example, `start` will be added to `ion-safe-area-left` in ltr and `ion-safe-area-right` in rtl. The `end` will be added to `ion-safe-area-right` in ltr and `ion-safe-area-left` in rtl.

The team wouldn't have to manually do the math with the mixin.

- [ ] Yes
- [x] No

## Other information

I only updated the `.item-divider-inner` style to use the mixin to start with. The point of this PR is to introduce the mixin. Another PR will be created to add the mixin to the styles.


core/src/components/item-divider/item-divider.scss
```diff
@@ -75,18 +75,7 @@
...
-    padding-right: calc(var(--ion-safe-area-right, 0px) + var(--inner-padding-end));
-    padding-left: var(--inner-padding-start);
-  }
-
-  @include rtl() {
-    padding-right: var(--inner-padding-start);
-    padding-left: calc(var(--ion-safe-area-left, 0px) + var(--inner-padding-end));
-  }
-  /* stylelint-enable */
+  @include safe-area-padding(null, var(--inner-padding-end));
```
brandyscarney, line 78: This is no longer using the `--inner-padding-start` property:

<table>
<tr>
  <td>Before</td>
  <td>After</td>
</tr>
<tr>
<td>

```css
.item-divider-inner {
  padding-right: calc(var(--ion-safe-area-right, 0px) + var(--inner-padding-end));
  padding-left: var(--inner-padding-start);
}

:host-context([dir=rtl]) .item-divider-inner {
  padding-right: var(--inner-padding-start);
  padding-left: calc(var(--ion-safe-area-left, 0px) + var(--inner-padding-end));
}

[dir=rtl] .item-divider-inner {
  padding-right: var(--inner-padding-start);
  padding-left: calc(var(--ion-safe-area-left, 0px) + var(--inner-padding-end));
}

@supports selector(:dir(rtl)) {
  .item-divider-inner:dir(rtl) {
    padding-right: var(--inner-padding-start);
    padding-left: calc(var(--ion-safe-area-left, 0px) + var(--inner-padding-end));
  }
}
```

</td>
<td>

```css
.item-divider-inner {
  padding-right: calc(var(--ion-safe-area-right, 0) + var(--inner-padding-end));
}

:host-context([dir=rtl]) .item-divider-inner {
  padding-left: calc(var(--ion-safe-area-left, 0) + var(--inner-padding-end));
  padding-right: unset;
}

[dir=rtl] .item-divider-inner {
  padding-left: calc(var(--ion-safe-area-left, 0) + var(--inner-padding-end));
  padding-right: unset;
}

@supports selector(:dir(rtl)) {
  .item-divider-inner:dir(rtl) {
    padding-left: calc(var(--ion-safe-area-left, 0) + var(--inner-padding-end));
    padding-right: unset;
  }
}
```

</td>
</tr>
</table>

brandyscarney (changes requested):

thetaPC:
Closing this because it's starting to get more complex. The complexity doesn't justify the amount of times that safe area is being used in Ionic.

PR #28231, opened 2023-09-25
fix(title): large title uses custom font on transition

liamdebeasi:
Issue number: N/A

---------

When testing Dynamic Font Scaling with a custom font I noticed that the large title does not respect `--ion-font-family` on transition.

## What is the new behavior?

- The cloned large title now respect `--ion-font-family`

Note: This happens in `main` too which is why I am merging into there instead of the Dynamic Font Scaling branch.

| `main` | branch |
| - | - |
| IMG_0182 | IMG_0183 |

- [ ] Yes
- [x] No

### DIFF
--- a/core/src/components/title/title.ios.scss
+++ b/core/src/components/title/title.ios.scss
@@ -61,6 +61,7 @@
 
 :host(.title-large.ion-cloned-element) {
   --color: #{$text-color};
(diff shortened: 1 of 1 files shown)


thetaPC (approved):
LGTM

PR #28237, opened 2023-09-26
fix(radio): radio mark icon scales with font

liamdebeasi:
Issue number: N/A

---------

Maria noted that the radio mark does not scale but should according to the iOS guidelines:

Simulator Screenshot - iPhone 14 Pro - 2023-09-26 at 11 44 44

## What is the new behavior?

- Radio mark scales with font

- [ ] Yes
- [x] No


thetaPC (approved):
LGTM

amandaejohnston (approved):

PR #28239, opened 2023-09-26
fix(select): add scaling to icon 

thetaPC:
Issue number: N/A

---------

- The icon on `ion-select` does not scale up when font size changes.

## What is the new behavior?

- The icon on `ion-select` does scale up when font size changes.
- Added screenshot test.

- [ ] Yes
- [x] No

## Other information

I didn't see any min or max when comparing it to a native iOS select. A native iOS select can be found through the Calendar app when adding a new event.


core/src/components/select/test/a11y/select.e2e.ts
```diff
@@ -12,3 +12,30 @@ configs({ modes: ['ios'], directions: ['ltr'] }).forEach(({ title, config }) =>
...
+    test.describe(title('select: font scaling'), () => {
+      test('should scale text on larger font sizes', async ({ page }) => {
+        await page.setContent(
+          `
+            <style>
+              html {
+                font-size: 310%;
+              }
+            </style>
+            <ion-select label="My Label" value="oranges" placeholder="Placeholder">
```
liamdebeasi, line 27: Can you put this in a wrapper with a width that makes the text not truncate? Right now I believe it's getting scaled down to the smaller width possible

PR #28245, opened 2023-09-27
fix(content): fullscreen offset is computed correctly with tab bar

### LINKED ISSUE #21130: bug: Platform MD ion-content wrong --offset-bottom
# Bug Report

**Ionic version:**
[x] **5.x**

**Where?**
In the Ionic 5 tabs example. 
On the **iOS** theme this **works**. 
On the **MD** theme it **does not work**.

**Current behavior:**
`<app-explore-container>` has a wrong y-position in the tab, which was initially loaded.
The `<ion-content>` has on reload of the current tab a wrong offset of `--offset-bottom:0px;`

**Expected behavior:**
The `<app-explore-container>` component should have the same y-position on every tab on the MD (Material Design) theme.
<ion-content> should have an offset of `--offset-bottom:57px;` on reload of the current tab, as the other tabs have, if you switch to some other tab.

**Steps to reproduce:**
Start a new Ionic 5 Tab Project and open Chrome with the Pixel 2 responsive emulator and switch between the tabs. The content position will change.

**Ionic info:**

```
Ionic:

   ionic (Ionic CLI)             : 4.10.2 (/usr/local/lib/node_modules/ionic)
   Ionic Framework               : @ionic/angular 5.0.7
   @angular-devkit/build-angular : 0.803.26
   @angular-devkit/schematics    : 8.3.26
   @angular/cli                  : 8.3.26

Cordova:

   Cordova Platforms     : ios 5.1.1
   Cordova Plugins       : cordova-plugin-ionic-keyboard 2.2.0, cordova-plugin-ionic-webview 4.2.0, (and 4 other plugins)

System:

   Android SDK Tools : 25.2.3 (/Users/stefanrein/Library/Android/sdk)
   ios-sim           : 8.0.2
   NodeJS            : v10.15.3 (/usr/local/bin/node)
   npm               : 6.14.3
   OS                : macOS
   Xcode             : Xcode 11.4 Build version 11E146


```

Example, when the Tab1 URL was initially loaded: http://localhost:4200/tabs/tab1

image

https://github.com/ionic-team/starters/tree/master/angular/official/tabs

liamdebeasi:
Issue number: resolves #21130

---------

IonContent sets `--offset-top` and `--offset-bottom` variables to allow the content to scroll under headers, footers, and tab bars. This is essential to creating the translucency effect on these components.

IonContent does this by computing its offsetHeight and offsetTop coordinates which take into account the dimensions of headers, footers, and tab bars. Occasionally, this code will run before the IonTabBar has been hydrated which means that the offset will be wrong because the IonTabBar will have a dimension of 0x0 prior to hydration.

This impacts Ionic Angular devs who are using the lazy loaded build of Ionic. React and Vue devs are not impacted because they are using the dist-custom-elements build of Ionic which does not have hydration.

## What is the new behavior?

- IonContent will re-run the offset computation code whenever the `ionTabBarLoaded` event is emitted. This event is emitted at most once per IonTabBar instance.

- [ ] Yes
- [x] No

## Other information

Dev build: `7.4.2-dev.11695831341.191bdf12`

Note: I did not write a test since this is fixing a race condition. I wasn't able to find a non-flaky way of testing this. You can test this in an Ionic Angular Tabs starter application with the dev build. The `--offset-bottom` variable on `ion-content` should be large enough such that the content will scroll under the tab bar. The translucency effect won't work just yet, but that is being fixed in https://github.com/ionic-team/ionic-framework/pull/28246.

### DIFF
--- a/core/src/components/content/content.tsx
+++ b/core/src/components/content/content.tsx
@@ -34,6 +34,9 @@ export class Content implements ComponentInterface {
   private isMainContent = true;
   private resizeTimeout: ReturnType<typeof setTimeout> | null = null;
 
+  private tabsElement: HTMLElement | null = null;
+  private tabsLoadCallback?: () => void;
+
   // Detail is used in a hot loop in the scroll event, by allocating it here
   // V8 will be able to inline any read/write to it since it's a monomorphic class.
   // https://mrale.ph/blog/2015/01/11/whats-up-with-monomorphism.html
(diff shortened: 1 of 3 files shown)


core/src/components/content/content.tsx
```diff
@@ -115,15 +118,61 @@ export class Content implements ComponentInterface {
...
+        closestTabs.addEventListener('ionTabBarLoaded', this.tabsLoadCallback);
+      }
+    }
   }
 
   disconnectedCallback() {
     this.onScrollEnd();
-  }
 
-  @Listen('appload', { target: 'window' })
```
liamdebeasi, line 124: `appload` does not always run after all components have been hydrated, so this listener wasn't actually helping at all.

sean-perkins (approved):

PR #28248, opened 2023-09-27
feat(toast): allow custom positioning relative to specific element

### LINKED ISSUE #17499: feat: allow for custom ion-toast positioning
# Feature Request

**Ionic version:**
[x] **4.x**

Currently with the implementation of toasts in Ionic v4, we have no way of adding some padding or margin to an `ion-toast`. I'd like to be able to customize this position in order to adhere to the material design principle outlined **[here](https://material.io/design/components/snackbars.html#placement)**. 

Ideally, this would be something that could be added in the CSS custom properties.

amandaejohnston:
Issue number: resolves #17499

---------

Currently, there isn't a way to position toasts such that they don't overlap navigation elements such as headers, footers, and FABs.

## What is the new behavior?

Added the new `positionAnchor` property, which specifies an element that the toast's position should be anchored to.

While the name can be tweaked, we should take care to keep the relation between it and the `position` property clear. The `position` acts as a sort of "origin" point, and the toast is moved from there to sit near the chosen anchor element. This is important because it helps clarify why the toast sits above the anchor for `position="bottom"` and vice versa.

I chose not to rename the `position` prop itself to avoid breaking changes.

Docs PR: https://github.com/ionic-team/ionic-docs/pull/3158

- [ ] Yes
- [x] No


core/src/components/toast/toast.tsx
```diff
@@ -275,7 +290,33 @@ export class Toast implements ComponentInterface, OverlayInterface {
...
+      /**
+       * If the anchor is defined as an ID, find the element.
+       * We do this on every present so the toast doesn't need
+       * to account for the surrounding DOM changing since the
+       * last time it was presented.
+       */
+
+      const foundEl = document.getElementById(positionAnchor);
+      if (foundEl === null) {
+        printIonWarning(`An anchor element with an ID of ${positionAnchor} was not found in the DOM.`, this.el);
```
liamdebeasi, line 308:
```suggestion
        printIonWarning(`An anchor element with an ID of "${positionAnchor}" was not found in the DOM.`, this.el);
```

The trigger controller wrap the ID in quotes, so we should probably do the same here: https://github.com/ionic-team/ionic-framework/blob/ac2c8e6c22da4d0d8224def24ddef56ee9d26246/core/src/utils/overlays.ts#L810

core/src/components/toast/toast.tsx
```diff
@@ -275,7 +290,33 @@ export class Toast implements ComponentInterface, OverlayInterface {
...
+      } else {
+        anchor = foundEl;
+      }
+    } else {
+      anchor = positionAnchor;
```
liamdebeasi, line 313: Should we add a check to verify that the value passed is an instance of `HTMLElement`? This would prevent devs from accidentally passing numbers, objects, etc.

core/src/components/toast/animations/ios.leave.ts
```diff
@@ -13,20 +13,23 @@ export const iosLeaveAnimation = (baseEl: HTMLElement, position: string): Animat
...
+  /**
+   * For top/bottom positions, we use to() instead of fromTo() so that if
+   * the positionAnchor prop was used, we don't need to recalculate the
+   * toast and/or anchor's position and duplicate work from the enter
+   * animation.
+   */
   switch (position) {
     case 'top':
-      wrapperAnimation.fromTo('transform', `translateY(${top})`, 'translateY(-100%)');
+      wrapperAnimation.to('transform', 'translateY(-100%)');
```
liamdebeasi, line 26: Implicit Keyframes are somewhat buggy on WebKit, so we may want to avoid it for now: https://caniuse.com/?search=implicit%20keyframes

It's gotten better since iOS 14, but you can see in the demo below that the leaving animation is still broken.

https://github.com/ionic-team/ionic-framework/assets/2721089/e349ff3a-22b2-4db8-98a5-8a483be0869f

core/src/components/toast/animations/md.enter.ts
```diff
@@ -1,20 +1,59 @@
...
+      /**
+       * Just box.top is the distance from the top edge of the screen
+       * to the top edge of the anchor. We want to calculate from the
+       * bottom edge of the screen instead.
+       */
+      offset += win.innerHeight - box.top;
+    }
+
+    /**
+     * We don't include safe area here because that should already be
```
liamdebeasi, line 48: Optional: We could move this to a utility function so both iOS/MD animations can use it. However, since it's only used twice this is optional.

liamdebeasi (changes requested):

core/src/components/toast/toast.tsx
amandaejohnston: This should be prevented by Typescript, since `positionAnchor` can only take `HTMLElement | string | undefined`.

liamdebeasi: What about devs not using TypeScript?

amandaejohnston: Fair enough!

core/src/components/toast/animations/md.enter.ts
amandaejohnston: I decided against this because the two are subtly different -- they have different starting offsets and, more importantly, the `bottom` positions are generally calculated using addition for `md` but subtraction for `ios`. This is because `md` uses `style.bottom`, meaning the position is calculated from the bottom edge of the screen, but `ios` uses `translateY` which calculates from the top edge instead.

I could certainly have the util function check the `mode` and act accordingly, but the context behind the difference would sit outside the util, making it harder to follow the code. There's probably some refactoring that could be done here to align the two modes, but that feels outside the scope of this PR.

core/src/components/toast/animations/ios.enter.ts
```diff
@@ -1,20 +1,58 @@
...
+  let top, bottom;
+
+  /**
+   * If positionAnchor is defined, add in the distance from the top
+   * of the screen to the target anchor edge. For position="top", the
+   * bottom anchor edge is targeted. For position="bottom", the top
+   * anchor edge is targeted.
+   */
+  if (positionAnchor && win) {
+    const box = positionAnchor.getBoundingClientRect();
```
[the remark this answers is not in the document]
amandaejohnston: Added in https://github.com/ionic-team/ionic-framework/pull/28248/commits/1e2a8b79ae194a89a62bd32970474664696ea437. LMK if you'd like me to tweak the messaging or put the function somewhere else 👀

core/src/components/toast/animations/md.enter.ts
amandaejohnston: Although, if I need to avoid implicit keyframes in the leave animation, in theory it would make more sense to do this calculation out in `toast.tsx` so I can pass it to both the enter and leave animations and save on calculations 🤔 Hmm. I'll play around with it some.

amandaejohnston: I've gone with the above idea. Let me know if you think it'd be worth it to make a tech debt ticket to sync up how the toast's position is calculated between modes, so the new `getAnimationPosition` util can be cleaned up a little. I'm not sure if it's worth the effort since the position-related CSS would need touching too 😅

core/src/components/toast/animations/utils.ts
```diff
@@ -0,0 +1,84 @@
...
+      top: `calc(${offset}px + var(--ion-safe-area-top, 0px))`,
+      bottom:
+        mode === 'md'
+          ? `calc(${offset}px + var(--ion-safe-area-bottom, 0px))`
+          : `calc(${offset}px - var(--ion-safe-area-bottom, 0px))`,
+    };
+  }
+}
+
+function warnIfAnchorIsHidden(positionAnchor: HTMLElement, toast: HTMLElement) {
```
liamdebeasi, line 77: Can we add a comment that explains what happens if the anchor is hidden?

core/src/components/toast/animations/md.enter.ts
liamdebeasi: I like your approach here. The additional benefit is if devs want to create a custom toast animation they can still take advantage of the anchor/position information since we pass that in.

core/src/components/toast/toast.tsx
```diff
@@ -56,6 +64,7 @@ export class Toast implements ComponentInterface, OverlayInterface {
   private readonly triggerController = createTriggerController();
   private customHTMLEnabled = config.get('innerHTMLTemplatesEnabled', ENABLE_HTML_CONTENT_DEFAULT);
   private durationTimeout?: ReturnType<typeof setTimeout>;
+  private lastPresentedPosition?: ToastAnimationPosition;
```
liamdebeasi, line 67: Can we add a comment explaining what this variable is for?

core/src/components/toast/toast.tsx
```diff
@@ -315,7 +349,16 @@ export class Toast implements ComponentInterface, OverlayInterface {
...
+      /**
+       * Fetch the cached position that was calculated back in the present
+       * animation. We always want to animate the dismiss from the same
+       * position the present stopped at, so the animation looks continuous.
+       */
+      {
+        position,
+        top: lastPresentedPosition?.top ?? '',
+        bottom: lastPresentedPosition?.bottom ?? '',
+      }
```
liamdebeasi, line 361: I know `lastPresentedPosition` is updated whenever `present` is called again, but can we still clear the reference in `dismiss`? The information is no longer relevant after the toast has been dismissed.

core/src/components/toast/toast.tsx
```diff
@@ -275,7 +293,21 @@ export class Toast implements ComponentInterface, OverlayInterface {
 
     await this.delegateController.attachViewToDom();
 
-    await present<ToastPresentOptions>(this, 'toastEnter', iosEnterAnimation, mdEnterAnimation, this.position);
+    const { el, position } = this;
+    const anchor = this.getAnchorElement();
+    const animationPosition = getAnimationPosition(position, anchor, getMode(), el);
```
liamdebeasi, line 298: The `getMode` call here will get the application mode, but it should get the component mode instead.

For example, if the app is in MD mode but the toast is in iOS mode then the toast currently does not appear. You can test this by setting the footerToast" mode to "ios".

Instead, we should call `getIonMode(this)` which will take into account the component's mode.

core/src/components/toast/animations/utils.ts
```diff
@@ -0,0 +1,84 @@
...
+ * Note that MD animates bottom-positioned toasts using style.bottom,
+ * which calculates from the bottom edge of the screen, while iOS uses
+ * translateY, which calculates from the top edge of the screen. This
+ * is why the bottom calculates differ slightly between modes.
+ */
+export function getAnimationPosition(
+  position: ToastPosition,
+  positionAnchor: HTMLElement | undefined,
+  mode: Mode,
+  toast: HTMLElement
```
liamdebeasi, line 20: I know the params are probably pretty obvious, but could we still document what they are for? Example: https://github.com/ionic-team/ionic-framework/blob/ac2c8e6c22da4d0d8224def24ddef56ee9d26246/core/src/utils/helpers.ts#L178-L182

liamdebeasi (changes requested):

liamdebeasi:
I reverted a flaky screenshot diff in https://github.com/ionic-team/ionic-framework/pull/28248/commits/5ded707681b73f8014f2fba1224e7bf90d6853ae, so make sure you pull latest

core/src/components/toast/animations/utils.ts
amandaejohnston: Added in https://github.com/ionic-team/ionic-framework/pull/28248/commits/82ac509b3fafcf62ba8bdd285d79c33b20850b14. LMK if you'd prefer more detail; I didn't want to add too much since the case is unsupported, so the specific behavior doesn't matter as much and might change in the future.

liamdebeasi (approved):
Great work! I'm fine with the naming convention.

sean-perkins (approved):
This feature works great on device 🎉 this is such a great improvement for the user experience!

PR #28250, opened 2023-09-27
test: remove deprecated getSnapshotSettings method

liamdebeasi:
Issue number: N/A

---------

The `getSnapshotSettings` method was used in the pre-generator test infrastructure. It was deprecated at the start of the generator migration.

## What is the new behavior?

- Now that the generator migration is complete, this method is no longer needed. As a result, I removed it. Developers should use the `screenshot` function: https://github.com/ionic-team/ionic-framework/blob/main/core/src/utils/test/playwright/docs/api.md#using-the-return-value-from-each-configuration

- [ ] Yes
- [x] No


brandyscarney (approved):
We should also probably update the Notion doc as it still mentions `getSnapshotSettings` and (optionally) remove it from this commented out code: https://github.com/ionic-team/ionic-framework/blob/eb41b556b57c97139b9c36dc3e3be3711d8afaca/core/src/components/picker-column/test/test.utils.ts#L42

liamdebeasi:
My plan (once I finish these changes) is to have Notion link to the GitHub guide so we have a single source of truth for our Playwright utils. I opted to move it to GitHub so community contributors can use it too. I'll remove the commented out reference though

PR #28258, opened 2023-09-28
fix(checkbox): add scaling to ios alert and checkbox

thetaPC:
Issue number: N/A

---------

Checkboxes inside alerts were not scaling.

## What is the new behavior?

- Checkboxes inside alerts are scaling.
- Switched from `math.div` to `calc` since there is a warning when using rem: `math.div() will only support number arguments in a future release.`
- Found the correct max size for checkboxes on iOS
- Dug deeper into the size for checkboxes on Android. I couldn't find a native example but did find an example for radio. Radios do not scale at all. I'm assuming that checkboxes follow that same logic. So checkboxes on Android will not scale.

- [ ] Yes
- [x] No

## Other information

A max width example can be found on the Health app on iOS.


liamdebeasi:
Synced the changes in. Re-running screenshots now

core/src/components/checkbox/checkbox.ios.vars.scss
thetaPC, line 1: Reverting to not scale. I dug deeper into the size for checkboxes on Android. I couldn't find a native example but did find an example for radio. Radios do not scale at all. I'm assuming that checkboxes follow that same logic. So checkboxes on Android will not scale.

thetaPC:
@liamdebeasi okay, now it's ready for review. Thank you!

core/src/components/checkbox/checkbox.ios.vars.scss
liamdebeasi: Thanks for catching this! Can confirm that checkboxes don't scale either: 
image

Mind updating the design doc for this when you get a chance?

core/src/components/checkbox/checkbox.ios.vars.scss
```diff
@@ -15,7 +15,7 @@ $checkbox-ios-background-color-on:        ion-color(primary, base) !default;
 $checkbox-ios-background-color-focused:   ion-color(primary, tint) !default;
 
 /// @prop - Size of the checkbox icon
-$checkbox-ios-icon-size:                  dynamic-font-max(26px, 2.38) !default;
+$checkbox-ios-icon-size:                  dynamic-font-max(26px, 2.535) !default;
```
liamdebeasi, line 18: Why did this change?

thetaPC: I was able to get PixelSnap installed and found that the checkboxes are actually bigger. The max is `66px`.

core/src/components/checkbox/checkbox.ios.vars.scss
thetaPC: Created a PR.

amandaejohnston (approved):

core/src/components/checkbox/checkbox.ios.vars.scss
sean-perkins: If the max is `66px`, shouldn't this be `2.538`?

thetaPC: Done

core/src/components/checkbox/checkbox.md.vars.scss
```diff
@@ -17,7 +17,7 @@ $checkbox-md-background-color-focused:   ion-color(primary, tint) !default;
 $checkbox-md-icon-background-color-on:   ion-color(primary, base) !default;
 
 /// @prop - Size of the checkbox icon
-$checkbox-md-icon-size:                  dynamic-font(18px) !default;
+$checkbox-md-icon-size:                  18px !default;
```
liamdebeasi, line 20: Can we add a code comment that explains why we do not use dynamic-font here?

liamdebeasi (approved):

core/src/components/checkbox/checkbox.md.vars.scss
thetaPC: done

liamdebeasi (approved):
Thank you!

Issue #28119, opened 2023-09-05
bug: Popover calculation seems to be incorect


nstekovic:
### Prerequisites

- [X] I have read the [Contributing Guidelines](https://github.com/ionic-team/ionic-framework/blob/main/.github/CONTRIBUTING.md#creating-an-issue).
- [X] I agree to follow the [Code of Conduct](https://ionicframework.com/code-of-conduct).
- [X] I have searched for [existing issues](https://github.com/ionic-team/ionic-framework/issues) that already report this problem, without success.

### Ionic Framework Version

v7.x

### Current Behavior

Currently, popover is determined using calculateWindowAdjustment and that relies on document body. This is all fine in theory but I think there might be a cases where we would like to actually exclude some elements from calculation instead of checking for `triggerTop + triggerHeight + contentHeight > bodyHeight` (https://github.com/ionic-team/ionic-framework/blob/main/core/src/components/popover/utils.ts#L865) only. 

Currently, calculations were done only comparing to body height and nothing that could be done here. For example I don't want my popover to overlap footer when opening.

![image](https://github.com/ionic-team/ionic-framework/assets/135704500/2b5394da-8cdc-42cb-9e49-101e79d418ab)

### Expected Behavior

Imagine that we want to take tabs footer into consideration or some sticky element above it. We do have ion-content (main section) above it that is scrollable. In mobile devices, content that opens popover would be very close to the footer or some element that is above footer which we'd like to also take into account when determining whether popover should open towards bottom or top.
We'd like to have behaviour as following "set all popovers side to be 'bottom' but if it would overlap with that element - open it towards top".

### Steps to Reproduce

- Clone github repo that is attached
- npm install 
- npm run start
- Once it runs the browser, toggle mobile view in developer console  and select any mobile device (same behaviour is observed when running ios simulator)
- In tab1 view, scroll down where your button would be near footer
- open popover and observe behaviour
If `triggerTop + triggerHeight + contentHeight > bodyHeight` is actually true - it would open towards top / if not it would open towards bottom.
This doesn't necessarily need to be true if our trigger height is smaller eg (16px and our content is 20px).

### Code Reproduction URL

https://github.com/nstekovic/ionic-popover-issue

### Ionic Info

```
Ionic:

   Ionic CLI                     : 7.1.1 (/Users/[user]/.nvm/versions/node/v18.17.0/lib/node_modules/@ionic/cli)
   Ionic Framework               : @ionic/angular 7.3.2
[...]

PR #28262, opened 2023-09-29
fix(popover): remove duplicate calculation to prevent popover from covering trigger

liamdebeasi:
Issue number: N/A

---------

As noted in https://github.com/ionic-team/ionic-framework/issues/28119, we are doing additional work when computing the popover positioning that doesn't need to happen.  When computing `triggerTop` we add the trigger height to the top coordinate of the trigger: https://github.com/ionic-team/ionic-framework/blob/ac2c8e6c22da4d0d8224def24ddef56ee9d26246/core/src/components/popover/utils.ts#L834

This is wrong because that would actually give us the bottom coordinate of the trigger. However, we then subtract `triggerHeight` from `triggerTop` to get the coordinate when computing the adjusted `top` value: https://github.com/ionic-team/ionic-framework/blob/ac2c8e6c22da4d0d8224def24ddef56ee9d26246/core/src/components/popover/utils.ts#L877

However, there are edge cases where that value is not negated (i.e. that code is not run) so it can cause layout issues.

## What is the new behavior?

- Removed the additional `triggerHeight` computation

- [ ] Yes
- [x] No

## Other information

Note: The screenshot updates here are correct. The popover should typically never cover the trigger element. Instead, the popover (correctly) is displayed above the trigger. This allows the full popover to be displayed relative to the trigger without covering it up.


core/src/components/datetime/test/position/datetime.e2e.ts-snapshots/datetime-position-popover-ios-ltr-Mobile-Safari-linux.png
amandaejohnston, line 1: I know you said the screenshot updates are correct, but they seem to show the opposite from what you described: the popover used to sit above the trigger, but now it's covering the trigger up slightly. Is that intentional? 🤔

core/src/components/popover/utils.ts
```diff
@@ -888,6 +886,8 @@ export const calculateWindowAdjustment = (
     }
   }
 
+  console.log(top, arrowTop);
```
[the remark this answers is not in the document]
liamdebeasi: Nope fixed in https://github.com/ionic-team/ionic-framework/pull/28262/commits/4cd23fb1dab86cee3d0b858e52d0e63b2fe3620a

core/src/components/datetime/test/position/datetime.e2e.ts-snapshots/datetime-position-popover-ios-ltr-Mobile-Safari-linux.png
liamdebeasi: Oh oops I got confused 🤔 I'm going to move this back to draft until I can look at this some more.

liamdebeasi:
As Amanda noted, popover does appear to be working correctly. I got confused during my initial analysis. Going to close.

PR #28265, opened 2023-09-29
fix(select): use correct aria-haspopup value

thetaPC:
Issue number: internal

---------

`ion-select` uses the `aria-haspopup="listbox"` even though the popup is a dialog.

## What is the new behavior?

- `ion-select` triggers a dialog, the value for `aria-haspopup` now reflects that

- [ ] Yes
- [x] No

## Other information

N/A

### DIFF
--- a/core/src/components/select/select.tsx
+++ b/core/src/components/select/select.tsx
@@ -1041,7 +1041,7 @@ Developers can use the "legacy" property to continue using the legacy form marku
         disabled={disabled}
         id={inputId}
         aria-label={this.ariaLabel}
-        aria-haspopup="listbox"
+        aria-haspopup="dialog"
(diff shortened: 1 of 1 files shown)


liamdebeasi (approved):
I tested the changes on TalkBack + Android, VoiceOver + iOS, VoiceOver + Safari for macOS, NVDA + Chrome, and NVDA + Firefox. They all work well! Can we hold off on merging this until https://github.com/ionic-team/ionic-docs/pull/3156 is merged? Would be good to have these changes ship together.

thetaPC:
@liamdebeasi I'll switch the base branch to `feature-7.5` since the docs is waiting on `FW-5128`. The work for `FW-5128` will target a minor release.

PR #27814, opened 2023-07-17
fix(angular): menu button is enabled with Angular Universal builds

sean-perkins:
Issue number: Resolves #27524 #22206

---------

## What is the current behavior?

After destructive hydration from an Angular Universal build, the `ion-menu` is disabled. This results in a few problematic behaviors:
- If used with an `ion-split-pane`, the split pane will not render with the menu.
- If used with an `ion-menu-button`, the menu icon will not render and the menu will not be accessible.

## What is the new behavior?

- Removes automatically disabling the `ion-menu` when in a server environment.
- Prevents setting an `ion-menu` that no longer exists in the DOM, as an active menu. This allows the most recent menu to be enabled when rendered from Angular Universal builds.

Loom recording overviewing the change: https://www.loom.com/share/5a149218adc84cc2af435fc1b1f45124?sid=249028f4-be3d-4d4a-948a-597da83be95a

## Does this introduce a breaking change?

- [ ] Yes
- [x] No

## Other information

Dev-build: `7.1.4-dev.11689625795.1d16532b`

Issue #18683, opened 2019-07-01
bug: navigating with multiple split panes causes first menu to never be reactivated


yogeshshsf:
Hi. i am using 2 split-pane, firstmenu/firstpage will be my home screen. from firstpage i am navigating to secondmenu/secondpage. now when i return back to firstmenu/firstpage, the split-pane is getting hide, can someone help me on this please. see the screenshot and code below.

firstmenu/firstpage: 

![image](https://user-images.githubusercontent.com/48043733/60429106-f7f94480-9c17-11e9-8fb8-ccf2e3c26413.png)

secondmenu/secondpage:

![image](https://user-images.githubusercontent.com/48043733/60429305-6fc76f00-9c18-11e9-9a49-4f23971ad7ed.png)

firstmenu/firstpage:

![image](https://user-images.githubusercontent.com/48043733/60429364-8b327a00-9c18-11e9-9774-b98073a9718d.png)

Below is the github repository to test this:

[https://github.com/yogeshshsf/SplitpaneDemo](url)

yogeshshsf:
Hi. i am using 2 split-pane, firstmenu/firstpage will be my home screen. from firstpage i am navigating to secondmenu/secondpage. now when i return back to firstmenu/firstpage, the split-pane is getting hide, can someone help me on this please. see the screenshot and code below.

Members
 0
18 posts
Report post 
Posted May 29 (edited)
Hi simon. i am using 2 split-pane, firstmenu/firstpage will be my home screen. from firstpage i am navigating to secondmenu/secondpage. now when i return back to firstmenu/firstpage, the split-pane is getting hide. this is stopper for my application, can you help on this please. see the screenshot and code below.﻿﻿

firstmenu/firstpage: 

![image](https://user-images.githubusercontent.com/48043733/60429106-f7f94480-9c17-11e9-8fb8-ccf2e3c26413.png)

liamdebeasi:
Thanks for the issue. I think this problem stems from the fact that we only allow 1 menu to be enabled at a time. If you want these two side panes to be always be open, I would recommend wrapping them in something other than `ion-menu`.

In your demo repo, I wrapped them each in a `div` rather than an `ion-menu` and the issue disappeared. Can you try this and let me know if it works for your use case?

If you really need to use `ion-menu`, I would recommend only having 1 instance of the menu.

ErlendEllingsen:
[...]

PR #28269, opened 2023-10-02
fix(menu) menus on the same side are not automatically disabled

### LINKED ISSUE #18974: bug: menu stop working in an application that has two main page with each a side menu
# Bug Report

**Ionic version:**
[x] **4.7.X**
[x] **4.8.X**
[x] **4.9.0**
[x] **4.10**

**Current behavior:**
Application has two main pages.  Each of these pages have a different side menu.  When navigating the application this way, the menu stop being displayed.

**Expected behavior:**
Side menu should not stop being displayed when navigating.

**Steps to reproduce:**
1. Open side menu from initial page and click "Page # 2 with menu # 2 
2. Navigation works, and side menu # 2 is displayed.
3. Open side menu and click "Page # 1 with menu # 1".
4. Page # 1 is correctly loaded except the menu # 1 is not displayed anymore.
2b5e74df-4e47-4481-9520-782732a0049d

**Related code:**

Sample application on GitHub:
https://github.com/iamkinetic/ionic4-multiple-menus-bug

**Ionic info:**

```
Ionic:

   Ionic CLI                     : 5.2.1 (C:\Users\iamkinetic\AppData\Roaming\npm\node_modules\ionic)
   Ionic Framework               : @ionic/angular 4.7.1
   @ionic/angular-toolkit        : 2.0.0

Utility:

   cordova-res : not installed
   native-run  : 0.2.7

System:

   NodeJS : v11.4.0 (C:\Program Files\nodejs\node.exe)
   npm    : 6.10.2
   OS     : Windows 10
```

liamdebeasi:
Issue number: resolves #18974

---------

When multiple menus on the same `side` are registered, all but the most recent menu are disabled. For example, if a user starts on PageA with a `start` menu and then navigates to PageB which also has a `start` menu, then the menu on PageA will be disabled. The problem is that if users navigates back to PageA they will be unable to open the menu on that view because it is still disabled. This behavior impacts any Ionic developer trying to open a menu whether by calling the `open` method on the menu itself or on the `menuController`.

After discussing with the team, we believe the original intent of this behavior was to prevent users from accidentally opening the wrong menu when calling `menuController.open('start')`. This API allows developers to reference a menu by side, and since it's possible to have multiple menus on the same side it's also possible to open the wrong menu when referencing by side only.

However, this API starts to break down pretty quickly in a navigation scenario.

Sample Repo: https://github.com/liamdebeasi/multiple-menu-bug-repro

## Scenario 1: Referencing Menu by Side

1. On the "home" route click "Open 'start' menu". Observe that the home page menu opens.
2. Close the menu and click "Go to Page Two".
3. On the "page-two" route click "Open 'start' menu". Observe that the page two menu opens.
4. Go back to "home".
5. Click "Open 'start' menu". Observe that nothing happens.
6. Click "Enable and Open 'start'" Menu". Observe that the home menu opens.

## Scenario 2: Referencing Menu by ID

1. On the "home" route click "Open '#menu1' menu". Observe that the home page menu opens.
2. Close the menu and click "Go to Page Two".
3. On the "page-two" route click "Open '#menu2' menu". Observe that the page two menu opens.
4. Go back to "home".
5. Click "Open '#menu1' menu". Observe that nothing happens.
6. Click "Enable and Open '#menu1'" Menu". Observe that the home menu opens.

## Scenario 3: Using 3 or more menus even when enabling menus

1. On the "home" route click "Open 'start' menu". Observe that the home page menu opens.
2. Close the menu and click "Go to Page Two".
3. On the "page-two" route click "Open 'start' menu". Observe that the page two menu opens.
4. Close the menu and click "Go to Page Three"
5. On the "page-three" route click "Open 'start' menu". Observe that the page three menu opens.
6. Go back to "page-two".
8. Click "Open 'start' menu". Observe that nothing happens.
9. Click "Enable and Open 'start' Menu". Observe that nothing happens.

The menu controller attempts to find an enabled menu on the specified side: https://github.com/ionic-team/ionic-framework/blob/a04a11be3571faa99c751edc034462e94a977e95/core/src/utils/menu-controller/index.ts#L79C12-L79C12

Step 6 is where this breaks down. In this scenario, the menus on "home" and "page-two" are disabled. This leads menu controller to use its fallback which tries to get the first menu registered on the specified side: https://github.com/ionic-team/ionic-framework/blob/a04a11be3571faa99c751edc034462e94a977e95/core/src/utils/menu-controller/index.ts#L86

This means that the menu controller would attempt to open the "home" menu even though the user is on "page-two" (because the start menu on "home" was the first to be registered).

## What is the new behavior?

- Menus are no longer automatically disabled when a new menu on the same side is registered
- Referencing menus by side when multiple menus with that side exist in the DOM will cause a warning to be logged

This change has a couple implications:

1. Developers no longer need to manually enable a menu as noted in https://ionicframework.com/docs/api/menu#multiple-menus. Note that continuing to manually enable the menus will not cause any adverse side effects and will effectively be a no-op.
2. Developers using the menuController to open a menu based on "side" may end up having the wrong menu get opened. 

Example before to this change:

1. Start on PageA with a `start` menu. Calling `menuController.open('start')` opens the menu on PageA.
2. Go to PageB with a `start` menu. Calling `menuController.open('start')` opens the menu on PageB because the menu on PageA is disabled.

Example after to this change:

1. Start on PageA with a `start` menu. Calling `menuController.open('start')` opens the menu on PageA.
2. Go to PageB with a `start` menu. Calling `menuController.open('start')` attempts to opens the menu on PageA because both menus are enabled. However, since PageA is hidden nothing will appear to happen.

- [ ] Yes
- [x] No

## Other information

I manually verified that removing the Angular Universal code does not regress the behavior fixed in https://github.com/ionic-team/ionic-framework/pull/27814. The menu is never automatically disabled, so the bug does not happen.

This is a partial fix for https://github.com/ionic-team/ionic-framework/issues/18683. Properly fixing this requires another change which is out of scope for this work.


core/src/components/menu/test/multiple/menu.e2e.ts
```diff
@@ -0,0 +1,74 @@
+import { expect } from '@playwright/test';
+import { configs, test } from '@utils/test/playwright';
+
+/**
+ * This behavior does not vary across modes/directions
+ */
+configs({ modes: ['ios'], directions: ['ltr'] }).forEach(({ title, config }) => {
+  test.describe(title('menu: multiple'), () => {
```
thetaPC, line 8: Please include an annotation for the issue. Not sure if it can go on the `describe` level or it should be inside the `beforeEach`.

thetaPC (changes requested):
One minor thing but should help with preventing tests from being removed.

core/src/components/menu/test/multiple/menu.e2e.ts
liamdebeasi: Thanks for the reminder! Added in https://github.com/ionic-team/ionic-framework/pull/28269/commits/55704489ea3ac57fa099d52d43f71e33f81b797c and https://github.com/ionic-team/ionic-framework/pull/28269/commits/90425eb61f8f3e916b977cb75d9037b468b6af28

thetaPC (approved):
LGTM

core/src/utils/menu-controller/index.ts
```diff
@@ -76,14 +78,31 @@ const createMenuController = () => {
...
       // there could be more than one menu on the same side
       // so first try to get the enabled one
-      const menuRef = find((m) => m.side === menu && !m.disabled);
-      if (menuRef) {
-        return menuRef;
+      const menuRefs = menus.filter((m) => m.side === menu && !m.disabled);
+      if (menuRefs.length >= 1) {
+        if (menuRefs.length > 1) {
+          printIonWarning(
+            `menuController queried for a menu on the "${menu}" side, but ${menuRefs.length} menus were found. The first menu reference will be used. If this is not the behavior you want then pass the ID of the menu instead of its side.`,
```
amandaejohnston, line 85: Optional: When I try setting the first `ion-menu-button` in the new `multiple` test to `menu="start"`, I get five copies of this warning in the console immediately on loading/refreshing the page. Would it be easy enough to only log the warning when the menu is actually opened? (If this isn't trivial, don't worry about it.)

core/src/utils/menu-controller/index.ts
amandaejohnston, line 1: Do we have the work tracked to update the docs in light of this change, particularly the Multiple Menus section?

liamdebeasi: Thanks for the reminder! I meant to do this as part of my work, but I forgot. Here's the PR: https://github.com/ionic-team/ionic-docs/pull/3186

core/src/utils/menu-controller/index.ts
liamdebeasi: Good call! I updated this in https://github.com/ionic-team/ionic-framework/pull/28269/commits/0787dc8e61948895ac4570d00b06e052c0739fb8. This only warns when opening a menu.

amandaejohnston: I'm no longer getting the warnings at all, even when opening the menu. I also noticed that with the two menu buttons set to `menu="start"` and `menu="secondary-menu"` respectively, both open the secondary menu, which doesn't seem right; shouldn't it be the primary one since that's first? I think I remember the `menu="start"` button opening the primary one before 🤔

liamdebeasi: The original intent of this was to only warn when users are explicitly calling `menuController.open('side')`, but I can update it so it gets called when `toggle` is called on the menuController (what the menu button does when you click it)

liamdebeasi: Fixed in https://github.com/ionic-team/ionic-framework/pull/28269/commits/01f4a272e2b93ce4fb92f14e557c7c85e98c6d8d

amandaejohnston (approved):

PR #28273, opened 2023-10-03
refactor(router-outlet): reuse lock controller

mapsandapps:
Issue number: Internal

---------

Code is duplicated between the router outlet and the overlays

## What is the new behavior?

- Code is reused between the router outlet and the overlays

- [ ] Yes
- [x] No


liamdebeasi (approved):
Thanks!

brandyscarney (approved):

PR #28277, opened 2023-10-03
fix(header): collapsible large title main header does not flicker on load

### LINKED ISSUE #27060: bug: Collapsible Large Titles Flicker
### Prerequisites

### Ionic Framework Version

v7.x

### Current Behavior

With version 7, the collapsed title flashes upon navigation

Does not happen with version 6

https://user-images.githubusercontent.com/15254340/228694868-c1a59b02-84a3-41eb-93ea-cae83d9294f7.mov

Please see video ^

### Expected Behavior

Would not flash, as in version 6

### Steps to Reproduce

upgrade to version 7

see flash on collapsed title

### Code Reproduction URL

_No response_

### Ionic Info

```
Ionic:

   Ionic CLI : 6.20.8

Utility:

   native-run  : 1.7.2

System:

   NodeJS : v18.15.0
   npm    : 9.5.0
   OS     : macOS
```

### Additional Information

_No response_

liamdebeasi:
Issue number: resolves #27060

---------

The main header is controlled by the header with `collapse="condense"` set: https://github.com/ionic-team/ionic-framework/blob/a04a11be3571faa99c751edc034462e94a977e95/core/src/components/header/header.tsx#L144

The collapse header will hide the main header and then show it once the user has scrolled enough. However, if the main header is rendered before the collapse header is rendered, then the main header will be visible for a brief moment before being hidden by the collapse header. This gives the perception of flicker that is reported on the linked issue.

## What is the new behavior?

- The main header will be hidden on load if it loads before the collapse header

The selector was written in a way such that once the collapse header loads, this CSS no longer applies (since the collapse header will add `.header-collapse-main` to the main header)

| `main` | branch |
| - | - |
| <video src="https://github.com/ionic-team/ionic-framework/assets/2721089/3cb11a57-e084-435a-89c2-e1c2afba04b1"></video> | <video src="https://github.com/ionic-team/ionic-framework/assets/2721089/c5caeb5e-3b33-4598-986f-bf097c46251c"></video> |

- [ ] Yes
- [x] No

## Other information

Note: `:has` browser compat is still fairly new. However, it is available on both Chromium and WebKit browsers (and has been for at least a year): https://caniuse.com/?search=%3Ahas

Given that this bug is a fairly minor UI glitch (as opposed to something that would cause an app to crash or otherwise malfunction), I think this is an acceptable tradeoff. As time goes on this will become less of a concern as more users update their devices.

Dev build: `7.4.3-dev.11696365694.156f41d3`

### DIFF
--- a/core/src/components/header/header.ios.scss
+++ b/core/src/components/header/header.ios.scss
@@ -122,6 +122,11 @@
  * header will have the ".header-collapse-main" class, and
  * this selector will no longer apply.
  *
+ * The :has(...) part of the selector ensures a couple things:
+ * 1. This will only apply within a page view since the content
+ * must be a subsequent-sibling of the header (~ ion-content).
+ * 2. This will only apply when that content has a collapse header (ion-header[collapse="condense"])
+ *
  * We use opacity: 0 to avoid a layout shift.
  */
 ion-header:not(.header-collapse-main):has(~ ion-content ion-header[collapse="condense"]) {
--- a/core/src/components/header/header.ios.scss
+++ b/core/src/components/header/header.ios.scss
@@ -107,3 +107,21 @@
 .header-collapse-condense-inactive.header-collapse-condense ion-toolbar.in-toolbar ion-buttons.buttons-collapse {
   visibility: hidden;
 }
+
+/**
+ * The main header is only hidden once the collapsible large
+ * title is configured. As a result, if the main header loads
+ * before the collapsible large title is configured then the
+ * main header will be visible briefly before being hidden
+ * by the collapsible large title.
+ *
+ * The following selector ensures that any main header
+ * on a page with a collapsible large title is hidden
+ * before the collapsible large title is configured.
+ * Once the collapsible large title is configured the main
+ * header will have the ".header-collapse-main" class, and
+ * this selector will no longer apply.
+ */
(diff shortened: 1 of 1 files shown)


thetaPC:
I'm still seeing it with the dev build in one of the community's [repo](https://github.com/seanaguinaga/large-title-repro/tree/main). The flickering is happening on the [name](https://github.com/seanaguinaga/large-title-repro/blob/main/src/pages/ViewMessage.tsx).


https://github.com/ionic-team/ionic-framework/assets/13530427/3769be58-e467-4950-9cb8-adda6789faa4

liamdebeasi:
That doesn't look like the same issue. This PR addresses an issue when the main header loads before the collapsing header, but in your demo the collapsing header is already loaded. I didn't look too much into why it was happening, but I did verify that the behavior is not being caused by my dev build.

thetaPC:
> That doesn't look like the same issue. This PR addresses an issue when the main header loads before the collapsing header, but in your demo the collapsing header is already loaded. I didn't look too much into why it was happening, but I did verify that the behavior is not being caused by my dev build.

Do we have a ticket tracking this other issue?

liamdebeasi:
No, but I'm also not confident this is an Ionic bug. I wasn't able to reproduce on my end in an Ionic starter app.

thetaPC (approved):
LGTM

PR #28279, opened 2023-10-03
fix(alert): stop Enter keypress for checkboxes

thetaPC:
Issue number: internal

---------

Checkboxes inside the `ion-alert` can toggle state with the `Enter` key. This does not follow [keyboard navigation specs](https://www.w3.org/WAI/ARIA/apg/patterns/checkbox/examples/checkbox/#kbd_label).

## What is the new behavior?

- Only the `Space` key toggles states on the checkboxes.
- Nothing happens when pressing the `Enter` key.

- [ ] Yes
- [x] No

## Other information

N/A


brandyscarney (approved):
Works well!

PR #28289, opened 2023-10-05
fix(label): remove overflow hidden

brandyscarney:
created to fix outstanding issues


brandyscarney:
This branch was created to test out fixes before I made them on https://github.com/ionic-team/ionic-framework/pull/28146

Issue #27748, opened 2023-07-05
bug: interactions not allowed on app until page transition animation has completed


aeharding:
### Prerequisites

- [X] I have read the [Contributing Guidelines](https://github.com/ionic-team/ionic-framework/blob/main/.github/CONTRIBUTING.md#creating-an-issue).
- [X] I agree to follow the [Code of Conduct](https://ionicframework.com/code-of-conduct).
- [X] I have searched for [existing issues](https://github.com/ionic-team/ionic-framework/issues) that already report this problem, without success.

### Ionic Framework Version

v7.x

### Current Behavior

You cannot scroll down a page during a page transition

### Expected Behavior

Page should accept interactions before page transition has completed

### Steps to Reproduce

1. Tap into a page or swipe back to trigger the page transition animation
2. Immediately try to scroll down before the page has finished transitioning

### Ionic Info

```
Ionic:

   Ionic CLI : 7.1.1

Utility:

   cordova-res : not installed globally
   native-run  : not installed globally

System:

   NodeJS : v18.15.0
   npm    : 9.5.0
   OS     : macOS Unknown
```

### Additional Information

This bug is made a bit more serious by https://bugs.webkit.org/show_bug.cgi?id=222654

You can experience this bug in production at https://wefwef.app

Please see a demo video (with narration) explaining the video in real time:

https://github.com/ionic-team/ionic-framework/assets/2166114/d5e489a0-fe66-4cd6-b0eb-ae4ed7ed88c2

And here's a video (without narration) demonstrating how long it takes for `pointer-events: none` to be removed from both pages by slowing down the animation to 10% real time:

https://github.com/ionic-team/ionic-framework/assets/2166114/61bba0b9-df03-44db-bc1f-93077b3d1359

Related bug: https://github.com/aeharding/wefwef/issues/18

liamdebeasi:
Thanks for the report. Can you please clarify what the expected behavior is here? Scrolling should be blocked until the transition is complete. This behavior aligns with how native iOS apps behave.

aeharding:
Hi there! Expected behavior is two things:

 1. Scroll gestures before the animation is completed shouldn't latch to the document root until you wait a few seconds without interaction (see my video demonstration). iOS native apps like Settings this does not happen. I believe that is due to https://bugs.webkit.org/show_bug.cgi?id=222654, however, so that is not the focus of this bug report. (However, it does make this bug report much more noticeable.)
[...]

PR #28302, opened 2023-10-08
Fix transition speed on swipe back

aeharding:
Issue number: #27748

---------

Swipe back can take too long. On the iOS settings app, swipe back is always shorter than tapping to go to a new page (even if you swipe back slowly). This is accomplished without making it visually seem "faster" by using a linear transition function for the back swipe.

A slow swipe back transition can make the app see laggy and slow when swiping back and then scrolling down in the page you swiped back to.

## What is the new behavior?

- Swipe back function is linear
- Maximum of 200ms to complete swipe back animation

- [ ] Yes
- [X] No

## Other information

This patch has been in use in [Voyager](https://github.com/aeharding/voyager) for the past couple months with great results.

### DIFF
--- a/core/src/components/router-outlet/router-outlet.tsx
+++ b/core/src/components/router-outlet/router-outlet.tsx
@@ -111,7 +111,9 @@ export class RouterOutlet implements ComponentInterface, NavOutlet {
             this.ani.easing('cubic-bezier(1, 0, 0.68, 0.28)');
             newStepValue += getTimeGivenProgression([0, 0], [1, 0], [0.68, 0.28], [1, 1], step)[0];
           } else {
-            newStepValue += getTimeGivenProgression([0, 0], [0.32, 0.72], [0, 1], [1, 1], step)[0];
+            this.ani.easing('linear');
+
+            newStepValue += step;
           }
 
           this.ani.progressEnd(shouldComplete ? 1 : 0, newStepValue, dur);
--- a/core/src/utils/gesture/swipe-back.ts
+++ b/core/src/utils/gesture/swipe-back.ts
@@ -70,7 +70,7 @@ export const createSwipeBackGesture = (
     let realDur = 0;
     if (missingDistance > 5) {
       const dur = missingDistance / Math.abs(velocity);
-      realDur = Math.min(dur, 540);
+      realDur = Math.min(dur, 200);
     }
 
     onEndHandler(shouldComplete, stepValue <= 0 ? 0.01 : clamp(0, stepValue, 0.9999), realDur);


core/src/components/router-outlet/router-outlet.tsx
```diff
@@ -111,7 +111,9 @@ export class RouterOutlet implements ComponentInterface, NavOutlet {
             this.ani.easing('cubic-bezier(1, 0, 0.68, 0.28)');
             newStepValue += getTimeGivenProgression([0, 0], [1, 0], [0.68, 0.28], [1, 1], step)[0];
           } else {
-            newStepValue += getTimeGivenProgression([0, 0], [0.32, 0.72], [0, 1], [1, 1], step)[0];
+            this.ani.easing('linear');
```
liamdebeasi, line 114: Can you clarify why you made this change? Swiping back does use a non-linear easing curve

aeharding: Sure! So when playing with the iOS Settings app, the transition easing function on swipe back appears to be linear, or at least much, much closer to linear than the open-a-new-page animation (which is more an ease-out function).

In my testing, if you shorten the duration from 540 to 200 without changing the easing function (from ease-out-ish to linear), the animation "appears" too fast relative to iOS Settings. By changing to linear, similar to what Apple uses for swipe back animation, the swipe back animation finishes faster (which fixes the problem in #27748) but critically doesn't "appear" much faster (because of the linear easing function).

This is all me trying to match what Apple is doing for swipe back - which from my trial an error testing is:

-  linear
- max of 200ms to complete.

aeharding: @liamdebeasi sound good! I agree, on closer inspection swipe back does appear to still be nonlinear. However, the swipe back animation does appear (to me at least) to be much closer to linear than the new page animation (as well as shorter duration).

https://github.com/ionic-team/ionic-framework/assets/2166114/266e9344-8d6e-476e-a73f-d3cd342413d5

liamdebeasi: Looks like keeping the curve + changing the max duration to 300ms (instead of 200ms) achieves a very similar result:

| native | ionic |
| - | - |
| <video src="https://github.com/ionic-team/ionic-framework/assets/2721089/30e37dbe-e711-43a3-96ac-1290a3a7270b"></video> | <video src="https://github.com/ionic-team/ionic-framework/assets/2721089/77161fc6-ba26-4f8d-a74e-58266c60fbd7"></video> |

liamdebeasi (changes requested):

core/src/components/router-outlet/router-outlet.tsx
aeharding: @liamdebeasi, can you try comparing a swipe back with the slowest momentum possible to trigger the swipe? That's when I find the timing function differences most apparent. I am happy to take a look this weekend too and whip up some comparison videos - but I figure you already have your environment setup so :)

aeharding: @liamdebeasi I believe slow swipes is the main problem, because it takes 540ms.

To your point on the slow swipe looking way too fast, that is where the change to the timing function that I originally proposed helps out. By changing the timing function to linear and max time to 200ms, it "appears" slower to the eye (and very similar to iOS, in slow swipe situation), but is actually faster at 200ms.

PR #28308, opened 2023-10-09
chore(repo): parameterize stencil nightly

rwaskiewicz:
Issue number: N/A

---------

when the stencil team wants to test a dev build of stencil against the framework team's ci, they  push a branch to this repo that overrides the stencil nightly job's npm tag (replacing 'nightly' with the dev build version) and manually kick off the nightly workflow. this commit would eliminate the need for that first step there

## What is the new behavior?

update the stencil nightly job to accept any npm release tag, rather than always default to 'nightly'. doing so allows the stencil team to reuse this workflow for cases where we'd like to test a _dev_ build of stencil by running it through the framework's ci process, without landing the feature in stencil first.

I was able to test that `nightly` gets set by default (and the field is required) by running the workflow from this branch.

I then also tested this against a dev build of Stencil:
Screenshot 2023-10-09 at 9 45 41 AM

Interestingly enough, it helped me catch something to consider if we were to accept the PR this dev build is accepted!

- [ ] Yes
- [x] No

## Other information

I considered making a separate workflow for this (rather than override/use nightly) - it didn't quite seem worth the maintenance effort 🤷

### DIFF
--- a/.github/workflows/stencil-nightly.yml
+++ b/.github/workflows/stencil-nightly.yml
@@ -8,7 +8,12 @@ on:
     # at 6:00 UTC (6:00 am UTC)
     - cron: '00 06 * * 1-5'
   workflow_dispatch:
-    # allows for manual invocations in the GitHub UI
+    inputs:
+      npm_release_tag:
+        required: true
+        type: string
+        description: What version should be pulled from NPM?
+        default: nightly
 
 # When pushing a new commit we should
 # cancel the previous test run to not
@@ -24,7 +29,7 @@ jobs:
       - uses: actions/checkout@v3
       - uses: ./.github/workflows/actions/build-core-stencil-prerelease
         with:
-          stencil-version: nightly
+          stencil-version: ${{ inputs.npm_release_tag || 'nightly' }}
 
   test-core-clean-build:
     needs: [build-core-with-stencil-nightly]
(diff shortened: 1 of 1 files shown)


.github/workflows/stencil-nightly.yml
```diff
@@ -24,7 +29,7 @@ jobs:
       - uses: actions/checkout@v3
       - uses: ./.github/workflows/actions/build-core-stencil-prerelease
         with:
-          stencil-version: nightly
+          stencil-version: ${{ inputs.npm_release_tag || 'nightly' }}
```
rwaskiewicz, line 32: We need these conditionals to default to 'nightly' for the cron job, since there's no 'input' to those

liamdebeasi: Wouldn't `npm_release_tag` default to "nightly" because of how you configured the input?

rwaskiewicz: I _think_ it wouldn't, since that input is tied to the `workflow_dispatch` event (and aren't shared to the cron).

rwaskiewicz: That said, we can back this out and see what happens tonight 😆

liamdebeasi: Ahh no you're right -- cron dispatches don't have access to inputs according to https://dev.to/mrmike/github-action-handling-input-default-value-5f2g

liamdebeasi (approved):

PR #28310, opened 2023-10-09
fix(footer, tab-bar): do not hide when webview resizing disabled

### LINKED ISSUE #28226: bug: Inconsistencies between Ionic and Native iOS 17
### Prerequisites

### Ionic Framework Version

v7.x

### Current Behavior

- [ ] Back button does not overlap page title (pretty sure this is an iOS 16 regression)

https://github.com/ionic-team/ionic-framework/assets/5757359/823092b9-bd62-46a4-9d05-d5bd3e2dfcba

- [ ] Keyboard resizes the layout with a significant delay instead of animating the layout size like on chat apps

https://github.com/ionic-team/ionic-framework/assets/5757359/3a16ea4b-d794-478e-918a-a2a080ed621f

- [ ] Page slide-to-go-back is very laggy, not instant like native (can't really add a video)
- [ ] Main tab is not highlighted when in the root path https://github.com/ionic-team/ionic-framework/issues/16949
- [ ] When rotated 90 degrees, the tabs text don't go horizontal https://github.com/ionic-team/ionic-framework/issues/27429

Quick demo video:

### Expected Behavior

The user should not feel like they are using a non-native inferior application. Despite building with web technologies, there should be a near-native experience. 

### Steps to Reproduce

Open various ionic templates on the device (ideally known apps, like whatsapp or tinder etc) and compare the ionic experience to the native one.

### Code Reproduction URL

_No response_

### Ionic Info

```
Ionic:

   Ionic CLI : 7.1.1

Utility:

   cordova-res : 0.15.4
   native-run  : 1.7.3

System:

   NodeJS : v18.15.0
   npm    : 9.6.7
   OS     : macOS Unknown
```

### Additional Information

I would like to release my app soon, and before I do that, I am doing some usability testing.
but the app does not feel like a native iOS 17 application.

liamdebeasi:
Issue number: resolves #28226

---------

In https://github.com/ionic-team/ionic-framework/pull/27417, we introduced the keyboard controller which allows the footer and tab bar to hide when the keyboard is opened. This prevents those elements from sitting on top of the keyboard.

However, we did not account for when webview resizing is disabled. In that scenario, these elements do not need to hide since they will never sit on top of the keyboard. Since these elements are hidden, the size of the content increases which causes a layout shift whenever the keyboard opens.

| main | branch |
| - | - |
| <video src="https://github.com/ionic-team/ionic-framework/assets/2721089/3e12d29d-756e-446c-92d0-41c2640c75e1"></video> | <video src="https://github.com/ionic-team/ionic-framework/assets/2721089/77f3cdfd-c844-43d5-ba6e-1884b61ff031"></video> |

## What is the new behavior?

- The tab bar and footer elements do not hide when webview resizing is set to "NONE".

- [ ] Yes
- [x] No

### DIFF
--- a/core/src/components/footer/footer.tsx
+++ b/core/src/components/footer/footer.tsx
@@ -3,6 +3,7 @@ import { Component, Element, Host, Prop, State, h } from '@stencil/core';
 import { findIonContent, getScrollElement, printIonContentErrorMsg } from '@utils/content';
 import type { KeyboardController } from '@utils/keyboard/keyboard-controller';
 import { createKeyboardController } from '@utils/keyboard/keyboard-controller';
+import { Keyboard, KeyboardResize } from '@utils/native/keyboard';
 
 import { getIonMode } from '../../global/ionic-global';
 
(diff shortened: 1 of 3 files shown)


core/src/components/footer/footer.tsx
```diff
@@ -52,18 +53,28 @@ export class Footer implements ComponentInterface {
...
 
-      this.keyboardVisible = keyboardOpen; // trigger re-render by updating state
-    });
+    /**
+     * If the resize mode is set to None then we don't want to
+     * hide the tab bar here since it will never sit on top
+     * of the keyboard. Hiding the tab bar will cause a layout shift
+     * in apps that have resize set to None.
+     */
+    if (resizeMode === undefined || resizeMode.mode !== KeyboardResize.None) {
```
liamdebeasi, line 64: I opted to add these checks in each component rather than integrating the logic into `createKeyboardController`. The purpose of this util is to notify subscribers of when the keyboard is open, so it seemed odd to disable that behavior based on the webview resize preference.

AmitMY, line 64: I believe that `getResizeMode` and this check should happen inside the controller, because the resize mode can change on-the-fly.
In my case, the resize mode is changed once the input is shown, and it is possible that this `connectedCallback` happens before that change.
In general, I believe it would be more robust to perform this check afterward.

liamdebeasi: Good point! I'll push a fix to address this.

AmitMY:
@liamdebeasi thanks again for this - any ETA for when this ships?

hedinasr:
any news on this PR ?

PR #28314, opened 2023-10-09
feat(a11y): add dynamic font scaling

### LINKED ISSUE #18592: feat: dynamic type text sizes not supported
# Bug Report

**Ionic version:**

4.11.0

**Current behavior:**

On iOS, ionic apps ignore the preferred font size that the user has set in the system preferences. (Apple calls this Dynamic Type Sizes https://developer.apple.com/design/human-interface-guidelines/ios/visual-design/typography/). Irrespective of what you set there, the text inside Ionic apps is always the same size.

(For comparison, when you build for Android, the Ionic app uses the font size the user has set in system preferences. https://material.io/design/usability/accessibility.html#layout-typography)

**Expected behavior:**

Ionic apps by default use the system-defined font size on all platforms.

**Steps to reproduce:**

On iOS:

1. Go to Phone settings -> Display & Brightness -> Text Size, and change the system font size. (Note that the text on that page resizes immediately.)
2. Run any Ionic app (e.g. the test app mentioned below, or the Moodle mobile app https://moodle.com/app/)
3. Note the size of the text.
4. Now go back to the phone settings and change to font size to something different.
5. Go back to the Ionic app, and note that the text size has not changed.

For comparison, you can do the same thing on an Android device, and the Ionic app will change font size, and the font size will match the system text, e.g. on the settings screen.

Note that this is important for accessibility.

**Related code:**

I made the simplest possible Ionic/Cordova app to test this:
https://github.com/timhunt/dynamicTypeTest

When run, it just displays this screen: https://github.com/timhunt/dynamicTypeTest/blob/master/src/pages/home/home.html which has this associated JS https://github.com/timhunt/dynamicTypeTest/blob/master/src/pages/home/home.ts#L14 to display a simple dialogue if you click the button. All this just uses the standard Ionic SCSS to style things. I have not added any SCSS of my own.

The https://github.com/timhunt/dynamicTypeTest/tree/master/screengrabs folder has some screen-grabs of the app running (in an emulator, but that does not matter) with different settings for the system preferred font size. For comparison, there are also screen-grabs of a page from an Apple / Google app. This shows that on Android does a good job of picking up the system font setting, but on iOS it completely ignores the system font setting. (The Ionic default styling seems to be a good match for the smallest system font size.)

**Other information:**

I have done my best to research this before raising an issue (and I was not sure if this was a bug report or a feature request. I'm sorry if I got this wrong.) The thing that surprises me most is how little discussion I can find about this. I feel sure that people must have thought about this before, but I have not found much. Pretty much the only thing I found was https://forum.ionicframework.com/t/ionic-get-the-mobile-os-font-size-in-app/101498/3 and the phonegap plugin it links to. I am surprised behaviour is so different between iOS and Android.

Thanks in advance for any information you can give up. Since we want this fixed, then given some suitable guidance, we might have resources to work on a pull request. (However, I have not done any development on the Ionic framework code before.)

Cross reference: here is the bug report on the Moodle Mobile app that lead me to start investigating this: https://tracker.moodle.org/browse/MOBILE-2097.)

**Ionic info:**

Ionic:

   ionic (Ionic CLI)  : 4.11.0 (C:\Users\[DELETED]\AppData\Roaming\npm\node_modules\ionic)
   Ionic Framework    : ionic-angular 3.9.5
   @ionic/app-scripts : 3.2.2

Cordova:

   cordova (Cordova CLI) : 9.0.0
   Cordova Platforms     : android 8.0.0, ios 5.0.1
   Cordova Plugins       : cordova-plugin-ionic-keyboard 2.1.3, cordova-plugin-ionic-webview 4.1.0, (and 4 other plugins)

System:

   NodeJS : v8.14.1 (C:\Program Files\nodejs\node.exe)
   npm    : 6.4.1
   OS     : Windows 7

Environment:

   HTTP_PROXY  : [DELETED]
   HTTPS_PROXY : [DELETED]

### LINKED ISSUE #24638: feat: global font size adjustment
### Prerequisites

Currently, there isn't a way to increase or decrease the font size by the same amount across the whole project. You can adjust the sizes on individual components, but this is annoying since you have to manually set the size for _every_ component type used.

### Describe the Use Case

My particular use case is a font that's just naturally small and hard to read at the default size ([Teko](https://fonts.google.com/specimen/Teko)), so I'd like to bump the font size across the whole app up by 5px or so. This feature also has accessibility applications, though -- you could add a size option to an app without causing non-typographic elements to scale as well.

I can think of a few possible fixes:
1. New CSS variables for small/med/large font sizes that everything either uses directly, or is set relative to via `calc` adjustments. Googling around, it seems like there used to be SCSS variables like this in Ionic 4, but they've been removed.
2. One new CSS variable, something like `--font-size-adjustment`, that gets added to all sizes. This could be automated with a mixin, similar to `padding` handling RTL.
3. Update all font sizes to use `em` or `rem` units, so a global base size could be set on the root.

### Describe Alternatives

As far as I can tell, the only alternatives are to either pick a different font, or manually update the font size for all components.

### Related Code

_No response_

### Additional Information

_No response_

liamdebeasi:
Issue number: resolves #24638, resolves #18592

---------

Developers have requested that Ionic Framework support the dynamic type feature on iOS for accessibility purposes. Ionic applications do not respond to font scaling on iOS which can create inaccessible applications particularly for users with low vision. Ionic apps on Android devices currently support the Android equivalent due to functionality in the Chromium webview.

Developers have also requested a way of adjusting the fonts in their Ionic UI components consistently.

## What is the new behavior?

- Ionic components now use `rem` instead of `px` where appropriate. This means devs can change the font size on `html` and the text in supported Ionic components will scale up/down appropriately
- Add support for Dynamic Type on iOS (the iOS version of Dynamic Font Scaling)

- [ ] Yes
- [x] No


thetaPC (commented):
LGTM

sean-perkins (approved):

matfantinel:
This was a breaking change to the app I work on. Suddenly all Ionic controls got too small and tweaking the font-size in the `html` tag made my other styles (that use rem) too big. Definitely not expected from a minor version bump. Not cool 😔

Issue #28186, opened 2023-09-16
bug: Ionic React lifecycle hooks should execute cleanup function when unmounted 


rabianr:
### Prerequisites

- [X] I have read the [Contributing Guidelines](https://github.com/ionic-team/ionic-framework/blob/main/.github/CONTRIBUTING.md#creating-an-issue).
- [X] I agree to follow the [Code of Conduct](https://ionicframework.com/code-of-conduct).
- [X] I have searched for [existing issues](https://github.com/ionic-team/ionic-framework/issues) that already include this feature request, without success.

### Describe the Feature Request

Like ```useEffect``` hook, ```useIonViewWillEnter``` and ```useIonViewDidEnter``` hook should also support cleanup function.

### Describe the Use Case

In React 18, ```useIonViewWillEnter``` and ```useIonViewDidEnter``` run twice in development. Which makes any fetch requests run twice too.
If I make a fetch request in ```useEffect```, I can pass an abort signal to the request and abort it in the cleanup function.
I tried to do the same in ```useIonViewWillEnter``` hook, but the cleanup function isn't supported, so there's no way to abort the request.

I also tried to use the ```useIonViewWillLeave``` hook, but this hook won't get fired until you navigate to another page.

**Code Reproduction**
https://stackblitz.com/edit/qhjlkw

sean-perkins:
Hello, thanks for reporting this issue!

Can you test with this dev-build and let me know if you experience any unexpected behaviors?

```
npm install @ionic/react@7.4.4-dev.11696895407.1bd7a380 @ionic/react-router@7.4.4-dev.11696895407.1bd7a380
```

rabianr:
> Hello, thanks for reporting this issue!
> 
> Can you test with this dev-build and let me know if you experience any unexpected behaviors?
> 
> ```
> npm install @ionic/react@7.4.4-dev.11696895407.1bd7a380 @ionic/react-router@7.4.4-dev.11696895407.1bd7a380
> ```

Hi, I can confirm that with that dev-build, the cleanup function get executed properly in my production project. Also, there is no unexpected behaviors so far.
Code Reproduction: https://stackblitz.com/edit/qhjlkw-4aknkp?file=package.json,src%2Fmain.tsx
Thank you for your support.

sean-perkins:
@rabianr thanks for testing! I'll get a PR open shortly for the changes. 

I have an updated dev-build if you want to continue testing with it: 
```
[...]

PR #28316, opened 2023-10-09
fix(react): lifecycle events are removed on page unmount

sean-perkins:
Issue number: N/A

---------

While debugging #28186, Maria and I identified that Ionic's lifecycle event listeners (`ionViewWillEnter`, etc.) were being registered multiple times on the same `.ion-page` element. This resulted in problematic behavior, where a user's implementation of our lifecycle hooks, would execute their callback multiple times. 

```ts
useIonViewWillEnter(() => {
  // This is called 2x for every time the `ionViewWillEnter` event is emitted (in React 18, dev mode)
  console.log('hello world'); 
});
```

When the Ionic lifecycle event listeners are registered in React, we bind the scope of the class to the callback function. When removing the event listener we additional use the `.bind` syntax. 

```tsx
componentDidMount() {
  element.addEventListener('ionViewWillEnter', this.ionViewWillEnter.bind(this));
} 

componentWillUnmount() {
  // This creates a new instance of the function to remove! It doesn't remove the original event listener.
  element.removeEventListener('ionViewWillEnter', this.ionViewWillEnter.bind(this)); 
}
```

The `.bind` method returns a new instance of the function. This means in the implementation we are creating a new instance of the function when both adding and removing the event listener - resulting in the `removeEventListener` to never remove the original event listener. 

This behavior only occurred in React 18 in dev mode, as a result of the mount/unmount behavior running 2x for `useEffect` hooks. 

## What is the new behavior?

- Ionic lifecycle event listeners are removed from element references when they are unmounted.
- User's lifecycle callback methods are only invoked once per event emission.

|Before|After|
|----|----|
|CleanShot 2023-10-09 at 18 32 08@2x|CleanShot 2023-10-09 at 18 29 37@2x|

- [ ] Yes
- [x] No

### DIFF
--- a/packages/react/src/routing/OutletPageManager.tsx
+++ b/packages/react/src/routing/OutletPageManager.tsx
@@ -28,8 +28,8 @@ export class OutletPageManager extends React.Component<OutletPageManagerProps> {
     /**
      * This binds the scope of the following methods to the class scope.
      * The `.bind` method returns a new function, so we need to assign it
-     * in the constructor to avoid creating a new function when removing the
-     * event listeners.
+     * in the constructor rather than when adding or removing the listeners
+     * to avoid creating a new function.
      */
     this.ionViewWillEnterHandler = this.ionViewWillEnterHandler.bind(this);
     this.ionViewDidEnterHandler = this.ionViewDidEnterHandler.bind(this);
--- a/packages/react/src/routing/OutletPageManager.tsx
+++ b/packages/react/src/routing/OutletPageManager.tsx
@@ -24,6 +24,17 @@ export class OutletPageManager extends React.Component<OutletPageManagerProps> {
     super(props);
 
     this.outletIsReady = false;
+
+    /**
+     * This binds the scope of the following methods to the class scope.
+     * The `.bind` method returns a new function, so we need to assign it
+     * in the constructor to avoid creating a new function when removing the
+     * event listeners.
+     */
+    this.ionViewWillEnterHandler = this.ionViewWillEnterHandler.bind(this);
+    this.ionViewDidEnterHandler = this.ionViewDidEnterHandler.bind(this);
+    this.ionViewWillLeaveHandler = this.ionViewWillLeaveHandler.bind(this);
+    this.ionViewDidLeaveHandler = this.ionViewDidLeaveHandler.bind(this);
   }
 
   componentDidMount() {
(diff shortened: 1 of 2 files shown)


thetaPC (approved):
LGTM

thetaPC:
@amandaejohnston it would be great to get your review. Since I was part of the triaging, my approval may be biased.

packages/react/src/routing/OutletPageManager.tsx
```diff
@@ -24,6 +24,17 @@ export class OutletPageManager extends React.Component<OutletPageManagerProps> {
     super(props);
 
     this.outletIsReady = false;
+
+    /**
+     * This binds the scope of the following methods to the class scope.
+     * The `.bind` method returns a new function, so we need to assign it
+     * in the constructor to avoid creating a new function when removing the
+     * event listeners.
+     */
```
amandaejohnston, line 33:
```suggestion
```
I don't think this comment adds much for people revisiting the code in the future. Binding handlers in the constructor like this is a common React pattern, but calling attention to it makes it seem like we're doing something weird which is mildly confusing. You have to know that we used to bind them when adding the listeners in `componentDidMount` (which wasn't correct anyway) for this to make sense.

packages/react/src/routing/PageManager.tsx
```diff
@@ -23,6 +23,17 @@ export class PageManager extends React.PureComponent<PageManagerProps> {
     this.ionPageElementRef = React.createRef();
     // React refs must be stable (not created inline).
     this.stableMergedRefs = mergeRefs(this.ionPageElementRef, this.props.forwardedRef);
+
+    /**
+     * This binds the scope of the following methods to the class scope.
+     * The `.bind` method returns a new function, so we need to assign it
+     * in the constructor to avoid creating a new function when removing the
+     * event listeners.
+     */
```
amandaejohnston, line 32:
```suggestion
```
Ditto to my other comment.

amandaejohnston (changes requested):

packages/react/src/routing/OutletPageManager.tsx
amandaejohnston: Maybe we add more context on what pattern we're trying to advise against, in that case? Something like this:
```suggestion
    /**
     * This binds the scope of the following methods to the class scope.
     * The `.bind` method returns a new function, so we need to assign it
     * in the constructor rather than when adding or removing the listeners
     * to avoid creating a new function.
     */
```
At the moment, it's not clear unless you have the context of this PR's changes. I still think this is a common enough React pattern that we don't need a comment at all, and that adding one is more confusing than it is helpful, but my background is in React so I acknowledge I could be too biased 😄

amandaejohnston (approved):

PR #28319, opened 2023-10-10
fix(react): cleanup functions are execute for lifecycle hooks

### LINKED ISSUE #28186: bug: Ionic React lifecycle hooks should execute cleanup function when unmounted

sean-perkins:
Issue number: Resolves #28186

---------

Ionic lifecycle hooks do not execute a cleanup function when the underlying `useEffect` is unmounted.

```ts
useEffect(() => {
  return () => {
     console.log('cleanup'); // called
   };
});

useIonViewWillEnter(() => {
  return () => {
     console.log('cleanup'); // never called
  };
});
```

Ionic's implementation registers the lifecycle callback to be handled at a later time, by the page managers. However, it does not keep a reference to the returned callback, so it cannot execute it when the `useEffect` is unmounted.  

## What is the new behavior?

- Ionic lifecycle hooks execute dev-specified cleanup functions

- [ ] Yes
- [x] No

## Other information

Dev-build: `7.4.4-dev.11696956070.1faa3cfe`

This PR builds on the changes in #28316.

### DIFF
--- a/packages/react/src/contexts/IonLifeCycleContext.tsx
+++ b/packages/react/src/contexts/IonLifeCycleContext.tsx
@@ -100,17 +100,6 @@ export const DefaultIonLifeCycleContext = class implements IonLifeCycleContextIn
       });
       // Remove all matching items from the array
       callbacks = callbacks.filter((x) => x.id !== callback.id);
-    } else {
-      /**
-       * If the destructor isn't registered in the array, then the
-       * user-provided callback was never invoked. This is usually
-       * the result of the useEffect hook in dev mode. So we manually
-       * invoke the callback to get the destructor to tear it down.
-       */
-      const destructor = callback();
-      if (typeof destructor === 'function') {
-        destructor();
-      }
     }
   }
 
--- a/packages/react/src/contexts/IonLifeCycleContext.tsx
+++ b/packages/react/src/contexts/IonLifeCycleContext.tsx
@@ -9,6 +9,10 @@ export interface IonLifeCycleContextInterface {
   ionViewWillLeave: () => void;
   onIonViewDidLeave: (callback: () => void) => void;
   ionViewDidLeave: () => void;
+  cleanupIonViewWillEnter: (callback: () => void) => void;
+  cleanupIonViewDidEnter: (callback: () => void) => void;
+  cleanupIonViewWillLeave: (callback: () => void) => void;
+  cleanupIonViewDidLeave: (callback: () => void) => void;
 }
 
 export const IonLifeCycleContext = /*@__PURE__*/ React.createContext<IonLifeCycleContextInterface>({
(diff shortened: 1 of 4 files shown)


packages/react/src/contexts/IonLifeCycleContext.tsx
```diff
@@ -63,8 +88,75 @@ export const DefaultIonLifeCycleContext = class implements IonLifeCycleContextIn
...
+      // Remove all matching items from the array
+      callbacks = callbacks.filter((x) => x.id !== callback.id);
+    } else {
+      /**
+       * If the destructor isn't registered in the array, then the
+       * user-provided callback was never invoked. This is usually
+       * the result of the useEffect hook in dev mode. So we manually
+       * invoke the callback to get the destructor to tear it down.
+       */
+      const destructor = callback();
```
sean-perkins, line 110: I'm questioning this now. If the `useEffect` was called and the lifecycle callback was registered, but the lifecycle event never fired, I _don't_ think we should be invoking it & then tearing it down. Likely it should never call either the lifecycle callback or the destructor in this case.

sean-perkins: Verified this was not needed.

packages/react/src/contexts/IonLifeCycleContext.tsx
```diff
@@ -63,8 +88,64 @@ export const DefaultIonLifeCycleContext = class implements IonLifeCycleContextIn
...
+   */
+  cleanupIonViewDidLeave(callback: LifeCycleCallback) {
+    this.teardownCallback(callback, this.ionViewDidLeaveDestructorCallbacks);
+  }
+
   ionViewWillEnter() {
-    this.ionViewWillEnterCallbacks.forEach((cb) => cb());
+    this.ionViewWillEnterCallbacks.forEach((cb) => {
+      const destructor = cb();
+      if (cb.id) {
```
thetaPC, line 145: Wouldn't it be better to create the variable, `destructor` inside the if statement? That way there isn't an unused variable if `cb.id` doesn't exist.

sean-perkins: Yes it would!

sean-perkins: Following up on why I did this - took me a second to recall. 

The callback may not have an `id`, but we still want to invoke the callback regardless. Otherwise the developers implementation within the lifecycle hook will never fire. 

We only want to store the destructor callback if the callback has an `id` - which is assigned from the lifecycle hooks. 

This was intentional.

thetaPC (approved):
LGTM

Issue #28121, opened 2023-09-05
bug: Ion-datetime when readonly=true, should able to navigate through calendar


FranciskoNeves:
### Prerequisites

- [X] I have read the [Contributing Guidelines](https://github.com/ionic-team/ionic-framework/blob/main/.github/CONTRIBUTING.md#creating-an-issue).
- [X] I agree to follow the [Code of Conduct](https://ionicframework.com/code-of-conduct).
- [X] I have searched for [existing issues](https://github.com/ionic-team/ionic-framework/issues) that already include this feature request, without success.

### Describe the Feature Request

 When readonly=true in ion-datetime component, we should be able to navigate between months and years.

### Describe the Use Case

If I'm using ion-datetime not to select dates, but to show multiple selected dates, if one of those dates is in another month, I can't see it.

### Describe Preferred Solution

With readonly=true, the selection of dates should be disabled, but not the navigation of the calendar.

### Describe Alternatives

Another solution would be to add a new property (viewOnlyMode, for example) to the component that disable selection of dates but let the user navigate to other months.

mapsandapps:
Hi, thank you for submitting the bug report. This was fixed & released in version [7.5.3](https://github.com/ionic-team/ionic-framework/releases/tag/v7.5.3) of Ionic Framework.

PR #28336, opened 2023-10-11
fix(datetime): allow calendar navigation in readonly mode; disallow keyboard navigation when disabled

mapsandapps:
Issue number: #28121 

---------

It is not possible to navigate between months when ion-datetime is in readonly mode. This means that if there are multiple dates selected, the user cannot browse to view them all.

Also, keyboard navigation is not prevented in `readonly` or `disabled` mode where it should be.

## What is the new behavior?

When `readonly`:
- Clicking the  month-year button changes the month & year in readonly mode
- Clicking the next & prev buttons changes the month in readonly mode
- Left and right arrow keys change the month in readonly mode
- Swiping/scrolling changes the month in readonly mode
- The selected date does not change when doing any of the above
- You cannot clear the value using keyboard navigation of the clear button in readonly mode

When `disabled`:
- You cannot navigate months via keyboard navigation of the month-year button in disabled mode
- You cannot navigate months using keyboard navigation of the previous & next buttons in disabled mode
- You cannot navigate months via the left and right arrow keys in disabled mode
- The selected date does not change when doing any of the above
- You cannot clear the value using keyboard navigation of the clear button in disabled mode

Known bug:
- It is still possible to navigate through dates in `prefers-wheel` when `disabled`. This bug existed prior to this PR. I created FW-5408 to track this.

- [ ] Yes
- [x] No


core/src/components/datetime/datetime.tsx
```diff
@@ -599,6 +599,14 @@ export class Datetime implements ComponentInterface {
   };
 
   private setActiveParts = (parts: DatetimeParts, removeDate = false) => {
+    /** if the datetime component is in readonly mode,
+     * allow browsing of the calendar without changing
+     * the set value
+     */
+    if (this.readonly) {
+      return;
+    }
```
amandaejohnston, line 608: I noticed that if you tab to the Clear button and hit Enter, the value still clears. You might need to add another `return` like this elsewhere.

To be fair, this behavior exists in `main` too, and then some -- you can do pretty much anything you want with the keyboard controls, haha.

amandaejohnston (changes requested):

core/src/components/datetime/datetime.tsx
```diff
@@ -599,6 +599,14 @@ export class Datetime implements ComponentInterface {
   };
 
   private setActiveParts = (parts: DatetimeParts, removeDate = false) => {
+    /** if the datetime component is in readonly mode,
+     * allow browsing of the calendar without changing
+     * the set value
+     */
+    if (this.readonly) {
```
liamdebeasi, line 606: The month/year picker button should be disabled. Clicking it reveals the wheel picker, and I can swipe up/down to change the month/year (which changes the `value`).

We should probably also hide the month/year picker when `readonly` is set so you don't get stuck in this state.

core/src/components/datetime/datetime.tsx
liamdebeasi: We should be able to fix this by making related buttons disabled. In particular:

1. The month/year picker button
2. The cancel/clear/done buttons that amanda mentioned.
3. The time button that opens the time picker
4. The calendar day buttons

The purpose of this ticket is to make readonly behave correctly, so we should consider doing this in the PR. However, if this work drastically increases the effort required to implement this then we can do it separately since this behavior does exist in main.

liamdebeasi (changes requested):

core/src/components/datetime/datetime.tsx
mapsandapps: It is a decent chunk of work, because of all the items you pointed out, and some behaviors are the same for `disabled` and `readonly` while others differ. And several tests to write. But I'm mostly done with it now, so I think I might as well continue.

I did inquire about this in slack, since it felt unrelated to the ticket.

core/src/components/datetime/datetime.tsx
mapsandapps: i'm confused by this comment. what's being requested in this ticket is a way to navigate months in readonly mode without changing the dates. i interpreted that as being able to use both the date picker and the previous & next buttons to change the month. the other code here should be disabling the date from actually being changed when the month view changes. if you're seeing something different, could you be more specific about the reproduction? i'm not seeing the value ever get changed in readonly mode.

liamdebeasi: In `main`, users are unable to click the month/year button to open the month/year picker when the date time is in readonly mode. This is good as opening the month/year picker would allow users to change the month and year.

https://github.com/ionic-team/ionic-framework/assets/2721089/a01241dd-6586-4a84-88f8-99f84e4dda8f

Your code regresses this behavior so users can click the month/year button to open the picker. This allows users to change the month and year which we do not want.

https://github.com/ionic-team/ionic-framework/assets/2721089/999f3ff9-81ee-46c6-940b-6b6860a32b78

I believe this issue is being caused by https://github.com/ionic-team/ionic-framework/pull/28336/files/09437b5d610388bc508f45ef7aac4255e8c32ebb#diff-83857955657186aba3b158e5561a577b94c5b9b8e74aff61fc97c78f4b77343cR201 since we are allowing pointer events.

amandaejohnston: For reference, MUI allows the picker to be opened, but no new values can be selected from there: https://stackblitz.com/edit/react-wjqnat?file=Demo.js I don't think we should entirely disable this button, but we should prevent the value and activeParts from changing when it is used.

mapsandapps: I can't reproduce what you're seeing, Liam. When I change the month and/or year via the picker on my branch, the selected date does not change. 🤔

liamdebeasi: > For reference, MUI allows the picker to be opened, but no new values can be selected from there: https://stackblitz.com/edit/react-wjqnat?file=Demo.js I don't think we should entirely disable this button, but we should prevent the value and activeParts from changing when it is used.

I don't think MUI is a great example to follow here. All of the buttons are enabled, so screen readers will consider them interactive even though they really should not be.

> I can't reproduce what you're seeing, Liam. When I change the month and/or year via the picker on my branch, the selected date does not change. 🤔

This is my code on the latest `FW-5096`:

```html
<ion-datetime readonly="true" value="2022-02-22T16:30:00"></ion-datetime>
```

1. Load the page. Observe that the button for "February 22nd, 2022" is solid blue (which indicates that it is selected).
2. Click the "February 2022" button. 
3. Click "March". 
4. Close the month/year picker. Observe that button for "March 22nd, 2022" is solid blue.

The `value` property doesn't change, but the UI still updates to indicate that March 22nd is selected which is not correct. At the very least, the wheel picker should not be interactive. However, I don't think this view should be accessible at all -- it doesn't provide any additional value for the user since the current month/year is already accessible.

mapsandapps: https://github.com/ionic-team/ionic-framework/assets/14926794/4f376028-ec81-495b-a36e-07c7184fb474

🤔

mapsandapps: > However, I don't think this view should be accessible at all -- it doesn't provide any additional value for the user since the current month/year is already accessible. 

The ticket mentions a case where multiple dates are selected across different months. The purpose of this PR is to allow you to navigate across those months so you can see all the selected dates without changing them.
(I suppose ideally, we could disable navigating across months outside of the range of the selected dates, similar to if there are min & max dates for the datetime, but I didn't think about that until now.)

core/src/components/datetime/datetime.tsx
```diff
@@ -1444,18 +1452,33 @@ export class Datetime implements ComponentInterface {
             <slot name="buttons">
               <ion-buttons>
                 {showDefaultButtons && (
-                  <ion-button id="cancel-button" color={this.color} onClick={() => this.cancel(true)}>
+                  <ion-button
+                    id="cancel-button"
+                    color={this.color}
+                    onClick={() => this.cancel(true)}
+                    disabled={this.disabled || this.readonly}
```
liamdebeasi, line 1459: I think it will be easier for other developers to understand this code if we move this check to its own variable. The reason is that we use it 3 times in this function, but it's not immediately clear why we check both disabled and readonly.

You could do something like this:

```js
const isButtonDisabled = this.disabled || this.readonly;

...

disabled={isButtonDisabled}
```

From there we'd also want comments above `isButtonDisabled` as to why we need to check disabled and readonly. Feel free to choose a different variable name.

core/src/components/datetime/datetime.tsx
```diff
@@ -2011,7 +2035,11 @@ export class Datetime implements ComponentInterface {
 
           <div class="calendar-next-prev">
             <ion-buttons>
-              <ion-button aria-label="Previous month" disabled={prevMonthDisabled} onClick={() => this.prevMonth()}>
+              <ion-button
+                aria-label="Previous month"
+                disabled={prevMonthDisabled || this.disabled}
```
liamdebeasi, line 2040: Referencing `this` variables incur a slight performance cost in Stencil, so if we want to use the variable multiple times we typically destructure `this` and reference the local variable instead to avoid duplicating the perf cost.

To fix this, you can do the following:

1. Destructure `disabled` at the top of `renderCalendarHeader`: `const { disabled } = this;`
2. Replace `this.disabled` with `disabled` in the function (to reference the local var instead).

core/src/components/datetime/datetime.scss
```diff
@@ -185,13 +185,31 @@ ion-picker-column-internal {
...
+   * Allow user to navigate months
+   * while in readonly mode
+   */
+  .calendar-action-buttons,
+  .calendar-body,
+  .datetime-year {
+    pointer-events: initial;
+  }
+
+  .calendar-body .calendar-day {
```
liamdebeasi, line 210: If the calendar day button is disabled do we still need this code?

core/src/components/datetime/test/disabled/datetime.e2e.ts
```diff
@@ -0,0 +1,156 @@
+import { expect } from '@playwright/test';
+import { configs, test } from '@utils/test/playwright';
+
+/**
+ * This behavior does not differ across
+ * modes/directions.
+ */
+configs({ modes: ['ios'], directions: ['ltr'] }).forEach(({ title, config }) => {
+  test.describe(title('datetime: disabled'), () => {
+    test('should not change value when date is clicked', async ({ page }) => {
```
liamdebeasi, line 10: This test still passes even if I re-enable the button (and remove the pointer-events code). I recommend using the [toBeDisabled](https://playwright.dev/docs/api/class-locatorassertions#locator-assertions-to-be-disabled) assertion instead. With this approach we can verify that a button is disabled (and therefore cannot be activated) rather than testing the side effect of a button being disabled.

core/src/components/datetime/test/disabled/datetime.e2e.ts
```diff
@@ -0,0 +1,156 @@
...
+        <ion-datetime value="2022-02-28" disabled></ion-datetime>
+    `,
+        config
+      );
+
+      await page.waitForSelector('.datetime-ready');
+      const calendarMonthYear = page.locator('ion-datetime .calendar-month-year');
+      await expect(calendarMonthYear).toHaveText('February 2022');
+
+      await calendarMonthYear.click({ force: true });
```
liamdebeasi, line 44: You can do a `toBeDisabled` assertion here too. Note that you'll need to grab the native button instead since it only works on native control elements:

```js
expect(calendarMonthYear.locator('button')).toBeDisabled();
```

Then you could get rid of the checks below it.

core/src/components/datetime/test/disabled/datetime.e2e.ts
```diff
@@ -0,0 +1,156 @@
...
+
+      const tabKey = browserName === 'webkit' ? 'Alt+Tab' : 'Tab';
+
+      await page.waitForSelector('.datetime-ready');
+      const calendarMonthYear = page.locator('ion-datetime .calendar-month-year');
+      const monthYearButton = page.locator('.calendar-month-year ion-item');
+      await expect(calendarMonthYear).toHaveText('February 2022');
+
+      await page.keyboard.press(tabKey);
+      await expect(monthYearButton).not.toBeFocused();
```
liamdebeasi, line 67: If we use the `toBeDisabled` assertion I mentioned for the test above we could probably get rid of this test altogether.

core/src/components/datetime/test/disabled/datetime.e2e.ts
```diff
@@ -0,0 +1,156 @@
...
+      );
+
+      const ionChange = await page.spyOnEvent('ionChange');
+
+      const calendarMonthYear = page.locator('ion-datetime .calendar-month-year');
+      await expect(calendarMonthYear).toHaveText('February 2022');
+
+      const nextMonthButton = page.locator('ion-datetime .calendar-next-prev ion-button + ion-button');
+      await nextMonthButton.click({ force: true });
+      await page.waitForChanges();
```
liamdebeasi, line 107: Ditto here on using the `toBeDisabled` check:

```js
await expect(nextMonthButton.locator('button')).toBeDisabled()
```

core/src/components/datetime/test/disabled/datetime.e2e.ts
```diff
@@ -0,0 +1,156 @@
...
+        `
+        <ion-datetime value="2022-02-22T16:30:00" disabled></ion-datetime>
+    `,
+        config
+      );
+
+      const monthYearButton = page.locator('.calendar-month-year ion-item');
+
+      await page.keyboard.press(tabKey);
+      await expect(monthYearButton).not.toBeFocused();
```
liamdebeasi, line 134: If we used the `toBeDisabled` check for the `should not navigate months via month-year button` test then we can probably get rid of this test. A disabled button implies that the button is not focusable.

core/src/components/datetime/test/disabled/datetime.e2e.ts
```diff
@@ -0,0 +1,156 @@
...
+        `
+
+        <ion-datetime value="2022-02-22T16:30:00" show-default-buttons="true" show-clear-button="true" disabled></ion-datetime>
+    `,
+        config
+      );
+
+      await page.waitForSelector('.datetime-ready');
+
+      const clearButton = page.locator('#clear-button button');
```
liamdebeasi, line 148: Let's use the `toBeDisabled` assertion instead which will cut down on the amount of actions we need to take in the test.

core/src/components/datetime/datetime.tsx
```diff
@@ -1414,7 +1422,13 @@ export class Datetime implements ComponentInterface {
...
 
   private renderFooter() {
-    const { showDefaultButtons, showClearButton } = this;
+    const { disabled, readonly, showDefaultButtons, showClearButton } = this;
+    /**
+     * The cancel, clear, and confirm buttons
+     * should not be interactive if the datetime
+     * is disabled or readonly.
+     */
+    const isButtonDisabled = disabled || readonly;
```
amandaejohnston, line 1431: Now that the buttons are outright disabled, the datetime looks visually different with `readonly="true"`:

disabled buttons

This doesn't seem right? In all our other components, `readonly` doesn't alter the visuals. The prop's description also states that the datetime "appears normal but is not interactive." If disabling the buttons is the easiest way to achieve the functionality we want, one possibility is to manually reset the buttons' opacity.

Either way, the prop's description should probably also be updated, since it's now technically interactive (you just can't select new dates).

core/src/components/datetime/test/disabled/datetime.e2e.ts
```diff
@@ -0,0 +1,91 @@
...
+        <ion-datetime value="2022-02-28" disabled></ion-datetime>
+    `,
+        config
+      );
+
+      await page.waitForSelector('.datetime-ready');
+
+      const febFirstButton = page.locator(`.calendar-day[data-day='1'][data-month='2']`);
+
+      await expect(febFirstButton).toBeDisabled;
```
amandaejohnston, line 22:
```suggestion
      await expect(febFirstButton).toBeDisabled();
```

core/src/components/datetime/test/readonly/datetime.e2e.ts
```diff
@@ -0,0 +1,155 @@
...
+        <ion-datetime value="2022-02-28" readonly></ion-datetime>
+    `,
+        config
+      );
+
+      await page.waitForSelector('.datetime-ready');
+
+      const febFirstButton = page.locator(`.calendar-day[data-day='1'][data-month='2']`);
+
+      await expect(febFirstButton).toBeDisabled;
```
amandaejohnston, line 22:
```suggestion
      await expect(febFirstButton).toBeDisabled();
```

amandaejohnston (changes requested):

core/src/components/datetime/datetime.tsx
mapsandapps: I adjusted the description, but I had a tough time figuring out how to word it. I'm open to suggestions for further improvements.

core/src/components/datetime/datetime.scss
```diff
@@ -185,13 +185,36 @@ ion-picker-column-internal {
...
+    pointer-events: initial;
+  }
+
+  /**
+   * Disabled buttons should have full opacity
+   * in readonly mode
+   */
+  .calendar-day[disabled],
+  .datetime-action-buttons ion-button[disabled] {
+    opacity: 1;
```
liamdebeasi, line 216: Heads up that this does change the behavior of datetime slightly. If you have a date disabled using `isDateEnabled` it no longer appears dimmed.

Here is the behavior in `main`: https://codepen.io/liamdebeasi/pen/WNPbLXd (Feb 5 is dimmed since it is explicitly disabled)

If all the buttons are disabled do we still want to show the dimmed state for `isDateEnabled`? For screen readers it won't make much of a difference because those tools will announce every button as dimmed.

liamdebeasi (approved):
I'm fine with the changes as-is, but I would like to make sure one of the changes was intentional.

core/src/components/datetime/datetime.scss
amandaejohnston: I think we should still show it if possible, I can definitely see someone reporting this as a bug. We might need a class for when a date is disabled due to `isDateEnabled`, or something along those lines 🤔

core/src/components/datetime/datetime.tsx
amandaejohnston: Maybe "the datetime appears normal but the selected date cannot be changed"?

core/src/components/datetime/datetime.scss
mapsandapps: cool, i can make that change! (and add a screenshot test). i'm not really sure what sort of situations people are using readonly mode for, but it does make more sense to me that specifically-disabled dates would appear different, even when readonly.

core/src/components/datetime/datetime.tsx
```diff
@@ -2011,7 +2042,11 @@ export class Datetime implements ComponentInterface {
 
           <div class="calendar-next-prev">
             <ion-buttons>
-              <ion-button aria-label="Previous month" disabled={prevMonthDisabled} onClick={() => this.prevMonth()}>
+              <ion-button
+                aria-label="Previous month"
+                disabled={prevMonthDisabled || disabled}
```
sean-perkins, line 2047: We should be updating the `prevMonthDisabled` constant with the additional check for `disabled`.

core/src/components/datetime/datetime.tsx
```diff
@@ -2021,7 +2056,11 @@ export class Datetime implements ComponentInterface {
                   flipRtl
                 ></ion-icon>
               </ion-button>
-              <ion-button aria-label="Next month" disabled={nextMonthDisabled} onClick={() => this.nextMonth()}>
+              <ion-button
+                aria-label="Next month"
+                disabled={nextMonthDisabled || disabled}
```
sean-perkins, line 2061: We should be updating the `nextMonthDisabled` constant with the additional check for disabled.

core/src/components/datetime/datetime.tsx
```diff
@@ -2094,16 +2145,21 @@ export class Datetime implements ComponentInterface {
...
 
             const dateIsoString = convertDataToISO(referenceParts);
-            let isCalDayDisabled = isCalMonthDisabled || disabled;
 
-            if (!isCalDayDisabled && isDateEnabled !== undefined) {
+            /**
+             * Days can be constrained by the min/max bounds or
+             * allowed dates. Constrained days may not be selected.
+             */
+            let isCalDayConstrained = isCalMonthDisabled || isDayDisabled;
```
sean-perkins, line 2153: ~ Nit: I wouldn't know what constrained means for a variable without looking at this comment. If the behavior is around the day being selected, a name focused around that may be more appropriate.

sean-perkins (changes requested):

core/src/components/datetime/datetime.tsx
mapsandapps: I was going off of "date constraints" which is how the docs refer to this feature. I couldn't think of any better way to distinguish between 

- dates that are specifically disabled (e.g. by min & max dates or specific allowed values) and 
- dates that are disabled for any reason (which includes the above but also if the component is disabled or readonly)

which is what I need to do here.

sean-perkins: After looking at the code, I think we can do this with a lot less modifications. But please double-check with the behavior you expect:

Revert this back:
```ts
let isCalDayDisabled = isCalMonthDisabled || isDayDisabled;

if (!isCalDayDisabled && isDateEnabled !== undefined) {
  try {
    /**
     * The `isDateEnabled` implementation is try-catch wrapped
     * to prevent exceptions in the user's function from
     * interrupting the calendar rendering.
     */
    isCalDayDisabled = !isDateEnabled(dateIsoString);
  } catch (e) {
    printIonError(
      'Exception thrown from provided `isDateEnabled` function. Please check your function and try again.',
      el,
      e
    );
  }
}
```

Add a boolean for the datetime being "disabled", you call this `isButtonDisabled` in other places, but here I think that could be confusing, so I opted for:
```ts
const isDatetimeDisabled = disabled || readonly;
```

You can then update your JSX bindings (I'll remove all the existing props that are not effected):

```jsx
<button
   disabled={isCalDayDisabled || isDatetimeDisabled}
  class={{
    'calendar-day-constrained': isCalDayDisabled && isDatetimeDisabled,
  }}
```

This should treat the button as disabled either if it is explicitly disabled **or** if the entire datetime is readonly and/or disabled. However it will optically be dimmed in readonly mode only if day is both disabled **and** the datetime is readonly and/or disabled. 

Advantages to this approach is we aren't coupling this constrained state to the disabled state of the day.

mapsandapps: i like that generally. but some of the specifics are confusing me. 

from other feedback on this PR, i've come to understand that our code style is to have logic in consts, not inside the JSX.

so instead of

```
<button
   disabled={isCalDayDisabled || isDatetimeDisabled}
  class={{
    'calendar-day-constrained': isCalDayDisabled && isDatetimeDisabled,
  }}
```

it should be e.g.

```
<button
   disabled={isButtonDisabled}
  class={{
    'calendar-day-constrained': isFoo,
  }}
```

i updated the PR with both your feedback and the other feedback in mind. still open to ideas for improved const names.

sean-perkins: There are a few places where abstracting inline expressions into a constant are useful:
1. The expression is re-used in multiple places in the implementation.
2. The expression is needing to access a value off the instance (`this.variableName`). 
3. The expression is either complicated or could benefit from an explicit variable to describe what the check is responsible for. 

I think it is fine either way here if you want to have new constants for abstracting `isCalDayDisabled && isDatetimeDisabled`. I mainly wanted to avoid intermixing the logic for checking if a date is enabled, with the behavior of making the day display at full opacity when in readonly mode.

core/src/components/datetime/datetime.tsx
```diff
@@ -2113,6 +2157,15 @@ export class Datetime implements ComponentInterface {
               }
             }
 
+            /**
+             * Some days are disabled through max & min or allowed dates
+             * and also by the component being readonly or disabled.
+             * These need to be displayed differently.
+             */
+            const isCalDayDoublyDisabled = isCalDayDisabled && isDatetimeDisabled;
```
sean-perkins, line 2165: I'd rename this back to `isCalDayConstrained`, then this PR is good to merge. 

"doubly disabled" is confusing in my opinion. An interactive can only be in a singular disabled state. 

Having the variable match the CSS class will probably be best for now, giving queues to both our dev team and 3rd party contributors to where they can look for the associated implementation.

sean-perkins (approved):

core/src/components/datetime/datetime.tsx
```diff
@@ -1957,11 +1986,12 @@ export class Datetime implements ComponentInterface {
...
 
   private renderCalendarHeader(mode: Mode) {
+    const { disabled } = this;
     const expandedIcon = mode === 'ios' ? chevronDown : caretUpSharp;
     const collapsedIcon = mode === 'ios' ? chevronForward : caretDownSharp;
 
-    const prevMonthDisabled = isPrevMonthDisabled(this.workingParts, this.minParts, this.maxParts);
-    const nextMonthDisabled = isNextMonthDisabled(this.workingParts, this.maxParts);
+    const prevMonthDisabled = isPrevMonthDisabled(this.workingParts, this.minParts, this.maxParts) || disabled;
+    const nextMonthDisabled = isNextMonthDisabled(this.workingParts, this.maxParts) || disabled;
```
amandaejohnston, line 1994:
```suggestion
    const prevMonthDisabled = disabled || isPrevMonthDisabled(this.workingParts, this.minParts, this.maxParts);
    const nextMonthDisabled = disabled || isNextMonthDisabled(this.workingParts, this.maxParts);
```
Nitpicky, but this way we can short-circuit the `isDisabled` functions.

amandaejohnston (approved):
Thanks for sorting through everything here!

PR #28343, opened 2023-10-12
fix(angular): do not create duplicate menuController instances

### LINKED ISSUE #28337: bug: ionic angular 7.5, cannot access menus with menuController
### Prerequisites

### Ionic Framework Version

v7.x

### Current Behavior

After upgrade @ionic/angular from 7.4.4 to 7.5.0 MenuController.enable() stop working.

### Expected Behavior

A menu initially disabled should be enabled after calling enable(true) from MenuController on app.component.ts.

### Steps to Reproduce

1. Create a new Ionic sidemenu app (angular).
2. Run `npm i @ionic/angular@7.4.4`
3. Disable ion-menu on app.component.html `<ion-menu contentId="main-content" type="overlay" [disabled]="true">`
4. Enable the menu on app.component.ts when platform is ready:
```
constructor(private menuCtrl: MenuController, private platform: Platform) {
    this.platform.ready().then(() => {
      this.menuCtrl.enable(true);
    });
  }
```
5. Run `ionic serve` and you will see the menu enabled (as expected).
6. Stop ionic serve and run `npm i @ionic/angular@7.5.0`
7. Run `ionic serve` again and the menu will be disabled.

### Code Reproduction URL

https://stackblitz.com/edit/angular-lgwu3i?file=src%2Fapp%2Fapp.component.ts

### Ionic Info

Ionic:

   @angular-devkit/build-angular : 16.2.6
   @angular-devkit/schematics    : 16.2.6
   @angular/cli                  : 16.2.6

Capacitor:

   Capacitor CLI      : 5.5.0
   @capacitor/core    : 5.5.0

Utility:

   cordova-res : 0.15.4
   native-run  : 1.7.3

System:

   NodeJS : v18.16.0 (C:\Program Files\nodejs\node.exe)
   npm    : 9.6.7
   OS     : Windows 10

### Additional Information

_No response_

liamdebeasi:
Issue number: resolves #28337

---------

Duplicate instances of `menuController` are being created in `@ionic/angular`. `ion-menu` registers itself in the `menuController` from `@ionic/core`, but the `MenuController` from `@ionic/angular` uses the `menuController` from `@ionic/core/components`. This is how the overlay providers work too. Normally, this is not a problem. However, `menuController` caches references to registered menus in each controller instances: https://github.com/ionic-team/ionic-framework/blob/dcbf45101f4c0dc3541cd4c19186daa3766a6ce7/core/src/utils/menu-controller/index.ts#L14

This means that since there are two different controllers, `menuController` B does not know about the menus in `menuController` A. The end result is that the menu controller used in developer applications did not have references to the registered menus, which gave the impression that the menu controller did not work.

## What is the new behavior?

- Updated the architecture of `MenuController` in Ionic Angular to accept a `menuController` instance. This allows `@ionic/angular` to pass the `menuController` from `@ionic/core` and for `@ionic/angular/standalone` to pass the `menuController` from `@ionic/core/components`.

Note: Overlay controllers don't **need** this change per-se since they don't cache references to overlays internally (they just query the DOM). However, I think it would be good to have a consistent architecture here, so I'll put up a separate PR that makes this change for overlays too.

- [ ] Yes
- [x] No

## Other information

Dev build: `7.5.1-dev.11697123035.1ee6b4a2`

### DIFF
--- a/core/src/components/menu/menu-interface.ts
+++ b/core/src/components/menu/menu-interface.ts
@@ -26,23 +26,23 @@ export interface MenuI {
 }
 
 export interface MenuControllerI {
-  registerAnimation: (name: string, animation: AnimationBuilder) => void;
-  get: (menu?: string | null, logOnMultipleSideMenus?: boolean) => Promise<HTMLIonMenuElement | undefined>;
-  getMenus: () => Promise<HTMLIonMenuElement[]>;
-  getOpen: () => Promise<HTMLIonMenuElement | undefined>;
-  isEnabled: (menu?: string | null) => Promise<boolean>;
-  swipeGesture: (shouldEnable: boolean, menu?: string | null) => Promise<HTMLIonMenuElement | undefined>;
-  isAnimating: () => Promise<boolean>;
-  isOpen: (menu?: string | null) => Promise<boolean>;
-  enable: (shouldEnable: boolean, menu?: string | null) => Promise<HTMLIonMenuElement | undefined>;
-  toggle: (menu?: string | null) => Promise<boolean>;
-  close: (menu?: string | null) => Promise<boolean>;
-  open: (menu?: string | null) => Promise<boolean>;
-  _getOpenSync: () => HTMLIonMenuElement | undefined;
-  _createAnimation: (type: string, menuCmp: MenuI) => Promise<Animation>;
-  _register: (menu: MenuI) => void;
-  _unregister: (menu: MenuI) => void;
-  _setOpen: (menu: MenuI, shouldOpen: boolean, animated: boolean) => Promise<boolean>;
+  registerAnimation(name: string, animation: AnimationBuilder): void;
+  get(menu?: string | null, logOnMultipleSideMenus?: boolean): Promise<HTMLIonMenuElement | undefined>;
+  getMenus(): Promise<HTMLIonMenuElement[]>;
+  getOpen(): Promise<HTMLIonMenuElement | undefined>;
+  isEnabled(menu?: string | null): Promise<boolean>;
+  swipeGesture(shouldEnable: boolean, menu?: string | null): Promise<HTMLIonMenuElement | undefined>;
+  isAnimating(): Promise<boolean>;
+  isOpen(menu?: string | null): Promise<boolean>;
+  enable(shouldEnable: boolean, menu?: string | null): Promise<HTMLIonMenuElement | undefined>;
+  toggle(menu?: string | null): Promise<boolean>;
+  close(menu?: string | null): Promise<boolean>;
+  open(menu?: string | null): Promise<boolean>;
+  _getOpenSync(): HTMLIonMenuElement | undefined;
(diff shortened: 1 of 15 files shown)


packages/angular/common/src/providers/menu-controller.ts
```diff
@@ -1,17 +1,23 @@
...
-import { menuController } from '@ionic/core/components';
 
 @Injectable({
   providedIn: 'root',
 })
 export class MenuController {
+  private menuController;
+
+  // TODO Now - Add type
+  constructor(private menuCtrl: any) {
```
sean-perkins, line 10: I think we should remove the `@Injectable` from this class in common. We aren't using it with dependency injection in the extending implementations. If we did based on the current implementation, Angular would not know how to resolve `menuCtrl` and would throw an exception. 

Since we are treating this as a base class of logic that is extended in the hydrated and custom elements implementations, I think we are safe to remove it. 

re: Type, could you import the type from of `menuController` from `@ionic/core/components`? We already have some instances in the common package where we are assuming the CE build exists in the distributed package.

liamdebeasi: Removed Injectable in base class in https://github.com/ionic-team/ionic-framework/pull/28343/commits/fb772c62627bf6c0c236e4328864055e3abba6d5

Re: the menuController type. We discussed offline, but the current controller interface was wrong so I fixed it in https://github.com/ionic-team/ionic-framework/pull/28343/commits/c102863f0267af17ab4451daa1edafaaf213a8f3

core/src/components/menu/menu-interface.ts
```diff
@@ -26,13 +26,23 @@ export interface MenuI {
...
+  close(menu?: string | null): Promise<boolean>;
+  open(menu?: string | null): Promise<boolean>;
+  _getOpenSync(): HTMLIonMenuElement | undefined;
   _createAnimation(type: string, menuCmp: MenuI): Promise<Animation>;
-  _setOpen(menu: MenuI, shouldOpen: boolean, animated: boolean): Promise<boolean>;
   _register(menu: MenuI): void;
   _unregister(menu: MenuI): void;
-
-  getMenus(): Promise<HTMLIonMenuElement[]>;
-  getOpenSync(): HTMLIonMenuElement | undefined;
```
liamdebeasi, line 35: `getOpenSync` doesn't actually exist on the menu controller. However, `_getOpenSync` does so I think this was a typo.

thetaPC (approved):
LGTM. Maybe add a comment on the tests that they were created for issue 28337, like the annotations in core.

core/src/components/menu/menu-interface.ts
thetaPC: Good catch!

amandaejohnston (approved):

sean-perkins (approved):
Looks good to me. Thanks for the quick turnaround on this!

PR #28353, opened 2023-10-13
fix(rtl): allow :host to use rtl()

thetaPC:
Issue number: N/A

---------

While working on a safe area padding mixin, I realized that `rtl()` wasn't being applied for `:host` when using Firefox or Safari. This is happening because the syntax for `:dir()` is wrong. The placement needs to be updated for Firefox and Safari to register it.

```scss
:host {
   @include rtl() { // <- won't work
      // styles
   }
}

// generates
:host-context([dir=rtl]) {
  // styles
}

:host:dir(rtl) { // <- wrong syntax
   // styles
}
```

```scss
:host(.class) {
   @include rtl() { // <- won't work
      // styles
   }
}

// generates
:host-context([dir=rtl]):host(.class) {
  // styles
}

:host-context([dir=rtl]).class {
   // styles
}

:host(.class):dir(rtl) { // <- wrong syntax
   // styles
}
```

## What is the new behavior?

I updated `rtl()` to use `:dir()` as the `addHostSelector` in the `add-root-selector` function. This generates all the correct selectors for Firefox and Safari. However, `:dir()` does not have the structure of `:host-context()` so I had to add a new parameter to `add-root-selector` to determine whether to use `:host-context()` or not. I set the default to `true` since the function originally used `:host-context()`.

An extra win is that the updated function will be ready for when `:host-context()` can be removed from the codebase.

```diff
:host {
   @include rtl() { // <- works
      // styles
   }
}

// generates
:host-context([dir=rtl]) {
  // styles
}

-  :host:dir(rtl) {
+  :host(:dir(rtl)) {
   // styles
}
```

```diff
:host(.class) {
   @include rtl() { // <- works
      // styles
   }
}

// generates
:host-context([dir=rtl]):host(.class) {
  padding-right: 40px;
}

:host-context([dir=rtl]).class {
   // styles
}

-  :host(.class):dir(rtl) {
+  :host(.class:dir(rtl)) {
   // styles
}
```

- [ ] Yes
- [x] No

## Other information

N/A


core/src/themes/ionic.functions.string.scss
```diff
@@ -85,24 +85,35 @@
 // Add Root Selector
 // --------------------------------------------------------------------------------
 // Adds a root selector using host based on the selector passed
+// $root: The selector to add the root to
+// $addHostSelector: The selector to add to the root
```
brandyscarney, line 89: Why are these the same description? Could we improve this? Maybe we should add examples of how to use `add-root-selector` like:

```
add-root-selector(&, "[dir=rtl]")
add-root-selector(".item", ":dir(rtl)", false)
```

I don't know if the above is accurate but I think it is.

brandyscarney (commented):
I tested it against my PR here where I ran into this issue: https://github.com/ionic-team/ionic-framework/pull/28371

It is working as expected in Firefox but I am still not seeing it resolved in Safari. 

Below is the change I see in the generated code for the following code snippet between main and this branch:

```css
.item { 
  @include ltr() {
    padding-left: 40px;
  }

  @include rtl() {
    padding-right: 40px;
  }
}
```

<table>
  <thead>
    <td><code>main</code></td>
    <td><code>branch</code></td>
  </thead>
  <tr>
    <td>
      <pre lang="css">
.item {
  padding-left: 40px;
}
:host-context([dir=rtl]) .item {
  padding-right: 40px;
}
[dir=rtl] .item {
  padding-right: 40px;
}
@supports selector(:dir(rtl)) {
  .item:dir(rtl) {
    padding-right: 40px;
  }
}
      </pre>
    </td>
    <td>
      <pre lang="css">
.item {
  padding-left: 40px;
}
:host-context([dir=rtl]) .item {
  padding-right: 40px;
}
[dir=rtl] .item {
  padding-right: 40px;
}
@supports selector(:dir(rtl)) {
  :dir(rtl) .item {
    padding-right: 40px;
  }
}
      </pre>
    </td>    
  </tr>
</table>

Is there something missing?

brandyscarney:
Maria and I looked into the issue in my last comment. It turns out that I was running into [this bug](https://bugs.webkit.org/show_bug.cgi?id=257133) while adding `?rtl=true` to the browser url of the test I was previewing. If I added `dir="rtl"` to the `html` I did not run into this. This bug was fixed in v17 of Safari but I am using v16.6 which is why I was seeing it and Maria was not. She is going to add a comment in the code in case anyone else runs into it while testing.

core/src/themes/ionic.functions.string.scss
thetaPC: They're very similar and can make it hard to distinguish. I'll update them.

brandyscarney (approved):

PR #28363, opened 2023-10-16
chore: remove unused sass variables

liamdebeasi:
Issue number: #

---------

## What is the new behavior?

-
-
-

- [ ] Yes
- [ ] No


brandyscarney (approved):
This looks good. The build passes and I can't see any issues with removing these. Should we update the linter here to check for unused variables:

https://github.com/ionic-team/ionic-framework/blob/feda7a0e963048d300eb17d4b9e1056f09088714/core/package.json#L85

Alternatively, we could make that part of the updates with the Sass guidelines in [FW-5394](https://ionic-cloud.atlassian.net/browse/FW-5394).

amandaejohnston (approved):

liamdebeasi:
I can look into adding a linter in a separate PR! I might also snag that for a hack day

PR #28368, opened 2023-10-17
docs(breaking): add v8 browser and platform support

liamdebeasi:
Issue number: N/A

---------

As part of our Ionic 8 work, the team would like to define which browsers and platforms we will support.

## What is the new behavior?

- Chromium (Chrome + Edge) support has changed from 79+ to 89+
- Firefox support has changed from 70+ to 75+
- Safari/WebKit support has changed from 14+ to 15+
- Angular support has changed from 14+ to 16+

A few notes on why I made these changes:

- These browsers have full support for [Web Animations Level 1](https://caniuse.com/?search=web%20animations). This allows us to remove the old CSS Animation fallback support from our animation library.
- These browsers have [native image lazy loading](https://caniuse.com/?search=image%20lazy%20loading) support (iOS has it starting in iOS 15.4) which allows us to reduce our dependency on `ion-img`.
- These browsers have full support for [CSS Logical Properties](https://caniuse.com/?search=logical%20properties) which will help improve our RTL support in Ionic.
- This also gives us ~3 years of Firefox and Chrome support by the time Ionic 8 comes out. Firefox 75 came out April 2020, and Chrome 89 came out February 2021.
- Dropping iOS 14 allows us to remove several iOS 14-related hacks in item, datetime, and more.
- Dropping iOS 14 still lets us support >94% of iPhones and >91% of iPads as of May 2023 (https://developer.apple.com/support/app-store/).
- Dropping Angular 14 and 15 allows us to remove several Angular specific hacks. Angular 14 LTS ends at the end of 2023, and Angular 15 LTS ends in May 2024, so these versions are not being supported by the Angular team for much longer. By the time Ionic 8 is out, Angular 17 should be released with Angular 18 coming soon.

- [x] Yes
- [ ] No

### DIFF
--- a/BREAKING.md
+++ b/BREAKING.md
@@ -12,3 +12,30 @@ This is a comprehensive list of the breaking changes introduced in the major ver
 - [Legacy](https://github.com/ionic-team/ionic-v3/blob/master/CHANGELOG.md)
 
 ## Version 8.x
+
+- [Browser and Platform Support](#version-8x-browser-platform-support)
+
+<h2 id="version-8x-browser-platform-support">Browser and Platform Support</h2>
+
+This section details the desktop browser, JavaScript framework, and mobile platform versions that are supported by Ionic 8.
+
+**Minimum Browser Versions**
+| Desktop Browser | Supported Versions |
+| --------------- | ----------------- |
+| Chrome          | 84+               |
+| Safari          | 15+               |
+| Firefox         | 75+               |
+| Edge            | 84+               |
+
+**Minimum JavaScript Framework Versions**
+| Framework | Supported Version     |
+| --------- | --------------------- |
+| Angular   | 16+                   |
+| React     | 17+                   |
+| Vue       | 3.0.6+                |
+
+**Minimum Mobile Platform Versions**
+| Platform | Supported Version      |
+| -------- | ---------------------- |
+| iOS      | 15+                    |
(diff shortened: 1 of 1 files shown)


BREAKING.md
```diff
@@ -12,3 +12,30 @@ This is a comprehensive list of the breaking changes introduced in the major ver
...
+| Safari          | 15+               |
+| Firefox         | 75+               |
+| Edge            | 84+               |
+
+**Minimum JavaScript Framework Versions**
+| Framework | Supported Version     |
+| --------- | --------------------- |
+| Angular   | 16+                   |
+| React     | 17+                   |
+| Vue       | 3.0.6+                |
```
liamdebeasi, line 35: I didn't see a big need to change these versions, so I left them alone for now.

brandyscarney (approved):

thetaPC (approved):
LGTM

liamdebeasi:
Re-requesting the team for review.

I wanted to propose changing the Chromium minimum version to 89+ instead of the previously noted 84+. The reason for this is targeting 89+ allows us to make use of [CSS Logical Properties](https://caniuse.com/?search=logical%20properties) for improved RTL support. We make use of some today, but we'd be able to use even more starting in v8. This could also reduce our dependence on custom mixins for certain properties.

brandyscarney (approved):

thetaPC (approved):
LGTM

PR #28369, opened 2023-10-17
fix(alert, action-sheet): show scrollbar for long list of options

### LINKED ISSUE #18487: bug: ion-select cannot handle long lists with a visible scrollbar for web users (alert view, Ionic 4.4.2)
With Ionic 4.4.2, the ion-select component (with default alert interface) cannot be configured or styled to show a scrollbar for larger lists of more than 8 or so items. This makes the component unusable on larger lists where web-based users need scrollbars.  

For lists of 10 to 20 items, there are temporary workarounds for changing the alert-wrapper max-height and adjusting the inner alert button height, but for larger 20+ lists, there doesn't seem to be any way to force a scrollbar on the inner content section of the alert (e.g. `overflow-y: scroll` on the alert-wrapper class). 

After a lot of troubleshooting, it was easier to switch over to a plain select component for the larger. If this can't be easily fixed, then to support future CSS styling, it would be useful if there were an inner div inside of alert-wrapper, and if alert-wrapper could be set to overflow-y:scroll. There is something preventing `overflow-y:scroll !important`  from taking effect on alert-wrapper, even in Chrome dev tools... maybe someone with more flex experience has a solution.

thetaPC:
Issue number: resolves #18487 

---------

Web-based users do not get a scrollbar when:

- alert has a long list of inputs (this also happens on `ion-select` with the alert interface)
- `ion-select` uses the action-sheet interface and has a long list of options

This makes it difficult for users to navigate through the options by forcing them to use their keyboards. Some users may also not be used to using their keyboards for navigation. Additionally, this can lead to potential confusion that there are no other options.

## What is the new behavior?

Web-based users get a scrollbar when:

- alert has a long list of inputs (this also happens on `ion-select` with the alert interface)
- `ion-select` uses the action-sheet interface and has a long list of options

- [ ] Yes
- [x] No

## Other information

The issue was filed for the alert interface but it's also happening on the action-sheet interface.

Dev build: 7.5.1-dev.11697570585.1774584d

### DIFF
--- a/core/src/components/alert/alert.scss
+++ b/core/src/components/alert/alert.scss
@@ -101,7 +101,7 @@
 // keyboard to navigate through the options.
 // This may not be intuitive for users who
 // are not familiar with keyboard navigation.
-@media (any-pointer: none) {
+@media (any-pointer: coarse) {
   .alert-checkbox-group::-webkit-scrollbar,
   .alert-radio-group::-webkit-scrollbar,
   .alert-message::-webkit-scrollbar {
--- a/core/src/components/alert/alert.scss
+++ b/core/src/components/alert/alert.scss
@@ -93,13 +93,14 @@
 
 // Scrollbars on mobile devices will be hidden.
 // Users can still scroll the content by swiping.
-// Desktop browsers will show scrollbars. This is
-// important with a long list of options since
-// this allows users to also have the option to
-// scroll with a mouse wheel or trackpad. Otherwise,
-// the user would have to use the keyboard to
-// navigate through the options. This may not be
-// obvious to the user.
+// If a user has a pointing device, such as a
+// mouse or trackpad, then scrollbars will be
+// visible. This allows users to scroll the
+// content with their pointing device.
+// Otherwise, the user would have to use the
+// keyboard to navigate through the options.
+// This may not be intuitive for users who
+// are not familiar with keyboard navigation.
 @media (any-pointer: none) {
   .alert-checkbox-group::-webkit-scrollbar,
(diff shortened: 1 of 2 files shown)


core/src/components/alert/alert.scss
```diff
@@ -91,10 +91,22 @@
...
+// Users can still scroll the content by swiping.
+// If a user has a pointing device, such as a
+// mouse or trackpad, then scrollbars will be
+// visible. This allows users to scroll the
+// content with their pointing device.
+// Otherwise, the user would have to use the
+// keyboard to navigate through the options.
+// This may not be intuitive for users who
+// are not familiar with keyboard navigation.
+@media (any-pointer: none) {
```
liamdebeasi, line 104: Do you mean to do `any-pointer: coarse` instead? `any-pointer: none` means that this CSS will only apply when there is no pointing device available. A `coarse` pointer has limited accuracy, so that will apply on touch screens.

With this change as-is the scrollbars show up on mobile devices.

thetaPC: My thought was that when no pointers are present then don't show the scrollbars. If no pointers are available then it's a mobile device. However, I just realized that touching the screen will count as a pointer. I had to dig a bit more to get a better description of coarse. I agree with changing it to `coarse`.

liamdebeasi (approved):
Implementation looks good. When merging, I recommend making it clear on the linked issue that scrollbars for action sheet and alert will now show up on non-touch devices, so during development developers should expect to see scrollbars (but not on actual mobile devices)

PR #28371, opened 2023-10-17
fix(list-header): apply safe area to proper side regardless of direction

brandyscarney:
Issue number: Internal

---------

The list header adds padding to the "start" side (`padding-left` in LTR and `padding-right` in RTL) based on the value of `--ion-safe-area-left`. It does not account for `--ion-safe-area-right` at all even though the list header can extend to the right side of the content.

## What is the new behavior?

- The `--ion-safe-area-left` always applies to the left side of the screen, regardless of direction. This means that in both LTR and RTL it applies as `padding-left`.
- Added support for `--ion-safe-area-right` which applies to `padding-right` in both LTR and RTL.
- Adds an e2e test which captures the list header with a button to ensure the proper padding is added for the safe area. 

- [ ] Yes
- [x] No

### Safe Area Left

| mode | direction | main | branch |
| ---| ---| ---| ---|
| `ios` | `LTR` |  ios-ltr-main | ios-ltr-branch |
| `ios` | `RTL` |  ios-rtl-main | ios-rtl-branch |
| `md` | `LTR` | md-ltr-main | md-ltr-branch |
| `md` | `RTL` | md-rtl-main  | md-rtl-branch |

### Safe Area Right

| mode | direction | main | branch |
| ---| ---| ---| ---|
| `ios` | `LTR` | ios-ltr-main |ios-ltr-branch |
| `ios` | `RTL` | ios-rtl-main | ios-rtl-branch |
| `md` |`LTR` | md-ltr-main | md-ltr-branch |
| `md` |`RTL` | md-rtl-main | md-rtl-branch |


core/src/components/list-header/test/basic/list-header.e2e.ts
```diff
@@ -12,3 +12,60 @@ configs().forEach(({ title, screenshot, config }) => {
     });
   });
 });
+
+/**
+ * This behavior needs to be tested in both modes and directions to
+ * make sure the safe area padding is applied only to that side
+ * regardless of direction
+ */
+configs().forEach(({ title, screenshot, config }) => {
```
liamdebeasi, line 21: Implementation looks good, I just have the same question as in https://github.com/ionic-team/ionic-framework/pull/28372#discussion_r1365789776

liamdebeasi (approved):

PR #28372, opened 2023-10-17
fix(tab-bar): apply safe area to proper side regardless of direction

brandyscarney:
Issue number: Internal

---------

The safe area padding (both left and right) swap sides when the app's direction changes from LTR to RTL. The `--ion-safe-area-left` should always apply to the left side of the device and the `--ion-safe-area-right` should always apply to the right side of the device.

## What is the new behavior?

- Updates the tab bar stylesheet to always set `padding-left` and `padding-right`
- Adds an e2e test for the basic directory which adds screenshots in both modes/directions for:
  - the default tab bar
  - a tab bar with safe area left applied
  - a tab bar with safe area right applied

- [ ] Yes
- [x] No


core/src/components/tab-bar/test/basic/tab-bar.e2e.ts
```diff
@@ -0,0 +1,108 @@
+import { expect } from '@playwright/test';
+import { configs, test } from '@utils/test/playwright';
+
+configs().forEach(({ title, screenshot, config }) => {
+  test.describe(title('tab-bar: basic'), () => {
+    test('should not have visual regressions', async ({ page }) => {
+      await page.setContent(
```
liamdebeasi, line 7: We already have test coverage for this here: https://github.com/ionic-team/ionic-framework/blob/068d0038602485085face4a443e473bf1d5cc227/core/src/components/tab-button/test/basic/tab-button.e2e.ts#L6
Maybe the test needs to move though since it's not obvious?

core/src/components/tab-bar/test/basic/tab-bar.e2e.ts
```diff
@@ -0,0 +1,108 @@
...
+
+/**
+ * This behavior needs to be tested in both modes and directions to
+ * make sure the safe area padding is applied only to that side
+ * regardless of direction
+ */
+configs().forEach(({ title, screenshot, config }) => {
+  test.describe(title('tab-bar: basic'), () => {
+    test.describe('safe area left', () => {
+      test('should have padding on the left only', async ({ page }) => {
```
liamdebeasi, line 41: Why do we have two different tests? I figured we could have one test with an uneven amount of safe area  (i.e 20px on the left 40px on the right) to make sure that the safe area is being added to the correct side

brandyscarney: I thought about separating them but I felt it was easier to catch regressions and visually see that there was _no padding_ on the one side.

liamdebeasi: I'm not sure I follow. If the we had 1 test that applied different amounts of left/right safe area we'd still see diffs. Reason why I'm asking about this is we effectively have 2 tests testing the same underlying functionality.

For example, if we applied 40px of left safe area padding and 20px of safe area padding we'd get the following screenshots and diff if this fix regresses in the future. 

(RTL direction, iOS mode, Chrome)

Base Screenshot | Regressed Screenshot | Diff |
| - | - | - |
| tab-bar-safe-area-left-ios-rtl-Mobile-Chrome-darwin | tab-bar-safe-area-left-ios-rtl-actual | tab-bar-safe-area-left-ios-rtl-diff |

This test applied the following CSS:

```css
 :root {
  --ion-safe-area-left: 40px;
  --ion-safe-area-right: 20px;
}
```

brandyscarney: Yeah I was just thinking it would be easier to tell if a change happened to the padding on the component itself but I suppose you would have to verify the diff either way. I can update it.

core/src/components/tab-bar/test/basic/tab-bar.e2e.ts
brandyscarney: I guess it's confusing because in Segment we keep the tests on `ion-segment` and have none in `ion-segment-button` but in Radio we have tests in both `ion-radio-group` and `ion-radio`. I could remove mine though, up to you.

liamdebeasi: I think we should remove the test here since it is duplicating test effort. However, I do agree that where some of these tests live are confusing. Maybe we can consolidate in a separate PR?

liamdebeasi (approved):
Good to go once my last comment is addressed.

PR #28376, opened 2023-10-18
fix(segment): avoid scrolling webkit bug

### LINKED ISSUE #28373: bug: rendering issue with ion-segment on iOS
### Prerequisites

### Ionic Framework Version

v7.x

### Current Behavior

After updating my app to "Ionic 7", I encountered a bug in "ion-segment": say you have a page with an `ion-segment` with 5 segments and the third is pre-selected through the `[(ngModel)]` or `[value]` binding. In addition to the value bindings `[scrollable]="true"` is also set (scrollable is actually used in my real application). Then if you navigate into this page it takes a moment to render the `ion-segment` component at the correct position.

As far as I can tell this bug is only reproducible on iOS (Simulator or Device). It's reproducible in Safari on my desktop computer but not in Chrome.

I've made a short clip to see the bug in action and also made the code to reproduce it available.

https://github.com/ionic-team/ionic-framework/assets/128096/70530696-95c2-4750-8e2e-318251e114d3

https://github.com/trytuna/IonicSegmentBug

### Expected Behavior

Should render without any visual glitches

### Steps to Reproduce

1. git clone https://github.com/trytuna/IonicSegmentBug
2. cd IonicSegmentBug
3. npm install
4. ionic build
5. npx cap sync ios
6. npx cap open ios
7. run it on an iPhone (e.g. 15 Pro iOS 17.0)
8. Navigate into the page with the bug by clicking on the blue "Bug"-Button

### Code Reproduction URL

https://github.com/trytuna/IonicSegmentBug

### Ionic Info

% ionic info
       
       Require stack:
       - /Users/timo/.nvm/versions/node/v16.17.1/lib/node_modules/@ionic/cli/lib/project/index.js
       - /Users/timo/.nvm/versions/node/v16.17.1/lib/node_modules/@ionic/cli/lib/index.js
       - /Users/timo/.nvm/versions/node/v16.17.1/lib/node_modules/@ionic/cli/index.js
       - /Users/timo/.nvm/versions/node/v16.17.1/lib/node_modules/@ionic/cli/bin/ionic

Ionic:

   Ionic CLI                     : 7.1.1 (/Users/timo/.nvm/versions/node/v16.17.1/lib/node_modules/@ionic/cli)
   @angular-devkit/build-angular : 16.2.6
   @angular-devkit/schematics    : 16.2.6
   @angular/cli                  : 16.2.6

Capacitor:

   Capacitor CLI      : 5.5.0
   @capacitor/core    : 5.5.0
   @capacitor/ios     : 5.5.0

Utility:

   native-run  : 1.7.3

System:

   NodeJS : v16.17.1 (/Users/timo/.nvm/versions/node/v16.17.1/bin/node)
   npm    : 9.8.0
   OS     : macOS Unknown

### Additional Information

_No response_

liamdebeasi:
Issue number: resolves #28373

---------

🚨 Reviewers: Please test this on Chrome, Firefox, and Safari

The fix in https://github.com/ionic-team/ionic-framework/commit/1167a9325fb930b6c727bc26889f5488d9620062 uncovered a WebKit bug where scrolling an element using `scrollIntoView` during an accelerated transition causes the segment to jump during the transition. `scrollIntoView` can cause parent elements to also scroll, and given that the entering view begins outside the viewport, it's possible this triggered some sort of WebKit bug where it was trying to scroll the element into view.

## What is the new behavior?

- Updated the internal implementation to use `scrollBy` instead of `scrollIntoView`. `scrollBy` does not attempt to scroll parent elements which avoids the WebKit issue.
- I also updated the initial scroll to be instant rather than smoothly scroll. If a segment is added to the DOM, I'd expect it to be added with the correct scroll position (after the first render that is), so animating on load seemed like a strange behavior. User interaction will continue to use smooth scrolling though.

- [ ] Yes
- [x] No

## Other information

Dev build: `7.5.2-dev.11697638908.1f04980a`

### DIFF
--- a/core/src/components/segment/segment.tsx
+++ b/core/src/components/segment/segment.tsx
@@ -145,7 +145,13 @@ export class Segment implements ComponentInterface {
      * before we can scroll.
      */
     raf(() => {
-      this.scrollActiveButtonIntoView();
+     /**
+      * When the segment loads for the first
+      * time we just want to snap the active button into
+      * place instead of scroll. Smooth scrolling should only
+      * happen when the user interacts with the segment.
+      */
+      this.scrollActiveButtonIntoView(false);
     });
 
     this.gesture = (await import('../../utils/gesture')).createGesture({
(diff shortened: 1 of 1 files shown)


amandaejohnston (approved):
Tested in Chrome, Firefox, and Safari, all on both LTR and RTL. Works like a charm 👍

PR #28377, opened 2023-10-18
fix(fab): apply safe area in positioning to proper side regardless of direction

amandaejohnston:
Issue number: Internal

---------

When calculating the fab's horizontal position, the safe area is taken into account. However, which safe area side is applied changes depending on whether the document's direction is LTR or RTL. This is incorrect as the left safe area padding will always be on the left side regardless of direction, and vice versa.

## What is the new behavior?

The left safe area is always applied to the fab's `left` position, and vice versa.

- [ ] Yes
- [x] No


core/src/components/fab/fab.scss
```diff
@@ -23,15 +23,29 @@
...
-    calc(#{$fab-content-margin} + var(--ion-safe-area-left, 0px)), null
-  );
+  /* stylelint-disable */
+  @include ltr() {
+    left: calc(#{$fab-content-margin} + var(--ion-safe-area-left, 0px));
+  }
+
+  @include rtl() {
+    right: calc(#{$fab-content-margin} + var(--ion-safe-area-right, 0px));
+    left: auto;
```
brandyscarney, line 33: Does `unset` work here?

amandaejohnston: Looks like it 👍

brandyscarney (approved):

PR #28386, opened 2023-10-19
chore(angular): remove radio value accessor

thetaPC:
Issue number: internal

---------

`ion-radio` does not emit the `ionSelect` event [anymore](https://github.com/ionic-team/ionic-framework/commit/a5229d90ca2a608e8bf4db0c8f71c86d481dd649#diff-46792eaa3f983b2b1dce0843e1f388a2d46705d7ed550fbfc8e95b5144861b16L58-L68). Angular still has a listener for this on radio.  The listener is not being for for anything,

## What is the new behavior?

- Removed the listeners on lazy and standalone
- Updated tests (for lazy) to verify that the `ion-radio` still works as intended. Standalone already tests `ion-radio`.

- [ ] Yes
- [x] No

## Other information

I verified that there are no build, test, etc issues for lazy and standalone. No issues found on ng14-16.

There were also no issues on the test app if a user is incorrectly using `ion-radio` with a form control:

```ts
// incorrect
<ion-radio-group>
   <ion-radio formControlName="radio" value="nes">Radio</ion-radio>
</ion-radio-group>
```


packages/angular/standalone/src/directives/proxies.ts
thetaPC, line 1: I'm not sure why this file was changed.

thetaPC:
I'm requesting a secondary approval since it's an Angular change.

liamdebeasi:
@sean-perkins Mind taking a look at this when you get a chance?

packages/angular/src/directives/control-value-accessors/radio-value-accessor.ts
```diff
@@ -17,13 +17,4 @@ export class RadioValueAccessorDirective extends ValueAccessor {
   constructor(injector: Injector, el: ElementRef) {
     super(injector, el);
   }
-
```
sean-perkins, line 20: Can you explain why you decided to remove the event listener vs. removing the value accessor implementation (directive) completely?

This change still allows `ion-radio` to be used as a form control, but it will never update a bound form control's value after the user selects it.

With the internal ticket, I believe the value accessor implementation for `ion-radio` can be removed completely, as long as we did not observe any issues with doing so. Developers should be binding their form controls to `ion-radio-group`, not `ion-radio`.

thetaPC: Should the explanation be placed within the code or the PR description?

thetaPC: Discussed via Slack

packages/angular/standalone/src/directives/radio.ts
```diff
@@ -4,7 +4,6 @@ import {
   Component,
```
sean-perkins, line 4: There are a few additional changes required to this file to align with removing the value accessor.

1. Remove the `providers` array registering this class as a value accessor:
```
providers: [
    {
      provide: NG_VALUE_ACCESSOR,
      useExisting: IonRadio,
      multi: true,
    },
  ],
```
2. `IonRadio` no longer needs to extend `ValueAccessor`
3. Remove the `super` call from the constructor
4. `Injector` appears to be unused once this updated, you can also remove that from the constructor
5. Remove unused imports as a result of these removals

sean-perkins (approved):

PR #28387, opened 2023-10-19
test(radio): skip Safari on legacy tab key tests

thetaPC:
Issue number: internal

---------

Legacy radio has tests that use `Tab` key presses. These tend to flake at unknown moments and were skipped until a fix can be implemented.

## What is the new behavior?

Unable to replicate the flakiness locally or on GitHub. However, it only fails on Safari so the tests were re-enabled except for Safari.

- [ ] Yes
- [x] No

## Other information

This only happens on legacy. The legacy form controls will also be removed in the future. Due to this, these tests will be removed anyways at that time.

### DIFF
--- a/core/src/components/radio/test/legacy/a11y/radio.e2e.ts
+++ b/core/src/components/radio/test/legacy/a11y/radio.e2e.ts
@@ -6,10 +6,11 @@ import { configs, test } from '@utils/test/playwright';
  */
 configs({ modes: ['ios'], directions: ['ltr'] }).forEach(({ title, config }) => {
   test.describe(title('radio: a11y'), () => {
-    // TODO(FW-5218)
-    test('tabbing should switch between radio groups', async ({ page, pageUtils }) => {
+    test.beforeEach(async ({ page, skip }) => {
+      skip.browser('webkit', 'Tabbing is flaky in Safari');
       await page.goto(`/src/components/radio/test/legacy/a11y`, config);
-
+    });
+    test('tabbing should switch between radio groups', async ({ page, pageUtils }) => {
       const firstGroupRadios = page.locator('#first-group ion-radio');
       const secondGroupRadios = page.locator('#second-group ion-radio');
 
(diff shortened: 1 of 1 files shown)


core/src/components/radio/test/legacy/a11y/radio.e2e.ts
```diff
@@ -6,10 +6,11 @@ import { configs, test } from '@utils/test/playwright';
  */
 configs({ modes: ['ios'], directions: ['ltr'] }).forEach(({ title, config }) => {
   test.describe(title('radio: a11y'), () => {
-    // TODO(FW-5218)
-    test.fixme('tabbing should switch between radio groups', async ({ page, pageUtils }) => {
+    test.beforeEach(async ({ page, skip }) => {
+      skip.browser('webkit', 'Tabbing is flaky in Safari');
```
amandaejohnston, line 10: You might need this fix? https://github.com/ionic-team/ionic-framework/blob/331c08aad542de158e53ed351705d4c396bb4e90/core/src/components/accordion/test/a11y/accordion.e2e.ts#L13

Since you were unable to replicate the flakiness, I would try adding the fix and re-enabling the test on all browsers. If it flakes again, we can always add the webkit skip back in.

amandaejohnston (changes requested):

core/src/components/radio/test/legacy/a11y/radio.e2e.ts
thetaPC: The `pressKeys` [function](https://github.com/ionic-team/ionic-framework/blob/331c08aad542de158e53ed351705d4c396bb4e90/core/src/components/radio/test/legacy/a11y/radio.e2e.ts#L16) has been customized to update the Tab if it's Webkit. So we [don't need to check](https://github.com/ionic-team/ionic-framework/blob/331c08aad542de158e53ed351705d4c396bb4e90/core/src/utils/test/press-keys.ts#L106) for browsers anymore.

amandaejohnston: Ah, forgot about that! Normally I'd still ask that we remove the webkit skip, but since this is a legacy test, it doesn't matter that much IMO.

amandaejohnston (approved):

PR #28396, opened 2023-10-23
refactor(utils): update slotMutationController to allow multiple slot names

amandaejohnston:
Issue number: Internal

---------

The internal `slotMutationController` utility can only watch one slot name at a time.

## What is the new behavior?

Added the option to specify multiple slot names, any of which will trigger the callback. This change is made in preparation for upcoming `start` and `end` slots for several form control components; it boosts performance by preventing us from needing to add multiple `createSlotMutationController` calls to each component.

- [ ] Yes
- [x] No


thetaPC (approved):
LGTM

liamdebeasi (approved):
Code looks good, but this should be a `refactor(util)` commit since it's an internal change.

amandaejohnston:
Ah, right, `perf` will show up in the changelog, huh 🤔 I figured `refactor` didn't fit since this is a behavior change (if an internal one) but it makes sense if the concern is more public API changes.

PR #28402, opened 2023-10-24
feat(toast): add swipe gesture implementation

liamdebeasi:
Issue number: N/A

---------

Toast cannot be swiped to dismiss.

## What is the new behavior?

 

- Implements a swipe gesture that allows the toast to be dismissed. This includes support for `positionAnchor`.
- Added a `getOffsetForMiddlePosition` util since we are doing this calculation in 3 different places.
- Added `OVERLAY_GESTURE_PRIORITY` so we can begin to get better consistency with our gesture priorities.

Note: There is a bug on iOS where swiping the toast and letting it snap back may cause a flicker. This is an underlying architecture bug that I plan to address in a separate PR. (I wasn't able to implement a fix today, but I felt it was important to get the other part of this implementation reviewed in the meantime)

⚠️ This PR focuses on the base functionality of the swipe gesture. This means things like the animation duration, the easing curve, the rubber band effect when dragging the wrong way are not part of this PR. I will address these separately. I wanted to allow reviewers to focus on the functionality before we start to tweak the parameters of this animation.

- [ ] Yes
- [x] No


core/src/components/toast/test/swipe-gesture/index.html
```diff
@@ -41,6 +56,24 @@
           message="Hello World"
           swipe-gesture="vertical"
         ></ion-toast>
+        <<<<<<< HEAD
```
brandyscarney, line 59: Looks like a bad merge here

liamdebeasi: Fixed, thanks! https://github.com/ionic-team/ionic-framework/pull/28402/commits/39f11763e2165e046d3e8dc8105228ec75894bab

core/src/components/toast/gestures/swipe-to-dismiss.ts
```diff
@@ -0,0 +1,283 @@
...
+     * value so the animation does not jump
+     * to the end when the user beings to drag.
+     */
+    .duration(100);
+
+  switch (el.position) {
+    case 'middle':
+      MAX_SWIPE_DISTANCE = hostElHeight + wrapperElBox.height;
+      swipeAnimation.keyframes(SWIPE_UP_DOWN_KEYFRAMES);
+      console.log(MAX_SWIPE_DISTANCE, SWIPE_UP_DOWN_KEYFRAMES, topPosition + wrapperElBox.height);
```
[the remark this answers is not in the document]
liamdebeasi: Oops yes I should have removed that 😂

thetaPC (approved):
LGTM, works great!

core/src/utils/overlays.ts
```diff
@@ -685,6 +685,7 @@ export const safeCall = (handler: any, arg?: any) => {
 
 export const BACKDROP = 'backdrop';
 export const GESTURE = 'gesture';
+export const OVERLAY_GESTURE_PRIORITY = 39;
```
sean-perkins, line 688: I like this approach - is the idea that future gesture priorities will be abstracted into this file?

sean-perkins (approved):

core/src/utils/overlays.ts
liamdebeasi: I don't know if it would makes sense to put all the gesture priorities in this file specifically, but I think long term it would be good to have them consolidated somewhere. Right now it's difficult to understand if changing a priority will cause other gestures used at the same time to break. Some of the gesture priorities seem pretty arbitrary, so I'd like to add some documentation around why they were chosen.

PR #28403, opened 2023-10-24
fix(item): apply safe area to proper side regardless of direction

mapsandapps:
Issue number: Internal

---------

The safe area for Item would change sides when the direction changed. e.g., if the safe area padding should be on the left, it would be on the left when the direction was LTR but would be on the right when the direction was RTL. It should always be on the left.

## What is the new behavior?

- The left safe area padding is on the left side of Item whether the direction is LTR or RTL.
- The right safe area padding is on the right side of Item whether the direction is LTR or RTL.
- There's now a screenshot test for left and right padding for both directions.

Note: There is a separate ticket for updating the safe area padding for ion-item-divider.

- [ ] Yes
- [x] No


core/src/components/item/item.scss
```diff
@@ -261,12 +268,19 @@ button, a {
...
-    calc(var(--ion-safe-area-right, 0px) + var(--inner-padding-end)),
-    var(--inner-padding-bottom),
-    var(--inner-padding-start)
-  );
+  @include padding(var(--inner-padding-top), null, var(--inner-padding-bottom));
+
+  /* stylelint-disable */
+  @include ltr() {
+    padding-right:calc(var(--ion-safe-area-right, 0px) + var(--inner-padding-end));
+    padding-left:var(--inner-padding-start);
```
amandaejohnston, line 276:
```suggestion
    padding-right: calc(var(--ion-safe-area-right, 0px) + var(--inner-padding-end));
    padding-left: var(--inner-padding-start);
```

amandaejohnston (approved):

core/src/components/item/item.scss
```diff
@@ -295,12 +309,18 @@ button, a {
 
 .item-bottom {
   @include margin(0);
-  @include padding(
-    0,
-    var(--inner-padding-end),
-    0,
-    calc(var(--padding-start) + var(--ion-safe-area-left, 0px))
-  );
+
```
brandyscarney, line 312: Was the top and bottom padding removed on purpose?

```scss
@include @padding(0, null, 0); 
```

core/src/components/item/item.scss
```diff
@@ -295,12 +309,18 @@ button, a {
...
+  /* stylelint-disable */
+  @include ltr() {
+    padding-left: calc(var(--padding-start) + var(--ion-safe-area-left, 0px));
+    padding-right: calc(var(--inner-padding-end) + var(--ion-safe-area-right, 0px));
+  }
+
+  @include rtl() {
+    padding-left: calc(var(--inner-padding-end) + var(--ion-safe-area-left, 0px));
+    padding-right: calc(var(--padding-start) + var(--ion-safe-area-right, 0));
```
brandyscarney, line 321:
```suggestion
    padding-right: calc(var(--padding-start) + var(--ion-safe-area-right, 0px));
```

core/src/components/item/test/basic/item.e2e.ts-snapshots/safe-area-ios-rtl-Mobile-Chrome-linux.png
brandyscarney, line 1: I think we might be applying the `--ion-safe-area-left` to the wrong element here. This might have been a problem prior to this PR but maybe we should look into it? 

If you notice on the LTR test the bottom border extends the full width to the right, even though `--ion-safe-area-right` is set to `10px`: 

image

When flipped, the bottom border is not extending fully to the left edge:

image

Happy to bring this up for team discussion to see if we should make this a separate issue!

brandyscarney (changes requested):
This is looking good, just a couple of things I found that I wanted to bring up!

liamdebeasi:
Removing myself as a reviewer. I'm getting added as a code-owner due the screenshot changes, but I think Amanda and Brandy have this covered 😄

core/src/components/item/test/basic/item.e2e.ts-snapshots/safe-area-ios-rtl-Mobile-Chrome-linux.png
mapsandapps: I think this was actually a mistake with this PR, not from before. Thanks for catching it!

core/src/components/item/test/basic/item.e2e.ts
```diff
@@ -10,5 +10,37 @@ configs().forEach(({ title, screenshot, config }) => {
...
+     * This behavior needs to be tested for all modes & directions
+     * Safe padding should stay on the same side when the direction changes
+     */
+    test('should have safe area padding', async ({ page }) => {
+      await page.setContent(
+        `
+        <style>
+          :root {
+            --ion-safe-area-left: 40px;
+            --ion-safe-area-right: 10px;
```
brandyscarney, line 24:
```suggestion
            --ion-safe-area-left: 40px;
            --ion-safe-area-right: 20px;
```

Nit pick but we have used these values in a few other places, could we update them to match? 

https://github.com/ionic-team/ionic-framework/blob/b7d1a5c86bf81b0808c343f35ba64214dba39e6f/core/src/components/list-header/test/basic/list-header.e2e.ts#L29-L30
https://github.com/ionic-team/ionic-framework/blob/b7d1a5c86bf81b0808c343f35ba64214dba39e6f/core/src/components/fab/test/safe-area/fab.e2e.ts#L11-L12

core/src/components/item/test/basic/item.e2e.ts
```diff
@@ -10,5 +10,37 @@ configs().forEach(({ title, screenshot, config }) => {
...
+            <ion-label> Single line text that should have ellipses when it doesn't all fit in the item</ion-label>
+          </ion-item>
+        </ion-list>
+      `,
+        config
+      );
+
+      const list = page.locator('ion-list');
+
+      await expect(list).toHaveScreenshot(screenshot('safe-area'));
```
brandyscarney, line 43:
```suggestion
      await expect(list).toHaveScreenshot(screenshot('item-safe-area'));
```

The test name should start with the component name. You will need to delete the screenshots that were added and run the update action again or rename them.

brandyscarney (commented):
The code itself looks good, great work! Just some minor feedback on the test.

brandyscarney (approved):

PR #28412, opened 2023-10-25
chore(dev-deps): upgrade to jest v29

rwaskiewicz:
Issue number: #

---------

Ionic Framework uses Jest 27. Previously, this was due to a limitation of Stencil. As of Stencil v4.7.0, this limitation no longer exists & the Framework can upgrade to the latest version of Jest.

## What is the new behavior?

There was a top level dependency on `@jest/core` that I could not find to be used anywhere. This dependency is included in `jest` by default, and has been removed as a top level/`package.json` dependency.

Otherwise, all jest packages have been upgraded to jest v29. This is two major version bumps and a switch to the jest-circus runtime (instead of jest-jasmine2). 

- [ ] Yes
- [x] No

## Other information
There is some inherent risk of _something_ breaking when someone like me who doesn't work on this project day-to-day bumps packages. I did this as a part of testing Jest 29 when implementing its support in Stencil and thought "Well, why not?". Feel free to reject this PR if you feel it's not worth the risk/effort at the moment.


liamdebeasi (approved):
I'm fine with this change. Our Jest tests currently pass, and the risk for Ionic developers seem pretty minimal since this is test-related.

Thanks for the PR!

PR #28415, opened 2023-10-25
fix(overlays): prevent scroll gestures when the overlay is presented

### LINKED ISSUE #23942: bug: overlays block scroll as soon as they are created, even if they do not get presented
### Prequisites

### Ionic Framework Version

- [ ] v4.x
- [X] v5.x
- [ ] v6.x

### Current Behavior

When I create a new Alert using an Alert controller like 

```typescript
  constructor(private _alertController: AlertController) {

      const alert = await this._alertController.create({
        header: `Just Header`,
        message: `Just a Question`,
        buttons: [
          {
            text: 'Cancel',
            role: 'cancel',
            cssClass: 'secondary'
          },
          {
            text: 'Continue',
            handler: () => {
              this.doSomething();
            }
          }
        ]
      });
```

class `backdrop-no-scroll` got added to the `body` and preventing scroll in the web version of Ionic.

### Expected Behavior

That class should be added after I call `alert.present()` I think, not while creating the alert, as for example, I can initialize the alert and `alert.present()` in a specific condition, instead of having all creation and presenting in the same code block.

### Steps to Reproduce

Just create an alert and don't call it or make a condition to call it.
```
      const alert = await this._alertController.create({
        header: `Just Header`,
        message: `Just a Question`,
        buttons: [
          {
            text: 'Cancel',
            role: 'cancel',
            cssClass: 'secondary'
          },
          {
            text: 'Continue',
            handler: () => {
              this.doSomething();
            }
          }
        ]
      });

      if (condition) {
        alert.present()
      }
```

### Code Reproduction URL

_No response_

### Ionic Info

Ionic:

   Ionic CLI                     : 6.17.1
   Ionic Framework               : @ionic/angular 5.8.0
   @angular-devkit/build-angular : 12.2.6
   @angular-devkit/schematics    : 12.2.6
   @angular/cli                  : 12.2.6
   @ionic/angular-toolkit        : 4.0.0

Cordova:

   Cordova CLI       : 10.0.0

Utility:

   native-run  : 1.4.1

System:

   NodeJS : v16.4.1 (/usr/bin/node)
   npm    : 7.23.0
   OS     : Linux 5.10

### Additional Information

_No response_

sean-perkins:
Issue number: Resolves #23942

---------

When an overlay is created (inserted in the DOM), but not presented, the scroll gesture is prevented. This behavior comes from the `connectedCallback` of `ion-backdrop`, where the gesture is prevented as soon as the backdrop is inserted in the DOM. 

This means in situations where a developer creates an overlay, but does not present it immediately, the user cannot scroll. This is not desired. 

## What is the new behavior?

- Scroll blocking behavior tied to the gesture has been removed from `ion-backdrop` and implemented into the overlays directly.
  - When an overlay is presented, scroll blocking is enabled on the `body` element (the user cannot scroll on the main content).
  - When the last presented overlay is dismissed, scroll blocking is disabled on the `body` element (the user can scroll on the main content).

- [x] Yes
- [ ] No

`ion-backdrop` no longer prevents scrolling on the main content when the backdrop is either inserted into the DOM or removed from the DOM. Developers using Ionic overlays do not need to migrate their implementations.

Developers with custom overlays using `ion-backdrop` internally can either use Ionic's gesture controller to disable scrolling when their overlay is presented/dismissed or can manually add the `backdrop-no-scroll` Ionic global class to the `body` element.


core/src/utils/test/overlays/overlays.e2e.ts
```diff
@@ -104,6 +104,53 @@ configs({ modes: ['ios'], directions: ['ltr'] }).forEach(({ title, config }) =>
...
+      const contentEl = page.locator('#main-content ion-content');
+
+      await expect(
+        await contentEl.evaluate(async (el: HTMLIonContentElement) => {
+          const contentScrollEl = await el.getScrollElement();
+          return contentScrollEl.scrollTop;
+        })
+      ).toBe(0);
+
+      await page.mouse.wheel(0, 100);
```
sean-perkins, line 126: I'm not entirely happy with the way that I had to test scrolling inside Playwright here. The `page.mouse.wheel` API is unavailable in mobile webkit/Safari. However any other emulation of adjusting `scrollTop`, or using `scrollIntoViewWhenNeeded` does not actually perform a scroll, just manually adjusts the scroll position.

core/src/components/loading/test/loading-id.spec.ts
```diff
@@ -1,10 +1,11 @@
 import { newSpecPage } from '@stencil/core/testing';
 
+import { Backdrop } from '../../backdrop/backdrop';
 import { Loading } from '../loading';
 
 it('loading should be assigned an incrementing id', async () => {
   const page = await newSpecPage({
-    components: [Loading],
+    components: [Loading, Backdrop],
```
sean-perkins, line 8: Backdrop was added to all affected unit tests (when the overlay is presented/dismissed), as the API for `.block()` and `.unblock()` causes an exception and a test failure if the component is not imported/defined.

core/src/utils/test/overlays/overlays.e2e.ts
```diff
@@ -255,4 +302,65 @@ configs({ modes: ['ios'], directions: ['ltr'] }).forEach(({ title, config }) =>
       await expect(modalInputOne).toBeFocused();
     });
   });
+
+  test.describe(title('overlays: create'), () => {
```
liamdebeasi, line 306: `mouse.wheel` doesn't wait for scrolling to finish before returning (https://playwright.dev/docs/api/class-mouse#mouse-wheel) so this might be flaky. As an alternative (and potential spec test) are we able to check the following:

1. Root does not have `overflow: hidden` evaluated styles when overlay is hidden
2. Root does have `overflow: hidden` evaluated styles when overlay is presented

liamdebeasi: I'm not sure if this is much better, but using https://developer.mozilla.org/en-US/docs/Web/API/Window/getComputedStyle could help. At the very least it would avoid any scrolling flakiness.

sean-perkins: Good thoughts! I'll try them out!

core/src/components/backdrop/backdrop.tsx
```diff
@@ -37,14 +37,22 @@ export class Backdrop implements ComponentInterface {
    */
   @Event() ionBackdropTap!: EventEmitter<void>;
 
-  connectedCallback() {
+  @Element() el!: HTMLElement;
+
+  /** @internal */
+  @Method()
+  async block() {
```
amandaejohnston, line 44: > This does change the behavior for using `ion-backdrop` with a custom implementation (outside of Ionic's overlays). Previous to this change, scrolling would be blocked when the backdrop was inserted in the DOM and enabled when removed from the DOM. After this change, that behavior will not occur and the scroll gesture will not be prevented unless the developer uses the internal API methods that our overlay implementation uses.
> 
> I don't consider this a breaking change since it is internal behavior (not an API breakage).

(Threading) I'm confused about this; doesn't this mean that people can no longer block scrolling with `ion-backdrop` if using it as part of a custom implementation? We do have demos for standalone backdrops in the docs -- https://ionicframework.com/docs/api/backdrop -- so I assume this is a supported use case that we would be breaking.

sean-perkins: The backdrop examples demonstrate blocking click interactions on the content behind, they do not mention anything regarding scroll blocking:

> The backdrop prevents clicking or tapping on the content behind it. 

This change would effect developers that have built custom overlays (not with Ionic), are using `ion-backdrop` within that implementation and are using Ionic's global styles to have `backdrop-no-scroll`. However they could use the internal methods, reflect the classes themselves when they present their overlay or we could decide to make those methods public API as part of a minor release. 

The scroll blocking behavior within `ion-backdrop` is largely specific to Ionic's overlay implementations. To support both use cases, we would need to be modifying the DOM multiple times to disable overflow when the backdrop is inserted into the DOM, enable overflow when the overlay is constructed and then disable overflow when the overlay is presented. I believe this is an unnecessary performance cost for the majority of Ionic use cases.

sean-perkins:
I think I need to rethink this approach a little in general - especially if multiple overlays are presented on the screen. I need to only re-enable scroll if there is _not_ a presented overlay/it is the last overlay. I'm going to work on that + investigate an approach that doesn't break advanced usages.

core/src/components/action-sheet/action-sheet.tsx
```diff
@@ -366,7 +367,7 @@ export class ActionSheet implements ComponentInterface, OverlayInterface {
         onIonActionSheetWillDismiss={this.dispatchCancelHandler}
         onIonBackdropTap={this.onBackdropTap}
       >
-        <ion-backdrop tappable={this.backdropDismiss} />
+        <ion-backdrop ref={(el) => (this.backdropEl = el)} tappable={this.backdropDismiss} />
```
sean-perkins, line 370: Looking at these changes you might ask "why?" or "where is this used?". 

The current overlay utils for [`present`](https://github.com/ionic-team/ionic-framework/blob/8e88269d8ee173cc89e578a9db79e0013b4669a4/core/src/utils/overlays.ts#L462-L468) and [`dismiss`](https://github.com/ionic-team/ionic-framework/blob/8e88269d8ee173cc89e578a9db79e0013b4669a4/core/src/utils/overlays.ts#L549C1-L557C2) accept an object that configures the util implementation. In our existing usage, [we pass the entire class instance (`this`) to the function](https://github.com/ionic-team/ionic-framework/blob/8e88269d8ee173cc89e578a9db79e0013b4669a4/core/src/components/action-sheet/action-sheet.tsx#L207). By adding the new `backdropEl` property on the class and assigning the reference in the `render()` function, the util can access it from the implementation without additional configuration. 

This logic applies to all overlays.

liamdebeasi: Out of curiosity, did we consider adding the class directly in the overlay.ts file? The `dismiss` function already checks to see if this is the last overlay we are dismissing. As possible implementation could be:

```js
import { BACKDROP_NO_SCROLL } from '@utils/gesture/gesture-controller';
const present = (...) => {
  document.body.classList.add(BACKDROP_NO_SCROLL);
  ...
}

const dismiss = (...) => {
  if (doc !== undefined && getPresentedOverlays(doc).length === 1) {
    setRootAriaHidden(false);
    document.body.classList.remove(BACKDROP_NO_SCROLL);
  }
}
```

The benefit of this is we don't need to worry about getting and clearing references to the backdrop element by grabbing inside the component instance.

I guess part of this is I don't totally understand why we need to create a new blocker for every overlay presented. I would have expected we'd only block scroll once.

sean-perkins: Yeah it looks like we could simplify and avoid getting the backdrop element reference by always blocking/unblocking the scroll as part of the overlay implementation. 

Originally I had thought that the behavior was dependent on the overlay having a backdrop, but they all technically render an `ion-backdrop`, it may just not be visible to the user. 

For the scroll blocking implementation we could either reflect the class ourselves like your code or use the `BlockerDelegate` (which does the same thing, but with an API). The only thing inside Ionic Framework that uses this, is this implementation. I'd be open to killing this API in favor of reflecting the class directly - if you agree.

liamdebeasi: I'm fine using the class directly, but lets hold off on killing it entirely because I believe the menu still uses it (even though it doesn't work on iOS, it still works on Android I believe)

liamdebeasi: Actually looks like it's part of the gesture public API, so we'll need to keep it regardless: https://github.com/ionic-team/ionic-framework/blob/89698b338fb05cde427c98720c238d2365abdaa7/core/src/utils/gesture/index.ts#L308

core/src/utils/overlays.ts
```diff
@@ -574,6 +579,11 @@ export const dismiss = async <OverlayDismissOptions>(
     if (role !== GESTURE) {
       await overlayAnimation(overlay, animationBuilder, overlay.el, opts);
     }
+
+    if (lastOverlay) {
+      document.body.classList.remove(BACKDROP_NO_SCROLL);
```
liamdebeasi, line 584: Out of curiosity, are we able to remove the class when we do `setRootAriaHidden(false)`? The underlying content shouldn't be scrollable since the backdrop will still catch pointer events until it's been removed from the DOM.

liamdebeasi (approved):
I'm good with the change as-is, just one question.

core/src/utils/overlays.ts
sean-perkins: Confirmed that we can! Originally I thought about avoiding the overflow layout repaint until after the dismiss transition, but it doesn't seem to make an observable impact to do it prior.

PR #28420, opened 2023-10-26
fix(item-divider): apply safe area to proper side regardless of direction

mapsandapps:
Issue number: Internal

---------

The safe area for Item Divider would change sides when the direction changed. e.g., if the safe area padding should be on the left, it would be on the left when the direction was LTR but would be on the right when the direction was RTL. It should always be on the left.

## What is the new behavior?

- The left safe area padding is on the left side of an Item Divider whether the direction is LTR or RTL.
- The right safe area padding is on the right side of an Item Divider whether the direction is LTR or RTL.
- Updated LTR and RTL screenshots for item to include an item divider

Similar to #28403 but for Item Divider.

- [ ] Yes
- [x] No


core/src/components/item-divider/item-divider.scss
```diff
@@ -30,13 +30,24 @@
   --inner-padding-start: 0px;
 
   @include font-smoothing();
+
+  /* stylelint-disable */
+  @include ltr() {
+    padding-top: var(--padding-top);
```
brandyscarney, line 36: Padding top and bottom don't need to be directional. You should be able to move those above the `ltr`/`rtl` mixins like:

```scss
@include padding(var(--padding-top), null, var(--padding-bottom), null);
```
```

File: core/src/components/item-divider/item-divider.scss [hunk 2]
```diff
@@ -31,12 +31,19 @@
 
   @include font-smoothing();
   @include margin(0);
-  @include padding(
-    var(--padding-top),
-    var(--padding-end),
-    var(--padding-bottom),
-    calc(var(--padding-start) + var(--ion-safe-area-left, 0px))
-  );
+  @include padding(var(--padding-top), null, var(--padding-bottom));
```
liamdebeasi, line 34: Nit: Can we add `null` at the end? I know it doesn't make any functional difference, but it threw me off only seeing 3 values here.
```suggestion
  @include padding(var(--padding-top), null, var(--padding-bottom), null);
```

core/src/components/item-divider/item-divider.scss
```diff
@@ -67,12 +74,19 @@
 
 .item-divider-inner {
   @include margin(0);
-  @include padding(
-    var(--inner-padding-top),
-    calc(var(--ion-safe-area-right, 0px) + var(--inner-padding-end)),
-    var(--inner-padding-bottom),
-    var(--inner-padding-start)
-  );
+  @include padding(var(--inner-padding-top), null, var(--inner-padding-bottom));
```
liamdebeasi, line 77: Nit: Can we add `null` at the end? I know it doesn't make any functional difference, but it threw me off only seeing 3 values here.
```suggestion
  @include padding(var(--inner-padding-top), null, var(--inner-padding-bottom), null);
```

core/src/components/item/test/basic/item.e2e.ts
```diff
@@ -30,6 +30,10 @@ configs().forEach(({ title, screenshot, config }) => {
             <div slot="helper">Helper</div>
           </ion-item>
 
+          <ion-item-divider>
```
liamdebeasi, line 33: Optional: I think it's confusing to add the item-divider to the item tests when item-divider has its own test directory in `item-divider/test`. I'll defer to you and Brandy though.

liamdebeasi (approved):
A few small changes, but otherwise looks good.

core/src/components/item/test/basic/item.e2e.ts
mapsandapps: My thinking was 

1. That it would be easier to understand what regression was occurring if Item & Item Divider are in the same test
2. It's more efficient to combine them into the same test
3. There are other tests in Item that include Item Dividers

but i can move them. It doesn't really matter. 🤷🏻

core/src/components/item-divider/test/basic/item-divider.e2e.ts
```diff
@@ -46,5 +46,32 @@ configs().forEach(({ title, screenshot, config }) => {
...
+        `
+        <style>
+          :root {
+            --ion-safe-area-left: 40px;
+            --ion-safe-area-right: 20px;
+          }
+        </style>
+        <ion-list>
+          <ion-item-divider>
+            <ion-label>Item Divider</ion-label>
```
brandyscarney, line 65:
```suggestion
            <ion-label>Item Divider</ion-label>
            <ion-button slot="end">Button</ion-button>
```

Could we add a button to the test to further verify the safe area is working when flipping?

brandyscarney (approved):

PR #28422, opened 2023-10-26
fix(angular,vue): range form value updates while dragging knob

### LINKED ISSUE #28256: bug: Range should update its value in forms while dragging
### Prerequisites

### Ionic Framework Version

v7.x

### Current Behavior

Currently the `ion-range` component subscribes to the `ionChange` event to update it's value when participating in a form:
- Angular: https://github.com/ionic-team/ionic-framework/blob/51b7ceb5bef81dc5367e313c96e8937483c06753/packages/angular/src/directives/control-value-accessors/select-value-accessor.ts#L6-L32
- Vue: https://github.com/ionic-team/ionic-framework/blob/51b7ceb5bef81dc5367e313c96e8937483c06753/core/stencil.config.ts#L138-L143

This is confusing to users and limits how the `ion-range` component can be used in applications. A use case could be a volume control that you want to immediately adjust the setting as a user drags on the range. This currently is not possible without manually subscribing to the `ionInput` value and assigning it to the form.

### Expected Behavior

`ion-range` should subscribe to the `ionInput` event instead of `ionChange` so that the value is immediately update as the user interacts with the control.

### Steps to Reproduce

1. Open the reproduction link
2. Drag the range knob and do not release
3. Observe: As you drag, the value does not update.
4. Release the range knob
5. Observe: The value updates to the knob position

### Code Reproduction URL

https://stackblitz.com/edit/angular-2nzaub?file=src%2Fapp%2Fexample.component.html,src%2Fapp%2Fapp.module.ts,src%2Fapp%2Fexample.component.ts

### Ionic Info

N/A

### Additional Information

_No response_

amandaejohnston:
Issue number: Resolves #28256

---------

In the form integrations for Angular and Vue, the value of a range does not update while the knob is actively being dragged, only when the knob is released.

## What is the new behavior?

The form integrations now update the range's value when the `ionInput` event fires, rather than `ionChange`.

- [ ] Yes
- [x] No

## Other information

I wasn't sure how to add reliable automated tests for this behavior. The difference only applies when actively dragging the knob, and we've had issues with such gestures being flaky in the past. I did add value displays to the test apps so the behavior can be manually tested.


packages/angular/test/base/src/app/lazy/form-controls/range/range.component.html
```diff
@@ -11,6 +11,7 @@
           <div slot="label">Range</div>
         </ion-range>
       </ion-item>
+      <p>Value: {{ form.controls['range'].value }}</p>
```
amandaejohnston, line 14: While talking to Sean about testing this, he had the idea to manually dispatch an `ionInput` event and check if the value changes, but that doesn't work here because the value is pulled directly from the component, which naturally doesn't update.

packages/angular/src/directives/control-value-accessors/text-value-accessor.ts
```diff
@@ -3,7 +3,7 @@ import { NG_VALUE_ACCESSOR } from '@angular/forms';
 import { ValueAccessor } from '@ionic/angular/common';
 
 @Directive({
```
sean-perkins, line 5: ~ We should probably consider refactoring this class in the next major release. It isn't a `TextValueAccessor` if we are using it for `ion-range`. If we are going to group all value accessors with a shared event, we might want to consider a `IonInputValueAccessor` class instead. Changing the class name will be a breaking change since this class can be used for dependency injection in Angular apps. 

Non-blocking for this PR, but something to consider.

sean-perkins (approved):

sean-perkins:
Oh one last thought on the PR subject/commit subject - should we change it to:

> fix(angular,vue): range form value updates while dragging

I think that would be organized better in the CHANGELOG.

liamdebeasi (approved):
Tested in both Angular and Vue -- works great!

packages/angular/src/directives/control-value-accessors/text-value-accessor.ts
amandaejohnston: I added a tech debt ticket for this 👍

PR #28439, opened 2023-10-31
fix(list): remove border from last item with item-sliding

### LINKED ISSUE #28435: bug: last `ion-item-sliding` should not have a `border-bottom`
### Prerequisites

### Ionic Framework Version

v7.x

### Current Behavior

The last item has a border bottom if they are `ion-item-sliding` but not if they are `ion-item`.
Screenshot 2023-10-30 at 3 02 13 PM

### Expected Behavior

The last item should not have a border bottom if they are `ion-item-sliding`, just like `ion-item`.
Screenshot 2023-10-30 at 3 02 06 PM

### Steps to Reproduce

Make two inset list, one with `ion-item-sliding` and one with regular `ion-item`.
The ion-item-sliding one will have a border-bottom on the last element, while the ion-item list won't have a border-bottom on the last element.

### Code Reproduction URL

https://stackblitz.com/run?file=package.json

### Ionic Info

Ionic:

   Ionic CLI       : 7.1.1 (/Users/sleblanc/.nvm/versions/node/v16.19.1/lib/node_modules/@ionic/cli)
   Ionic Framework : @ionic/vue 7.5.0

Capacitor:

   Capacitor CLI      : 5.5.0
   @capacitor/core    : 5.5.0
   @capacitor/ios     : 5.5.0

Utility:

   native-run (update available: 1.7.4) : 1.7.3

System:

   NodeJS : v16.19.1 (/Users/sleblanc/.nvm/versions/node/v16.19.1/bin/node)
   npm    : 8.19.3
   OS     : macOS Ventura 14.1

### Additional Information

_No response_

liamdebeasi:
Issue number: resolves #28435

---------

The item in the last item-sliding still has a border in an inset list.

## What is the new behavior?

- Item in last item-sliding no longer has a border

I originally only added `ion-item-sliding:last-of-type ion-item` but I discovered that the original `ion-item:last-child` causes items in item-sliding where the item is the last element in the item-sliding container to not have a border, so the original fix was incomplete.

I added comments as to what each line does and why we didn't just do `ion-item:last-child`.

- [ ] Yes
- [x] No

## Other information

--------

Co-authored-by: Brandy Carney <brandyscarney@users.noreply.github.com>


core/src/components/list/list.ios.scss
```diff
@@ -19,12 +19,25 @@
...
+ * These selectors ensure the first item in the list
+ * has the correct border.
+ * We need to consider the following scenarios:
+ 1. The only item in a list.
+ 2. The last item in a list as long as it is not the only item.
+ 3. The item in the last item-sliding in a list.
+ * Note that we do not select ion-item:last-of-type
+ * because that will cause the borders to disappear for
+ * when for items in item-sliding where the item is the last
+ * element in the item-sliding container.
```
brandyscarney, line 32:
```suggestion
 * Note that we do not select ion-item:last-of-type
 * because that will cause the borders to disappear on
 * items in an item-sliding when the item is the last
 * element in the item-sliding container.
```

core/src/components/list/list.md.scss
```diff
@@ -23,12 +23,40 @@
...
+ * These selectors ensure the first item in the list
+ * has the correct radius.
+ * We need to consider the following scenarios:
+ 1. The only item in a list.
+ 2. The first item in a list as long as it is not the only item.
+ 3. The item in the first item-sliding in a list.
+ * Note that we do not select ion-item:first-of-type
+ * because that will cause the borders to disappear for
+ * when for items in item-sliding where the item is the last
+ * element in the item-sliding container.
```
brandyscarney, line 36:
```suggestion
 * Note that we do not select ion-item:first-of-type
 * because that will cause the borders to disappear on
 * items in an item-sliding when the item is the first
 * element in the item-sliding container.
```

core/src/components/list/list.md.scss
```diff
@@ -23,12 +23,40 @@
...
+ * These selectors ensure the first item in the list
+ * has the correct radius and border.
+ * We need to consider the following scenarios:
+ 1. The only item in a list.
+ 2. The last item in a list as long as it is not the only item.
+ 3. The item in the last item-sliding in a list.
+ * Note that we do not select ion-item:last-of-type
+ * because that will cause the borders to disappear for
+ * when for items in item-sliding where the item is the last
+ * element in the item-sliding container.
```
brandyscarney, line 54:
```suggestion
 * Note that we do not select ion-item:last-of-type
 * because that will cause the borders to disappear on
 * items in an item-sliding when the item is the last
 * element in the item-sliding container.
```

core/src/components/list/test/lines/list.e2e.ts
```diff
@@ -26,3 +26,154 @@ configs().forEach(({ title, screenshot, config }) => {
...
   });
 });
+
+/**
+ * Padding and border color ensures the bottom border can be easily seen if it regresses.
+ * The background color ensures that any border radius values can be seen.
+ */
+configs({ directions: ['ltr'] }).forEach(({ title, screenshot, config }) => {
+  test.describe(title('list: lines with children'), () => {
+    test('only item in inset list should not have line', async ({ page }) => {
```
brandyscarney, line 36: Channeling my inner Maria to ask for a test annotation. 🙂

liamdebeasi: sorry I'll add that 😂

core/src/components/list/list.ios.scss
```diff
@@ -19,12 +19,25 @@
   @include border-radius($list-inset-ios-border-radius);
 }
 
-.list-ios.list-inset ion-item:last-child {
+/**
+ * These selectors ensure the first item in the list
```
amandaejohnston, line 23:
```suggestion
 * These selectors ensure the last item in the list
```
At least, I think this is what you meant? Ditto for the other instances of this comment.

core/src/components/list/list.ios.scss
```diff
@@ -19,12 +19,25 @@
...
+ * We need to consider the following scenarios:
+ 1. The only item in a list.
+ 2. The last item in a list as long as it is not the only item.
+ 3. The item in the last item-sliding in a list.
+ * Note that we do not select ion-item:last-of-type
+ * because that will cause the borders to disappear on
+ * items in an item-sliding when the item is the last
+ * element in the item-sliding container.
+ */
+.list-ios.list-inset ion-item:only-child,
```
amandaejohnston, line 34: Shouldn't this be `:only-of-type`? The other lines in this rule use `-type` selectors instead of `-child` ones, implying that we still want to remove the border even if there are other, non-item elements under the last item. Ditto for the other instances of this pattern.

core/src/components/list/list.md.scss
```diff
@@ -23,12 +23,40 @@
...
+ * We need to consider the following scenarios:
+ 1. The only item in a list.
+ 2. The last item in a list as long as it is not the only item.
+ 3. The item in the last item-sliding in a list.
+ * Note that we do not select ion-item:last-of-type
+ * because that will cause the borders to disappear on
+ * items in an item-sliding when the item is the last
+ * element in the item-sliding container.
+ */
+.list-md.list-inset ion-item:only-child,
```
amandaejohnston, line 56: This piece is also present in the selector right above this, so the styles are getting set and then immediately overwritten.

amandaejohnston (changes requested):

core/src/components/list/list.ios.scss
liamdebeasi: Fixed in https://github.com/ionic-team/ionic-framework/pull/28439/commits/127e374bf65417bdadb178fff9378947257ca5e7

core/src/components/list/list.md.scss
liamdebeasi: Good catch! Fixed in https://github.com/ionic-team/ionic-framework/pull/28439/commits/1c83d15b38520144484229a245c544afc8fe2460.

core/src/components/list/list.ios.scss
liamdebeasi: This rule is targeting when there is only 1 item inside of a list (Regardless of item-sliding usage). Doing `ion-item:only-of-type` causes lines to disappear when the item is using in `ion-item-sliding`.

Example: 
image

Markup:

```html
<ion-list inset="true">
  <ion-item-sliding>
    <ion-item-options slot="start">
      <ion-item-option color="warning">
        <ion-icon slot="icon-only" name="pin"></ion-icon>
      </ion-item-option>
    </ion-item-options>
    <ion-item>
      <ion-label>Item 2</ion-label>
    </ion-item>
  </ion-item-sliding>
</ion-list>
```

brandyscarney: I created a Codepen of this if it helps you visualize it better: https://codepen.io/brandyscarney/pen/MWLewmr

amandaejohnston: Ahh, makes sense 👍

amandaejohnston (approved):

core/src/components/list/list.ios.scss
liamdebeasi: I added a comment in https://github.com/ionic-team/ionic-framework/pull/28439/commits/448b88c0a29cccb8989d2caea79a60522d0abdb4 to clarify

core/src/components/list/list.md.scss
```diff
@@ -23,12 +23,46 @@
...
+  --border-width: 0;
+  --inner-border-width: 0;
+}
+
+/**
+ * The only item in a list should have a border radius
+ * on all corners.
+ */
+.list-md.list-inset ion-item:only-child {
+  --border-radius: #{$list-inset-md-border-radius};
```
brandyscarney, line 69: This change seems broken to me:

| main | branch |
| ---| ---|
| main(Pixel 5) | branch(Pixel 5) |

Steps to reproduce:

1) Modify `core/src/components/list/test/inset/index.html` to:

```diff
- <ion-content id="content" color="light">
+ <ion-content id="content" color="tertiary">
+   <ion-list inset>
+     <ion-item>
+       <ion-label>Text Items</ion-label>
+     </ion-item>
+   </ion-list>
+
    <ion-list inset>
```

Add the following `style` to the `head`:

```html
<style>
  ion-item {
    --background: red;
  }
</style>
```

Increase the border radius on `md` to see the change better:

```scss
$list-inset-md-border-radius:          10px !default;
```

liamdebeasi: I'm just following what was there before for border radius: https://github.com/ionic-team/ionic-framework/blob/a4551470a9700c3278d0005f083e7a8b6ddd04b5/core/src/components/list/list.md.scss#L26-L33

Demo: https://codepen.io/liamdebeasi/pen/eYxzXjw

Note that the last-child's border radius never showed up because it was formatted incorrectly: 
image

(either everything should have a comma, or nothing should)

brandyscarney: Hmm okay. I think this might be broken behavior then on `md`. This code used to be required when the items lined up with the top / bottom of the list but since [Material Design added padding](https://m2.material.io/components/lists#specs) to the top and bottom of a list it doesn't line up properly. It looks like we added that in [5 years ago](https://github.com/ionic-team/ionic-framework/commit/e416c23b3b946f34b0e54d4d236991727dc353a2#diff-a62a9570aa3f44e50c23e047c148e994010732cf3af17f9257672d0f68b30436R20) but didn't account for the border radius since item generally has the same background as list. 

[Material Design 3](https://m3.material.io/components/lists/specs#ee01024f-ea3b-47f4-81af-4f90d12691d6) also has this padding at the top and bottom of their list. Should we create a tech debt ticket to take a look at this or add it to our MD3 work? 

It does look like it should still have this code on `ios` though. We also don't seem to have any tests verifying this behavior.

brandyscarney (approved):
Code looks good, just need to create or update a ticket to take a look at the border radius for items in an inset list on `md`. 👍

Issue #26297, opened 2022-11-16
feat: standalone input component should support icons


EinfachHans:
### Prerequisites

- [X] I have read the [Contributing Guidelines](https://github.com/ionic-team/ionic-framework/blob/main/.github/CONTRIBUTING.md#creating-an-issue).
- [X] I agree to follow the [Code of Conduct](https://ionicframework.com/code-of-conduct).
- [X] I have searched for [existing issues](https://github.com/ionic-team/ionic-framework/issues) that already include this feature request, without success.

### Describe the Feature Request

This is a Feature Request to v7 and #26283 

To fully be able to get rid of the deprecated usage of form elements within the ion item it would be great to support icon's within these elements as well.

### Describe the Use Case

This are inputs in our current app:

![D0BB0629-2C4C-4877-B78F-B8F23BF8D3D1](https://user-images.githubusercontent.com/20899975/202202968-8ddd1080-e253-4054-a874-10e08858dd4e.jpeg)

Many Password Input's have a toggle icon to show the password at the end. Also a start icon can be great to personalize the form element more. 

### Describe Preferred Solution

I would prefer a slot solution where deverlopers are able to place something at the start or end slot, similar to the item. With this it would be easier to add a click handler to the icon (to for example toggle the input mode to show or hide a password)

Also it should be somehow possible to not focus the element on icon click. If i think of the password toggle icon, a click on that icon should not focus the element.

### Describe Alternatives

An alternative could be two properites `startIcon` and `endIcon`, but i don't really think this is the right way, as then there also should be two events for clicking these.

TiBz0u:
Hey @EinfachHans , thanks for your this feature request.
For my information, how do you achieve this design currently?

I full agree with the necessity of icon event (ie: password reveal, tooltip, data masking, ...)

Regarding design of input with icon, it often refers to "inputGroup". Maybe it must be separated in a new component.
https://getbootstrap.com/docs/5.2/forms/input-group/
https://www.primefaces.org/primeng/inputgroup
https://www.primefaces.org/primeng/inputnumber

Another nice feature could be a picker in front of the input (ie: country selector for phone number).

Kr

WolfWalter:
If this gets implemented, it would be great if it would also be possible to add multiple icons at the end. The use case for that is a password field with a show/hide button and a clear button.

SvenBudak:
I was also very surprised today that there is no slot for buttons in inputs. Are there any updates on this?

chrisk8er:
[...]

PR #28441, opened 2023-10-31
feat(textarea): add start and end slots

amandaejohnston:
Issue number: Part of #26297

---------

With the modern form control syntax, it is not possible to add icon buttons or other decorators to the sides of `ion-textarea`, as you can with `ion-item`.

## What is the new behavior?

`start` and `end` slots added. While making this change, I also tweaked the CSS selectors responsible for translating the label above the input with `"stacked"` and `"floating"` placements, merging this logic into a single class managed by the TSX. I needed to add a new class for whether slot content is present, so the selectors were getting unwieldy otherwise.

Docs PR TBA; I plan on knocking out all three components at once when the features are all complete, to make dev builds easier to manage.

- [ ] Yes
- [x] No


core/src/components/textarea/textarea.tsx
```diff
@@ -705,55 +707,98 @@ Developers can use the "legacy" property to continue using the legacy form marku
...
-            </textarea>
+          <div class="textarea-wrapper-inner">
+            {/**
+             * Some elements have their own padding styles which may
+             * interfere with slot content alignment (such as icon-
+             * only buttons setting --padding-top=0). To avoid this,
+             * we wrap both the start and end slots in separate
+             * elements and apply our padding styles to that instead.
+             */}
+            <div class="start-slot-wrapper">
```
amandaejohnston, line 766: If desired, I can hide these wrappers via CSS or avoid rendering them entirely if there isn't any slotted content provided. I figured it'd be better to keep things simple since keeping them doesn't hurt anything aside from having an extra node in the DOM, but I'm cool with making that change.

core/src/components/textarea/textarea.scss
```diff
@@ -330,12 +330,8 @@
 // ----------------------------------------------------------------
 
 .native-wrapper {
-  display: flex;
-
   position: relative;
 
-  flex-grow: 1;
```
amandaejohnston, line 337: These weren't doing anything and were likely a copy-paste goof while adding the modern syntax; `.native-wrapper` is set to `display: grid` a few lines below this.

thetaPC (approved):
LGTM, with minor change request. Everything works as intended!

PR #28443, opened 2023-10-31
refactor(datetime): render button for month/year toggle

sean-perkins:
Issue number: Internal

---------

`ion-datetime` uses an `ion-item` to render the month/year toggle button inside of the header.

## What is the new behavior?

- `ion-datetime` uses a `button` element for the month/year toggle button

- [x] Yes
- [ ] No

Impact and migration path is noted in the `BREAKING.md`. 

## Other information

- `translucent` is not a valid CSS value for `background`. This was always intended to be `transparent`.


core/src/components/datetime/test/a11y/datetime.e2e.ts-snapshots/datetime-scale-md-ltr-Mobile-Chrome-linux.png
sean-perkins, line 1: I tried a few different approaches to match the previous behavior and could not find a viable path. I believe the proposed image diff is an improvement over the previous. The contents are vertically centered and text wraps appropriately. 

If we wanted month such as "June" to be on a single line at this scale we should consider reducing the padding to accomplish this.

core/src/components/datetime/datetime.scss
```diff
@@ -488,6 +488,32 @@ ion-picker-column-internal {
...
 }
+
+.calendar-month-year-toggle {
+  @include text-inherit();
+
+  position: relative;
+
+  border: 0;
+
+  background: var(--background);
```
sean-perkins, line 502: I've tried to account for some of the CSS variable styling APIs that existed on `ion-item` that developers _could_ be targeting via shadow parts to customize the appearance.

I think in v8 we should remove these completely in favor of the built-in properties.

e.g.:
```diff
ion-datetime::part(month-year-button) {
-  --background: blue;
+  background: blue;
}
```

If others agree, I can create the tasks and link them appropriately in the code.

core/src/components/datetime/test/a11y/datetime.spec.ts
```diff
@@ -27,24 +27,26 @@ describe('datetime', () => {
```
sean-perkins, line 27: I can remove the `IonItem` completely from this spec test.

core/src/components/datetime/datetime.scss
amandaejohnston: Since we're not covering _all_ the CSS variables from `ion-item`, wouldn't this be a breaking change as-is?

core/src/components/datetime/test/a11y/datetime.e2e.ts-snapshots/datetime-scale-ios-ltr-Mobile-Chrome-linux.png
amandaejohnston, line 1: Looks like we lost 2px of space underneath the month/year button. Might be as simple as bumping the margin up a little?

sean-perkins: Originally I had simplified some of the nested container styling from `ion-item` when flattening to just the `button > label`. However, it looks like a slotted label has explicit margin that I didn't account for. By introducing that margin and removing the padding from the `button` container, it results in no visual regressions in the common datetime implementations and here with font scaling 👍

core/src/components/datetime/test/a11y/datetime.e2e.ts-snapshots/datetime-scale-md-ltr-Mobile-Chrome-linux.png
brandyscarney: I took a look at this and while I agree that we should let the text wrap so everything should fit, I found that the buttons in the `calendar-next-prev` div are pushed over and cut off on the right with long months still. 

Notice the focus state is cut off for September:

| June | September |
| ---| ---|
| June | September |

And see how the buttons aren't aligned here:

Group 1

I am attempting to find a solution for this though!

brandyscarney: @sean-perkins Could you try adding the following to keep the buttons from shifting based on the month and to prevent the month icon from shrinking:

```diff
:host .calendar-action-buttons .calendar-month-year-toggle ion-icon {
  @include padding(0, 0, 0, 4px);

+  flex-shrink: 0;
}

+ .calendar-month-year {
+   min-width: 0;
+ }
```

| Before | After |
| ---| ---|
| before | after |

Let me know if you'd like to discuss!

core/src/components/datetime/datetime.md.scss
```diff
@@ -140,3 +140,23 @@
...
+.calendar-month-year-toggle {
+  @include padding(12px, 16px, 12px, calc(var(--padding-start) + var(--ion-safe-area-left, 0px)));
+
+  min-height: 48px;
+}
+
+.calendar-month-year-toggle ion-ripple-effect {
+  color: currentColor;
+}
+
```
brandyscarney, line 153: The focus state seems to be missing now:

| `main` | `FW-3547` |
| ---| ---|
| main | FW-3547 |

core/src/components/datetime/datetime.ios.scss
```diff
@@ -286,3 +286,9 @@
 :host .datetime-action-buttons {
   width: 100%;
 }
```
brandyscarney, line 288: The focus state is missing from iOS as well (see my comment about MD below)

core/src/components/datetime/datetime.scss
sean-perkins: I think our intention is we will target this to v8, so I can drop all the CSS variables that came from the migration from `ion-item`. 

I have a little work to get all of that ready for review 🙂

BREAKING.md
```diff
@@ -46,4 +47,16 @@ This section details the desktop browser, JavaScript framework, and mobile platf
 
 <h4 id="version-8x-content">Content</h4>
 
-- Content no longer sets the `--background` custom property when the `.outer-content` class is set on the host.
\ No newline at end of file
+- Content no longer sets the `--background` custom property when the `.outer-content` class is set on the host.
+
+<h4 id="version-8x-content">Datetime</h4>
```
brandyscarney, line 52:
```suggestion
<h4 id="version-8x-datetime">Datetime</h4>
```

core/src/components/datetime/datetime.md.scss
```diff
@@ -30,15 +30,43 @@
 
 // Calendar / Header / Action Buttons
 // -----------------------------------
-:host .datetime-calendar .calendar-action-buttons ion-item {
-  --padding-start: #{$datetime-md-header-padding};
+
+:host .calendar-action-buttons .calendar-month-year-toggle {
+  color: #{$text-color-step-350};
```
amandaejohnston, line 35: Can this be folded into the `.calendar-month-year-toggle` rule below?

core/src/components/datetime/datetime.scss
```diff
@@ -267,19 +267,24 @@ ion-picker-column-internal {
   justify-content: space-between;
 }
 
-:host .calendar-action-buttons ion-item,
+:host .calendar-action-buttons .calendar-month-year-toggle {
```
amandaejohnston, line 270: Does this rule need to be so specific now, or can we remove the `:host .calendar-action-buttons` piece? Ditto for the `ion-label` and `ion-icon` styles below. We may also be able to fold this first one into the `.calendar-month-year-toggle` rule near the bottom of this file.

core/src/components/datetime/datetime.scss
```diff
@@ -267,19 +267,24 @@ ion-picker-column-internal {
...
+  background: transparent;
+}
+
 :host .calendar-action-buttons ion-button {
-  --background: translucent;
+  --background: transparent;
 }
 
-:host .calendar-action-buttons ion-item ion-label {
+:host .calendar-action-buttons .calendar-month-year-toggle ion-label {
```
amandaejohnston, line 278: Actually, we're not using an `ion-label` here anymore, are we? It's just a regular `label`. I think this rule can be removed entirely.

core/src/components/datetime/test/a11y/datetime.e2e.ts-snapshots/datetime-scale-md-ltr-Mobile-Chrome-linux.png
amandaejohnston, line 1: It looks like the next/prev month buttons have shifted over slightly here, so they're not longer aligned with the right edge of the calendar.

core/src/components/datetime/datetime.md.scss
liamdebeasi: Fixed in https://github.com/ionic-team/ionic-framework/pull/28443/commits/ccd7a4223a6a4b1a2c20c2c6ef5a19982a2dde03

core/src/components/datetime/datetime.scss
liamdebeasi: Fixed in https://github.com/ionic-team/ionic-framework/pull/28443/commits/d2d74f2349675e03c1c6f11cf9e523579ec94dd0

core/src/components/datetime/datetime.md.scss
liamdebeasi: Fixed in https://github.com/ionic-team/ionic-framework/pull/28443/commits/e8e93cb8a269c6870dc62bbb0420233825bc0ea7

liamdebeasi:
In addition to resolving Amanda's feedback, I also fixed an issue where `cursor: pointer` was not being set on the button (this is something we have in `main`): https://github.com/ionic-team/ionic-framework/pull/28443/commits/e951ca1f8afc8697787dd29c45d57462c245efec

core/src/components/datetime/datetime.tsx
```diff
@@ -2000,44 +1999,28 @@ export class Datetime implements ComponentInterface {
...
-                    const monthYearAriaLabel = this.showMonthAndYear ? 'Hide year picker' : 'Show year picker';
-                    btn.setAttribute('aria-label', monthYearAriaLabel);
-                  }
-                }
-              }}
+              aria-label={this.showMonthAndYear ? 'Hide year picker' : 'Show year picker'}
+              onClick={() => this.toggleMonthAndYearView()}
             >
-              <ion-label>
+              <span id="toggle-wrapper">
```
liamdebeasi, line 2013: We apply `display: inline-flex` here so I opted to use `span` which is for inline styles instead of a `div` since that is for block styles (doesn't really matter since we override `display`, but I felt `span` was still semantically closest).

core/src/components/datetime/test/a11y/datetime.e2e.ts-snapshots/datetime-scale-md-ltr-Mobile-Chrome-linux.png
liamdebeasi: I took a look at this, and the new screenshots appear to be correct. It's hard to see, but the buttons are being pushed off screen in `main` whereas they are fully on screen in this branch. I added a red background to the buttons so it's easy to see:

| main | branch |
| - | - |
| Screenshot 2023-11-22 at 12 35 33 PM | Screenshot 2023-11-22 at 12 36 41 PM |

liamdebeasi: It looks like Sean made the requested changes as https://github.com/ionic-team/ionic-framework/pull/28443/files#diff-83857955657186aba3b158e5561a577b94c5b9b8e74aff61fc97c78f4b77343cR485 and https://github.com/ionic-team/ionic-framework/pull/28443/files#diff-83857955657186aba3b158e5561a577b94c5b9b8e74aff61fc97c78f4b77343cR524. I'm not able to repro the issue anymore, so I'm going to mark this as resolved.

core/src/components/datetime/datetime.scss
liamdebeasi: We're making this a breaking change in v8, so it looks like the feedback here is resolved.

liamdebeasi (approved):
I addressed all the remaining feedback, and I fixed a bug I found. The new behavior looks good. Great work!

core/src/components/datetime/datetime.tsx
sean-perkins: What's the reasoning for the use of an `id` here vs. a class? I know we've used id in the past for associating labels to controls, but rarely for styling specific concerns.

liamdebeasi: No big reason. I just figured there was only going to be one instance of this, so an `id` made the most sense.

I can also get rid of the id altogether and do `.calendar-month-year-toggle > span` though if we end up adding another span then this may cause issues.

amandaejohnston (approved):

core/src/components/datetime/datetime.tsx
sean-perkins: I think an `id` is fine here 👍 I was just curious.

PR #28450, opened 2023-11-01
chore: packages do not explicitly run tsc

liamdebeasi:
Issue number: N/A

---------

Rollup has a plugin to handle TypeScript compilation that we are not using. As a result, we are maintaining unnecessary infrastructure such as compiling to `dist-transpiled`(TS to JS) and then compiling the results to `dist` (JS files to bundled code).

## What is the new behavior?

- Vue, Vue Router, React, and React Router now make use of the Rollup Typescript plugin to handle TypeScript compilation for us. This lets us remove the `dist-transpiled` directory altogether and the manual `tsc` invocation.

- [ ] Yes
- [x] No

## Other information

Developers should not see any changes in how they consume Ionic. This just hands the TypeScript compilation step off to the Rollup plugin.

Dev build: `7.5.3-dev.11698859252.15ac163e`


thetaPC (approved):
LGTM

PR #28452, opened 2023-11-01
feat(radio-group): add compareWith property

### LINKED ISSUE #18943: feat: Add ion-radio-group compareWith property
# Feature Request

**Ionic version:**
[x] **4.x**

Support for using objects as values for `ion-radio` elements within an `ion-radio-group` and having the default value checked.

Add a `compareWith` property to `ion-radio-group` just like on `ion-select`.

**Additional Context**
Is there a reason the select does support this and the radio group doesn't? It seems to me both are merely different UI solutions for a similar technical use-case (select one of a list of values).

mapsandapps:
Issue number: resolves #18943

---------

Radio Group only allows for a strict equality check between the value and the options.

## What is the new behavior?

- Radio Group has a new optional property `compareWith` which can be a function or string.
- If `compareWith` is a string, the Radio Group will compare the value of that key.
- If `compareWith` is a function, the Radio Group will use that function to determine the selected option.
- Refactored Select a bit to reuse functionality between it and Radio Group.

- [ ] Yes
- [x] No

## Other information

Docs PR is incoming.


core/src/components/radio-group/test/compare-with/radio-group.e2e.ts
```diff
@@ -0,0 +1,91 @@
+import { expect } from '@playwright/test';
+import { configs, test } from '@utils/test/playwright';
+
+/**
+ * This behavior does not vary across modes/directions.
+ */
+configs({ modes: ['ios'], directions: ['ltr'] }).forEach(({ title, config }) => {
+  test.describe(title('radio-group: compare-with'), () => {
```
liamdebeasi, line 8: Since we're testing the logic of radio group rather than the rendering, we can probably make these spec tests instead (they'll run a lot faster than the E2E tests too)

I migrated the first two E2E tests to give you an idea of how they should look:

```tsx
import { h } from '@stencil/core';
import { RadioGroup } from '../radio-group';
import { Radio } from '../../radio/radio';

describe('radio-group: compareWith', () => {
  it('should correctly set value when using compareWith string', async () => {
      components: [RadioGroup, Radio],
      template: () => (
        <ion-radio-group compare-with="value" value={{ label: 'Blue', value: 'blue' }}>
          <ion-radio value={{ label: 'Red', value: 'red' }}>Red</ion-radio>
          <ion-radio value={{ label: 'Blue', value: 'blue' }}>Blue</ion-radio>
          <ion-radio value={{ label: 'Green', value: 'green' }}>Green</ion-radio>
        </ion-radio-group>
      ),
    });

    const radioGroup = page.body.querySelector('ion-radio-group')!;

    expect(radioGroup.value).toEqual({
      label: 'Blue',
      value: 'blue',
    });
  });
  it('should correctly set value when using compareWith function', async () => {
      components: [RadioGroup, Radio],
      template: () => (
        <ion-radio-group compare-with={(a, b) => a.value === b.value} value={{ label: 'Blue', value: 'blue' }}>
          <ion-radio value={{ label: 'Red', value: 'red' }}>Red</ion-radio>
          <ion-radio value={{ label: 'Blue', value: 'blue' }}>Blue</ion-radio>
          <ion-radio value={{ label: 'Green', value: 'green' }}>Green</ion-radio>
        </ion-radio-group>
      ),
    });

    const radioGroup = page.body.querySelector('ion-radio-group')!;

    expect(radioGroup.value).toEqual({
      label: 'Blue',
      value: 'blue',
    });
  });
});
```

(Note I'm using TSX here, so you'll want to make sure the test file has a `.tsx` extension not `.ts`. See https://stenciljs.com/docs/unit-testing#newspecpage for more info.

liamdebeasi: Let's also add a test for when `compareWith` is `null`

liamdebeasi (changes requested):
Functionality looks great. My comments are focused around the tests.

core/src/components/radio-group/test/compare-with/radio-group.e2e.ts
mapsandapps: I'm instead changing these to all have a click that changes the value. Rewriting the tests like your example made it clear that I was just testing that the value was what I had set it to... which isn't really what I intended to test.

```
    radioGroup.value = options[1];
    expect(radioGroup.value).toEqual(options[1]);
```

is basically the test I wrote 🤦🏻

liamdebeasi: Glad to hear that we are able to improve the test coverage, though my main point was that writing E2E tests are a bit overkill for what we need because aren't benefitting from cross-browser testing. They are fine as is, but they're going to take significantly longer than spec tests.

For example, running only the first E2E test takes 4.5s on my computer whereas writing it as a spec test takes 0.43s:

```js
import { h } from '@stencil/core';
import { RadioGroup } from '../../radio-group';
import { Radio } from '../../../radio/radio';

describe('radio-group: compareWith', () => {
  it('should correctly set value when using compareWith string', async () => {
      components: [RadioGroup, Radio],
      template: () => (
        <ion-radio-group compare-with="value">
          <ion-radio value={{ label: 'Red', value: 'red' }}>Red</ion-radio>
          <ion-radio value={{ label: 'Blue', value: 'blue' }}>Blue</ion-radio>
          <ion-radio value={{ label: 'Green', value: 'green' }}>Green</ion-radio>
        </ion-radio-group>
      ),
    });

    const radioGroup = page.body.querySelector('ion-radio-group')!;
    const radios = page.body.querySelectorAll('ion-radio');

    await radios[2].click();
    await page.waitForChanges();

    expect(radioGroup.value).toEqual({
      label: 'Green',
      value: 'green',
    });
    expect(radios[2]).toHaveAttribute('aria-checked', 'true');
  });
});
```

core/src/components/radio-group/radio-group.tsx
```diff
@@ -21,6 +21,14 @@ export class RadioGroup implements ComponentInterface {
    */
   @Prop() allowEmptySelection = false;
 
+  /**
+   * This property allows developers to specify a custom function or property
+   * name for comparing objects when determining the selected option in the
+   * ion-radio-group. When not specified, the default behavior will use object
+   * equality (===) for comparison.
```
liamdebeasi, line 28:
```suggestion
   * ion-radio-group. When not specified, the default behavior will use strict
   * equality (===) for comparison.
```

I think this is actually called "strict equality" according to https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Strict_equality

core/src/components/select/select.tsx
```diff
@@ -82,7 +82,10 @@ export class Select implements ComponentInterface {
   @Prop({ reflect: true }) color?: Color;
 
   /**
-   * A property name or function used to compare object values
+   * This property allows developers to specify a custom function or property
+   * name for comparing objects when determining the selected option in the
+   * ion-select. When not specified, the default behavior will use object
```
liamdebeasi, line 87: Same as other comment: I think this is called "strict equality" according to https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Strict_equality

core/src/utils/forms/compare-with-utils.ts
```diff
@@ -0,0 +1,37 @@
+type CompareFn = (currentValue: any, compareValue: any) => boolean;
+
+export const compareOptions = (
```
liamdebeasi, line 3: Can we add a JSDoc for compareOptions too like we did for isOptionSelected?

core/src/components/radio-group/test/compare-with/radio-group.e2e.ts
mapsandapps: Oh, ok. None of the spec tests I looked at had any interactivity in them (e.g. clicking), so I didn't know that was the sort of thing we would want to move into those. I can rewrite them 👍🏻

core/src/utils/forms/compare-with-utils.ts
```diff
@@ -0,0 +1,44 @@
+type CompareFn = (currentValue: any, compareValue: any) => boolean;
+
+/**
+ * Uses the compareWith param to compare two values to determine if they are equal.
+ *
+ * @param currentValue A current value of the control.
```
liamdebeasi, line 6:
```suggestion
 * @param currentValue The current value of the control.
```

Nit: We say "The current" for the function below

liamdebeasi (approved):
One naming change, but otherwise looks good. Great job!

thetaPC (approved):
LGTM

liamdebeasi:
@mapsandapps Don't forget to close https://github.com/ionic-team/ionic-framework/issues/18943 too. GitHub doesn't autoclose linked issues when merging into the non-primary branch.

mapsandapps:
@liamdebeasi thanks for the reminder!

core/src/utils/forms/compare-with-utils.ts
```diff
@@ -0,0 +1,44 @@
...
+ */
+export const compareOptions = (
+  currentValue: any,
+  compareValue: any,
+  compareWith?: string | CompareFn | null
+): boolean => {
+  if (typeof compareWith === 'function') {
+    return compareWith(currentValue, compareValue);
+  } else if (typeof compareWith === 'string') {
+    return currentValue[compareWith] === compareValue[compareWith];
```
dietergeerts, line 18: `currentValue` can be `null` with Angular forms, so this throws errors when using the property string variant :(. This is still an issue in the latest `8.6.4` version.

dietergeerts: While I was further using this component, I also see the issue that deselection doesn't work initially. This is because this custom compare isn't applied in the logic to see if a value should be deselected.

liamdebeasi: @dietergeerts Can you file an issue and include a reproduction? The team doesn't typically monitor closed PRs.

dietergeerts: Created with reproduction: https://github.com/ionic-team/ionic-framework/issues/30560

PR #28453, opened 2023-11-01
refactor: introduce watch command

liamdebeasi:
Issue number: N/A

---------

This is part of my work for hack day, but I wanted to tackle the following problems:

1. There is no easy way to have angular, angular-server, vue, vue-router, react, and react-router automatically recompile when you make a change. Right now you have to manually run `npm run build` every time you make a change.

## What is the new behavior?

- Adds `npm run build.watch` to let developers easily watch for changes in each project.

- [ ] Yes
- [x] No


packages/react/scripts/treeshaking.js
liamdebeasi, line 1: This file hasn't worked in a while as far as I can tell

packages/react-router/rollup.config.mjs
```diff
@@ -1,5 +1,3 @@
-import sourcemaps from 'rollup-plugin-sourcemaps';
```
liamdebeasi, line 1: These plugins didn't do anything to the final result, and some of them had peer dependency issues with Rollup 4, so I just removed them.

packages/react/src/components/inner-proxies.ts
```diff
@@ -7,7 +7,7 @@ import { defineCustomElement as defineIonTabButton } from '@ionic/core/component
 import type { JSX as IoniconsJSX } from 'ionicons';
 import { defineCustomElement as defineIonIcon } from 'ionicons/components/ion-icon.js';
 
-import { /*@__PURE__*/ createReactComponent } from './react-component-lib';
+import { createReactComponent } from './react-component-lib';
```
liamdebeasi, line 10: Rollup was warning that this was an invalid usage of the PURE comment. As far as I can tell, `createReactComponent` is used by all the components in this file, so adding PURE here doesn't seem to give us anything.

thetaPC:
What is your recommendation to test it?

liamdebeasi:
Easiest approach is to run `npm run build.watch` and make sure that the app re-compiles when you make a change in `src`. Note that this doesn't add any new functionality -- the watch command is provided by rollup. All I did here was add a command that's a bit easier to remember.

thetaPC (approved):
LGTM

packages/react-router/package.json
```diff
@@ -21,6 +21,7 @@
   },
   "scripts": {
     "build": "npm run clean && npm run compile",
+    "build.watch": "npm run compile -- --watch",
```
amandaejohnston, line 24: Should the new `build.watch` commands include the other steps from the original `build`s? For example, the `build` here also includes `npm run clean`.

liamdebeasi: The `--watch` flag only works for compilation not other commands. You could have `npm run clean` run at the beginning, but it wouldn't re-run every time there was a change.

amandaejohnston (approved):

PR #28459, opened 2023-11-02
refactor: update to rollup 4

liamdebeasi:
Issue number: N/A

---------

The vue, vue-router, react, and react-router packages are bundled with a version of rollup that are 2 major versions out of date.

## What is the new behavior?

- Update deps to rollup 4 and made any necessary changes as a result of breaking changes
- Removed the react projects old treeshaking script. The rollup dep used was incompatible with Rollup 4, and the script didn't work to begin with.
- Removed the source maps and resolve plugins. These did not make any difference in the final result, and source maps are still included in the final build.
- Removed a PURE annotation from an import. Rollup 4 warns about this, and this PURE annotation does not seem to be needed since every import in this file uses "createReactComponent" to begin with.

- [ ] Yes
- [x] No


thetaPC (approved):
LGTM

Issue #25243, opened 2022-05-05
feat(nav-animations): export default ios and md transitions 


DwieDima:
### Prerequisites

- [X] I have read the [Contributing Guidelines](https://github.com/ionic-team/ionic-framework/blob/main/.github/CONTRIBUTING.md#creating-an-issue).
- [X] I agree to follow the [Code of Conduct](https://ionicframework.com/code-of-conduct).
- [X] I have searched for [existing issues](https://github.com/ionic-team/ionic-framework/issues) that already include this feature request, without success.

### Describe the Feature Request

The default page transitions for ios and md should be exposed to the developers so we can use them to create custom animations.

### Describe the Use Case

I have a page transition, where the forward animation is just a fade-in effect. The back navigation should stay as it is (on ios transition from right to left).
Currently i have to inspect the ionic core code and paste the code to my method where i apply the animation.

```typescript
    const enterAnimation = (baseEl: HTMLElement, opts: any): Animation => {
      if (opts.direction === 'forward') {
        return this.animationController
          .create()
          .addElement(opts.enteringEl)
          .duration(1000)
          .easing('ease-in')
          .fromTo('opacity', '0', '1');
      }
      // default ios back transition
      return null;
    };
```

### Describe Preferred Solution

the ability to import the default ios transition to my page

```typescript

  import { ..., iosTransitionAnimation } from '@ionic/core'

  private async navigateToPage(): Promise<void> {
    const enterAnimation = (baseEl: HTMLElement, opts: any): Animation => {
      if (opts.direction === 'forward') {
        return this.animationController
          .create()
          .addElement(opts.enteringEl)
          .duration(1000)
          .easing('ease-in')
          .fromTo('opacity', '0', '1');
      }
      // default ios back transition
      return iosTransitionAnimation.direction('reverse');
    };

    await this.ionNav.push(TransferPointsPage, null, {
      animationBuilder: enterAnimation,
    });
  }
```

sean-perkins:
Hello @DwieDima thanks for opening this feature request! I think this has a lot of overlap with #24964 and would like to combine the two feature requests. 

Can you upvote that feature request to let us know you want it?

I will close this request out, but will comment on the other request regarding details from this thread. Thanks!

Issue #24964, opened 2022-03-19
feat: Expose built-in animations


crhistianramirez:
### Prerequisites

- [X] I have read the [Contributing Guidelines](https://github.com/ionic-team/ionic-framework/blob/main/.github/CONTRIBUTING.md#creating-an-issue).
- [X] I agree to follow the [Code of Conduct](https://ionicframework.com/code-of-conduct).
- [X] I have searched for [existing issues](https://github.com/ionic-team/ionic-framework/issues) that already include this feature request, without success.

### Describe the Feature Request

Export the already created animations for modals so that they can be publicly consumed by users of this library

### Describe the Use Case

I understand that I can build my own animations which is great but sometimes I just want to tweak an existing animation slightly. Today the only way to do that is to bring the source code in to my project. 

### Describe Preferred Solution

Ideally, I would like to simply import the already created animations and then call the methods on the Animation object to enhance it for my purposes

### Related Code

Here is an example of how I might expect this to work:

```typescript
import { sheetEnterAnimation } from '@ionic/core';

export const myEnterAnimation = (
  baseEl: HTMLElement,
  opts: ModalAnimationOptions
): Animation => {
  const animation = sheetEnterAnimation(baseEl, opts);

  // here I'm moving the modal so it renders somewhere else but you can imagine any other tweaks
  // that are possible by hooking into the Animation object
  animation.beforeAddWrite(() => {
    const modal = document.querySelector('ion-modal');
    const appTabs = document.querySelector('app-tabs');
    appTabs.appendChild(modal);
  });

  return animation;
};
```

3zzy:
This is absolutely needed, would potentially also resolve [my issue about page transitions](https://github.com/ionic-team/ionic-framework/issues/24995) if I could simply reuse the existing animation.

PR #28464, opened 2023-11-03
refactor: exportable md animations for modal

reslear:
It is convenient from the point of view when ios style need md style animation (https://ionicframework.com/docs/api/modal#enteranimation) that would not copy the code in the application and always use the latest changes.

exportable `mdEnterModalAnimation` and `mdLeaveModalAnimation`


liamdebeasi:
Hey there, could you please create an issue for this? We'd like to understand the use case for this feature before we expose this API. Thanks!

DwieDima:
> Hey there, could you please create an issue for this? We'd like to understand the use case for this feature before we expose this API. Thanks!

I think there is already an open issue which address this PR: #24964 (#25243)

OS-paulvisciano (approved):

PR #28473, opened 2023-11-06
refactor(angular): animation controller uses correct core instance

sean-perkins:
Issue number: Internal

---------

The `AnimationController` does not use the correct underlying instance of the utilities from either the lazy or custom elements build, depending on if the developer is using `@ionic/angular` or `@ionic/angular/standalone`. It will always use the lazy instance. 

## What is the new behavior?

- `AnimationController` uses the instance of the utilities based on it's implementation type, e.g. `@ionic/angular/standalone` uses the custom elements build with the utilities from `@ionic/core/components`.
- `@ionic/angular` and `@ionic/angular/standalone` now export their own specific implementation of `AnimationController`

- [ ] Yes
- [x] No

Ionic was re-exporting the `AnimationController` from both `@ionic/angular` and `@ionic/angular/standalone` entry points. Developers will not need to update their implementations or change import paths to take advantage of this change.


thetaPC (approved):
LGTM

PR #28474, opened 2023-11-06
refactor(angular): action sheet provider imports correct instance from core

liamdebeasi:
Issue number: N/A

---------

## What is the new behavior?

- Removed the common action sheet provider in favor of separate ones in src/standalone

- [ ] Yes
- [x] No


thetaPC (approved):
LGTM

thetaPC (approved):
LGTM

sean-perkins (approved):

PR #28477, opened 2023-11-06
refactor(angular): gesture controller uses correct core instance

sean-perkins:
Issue number: Internal

---------

The `GestureController` provider does not use the correct underlying instance of the utilities from either the lazy or custom elements build, depending on if the developer is using `@ionic/angular` or `@ionic/angular/standalone`. It will always use the lazy instance. 

This applied to the `createGesture` function.

## What is the new behavior?

- `GestureController` uses the instance of the utilities based on it's implementation type, e.g. `@ionic/angular/standalone` uses the custom elements build with the utilities from `@ionic/core/components`.
- `@ionic/angular` and `@ionic/angular/standalone` now export their own specific implementation of `GestureController`

- [ ] Yes
- [x] No


thetaPC (commented):
Should we add tests?

thetaPC:
@sean-perkins if we can setup the test apps with unit tests in the future then we should add the test that you suggested. For now, we can not leave it as is.

thetaPC (approved):
LGTM

PR #28485, opened 2023-11-07
refactor(angular): popover controller uses correct core instance

thetaPC:
Issue number: internal

---------

We also noticed that the common provider does not provide much value and it's easier to just have two separate implementations in src and standalone. (There wasn't much code we could de-duplicate)

## What is the new behavior?

- Removed the common popover provider in favor of separate ones in src/standalone

- [ ] Yes
- [x] No

## Other information

I didn't add tests because there's already existing ones for popover controller.


packages/angular/test/base/src/app/lazy/providers/providers.component.ts
```diff
@@ -28,7 +28,7 @@ export class ProvidersComponent {
     alertCtrl: AlertController,
     loadingCtrl: LoadingController,
     private menuCtrl: MenuController,
-    pickerCtrl: PickerController,
+    private pickerCtrl: PickerController,
```
sean-perkins, line 31: This diff shouldn't be needed. `pickerCtrl` is not being accessed outside of the constructor.

sean-perkins (approved):

PR #28488, opened 2023-11-08
fix(radio-group): emit value change on componentDidLoad

### LINKED ISSUE #28356: bug: [v7.5] ion-radio-group doesn't work with variables with standalone components
### Prerequisites

### Ionic Framework Version

v7.5

### Current Behavior

I refactored my app to use the new _**standalone**_ components. I found, that the radio-group doesn't work, if the "value" property comes from a variable.

I forked the original demo code from stackblitz, here is my updated code: https://stackblitz.com/edit/angular-vquurr

You can see: when we pass the value property from a variable, it doesn't work. With simple string, it works as expected.

### Expected Behavior

The radio group works as it works with module based version.

### Steps to Reproduce

See my code

### Code Reproduction URL

https://stackblitz.com/edit/angular-vquurr

### Ionic Info

Ionic 7.5

### Additional Information

_No response_

thetaPC:
Issue number: resolves #28356 

---------

`ion-radio-group` would not set the radio value when using it as a Angular standalone component and data binding:

```html
<ion-radio-group [value]="fromValue">
  <ion-radio value="dogs">Dogs</ion-radio><br />
  <ion-radio value="cats">Cats</ion-radio><br />
  <ion-radio value="turtles">Turtles</ion-radio><br />
  <ion-radio value="fish">Fish</ion-radio><br />
</ion-radio-group>
```

This is happening because the value is set before the [value watcher](https://github.com/ionic-team/ionic-framework/blob/c5dd622bbe353f4d43752ede702f60385160d5fc/core/src/components/radio-group/radio-group.tsx#L34) has been configured. The event, `ionValueChange`, does not get [dispatched](https://github.com/ionic-team/ionic-framework/blob/c5dd622bbe353f4d43752ede702f60385160d5fc/core/src/components/radio-group/radio-group.tsx#L37).

## What is the new behavior?

Run `valueChanged()` in `componentDidLoad()`.
- `valueChanged()` function is tied to the [value watcher](https://github.com/ionic-team/ionic-framework/blob/c5dd622bbe353f4d43752ede702f60385160d5fc/core/src/components/radio-group/radio-group.tsx#L34) so it will [dispatch](https://github.com/ionic-team/ionic-framework/blob/c5dd622bbe353f4d43752ede702f60385160d5fc/core/src/components/radio-group/radio-group.tsx#L37) the `ionValueChange`.

- [ ] Yes
- [x] No

## Other information

In our testing, we observed that the issue described below only occurs when assigning a value to the radio group within the primary content, such as rendering within the app component template.

When the template is isolated to a route, the value is assigned correctly. To address this issue, we need to ensure that the watcher is called after the component has finished loading, allowing the emit to be dispatched correctly.

Dev build: 7.5.4-dev.11699404450.136700d7

### DIFF
--- a/core/src/components/radio-group/radio-group.tsx
+++ b/core/src/components/radio-group/radio-group.tsx
@@ -52,7 +52,7 @@ export class RadioGroup implements ComponentInterface {
   @Event() ionValueChange!: EventEmitter<RadioGroupChangeEventDetail>;
 
   componentDidLoad() {
-    this.setRadioTabindex(this.value);
+    this.valueChanged(this.value);
   }
 
   private setRadioTabindex = (value: any | undefined) => {
--- a/core/src/components/radio-group/radio-group.tsx
+++ b/core/src/components/radio-group/radio-group.tsx
@@ -52,6 +52,15 @@ export class RadioGroup implements ComponentInterface {
   @Event() ionValueChange!: EventEmitter<RadioGroupChangeEventDetail>;
 
   componentDidLoad() {
+    /**
+     * There's an issue when assigning a value to the radio group
+     * within the Angular primary content (rendering within the
+     * app component template). When the template is isolated to a route,
+     * the value is assigned correctly.
+     * To address this issue, we need to ensure that the watcher is
+     * called after the component has finished loading,
+     * allowing the emit to be dispatched correctly.
+     */
     this.valueChanged(this.value);
   }
 


sean-perkins:
Can we add an e2e test that has a template similar to your example usage and confirms that the `ion-radio-group` value matches the expected value?

medbenmakhlouf:
I think the same problem is happening with `<ion-accordion-group [value]="fromValue">`, maybe the fix could be similar ?

thetaPC:
> I think the same problem is happening with `<ion-accordion-group [value]="fromValue">`, maybe the fix could be similar ?

@medbenmakhlouf Appreciate you flagging the issue! To help us manage and fix it more effectively, it would be great if you could open a new GitHub issue with a reproduction. This way, we can keep track of the problem and work towards a solution.

medbenmakhlouf:
@thetaPC , just did [here](https://github.com/ionic-team/ionic-framework/issues/28506). Thanks!

sean-perkins (approved):
Changes look good 👍

PR #28495, opened 2023-11-08
fix(alert): date inputs render correctly in mobile safari

### LINKED ISSUE #28494: bug: Alert date input does not render correctly on mobile safari
### Prerequisites

### Ionic Framework Version

v7.x

### Current Behavior

On iOS/Mobile Safari, a date input within an alert does not render with the correct height. This results in a squished appearance:

### Expected Behavior

The date input should render correctly regardless if the date has a selection or not:

### Steps to Reproduce

1. Clone Ionic Framework repository
2. Run the Stencil project in `core/`
3. Open: http://localhost:3333/src/components/alert/test/basic?ionic:mode=ios on an iPhone
4. Click "Prompt" button to present an alert with a date input
5. Observe: Input does not render correctly.

### Code Reproduction URL

http://localhost:3333/src/components/alert/test/basic?ionic:mode=ios

### Ionic Info

N/A

### Additional Information

_No response_

sean-perkins:
Issue number: resolves #28494

---------

The appearance of the input is being set to `none` which clears the browser appearance settings for height on the control. This results in the `<input type="date" />` to render without a height. 

## What is the new behavior?

- Minimum height is assigned to the alert input, forcing it to render at it's expected height
- Date input renders correctly in Mobile Safari

- [ ] Yes
- [x] No

## Other information

We have screenshots for this component, but Playwright runs an emulation of Mobile Safari that does not reproduce this issue that is only present on device. There for we cannot write a test on this change.


core/src/components/alert/alert.ios.scss
```diff
@@ -105,6 +105,10 @@
   &::-ms-clear {
     display: none;
   }
+
+  &::-webkit-date-and-time-value {
```
sean-perkins, line 109: This pseudo selector only appears on Mobile Safari. For Safari it is `::-webkit-date-and-time`. So this fix is specifically targeted for the problem we face on Mobile Safari.

amandaejohnston: If the problem is that `appearance` is being set to `none`, would it make sense to set `appearance: auto` instead to counteract it without needing to set a specific height?

Either way, can we add a comment explaining why the change is needed, and why we didn't use `::webkit-date-and-time`?

core/src/components/alert/test/basic/alert.e2e.ts-snapshots/alert-prompt-ios-ltr-Mobile-Safari-linux.png
amandaejohnston, line 1: It looks like 18px was too tall since the height of the inputs has increased in these screenshots. Or is the height different per device?

amandaejohnston (changes requested):

core/src/components/alert/alert.ios.scss
sean-perkins: [`appearance: auto`](https://developer.mozilla.org/en-US/docs/Web/CSS/appearance#auto) would act the same as removing the explicit `appearance` style and letting the vendor-specific styling take effect. 

On iOS an input of `type="date"` has explicit width and height styles on the pseudo element:
```css
height: 18px;
width: 86px;
```

Would a comment saying that the pseudo selector comes from iOS be sufficient? I don't know how else to describe why we didn't use `-webkit-date-and-time` other than it isn't the pseudo selector rendered:

CleanShot 2023-11-13 at 12 48 10

core/src/components/alert/test/basic/alert.e2e.ts-snapshots/alert-prompt-ios-ltr-Mobile-Safari-linux.png
sean-perkins: `18px` is the height of the input from Mobile Safari on device. Playwright emulates the environment, but does not match 1:1 in this case. 

If we set a lower height to match Web Safari, it would end up shifting the layout after a user interacts with the input on device. Instead, we set the explicit height to match iOS which will result in no layout shifts on web or iOS, but will have a few pixel height increase for web usages. 

We could additionally try setting a media query to target mobile devices vs. web, if that is desired.

core/src/components/alert/alert.ios.scss
amandaejohnston: I would include that (1) we need to set an explicit height to fix mobile Safari squishing the component down and (2) the pseudo-selector is specific to mobile Safari, similar to the original PR comment you put here.

core/src/components/alert/test/basic/alert.e2e.ts-snapshots/alert-prompt-ios-ltr-Mobile-Safari-linux.png
amandaejohnston: Makes sense 👍 I'm good with this as-is in that case.

amandaejohnston (approved):

PR #28503, opened 2023-11-09
chore: remove unused code

liamdebeasi:
Issue number: N/A

---------

While evaluating step colors for high contrast themes I discovered that this code is not actually applied anywhere.

## What is the new behavior?

- Remove unused code

- [x] Yes
- [ ] No

### DIFF
--- a/core/src/components/content/content.scss
+++ b/core/src/components/content/content.scss
@@ -54,10 +54,6 @@
   color: current-color(contrast);
 }
 
-:host(.outer-content) {
-  --background: #{$background-color-step-50};
-}
-
 #background-content {
   @include position(calc(var(--offset-top) * -1), 0px,calc(var(--offset-bottom) * -1), 0px);
 
--- a/BREAKING.md
+++ b/BREAKING.md
@@ -15,7 +15,7 @@ This is a comprehensive list of the breaking changes introduced in the major ver
 
 - [Browser and Platform Support](#version-8x-browser-platform-support)
 - [Components](#version-8x-components)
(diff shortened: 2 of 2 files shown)


core/src/components/content/content.scss
```diff
@@ -54,10 +54,6 @@
   color: current-color(contrast);
 }
 
-:host(.outer-content) {
```
liamdebeasi, line 57: This class is applied in tests, but the applicable CSS is also applied in those tests

brandyscarney: It looks like the work to change the tests was done here but not removing the actual style from content: https://github.com/ionic-team/ionic-framework/pull/16589

It was never documented as a breaking change here though: https://github.com/ionic-team/ionic-framework/blob/main/BREAKING_ARCHIVE/v4.md

Something to keep in mind as developers could be still using this since it continued to work.

liamdebeasi: Should I target v8 then? Or maybe v7.6?

brandyscarney: Possibly v8 if we want to be extra cautious. I don't know who might still be using it.

amandaejohnston: I would be good with targeting v8 and documenting as a breaking change. Since this is just removing a few lines of code, I don't think there's a strong reason to release sooner.

liamdebeasi: Ok all set. This code now targets v8

thetaPC (approved):
LGTM

PR #28510, opened 2023-11-10
fix(accordion-group): correct accordion is open on load

### LINKED ISSUE #28506: bug: [v7.5] ion-accordion-group doesn't work with variables with standalone components
### Prerequisites

### Ionic Framework Version

v7.x

### Current Behavior

I refactored my app to use the new standalone components. I found, that the accordion-group doesn't work, if the "value" property comes from a variable.

I forked the original demo code from stackblitz, here is my updated code: https://stackblitz.com/edit/angular-vquurr-sddthd

You can see: when we pass the value property from a variable, it doesn't work. With simple string, it works as expected.****

### Expected Behavior

The accordion group works as it works with module based version.

### Steps to Reproduce

See my code

### Code Reproduction URL

https://stackblitz.com/edit/angular-vquurr-sddthd

### Ionic Info

v7.5

### Additional Information

_No response_

medbenmakhlouf:
Issue number: resolves #28506 

---------

ion-accordion-group would not set the value when using it as a Angular standalone component and data binding:

``` html
<ion-accordion-group #accordionGroup [value]="fromValue">
  <ion-accordion value="turtles">
    <ion-item slot="header" color="light">
      <ion-label>First Accordion</ion-label>
    </ion-item>
    <div class="ion-padding" slot="content">First Content</div>
  </ion-accordion>
  <ion-accordion value="second">
    <ion-item slot="header" color="light">
      <ion-label>Second Accordion</ion-label>
    </ion-item>
    <div class="ion-padding" slot="content">Second Content</div>
  </ion-accordion>
</ion-accordion-group>

```

The problem here is Angular is setting the value of the accordion group after the component has been initialized (but not loaded) and before the component watchers are setup, so [valueChanged](https://github.com/ionic-team/ionic-framework/blob/d69ad434827ea56b777cc3af4e591362ce8bcd99/core/src/components/accordion-group/accordion-group.tsx#L78) does not fire automatically.

## What is the new behavior?

- Run valueChanged() in componentDidLoad().
-
-

- [ ] Yes
- [x] No

### DIFF
--- a/core/src/components/accordion-group/accordion-group.tsx
+++ b/core/src/components/accordion-group/accordion-group.tsx
@@ -181,13 +181,13 @@ export class AccordionGroup implements ComponentInterface {
       this.readonlyChanged();
     }
     /**
-     * There's an issue when assigning a value to the accordion group
-     * within the Angular primary content (rendering within the
-     * app component template). When the template is isolated to a route,
-     * the value is assigned correctly.
-     * To address this issue, we need to ensure that the watcher is
-     * called after the component has finished loading,
-     * allowing the emit to be dispatched correctly.
+     * When binding values in frameworks such as Angular
+     * it is possible for the value to be set after the Web Component
+     * initializes but before the value watcher is set up in Stencil.
+     * As a result, the watcher callback may not be fired.
+     * We work around this by manually calling the watcher
+     * callback when the component has loaded and the watcher
+     * is configured.
      */
     this.valueChanged();
   }
--- a/core/src/components/accordion-group/accordion-group.tsx
+++ b/core/src/components/accordion-group/accordion-group.tsx
@@ -180,6 +180,16 @@ export class AccordionGroup implements ComponentInterface {
     if (this.readonly) {
       this.readonlyChanged();
     }
+    /**
+     * There's an issue when assigning a value to the accordion group
+     * within the Angular primary content (rendering within the
+     * app component template). When the template is isolated to a route,
+     * the value is assigned correctly.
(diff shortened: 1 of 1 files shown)


core/src/components/accordion-group/accordion-group.tsx
```diff
@@ -180,6 +180,16 @@ export class AccordionGroup implements ComponentInterface {
...
       this.readonlyChanged();
     }
+    /**
+     * There's an issue when assigning a value to the accordion group
+     * within the Angular primary content (rendering within the
+     * app component template). When the template is isolated to a route,
+     * the value is assigned correctly.
+     * To address this issue, we need to ensure that the watcher is
+     * called after the component has finished loading,
+     * allowing the emit to be dispatched correctly.
```
liamdebeasi, line 190:
```suggestion
     * When binding values in frameworks such as Angular
     * it is possible for the value to be set after the Web Component
     * initializes but before the value watcher is set up in Stencil.
     * As a result, the watcher callback may not be fired.
     * We work around this by manually calling the watcher
     * callback when the component has loaded and the watcher
     * is configured.
```

My proposed comment provides more context on the root issue which I think will be helpful for team members working on this project.

liamdebeasi (changes requested):
Here's a dev build with your fix: `7.5.5-dev.11699885804.1d8011d5`

I tested it in your sample app, and the fix is resolved. Thanks for contributing! I left one suggestion on the code comment, but everything else looks good.

liamdebeasi (approved):

liamdebeasi:
Merged. Thank you for the PR!

PR #28511, opened 2023-11-10
fix(datetime): allow disabling datetime with prefer-wheel

mapsandapps:
Issue number: Internal

---------

It is possible to navigate the columns of a disabled Datetime with `prefer-wheel` via the keyboard.

https://github.com/ionic-team/ionic-framework/assets/14926794/9c9dafc4-4b77-45a6-a276-70201c5c3ea5

## What is the new behavior?

- Picker Column Internal has a disabled state that disables the full column
- When a Datetime is disabled with `prefer-wheel`, the columns in the Datetime will be disabled
- It is no longer possible to navigate the wheels in a disabled Datetime via the keyboard

Comparison of native & Ionic components:

Screenshot 2023-11-10 at 10 58 25 AM

- [ ] Yes
- [x] No


core/src/components/datetime/test/disabled/datetime.e2e.ts
```diff
@@ -99,5 +99,20 @@ configs({ modes: ['ios'], directions: ['ltr'] }).forEach(({ title, config, scree
 
       await expect(calendarMonthYear).toHaveText('February 2022');
     });
+
+    test('picker should be disabled in prefer wheel mode', async ({ page }) => {
```
sean-perkins, line 103: I think we could migrate this test to a spec test.

core/src/components/picker-column-internal/picker-column-internal.tsx
```diff
@@ -409,12 +414,12 @@ export class PickerColumnInternal implements ComponentInterface {
...
   get activeItem() {
     return getElementRoot(this.el).querySelector(
-      `.picker-item[data-value="${this.value}"]:not([disabled])`
+      `.picker-item[data-value="${this.value}"]:not(.picker-item-disabled)`
     ) as HTMLElement | null;
   }
 
   render() {
-    const { items, color, isActive, numericInput } = this;
+    const { items, color, disabled: pickerDisabled, isActive, numericInput } = this;
```
sean-perkins, line 422: Can you share why you are renaming the `disabled` variable here? I don't see any naming conflicts in the local scope of the `render` function.

core/src/components/picker-column-internal/test/disabled/picker-column-internal.e2e.ts
```diff
@@ -35,7 +35,7 @@ configs({ directions: ['ltr'] }).forEach(({ title, screenshot, config }) => {
  * This behavior does not vary across modes/directions.
  */
 configs({ modes: ['ios'], directions: ['ltr'] }).forEach(({ title, config }) => {
-  test.describe(title('picker-column-internal: disabled'), () => {
+  test.describe(title('picker-column-internal: disabled items'), () => {
```
sean-perkins, line 38: With changing the test name, you will need to manually delete the previous screenshots: 
- https://github.com/ionic-team/ionic-framework/blob/FW-5408/core/src/components/picker-column-internal/test/disabled/picker-column-internal.e2e.ts-snapshots/picker-internal-disabled-ios-ltr-Mobile-Chrome-linux.png
- https://github.com/ionic-team/ionic-framework/blob/FW-5408/core/src/components/picker-column-internal/test/disabled/picker-column-internal.e2e.ts-snapshots/picker-internal-disabled-ios-ltr-Mobile-Firefox-linux.png
- https://github.com/ionic-team/ionic-framework/blob/FW-5408/core/src/components/picker-column-internal/test/disabled/picker-column-internal.e2e.ts-snapshots/picker-internal-disabled-ios-ltr-Mobile-Safari-linux.png
- https://github.com/ionic-team/ionic-framework/blob/FW-5408/core/src/components/picker-column-internal/test/disabled/picker-column-internal.e2e.ts-snapshots/picker-internal-disabled-md-ltr-Mobile-Chrome-linux.png
- https://github.com/ionic-team/ionic-framework/blob/FW-5408/core/src/components/picker-column-internal/test/disabled/picker-column-internal.e2e.ts-snapshots/picker-internal-disabled-md-ltr-Mobile-Firefox-linux.png
- https://github.com/ionic-team/ionic-framework/blob/FW-5408/core/src/components/picker-column-internal/test/disabled/picker-column-internal.e2e.ts-snapshots/picker-internal-disabled-md-ltr-Mobile-Safari-linux.png

core/src/components/picker-column-internal/test/disabled/picker-column-internal.e2e.ts
```diff
@@ -159,3 +159,40 @@ configs({ modes: ['ios'], directions: ['ltr'] }).forEach(({ title, config }) =>
...
+    });
+
+    test('disabled column should not have visual regressions', async ({ page }) => {
+      const disabledColumn = page.locator('#column-disabled');
+      await disabledColumn.evaluate((el: HTMLIonPickerColumnInternalElement) => {
+        el.value = 11;
+      });
+      await page.waitForChanges();
+
+      await expect(disabledColumn).toHaveScreenshot(screenshot('picker-internal-disabled-column'));
```
sean-perkins, line 179: We should dig into this test more. Playwright is already reporting it as flaky.

sean-perkins (changes requested):

core/src/components/picker-column-internal/picker-column-internal.tsx
mapsandapps: oh, i just did this to reduce confusion when it's used below, e.g. `const isItemDisabled = pickerDisabled || item.disabled || false;` seemed more clear to me than `const isItemDisabled = disabled || item.disabled || false;`

core/src/components/picker-column-internal/test/disabled/picker-column-internal.e2e.ts
sean-perkins: 🤦 good catch - I don't know why I thought the `test.describe` title change was a screenshot name change.

core/src/components/picker-column-internal/picker-column-internal.scss
```diff
@@ -77,6 +77,14 @@
   cursor: default;
 }
 
+:host(.picker-column-disabled) {
+  overflow-y: hidden;
+
+  .picker-item {
```
sean-perkins, line 83: I think we could group this selector with the following styles to have the opacity managed from a single style:
```scss
:host .picker-item.picker-item-disabled,
:host .picker-item.picker-column-disabled {
  opacity: 0.4;
}
```

sean-perkins (approved):
Behavior looks great locally 👍 I had one suggestion for organizing the styles, but everything else looks good to me. Great work!

core/src/components/picker-column-internal/picker-column-internal.scss
```diff
@@ -86,7 +91,7 @@
 }
 
 @media (any-hover: hover) {
-  :host(:focus) {
+  :host(:focus:not(.picker-column-disabled)) {
```
liamdebeasi, line 94: Focusable items can't be focused if they are disabled (for example, `<button>`). I understand this is happening because we set `tabindex={0}` on the host, but it might make more sense to do `tabindex={disabled ? null : 0}` in the template instead. It's not a huge deal, but the selector here is confusing given how regular interactive elements work.

core/src/components/picker-column-internal/picker-column-internal.tsx
```diff
@@ -443,6 +451,8 @@ export class PickerColumnInternal implements ComponentInterface {
           &nbsp;
         </div>
         {items.map((item, index) => {
+          const isItemDisabled = pickerDisabled || item.disabled || false;
```
liamdebeasi, line 454: Why do we do `|| false`? If `pickerDisabled` is `false` and `item.disabled` is `false` then `isItemDisabled` should evaluated to `false`.

I see we do this same thing later on in the file, which we shouldn't need to do either. (I know you didn't make the later change, but just wanted to call it out)

liamdebeasi: Also, why do we use `isItDisabled` when setting the `disabled` state of the button, but not when setting the `picker-item-disabled` class?

mapsandapps: That `|| false` bothered me too. But the class object doesn't care for the possibility of `undefined`.

Screenshot 2023-11-17 at 8 18 15 AM


I could alternatively do something like `const isItemDisabled = Boolean(pickerDisabled || item.disabled);` or `const isItemDisabled = (pickerDisabled || item.disabled) as boolean;` but I'm not sure that that feels better :shrug:

---

i did some refactoring to make the use of `isItemDisabled` more consistent

core/src/components/picker-column-internal/picker-column-internal.scss
mapsandapps: Great idea, thanks! This change helped me be able to refactor some other stuff that hopefully makes everything more clear.

core/src/components/picker-column-internal/picker-column-internal.tsx
```diff
@@ -408,13 +413,17 @@ export class PickerColumnInternal implements ComponentInterface {
   };
 
   get activeItem() {
-    return getElementRoot(this.el).querySelector(
-      `.picker-item[data-value="${this.value}"]:not([disabled])`
-    ) as HTMLElement | null;
+    // If the whole picker column is disabled, the current value should appear active
+    // If the current value item is specifically disabled, it should not appear active
+    const selector = `.picker-item[data-value="${this.value}"]:not(${
```
liamdebeasi, line 418: Why did this change? `.picker-item-empty` will never have the `data-value` attribute to begin with, so this selector will never match a `.picker-item-empty` element. Additionally, we disable the `.picker-item` button if the component's `disabled` prop is `true`. As a result, the original selector should still be sufficient.

mapsandapps: I can remove the `.picker-item-empty` part. But the behavior does need to be different for a disabled column vs disabled items (as far as I understand.) The behavior previously was that it should not scroll to a disabled item. But when the whole column is disabled, it *should* scroll to a disabled item. e.g. if the default value in a calendar picker is today, it should scroll to that even if the column is disabled. You can see a native example of that in my text in the PR description.

liamdebeasi: Ok makes sense! Does the selector need to be `.picker-item-disabled` or are we able to keep using `[disabled]`?

mapsandapps: Oh, true, we can go back to `[disabled]` after the refactoring I did previously.

core/src/components/datetime/test/disabled/datetime.spec.tsx
```diff
@@ -0,0 +1,37 @@
...
+
+    await page.waitForChanges();
+
+    const datetime = page.body.querySelector('ion-datetime')!;
+    const columns = datetime.shadowRoot!.querySelectorAll('ion-picker-column-internal');
+
+    await expect(columns.length).toEqual(4);
+
+    columns.forEach((column) => {
+      expect(column.getAttribute('disabled')).not.toBeNull();
```
liamdebeasi, line 34: I don't understand why we use `toBeNull` here if the property type is `boolean | undefined`. If `disabled={undefined}` then this check would still pass. Wouldn't we want `toBeUndefined` instead?

core/src/components/picker-column-internal/picker-column-internal.tsx
```diff
@@ -423,13 +430,16 @@ export class PickerColumnInternal implements ComponentInterface {
...
     return (
       <Host
         exportparts={`${PICKER_ITEM_PART}, ${PICKER_ITEM_ACTIVE_PART}`}
-        tabindex={0}
+        disabled={pickerDisabled}
+        tabindex={pickerDisabled ? null : 0}
         class={createColorClasses(color, {
           [mode]: true,
           ['picker-column-active']: isActive,
+          ['picker-column-disabled']: pickerDisabled,
```
liamdebeasi, line 442: This class doesn't seem needed anymore. The only place it's used is in the CSS, and that could be swapped out for the `[disabled]` attribute.

liamdebeasi (changes requested):
The behavior looks good now, though there are some places where we can clean up our code.

core/src/components/picker-column-internal/picker-column-internal.tsx
mapsandapps: Sure. I went ahead and removed the `.picker-item-disabled` class while I was in there, by the same reasoning, even though it predates this PR.

liamdebeasi: I was referring to `picker-column-disabled` not `picker-item-disabled`. The `picker-column-disabled` class was added in this PR.

core/src/components/datetime/test/disabled/datetime.spec.tsx
liamdebeasi: The problem is you aren't including the picker components in your test, so they aren't getting defined. If you log each `column` in the `forEach` loop you'll see that the result is an empty object (`{}`).

Your `hasAttribute` works because even if the component isn't defined, the DOM node is still there which _does_ have a disabled attribute.

`getAttribute` should never return `undefined` according to https://developer.mozilla.org/en-US/docs/Web/API/Element/getAttribute which is why your `toBeUndefined` check was not working with `getAttribute`.

So you could still use `toBeNull` and be fine, but I think this is confusing because when thinking about the `disabled` property it can never be null.

------

I think it would be more clear to do the following:

1. Import and pass the picker components to the `components` array.

```diff
+ import { PickerColumnInternal } from '../../../picker-column-internal/picker-column-internal';
+ import { PickerInternal } from '../../../picker-internal/picker-internal';

...

- components: [Datetime],
+ components: [Datetime, PickerColumnInternal, PickerInternal],
```


2. Check the component property instead of the attribute:

```diff
- expect(column.hasAttribute('disabled')).toBe(true);
+ expect(column.disabled).toBe(true);
```

core/src/components/picker-column-internal/test/disabled/picker-column-internal.e2e.ts
mapsandapps: It's still flaking. I can't reproduce locally 😕

liamdebeasi: Can we try loading the test with the `value` already set? The problem might be that Playwright isn't waiting for the scroll to finish before taking a screenshot.

mapsandapps: Done in 46616677852bdaaa26439ff2b3183b970c805926 and I also made a change in 7a3771a220170a5ed91dce629d44bd41d281bef2 that hopefully will help.

liamdebeasi: I meant that we should try loading the picker with `value="11"` already set. So you should be able to do `value="11"` here: https://github.com/ionic-team/ionic-framework/blob/46616677852bdaaa26439ff2b3183b970c805926/core/src/components/picker-column-internal/test/disabled/index.html#L56 and then remove this line: https://github.com/ionic-team/ionic-framework/blob/46616677852bdaaa26439ff2b3183b970c805926/core/src/components/picker-column-internal/test/disabled/index.html#L81

For https://github.com/ionic-team/ionic-framework/commit/7a3771a220170a5ed91dce629d44bd41d281bef2 this causes a regression where the selected item is no longer highlighted when the picker column is disabled. You can see this at `src/components/picker-column-internal/test/disabled` on MD mode. The "11" option in "Column disabled" is no longer highlighted in blue.

According to your comment in https://github.com/ionic-team/ionic-framework/pull/28511/files#diff-5e7b473017ccb3a8be0fa9c7c9d8e1fc393b64ddfe719d5f3274e212ba09b3fdR416 this item should be highlighted blue.

mapsandapps: Updated. Also, I added a screenshot for the disabled calendar wheel to capture that intended behavior.

liamdebeasi: > I don't understand how setting the value directly changes anything for the scrolling if the items don't get set until the script? It can't scroll until the items are there any more than it can scroll if the value isn't there, right? 

The issue that I raised was the value was being set after the component loaded: https://github.com/ionic-team/ionic-framework/pull/28511/files/b7e42befeabbbb96044bbaa0aeb6ba6668baa547#diff-875e8dba2863da31c37c64b445f9c01418f2ee542625acf814f7a8cab7464b19R174

Your old test loaded the picker with `value="12"`. The picker spends time scrolling to value 12. The test then set the value to `11`, so the picker then needed to spend additional time scrolling to value 11. This additional value update can take a few hundred ms especially on CI.

By setting up the test with the correct configuration on load we a) reduce the run time of the test and b) reduce the likelihood of flaky tests.

core/src/components/picker-column-internal/picker-column-internal.tsx
liamdebeasi: Re-read this and looks like I misunderstood your original comment. All of that sounds good!

liamdebeasi (approved):
I made one change in https://github.com/ionic-team/ionic-framework/pull/28511/commits/fab89a7aed8c2fe50ce0f176e090669b7edd2793

This test is not needed because we already have the following tests:

1. A test to verify that datetime sets the `disabled` prop on columns correctly: https://github.com/ionic-team/ionic-framework/pull/28511/files#diff-68455eb0492521a5359f114d63eedc2c4d3731e47f7fa6f349ba54456bcc18a2
2. A test to verify that a column renders correctly when `disabled` is set: https://github.com/ionic-team/ionic-framework/pull/28511/files#diff-875e8dba2863da31c37c64b445f9c01418f2ee542625acf814f7a8cab7464b19

Datetime does not apply any disabled-specific visual styles for the wheel picker. It does apply `pointer-events: none`, but the test I removed was not checking that.

The rest of this is good to go. Great work!

PR #28513, opened 2023-11-10
chore: adds angular test app for v17

mapsandapps:
Issue number: Internal

---------

There is no Angular v17 test app.

## What is the new behavior?

There is an Angular v17 test app.

New overrides:
- angular.json: `browserTarget` was changed to `buildTarget` and to replace `@nguniversal` with `@angular/ssr` and `@angular-devkit/build-angular`
- server.ts: Replace uses of `@nguniversal` with `@angular/ssr`

New change to base:
- polyfills.ts: Import [changed](https://github.com/angular/angular/blob/main/CHANGELOG.md#zonejs) from `import 'zone.js/dist/zone';` (this change is supported for all versions of Angular that we support, so should be changed in the base file)

All other files were duplicated from their v16 counterparts.

- [ ] Yes
- [x] No


packages/angular/test/apps/ng17/src/polyfills.ts
```diff
@@ -0,0 +1,56 @@
...
+ *
+ */
+
+import './zone-flags';
+
+/***************************************************************************************************
+ * Zone JS is required by default for Angular itself.
+ */
+
+import 'zone.js';  // Included with Angular CLI.
```
mapsandapps, line 51: We now need to overwrite this file because this import changed. [[changelog](https://github.com/angular/angular/blob/main/packages/zone.js/CHANGELOG.md#breaking-changes)]

packages/angular/test/apps/ng17/angular.json
```diff
@@ -0,0 +1,156 @@
...
+              "sourceMap": true,
+              "namedChunks": true
+            }
+          },
+          "defaultConfiguration": "production"
+        },
+        "serve": {
+          "builder": "@angular-devkit/build-angular:dev-server",
+          "options": {
+            "buildTarget": "test-app:build"
```
mapsandapps, line 81: We need to overwrite this file because `browserTarget` was changed to `buildTarget`.

packages/angular/test/apps/ng17/server.ts
```diff
@@ -0,0 +1,71 @@
+
+import 'zone.js/node';
+
+import { APP_BASE_HREF } from '@angular/common';
+import { CommonEngine } from '@angular/ssr';
```
mapsandapps, line 5: We need to overwrite this file to replace uses of `@nguniversal` with `@angular/ssr`

sean-perkins (commented):
The `zone.js` import change happend during v14, but we missed migrating it to our test apps. 

e.g.: https://github.com/ionic-team/starters/commit/19a0b5c832ca90e63dda455aba11948ee0dabc4f#diff-4e4a3c7371faa81b32df8ae9be1815ed723e74e713e553de6702da3f439fc4f9

Can we abstract that change into the `base/` files and remove it from the specific ng17 test app files?

Everything else looks great!

sean-perkins (approved):
Great work!

thetaPC (approved):
LGTM

PR #28521, opened 2023-11-13
refactor(angular): picker controller uses correct core instance

mapsandapps:
Issue number: Internal

---------

## What is the new behavior?

- Removed the common picker provider in favor of separate ones in src/standalone

- [ ] Yes
- [x] No


sean-perkins:
Running `npm run eslint -- --fix` will fix the build errors for the import group ordering. Tomorrow I can update the commands to do this automatically when running `npm run lint`.

sean-perkins (approved):
Looks good, great work!

PR #28529, opened 2023-11-14
chore(core): type checking for unit tests

sean-perkins:
Issue number: N/A

---------

Type checking inside of the Stencil unit tests have been disabled for a long time. This has resulted in a difficult developer experience and numerous issues (both types and implementation) within our unit tests.

## What is the new behavior?

- Type checking is now enabled for all Stencil unit tests
- Tests have been updated to resolve type errors and implementation errors
  - Many `as any` casts were introduced, as many legacy tests test invalid configurations of functions that require it (for example passing `undefined` to an argument that cannot be `undefined`).

- [ ] Yes
- [x] No

## Other information

To test this PR you can checkout the branch locally. Install dependencies in the `/core` directory to make sure you are on at least `@stencil/core@4.7.2`. 

Opening either a `.spec.ts` or `.spec.tsx` file, validate that your IDE detects types and can provide auto completions for jest global types. 

If you desire, you can provide an invalid type and try building the project - you will observe the build will fail due to the invalid type.


core/package.json
```diff
@@ -64,6 +64,7 @@
     "jest": "^29.7.0",
     "jest-cli": "^29.7.0",
     "prettier": "^2.6.1",
+    "puppeteer": "21.1.1",
```
liamdebeasi, line 67: Why do we need puppeteer as a dev dependency?

core/src/components/modal/test/can-dismiss/modal.spec.tsx
```diff
@@ -44,6 +43,7 @@ describe('modal: canDismiss', () => {
         components: [Modal],
         template: () => (
           <ion-modal
+            overlayIndex={1}
```
liamdebeasi, line 46: Why do we need to set overlayIndex?

liamdebeasi (commented):
Changes overall look good, just had a couple questions.

core/src/components/popover/test/util.spec.ts
```diff
@@ -17,45 +17,45 @@ describe('isTriggerElement', () => {
 
 describe('getIndexOfItem', () => {
   it('should return the correct index in an array of ion-items', () => {
-    const array = createArrayOfElements(['ion-item', 'ion-item', 'ion-item']);
+    const array = createArrayOfElements(['ion-item', 'ion-item', 'ion-item']) as HTMLIonItemElement[];
```
amandaejohnston, line 20: Optional: You could get fancy with this by adding a generic type parameter to `createArrayOfElements`. Not super important, just feels nicer than casting 😄

amandaejohnston (approved):
I had the same questions as Liam, but otherwise this LGTM.

core/src/components/datetime/test/data.spec.ts
```diff
@@ -364,15 +365,15 @@ describe('generateTime()', () => {
         year: undefined,
         hour: 19,
         minute: 50,
-      };
+      } as unknown as DatetimeParts;
```
brandyscarney, line 368: Why do we do `as unknown` here instead of something like 

```suggestion
      } as Partial<DatetimeParts>;
```

core/src/global/test/config-controller.spec.ts
```diff
@@ -18,14 +19,14 @@ describe('Config', () => {
...
-    });
-    expect(config.getBoolean('bool0')).toEqual(false);
-    expect(config.getBoolean('bool1')).toEqual(false);
-    expect(config.getBoolean('bool2')).toEqual(true);
-    expect(config.getBoolean('bool3')).toEqual(true);
-    expect(config.getBoolean('bool4')).toEqual(false);
-    expect(config.getBoolean('bool5')).toEqual(false);
-    expect(config.getBoolean('bool6')).toEqual(true);
+    } as any);
+    expect(config.getBoolean('bool0' as any)).toEqual(false);
```
brandyscarney, line 23: Should these be typed as `keyof IonicConfig` instead?

```suggestion
    expect(config.getBoolean('bool0' as keyof IonicConfig)).toEqual(false);
```

core/src/utils/test/hardware-back-button.spec.ts
```diff
@@ -1,11 +1,12 @@
+import type { BackButtonEvent } from '../../../src/interface';
 import { startHardwareBackButton } from '../hardware-back-button';
 
 describe('Hardware Back Button', () => {
   beforeEach(() => startHardwareBackButton());
   it('should call handler', () => {
     const cbSpy = jest.fn();
     document.addEventListener('ionBackButton', (ev) => {
```
brandyscarney, line 8: Why not set the event type here?

```suggestion
    document.addEventListener('ionBackButton', (ev: BackButtonEvent) => {
```

core/src/utils/test/hardware-back-button.spec.ts
```diff
@@ -1,11 +1,12 @@
...
-      ev.detail.register(0, cbSpy);
+      (ev as BackButtonEvent).detail.register(0, cbSpy);
```
brandyscarney, line 9: Then remove it here?

```suggestion
      ev.detail.register(0, cbSpy);
```

brandyscarney (commented):
Looks good! I had a few questions about some casting.

core/package.json
sean-perkins: Stencil regressed this behavior in v3.3.0: https://github.com/ionic-team/stencil/issues/4526.

At this point it is required or else you get a build error when `tsc` runs.

core/src/components/modal/test/can-dismiss/modal.spec.tsx
sean-perkins: We have it explicitly set as a required field: https://github.com/ionic-team/ionic-framework/blob/f143bd0a11941739ee616908891509b48bd48463/core/src/components/modal/modal.tsx#L103

core/src/components/popover/test/util.spec.ts
sean-perkins: I tried looking into this originally, but later usages passes an item and a div. So we'd have to do potentially a generic that works as a tuple.. and that got way more involved than seemed worth it here.

core/src/components/action-sheet/test/basic/action-sheet.spec.tsx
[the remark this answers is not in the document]
sean-perkins: `*.tsx` is normally reserved for when you are doing bindings to the attributes in the template. 

e.g.:
```tsx
  components: [Toast],
  template: () => <ion-toast htmlAttributes={{ 'data-testid': 'basic-toast' }}></ion-toast>,
});
```

core/src/components/datetime/test/data.spec.ts
sean-perkins: All downstream consumers of this argument require the entire `DatetimeParts` object. But for what we are testing here we are only using a minimal representation of that object. 

`as unknown as DatetimeParts` allows us to say "we know what we are doing" by forcing the object to a shape that isn't valid to the typings. 

`as Partial<DatetimeParts>` would require a huge overall of the type implementations of all the utilities and assertions that properties are defined before using them. 

I question why we are checking so many things with invalid arguments. That's what typescript gives us for free. But I only resolved the typing errors as part of this.

core/src/global/test/config-controller.spec.ts
sean-perkins: `bool0` is not a key of that interface. So we would either need to do `as unknown as keyof IonicConfig` or cast to `any` here. Since we aren't using the key anywhere else, I opted for `any`. 

These legacy tests should probably get an overhaul in the future to use valid keys or provide an interface shape for the config that matches what the test is accessing.

core/src/utils/test/hardware-back-button.spec.ts
sean-perkins: You can't. It'll result in a typing error since `BackButtonEvent` does not match the type signature that comes from `.addEventListener`.

sean-perkins: FWIW this is why we have this logic for this custom type: https://github.com/ionic-team/ionic-framework/blob/f143bd0a11941739ee616908891509b48bd48463/core/src/utils/browser/index.ts#L30-L41

But we don't get that when using the `Window` global type.

core/package.json
liamdebeasi: Does that mean this can be removed once the issue is fixed on Stencil's end?

core/src/components/datetime/test/data.spec.ts
liamdebeasi: We check invalid arguments because not everyone uses TypeScript. For example, using Vue w/o TypeScript is common enough that we have a page on our docs that show how to remove TypeScript: https://ionicframework.com/docs/vue/quickstart#build-your-way-with-typescript-or-javascript

liamdebeasi (approved):
One comment in https://github.com/ionic-team/ionic-framework/pull/28529#discussion_r1395752505. This is good to go once addressed.

core/src/global/test/config-controller.spec.ts
brandyscarney: Oh okay. When I tested it out locally it worked for me but maybe I needed to refresh VsCode or something.

brandyscarney (approved):

core/src/components/datetime/test/data.spec.ts
sean-perkins: Is this still fairly common in Vue? Our starters and playgrounds both use Typescript. Having unit test coverage on behavior that is driven by the consuming package vs. the implementation of the code is strange to me, especially in utility functions that developers aren't accessing directly. 

If `ion-datetime` can accept potentially a `null` or `undefined` property - I think that makes perfect sense to test & would be possible without manual casting since the type signature supports it.

liamdebeasi: yeah that's fair -- I could see us testing the public APIs with invalid inputs but expect that the internal utils always use valid inputs

core/package.json
sean-perkins: Correct, created FW-5617 to track.

PR #28188, opened 2023-09-16
fix(angular): transition animation plays when using browser back and forward buttons

hoi4:
Issue number: resolves #16569

---------

This PR resolves a few issues around navigation transition animations and the navigation stack when using the browser back and forward buttons.

1. The respective transition animation is now applied automatically (without e.g. the workaround described [here](https://github.com/ionic-team/ionic-framework/issues/28176#issuecomment-1720914437))

2. Sometimes when navigating forward with the browser button, pages in the middle of the current tab's stack were incorrectly removed. This caused an issue when toggling between pages via the browser back and forward button (especially visible when using custom transition animations).  

I added some logic to use the current navigation stack as basis for the direction of the animation that will be applied.

Short demonstration video of the changes (using the `ng16` test app):
https://github.com/ionic-team/ionic-framework/assets/23243256/40aba3ec-0c56-44f1-8b82-ba5a2903b414

## Does this introduce a breaking change?

- [ ] Yes
- [x] No

At least to my limited knowledge. The tests are still passing. I just had to increase the timeouts slightly to properly wait for the page transitions to finish.

PR #28532, opened 2023-11-14
test(angular): remove routing waits in tests

liamdebeasi:
Issue number: N/A

---------

Our Angular E2E tests are brittle because they rely on arbitrary `cy.wait` calls to account for asynchronous routing. This leads to flaky tests on CI and seemingly random test failures when we make adjustments to the Ionic Anguar routing integration (see: https://github.com/ionic-team/ionic-framework/pull/28188)

Additionally, our test execution for the navigation tests is quite slow because transitions are enabled. As a result, we need to wait hundreds of ms per test just for the transitions to finish.

## What is the new behavior?

- Updated the `testStack` command to use a new `getStack` [Cypress query](https://docs.cypress.io/api/cypress-api/custom-queries). These queries come with automatic retrying built-in. By leveraging this query in the `testStack` command, we can avoid the arbitrary waits.
- Added `ionic:_testing=true` query strings to the navigation tests. This causes Ionic to disable any transitions so the tests execute faster.
- Removed most of the arbitrary `cy.wait` calls. I kept the swipe to go back `cy.wait` -- I wasn't quite sure how to reduce flakiness on that one.

- [ ] Yes
- [x] No

### DIFF
--- a/packages/angular/test/base/cypress/support/commands.js
+++ b/packages/angular/test/base/cypress/support/commands.js
@@ -41,11 +41,21 @@ Cypress.Commands.add('ionSwipeToGoBack', (complete = false, selector = 'ion-rout
   cy.wait(150);
 });
 
+/**
+ * getStack is a query because it has automatic
+ * retries built in which will let us account for
+ * async routing without having to use
+ * arbitrary cy.wait calls.
+ */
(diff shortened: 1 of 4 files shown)


sean-perkins (approved):
I think it's a good call leaving the wait for the swipe to go back.

PR #28543, opened 2023-11-16
refactor(angular): loading controller uses correct core instance

mapsandapps:
Issue number: Internal

---------

## What is the new behavior?

- Removed the common loading provider in favor of separate ones in src/standalone

- [ ] Yes
- [x] No


packages/angular/standalone/src/providers/loading-controller.ts
```diff
@@ -0,0 +1,17 @@
+import { Injectable } from '@angular/core';
+import { OverlayBaseController } from '@ionic/angular/common';
+import type { LoadingOptions } from '@ionic/core';
+import { loadingController } from '@ionic/core';
```
sean-perkins, line 4: This is referencing the lazy instance of the loading controller, which is not desired. These imports should come from `@ionic/core/components`.

sean-perkins (changes requested):

thetaPC:
@mapsandapps is this ready to review?

mapsandapps:
@thetaPC yep!

sean-perkins (approved):

PR #28583, opened 2023-11-27
feat(input, textarea, select): add start and end slots

### LINKED ISSUE #26297: feat: standalone input component should support icons

amandaejohnston:
Issue number: Resolves #26297

---------

With the modern form control syntax, it is not possible to add icon buttons or other decorators to the sides of `ion-input`, `ion-textarea`, or `ion-select`, as you can with `ion-item`.

## What is the new behavior?

`start` and `end` slots added to each component.

This PR is a combination of several others that were already approved. If needed, it might be easiest to review the PRs individually by looking at the commit history here.

- [ ] Yes
- [x] No

## Other information

Docs PR: https://github.com/ionic-team/ionic-docs/pull/3271

Dev build: `7.5.4-dev.11701112913.1ea61220`


core/src/components/textarea/test/slot/index.html
```diff
@@ -51,51 +51,184 @@
       <ion-content id="content" class="ion-padding">
         <div class="grid">
           <div class="grid-item">
-            <h2>No Fill / Start</h2>
+            <h2>No Fill / Start Label</h2>
```
liamdebeasi, line 54: I know in the select/input PRs we discussed using an icon instead of a button in the start slot since that's not a very common pattern. Can we do the same here for the a11y and slot files for `textarea`?

liamdebeasi (approved):
One template change in the tests, but otherwise looks good. I'm also fine with the API naming -- start/end seems to make the most sense here. Great work!

core/src/components/input/input.scss
```diff
@@ -652,3 +653,14 @@
...
+// Start/End Slots
+// ----------------------------------------------------------------
+
+::slotted([slot="start"]) {
+  margin-inline-end: $form-control-label-margin;
+}
+
+::slotted([slot="end"]) {
+  margin-inline-start: $form-control-label-margin;
+}
```
thetaPC, line 666:
```suggestion
}

```
Missing new line

core/src/components/select/select.scss
```diff
@@ -545,3 +580,23 @@ button {
...
+  flex-shrink: 0;
+}
+
+::slotted([slot="start"]) {
+  margin-inline-end: $form-control-label-margin;
+}
+
+::slotted([slot="end"]) {
+  margin-inline-start: $form-control-label-margin;
+}
```
thetaPC, line 602:
```suggestion
}

```
Missing new line.

thetaPC (changes requested):
Thank you for the clear comments! Made it easier to review.

core/src/components/textarea/test/slot/index.html
amandaejohnston: Updated for input and textarea in https://github.com/ionic-team/ionic-framework/pull/28583/commits/ff3c9c80ba08ce6132be20391eaf8e7a4c25b1db 👍 I also added some headers to the `slot` tests to make them easier to parse.

klill1:
Had a question about using the start and end slots for buttons or icons. If the label placement is set to floating and there is ion-button with slot inside the input then the label is already floating and not inside the field if the button is there and there is no value on the field. Is this a bug or there is a workaround for that?

PR #28584, opened 2023-11-27
refactor(picker): deprecate ion-picker and ion-picker-column

liamdebeasi:
Issue number: Internal

---------

As part of our work to expose `ion-picker-internal` as an inline picker, we'd like to preserve the legacy picker so devs can continue to access it if they are not ready to migrate.

## What is the new behavior?

- `ion-picker` and `ion-picker-column` have been renamed to `ion-picker-legacy` and `ion-picker-legacy-column`, respectively.
- Deprecated `ion-picker-legacy`, `pickerController` (in Angular, Core, and Vue, `useIonPicker` in React)

- [x] Yes
- [ ] No

## Other information

A few implementation notes:

1. I opted to only put the deprecation warning in `ion-picker-legacy` and not `ion-picker-legacy-column`. If devs are using `ion-picker-legacy` they are typically using `ion-picker-legacy-column` too and vice versa.
2. I intentionally did not change the file names for the picker and picker column. There should be no naming collisions with the internal picker since these are in separate directories. Also, the names of the files are internal details and do not impact how developers build Ionic apps.
3. I decided to call the components `ion-picker-legacy` and `ion-picker-legacy-column` to hopefully make the migration to the inline picker a bit easier. My reasoning is (apart from the other API changes) you could at least quickly do a find and replace for `ion-picker-legacy` to replace `ion-picker-legacy` and `ion-picker-legacy-column` with `ion-picker` and `ion-picker-column`, respectively. On the other hand, if we called the component `ion-picker-column-legacy` then you'd need to do one find and replace for `ion-picker-legacy` and then another find and replace for `ion-picker-column-legacy`.

Dev build: `7.5.7-dev.11701188221.15180cb4`
Core functionality in a stackblitz: https://stackblitz.com/edit/bs8p6f?file=index.html


BREAKING.md
```diff
@@ -59,4 +60,9 @@ This section details the desktop browser, JavaScript framework, and mobile platf
...
       +  background: red;
       }
-      ```
\ No newline at end of file
+      ```
+<h2 id="version-8x-picker">Picker</h2>
+
+- `ion-picker` and `ion-picker-column` have been renamed to `ion-picker-legacy` and `ion-picker-legacy-column`, respectively. This change was made to accommodate the new inline picker component while allowing developers to continue to use the legacy picker during this migration period.
+  - Only the component names have been changed. Usages such as `ion-picker` or `IonPicker` should be changed to `ion-picker-legacy` and `IonPickerLegacy`, respectively.
+  - Non-component usages such as `pickerController` or `useIonPicker` remain unchanged.The new picker displays inline with your page content and does not have equivalents for these non-component usages.
```
sean-perkins, line 68:
```suggestion
  - Non-component usages such as `pickerController` or `useIonPicker` remain unchanged. The new picker displays inline with your page content and does not have equivalents for these non-component usages. 
```

core/src/components/picker-legacy/picker.tsx
```diff
@@ -30,7 +31,7 @@ import type { PickerButton, PickerColumn } from './picker-interface';
  * @virtualProp {"ios" | "md"} mode - The mode determines which platform styles to use.
  */
 @Component({
```
sean-perkins, line 33: Should we consider renaming the class name for `Picker` and `PickerColumnCmp` so that after migrating the internal implementation we don't have competing names for the modern vs. deprecated approaches?

core/src/components/picker-column-legacy/picker-column.tsx
```diff
@@ -6,13 +6,13 @@ import { getClassMap } from '@utils/theme';
...
 import type { Gesture, GestureDetail } from '../../interface';
-import type { PickerColumn } from '../picker/picker-interface';
+import type { PickerColumn } from '../picker-legacy/picker-interface';
 
 /**
  * @internal
  */
 @Component({
-  tag: 'ion-picker-column',
+  tag: 'ion-picker-legacy-column',
```
sean-perkins, line 15: ~ The naming of `ion-picker-legacy-column` vs. `ion-picker-column-legacy` caught my eye, especially as the folder name is called `picker-column-legacy`. I'd be curious if others flagged this while reviewing. 

I can understand why we landed on `ion-picker-legacy-column` from a selector perspective, but I think we should be consistent in folder naming since to date the selector has mostly matched the folder name (without `ion-`).

core/src/components/picker-legacy/picker.tsx
liamdebeasi: Wouldn't that only be a concern if you're developing Ionic and importing both the deprecated and modern approaches in the same folder? IIRC the class names aren't exposed to developers.

core/src/components/picker-column-legacy/picker-column.tsx
liamdebeasi: Oops that was an oversight on my part. Fixed in https://github.com/ionic-team/ionic-framework/pull/28584/commits/41b520bc60dacba83e2247015bc94cf468a938d8.

core/src/components/picker-legacy/picker.tsx
mapsandapps: Informational question: I assume we _do_ want to rename the `PickerInternal` class to just `Picker` even though it's not exposed to developers? From this, it sounds like that shouldn't cause any issues.

core/src/components/picker-legacy-column/test/standalone/picker-column.e2e.ts
```diff
@@ -4,13 +4,13 @@ import { configs, test } from '@utils/test/playwright';
 configs({ modes: ['ios'], directions: ['ltr'] }).forEach(({ title, config }) => {
   test.describe(title('picker-column'), () => {
```
mapsandapps, line 5: Do these titles need to be unique? If we change the internal ones but not these, will that cause any problems?

core/src/components/picker-legacy/picker.tsx
```diff
@@ -202,6 +203,11 @@ export class Picker implements ComponentInterface, OverlayInterface {
   }
 
   componentDidLoad() {
+    printIonWarning(
+      'ion-picker-legacy and ion-picker-legacy-column have been deprecated in favor of the ion-picker and ion-picker-column components. These new components display inline with your page content allowing for more presentation flexibility than before.',
```
mapsandapps, line 207: Is this wording perhaps a bit confusing given that the developer would have just changed all their uses _from_ "ion-picker"? I've provided one suggestion.

```suggestion
      'ion-picker-legacy and ion-picker-legacy-column have been deprecated in favor of new versions of the ion-picker and ion-picker-column components. These new components display inline with your page content allowing for more presentation flexibility than before.',
```

packages/react/src/components/IonPickerLegacy.tsx
```diff
@@ -1,9 +1,9 @@
...
 import { createInlineOverlayComponent } from './createInlineOverlayComponent';
 
-export const IonPicker = /*@__PURE__*/ createInlineOverlayComponent<JSX.IonPicker, HTMLIonPickerElement>(
-  'ion-picker',
-  defineCustomElement
-);
+export const IonPickerLegacy = /*@__PURE__*/ createInlineOverlayComponent<
+  JSX.IonPickerLegacy,
+  HTMLIonPickerLegacyElement
+>('ion-picker', defineCustomElement);
```
mapsandapps, line 9: This is the tag of the component, right? So this should have changed?

packages/react/src/hooks/useIonPicker.tsx
```diff
@@ -9,9 +9,10 @@ import { useController } from './useController';
 /**
  * A hook for presenting/dismissing an IonPicker component
  * @returns Returns the present and dismiss methods in an array
+ * @deprecated Use the inline ion-picker component instead.
  */
 export function useIonPicker(): UseIonPickerResult {
-  const controller = useController<PickerOptions, HTMLIonPickerElement>(
+  const controller = useController<PickerOptions, HTMLIonPickerLegacyElement>(
     'IonPicker',
```
mapsandapps, line 16: Just to confirm my understanding of why we don't rename 'IonPicker' here in line 16: We don't need to rename this because it's used for the hook & controller, which we in turn don't need to rename because there's no equivalent in IonPickerInternal. Do I have that right?

mapsandapps (commented):
A buncha questions (some are more about the future internal rename than this rename).

core/src/components/picker-legacy-column/picker-column.tsx
```diff
@@ -6,13 +6,13 @@ import { getClassMap } from '@utils/theme';
```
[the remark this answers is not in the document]
liamdebeasi: > The ticket & design doc were unclear about whether or not this file & others like it should be renamed to e.g. picker-legacy-column.tsx. I don't see any reason it would matter, I just wanted to check that we aren't renaming them.

This is addressed at the top of the PR:

"I intentionally did not change the file names for the picker and picker column. There should be no naming collisions with the internal picker since these are in separate directories. Also, the names of the files are internal details and do not impact how developers build Ionic apps."

> As a side question: I assume we do want to rename e.g. picker-internal-column.tsx to picker-column.tsx?

Yea I think that makes sense.

packages/react/src/components/IonPickerLegacy.tsx
liamdebeasi: Yes, this is what developers import when they want to use the picker.

packages/react/src/hooks/useIonPicker.tsx
liamdebeasi: Correct!

core/src/components/picker-legacy-column/test/standalone/picker-column.e2e.ts
liamdebeasi: Titles need to be unique within a file, so since the legacy and modern tests are separate files we should be 
fine.

core/src/components/picker-legacy/picker.tsx
liamdebeasi: > Informational question: I assume we do want to rename the PickerInternal class to just Picker even though it's not exposed to developers? From this, it sounds like that shouldn't cause any issues.

Yeah I think renaming `PickerInternal` to `Picker` makes sense. It shouldn't have any impact on developers, but I think from a maintenance point of view it could be confusing for a team member to see "Internal" on a public component.

core/src/components/picker-legacy/test/basic/picker.e2e.ts
```diff
@@ -15,7 +15,7 @@ configs().forEach(({ title, screenshot, config }) => {
       await expect(page).toHaveScreenshot(screenshot(`picker-basic`));
```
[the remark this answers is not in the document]
liamdebeasi: I wasn't planning on updating the screenshot file names just to keep the number of diffs low. However, screenshot file names need to be unique to the test file, so having two screenshots named `picker-basic` should be fine as long as they are in different test files.

packages/react/src/components/IonPickerLegacy.tsx
mapsandapps: Sorry for the confusion. I'm inquiring about line 9.

https://github.com/ionic-team/ionic-framework/blob/cf7882172a13d9253d6d49e4becfcf5641df68df/packages/react/src/components/IonPickerLegacy.tsx#L9

liamdebeasi: Oh thanks! Yes, that should say `ion-picker-legacy`. Fixed in https://github.com/ionic-team/ionic-framework/pull/28584/commits/5180cb4c2c16861ee6bc6d83ddd200d26eacc405

mapsandapps (approved):
LGTM

packages/react/src/components/IonPickerLegacy.tsx
amandaejohnston, line 1: Threading: If I copy the React code from the [basic example](https://ionicframework.com/docs/api/picker#inline-pickers-recommended) into an app, install the dev build, and switch to the `IonPickerLegacy` component, the picker doesn't open when I click the button.

liamdebeasi: Ah sorry, the dev build was from before https://github.com/ionic-team/ionic-framework/pull/28584/commits/5180cb4c2c16861ee6bc6d83ddd200d26eacc405. I'll create a new dev build now.

liamdebeasi: Here's a new one: 7.5.7-dev.11701188221.15180cb4

sean-perkins (approved):
I don't have any outstanding feedback, good work!

packages/react/src/components/IonPickerLegacy.tsx
amandaejohnston: Working now, thank you!

amandaejohnston (approved):

Issue #28300, opened 2023-10-07
feat: checkbox, radio, range, toggle expose label shadow part


muuvmuuv:
### Prerequisites

- [X] I have read the [Contributing Guidelines](https://github.com/ionic-team/ionic-framework/blob/main/.github/CONTRIBUTING.md#creating-an-issue).
- [X] I agree to follow the [Code of Conduct](https://ionicframework.com/code-of-conduct).
- [X] I have searched for [existing issues](https://github.com/ionic-team/ionic-framework/issues) that already include this feature request, without success.

### Describe the Feature Request

As for all other input elements like ion-select and ion-input, ion-range should also get the new implementation of labels with a part attribute attached for styling and customizations.

![image](https://github.com/ionic-team/ionic-framework/assets/11534760/78f9b6df-a1cc-42fc-8956-5548f95b99c0)

### Describe the Use Case

I am building a consistent styling for all input elements in our app where the label is outside the input box, but the input box has a shadow and other styling. This works for all other elements pretty well, thanks for the work in that.

![image](https://github.com/ionic-team/ionic-framework/assets/11534760/c487657b-af34-4e55-b077-4b2a497a0460)

### Describe Preferred Solution

Slots

### Describe Alternatives

No shadow-dom

### Additional Information

~~I remember that there have been a lot of label issues opened, but couldn't find one related to slots and ion-range.~~

muuvmuuv:
Ah yes, sorry, it's not clear because it is both named “slot”. I mean for styling, that I have a slot in HTML to select via CSS when it is in the shadow-dom. Alternatively, avoid shadow-dom like it's in some other ion components when using labels. I think it is really inconsistent between components.

liamdebeasi:
When using the label slot all of the content is in the light DOM, so you can style it directly: https://codepen.io/liamdebeasi/pen/qBLgdpR

muuvmuuv:
Liam I am so sorry, I meant "part"... hard week, pretty under stress with migrating our forms module and implementing other features.

Here is my repl. You see that range has shadow-dom without [part], ion-input has no shadow-dom and ion-select has shadow-dom but with [part]. That is what I meant with inconsistency.

https://codepen.io/muuvmuuv-the-selector/pen/ExGrjRg

liamdebeasi:
Thanks for the clarification. So the reason why there appears to be styling inconsistency is two reasons:

1. **You're using different APIs meant for different use cases.**

[...]

PR #26283, opened 2022-11-14
feat(input): component can be used outside of ion-item

liamdebeasi:
Note: Docs will be coming in a separate PR.

## Pull request checklist

Please check if your PR fulfills the following requirements:
- [x] Tests for the changes have been added (for bug fixes / features)
- [ ] Docs have been reviewed and added / updated if needed (for bug fixes / features)
  - Some docs updates need to be made in the `ionic-docs` repo, in a separate PR. See the [contributing guide](https://github.com/ionic-team/ionic-framework/blob/main/.github/CONTRIBUTING.md#modifying-documentation) for details.
- [x] Build (`npm run build`) was run locally and any changes were pushed
- [x] Lint (`npm run lint`) has passed locally and any fixes were made for failures

## Pull request type

 

 

Please check the type of change your PR introduces:
- [ ] Bugfix
- [x] Feature
- [ ] Code style update (formatting, renaming)
- [ ] Refactoring (no functional changes, no api changes)
- [ ] Build related changes
- [ ] Documentation content changes
- [ ] Other (please describe): 

## What is the current behavior?

Issue URL: resolves https://github.com/ionic-team/ionic-framework/issues/20153 resolves https://github.com/ionic-team/ionic-framework/issues/19084 resolves https://github.com/ionic-team/ionic-framework/issues/22736

See https://github.com/ionic-team/ionic-framework/discussions/25661

## What is the new behavior?

- `ion-input` can now be used independent of `ion-item`.
- Deprecated the `counter`, `counterFormatter`, `fill`, and `shape` properties on `ion-item`
- Deprecated the `helper` and `error` slots on `ion-item`.

Note: Each commit links back to a PR that describes the different functionality added. I opted to do this instead of adding all the changes directly into this branch.

## Does this introduce a breaking change?

- [ ] Yes
- [x] No

PR #28585, opened 2023-11-27
feat(toggle): expose label wrapper as shadow part

liamdebeasi:
Issue number: Part of #28300

---------

Developers can access the contents of the label in `ion-toggle`. However, they do not access the label wrapper. As a result, they are unable to customize the position of the label wrapper relative to the control (margin, padding, etc).

## What is the new behavior?

- Adds a `label` shadow part to `ion-toggle`

- [ ] Yes
- [x] No

## Other information

As part of this work, I investigated the `pointer-events` on `.label-text-wrapper` to see if we need to move it higher up in the DOM: https://github.com/ionic-team/ionic-framework/blob/01130e12e1d73bbf558da9d4dffd7122822ff39c/core/src/components/toggle/toggle.scss#L113

This code was originally added for `ion-input`: https://github.com/ionic-team/ionic-framework/pull/26283/commits/9a9568c8c32150f4df25cb5cde6f1efb02f152fe. I later added this same logic when updating toggle https://github.com/ionic-team/ionic-framework/pull/26357. However, I was not able to reproduce the original issue with `ion-toggle` (tested iOS 16, iOS 17, Android 13 with screen readers on). This leads me to believe that the patch is only needed for text input controls.

I removed the code from `ion-toggle`. However, authors should still verify if this code is needed for each of the remaining features (checkbox, radio, and range).


core/src/components/toggle/test/toggle.spec.ts
```diff
@@ -40,6 +40,19 @@ describe('toggle', () => {
...
+  describe('shadow parts', () => {
+    it('should have shadow parts', async () => {
+      const page = await newSpecPage({
+        components: [Toggle],
+        html: `<ion-toggle></ion-toggle>`,
+      });
+      const toggle = page.body.querySelector('ion-toggle')!;
+      expect(toggle.shadowRoot!.querySelector('[part="label"]')).not.toBe(null);
+      expect(toggle.shadowRoot!.querySelector('[part="track"]')).not.toBe(null);
+      expect(toggle.shadowRoot!.querySelector('[part="handle"]')).not.toBe(null);
```
liamdebeasi, line 53: Not strictly necessary for this PR, but I figured while I was here I might as well check the other shadow parts too.

mapsandapps (approved):
Looks good 👍🏻

brandyscarney (approved):

PR #28591, opened 2023-11-28
feat(picker-column-option): add the new component

thetaPC:
Issue number: #

---------

## What is the new behavior?

-
-
-

- [ ] Yes
- [ ] No

### DIFF
--- a/core/src/components/picker-column-option/picker-column-option.tsx
+++ b/core/src/components/picker-column-option/picker-column-option.tsx
@@ -1,0 +1,55 @@
+import type { ComponentInterface } from '@stencil/core';
+import { Component, Element, Host, Prop, h } from '@stencil/core';
+import type { Attributes } from '@utils/helpers';
+import { inheritAriaAttributes } from '@utils/helpers';
+
+import { getIonMode } from '../../global/ionic-global';
+
+@Component({
+  tag: 'ion-picker-column-option',
+  shadow: true,
+})
+export class PickerColumnOption implements ComponentInterface {
+  private optionId = `ion-picker-opt-${pickerOptionIds++}`;
+
+  private inheritedAttributes: Attributes = {};
+
+  @Element() el!: HTMLElement;
+
+  /**
+   * If `true`, the user cannot interact with the select option.
+   */
+  @Prop() disabled = false;
+
+  /**
+   * The text value of the option.
+   */
+  @Prop() value?: any | null;
+
+  componentWillLoad() {
+    this.inheritedAttributes = inheritAriaAttributes(this.el);
+  }
+
+  render() {
+    const { value, disabled, inheritedAttributes } = this;
+    const ariaLabel = inheritedAttributes['aria-label'] || null;
+
+    return (
+      <Host id={this.optionId} class={getIonMode(this)}>
+        <button
+          tabindex="-1"
+          aria-label={ariaLabel}
+          class={{
+            'picker-opt': true,
+            'picker-opt-disabled': !!disabled,
+          }}
+          disabled={disabled}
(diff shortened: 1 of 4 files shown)


amandaejohnston:
Made a dev build: `7.5.7-dev.11701270755.14f050ad`

core/src/components/picker-column-option/picker-column-option.tsx
```diff
@@ -0,0 +1,55 @@
...
+    const ariaLabel = inheritedAttributes['aria-label'] || null;
+
+    return (
+      <Host id={this.optionId} class={getIonMode(this)}>
+        <button
+          tabindex="-1"
+          aria-label={ariaLabel}
+          class={{
+            'picker-opt': true,
+            'picker-opt-disabled': !!disabled,
```
amandaejohnston, line 44: We may want to use `picker-item`/`picker-item-disabled` instead to match the existing implementation: https://github.com/ionic-team/ionic-framework/blob/4f050ad437e4087b05edba8686691f61c9cd70c2/core/src/components/picker-column/picker-column.tsx#L470 This would make it easier to copy styles over later.

amandaejohnston (commented):
Tested in core and all three frameworks, looks good 👍

core/src/components/picker-column-option/picker-column-option.tsx
```diff
@@ -0,0 +1,55 @@
...
+   */
+  @Prop() disabled = false;
+
+  /**
+   * The text value of the option.
+   */
+  @Prop() value?: any | null;
+
+  componentWillLoad() {
+    this.inheritedAttributes = inheritAriaAttributes(this.el);
```
liamdebeasi, line 30: As per the ticket, we should be taking advantage of Stencil's watch decorator for attributes: https://stenciljs.com/docs/reactive-data#watching-native-html-attributes

liamdebeasi (changes requested):

core/src/components/picker-column-option/picker-column-option.tsx
amandaejohnston: Oh, nevermind, I see you were copying this: https://github.com/ionic-team/ionic-framework/blob/4f050ad437e4087b05edba8686691f61c9cd70c2/core/src/components/picker-legacy-column/picker-column.tsx#L420 I think I just got confused on which implementation is which 😆

amandaejohnston (approved):
LGTM aside from Liam and Sean's comments.

core/src/components/picker-column-option/picker-column-option.tsx
thetaPC: Thank you both for the deeper explanations. I now have a better idea on how this stuff works.

core/src/components/picker-column-option/picker-column-option.tsx
```diff
@@ -0,0 +1,77 @@
...
+     * the first render.
+     */
+    this.ariaLabel = inheritedAttributes['aria-label'] || null;
+  }
+
+  render() {
+    const { value, disabled, ariaLabel } = this;
+
+    return (
+      <Host id={this.optionId} class={getIonMode(this)}>
```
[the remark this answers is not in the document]
thetaPC: I added them for future usage. I'll remove them since this is a stubbed version.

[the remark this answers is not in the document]
liamdebeasi: If they are for future usage I'm fine keeping them, I was just hoping we could document why it was needed.

liamdebeasi (approved):
Thank you!

PR #28593, opened 2023-11-28
chore(playwright): dark mode support

sean-perkins:
Issue number: N/A

---------

Testing dark mode is manual per test in Playwright. Ionic developer needs to setup the variables and assign them to a selector that applies in the class. 

## What is the new behavior?

- The `.setContent` API will now work with a new config option to test dark mode automatically without additional configuration/test set-up. 
- Default theme is no theme (fallback theme)
- Screenshot names and test titles remain the same for all existing tests. Only tests that opt into a theme will be pre-pended with `-dark` or `-light` (as an example for current themes). 

- [ ] Yes
- [x] No


core/scripts/testing/variables.css
brandyscarney, line 1: Do we want this file to exist as `variables.css` or should we add them as theme files where their name matches the theme names like this: https://github.com/ionic-team/ionic-framework/tree/main/core/src/themes/test/css-variables/css

That way if we ever want to include other themes (such as a high contrast theme) we just have to add a new file with that name?

brandyscarney (commented):
Code makes sense, just a question on how we want to support multiple themes if we decide to add more in the future.

core/scripts/testing/variables.css
sean-perkins: The code currently behaves differently than this, but I can update it to match this idea. It will set a requirement that the theme name has to match the stylesheet name (since we will be conditionally importing the stylesheet), but I think the idea is better ergonomically. 

New themes will require both a new stylesheet and an update to the type signature for themes.

sean-perkins: Addressed here: https://github.com/ionic-team/ionic-framework/pull/28593/commits/44fc0372d7d93d5e5e2c4c13acff52d245bebb2c

Inverted all of the existing logic so `light` theme will operate as it does today for titles/screenshots, but any new themes will format correctly. This will reduce the amount of work to support additional themes.

core/src/utils/test/playwright/generator.ts
```diff
@@ -27,19 +30,27 @@ interface TestConfigOption {
...
 const generateTitle = (title: string, config: TestConfig): string => {
-  const { mode, direction } = config;
+  const { mode, direction, theme } = config;
 
-  return `${title} - ${mode}/${direction}`;
+  if (theme === 'light') {
+    return `${title} - ${mode}/${direction}`;
+  }
+
+  return `${title} - ${mode}/${direction}/${theme}`;
```
liamdebeasi, line 39: Can we add a comment as to why we have this conditional? Same thing for `generateScreenshotName` below. (I know why, I'm just concerned this context will be lost over time)

core/src/utils/test/playwright/page/utils/set-content.ts
```diff
@@ -19,25 +19,30 @@ export const setContent = async (page: Page, html: string, testInfo: TestInfo, o
...
     <!DOCTYPE html>
-    <html dir="${direction}">
+    <html dir="${direction}" lang="en">
       <head>
+        <title>Ionic Playwright Test</title>
         <meta charset="UTF-8" />
         <meta name="viewport" content="width=device-width, initial-scale=1.0, minimum-scale=1.0" />
         <link href="${baseUrl}/css/ionic.bundle.css" rel="stylesheet" />
         <link href="${baseUrl}/scripts/testing/styles.css" rel="stylesheet" />
+        <link href="${baseUrl}/scripts/testing/themes/${theme}.css" rel="stylesheet" />
```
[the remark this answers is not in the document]
sean-perkins: `theme` is currently always defined in this capacity since we assign a default of `light` when the test config doesn't specify an explicit theme to test against. 

I'll rework this a bit so that `default` is an available option for the fallback state. If the theme is `default` it will opt-out of importing a theme stylesheet.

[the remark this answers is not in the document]
sean-perkins: Addressed here: https://github.com/ionic-team/ionic-framework/pull/28593/commits/54289d8757cb579a37d823172fab74c04190dfcc (both feedback comment)

core/src/utils/test/playwright/generator.ts
```diff
@@ -1,6 +1,13 @@
...
 export type Direction = 'ltr' | 'rtl';
-export type Theme = 'light' | 'dark';
+/**
+ * The theme to use for the playwright test.
+ *
+ * - `default`: The fallback theme values.
+ * - `light`: The light theme values.
+ * - `dark`: The dark theme values.
+ */
+export type Theme = 'default' | 'light' | 'dark';
```
liamdebeasi, line 10: Bikeshedding: I wonder if "none" or "fallback" would make more sense here? I think of light mode as the default theme. Not a blocker for me, so if you think "default" is still the best option I'm fine with that.

liamdebeasi (approved):
One non-blocking comment. Otherwise looks good. Thanks for doing this!

core/src/utils/test/playwright/generator.ts
sean-perkins: I'm not wedded to the naming here, also agree `default` isn't clear. I think `none` could make sense 👍

brandyscarney (approved):

PR #28594, opened 2023-11-28
fix(item): allow item to grow when it is used in a flex container

brandyscarney:
Issue number: N/A

---------

When an item containing a select is used inside of a flex container, the item collapses to `0px` width. Example code:

```html
<div style="display: flex">
  <ion-item>
    <ion-select aria-label="fruit" placeholder="Select fruit">
      <ion-select-option value="apples">Apples</ion-select-option>
      <ion-select-option value="oranges">Oranges</ion-select-option>
      <ion-select-option value="bananas">Bananas</ion-select-option>
    </ion-select>
  </ion-item>
</div>
```

> [!NOTE]
> This bug was introduced in `feature-7.6` by [this commit](https://github.com/ionic-team/ionic-framework/commit/6438e3e919c665569b731a2d74fe1547b4f3c1cc#diff-7e97878a82f02919b555a816b1620fd8aff34fc94b7f96b7810807b58ce6c777R88). It cannot be reproduced on `main`. 

## What is the new behavior?
Sets the flex property to `1` on `ion-item` so that it will grow inside of a flex container, resulting in the select being displayed. This property is ignored when inside of a block container. 

- [ ] Yes
- [x] No

## Other information

| `feature-7.6` | branch |
| ---| ---|
| feature-7 6 | branch |


amandaejohnston (commented):
Fix looks good. Should we add a test for this?

brandyscarney:
I meant to add that in the description - I can add a test if we want one. Should it go in item?

amandaejohnston:
Is there a way to show the change in behavior without `ion-select`? If not, I think it should go in select, but I could be convinced either way 🤷

brandyscarney:
Yeah I couldn't reproduce it outside of select so I added it to the select tests here: https://github.com/ionic-team/ionic-framework/pull/28594/commits/6d0d4902c1c601af98040fa5d8e0fb58f09bc8d2

Running screenshots now.

core/src/components/picker-internal/test/basic/picker-internal.e2e.ts-snapshots/picker-internal-popover-diff-md-ltr-Mobile-Firefox-linux.png
amandaejohnston, line 1: This screenshot was flaky, you'll probably want to revert it 👀

amandaejohnston (approved):

core/src/components/picker-internal/test/basic/picker-internal.e2e.ts-snapshots/picker-internal-popover-diff-md-ltr-Mobile-Firefox-linux.png
brandyscarney: Ah yeah I had to revert one with the last run too. 😩

sean-perkins (approved):

liamdebeasi:
Removing myself from review since there are two approvals already. Lmk if you would still like me to look though!

PR #28601, opened 2023-11-29
Fw 5614

liamdebeasi:
Issue number: #

---------

## What is the new behavior?

-
-
-

- [ ] Yes
- [ ] No

### DIFF
--- a/core/api.txt
+++ b/core/api.txt
@@ -1077,6 +1077,7 @@ ion-range,css-prop,--pin-color
 ion-range,part,bar
 ion-range,part,bar-active
 ion-range,part,knob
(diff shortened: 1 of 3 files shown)


sean-perkins (approved):
Great work 👍 filling in the unit test coverage for the other shadow parts was a good improvement too!

I wonder if we should create a test util or custom selector to cut down on the boiler plate to follow this mentality with all our shadow dom components. 

e.g.:
```ts
expect(range).toHaveShadowPart('label');
```

Not necessary as part of this!

mapsandapps (approved):

PR #28603, opened 2023-11-29
feat(picker-column): add prefix and suffix slots

amandaejohnston:
Issue number: Internal

---------

The new picker component does not have a counterpart for the `prefix` and `suffix` properties on the old picker column.

## What is the new behavior?

- `prefix` and `suffix` slots added.
- Picker items wrapped in a new `.picker-opts` element for positioning purposes, similar to the legacy picker column. This necessitated additional changes to functionality and styling, since previously the items were direct descendants of the host.
- New `setFocus` method added. This gets around a browser bug in Firefox where calling `focus()` on the column doesn't correctly move focus, as the nearest focusable element (`.picker-opts`) is in the shadow DOM.

- [ ] Yes
- [x] No

While you can no longer call `focus()` on the column in Firefox and must use `setFocus()` instead, this isn't considered a breaking change because this component isn't public API yet.


core/src/components/datetime/datetime.scss
```diff
@@ -25,19 +25,6 @@
...
- * between months, sometimes the allowed
- * dates may be filtered. As a result, it
- * is possible to get a layout shift as
- * the picker column will shrink to fit the
- * widest item in the column. Setting a minimum
- * width avoids this layout shifting.
- */
-ion-picker-column {
-  min-width: 26px;
-}
```
amandaejohnston, line 39: I had to remove this from datetime and put it in picker-column so it can be moved to `.picker-opts`, which is in the shadow DOM.

core/src/components/datetime/test/custom/datetime.e2e.ts-snapshots/datetime-custom-time-picker-ios-ltr-Mobile-Chrome-linux.png
amandaejohnston, line 1: These time picker changes are from the column widths increasing due to the `min-width: 26px` applying in all cases instead of just on wheel datetimes. IMO this is an acceptable trade-off; see the new comments on `picker-column.scss` for more context on why the change was necessary.

core/src/components/picker-column/test/disabled/picker-column.e2e.ts-snapshots/picker-disabled-column-md-ltr-Mobile-Chrome-linux.png
amandaejohnston, line 1: See https://github.com/ionic-team/ionic-framework/pull/28603#discussion_r1411251941 re: these disabled column changes.

core/src/components/datetime/test/custom/datetime.e2e.ts-snapshots/datetime-custom-time-picker-ios-ltr-Mobile-Chrome-linux.png
amandaejohnston, line 1: These time picker changes are from the column widths increasing due to the `min-width: 26px` applying in all cases instead of just on wheel datetimes. IMO this is an acceptable trade-off; see the new comments on `picker-column.scss` for more context on why the change was necessary.

core/src/components/datetime/test/position/index.html
```diff
@@ -25,7 +25,7 @@
       <ion-content class="ion-padding">
         <ion-button id="open-datetime" onclick="presentPopover(defaultPopover, event)">Present Popover</ion-button>
         <ion-popover class="datetime-popover" id="default-popover">
-          <ion-datetime value="2022-05-03"></ion-datetime>
+          <ion-datetime value="2022-05-03T14:23:00.000Z"></ion-datetime>
```
amandaejohnston, line 28: I made this change because of a minor regression where the hour column would jump down to 11 on opening the popover for the first time, on Firefox only. I decided this was acceptable because the missing time isn't a supported use case anyway, and the test is checking the position of the popover itself (which isn't affected). So I added a defined time to skirt the odd behavior.

core/src/components/picker-column/picker-column.md.scss
```diff
@@ -1,6 +1,8 @@
 @import "./picker-column.scss";
 @import "../../themes/ionic.globals.md";
 
-:host .picker-item-active {
+:host .picker-item-active,
+::slotted([slot="prefix"]),
+::slotted([slot="suffix"]) {
```
amandaejohnston, line 6: This is to mirror the legacy picker, though I wasn't sure if we wanted to keep it.

core/src/components/picker-column/picker-column.scss
```diff
@@ -4,13 +4,42 @@
...
+   * Note that we need this min-width set on the same
+   * element as the vertical padding (i.e. not on the
+   * host).
+   */
+  min-width: 26px;
+
   outline: none;
 
-  font-size: 22px;
+  text-align: inherit;
```
amandaejohnston, line 41: Datetime adjusts the text alignment of specific columns in its pickers; I added this so we can continue to specify the `text-align` on the host from other components.

core/src/components/datetime/test/custom/datetime.e2e.ts-snapshots/datetime-custom-time-picker-ios-ltr-Mobile-Chrome-linux.png
amandaejohnston, line 1: These time picker changes are from the column widths increasing due to the `min-width: 26px` applying in all cases instead of just on wheel datetimes. IMO this is an acceptable trade-off; see the new comments on `picker-column.scss` for more context on why the change was necessary.

core/src/components/datetime/test/position/datetime.e2e.ts-snapshots/datetime-position-base-ios-ltr-Mobile-Chrome-linux.png
amandaejohnston, line 1: See https://github.com/ionic-team/ionic-framework/pull/28603#discussion_r1411358559 for why the screenshots for this test changed.

core/src/components/picker-column/test/disabled/picker-column.e2e.ts-snapshots/picker-disabled-column-md-ltr-Mobile-Chrome-linux.png
amandaejohnston, line 1: See https://github.com/ionic-team/ionic-framework/pull/28603#discussion_r1411375718 for why these changed.

core/src/components/picker-column/picker-column.tsx
```diff
@@ -434,62 +454,71 @@ export class PickerColumn implements ComponentInterface {
...
-        </div>
-        <div class="picker-item picker-item-empty" aria-hidden="true">
-          &nbsp;
-        </div>
-        <div class="picker-item picker-item-empty" aria-hidden="true">
-          &nbsp;
+        <slot name="prefix"></slot>
+        <div
+          class="picker-opts"
+          tabindex={pickerDisabled ? undefined : 0}
```
liamdebeasi, line 466:
```suggestion
          tabindex={pickerDisabled ? null : 0}
```

The original code used `null`

core/src/components/picker-column/picker-column.md.scss
liamdebeasi: I'm fine keeping it, though my 2 cents is using the color makes me think it's interactive/selected when it is not.

core/src/components/picker-column/picker-column.scss
```diff
@@ -22,17 +57,46 @@
...
+.picker-opts::-webkit-scrollbar {
   display: none;
 }
 
+::slotted([slot="prefix"]), ::slotted([slot="suffix"]) {
+  display: flex;
+
+  flex: 1;
+  align-items: center;
+}
```
liamdebeasi, line 74: Currently the slot contains are equal width, so as one slot grows in width so does the other one (even if it doesn't necessarily need it). Just wanted to make sure this is an intended behavior:
 
image


This behavior seems to align with how the current `ion-picker` works though the difference is the prefix/suffix always evenly take up the remaining white space.

core/src/components/picker-column/picker-column.scss
```diff
@@ -22,17 +57,46 @@
...
+
+::slotted([slot="prefix"]) {
```
liamdebeasi, line 76: We might want to have text truncate with ellipses otherwise it will spill outside of the highlight area: 
image


Prefix/suffix text should generally be short anyways.

core/src/components/picker/test/basic/picker.e2e.ts-snapshots/picker-popover-diff-md-ltr-Mobile-Chrome-linux.png
liamdebeasi, line 1: Was this supposed to change?

liamdebeasi (changes requested):
Code is overall looking good! There are some adjustments we should consider making for when we have long text. I know prefix/suffix should generally be short, but right now the layout breaks a bit when there's long text.

core/src/components/picker-column/picker-column.tsx
amandaejohnston: That gives me a type error:
```
Type 'number | null' is not assignable to type 'string | number | undefined'.
  Type 'null' is not assignable to type 'string | number | undefined'.
```

XoL1507 (approved):

core/src/components/picker-column/picker-column.scss
amandaejohnston: Fixed in https://github.com/ionic-team/ionic-framework/pull/28603/commits/cce4c0fabbd907629f7f6384082de8234ff9db9a.

core/src/components/picker-column/picker-column.scss
amandaejohnston: Done in https://github.com/ionic-team/ionic-framework/pull/28603/commits/4547e18428b47c13eb66ecd94884a4e7c0d79638.

core/src/components/picker/test/basic/picker.e2e.ts-snapshots/picker-popover-diff-md-ltr-Mobile-Chrome-linux.png
amandaejohnston: Not sure; I'll try reverting it since I've seen this test flake before.

core/src/components/picker-column/picker-column.md.scss
amandaejohnston: That's true 🤔 I'll revert this for now. We can always adjust it if people report it later.

core/src/components/picker/test/basic/picker.e2e.ts-snapshots/picker-popover-diff-md-ltr-Mobile-Chrome-linux.png
amandaejohnston: Well, it came back 😆 I'm admittedly not sure why this changed, but I don't see any visual difference, so I think it's okay.

core/src/components/picker-column/picker-column.tsx
liamdebeasi: Spoke with Sean. Reason for this is Stencil's Host functional component isn't typed the same way as the JSX elements. You can give it just about anything (including tabindex={null}. tabindex={undefined} is correct here because the JSX elements have stricter types.

core/src/components/picker-column/picker-column.scss
amandaejohnston: Added in https://github.com/ionic-team/ionic-framework/pull/28603/commits/15dd7a8df15d2cae5d7d86dc4a7de97285410944, updating screenshots now 👍

core/src/components/picker-column/picker-column.scss
```diff
@@ -22,17 +60,49 @@
...
+  justify-content: end;
+
+  /**
+   * The prefix's location is not RTL-aware, so we need an explicit right padding.
+   * This is added in addition to the column's own padding so that there is white
+   * space between the slot and the column's focus highlight.
+   */
+  /* stylelint-disable-next-line property-disallowed-list */
+  padding-right: 16px;
```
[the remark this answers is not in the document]
amandaejohnston: Done across https://github.com/ionic-team/ionic-framework/pull/28603/commits/7b9f26146f0f0bc33187398dea8d58b1326844a6 and https://github.com/ionic-team/ionic-framework/pull/28603/commits/452ee5b52d6003616e210789c4eded14f35353e1, updating screenshots now 👍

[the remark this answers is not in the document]
amandaejohnston: Ah, that `width: 100%` broke stuff, one sec while I fix that ⌛

thetaPC (approved):
LGTM

liamdebeasi (approved):
Great job!

sean-perkins:
Going to remove myself from the reviewers since this already has +2. Let me know if you do want me to look through the changes!

PR #28607, opened 2023-11-29
feat(radio): add shadow part for label

mapsandapps:
Issue number: Part of #28300 

---------

Developers are unable to adjust margin, width, etc. of the radio label

## What is the new behavior?

- Radio label has a shadow part.

- [ ] Yes
- [x] No

## Other information

As part of this work, I investigated moving `pointer-events: none` up the DOM tree so developers wouldn't be able to override it with the shadow part. In my testing, I was unable to see any difference in behavior with vs without `pointer-events: none`. Therefore, I removed it entirely.

### DIFF
--- a/core/api.txt
+++ b/core/api.txt
@@ -1029,6 +1029,7 @@ ion-radio,css-prop,--color
 ion-radio,css-prop,--color-checked
 ion-radio,css-prop,--inner-border-radius
 ion-radio,part,container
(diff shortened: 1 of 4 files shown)


amandaejohnston (approved):
I tested TalkBack and VoiceOver, but I couldn't find any behavior differences either, so I think we're good 👍

PR #28609, opened 2023-11-30
feat: WCAG AA compliant dark theme colors

sean-perkins:
Issue number: Internal

---------

Ionic's proposed color palette for dark theme is not WCAG AA compliant. 

## What is the new behavior?

- Updates the dark theme (`dark.css`) file with the proposed color palette for Level AA color contrast compliance.
- Adds tests for verifying color contrast requirements for the following cases:
  - Base color on dark theme background (`#00000` on iOS an  `#121212` on MD)
  - Contrast color on base color
  - Base color on base color with 0.08 opacity
  - Base color on base color with 0.12 opacity
  - Base color on base color with 0.16 opacity

- [ ] Yes
- [x] No


core/scripts/testing/themes/dark.css
```diff
@@ -53,19 +53,19 @@
   --ion-color-dark-shade: #d7d8da;
   --ion-color-dark-tint: #f5f6f9;
 
-  --ion-color-medium: #989aa2;
-  --ion-color-medium-rgb: 152, 154, 162;
+  --ion-color-medium: #86919c;
```
sean-perkins, line 56: Medium and light are extremely close, even though they pull from the carbon vs. gray design tokens. 

I'm open to suggestions here on how we can strike a balance of color contrast with a wider color separation between light and medium.

core/scripts/testing/styles.css
```diff
@@ -48,3 +48,35 @@ html.ios.ios {
   --ion-font-family: -apple-system, BlinkMacSystemFont, "iosTestingFont", sans-serif;
   font-family: -apple-system, BlinkMacSystemFont, "iosTestingFont", sans-serif;
 }
+
```
sean-perkins, line 51: These classes are going to be useful for light (fallback) theme color testing as well.

I can move them to just the explicit tests if desired, but placing them in the shared test stylesheet cuts down on a lot of repeat boilerplate. These classes may also prove to be useful in testing other color contrast feature work.

sean-perkins: Never mind, these selectors directly conflict with the same selectors we use for color support in components. Adjusted the approach to just inject the styles for these tests: https://github.com/ionic-team/ionic-framework/pull/28609/commits/61dfd177718a8a110c70b0048be3bdb99d9c293f

liamdebeasi (approved):
One test change, but otherwise looks good.

core/scripts/testing/themes/dark.css
```diff
@@ -53,19 +53,19 @@
...
-  --ion-color-light: #222428;
-  --ion-color-light-rgb: 34, 36, 40;
-  --ion-color-light-contrast: #ffffff;
-  --ion-color-light-contrast-rgb: 255, 255, 255;
-  --ion-color-light-shade: #1e2023;
-  --ion-color-light-tint: #383a3e;
+  --ion-color-medium-shade: #768089;
+  --ion-color-medium-tint: #929ca6;
+
+  --ion-color-light: #959595;
```
brandyscarney, line 63: The `light` color in `dark` mode should actually be a dark hex value:

Screenshot 2023-12-01 at 11 41 22 AM

Resolving this will probaly also fix the issue of `light` and `medium` being too close.

core/scripts/testing/themes/dark.css
brandyscarney: I actually think the `#959595` color you used for light looks closer to what was there. Does that one work?

core/scripts/testing/themes/dark.css
sean-perkins: Thank you! I'll also create a test template to preview all these color changes - I should have done that originally!

sean-perkins: Ok this is ready for another look: https://github.com/ionic-team/ionic-framework/pull/28609/commits/c87940a29c91f3c4992b1edea5c2e3bdca9fbe92

XoL1507 (changes requested):
sp/FW-5622

XoL1507 (changes requested):
sp/FW-5622

brandyscarney (approved):

PR #28610, opened 2023-11-30
chore(jest): toHaveShadowPart custom matcher

sean-perkins:
Issue number: N/A

---------

## What is the new behavior?

- Adds a custom matcher to the jest configuration for Stencil to enable using `expect(el).toHaveShadowPart('partName')`

- [ ] Yes
- [x] No


sean-perkins:
Let me know if you think this approach is worth pursuing/merging. I think custom matchers could help reduce some of the test boilerplate required for these common scenarios.

Issue #28611, opened 2023-11-30
bug: RootParams of `ion-nav` not detected if in kebab-case


Ericlm:
### Prerequisites

- [X] I have read the [Contributing Guidelines](https://github.com/ionic-team/ionic-framework/blob/main/.github/CONTRIBUTING.md#creating-an-issue).
- [X] I agree to follow the [Code of Conduct](https://ionicframework.com/code-of-conduct).
- [X] I have searched for [existing issues](https://github.com/ionic-team/ionic-framework/issues) that already report this problem, without success.

### Ionic Framework Version

v7.x

### Current Behavior

I created an Ionic-Vue project, and when using the `ion-nav` component, the root component doesn't get any props if `ion-nav` uses kebab-case but it works perfectly with camelCase.

### Expected Behavior

The `ion-nav` component should not be sensitive to casing, just like other Vue components, and recommended by the linter rules: https://eslint.vuejs.org/rules/attribute-hyphenation.html#vue-attribute-hyphenation

### Steps to Reproduce

1. Clone https://github.com/Ericlm/ionic-vue-nav-root-params
2. Go to `Tab1Page.vue` file and look for the two `IonNav` components in the template
3. The second one doesn't work (because it is in kebab-case)

### Code Reproduction URL

https://github.com/Ericlm/ionic-vue-nav-root-params

### Ionic Info

`ionic info` doesn't seem to work with my project (?)
```
[ERROR] Invalid project type: vue-vite (project config: ./ionic.config.json).
        
        Project type must be one of: angular, ionic-angular, ionic1, custom, react, vue

Bad project type: undefined
```

Ericlm:
*Maybe the same issue with `componentProps` on `ion-nav-link`.

liamdebeasi:
Can you make your repo public?

Ericlm:
I may take a look at it! 🙂

liamdebeasi:
Thanks for the issue. This has been resolved via https://github.com/ionic-team/ionic-framework/pull/28615, and a fix will be available in an upcoming release of Ionic Framework.

PR #28615, opened 2023-11-30
fix(nav): adds component props to allow kebab-case

Ericlm:
Issue number: #28611

---------

It's not possible to pass props that are not camelCase to the `IonNav` component.

## What is the new behavior?

- It is now possible to set a props with kebab-case instead of camelCase (for example, `root-params` instead of `rootParams`)

- [ ] Yes
- [X] No

## Other information

⚠️ This is my first PR for ionic so I hope I didn't miss important steps into the process. I also checked on my project that the fix is working well. Thank you! 🙂

### DIFF
--- a/packages/vue/src/components/IonNav.ts
+++ b/packages/vue/src/components/IonNav.ts
@@ -4,7 +4,7 @@ import { defineComponent, h, shallowRef } from "vue";
 
 import { VueDelegate } from "../framework-delegate";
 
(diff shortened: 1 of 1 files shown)


sean-perkins (approved):
Great work! Thanks for your contribution!

liamdebeasi (approved):
Verified fix using dev build `7.5.8-dev.11701374238.101cd738`. Thanks for contributing this fix!

PR #28618, opened 2023-12-01
fix(angular): add missing menu controller methods

### LINKED ISSUE #20053: bug: angular, isAnimating() not defined on menu controller provider
**Ionic version:**
4.11.5

**Current behavior:**
when i import menu controller into a page component, i saw from the official ionic website listed got a method to detect is the menu animating or not. [Link to Doc](https://ionicframework.com/docs/api/menu-controller)

image

but i couldn't find the method when writing code, IDE prompting method not found during compile and browser also say method not found.

**Expected behavior:**
it should be able to find the method and return the menu is animating or not.

**Steps to reproduce:**
1) just import the menu controller in any module, and write like ```this.menu.isAnimating()```, it will show not found error

**Related code:**

**Ionic info:**

image

image

**Ionic Info**
```
Ionic:

   ionic (Ionic CLI)             : 4.11.0 (C:\Users\oks\AppData\Roaming\npm\node_modules\ionic)
   Ionic Framework               : @ionic/angular 4.11.5
   @ionic/angular-toolkit        : 2.1.1

Cordova:

   Cordova Platforms     : browser 5.0.4
   Cordova Plugins       : cordova-plugin-ionic-keyboard 2.2.0, cordova-plugin-ionic-webview 4.1.3, (and 5 other plugins)

System:

   NodeJS : v10.15.3 (D:\Program\nodejs\node.exe)
   npm    : 6.4.1
   OS     : Windows 10
```

thetaPC:
Issue number: resolves #20053 

---------

There are a few methods that that are missing for the menu in the Angular packages. This leads to users to not being able to use methods like `isAnimating()`. 

## What is the new behavior?

- The missing methods have been added by implementing `MenuControllerI`.

- [ ] Yes
- [x] No

## Other information

Dev build: 7.5.8-dev.11701461830.1be851fd

### DIFF
--- a/packages/angular/common/src/providers/menu-controller.ts
+++ b/packages/angular/common/src/providers/menu-controller.ts
@@ -1,5 +1,4 @@
 import type { MenuControllerI, AnimationBuilder, MenuI, Animation } from '@ionic/core/components';
-// import { Animation } from '@ionic/core/components';
 
 export class MenuController implements MenuControllerI {
   constructor(private menuController: MenuControllerI) {}
--- a/packages/angular/common/src/providers/menu-controller.ts
+++ b/packages/angular/common/src/providers/menu-controller.ts
@@ -1,6 +1,7 @@
-import type { MenuControllerI } from '@ionic/core/components';
+import type { MenuControllerI, AnimationBuilder, MenuI } from '@ionic/core/components';
+import { Animation } from '@ionic/core/components';
 
-export class MenuController {
+export class MenuController implements MenuControllerI {
   constructor(private menuController: MenuControllerI) {}
 
   /**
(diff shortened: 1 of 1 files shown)


packages/angular/common/src/providers/menu-controller.ts
```diff
@@ -1,6 +1,7 @@
-import type { MenuControllerI } from '@ionic/core/components';
+import type { MenuControllerI, AnimationBuilder, MenuI } from '@ionic/core/components';
+import { Animation } from '@ionic/core/components';
```
sean-perkins, line 2: Can `Animation` be moved to a type import in the previous import statement?

sean-perkins (changes requested):

packages/angular/common/src/providers/menu-controller.ts
thetaPC: Good catch, thanks!

packages/angular/common/src/providers/menu-controller.ts
```diff
@@ -1,6 +1,7 @@
-import type { MenuControllerI } from '@ionic/core/components';
+import type { MenuControllerI, AnimationBuilder, MenuI, Animation } from '@ionic/core/components';
+// import { Animation } from '@ionic/core/components';
```
sean-perkins, line 2: This needs to be removed.

sean-perkins (changes requested):

packages/angular/common/src/providers/menu-controller.ts
thetaPC: 🤦‍♀️

sean-perkins (approved):
Great work!

PR #28621, opened 2023-12-01
feat(picker-column): add styles, disabled and active states

liamdebeasi:
Issue number: Internal

---------

We'd like to introduce a new API to the picker column to allow developers to manipulate the picker options right in the DOM. Currently developers need to pass an object which is quite cumbersome as you need to re-pass that object anytime you want to change a single option.

## What is the new behavior?

This PR does some of the prep work to make integrating the picker column option into picker column easier. It does the following things:

1. Copies the base button styles from picker-column over to picker-column-option
2. Creates active and disabled states (plus styles)
3. Add tests for base styles

Note: There are duplicate styles across picker-column and picker-column-option that will be cleaned up once picker-column-option is integrated and shown to be functional.

- [ ] Yes
- [x] No

### DIFF
--- a/core/src/components/picker-column-option/picker-column-option.ios.scss
+++ b/core/src/components/picker-column-option/picker-column-option.ios.scss
@@ -1,1 +1,1 @@
-@import "./picker-column-option.scss";
+@import "./picker-column-option";


core/src/components/picker-column-option/picker-column-option.tsx
```diff
@@ -50,19 +65,17 @@ export class PickerColumnOption implements ComponentInterface {
...
-          class={{
-            'picker-opt': true,
-            'picker-opt-disabled': !!disabled,
-          }}
-          disabled={disabled}
-        >
+      <Host
+        class={createColorClasses(color, {
+          [mode]: true,
+          ['option-disabled']: disabled,
```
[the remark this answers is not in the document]
liamdebeasi: Correct, picker-column is in charge of adding that.

I opted to change the name since `picker-item-active` doesn't align with the component name of `ion-picker-column-option` (item vs option). The `PICKER_ITEM_ACTIVE_CLASS` usage will be cleaned up in https://github.com/ionic-team/ionic-framework/pull/28622

thetaPC (approved):
LGTM

core/src/components/picker-column-option/picker-column-option.ios.scss
```diff
@@ -0,0 +1 @@
+@import "./picker-column-option.scss";
```
sean-perkins, line 1: Our existing components leave off the file extension with importing another SASS file.

core/src/components/picker-column-option/picker-column-option.md.scss
```diff
@@ -0,0 +1,5 @@
+@import "./picker-column-option.scss";
```
sean-perkins, line 1: Our existing components leave off the file extension with importing another SASS file.

sean-perkins (approved):
Few minor clean-up suggestions, rest looks good.

PR #28622, opened 2023-12-01
feat: integrate column-option into column

liamdebeasi:
Issue number: Internal

---------

We'd like to introduce a new API to the picker column to allow developers to manipulate the picker options right in the DOM. Currently developers need to pass an object which is quite cumbersome as you need to re-pass that object anytime you want to change a single option.

## What is the new behavior?

- `ion-picker-column` now uses `ion-picker-column-option` for displaying options in a column
- Scrolling the column and clicking an enabled option should update the value of the column/emit ionChange
- Loading the picker with a valid value already set

There are a few things missing from this PR. I plan to address this separately and have disabled their tests for now:

1. Ensure column option is disabled if the parent column is disabled
2. Keyboard entry for numbers
3. Setting a value initially if nothing is set/the value is out of bounds -- There are several ways I can tackle this, but based on the feedback on this PR I may change my approach so figured it would be best to hold off for now.

Notes to reviewers:
- There is likely duplicate/unneeded. I plan to focus on getting functionality implemented + tests passing before I go ripping other code out. I added TODO comments  so I can keep track of this.
- There should be no visual diffs as a result of this change
- When updating the datetime tests I tried to rely on public APIs for our components as much as possible instead of diving into the Shadow DOM. My hope is that this reduces the need for test updates in the event we make more internal changes. Let me know if you have ideas on how I can improve the test changes in this PR!

- [x] Yes
- [ ] No

The shadow parts on the picker column options have been removed since they are in the light DOM now. The `wheel-item` and `active` parts remain when targeting the wheel inside of `ion-datetime`.


core/src/components/picker-column/picker-column.tsx
```diff
@@ -352,23 +365,7 @@ export class PickerColumn implements ComponentInterface {
...
-          if (dataIndex === null) {
-            return;
-          }
-
-          const index = parseInt(dataIndex, 10);
-          const selectedItem = this.items[index];
-
-          if (selectedItem.value !== this.value) {
-            this.setValue(selectedItem.value);
-          }
```
liamdebeasi, line 371: Before, we were dealing with `button` elements so we had to encode the button index so we could then find the associated item value. We don't need to do that with the option component since we can just look at the `value` property directly.

core/src/components/picker-column/picker-column.tsx
```diff
@@ -325,8 +338,8 @@ export class PickerColumn implements ComponentInterface {
           }
         }
 
-        activeEl = activeElement;
-        this.setPickerItemActiveState(activeElement, true);
+        activeEl = newActiveElement;
```
liamdebeasi, line 341: The name change isn't needed to land this PR, but I couldn't immediately tell the different between `activeEl` and `activeElement`

core/src/components/picker-column/picker-column.tsx
```diff
@@ -292,21 +294,32 @@ export class PickerColumn implements ComponentInterface {
...
-        if (activeElement === null || activeElement.disabled) {
+        /**
+         * This null check is important because activeEl
+         * can be undefined but newActiveElement can be
+         * null since we are using a querySelector.
+         * newActiveElement !== activeEl would return true
+         * below if newActiveElement was null but activeEl
+         * was undefined.
+         */
+        if (newActiveElement === null || newActiveElement.disabled) {
```
liamdebeasi, line 314: I considered making `activeEl` null instead of undefined, but I decided against it. `Array.prototype.find` returns `undefined` if an array option was not found, and `querySelector` returns `null` if the element was not found in the DOM. I decided it would be better to add this comment highlighting the importance of the `newActiveElement === null` check instead of having the other methods return a value they were not designed to return.

core/src/components/picker-column/picker-column.tsx
```diff
@@ -167,12 +173,8 @@ export class PickerColumn implements ComponentInterface {
    */
   @Method()
   async setValue(value?: string | number) {
-    const { items } = this;
     this.value = value;
-    const findItem = items.find((item) => item.value === value && item.disabled !== true);
-    if (findItem) {
-      this.ionChange.emit(findItem);
-    }
```
liamdebeasi, line 175: Not needed anymore. We could have this filter through the existing `ion-picker-column-option` elements, but none of our other grouped components do this (accordion, radio group, etc)

XoL1507 (approved):
ld/5580-integration

core/src/components/picker-column-option/picker-column-option.tsx
```diff
@@ -64,6 +64,33 @@ export class PickerColumnOption implements ComponentInterface {
...
+   * parent picker column when it is loaded
+   * so the picker column can ensure it is
+   * centered in the view.
+   */
+  componentDidLoad() {
+    const parentPickerColumn = this.el.closest('ion-picker-column');
+    if (parentPickerColumn !== null && this.value === parentPickerColumn.value) {
+      parentPickerColumn.scrollActiveItemIntoView();
+    }
+  }
```
liamdebeasi, line 79: This is new behavior. With the internal picker, the options were part of `ion-picker-column-internal`, so we knew exactly when the options were loaded. With the `ion-picker-column-option` approach, these options can be loaded at any time (such as in an `*ngIf`), so we need the options to tell the column when it is ready to be scrolled into view.

core/src/components/picker-column/test/disabled/picker-column.e2e.ts
```diff
@@ -1,123 +1,75 @@
...
       `,
         config
       );
 
-      const pickerItems = page.locator('ion-picker-column .picker-item:not(.picker-item-empty, [disabled])');
+      const pickerItems = page.locator('ion-picker-column ion-picker-column-option button:not([disabled])');
 
       expect(await pickerItems.count()).toBe(3);
     });
+    // TODO FW-5580 move this to a spec test in picker-column-option
```
liamdebeasi, line 28: This is much easier to test using a spec test now, so I'd like to update this in a future PR.

core/src/components/picker-column/test/disabled/picker-column.e2e.ts
```diff
@@ -179,8 +123,9 @@ configs({ directions: ['ltr'] }).forEach(({ title, screenshot, config }) => {
 /**
  * This behavior does not vary across modes/directions.
  */
+// TODO FW-5580 fix this
 configs({ modes: ['ios'], directions: ['ltr'] }).forEach(({ title, config }) => {
-  test.describe(title('picker-column: disabled column'), () => {
+  test.describe.skip(title('picker-column: disabled column'), () => {
```
liamdebeasi, line 128: This will be handled in a future PR

core/src/components/picker/test/keyboard-entry/picker.e2e.ts
```diff
@@ -5,8 +5,9 @@ import type { E2ELocator } from '@utils/test/playwright/page/utils/locator';
 /**
  * This behavior does not vary across modes/directions.
  */
+// TODO FW-5580 fix this functionality
 configs({ modes: ['ios'], directions: ['ltr'] }).forEach(({ title, config }) => {
-  test.describe(title('picker: keyboard entry'), () => {
+  test.describe.skip(title('picker: keyboard entry'), () => {
```
liamdebeasi, line 10: This will be handled in a future PR

core/src/components/picker-column/picker-column.tsx
```diff
@@ -130,20 +134,20 @@ export class PickerColumn implements ComponentInterface {
...
+      const firstOption = el.querySelector('ion-picker-column-option');
+
+      /**
+       * If the picker column does not have an active item and the current value
+       * does not match the first item in the picker column, that means
+       * the value is out of bounds. In this case, we assign the value to the
+       * first item to match the scroll position of the column.
+       *
+       */
+      if (firstOption !== null && firstOption.value !== value) {
```
liamdebeasi, line 149: I updated this to get the project to compile, but this likely needs more changes. This is one of the behaviors that is not working yet.

core/src/components/picker-column/picker-column.tsx
```diff
@@ -412,15 +411,25 @@ export class PickerColumn implements ComponentInterface {
   };
 
   get activeItem() {
-    // If the whole picker column is disabled, the current value should appear active
-    // If the current value item is specifically disabled, it should not appear active
-    const selector = `.picker-item[data-value="${this.value}"]${this.disabled ? '' : ':not([disabled])'}`;
+    const { value } = this;
```
liamdebeasi, line 414: We used to encode the value on the button element itself, but we don't do that anymore. As a result, we need to iterate through the components and check the value/disabled properties.

core/src/components/picker-column/picker-column.tsx
```diff
@@ -70,7 +72,7 @@ export class PickerColumn implements ComponentInterface {
   /**
    * Emitted when the value has changed.
    */
-  @Event() ionChange!: EventEmitter<PickerColumnItem>;
+  @Event() ionChange!: EventEmitter<string | number | undefined>;
```
liamdebeasi, line 75: I changed this to get the app to compile, but it likely needs more changes. It may changed based on my keyboard entry work in a future PR.

core/src/components/picker-column/picker-column.tsx
```diff
@@ -292,21 +297,67 @@ export class PickerColumn implements ComponentInterface {
...
+         * sure we are choosing the correct host otherwise
+         * the elements returns by elementsFromPoint will be
+         * retargeted. To account for this, we check to see
+         * if the picker column has a parent shadow root. If
+         * so, we use that shadow root when doing elementsFromPoint.
+         * Otherwise, we just use the document.
+         */
+        const rootNode = el.getRootNode();
+        const hasParentShadow = rootNode instanceof ShadowRoot;
+        const referenceNode = hasParentShadow ? (rootNode as ShadowRoot) : doc;
```
liamdebeasi, line 319: This changed because we moved the options from inside of picker-column's shadow DOM to outside of it.

core/src/components/picker-column-option/picker-column-option.md.scss
```diff
@@ -1,5 +1,5 @@
 @import "./picker-column-option.scss";
 
-:host(.option-active) button {
```
liamdebeasi, line 3: This allows developers to override the active color

core/src/components/picker-column-option/picker-column-option.tsx
```diff
@@ -75,8 +124,8 @@ export class PickerColumnOption implements ComponentInterface {
           ['option-disabled']: disabled,
         })}
       >
-        <button tabindex="-1" aria-label={ariaLabel} disabled={disabled}>
-          <slot>{value}</slot>
```
liamdebeasi, line 79: This causes issues with Playwright's `toHaveText` since both the slotted content and the default content are accounted for. Given that we don't have this behavior today I've opted to remove it.

sean-perkins (changes requested):

core/src/components/picker-column/picker-column.tsx
```diff
@@ -70,7 +73,7 @@ export class PickerColumn implements ComponentInterface {
   /**
    * Emitted when the value has changed.
    */
-  @Event() ionChange!: EventEmitter<PickerColumnItem>;
+  @Event() ionChange!: EventEmitter<{ value: string | number | undefined }>;
```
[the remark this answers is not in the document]
liamdebeasi: Sorry, I had commented regarding this in https://github.com/ionic-team/ionic-framework/pull/28622#discussion_r1412601566, but it was marked as outdated.

TL;DR; This will likely change again in order to address the keyboard entry behavior, so I'd prefer to keep this as-is for now. Whatever type we end up converging on I'll make sure the implementations of it are cleaned up.

core/src/components/picker-column/picker-column.tsx
```diff
@@ -412,22 +439,33 @@ export class PickerColumn implements ComponentInterface {
   };
 
   get activeItem() {
-    // If the whole picker column is disabled, the current value should appear active
-    // If the current value item is specifically disabled, it should not appear active
-    const selector = `.picker-item[data-value="${this.value}"]${this.disabled ? '' : ':not([disabled])'}`;
+    const { value } = this;
+    const options = Array.from(
+      this.el.querySelectorAll('ion-picker-column-option') as NodeListOf<HTMLIonPickerColumnOptionElement>
```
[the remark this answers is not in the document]
liamdebeasi: https://github.com/ionic-team/ionic-framework/pull/28622/commits/d48c41db5ba2c169b3906ef5d8d5e2367b47872e

core/src/components/picker-column/picker-column.tsx
```diff
@@ -412,22 +439,33 @@ export class PickerColumn implements ComponentInterface {
...
+    );
+    return options.find((option: HTMLIonPickerColumnOptionElement) => {
```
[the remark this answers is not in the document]
liamdebeasi: https://github.com/ionic-team/ionic-framework/pull/28622/commits/e4797f73f47ecc2db1446aeceac999ec89dda27b

core/src/components/picker-column/picker-column.tsx
```diff
@@ -412,22 +439,33 @@ export class PickerColumn implements ComponentInterface {
...
+       * If the whole picker column is disabled, the current value should appear active
+       * If the current value item is specifically disabled, it should not appear active
+       */
+      if (!this.disabled && option.disabled) {
+        return false;
+      }
 
-    return getElementRoot(this.el).querySelector(selector) as HTMLElement | null;
+      return option.value === value;
+    }) as HTMLIonPickerColumnOptionElement | undefined;
```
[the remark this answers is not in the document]
liamdebeasi: https://github.com/ionic-team/ionic-framework/pull/28622/commits/a98e85d632fb82bc98654002a0c401047401bc8a

core/src/components/picker-column/picker-column.tsx
```diff
@@ -101,7 +104,9 @@ export class PickerColumn implements ComponentInterface {
          * Because this initial call to scrollActiveItemIntoView has to fire before
          * the scroll listener is set up, we need to manage the active class manually.
          */
-        const oldActive = getElementRoot(el).querySelector(`.${PICKER_ITEM_ACTIVE_CLASS}`);
+        const oldActive = getElementRoot(el).querySelector(
```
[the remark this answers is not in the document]
liamdebeasi: https://github.com/ionic-team/ionic-framework/pull/28622/commits/33a21d367530584da9fe3ac11540768741f02e11

core/src/components/picker-column/picker-column.tsx
```diff
@@ -292,21 +297,59 @@ export class PickerColumn implements ComponentInterface {
...
+          return;
+        }
+
+        const elementsAtPoint = referenceNode.elementsFromPoint(centerX, centerY) as HTMLIonPickerColumnOptionElement[];
+
+        /**
+         * elementsFromPoint can returns multiple elements
+         * so find the relevant picker column option if one exists.
+         */
+        const newActiveElement = elementsAtPoint.find((el) => el.tagName === 'ION-PICKER-COLUMN-OPTION') as
```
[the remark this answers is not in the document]
liamdebeasi: https://github.com/ionic-team/ionic-framework/pull/28622/commits/0ef015b5be4d4bbdc1ec4976081a383a760da2e6

core/src/components/picker-column/picker-column.tsx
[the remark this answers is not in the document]
sean-perkins: Sounds good, I'm ok with addressing this later 👍

core/src/components/datetime/datetime.tsx
```diff
@@ -1647,7 +1646,18 @@ export class Datetime implements ComponentInterface {
 
           ev.stopPropagation();
         }}
-      ></ion-picker-column>
+      >
+        {items.map((item: PickerColumnItem) => (
```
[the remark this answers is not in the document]
liamdebeasi: https://github.com/ionic-team/ionic-framework/pull/28622/commits/f9d01b90410c137aa57b8a36ae5f09aa68a2e211

core/src/components/datetime/datetime.tsx
```diff
@@ -1764,7 +1774,18 @@ export class Datetime implements ComponentInterface {
 
           ev.stopPropagation();
         }}
-      ></ion-picker-column>
+      >
+        {days.map((day: PickerColumnItem) => (
```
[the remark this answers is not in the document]
liamdebeasi: https://github.com/ionic-team/ionic-framework/pull/28622/commits/f9d01b90410c137aa57b8a36ae5f09aa68a2e211

core/src/components/datetime/datetime.tsx
```diff
@@ -1853,7 +1883,18 @@ export class Datetime implements ComponentInterface {
 
           ev.stopPropagation();
         }}
-      ></ion-picker-column>
+      >
+        {years.map((year: PickerColumnItem) => (
```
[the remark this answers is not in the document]
liamdebeasi: https://github.com/ionic-team/ionic-framework/pull/28622/commits/f9d01b90410c137aa57b8a36ae5f09aa68a2e211

core/src/components/datetime/datetime.tsx
```diff
@@ -1916,7 +1956,18 @@ export class Datetime implements ComponentInterface {
 
           ev.stopPropagation();
         }}
-      ></ion-picker-column>
+      >
+        {hoursData.map((hour: PickerColumnItem) => (
```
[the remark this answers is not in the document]
liamdebeasi: https://github.com/ionic-team/ionic-framework/pull/28622/commits/f9d01b90410c137aa57b8a36ae5f09aa68a2e211

core/src/components/datetime/datetime.tsx
```diff
@@ -1945,7 +1995,18 @@ export class Datetime implements ComponentInterface {
 
           ev.stopPropagation();
         }}
-      ></ion-picker-column>
+      >
+        {minutesData.map((minute: PickerColumnItem) => (
```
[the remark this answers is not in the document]
liamdebeasi: https://github.com/ionic-team/ionic-framework/pull/28622/commits/f9d01b90410c137aa57b8a36ae5f09aa68a2e211

core/src/components/datetime/datetime.tsx
```diff
@@ -1981,7 +2041,18 @@ export class Datetime implements ComponentInterface {
 
           ev.stopPropagation();
         }}
-      ></ion-picker-column>
+      >
+        {dayPeriodData.map((dayPeriod: PickerColumnItem) => (
```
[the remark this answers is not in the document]
liamdebeasi: https://github.com/ionic-team/ionic-framework/pull/28622/commits/f9d01b90410c137aa57b8a36ae5f09aa68a2e211

core/src/components/datetime/datetime.tsx
```diff
@@ -2529,3 +2600,5 @@ export class Datetime implements ComponentInterface {
 }
 
 let datetimeIds = 0;
+const WHEEL_ITEM_PART = 'wheel-item';
+const WHEEL_ITEM_ACTIVE_PART = `${WHEEL_ITEM_PART} active`;
```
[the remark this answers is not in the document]
liamdebeasi: https://github.com/ionic-team/ionic-framework/pull/28622/commits/cf3d028d696ac05f91993144d9356116ae4fb062

core/src/components/datetime/datetime.tsx
```diff
@@ -1647,7 +1646,18 @@ export class Datetime implements ComponentInterface {
 
           ev.stopPropagation();
         }}
-      ></ion-picker-column>
+      >
+        {items.map((item: PickerColumnItem) => (
+          <ion-picker-column-option
```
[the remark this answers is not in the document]
liamdebeasi: Wouldn't that mean we would be inconsistent with how `ion-picker-column` is used in datetime? Right now each `render*` function renders an `ion-picker-column` directly in the return statement. The proposed change would keep rendering the duplicate `ion-picker-column` inline in each `render*` function but move the `ion-picker-column-option` components to a separate function.

I'm fine making this change such that we are consistent, but I'd advocate for that to be done separately.

core/src/components/picker-column-option/picker-column-option.tsx
```diff
@@ -60,12 +67,54 @@ export class PickerColumnOption implements ComponentInterface {
...
+    }
+  }
+
+  /**
+   * When an option is clicked update the
+   * parent picker column value. This
+   * component will handle centering the option
+   * in the column view.
+   */
+  onClick() {
```
amandaejohnston, line 109: When clicking an option to select it (vs. scrolling to it), I'm no longer getting a smooth scroll animation; it just snaps right to the option. Is that expected?

core/src/components/picker/test/basic/picker.e2e.ts
```diff
@@ -9,11 +9,11 @@ configs({ directions: ['ltr'] }).forEach(({ title, screenshot, config }) => {
...
       await page.goto(`/src/components/picker/test/basic`, config);
 
-      const fullStack = page.locator('#inline button[data-value="full-stack"]');
-      const onion = page.locator('#inline button[data-value="onion"]');
+      const fullStack = page.locator('#inline ion-picker-column-option[data-test-value="full-stack"]');
+      const onion = page.locator('#inline ion-picker-column-option[data-test-value="onion"]');
 
-      await expect(fullStack).toHaveClass(/picker-item-active/);
-      await expect(onion).toHaveClass(/picker-item-active/);
+      await expect(fullStack).toHaveClass(/option-active/);
```
thetaPC, line 15: It seems to have taken two tries to pass. I wonder if there's a way to minimize the retries. This isn't a blocker but can be worth a conversation.

core/src/components/picker/test/basic/index.html
thetaPC, line 1: I'm not able to click on any of the options for `inline`, `one numeric input`, and `two numeric input`. I can only access them through the keys. The overlays work fine without issues.

core/src/components/picker-column-option/picker-column-option.tsx
liamdebeasi: good catch! fixed in https://github.com/ionic-team/ionic-framework/pull/28622/commits/da2db5ca63f4a773682ddb311086d5b73281e571

core/src/components/picker/test/basic/index.html
liamdebeasi: Steps to repro? It's working fine on my end:

https://github.com/ionic-team/ionic-framework/assets/2721089/426c7484-ad16-4793-a5cb-03c0538d9dfe

amandaejohnston (approved):

core/src/components/picker/test/basic/index.html
thetaPC: This is happening for iOS mode on Firefox. No issues on Chrome and Safari.

https://github.com/ionic-team/ionic-framework/assets/13530427/30e5dfbf-12ab-439e-8e77-fa8d3b92584e

liamdebeasi: Discussed offline. Maria is not able to reproduce this anymore, so going to mark this as resolved. If you can reproduce again let me know!

thetaPC (approved):
LGTM, smooth on click also works

liamdebeasi:
Spoke with Sean offline and he's good with merging.

PR #28630, opened 2023-12-04
test(overlays): add Axe tests for color contrast

amandaejohnston:
Issue number: Internal

---------

- The action-sheet, alert, and toast components have tests for Axe violations, but the `color-contrast` rule is disabled.
- The select-popover component does not test for Axe violations.
- For all of the above plus loading, dark mode test for color contrast are not included.

## What is the new behavior?

- Existing Axe tests pulled out into separate `configs` blocks that test both light and dark themes. Markup has also been converted to use `page.setContent` instead of `page.goto` to support automatic theme switching.
- `color-contrast` rule re-enabled on action-sheet, alert, and toast. The rule was disabled due to contrast issues with the page content covered up by the backdrop, but with the new setup, the overlay is the only thing on the page.
- For select-popover, the Axe test has been limited to only color contrast testing for now. The component had several Axe failures that will be addressed separately to keep scope down. (We didn't previously have any Axe testing on this component, which is why this wasn't caught.)
- HTML file for the loading a11y test removed, since we're no longer using it.

- [ ] Yes
- [x] No


core/src/components/loading/test/a11y/loading.e2e.ts
```diff
@@ -2,27 +2,48 @@ import AxeBuilder from '@axe-core/playwright';
 import { expect } from '@playwright/test';
 import { configs, test } from '@utils/test/playwright';
 
-configs({ modes: ['ios'], directions: ['ltr'] }).forEach(({ title, config }) => {
+configs({ modes: ['ios'], directions: ['ltr'], themes: ['light', 'dark'] }).forEach(({ title, config }) => {
```
liamdebeasi, line 5: Optional: We don't use the `index.html` file in this test anymore. Should we remove it?

core/src/components/toast/test/a11y/toast.e2e.ts
```diff
@@ -33,7 +49,11 @@ configs({ directions: ['ltr'] }).forEach(({ title, screenshot, config }) => {
       /**
        * IonToast overlays the entire screen, so
        * Axe will be unable to verify color contrast
-       * on elements under the toast.
+       * on elements under the toast. To avoid needing
+       * to spin up a controller toast using page.setContent
+       * (to support automatic dark theme testing), color
+       * contrast is checked on an inline toast in a
+       * separate test above.
        */
```
liamdebeasi, line 57: Why does the color-contrast test work on the inline toast but not the controller toast?

liamdebeasi (commented):
Looks good! One non-blocking suggestion and one question.

core/src/components/toast/test/a11y/toast.e2e.ts
liamdebeasi: Something like this might work: https://github.com/ionic-team/ionic-framework/blob/fc88613fefa019a3b695a2c6e10c85cd3ce79ae8/core/src/components/toast/test/standalone/index.html#L16C1-L19

amandaejohnston: Was able to get it working 👍

liamdebeasi (approved):
Great work!

PR #28631, opened 2023-12-04
fix(select): do not collapse to width: 0 when placed in flex container

liamdebeasi:
Issue number: Internal

---------

We currently apply a workaround to `ion-select` so it can wrap correctly inside of `ion-item`: https://github.com/ionic-team/ionic-framework/blob/357b8b2beb29b95d53ef043af349067be1d32658/core/src/components/select/select.scss#L99-L103

However, this causes issues when a parent element has `display: flex` because the `ion-select` width becomes 0.

## What is the new behavior?

- In order to get the desired behavior, we need the `ion-select` (and other elements in the default slot) to either truncate or wrap within its own container and then have the entire container (i.e. the entire `ion-select`) wrap to the next line once the container is too small.

To achieve this, I needed to set a min-width on `.item-inner` to define the point at which the element should wrap to the next line. I also changed the flex basis from `auto` to `0` which means the initial main size of the flex item will be 0px. In reality, this will be `--inner-min-width` since we also set `min-width: var(--inner-min-width)`. I used `0` for simplicity but I can change this to use the CSS variable if that's more clear. Since we also set `flex-grow: 1` we indicate that the element can grow from that basis (but it cannot shrink).

I chose `--inner-min-width: 4rem` to minimize the number of diffs. We can certainly change this, but it may cause some diffs as certain elements will start wrapping sooner. I also chose to use `rem` because having a fixed min-width means that fewer characters are going to fit in the same space as text scales.

I made this a CSS variable but left it undocumented. If developers need a way of changing this `min-width` they can request it and we can easily expose this variable. However, I think `4rem` is small enough that this should be sufficient.

- [ ] Yes
- [x] No

## Other information

The visual diffs here are correct. The table below shows the screenshot group and an explanation for why the changes are correct.

| Path | Example | Details |
| - | - | - |
| `disabled` | [Link](https://github.com/ionic-team/ionic-framework/pull/28631/files#diff-d529716f95f7a7aa82c88588104220775b728af67077f48cd47a8afa04423143) | The searchbar is able to shrink slightly to fit on the same line as the checkbox at the bottom. |
| `highlight` | [Link](https://github.com/ionic-team/ionic-framework/pull/28631/files#diff-0b64f24c91393923701d1ced4e330a1c6b926d72ee461b8ab1e135e708be3457) | We're changing how small the main content can get, so the input is only wrapping once it gets to `--inner-min-width`. |
| `legacy/fill` | [Link](https://github.com/ionic-team/ionic-framework/pull/28631/files#diff-2ef8dbfa5e69e2b96c3e1ed29ab962f08cf5ba2aaf2af773e40bd143e38a4bef) | We're changing how small the main content can get, so the input is only wrapping once it gets to `--inner-min-width`. |
| `slotted-inputs` | [Link](https://github.com/ionic-team/ionic-framework/pull/28631/files#diff-2f173c7303969d6a6c58f30a618cebc3caf918d3761fc83df5642fd48dfabd7b) | We're changing how small the main content can get, so the range is only wrapping once it gets to `--inner-min-width`. |

`slotted-inputs` note: I'd argue many of these examples are not best practices. For example, adding a range in the start slot and the end slot is a bit unusual. I'm not aware of any native apps that implement this pattern.

popover note: I [removed the `ion-item` from the `popover/test/async` test](https://github.com/ionic-team/ionic-framework/pull/28631/commits/331fcb859c4ad2d79911e8bb39341cea5695ec2f). There was a diff because the min-width increased, but IMO that component should not be used in the popover test since we want to test the popover, not the item.

--------

Demo:

| `feature-7.6` | `branch` |
| - | - |
| <video src="https://github.com/ionic-team/ionic-framework/assets/2721089/693d4947-fa33-460d-bc7f-7b96b6338032"></video> | <video src="https://github.com/ionic-team/ionic-framework/assets/2721089/df35ca73-87aa-4e76-9bb7-99f0f2810640"></video> |

(In this demo I updated the `ion-select` to wrap within its own container first instead of truncate. We may want to consider doing this by default, but I think this is out of scope for this task)


amandaejohnston (approved):
Confirmed that this works with the extra `ion-item` 👍 Once this merges, I'll get a separate PR up to remove the `min-width` patch from the select slot playground.

core/src/components/item/test/disabled/item.e2e.ts-snapshots/item-disabled-diff-ios-ltr-Mobile-Chrome-linux.png
brandyscarney, line 1: Is it possible to fix this so the month and year select are still on the same line? This is a regression even from `main`: https://github.com/ionic-team/ionic-framework/blob/main/core/src/components/item/test/disabled/item.e2e.ts-snapshots/item-disabled-diff-ios-ltr-Mobile-Chrome-linux.png

| main | branch |
| ---| ---|
| image | image |

core/src/components/popover/test/async/index.html
brandyscarney, line 1: Was this change needed to prevent the popover width collapsing?

core/src/components/item/test/slotted-inputs/item.e2e.ts-snapshots/item-slotted-inputs-ios-ltr-Mobile-Chrome-linux.png
brandyscarney, line 1: This change causes the `.item-inner` elements to wrap instead of staying on the same line, here is a comparison with `main`:

| main | branch |
| ---| ---|
| main | branch |

Is this desired?

core/src/components/item/test/legacy/fill/item.e2e.ts-snapshots/item-fill-diff-ios-ltr-Mobile-Chrome-linux.png
brandyscarney, line 1: This seems to revert it back to how it looks on `main` which is probably fine: https://github.com/ionic-team/ionic-framework/blob/main/core/src/components/item/test/legacy/fill/item.e2e.ts-snapshots/item-fill-diff-ios-ltr-Mobile-Chrome-linux.png

core/src/components/item/test/highlight/item.e2e.ts-snapshots/item-highlight-diff-ios-ltr-Mobile-Chrome-linux.png
brandyscarney, line 1: This seems to revert it back to how it looks on `main` which is probably fine: https://github.com/ionic-team/ionic-framework/blob/main/core/src/components/item/test/highlight/item.e2e.ts-snapshots/item-highlight-diff-ios-ltr-Mobile-Chrome-linux.png

core/src/components/item/item.scss
```diff
@@ -51,6 +51,15 @@
...
    */
+
+  /**
+   * We change the minimum width as the
+   * font size changes. Using a fixed minimum
+   * width means that fewer and fewer characters
+   * can be displayed in the same space as the
+   * text grows.
+   */
+  --inner-min-width: 4rem;
```
brandyscarney, line 62: Why do we need a CSS variable for this? Are we expecting users to update it?

liamdebeasi: This is explained in the original post: 

> I made this a CSS variable but left it undocumented. If developers need a way of changing this min-width they can request it and we can easily expose this variable.

This was a way to minimize risk in case people are using narrow lists that are less than the minimum width. I'm fine removing it if you don't think it's needed.

core/src/components/popover/test/async/index.html
liamdebeasi: This is explained in the original post:

> popover note: I [removed the ion-item from the popover/test/async test](https://github.com/ionic-team/ionic-framework/pull/28631/commits/331fcb859c4ad2d79911e8bb39341cea5695ec2f). There was a diff because the min-width increased, but IMO that component should not be used in the popover test since we want to test the popover, not the item.

core/src/components/item/item.scss
brandyscarney: Oh oops I should've read that, sorry. 😆

core/src/components/item/test/slotted-inputs/item.e2e.ts-snapshots/item-slotted-inputs-ios-ltr-Mobile-Chrome-linux.png
liamdebeasi: Discussed offline. This is intentional: We don't specify that content in the start slot can shrink, so the browser will use the intrinsic size of the range here. Additionally, we are specifying that the main slotted content should wrap when it hits the min-width whereas it was never wrapping in main.

core/src/components/item/test/disabled/item.e2e.ts-snapshots/item-disabled-diff-ios-ltr-Mobile-Chrome-linux.png
liamdebeasi: Fixed in https://github.com/ionic-team/ionic-framework/pull/28631/commits/5be6a0db81d7fc05c2aacf40bc304ffff0717b11. Re-running screenshots now.

core/src/components/select/test/item/select.e2e.ts
```diff
@@ -45,11 +45,13 @@ configs({ directions: ['ltr'] }).forEach(({ title, screenshot, config }) => {
       await page.setContent(
         `
         <div id="container" style="display: flex">
-          <ion-item>
-            <ion-select label="Fruit" value="apple">
-              <ion-select-option value="apple">Apple</ion-select-option>
-            </ion-select>
-          </ion-item>
+          <ion-list>
```
liamdebeasi, line 48: Brandy and I discussed that it would be good to account for `ion-list` here too. Amanda had found an edge case where [Brandy's previous fix](https://github.com/ionic-team/ionic-framework/pull/28594) did not work, so we wanted to test for that here.

brandyscarney (approved):
Talked through this with Liam - the change makes sense & we can always remove the CSS variable later since it is internal. Great job!

core/src/components/item/item.scss
```diff
@@ -80,15 +89,6 @@
...
   position: relative;
 
-  // When an item containing a select is inside of a
-  // flex container the item will collapse to 0px
-  // width due to the select setting the width to 0px.
-  // By setting the flex property here, we are
-  // allowing the item to grow to fill the flex container.
-  // If the item is inside of a block container this
-  // property will be ignored.
-  flex: 1;
```
liamdebeasi, line 90: This is the old fix that brandy had that is no longer needed.

liamdebeasi:
Note: The screenshot diffs in https://github.com/ionic-team/ionic-framework/pull/28594 are correct. I removed brandy's old fix in https://github.com/ionic-team/ionic-framework/pull/28631/commits/55a091974bd58db10703b47b8368fd76170bd5dc since it is no longer needed. As a result, items in a flex container no longer grow by default. This test is verifying that the select widths don't collapse to 0px.

PR #28639, opened 2023-12-05
fix(input, textarea): clearOnInput ignores key modifiers

### LINKED ISSUE #28633: bug: IonInput type="password" clears input field on modifier key press
### Prerequisites

### Ionic Framework Version

v7.x

### Current Behavior

When IonInput is given a value of `type="password"` in `@ionic/react`, the input will clear text unexpectedly when certain keys are pressed. This is most notable when using a keyboard to navigate between controls. This behavior is not observed with the native `input` element with `type` set to password or the default `IonInput`.

### Expected Behavior

`IonInput`s should be accessible via keyboard. Pressing modifier keys (e.g., shift, command) should not clear a password input (unless combined with another key, e.g., shift + p).

### Steps to Reproduce

1. Create an `IonInput` with `type="password"`
2. Enter a value (either via keyboard or via password manager autofill)
3. Tab to the next focusable element
4. Tab back to the password input
5. Tab to the next focusable element
6. IonInput with `type="password"` is empty

### Code Reproduction URL

https://stackblitz.com/edit/vxete3-yk4b3w?file=src%2Fmain.tsx

### Ionic Info

Via StackBlitz:

Ionic:

   Ionic CLI : 5.4.16

Utility:

   cordova-res : not installed
   native-run  : not installed

System:

   NodeJS : v18.18.0
   npm    : 9.4.2
   OS     : Linux 5.0

### Additional Information

First noticed on a project using @ionic/react: 7.0.4 and validated on latest available release in StackBlitz  (i.e., "@ionic/react": "^7.5.7" and "@ionic/react-router": "^7.5.7").

liamdebeasi:
Issue number: resolves #28633

---------

In https://github.com/ionic-team/ionic-framework/pull/28005 I introduced a fix that causes "clearOnEdit" to not clear input/textarea when the Tab key was pressed. However, there were several edge cases I missed. For instance, pressing the "shift" key and then the "tab" key would still clear the input because "shift" was not explicitly excluded.

## What is the new behavior?

- Input and textarea now explicitly ignores modifier keys that do not change the value of the input
- `didInputClearOnEdit` is not set to `true` when an ignored key is pressed. Otherwise, pressing an ignored key and then an accepted key would not cause the input to be cleared. For example, pressing "shift", releasing "shift", then pressing "A" should cause the input to still get cleared.
- Added test coverage

- [ ] Yes
- [x] No

## Other information

Playwright bug report for https://github.com/ionic-team/ionic-framework/pull/28639/commits/a9f34a54f1bf5b7dfdc1401f5a339f55c4b9093b: https://github.com/microsoft/playwright/issues/28495

### DIFF
--- a/core/src/components/input/test/clear-on-edit/input.e2e.ts
+++ b/core/src/components/input/test/clear-on-edit/input.e2e.ts
@@ -30,7 +30,11 @@ configs({ modes: ['ios'], directions: ['ltr'] }).forEach(({ title, config }) => 
     });
 
     IGNORED_KEYS.forEach((ignoredKey: string) => {
-      test(`should not clear when ${ignoredKey} is pressed`, async ({ page }) => {
+      test(`should not clear when ${ignoredKey} is pressed`, async ({ page, skip }) => {
+        skip.browser(
+          (browserName: string) => browserName === 'firefox' && ignoredKey === 'Meta',
+          'Firefox incorrectly adds "OS" to the input when pressing the Meta key on Linux'
+        );
         await page.setContent(`<ion-input value="abc" clear-on-edit="true" aria-label="input"></ion-input>`, config);
 
         const input = page.locator('ion-input');
--- a/core/src/components/input/test/clear-on-edit/input.e2e.ts
+++ b/core/src/components/input/test/clear-on-edit/input.e2e.ts
@@ -18,6 +18,17 @@ configs({ modes: ['ios'], directions: ['ltr'] }).forEach(({ title, config }) => 
       await expect(input).toHaveJSProperty('value', 'h');
     });
 
+    test('should not clear the input if it does not have an initial value when typing', async ({ page }) => {
+      await page.setContent(`<ion-input label="input" value="" clear-on-edit="true"></ion-input>`, config);
+
+      const input = page.locator('ion-input');
+
+      await input.click();
+      await input.type('hello world');
+
+      await expect(input).toHaveJSProperty('value', 'hello world');
+    });
+
     IGNORED_KEYS.forEach((ignoredKey: string) => {
       test(`should not clear when ${ignoredKey} is pressed`, async ({ page }) => {
         await page.setContent(`<ion-input value="abc" clear-on-edit="true" aria-label="input"></ion-input>`, config);
--- a/core/src/components/input/test/clear-on-edit/input.e2e.ts
+++ b/core/src/components/input/test/clear-on-edit/input.e2e.ts
@@ -1,6 +1,8 @@
 import { expect } from '@playwright/test';
 import { test, configs } from '@utils/test/playwright';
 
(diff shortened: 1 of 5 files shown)


core/src/components/input/test/clear-on-edit/input.e2e.ts
```diff
@@ -1,6 +1,8 @@
 import { expect } from '@playwright/test';
 import { test, configs } from '@utils/test/playwright';
 
+const IGNORED_KEYS = ['Enter', 'Tab', 'Shift', 'Meta', 'Alt', 'Control'];
```
liamdebeasi, line 4: I could export this array from each component file and then import it in the test file, but wasn't sure if we were comfortable with that level of coupling. For example, if that array ever gets moved/deleted (even if the underlying functionality remains) these tests will start to fail.

liamdebeasi:
Reviewers: This task was completed outside of my sprint work, so sprint work-related PRs take higher priority.

core/src/components/input/test/clear-on-edit/input.e2e.ts
```diff
@@ -16,28 +18,53 @@ configs({ modes: ['ios'], directions: ['ltr'] }).forEach(({ title, config }) =>
       await expect(input).toHaveJSProperty('value', 'h');
     });
 
-    test('should not clear when enter is pressed', async ({ page }) => {
-      await page.setContent(`<ion-input value="abc" clear-on-edit="true" aria-label="input"></ion-input>`, config);
+    test('should not clear the input if it does not have an initial value when typing', async ({ page }) => {
```
liamdebeasi, line 21: This new test (and the same textarea test) was copied over  from the legacy directory. I accidentally regressed a behavior that only the legacy tests caught. Since the behavior is still relevant for the modern form control syntax, I decided to copy them over so we don't lose test coverage when we remove the legacy form syntax.

core/src/components/input/test/clear-on-edit/input.e2e.ts
thetaPC: True, but we also run into the potential issue of if the array is updated in the components then we have to remember to update it in this file. It might be best to export it since we can guarantee the right keys. The tests may fail but at least it would be an easy fix vs not testing the correct array.

thetaPC (commented):
Works as intended! Just one comment.

core/src/components/input/test/clear-on-edit/input.e2e.ts
liamdebeasi: Wouldn't that be expected anyways? i.e. if you make a change/fix a bug I'd expect a new test to be written.

thetaPC: Good point!

thetaPC: I vote for leaving it as is.

thetaPC (approved):
LGTM

PR #28640, opened 2023-12-05
test(many): add axe testing for misc components

thetaPC:
Issue number: internal

---------

The following components use `ion-color()` and are not being tested for Axe violations.

- picker-internal
- progress-bar
- router-link
- badge
- label
- refresher
- typography

## What is the new behavior?

- All the components check for accessibility violations
- Added `aria-label` to all the `ion-spinner` components within `ion-refresher` in order for it to pass an [Axe violation](https://dequeuniversity.com/rules/axe/4.8/aria-progressbar-name?application=playwright)

- [ ] Yes
- [x] No

## Other information

N/A


core/src/components/refresher-content/refresher-content.tsx
```diff
@@ -108,7 +108,7 @@ export class RefresherContent implements ComponentInterface {
           {this.pullingIcon && hasSpinner && (
             <div class="refresher-pulling-icon">
               <div class="spinner-arrow-container">
-                <ion-spinner name={this.pullingIcon as SpinnerTypes} paused></ion-spinner>
+                <ion-spinner name={this.pullingIcon as SpinnerTypes} aria-label="pulling icon" paused></ion-spinner>
```
thetaPC, line 111: `ion-refresher-content` is [failing](https://dequeuniversity.com/rules/axe/4.8/aria-progressbar-name?application=playwright) to pass Axe testing due to not passing an `aria-label` to `ion-spinner`.

core/src/components/refresher-content/refresher-content.tsx
```diff
@@ -127,7 +127,7 @@ export class RefresherContent implements ComponentInterface {
         <div class="refresher-refreshing">
           {this.refreshingSpinner && (
             <div class="refresher-refreshing-icon">
-              <ion-spinner name={this.refreshingSpinner}></ion-spinner>
+              <ion-spinner name={this.refreshingSpinner} aria-label="refreshing icon"></ion-spinner>
```
thetaPC, line 130: `ion-refresher-content` is [failing](https://dequeuniversity.com/rules/axe/4.8/aria-progressbar-name?application=playwright) to pass Axe testing due to not passing an `aria-label` to `ion-spinner`.

core/src/components/refresher/test/a11y/refresher.e2e.ts
```diff
@@ -0,0 +1,32 @@
...
+
+      const target = page.locator('body');
+
+      await page.waitForSelector('ion-refresher.hydrated', { state: 'attached' });
+
+      await dragElementBy(target, page, 0, 320, undefined, undefined, false);
+
+      await page.waitForSelector('.arrow-container', { state: 'attached' });
+
+      const results = await new AxeBuilder({ page }).disableRules('region').analyze();
```
thetaPC, line 27: The Axe test fails because a landmark role like `<main> is missing. However, when I add it in then it fails for having a nested `<main>`. The `ion-content` adds it in but still fails because it's not at the top level.

core/src/components/label/test/a11y/label.e2e.ts
```diff
@@ -82,3 +83,28 @@ configs({ directions: ['ltr'] }).forEach(({ title, screenshot, config }) => {
...
+  test.describe(title('label: a11y for ion-color()'), () => {
+    test('should not have accessibility violations when focused', async ({ page, pageUtils }) => {
+      /**
+       * All page content should be contained by landmarks (main, nav, etc.)
+       * By containing the badge in a main element, we can avoid this violation.
+       */
+      await page.setContent(
+        `
+          <main>
+            <ion-input label="Stacked" label-placement="stacked"></ion-input>
```
liamdebeasi, line 97: This does not use `ion-label` internally. You'll need to use `ion-label` directly instead.

```html
<ion-label color="primary">Label</ion-label>
```

core/src/components/badge/test/a11y/badge.e2e.ts
```diff
@@ -23,3 +24,27 @@ configs({ directions: ['ltr'] }).forEach(({ title, screenshot, config }) => {
...
+    test('should not have accessibility violations', async ({ page }) => {
+      /**
+       * All page content should be contained by landmarks (main, nav, etc.)
+       * By containing the badge in a main element, we can avoid this violation.
+       */
+      await page.setContent(
+        `
+        <main>
+          <ion-badge>123</ion-badge>
+          <ion-badge class="ion-color">123</ion-badge>
```
liamdebeasi, line 39: Why do we add `ion-color` here? `ion-badge` should use the primary color palette by default.

core/src/components/progress-bar/test/a11y/progress-bar.e2e.ts
```diff
@@ -0,0 +1,28 @@
...
+    test('should not have accessibility violations', async ({ page }) => {
+      /**
+       * All page content should be contained by landmarks (main, nav, etc.)
+       * By containing the badge in a main element, we can avoid this violation.
+       */
+      await page.setContent(
+        `
+          <main>
+            <ion-progress-bar type="indeterminate" aria-label="default"></ion-progress-bar>
+            <ion-progress-bar type="indeterminate" aria-label="class" color="ion-color"></ion-progress-bar>
```
liamdebeasi, line 16: Same question as badge

core/src/components/router-link/test/a11y/router-link.e2e.ts
```diff
@@ -0,0 +1,27 @@
...
+    test('should not have accessibility violations', async ({ page }) => {
+      /**
+       * All page content should be contained by landmarks (main, nav, etc.)
+       * By containing the badge in a main element, we can avoid this violation.
+       */
+      await page.setContent(
+        `
+          <main>
+            <ion-router-link href="#">Router Link</ion-router-link>
+            <ion-router-link href="#" class="ion-color">Router Link</ion-router-link>
```
liamdebeasi, line 16: Same question as badge

core/src/components/refresher/test/a11y/refresher.e2e.ts
```diff
@@ -0,0 +1,39 @@
...
+            <ion-refresher-content></ion-refresher-content>
+          </ion-refresher>
+        </ion-content>
+      `,
+        config
+      );
+
+      const target = page.locator('body');
+
+      await page.waitForSelector('ion-refresher.hydrated', { state: 'attached' });
```
liamdebeasi, line 21: Why is this needed?

core/src/components/refresher/test/a11y/refresher.e2e.ts
```diff
@@ -0,0 +1,39 @@
...
+
+      await dragElementBy(target, page, 0, 320, undefined, undefined, false);
+
+      await page.waitForSelector('.arrow-container', { state: 'attached' });
```
liamdebeasi, line 25: Why is this needed?

core/src/components/refresher/test/a11y/refresher.e2e.ts
liamdebeasi: I'm not sure I understand this comment. `setContent` adds the HTML as a direct descendant of `body`, so there shouldn't be a nested `main` element: https://github.com/ionic-team/ionic-framework/blob/e886e3ff2fcb8a3586a62881c5fc848f3074235d/core/src/utils/test/playwright/page/utils/set-content.ts#L57

thetaPC: The nested (second) `<main>` comes from `ion-content` [logic](https://github.com/ionic-team/ionic-framework/blob/c680ad99885ce7600d7447a5aadedd2b3f79adf9/core/src/components/content/content.tsx#L431).

So the following code:
```ts
await page.setContent(
  `
  <main>
      <ion-content>
      <ion-refresher class="refresher-native" slot="fixed">
        <ion-refresher-content></ion-refresher-content>
      </ion-refresher>
    </ion-content>
   </main>
`,
  config
);
```

would be:
```tsx
<main> // This main solves the violation for not having a landmark
   <ion-content>
       <main> // This main triggers a violation for being nested
           // stuff
       </main>
    </ion-content>
</main>
```

core/src/components/badge/test/a11y/badge.e2e.ts
thetaPC: Ah, this was accidentally left in. Thank you for catching it!

core/src/components/refresher/test/a11y/refresher.e2e.ts
liamdebeasi: Just so I understand: Only using `ion-content` is not sufficient because `main` is nested within `ion-content` when it really should be the other way around.

What about adding `role="main"` to the host of `ion-content`? That might be a better solution that nested a `main` element inside of the Shadow DOM.

core/src/components/progress-bar/test/a11y/progress-bar.e2e.ts
thetaPC: Same [response](https://github.com/ionic-team/ionic-framework/pull/28640#discussion_r1422983693)

core/src/components/router-link/test/a11y/router-link.e2e.ts
thetaPC: Same [response](https://github.com/ionic-team/ionic-framework/pull/28640#discussion_r1422983693)

core/src/components/refresher/test/a11y/refresher.e2e.ts
thetaPC: I was following the [structure](https://github.com/ionic-team/ionic-framework/blob/888a30fba5790c2b21b669aa6a58937914d5336e/core/src/components/refresher/test/test.utils.ts#L17) from `pullToRefresh()` but it still works without it. I'll remove it.

core/src/components/refresher/test/a11y/refresher.e2e.ts
thetaPC: I wanted to verify that the arrow container was visible when dragging since gestures tend to be flaky. Should I remove it?

liamdebeasi: Rather than check something that is dependent on the internal markup of the component, it might be better to check the state of the component: https://github.com/ionic-team/ionic-framework/blob/fc88613fefa019a3b695a2c6e10c85cd3ce79ae8/core/src/components/refresher/refresher.tsx#L814

For example, checking the `refresher-refreshing` class verifies that a pull to refresh gesture was completed and the component is in the process of refreshing.

thetaPC: I'll have to use `refreshing-pulling` since I'm testing the contrast while it's being pulled.

core/src/components/label/test/a11y/label.e2e.ts
thetaPC: The modern syntax doesn't use `<ion-label>` so I opted to test when the label is [stacked/floating](https://github.com/ionic-team/ionic-framework/blob/c5a67cdabe7c0b03042161f2d467db877614e748/core/src/components/input/input.md.scss#L71-L74). Since the goal was to test the label, I had the test in this file instead of input.

liamdebeasi: The goal of this task is to test `ion-label`, not labels in general. We already verify that the label in `ion-input` has sufficient contrast in https://github.com/ionic-team/ionic-framework/blob/7c1f8cc8afb43dea2f4cc109c88d06e26fc60276/core/src/components/input/test/a11y/input.e2e.ts#L37-L43

thetaPC: Understood, change made in https://github.com/ionic-team/ionic-framework/pull/28640/commits/ae56a3c0506f976baaae9266f4d36c96bcf4a8b6

core/src/components/refresher/test/a11y/refresher.e2e.ts
liamdebeasi: I don't think we need to make any changes in this PR, but we should make a tech debt ticket to evaluate doing `<ion-content role="main">` from `content.tsx` instead of nesting the `<main>` element. We can then add a `TODO` comment above this disabled rule.

liamdebeasi (approved):
Good to go once the tech debt ticket is made and a TODO is added to the code in question: https://github.com/ionic-team/ionic-framework/pull/28640#discussion_r1424055077

liamdebeasi (approved):
thanks for making the ticket!

PR #28641, opened 2023-12-05
chore(playwright): throw error when using themes config with page.goto

sean-perkins:
Issue number: N/A

---------

Team was getting confused why the themes configuration did not have any effect when using a separate template and testing with `page.goto`. 

Currently the `themes` configuration is only supported when using `page.setContent`. 

## What is the new behavior?

- Throws an error when using a `themes` configuration with `page.goto`. 

Note: `light` theme is the "fallback" or "no theme" state. This is why we explicitly check if the theme is not that value, which means the test is configured to test against a specific theme that would not be set. 

- [ ] Yes
- [x] No

### DIFF
--- a/core/src/utils/test/playwright/playwright-declarations.ts
+++ b/core/src/utils/test/playwright/playwright-declarations.ts
@@ -6,7 +6,7 @@ import type { LocatorOptions, E2ELocator } from './page/utils/locator';
 
 export interface E2EPageOptions extends PageOptions, TestConfig {
   /**
-   * Indicates if the page content is set inline with `.setContent`. This setting can 
+   * Indicates if the page content is set inline with `.setContent`. This setting can
    * only be true or not set.
    */
   inline?: true;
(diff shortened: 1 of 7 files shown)


sean-perkins:
Ahh shoot, the `goto` function is used internally by the `.setContent` flow. I'll need to rework this a bit after my meeting.

core/src/utils/test/playwright/playwright-declarations.ts
```diff
@@ -4,7 +4,13 @@ import type { TestConfig } from './generator';
 import type { EventSpy } from './page/event-spy';
 import type { LocatorOptions, E2ELocator } from './page/utils/locator';
 
-export interface E2EPageOptions extends PageOptions, TestConfig {}
+export interface E2EPageOptions extends PageOptions, TestConfig {
+  /**
+   * Indicates if the page content is set inline with `.setContent`. This setting can
+   * only be true or not set.
+   */
+  inline?: true;
```
sean-perkins, line 12: The use of `true` here is intentional. This configuration should only be enabled or be not defined.

amandaejohnston (approved):
Thank you!

PR #28653, opened 2023-12-06
fix(picker): disabled column works with options

liamdebeasi:
Issue number: N/A

---------

As noted in https://github.com/ionic-team/ionic-framework/pull/28622, disabled columns w/ options was not migrated to minimize the amount of code changes that needed to be reviewed at once. This would be updated in a separate PR.

## What is the new behavior?

- This PR ensures that disabling a column makes the column not scrollable and the options non-interactive

The previous approach of coupling the column disabled prop with the option disabled prop was modified as it resulted in a brittle architecture. We've opted for a solution that achieves the same goal while reducing the tight coupling. We found that the tight coupling would cause a customized disabled property value on the option to be overridden with the value of the parent column's disabled value.

- [ ] Yes
- [x] No


mapsandapps (approved):
works & makes sense 👍🏻

amandaejohnston (approved):

PR #28654, opened 2023-12-06
feat(picker): add css variables for picker highlight and fade

brandyscarney:
Issue number: N/A

---------

Picker has the following CSS variables for styling: `--wheel-fade-background-rgb` and `--wheel-highlight-background`

These variables make sense when used from `ion-datetime`, but picker does not need the `wheel` prefix.

### Datetime

- Adds `--wheel-highlight-border-radius` 
- Keeps `--wheel-fade-background-rgb` and `--wheel-highlight-background`
- Assigns these variables to picker's new variables

```scss
ion-picker {
  --highlight-background: var(--wheel-highlight-background);
  --highlight-border-radius: var(--wheel-highlight-border-radius);
  --fade-background-rgb: var(--wheel-fade-background-rgb);
}
```

### Picker 

Renames `--wheel-fade-background-rgb` and `--wheel-highlight-background` and adds another CSS variables to `ion-picker`:

- Renames `--wheel-fade-background-rgb` to `--fade-background-rgb` 
- Renames `--wheel-highlight-background` to `--highlight-background`
- Adds `--highlight-border-radius`

- [ ] Yes
- [x] No

## Other information

PR to update the [Datetime docs](https://github.com/ionic-team/ionic-docs/pull/3303)


core/src/components/datetime/test/custom/index.html
brandyscarney, line 1: This test is probably not required in datetime anymore unless we just want to make sure it is still customizable from datetime.

core/api.txt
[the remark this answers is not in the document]
liamdebeasi: Agreed on not using the "wheel" term in ion-picker. However, would it beneficial to keep the variable in `ion-datetime`? When setting `--highlight-background` on `ion-datetime` it isn't obvious that the highlight is for the wheel.

liamdebeasi (commented):
Had some questions on the naming/how this is used in datetime

core/api.txt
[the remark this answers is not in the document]
amandaejohnston: It might be easiest to add the `wheel-` prefixed variables to datetime specifically, and then have datetime's code just pass the values along to the picker's non-prefixed ones. That way the variables are documented on both the datetime and picker component, without people having to know that datetime uses a picker internally.

core/src/components/datetime/test/custom/index.html
amandaejohnston: If we go with my suggestion in https://github.com/ionic-team/ionic-framework/pull/28654/files#r1418090200, we'll want to keep the tests as-is but use the prefixed variables for the wheels instead. Either way, we'll want to keep the two "Grid Style" datetimes here since they're testing other things.

brandyscarney: Updated in https://github.com/ionic-team/ionic-framework/pull/28654/commits/221ff2a29c6a2d66e29b3644f3fa6ab880db776d

amandaejohnston (approved):

core/api.txt
```diff
@@ -428,6 +428,7 @@ ion-datetime,css-prop,--background-rgb
 ion-datetime,css-prop,--title-color
 ion-datetime,css-prop,--wheel-fade-background-rgb
 ion-datetime,css-prop,--wheel-highlight-background
+ion-datetime,css-prop,--wheel-highlight-border-radius
```
liamdebeasi, line 431: The border radius is not showing up on iOS anymore: 
image

brandyscarney: Fixed https://github.com/ionic-team/ionic-framework/pull/28654/commits/294c3043a3ce7bbd7e1e91920d5bdd35a5ff0352

liamdebeasi (approved):
Great job!

PR #28656, opened 2023-12-06
tests(many): add tests for color contrast for button components

mapsandapps:
Issue number: Internal

---------

Uses of `ion-color()` in Button, Back Button, FAB, Menu Button, Item Option, and Segment do not have tests for color contrast.

## What is the new behavior?

- Uses of `ion-color()` in Button, Back Button, FAB, Menu Button, Item Option, and Segment have tests for color contrast for both light and dark themes.

- [ ] Yes
- [x] No


core/src/components/back-button/test/a11y/back-button.e2e.ts
```diff
@@ -1,6 +1,32 @@
+import AxeBuilder from '@axe-core/playwright';
 import { expect } from '@playwright/test';
 import { configs, test } from '@utils/test/playwright';
 
+/**
+ * Only ios mode uses ion-color() for the back button
+ */
+configs({ directions: ['ltr'], modes: ['ios'], themes: ['light', 'dark'] }).forEach(({ title, config }) => {
+  test.describe(title('back-button: contrast'), () => {
```
thetaPC, line 9:
```suggestion
  test.describe(title('back-button: a11y'), () => {
```

Let's keep it consistent with the other tests especially since since the test is checking for all violations. It's not specific to contrast. The same goes for the rest.

core/src/components/fab-button/test/a11y/fab-button.e2e.ts
```diff
@@ -1,6 +1,72 @@
...
+      const results = await new AxeBuilder({ page }).analyze();
+      expect(results.violations).toEqual([]);
+    });
+  });
+
+  test.describe(title('fab: translucent contrast'), () => {
+    test('should not have accessibility violations', async ({ page }) => {
+      await page.setContent(
+        `
+        <ion-fab-button class="ion-focused">FAB</ion-fab-button>
```
thetaPC, line 48: Since the test focuses on translucent then can this be removed?

thetaPC (changes requested):

core/src/components/back-button/test/a11y/back-button.e2e.ts
mapsandapps: I guess I could. My reasoning for changing it (although perhaps I should have changed it to something else) is that we aren't really testing the accessibility of these components very much when we aren't testing very many versions of them. For example, this test only runs in ios mode. I didn't want to give the impression that this is _the_ a11y test for this component like some other components have.

core/src/components/fab-button/test/a11y/fab-button.e2e.ts
mapsandapps: Yes, thank you! Fixed in [9684765](https://github.com/ionic-team/ionic-framework/pull/28656/commits/96847650da993f6c4c73bcf00f33e069b2696495).

core/src/components/back-button/test/a11y/back-button.e2e.ts
mapsandapps: works for me! fixed in [8b78e5a](https://github.com/ionic-team/ionic-framework/pull/28656/commits/8b78e5a6bafa1ed41907612e6c8071956d291cba).

core/src/components/back-button/test/a11y/back-button.e2e.ts
```diff
@@ -1,6 +1,32 @@
+import AxeBuilder from '@axe-core/playwright';
 import { expect } from '@playwright/test';
 import { configs, test } from '@utils/test/playwright';
 
+/**
+ * Only ios mode uses ion-color() for the back button
+ */
```
thetaPC, line 7: Should we also remove these since the latest change provides more info on the scenario?

mapsandapps: We can. My reasoning was that we often put a comment explaining why we're only testing in one mode or with one direction, even if it's the default reason (behavior doesn't differ between modes/directions). So I figured it made sense to have a comment here, especially since it's a different reason than normal. But I'd be fine with removing it.

thetaPC: Got it, I'm okay with either or.

thetaPC (approved):
LGTM

core/src/components/button/test/a11y/button.e2e.ts
```diff
@@ -1,6 +1,74 @@
...
+
+      const results = await new AxeBuilder({ page }).analyze();
+      expect(results.violations).toEqual([]);
+    });
+
+    test('button in toolbar should not have accessibility violations', async ({ page }) => {
+      await page.setContent(
+        `
+        <ion-toolbar>
+          <ion-button fill="outline" class="ion-activated"> Start </ion-button>
```
liamdebeasi, line 40:
```suggestion
          <ion-button fill="outline" class="ion-activated">Start</ion-button>
```

Nit: Just to keep it consistent with the other tests 😄

core/src/components/fab-button/test/a11y/fab-button.e2e.ts
```diff
@@ -1,6 +1,71 @@
...
+        `
+        <ion-fab-button translucent="true">FAB</ion-fab-button>
+        <ion-fab-button class="ion-focused" translucent="true"> FAB </ion-fab-button>
+        <ion-fab activated="true">
+          <ion-fab-button>
+            Open
+          </ion-fab-button>
+          <ion-fab-list side="bottom">
+            <ion-fab-button translucent="true"> Down </ion-fab-button>
+            <ion-fab-button class="ion-focused" translucent="true"> Down </ion-fab-button>
```
liamdebeasi, line 56:
```suggestion
        <ion-fab-button class="ion-focused" translucent="true">FAB</ion-fab-button>
        <ion-fab activated="true">
          <ion-fab-button>
            Open
          </ion-fab-button>
          <ion-fab-list side="bottom">
            <ion-fab-button translucent="true">Down</ion-fab-button>
            <ion-fab-button class="ion-focused" translucent="true">Down</ion-fab-button>
```

liamdebeasi (approved):
Good to go on my end once comments are addressed.

PR #28657, opened 2023-12-06
chore: clean up TODOs

liamdebeasi:
Issue number: N/A

---------

There were a number of TODOs to clean up the remaining code for this feature.

## What is the new behavior?

- Many of the changes are autogenerated diffs. I added comments to diffs that I felt were not obvious for why they changed.

- [ ] Yes
- [x] No

### DIFF
--- a/core/src/components/datetime/utils/data.ts
+++ b/core/src/components/datetime/utils/data.ts
@@ -1,5 +1,4 @@
 import type { Mode } from '../../../interface';
-import type { PickerColumnItem } from '../../picker-column/picker-column-interfaces';
 import type { DatetimeParts, DatetimeHourCycle } from '../datetime-interface';
 
 import { isAfter, isBefore, isSameDay } from './comparison';
@@ -13,6 +12,12 @@ import {
 } from './format';
 import { getNumDaysInMonth, is24Hour, getHourCycle } from './helpers';
 import { getNextMonth, getPreviousMonth, getInternalHourValue } from './manipulation';
+
+export interface WheelColumnOption {
+  text: string;
+  value: string | number;
+  disabled?: boolean;
+}
 
 /**
  * Returns the current date as
@@ -313,7 +318,7 @@ export const getMonthColumnData = (
   formatOptions: Intl.DateTimeFormatOptions = {
     month: 'long',
   }
-): PickerColumnItem[] => {
+): WheelColumnOption[] => {
   const { year } = refParts;
   const months = [];
 
@@ -391,7 +396,7 @@ export const getDayColumnData = (
   formatOptions: Intl.DateTimeFormatOptions = {
     day: 'numeric',
   }
-): PickerColumnItem[] => {
+): WheelColumnOption[] => {
   const { month, year } = refParts;
   const days = [];
 
@@ -438,7 +443,7 @@ export const getYearColumnData = (
   minParts?: DatetimeParts,
   maxParts?: DatetimeParts,
   yearValues?: number[]
-): PickerColumnItem[] => {
+): WheelColumnOption[] => {
   let processedYears = [];
   if (yearValues !== undefined) {
     processedYears = yearValues;
@@ -466,7 +471,7 @@ export const getYearColumnData = (
 
 interface CombinedDateColumnData {
   parts: DatetimeParts[];
-  items: PickerColumnItem[];
+  items: WheelColumnOption[];
 }
 
 /**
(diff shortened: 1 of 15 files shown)


core/src/components/picker-column/picker-column.tsx
```diff
@@ -46,23 +39,11 @@ export class PickerColumn implements ComponentInterface {
    */
   @Prop() disabled = false;
 
-  /**
-   * A list of options to be displayed in the picker
-   */
-  @Prop() items: PickerColumnItem[] = [];
```
liamdebeasi, line 52: Removed in favor of `ion-picker-column-option`

core/src/components/picker-column/picker-column.tsx
```diff
@@ -46,23 +39,11 @@ export class PickerColumn implements ComponentInterface {
...
    * The selected option in the picker.
    */
   @Prop({ mutable: true }) value?: string | number;
 
-  /**
-   * The color to use from your application's color palette.
-   * Default options are: `"primary"`, `"secondary"`, `"tertiary"`, `"success"`, `"warning"`, `"danger"`, `"light"`, `"medium"`, and `"dark"`.
-   * For more information on colors, see [theming](/docs/theming/basics).
-   */
-  @Prop({ reflect: true }) color?: Color = 'primary';
```
liamdebeasi, line 64: We could have picker-column-option inherit the color from picker-column, but that behavior was not scoped in the design document. As a result, I removed this prop.

core/src/components/picker-column/picker-column.tsx
```diff
@@ -473,26 +452,17 @@ export class PickerColumn implements ComponentInterface {
...
-     * exportparts is needed so ion-datetime can expose the parts
-     * from two layers of shadow nesting. If this causes problems,
-     * the attribute can be moved to datetime.tsx and set on every
-     * instance of ion-picker-column there instead.
-     * TODO FW-5580 remove exportparts
-     */
-
     return (
       <Host
-        exportparts={`${PICKER_ITEM_PART}, ${PICKER_ITEM_ACTIVE_PART}`}
```
liamdebeasi, line 489: The column options exist in the light DOM now by default. For pickers in datetime, datetime will continue to expose shadow parts.

core/src/components/picker-column/test/disabled/picker-column.e2e.ts
```diff
@@ -105,9 +68,8 @@ configs({ modes: ['ios'], directions: ['ltr'] }).forEach(({ title, config }) =>
 /**
  * This behavior does not vary across directions.
  */
-// TODO FW-5580 fix this
 configs({ directions: ['ltr'] }).forEach(({ title, screenshot, config }) => {
-  test.describe.skip(title('picker-column: disabled column rendering'), () => {
+  test.describe(title('picker-column: disabled column rendering'), () => {
```
liamdebeasi, line 72: I forgot to unskip these tests when working on previous PRs. However, I ran them all locally and verified that they passed during development.

core/src/components/datetime/utils/data.ts
```diff
@@ -14,6 +13,12 @@ import {
 import { getNumDaysInMonth, is24Hour, getHourCycle } from './helpers';
 import { getNextMonth, getPreviousMonth, getInternalHourValue } from './manipulation';
 
+export interface WheelColumnOption {
```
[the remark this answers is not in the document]
liamdebeasi: Also decided to call it "Wheel" since that's the terminology we use with datetime. No functionality has changed though.

core/src/components/picker-column/picker-column.scss
```diff
@@ -112,66 +112,10 @@
   justify-content: start;
 }
 
-// TODO FW-5580 remove this
-:host .picker-item {
```
liamdebeasi, line 116: All `.picker-item` selectors should go away since we are using `ion-picker-column-option` instead.

core/src/components/picker-column/picker-column.scss
```diff
@@ -64,6 +64,41 @@
   scrollbar-width: none;
 }
 
+.picker-item-empty {
```
liamdebeasi, line 67: I originally thought the old `.picker-item` CSS wasn't needed. Turns out it's still needed but just for the spacer elements. I updated the selector to make that more clear.

amandaejohnston (approved):

PR #28666, opened 2023-12-07
feat(picker): integrate picker-column-option with picker

liamdebeasi:
Issue number: N/a

---------

The team would like to add a way of using the picker inline with page content.

## What is the new behavior?

- This PR integrates the column option component with the rest of the picker so that is functional. This PR is a combination of commits from previously reviewed PRs.

- [ ] Yes
- [x] No


thetaPC (approved):
LGTM, didn't encounter any issues. Great job!

PR #28671, opened 2023-12-08
fix(react): replacing route uses new route direction and animation

### LINKED ISSUE #24260: bug: Ionic React navigate with push from useIonRouter ignores direction when used as "replace"
### Prerequisites

### Ionic Framework Version

- [ ] v4.x
- [X] v5.x
- [ ] v6.x

### Current Behavior

using Ionic React and `useIonRouter` the direction is not respected if the route action is "replace". It always goes "forward"
If the action is "push" (the default) all directions respected.

### Expected Behavior

Direction of the router should behave based on function call - no matter the route action.

### Steps to Reproduce

In the repo, fire up the app and you should go to the EXAMPLE menu item. Here there are three links to detail pages. 
Detail 1 has a button to redirect using "replace" with direction "none" - this will redirect but animate using "forward"
Detail 2 has a button to redirect using default behavior ("push") with direction "none" - this will redirect and have no page transition animation.

### Code Reproduction URL

https://github.com/babycourageous/ionic-use-ion-router-direction-bug

### Ionic Info

Ionic:

   Ionic CLI       : 6.17.1 (/Users/renedellefont/.fnm/node-versions/v14.17.6/installation/lib/node_modules/@ionic/cli)
   Ionic Framework : @ionic/react 5.8.4

Capacitor:

   Capacitor CLI      : 3.2.5
   @capacitor/core    : 3.2.5

Utility:

   native-run  : 1.5.0

System:

   NodeJS : v14.17.6 (/Users/renedellefont/.fnm/node-versions/v14.17.6/installation/bin/node)
   npm    : 6.14.15
   OS     : macOS Big Sur

### Additional Information

Just let me know if there's any more info you need!

aeharding:
Issue number: resolves #24260

---------

When replacing a route (`router.push(newRoute, 'none', 'replace')`) the `RouterDirection` from the route being replaced is being used (if it exists) instead of the new one the user specifies.

## What is the new behavior?

User-specified `RouteDirection` is used, if it exists. If it doesn't it falls back to the `RouteDirection` of the route being replaced.

- [ ] Yes
- [X] No

## Other information

Please see the following comment for why I think the current behavior is incorrect, and why this change is needed: https://github.com/ionic-team/ionic-framework/issues/24260#issuecomment-1078960780

### DIFF
--- a/packages/react-router/src/ReactRouter/IonRouter.tsx
+++ b/packages/react-router/src/ReactRouter/IonRouter.tsx
@@ -187,7 +187,7 @@ class IonRouterInner extends React.PureComponent<IonRouteProps, IonRouteState> {
           routeInfo.lastPathname = currentRouteInfo?.pathname || routeInfo.lastPathname;
           routeInfo.prevRouteLastPathname = currentRouteInfo?.lastPathname;
           routeInfo.pushedByRoute = pushedByRoute;
-          routeInfo.routeDirection = currentRouteInfo?.routeDirection || routeInfo.routeDirection;
+          routeInfo.routeDirection = routeInfo.routeDirection || currentRouteInfo?.routeDirection;
           routeInfo.routeAnimation = currentRouteInfo?.routeAnimation || routeInfo.routeAnimation;
         }
 
--- a/packages/react-router/src/ReactRouter/IonRouter.tsx
+++ b/packages/react-router/src/ReactRouter/IonRouter.tsx
@@ -188,7 +188,7 @@ class IonRouterInner extends React.PureComponent<IonRouteProps, IonRouteState> {
           routeInfo.prevRouteLastPathname = currentRouteInfo?.lastPathname;
           routeInfo.pushedByRoute = pushedByRoute;
           routeInfo.routeDirection = routeInfo.routeDirection || currentRouteInfo?.routeDirection;
-          routeInfo.routeAnimation = currentRouteInfo?.routeAnimation || routeInfo.routeAnimation;
+          routeInfo.routeAnimation = routeInfo.routeAnimation || currentRouteInfo?.routeAnimation;
         }
 
         this.locationHistory.add(routeInfo);
--- a/packages/react-router/src/ReactRouter/IonRouter.tsx
+++ b/packages/react-router/src/ReactRouter/IonRouter.tsx
@@ -187,6 +187,7 @@ class IonRouterInner extends React.PureComponent<IonRouteProps, IonRouteState> {
           routeInfo.lastPathname = currentRouteInfo?.pathname || routeInfo.lastPathname;
           routeInfo.prevRouteLastPathname = currentRouteInfo?.lastPathname;
           routeInfo.pushedByRoute = pushedByRoute;
+
           /**
            * When replacing routes we should still prefer
            * any custom direction/animation that the developer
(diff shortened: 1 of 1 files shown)


packages/react-router/src/ReactRouter/IonRouter.tsx
```diff
@@ -187,7 +187,7 @@ class IonRouterInner extends React.PureComponent<IonRouteProps, IonRouteState> {
           routeInfo.lastPathname = currentRouteInfo?.pathname || routeInfo.lastPathname;
           routeInfo.prevRouteLastPathname = currentRouteInfo?.lastPathname;
           routeInfo.pushedByRoute = pushedByRoute;
-          routeInfo.routeDirection = currentRouteInfo?.routeDirection || routeInfo.routeDirection;
+          routeInfo.routeDirection = routeInfo.routeDirection || currentRouteInfo?.routeDirection;
           routeInfo.routeAnimation = currentRouteInfo?.routeAnimation || routeInfo.routeAnimation;
```
aeharding, line 191: Note: May want to do the same thing for routeAnimation, but I was trying to be surgical because I haven't tested a change to `routeAnimation` and I am not an Ionic routing expert :)

liamdebeasi:
Thanks! Some more context for why this bug exists. When we are replacing, we need to copy over some of the previous state. For example, consider a history of `A --> B`. When replacing to `C`, we are replacing the state of `B` with the state of `C`. As result, the previous route for `C` is `A` (since `B` no longer exists).

Due to this requirement, we correctly copy information from `currentRouteInfo` first before falling back to `routeInfo`: https://github.com/ionic-team/ionic-framework/blob/37290df7ef6cf8eaa625dae8434afb3751e13a6c/packages/react-router/src/ReactRouter/IonRouter.tsx#L174-L189

However, we should not do that for `routeDirection` and `routeAnimation`. The reason is that developers may have specified a custom animation or direction to use with this replace operation. We should prefer that first, and then fall back to the data in `currentRouteInfo` if the developer does not pass any information.

liamdebeasi (changes requested):

packages/react-router/src/ReactRouter/IonRouter.tsx
```diff
@@ -187,8 +187,8 @@ class IonRouterInner extends React.PureComponent<IonRouteProps, IonRouteState> {
           routeInfo.lastPathname = currentRouteInfo?.pathname || routeInfo.lastPathname;
           routeInfo.prevRouteLastPathname = currentRouteInfo?.lastPathname;
           routeInfo.pushedByRoute = pushedByRoute;
-          routeInfo.routeDirection = currentRouteInfo?.routeDirection || routeInfo.routeDirection;
-          routeInfo.routeAnimation = currentRouteInfo?.routeAnimation || routeInfo.routeAnimation;
+          routeInfo.routeDirection = routeInfo.routeDirection || currentRouteInfo?.routeDirection;
+          routeInfo.routeAnimation = routeInfo.routeAnimation || currentRouteInfo?.routeAnimation;
```
liamdebeasi, line 191:
```suggestion
          /**
           * When replacing routes we should still prefer
           * any custom direction/animation that the developer
           * has specified when navigating first instead of relying
           * on previously used directions/animations.
           */
          routeInfo.routeDirection = routeInfo.routeDirection || currentRouteInfo?.routeDirection;
          routeInfo.routeAnimation = routeInfo.routeAnimation || currentRouteInfo?.routeAnimation;
```

Sorry, forgot to add a comment 😄

liamdebeasi (approved):
Good to go once the final change is made. I'm making a dev build now and will post this in the linked thread.

thetaPC (approved):
LGTM, works as intended.

sean-perkins (approved):
Thank you for fixing this!

PR #28676, opened 2023-12-08
fix(input, textarea, select): reduce padding on slotted buttons

amandaejohnston:
Issue number: N/A

---------

Buttons slotted in the start/end slots of input, textarea, or select have a lot of excess padding that can cause them to look misaligned from other pieces such as the control's label, especially when using `fill="clear"`.

## What is the new behavior?

Padding removed, and a border radius added so that non-clear buttons as well as the focus state on clear buttons still look okay. Existing screenshot tests have also been updated to include buttons (non-clear ones so you can see the border radius in the screenshots).

Let me know if you think this should go on a feature branch instead. I chose `main` because the current experience looks somewhat broken and we just did a minor release recently, but this could be considered a notable enough behavior change.

- [ ] Yes
- [x] No


core/src/components/input/input.scss
```diff
@@ -659,8 +659,21 @@
...
+::slotted(ion-button[slot="start"].button-has-icon-only),
+::slotted(ion-button[slot="end"].button-has-icon-only) {
+  --border-radius: 50%;
+  --padding-start: 0;
+  --padding-end: 0;
+  --padding-top: 0;
+  --padding-bottom: 0;
+  
+  aspect-ratio: 1;
+}
```
thetaPC, line 679:
```suggestion
}

```
Missing new line

core/src/components/input/input.scss
```diff
@@ -659,8 +659,21 @@
...
+
+::slotted(ion-button[slot="start"].button-has-icon-only),
+::slotted(ion-button[slot="end"].button-has-icon-only) {
+  --border-radius: 50%;
+  --padding-start: 0;
+  --padding-end: 0;
+  --padding-top: 0;
+  --padding-bottom: 0;
+  
+  aspect-ratio: 1;
```
thetaPC, line 678: Let's add a comment of why this is needed. Since it's connected to the hover and focus, it might get accidentally removed because it's not evident on first glance.
The same for the others.

core/src/components/select/select.scss
```diff
@@ -592,8 +592,21 @@ button {
...
+::slotted(ion-button[slot="start"].button-has-icon-only),
+::slotted(ion-button[slot="end"].button-has-icon-only) {
+  --border-radius: 50%;
+  --padding-start: 0;
+  --padding-end: 0;
+  --padding-top: 0;
+  --padding-bottom: 0;
+
+  aspect-ratio: 1;
+}
```
thetaPC, line 612:
```suggestion
}

```

core/src/components/textarea/textarea.scss
```diff
@@ -753,8 +753,21 @@
...
+::slotted(ion-button[slot="start"].button-has-icon-only),
+::slotted(ion-button[slot="end"].button-has-icon-only) {
+  --border-radius: 50%;
+  --padding-start: 0;
+  --padding-end: 0;
+  --padding-top: 0;
+  --padding-bottom: 0;
+
+  aspect-ratio: 1;
+}
```
thetaPC, line 773:
```suggestion
}

```

thetaPC (changes requested):

core/src/components/input/input.scss
amandaejohnston: Added in https://github.com/ionic-team/ionic-framework/pull/28676/commits/582d4827e306d8a97bab5803cdcfc94a5bf98401 👍

thetaPC (approved):
LGTM

core/src/components/input/input.scss
```diff
@@ -659,8 +659,29 @@
...
+
+/**
+ * Slotted buttons have a lot of default padding that can
+ * cause them to look misaligned from other pieces such
+ * as the control's label, especially when using a clear
+ * fill. We also make them circular to ensure that non-
+ * clear buttons and the focus/hover state on clear ones
+ * don't look too crowded.
+ */
+::slotted(ion-button[slot="start"].button-has-icon-only),
```
brandyscarney, line 678: Threading for discussion

Overall the changes look good, but I think we should adjust some of the values. For example, in `md` we could use the following:

```css
::slotted(ion-button[slot="start"].button-has-icon-only),
::slotted(ion-button[slot="end"].button-has-icon-only) {
  --border-radius: 50%;
  --padding-start: 8px;
  --padding-end: 8px;
  --padding-top: 8px;
  --padding-bottom: 8px;

  width: 40px;
  height: 40px;
}
```

This would match [MUI](https://mui.com/material-ui/react-text-field/#icons) and [Material Web](https://material-web.dev/components/text-field/#icons).

| Before | After |
| ---| ---|
| before | after | 

I think iOS buttons look too large but it would take more investigating to find out the right size.

I am okay with holding off on these updates until MD3 & iOS 17 but I do think we should add them in the design docs if that's the decision.

brandyscarney (approved):

core/src/components/input/input.scss
amandaejohnston: I'll go ahead and make the change for MD. For iOS, looking at the [iOS 17 Figma](https://www.figma.com/community/file/1248375255495415511), I don't see any specs for using a button like this. I believe it's an MD-exclusive feature, at least as far as the specs go, so we have some design freedom. If you want, I can reduce the size of the buttons to match MD?

amandaejohnston: I've reduced the iOS button size (using em instead of px to match the default `ios` mode button styling), although it looks a little crowded with a larger icon like the lock. I'm a little hesitant to go reducing the font size of the icon since the new value would be pretty arbitrary. Do you have any thoughts? I can always revert the iOS button size for now and create a spike to investigate later.

core/src/components/input/input.md.scss
```diff
@@ -117,3 +117,27 @@
...
+ * as the control's label, especially when using a clear
+ * fill. We also make them circular to ensure that non-
+ * clear buttons and the focus/hover state on clear ones
+ * don't look too crowded.
+ */
+::slotted(ion-button[slot="start"].button-has-icon-only),
+::slotted(ion-button[slot="end"].button-has-icon-only) {
+  --border-radius: 50%;
+  --padding-start: 8px;
+  --padding-end: 8px;
```
brandyscarney, line 136: Does this change need to be made to `select.scss` and `textarea.scss` too?

core/src/components/input/input.scss
amandaejohnston: We agreed in Slack to revert the iOS size change and make a spike for later, since we're running up on the end of the sprint.

core/src/components/input/input.md.scss
amandaejohnston: Oh, yes! Haha, thank you for the reminder 😆

brandyscarney (approved):

PR #28682, opened 2023-12-11
fix(button): wrap text by default

### LINKED ISSUE #8700: bug: long text with max width not truncating in button

mapsandapps:
Issue number: resolves #8700

---------

Button text is truncated by default.

## What is the new behavior?

- Button text wraps by default.

- [x] Yes
- [ ] No

Starting in [Ionic Framework 7.2](https://github.com/ionic-team/ionic-framework/releases/tag/v7.2.0), we encouraged developers to [opt-in](https://github.com/ionic-team/ionic-framework/pull/27547) to text wrapping in buttons by setting `ion-text-wrap`.

This PR is targeting Ionic Framework v8, where text wrapping of buttons will be the default behavior.

This may cause visual changes across apps using Ionic Framework. Buttons with long text may now be taller than before, leading to layout shifts.

If developers wish to prevent text wrapping, they may use the [.ion-text-nowrap](https://ionicframework.com/docs/layout/css-utilities#text-alignment) class.


core/src/components/breadcrumbs/test/basic/breadcrumbs.e2e.ts-snapshots/breadcrumb-diff-ios-ltr-Mobile-Chrome-linux.png
mapsandapps, line 1: We could add the nowrap class to these buttons so they don't wrap, if that's preferred over updating the snapshots.

BREAKING.md
```diff
@@ -45,6 +46,10 @@ This section details the desktop browser, JavaScript framework, and mobile platf
 
 <h2 id="version-8x-components">Components</h2>
 
+<h4 id="version-8x-button">Button</h4>
+
+- Button text wraps by default.
```
brandyscarney, line 51: Can we add a note here with instructions to "undo" it:

```suggestion
- Button text now wraps by default. If this behavior is not desired, add the `ion-text-nowrap` class from the [CSS Utilities](https://ionicframework.com/docs/layout/css-utilities).
```

core/src/components/breadcrumbs/test/basic/breadcrumbs.e2e.ts-snapshots/breadcrumb-diff-ios-ltr-Mobile-Chrome-linux.png
brandyscarney: I am okay with keeping this. I think we should avoid custom css where possible.

brandyscarney (approved):
This looks great!

Issue #28680, opened 2023-12-11
feat: Improved types for react useIonModal


aeharding:
### Prerequisites

- [X] I have read the [Contributing Guidelines](https://github.com/ionic-team/ionic-framework/blob/main/.github/CONTRIBUTING.md#creating-an-issue).
- [X] I agree to follow the [Code of Conduct](https://ionicframework.com/code-of-conduct).
- [X] I have searched for [existing issues](https://github.com/ionic-team/ionic-framework/issues) that already include this feature request, without success.

### Describe the Feature Request

useIonModal should type props parameter.

### Describe the Use Case

useIonModal currently allows `any` for component props. This is dangerous and has caused many crashes and unintended behaviors in my experience.

Currently:

```
export declare function useIonModal(component: ReactComponentOrElement, componentProps?: any): UseIonModalResult;
```

### Describe Preferred Solution

```
export declare function useIonModal<P>(component: React.ComponentClass<P> | React.FC<P>, componentProps?: P): UseIonModalResult;
export declare function useIonModal(component: React.ReactElement, componentProps?: any): UseIonModalResult;
```

### Describe Alternatives

Leaving as-is

### Additional Information

If you would accept this change, I would happily make a PR

aeharding:
FYI I see #27548, this seemed a bit more open-ended so I decided to make a separate issue in smaller scope (fixing the types).

liamdebeasi:
Thanks! I think this would be a good change to make. We'll want to do make this change for both `useIonModal` and `useIonPopover`, and `T` should continue to be `any` if not set to avoid breaking changes.

aeharding:
@liamdebeasi, in my testing Typescript automatically infers T based on the component you pass it. By making the proposed change I was able to see type errors with incorrect params in my usage without adding any explicit generic declarations to my code :)

mikelhamer:
Is this still relevant?

aeharding:
Yes, and my PR was close to being merged, but @liamdebeasi left Ionic so no progress since

mikelhamer:
Ah, thank you. I'm new here and digging around for ways to help. So your PR partially completed the feature, but there is some left to do?

PR #28685, opened 2023-12-11
feat(react): add types to useIonModal, useIonPopover

aeharding:
Issue number: #28680

---------

Params (second parameter) is of type `any`

## What is the new behavior?

Params is now typed according to the modal/popover component passed in the first parameter.

- [ ] Yes
- [X] No Note: It's technically not breaking, but many projects will likely see type errors due to incorrect types

## Other information

None


liamdebeasi:
Spoke with Alexander on Discord. There's a problem with this implementation, and he plans to open a new PR when he's back in the office.

PR #28689, opened 2023-12-11
feat(picker): add inline picker

### LINKED ISSUE #15710: Picker: Add cascading columns support to IonPicker component.
Current IonPicker component, just as acting in IonDatetime component, can support multi columns, however, each column is standalone, which has no relationship with others.

In fact, cascading relationship between columns sometimes is very common. 
For example, when we choose country | city | address, such choosing action is suitable for Picker UI. 

We can find some good solutions from other JS package to support cascading issues, but we hope it can be supported in ionic's own component.

https://github.com/raychenfj/ion-multi-picker 
##which now does not support V4
image

https://didi.github.io/cube-ui/example/#/cascade-picker

To realize it, it is very sample. Just emit a change event when user scroll one column end, and then offer an API method, to allow user update some column data. Sure, if all of these can be configured, it will be better.

Thanks for your great jobs.

### LINKED ISSUE #24905: feat: add support for inline ion-picker
### Prerequisites

Hello, I would like a feature to enable the inclusion of the ion-picker component within the content of a page without the need to open it in a modal, I managed to implement this feature with ion-datetime but it only works with dates, years and months and I want to use ion-picker to create an ios-style select and only that with ion-picker I can create but it only opens in modal, is this feature currently possible or could it be developed?

ion-picker

I am currently using the following code in angular/ionic 6

**home.page.html** 
```

<ion-content>
  <ng-container >

    <ion-button expand="block" (click)="showPickerController()">Show Picker Controller</ion-button>

    <ion-list>     
      <ng-container *ngFor="let type of ['year']">   
            <ion-item slot="header">
              <ion-label>ion-datetime</ion-label>
              <ion-text slot="end"></ion-text>
            </ion-item>
            <ion-datetime
              size="cover"
              slot="content"
              [presentation]="type"
            >
            </ion-datetime>
      </ng-container>
    </ion-list>
  </ng-container>
</ion-content>
```

**home.page.ts**

```

import { Component } from '@angular/core';
import { PickerController, PickerOptions } from '@ionic/angular';


@Component({
  selector: 'app-home',
  templateUrl: 'home.page.html',
  styleUrls: ['home.page.scss'],
  
})
export class HomePage {


  constructor(private pickerCtrl: PickerController) {}

  async showPickerController() {
    let opts: PickerOptions = {
      buttons: [
        {
          text: 'Ok'
        }
      ],
      columns: [
        {
          name: 'category',
          options: [
            { text: 'Dog', value: 'dog' },
            { text: 'Cat', value: 'cat' },
            { text: 'Fish', value: 'fish' },
            { text: 'Rabbit', value: 'rabbit' },
          ]
        }
      ]
    };
    
    let picker = await this.pickerCtrl.create(opts);
    picker.present();
    picker.onDidDismiss().then(async data => {
      console.log('dismiss');
    });
  }
}
```

### Describe the Use Case

**Create an inshot-style font select**

select-inshot

_No response_

### Describe Alternatives

_No response_

### Related Code

_No response_

### Additional Information

_No response_

### LINKED ISSUE #26840: bug: ion-picker allows selection of disabled options between enabled options
### Prerequisites

### Ionic Framework Version

- [ ] v4.x
- [ ] v5.x
- [X] v6.x
- [ ] v7.x
- [ ] Nightly

### Current Behavior

When using an ion-picker with a list of options that begins with 1+ enabled, contains 1+ disabled options in the middle, then ends with 1+ enabled options, the selector doesn't correctly auto-select an enabled option. Further the selector allows the user to select a disabled option.

### Expected Behavior

Users should not be able to select a disabled PickerColumnOption.

PickerColumn should auto-select a valid option if the user attempts to choose a disabled option.

### Steps to Reproduce

1. Create a simple ion-picker with a set of options - recommend 7 options to easily see the issue.
2. Set the options as follows:
- 0: disabled = false
- 1: disabled = false
- 2: disabled = true
- 3: disabled = true
- 4: disabled = true
- 5: disabled = false
- 6: disabled = 6
3. Attempt to select a disabled option.

*Note, you may need to adjust the CSS for .picker-opt-disabled in order to see the disabled options more easily.

### Code Reproduction URL

https://github.com/austinrzg/angular-fqvgd8

### Ionic Info

Ionic:

   Ionic CLI : 5.4.16

Utility:

   cordova-res : not installed
   native-run  : not installed

System:

   NodeJS : v16.18.1
   npm    : 8.1.2
   OS     : Windows 10

### Additional Information

_No response_

liamdebeasi:
Issue number: resolves #24905, resolves #26840, resolves #15710

---------

The current picker is only displayed as an overlay which makes it impossible to use inline with the rest of the application. Additionally, there are several bugs and missing features. For example, it is impossible to know when the value of the picker column has changed using public APIs. This is valuable for updating other columns in response to that data.

## What is the new behavior?

- Introduces `ion-picker` as an inline wheel picker component. The overlay picker is still available using `ion-picker-legacy`.
- The implementation of this component resolves linked bugs. It also resolves linked features by exposing an `ionChange` event for developers to listen for.

- [ ] Yes
- [x] No

## Other information

Dev-build: `7.5.8-dev.11702398696.1ab62ea9`


BREAKING.md
amandaejohnston, line 1: (Threading) I noticed a couple of oddities with the datetime `custom` test, compared to `main` 🤔

The time picker popover no longer has its fade background styled:
missing fade bg

And on the wheel picker datetime, when scrolling the options, there is a delay before the active item styles apply to the newly selected item (most obvious when using the scroll wheel):

https://github.com/ionic-team/ionic-framework/assets/90629384/89497461-2163-49fe-b46b-260e9bcedbe8

amandaejohnston (changes requested):

BREAKING.md
liamdebeasi: First issue is fixed in https://github.com/ionic-team/ionic-framework/pull/28689/commits/e4797a46d2a5672bdcffc7eea6199c758af5aaca.

The second issue is working as intended. The old behavior added the Shadow Part every time the scroll listener fired and the centered item changed: https://github.com/ionic-team/ionic-framework/blob/8ee23d20d5cc7419ce15f047b92d2f826d3eb681/core/src/components/picker-column-internal/picker-column-internal.tsx#L206

However, since the shadow part is now set from `ion-datetime this part is only set when the value of the picker column changes which is done in a timeout: https://github.com/ionic-team/ionic-framework/blob/8ee23d20d5cc7419ce15f047b92d2f826d3eb681/core/src/components/picker-column-internal/picker-column-internal.tsx#L332

amandaejohnston: We discussed the second issue in Slack and agreed to wait and see if we get feedback on it before making changes.

amandaejohnston (approved):

PR #28693, opened 2023-12-12
chore: add resolves to PR template

liamdebeasi:
I've noticed a few instances where the team forgets to add the "resolves" keyword (or similar "closes"/"fixes" keywords) when linking a PR to an issue. GitHub will not link the PR with the issue and will not close the issue on PR merge unless one of these keywords is used.

I've updated the PR template to add the "resolves" keyword so team members don't need to worry about remembering to add it.


mapsandapps (approved):
Thank you!

PR #28698, opened 2023-12-13
chore: update PR template for breaking changes

liamdebeasi:
The team had feedback during a sprint retro that the breaking change process documentation was a bit spread out. In particular, some team members were not clear that you need to a) update the breaking.md file and b) add the "BREAKING CHANGE" comment to the commit description.

This PR adds both of these things are comments for awareness.


mapsandapps (approved):
Thank you!

PR #28700, opened 2023-12-13
fix(item): label does not expand indefinitely

liamdebeasi:
Issue number: N/A

---------

While working on an unrelated task, I noticed the list items in the list starter app run off the page. This is due to the text wrapping behavior in v7.6 where we allows the labels to grow/shrink. However, we do not set a limit on how much they can grow. This means that child elements inside of `ion-label` with long text can cause the label container to grow indefinitely. 

## What is the new behavior?

- Labels in item have a max-width of 100%

- [ ] Yes
- [x] No

## Other information

Dev build: `7.6.1-dev.11702498225.1ad6eda6`

| before | after |
| - | - |
| localhost_8102_home(Samsung Galaxy S8+) | localhost_8102_home(Samsung Galaxy S8+) (1) |


brandyscarney (approved):
This looks good but should we put this in the `item` test since it only applies when a label is inside of an item? I am good with either way though!

PR #28705, opened 2023-12-14
feat: add experimental hardware back button support in browsers

### LINKED ISSUE #28703: feat: Use CloseWatcher API for modals etc
### Prerequisites

It could be useful for ionic to make use of the new close watcher api [1][2] to listen for close signals to close components such as alert dialogs. 

[1] https://html.spec.whatwg.org/multipage/interaction.html#the-closewatcher-interface
[2] https://developer.chrome.com/blog/new-in-chrome-120#close-watcher

### Describe the Use Case

This API allows listening for escape key presses, along with more platform specific close signals such as Android's back button/ gesture. This would provide a more native feel (matching both the native platform and the web).

It would also improve accesibility as it supports Talkback back gestures too.

The close watcher API being used when supported feels like the best solution.

### Describe Alternatives

_No response_

### Related Code

_No response_

### Additional Information

_No response_

liamdebeasi:
Issue number: resolves #28703

---------

Users are unable to dismiss modals and menus via the hardware back button in a mobile web browser/PWA.

## What is the new behavior?

- This PR adds a new feature that allows developers to leverage the CloseWatcher API. This enables hardware back button support in web browsers/PWAs.

- [ ] Yes
- [x] No

## Other information

Dev build: `7.6.1-dev.11702569564.1746fa9f`

Full design doc: https://github.com/ionic-team/ionic-framework-design-documents/pull/205

### DIFF
--- a/core/src/components/app/app.tsx
+++ b/core/src/components/app/app.tsx
@@ -2,6 +2,7 @@ import type { ComponentInterface } from '@stencil/core';
 import { Build, Component, Element, Host, Method, h } from '@stencil/core';
 import type { FocusVisibleUtility } from '@utils/focus-visible';
 import { shoudUseCloseWatcher } from '@utils/hardware-back-button';
+import { printIonWarning } from '@utils/logging';
 import { isPlatform } from '@utils/platform';
 
 import { config } from '../../global/config';
@@ -39,6 +40,16 @@ export class App implements ComponentInterface {
         if (config.getBoolean('hardwareBackButton', supportsHardwareBackButtonEvents)) {
           hardwareBackButtonModule.startHardwareBackButton();
         } else {
+          /**
+           * If an app sets hardwareBackButton: false and experimentalCloseWatcher: true
+           * then the close watcher will not be used.
+           */
+          if (shoudUseCloseWatcher) {
+            printIonWarning(
+              'experimentalCloseWatcher was set to `true`, but hardwareBackButton was set to `false`. Both config options must be `true` for the Close Watcher API to be used.'
+            );
+          }
+
           hardwareBackButtonModule.blockHardwareBackButton();
         }
         if (typeof (window as any) !== 'undefined') {
--- a/core/src/components/app/app.tsx
+++ b/core/src/components/app/app.tsx
@@ -36,7 +36,7 @@ export class App implements ComponentInterface {
           import('../../utils/input-shims/input-shims').then((module) => module.startInputShims(config, platform));
         }
         const hardwareBackButtonModule = await import('../../utils/hardware-back-button');
-        const supportsHardwareBackButtonEvents = isHybrid || shoudUseCloseWatcher;
+        const supportsHardwareBackButtonEvents = isHybrid || shoudUseCloseWatcher();
         if (config.getBoolean('hardwareBackButton', supportsHardwareBackButtonEvents)) {
           hardwareBackButtonModule.startHardwareBackButton();
         } else {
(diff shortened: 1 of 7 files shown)


core/src/utils/browser/index.ts
```diff
@@ -72,7 +72,32 @@ type IonicEvents = {
   ): void;
 };
 
-type IonicWindow = Window & IonicEvents;
+export interface CloseWatcher extends EventTarget {
+  new (options?: CloseWatcherOptions): any;
```
liamdebeasi, line 76: Having this return `CloseWatcher` gives me the following lint error:

```
/Users/liamdebeasi/Ionic/ionic/core/src/utils/browser/index.ts
  76:3  error  Interfaces cannot be constructed, only classes  @typescript-eslint/no-misused-new

✖ 1 problem (1 error, 0 warnings)
```

However, using `any` still gives us type checking:

```
[ ERROR ]  TypeScript: src/utils/hardware-back-button.ts:120:16
           Property 'liam' does not exist on type
           'CloseWatcher'.

    L119:   */
    L120:  watcher!.liam = true;
    L121:  watcher!.requestClose();
```

core/src/components/app/app.tsx
```diff
@@ -1,6 +1,7 @@
 import type { ComponentInterface } from '@stencil/core';
 import { Build, Component, Element, Host, Method, h } from '@stencil/core';
 import type { FocusVisibleUtility } from '@utils/focus-visible';
+import { shoudUseCloseWatcher } from '@utils/hardware-back-button';
```
liamdebeasi, line 4: I'm still deciding if we should target Ionic 8 vs. Ionic 7.7. Anyone have strong feelings one way or the other?

thetaPC: No strong feelings since it's an opt-in feature. Maybe 7.7 b/c the sooner it comes out then the sooner the community can provide feedback if needed.

amandaejohnston: ^ Agreed.

core/src/utils/config.ts
```diff
@@ -204,6 +204,14 @@ export interface IonicConfig {
    */
   platform?: PlatformConfig;
 
+  /**
+   * @experimental
+   * If `true`, the [CloseWatcher API](https://github.com/WICG/close-watcher) will be used to handle
+   * all Escape key and hardware back button presses to dismiss menus and overlays and to navigate.
+   * Note that the `hardwareBackButton` config option must also be `true`.
```
amandaejohnston, line 211: Optional: Could we log a warning if the CloseWatcher config is used but `hardwareBackButton` is still false?

amandaejohnston (approved):

thetaPC (changes requested):

core/src/utils/config.ts
liamdebeasi: Good idea! Added in https://github.com/ionic-team/ionic-framework/pull/28705/commits/f34e136df164854d6e9189ad9f2250e972f45539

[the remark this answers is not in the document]
liamdebeasi: In the event that `hardwareBackButton` is `false` and `experimentalCloseWatcher` is `true` there would be a warning logged instead: https://github.com/ionic-team/ionic-framework/pull/28705#discussion_r1431890288. Does that answer your question?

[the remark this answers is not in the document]
thetaPC: Yes, thank you!

thetaPC (approved):
LGTM

liamdebeasi:
Will put up docs in a separate PR

PR #28709, opened 2023-12-15
fix(content): main landmark is set correctly

liamdebeasi:
Issue number: Internal

---------

Maria noted that the way we set the main landmark is not correct and causes accessibility violations.

## What is the new behavior?

- Instead of rendering a main element inside of ion-content, the content itself is the main landmark.

- [ ] Yes
- [x] No

### DIFF
--- a/core/src/components/content/content.tsx
+++ b/core/src/components/content/content.tsx
@@ -428,12 +428,12 @@ export class Content implements ComponentInterface {
     const mode = getIonMode(this);
     const forceOverscroll = this.shouldForceOverscroll();
     const transitionShadow = mode === 'ios';
-    const TagType = isMainContent ? 'main' : ('div' as any);
 
     this.resize();
 
     return (
       <Host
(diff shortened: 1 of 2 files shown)


core/src/components/content/content.tsx
```diff
@@ -446,19 +446,19 @@ export class Content implements ComponentInterface {
...
-        <TagType
+        <div
           class={{
             'inner-scroll': true,
             'scroll-x': scrollX,
             'scroll-y': scrollY,
             overscroll: (scrollX || scrollY) && forceOverscroll,
           }}
-          ref={(scrollEl: HTMLElement) => (this.scrollEl = scrollEl!)}
+          ref={(scrollEl) => (this.scrollEl = scrollEl)}
```
liamdebeasi, line 456: The types can be inferred here since we are using a div element directly

sean-perkins (approved):

PR #28717, opened 2023-12-18
fix(refresher): mode property can be used in typescript

### LINKED ISSUE #28716: bug(refresher): Can't bind to 'mode' since it isn't a known property of 'ion-refresher'
### Prerequisites

### Ionic Framework Version

v7.x

### Current Behavior

Currently i cannot apply `mode` dynamically using angular value binding for `ion-refresher`.

### Expected Behavior

`mode` can be applied dynamically

### Steps to Reproduce

see stackblitz

### Code Reproduction URL

https://stackblitz.com/edit/angular-2fmpy9?file=src%2Fapp%2Fapp.module.ts,src%2Ftheme%2Fvariables.css,src%2Fapp%2Fexample.component.ts,src%2Fapp%2Fexample.component.html,src%2Fapp%2Fapp.component.css

### Ionic Info

latest

### Additional Information

_No response_

liamdebeasi:
Issue number: resolves #28716

---------

All components that have per-mode stylesheets can have their mode adjusted by setting `mode` on the component. We use the `setMode` function to determine which mode to use on the component: https://github.com/ionic-team/ionic-framework/blob/516b84475e5d78060f35fa2c4821efc712536353/core/src/global/ionic-global.ts#L75

While this works on refresher, it is missing the `virtualProp` jsdoc comment which causes it to not have the appropriate type information. As a result, when developers try to use a JS binding for `mode`, they will get compilation errors.

## What is the new behavior?

- Adds virtualProp for mode to refresher

- [ ] Yes
- [x] No

## Other information

I am considering this a bug fix instead of a feature. In non-TypeScript environments you can set `mode` on `ion-refresher` and it does change the mode. What's missing here is the type information associated with it.

Dev build: `7.6.2-dev.11702914017.1ae72da5`

### DIFF
--- a/core/api.txt
+++ b/core/api.txt
@@ -1087,6 +1087,7 @@ ion-range,part,tick-active
 ion-refresher,none
 ion-refresher,prop,closeDuration,string,'280ms',false,false
 ion-refresher,prop,disabled,boolean,false,false,false
+ion-refresher,prop,mode,"ios" | "md",undefined,false,false
 ion-refresher,prop,pullFactor,number,1,false,false
 ion-refresher,prop,pullMax,number,this.pullMin + 60,false,false
 ion-refresher,prop,pullMin,number,60,false,false
--- a/core/src/components.d.ts
+++ b/core/src/components.d.ts
@@ -2378,6 +2378,10 @@ export namespace Components {
          */
         "getProgress": () => Promise<number>;
         /**
+          * The mode determines which platform styles to use.
+         */
+        "mode"?: "ios" | "md";
+        /**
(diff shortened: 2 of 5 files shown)


DwieDima:
wow that was fast! 🚀
could you release the fix on both, ionic 6 and 7?

liamdebeasi:
This will only be an Ionic 7 fix. We only back port fixes to Ionic 6 for those who have an "Extended Support" plan with Ionic: https://ionicframework.com/docs/reference/support#framework-maintenance-and-support-status

thetaPC (approved):
LGTM, works as expected

Issue #28694, opened 2023-12-13
feat: Make ion-input counter formatter more beginner-friendly


johannesvollmer:
### Prerequisites

- [X] I have read the [Contributing Guidelines](https://github.com/ionic-team/ionic-framework/blob/main/.github/CONTRIBUTING.md#creating-an-issue).
- [X] I agree to follow the [Code of Conduct](https://ionicframework.com/code-of-conduct).
- [X] I have searched for [existing issues](https://github.com/ionic-team/ionic-framework/issues) that already include this feature request, without success.

### Describe the Feature Request

We suggest that the `counterFormatter` attribute on `ion-input` can be passed an interface, in addition to a callback. This will avoid a potential pitfall, thus making ionic more beginner friendly.

We added internationalization to our ion input counter formatter. But the current design is inconvenient, as it requires the use of `bind` or an unidiomatic arrow function in the class.

### Describe the Use Case

The [example](https://ionicframework.com/docs/api/input#input-counter) in the documentation is error-prone:
 
```html
<ion-input
  [counter]="true"
  maxlength="20"
  [counterFormatter]="customCounterFormatter"
></ion-input>
```

```ts
@Component(...)
export class ExampleComponent {
  customCounterFormatter(inputLength: number, maxLength: number) {
    return `${maxLength - inputLength} characters remaining`;
  }
}
```

As soon as you add a class member, the code breaks, as the function looses the `this` context. It's not beginner friendly, and results in combersome code.

For example, when adding internationalization:

```ts
export class ExampleComponent {
  translated: string = "characters remaining" // probably injected dynamically

  customCounterFormatter(inputLength: number, maxLength: number) {
    return `${maxLength - inputLength} {this.translated}`;
  }
}
```

This now results in an error: `Exception in provided 'counterFormatter'`, as `this` is not bound when ionic calls the formatter callback. 

### Describe Preferred Solution

Introduce an 

```ts
interface CounterFormatter {
    format(current: number, limit: number): string
}
```

It should be possible to pass an object of type `CounterFormatter` to the `counterFormatter` attribute on the `ion-input`.
We prefer this solution because it is the most explicit one.

### Describe Alternatives

Either
- `callback.bind(component)` to the component before calling the callback inside the `ion-input` (100% backwards compatible, but not as explicit)
- change the [example](https://ionicframework.com/docs/api/input#input-counter) to include a call `bind` or an arrow function

### Additional Information

[...]

PR #28720, opened 2023-12-18
docs: account for this context

liamdebeasi:
Issue number: N/A

---------

In https://github.com/ionic-team/ionic-framework/issues/28694 there was some confusion around how to access `this` inside of a callback function passed to a property on Ionic components. The root issue was due to how the `this` context is determined with developers being responsible for setting the appropriate `this` context.

## What is the new behavior?

- While this isn't an Ionic bug, I think it's worth calling out this behavior so developers are aware of how to account for it.

- [ ] Yes
- [x] No

## Other information

Note: The link in the docs will not work until https://github.com/ionic-team/ionic-docs/pull/3333 is merged.

### DIFF
--- a/core/src/components/datetime/datetime.tsx
+++ b/core/src/components/datetime/datetime.tsx
@@ -342,9 +342,8 @@ export class Datetime implements ComponentInterface {
    * selected (i.e. unused for exactly 1). By default, the header
    * text is set to "numberOfDates days".
    *
-   * Developers who wish to access "this" inside of the function
-   * should either use an arrow function or manually bind "this"
-   * using `.bind(this)` when passing the function to titleSelectedDatesFormatter.
+   * See https://ionicframework.com/docs/troubleshooting/runtime#accessing-this
+   * if you need to access `this` from within the callback.
    */
   @Prop() titleSelectedDatesFormatter?: TitleSelectedDatesFormatter;
 
--- a/core/src/components/datetime/datetime.tsx
+++ b/core/src/components/datetime/datetime.tsx
@@ -341,6 +341,10 @@ export class Datetime implements ComponentInterface {
    * dates are selected. Only used if there are 0 or more than 1
    * selected (i.e. unused for exactly 1). By default, the header
    * text is set to "numberOfDates days".
+   *
+   * Developers who wish to access "this" inside of the function
+   * should either use an arrow function or manually bind "this"
+   * using `.bind(this)` when passing the function to titleSelectedDatesFormatter.
    */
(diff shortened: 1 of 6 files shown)


core/src/components/datetime/datetime.tsx
```diff
@@ -341,6 +341,10 @@ export class Datetime implements ComponentInterface {
    * dates are selected. Only used if there are 0 or more than 1
    * selected (i.e. unused for exactly 1). By default, the header
    * text is set to "numberOfDates days".
+   *
+   * Developers who wish to access "this" inside of the function
```
thetaPC, line 345:
```suggestion
   * Angular developers who wish to access "this" inside of the function
```
Should it specify the framework? My concern is that other framework users might get confused with the comment.

amandaejohnston (commented):
I question if this is the best way to document this tip, for a few reasons:

1. It creates a lot of duplicate text, which is annoying to maintain.
2. I can see us easily forgetting to add the same tip to any future function props.
3. It seems to be Angular-specific -- I was unable to access the right `this` in [React](https://stackblitz.com/edit/am4g8o?file=src%2Fmain.tsx) (using an arrow function) or [Vue](https://stackblitz.com/edit/madcxg?file=src%2Fcomponents%2FExample.vue) (using `.bind(this)`). If it does actually apply to React or Vue, but my example is just subtly wrong, that only speaks to the need for example code somewhere.
 
Maybe we could put it in [Development Tips](https://ionicframework.com/docs/developing/tips) instead? Or a page in the Angular-specific docs if it really is only relevant to Angular? I know that's not as discoverable, but my understanding is that this isn't an Ionic-specific issue, so I'm unsure if we need to give it much priority in the docs.

thetaPC (approved):
LGTM

core/src/components/modal/modal.tsx
```diff
@@ -264,6 +264,10 @@ export class Modal implements ComponentInterface, OverlayInterface {
    *
    * If the value is `true` or the value's function returns `true`, the modal will close when trying to dismiss.
    * If the value is `false` or the value's function returns `false`, the modal will not close when trying to dismiss.
+   *
+   * Developers passing a function who wish to access "this" inside of the function
+   * should either use an arrow function or manually bind "this"
+   * using `.bind(this)` when passing the function to canDismiss.
```
amandaejohnston, line 270:
```suggestion
   * See https://ionicframework.com/docs/troubleshooting/runtime#accessing-this
   * if you need to access `this` from within the callback.
```

amandaejohnston (approved):
LGTM aside from a component that was missed when updating to the link.

PR #28723, opened 2023-12-18
feat(action-sheet): add disabled button

liamdebeasi:
Issue number: N/A
---------

Action sheet buttons cannot be disabled. This behavior exists in iOS 17.

## What is the new behavior?

- Action sheet buttons can be disabled

- [ ] Yes
- [x] No


core/src/components/action-sheet/action-sheet-interface.ts
```diff
@@ -26,4 +26,5 @@ export interface ActionSheetButton<T = any> {
   htmlAttributes?: { [key: string]: any };
   handler?: () => boolean | void | Promise<boolean | void>;
   data?: T;
+  disabled?: boolean;
```
sean-perkins, line 29: ~ We should add a JSDoc comment discussing this property. Specifically to use it for buttons that are not the cancel action. This information should come through wherever the TS language server can parse the types, avoiding developers solely having to rely on the documentation on the site.

sean-perkins (approved):
Features looks good. I agree with your approach of disabling the functionality vs. hacking around it in CSS. 

One comment, non-blocking, but I think a nice improvement.

amandaejohnston (approved):

core/src/components/action-sheet/action-sheet-interface.ts
liamdebeasi: Added in https://github.com/ionic-team/ionic-framework/pull/28723/commits/cff8ce7d413265b00b086d7e78bb4714db62b557

PR #28724, opened 2023-12-18
feat(action-sheet): align with iOS spec

liamdebeasi:
Updates the Action Sheet border radius to align with the iOS 17 spec


liamdebeasi:
Moving to draft. I need to manually generate screenshots since the diff is so small.

liamdebeasi:
Upon further investigation, this change is not needed. 

| 13px | 14px |
| - | - |
| action-sheet-13-overlay | action-sheet-14-overlay |

It's a small difference, but 14px is definitely not correct here. Tested on an iPhone 12 Pro on iOS 17. I opened the "New Event" modal in the Calendar app. Type something into the event title and hit the cancel button. An action sheet will then show up. I used this to compare.

PR #28726, opened 2023-12-18
feat(alert): update styles to iOS 17 specs

amandaejohnston:
Issue number: N/A

---------

Alerts don't match the iOS 17 spec for styling.

## What is the new behavior?

- Overlay border radius increased.
- Input font size increased.
- Input padding increased.

- [ ] Yes
- [x] No

## Other information

The difference in the MD screenshots is the box shadow on the header in the background. This was a change made a while back that probably wasn't caught due to the pixel threshold on the screenshot update job. Deleting all screenshots and running again got it updated.


amandaejohnston:
One sec, need to add a border radius update that was missed in the design doc 👀

amandaejohnston:
Okay, should be good now 👍

core/src/components/alert/test/basic/alert.e2e.ts-snapshots/alert-basic-md-ltr-Mobile-Firefox-linux.png
amandaejohnston, line 1: The difference here is the box shadow on the header in the background. This was a change made a while back that probably wasn't caught due to the pixel threshold on the screenshot update. Deleting all screenshots and running again got it updated.

core/src/components/alert/test/basic/alert.e2e.ts-snapshots/alert-confirm-md-ltr-Mobile-Chrome-linux.png
amandaejohnston, line 1: Another header box shadow change.

liamdebeasi:
After discussing with Amanda offline, I made a few adjustments so the alert inputs can be more closely aligned. It's likely that there are other areas where we do not perfectly align with the spec. However, I opted to focus on the inputs and wrapper border radius since that was the agreed upon scope.

I tested this by overlaying an Ionic alert with a native alert. The following image verifies that the wrapper border radius value is aligned:

Screenshot 2023-12-19 at 4 20 53 PM

The following image verifies that the input styles are aligned:

Screenshot 2023-12-19 at 4 21 03 PM

liamdebeasi (approved):
I've made my changes. I verified them against a real iOS device, but I'd appreciate it if you could take a quick look too before you merge 😄

amandaejohnston:
Looks great to me, thanks again!

PR #28728, opened 2023-12-18
feat(searchbar): update styles to iOS 17 specs

sean-perkins:
Issue number: Internal

---------

The searchbar implementation for Ionic's iOS mode does not align with the appearance of a native search control in iOS 17.

## What is the new behavior?

- Searchbar background color opacity has increased.
- Searchbar background color has changed from the text color by default, to `118, 118, 128`
- The spacing between the searchbar and the cancel button has increased to `16px`.

- [x] Yes
- [ ] No

Impact is minimal. Developers can customize the appearance of the control if they desire a similar appearance to before. Changes are noted in the breaking changes guide.


BREAKING.md
```diff
@@ -70,4 +71,11 @@ This section details the desktop browser, JavaScript framework, and mobile platf
 
 - `ion-picker` and `ion-picker-column` have been renamed to `ion-picker-legacy` and `ion-picker-legacy-column`, respectively. This change was made to accommodate the new inline picker component while allowing developers to continue to use the legacy picker during this migration period.
   - Only the component names have been changed. Usages such as `ion-picker` or `IonPicker` should be changed to `ion-picker-legacy` and `IonPickerLegacy`, respectively.
-  - Non-component usages such as `pickerController` or `useIonPicker` remain unchanged. The new picker displays inline with your page content and does not have equivalents for these non-component usages. 
\ No newline at end of file
+  - Non-component usages such as `pickerController` or `useIonPicker` remain unchanged. The new picker displays inline with your page content and does not have equivalents for these non-component usages. 
+
+<h4 id="version-8x-searchbar">Searchbar</h2>
```
thetaPC, line 76: Shouldn't this be a `h3`?

sean-perkins: The component sub-levels are currently broken out as h4 elements: https://github.com/ionic-team/ionic-framework/blob/58c45515562b9c85cb2aa3dff198a903af95387f/BREAKING.md?plain=1#L51

The "Picker" heading was the only one that didn't follow the existing pattern & has been addressed here: https://github.com/ionic-team/ionic-framework/pull/28725. 

This aligns with the formatted we followed for v7: https://github.com/ionic-team/ionic-framework/blob/58c45515562b9c85cb2aa3dff198a903af95387f/BREAKING_ARCHIVE/v7.md?plain=1#L65-L67

thetaPC: Hmm, but this would fail accessibility checks for headings.

sean-perkins: The axe devtools pass on the page(outside of a color contrast issue from GitHub). The heading ranks mentions:
> Skipping heading ranks can be confusing and should be avoided where possible

If you want to open that conversation to the entire team in Slack, I think that's great. That shouldn't block this work though.

thetaPC (approved):
LGTM, verified against the iOS native Calendar app

brandyscarney:
It looks like the cancel text also got larger. Are you seeing this too?

Screenshot 2023-12-20 at 5 34 32 PM


Another thing I noticed when looking at the Searchbar in Safari on iOS is that when the text centers the search icon doesn't center with it. I am not sure if this is new with iOS 17 and I can't find another app where it centers but we might want to add this as "out of scope" for now. 

IMG_9207

brandyscarney (changes requested):
It looks like the cancel text also got larger. I can reproduce it using the images in your description as well as my own images taken from my device:

Screenshot 2023-12-20 at 5 34 32 PM


Another thing I noticed when looking at the Searchbar in Safari on iOS is that when the text centers the search icon doesn't center with it. I am not sure if this is new with iOS 17 and I can't find another app where it centers but we might want to add this as "out of scope" for now. 

IMG_9207

sean-perkins:
@brandyscarney good catch on the cancel button text. Updating the screenshots now.

I'm not sure i follow the Safari comment. Developers can add `text-align: center` to the searchbar to accomplish a similar display style (icon remains left, text centers, but to the remaining space - not the entire container). However I don't think the Safari searchbar is stock UI. It has a custom implementation where on focus on the input covers the search icon.

brandyscarney:
You're right, I can't find any other examples where the searchbar behaves that way in iOS. I also tried reproducing it in SwiftUI but their default is left aligned. I see what you're saying that users can center it, but it doesn't give the animation like we have in our [docs](https://ionicframework.com/docs/api/searchbar#basic-usage) when it is centered that way (see the last example). It might be good to note it either way, that we chose not to change the animated implementation to leave the icon static.

liamdebeasi:
Looks like I got tagged as a reviewer due to a flaky screenshot update. Going to remove myself from a review since this PR already has 2 reviewers.

liamdebeasi (approved):

PR #28733, opened 2023-12-18
feat(picker): update styles to iOS 17 specs

mapsandapps:
⚠️ I'm gonna check against the device again before reopening this

Issue number: Internal

---------

The Picker styles do not match iOS 17.

## What is the new behavior?

- Active Picker highlight border radius has been decreased
- Active Picker highlight height has been increased

- [x] Yes
- [ ] No


mapsandapps:
Closing based on findings in https://github.com/ionic-team/ionic-framework-design-documents/pull/214

PR #28745, opened 2023-12-20
fix(nav, router-outlet): ios page transition does not cover menu on larger screens

### LINKED ISSUE #28737: bug: page transition slides over menu when visible on larger screen
### Prerequisites

### Ionic Framework Version

v7.x

### Current Behavior

On an iPad if you have a split pane and are using the ion-router-outlet and issue a push command in Vue, it used to just animate the router page area of the screen but now the transition slides over the entire screen including the menu. The web and android tablet builds only have a transition animation in the page section (they move up and down), it is just iPad builds that navigate over top of the menu as it pushes towards either the left or right.

### Expected Behavior

This really should only transition between the two pages and not push the view over top of the menu. 

### Steps to Reproduce

1. Have a split-pane project in vue with a menu using ion-router-outlet
2. Make two views and navigate from one view to the other using a push command

### Code Reproduction URL

_No response_

### Ionic Info

Ionic:

   Ionic CLI       : 7.1.1 (/opt/homebrew/lib/node_modules/@ionic/cli)
   Ionic Framework : @ionic/vue 7.6.1

Capacitor:

   Capacitor CLI      : 5.5.1
   @capacitor/android : 5.5.1
   @capacitor/core    : 5.5.1
   @capacitor/ios     : 5.5.1

Utility:

   native-run (update available: 2.0.0) : 1.7.4

System:

   NodeJS : v20.8.0 (/opt/homebrew/Cellar/node/20.8.0/bin/node)
   npm    : 10.2.5
   OS     : macOS Unknown

### Additional Information

_No response_

liamdebeasi:
Issue number: resolves #28737

---------

In https://github.com/ionic-team/ionic-framework/pull/28246 we removed the overflow on Nav and Router Outlet to allow content to flow outside of these containers. This allows the translucent tab effect to work (otherwise content would be clipped and there would be no translucency). However, this had the unintended side effect of causing page transitions to flow outside of these components. This is most noticeable on larger displays when using SplitPane because the page can cover the menu mid-transition.

## What is the new behavior?

- Overflow is no longer allowed on the main content. I originally set the overflow on the router outlet/nav itself, but that caused the translucent tab bar behavior to regress. Since this issue only happens with split pane, I decided to apply the fix there.

- [ ] Yes
- [x] No

### Dev build
`7.6.3-dev.11703103144.148eb1f6`

⚠️ Please test in a physical app too

### Before/After Demo

| `main` | branch |
| - | - |
| <video src="https://github.com/ionic-team/ionic-framework/assets/2721089/2f3f0d74-9a7e-4ebe-b58a-6e1a6ea3636e"></video> | <video src="https://github.com/ionic-team/ionic-framework/assets/2721089/cdbf8fb5-e217-48cf-8c1e-4bdecee4de4c"></video> |

### Screenshot Diffs

The `menu-custom` screenshot diffs ([Example](https://github.com/ionic-team/ionic-framework/pull/28745/files)) are correct. By adding `overflow: hidden`, the box shadow from the "Content" header no longer flows outside of the container. 

The menu [border on MD has an opacity of 0.18](https://github.com/ionic-team/ionic-framework/blob/e5226016a0f0b066a7bd7fc9997f905d3b87fbc4/core/src/components/menu/menu.md.vars.scss#L7), so the border color was mixing with the color of the box shadow from the header.

Since the shadow no longer flows outside of the container, the border color does not mix with the box shadow color.

### Does this break translucent tabs when the split pane is inside of a tab?

No. The [split pane has `contain: strict` set](https://github.com/ionic-team/ionic-framework/blob/e5226016a0f0b066a7bd7fc9997f905d3b87fbc4/core/src/components/split-pane/split-pane.scss#L24) which prevents content from flowing under the tab bar, so the translucent tab bar never worked with this layout.


thetaPC (approved):
LGTM, works great!

PR #28754, opened 2023-12-21
feat(tab-button): update styles to iOS 17 specs

mapsandapps:
Issue number: Internal

---------

Tab Button icons on iOS are all 30px font size with or without a label.

Screenshot 2023-12-21 at 2 12 56 PM

## What is the new behavior?

- Tab button icons are still 30px font size when there is no label
- Tab button icons are 24px font size when there is a label

Screenshot 2023-12-21 at 2 12 29 PM

- [ ] Yes
- [x] No


sean-perkins (approved):
Font size for the label looks perfect when compared against the App Store tab bar: 

Group 14

Our icons are slightly smaller, but our icons in general have odd bounding boxes that will make it difficult to get a perfect fit for all icons with the same font size when compared against native.

PR #28759, opened 2023-12-22
feat(progress-bar): update styles to iOS 17 specs

sean-perkins:
Issue number: Internal

---------

Ionic's progress bar implementation deviates from the appearance of a progress bar on iOS 17.

## What is the new behavior?

- Progress bar is updated to have rounded corners (around the container).
- Increases progress bar height to `4px`
- Solid progress bars (progress bars with a buffer value of `1`) will have a consistent background appearance. 
  - Buffer progress bars will continue to have the colored background appearance as they do in v7. 

- [x] Yes
- [ ] No

### Breaking Changes

- `--buffer-background` has been removed in favor of `--background`. 

## Other information

Accomplishing rounded caps for the inner progress with Ionic's scale transformation was not viable. To maintain 60fps animations for the progress, rounded corners were applied to the container.


core/src/components/progress-bar/progress-bar.ios.vars.scss
```diff
@@ -0,0 +1,8 @@
+// iOS Progress bar
+// --------------------------------------------------
+
+/// @prop - Height of the progress bar
+$progress-bar-ios-height: 3px !default;
```
mapsandapps, line 5: Both in Figma and on the simulator, this appears to be 4px. What did you find that has a value of 3?

core/src/components/progress-bar/progress-bar.ios.scss
```diff
@@ -1,8 +1,22 @@
 @import "./progress-bar";
+@import "./progress-bar.ios.vars";
 
 // iOS Progress bar
 // --------------------------------------------------
 
 :host {
-  height: 3px;
+  @include border-radius($progress-bar-ios-border-radius);
```
mapsandapps, line 8: In Figma and Xcode, this border radius applies both to the track (container) and the progress (value) bars. I think I remember some discussion about how we've decided not to round the corner of the progress bar because of animation challenges. Could we add that information to the design doc?

This PR:
Screenshot 2023-12-27 at 10 00 55 AM


Xcode:
Screenshot 2023-12-27 at 10 17 31 AM

mapsandapps (changes requested):

core/src/components/progress-bar/progress-bar.ios.vars.scss
sean-perkins: This was bad refactoring on my part, should be `4px` 🤦

core/src/components/progress-bar/progress-bar.ios.scss
sean-perkins: I can definitely add that extra context in the design document 👍 I put a note in the other info section of the PR for a brief amount of context:
> Accomplishing rounded caps for the inner progress with Ionic's scale transformation was not viable. To maintain 60fps animations for the progress, rounded corners were applied to the container.

I tried clip-path, pseudo elements, SVG cap elements for the edges of the progress, etc. All get skewed by the scale transformation on the actual progress bar. We are looking at a much larger overhaul to change the implementation from a scale transformation to something else (likely a translate would suffice).

mapsandapps (approved):
looks good 👍🏻

core/src/components/progress-bar/test/buffer/index.html
```diff
@@ -14,14 +14,6 @@
     <script type="module" src="../../../../../dist/ionic/ionic.esm.js"></script>
 
     <style>
-      .custom-bar-background {
-        --buffer-background: red;
-      }
-
-      .no-bar-background {
-        --buffer-background: none;
-      }
```
liamdebeasi, line 23: These classes were not applied anywhere in the file

liamdebeasi (approved):

PR #28767, opened 2023-12-29
feat(themes): add dark theme files and mixins

mapsandapps:
Issue number: Internal

---------

The dark theme is not exported for use in Ionic Framework apps

## What is the new behavior?

- The dark theme is usable in Ionic Framework apps via any of the following:
  - The `ion-theme-dark` class (once the developer assigns this class to the `html` element in their app) via dark.class.css
  - The `prefers-color-scheme` media query via dark.system.css
  - Throughout the full app via dark.always.css

- [ ] Yes
- [x] No

## Other information

Usage:

```
@import '~@ionic/core/css/themes/dark.always.css';
/* @import '~@ionic/core/css/themes/dark.class.css'; */
/* @import '~@ionic/core/css/themes/dark.system.css'; */
```

replace `~@ionic/core` with the relevant package

Dev build: `7.5.8-dev.11704294569.1abc989d`

Note to reviewers: the design doc says to include these files in utils.bundle.scss. Doing this causes the dark theme to get applied by default because it includes the "always" theme in the bundled css used in Framework. Adding the imports in that file is not required for the files to get included in the built package; users can import them without them being included there.

We can bikeshed the name of the "system" file.

### DIFF
--- a/core/src/css/themes/dark.always.scss
+++ b/core/src/css/themes/dark.always.scss
@@ -1,0 +1,13 @@
+@import "./dark";
+
+:root {
+  @include dark-base-theme();
+}
+
+.ios body {
+  @include dark-ios-theme();
+}
+
+.md body {
+  @include dark-md-theme();
+}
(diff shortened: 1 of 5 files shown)


liamdebeasi (approved):
Great job!

Issue #5536, opened 2016-02-19
<ion-input> keyboard attach problem


denkomanceski:
#### Short description of the problem:
When I focus an `<ion-input>` which is positioned lower than the height of the keyboard, the input is getting hidden by the keyboard (It is not floating above so I cant see what I am typing). Also when I type 1 letter, the keyboard disappears. Another thing is that when I focus an input and the keyboard comes out, and than I try to focus another input field, the keyboard is disappearing and appearing again. 

#### What behavior are you expecting?
I am expecting the input to get scrolled up above the keyboard so I can see what I am typing. Also I should be able to type more than 1 letter

**Steps to reproduce:**
1.  Create List with 5-6 ion-items which are containing ion inputs. Than try to focus the lowest positioned (6th in this case) input, and see that you wont be able to see the input above the keyboard and when you try to type 1 letter the keyboard will disappear

**Ionic Version**: 2

**Browser & Operating System**: iOS / Android / Chrome
I have this problem on Android Nexus 5 (I am also using crosswalk). It works fine in Ionic View, but still on focusing different input the keyboard is disappearing and appearing again.

denkomanceski:
any progress on this ?

privetr:
A bug which is similar to this one : 

On Android 6.0, Ionic 2 : 
When an input which is at the bottom of the page is focused, Keyboard is opened. The Input goes up and we can see what we write on. But when we close the keyboard, the input stays at the same place, so there is a big blank at the end of the page (See attached images).

This seems critical. 
![screenshot1](https://cloud.githubusercontent.com/assets/8597289/13334931/d63a0cac-dc0e-11e5-95d7-df1fb1b9e624.png)
![screenshot2](https://cloud.githubusercontent.com/assets/8597289/13334932/d72bb778-dc0e-11e5-8a1e-05a3fdac2966.png)
![screenshot3](https://cloud.githubusercontent.com/assets/8597289/13334933/d7326636-dc0e-11e5-92dd-a32834e87022.png)

matitalatina:
Same bug here...
Now it's completely inusable.
![scrollbug](https://cloud.githubusercontent.com/assets/2093741/13723786/cc628f16-e86f-11e5-9224-d01ccfb74f34.gif)

EvanWillms:
[...]

PR #28772, opened 2024-01-02
fix(input): slotted buttons are clickable

### LINKED ISSUE #28762: bug: click event not fired on slotted buttons inside an ion-input that has focus
### Prerequisites

### Ionic Framework Version

v7.x

### Current Behavior

When using ion-button elements in the "end" slot of an ion-input, the buttons do not fire the click event when clicked while the ion-input has focus.

### Expected Behavior

The click event should fire whether the input control has focus or not.

### Steps to Reproduce

1. Add an ion-input like the following
   ```html
   <ion-item>
     <ion-input label="Input with buttons">
       <ion-button slot="end" onClick={...} >Click Me</ion-button>
     </ion-input>
   </ion-item>
   ```
2. Click the button when the input control does not have focus - onClick fires
3. Click the button when the input control has focus - onClick does not fire

### Code Reproduction URL

_No response_

### Ionic Info

Ionic:

   Ionic CLI : 5.4.16

Utility:

   cordova-res : not installed
   native-run  : not installed

System:

   NodeJS : v20.10.0
   npm    : 10.2.3
   OS     : Linux 6.1

### Additional Information

_No response_

liamdebeasi:
Issue number: resolves #28762

---------

Focused inputs have `pointer-events: none`. This code was added in https://github.com/ionic-team/ionic-framework/commit/e27452b789232c0a3222a03e75de1a9af19d3411 to address https://github.com/ionic-team/ionic-framework/issues/5536. However, this causes slotted buttons to not be clickable when the input is focused because pointer events have been removed.

This also causes the input to blur whenever you tap the label text.

This behavior only exists on `ion-input`. Textarea is not affected.

## What is the new behavior?

 This code was added several years ago for the (now) legacy syntax. I don't have full confidence that this won't break the legacy input because there's not a lot of context around why specifically this code was added. As a result, I scoped these styles only to the legacy input.
- Slotted buttons can now be clicked when the input is focused.
- Tapping the label when an input is focused keeps the input focused.

- [ ] Yes
- [x] No

## Other information

**Testing:**

1. I manually verified that scroll assist still works with the modern input.
2. I manually verified that the input remains focused after tapping the label.

Dev build: `7.6.3-dev.11704229014.12cdc767`

### DIFF
--- a/core/src/components/input/input.scss
+++ b/core/src/components/input/input.scss
@@ -259,15 +259,16 @@
 
 // Input Has focus
 // --------------------------------------------------
-// When the input has focus, then the input cover should be hidden
 
-:host(.has-focus) {
+// TODO FW-2764 Remove this
+:host(.has-focus.legacy-input) {
   pointer-events: none;
 }
 
-:host(.has-focus) input,
-:host(.has-focus) a,
-:host(.has-focus) button {
+// TODO FW-2764 Remove this
+:host(.has-focus.legacy-input) input,
+:host(.has-focus.legacy-input) a,
+:host(.has-focus.legacy-input) button {
   pointer-events: auto;
 }
 


core/src/components/input/input.scss
```diff
@@ -259,15 +259,16 @@
 
 // Input Has focus
 // --------------------------------------------------
-// When the input has focus, then the input cover should be hidden
```
liamdebeasi, line 262: The associated code for this comment was removed in https://github.com/ionic-team/ionic-framework/commit/d817cc3b30ae648730b063120b8d55b9ec09a53d#diff-25a4832c20c7342508f6e1921d7e0db8c6c7ef1b5053f643d665c8f45a84cd19R146.

thetaPC (approved):
LGTM, works well

PR #28773, opened 2024-01-02
fix(item): only default slot content wraps

### LINKED ISSUE #28769: bug: item start and end slots wrap to next line
### Prerequisites

### Ionic Framework Version

v7.x

### Current Behavior

If an element other than `ion-label` is used in the default slot and the width is long, `slot=end` will break lines.

ex:

```html
<ion-list>
  <ion-item>
    <ion-text style="background-color: lightgreen">
      A "Hello, World!" program is generally a simple computer program.
    </ion-text>
    <div slot="end" style="background-color: yellow">[slot=end]</div>
  </ion-item>
</ion-list>
```

スクリーンショット 2024-01-02 午後3 59 35

This has been occurring since v7.6.0 and does not occur in v7.5.8. In complex apps, `ion-grid` may be placed under `ion-item`, and this behavior can be a major problem.

### Expected Behavior

Equivalent behavior to v7.5.8.

スクリーンショット 2024-01-02 午後3 58 56

### Steps to Reproduce

1. `npm i @ionic/angular@7.5.8` or `npm i @ionic/angular@7.6.0`

### Code Reproduction URL

_No response_

### Ionic Info

Ionic:

   Ionic CLI                     : 7.1.1 (/opt/homebrew/lib/node_modules/@ionic/cli)
   Ionic Framework               : @ionic/angular 7.6.0
   @angular-devkit/build-angular : 17.0.8
   @angular-devkit/schematics    : 17.0.8
   @angular/cli                  : 17.0.8

Capacitor:

   Capacitor CLI      : 5.6.0
   @capacitor/core    : 5.6.0

Utility:

   cordova-res : 0.15.4
   native-run  : 2.0.0

System:

   NodeJS : v20.10.0 (/usr/local/bin/node)
   npm    : 10.2.3
   OS     : macOS Unknown

### Additional Information

_No response_

liamdebeasi:
Issue number: resolves #28769

---------

As part of https://github.com/ionic-team/ionic-framework/pull/28146, we allowed text wrapping inside of `ion-item` for accessibility purposes. One of the behaviors we added was to allow start, default, and end slotted containers to wrap to the next line to align with the iOS spec. However, this decision was based on an incorrect assumption.

The following screenshot shows the Settings app on iOS:

| default scale | 310% scale |
| - | - |
| Frame 4 | Frame 5 |

At the default scale, the blue icon is in the iOS equivalent of the "start" slot, "Bluetooth" is in the default slot, and "On" is in the "end" slot. We incorrectly assumed the same markup was true when scaling the text up. However, at 310% scale the icon, "Bluetooth" text, and "On" text all become part of the default slot in a single container that wraps. You can tell because the bottom border runs underneath the blue icon at 310% whereas it does not at the default scale. This allows the text to wrap underneath the blue icon. When we originally implemented #28146 we thought that this meant the start, default, and end slot containers should wrap to the next line.

I further validated this behavior by creating an app with Swift UI. I created a list of items where each item has the native equivalent of a checkbox in the `start` slot and multiple `ion-labels` in the default slot of the item:

| Default Scale | 310% Scale |
| - | - |
| IMG_3133 | IMG_3134 |

The content within each label wraps within the container, but the containers themselves never wrap to the next line.

Demo code:

```swift
import SwiftUI

struct ContentView: View {
    struct Item: Identifiable, Hashable {
        let id = UUID()
    }

    private var items = [
        Item(),
        Item(),
        Item(),
        Item(),
        Item()
    ]

    @State private var multiSelection = Set<UUID>()

    var body: some View {
        NavigationView {
            List(items, selection: $multiSelection) {_ in
                HStack {
                    Text("Column 1 with really long text")
                    Text("Column 2 with really long text")
                    Text("Column 3 with really long text")
                    Text("Column 4 with really long text")
                }
            }
            .toolbar { EditButton() }
        }
        Text("\(multiSelection.count) selections")
    }
}

#Preview {
    ContentView()
}
```

## What is the new behavior?

- This PR removes the ability for the start, default, and end slot containers to wrap to the next line. This behavior aligns with pre-v7.6.0 behaviors. The containers inside of the default slot will not wrap to the next line. However, content within each container (such as text within an `ion-label`) will continue to wrap to meet the team's accessibility requirements.

- [ ] Yes
- [x] No

## Other information

Dev build: `7.6.5-dev.11704916749.1e64a3a7`


core/src/components/item/test/highlight/item.e2e.ts-snapshots/item-highlight-diff-md-ltr-Mobile-Chrome-linux.png
liamdebeasi, line 1: This diff is correct. The icons were in start/end slots which now no longer wrap to the next line.

core/src/components/item/test/legacy/fill/item.e2e.ts-snapshots/item-fill-diff-md-ltr-Mobile-Chrome-linux.png
liamdebeasi, line 1: This diff is correct. The icons were in start/end slots which now no longer wrap to the next line.

core/src/components/item/item.scss
```diff
@@ -51,15 +51,6 @@
...
    */
-
-  /**
-   * We change the minimum width as the
-   * font size changes. Using a fixed minimum
-   * width means that fewer and fewer characters
-   * can be displayed in the same space as the
-   * text grows.
-   */
-  --inner-min-width: 4rem;
```
liamdebeasi, line 62: This API was only needed to control when the default slot content should stop wrapping within its container and instead have the container wrap to the next line.

core/src/components/item/test/highlight/item.e2e.ts-snapshots/item-highlight-diff-md-ltr-Mobile-Chrome-linux.png
liamdebeasi, line 1: This diff is correct. The icons were in start/end slots which now no longer wrap to the next line.

core/src/components/item/test/legacy/fill/item.e2e.ts-snapshots/item-fill-diff-md-ltr-Mobile-Chrome-linux.png
liamdebeasi, line 1: This diff is correct. The icons were in start/end slots which now no longer wrap to the next line.

core/src/components/button/test/wrap/button.e2e.ts-snapshots/button-wrap-item-button-ios-ltr-Mobile-Firefox-linux.png
liamdebeasi, line 1: This is the same "wiggling" that we saw in https://github.com/ionic-team/ionic-framework/pull/28146 being undone.

core/src/components/item/test/slotted-inputs/item.e2e.ts-snapshots/item-slotted-inputs-ios-ltr-Mobile-Chrome-linux.png
liamdebeasi, line 1: "Range" is part of the default slot, and "End" is part of the end slot. Previously the "end" slot container would wrap to the next line. However, this was removed because it was incorrect. Now both the default and end slots will appear on the same line, and the default slot will shrink.

core/src/components/note/test/a11y/note.e2e.ts-snapshots/note-wrapping-label-scale-ios-ltr-Mobile-Chrome-linux.png
liamdebeasi, line 1: These screenshot changes are a result of https://github.com/ionic-team/ionic-framework/pull/28773/commits/c602b30f77e283049914acfed00b8bae197aeeab. This screenshot tries to capture the text wrapping behavior found on native iOS when scaled up to 310%. However, as explained in the PR description, the label and note should really be in the same slot so they can wrap within the container.

thetaPC (approved):
LGTM, works as discussed

core/src/components/note/test/a11y/note.e2e.ts-snapshots/note-wrapping-label-scale-ios-ltr-Mobile-Chrome-linux.png
brandyscarney: How would developers achieve this? We have [notes documented as being slotted in items](https://ionicframework.com/docs/api/note#item-notes). The slotted notes with long labels are expanding outside of the viewport height at 310% with these changes:

| 100% | 310% |
| ---| ---|
|  |  |

vs. inside of the label at 100% they are not clinging to the sides:

| 100% | 310% |
| ---| ---|
|  |  |

liamdebeasi:
I need to do some clean up and address feedback, so I'll re-request for review when I'm reading 😄

core/src/components/button/test/wrap/button.e2e.ts-snapshots/button-wrap-item-button-ios-ltr-Mobile-Firefox-linux.png
liamdebeasi, line 1: This is the same "wiggling" that we saw in https://github.com/ionic-team/ionic-framework/pull/28146 being undone.

core/src/components/item/test/slotted-inputs/item.e2e.ts-snapshots/item-slotted-inputs-ios-ltr-Mobile-Chrome-linux.png
liamdebeasi, line 1: "Range" is part of the default slot, and "End" is part of the end slot. Previously the "end" slot container would wrap to the next line. However, this was removed because it was incorrect. Now both the default and end slots will appear on the same line, and the default slot will shrink.

core/src/components/item/test/highlight/item.e2e.ts-snapshots/item-highlight-diff-md-ltr-Mobile-Chrome-linux.png
liamdebeasi, line 1: This diff is correct. The icons were in start/end slots which now no longer wrap to the next line.

core/src/components/item/test/legacy/fill/item.e2e.ts-snapshots/item-fill-diff-md-ltr-Mobile-Chrome-linux.png
liamdebeasi, line 1: This diff is correct. The icons were in start/end slots which now no longer wrap to the next line.

core/src/components/note/test/a11y/note.e2e.ts-snapshots/note-wrapping-label-scale-ios-ltr-Mobile-Chrome-linux.png
liamdebeasi: So this is an area where we don't have a great solution right now. I'm not aware of how developers can easily  have a long note go from the start slot to the default slot when the font size changes apart from putting the content in both slots and selectively hiding one based on the font size using [`mod()` similar to what we do in datetime](https://github.com/ionic-team/ionic-framework/blob/9262f7da15969706039e9d5f11ef5c7bd73ced0d/core/src/components/datetime/datetime.ios.scss#L76-L130).

However, I'd argue having both long notes and long labels in the same item isn't something we should encourage. 

According to https://ionicframework.com/docs/api/item#metadata, any notes should be short/to the point such as a count or one/two words.

If developers follow that best practice, they'd get something like this:

| Default Scale | 310% Scale |
| - | - | 
| IMG_3135 | IMG_3136 |

liamdebeasi:
I made a few more changes, so re-requesting review.

core/src/components/note/test/a11y/note.e2e.ts-snapshots/note-wrapping-label-scale-ios-ltr-Mobile-Chrome-linux.png
liamdebeasi: One other idea is you could use a [ResizeObserver](https://developer.mozilla.org/en-US/docs/Web/API/ResizeObserver) on an element/container and then do something when that fires? I doubt that would be a very performant way of doing it though.

amandaejohnston: My main concern is that in moving the notes inside the labels, we're testing a pattern we don't actually expect developers to use, if there isn't a good way to accomplish the swap.

liamdebeasi: What do you propose instead?

thetaPC (approved):
Still LGTM, re-request me if the code changes due to this [thread](https://github.com/ionic-team/ionic-framework/pull/28773#discussion_r1462302438).

core/src/components/note/test/a11y/note.e2e.ts-snapshots/note-wrapping-label-scale-ios-ltr-Mobile-Chrome-linux.png
amandaejohnston: I would revert the two tests to keep the note slotted, since that's the pattern devs will actually be using.

liamdebeasi: All set https://github.com/ionic-team/ionic-framework/pull/28773/commits/ef4fa21d872fd19630e80cd25382a789b1547d5a https://github.com/ionic-team/ionic-framework/pull/28773/commits/52520e5c26eb091172c0e6ece871c41ad818d5e9

amandaejohnston: Looks like the issue Brandy mentioned in https://github.com/ionic-team/ionic-framework/pull/28773#discussion_r1455885823 is still present; some of the notes have their heights massively expanded with this change. Is that supposed to happen? 😆

liamdebeasi: That's why I suggested having the items in `ion-label` so they can wrap and why I mentioned that the test isn't a great example of best practices. The "Really really long note" text is in the start slot, and the default slot is collapsing to account for the width of the start slot. The text inside of the `ion-label` is wrapping to the next line, so you essentially have 1 character per line (but the width is so small so you don't see it).

image

The orange/yellow is the `ion-label`.

amandaejohnston: Ahh, I see. In that case, should we update the test to avoid these long notes in the first place? Wouldn't be the first time we've adjusted tests to match best practices, and this one is particularly sticky since it leads to unnecessarily massive screenshots.

liamdebeasi: Updated in https://github.com/ionic-team/ionic-framework/pull/28773/commits/d9e034cca9e7fc7b6cba0597e7bf5c7f23759ec4. Mind taking another look?

amandaejohnston: Muuuch better, thank you!!

amandaejohnston (approved):

PR #28783, opened 2024-01-04
chore: remove matchMedia check

liamdebeasi:
Issue number: Internal

---------

As part of FW-2832 the team has an initiative to remove much of the `any` usage in favor of stronger types. This will make modifications to this codebase safer as we will have access to proper type checking.

## What is the new behavior?

- Removed the `any` case from window.matchMedia. This causes the linter to fail because `window.matchMedia` is always defined. I removed the check altogether. `matchMedia` is supported on all browsers that Ionic 7 supports. This should not cause any changes for Ionic developers, but I want to ship this a) in a separate patch and b) in a minor release to de-risk just in case there's an edge case I am not considering

- [ ] Yes
- [x] No

### DIFF
--- a/core/src/components/split-pane/split-pane.tsx
+++ b/core/src/components/split-pane/split-pane.tsx
@@ -118,6 +118,7 @@ export class SplitPane implements ComponentInterface {
     };
 
     const mediaList = window.matchMedia(mediaQuery);
+    // TODO FW-5869
     (mediaList as any).addListener(callback as any);
     this.rmL = () => (mediaList as any).removeListener(callback as any);
     this.visible = mediaList.matches;
--- a/core/src/components/split-pane/split-pane.tsx
+++ b/core/src/components/split-pane/split-pane.tsx
@@ -112,17 +112,15 @@ export class SplitPane implements ComponentInterface {
       return;
     }
 
-    if ((window as any).matchMedia) {
-      // Listen on media query
-      const callback = (q: MediaQueryList) => {
-        this.visible = q.matches;
-      };
+    // Listen on media query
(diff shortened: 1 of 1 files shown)


core/src/components/split-pane/split-pane.tsx
```diff
@@ -112,17 +112,15 @@ export class SplitPane implements ComponentInterface {
...
-      this.rmL = () => (mediaList as any).removeListener(callback as any);
-      this.visible = mediaList.matches;
-    }
+    // Listen on media query
+    const callback = (q: MediaQueryList) => {
+      this.visible = q.matches;
+    };
+
+    const mediaList = window.matchMedia(mediaQuery);
+    (mediaList as any).addListener(callback as any);
```
sean-perkins, line 121: Unrelated to your changes, but `addListener` and `removeListener` are [deprecated APIs](https://developer.mozilla.org/en-US/docs/Web/API/MediaQueryList/addListener). Thoughts on tracking a technical debt to resolve those or just updating them?

I believe it would be to the effect of:
```ts
const callback = (q: MediaQueryListEvent) => {
  this.visible = q.matches;
};

const mediaList = window.matchMedia(mediaQuery);
mediaList.addEventListener('change', callback);
this.rmL = () => mediaList.removeEventListener('change', callback);
```

liamdebeasi: Good idea. I don't have time at the moment to test and verify there are no behavior differences (there shouldn't be, but good to check just in case). I am tracking this in a ticket now.

sean-perkins (approved):

PR #28795, opened 2024-01-08
test: always use official dark mode theme

liamdebeasi:
Issue number: N/A

---------

Our theme tests currently consume an internal dark theme. This was needed to add light/dark theme tests prior to us shipping separate stylesheets to developers. However, now that these stylesheets have been created we should always be testing with them so we can catch any bugs in the dark mode stylesheet.

## What is the new behavior?

- Remove the custom dark.css file and removed references to it
- Added a TODO to do the same thing for the light theme once FW-5862 is done.

- [ ] Yes
- [x] No


core/src/themes/test/css-variables/index.html
```diff
@@ -983,12 +983,18 @@ <h2>Street Fighter II</h2>
 
         // The default and dark themes are official Ionic
         // themes so they are located outside of this test
-        if (theme == 'default' || theme == 'dark') {
+        let cssFile = `${themeFilesDir}/${theme}.css`;
+
+        // TODO FW-5862 update this to not
```
amandaejohnston, line 988: Noticed that the tech debt ticket doesn't mention this, only the `colors` test 👀

amandaejohnston (approved):

core/src/themes/test/css-variables/index.html
liamdebeasi: Good catch. Fixed!

PR #28801, opened 2024-01-09
fix(menu): menu can be encapsulated in a component

### LINKED ISSUE #16304: bug: wrapped menu does not work with content property
**Ionic Info**

```
Ionic:

   ionic (Ionic CLI)             : 4.3.1 (/usr/local/lib/node_modules/ionic)
   Ionic Framework               : @ionic/angular 4.0.0-beta.15
   @angular-devkit/build-angular : 0.8.7
   @angular-devkit/schematics    : 0.8.7
   @angular/cli                  : 6.2.7
   @ionic/angular-toolkit        : 1.1.0

System:

   NodeJS : v8.11.3 (/usr/local/bin/node)
   npm    : 6.4.1
   OS     : macOS
```

**Describe the Bug**

When putting the menu inside a custom component, (app-menu) to make the app component less busy, I get this result : 

https://ibb.co/kUDFOV

But normally, I have to have the initial result : 

https://ibb.co/cFcLqA

**Steps to Reproduce**

1 - Generate a new app with menu
2 - Put the ion-menu in the app-menu
`
<ion-menu contentId="content1" >
  <ion-header>
    <ion-toolbar>
      <ion-title>Menu</ion-title>
    </ion-toolbar>
  </ion-header>
  <ion-content>
    <ion-list>
      <ion-menu-toggle auto-hide="false" *ngFor="let p of appPages">
        <ion-item [routerDirection]="'root'" [routerLink]="[p.url]">
          <ion-icon slot="start" [name]="p.icon"></ion-icon>
          <ion-label>
            {{p.title}}
          </ion-label>
        </ion-item>
      </ion-menu-toggle>
    </ion-list>
  </ion-content>
</ion-menu>
`
3 - Put the following in app.component:
`
<ion-app>
  <ion-split-pane>
  <app-layout-menu contentId="content1" [content1]="content"></app-layout-menu>
    <ion-router-outlet id="content1" main [swipeBackEnabled]="true"></ion-router-outlet>
  </ion-split-pane>
</ion-app>ion
`

Thank you for your answers !

### LINKED ISSUE #20681: bug: SplitPane not working when Menu is wrapped in a Component
# Bug Report
When wrappen the  menu inside a SpliePane in it's own component (e.g. `<my-menu>`) the SplitPane is not working as expected (2 Panes are shown)

**Ionic version:**

[x] **5.x**
latest

**Current behavior:**

When wrapping the menu inside a SplitPane in it's own component (e.g. <my-menu>)
2 Panes are shwon

**Expected behavior:**

The SplitPane should work as if i would put the `<ion-menu>` right underneath the SplitPane

**Steps to reproduce:**

**Related code:**
not working:
```
<ion-split-pane when="xl" contentId="main">
  <my-menu></my-menu>

  <ion-router-outlet main id="main"></ion-router-outlet>
</ion-split-pane>
```

working:
```
<ion-split-pane when="xl" contentId="main">
  <ion-menu>...</ion-menu>

  <ion-router-outlet main id="main"></ion-router-outlet>
</ion-split-pane>
```

```
insert short code snippets here
```

**Other information:**

**Ionic info:**

```
insert the output from ionic info here
```

liamdebeasi:
Issue number: resolves #16304, resolves #20681

---------

Split Pane assumes that Menu is a child of the Split Pane. This means that CSS selectors and JS queries only check for children instead of descendants. When a Menu is encapsulated in a component that component can itself show up as an element in the DOM depending on the environment. For example, both Angular and Web components show up as elements in the DOM. This causes the Menu to not work because it is no longer a child of the Split Pane.

## What is the new behavior?

- Menu can now be used as a descendant of Split Pane. The following changes make this possible:
1. When the menu is loaded, it queries the ancestor Split Pane. If one is found, it sets `isPaneVisible` depending on if the split pane is visible or not.
2. Any changes to the split pane visibility state after the menu is loaded will be handled through the `ionSplitPaneVisible` event. 
3. A menu is now considered to be a pane if it is inside of a split pane
4. Related CSS no longer uses `::slotted` which only targets children. The CSS was moved into the menu stylesheets. I consider the coupling here to be valuable as menu and split pane are often used together. We also already have style coupling between the two components, so this isn't anything new.

- [ ] Yes
- [x] No

There are no known changes to the public API or public behavior. However, there are two risks here:

1. There is an unknown risk into how this change impacts nested split panes. This isn't an explicitly supported feature, but it technically works in Ionic now. We don't know if anyone is actually implementing this pattern. We'd like to evaluate use cases for nested split panes submitted by the community before we try to account for this functionality in menu and split pane.
5. There may be some specificity changes due to the new CSS. CSS was moved from split pane to menu so it could work with encapsulated components:

`:host(.split-pane-visible) ::slotted(.split-pane-side)` has a specificity of 0-1-1

`:host(.menu-page-visible.menu-pane-side)` has a specificity of 0-1-0

There shouldn't be any changes needed to developer code since the selectors are in the Shadow DOM and developer styles are in the Light DOM. However, we aren't able to anticipate every edge case so we'd like to place this in a major release just to be safe.

## Other information

Dev build: `7.6.2-dev.11704922151.1fd3369f`


core/src/components/menu/menu.tsx
```diff
@@ -288,23 +312,14 @@ export class Menu implements ComponentInterface, MenuI {
   }
 
   @Listen('ionSplitPaneVisible', { target: 'body' })
-  onSplitPaneChanged(ev: CustomEvent) {
-    const { target } = ev;
-    const closestSplitPane = this.el.closest('ion-split-pane');
+  async onSplitPaneChanged(ev: CustomEvent<{ visible: boolean }>) {
```
amandaejohnston, line 315: Why was this made async? It doesn't `await` anything.

liamdebeasi: Good catch! That was left over from some older attempts at fixing the issue. Removed in https://github.com/ionic-team/ionic-framework/pull/28801/commits/166df6c3922fba3f5d3d6c28737c8cfebbf6285d

core/src/components/menu/menu.ios.scss
```diff
@@ -19,3 +19,20 @@
...
+  min-width: var(--side-min-width);
+  max-width: var(--side-max-width);
+}
+
+:host(.menu-pane-visible.split-pane-side[side=end]) {
+  @include border(0, 0, 0, var(--border));
+
+  min-width: var(--side-min-width);
+  max-width: var(--side-max-width);
+}
```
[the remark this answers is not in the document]
liamdebeasi: Good idea. Updated in https://github.com/ionic-team/ionic-framework/pull/28801/commits/60662f450813ce95bdfeca29b2e2ccda4e9621b0

amandaejohnston (approved):

PR #28804, opened 2024-01-10
feat(toast): high contrast support for button text

sean-perkins:
Issue number: Internal

---------

With the high contrast themes, the `ion-toast` buttons do not have sufficient color contrast to satisfy Level AAA. 

## What is the new behavior?

- `ion-toast` buttons satisfy Level AAA color contrast with the high contrast themes

- [ ] Yes
- [x] No

### Testing Locally

In `core/src/components/toast/test/buttons/index.html` add the stylesheet import either the dark or light high contrast theme:

Dark (high contrast):
```css
<link href="../../../../../css/themes/high-contrast-dark.always.css" rel="stylesheet" />
```

Light (hight contrast):
```css
<link href="../../../../../css/themes/high-contrast-light.always.css" rel="stylesheet" />
```


core/src/components/toast/test/a11y/toast.e2e.ts
```diff
@@ -230,3 +230,48 @@ configs({ directions: ['ltr'] }).forEach(({ title, screenshot, config }) => {
...
+            { text: 'Cancel', role: 'cancel' },
+            { text: 'OK' }
+          ];
+        </script>
+      `,
+          config
+        );
+      });
+
+      test('should not have visual regressions', async ({ page }) => {
```
amandaejohnston, line 256: Do we need screenshot tests for this? I don't think we added any per-component screenshots for dark mode, since the color updates were handled in dedicated theme tests. (We should definitely keep the Axe test though.)

sean-perkins: `ion-toast` has custom styles for the different high contrast themes that are not covered by the theme tests.

amandaejohnston: Ah, true! Fair enough 👍

amandaejohnston (approved):

brandyscarney (approved):

PR #28809, opened 2024-01-10
fix(datetime): set default background color correctly

mapsandapps:
Issue number: None
---------

In my work for FW-5822, I found that I had unexpected screenshot diffs. `--ion-color-light` was not defined in our test infrastructure, and the white fallback for the datetime background was causing snapshots to have a white background when they should have had a light one.

## What is the new behavior?

- Datetime has the correct background color even when the `--ion-color-light` token is not defined

- [ ] Yes
- [x] No

Introduces style changes if the token is undefined, but it should always be defined in real-world uses.


liamdebeasi (approved):
Thanks for catching and fixing this!

PR #28812, opened 2024-01-10
fix(toggle): set switch icon color correctly

mapsandapps:
Issue number: None

---------

Similar to #28809, in my work for FW-5822, I found that I had unexpected screenshot diffs. `--ion-color-dark` was not defined in our test infrastructure, and this toggle icon was therefore displaying in the wrong color in the tests.

## What is the new behavior?

- Toggle has the correct icon color even when the --ion-color-dark token is not defined

- [ ] Yes
- [x] No

Introduces style changes if the token is undefined, but it should always be defined in real-world uses.


liamdebeasi (approved):
Thank you!

PR #28823, opened 2024-01-12
chore(alert): remove ion-buttons from tests

thetaPC:
Issue number: internal

---------

`ion-button` is being used as a way to open the alerts. It's not being used to test functionality.

When `ion-button` is updated then the screenshots in the alert tests must be updated. This shouldn't happen when `ion-button` isn't necessary to have in these tests.

## What is the new behavior?

- Replaced `ion-button` with `button`

- [ ] Yes
- [x] No

## Other information

N/A


core/src/components/alert/test/translucent/index.html
```diff
@@ -27,12 +27,10 @@
...
-        <ion-button expand="block" onclick="presentAlert()">Alert</ion-button>
-        <ion-button expand="block" color="secondary" onclick="presentAlertLongMessage()">Alert Long Message</ion-button>
-        <ion-button expand="block" color="danger" onclick="presentAlertMultipleButtons()"
-          >Multiple Buttons (>2)</ion-button
-        >
-        <ion-button expand="block" color="light" onclick="presentAlertNoMessage()">Alert No Message</ion-button>
+        <button expand="block" class="blue" onclick="presentAlert()">Alert</button>
+        <button expand="block" class="light-blue" onclick="presentAlertLongMessage()">Alert Long Message</button>
+        <button expand="block" class="red" onclick="presentAlertMultipleButtons()">Multiple Buttons (>2)</button>
+        <button expand="block" class="light" onclick="presentAlertNoMessage()">Alert No Message</button>
```
liamdebeasi, line 33: `expand="block"` is an `ion-button` API, so it should be removed if we are going to use `button` elements. (Same for all other instances of this)

liamdebeasi (changes requested):
Overall looks good! One comment on the `expand` usage since we don't need it anymore.

liamdebeasi (approved):
Thank you!

PR #28828, opened 2024-01-15
feat(item-sliding): modern animation to reveal options

sean-perkins:
Issue number: Internal

---------

Currently when swiping to reveal options, the options are translated along with the swipe distance to reveal the last to first option. 

https://github.com/ionic-team/ionic-framework/assets/13732623/084cefb8-9ac0-4f37-a4be-d8cddc7ee847

This is different than the behavior we observe in iOS 17 and Material Design, where the options are all revealed at a consistent interval with the swipe distance. 

### Item Sliding

- There is a new property called `animationType` available on `ion-item-sliding` to customize the animation behavior of the sliding item options.
  - `animationType="legacy"` will maintain the previous animation behavior that developers expect in their applications in Ionic v7.
  - `animationType="modern"` will opt-in to the modern animation appearance as seen in iOS 17. This animation does leverage the Web Animation API and developers should consult caniuse if targeting older devices. If Web Animations API is not supported on the device, the sliding item will fallback to the legacy implementation and log a warning to the console.

|LTR|RTL|
|---|---|
|<video src="https://github.com/ionic-team/ionic-framework/assets/13732623/57b1b528-0c52-40f8-9f60-329ef793b1da"></video>|<video src="https://github.com/ionic-team/ionic-framework/assets/13732623/bde30ad9-7502-4551-a442-c89a91b4e7a4"></video>|

### Ionic Animations

To support the animation behavior of this feature, the Ionic Animations utility was updated to add support for [`commitStyles`](https://developer.mozilla.org/en-US/docs/Web/API/Animation/commitStyles). This feature reflects the state of styles from the animation to the element node. This enables parsing the computed transformations to leverage in the animation. The styles assigned by the Web Animations API take priority over the reflected state. 

The constants for detecting if the environment supports the Web Animations API have been exported.

- [ ] Yes
- [x] No

### Verifying this PR

1. Test on a physical device or mobile emulator.
2. Verify that swiping to reveal options represents the preview above, where option are all revealed gradually with the swipe distance.
3. Verify the same behavior in RTL mode.
4. Verify the release and cancel behavior with quickly swiping to show/collapse items as well as releasing the swipe gesture prior to the threshold that shows/collapses the item. 

### Deep dive into the implementation

This deep dive focuses on the item options animation within this PR. It assumes you are familiar to how the existing item sliding animation behaves to translate the item to reveal options. 

**1. Prepare item options for animation**

 In order to simplify the calculations of the animation, we relocate each item option off the visible viewport to be on the edge of the screen. For example in LTR text direction, the items on the end of the container will be located on the right edge of the screen, off the viewport. 

<figure>
  
  <figcaption>This image demonstrates the transformation on the x-axis to move the end options in LTR text decoration.</figcaption>
</figure>

To determine the distance to move each option, we need to define some initial values and their purposes:
- `o`: This is an incrementing value that accounts for the _total_ width of all previous processed item options. For example when it processes the first item option, this resolves `0`, the next item option this would equal the width of the previous item option. On the next item option this would equal the width of the two previous item options and so on.
- `tW`: The total width of all the item options. 
- `oW`: The width of the current item option.

With this information we can equate the distance as follows:

_Side: Start_
- Left-to-right text decoration: `distance = -1 * (oW + o)`
- Right-to-left text decoration: `distance = -1 * (tw - o)`

_Side: End_
- Left-to-right text decoration: `distance = tW - o`
- Right-to-left text decoration: `distance = o + oW`

**2. Animate each option during swipe gesture**

As the user swipes to reveal options, there is an open amount of space to reveal all the options. We need to give each option roughly the same amount of visible viewport space.

<figure>
  
  <figcaption>This image demonstrates the transformation on the x-axis to move each option into the visible open space from the swipe gesture. This graphic is representative of end item options in LTR. The stacking order of the elements is important to note, so that items option are pulled from underneath each other.</figcaption>
</figure>

We are able to defer the complexity of computing the transform value as the user is dragging the item to the Web Animations API by transforming between two fixed points: the position off the viewport and the original render position, and animate based on the progress of the open amount divided by the options total width.

As a result of using a transform between two points, the amount of space that each option fills in the available space will be dependent on the width of the option. The animation effect scales evenly with options of equal width.


core/src/components/item-sliding/test/basic/item-sliding.e2e.ts-snapshots/item-sliding-end-ios-ltr-Mobile-Chrome-linux.png
sean-perkins, line 1: From what I can tell, all the diffs that are not from the `modern-item-sliding-*` screenshots, are minor browser differences between when the screenshots were originally captured and now. 

Commits prior to the diff that adds the screenshots are passing CI (e.g.: https://github.com/ionic-team/ionic-framework/pull/28828/commits/79f188fdb743ecc7f26253a8b5a80c669b2fdfa3) and the contents of this PR should not impact the height changes or the text alignment shifts we observe in the diffs. 

I don't know why playwright is detecting them now. It is possible that when a shard has screenshot diffs (e.g. the additions of this feature) that other images are updated, even if they fall below the tolerance to constitute a "failing diff".

core/src/components/item-sliding/test/basic/index.html
thetaPC, line 1: https://github.com/ionic-team/ionic-framework/assets/13530427/b42404a9-04b4-47d9-b3d2-431ea966f365

When dragging to the right, I see that BTN 1 smoothly expands to be the only button visible. When I drag back to the left, BTN 1 snaps back to the same width as the others. Is this the expected behavior? I would assume that it would smoothly decrease width.

thetaPC:
Dev build: 7.6.2-dev.11705694334.11b34c62

core/src/components/item-sliding/item-sliding.tsx
thetaPC, line 1: I'm using a phone simulator. Opening transition follows the expected outcome. However, the closing transition does not. Notice in the video that the buttons are closing one at a time. App was built with React.


https://github.com/ionic-team/ionic-framework/assets/13530427/338ee483-1f60-493e-a683-1ede0409e1c2

core/src/components/item-sliding/item-sliding.tsx
```diff
@@ -462,15 +503,62 @@ export class ItemSliding implements ComponentInterface {
 
       openSlidingItem = undefined;
       style.transform = '';
+
+      if (animationType === 'modern') {
```
liamdebeasi, line 507: Can we extract these keywords to consts or a type or something? When we eventually bike shed this will make potentially changing the names easier.

core/src/utils/animation/animation.ts
```diff
@@ -38,6 +38,15 @@ type AnimationOnStopCallback = AnimationOnFinishCallback;
...
 type AnimationReadWriteCallback = () => void;
 
+export const supportsAnimationEffect =
+  typeof (AnimationEffect as any) === 'function' ||
+  (win !== undefined && typeof (win as any).AnimationEffect === 'function');
+
+export const supportsWebAnimations =
+  typeof (Element as any) === 'function' &&
+  typeof (Element as any).prototype!.animate === 'function' &&
+  supportsAnimationEffect;
```
liamdebeasi, line 48: How does this differ from the consts of the same name inside of `createAnimation`?

core/src/components/item-sliding/item-sliding.tsx
```diff
@@ -462,15 +503,62 @@ export class ItemSliding implements ComponentInterface {
...
+    const progress = Math.abs(openAmount / optsWidth);
+
+    options.forEach((option, index) => {
+      let isFinal = false;
+      let zIndex: number | undefined;
+      let viewportOffset = 0;
+
+      const isReset = openAmount === 0;
+
+      if (side === 'start') {
```
liamdebeasi, line 546: When dragging to close the options stop collapsing:


https://github.com/ionic-team/ionic-framework/assets/2721089/ca18f8b3-a25f-4115-8d5f-e380a9587b38

Tested in Safari for macOS.

core/src/components/item-sliding/item-sliding.tsx
```diff
@@ -462,15 +503,62 @@ export class ItemSliding implements ComponentInterface {
...
+    let optionWidthOffset = 0;
+
+    const side = this.state === SlidingState.Start ? 'start' : 'end';
+
+    const options = Array.from(this.getOptions(side)?.querySelectorAll('ion-item-option') || []);
+    const optsWidth = this.getOptionsWidth(side);
+
+    const progress = Math.abs(openAmount / optsWidth);
+
+    options.forEach((option, index) => {
```
liamdebeasi, line 539: The calculation doesn't seem correct when quickly dragging:


https://github.com/ionic-team/ionic-framework/assets/2721089/972bb09e-139b-4a45-96a8-e12ee5fd4d66

core/src/components/item-sliding/item-sliding.tsx
```diff
@@ -1,13 +1,18 @@
...
+import { supportsWebAnimations } from '@utils/animation/animation';
 import { findClosestIonContent, disableContentScrollY, resetContentScrollY } from '@utils/content';
 import { isEndSide } from '@utils/helpers';
+import { printIonWarning } from '@utils/logging';
+import { isRTL } from '@utils/rtl';
 import { watchForOptions } from '@utils/watch-options';
 
 import { getIonMode } from '../../global/ionic-global';
-import type { Gesture, GestureDetail } from '../../interface';
+import { type Gesture, type GestureDetail } from '../../interface';
```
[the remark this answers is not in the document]
sean-perkins: This change was an automatic code change through VSCode's typescript language service. In this scenario they accomplish the same end result. The compiler will fully erase the references so they are not preset at runtime. 

The advantage of the `{ type Name }` vs `type { }` is you can have a single import declaration from a single module, even if you have a mix of type and value referenced imports. 

I _believe_ this is a newer typescript change.

PR #28830, opened 2024-01-15
fix(list): remove uneeded border radius from items in inset list

amandaejohnston:
Issue number: N/A

---------

In MD mode, items in an inset list currently receive their own border radius, separate from the border radius also applied to the inset list itself. The original intent was likely to ensure that the corners of the items don't spill out of the list. However, the item's border radius is no longer needed, for two reasons:
1. MD has since added top/bottom padding to their lists ([spec](https://m2.material.io/components/lists#specs)), which we've mirrored. This means the borders no longer line up anyway.
2. Even if a developer removes the list's vertical padding (we set it on the host [here](https://github.com/ionic-team/ionic-framework/blob/535b8ed72486a1e94125c72b12d5b447b630f8c4/core/src/components/list/list.md.scss#L9), so it's easy to overwrite), inset lists currently have `overflow: hidden` set [here](https://github.com/ionic-team/ionic-framework/blob/535b8ed72486a1e94125c72b12d5b447b630f8c4/core/src/components/list/list.scss#L20), so the list's border radius still applies correctly even with the items' border radius removed.

## What is the new behavior?

Item border radius in inset lists removed.

When testing locally, you may find it useful to increase the MD border radius value [here](https://github.com/ionic-team/ionic-framework/blob/535b8ed72486a1e94125c72b12d5b447b630f8c4/core/src/components/list/list.md.vars.scss#L47) so it's easier to see.

- [ ] Yes
- [x] No

### DIFF
--- a/core/src/components/list/list.md.scss
+++ b/core/src/components/list/list.md.scss
@@ -24,24 +24,8 @@
 }
 
 /**
- * These selectors ensure the first item in the list
- * has the correct radius.
- * We need to consider the following scenarios:
- 1. The first item in a list as long as it is not the only item.
- 2. The item in the first item-sliding in a list.
- * Note that we do not select "ion-item-sliding ion-item:first-of-type"
- * because that will cause the borders to disappear on
- * items in an item-sliding when the item is the first
- * element in the item-sliding container.
- */
-.list-md.list-inset ion-item:not(:only-of-type):first-of-type,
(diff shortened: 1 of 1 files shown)


liamdebeasi (approved):
Looks good! I would consider this a fix though since it's removing buggy behavior.

PR #28839, opened 2024-01-16
fix(select): click handlers on slotted content fire

### LINKED ISSUE #28818: bug: v7 - React - onClick doesn't work in ionButton inside ionSelect
### Prerequisites

### Ionic Framework Version

v7.x

### Current Behavior

The new functionality insert ionButton inside ionSelect work noramly to visualize but when is insert a function like the onClick dosen't work, the event isn't fired, have othen events that work normaly like onTouch(Start or End).

### Expected Behavior

Just when insert the onClick and click the event be fire.

### Steps to Reproduce

```
<IonSelect
        label='Level 1'
        placeholder="Select Level 1"
        labelPlacement="floating"
        fill="solid"
        value={'test'}
        onIonChange={e=>{}}
>
       <IonSelectOption value="test">Opt</IonSelectOption>
       <IonButton fill="clear" slot="end" onClick={e=>{alert('fire')}}>
          <IonIcon slot="icon-only" color='dark' icon={addCircle}></IonIcon>
       </IonButton>
</IonSelect>
```

### Code Reproduction URL

_No response_

### Ionic Info

       Require stack:
       - C:\Users\Arcano\AppData\Roaming\npm\node_modules\@ionic\cli\lib\project\index.js
       - C:\Users\Arcano\AppData\Roaming\npm\node_modules\@ionic\cli\lib\index.js
       - C:\Users\Arcano\AppData\Roaming\npm\node_modules\@ionic\cli\index.js
       - C:\Users\Arcano\AppData\Roaming\npm\node_modules\@ionic\cli\bin\ionic

Ionic:

   Ionic CLI       : 7.1.1 (C:\Users\Arcano\AppData\Roaming\npm\node_modules\@ionic\cli)
   Ionic Framework : @ionic/react 7.6.4

Capacitor:

   Capacitor CLI      : 5.4.1
   @capacitor/android : 5.4.1
   @capacitor/core    : 5.4.1

Utility:

   native-run  : 1.7.3

System:

   NodeJS : v18.16.0 (C:\Program Files\nodejs\node.exe)
   npm    : 9.5.1
   OS     : Windows 10

### Additional Information

_No response_

liamdebeasi:
Issue number: resolves #28818

---------

Select has logic in place to prevent the overlay from opening when clicking the slotted content. As part of this, we need to deal with the `<label>` element dispatching another click that then bubbles up and causes the overlay to open when clicking the slotted content. We currently deal with this by calling `preventDefault` which prevents this extra event from being dispatched only when clicking the slotted content.

We also call `stopPropagation`. This code is not necessary, and in most cases should not have any adverse effects on developer applications. However, this does impact React applications due to how React handles events. When a developer places `onClick` on a slotted element, a native "click" listener is not immediately applied to that element. Instead, React adds a native "click" listener to the root element of an application. When that listener's callback fires, React will dispatch a synthetic click event on the slotted element. Since we are calling `stopPropagation`, the click event never bubbles up to this root node's listener. As a result, the synthetic event is never dispatched on the slotted element, and the developed `onClick` callback never fires.

## What is the new behavior?

- Select no longer prevents events from bubbling. The existing `event.preventDefault` is sufficient to prevent the label from dispatching another click (thereby causing the select overlay to open) when clicking the slotted content.

- [ ] Yes
- [x] No

## Other information

Dev build: `7.6.5-dev.11705430252.1023daf3`

### DIFF
--- a/core/src/components/select/select.tsx
+++ b/core/src/components/select/select.tsx
@@ -772,12 +772,26 @@ export class Select implements ComponentInterface {
       this.open(ev);
     } else {
       /**
-       * Prevent clicks to the start/end slots from opening the select.
-       * We ensure the target isn't this element in case the select is slotted
-       * in, for example, an item. This would prevent the select from ever
-       * being opened since the element itself has slot="start"/"end".
+       * Clicking a slotted element also causes a click
+       * on the <label> element (since it wraps the slots).
+       * Clicking <label> dispatches another click event on
+       * the native form control that then bubbles up to this
+       * listener. This additional event targets the host
+       * element, so the select overlay is opened.
+       *
+       * When the slotted elements are clicked (and therefore
+       * the ancestor <label> element) we want to prevent the label
+       * from dispatching another click event.
+       *
+       * Do not call stopPropagation() because this will cause
+       * click handlers on the slotted elements to never fire in React.
+       * When developers do onClick in React a native "click" listener
+       * is added on the root element, not the slotted element. When that
+       * native click listener fires, React then dispatches the synthetic
+       * click event on the slotted element. However, if stopPropagation
+       * is called then the native click event will never bubble up
+       * to the root element.
        */
-      ev.stopPropagation();
       ev.preventDefault();
     }
   };
--- a/core/src/components/select/select.tsx
+++ b/core/src/components/select/select.tsx
@@ -772,6 +772,11 @@ export class Select implements ComponentInterface {
       this.open(ev);
     } else {
       /**
+       * Prevent clicks to the start/end slots from opening the select.
+       * We ensure the target isn't this element in case the select is slotted
+       * in, for example, an item. This would prevent the select from ever
+       * being opened since the element itself has slot="start"/"end".
+       *
        * Clicking a slotted element also causes a click
        * on the <label> element (since it wraps the slots).
        * Clicking <label> dispatches another click event on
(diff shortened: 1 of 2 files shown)


core/src/components/select/select.tsx
```diff
@@ -772,12 +772,26 @@ export class Select implements ComponentInterface {
       this.open(ev);
     } else {
       /**
-       * Prevent clicks to the start/end slots from opening the select.
-       * We ensure the target isn't this element in case the select is slotted
-       * in, for example, an item. This would prevent the select from ever
-       * being opened since the element itself has slot="start"/"end".
```
amandaejohnston, line 778: Can we ensure this context is preserved somewhere, whether in this specific comment or elsewhere (such as near the `target === this.el` check above)?

core/src/components/select/test/select.spec.tsx
```diff
@@ -69,3 +69,31 @@ describe('ion-select', () => {
     expect(slotEl).toBe(null);
   });
 });
+
+describe('select: slot interactivity', () => {
```
amandaejohnston, line 73: Should we have a test in the React app too, since the problem was specific to that?

liamdebeasi: The bug most easily reproduces in React, but it's not specific to the React environment which is why I decided to make this a test in core instead of a test in React.

core/src/components/select/select.tsx
liamdebeasi: re-added in https://github.com/ionic-team/ionic-framework/pull/28839/commits/03bd74dd9909f89b1fa8e76e7fcc4743bf1ffa2a

amandaejohnston (approved):

PR #28847, opened 2024-01-17
fix(datetime): do not animate to new value when multiple values in different months are set

### LINKED ISSUE #28602: bug: Calendar month changes when selecting date on Multiple date mode
### Prerequisites

### Ionic Framework Version

v7.x

### Current Behavior

When using the Calendar on multiple date mode, if the initial month is November, selecting a date from a different month navigates the view back to the initial month. 

### Expected Behavior

The calendar month should not navigate automatically when selecting dates from a month that different from the default one.

### Steps to Reproduce

1. Open page with multiple date calendar
2. Navigate to a different month
3. select a date

### Code Reproduction URL

https://ionicframework.com/docs/api/datetime#multiple-date-selection

### Ionic Info

```
Ionic:

   Ionic CLI                     : 6.20.9 (/Users/jonathan/.nvm/versions/node/v16.14.0/lib/node_modules/@ionic/cli)
   @angular-devkit/build-angular : 16.2.10
   @angular-devkit/schematics    : 16.2.10
   @angular/cli                  : 16.2.10

Capacitor:

   Capacitor CLI      : 5.5.1
   @capacitor/android : 5.5.1
   @capacitor/core    : 5.5.1
   @capacitor/ios     : 5.5.1

Utility:

   cordova-res : 0.15.4
   native-run  : 2.0.0

System:

   NodeJS : v16.14.0 (/Users/jonathan/.nvm/versions/node/v16.14.0/bin/node)
   npm    : 8.3.1
   OS     : macOS Unknown
```

### Additional Information

_No response_

amandaejohnston:
Issue number: resolves #28602

---------

We animate to the new date when the value is changed, but we do not account for multiple selection. This behavior is valuable for when the value is set asynchronously on a different month than what it shown.

However, this is confusing when there are multiple dates selected in different months, because we do not reasonably know which date to animate to.

## What is the new behavior?

Datetime no longer animates to the new value if more than one date is selected, and the selected dates aren't all in the same month.

An alternative strategy would be to always animate unless one of the selected dates is in the month currently being viewed. However, this would still mean guessing at which value the user wants to see, which we are trying to avoid.

Based on this PR: https://github.com/ionic-team/ionic-framework/pull/28605

- [ ] Yes
- [x] No

## Other information

Co-authored-by: Liam DeBeasi <liamdebeasi@users.noreply.github.com>

### DIFF
--- a/core/src/components/datetime/test/multiple/datetime.e2e.ts
+++ b/core/src/components/datetime/test/multiple/datetime.e2e.ts
@@ -159,6 +159,10 @@ configs({ modes: ['md'], directions: ['ltr'] }).forEach(({ title, config }) => {
     });
 
     test('should scroll to new month when value is updated with multiple dates in the same month', async ({ page }) => {
+    test.info().annotations.push({
+      type: 'issue',
+      description: 'https://github.com/ionic-team/ionic-framework/issues/28602',
+    });
       const datetime = await datetimeFixture.goto(config, MULTIPLE_DATES_SEPARATE_MONTHS);
       await datetime.evaluate((el: HTMLIonDatetimeElement, dates: string[]) => {
         el.value = dates;
--- a/core/src/components/datetime/test/multiple/datetime.e2e.ts
+++ b/core/src/components/datetime/test/multiple/datetime.e2e.ts
@@ -158,6 +158,18 @@ configs({ modes: ['md'], directions: ['ltr'] }).forEach(({ title, config }) => {
       });
     });
 
+    test('should scroll to new month when value is updated with multiple dates in the same month', async ({ page }) => {
+      const datetime = await datetimeFixture.goto(config, MULTIPLE_DATES_SEPARATE_MONTHS);
+      await datetime.evaluate((el: HTMLIonDatetimeElement, dates: string[]) => {
+        el.value = dates;
+      }, MULTIPLE_DATES);
+
+      await page.waitForChanges();
+
+      const monthYear = datetime.locator('.calendar-month-year');
+      await expect(monthYear).toHaveText(/June 2022/);
+    });
+
     test('should not scroll to new month when value is updated with dates in different months', async ({ page }) => {
       const datetime = await datetimeFixture.goto(config, MULTIPLE_DATES);
       await datetime.evaluate((el: HTMLIonDatetimeElement, dates: string[]) => {
--- a/core/src/components/datetime/test/multiple/datetime.e2e.ts
+++ b/core/src/components/datetime/test/multiple/datetime.e2e.ts
@@ -5,7 +5,7 @@ import { configs, test } from '@utils/test/playwright';
 
 const SINGLE_DATE = '2022-06-01';
 const MULTIPLE_DATES = ['2022-06-01', '2022-06-02', '2022-06-03'];
-const MULTIPLE_DATES_SEPARATE_MONTHS = ['2022-04-01', '2022-05-01', '2022-06-01'];
+const MULTIPLE_DATES_SEPARATE_MONTHS = ['2022-03-01', '2022-04-01', '2022-05-01'];
 
 interface DatetimeMultipleConfig {
   multiple?: boolean;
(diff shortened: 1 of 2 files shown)


core/src/components/datetime/test/multiple/datetime.e2e.ts
```diff
@@ -5,7 +5,7 @@ import { configs, test } from '@utils/test/playwright';
 
 const SINGLE_DATE = '2022-06-01';
 const MULTIPLE_DATES = ['2022-06-01', '2022-06-02', '2022-06-03'];
-const MULTIPLE_DATES_SEPARATE_MONTHS = ['2022-04-01', '2022-05-01', '2022-06-01'];
+const MULTIPLE_DATES_SEPARATE_MONTHS = ['2022-03-01', '2022-04-01', '2022-05-01'];
```
amandaejohnston, line 8: This was changed to avoid any of these dates being in the same month as the `MULTIPLE_DATES` set above.

core/src/components/datetime/test/multiple/datetime.e2e.ts
```diff
@@ -158,10 +158,16 @@ configs({ modes: ['md'], directions: ['ltr'] }).forEach(({ title, config }) => {
       });
     });
 
-    test('multiple default values across months should display at least one value', async () => {
-      const datetime = await datetimeFixture.goto(config, MULTIPLE_DATES_SEPARATE_MONTHS);
+    test('should not scroll to new month when value is updated with dates in different months', async ({ page }) => {
```
amandaejohnston, line 161: I didn't add a test for ensuring it *does* scroll when all dates are in the same month because that's already covered by [this screenshot test](https://github.com/ionic-team/ionic-framework/blob/09bc77eb3dcca2b28f35b714a1cef64340917365/core/src/components/datetime/test/multiple/datetime.e2e.ts#L86).

liamdebeasi, line 161: Why are we removing the old test in favor of this new one?

liamdebeasi (commented):
Fix works well, just one question on the test.

core/src/components/datetime/test/multiple/datetime.e2e.ts
amandaejohnston: The old test checked whether the datetime scrolled to one of the selected months if multiple values from different months were provided. We're removing precisely that behavior, so the test would fail if kept.

amandaejohnston: Ah, true; I'd assumed loading with the values already set was impossible in pure JS since you have to do it in a separate `<script>` tag, but looking at a React app with the `value` prop initialized, the behaviors are equivalent. I'll add a new test 👀

amandaejohnston: Added in https://github.com/ionic-team/ionic-framework/pull/28847/commits/4369f2adda981eb15f2c9a86f9d6e818812a10fd.

core/src/components/datetime/test/multiple/datetime.e2e.ts
```diff
@@ -158,10 +158,28 @@ configs({ modes: ['md'], directions: ['ltr'] }).forEach(({ title, config }) => {
       });
     });
 
-    test('multiple default values across months should display at least one value', async () => {
+    test('should scroll to new month when value is updated with multiple dates in the same month', async ({ page }) => {
```
liamdebeasi, line 161:
```suggestion
    test('should scroll to new month when value is updated with multiple dates in the same month', async ({ page }) => {
    test.info().annotations.push({
      type: 'issue',
      description: 'https://github.com/ionic-team/ionic-framework/issues/28602',
    });
```

liamdebeasi (approved):
Test was missing an annotation, but otherwise looks good!

core/src/components/datetime/test/multiple/datetime.e2e.ts
```diff
@@ -158,10 +158,28 @@ configs({ modes: ['md'], directions: ['ltr'] }).forEach(({ title, config }) => {
...
+        el.value = dates;
+      }, MULTIPLE_DATES);
+
+      await page.waitForChanges();
+
+      const monthYear = datetime.locator('.calendar-month-year');
+      await expect(monthYear).toHaveText(/June 2022/);
+    });
+
+    test('should not scroll to new month when value is updated with dates in different months', async ({ page }) => {
```
liamdebeasi, line 173:
```suggestion
    test('should not scroll to new month when value is updated with dates in different months', async ({ page }) => {
    test.info().annotations.push({
      type: 'issue',
      description: 'https://github.com/ionic-team/ionic-framework/issues/28602',
    });
```

PR #28850, opened 2024-01-18
fix(overlays): do not return focus if application has already moved focus manually

### LINKED ISSUE #28849: bug: ionic should not re-focus an element when focus already moved on overlay dismiss by developer
### Prerequisites

Allow developer to set custom focus during overlay dismissal.

### Describe the Use Case

In the `handler` of an `ion-action-sheet`  the developer may want to restore focus to a custom element, like a text box.

Ionic overlay should detect whether focus has been set by developer, and **not** restore focus if that is so.

### Describe Alternatives

None.

### Reproduction

https://github.com/aeharding/test-focus

### Additional Information

Not sure if this is considered a feature or a bug 😅

aeharding:
Issue number: resolves #28849

---------

If the developer tries to set focus to a custom element on overlay dismissal, Ionic will always override that focus.

## What is the new behavior?

- If focus is already set by developer during dismissal, then Ionic will not restore focus to previous element

- [ ] Yes
- [x] No

## Other information

In the before video, you can see the text box is focused by developer code when "Mention User" is tapped, which opens the keyboard. Shortly after that, when the bottom sheet fully dismisses, Ionic focuses the button, removing focus from the text box and hiding the keyboard.

In the after, Ionic detects that the developer has already focused the text box and does not change that focus.

|Before|After|
|---|---|
|<video src="https://github.com/ionic-team/ionic-framework/assets/2166114/47d55eff-29af-4019-ac3c-00f9fe722ca7"></video>| <video src="https://github.com/ionic-team/ionic-framework/assets/2166114/508ae466-d037-41eb-b518-92338a122b22"></video>|

### DIFF
--- a/core/src/utils/overlays.ts
+++ b/core/src/utils/overlays.ts
@@ -496,7 +496,7 @@ export const present = async <OverlayPresentOptions>(
    * from returning focus as a result.
    */
   if (overlay.el.tagName !== 'ION-TOAST') {
-    focusPreviousElementOnDismiss(overlay.el);
+    focusPreviousElementOnDismissIfNeeded(overlay.el);
   }
 
   /**
@@ -520,7 +520,7 @@ export const present = async <OverlayPresentOptions>(
  * to where they were before they
  * opened the overlay.
  */
-const focusPreviousElementOnDismiss = async (overlayEl: any) => {
+const focusPreviousElementOnDismissIfNeeded = async (overlayEl: any) => {
   let previousElement = document.activeElement as HTMLElement | null;
   if (!previousElement) {
     return;
@@ -533,7 +533,17 @@ const focusPreviousElementOnDismiss = async (overlayEl: any) => {
   }
 
   await overlayEl.onDidDismiss();
-  previousElement.focus();
+
+  /**
+   * If the user has already removed focus
+   * from the overlay (For example, focusing
+   * a text box after tapping a button in an
+   * action sheet) then don't restore focus
+   * to previous element
+   */
+  if (document.activeElement === null || document.activeElement === document.body) {
+    previousElement.focus();
+  }
 };
 
 export const dismiss = async <OverlayDismissOptions>(
--- a/core/src/utils/overlays.ts
+++ b/core/src/utils/overlays.ts
@@ -496,7 +496,7 @@ export const present = async <OverlayPresentOptions>(
    * from returning focus as a result.
    */
   if (overlay.el.tagName !== 'ION-TOAST') {
-    focusPreviousElementOnDismissIfNeeded(overlay.el);
+    restoreElementFocus(overlay.el);
   }
 
   /**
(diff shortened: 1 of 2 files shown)


core/src/utils/overlays.ts
```diff
@@ -533,7 +533,17 @@ const focusPreviousElementOnDismiss = async (overlayEl: any) => {
...
-  previousElement.focus();
+
+  /**
+   * If the user has already removed focus
+   * from the overlay (For example, focusing
+   * a text box after tapping a button in an
+   * action sheet) then don't restore focus
+   * to previous element
+   */
+  if (document.activeElement === null || document.activeElement === document.body) {
```
liamdebeasi, line 544: We'll need a test for this. Something like the following should suffice:

```js
test('should not return focus to another element if focus already manually returned', async ({
  page,
  skip,
}, testInfo) => {
  skip.browser(
    'webkit',
    'WebKit does not consider buttons to be focusable, so this test always passes since the input is the only focusable element.'
  );
  testInfo.annotations.push({
    type: 'issue',
    description: 'https://github.com/ionic-team/ionic-framework/issues/28849',
  });
  await page.setContent(
    `
    <button id="open-action-sheet">open</button>
    <ion-action-sheet trigger="open-action-sheet"></ion-action-sheet>
    <input id="test-input" />

    <script>
      const actionSheet = document.querySelector('ion-action-sheet');

      actionSheet.addEventListener('ionActionSheetWillDismiss', () => {
        requestAnimationFrame(() => {
          document.querySelector('#test-input').focus();
        });
      });
    </script>
  `,
    config
  );

  const ionActionSheetDidPresent = await page.spyOnEvent('ionActionSheetDidPresent');
  const actionSheet = page.locator('ion-action-sheet');
  const input = page.locator('#test-input');
  const trigger = page.locator('#open-action-sheet');

  // present action sheet
  await trigger.click();
  await ionActionSheetDidPresent.next();

  // dismiss action sheet
  await actionSheet.evaluate((el: HTMLIonActionSheetElement) => el.dismiss());

  // verify focus is in correct location
  await expect(input).toBeFocused();
});
```

You can put this in `src/utils/test/overlays/overlays.e2e.ts`

liamdebeasi, line 544: Can we also add a comment that explains focus is always moved to the body when the overlay is dismissed? Maintainers may not know why we specifically check the body here.

core/src/utils/overlays.ts
```diff
@@ -496,7 +496,7 @@ export const present = async <OverlayPresentOptions>(
    * from returning focus as a result.
    */
   if (overlay.el.tagName !== 'ION-TOAST') {
-    focusPreviousElementOnDismiss(overlay.el);
+    focusPreviousElementOnDismissIfNeeded(overlay.el);
```
liamdebeasi, line 499: Maybe we can call this `restoreElementFocus`? `focusPreviousElementOnDismissIfNeeded` is a bit wordy.

liamdebeasi (changes requested):
Overall looks good! A few changes for tests and maintainability.

core/src/utils/overlays.ts
aeharding: I added some MDN references. Apologies if its now a bit too wordy, feel free to make further suggestions!

aeharding: Done thanks!

core/src/utils/overlays.ts
aeharding: Done!

liamdebeasi (approved):
Thank you!

PR #28865, opened 2024-01-22
fix(accordion): prevent opening of readonly accordion using keyboard

### LINKED ISSUE #28344: bug: readonly accordion can be opened using keyboard controls
### Prerequisites

### Ionic Framework Version

v7.x

### Current Behavior

If you have an `ion-accordion` with `readonly="true"`, tab to it, and press Enter, the accordion will open. This does not occur if `readonly` is set on the parent `ion-accordion-group`.

### Expected Behavior

Readonly accordions should not be interactable, even with keyboard controls.

### Steps to Reproduce

1. Navigate to https://ionicframework.com/docs/api/accordion#individual-accordion-1
2. Use the Tab key to focus the second accordion
3. Press Enter; the accordion will open

### Code Reproduction URL

https://ionicframework.com/docs/api/accordion#individual-accordion-1

### Ionic Info

N/A (reproducible from v7 docs)

### Additional Information

`ion-accordion-group` doesn't exhibit this behavior because of this piece: https://github.com/ionic-team/ionic-framework/blob/f14a59c5e0670ed7cc9ce1a73a087a5af13266e2/core/src/components/accordion-group/accordion-group.tsx#L208-L210

Fixing this in `ion-accordion` should hopefully be as simple as `return`ing early if `readonly` is set here: https://github.com/ionic-team/ionic-framework/blob/f14a59c5e0670ed7cc9ce1a73a087a5af13266e2/core/src/components/accordion/accordion.tsx#L384

mapsandapps:
Issue number: resolves #28344

---------

When an Accordion is inside an Accordion Group, and the Accordion Group is enabled and not readonly but the Accordion is disabled and/or readonly, it is possible to navigate to the accordion and open it using the keyboard. This should not be allowed, just like opening it on click is not allowed.

## What is the new behavior?

- A disabled Accordion inside an Accordion Group may not be opened via the keyboard.
- A readonly Accordion inside an Accordion Group may not be opened via the keyboard.

- [ ] Yes
- [x] No


core/src/components/accordion/test/disabled/accordion.e2e.ts
```diff
@@ -0,0 +1,136 @@
...
+
+      await expect(accordion).toHaveClass(/accordion-collapsed/);
+
+      accordion.click();
+      await page.waitForChanges();
+
+      await expect(accordion).toHaveClass(/accordion-collapsed/);
+    });
+
+    test('should not open accordion via keyboard navigation when group is disabled', async ({ page, browserName }) => {
```
liamdebeasi, line 79: I believe pressing "Enter" on an interactive element will also dispatch a `click` event, so testing that the accordion does not expand on click should be sufficient here (i.e. we don't need to test click and Enter)

core/src/components/accordion/test/disabled/accordion.e2e.ts
```diff
@@ -0,0 +1,136 @@
...
+      await page.keyboard.press(tabKey);
+      await page.waitForChanges();
+
+      await page.keyboard.press('Enter');
+      await page.waitForChanges();
+
+      await expect(accordion).toHaveClass(/accordion-collapsed/);
+    });
+
+    test('should not open accordion via keyboard navigation when accordion is disabled', async ({
```
liamdebeasi, line 106: I believe pressing "Enter" on an interactive element will also dispatch a `click` event, so testing that the accordion does not expand on click should be sufficient here (i.e. we don't need to test click and Enter)

core/src/components/accordion/test/readonly/accordion.e2e.ts
```diff
@@ -0,0 +1,136 @@
...
+
+      await expect(accordion).toHaveClass(/accordion-collapsed/);
+
+      accordion.click();
+      await page.waitForChanges();
+
+      await expect(accordion).toHaveClass(/accordion-collapsed/);
+    });
+
+    test('should not open accordion via keyboard navigation when group is readonly', async ({ page, browserName }) => {
```
liamdebeasi, line 79: I believe pressing "Enter" on an interactive element will also dispatch a `click` event, so testing that the accordion does not expand on click should be sufficient here (i.e. we don't need to test click and Enter)

core/src/components/accordion/test/readonly/accordion.e2e.ts
```diff
@@ -0,0 +1,136 @@
...
+      await page.keyboard.press(tabKey);
+      await page.waitForChanges();
+
+      await page.keyboard.press('Enter');
+      await page.waitForChanges();
+
+      await expect(accordion).toHaveClass(/accordion-collapsed/);
+    });
+
+    test('should not open accordion via keyboard navigation when accordion is readonly', async ({
```
liamdebeasi, line 106: I believe pressing "Enter" on an interactive element will also dispatch a `click` event, so testing that the accordion does not expand on click should be sufficient here (i.e. we don't need to test click and Enter)

liamdebeasi (changes requested):
Implementation looks good, and tests overall look good. I think there a few redundant tests which I've commented on below.

core/src/components/accordion/test/disabled/accordion.e2e.ts
mapsandapps: That's not what I'm seeing. If I revert the bugfix I made in this PR and run the tests, the "should not open accordion via keyboard navigation when accordion is readonly" test fails, and no other tests fail. This is what we expect, since this is the one case this PR addresses. The other tests I added are for behavior that was already correct, but didn't have test coverage.

If the keyboard & mouse tests are redundant with each other, I would expect either both to pass or both to fail in any situation.

Put another way: if I removed the keyboard tests (which I think is what you're suggesting), all the tests would pass without my bugfix being applied.

liamdebeasi: Oh yes I see what you are saying! In `main` all we do is set `pointer-events: none` on the header which prevents click events. But we never disable the underlying element, so you can still tab to an `ion-item` and then hit "Enter". 

Keeping these tests make sense 👍  Thanks for clarifying

liamdebeasi (approved):
Thanks!

PR #28873, opened 2024-01-24
chore(ci): auto assign issues

liamdebeasi:
Issue number: resolves #

---------

## What is the new behavior?

-
-
-

- [ ] Yes
- [ ] No

### DIFF
--- a/.github/workflows/assign-issues.yml
+++ b/.github/workflows/assign-issues.yml
@@ -1,0 +1,19 @@
+name: Assign issues to triage
+
+on:
+  issues:
+    types: [opened]
+
+jobs:
+  auto-assign:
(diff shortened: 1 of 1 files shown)


sean-perkins (approved):

mapsandapps:
the [docs](https://github.com/marketplace/actions/auto-assign-issue#inputs) say 

>  if using the teams input parameter, you need to use a personal access token with read:org scope (the default GITHUB_TOKEN is not enough).

so I don't know if this will work? But we can always merge it and see.

liamdebeasi:
Thank you! I wanted to use the Ionitron token but I mis-remembered what it was called 🤦

mapsandapps (approved):

PR #28889, opened 2024-01-26
fix(item-sliding): newly added options are clickable

### LINKED ISSUE #28662: bug: ion-item-sliding, updating options async while item is open should trigger item to move to accommodate new option set
### Prerequisites

### Ionic Framework Version

v7.x

### Current Behavior

I’m porting an app from ionic 6/cordova to ionic 7/capacitor and something that used to work on the old app is not functioning in the new one.

The scenario is that I’m using an ion-list of ion-item-sliding and I want to be able to dynamically add extra options into an open sliding row.

When I add an option to the row, it appears OK, but the ion-item-sliding text may now fully/partially cover it. So to get around this, I programmatically close() and open() the ion-item-sliding and the options are now fully visible.

Only problem now is that they are no longer clickable and the sliding item itself can’t be manually closed.

I’ve recreated the issue in this app based on the blank example…

Slide the item open. Click on the button to add extra option. Options appears OK… but no events.

### Expected Behavior

After the close()/open() sequence the ion-item-sliding options should be enabled and actionable.

### Steps to Reproduce

Slide the item open. Click on the button to add extra option. Options appears OK… but tapping/clicking on the options do not generate any events, they appear disabled..

### Code Reproduction URL

https://github.com/john-bunyan/ion-item-sliding

### Ionic Info

ionic info   
[WARN] Error loading @ionic/angular package.json: Error [ERR_PACKAGE_PATH_NOT_EXPORTED]: Package subpath './package' is
       not defined by "exports" in /Users/iss/Development/Slider/Slider/node_modules/@ionic/angular/package.json
[WARN] Error loading @capacitor/ios package.json: Error: Cannot find module '@capacitor/ios/package'
       
       Require stack:
       - /usr/local/lib/node_modules/@ionic/cli/lib/project/index.js
       - /usr/local/lib/node_modules/@ionic/cli/lib/index.js
       - /usr/local/lib/node_modules/@ionic/cli/index.js
       - /usr/local/lib/node_modules/@ionic/cli/bin/ionic
[WARN] Error loading @capacitor/android package.json: Error: Cannot find module '@capacitor/android/package'
       
       Require stack:
       - /usr/local/lib/node_modules/@ionic/cli/lib/project/index.js
       - /usr/local/lib/node_modules/@ionic/cli/lib/index.js
       - /usr/local/lib/node_modules/@ionic/cli/index.js
       - /usr/local/lib/node_modules/@ionic/cli/bin/ionic

Ionic:

   Ionic CLI                     : 6.19.1 (/usr/local/lib/node_modules/@ionic/cli)
   @angular-devkit/build-angular : 17.0.5
   @angular-devkit/schematics    : 17.0.5
   @angular/cli                  : 17.0.5

Capacitor:

   Capacitor CLI      : 5.5.1
   @capacitor/core    : 5.5.1

Utility:

   native-run (update available: 2.0.0) : 1.7.4

System:

   NodeJS : v18.16.1 (/Users/iss/.nvm/versions/node/v18.16.1/bin/node)
   npm    : 9.5.1
   OS     : macOS

### Additional Information

_No response_

bmarcelino-fe:
Issue number: resolves #28662

---------

When navigated via tab-key, the ion-item is not highlighted correctly after switching from button=false to button=true.

When item sliding is opened we set specific transform amounts so the Item does not cover the Item Options. However, adding more Options does not cause this specific transform to update. As a result, newly added Options are still covered by the Item.

## What is the new behavior?

Now the `item-sliding` reacts to any option or element added as `item-option` in component.

- [ ] Yes
- [x] No


core/src/components/item-sliding/test/basic/item-sliding.e2e.ts
```diff
@@ -68,6 +68,28 @@ configs({ modes: ['md'], directions: ['ltr'] }).forEach(({ title, screenshot, co
...
+      await button.click();
+
+      // Check if the element added is clicked
+      const elementAdded = page.locator('#element-added-delete');
+      await elementAdded.click();
+
+      await page.waitForChanges();
+
+      // item-sliding doesn't have an easy way to tell whether it's fully open so just screenshot it
+      await expect(item).toHaveScreenshot(screenshot(`dynamic-element-clicked`));
```
sean-perkins, line 90: I think we may be able to avoid a screenshot test here. 

The behavior we want to validate with this PR is if the newly inserted button is focusable. If it cannot receive focus after trying to click the locator, the test will fail.

Something similar to:
```ts
test('the dynamic element should be clicked', async ({ page }) => {
  await page.setContent(`
    /* We can move the specific behavior we are testing here 
      to avoid visual changes to the existing screenshots captured for 
      src/components/item-sliding/test/basic.
      The content here will be evaluated as HTML. Wrap JS in a <script> */
  `, config);

  // Drag the element
  // Click the add button
  // Find the added button
  // Click the button
  // Check if the added button has focused
});
```

The Playwright assertion that can be used here is [`toBeFocused`](https://playwright.dev/docs/api/class-locatorassertions#locator-assertions-to-be-focused). 

Example of an E2E test using `.setContent`: https://github.com/ionic-team/ionic-framework/blob/bf7922c8b37b32dbf90650cb74a2e77bedf0c118/core/src/components/item/test/basic/item.e2e.ts#L18-L39

core/src/components/item-sliding/test/basic/item-sliding.e2e.ts
```diff
@@ -68,6 +68,28 @@ configs({ modes: ['md'], directions: ['ltr'] }).forEach(({ title, screenshot, co
       await expect(item).toHaveScreenshot(screenshot(`item-sliding-gesture`));
     });
 
+    test('the dynamic element should be clicked', async ({ page }) => {
+      await page.goto(`/src/components/item-sliding/test/basic`, config);
```
liamdebeasi, line 72: We should add an annotation. They let us associate a test with an issue so team members have an understanding of why the test exists: 
```suggestion
      test.info().annotations.push({
        type: 'issue',
        description: 'https://github.com/ionic-team/ionic-framework/issues/28662',
      });
      await page.goto(`/src/components/item-sliding/test/basic`, config);
```

liamdebeasi: Note that you might need to re-run the linter.

core/src/components/item-sliding/item-sliding.tsx
```diff
@@ -407,16 +407,17 @@ export class ItemSliding implements ComponentInterface {
   }
 
   private setOpenAmount(openAmount: number, isFinal: boolean) {
+    const { el } = this;
+
     if (this.tmr !== undefined) {
       clearTimeout(this.tmr);
+      el.classList.remove('item-sliding-closing');
```
liamdebeasi, line 414: There was another edge case I did not catch during our debug session. It looks like we clear the timeout in 2 different places: https://github.com/ionic-team/ionic-framework/blob/9448783bb19b187f867054c86d215e3dc97952c1/core/src/components/item-sliding/item-sliding.tsx#L315-L316

As a result, this bug can appear in two different ways. To fix this properly, we can make a `clearCloseTimeout` method that runs the following code:

```js
if (this.tmr !== undefined) {
  clearTimeout(this.tmr);
  this.tmr = undefined;
  this.el.classList.remove('item-sliding-closing');
}
```

From here, we'd need to call `this.clearCloseTimeout()` in [setOpenAmount](https://github.com/ionic-team/ionic-framework/blob/9448783bb19b187f867054c86d215e3dc97952c1/core/src/components/item-sliding/item-sliding.tsx#L410) and [onStart](https://github.com/ionic-team/ionic-framework/blob/9448783bb19b187f867054c86d215e3dc97952c1/core/src/components/item-sliding/item-sliding.tsx#L314).

liamdebeasi (changes requested):

core/src/components/item-sliding/item-sliding.tsx
bmarcelino-fe: A method was created called `resetCloseState()` to hold the code that is called in two different times and updated the code to call this new method.

core/src/components/item-sliding/test/basic/item-sliding.e2e.ts
bmarcelino-fe: The test case was reviewed to avoid the screenshot and check only if the element is clicked. The test was updated too define the content that will be tested.

core/src/components/item-sliding/test/basic/item-sliding.e2e.ts
bmarcelino-fe: Thanks for the suggestion. Already applied it and commited.

core/src/components/item-sliding/item-sliding.tsx
```diff
@@ -296,6 +296,15 @@ export class ItemSliding implements ComponentInterface {
     return !!(this.rightOptions || this.leftOptions);
   }
 
+  // Method to reset item-sliding state
```
liamdebeasi, line 299:
```suggestion
/**
  * Resets any state related to the item closing.
  * We run code in a timeout to clean up the item state
  * after the closing animation has finished. 
  * As part of the item sliding closing, we apply the
  * "item-sliding-closing" class which makes item options
  * not clickable so they can't be accidentally clicked 
  * during the close animation. This class is removed
  * once the animation is complete.
  * 
  * However, if the animation is interrupted (such as when `open()`
  * is called during the close animation), the timeout needs to be cleared.
  * We also need to manually remove the previously noted class otherwise item
  * options will not be clickable.
  */
```

This comment isn't very descriptive of why this code is needed. Maybe we can try something like my suggestion?

liamdebeasi (commented):
The changes look good! I left one comment below. Also, you'll need to run `npm run lint` to clean up any white space issues. You can commit and push the result of the lint.

core/src/components/item-sliding/item-sliding.tsx
bmarcelino-fe: Sounds good for me. I will add this improvement. and run the lint command.

liamdebeasi (approved):
I made a few simplifications to your E2E test -- we only need to include the code necessary to run the test. I was able to remove extra properties and API calls. However, the test ultimately does the same steps.

This look great. Thanks for taking care of this!

liamdebeasi (commented):
It looks like the linked thread actually has 2 separate issues. One issue is fixed by the code here, but the other one should be fixed separately. I'll work on fixing that other issue so we can merge these two PRs around the same time.

PR #28892, opened 2024-01-26
fix(item): ensure button focus state on property change

### LINKED ISSUE #28525: bug: setting button="true" after load does not make item focusable
### Prerequisites

### Ionic Framework Version

v7.x

### Current Behavior

when navigated via tab-key, the ion-item is not highlighted correctly after switching from `button=false` to `button=true`.

### Expected Behavior

ion-item should adjust it's highlighting styles when the button property changes.
the css classes `ion-focusable` and `ion-focused` are no longer present when the bug appears.

my wild guess is that `this.focusable` is initialized once, but never updated:
https://github.com/ionic-team/ionic-framework/blob/v7.5.0/core/src/components/item/item.tsx#L265

### Steps to Reproduce

It's important to start with `foo = false` as initial value to reproduce the bug.
```html
<ion-button (click)="foo = !foo">toggle</ion-button>
<ion-item [button]="foo">
  <ion-text>button mode active: {{foo}}</ion-text>
</ion-item>
``` 

### Code Reproduction URL

_No response_

### Ionic Info

Ionic:

   @angular-devkit/build-angular : 16.2.10
   @angular-devkit/schematics    : 16.2.1
   @angular/cli                  : 16.2.1

Capacitor:

   Capacitor CLI      : 4.8.1
   @capacitor/android : 4.8.1
   @capacitor/core    : 4.8.1
   @capacitor/ios     : 4.8.1

Cordova:

Utility:

   native-run  : 1.7.4

System:

   NodeJS : v18.18.2 (C:\Program Files\nodejs\node.exe)
   npm    : 9.8.1
   OS     : Windows 10

BenOsodrac:
Issue number: resolves #28525

---------

When navigated via tab-key, the ion-item is not highlighted correctly after switching from button=false to button=true.

## What is the new behavior?

Now, when dynamically changing the the `button` option to `True`, there's a `@watch` callback that will make sure the internal `isFocusable` `@state` will be updated.  

- A new unit test was added to item/test. As there was still any unit test for the `ion-item`, a new files was created - `item.spec.tsx`

- [ ] Yes
- [x] No

## Other information

New behavior in runtime:

focusable

### DIFF
--- a/core/src/components/item/test/item.spec.tsx
+++ b/core/src/components/item/test/item.spec.tsx
@@ -3,18 +3,20 @@ import { newSpecPage } from '@stencil/core/testing';
 
 import { Item } from '../item';
 
-it('should change focusable option after switching button option status', async () => {
-  const page = await newSpecPage({
-    components: [Item],
-    template: () => <ion-item button={false}></ion-item>,
+describe('item', () => {
+  it('should change focusable option after switching button option status', async () => {
+    const page = await newSpecPage({
+      components: [Item],
+      template: () => <ion-item button={false}></ion-item>,
+    });
+
+    const item = page.body.querySelector('ion-item')!;
+    // Change button attribute to true
+    item.setAttribute('button', 'true');
+
+    await page.waitForChanges();
+
+    // Check if it has the expected class that gives the highlight style to .item-highlight element
+    expect(item).toHaveClass('ion-focusable');
   });
-
-  const item = page.body.querySelector('ion-item')!;
-  // Change button attribute to true
-  item.setAttribute('button', 'true');
-
-  await page.waitForChanges();
-
-  // Check if it has the expected class that gives the highlight style to .item-highlight element
-  expect(item).toHaveClass('ion-focusable');
 });
--- a/core/src/components/item/test/item.spec.tsx
+++ b/core/src/components/item/test/item.spec.tsx
@@ -1,0 +1,20 @@
+import { h } from '@stencil/core';
+import { newSpecPage } from '@stencil/core/testing';
+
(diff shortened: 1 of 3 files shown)


core/src/components/item/test/item.spec.tsx
```diff
@@ -0,0 +1,20 @@
+import { h } from '@stencil/core';
+import { newSpecPage } from '@stencil/core/testing';
+
+import { Item } from '../item';
+
+it('should change focusable option after switching button option status', async () => {
```
sean-perkins, line 6: I will refactor this when I address the parent comment, for tests we try to nest them under a describe block for the component name. 

For example:
```ts
describe('item', () => {
  /* existing test */
});
```

It helps when reviewing the logs locally or in CI to know exactly which component the spec test is responsible for.

sean-perkins (commented):
Great work! I am able to verify locally that the `ion-button` is able to receive focus and appropriately styling when the `button` property is dynamically set. 

I am going to reset `core/src/components/item/test/buttons/index.html` back to `origin/main`. We have existing screenshots that use the entire page template for the screenshot: https://github.com/ionic-team/ionic-framework/blob/main/core/src/components/item/test/buttons/item.e2e.ts#L13 and adding new elements will cause "visual regressions" to the existing screenshot references. Since we are not adding new E2E tests, we don't necessarily need this diff beyond local testing.

sean-perkins (approved):

core/src/components/item/test/item.spec.tsx
BenOsodrac: Ah, got it! 👍

PR #28893, opened 2024-01-26
test(ci): run tests in docker container

liamdebeasi:
Issue number: Internal

---------

The team currently faces two challenges:

1. Verifying visual changes locally is difficult because we cannot use the existing ground truths as they were generated in Linux environments and most of our team uses either macOS or Windows. While team members can generate ground truths in the correct environment, they need to remember to do that first before making changes.
2. Updating visual diffs is time consuming and can only be done by team members. Our GitHub Action runs the entire test suite which can take ~10 even if only a handful of screenshots are generated. Additionally, this job can only be run by team members meaning community contributors cannot update/add screenshots. This limits them to non-visual tasks when contributing. In the event that they do want to make visual changes, the team needs to copy all their code into a branch and manually run screenshot diffs for them.

## What is the new behavior?

- This PR introduces the ability to run all Playwright tests inside of a Docker container using an image with Playwright dependencies. The container will have access to the local project, so developers can make changes and then run tests in the container after the changes are compiled. This enables anyone to propose new screenshot changes. However, the "update screenshot" job will still be available for folks who do not want/are unable to use docker.
- There are some typeface differences between GH Actions and the Docker image which is why there are a handful of screenshots that needed to be updated.

One risk here is that the Playwright npm and Docker image versions must be kept in sync. As a result, I also updatRenovate to allow us to auto update the npm and Docker image versions at the same time.

- [ ] Yes
- [x] No

## Other information

⚠️ There are still some issues I need to sort out with mounting the local project on Windows. However, using Ubuntu with the linux subsystem for windows can be used as a workaround. I'd like to merge this so we can start testing it in our day-to-day workflow and ironing out any bugs.


.gitignore
```diff
@@ -69,6 +69,8 @@ core/www/
 core/test-results/
 core/playwright-report/
 core/**/*-snapshots
```
sean-perkins, line 71: All screenshots are currently ignored locally. Do we need to reconsider this regex pattern to allow screenshots generated from Linux to be tracked by source control?

I can see this being a point of confusion for developers, that they would need to forcibly track the new screenshots.

liamdebeasi: Good catch - fixed. Tested this locally by making a change to chip and screenshots updated inside of Docker are tracked correctly. Screenshots updated outside of Docker are not tracked.

image

sean-perkins (approved):
This looks good to me. Excited to have this option for forks/contributors as well as the added bonus of not having to wait for CI.

PR #28901, opened 2024-01-29
refactor(angular-server): build with angular 16

liamdebeasi:
[Issue number: Internal

---------

Ionic 8 drops support for Angular 14 and 15. As a result, Ionic Angular Server should be built with Angular 16 to account for supported versions of Angular.

## What is the new behavior?

- Updates Ionic Angular Server to build with Angular 16

Note: The ng14 and ng15 tests will fail because they are not compatible. These test apps are being removed as part of FW-5868. However, the rest of this PR can be reviewed. Once the test apps are removed, I will sync in the changes and all tests should pass.

- [ ] Yes
- [x] No

## Other information

⚠️ Blocked on merging until FW-5868 is complete.


thetaPC (approved):
LGTM, works fine locally

PR #28902, opened 2024-01-29
refactor(angular): build with angular 16

liamdebeasi:
[Issue number: Internal

---------

Ionic 8 drops support for Angular 14 and 15. As a result, Ionic Angular should be built with Angular 16 to account for supported versions of Angular.

## What is the new behavior?

- Updates Ionic Angular to build with Angular 16

Note: The ng14 and ng15 tests will fail because they are not compatible. These test apps are being removed as part of FW-5868. However, the rest of this PR can be reviewed. Once the test apps are removed, I will sync in the changes and all tests should pass.

- [ ] Yes
- [x] No

## Other information

⚠️ Blocked on merging until FW-5868 is complete.


packages/angular/common/src/directives/navigation/router-outlet.ts
```diff
@@ -42,7 +42,9 @@ import { RouteView, StackDidChangeEvent, StackWillChangeEvent, getUrl, isTabSwit
   inputs: ['animated', 'animation', 'mode', 'swipeGesture'],
 })
 // eslint-disable-next-line @angular-eslint/directive-class-suffix
-export class IonRouterOutlet implements OnDestroy, OnInit {
+export abstract class IonRouterOutlet implements OnDestroy, OnInit {
+  abstract outletContent: any;
```
liamdebeasi, line 46: Previously we were passing in the outletContent via the constructor. However, newer versions of Angular warn that this will override the base implementation. However, this is what we want so I made the base class an abstract class to indicate that extending classes need to implement outletContent.

packages/angular/common/src/directives/navigation/router-outlet.ts
```diff
@@ -296,7 +297,8 @@ export class IonRouterOutlet implements OnDestroy, OnInit {
 
       // Calling `markForCheck` to make sure we will run the change detection when the
       // `RouterOutlet` is inside a `ChangeDetectionStrategy.OnPush` component.
-      enteringView = this.stackCtrl.createView(this.activated, activatedRoute);
+      // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
+      enteringView = this.stackCtrl.createView(this.activated!, activatedRoute);
```
liamdebeasi, line 306: The build started to fail because `this.activated` was potentially `null`, but `createView` does not accept `null` for this param. This has always been a problem, but it only started to fail with Angular 16.

thetaPC (approved):
LGTM, no issues locally. Thank you for the comments!

PR #28906, opened 2024-01-29
refactor(action-sheet): use new step color tokens

amandaejohnston:
Issue number: Internal

---------

Action sheet uses an `--ion-color-step-*` token directly. These tokens will be deprecated soon.

## What is the new behavior?

New token added, while keeping the old one for backwards compatibility.

- [ ] Yes
- [x] No


mapsandapps (approved):
LGTM 👍🏻

PR #28907, opened 2024-01-29
fix(overlays): tear down animations after dismiss

### LINKED ISSUE #28352: bug: `IonModal` with `presentingElement` doesn't render on iOS
### Describe the Bug

When using the ["Card modal"](https://ionicframework.com/docs/api/modal#card-modal) component, we noticed that it _doesn't render_ its content after _switching the viewport orientation_ back and forth:

https://github.com/ionic-team/ionic-docs/assets/68850090/63cfd525-0575-4f3c-90e4-4c116b0a0972

Unfortunately, using the "Layers" tool I didn't notice any irregularities:
image 
image

However, **_the issue doesn't happen_** if you rotate the device with the opened modal:

https://github.com/ionic-team/ionic-docs/assets/68850090/d8c65b17-724f-4432-9c83-c0eb584c984f

Or you don't open the modal in the horizontal view:

https://github.com/ionic-team/ionic-docs/assets/68850090/776fabd1-5be8-423a-9b79-eb17bf2187fe

---

Moreover, the issue doesn't happen at all when the `presentingElement` is removed from the `IonModal`, thus removing the "Card effect":
image

https://github.com/ionic-team/ionic-docs/assets/68850090/bcfd3c2d-04c4-4814-b276-e1df9e5037c5

### Expected Behavior

The modal is consistently opened and rendered both in the vertical and horizontal modes. Prior openings of the modal shouldn't impact its rendering in the following ones. In short: should work like in the docs and demos 👌🏻 

### Steps to Reproduce

This bug can be reproduced using the official ["Card Modal" example](https://ionicframework.com/docs/api/modal#card-modal):
1. Open the example in Stackblitz 
2. Open the preview in a new tab
3. Enable mobile view testing
4. Choose an iPhone device
5. Vertical Open -> Close
6. Rotate the view
7. Horizontal Open -> Close
8. Rotate the view
9. Vertical Open

Then you'll see that the modal's content isn't rendered.

The issue happens not only on the Web but on real iOS devices as well.

### Operating System

Macbook Pro, M1 2021, MacOS Sonoma 14.0

### Browser

Chrome

### Version

117.0.5938.149

mapsandapps:
Issue number: resolves #28352

---------

Not all animations are getting properly destroyed after they run. This means that styles can get stuck at the end value of their animation. 

Specifically for this reproduction (as reported in the bug ticket), if the modal animates from 1 to 0 opacity and then the modal gets reopened with a different animation where the opacity should be 1 throughout (i.e. the opacity isn't supposed to animate at all), the modal is invisible because the opacity got stuck at 0 and never got reset to the default value of 1.

This bug is probably causing some incorrect behavior on other edge cases with overlays, but this is the only one I've identified.

### Reproduction steps
Note: you cannot reproduce this when using a modalController

1. Open a modal, e.g. [this one](http://localhost:3333/src/components/modal/test/card-nav?ionic:mode=ios) in `ios` mode on a screen wider than 768px
1. Close the modal
1. Open the same modal on a screen narrower than 768px
1. See that the modal does not appear

## What is the new behavior?

- Animations are properly destroyed after the animation completes
- The modal now appears as expected after following the reproduction steps above

- [ ] Yes
- [x] No

### DIFF
--- a/core/src/components/modal/test/animations/modal.e2e.ts
+++ b/core/src/components/modal/test/animations/modal.e2e.ts
@@ -11,7 +11,12 @@ configs({ modes: ['ios'], directions: ['ltr'] }).forEach(({ config, title }) => 
         config
       );
     });
-    test('card modal should clean up animations on dismiss', async ({ page }) => {
+    test('card modal should clean up animations on dismiss', async ({ page }, testInfo) => {
+      testInfo.annotations.push({
+        type: 'issue',
+        description: 'https://github.com/ionic-team/ionic-framework/issues/28352',
+      });
+
       const ionModalDidDismiss = await page.spyOnEvent('ionModalDidDismiss');
 
       const modal = page.locator('ion-modal');
--- a/core/src/components/modal/test/animations/modal.e2e.ts
+++ b/core/src/components/modal/test/animations/modal.e2e.ts
@@ -2,41 +2,39 @@ import { expect } from '@playwright/test';
 import { configs, test } from '@utils/test/playwright';
 
 configs({ modes: ['ios'], directions: ['ltr'] }).forEach(({ config, title }) => {
-  test.describe(title('card modal - animations'), () => {
+  test.describe(title('modal: animations'), () => {
     test.beforeEach(async ({ page }) => {
       await page.setContent(
         `
-        <ion-button id="open-modal">Open Modal</ion-button>
-
-        <ion-modal trigger="open-modal"></ion-modal>
+        <ion-modal is-open="true" trigger="open-modal"></ion-modal>
       `,
         config
       );
     });
-    test('should have animations and then clean them up', async ({ page }) => {
-      page.locator('#open-modal').click();
-
-      await page.waitForChanges();
+    test('card modal should clean up animations on dismiss', async ({ page }) => {
+      const ionModalDidDismiss = await page.spyOnEvent('ionModalDidDismiss');
 
       const modal = page.locator('ion-modal');
 
-      let animations = await page.evaluate(() => {
+      const initialAnimations = await page.evaluate(() => {
         return document.querySelector('ion-modal')!.shadowRoot!.getAnimations();
       });
 
       // While the modal is open, it should have animations
-      await expect(animations.length).toBeGreaterThan(0);
+      await expect(initialAnimations.length).toBeGreaterThan(0);
 
-      await modal.evaluate(async (el: HTMLIonModalElement) => {
-        await el.dismiss();
+      await modal.evaluate((el: HTMLIonModalElement) => {
+        el.dismiss();
       });
 
-      animations = await page.evaluate(() => {
+      await ionModalDidDismiss.next();
+
+      const currentAnimations = await page.evaluate(() => {
         return document.querySelector('ion-modal')!.shadowRoot!.getAnimations();
       });
 
       // Once the modal has finished closing, there should be no animations
-      await expect(animations.length).toBe(0);
+      await expect(currentAnimations.length).toBe(0);
     });
   });
 });
--- a/core/src/components/modal/test/animations/modal.e2e.ts
+++ b/core/src/components/modal/test/animations/modal.e2e.ts
@@ -26,7 +26,7 @@ configs({ modes: ['ios'], directions: ['ltr'] }).forEach(({ config, title }) => 
       });
 
       // While the modal is open, it should have animations
-     expect(initialAnimations.length).toBeGreaterThan(0);
+      expect(initialAnimations.length).toBeGreaterThan(0);
 
       await modal.evaluate((el: HTMLIonModalElement) => {
         el.dismiss();
@@ -39,7 +39,7 @@ configs({ modes: ['ios'], directions: ['ltr'] }).forEach(({ config, title }) => 
       });
 
       // Once the modal has finished closing, there should be no animations
-     expect(currentAnimations.length).toBe(0);
+      expect(currentAnimations.length).toBe(0);
     });
   });
 });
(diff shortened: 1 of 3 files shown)


core/src/components/modal/test/animations/modal.e2e.ts
```diff
@@ -0,0 +1,42 @@
+import { expect } from '@playwright/test';
+import { configs, test } from '@utils/test/playwright';
+
+configs({ modes: ['ios'], directions: ['ltr'] }).forEach(({ config, title }) => {
+  test.describe(title('card modal - animations'), () => {
```
sean-perkins, line 5: Can we prefix the title with `modal:` to match similar e2e tests?

I'd recommend this block to be:
```
modal: animations
```

and then then actual test case to be:
```
card modal should clean up animations on dismiss
```

core/src/components/modal/test/animations/modal.e2e.ts
```diff
@@ -0,0 +1,42 @@
...
+    test.beforeEach(async ({ page }) => {
+      await page.setContent(
+        `
+        <ion-button id="open-modal">Open Modal</ion-button>
+
+        <ion-modal trigger="open-modal"></ion-modal>
```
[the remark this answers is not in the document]
mapsandapps: updated in 08f301167f

core/src/components/modal/test/animations/modal.e2e.ts
mapsandapps: updated in 08f301167f thanks!

core/src/components/modal/test/animations/modal.e2e.ts
```diff
@@ -0,0 +1,40 @@
...
+    test.beforeEach(async ({ page }) => {
+      await page.setContent(
+        `
+        <ion-modal is-open="true" trigger="open-modal"></ion-modal>
+      `,
+        config
+      );
+    });
+    test('card modal should clean up animations on dismiss', async ({ page }) => {
+      const ionModalDidDismiss = await page.spyOnEvent('ionModalDidDismiss');
```
thetaPC, line 15: Add an annotation that keeps track of what GH issue this test is linked it.

```ts
testInfo.annotations.push({
  type: 'issue',
  description: 'https://github.com/ionic-team/ionic-framework/issues/28352',
});
```

thetaPC (changes requested):

core/src/components/modal/test/animations/modal.e2e.ts
mapsandapps: thank you! i always forget this is a thing. added in 26fb033434

thetaPC (approved):
LGTM

sean-perkins (approved):
Fantastic work 👏

core/src/components/modal/test/animations/modal.e2e.ts
```diff
@@ -0,0 +1,45 @@
...
+
+      const modal = page.locator('ion-modal');
+
+      const initialAnimations = await page.evaluate(() => {
+        return document.querySelector('ion-modal')!.shadowRoot!.getAnimations();
+      });
+
+      // While the modal is open, it should have animations
+      await expect(initialAnimations.length).toBeGreaterThan(0);
```
liamdebeasi, line 29:
```suggestion
     expect(initialAnimations.length).toBeGreaterThan(0);
```

This assertion is not async

core/src/components/modal/test/animations/modal.e2e.ts
```diff
@@ -0,0 +1,45 @@
...
+      });
+
+      await ionModalDidDismiss.next();
+
+      const currentAnimations = await page.evaluate(() => {
+        return document.querySelector('ion-modal')!.shadowRoot!.getAnimations();
+      });
+
+      // Once the modal has finished closing, there should be no animations
+      await expect(currentAnimations.length).toBe(0);
```
liamdebeasi, line 42:
```suggestion
     expect(currentAnimations.length).toBe(0);
```

This assertion is not async

core/src/components/modal/test/animations/modal.e2e.ts
```diff
@@ -0,0 +1,45 @@
...
+
+      await modal.evaluate((el: HTMLIonModalElement) => {
+        el.dismiss();
+      });
+
+      await ionModalDidDismiss.next();
+
+      const currentAnimations = await page.evaluate(() => {
+        return document.querySelector('ion-modal')!.shadowRoot!.getAnimations();
+      });
```
liamdebeasi, line 39: Nit: You can use the existing locator to run your evaluation:

```suggestion
      const currentAnimations = await modal.evaluate((el: HTMLIonModalElement) => {
        return el.shadowRoot!.getAnimations();
      });
```

core/src/components/modal/test/animations/modal.e2e.ts
```diff
@@ -0,0 +1,45 @@
...
+        description: 'https://github.com/ionic-team/ionic-framework/issues/28352',
+      });
+
+      const ionModalDidDismiss = await page.spyOnEvent('ionModalDidDismiss');
+
+      const modal = page.locator('ion-modal');
+
+      const initialAnimations = await page.evaluate(() => {
+        return document.querySelector('ion-modal')!.shadowRoot!.getAnimations();
+      });
```
liamdebeasi, line 26: Nit: You can use the existing locator to run your evaluation:

```suggestion
      const initialAnimations = await modal.evaluate((el: HTMLIonModalElement) => {
        return el.shadowRoot!.getAnimations();
      });
```

liamdebeasi (approved):
Great job! I tested this in another app I was debugging, and it fixes one of the issues there as well.

PR #28915, opened 2024-01-30
fix(action-sheet): iOS dismiss animation respects safe area

### LINKED ISSUE #28541: bug: Action sheet does not completely translate off bottom of page with safe areas
### Prerequisites

### Ionic Framework Version

v7.x

### Current Behavior

When you dismiss an action sheet on a screen with safe areas, the action sheet does not fully transition off the screen before being destroyed.

Once you see it, you cannot unsee :)

### Expected Behavior

Action sheet should fully transition off screen before destroy

### Steps to Reproduce

1. On device with safe areas, present action sheet
2. Dismiss
3. Observe the very top of action sheet at the bottom of the device viewport disappears because it is not fully off page before being removed

### Code Reproduction URL

https://github.com/aeharding/action-sheet-dismiss

### Ionic Info

Ionic:

   Ionic CLI       : 7.1.1 (/Users/aeharding/.nvm/versions/node/v20.8.1/lib/node_modules/@ionic/cli)
   Ionic Framework : @ionic/react 7.5.5

Capacitor:

   Capacitor CLI      : 5.5.1
   @capacitor/core    : 5.5.1

Utility:

System:

   NodeJS : v20.8.1 (/Users/aeharding/.nvm/versions/node/v20.8.1/bin/node)
   npm    : 10.1.0
   OS     : macOS Unknown

### Additional Information

https://github.com/ionic-team/ionic-framework/assets/2166114/69f49007-8b37-45d4-98eb-cf14e0aad445

sean-perkins:
Issue number: resolves #28541

---------

On iOS when dismissing the `ion-action-sheet`, the animation does not account for the safe area of the device. This results in the action sheet not completely animating off the visible viewport on a device with safe area enabled. 

## What is the new behavior?

- The `ion-action-sheet` will animate consistently off the viewport when dismissing, including the safe area.

To better support custom animations not needing to account for the safe area, the safe area has been added to the padding of the action sheet container. This results in the height increasing based on the bottom safe area and animating correctly when translating between [`100%` and `0%`](https://github.com/ionic-team/ionic-framework/blob/9d57758e3ec1095c6eb951d2640b95084b711bbc/core/src/components/action-sheet/animations/ios.enter.ts#L23).

|Before|After|
|---|---|
|<video src="https://github.com/ionic-team/ionic-framework/assets/13732623/d9475a54-f1b9-4283-ae1a-86e2b71a9d9e"></video>|<video src="https://github.com/ionic-team/ionic-framework/assets/13732623/d051b4d9-4f45-4a51-b522-510f8cddef74"></video>|

In the recorded examples the bottom safe area is exaggerated to show the dismiss animation differences. 

- [ ] Yes
- [x] No

## Other information

To test this PR, I would recommend adding the following styles to the action sheet basic test:
```html
<style>
ion-action-sheet {
  --ion-safe-area-bottom: 100px;
}
</style>
```

You can then open dev-tools and slow the animation speed to 10%:

CleanShot 2024-01-29 at 19 49 32@2x

- Open and dismiss the action sheet tied to the "Basic" button. 
- Verify that the action sheet animates completely off the view when dismissed.
- Open the and dismiss the action sheet tied to the "Custom CSS Class" button. This implementation customizes the `--height` of the action sheet. 
- Verify that the action sheet animates completely off the view when dismissed.


core/src/components/action-sheet/test/basic/index.html
```diff
@@ -62,7 +62,7 @@
         --ion-safe-area-bottom: 40px;
 
         --max-width: 300px;
-        --height: 325px;
+        --height: calc(325px + var(--ion-safe-area-bottom));
```
sean-perkins, line 65: This avoid unnecessary diffs with our basic test, except for places we are explicitly setting the `--ion-safe-area-bottom` to a custom value (action-sheet-custom-diff screenshots).

core/src/components/action-sheet/test/basic/action-sheet-rendering.e2e.ts-snapshots/action-sheet-custom-diff-md-ltr-Mobile-Chrome-linux.png
sean-perkins, line 1: These diffs are a result of: https://github.com/ionic-team/ionic-framework/pull/28915/files#r1471958550. We either get diffs on MD or iOS. I could see it making sense to only have diffs on iOS for this change. 

Let me know if you think it is worth changing this.

core/src/components/action-sheet/action-sheet.ios.scss
```diff
@@ -26,7 +26,7 @@
 // ---------------------------------------------------
 
 .action-sheet-wrapper {
-  @include margin(var(--ion-safe-area-top, 0), auto, var(--ion-safe-area-bottom, 0), auto);
+  @include margin(var(--ion-safe-area-top, 0), auto, null, auto);
```
liamdebeasi, line 29: With this PR, the bottom safe area may change how much of the action sheet content is visible, but it does not change the computed height of the action sheet wrapper. In other words, if `--height: 300px` but `--ion-safe-area-bottom: 40px` then the visible area will be `260px` even though the overall container size remains at `300px`.

This behavior aligns with how we do things on MD, but this is a change in behavior to how the iOS action sheet used to work. I'm unclear if this is what we want though The `--height` docs state that this controls the "height of the action sheet". I would assume that the "action sheet" refers to the visible area.

Is there a way we can implement this where the visible area of the action remains === `--height`? Alternatively, would you consider the current behavior wrong and the new behavior (where wrapper height includes the bottom safe area value)?

liamdebeasi: Changes discussed on our call:

```
.action-sheet-wrapper {
  @include margin(var(--ion-safe-area-top, 0), auto, null, auto);
  @include padding(null, null, var(--ion-safe-area-bottom, 0), null);

  box-sizing: content-box;
}
```

We'd like to add a note as to why box-sizing is needed here. We should also verify that this does not have an adverse effects for descendant element sizing. We also discussed reverting the test file change for `--height`.

core/src/components/action-sheet/action-sheet.ios.scss
```diff
@@ -26,7 +26,13 @@
...
+  @include padding(null, null, var(--ion-safe-area-bottom, 0), null);
```
liamdebeasi, line 30:
```suggestion
  /**
   * Bottom safe area is applied as padding so that it impacts the bounding box.
   * When the action sheet is shown/hidden, this element is transformed by translating
   * 100% of its height. This translation needs to include the bottom safe area
   * otherwise part of the action sheet will still be visible at the end of
   * the show transition.
   *
   * If this code is changed, reviewers should verify that the action
   * sheet still translates out of the viewport completely when the bottom
   * safe area is a positive value.
   */
  @include padding(null, null, var(--ion-safe-area-bottom, 0), null);
```

Maybe something like this?

liamdebeasi (approved):
One non-blocking comment. Great job!

core/src/components/action-sheet/action-sheet.ios.scss
sean-perkins: Love it, thank you for helping out with the comment! Running behind today 😅

PR #28923, opened 2024-01-30
refactor(alert): use correct color step tokens

amandaejohnston:
Issue number: Internal

---------

Alert is using the wrong type of step tokens for a couple styles -- only text elements should use `$text-color-step-*`, while everything else (including borders) should use `$background-color-step-*`.

## What is the new behavior?

Tokens updated, including changing the step values so the actual color resolved remains the same.

- [ ] Yes
- [x] No


thetaPC (approved):
LGTM

PR #28926, opened 2024-01-30
refactor(modal): update to use new step token

brandyscarney:
Issue number: internal

---------

The modal handle only uses the `--ion-color-step-350` variable with a fallback to `#c0c0be`.

## What is the new behavior?
The modal handle falls back to using the `--ion-background-color-step-350` variable if it is defined but prioritizes `--ion-color-step-350` if it is also defined. This is the new step color token added for high-contrast themes.

- [ ] Yes
- [x] No

## Other information

No visual diffs are expected.

### DIFF
--- a/core/src/components/modal/modal.scss
+++ b/core/src/components/modal/modal.scss
@@ -129,7 +129,7 @@ ion-backdrop {
 
   border: 0;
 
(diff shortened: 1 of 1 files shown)


thetaPC (approved):
LGTM

PR #28941, opened 2024-01-31
refactor(refresher): use new step color tokens

mapsandapps:
Issue number: Internal

---------

The refresher uses the --ion-color-step tokens or a fallback.

## What is the new behavior?

- The refresher colors use the --ion-color-step tokens or the --ion-text-color-step tokens or a fallback
- The refresher background colors use the --ion-color-step tokens or the --ion-background-color-step tokens or a fallback

- [ ] Yes
- [x] No

### DIFF
--- a/core/src/components/refresher/refresher.ios.vars.scss
+++ b/core/src/components/refresher/refresher.ios.vars.scss
@@ -7,7 +7,7 @@ $refresher-ios-icon-color:            $text-color !default;
 $refresher-ios-text-color:            $text-color !default;
 
 /// @prop - Color of the native refresher spinner
-$refresher-ios-native-spinner-color:  var(--ion-color-step-450, var(--ion-text-color-step-450, #747577)) !default;
+$refresher-ios-native-spinner-color:  var(--ion-color-step-550, var(--ion-text-color-step-450, #747577)) !default;
 
 /// @prop - Width of the native refresher spinner
 $refresher-ios-native-spinner-width:  32px !default;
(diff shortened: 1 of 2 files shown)


core/src/components/refresher/refresher.md.vars.scss
```diff
@@ -10,10 +10,10 @@ $refresher-md-text-color:                     $text-color !default;
 $refresher-md-native-spinner-color:           #{ion-color(primary, base)} !default;
 
 /// @prop - Border of the native refresher spinner
-$refresher-md-native-spinner-border:          1px solid var(--ion-color-step-200, #ececec) !default;
+$refresher-md-native-spinner-border:          1px solid var(--ion-color-step-800, var(--ion-text-color-step-200, #ececec)) !default;
```
liamdebeasi, line 13: The border is considered a background, so we don't need to invert the numbers

liamdebeasi (changes requested):

core/src/components/refresher/refresher.ios.vars.scss
```diff
@@ -7,7 +7,7 @@ $refresher-ios-icon-color:            $text-color !default;
 $refresher-ios-text-color:            $text-color !default;
 
 /// @prop - Color of the native refresher spinner
-$refresher-ios-native-spinner-color:  var(--ion-color-step-450, #747577) !default;
+$refresher-ios-native-spinner-color:  var(--ion-color-step-550, var(--ion-text-color-step-450, #747577)) !default;
```
liamdebeasi, line 10: This should use a background color token too. We are using the `color` property here, but it's being applied to an SVG which is confusing. Basically the only time we need to use `--ion-text-color-step` is for setting the color of text.

The example in the design document is a bit confusing, so I clarified this on the design doc: https://github.com/ionic-team/ionic-framework-design-documents/pull/238

Also, we don't need to modify the original `--ion-color-step` number.

liamdebeasi (approved):
Good to go when the below comment is addressed

liamdebeasi (approved):

PR #28942, opened 2024-01-31
refactor: remove legacy tokens from dark theme

liamdebeasi:
Issue number: Internal

---------

The current dark theme still has references to the legacy stepped color tokens.

## What is the new behavior?

- Removes references the legacy stepped color tokens. The dark theme already includes references to the new stepped color tokens

As part of this work I also realized I needed to add support  for `--ion-text-color-step` and `--ion-background-color-step` to the default iOS and MD themes, so I did that in this PR (without it we would have gotten screenshot diffs).

- [ ] Yes
- [x] No


brandyscarney (approved):

brandyscarney (changes requested):
Actually removing my approval, will comment why

core/src/css/themes/dark.scss
brandyscarney, line 1: We need to update the usage here:

```scss
  & ion-modal {
    --ion-background-color: var(--ion-color-step-100);
    --ion-toolbar-background: var(--ion-color-step-150);
    --ion-toolbar-border-color: var(--ion-color-step-250);
  }
```

liamdebeasi: Good catch! fixed in https://github.com/ionic-team/ionic-framework/pull/28942/commits/cbbed8028e133064aed541d70443a01a85766d97

brandyscarney (approved):

PR #28943, opened 2024-01-31
test(angular): remove ng14 and ng15 test apps

brandyscarney:
Issue number: internal

---------

Test apps exist for ng14 and ng15

## What is the new behavior?
Removes Angular test apps for ng14 and ng15

- [ ] Yes
- [x] No


.github/workflows/build.yml
thetaPC, line 1: Should the `stencil-nightly.yml` also be [updated](https://github.com/ionic-team/ionic-framework/blob/c5c78a32cf296caf51e8ee558cdb9b0dc8f7fbd4/.github/workflows/stencil-nightly.yml#L153)?

packages/angular/test/README.md
```diff
@@ -108,9 +108,9 @@ Note: You may encounter some other peer dependency issues not covered by the Ang
...
 
-5. Create a directory in `apps` named `ng14`.
-6. Copy the modified files to the `apps/ng14` directory.
+5. Create a directory in `apps` named `ng17`.
+6. Copy the modified files to the `apps/ng17` directory.
 7. Open `./github/workflows/build.yml` and find the `test-angular-e2e` job.
 8. Find the `apps` field under `matrix`.
-9. Add "ng14" to the `apps` field.
+9. Add "ng17" to the `apps` field.
 10. Commit these changes and push.
```
mapsandapps, line 116: I think these should probably stay the same, since they're continuing the example of making a new test app for Angular 14. (Either that, or the sections above that reference 13 & 14 should change, but I think it might be easiest to just keep them with these old numbers, since they're referring to making a new app anyway.)

.github/workflows/build.yml
brandyscarney: I noticed that but I wasn't sure what it was for because it doesn't even have `ng16` or `ng17` so I left it alone.

mapsandapps (changes requested):
Just one comment. I think everything else is good 👍🏻

packages/angular/test/README.md
brandyscarney: Thank you for catching that! I did a find and replace not realizing that section was showing how to add a new one, ha. Fixed here: https://github.com/ionic-team/ionic-framework/pull/28943/commits/fe3da871bf2fd1b08ad20b8c39fa5369768f7941

mapsandapps (approved):

.github/workflows/build.yml
liamdebeasi: I'll fix this in `main` since that's the branch the stencil nightlies run off

liamdebeasi: https://github.com/ionic-team/ionic-framework/pull/28945

liamdebeasi (approved):

.github/workflows/build.yml
thetaPC: Thanks!

thetaPC (approved):
LGTM

PR #28946, opened 2024-01-31
refactor(input): remove accept property

thetaPC:
Issue number: internal

---------

`ion-input` has a deprecated property: `accept`. This property was marked deprecated because the `file` value for `type` is not a valid value in Ionic Framework.

## What is the new behavior?

- Removed the `accept` property.
- Updated the breaking change to include this change.

- [x] Yes
- [ ] No

`accept` can only be used when the input has a type of `file`. However, `file` is not a valid value in Ionic Framework so the `accept` property was removed.

## Other information

[Related PR](https://github.com/ionic-team/ionic-framework/pull/25501)

### DIFF
--- a/BREAKING.md
+++ b/BREAKING.md
@@ -138,6 +138,7 @@ This allows components to inherit the color properly when used outside of Ionic 
 <h4 id="version-8x-input">Input</h4>
 
 - `size` has been removed from the `ion-input` component. Developers should use CSS to specify the visible width of the input.
+- `accept` has been removed from the `ion-input` component.
 
 <h4 id="version-8x-nav">Nav</h4>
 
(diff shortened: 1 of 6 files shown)


liamdebeasi (approved):
Change looks good. Don't forget to add a `BREAKING CHANGE` message to the bottom of the commit description when merging: https://github.com/ionic-team/ionic-framework/blob/main/.github/CONTRIBUTING.md#examples

Also, the commit should probably be `refactor` here instead of `feat` since we are removing a feature, not adding one.

liamdebeasi:
Also make sure the scope of the commit subject is `refactor(input)` and not `feat(input)` (since we are removing a feature not adding one)

PR #28949, opened 2024-01-31
chore(angular): remove type merging for schematic options

thetaPC:
Issue number: internal

---------

`schematicOpts` has an intersection type in order to be backward compatible.

## What is the new behavior?

- Remove the added type since `SchematicOptions` type now includes `[property: string]: any;` with Angular 16.

- [ ] Yes
- [x] No

`schematicOpts` has the same type, just a different way of getting it.

## Other information

N/A


sean-perkins (approved):
Perfect!

PR #28950, opened 2024-01-31
fix(overlays): focus is returned to last focus element when focusing toast

### LINKED ISSUE #28261: bug: focus trapping is not disabled on non-top overlay which causes side effects with toast
### Prerequisites

### Ionic Framework Version

v6.x

### Current Behavior

When a modal is open, and then show a toast message: A click on the toast scroll the modal to the next e.g. button.

demo

It's a critical bug in my opinion or case. Because it always scrolls to the next focusable element either vertically or horizontally. If you have content with overflow: hidden and custom scroll button (like next in an intro slider) you jump to the element which is in overflow hidden. But anyway, this is a weird bug which needs to be fixed. I cannot figure out what the problem is. Maybe you have an idea?

### Expected Behavior

Don't affect the scroll position. A click on toast (regardless if close button or everywhere else) should not scroll anything.

### Steps to Reproduce

1. Create a modal with scrollable content.
2. Add a ion-button to the end of content.
3. Add an element on top which is not focusable. In this case just a div with click event.
4. The click event of div opens a toast message.
5. Click on the message. In this demo you have to click twice. In my project a single click also causes the issue. Not sure why, but it does not matter. The issue can be reproduce in this demo here.
6. See that a click on the toast scroll the content of the demo. It's maybe because of focus. But why? It should not happen!

Tip: Use mobile view in dev tools (Chrome "Toggle device toolbar". I'm on Windows 10 btw. But this bug is also present in Android).

[Demo](https://ionic-6-toast-scroll-bug.stackblitz.io)

Again: In my project, this bug is also triggered on a single click e.g. on X. In this demo here you need to click twice. Anywhere on toast body. The second click can be on X. But it does not matter where you click. 

### Code Reproduction URL

https://stackblitz.com/edit/ionic-6-toast-scroll-bug?file=src%2Fapp%2Fapp.component.html

### Ionic Info

Ionic 6.6.1 (Angular 14). Not yet tested in Ionic 7. But I suspect the same. It's a candidate for patch fix in v6.

liamdebeasi:
Issue number: resolves #28261

---------

When moving focus from a focus-trapped overlay to a toast, focus is moved back to the overlay. This is the correct behavior as focus should never leave a focus-trapped overlay (unless the overlay is dismissed or focus is moved to a _new_ top-most overlay). However, the way we return focus is a bit unexpected because it always returns focus to the last focusable element in the overlay. 

This means that if you were focused on the first focusable element, presented the toast, and then focused the toast, focus might not be moved back to that first focusable element. In the case of the linked issue, this was causing an unexpected scroll so that the last focused element could be in view.

## What is the new behavior?

- This fix adds an exception for `ion-toast` (as it is the only overlay that is **not** focus trapped) that ensures that focus is moved back to the last focus element.

- [ ] Yes
- [x] No

## Other information

Dev build: `7.7.1-dev.11707253408.186eea70`

Note: We don't recommend this pattern in general because it would be impossible for a screen reader user to focus the toast. However, we can at least improve the experience for developers who continue to implement this pattern by returning focus in a more predictable manner.

Docs: https://github.com/ionic-team/ionic-docs/pull/3432

Testing: 

Reviewers should manually test the following behaviors:

1. Create a modal with 2 buttons. Have one of the buttons present a toast. Open the toast and verify that you can still Tab to cycle through the buttons in the modal.
2. Create a modal with 2 buttons. Have one of the buttons present a toast. Open the toast. Move focus to the toast and verify that you can still Tab to cycle through the buttons in the modal (once focus is returned to the modal).


core/src/utils/overlays.ts
```diff
@@ -131,38 +137,55 @@ export const createOverlay = <T extends HTMLIonOverlayElement>(
...
-  if (firstInput) {
-    focusElement(firstInput);
-  } else {
-    // Focus overlay instead of letting focus escape
-    overlay.focus();
-  }
+  focusElementInOverlay(firstInput, overlay);
 };
 
-const isOverlayHidden = (overlay: Element) => overlay.classList.contains('overlay-hidden');
```
liamdebeasi, line 152: there should be no behavior change to focusFirstDescendant and focusLastDescendant. All I did here was abstract the focusing logic to `focusElementInOverlay` and had those functions call it instead. I needed to use `focusElementInOverlay` to implement the fix, so I figured to consolidate otherwise we're writing the same code 3x.

core/src/utils/overlays.ts
```diff
@@ -131,38 +137,55 @@ export const createOverlay = <T extends HTMLIonOverlayElement>(
...
   if (shadowRoot) {
     // If there are no inner focusable elements, just focus the host element.
-    lastInput = shadowRoot.querySelector(focusableQueryString) || lastInput;
+    elementToFocus = shadowRoot.querySelector<HTMLElement>(focusableQueryString) || hostToFocus;
   }
 
-  if (lastInput) {
-    lastInput.focus();
+  if (elementToFocus) {
+    focusVisibleElement(elementToFocus);
```
liamdebeasi, line 188: One thing I noticed was that `focusLastDescendant` called `.focus()` whereas `focusFirstDescendant` calls `focusVisibleElement`. I believe this was an oversight on my end when I originally implemented this since I was only concerned about moving focus to the first element in the select popover: https://github.com/ionic-team/ionic-framework/pull/23474/commits/8be9ae1f7e218e229d1e5a9981bec26f644e0b1f 

However, focusVisibleElement also calls `.focus` so there should be no behavior change.

liamdebeasi:
I decided to consolidate some code under the hood otherwise I was going to write the same code 3x. This isn't strictly necessary to ship this fix (though it makes it easier to implement the fix), so if the reviewers would prefer I pull this out to a separate PR let me know!

amandaejohnston (commented):
Functionality looks great in my local testing 👍 Just had a question about the tests.

core/src/utils/test/overlays/overlays.e2e.ts
```diff
@@ -254,5 +254,125 @@ configs({ modes: ['ios'], directions: ['ltr'] }).forEach(({ title, config }) =>
...
+
+    test('focusing toast from a shadow overlay should return focus to the last focused element', async ({
+      page,
+      skip,
+    }) => {
+      test.info().annotations.push({
+        type: 'issue',
+        description: 'https://github.com/ionic-team/ionic-framework/issues/28261',
+      });
+      skip.browser('webkit', 'WebKit does not consider buttons to be focusable');
```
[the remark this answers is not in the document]
liamdebeasi: Added in https://github.com/ionic-team/ionic-framework/pull/28950/commits/5d0121bd9eda801f9d618c30df58ba5ae2c5297c. I only did it for the shadow component tests (modal/popover). Doing the same for the scoped overlays is tricky because we don't have an easy way of connecting a non-button to an overlay since we are restricted in what we can display in these overlays.

However, I think the current coverage should be sufficient. In the event that this breaks due a browser bug, the shadow test should at least fail.

amandaejohnston (approved):

PR #28951, opened 2024-01-31
refactor(angular, react, vue): add missing test app msg

thetaPC:
Issue number: N/A

---------

The packages, Angular, React, and Vue, will accept any `app_dir` parameter when passing it to the build script. This can lead to a user to entering a directory that doesn't exist within `test/apps`. If this happens, then the build folder will include an incomplete test app.

For example:
1. `./build ng13`
2. This creates a `build/ng13` folder, but the folder only contains the contents from the `base` folder. This causes it to be incomplete.

## What is the new behavior?

- Provides a message that the provided app directory doesn't exist.
- Does not create the incomplete app when an invalid app directory is given.

- [ ] Yes
- [x] No

## Other information

N/A

### DIFF
--- a/packages/angular/test/build.sh
+++ b/packages/angular/test/build.sh
@@ -22,6 +22,12 @@ FULL_BASE_DIR="${BASE_DIR}/."
 # The full path to the built application.
 BUILD_APP_DIR="${BUILD_DIR}/${APP_DIR}/"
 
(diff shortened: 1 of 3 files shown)


amandaejohnston (approved):
Tested locally and works great, thank you!

Issue #23037, opened 2021-03-11
bug: Duplicate Alert button handler 'role' values cause incorrect handlers to be executed


lincolnthree:
# Bug Report

**Ionic version:**

[ ] **4.x**
[x] **5.x**

**Current behavior:**
If two buttons in an alert both have the 'cancel' role, only the first handler will ever be called, regardless of which button is pressed. Changing the roles to be different will result in expected behavior.

**Expected behavior:**
The handler for the pressed button should be called, regardless of its 'role'.

**Steps to reproduce:**
1. See code below.
2. Click "BUTTON 1"
3. Observe that the handler for "BUTTON 2" will have been called.

**Related code:**

```
const alert = await this.alertCtrl.create({
            header: 'Header',
            message:'Some message here',
            buttons:[{
              text: 'BUTTON 1',
              role: 'cancel',
              handler: () => {
               console.log('BUTTON 1');
              }
            }, {
              text: 'BUTTON 2',
              role: 'cancel',
              handler: () => {
               console.log('BUTTON 2');
              }
            }]
          });
          alert.present();
```

Aside... can we get an updated StackBlitz/Plunkr for the latest version of Ionic Angular, etc? :) They are still on 3 & 4.

**Other information:**

**Ionic info:**

```
sharktop:topdecked-unified lincoln$ ionic info

Ionic:

   Ionic CLI                     : 6.12.4 (/usr/local/lib/node_modules/@ionic/cli)
   Ionic Framework               : @ionic/angular 5.6.0
   @angular-devkit/build-angular : 0.1200.0-next.3
   @angular-devkit/schematics    : 12.0.0-next.3
   @angular/cli                  : 12.0.0-next.3
   @ionic/angular-toolkit        : 3.1.0

Capacitor:

   Capacitor CLI   : 3.0.0-beta.6
   @capacitor/core : 3.0.0-beta.6

Utility:

   cordova-res : 0.15.3
   native-run  : 1.3.0

System:

   NodeJS : v14.15.0 (/usr/local/bin/node)
   npm    : 6.14.9
   OS     : macOS Catalina

```

PR #28952, opened 2024-02-01
[fix](alert): Call pressed button handler, regardless its 'role'

joselrio:
Issue number: resolves https://github.com/ionic-team/ionic-framework/issues/23037

---------

If two buttons in an alert both have the cancel role, only the first handler will ever be called, regardless of which button is pressed.

## What is the new behavior?
Now the handler for the pressed button is called, regardless of its 'role'.

- [ ] Yes
- [X] No

## Other information

> Before the fix:
WithoutTheFix

> After the fix:
WithTheFix


core/src/components/alert/alert.tsx
```diff
@@ -462,12 +462,12 @@ export class Alert implements ComponentInterface, OverlayInterface {
   private async buttonClick(button: AlertButton) {
     const role = button.role;
     const values = this.getValues();
-    if (isCancel(role)) {
+    if (isCancel(role) && button.handler === undefined) {
```
sean-perkins, line 465: I'll need to ask the team, but this change does change the current behavior of a `cancel` role button with a custom handler that returns `false`. 

Previously, it would have dismissed the overlay. With the proposed changes it would not dismiss. 

I _think_ the change in behavior is valid/expected, but never hurts to get a second opinion 🙂

sean-perkins: Hello! After speaking with the team, I'd advise we adjust the implementation pattern here. We should de-risk and validate a use case for needing to prevent a dismiss with a cancel role, before making that change. 

The proposed fix for the original issue should execute the handler callback, but not prevent dismissing. Something to the effect of:
```ts
private async buttonClick(button: AlertButton) {
  const role = button.role;
  const values = this.getValues();
  const returnData = await this.callButtonHandler(button, values);
  if (isCancel(role)) {
    /**
     * Cancel roles should always dismiss the alert,
     * even if the handler function returns `false`.
     */
    return this.dismiss({ values, ...returnData }, role);
  }
  if (returnData !== false) {
    return this.dismiss({ values, ...returnData }, role);
  }
  return false;
}
```

While investigating this issue further, I found that [`ion-action-sheet`](https://github.com/ionic-team/ionic-framework/blob/51c729eafc3b290a5daddf7f0ccffd0a2a9fe2aa/core/src/components/action-sheet/action-sheet.tsx#L252-L262), [`ion-picker`](https://github.com/ionic-team/ionic-framework/blob/51c729eafc3b290a5daddf7f0ccffd0a2a9fe2aa/core/src/components/picker/picker.tsx#L296-L306), and [`ion-toast`](https://github.com/ionic-team/ionic-framework/blob/51c729eafc3b290a5daddf7f0ccffd0a2a9fe2aa/core/src/components/toast/toast.tsx#L523-L533) all have very similar implementations and likely the same problematic behavior. 

We can discuss more in Slack, but I think addressing the fix across all these implementations would be great to knock out together, once we validate the path for one. I can help out here if desired.

sean-perkins:
@joselrio changes look great! 

Are we able to write an automated test using Playwright to validate the button handler is called?

Feel free to ping me on Slack if you have questions on how to write the test.

joselrio:
Hi @sean-perkins,

Can you please take a look at the latest stuff I made to this?
Didn't implement ActionSheet cancel button tests since at this context having more than one cancel button is not allowed.
Let me know if any change is need.

Cheers,
JR

sean-perkins:
Hello @joselrio can you incorporate the changes that I applied here: https://github.com/ionic-team/ionic-framework/pull/28952/commits/91b5b8b3bb82da20fd0cfed373c478809f01e12d#diff-5ce56bbcb3bd1b1e868dd811c6c96de0bb2d8f72d3dde47fb2d83c48989b6159 for the picker component tests, to the remaining overlays?

A few key points on why we would prefer this change:
1. Our changes in this PR do not differ based on the text direction or the `mode` (visual appearance at this time). As a result, we don't need to test against those conditions in our E2E test. This means our tests will run faster!
2. Building on above, our changes in this PR are logical - does a callback function get called when we expect it to? We don't need screenshots to verify this. Screenshots are significantly slower to execute than parsing a condition of the DOM. 
3. This one we do inconsistently today in Ionic Framework, but we are actively trying to reduce coupling between components and tests, where they are not needed. For example, testing this behavior for `ion-picker`, does not depend on `ion-button`. We can just use a HTML `button` in this case. This is most important with screenshots, but by reducing coupling, we have significantly less risk of something else breaking/changing in the future. We also request less JS per test as a result, which improves performance of the tests. 

Let me know if you have any questions or concerns!

PR #28956, opened 2024-02-01
feat: add high contrast implementation

liamdebeasi:
Issue number: Internal

---------

Users do not have a way of increasing the contrast in Ionic apps. This is valuable for people with low vision as increasing the contrast between foreground and background content helps improve readability.

## What is the new behavior?

- Adds a high contrast light and high contrast dark theme. As with our other themes, developers can choose between system, class, and always stylesheets.

While we aim to improve contrast for text and UI components, this feature prioritizes text in the event that both text and UI component cannot be improved without one negatively impacting the other.

- [ ] Yes
- [x] No

## Other information

Dev build: `7.6.7-dev.11706894781.1cd59fde`

Testing instructions:

1. Open `src/themes/test/css-variables`. Activate the high contrast light and dark themes to verify that contrast does increase.
2. Use the dev build to integrate the theme into a test app (conference app, starter app, etc).

I'd recommend using these imports:

```css
@import "@ionic/angular/css/themes/high-contrast.system.css";
@import "@ionic/angular/css/themes/high-contrast-dark.system.css";
```
Note: Make sure this is imported **after** `core.scss`


core/src/components/item/item.ios.vars.scss
```diff
@@ -25,7 +25,7 @@ $item-ios-paragraph-margin-start:            $item-ios-paragraph-margin-end !def
 $item-ios-paragraph-font-size:               dynamic-font(14px) !default;
 
 /// @prop - Color of the item paragraph
-$item-ios-paragraph-text-color:              rgba($text-color-rgb, .4) !default;
+$item-ios-paragraph-text-color:              var(--ion-text-color-step-550, #a3a3a3) !default;
```
liamdebeasi, line 28: Not needed to ship this feature, so let me know if you'd prefer I add this in a separate PR.

Devs can add a `p` inside of an item. The problem is that `p` won't participate in high contrast because it's using the `--ion-text-color-rgb` var with an alpha channel instead of a step color. This change allows it to participate in high contrast. The values I chose are very close to the original computed values.

core/src/css/themes/high-contrast-light.scss
```diff
@@ -89,12 +100,48 @@ $colors:  (
...
+    --ion-text-color-step-700: #{mix($lightest-text-color, $darkest-text-color, 70%)};
+    --ion-text-color-step-750: #{mix($lightest-text-color, $darkest-text-color, 75%)};
+    --ion-text-color-step-800: #{mix($lightest-text-color, $darkest-text-color, 80%)};
+    --ion-text-color-step-850: #{mix($lightest-text-color, $darkest-text-color, 85%)};
+    --ion-text-color-step-900: #{mix($lightest-text-color, $darkest-text-color, 90%)};
+    --ion-text-color-step-950: #{mix($lightest-text-color, $darkest-text-color, 95%)};
   }
+}
 
+@mixin high-contrast-light-md-theme() {
```
liamdebeasi, line 137: Toast (for both light and dark high contrast themes) was implemented incorrectly because it was scoped to all toasts when it should only have been scoped to MD toasts. This resulted in screenshot diffs.

core/src/css/themes/high-contrast-light.scss
```diff
@@ -89,12 +100,48 @@ $colors:  (
...
   // Toast
   // --------------------------------------------------
 
+  & ion-toast {
+    --color: var(--ion-background-color);
+  }
+
```
liamdebeasi, line 144: One thing I missed during the design phase of this toast workaround is that the text stepped colors would also be changing. I built this solution before I worked on the new text stepped colors architecture.

The problem is that `--color` by default uses a stepped color. But since those stepped colors are changing in high contrast, the way they mix using `color-mix` also changes. This resulted in text not having sufficient contrast. As a result, all text in high contrast mode will use the background color of the page.

Note that this is not a type. The MD toast background is a dark gray in light mode, so using the page background color (white) will ensure correct contrast. Similarly, the background is white in dark mode (where the page background is dark gray).

liamdebeasi: I had to make the same adjustments in the high contrast dark theme too

sean-perkins (approved):
Feature looks great to me with my testing! I noted the gotcha I ran into with the import order on the stylesheets in our internal thread. We'll have that accounted for in the documentation and starter templates.

amandaejohnston (approved):
Looks great!

PR #28958, opened 2024-02-02
chore(angular, react, vue): update readmes for pnpm usage

thetaPC:
Issue number: internal

---------

The Readmes are all using `npm` in the building sections.

## What is the new behavior?

- The Readmes are all using `pnpm` in the building sections.

- [ ] Yes
- [x] No

## Other information

N/A

### DIFF
--- a/packages/angular/README.md
+++ b/packages/angular/README.md
@@ -21,7 +21,7 @@ Ionic Angular specific building blocks on top of [@ionic/core](https://www.npmjs
 ## Testing ng-add in ionic
 
 1. Pull the latest from `main`
-2. Build ionic/angular: `npm run build`
+2. Build ionic/angular: `pnpm build`
 3. Run `npm link` from `ionic/angular/dist` directory
 4. Create a blank angular project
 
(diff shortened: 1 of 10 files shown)


packages/angular/README.md
```diff
@@ -72,3 +72,35 @@ Ionic developers can access this by importing from `@ionic/angular/standalone`.
...
+pnpm build
+```
+
+> [!WARNING]
+> If you are receiving errors about missing packages, you may need to run `pnpm install --force` to force the installation of the missing packages.
+
+3. Install dependencies in `@ionic/angular`:
+
+```shell
+cd packages/angular && pnpm install
```
sean-perkins, line 96: `pnpm install` will only be needed the first time to establish the symlink between core and angular. I wonder if we just recommend `pnpm install -r` as the first step to install all dependencies in the workspace and that will simplify the flow for development to just building the specific packages needed.

packages/angular/test/README.md
```diff
@@ -6,19 +6,16 @@ Ionic Framework supports multiple versions of Angular. As a result, we need to v
...
 
-1. Build the `core` and `packages/angular` directories using `npm run build`.
-2. [Build the Angular test app](#test-app-build-structure).
-3. Navigate to the built test app directory (e.g. `packages/angular/test/build/ng14`).
-4. Install dependencies using `npm install`.
-5. Sync your local changes using `npm run sync`.
+1. [Build](../README.md#building) the `core` and `packages/angular` projects using `pnpm build`.
+2. Watch for changes made in `packages/angular` by using `pnpm build.watch`.
+3. [Build the Angular test app](#test-app-build-structure).
+4. Navigate to the built test app directory (e.g. `packages/angular/test/build/ng14`).
```
sean-perkins, line 12: `ng14` will be gone soon, maybe we should use `ng17` for now?

packages/angular/README.md
thetaPC: I thought about that too. My hesitation is if a developer is only wanting to work in Angular, should they also have to wait for the dependencies like `@ionic/react`? It doesn't take very long do the installs though, so I'm open to using `pnpm install -r`.

amandaejohnston (approved):
Thank you!

sean-perkins (approved):
I'm good with the current changes. If we need to tweak any language/steps in the future, we can always do that. 

Thanks for updating these!

packages/angular/README.md
thetaPC: For record: We are leaving it as is: https://github.com/ionic-team/ionic-framework/pull/28958#pullrequestreview-1860465599.

PR #28965, opened 2024-02-02
fix(select): popover can be scrolled

liamdebeasi:
Issue number: resolves #

---------

## What is the new behavior?

-
-
-

- [ ] Yes
- [ ] No

### DIFF
--- a/core/src/components/select-popover/select-popover.scss
+++ b/core/src/components/select-popover/select-popover.scss
@@ -8,3 +8,14 @@ ion-list-header,
 ion-label {
   @include margin(0);
 }
+
+
+/**
+ * The main content inside of a popover has overflow: hidden
+ * so that scrolling scroll a descendant ion-content.
+ * However, select-popover does not use ion-content so we need
+ * to explicitly make this container scrollable.
+ */
+:host {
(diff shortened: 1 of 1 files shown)


thetaPC (approved):
LGTM, works as intended

tgangso:
FYI. This did not resolve the issue for me when using a component in the popover. Example:

this.popover.create({
      component: XPopoverComponent,
      
But it works when I add the following CSS to the XPopoverComponent CSS file.

:host {
  overflow-y: auto;
}

capc0:
@liamdebeasi I can confirm @tgangso issue with version 7.7.5

A workaround is to set the following in a global style set:
```css
ion-popover>* {
    overflow-y: auto;
}
```

Should I open another issue?

thetaPC:
Thank you for bringing this to our attention!

Yes, please open a new issue with a minimal repro so we can investigate this.

capc0:
> Thank you for bringing this to our attention!
> 
> Yes, please open a new issue with a minimal repro so we can investigate this.

fyi I have created the following issue to track this: https://github.com/ionic-team/ionic-framework/issues/29211

PR #28966, opened 2024-02-02
refactor(dynamic-font): remove --ion-default-dynamic-font

thetaPC:
Issue number: internal

---------

Dynamic font scaling was not enabled by default.

## What is the new behavior?

Dynamic font scaling is enabled by default.

- Updated the breaking change file to include the change.

- [x] Yes
- [ ] No

`--ion-default-dynamic-font` has been removed completely and developers need to use `--ion-dynamic-font` instead. Plus, dynamic font scaling is now enabled by default.

## Other information

N/A

### DIFF
--- a/BREAKING.md
+++ b/BREAKING.md
@@ -138,6 +138,7 @@ This allows components to inherit the color properly when used outside of Ionic 
 <h4 id="version-8x-input">Input</h4>
 
 - `size` has been removed from the `ion-input` component. Developers should use CSS to specify the visible width of the input.
+- `accept` has been removed from the `ion-input` component. This was previously used in conjunction with the `type="file"`. However, the `file` value for `type` is not a valid value in Ionic Framework.
 
 <h4 id="version-8x-nav">Nav</h4>
 
--- a/BREAKING.md
+++ b/BREAKING.md
@@ -125,6 +125,7 @@ Developers who had previously chosen dynamic font scaling by activating it in th
 
 Developers who want to disable dynamic font scaling can set `--ion-dynamic-font: initial;` in their global stylesheets. However, this is not recommended because it may introduce accessibility challenges for users who depend on enlarged font sizes.
 
+For more information on the dynamic font, refer to the [Dynamic Font Scaling documentation](https://ionicframework.com/docs/layout/dynamic-font-scaling).
 <h2 id="version-8x-components">Components</h2>
 
 <h4 id="version-8x-button">Button</h4>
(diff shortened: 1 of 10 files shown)


BREAKING.md
```diff
@@ -113,6 +115,19 @@ body {
...
+The `core.css` file has been updated to enable dynamic font scaling by default:
+
+```diff
+html {
+- --ion-default-dynamic-font: -apple-system-body;
++ --ion-dynamic-font: -apple-system-body;
+}
+```
+
+The `--ion-default-dynamic-font` variable has been removed and replaced with `--ion-dynamic-font`. This change may have unintentional side effects in apps if the font was not expected to scale dynamically.
```
brandyscarney, line 129: I think we need to include something for users that were enabling the dynamic font. In our [guide for enabling dynamic font scaling](https://ionicframework.com/docs/layout/dynamic-font-scaling#enabling-dynamic-font-scaling-in-ionic) we tell the developers how to enable it. Maybe we should say somewhere that developers can remove this code and it will still work? 

I am just not sure how showing the code snippet above will be helpful to developers without seeing how `--ion-dynamic-font` is used or linking to what dynamic font is. We could show them how to disable it, or just let them know why they should keep it enabled and that they shouldn't be changing the `font-size` of their apps on the `html` element.

thetaPC: Hopefully this sounds better.

BREAKING.md
```diff
@@ -113,6 +115,16 @@ body {
...
+<h4 id="version-8x-dynamic-font">Dynamic Font</h4>
+
+The `core.css` file has been updated to enable dynamic font scaling by default.
+
+The `--ion-default-dynamic-font` variable has been removed and replaced with `--ion-dynamic-font`.
+
+Developers who had previously chosen dynamic font scaling by activating it in their global stylesheets can revert to the default setting by removing their custom CSS. In doing so, their application will seamlessly continue utilizing dynamic font scaling as it did before. It's essential to note that altering the font-size of the html element should be avoided, as it may disrupt the proper functioning of dynamic font scaling.
+
+Developers who want to disable dynamic font scaling can set `--ion-dynamic-font: initial;` in their global stylesheets. However, this is not recommended because it may introduce accessibility challenges for users who depend on enlarged font sizes.
+
```
brandyscarney, line 127:
```suggestion

For more information on the dynamic font, refer to the [Dynamic Font Scaling documentation](https://ionicframework.com/docs/layout/dynamic-font-scaling).
```

brandyscarney (approved):

PR #28968, opened 2024-02-02
refactor(refresher-content): remove the aria-label

thetaPC:
Issue number: N/A

---------

`refresher-content` passes `aria-label` to the `ion-spinner`. This was done to improve web accessibility.

## What is the new behavior?

We realized that `ion-spinner` may not have the correct `role`. We plan to investigate to determine if we should continue using `role="progressbar"` or if the component is a decorative component. Based on that research, we will update the component.

In the mean time, the code reverts back to what v7 currently has until a proper solution is implemented.

- Removed the `aria-label` because the team will re-evaluate if the spinners should use `aria-hidden` instead.

- [ ] Yes
- [x] No

## Other information

N/A

### DIFF
--- a/core/src/components/refresher/test/a11y/refresher.e2e.ts
+++ b/core/src/components/refresher/test/a11y/refresher.e2e.ts
@@ -23,6 +23,7 @@ configs({ directions: ['ltr'], modes: ['md'], themes: ['light', 'dark'] }).forEa
 
       await expect(refresher).toHaveClass(/refresher-pulling/);
 
+      // TODO(FW-5937): Remove the disableRules once the ticket is resolved.
       const results = await new AxeBuilder({ page }).disableRules('aria-progressbar-name').analyze();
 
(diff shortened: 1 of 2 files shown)


core/src/components/refresher/test/a11y/refresher.e2e.ts
```diff
@@ -23,7 +23,7 @@ configs({ directions: ['ltr'], modes: ['md'], themes: ['light', 'dark'] }).forEa
 
       await expect(refresher).toHaveClass(/refresher-pulling/);
 
-      const results = await new AxeBuilder({ page }).analyze();
+      const results = await new AxeBuilder({ page }).disableRules('aria-progressbar-name').analyze();
```
liamdebeasi, line 26: Is there a tech debt ticket we reference here?

thetaPC: :face_palm: forgot that

PR #28976, opened 2024-02-05
fix(angular RouterLinkDelegateDirective): Make routerLink respect target and keyboard modifiers

### LINKED ISSUE #26394: feat: Have the feature of "href" and "router-link" in a unified way for web and mobile apps
### Prerequisites

For now, if we use "router-link" args on ion-button or other tags in the app, this does not allow to do a "shift+click" to open a link in a new tab. For this, we need to use "href" args.

But using "href" args seems to break the "swipe-to-navigage-back" behavior on iOS.

Is it possible to reconcile both? Either have the swipe-to-navigate-back gesture work even when following href links, or have a way to selectively choose the "href-behavior" when compiling for the web, and the "router-link" behavior when compiling on mobile?

### Describe the Use Case

This would allow people to have the best of both worlds: allow to open views in new tabs when needed on web UIs, and allow to have the "swipe-to-go-back" gesture on iOS

_No response_

### Describe Alternatives

_No response_

### Related Code

_No response_

### Additional Information

_No response_

j-oppenhuis:
Issue number: resolves #26394

---------

This fixes using routerLink in combination with a target different than _self. It also fixes a bug where CTRL + click or shift + click does not work like expected in combination with routerLink in Angular.

## What is the new behavior?
When using a target other than _self and keyboard modifiers we are now stopping execution of navigating by routeDirection. This is not needed because the app will reload either way. 

- [ ] Yes
- [x] No

### DIFF
--- a/packages/angular/common/src/directives/navigation/router-link-delegate.ts
+++ b/packages/angular/common/src/directives/navigation/router-link-delegate.ts
@@ -49,6 +49,29 @@ export class RouterLinkDelegateDirective implements OnInit, OnChanges {
    */
   @HostListener('click', ['$event'])
   onClick(ev: UIEvent): void {
+    /**
+     * Using a target different from _self will
+     * open another browser tab.
+     * Therefore, use default
+     * behaviour of this event.
+     */
+    if (this.elementRef.nativeElement.target === undefined || this.elementRef.nativeElement.target !== '_self') {
+      return;
+    }
+
+    /**
+     * Using a keyboard modifier will
+     * open another browser tab.
+     * Therefore, use default
+     * behaviour of this event.
+     */
+    if (
+      (ev instanceof MouseEvent || ev instanceof TouchEvent)
+      && (ev.altKey || ev.ctrlKey || ev.metaKey || ev.shiftKey)
+    ) {
+      return;
+    }
+
     this.navCtrl.setDirection(this.routerDirection, undefined, undefined, this.routerAnimation);
 
     /**


thetaPC (changes requested):
Thank you for submitting a PR to the project! We appreciate your contribution.

There are a few things that need to be addressed before we can move forward with the PR.

1. Command clicking is not following expected browser behavior. While I’m able to `CMD` + `click` on the button and the link opens in a new tab, the active page navigates to the same url.
2. The `OPTION` key is not being taken into account when clicking on the button if the user is using a Mac. The user shouldn't be able to open the link in a new tab and the page should not redirect to the link passed to `routerLink`.
3. As a noted in the first item, the page should not redirect to the link passed to `routerLink`. The redirection is happening because of a [behavior](https://github.com/ionic-team/ionic-framework/blob/f885a5526addab8abf93a95de8980e4ef78cbd36/core/src/components/button/button.tsx#L284) that is implemented in `ion-button`. We would want to prevent Ionic's click listener from triggering when a new tab is opened.
4. We also want to check the operating system of the user to tie the correct key to the correct action. The `OPTION` key should only be taken into account when the user is using a Mac.

Let me know if you have any questions! Thank you for your time and effort!

j-oppenhuis:
Thank you for your feedback. I understand that it is more complex than my initial idea. Instead of changing routerLink functionality, we should probably change the `openURL` function. Am I correct? If that is the case than we should indeed also pass in the operating system and the target. But is that a good solution?

j-oppenhuis:
I have looked into the code and I think my suggestion is incorrect. But it seems we are not able to prevent the button or item  navigation by using stoppropagation. The only thing I can think of is adding a `routerLink` `prop` to the button and item components. Then we could make a distinction between desktop and mobile. When the current OS is desktop, the anchor could handle navigation. Otherwise, we could use openURL like we use now. Is that something that could work?

ShaneK:
Hey, thanks for your PR!

We've addressed this issue in another PR ( https://github.com/ionic-team/ionic-framework/pull/31230 ), but have given you co-author credit for it thanks to your work on this one. This issue's fix will be deployed in the next release of Ionic Framework.

PR #28977, opened 2024-02-05
refactor(toast): remove cssClass from ToastButton

thetaPC:
Issue number: internal

---------

The deprecated `cssClass` is still part of `ToastButton` even though the buttons can be access through shadow parts.

## What is the new behavior?

- Removed `cssClass` from `ToastButton`
- Updated the Breaking Change file to inform developers that they can add their styles through the "button" shadow part.

- [x] Yes
- [ ] No

## Other information

N/A

### DIFF
--- a/BREAKING.md
+++ b/BREAKING.md
@@ -153,3 +153,5 @@ This allows components to inherit the color properly when used outside of Ionic 
 <h4 id="version-8x-toast">Toast</h4>
 
 - `cssClass` has been removed from the `ToastButton` interface. This was previously used to apply a custom class to the toast buttons. Developers can use the "button" shadow part to style the buttons.
+
+For more information on styling toast buttons, refer to the [Toast Theming documentation](https://ionicframework.com/docs/api/toast#theming).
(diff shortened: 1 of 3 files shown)


BREAKING.md
```diff
@@ -149,3 +149,7 @@ This allows components to inherit the color properly when used outside of Ionic
 - `ion-picker` and `ion-picker-column` have been renamed to `ion-picker-legacy` and `ion-picker-legacy-column`, respectively. This change was made to accommodate the new inline picker component while allowing developers to continue to use the legacy picker during this migration period.
   - Only the component names have been changed. Usages such as `ion-picker` or `IonPicker` should be changed to `ion-picker-legacy` and `IonPickerLegacy`, respectively.
   - Non-component usages such as `pickerController` or `useIonPicker` remain unchanged. The new picker displays inline with your page content and does not have equivalents for these non-component usages.
+
+<h4 id="version-8x-toast">Toast</h4>
+
+- `cssClass` has been removed from the `ToastButton` interface. This was previously used to apply a custom class to the toast buttons. Developers can use the "button" shadow part to style the buttons.
```
brandyscarney, line 155: Can we include an example or link to https://ionicframework.com/docs/api/toast#theming?

core/src/components/toast/toast-interface.ts
```diff
@@ -4,7 +4,6 @@ import type { IonicSafeString } from '../../utils/sanitization';
 export interface ToastOptions {
   header?: string;
   message?: string | IonicSafeString;
-  cssClass?: string | string[];
```
brandyscarney, line 7: I don't think this was supposed to be removed, `cssClass` still exists for Toast: https://github.com/ionic-team/ionic-framework/blob/FW-4923/core/src/components/toast/toast.tsx#L123

brandyscarney (changes requested):

core/src/components/toast/toast-interface.ts
thetaPC: Good catch! Thank you!!

brandyscarney (approved):

PR #28982, opened 2024-02-06
feat(menu): apply strict type for menu type

sean-perkins:
Issue number: N/A

---------

`ion-menu` currently accepts any `string` to the `type` property, even though only `overlay`, `push` and `reveal` are only supported.

## What is the new behavior?

- Adds more explicit types to the `type` property on `ion-menu`

- [ ] Yes
- [x] No


amandaejohnston (commented):
Should this be added to the breaking changes guide? I'm genuinely not sure; I know that any other strings wouldn't have worked anyway, but *technically* people could be using other things which would now error.

sean-perkins:
@amandaejohnston I opted to leave it out of the breaking changes, since only misconfigurations should be affected. We aren't removing any options that previously existed. Typescript compiler/type checking should inform the developer clearly of the accepted type. 

Still targeted it for v8, since type changes (introducing strictness) are considered breaking.

amandaejohnston (approved):

PR #28430, opened 2023-10-29
Fix( popover-arrow ): render above backdrop

Sahil-B07:
Issue number: ionic-team/ionic-framework#28985
fixes: ionic-team/ionic-framework#28985

---------

## What is the current behavior?
when the popover is presented in iOS the arrow is lighter than the popover container as a result of rendering under the backdrop.
Screenshot 2023-10-29 at 11 13 39 PM

## What is the new behavior?
Now the arrow is visible and is rendering above/in-Front-of backdrop.
Screenshot 2023-10-29 at 11 11 53 PM

PR #28986, opened 2024-02-06
fix(popover): render arrow above backdrop

### LINKED ISSUE #28985: bug: Popover arrow is rendered under the backdrop
### Prerequisites

### Ionic Framework Version

v7.x

### Current Behavior

**This issue only seems to reproduce on Mac**

As seen in the [Side and Alignment Demo](https://ionicframework.com/docs/api/popover#side-and-alignment-demo), when the popover is presented in iOS the arrow is lighter than the popover container as a result of rendering under the backdrop. 

CleanShot 2023-06-02 at 12 16 36

Additionally, while hard to visibly see based on the background color, the arrow is not tight to the container and is ~1px offset from where it should render.

### Expected Behavior

1. The popover arrow should have the same visible background state as the popover container. With brief exploration, the z-index of `.popover-arrow` should be `11` or greater. 
2. The popover arrow should be tight to the container, as seen in the [Human Interface Guidelines](https://developer.apple.com/design/human-interface-guidelines/popovers). With brief exploration, the fractional units used for the `top` value are causing the arrow to render incorrectly. Flooring the value results in the correct display.

Here is a modified example of the expected visual state:
CleanShot 2023-06-02 at 12 22 25

### Steps to Reproduce

1. Visit the [Side and Alignment Demo](https://ionicframework.com/docs/api/popover#side-and-alignment-demo)
2. With iOS active, Click "Side=Top, Alignment=Center" button
3. Observe: The arrow for the presented popover is not the same background color as the popover container.
4. Observe: The arrow is offset from the container.

### Code Reproduction URL

_No response_

### Ionic Info

N/A

### Additional Information

Tested on Google Chrome: Version 113.0.5672.126 (Official Build) (arm64)

This issue is not captured in our visual regression tests: https://github.com/ionic-team/ionic-framework/blob/main/core/src/components/popover/test/arrow/popover.e2e.ts-snapshots/popover-arrow-ios-ltr-Mobile-Chrome-linux.png

thetaPC:
Issue number: resolves #28985

---------

When a popover is being rendered in iOS mode, the arrow renders under the backdrop.

Screenshot 2024-02-06 at 11 59 58 AM

## What is the new behavior?

- The arrow renders at the same level as the content.

Screenshot 2024-02-06 at 11 59 51 AM

- [ ] Yes
- [x] No

## Other information

Credit [Sahil-B07](https://github.com/Sahil-B07) for providing the [fix](https://github.com/ionic-team/ionic-framework/pull/28430). A new PR had to be created in order to update snapshots.

---------

Co-authored-by: Sahil Bhor <92709590+Sahil-B07@users.noreply.github.com>


core/src/components/popover/popover.ios.scss
```diff
@@ -35,6 +35,7 @@
   height: 10px;
 
   overflow: hidden;
+  z-index: 12;
```
liamdebeasi, line 38: Can we add a comment here describing why this is necessary? Also is there a reason why we specifically chose 12?

core/src/components/popover/popover.ios.scss
```diff
@@ -35,6 +35,13 @@
   height: 10px;
 
   overflow: hidden;
+  /*
+   * Required for the arrow to render above the backdrop.
+   * Otherwise, the arrow will appear slightly transparent.
+   * The value is set to 11 since it's the minimum value that
+   * will allow the arrow to render above the backdrop.
+   */
+  z-index: 11;
```
thetaPC, line 44: I switched it from 12 to 11. I was on the fence between these. I had 12 because it was one value above the minimum. This would allow for one pixel buffer just in case. But the more I thought about it, the more I don't think we need the buffer.

core/src/components/popover/popover.ios.scss
thetaPC: Linking my [comment](https://github.com/ionic-team/ionic-framework/pull/28986#discussion_r1481793254) in this thread since it's related.

liamdebeasi (approved):
Looks good! Be sure to give co-author credit when merging.

Issue #28981, opened 2024-02-05
bug: ion-loading is not removed from dom if never presented


aparajita:
### Prerequisites

- [X] I have read the [Contributing Guidelines](https://github.com/ionic-team/ionic-framework/blob/main/.github/CONTRIBUTING.md#creating-an-issue).
- [X] I agree to follow the [Code of Conduct](https://ionicframework.com/code-of-conduct).
- [X] I have searched for [existing issues](https://github.com/ionic-team/ionic-framework/issues) that already report this problem, without success.

### Ionic Framework Version

v7.x

### Current Behavior

Hey @liamdebeasi, when I dismiss an ion-loading created by a controller that has never been presented, it is not removed from the dom, leading to a fairly large memory leak.

### Expected Behavior

One of two fixes:

- When an ion-loading is dismissed, it should be removed from the dom. This is the ideal.
- Clearly document that if `present()` is not called after `loadingController.create()`, `dismiss()` will be a no-op and there will be a memory leak.

### Steps to Reproduce

1. Create an ion-loading with `loadingController.create()`. An ion-loading is added to the dom.
3. Do not call `present()`.
4. Call `dismiss()`. The ion-loading is not removed from the dom.

### Code Reproduction URL

https://stackblitz.com/edit/a1fkix?file=src%2Fcomponents%2FExample.vue

### Ionic Info

Ionic:

   Ionic CLI : 7.2.0 (/monorepo/node_modules/.pnpm/@ionic+cli@7.2.0/node_modules/@ionic/cli)

Capacitor:

   Capacitor CLI      : 5.6.0
   @capacitor/android : 5.6.0 (/monorepo/node_modules/.pnpm/@capacitor+android@5.6.0_@capacitor+core@5.6.0/node_modules/@capacitor/android)
   @capacitor/core    : 5.6.0 (/monorepo/node_modules/.pnpm/@capacitor+core@5.6.0/node_modules/@capacitor/core)
   @capacitor/ios     : 5.6.0 (/monorepo/node_modules/.pnpm/@capacitor+ios@5.6.0_@capacitor+core@5.6.0/node_modules/@capacitor/ios)

Utility:

   cordova-res : not installed globally
   native-run  : not installed globally

System:

   NodeJS : v21.4.0 (/path/to/Library/pnpm/nodejs/21.4.0/bin/node)
   npm    : 10.2.4
   OS     : macOS Unknown

### Additional Information

Here's the context:

[...]

PR #28989, opened 2024-02-07
docs(overlays): clarify how to remove an overlay

liamdebeasi:
In https://github.com/ionic-team/ionic-framework/issues/28981 there was some confusion surrounding how to remove an overlay from the DOM if it was never presented. The `dismiss` method will remove the overlay from the DOM, but only if the overlay is visible. Otherwise, it's a no-op.

This PR updates the `dismiss` method docs for each overlay component to note that developers can use the browser's remove method to remove the element from the DOM.


thetaPC (approved):
LGTM

PR #28995, opened 2024-02-07
refactor(config): enable stencil's experimental slot fixes

rwaskiewicz:
Issue number: N/A

---------

the stencil team has been working on fixing multiple issues with slots
elements in the stencil code base. to make these changes backwards
compatible and communicate that they were volatile, we added two
configuration flags for these fixes that were prefixed with the word
"experimental".

now that the effort to provide these fixes has largely solidified, the
features behind these flags are slightly less volatile. while the
"experimental" aspect still technically holds true, we've requested
the Framework team to enable these flags in a v8 beta. the stencil
team expects these flags to be set to `true` by default in stencil
v5, which ought to help prepare for future migrations the ionic
framework has to undergo.

Previously, Stencil would allow content to project through to a component even when a slot was not present. However, with the changes in `extras.experimentalScopedSlotChanges`, this behavior was changed to hide elements without a destination slot - matching the behavior of a shadow encapsulated component.

As such, elements projected through the `ion-label` or `ion-buttons` components would no longer be visible in rendered output.

## What is the new behavior?

This commit adds an explicit `slot` tag to components in core leveraging "scoped" encapsulation with no rendered content (i.e. elements with only a `Host` tag and styles).

This fixes an issue with the `ion-input` component not re-rendering in some cases when using the label slot functionality.

HTML element patches in Stencil that are enabled by the `experimentalSlotFixes` flag result in DOM manipulations that won't trigger the current mutation observer configuration and callback.

- [] Yes
- [x] No

## Other information

Would you like us to update the commit message to include the `BREAKING CHANGE:` comment? Unsure of the actual impact on end users here. 

Similarly, would you like us to update the commit message here from `chore()` to something else?


core/src/utils/slot-mutation-controller.ts
```diff
@@ -50,6 +50,7 @@ export const createSlotMutationController = (
 
     hostMutationObserver.observe(el, {
       childList: true,
+      subtree: true,
```
thetaPC, line 53: Let's add a comment explaining why this was adding. That way it's not accidentally removed in the future.

core/stencil.config.ts
```diff
@@ -257,4 +257,8 @@ export const config: Config = {
   globalScript: 'src/global/ionic-global.ts',
   enableCache: true,
   transformAliasedImportPaths: true,
+  extras: {
+    experimentalSlotFixes: true,
+    experimentalScopedSlotChanges: true,
```
thetaPC, line 262: Let's add a comment noting that these are needed until a stable version is released.

thetaPC (changes requested):

thetaPC:
This is not considered a breaking change so no additional work needs to be done for that.

rwaskiewicz:
@thetaPC Thanks! Added those comments in https://github.com/ionic-team/ionic-framework/pull/28995/commits/a85044ae8c5967e51183d57ef440082545e618a1. I'll probably squash and amend the commit once this gets approved (before we merge it).

core/src/utils/slot-mutation-controller.ts
```diff
@@ -50,6 +50,18 @@ export const createSlotMutationController = (
 
     hostMutationObserver.observe(el, {
       childList: true,
+      /**
+       * This fixes an issue with the `ion-input` component
+       * not re-rendering in some cases when using the
+       * label slot functionality.
```
thetaPC, line 56:
```suggestion
       * This fixes an issue with the `ion-input` and
       * `ion-textarea` not re-rendering in some cases
       * when using the label slot functionality.
```

Tiny nitpick.

thetaPC (changes requested):

core/src/utils/slot-mutation-controller.ts
rwaskiewicz: For my own understanding, where/how does `ion-textarea` come into play here? TBH, I just lifted the commit message from https://github.com/ionic-team/ionic-framework/pull/28995/commits/13042029c96a7564d37a7d75ca3cdb6ae6026246. I'm not opposed to this change, just wanna double check that it is/was an issue we were seeing.

PR #28997, opened 2024-02-07
fix(overlays): ensure that only topmost overlay is announced by screen readers

### LINKED ISSUE #23472: bug: Screen readers read content behind overlays if is present more than 1 overlay
# Bug Report

**Ionic version:**
[ ] **4.x**
[x] **5.x**

**Current behavior:**
Screen readers read content behind overlays if is present more than 1 overlay and, after closing the last overlay, the aria-hidden applied on the ion-router-outlet component is removed despite the first modal is already present.

**Expected behavior:**
if I open another overlay (inside an overlay for example) the screen reader doesn't have to read the hidden overlay and if I close the second overlay, the ion-router-outlet  component must have the attribute aria-hidden if an overlay is present

**Steps to reproduce:**

open first overlay -> open second overlay from first overlay

A sample application via GitHub
https://stackblitz.com/edit/ionic-bug-overlays

**Ionic info:**

```
Ionic:

   Ionic CLI                     : 6.16.1 
   Ionic Framework               : @ionic/angular 5.6.9
   @angular-devkit/build-angular : 0.1100.7
   @angular-devkit/schematics    : 11.0.7
   @angular/cli                  : 11.0.7
   @ionic/angular-toolkit        : 3.1.1

Cordova:

   Cordova Platforms : android 9.0.0, ios 5.1.1
   Cordova Plugins   : cordova-plugin-ionic-keyboard 2.1.3, cordova-plugin-ionic-webview 5.0.0, (and 41 other plugins)

Utility:

   native-run (update available: 1.4.0) : 1.2.2

System:

   Android SDK Tools : 26.1.1 
   ios-deploy        : 1.10.0
   ios-sim           : 8.0.2
   NodeJS            : v12.19.0 (/usr/local/bin/node)
   npm               : 6.14.13
   OS                : macOS Big Sur
   Xcode             : Xcode 12.5 Build version 12E262
```

amandaejohnston:
Issue number: resolves #23472

---------

If multiple overlays are presented at the same time, none of them receive `aria-hidden="true"`. This means that screen readers can read contents from overlays behind the current one, which can be confusing for users.

The original issue also reports router outlets getting `aria-hidden` removed when any overlay is dismissed, not just the last one, but we've since fixed that: https://github.com/ionic-team/ionic-framework/blob/35ab6b4816bd627239de8d8b25ce0c86db8c74b4/core/src/utils/overlays.ts#L573-L576

## What is the new behavior?

All overlays besides the topmost one now receive `aria-hidden="true"`. This means that screen readers will only announce the topmost overlay.

- [ ] Yes
- [x] No


core/src/utils/test/overlays/index.html
```diff
@@ -62,7 +62,7 @@
             </ion-toolbar>
           </ion-header>
           <ion-content class="ion-padding">
-            Modal Content
+            Modal ${id}
```
amandaejohnston, line 65: This isn't necessary for the fix, but I found it very useful while debugging since the page lets you create as many modals as you want, and this makes it easier to tell which one you're looking at. I can revert if we'd rather keep things slim.

core/src/utils/test/overlays/overlays.spec.ts
```diff
@@ -5,127 +5,194 @@ import { Nav } from '../../../components/nav/nav';
...
 import { setRootAriaHidden } from '../../overlays';
 
-describe('setRootAriaHidden()', () => {
-  it('should correctly remove and re-add router outlet from accessibility tree', async () => {
-    const page = await newSpecPage({
-      components: [RouterOutlet],
-      html: `
-        <ion-router-outlet></ion-router-outlet>
-      `,
+describe('overlays', () => {
```
amandaejohnston, line 8: The first `describe` block is unchanged; the second one contains my new tests. I wrapped both in a single parent `describe` as a workaround for a Stencil bug discussed here: https://outsystems.slack.com/archives/C06BR3RGLHM/p1707342287102049 We use this pattern in plenty of other spec tests, so I don't see this needing a comment or tech debt ticket, but LMK if you'd like me to add one.

amandaejohnston:
We might have a better fix for the spec test issue. Moving to draft for a sec.

amandaejohnston:
Okay, all good now 👍

sean-perkins (approved):
Tested the changes locally with voice over, works great!

Can we update the PR description to describe the new behavior impact? For example:
> Screen readers will announce only the top-most visible overlay.

Suggestion for the PR title, can we use "announced" language vs. visible?

Issue #28487, opened 2023-11-07
bug: ion-range should not update value or lock scrolling until touchEnd or drag within the bar


cmaas:
### Prerequisites

- [X] I have read the [Contributing Guidelines](https://github.com/ionic-team/ionic-framework/blob/main/.github/CONTRIBUTING.md#creating-an-issue).
- [X] I agree to follow the [Code of Conduct](https://ionicframework.com/code-of-conduct).
- [X] I have searched for [existing issues](https://github.com/ionic-team/ionic-framework/issues) that already report this problem, without success.

### Ionic Framework Version

v7.x

### Current Behavior

If you click/tap on an ion-range component somewhere on the bar (not on the knob), the knob will immediately jump to the place where you clicked or touched down. See video. The circle indicates the beginning of a click (onTouchStart):

![CleanShot 2023-11-07 at 21 05 54](https://github.com/ionic-team/ionic-framework/assets/6094364/829ab74d-a40e-4f98-865a-e1091f16095c)

This is "too sensitive", because it causes accidentally changing the ion-range value if the user just wants to scroll in some cases. In my app, there are several `ion-range` elements further down on the screen that get accidentally changed if the user merely wants to scroll.

On **native** iOS 17 and Android 14 UI elements, the behavior is different: if you touch down on the knob, you can start dragging it. But if you touch down on the bar somewhere else, the knob will move there only "onTouchEnd" when you lift your finger. This reduces the sensitivity of this element and prevents accidents. Furthermore, the value of the element will only be changed if you lift your finger while it's still hovering over it. If you touch down on the bar and move your finger somewhere else while keeping it touched down, the parent view starts to scroll vertically (most of the time at least).

The behavior is pretty consistent between native iOS and Android elements and feels quite off in Ionic.

### Expected Behavior

ion-range should mimic native behavior and trigger value adjustments only "onTouchEnd" and only, if the finger is still over the component. If the user touches down on the bar, but moves the finger to another place on the screen and lifts the finger there, ion-range value should not be changed.

### Steps to Reproduce

[...]

PR #29005, opened 2024-02-08
refactor(range): update value on touchEnd or drag

thetaPC:
Issue number: resolves #

---------

## What is the new behavior?

-
-
-

- [ ] Yes
- [ ] No

### DIFF
--- a/core/src/components/range/range.tsx
+++ b/core/src/components/range/range.tsx
@@ -60,6 +60,7 @@ export class Range implements ComponentInterface {
   private initialContentScrollY = true;
   private originalIonInput?: EventEmitter<RangeChangeEventDetail>;
   private legacyFormController!: LegacyFormController;
+  private isScrolling = false;
 
   // This flag ensures we log the deprecation warning at most once.
   private hasLoggedDeprecationWarning = false;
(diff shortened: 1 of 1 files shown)


core/src/components/range/range.tsx
```diff
@@ -418,42 +424,101 @@ export class Range implements ComponentInterface {
...
+     * The `pressedKnob` can be undefined if the user never
+     * dragged the knob. They just tapped on the bar.
+     *
+     * This is necessary to determine which knob the user is changing,
+     * especially when it's a dual knob.
+     * Plus, it determines when to apply certain styles.
+     */
+    if (this.pressedKnob === undefined) {
+      this.setPressedKnob(currentX);
+    }
```
amandaejohnston, line 504: I'm no longer seeing the `ionKnobMoveStart` event emit when clicking on an empty part of the range to move the knob (without dragging), only `ionKnobMoveEnd`. In `main`, both events emit when clicking (first start, then end).

We might be able to emit the start event here? I assume it's not firing since the gesture never technically started.

core/src/components/range/range.tsx
```diff
@@ -922,6 +1014,7 @@ const renderKnob = (
           ev.stopPropagation();
         }
       }}
+      // onPointerMove={handlePointerMove}
```
amandaejohnston, line 1017: Was this left in intentionally?

amandaejohnston (changes requested):
Great work on this! I tested the behavior against my findings in https://github.com/ionic-team/ionic-framework/issues/28487#issuecomment-1804077261 and everything looks great. Just noticed a thing with the events.

core/src/components/range/range.tsx
thetaPC: Great catch! https://github.com/ionic-team/ionic-framework/pull/29005/commits/3f4e38270582b8417d16e54a661bc93adebedc96

core/src/components/range/range.tsx
amandaejohnston: I'm now seeing the events both fire twice total on release of a drag gesture:

https://github.com/ionic-team/ionic-framework/assets/90629384/04318f35-2b81-4206-8a75-fcb580cf4f9e

thetaPC: Again good catch! https://github.com/ionic-team/ionic-framework/pull/29005/commits/8fe6d79b6fa5eff5823ec24b2ee1c09e44b82c0c

thetaPC: @amandaejohnston ready for review

amandaejohnston: Looks good now, thank you!

amandaejohnston (approved):

PR #29007, opened 2024-02-09
refactor(item): remove build conditional for test environment

gnbm:
- Fix Internal issue (FW-5205) - Removed the build conditional in the `item` component since the previous issue is already fixed in Stencil.

---------

- No changes were made in terms of behavior.

## What is the new behavior?

- No changes were made in terms of behavior so it will work as before.

- [ ] Yes
- [x] No

## Other information

- Spec tests still pass even after this change (we can see below the results before and after the change)
    - Ran the command `npm run test.spec src/components/item/test/item.spec.ts src/components/item/test/item.spec.tsx`
    
image

### DIFF
--- a/core/src/components/item/item.tsx
+++ b/core/src/components/item/item.tsx
@@ -1,5 +1,5 @@
 import type { ComponentInterface } from '@stencil/core';
-import { Build, Component, Element, Host, Listen, Prop, State, Watch, forceUpdate, h } from '@stencil/core';
+import { Component, Element, Host, Listen, Prop, State, Watch, forceUpdate, h } from '@stencil/core';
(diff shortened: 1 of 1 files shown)


sean-perkins:
Thanks @gnbm! We have strict lint rules for unused imports. Removing the `Build` import member from the file should resolve the build issues on this PR 👍

gnbm:
> Thanks @gnbm! We have strict lint rules for unused imports. Removing the `Build` import member from the file should resolve the build issues on this PR 👍

Thanks for the feedback @sean-perkins 💪🏻
Missed that but already committed that and the build seems to be ok now.

sean-perkins (approved):
🎉

PR #29010, opened 2024-02-09
feat: add high contrast themes

liamdebeasi:
:warning: This is a combination of previously approved PRs with the exception of https://github.com/ionic-team/ionic-framework/pull/29010/commits/fe9dca513c143423f95903e7415abe885446444b. This change was made as a result of https://github.com/ionic-team/ionic-framework-design-documents/pull/248.

Issue number: Internal

---------

Users do not have a way of increasing the contrast in Ionic apps. This is valuable for people with low vision as increasing the contrast between foreground and background content helps improve readability.

## What is the new behavior?

- Adds a high contrast light and high contrast dark theme. As with our other themes, developers can choose between system, class, and always stylesheets.

While we aim to improve contrast for text and UI components, this feature prioritizes text in the event that both text and UI component cannot be improved without one negatively impacting the other.

- [ ] Yes
- [x] No

## Other information

Dev build: `7.6.7-dev.11706894781.1cd59fde`

Testing instructions:

1. Open `src/themes/test/css-variables`. Activate the high contrast light and dark themes to verify that contrast does increase.
2. Use the dev build to integrate the theme into a test app (conference app, starter app, etc).

I'd recommend using these imports:

```css
@import "@ionic/angular/css/themes/high-contrast.system.css";
@import "@ionic/angular/css/themes/high-contrast-dark.system.css";
```
Note: Make sure this is imported **after** `core.scss`


amandaejohnston (approved):
Woohoo!

core/src/css/themes/high-contrast.scss
```diff
@@ -0,0 +1,160 @@
...
+
+$primary:         #003FAE !default;
+$secondary:       #01487B !default;
+$tertiary:        #3400e6 !default;
+$success:         #004314 !default;
+$warning:         #5F4100 !default;
+$danger:          #9C000C !default;
+$light:           #f4f5f8 !default;
+$medium:          #444446 !default;
+$dark:            #222428 !default;
```
brandyscarney, line 12:
```suggestion
$primary:         #003fae !default;
$secondary:       #01487b !default;
$tertiary:        #3400e6 !default;
$success:         #004314 !default;
$warning:         #5f4100 !default;
$danger:          #9c000c !default;
$light:           #f4f5f8 !default;
$medium:          #444446 !default;
$dark:            #222428 !default;
```

Optional nit pick but we should keep these all the same case

brandyscarney (approved):
I do have one requested change but it isn't critical so I am approving still.

PR #29011, opened 2024-02-09
refactor: update styles according to design document

liamdebeasi:
This PR makes adjustments to the standard contrast light and dark themes as defined in https://github.com/ionic-team/ionic-framework-design-documents/pull/248.


core/src/css/themes/dark.scss
brandyscarney, line 1: We should auto-generate these like the high contrast theme, but not necessary for this PR 👍

brandyscarney (approved):

PR #29014, opened 2024-02-12
test(toggle): remove legacy tests

liamdebeasi:
This PR removes the tests for the legacy toggle syntax. A separate PR will be used to remove the implementation for the legacy syntax.


core/src/components/item/test/legacy/disabled/index.html
```diff
@@ -59,7 +59,6 @@
 
           <ion-item>
             <ion-label>Disabled Toggle</ion-label>
-            <ion-toggle disabled checked slot="end" legacy="true"></ion-toggle>
           </ion-item>
```
brandyscarney, line 62: Can we remove this entire chunk since it is just a label now?

core/src/components/item/test/legacy/disabled/index.html
```diff
@@ -145,7 +144,6 @@
           <ion-item>
             <ion-checkbox slot="start" legacy="true"></ion-checkbox>
             <ion-label>Checkbox + Toggle</ion-label>
-            <ion-toggle disabled value="45"></ion-toggle>
```
brandyscarney, line 148: Same as other PRs - should we remove this entire item?

brandyscarney (changes requested):

brandyscarney (approved):

PR #29015, opened 2024-02-12
refactor(toggle): remove legacy syntax

liamdebeasi:
This pull request includes the changes to remove the `legacy` property for the toggle as part of https://github.com/ionic-team/ionic-framework/pull/29037. That pull request specifically focuses on updating tests to remove any legacy toggle usage. The internal ticket suggested separating these changes into individual pull requests. Please refer to the mentioned pull request for a detailed description of the combined changes from both pull requests. This will be merged into that pull request upon approval.


core/src/utils/test/aria.spec.ts
```diff
@@ -5,7 +5,8 @@ import { Label } from '../../components/label/label';
 import { Toggle } from '../../components/toggle/toggle';
 import { getAriaLabel } from '../helpers';
 
-describe('getAriaLabel()', () => {
+// TODO FW-5969
+describe.skip('getAriaLabel()', () => {
```
liamdebeasi, line 9: These tests use toggle which are failing now because we removed support for the legacy form syntax. I'm not 100% sure if this utility is still needed, so I made a tech debt ticket to investigate after all of the form controls have been migrate.

PR #29019, opened 2024-02-12
refactor(input): remove legacy property and styling

brandyscarney:
This pull request includes the changes to remove the legacy property for the input as part of https://github.com/ionic-team/ionic-framework/pull/29017. That pull request specifically focuses on updating tests to remove any legacy input usage. The internal ticket suggested separating these changes into individual pull requests. Please refer to the mentioned pull request for a detailed description of the combined changes from both pull requests. This will be merged into that pull request upon approval.


core/src/components/input/input.ios.vars.scss
```diff
@@ -7,17 +7,5 @@
...
-$input-ios-padding-top:                    $item-ios-padding-top !default;
-
-/// @prop - Margin end of the input
-$input-ios-padding-end:                    ($item-ios-padding-end * 0.5) !default;
-
-/// @prop - Margin bottom of the input
-$input-ios-padding-bottom:                 $item-ios-padding-bottom !default;
-
-/// @prop - Margin start of the input
-$input-ios-padding-start:                  0 !default;
```
brandyscarney, line 20: These variables were only used by the legacy input

core/src/components/input/input.md.vars.scss
```diff
@@ -7,18 +7,6 @@
...
-
-/// @prop - Margin end of the input
-$input-md-padding-end: 0 !default;
-
-/// @prop - Margin bottom of the input
-$input-md-padding-bottom: 10px !default;
-
-/// @prop - Margin start of the input
-$input-md-padding-start: ($item-md-padding-start * 0.5) !default;
-
```
brandyscarney, line 21: These variables were only used by the legacy input

core/src/components/input/input.scss
```diff
@@ -257,22 +229,6 @@
...
-:host(.has-focus.legacy-input) {
-  pointer-events: none;
-}
-
-// TODO FW-2764 Remove this
-:host(.has-focus.legacy-input) input,
-:host(.has-focus.legacy-input) a,
-:host(.has-focus.legacy-input) button {
-  pointer-events: auto;
-}
```
liamdebeasi, line 273: Out of curiosity, can the other `:host-context` stuff in this file be removed too?

brandyscarney: Good catch! Removed here and looks like everything is still working properly: https://github.com/ionic-team/ionic-framework/pull/29019/commits/d1a7adc1aba3d544b5d829f548fd63b2ccb29b6f

This should also be removed from `textarea`: https://github.com/ionic-team/ionic-framework/blob/feature-8.0/core/src/components/textarea/textarea.ios.scss#L19-L20

We could just do this when we go through `item` though, I added a note on FW-5289

liamdebeasi (approved):

PR #29020, opened 2024-02-12
test(checkbox): remove legacy tests

liamdebeasi:
This PR removes the tests for the legacy checkbox syntax. A separate PR will be used to remove the implementation for the legacy syntax.


core/src/components/item/test/disabled/index.html
```diff
@@ -82,13 +78,7 @@
...
 
           <ion-item>
-            <ion-checkbox aria-label="Checkbox" disabled slot="start"></ion-checkbox>
-            <ion-input aria-label="Input" placeholder="Disabled Checkbox w/ Input"></ion-input>
-          </ion-item>
-
-          <ion-item>
-            <ion-checkbox aria-label="Checkbox" disabled slot="end"></ion-checkbox>
-            <ion-input aria-label="Input" placeholder="Disabled Checkbox w/ Input"></ion-input>
+            <ion-input aria-label="Input" placeholder="Input"></ion-input>
```
amandaejohnston, line 81: The items changed here were all originally checking multiple different inputs in one item, with one being disabled and the other not. This is lost now that the checkboxes have been removed. If we don't think these checks are still useful, we should remove the items entirely.

core/src/components/item/test/legacy/disabled/index.html
```diff
@@ -86,13 +81,11 @@
           </ion-list-header>
 
           <ion-item>
-            <ion-checkbox disabled slot="start" legacy="true"></ion-checkbox>
-            <ion-input placeholder="Disabled Checkbox w/ Input" legacy="true"></ion-input>
+            <ion-input placeholder="Input" legacy="true"></ion-input>
```
amandaejohnston, line 84: Same thoughts as the comment above.

amandaejohnston (changes requested):

core/src/components/item/test/disabled/index.html
liamdebeasi: Brandy is taking care of this in https://github.com/ionic-team/ionic-framework/pull/29017 which is why I didn't remove the inputs.

sean-perkins (approved):

core/src/components/item/test/disabled/index.html
amandaejohnston: Brandy's PR only covers the `ion-input`s; other items changed here use radios, a range, a toggle, a button, and a searchbar, which this comment still applies to. Even accounting for future PRs that remove the other legacy components, there's still the button and searchbar.

Maybe there should be a separate PR that removes the section entirely so we don't have as many screenshot diffs for what is intended to be a temporary state? Both here and in the other PRs removing legacy component usage from this test.

amandaejohnston: I also noticed that Brandy's PR just removes the relevant items completely. That might be a better alternative; it feels less confusing than leaving the items in a temporary state and changing the labels.

liamdebeasi: Good idea - updated in https://github.com/ionic-team/ionic-framework/pull/29020/commits/19af7b5b5bcb2bb49bcc7186cd122119c3c895a3

core/src/components/item/test/legacy/disabled/index.html
liamdebeasi: https://github.com/ionic-team/ionic-framework/pull/29020/commits/19af7b5b5bcb2bb49bcc7186cd122119c3c895a3

amandaejohnston (approved):

PR #29023, opened 2024-02-12
test(range): remove legacy tests

liamdebeasi:
This PR removes the tests for the legacy range syntax. A separate PR will be used to remove the implementation for the legacy syntax.


core/src/components/item/test/disabled/index.html
```diff
@@ -134,8 +128,7 @@
 
           <ion-item>
             <ion-checkbox slot="start" aria-label="Checkbox"></ion-checkbox>
-            <ion-label>Checkbox + Range</ion-label>
-            <ion-range disabled value="45" aria-label="Range"></ion-range>
```
brandyscarney, line 138: We should probably just remove this entirely since it will be removed with checkbox anyway and it doesn't check what it was testing anymore

core/src/components/item/test/legacy/disabled/index.html
```diff
@@ -138,8 +133,7 @@
 
           <ion-item>
             <ion-checkbox slot="start" legacy="true"></ion-checkbox>
-            <ion-label>Checkbox + Range</ion-label>
-            <ion-range disabled value="45"></ion-range>
+            <ion-label>Checkbox</ion-label>
```
brandyscarney, line 136: Same here - this was testing multiple inputs but won't test that anymore, we should remove entirely

brandyscarney (commented):
I added comments to remove some usage but it's really optional since it will just be removed anyway with checkbox

sean-perkins (approved):

brandyscarney (approved):

PR #29025, opened 2024-02-12
refactor(select): remove legacy property and styling #29019

brandyscarney:
This pull request includes the changes to remove the legacy property for the select as part of https://github.com/ionic-team/ionic-framework/pull/29024. That pull request specifically focuses on updating tests to remove any legacy select usage. The internal ticket suggested separating these changes into individual pull requests. Please refer to the mentioned pull request for a detailed description of the combined changes from both pull requests. This will be merged into that pull request upon approval.


core/src/components/select/select.ios.vars.scss
```diff
@@ -4,18 +4,6 @@
...
-
-/// @prop - Padding end of the select
-$select-ios-padding-end:            ($item-ios-padding-end * 0.5) !default;
-
-/// @prop - Padding bottom of the select
-$select-ios-padding-bottom:         $item-ios-padding-bottom !default;
-
-/// @prop - Padding start of the select
-$select-ios-padding-start:          $item-ios-padding-start !default;
-
```
brandyscarney, line 18: These variables were only used by the legacy select

core/src/components/select/select.tsx
```diff
@@ -1138,9 +921,86 @@ Developers can use the "legacy" property to continue using the legacy form marku
   }
 
   render() {
```
brandyscarney, line 923: I didn't change the `renderSelect()` function but I moved it down in the file to where the `render()` function was which is why there is a larger diff

liamdebeasi (approved):

PR #29029, opened 2024-02-13
refactor(angular): apply range to numeric value accessor

sean-perkins:
Issue number: Internal

---------

The ticket describes renaming `TextValueAccessorDirective` since `ion-range` is not a text-based control, however I think this was an incorrect assumption made during the original implementation. 

`ion-range` is a numeric based value accessor (either as a single number or an object accepting a numeric start/end value). 

## What is the new behavior?

- Migrates the usage of `ion-range` value accessor implementation to the `NumericValueAccessorDirective`
- Adds tests for validating the value accessor is functioning as expected

- [ ] Yes
- [x] No


packages/angular/test/base/src/app/lazy/inputs/inputs.component.html
```diff
@@ -99,6 +99,12 @@
       <ion-note slot="end">{{radio}}</ion-note>
     </ion-item>
 
+    <ion-item>
+      <ion-range [(ngModel)]="range" min="0" max="100" id="first-range">
```
thetaPC, line 103: Would there be a benefit to also having a range mirror like how the others are structured?

thetaPC (approved):
LGTM, didn't see any issues locally

packages/angular/test/base/src/app/lazy/inputs/inputs.component.html
sean-perkins: What functionality is the approach with mirroring/duplicating the implementation validating?

thetaPC: From what I can tell, it's to validate that the second component can update dynamically. We should probably check with the team of the reasoning behind the mirroring since it's before my time.

sean-perkins: Took another look with Liam, the mirror appears to be used when interacting with the component (e.g. if the test was dragging the range knob) and then reflecting that state to a clone of that component. 

However here we are not interacting with the component, we are checking if assigning the form control value results in the value updating. 

I think we are good to avoid the range mirror here and re-evaluate if we add more interaction based tests in the future.

PR #29032, opened 2024-02-13
chore(datetime): remove safari 14 intersection observer workaround

sean-perkins:
Issue number: Internal

---------

Ionic Framework has a patch implementation to workaround a Safari 14 bug that resulted in the calendar month jumping when opening/closing the month/year interface. 

This bug was fixed in Safari 15.

## What is the new behavior?

- Ionic v8 no longer supports Safari 14, so we can safely remove this patch code
- Calendar month does not jump when opening/closing the month/year interface with Safari/iOS 15-17

- [ ] Yes
- [x] No

## Other information

I tested on Sims for iOS 15-17. Tested against a physical device for iOS 17 as well.

### DIFF
--- a/core/src/components/datetime/datetime.scss
+++ b/core/src/components/datetime/datetime.scss
@@ -121,39 +121,6 @@
 }
 
 /**
- * Safari 14 has an issue where a scroll event
- * is incorrectly fired when unhiding the calendar content.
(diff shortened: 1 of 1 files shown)


liamdebeasi (approved):
Tested on iOS 15-17 and works well

PR #29035, opened 2024-02-13
refactor(datetime): remove safari 14 hack

liamdebeasi:
Issue number: Internal

---------

This code was added to account for a Safari 14 issue that would cause datetime to render incorrectly. The Safari bug was fixed starting in Safari 15. However, Ionic v7 supported back to Safari 14.

More context is available [here](https://github.com/ionic-team/ionic-framework/pull/24421#discussion_r796113907)

## What is the new behavior?

- Removed the Safari 14 workarounds since Ionic v8 does not support Safari 14

I tested on iOS 15, 16, and 17 to verify that removing this hack does not cause any issues. Also verified that removing this hack does reproduce the issue on iOS 14 (to verify that I am reproducing the original issue correctly) 

- [ ] Yes
- [x] No

### DIFF
--- a/core/src/components/datetime/datetime.tsx
+++ b/core/src/components/datetime/datetime.tsx
@@ -1662,14 +1662,6 @@ export class Datetime implements ComponentInterface {
         disabled={disabled}
         value={todayString}
         onIonChange={(ev: CustomEvent) => {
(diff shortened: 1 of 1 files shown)


sean-perkins (approved):
Tested against Sims for iOS 15-17 and against a physical device for iOS 17. Changes look good 👍

PR #29037, opened 2024-02-13
refactor(toggle): remove legacy property and support for legacy syntax

liamdebeasi:
Issue number: internal

---------

In Ionic Framework v7, we [simplified the toggle syntax](https://ionic.io/blog/ionic-7-is-here#simplified-form-control-syntax) so that it was no longer required to be placed inside of an `ion-item`. We maintained backwards compatibility by adding a `legacy` property which allowed it to continue to be styled properly when written in the following way:

```html
<ion-item>
  <ion-label>Label</ion-label>
  <ion-toggle></ion-toggle>
</ion-item>
```

While this was supported in v7, console warnings were logged to notify developers that they needed to update this syntax for the best accessibility experience.

## What is the new behavior?

- Removes the `legacy` property and support for the legacy syntax. Developers should follow the [migration guide](https://ionicframework.com/docs/api/toggle#migrating-from-legacy-toggle-syntax) in the toggle documentation to update their apps. The new syntax requires a `label` or `aria-label` on `ion-toggle `:
    ```html
    <ion-item>
      <ion-toggle>Label</ion-toggle>
    </ion-item>
    ```
- Removes the legacy tests under `toggle/test/legacy/` and all related screenshots
- Removes the toggle usage in `item/test/disabled`, `item/test/legacy/disabled`

- [x] Yes
- [ ] No

1. Developers have had console warnings when using the legacy syntax since the v7 release. The migration guide for the new toggle syntax is outlined in the [Toggle documentation](https://ionicframework.com/docs/api/toggle#migrating-from-legacy-toggle-syntax).
2. This change has been documented in the Breaking Changes document with a link to the migration guide. 

BREAKING CHANGE:

The `legacy` property and support for the legacy syntax, which involved placing an `ion-toggle` inside of an `ion-item` with an `ion-label`, have been removed from toggle. For more information on migrating from the legacy toggle syntax, refer to the [Toggle documentation](https://ionicframework.com/docs/api/toggle#migrating-from-legacy-toggle-syntax).


core/src/components/item/test/disabled/index.html
```diff
@@ -85,7 +81,6 @@
 
           <ion-item>
             <ion-checkbox aria-label="Checkbox" slot="start"></ion-checkbox>
-            <ion-toggle disabled value="45">Checkbox + Toggle</ion-toggle>
```
amandaejohnston, line 88: Could we remove the whole item here instead?

amandaejohnston (approved):

core/src/components/item/test/disabled/index.html
liamdebeasi: I was planning on removing this in my checkbox PR if that's ok? (The checkbox work is blocked by the toggle work)

amandaejohnston: That works 👍

packages/angular/src/directives/proxies.ts
brandyscarney, line 1: We also need to remove it here I believe: https://github.com/ionic-team/ionic-framework/blob/FW-2990/packages/angular/standalone/src/directives/toggle.ts#L26

brandyscarney (changes requested):

packages/angular/src/directives/proxies.ts
liamdebeasi: https://github.com/ionic-team/ionic-framework/pull/29037/commits/c8f520bea56bfe399f3cae77734d3b36af50965b

brandyscarney (approved):

PR #29040, opened 2024-02-13
refactor(range): remove legacy property and support for legacy syntax

liamdebeasi:
Issue number: internal

---------

In Ionic Framework v7, we [simplified the range syntax](https://ionic.io/blog/ionic-7-is-here#simplified-form-control-syntax) so that it was no longer required to be placed inside of an `ion-item`. We maintained backwards compatibility by adding a `legacy` property which allowed it to continue to be styled properly when written in the following way:

```html
<ion-item>
  <ion-label>Label</ion-label>
  <ion-range></ion-range>
</ion-item>
```

While this was supported in v7, console warnings were logged to notify developers that they needed to update this syntax for the best accessibility experience.

## What is the new behavior?

- Removes the `legacy` property and support for the legacy syntax. Developers should follow the [migration guide](https://ionicframework.com/docs/api/range#migrating-from-legacy-range-syntax) in the range documentation to update their apps. The new syntax requires a `label` or `aria-label` on `ion-range `:
    ```html
    <ion-item>
      <ion-range label="Label"></ion-range>
    </ion-item>
    ```
    
- Alternatively, the "label" slot can be used for developer who require custom HTML labels:
  ```html

  <ion-item>
    <ion-range>
      <div slot="label">Label</div>
    </ion-range>
  </ion-item>
  ```
- Removes the legacy tests under `range/test/legacy/` and all related screenshots
- Removes the range usage in `item/test/disabled`, `item/test/legacy/disabled`

- [x] Yes
- [ ] No

1. Developers have had console warnings when using the legacy syntax since the v7 release. The migration guide for the new range syntax is outlined in the [Range documentation](https://ionicframework.com/docs/api/range#migrating-from-legacy-range-syntax).
2. This change has been documented in the Breaking Changes document with a link to the migration guide. 

BREAKING CHANGE:

The `legacy` property and support for the legacy syntax, which involved placing an `ion-range` inside of an `ion-item` with an `ion-label`, have been removed from range. For more information on migrating from the legacy range syntax, refer to the [Range documentation](https://ionicframework.com/docs/api/range#migrating-from-legacy-range-syntax).


brandyscarney (changes requested):
Do we need to update this: https://github.com/ionic-team/ionic-framework/blob/5bb0e8f9f709a7067a067d2527db1eec2f5f8303/core/src/components/range/test/range-events.e2e.ts#L74

core/src/components/range/range.tsx
```diff
@@ -861,8 +767,7 @@ Developers can dismiss this warning by removing their usage of the "legacy" prop
   }
 
   render() {
-    const { legacyFormController } = this;
-    return legacyFormController.hasLegacyControl() ? this.renderLegacyRange() : this.renderRange();
+    return this.renderRange();
```
brandyscarney, line 770: Can we just move the `renderRange` function into `render` instead?

liamdebeasi:
I updated the legacy path in https://github.com/ionic-team/ionic-framework/pull/29040/commits/5cdedc93848b401e981d6e956d1536ad53535436. However, this test is skipped so there should be no noticeable difference on CI right now.

core/src/components/range/range.tsx
liamdebeasi: Fixed in https://github.com/ionic-team/ionic-framework/pull/29040/commits/a65ec85a4d04a6ac46f66633f86ce792dead1e05

brandyscarney (approved):

PR #29055, opened 2024-02-14
refactor(item): remove deprecated highlight variables used with legacy form controls

brandyscarney:
Issue number: internal

---------

In Ionic Framework v7, we [simplified the form control syntax](https://ionic.io/blog/ionic-7-is-here#simplified-form-control-syntax), eliminating the requirement to place form controls inside of an `ion-item`. We ensured backwards compatibility by introducing a `legacy` property for the form controls and keeping, but deprecating, the following CSS variables on item:

```css
--highlight-color-focused
--highlight-color-invalid
--highlight-color-valid
--highlight-height
```

While this was supported in v7, console warnings were logged to notify developers that they needed to update to the modern syntax if they were using form controls in an item for the best accessibility experience.

## What is the new behavior?

Removes the `--highlight-color-focused`,  `--highlight-color-invalid`,  `--highlight-color-valid`,  and `--highlight-height` variables from item.

- [x] Yes
- [ ] No

Steps taken to mitigate this breaking change: 
1. Developers have had console warnings when using the legacy syntax since the v7 release and the variables were marked as `deprecated`. 
2. The removal of these CSS variables has been documented in the Breaking Changes document with a link to the migration guides for the affected form controls. 

BREAKING CHANGE:

The following CSS variables have been removed from item: `--highlight-height`, `--highlight-color-focused`, `--highlight-color-valid`, and `--highlight-color-invalid`. These variables were used on the bottom border highlight of an item when the form control inside of that item was focused. The form control syntax was [simplified in v7](https://ionic.io/blog/ionic-7-is-here#simplified-form-control-syntax) so that inputs, selects, and textareas would no longer be required to be used inside of an item. 

If you have not yet migrated to the modern form control syntax, migration guides for each of the form controls that added a highlight to item can be found below:
- [Input migration documentation](https://ionicframework.com/docs/api/input#migrating-from-legacy-input-syntax)
- [Select migration documentation](https://ionicframework.com/docs/api/select#migrating-from-legacy-select-syntax)
- [Textarea migration documentation](https://ionicframework.com/docs/api/textarea#migrating-from-legacy-textarea-syntax)

The highlight variables should then be moved from the item to the form control:

```diff
- ion-item {
+ ion-input,
+ ion-textarea,
+ ion-select {
  --highlight-color-focused: purple;
  --highlight-color-valid: blue;
  --highlight-color-invalid: orange;
  --highlight-height: 6px;
}
```

> [!NOTE]
> The input and textarea components are scoped, which means they will automatically scope their CSS by appending each of the styles with an additional class at runtime. Overriding scoped selectors in CSS requires a [higher specificity](https://developer.mozilla.org/en-US/docs/Web/CSS/Specificity) selector. Targeting the `ion-input` or `ion-textarea` for customization will not work; therefore we recommend adding a class and customizing it that way.


BREAKING.md
```diff
@@ -162,6 +163,51 @@ For more information on the dynamic font, refer to the [Dynamic Font Scaling doc
...
+will instead be applied on the form controls:
+
+```css
+ion-input,
+ion-textarea,
+ion-select {
+  --highlight-color-focused: purple;
+  --highlight-color-valid: blue;
+  --highlight-color-invalid: orange;
+  --highlight-height: 6px;
```
brandyscarney, line 204: Note: this will not currently work due to `--highlight-height` not being migrated to the new form controls. I will be updating the design doc to get this added.

amandaejohnston (approved):

brandyscarney:
Requesting a new review because I found some places where `item-highlight` and `item-inner-highlight` weren't removed from the styles.

amandaejohnston (approved):

BREAKING.md
brandyscarney: Added by https://github.com/ionic-team/ionic-framework/commit/13b7f8ac3ab3b8d436400993d9a7c62d1670f475

PR #29058, opened 2024-02-15
feat(react): type check component props for useIonModal, useIonPopover

### LINKED ISSUE #28680: feat: Improved types for react useIonModal

aeharding:
Issue number: resolves #28680

---------

useIonModal and useIonPopover do not type check component parameters. This can cause bugs in applications that heavily use these hooks.

## What is the new behavior?

- useIonModal and useIonPopover have new types
- If a component does not have props, second parameter can be omitted (`extends undefined`)
- Passing a JSX Element keeps old functionality (`any` for props)

- [ ] Yes
- [X] No

This may be considered a breaking change, because typescript will now complain if component props do not match up. The majority of my `useIonModal` usages required tweaks.


aeharding:
I took a look at the sample app, and found out that the typescript issue can be fixed by simply upgrading typescript.

I'm not sure if there's any minimum required typescript version for Ionic, so any guidance is appreciated!

packages/react/src/hooks/useIonModal.ts
```diff
@@ -8,14 +8,18 @@ import type { ReactComponentOrElement } from '../models/ReactComponentOrElement'
...
  * @param component The component that the modal will show. Can be a React Component, a functional component, or a JSX Element
  * @param componentProps The props that will be passed to the component, if required
  * @returns Returns the present and dismiss methods in an array
  */
+export function useIonModal(component: JSX.Element, componentProps?: any): UseIonModalResult;
+export function useIonModal<P extends undefined>(component: React.ComponentClass<P> | React.FC<P>): UseIonModalResult;
+export function useIonModal<P extends Record<string, never>>(
+  component: React.ComponentClass<P> | React.FC<P>
+): UseIonModalResult;
+export function useIonModal<P>(component: React.ComponentClass<P> | React.FC<P>, componentProps: P): UseIonModalResult;
```
sean-perkins, line 22: Curious on this approach vs. something similar to:

```ts
export function useIonModal<Props = any>(component: JSX.Element | React.ComponentClass<Props> | React.FC<Props>, componentProps?: Props): UseIonModalResult {
```

Appears to accomplish the same level of type inference & type checking.

aeharding: The main reason I did the way I did is so that Typescript requires the second parameter if the component has props.

With your snippet, typescript does not complain if you do not pass props to a component that requires props.

sean-perkins: Perfect, thanks for the quick response 👍 

I'll chat with Liam on Monday more, but the high chance of type errors in an existing project from this change, makes me lean towards a major - even though this is the correct level of strictness for the implementation. 

Fortunately we are in beta.0 right now, so that means a much shorter window to a release.

packages/react/src/hooks/useIonModal.ts
```diff
@@ -8,14 +8,18 @@ import type { ReactComponentOrElement } from '../models/ReactComponentOrElement'
...
 
-// TODO(FW-2959): types
-
 /**
  * A hook for presenting/dismissing an IonModal component
  * @param component The component that the modal will show. Can be a React Component, a functional component, or a JSX Element
  * @param componentProps The props that will be passed to the component, if required
  * @returns Returns the present and dismiss methods in an array
  */
+export function useIonModal(component: JSX.Element, componentProps?: any): UseIonModalResult;
```
aeharding, line 17: I wasn't sure on this - does it make sense to allow passing props to a `JSX.Element`? I don't think so, but I didn't want to break apps if there is some behavior I am unaware of.

If `JSX.Element` should never take `componentProps`, then I can update to remove the second parameter from this definition.

liamdebeasi:
Ok all set: https://github.com/ionic-team/ionic-framework/pull/29158

PR #29063, opened 2024-02-15
feat(textarea): add --highlight-height variable

brandyscarney:
Issue number: internal

---------

In v7, using the legacy syntax, the height of the highlight on an item could be adjusted using the `--highlight-height` variable. This variable was not added to textarea and therefore would not work using the modern syntax. 

## What is the new behavior?
Adds the `--highlight-height` variable so that developers can customize this height in `md` mode. Since the highlight element is not added for `ios` mode, this variable won't do anything for `ios`. Note that this diverges from the v7 behavior, where setting `--highlight-height` enabled the highlight for `ios`. A design document outlining this has been proposed here: https://github.com/ionic-team/ionic-framework-design-documents/pull/252 

- [ ] Yes
- [x] No


core/src/components/textarea/textarea.ios.scss
```diff
@@ -12,6 +12,7 @@
   --padding-end: 0px;
   --padding-bottom: #{$textarea-padding-bottom};
   --padding-start: 0px;
+  --highlight-height: 0px;
```
brandyscarney, line 15: This doesn't need to be defined since the highlight is defined in `md` but I kept it here in case we ever move the highlight to the base file.

sean-perkins (approved):

PR #29059, opened 2024-02-15
feat(datetime-button): formatOptions for Datetime Button

mapsandapps:
Issue number: Internal

---------

## What is the current behavior?

The Datetime Button has default date formatting that cannot be set by the developer.

## What is the new behavior?

- The developer can customize the date and time formatting for the Datetime Button
- A warning will be logged if they do not include the `date` or `time` object for formatOptions as needed for the presentation of the Datetime

## Does this introduce a breaking change?

- [ ] Yes
- [x] No

PR #29065, opened 2024-02-15
feat(datetime): formatOptions property for Datetime

mapsandapps:
Issue number: Internal

---------

The Datetime header, Datetime time button, and Datetime Button have default date formatting that cannot be set by the developer.

## What is the new behavior?

- The developer can customize the date and time formatting for the Datetime header and time button
- The developer can customize the date and time formatting for the Datetime Button
- A warning will appear in the console if they try to provide a time zone (the time zone will not get used)
- A warning will be logged if they do not include the date or time object for formatOptions as needed for the presentation of the Datetime

- [ ] Yes
- [x] No

## Other information

These changes have been reviewed in #29009 and #29059. This PR just adds them to the feature branch now that the separate tickets are complete.

### DIFF
--- a/core/src/components/datetime/utils/format.ts
+++ b/core/src/components/datetime/utils/format.ts
@@ -17,19 +17,19 @@ const getFormattedDayPeriod = (dayPeriod?: string) => {
  * confusion.
  */
 export const stripTimeZone = (formatOptions: Intl.DateTimeFormatOptions): Intl.DateTimeFormatOptions => {
-  /**
-   * We do not want to display the time zone name
-   */
-  delete formatOptions.timeZoneName;
-
-  /**
-   * Setting the time zone to UTC ensures that the value shown is always the
-   * same as what was selected and safeguards against older Safari bugs with
-   * Intl.DateTimeFormat.
-   */
-  formatOptions.timeZone = 'UTC';
-
-  return formatOptions;
+  return {
+    ...formatOptions,
+    /**
+     * Setting the time zone to UTC ensures that the value shown is always the
+     * same as what was selected and safeguards against older Safari bugs with
+     * Intl.DateTimeFormat.
+     */
+    timeZone: 'UTC',
+    /**
+     * We do not want to display the time zone name
+     */
+    timeZoneName: undefined,
+  };
 };
 
 export const getLocalizedTime = (
--- a/core/src/components/datetime/utils/format.ts
+++ b/core/src/components/datetime/utils/format.ts
@@ -11,7 +11,33 @@ const getFormattedDayPeriod = (dayPeriod?: string) => {
   return dayPeriod.toUpperCase();
 };
 
-export const getLocalizedTime = (locale: string, refParts: DatetimeParts, hourCycle: DatetimeHourCycle): string => {
+/**
+ * Including time zone options may lead to the rendered text showing a
+ * different time from what was selected in the Datetime, which could cause
+ * confusion.
+ */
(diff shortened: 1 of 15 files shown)


amandaejohnston (approved):

core/src/components/datetime/utils/format.ts
```diff
@@ -11,7 +11,33 @@ const getFormattedDayPeriod = (dayPeriod?: string) => {
...
+/**
+ * Including time zone options may lead to the rendered text showing a
+ * different time from what was selected in the Datetime, which could cause
+ * confusion.
+ */
+export const stripTimeZone = (formatOptions: Intl.DateTimeFormatOptions): Intl.DateTimeFormatOptions => {
+  /**
+   * We do not want to display the time zone name
+   */
+  delete formatOptions.timeZoneName;
```
sean-perkins, line 23: I do not think we should be using `delete` here. 

This will mutate the original reference passed into the function. 

For example, check the console logs in this example: https://codepen.io/sean-perkins/pen/BabMMBK.

You will see that after, the object passed into the helper function no longer reflects it's original value.

Instead, we should be creating a new object _without_ the timezone.

```ts
return {
  ...formatOptions
  timeZone: 'UTC',
  timeZoneName: undefined
}
```

sean-perkins (changes requested):

core/src/components/datetime/utils/format.ts
mapsandapps: fixed in d9239e1b56

core/src/components/datetime/utils/warn.ts
```diff
@@ -0,0 +1,52 @@
+import { printIonWarning } from '@utils/logging';
```
sean-perkins, line 1: ~ Optional: I think `validate.ts` would be a better name for what this file is responsible for and the functions within it. 

"Warn" is what is happening as a side-effect of validating the data coming in.

core/src/components/datetime/utils/warn.ts
```diff
@@ -0,0 +1,52 @@
...
+ * confusion.
+ */
+export const warnIfTimeZoneProvided = (formatOptions?: FormatOptions) => {
+  if (
+    formatOptions?.date?.timeZone ||
+    formatOptions?.date?.timeZoneName ||
+    formatOptions?.time?.timeZone ||
+    formatOptions?.time?.timeZoneName
+  ) {
+    printIonWarning('Datetime: "timeZone" and "timeZoneName" are not supported in "formatOptions".');
```
sean-perkins, line 17: In most places that we use `printIonWarning`, we additionally log the element reference that is associated with the warning. This helps developers know which element is throwing the warning, when they have multiple on the view. 

I think it'd be worth passing the element reference into these functions (all in this file) and log the element reference as the second argument:

```ts
printIonWarning("Your message", el);
```

sean-perkins (changes requested):
Feature works great, nice work!

Just a few clean-up items listed. 

I think an example in one of our HTML templates that demonstrates the format options feature could be useful as well (doesn't need to be connected to a playwright test).

core/src/components/datetime/utils/warn.ts
mapsandapps: updated in 85a7d73763

core/src/components/datetime/utils/warn.ts
mapsandapps: updated in 85a7d73763

also added the HTML templates in 016b60a7c2

sean-perkins (approved):
Great work 🎉

PR #29068, opened 2024-02-16
fix(popover): adjust position to account for approximate safe area

### LINKED ISSUE #28411: feat: make it possible to add a safe-area for select-popovers
### Prerequisites

### Describe the Use Case

image

### Describe Alternatives

### Related Code

_No response_

### Additional Information

_No response_

amandaejohnston:
Issue number: resolves #28411

---------

Popover position does not account for safe area. iOS has some existing logic that attempts to do so, but it always uses a hardcoded value of 25px for the margin (or 0 for `size="cover"` popovers).

## What is the new behavior?

The most straightforward way of accounting for safe area would be to use `window.getComputedStyle()` to calculate the safe area margins when the popover's position is being determined. However, this function is very expensive, so using it would introduce performance issues. Until a better strategy can be found that properly accounts for all cases, this PR approximates things by making a guess at the safe area margins.

Note that the position of the arrow on iOS popovers has not been adjusted to account for safe area in the same way as the popover content. I chose to leave it like this because the behavior in `main` doesn't adjust the arrow either, even with the 25px guess at the safe area margin. Moving the arrow will also add complexity to account for the different values of `side`, which increases the risk of bugs. If it does need fixing, it would probably be best to do that as separate scope.

This PR has been targeted to a minor release to de-risk the behavior change.

The easiest way to test the behavior is with the `adjustment` test, as this lets you present the popover from anywhere on the screen. Checking the new "show safe area approximation" checkbox highlights the safe area margin guesses in blue.

- [ ] Yes
- [x] No


core/src/components/popover/animations/ios.enter.ts
```diff
@@ -122,20 +109,8 @@ export const iosEnterAnimation = (baseEl: HTMLElement, opts?: any): Animation =>
...
-      const safeAreaRight = ' - var(--ion-safe-area-right, 0)';
-
-      let leftValue = `${left}px`;
-
-      if (checkSafeAreaLeft) {
-        leftValue = `${left}px${safeAreaLeft}`;
-      }
-      if (checkSafeAreaRight) {
-        leftValue = `${left}px${safeAreaRight}`;
-      }
```
amandaejohnston, line 135: While removing these looks like a step backwards at a glance, they weren't actually working as intended because the `checkSafeAreaLeft/Right` variables were based on the hardcoded 0/25px guess at the margins. I've removed them to ensure the MD and iOS safe area behavior lines up as closely as possible.

thetaPC:
@amandaejohnston is this ready for review? I noticed that a screenshot test is failing.

amandaejohnston:
@thetaPC Fixed the screenshots 👍 This is indeed ready for review.

core/src/components/popover/utils.ts
```diff
@@ -837,32 +837,72 @@ export const calculateWindowAdjustment = (
...
+    /**
+     * We only want to check horizontal safe area on landscape.
+     * Most devices do not have horizontal safe area margins in
+     * portrait mode, so enforcing it would lead to popovers
+     * being misaligned with the trigger when we don't want them
+     * to move.
+     */
+    if (win.matchMedia('(orientation: landscape)').matches) {
+      horizontalSafeAreaApprox = win.innerWidth * 0.05;
+    }
```
amandaejohnston, line 865: I'm not really sure about this. I wanted to cover phones rotated into landscape mode, but it also causes the horizontal safe area to kick in on desktop, and possibly tablets when we don't want it to. I just couldn't think of a better way to cover things that doesn't make too many assumptions about more specific screen sizes.

One tweak that seemed promising is to add a max value for the margins, to avoid them being too large on wider devices, but I'm not sure if that's a route we'd want to take or what the max value should be.

Coming back to this PR after a while, I would be open to closing it until we can investigate a more robust solution, possibly re-opening it later if we don't determine the in-progress solution to be solidly feasible.

core/src/components/popover/test/adjustment/popover.e2e.ts
```diff
@@ -33,5 +33,45 @@ configs({ modes: ['ios'], directions: ['ltr'] }).forEach(({ title, config }) =>
...
+
+      const popoverContent = page.locator('ion-popover .popover-content');
+      const box = (await popoverContent.boundingBox())!;
+
+      /**
+       * The safe area approximation should move the y position by 5%
+       * of the screen height. We use 10px as the threshold to give
+       * wiggle room and help prevent flakiness.
+       */
+      expect(box.x < 10).toBe(true);
```
liamdebeasi, line 52: These should be screenshot tests according to our [best practices guide](https://github.com/ionic-team/ionic-framework/blob/main/core/src/utils/test/playwright/docs/best-practices.md#avoid-tests-that-compare-computed-values).

liamdebeasi, line 52: These should be screenshot tests according to our [best practices guide](https://github.com/ionic-team/ionic-framework/blob/main/core/src/utils/test/playwright/docs/best-practices.md#avoid-tests-that-compare-computed-values).

(same as below test)

core/src/components/popover/test/arrow/popover.e2e.ts-snapshots/popover-arrow-ios-ltr-Mobile-Chrome-linux.png
liamdebeasi, line 1: This doesn't look correct. In the old screenshot the arrow is aligned with the popover body for the top-most popover. That's no longer true in the new screenshot.

core/src/components/popover/animations/ios.enter.ts
liamdebeasi: So I understand: Since we are approximating the safe area bounds we no longer need these checks since it's already handled elsewhere?

core/src/components/popover/utils.ts
```diff
@@ -837,32 +837,72 @@ export const calculateWindowAdjustment = (
...
+   *
+   * 5% is derived from the iPhone 14 top safe area margin (47pt / 844 pt ~= 0.05).
+   * Source: https://useyourloaf.com/blog/iphone-14-screen-sizes/
+   *
+   * TODO(FW-5982): Investigate a more robust solution that uses the actual
+   * safe area margins through alternate means.
+   */
+  let horizontalSafeAreaApprox = 0;
+  let verticalSafeAreaApprox = 0;
+  if (win?.matchMedia !== undefined) {
```
liamdebeasi, line 853: `matchMedia` should always be defined in the [supported browsers](https://developer.mozilla.org/en-US/docs/Web/API/Window/matchMedia). Are there other scenarios where this is not defined?

liamdebeasi (changes requested):

core/src/components/popover/utils.ts
liamdebeasi: I think this approach makes sense. The top/bottom safe areas always apply, but the left/right areas generally only apply for landscape. This approach is trying to get us 90% of the way there, so I think it's fair that this may not be perfectly accurate in all scenarios.

core/src/components/popover/utils.ts
```diff
@@ -837,32 +837,72 @@ export const calculateWindowAdjustment = (
...
     originX = 'right';
   }
 
+  /**
+   * Ensure the popover doesn't sit above the safe area approxmation.
+   * If popover is on the left or right of the trigger, we should not
+   * adjust top margins.
+   */
+  if (side === 'top' || side === 'bottom') {
+    top = Math.max(top, verticalSafeAreaApprox + bodyPadding);
```
liamdebeasi, line 892: Can we write a test for this?

amandaejohnston:
Closing for now due to shifting priorities. This may be reopened later after we've had a chance to discuss the strategy internally.

core/src/components/popover/animations/ios.enter.ts
amandaejohnston: Right; the adjustment is all done in `calculateWindowAdjustment` now.

core/src/components/popover/utils.ts
amandaejohnston: Met with the team; we agreed to proceed with this approach, but add an undocumented escape hatch that devs can use to revert to the old behavior if they run into problems.

Webrow:
I am unsure of how to read this thread with the vercel bot inbetween. 
Has this been merged and released yet? Or is this still waiting for steps. 
We are currently on 8.2.7. and still have the bottom and top of the popover list "falling off the screen"

thetaPC:
Hello @Webrow!

The issue still exists within Ionic Framework. I would recommend subscribing to its GitHub [issue](https://github.com/ionic-team/ionic-framework/issues/28411) to receive updates.

ShaneK:
This issue was resolved in https://github.com/ionic-team/ionic-framework/pull/30949, which is now out in Ionic Framework v8.7.18!

PR #29074, opened 2024-02-21
fix(core): add `slot` elements to scoped "wrapper" components

tanner-reits:
Issue number: N/A

---------

Previously, Stencil would allow content to project through to a component even when a slot was not present. However, with the changes in `extras.experimentalScopedSlotChanges`, this behavior was changed to hide elements without a destination slot - matching the behavior of a shadow encapsulated component.

As such, elements projected through the `ion-label` or `ion-buttons` components would no longer be visible in rendered output.

## What is the new behavior?

This commit adds an explicit `slot` tag to components in core leveraging "scoped" encapsulation with no rendered content (i.e. elements with only a `Host` tag and styles).

- [ ] Yes
- [X] No

### DIFF
--- a/core/src/components/buttons/buttons.tsx
+++ b/core/src/components/buttons/buttons.tsx
@@ -34,7 +34,9 @@ export class Buttons implements ComponentInterface {
           [mode]: true,
           ['buttons-collapse']: this.collapse,
         }}
-      ></Host>
+      >
+        <slot></slot>
+      </Host>
     );
   }
 }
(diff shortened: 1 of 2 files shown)


rwaskiewicz:
@tanner-reits should we just merge this into the current target (my branch) and update my PR? Alternatively, I think you should be able to cherry-pick my commit onto your PR now that you have write permissions. Either way, I think that'll help give the Framework team a little more context as to what we're doing/that we're targeting a feature branch for both of our changes here. WDYT?

tanner-reits:
@rwaskiewicz I'll go ahead and merge this (don't know why I didn't do that earlier)

PR #29078, opened 2024-02-22
chore(ci): add `--debug --verbose` to stencil nightly build

alicewriteswrongs:
Set the `--debug` and `--verbose` flags on the Stencil Nightly CI build.

The Stencil nightly build doesn't provide all the information that it could!

## What is the new behavior?

This sets the `--debug` and `--verbose` flags when calling `npm run build` in the 'Build Ionic Core with Stencil Prerelease' github action, which is used by the workflow which installs nightly Stencil builds and builds framework with it, thereby checking for regressions.

This change will just ensure that this nightly build provides a bit more information.

- [ ] Yes
- [x] No

### DIFF
--- a/.github/workflows/actions/build-core-stencil-prerelease/action.yml
+++ b/.github/workflows/actions/build-core-stencil-prerelease/action.yml
@@ -22,7 +22,7 @@ runs:
       run: npm i @stencil/core@${{ inputs.stencil-version }}
       shell: bash
     - name: Build Core
-      run: npm run build -- --ci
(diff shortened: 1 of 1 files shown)


mapsandapps (approved):
good to merge once we get the builds fixed 😬

rwaskiewicz:
Created https://github.com/ionic-team/ionic-framework/pull/29079 to fix CI, if it doesn't magically fix itself in the mean time

rwaskiewicz:
@alicewriteswrongs should be good to go after a rebase, #29079 landed

Issue #27888, opened 2023-07-28
bug: Tapping an <IonItem> with a focusable child on Android unexpectedly focuses child and scrolls the page down


aeharding:
### Prerequisites

- [X] I have read the [Contributing Guidelines](https://github.com/ionic-team/ionic-framework/blob/main/.github/CONTRIBUTING.md#creating-an-issue).
- [X] I agree to follow the [Code of Conduct](https://ionicframework.com/code-of-conduct).
- [X] I have searched for [existing issues](https://github.com/ionic-team/ionic-framework/issues) that already report this problem, without success.

### Ionic Framework Version

v7.x

### Current Behavior

On Android, tapping an `<ion-item>` will focus the first focusable child. This can cause the page to unexpectedly scroll around on large `<ion-item>`s. This only happens on Android.

### Expected Behavior

Child should not receive focus.

### Steps to Reproduce

1. `git clone git@github.com:aeharding/ion-item-focus-issue.git`
2. `yarn`
3. `yarn start`
4. Open page on Android Chrome
5. Tap first Ion Item. Observe button at bottom of item receives focus
6. Additionally, observe page scrolls down to the button receiving focus if the `<button>` is off-screen
7. Finally, observe changing `<IonItem>` to a `<div>` causes the issue to stop

### Code Reproduction URL

https://github.com/aeharding/ion-item-focus-issue

### Ionic Info

```
[WARN] You are not in an Ionic project directory. Project context may be missing.

Ionic:

   Ionic CLI : 7.1.1

Utility:

   cordova-res : not installed globally
   native-run  : not installed globally

System:

   NodeJS : v18.15.0
   npm    : 9.5.0
   OS     : macOS Unknown
```

### Additional Information

End user bug report: https://github.com/aeharding/voyager/issues/287

Video:

https://github.com/ionic-team/ionic-framework/assets/2166114/11f90d7f-309f-484f-a978-d74d66449f16

liamdebeasi:
This is to be expected. `ion-item` uses a feature called "delegatesFocus". When the item itself is tapped focus will be forwarded to the first focusable element. In this case, this is the button you have defined in `ion-item`.

This feature was added for the legacy form control syntax a few years ago, though we plan to remove it once the legacy form syntax has been removed in an upcoming major release of Ionic. 

We're tracking this in https://github.com/ionic-team/ionic-framework/issues/21982, so I am going to merge the two threads.

aeharding:
[...]

Issue #25875, opened 2022-09-02
bug: icon click in item focuses input


EinfachHans:
### Prerequisites

- [X] I have read the [Contributing Guidelines](https://github.com/ionic-team/ionic-framework/blob/main/.github/CONTRIBUTING.md#creating-an-issue).
- [X] I agree to follow the [Code of Conduct](https://ionicframework.com/code-of-conduct).
- [X] I have searched for [existing issues](https://github.com/ionic-team/ionic-framework/issues) that already report this problem, without success.

### Ionic Framework Version

- [ ] v4.x
- [ ] v5.x
- [X] v6.x
- [ ] Nightly

### Current Behavior

We have an input in our app which we now want to extend by a camera scanner, so we added an icon at the iiems end to open the scanner. Clicking this icon will focus the input and show the keyboard.

This should be change-able by preventing the click event, but that doesn't work 🤔

### Expected Behavior

When using 

```
event.preventDefault();
event.stopPropagation();
```

the input should not focus

### Steps to Reproduce

See attached Stackblitz, just click on the camera icon

### Code Reproduction URL

https://stackblitz.com/edit/ionic6-angular13-ve7khj?file=src%2Fapp%2Fapp.component.html

### Ionic Info

From the app:

Ionic:

   Ionic CLI                     : 6.20.1 (/Users/hans/.nvm/versions/node/v16.14.2/lib/node_modules/@ionic/cli)
   Ionic Framework               : @ionic/angular 6.2.3
   @angular-devkit/build-angular : 14.1.3
   @angular-devkit/schematics    : 14.1.3
   @angular/cli                  : 14.1.3
   @ionic/angular-toolkit        : 7.0.0

Capacitor:

   Capacitor CLI      : 4.0.1
   @capacitor/android : 4.0.1
   @capacitor/core    : 4.0.1
   @capacitor/ios     : 4.0.1

Cordova:

   Cordova CLI       : 10.0.0 (cordova-lib@10.1.0)
   Cordova Platforms : not available
   Cordova Plugins   : not available

Utility:

   cordova-res                          : 0.15.4
   native-run (update available: 1.7.0) : 1.6.0

System:

   Android SDK Tools : 25.2.3 (/Users/hans/Library/Android/sdk)
   ios-deploy        : 1.10.0
   NodeJS            : v16.14.2 (/Users/hans/.nvm/versions/node/v16.14.2/bin/node)
   npm               : 8.17.0
   OS                : macOS Monterey
[...]

Issue #25273, opened 2022-05-10
bug: delegatesFocus patch for ios causes rendering issues with item sliding in chrome


deving1995:
### Prerequisites

- [X] I have read the [Contributing Guidelines](https://github.com/ionic-team/ionic-framework/blob/main/.github/CONTRIBUTING.md#creating-an-issue).
- [X] I agree to follow the [Code of Conduct](https://ionicframework.com/code-of-conduct).
- [X] I have searched for [existing issues](https://github.com/ionic-team/ionic-framework/issues) that already report this problem, without success.

### Ionic Framework Version

- [ ] v4.x
- [ ] v5.x
- [X] v6.x
- [ ] Nightly

### Current Behavior

When you click to drag an ion-item inside of an ion-item-sliding and you have an ion-input in the ion-item if you start the drag by hovering over the ion-label, then end the drag while hovering over the ion-input the slider abruptly jiggles to the left, slams shut, and when you drag it to open it again the ion-item-option is missing.

You can fix the ion-item-sliding by dragging it back to the left even though it is already shut.

I believe the issue is that the ion-item-option is somehow being pushed out of view.

### Expected Behavior

I expect that when I am intending to drag an ion-item inside an ion-item-sliding that releasing the mouse while hovering over an ion-input will not cause the ion-item-sliding to close, or for the ion-item-option to disappear when re-opening.

### Steps to Reproduce

1. Click the `ion-label` of the `ion-item`
    - Do NOT let go of the mouse
2. Drag the `ion-item` to the right until the cursor turns into a text-cursor
3. Release the mouse
4. Open the `ion-item-sliding` back up to see that the `ion-item-option` is now gone

```html
<ion-item-sliding>
  <ion-item-options side="start">
    <ion-item-option>Test</ion-item-option>
  </ion-item-options>
  <ion-item>
    <ion-label>Test Slider</ion-label>
    <ion-input></ion-input>
  </ion-item>
</ion-item-sliding>
```

Before option disappears
![image](https://user-images.githubusercontent.com/87441052/167725116-90260e5d-672f-47c6-827e-b3cf55cef9a3.png)

After option disappears
![image](https://user-images.githubusercontent.com/87441052/167725288-df2274ec-f892-40e8-82a9-c2ef79ecdb61.png)

And here's a gif of the issue in action
![slider-issue](https://user-images.githubusercontent.com/87441052/167725577-ceaacc95-151f-4851-bc65-2a50bf95da21.gif)

[...]

Issue #24956, opened 2022-03-17
bug: Cannot select text in IonItem IonLabel when there is a IonButton present (Ionic 5, VueJS)


maelp:
### Prerequisites

- [X] I have read the [Contributing Guidelines](https://github.com/ionic-team/ionic-framework/blob/main/.github/CONTRIBUTING.md#creating-an-issue).
- [X] I agree to follow the [Code of Conduct](https://ionicframework.com/code-of-conduct).
- [X] I have searched for [existing issues](https://github.com/ionic-team/ionic-framework/issues) that already report this problem, without success.

### Ionic Framework Version

- [ ] v4.x
- [X] v5.x
- [ ] v6.x

### Current Behavior

When adding an ion-item like this, I can select the text of the ion-label in a browser using the mouse (it highlights the text that I select and I can copy it)

```
        <ion-list v-if="items.length > 0">
          <ion-item v-for="item in items" :key="item.key">
            <ion-label>{{ item.id }}</ion-label>
          </ion-item>
        </ion-list>
```

but if I add a ion-button to it, the text can no longer be selected

```
        <ion-list v-if="items.length > 0">
          <ion-item v-for="item in items" :key="item.key">
            <ion-label>{{ item.id }}</ion-label>
            <ion-button slot="end">action</ion-button>
          </ion-item>
        </ion-list>
```

### Expected Behavior

The text should be selectable and copyable in a browser independently of the presence of a ion-button

### Steps to Reproduce

- create a ionic item such as above
- try to select the text

### Ionic Info

Ionic:

   Ionic CLI       : 6.18.1 (/Users/primet/.node/lib/node_modules/@ionic/cli)
   Ionic Framework : @ionic/vue 5.8.5

Capacitor:

   Capacitor CLI      : 3.3.1
   @capacitor/android : 3.4.3
   @capacitor/core    : 3.4.1
   @capacitor/ios     : 3.3.1

Utility:

   cordova-res (update available: 0.15.4) : 0.15.3
   native-run                             : 1.5.0

System:

   NodeJS : v16.14.0 (/usr/local/bin/node)
   npm    : 8.3.1
   OS     : macOS Monterey

amandaesmith3:
Thanks for the issue, I'm able to replicate this in core on the latest Ionic version. A couple things I noticed:
- The cursor changes to an I-beam as normal when hovering over the un-highlightable text.
- Setting `display: none` on the button makes the issue go away, even if the button is still present in the markup.

[...]

PR #29091, opened 2024-02-26
refactor(item): do not automatically delegate focus

### LINKED ISSUE #21982: bug: interactive items in ion-item unexpectedly being focused with delegatesFocus

liamdebeasi:
Issue number: resolves #21982

---------

### History

`delegatesFocus` was originally added to `ion-item` so [clicking an `ion-label` moves focus to the associated `ion-input`](https://github.com/ionic-team/ionic-framework/pull/21170). When using the legacy form syntax, labels were not automatically associated with inputs. Given the following markup, clicking `ion-label` would not focus the input:

```html
<ion-item>
  <ion-label>Input</ion-label>
  <ion-input></ion-input>
</ion-item>
```

`delegatesFocus` fixed this issue by moving focus to `ion-input` when a non-focusable part of the item is clicked. In this case, `ion-label` is non-focusable.

### Problems with Delegates Focus

fortunately, delegatesFocus was also buggy in browsers such as WebKit. We needed to [add workarounds](https://github.com/ionic-team/ionic-framework/pull/21930) which added technical debt to the project. We've slowly been able to remove these workarounds as the implementations in browsers improved.

In particular, we needed to add a workaround so that clicking the start/end padding in `ion-item` caused the input to be focused. The following screenshot demonstrates the start/end padding in the item. The color key is as follows:

White: `ion-input`

Blue: `ion-label`

Green: `ion-item`

Without `delegatesFocus`, clicking the green area does not cause the input to be focused. A user does not know about this implementation detail, so to them it seems like the input will sometimes not focus. In reality, this is happening because they sometimes tap the padding and other times they tap the label/input.

Screenshot 2024-02-26 at 12 18 58 PM

However, browser issues remain with this implementation. Developers have [complained](https://github.com/ionic-team/ionic-framework/issues/21982) [of](https://github.com/ionic-team/ionic-framework/issues/27888) [unexpected](https://github.com/ionic-team/ionic-framework/issues/25875) [focus](https://github.com/ionic-team/ionic-framework/issues/25273) [and](https://github.com/ionic-team/ionic-framework/issues/24956) [selection](https://github.com/ionic-team/ionic-framework/issues/23002) issues which resulted in even more [technical debt added to try and resolve the issues](https://github.com/ionic-team/ionic-framework/pull/22049).

### Current Day

With the introduction of the new form control syntax, labels are automatically associated with their inputs. This means that when a label is clicked, the browser will automatically focus the associated input. This behavior is [built in to the browser with `label` and `input`/`textarea` elements](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/Input).

With the [Entire item should be clickable with form controls](https://github.com/ionic-team/ionic-framework/commit/5b7e422dc0bfd4d58fb31f62715af47e62dabb57) feature we updated `ion-item` so that clicking the start/end padding focuses the input inside of the item.

Input/textarea were not accounted for in this scope, but by extending this functionality to input/textarea we would no longer need `delegatesFocus` for input/textarea.

## What is the new behavior?

- Item no longer automatically delegates focus to the first focusable element
- Extended the  [Entire item should be clickable with form controls](https://github.com/ionic-team/ionic-framework/commit/5b7e422dc0bfd4d58fb31f62715af47e62dabb57) feature so clicking the padding focuses input/textarea.

- [x] Yes
- [ ] No

## Other information

Design doc: https://github.com/ionic-team/ionic-framework-design-documents/commit/e09a8f309569a2a5383738d2bd727000b691871c

Only removing the delegatesFocus logic from `ion-item` causes focus trapping in `ion-menu` to fail. Upon further investigation I discovered that menu focus trapping only worked by chance. In our E2E test, we had an `ion-item` in a menu with an `ion-button` inside of it. There are a couple problems:

1. Menu moves focus to the `ion-item` because it has an `.ion-focusable` class. For some reason we consider the item focusable if it [has an focusable elements inside of it](https://github.com/ionic-team/ionic-framework/blob/5a5f330705f77a8971119fad448101a5504d8e5e/core/src/components/item/item.tsx#L319-L322). With the delegatesFocus logic, focus was being moved to the `ion-button` so the tests passed. However, this broke because focus was no longer delegated. The `ion-item` itself in this case is not a button item, so it cannot be focused. As a result, focus escaped from the menu. I'm not 100% sure why we consider an item to be focusable with this criteria, but I felt that changing that was too risky to be considered in scope.

2. I then tried to remove the `ion-item` from the test and just have an `ion-button`. I discovered something that I am confident is a bug which is that we always focus the host element and never try to focus the native button/input inside of a component's Shadow DOM. For the overlay focus trapping we have special logic to ensure the [correct interactive element is focused](https://github.com/ionic-team/ionic-framework/blob/5a5f330705f77a8971119fad448101a5504d8e5e/core/src/utils/overlays.ts#L182-L185). However, that logic does not exist in the menu. As a result, `ion-button` was being focused, but it's actually the native `button` inside of the shadow root that needs to be focused.

-----

To fix this, I created a `focus-trap` file with utilities for focusing the first/last descendant in a focus trapped container. The implementation of this is the same implementation used in the overlay focus trapping. However, I moved it to this util file so that the menu can take advantage of that logic too (and therefore address the bug in point 2 above). There should be no behavior changes for the overlay components.

I removed the `ion-item` usage from the basic menu test which is why there are visual diffs.


liamdebeasi:
The menu changes are required in order to ship the `delegatesFocus` change. However, I could move it to a separate PR if folks would prefer. I decided to keep them combined because we have solid test coverage with focus trapping, so I feel confident that this does not regress anything. Also, this is shipping in Ionic 8 so we'll have a beta period too.

thetaPC (approved):
LGTM, didn't have issues locally

core/src/utils/focus-trap.ts
```diff
@@ -0,0 +1,82 @@
...
+
+/**
+ * Focuses the first descendant in a context
+ * that can receive focus. If none exists,
+ * a fallback element will be focused.
+ * This fallback is typically an ancestor
+ * container such as a menu or overlay so focus does not
+ * leave the container we are trying to trap focus in.
+ */
+export const focusFirstDescendant = <T extends HTMLElement>(ref: Element, fallbackElement: T) => {
```
sean-perkins, line 26: I think we can simplify some of the usage pattern here by using an additional type parameter. 

For example:
```ts
export const focusFirstDescendant = <R extends Element, T extends HTMLElement>(ref: R, fallbackElement: T) => {
  const firstInput = ref.querySelector<HTMLElement>(focusableQueryString);

  focusElementInContext(firstInput, fallbackElement);
};
```

Takes the mark-up that used to be:
```ts
focusFirstDescendant<HTMLIonMenuElement>(el, el);
```

and allows us to do:
```ts
focusFirstDescendant(el, el);
```

The type parameters will be computed based on the type of the arguments passed into the utility function:

CleanShot 2024-02-27 at 23 45 48@2x

You will also notice that I updated the usage of the `.querySelector` to use a type parameter instead of casting. This is a much safer type inference vs. assigning the type signature manually.

sean-perkins, line 26: Separate comment, but on the same function. Why is the fallback argument required if in most of the implementations we supply the same element reference to both the `ref` and `fallbackElement`? 

Could we make it so that if you pass a `ref` without a `fallbackElement`, it uses the `ref` as the fallback?

sean-perkins: I do think it is strange that `ref` and `fallbackElement` are different type signatures, while being a reference to the same element in some cases. Is there a reason why it is `Element` vs. `HTMLElement`?

liamdebeasi: It's possible this was an oversight when originally implemented. For this I tried to minimize refactoring and re-use what we already had as much as possible.

liamdebeasi: Ok I made a couple changes:

1. [Converged on `HTMLElement` instead of `Element` and `HTMLElement`](https://github.com/ionic-team/ionic-framework/pull/29091/commits/3f9f94c2416e6d6bbf9e65573687938208e06f4f).
2. [Added additional type parameter as you suggested.](https://github.com/ionic-team/ionic-framework/pull/29091/commits/37ad77a304fdbceb47fd6c5f2888bc7897f48d1c)

liamdebeasi: Updated in https://github.com/ionic-team/ionic-framework/pull/29091/commits/b1917462ab61c9e73986e190c7a903ecf044ffad

core/src/components/item/item.tsx
```diff
@@ -359,7 +357,7 @@ export class Item implements ComponentInterface, AnchorInterface, ButtonInterfac
 
   private getFirstInteractive() {
     const controls = this.el.querySelectorAll<HTMLElement>(
-      'ion-toggle:not([disabled]), ion-checkbox:not([disabled]), ion-radio:not([disabled]), ion-select:not([disabled])'
+      'ion-toggle:not([disabled]), ion-checkbox:not([disabled]), ion-radio:not([disabled]), ion-select:not([disabled]), ion-input:not([disabled]), ion-textarea:not([disabled])'
```
amandaejohnston, line 360: Nitpick: The padding around the input now has `cursor: pointer`, even though the rest of the control (label, input) doesn't; the inconsistency looks strange. For the other controls, it's consistent since `cursor: pointer` is used over the entire control.

amandaejohnston (changes requested):

core/src/components/item/item.tsx
amandaejohnston: That `cursor: pointer` is coming from the `item-has-interactive-control` class on `ion-item`, and then it gets overwritten by `label`'s default styles. Either we could add `cursor: pointer` to the label --
```css
/* possibly restricted to just labels in inputs/textareas */
.item-has-interactive-control label {
  cursor: pointer;
}
```
-- or make items with inputs/textareas not get `cursor: pointer` at all, likely by having it not add the class in that case. I'm not really sure which is best though 🤔

liamdebeasi: In https://github.com/ionic-team/ionic-framework/pull/29091/commits/143fb7d602f20be3a6e7d683e3c5710c01ce0afd I updated the code so that inputs and textures do not have the cursor effect. I updated the name of the class to be more descriptive of what it's doing, but I can change the name if you have other ideas.

liamdebeasi: I could also do this with CSS and have it be `.item-has-interactive-control:not(:has(ion-input, ion-textarea))` but the specificity gets a little complicated, so I opted to just apply the class using logic in JS (since the class is already applied using JS logic)

amandaejohnston: Maybe something like `item-has-click-toggle-control` or `item-control-needs-pointer-cursor`? "Needs cursor" by itself is confusing to me since in theory everything needs a cursor, like initially clicking into the input.

amandaejohnston (approved):
LGTM once the class name is adjusted. Might also want to update the `firstInteractiveNeedsCursor` variable to match.

core/src/components/item/item.tsx
liamdebeasi: Clarified in https://github.com/ionic-team/ionic-framework/pull/29091/commits/4362c3bb79fed288cbef47cfea03d316e12bde11 and https://github.com/ionic-team/ionic-framework/pull/29091/commits/0c9311c3af6de9ebd22bf4e26adcbd93db1a09a3

PR #29101, opened 2024-02-29
perf(picker): avoid flicker on ios

liamdebeasi:
Issue number: N/A (see below)

---------

Mike found an issue while testing the Ionic 8 beta where there's a flicker when presenting a picker in a modal on iOS. [We've seen this issue before](https://github.com/ionic-team/ionic-framework/blob/de13633a182d963876434db773aa346833f956fd/core/src/utils/forms/notch-controller.ts#L135-L144), and the reason why it's happening is there's a quirk in WebKit where the IntersectionObserver callback is fired _after_ an accelerated animation completes given a particular IO configuration.

The end result is there's a delay before each picker column is scrolled to the correct place.

In particular, the modal enter animation on iOS is an accelerated animation, and we use an IO to control when the picker columns [should scroll their active options into view](https://github.com/ionic-team/ionic-framework/blob/60056643a99aea9817d2cd205fd011ab5967a995/core/src/components/picker-column/picker-column.tsx#L107).

## What is the new behavior?

- The root of the intersection observer is now the parent picker element which avoids the WebKit quirk.

| `feature-8.0` | branch |
| - | - |
| <video src="https://github.com/ionic-team/ionic-framework/assets/2721089/7478512b-a691-405e-aafa-e9e377fc47ac"></video> | <video src="https://github.com/ionic-team/ionic-framework/assets/2721089/5092e050-733a-4965-8f16-317d251af46a"></video> |

- [ ] Yes
- [x] No

## Other information

Dev build: `8.0.0-dev.11709226349.179192de`

Note: Given that this bug is due to a WebKit quirk and is timing-related, I did not add a test.


core/src/components/picker-column/picker-column.tsx
```diff
@@ -87,13 +87,16 @@ export class PickerColumn implements ComponentInterface {
    * height of 0px.
    */
   componentWillLoad() {
+    const parentEl = (this.parentEl = this.el.closest('ion-picker') as HTMLIonPickerElement | null);
```
thetaPC, line 90: Suggestion: Add a comment of the difference between this `parentEl` and `this.parentEl`. If a dev has limited time, they might not be able to quickly determine the differences.

thetaPC (approved):
LGTM, it works great! The flicker is no longer there.

core/src/components/picker-column/picker-column.tsx
liamdebeasi: Good idea. Added in https://github.com/ionic-team/ionic-framework/pull/29101/commits/36966c61a46014c958362043333b5a15bedb5744

PR #29102, opened 2024-02-29
refactor(item): remove deprecated apis

liamdebeasi:
Issue number: Internal

---------

Item the following deprecated APIs:

**Props**

1. fill
2. shape
3. counter
4. counterFomatter

**Slots**

1. helper
2. error

## What is the new behavior?

- These properties and slots (including implementations) have been removed.

- [x] Yes
- [ ] No

## Other information

BREAKING CHANGE:

- The `helper` slot has been removed. Developers should use the `helperText` property on `ion-input` and `ion-textarea`.
- The `error` slot has been removed. Developers should use the `errorText` property on `ion-input` and `ion-textarea`.
- Counter functionality has been removed including the `counter` and `counterFormatter` properties. Developers should use the properties of the same name on `ion-input` and `ion-textarea`.
- The `fill` property has been removed. Developers should use the property of the same name on `ion-input`, `ion-select`, and `ion-textarea`.
- The `shape` property has been removed. Developers should use the property of the same name on `ion-input`, `ion-select`, and `ion-textarea`.


core/src/components/item/item.md.scss
```diff
@@ -32,65 +32,22 @@
   text-transform: none;
 }
 
-:host(.item-fill-outline) {
-  --highlight-height: 2px;
-}
-
 // Item Fill: None
 // --------------------------------------------------
```
amandaejohnston, line 36:
```suggestion
```
This comment is no longer relevant.

core/src/components/item/item.ios.scss
```diff
@@ -204,18 +202,10 @@
...
 }
 
-
 // iOS Stacked & Floating Inputs
 // --------------------------------------------------
 
 :host(.item-label-floating),
 :host(.item-label-stacked) {
   --min-height: 68px;
 }
```
amandaejohnston, line 211: Was this missed? The corresponding styles in MD have been removed in this PR.

I was a little confused about why the `item-label-` styles are being removed in this PR, but it looks like the class hasn't been on item for a while anyway, so we're just removing the remaining styles now since it's convenient? Is that accurate?

core/src/components/item/item.md.scss
```diff
@@ -32,65 +32,22 @@
...
   border-style: var(--border-style);
   border-color: var(--highlight-background);
 }
 
-:host(.item-fill-solid.item-interactive.ion-focus) .item-native,
-:host(.item-fill-solid.item-interactive.item-has-focus) .item-native,
-:host(.item-fill-solid.item-interactive.ion-touched.ion-invalid) .item-native {
+:host(.item-interactive.ion-focus) .item-native,
+:host(.item-interactive.item-has-focus) .item-native,
+:host(.item-interactive.ion-touched.ion-invalid) .item-native {
```
amandaejohnston, line 50: Since these styles previously only applied on `fill="solid"`, wouldn't we want to remove them completely rather than have them apply in all cases?

core/src/components/item/item.scss
```diff
@@ -412,7 +412,7 @@ button, a {
   box-sizing: border-box;
 }
 
-:host(.item-label-stacked),
```
amandaejohnston, line 415: Why are we only removing the `stacked` style here? Other parts of the PR removed both.

core/src/components/item/item.scss
```diff
@@ -521,35 +515,6 @@ button, a {
...
-:host(.item-interactive.ion-invalid) ::slotted([slot="error"]) {
-  display: block;
-}
-
-// Item Datetime
-// --------------------------------------------------
-
-:host(:not(.item-label)) ::slotted(ion-datetime) {
-  --padding-start: 0;
-}
```
amandaejohnston, line 543: Why was this style removed? It doesn't look related to any of the deprecated APIs.

core/src/components/item/test/basic/item.e2e.ts
```diff
@@ -26,8 +26,7 @@ configs().forEach(({ title, screenshot, config }) => {
         </style>
         <ion-list>
           <ion-item>
-            <ion-label>Item with helper</ion-label>
-            <div slot="helper">Helper</div>
+            <ion-label>Item</ion-label>
```
amandaejohnston, line 29: I think this item can be removed entirely; it doesn't seem to be adding anything to the test now.

amandaejohnston (changes requested):

core/src/components/item/item.ios.scss
liamdebeasi: No, these should stay. These have to do with `ion-label` position property. We may need to discuss if the floating/stacked/fixed variants are still needed on `ion-label`, but the behaviors were not deprecated originally so I kept them in.

core/src/components/item/item.md.scss
liamdebeasi: fixed in https://github.com/ionic-team/ionic-framework/pull/29102/commits/e65dbf6f9a06a001fae0031e44cfe05d063c6a13

core/src/components/item/item.scss
liamdebeasi: This is part of the form control + item logic. As part of this refactor we are removing the tight coupling between items and form controls.

core/src/components/textarea/textarea.scss
```diff
@@ -116,10 +116,6 @@
   align-self: baseline;
 }
 
-:host-context(ion-item:not(.item-label)) {
-  --padding-start: 0;
-}
```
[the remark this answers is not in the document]
liamdebeasi: This only applied with the legacy textarea syntax. With the modern syntax the padding start is always 0.

core/src/components/item/item.scss
liamdebeasi: Fixed in https://github.com/ionic-team/ionic-framework/pull/29102/commits/1e480e7e3c45629b89587142766d926051a00d32

core/src/components/item/test/basic/item.e2e.ts
liamdebeasi: I don't understand. This test verifies that safe area is being added properly, but I didn't remove any of that logic.

core/src/components/item/item.md.scss
```diff
@@ -298,24 +203,6 @@
...
 }
 
-
-// Material Design Floating/Stacked Label
-// --------------------------------------------------
-
-:host(.item-label-stacked) ::slotted([slot="end"]),
-:host(.item-label-floating) ::slotted([slot="end"]) {
-  @include margin($item-md-label-slot-end-margin-top, $item-md-label-slot-end-margin-end, $item-md-label-slot-end-margin-bottom, $item-md-label-slot-end-margin-start);
-}
```
amandaejohnston, line 308: Why were these removed if we're keeping the `ion-label` position styles?

core/src/components/item/item.md.scss
```diff
@@ -348,14 +235,6 @@
 // }
 
 
-// Material Design Stacked & Floating Inputs
-// --------------------------------------------------
-
-:host(.item-label-floating),
-:host(.item-label-stacked) {
-  --min-height: 55px;
-}
```
amandaejohnston, line 357: Why were these removed if we're keeping the `ion-label` position styles?

core/src/components/item/test/basic/item.e2e.ts
amandaejohnston: There are now two items in these screenshots, but we only need one to test the safe area. Previously the modified item was checking how the `helper` slot handles safe area; now it's just a regular item, so the other item in the screenshots handles things by itself.

mapsandapps (changes requested):
I think I found two other things that can be removed:

- The `item-md-input-counter-letter-spacing` variable in [item.md.vars.scss](https://github.com/ionic-team/ionic-framework/blob/1e480e7e3c45629b89587142766d926051a00d32/core/src/components/item/item.md.vars.scss#L72-L73)
- The `.item-bottom` styles in [item.scss](https://github.com/ionic-team/ionic-framework/blob/1e480e7e3c45629b89587142766d926051a00d32/core/src/components/item/item.scss#L309-L332)

core/src/components/item/item.md.scss
liamdebeasi: Good catch. I had originally removed this before I realized that I should not have. These instances were the ones I missed when reverting.

core/src/components/item/item.md.scss
liamdebeasi: Fixed in https://github.com/ionic-team/ionic-framework/pull/29102/commits/02e5958b825b3a07174b178fd29b80b34feec0be

liamdebeasi:
Bottom styles and variables removed in https://github.com/ionic-team/ionic-framework/pull/29102/commits/d48224689337376c6cd38978c5675b61be44d0bb and https://github.com/ionic-team/ionic-framework/pull/29102/commits/f98e95ded4650f7b1680cf3eca2c53d0002d646c.

core/src/components/item/test/basic/item.e2e.ts
liamdebeasi: Ahh ok I see. I can fix

liamdebeasi: Fixed https://github.com/ionic-team/ionic-framework/pull/29102/commits/cbdfdd677819440d28a9837a96cb34cf42bcfc50

mapsandapps (approved):

Issue #11484, opened 2017-05-03
ion-searchbar focus issues


Rockforced:
**Ionic version:**  (check one with "x")
[ ] **1.x**
[ ] **2.x**
[X ] **3.x**

**I'm submitting a ...**  (check one with "x")
[X] bug report
[ ] feature request
[ ] support request => Please do not submit support requests here, use one of these channels: https://forum.ionicframework.com/ or http://ionicworldwide.herokuapp.com/

**Current behavior:**
ion-searchbar has focus issues. Does not lose focus when touching other areas of the screen. Incidentally the keyboard stays up as well and persists through screen changes, effectively breaking the app.

https://www.youtube.com/watch?v=kXSUjTJPRqM   <- watch

**Expected behavior:**
The search bar should lose focus when other areas of the screen are touched. Further, the search bar should lose focus when transitioning screens.

**Other information:**
https://www.youtube.com/watch?v=kXSUjTJPRqM

**Ionic info:** (run `ionic info` from a terminal/cmd prompt and paste output below):
Cordova CLI: 6.4.0
Ionic Framework Version: 3.1.0
Ionic CLI Version: 2.1.14
Ionic App Lib Version: 2.1.7
Ionic App Scripts Version: 1.3.4
ios-deploy version: Not installed
ios-sim version: Not installed
OS: Windows 10
Node Version: v4.4.1
Xcode version: Not installed

jgw96:
Hello, thanks for using Ionic! We will look into this.

joewoodhouse:
Any update on when these keyboard fixes will be released? The ion-searchbar is completely unusable on iOS without this (as well as a number of other keyboard related things).

Issue #8933, opened 2016-10-27
Searchbar Not Blurring


duncanreid:
#### Short description of the problem:

RC1. Device: iOS 10.0.2.

This only happens when testing on the device, it works as expected while testing in Chrome.

Searchbar input will not blur when tapping anywhere outside of the component, so the keyboard will not go away and the ionBlur event is not firing.  It will blur if you tap on a button or something else that gains focus. The only way I have been able to work around this is to add click listener directives to all of the elements around the Searchbar.  

I notice that the Searchbar is using a normal input as opposed to an ion-input, so it also inhibits the cursor peekaboo effect where the cursor can still be seen while scrolling underneath fixed elements like headers.
#### What behavior are you expecting?

Tapping anywhere outside the Search bar will remove the keyboard and fire the (ionBlur) event.

**Steps to reproduce:**
1. Fire up a basic v2 tabs project
2. Add a Searchbar
3. Tap Searchbar input, tap anywhere outside after keyboard is shown.

**Which Ionic Version?** 1.x or 2.x
2.X RC1

Cordova CLI: 6.3.0
Gulp version:  CLI version 3.9.0
Gulp local:
Ionic Framework Version: 2.0.0-rc.1
Ionic CLI Version: 2.1.1
Ionic App Lib Version: 2.1.1
Ionic App Scripts Version: 0.0.36
ios-deploy version: 1.8.6
ios-sim version: 5.0.6
OS: Mac OS X El Capitan
Node Version: v6.2.2
Xcode version: Xcode 8.0 Build version 8A218a

jgw96:
@manucorporat im thinking this is probably caused by the backrdop on iOS right?

manucorporat:
@jgw96 I don't think so, hard to say without testing

danielcrk-cn:
If it helps, the blurring works fine on Android but not on iOS.

Xal3ph:
We are also seeing this in 2.x RC2

biesbjerg:
Just to add to this issue, I'm seeing a delay (300ms tap delay?) when tapping the searchbar to gain focus, and similar when tapping clear or cancel.

Xal3ph:
We added '(click)="return:"' or '(click)="{}"' to the 'ion-content' and this fixed it for us.

biesbjerg:
[...]

PR #29104, opened 2024-02-29
refactor(input-shims): disable input blurring util by default

### LINKED ISSUE #29072: bug: input blurring utility causes non input and textarea elements to be blurred
### Prerequisites

### Ionic Framework Version

v7.x

### Current Behavior

In an Ionic Angular application with video.js installed at latest (8.10.x) when running in either Chrome with DevTools in mobile device mode, or actually running on a device (simulator in xcode or deployed to a mobile device) the playback rates button will open and then immediately close the menu with the rate buttons in it. Sometimes a longer press will cause it to stay open, but a tap is always an immediate open/close. This ONLY happens in an Ionic app. Outside of Ionic the videojs menus work perfectly.

### Expected Behavior

When clicking on a videojs control button Ionic doesn't interfere with videojs behavior.

### Steps to Reproduce

1. Follow the instructions in the readme on the linked repo

### Code Reproduction URL

https://github.com/jkyoutsey/ionic-videojs-playback-rates-flicker-bug

### Ionic Info

Ionic:

   Ionic CLI : 7.2.0

Utility:

   native-run  : 2.0.1

System:

   NodeJS : v20.11.1
   npm    : 10.2.4
   OS     : macOS Unknown

### Additional Information

_No response_

liamdebeasi:
Issue number: resolves #29072

---------

The input blurring utility is causing adverse side effects in developer applications by causing certain UI components to be blurred. This utility was [added back in 2017](https://github.com/ionic-team/ionic-v3/commit/c10f72b1e2614292e547c68f6f26515efb1cd691).

That commit claims to fix a number of issues, but https://github.com/ionic-team/ionic-framework/issues/8933 and https://github.com/ionic-team/ionic-framework/issues/11484 seem most relevant here.

While the purpose of this utility is not definitively known it appears that this was created to solve an issue on iOS where the searchbar was not being blurred when tapping outside of the input. The linked issues refer to cases where inputs are not blurred when they should be. This aligns with the input blurring utility behavior where it only blurs elements and never focuses them.

Additionally, the two linked issues only happened on iOS which aligns with the default behavior of the input blurring utility which is to only be enabled on iOS.

I tested the searchbar on iOS with this utility **disabled** and I was able to blur the searchbar by tapping outside the input. It seems that this utility was created to work around a WebKit issue that has since been resolved.

https://github.com/ionic-team/ionic-framework/assets/2721089/7772688b-a0d4-476e-be72-931cc07cd93a

## What is the new behavior?

- Given that I am not 100% sure that this utility does what I think it does, I'd like to propose we disable this feature by default starting in Ionic 8. This will resolve the linked issue but also give developers an escape hatch (by manually re-enabling it in their apps) if disabling the utility does cause issues. The team can evaluate remove the code altogether if disabling it does not have any known adverse side effects.

- [ ] Yes
- [x] No

I don't consider this a breaking change because a) `inputBlurring` is a private API and b) there current thinking is that there should be no behavior change (other than the buggy behaviors going away) since this utility exists to solve a WebKit issue that no longer exists.

## Other information

Dev build: `8.0.0-dev.11709245047.1565a499`


core/src/utils/input-shims/input-shims.ts
```diff
@@ -33,7 +33,15 @@ export const startInputShims = async (config: Config, platform: 'ios' | 'android
   const keyboardHeight = config.getNumber('keyboardHeight', 290);
   const scrollAssist = config.getBoolean('scrollAssist', true);
   const hideCaret = config.getBoolean('hideCaretOnScroll', isIOS);
-  const inputBlurring = config.getBoolean('inputBlurring', isIOS);
+
+  /**
+   * The team is evaluating if inputBlurring is still needed. As a result
+   * this feature is disabled by default as of Ionic 8.0. Developers are
+   * able to re-enabled it temporarily. The team may remove this utility
```
amandaejohnston, line 40:
```suggestion
   * able to re-enable it temporarily. The team may remove this utility
```

core/src/utils/input-shims/input-shims.ts
```diff
@@ -113,6 +121,7 @@ export const startInputShims = async (config: Config, platform: 'ios' | 'android
     }
   };
 
+  console.log(inputBlurring, INPUT_BLURRING);
```
amandaejohnston, line 124:
```suggestion
```

amandaejohnston (approved):

PR #26900, opened 2023-03-02
fix(button): show correct activated state for ios

brandyscarney:
## Pull request checklist

Please check if your PR fulfills the following requirements:
- [x] Tests for the changes have been added (for bug fixes / features)
- [x] Docs have been reviewed and added / updated if needed (for bug fixes / features)
  - Some docs updates need to be made in the `ionic-docs` repo, in a separate PR. See the [contributing guide](https://github.com/ionic-team/ionic-framework/blob/main/.github/CONTRIBUTING.md#modifying-documentation) for details.
- [x] Build (`npm run build`) was run locally and any changes were pushed
- [x] Lint (`npm run lint`) has passed locally and any fixes were made for failures

## Pull request type

 

 

Please check the type of change your PR introduces:
- [x] Bugfix
- [ ] Feature
- [ ] Code style update (formatting, renaming)
- [ ] Refactoring (no functional changes, no api changes)
- [ ] Build related changes
- [ ] Documentation content changes
- [ ] Other (please describe): 

## What is the current behavior?

- The activated state does not show up on iOS while hovering
- Outline buttons in a toolbar aren't showing activated state at all

Issue URL: fixes #22468

## What is the new behavior?

- Display activated states for buttons when pressing while hovering
- Show the activated state for outline buttons in a toolbar (updated to match what an outline button with color was already doing) 
- While fixing the above there were some issues that popped up, resolved all found issues 

## Does this introduce a breaking change?

- [ ] Yes
- [x] No

## Other information

The following tests should show no differences:

- button/basic
- button/clear
- button/outline

The following tests will show fixes where activated was not being applied at all for outline buttons, regardless of hover state. I made edits to these tests on my branch in order to check more use cases, so these diffs won't show up in screenshot comparisons:

| test | main | branch | 
| ---| ---| ---|
| toolbar/basic | toolbar - basic 1 | toolbar - basic 2 | 
| toolbar/colors | toolbar - colors 2 | toolbar - colors 1 |
| themes/default | themes - default | themes - default |
| themes/oceanic | themes - oceanic | themes - oceanic |
| themes/vibrant | themes - vibrant |  themes - vibrant |
| themes/dark |  themes - dark | themes - dark |

PR #29106, opened 2024-03-01
docs(contributing): add more context about focus, hover, and active

liamdebeasi:
The team discussed these utilities for some upcoming features work. This PR reflects some of the things I talked about during the discussion for future reference.


.github/COMPONENT-GUIDE.md
```diff
@@ -93,6 +93,10 @@ The following styles should be set for the CSS to work properly. Note that the `
 
 The activated state should be enabled for elements with actions on "press". It usually changes the opacity or background of an element.
 
+`:active` should not be used here as it is not received on mobile Safari unless the element has a `touchstart` listener (which we don't necessarily want to have to add to every element). From [Safari Web Content Guide](https://developer.apple.com/library/archive/documentation/AppleApplications/Reference/SafariWebContent/AdjustingtheTextSize/AdjustingtheTextSize.html):
+
+> On iOS, mouse events are sent so quickly that the down or active state is never received. Therefore, the :active pseudo state is triggered only when there is a touch event set on the HTML element
```
brandyscarney, line 98:
```suggestion
> [!WARNING]
>`:active` should not be used here as it is not received on mobile Safari unless the element has a `touchstart` listener (which we don't necessarily want to have to add to every element). From [Safari Web Content Guide](https://developer.apple.com/library/archive/documentation/AppleApplications/Reference/SafariWebContent/AdjustingtheTextSize/AdjustingtheTextSize.html):
>
>> On iOS, mouse events are sent so quickly that the down or active state is never received. Therefore, the `:active` pseudo state is triggered only when there is a touch event set on the HTML element
```

.github/COMPONENT-GUIDE.md
```diff
@@ -212,7 +216,9 @@ TODO
...
+
+Note that the [`:focus-visible`](https://developer.mozilla.org/en-US/docs/Web/CSS/:focus-visible) pseudo-class mostly does the same thing as our JavaScript-driven utility. However, it does not work well with Shadow DOM components as the element that receives focus is typically inside of the Shadow DOM, but we usually want to set the `:focus-visible` state on the host so we can style other parts of the component. Using other combinations such as `:has(:focus-visible)` does not work because `:has` does not pierce the Shadow DOM (as that would leak implementation details about the Shadow DOM contents). `:focus-within` does work with the Shadow DOM, but that has the same problem as `:focus` that was mentioned before. Unfortunately, a [`:focus-visible-within` pseudo-class does not exist yet](https://github.com/WICG/focus-visible/issues/151).
```
brandyscarney, line 221:
```suggestion
The focused state should be enabled for elements with actions when tabbed to via the keyboard. This will only work inside of an `ion-app`. It usually changes the opacity or background of an element. 

> [!WARNING]
> Do not use `:focus` because that will cause the focus to apply even when an element is tapped (because the element is now focused). Instead, we only want the focus state to be shown when it makes sense which is what the `.ion-focusable` utility mentioned below does.

> [!NOTE]
> The [`:focus-visible`](https://developer.mozilla.org/en-US/docs/Web/CSS/:focus-visible) pseudo-class mostly does the same thing as our JavaScript-driven utility. However, it does not work well with Shadow DOM components as the element that receives focus is typically inside of the Shadow DOM, but we usually want to set the `:focus-visible` state on the host so we can style other parts of the component. Using other combinations such as `:has(:focus-visible)` does not work because `:has` does not pierce the Shadow DOM (as that would leak implementation details about the Shadow DOM contents). `:focus-within` does work with the Shadow DOM, but that has the same problem as `:focus` that was mentioned before. Unfortunately, a [`:focus-visible-within` pseudo-class does not exist yet](https://github.com/WICG/focus-visible/issues/151).
```

.github/COMPONENT-GUIDE.md
```diff
@@ -275,7 +281,7 @@ ion-button {
 
 ### Hover
 
-The [hover state](https://developer.mozilla.org/en-US/docs/Web/CSS/:hover) happens when a user moves their cursor on top of an element without pressing on it. It should not happen on mobile, only on desktop devices that support hover.
+The [hover state](https://developer.mozilla.org/en-US/docs/Web/CSS/:hover) happens when a user moves their cursor on top of an element without pressing on it. It should not happen on mobile, only on desktop devices that support hover. Note that [some Android devices incorrectly report their inputs](https://issues.chromium.org/issues/40855702) which can result in certain devices receiving hover events when they should not.
```
brandyscarney, line 284:
```suggestion
The [hover state](https://developer.mozilla.org/en-US/docs/Web/CSS/:hover) happens when a user moves their cursor on top of an element without pressing on it. It should not happen on mobile, only on desktop devices that support hover. 

> [!NOTE]
> Some Android devices [incorrectly report their inputs](https://issues.chromium.org/issues/40855702) which can result in certain devices receiving hover events when they should not.
```

brandyscarney (commented):
Feel free to reject or change my suggestions, I just thought it might be easier to parse if we split out things that are informative rather than required implementation steps.

mapsandapps:
Hi, this doc contains the following statements:

> Order is important! Activated should be after the focused & hover states.

> Order is important! Focused should be after the activated and before the hover state.

> Order is important! Hover should be before the activated state.

These don't match each other 😅 

Button has them in this order:
1. focused
2. hover
3. activated

brandyscarney:
Yeah it looks like I forgot to update the `focused` one when I moved them: https://github.com/ionic-team/ionic-framework/pull/26900/files

liamdebeasi:
@brandyscarney would you be able to fix in a separate PR if you have time?

brandyscarney (approved):

PR #29111, opened 2024-03-04
docs(contributing): update component guide with order for states

brandyscarney:
Updates the component guide to use the proper order for the `focused`, `hover`, and `activated` states. Reorders the sections based on the recommended order and adds emphasis and blockquotes to call out the order. 

I have created ticket [FW-6039](https://outsystemsrd.atlassian.net/browse/FW-6039) for investigating components that don't follow the recommended order here.

### DIFF
--- a/.github/COMPONENT-GUIDE.md
+++ b/.github/COMPONENT-GUIDE.md
@@ -143,77 +143,6 @@ The following CSS _at the bare minimum_ should be added for the disabled class, 
 
 TODO
 
-### Activated
-
-The activated state should be enabled for elements with actions on "press". It usually changes the opacity or background of an element.
-
-> [!WARNING]
->`:active` should not be used here as it is not received on mobile Safari unless the element has a `touchstart` listener (which we don't necessarily want to have to add to every element). From [Safari Web Content Guide](https://developer.apple.com/library/archive/documentation/AppleApplications/Reference/SafariWebContent/AdjustingtheTextSize/AdjustingtheTextSize.html):
->
->> On iOS, mouse events are sent so quickly that the down or active state is never received. Therefore, the `:active` pseudo state is triggered only when there is a touch event set on the HTML element
-
-> Make sure the component has the correct [component structure](#component-structure) before continuing.
-
-#### JavaScript
-
-The `ion-activatable` class needs to be set on an element that can be activated:
-
-```jsx
-render() {
-  return (
-    <Host class='ion-activatable'>
-      <slot></slot>
-    </Host>
-  );
-}
-```
-
-Once that is done, the element will get the `ion-activated` class added on press after a small delay. This delay exists so that the active state does not show up when an activatable element is tapped while scrolling.
-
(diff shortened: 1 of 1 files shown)


.github/COMPONENT-GUIDE.md
```diff
@@ -336,6 +267,78 @@ ion-button {
...
+### Activated
+
+The activated state should be enabled for elements with actions on "press". It usually changes the opacity or background of an element.
+
+> [!WARNING]
+>`:active` should not be used here as it is not received on mobile Safari unless the element has a `touchstart` listener (which we don't necessarily want to have to add to every element). From [Safari Web Content Guide](https://developer.apple.com/library/archive/documentation/AppleApplications/Reference/SafariWebContent/AdjustingtheTextSize/AdjustingtheTextSize.html):
+>
+>> On iOS, mouse events are sent so quickly that the down or active state is never received. Therefore, the `:active` pseudo state is triggered only when there is a touch event set on the HTML element
+
+> Make sure the component has the correct [component structure](#component-structure) before continuing.
```
sean-perkins, line 279:
```suggestion
Make sure the component has the correct [component structure](#component-structure) before continuing.
```

I don't think you wanted to use a quote block here.

.github/COMPONENT-GUIDE.md
```diff
@@ -336,6 +267,78 @@ ion-button {
...
+
+#### JavaScript
+
+The `ion-activatable` class needs to be set on an element that can be activated:
+
+```jsx
+render() {
+  return (
+    <Host class='ion-activatable'>
```
sean-perkins, line 288:
```suggestion
    <Host class="ion-activatable">
```

~ Our prettier formatting is double quotes, figured we should be consistent in the code snippets here.

.github/COMPONENT-GUIDE.md
```diff
@@ -336,6 +267,78 @@ ion-button {
...
+  );
+}
+```
+
+Once that is done, the element will get the `ion-activated` class added on press after a small delay. This delay exists so that the active state does not show up when an activatable element is tapped while scrolling.
+
+In addition to setting that class, `ion-activatable-instant` can be set in order to have an instant press with no delay:
+
+```jsx
+<Host class='ion-activatable ion-activatable-instant'>
```
sean-perkins, line 300:
```suggestion
<Host class="ion-activatable ion-activatable-instant">
```

sean-perkins (changes requested):

.github/COMPONENT-GUIDE.md
brandyscarney: I didn't add this section with this PR, I just moved the section which is why it looks like this was a new addition. I can update all of these though.

brandyscarney: Updated in https://github.com/ionic-team/ionic-framework/pull/29111/commits/c1797b87b65aeebf1d30eecd244e272b8629e8aa

sean-perkins (approved):

PR #29123, opened 2024-03-07
feat: remove css animations support for ionic animations

sean-perkins:
Issue number: Internal

---------

Ionic Framework provides a small utility wrapper around the Web Animations API. Historically not all browsers that Ionic Framework supported, had support for the Web Animations API. To offer backwards compatibility, Ionic Framework provided fallback behaviors for the different wrapped APIs.

## What is the new behavior?

- Removes the legacy CSS animations fallback behavior from the Web Animations API animation utility. Maintaining a few no-op behaviors for test environments. 
- Resolved a few internal type usages that were casting to any
- Removed spec tests that were testing the fallback CSS animations behavior and/or already had test coverage from other unit tests.

- [x] Yes
- [ ] No

All modern browsers support the Web Animations API today. If a developer needs to target an older browser that does not support Web Animations, they should either use [a polyfill](https://github.com/web-animations/web-animations-js), or implement the fallback behavior themselves.


core/src/utils/animation/animation.ts
```diff
@@ -979,8 +860,6 @@ export const createAnimation = (animationId?: string): Animation => {
     if (supportsWebAnimations) {
       setAnimationStep(0);
       updateWebAnimation();
-    } else {
```
liamdebeasi, line 982: Can we add some comments where we still use the `supportsWebAnimations` checks (or even a comment where the const is defined) that explains why we still use this support check?

core/src/utils/animation/animation.ts
```diff
@@ -950,21 +840,12 @@ export const createAnimation = (animationId?: string): Animation => {
          * element too quickly
          */
         raf(() => {
-          clearCSSAnimationPlayState();
```
liamdebeasi, line 953: Do we still need the code above? It looks to be specific to CSS animations and display: none

liamdebeasi (changes requested):

core/src/utils/animation/animation.ts
sean-perkins: Some level of fallback behavior is required to satisfy existing unit tests: https://github.com/ionic-team/ionic-framework/actions/runs/8333321414/job/22804541495.

I think we have two options:
1. Have the minimal code to satisfy the unit test: https://github.com/ionic-team/ionic-framework/pull/29123/commits/72a96622c99c2d5271ba277fa784e31eb66962e3
2. Remove the unit tests

I've done option 1 to reduce the number of changes. However I question the value of these tests at all. They would have only tested the fallback CSS behavior, not the web animation behavior. 

If you are in favor, I'd propose removing those (can be either part of this PR or a separate ticket).

liamdebeasi (approved):

PR #29132, opened 2024-03-09
feat: add ionic theme architecture

sean-perkins:
Issue number: Internal

---------

## What is the new behavior?

Adds the base architecture to add a new theme configuration to Ionic Framework components. 
- Components can now specify an additional stylesheet for the `ionic` theme. 
- Developers can specify the `theme` and `mode` independently to control look and feel of a component.

Test infrastructure has been updated to add support for testing the theme configuration with Playwright. 
- Existing `themes` test configuration has been renamed to `palettes`

This PR is just the initial effort to decouple Ionic's architecture to separate look and feel and allow our dev team to start introducing the new component appearance to the UI. There will be additional changes required to completely add support for the Ionic theme. These changes are targeted against the `next` branch and are not expected to be used in a production environment at this time. 

- [ ] Yes
- [x] No


core/src/components/action-sheet/action-sheet-interface.ts
```diff
@@ -8,7 +8,15 @@ export interface ActionSheetOptions {
   backdropDismiss?: boolean;
   translucent?: boolean;
   animated?: boolean;
+  /**
+   * @deprecated To change the default appearance of the popover, use the `theme` option.
```
sean-perkins, line 12: Talked with Liam, we will pull this deprecation back for now. I'll need to add additional runtime validations for now to make sure developers are not using an invalid `mode` and `theme` configuration. 

This applies to all controller overlays.

bmarcelino-fe (approved):

core/src/global/ionic-global.ts
```diff
@@ -1,16 +1,114 @@
...
+import { getMode, setMode, setPlatformHelpers, getElement } from '@stencil/core';
+import { printIonWarning } from '@utils/logging';
 
-import type { IonicConfig, Mode } from '../interface';
+import type { IonicConfig, Mode, Theme } from '../interface';
 import { isPlatform, setupPlatforms } from '../utils/platform';
 
 import { config, configFromSession, configFromURL, saveConfig } from './config';
 
-// TODO(FW-2832): types
```
amandaejohnston, line 8: This file still has several `any` types that need replacing, so I don't think this TODO should be removed yet.

core/src/global/ionic-global.ts
```diff
@@ -1,16 +1,114 @@
...
+/**
+ * Returns the theme value of the element reference or the closest
+ * parent with a valid theme.
+ *
+ * @param ref The element reference to look up the theme for.
+ * @returns The theme value for the element reference, defaults to
+ * the default theme if it cannot be determined.
+ */
+export const getIonTheme = (ref?: any): Theme => {
+  return (ref && getMode(ref)) || defaultTheme;
```
amandaejohnston, line 111: Should this be using `getMode()`? Doesn't that mean that the theme is read from the `mode` prop?

amandaejohnston, line 111: The design doc specified that if the `theme` isn't given, the fallback should match the `mode`. Right now, it's always defaulting to `md`.

core/src/utils/test/playwright/generator.ts
```diff
@@ -38,61 +46,89 @@ interface TestConfigOption {
...
+  processedModes.forEach((mode) => {
+    processedTheme.forEach((theme) => {
+      if (mode === 'md' && theme === 'ios') {
+        // Prevent using Android platform with iOS theme
+        return;
+      }
+      if (mode === 'ios' && theme === 'md') {
+        // Prevent using iOS platform with Material Design theme
+        return;
+      }
```
amandaejohnston, line 127: Would it be possible to use the new `isModeValidForTheme` util in `ionic-global.ts` instead, to make this code easier to maintain?

amandaejohnston (changes requested):

core/src/utils/config.ts
```diff
@@ -18,10 +18,18 @@ export interface IonicConfig {
   rippleEffect?: boolean;
 
   /**
-   * The mode determines which platform styles to use for the whole application.
+   * The mode determines the platform behaviors to use for the whole application.
+   *
+   * Without a theme specified, mode will fallback to determining which platform
+   * styles to use for the whole application.
```
brandyscarney, line 24:
```suggestion
   * The mode determines the platform behaviors to use for the entire application.
   *
   * Without a specified theme, the mode will be used to determine which platform
   * styles to use throughout the application.
```

I don't know if this is what you are trying to say, I am a bit confused here.

core/src/global/ionic-global.ts
sean-perkins: Stencil calls the different `styleUrl` configurations as `modes` and provide a utility method to parse that state from the component with `getMode`: https://github.com/ionic-team/stencil/blob/b6b96175c5e6a7d3477ed5fc2d4ddfc17827dd63/src/runtime/mode.ts#L10

Not be confused with what we call `getIonMode`.

core/src/components/buttons/buttons.tsx
```diff
@@ -1,13 +1,17 @@
 import type { ComponentInterface } from '@stencil/core';
 import { Component, Host, Prop, h } from '@stencil/core';
 
-import { getIonMode } from '../../global/ionic-global';
+import { getIonTheme } from '../../global/ionic-global';
 
+/**
+ * @virtualProp {"ios" | "md" | "ionic"} theme - The visual appearance of the component.
```
thetaPC, line 7: Does this need a virtual prop for `mode`?

core/src/global/ionic-global.ts
sean-perkins: The remaining usages of `any` are valid and unlikely to be provided a more specific type. 

- `ref` - Stencil specific instance, there is no type for this. Stencil internally types the argument of `getMode`, which is `ref` as an `any`.
- `window` - we are applying a custom property to the window global. Casting as any is how we do this for Capacitor as well.
- `platformHelpers` - this type also is not exposed by Stencil and is passed to a Stencil utility function.

core/src/components/action-sheet/action-sheet-interface.ts
sean-perkins: I don't believe configuring `mode` makes sense from the overlay controllers. That is what the "for now" refers to. We will do further exploration if developers really only need to configure the visual appearance and would use `theme` in the future to do that. Changing the platform behavior should be something developers rarely or never need to do. 

Overlay controllers have always had more structured and limited APIs. The advantage of _not_ exposing `mode` is that we don't need additional JS overhead of validating configurations.

I'm still working on the runtime validations locally. Once that is ready I will push that work up, along with the updated overlay interfaces and deprecation removals.

core/src/components/infinite-scroll-content/infinite-scroll-content.tsx
```diff
@@ -4,15 +4,19 @@ import { ENABLE_HTML_CONTENT_DEFAULT } from '@utils/config';
 import { sanitizeDOMString } from '@utils/sanitization';
 
 import { config } from '../../global/config';
-import { getIonMode } from '../../global/ionic-global';
+import { getIonTheme } from '../../global/ionic-global';
 import type { IonicSafeString } from '../../utils/sanitization';
 import type { SpinnerTypes } from '../spinner/spinner-configs';
 
+/**
+ * @virtualProp {"ios" | "md" | "ionic"} theme - The visual appearance of the component.
```
thetaPC, line 12: Missing mode virtual prop?

core/src/utils/test/playwright/generator.ts
```diff
@@ -8,15 +10,16 @@ export type Direction = 'ltr' | 'rtl';
...
-export type Theme = 'light' | 'dark' | 'high-contrast' | 'high-contrast-dark';
+export type ThemeMode = 'light' | 'dark' | 'high-contrast' | 'high-contrast-dark';
 
 export type TitleFn = (title: string) => string;
 export type ScreenshotFn = (fileName: string) => string;
 
 export interface TestConfig {
-  mode: Mode;
   direction: Direction;
+  themeMode: ThemeMode;
```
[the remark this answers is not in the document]
sean-perkins: I think our whole language model is unsalvageable with our desired goal. This was just to unblock the testing infra to use `theme` for what we are calling themes. 

I'm planning on having a session with the team to align on short term and long term language. I like the idea of `colorScheme`, but we should try to get this language organized before v8 ships... since we are shipping a lot of feature scope under something we are calling "themes" (light, dark, high contrast).

core/src/utils/test/playwright/generator.ts
sean-perkins: I can look further into this. Originally we avoided coupling any source code to the test infrastructure to avoid bundling the test infra in the published package.

core/src/components/loading/loading-interface.ts
```diff
@@ -11,7 +11,12 @@ export interface LoadingOptions {
   translucent?: boolean;
   animated?: boolean;
   backdropDismiss?: boolean;
+  /**
+   * @deprecated To change the default appearance of the popover, use the `theme` option.
```
thetaPC, line 15:
```suggestion
   * @deprecated To change the default appearance of the loading component, use the `theme` option.
```

brandyscarney (commented):
I need to take a closer look at the implementation in `ionic-global.ts` and test it out but I reviewed all of the other files.

core/src/components/menu-button/menu-button.tsx
```diff
@@ -109,10 +111,11 @@ export class MenuButton implements ComponentInterface, ButtonInterface {
         <button {...attrs} disabled={disabled} class="button-native" part="native" aria-label={ariaLabel}>
           <span class="button-inner">
             <slot>
-              <ion-icon part="icon" icon={menuIcon} mode={mode} lazy={false} aria-hidden="true"></ion-icon>
+              {/* TODO double check the mode/theme attribute */}
```
thetaPC, line 114: Does this have a Jira ticket? Add it, if possible.

core/src/components/modal/modal-interface.ts
```diff
@@ -11,7 +11,12 @@ export interface ModalOptions<T extends ComponentRef = ComponentRef> {
   animated?: boolean;
   canDismiss?: boolean | ((data?: any, role?: string) => Promise<boolean>);
 
+  /**
+   * @deprecated To change the default appearance of the popover, use the `theme` option.
```
thetaPC, line 15:
```suggestion
   * @deprecated To change the default appearance of the modal, use the `theme` option.
```

core/src/components/nav/nav-interface.ts
```diff
@@ -34,7 +34,12 @@ export interface RouterOutletOptions {
   showGoBack?: boolean;
   direction?: NavDirection;
   deepWait?: boolean;
+  /**
+   * @deprecated To change the default appearance of the popover, use the `theme` option.
```
thetaPC, line 38:
```suggestion
   * @deprecated To change the default appearance of the nav, use the `theme` option.
```

core/src/components/searchbar/searchbar.tsx
```diff
@@ -570,14 +572,15 @@ export class Searchbar implements ComponentInterface {
...
+        tabIndex={theme === 'ios' && !shouldShowCancelButton ? -1 : undefined}
         onMouseDown={this.onCancelSearchbar}
         onTouchStart={this.onCancelSearchbar}
         class="searchbar-cancel-button"
       >
         <div aria-hidden="true">
-          {mode === 'md' ? (
-            <ion-icon aria-hidden="true" mode={mode} icon={this.cancelButtonIcon} lazy={false}></ion-icon>
+          {theme === 'md' ? (
+            // TODO look into the mode/theme
```
thetaPC, line 582: Jira ticket?

core/src/components/toast/toast-interface.ts
```diff
@@ -17,7 +17,12 @@ export interface ToastOptions {
   layout?: ToastLayout;
 
   color?: Color;
+  /**
+   * @deprecated To change the default appearance of the popover, use the `theme` option.
```
thetaPC, line 21:
```suggestion
   * @deprecated To change the default appearance of the toast, use the `theme` option.
```

thetaPC (changes requested):

.github/COMPONENT-GUIDE.md
thetaPC, line 1: Should it also give an explanation between `theme` and `mode`?

core/src/components/backdrop/backdrop.tsx
```diff
@@ -1,13 +1,17 @@
 import type { ComponentInterface, EventEmitter } from '@stencil/core';
 import { Component, Event, Host, Listen, Prop, h } from '@stencil/core';
 
-import { getIonMode } from '../../global/ionic-global';
+import { getIonTheme } from '../../global/ionic-global';
 
+/**
+ * @virtualProp {"ios" | "md" | "ionic"} theme - The visual appearance of the component.
```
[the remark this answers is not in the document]
sean-perkins: 1. It was missing originally.
2. `theme` replaces the `mode` implementation for backdrop. It only uses the `theme` to derive the style urls applied. There is no implementation associated with the `mode` and backdrop.

core/src/components/buttons/buttons.tsx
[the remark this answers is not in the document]
sean-perkins: Same answer as: https://github.com/ionic-team/ionic-framework/pull/29132#discussion_r1520488732

However we _could_ list the virtual prop since the logic crawls to the nearest element with a `mode` attribute.

core/src/components/menu-button/menu-button.tsx
```diff
@@ -109,10 +111,11 @@ export class MenuButton implements ComponentInterface, ButtonInterface {
...
+              <ion-icon part="icon" icon={menuIcon} mode={theme} lazy={false} aria-hidden="true"></ion-icon>
```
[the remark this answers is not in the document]
sean-perkins: Seems to be related to applying proper styling when used inside `ion-buttons`: https://github.com/ionic-team/ionic-framework/commit/811eee70f83e5bec6e79ef8dd22fb8e9c0b2ce07

However I don't see why specifying the `mode` would be needed for that.

core/src/components/thumbnail/thumbnail.tsx
```diff
@@ -1,7 +1,7 @@
 import type { ComponentInterface } from '@stencil/core';
 import { Component, Host, h } from '@stencil/core';
 
-import { getIonMode } from '../../global/ionic-global';
+import { getIonTheme } from '../../global/ionic-global';
```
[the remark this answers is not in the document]
sean-perkins: `ion-thumbnail` does not have different styles applied for the different themes, so documenting the property is not required.

core/src/utils/test/playwright/generator.ts
```diff
@@ -28,6 +31,11 @@ interface TestUtilities {
 interface TestConfigOption {
   modes?: Mode[];
   directions?: Direction[];
+  themeModes?: ThemeMode[];
+  /**
+   * The individual themes (iOS, Material Design and Ionic) to test
+   * against. If unspecified, defaults iOS and Material Design
```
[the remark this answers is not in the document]
sean-perkins: This would create a significant number of new screenshots. This behavior only exists for backwards compatibility. 

Ideally we do not test against any specific theme by default and the developer specifies the theme they would like to test against.

core/src/components/accordion-group/accordion-group.tsx
```diff
@@ -2,18 +2,20 @@ import type { ComponentInterface, EventEmitter } from '@stencil/core';
...
- * @virtualProp {"ios" | "md"} mode - The mode determines which platform styles to use.
+ * @virtualProp {"ios" | "md"} mode - The mode determines the platform behaviors of components.
+ * @virtualProp {"ios" | "md" | "ionic"} theme - The visual appearance of the component.
  */
 @Component({
   tag: 'ion-accordion-group',
   styleUrls: {
     ios: 'accordion-group.ios.scss',
     md: 'accordion-group.md.scss',
+    ionic: 'accordion-group.md.scss',
```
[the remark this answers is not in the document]
sean-perkins: Yes. Every component using `styleUrls` needs to assign a default stylesheet for the ionic theme. Otherwise the component will have no style representation. 

As developers implement the ionic theme, they will create a new stylesheet and update this reference.

core/src/utils/test/playwright/generator.ts
```diff
@@ -38,61 +46,89 @@ interface TestConfigOption {
...
+     * when the theme and mode are the same.
      */
-    return `${title} - ${mode}/${direction}`;
+
+    if (themeMode === 'light') {
+      /**
+       * Ionic has many existing tests that existed prior to
+       * the introduction of theme testing. To maintain backwards
+       * compatibility, we will not include the theme in the test
+       * title if the theme is set to light.
```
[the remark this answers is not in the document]
sean-perkins: None currently. It would be less of a burden for test titles and a huge pain point for screenshot renames.

core/src/components/action-sheet/action-sheet.tsx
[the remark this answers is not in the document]
sean-perkins: Great list 👍 

- `presentingElement` is still based on mode detection currently. We may adjust when working on overlays as part of the new theme.

core/src/components/input/input.tsx
```diff
@@ -768,7 +770,7 @@ export class Input implements ComponentInterface {
                 }}
                 onClick={this.clearTextInput}
               >
-                <ion-icon aria-hidden="true" icon={mode === 'ios' ? closeCircle : closeSharp}></ion-icon>
+                <ion-icon aria-hidden="true" icon={theme === 'ios' ? closeCircle : closeSharp}></ion-icon>
```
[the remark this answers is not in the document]
brandyscarney: It might've been written before we added support for those properties on `ion-icon`, I was just checking!

core/src/components/accordion/accordion.tsx
```diff
@@ -14,7 +14,8 @@ const enum AccordionState {
 }
 
 /**
- * @virtualProp {"ios" | "md"} mode - The mode determines which platform styles to use.
+ * @virtualProp {"ios" | "md"} mode - The mode determines the platform behaviors of components.
```
brandyscarney, line 17: Nit: can we change this to 

```suggestion
 * @virtualProp {"ios" | "md"} mode - The mode determines the platform behaviors of the component.
```

to match the theme description

core/src/global/ionic-global.ts
```diff
@@ -64,27 +165,53 @@ export const initialize = (userConfig: IonicConfig = {}) => {
...
 
-  const isIonicElement = (elm: any) => elm.tagName?.startsWith('ION-');
-
-  const isAllowedIonicModeValue = (elmMode: string) => ['ios', 'md'].includes(elmMode);
-
   setMode((elm: any) => {
     while (elm) {
+      const theme = (elm as any).theme || elm.getAttribute('theme');
+
+      if (theme) {
```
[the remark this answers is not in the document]
sean-perkins: What dev audience are you thinking of here? This behavior is consistent with what `modes` were doing previously, where we iterate over all element nodes and cascade the theme/mode to elements that do not have it explicitly specified (in the Stencil metadata).

core/src/components/action-sheet/action-sheet-interface.ts
amandaejohnston: Doesn't this mean that developers can no longer change the `mode` on controller overlays, but they still can on inline ones? Wouldn't we want the two APIs to be as in sync as possible?

core/src/components/action-sheet/action-sheet.tsx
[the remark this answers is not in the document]
amandaejohnston: On closer inspection, `presentingElement` is dependent on both mode *and* theme currently. It is used to tell whether a card modal should be rendered [here](https://github.com/ionic-team/ionic-framework/blob/sp/look-and-feel/core/src/components/modal/modal.tsx#L914), which is then used to configure the classes` modal-default` and `modal-card`. The styling for these classes is theme-dependent.

core/src/components/header/header.tsx
```diff
@@ -38,15 +40,15 @@ export class Header implements ComponentInterface {
 
   /**
    * Describes the scroll effect that will be applied to the header.
-   * Only applies in iOS mode.
+   * Only applies when the team is `"ios"`.
```
amandaejohnston, line 43:
```suggestion
   * Only applies when the theme is `"ios"`.
```

core/src/utils/test/playwright/generator.ts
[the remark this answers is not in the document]
amandaejohnston: We changed it to `palette` in separate PRs, so this is good to go 👍

core/src/components/buttons/buttons.tsx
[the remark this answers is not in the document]
amandaejohnston: Bump? Wondering what your thoughts are on this. At minimum we should probably update the `collapse` comment to specify that both the mode and theme need to be set.

core/src/global/ionic-global.ts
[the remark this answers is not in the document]
amandaejohnston: To be fair, I had to ask about it 😆 A comment would be useful for other team members, including people hired in the future.

core/src/components/action-sheet/action-sheet-interface.ts
amandaejohnston: Nevermind, just noticed you put the mode prop back through `OverlayOptions` 👍

core/src/components/header/header.tsx
sean-perkins: 🤣 been typing `theme` so much, that `team` looked too similar 😭

sean-perkins: Addressed here: https://github.com/ionic-team/ionic-framework/pull/29132/commits/caa5490a5679a0357dea81d3b477c1af5322ffba

amandaejohnston (changes requested):
Sorry about the multiple rounds of reviews, didn't know Github would submit some of my comments early 😅

I also have a couple earlier comments that haven't been fully addressed yet, re: `presentingElement` on modal and `collapse` on buttons.

core/src/components/buttons/buttons.tsx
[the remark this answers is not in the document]
sean-perkins: Created `FW-6047` to track. This doesn't need to be solved right now. We should do a focused investigation around that experience and determine if we support mixing `mode` and `theme` with that feature set and if we do, what changes we need to make to resolve that.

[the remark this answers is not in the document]
amandaejohnston: Works for me 👍

core/src/components/input/test/basic/index.html
```diff
@@ -48,7 +48,7 @@
         <div class="grid">
           <div class="grid-item">
             <h2>Default</h2>
-            <ion-input value="hi@ionic.io" label="Email"></ion-input>
+            <ion-input value="hi@ionic.io" label="Email" clear-input="true"></ion-input>
```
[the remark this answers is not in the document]
sean-perkins: Nope. I cannot remember why I made this change & it doesn't fit with the commit I had it in. I think I was testing some inconsistencies with the mode fallback behaviors and forgot to revert this.

core/src/global/ionic-global.ts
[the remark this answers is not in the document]
sean-perkins: Added a comment here: https://github.com/ionic-team/ionic-framework/pull/29132/commits/0b2cc494da41b5e943cf90460ba38e138acf58cf

core/src/utils/test/playwright/generator.ts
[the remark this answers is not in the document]
thetaPC: Sounds good to me

core/src/utils/test/playwright/generator.ts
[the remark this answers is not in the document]
thetaPC: I don't even want to think about the screenshot renames 😭

core/src/components/accordion-group/accordion-group.tsx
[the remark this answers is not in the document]
thetaPC: Makes sense, good choice to use `md`.

thetaPC (commented):
My comment is a suggestion. Regardless of what you decide, request me when it's ready.

amandaejohnston (approved):
Can we create another ticket to look into the `presentingElement` issue I mentioned in https://github.com/ionic-team/ionic-framework/pull/29132#discussion_r1525221442, or update the existing one for `collapse` to cover it as well? Otherwise this LGTM.

sean-perkins:
@amandaejohnston created FW-6049 to track. Feel free to create spikes for tasks like that as you think they make sense, especially if you have context from your comments.

core/src/components/router-outlet/router-outlet.tsx
```diff
@@ -19,6 +19,9 @@ import type {
 import type { RouterOutletOptions, SwipeGestureHandler } from '../nav/nav-interface';
 import type { RouteID, RouterDirection, RouteWrite, NavOutlet } from '../router/utils/interface';
 
+/**
+ * @virtualProp {"ios" | "md" | "ionic"} theme - The theme determines the visual appearance of the component.
+ */
```
mapsandapps, line 24: This component also doesn't have a virtualProp for mode. Should it?

mapsandapps (changes requested):
As well as the inline comments, it looks like the following components are missing an ionic theme in the `styleUrls`:
- Datetime Button
- Picker Column Option

core/src/components/router-outlet/router-outlet.tsx
sean-perkins: This component has an explicit property for `mode` on L45.

core/src/components/infinite-scroll-content/infinite-scroll-content.tsx
```diff
@@ -4,15 +4,19 @@ import { ENABLE_HTML_CONTENT_DEFAULT } from '@utils/config';
...
 
 import { config } from '../../global/config';
-import { getIonMode } from '../../global/ionic-global';
+import { getIonTheme } from '../../global/ionic-global';
 import type { IonicSafeString } from '../../utils/sanitization';
 import type { SpinnerTypes } from '../spinner/spinner-configs';
 
+/**
+ * @virtualProp {"ios" | "md" | "ionic"} theme - The theme determines the visual appearance of the component.
+ */
```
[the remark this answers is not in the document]
sean-perkins: It won't hurt anything to add it, but it didn't exist before.

mapsandapps (approved):
Looks good!

brandyscarney (approved):

PR #29136, opened 2024-03-11
chore: reproduction URL is required

liamdebeasi:
The team would like to make issue URLs required for bug reports. Having a runnable sample makes it easy for the team to quickly triage bugs. Previously we accepted code snippets, but recently we've found that the snippets are insufficient to reproduce the issue.


liamdebeasi:
Closing in favor of https://github.com/ionic-team/ionic-framework/pull/29135

PR #29140, opened 2024-03-12
fix(overlay): do not hide overlay if toast is presented

### LINKED ISSUE #29139: bug: Overlays now have aria-hidden=true when displaying a Toast
### Prerequisites

### Ionic Framework Version

v7.x

### Current Behavior

In issue https://github.com/ionic-team/ionic-framework/issues/23472, the behaviour of overlays was modified to add `aria-hidden=true` to existing overlays when displaying a new overlay. IMO this makes sense for modals and popovers, but this shouldn't be done when displaying a Toast. Toasts are overlays that don't block the interaction with the other elements in the DOM, so the rest of overlays should remain active and visible.

### Expected Behavior

Displaying a Toast should not add aria-hidden=true to other overlays.

### Steps to Reproduce

1. In an Ionic app, display any modal using ModalController.create.
2. While the modal is displayed, show a toast with a long duration using ToastController.create.
3. Inspect the DOM and search for the ion-modal. Check that it has aria-hidden="true" while the Toast is displayed.

### Code Reproduction URL

https://github.com/dpalou/overlays-toast-bug

### Ionic Info

Ionic:

   Ionic CLI                     : 7.2.0 (/Users/dpalou/.nvm/versions/node/v18.18.2/lib/node_modules/@ionic/cli)
   Ionic Framework               : @ionic/angular 7.7.4
   @angular-devkit/build-angular : 17.2.3
   @angular-devkit/schematics    : 17.2.3
   @angular/cli                  : 17.2.3

Capacitor:

   Capacitor CLI      : 5.7.2
   @capacitor/core    : 5.7.2

Utility:

   cordova-res : 0.15.4
   native-run  : 2.0.1

System:

   NodeJS : v18.18.2 (/Users/dpalou/.nvm/versions/node/v18.18.2/bin/node)
   npm    : 9.8.1
   OS     : macOS Unknown

### Additional Information

_No response_

liamdebeasi:
Issue number: resolves #29139 

---------

When implementing https://github.com/ionic-team/ionic-framework/pull/28997 we did not consider the case where a Toast could be presented. When presenting a Toast after presenting a Modal the linked change causes the Modal to be hidden from screen readers.

## What is the new behavior?

- If the top-most overlay is a Toast then the closest non-Toast overlay is also not hidden from screen readers. 

- [ ] Yes
- [x] No

## Other information

Dev build: `7.7.5-dev.11710260658.1fc29a6c`

### DIFF
--- a/core/src/utils/overlays.ts
+++ b/core/src/utils/overlays.ts
@@ -975,7 +975,7 @@ const hideOverlaysFromScreenReaders = (newTopMostOverlay: HTMLIonOverlayElement)
     if (nextPresentedOverlay.hasAttribute('aria-hidden')) {
       presentedOverlay.setAttribute('aria-hidden', 'true');
       /**
-       * If the next overlay is a Toast this does not have aria-hidden then current overlay
+       * If the next overlay is a Toast that does not have aria-hidden then current overlay
        * should not have aria-hidden either so focus can remain in the current overlay.
        */
     } else if (nextPresentedOverlay.tagName !== 'ION-TOAST') {
--- a/core/src/utils/overlays.ts
+++ b/core/src/utils/overlays.ts
@@ -992,7 +992,6 @@ const hideOverlaysFromScreenReaders = (newTopMostOverlay: HTMLIonOverlayElement)
 const revealOverlaysToScreenReaders = () => {
   if (doc === undefined) return;
 
-  // If there are other overlays presented, unhide the new topmost one from screen readers.
   const overlays = getPresentedOverlays(doc);
 
   for (let i = overlays.length - 1; i >= 0; i--) {
--- a/core/src/utils/overlays.ts
+++ b/core/src/utils/overlays.ts
@@ -1003,15 +1003,14 @@ const revealOverlaysToScreenReaders = () => {
      * could be more Toasts underneath. Additionally, we need to unhide the closest non-Toast
      * overlay too so focus can move there since focus is never automatically moved to the Toast.
      */
-    if (currentOverlay.tagName === 'ION-TOAST') {
-      currentOverlay.removeAttribute('aria-hidden');
-      /**
-       * If we found a non-Toast element then we can just remove aria-hidden and stop searching entirely
-       * since this overlay should always receive focus. As a result, all underlying overlays should still
-       * be hidden from screen readers.
-       */
-    } else {
-      currentOverlay.removeAttribute('aria-hidden');
+    currentOverlay.removeAttribute('aria-hidden');
+    
+    /**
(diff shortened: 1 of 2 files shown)


core/src/utils/overlays.ts
```diff
@@ -966,3 +951,69 @@ export const createTriggerController = () => {
...
+  if (doc === undefined) return;
+
+  const overlays = getPresentedOverlays(doc);
+
+  for (let i = overlays.length - 1; i >= 0; i--) {
+    const presentedOverlay = overlays[i];
+    const nextPresentedOverlay = overlays[i + 1] ?? newTopMostOverlay;
+
+    // If next overlay has aria-hidden then all remaining overlays will have it too.
+    if (nextPresentedOverlay.hasAttribute('aria-hidden')) {
```
liamdebeasi, line 975: This may cause issues if the developer decides to present a new overlay with `aria-hidden`. However, I don't know how realistic that scenario is because then screen readers would never be able to focus the new overlay even without this change.

amandaejohnston, line 975: This and the `else` block both do the same thing, so the two conditions can be merged into a single `||` statement.

core/src/utils/overlays.ts
```diff
@@ -966,3 +951,69 @@ export const createTriggerController = () => {
...
+
+/**
+ * When dismissing an overlay we need to reveal the new top-most overlay to screen readers.
+ * If the top-most overlay is a Toast we potentially need to reveal more overlays since
+ * focus is never automatically moved to the Toast.
+ */
+const revealOverlaysToScreenReaders = () => {
+  if (doc === undefined) return;
+
+  // If there are other overlays presented, unhide the new topmost one from screen readers.
```
amandaejohnston, line 995:
```suggestion
```
This comment no longer has anything to do with the code it's following, and is made redundant by the function's top-level comments.

core/src/utils/overlays.ts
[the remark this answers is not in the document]
liamdebeasi: Here's an example:

```
(bottom-most overlay)
modal one
toast one
(top-most overlay)
```

When `toast one` is presented, `modal one` does **not** have `aria-hidden="true"`. If we then present `modal two` then `modal-one` should be hidden from screen readers. However, if we only checked if the next overlay was not a Toast then this wouldn't happen.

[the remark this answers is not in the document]
amandaejohnston: Ah, got it! Thank you, that makes sense 👍

amandaejohnston (approved):

sean-perkins (approved):

PR #29141, opened 2024-03-12
feat(input): add showPasswordToggle, hidePasswordIcon and showPasswordIcon properties

OS-giulianasilva:
Issue number: Internal

---------

The ion-input does not provide an ability to toggle password visibility on input

## What is the new behavior?
- Add three new properties:
    - `showPasswordToggle` to enable the passwordToggle button in the input that when clicked toggles the input type between `"text"` and `"password"`, allowing the user to view or hide the input value.
    - `showPasswordIcon` to set the icon to be used to represent showing a password.
    - `hidePasswordIcon` to set the icon to be used to represent hiding a password.

- [ ] Yes
- [x] No

## Other information
Behavior on MD:
passwordToggleMD
Behavior on iOS:
passwordToggleios


core/src/components/input/test/password-toggle/input.e2e.ts
```diff
@@ -0,0 +1,68 @@
+import { expect } from '@playwright/test';
+import { configs, test } from '@utils/test/playwright';
+
+configs().forEach(({ title, screenshot, config }) => {
+  test.describe(title('input: password toggle button'), () => {
+    test('should render password icon when show-password-toggle', async ({ page }) => {
```
OS-giulianasilva, line 6: I was not able to commit the screenshots, so I am not sure if I should keep this test.

core/src/components/input/input.tsx
```diff
@@ -836,6 +857,30 @@ export class Input implements ComponentInterface {
                 <ion-icon aria-hidden="true" icon={mode === 'ios' ? closeCircle : closeSharp}></ion-icon>
               </button>
             )}
+            {this.showPasswordToggle && !readonly && !disabled && (
```
sean-perkins, line 860: The design document described rendering this content as the default content of the `end` slot. 

Did we experience an issue or additional consideration to not render it as the default slot content?

core/src/components/input/input.tsx
```diff
@@ -836,6 +857,30 @@ export class Input implements ComponentInterface {
...
+              <button
+                aria-label="show password"
+                aria-checked={this.type === 'text' ? 'true' : 'false'}
```
sean-perkins, line 863: Since the implementation is accessing `type` multiple times in the function, it would be more performant to deconstruct `type` above and reference that variable reference. 

Accessing the variable from `this` will look up that reference in memory each time you access the property from it. 

For example:

```tsx
const { type } = this;

return (
  <button aria-checked={type === 'text' ? 'true' : 'false'}></button>
);
```

sean-perkins (commented):
Couple of initial questions, great work so far!

I've updated the targeted branch to `feature-8.0` since we will not be releasing any additional minor release changes other than what we've already committed and v8 is close 🎉 

You will need to either merge `feature-8.0` into your branch and resolve the merge conflicts or rebase. `api.txt`, `components.d.ts` and the `proxies.ts` are auto-generated from `npm run build`, so you can take whatever diff and then re-generate the results. 

Once we are up to date with `feature-8.0` I can help get screenshots generated and merged into your fork. The team is actively working to remove this restraint in the near future so you can update screenshots locally.

core/src/components/input/input.scss
```diff
@@ -202,11 +203,11 @@
 
   color: $text-color-step-400;
 
-  visibility: hidden;
   appearance: none;
 }
 
-:host(.in-item-color) .input-clear-icon {
+:host(.in-item-color) .input-clear-icon,
+:host(.in-item-color) .input-password-toggle {
```
JoaoFerreira-FrontEnd, line 210: let's break this into

```
:host(.in-item-color){
    .input-clear-icon,
    .input-password-toggle {  
    } 
}
```

JoaoFerreira-FrontEnd (changes requested):

core/src/components/input/input.tsx
OS-giulianasilva: Actually, this is something that I overlooked in the design document. But now testing it as the default content of the end slot, I noticed that the password toggle button will be hidden when the end slot is used. 

I think this would impact some use-cases. For example, if the developer wants to use the password toggle and also add a green checkmark icon to the end slot to represent the validation of the strength of the password, it will not be possible to achieve.

Please let me know your thoughts on it.

core/src/components/input/input.tsx
OS-giulianasilva: Done!

core/src/components/input/input.scss
OS-giulianasilva: I would like to keep it as it is, because it is more readable and also matches the current CSS structure of the code.

JoaoFerreira-FrontEnd: @sean-perkins what do you think about this topic?

core/src/components/input/input.tsx
liamdebeasi: > But now testing it as the default content of the end slot, I noticed that the password toggle button will be hidden when the end slot is used.

This is an intentional behavior. If you place the toggle password content in the end slot that's considered the "default content". Developers can override the default behavior to add their own logic (such as a custom password toggle button).

> I think this would impact some use-cases. For example, if the developer wants to use the password toggle and also add a green checkmark icon to the end slot to represent the validation of the strength of the password, it will not be possible to achieve.

This was [discussed](https://outsystemsrd.atlassian.net/wiki/spaces/RDMBVS/pages/3939729499?focusedCommentId=3990618567#comment-3990618567) in the design document. The concern from design was that you could potentially have 4 icons on the same line which will reduce the visible viewport for the input text. For example, you could have a start icon, an end icon, a clear input icon, and a validation icon all on the same line.

The idea here is that by using the end slot we can leverage the existing infrastructure and avoid the problem altogether.

core/src/components/input/input.tsx
```diff
@@ -238,6 +243,17 @@ export class Input implements ComponentInterface {
    */
   @Prop() shape?: 'round';
 
+  /**
+   * Set the icon that can be used to represent showing a password. Defaults to `eyeOutline`.
+   */
+  @Prop() showPasswordIcon?: string;
```
liamdebeasi, line 249: Why are we switching between icon variants here? The show password icon uses the outline variant, but the hide password uses the filled variant. Visually it looks a bit strange switching between variants.

core/src/components/input/input.tsx
```diff
@@ -541,6 +558,10 @@ export class Input implements ComponentInterface {
     this.emitInputChange(ev);
   };
 
+  private showPassword = () => {
```
liamdebeasi, line 561:
```suggestion
  private togglePasswordVisibility = () => {
```

Nit: This name isn't entirely accurate because depending on `this.type` it will either show or hide the password.

core/src/components/input/input.tsx
```diff
@@ -771,6 +792,30 @@ export class Input implements ComponentInterface {
                 <ion-icon aria-hidden="true" icon={mode === 'ios' ? closeCircle : closeSharp}></ion-icon>
               </button>
             )}
+            {this.showPasswordToggle && !readonly && !disabled && (
+              <button
+                aria-label="show password"
```
liamdebeasi, line 797: This aria label should change based on whether or not the password is visible

core/src/components/input/input.tsx
```diff
@@ -771,6 +792,30 @@ export class Input implements ComponentInterface {
...
+                aria-checked={type === 'text' ? 'true' : 'false'}
+                aria-controls={this.inputId}
```
liamdebeasi, line 799:
```suggestion
                aria-controls={inputId}
```

We've already deconstructed this before, so we can access the local variable instead.

core/src/components/input/test/password-toggle/input.e2e.ts
liamdebeasi: Let's keep it for now. We have some work in review right now that will unblock you on this front.

core/src/components/input/input.scss
liamdebeasi: Most CSS we've written in Ionic avoids nested to keep the code readable. However, I think either would be fine -- really comes down to a matter of preference.

core/src/components/input/input.scss
```diff
@@ -217,12 +218,13 @@
...
-.input-clear-icon:focus {
+.input-clear-icon:focus,
+.input-password-toggle:focus {
   opacity: 0.5;
 }
 
-:host(.has-value) .input-clear-icon {
-  visibility: visible;
+:host(:not(.has-value)) .input-clear-icon {
+  visibility: hidden;
```
liamdebeasi, line 227: Why was this change made?

core/src/components/input/test/password-toggle/input.e2e.ts
```diff
@@ -0,0 +1,68 @@
...
+
+      await toggle.click();
+      await page.waitForChanges();
+
+      await expect(nativeInput).toHaveAttribute('type', 'text');
+
+      await toggle.click();
+      await page.waitForChanges();
+
+      await expect(nativeInput).toHaveAttribute('type', 'password');
```
liamdebeasi, line 38: This could be a spec test. We don't necessarily benefit from having cross-browser testing here.

core/src/components/input/test/password-toggle/input.e2e.ts
```diff
@@ -0,0 +1,68 @@
...
+      const input = page.locator('ion-input');
+      const toggle = input.locator('.input-password-toggle');
+      const icon = toggle.locator('ion-icon');
+
+      await expect(icon).toHaveJSProperty('icon', 'bus');
+
+      await toggle.click();
+      await page.waitForChanges();
+
+      await expect(icon).toHaveJSProperty('icon', 'cafe');
```
liamdebeasi, line 65: Same as previous comment, we don't really value from having cross-browser testing here so this could be a spec test.

liamdebeasi (changes requested):
Great work so far!

liamdebeasi:
Also you'll want to run `npm run lint.fix` from `core` and push your changes so the `test-core-lint` step passes.

core/src/components/input/input.tsx
liamdebeasi: Ah I see! I missed that during my review. I'll converse with the team and see if we can come to a consensus.

core/src/components/input/input.tsx
OS-giulianasilva: I think this is not the case when we use the `switch` role. If the aria-label changes, the screen reader will read: 
> show password on
> hide password off

I don't know if it would be completely wrong, but it isn't very clear to me.

If we want the aria-label to change, I should drop the `switch` role and use it as a simple button. Both approaches are commonly used, and I'm fine with either.

core/src/components/input/input.tsx
OS-giulianasilva: True!

core/src/components/input/input.tsx
OS-giulianasilva: Done!

core/src/components/input/input.scss
OS-giulianasilva: Because otherwise, the clear icon would not disappear when the input is erased.

core/src/components/input/test/password-toggle/input.e2e.ts
OS-giulianasilva: Makes sense! I will take care of this tomorrow.

core/src/components/input/test/password-toggle/input.e2e.ts
OS-giulianasilva: Same as above.

core/src/components/input/input.tsx
liamdebeasi: Oh true that would be confusing 🤔 Ok let's keep it as-is for now.

core/src/components/input/input.scss
liamdebeasi: The clear icon currently disappears when the input is erased on the latest stable version of Ionic: https://ionicframework.com/docs/api/input#clear-options

If that's not the case in this branch, then there may be other code impacting that.

core/src/components/input/input.tsx
OS-giulianasilva: Let's talk about this internally.

core/src/components/input/test/password-toggle/input.e2e.ts
OS-giulianasilva: Done!

core/src/components/input/test/password-toggle/input.e2e.ts
OS-giulianasilva: Done!

thetaPC:
Thank you for contributing! I'm going to close this PR since the work was implemented through PR https://github.com/ionic-team/ionic-framework/pull/29141.

PR #29145, opened 2024-03-12
refactor(range, select): prefer labels passed by developer

liamdebeasi:
Issue number: resolves #

---------

## What is the new behavior?

-
-
-

- [ ] Yes
- [ ] No


.github/COMPONENT-GUIDE.md
brandyscarney, line 1: There is still a `getAriaLabel` usage in this file under Labels

liamdebeasi: Fixed! https://github.com/ionic-team/ionic-framework/pull/29145/commits/5c92b1acc2da92b7fa456394c430bb042ea08704

.github/COMPONENT-GUIDE.md
```diff
@@ -439,53 +439,35 @@ render() {
...
 
-A helper function has been created to get the proper `aria-label` for the checkbox. This can be imported as `getAriaLabel` like the following:
+Labels should be passed directly to the component in the form of either visible text or an `aria-label`. The visible text can be set inside of a `label` element, and the `aria-label` can be set directly on the interactive element.
 
-```tsx
-const { label, labelId, labelText } = getAriaLabel(el, inputId);
-```
-
-where `el` and `inputId` are the following:
+In the following example the `aria-label` can be inherited from the Host using the `inheritAttributes` or `inheritAriaAttributes` utilities. This allows developers to set `aria-label` on the host element since they do not have access to inside the shadow root.
```
brandyscarney, line 444: Should we add a note on why we are using `inheritAttributes` vs `inheritAriaAttributes`? Something like:

```
> [!NOTE]
> Use `inheritAttributes` to specify which attributes should be inherited or `inheritAriaAttributes` to inherit all of the possible `aria` attributes.
```

.github/COMPONENT-GUIDE.md
```diff
@@ -567,57 +549,37 @@ render() {
...
 
-A helper function has been created to get the proper `aria-label` for the switch. This can be imported as `getAriaLabel` like the following:
+Labels should be passed directly to the component in the form of either visible text or an `aria-label`. The visible text can be set inside of a `label` element, and the `aria-label` can be set directly on the interactive element.
 
-```tsx
-const { label, labelId, labelText } = getAriaLabel(el, inputId);
-```
-
-where `el` and `inputId` are the following:
+In the following example the `aria-label` can be inherited from the Host using the `inheritAttributes` or `inheritAriaAttributes` utilities. This allows developers to set `aria-label` on the host element since they do not have access to inside the shadow root.
```
brandyscarney, line 554: Should we add a note on why we are using `inheritAttributes` vs `inheritAriaAttributes`? Something like:

```
> [!NOTE]
> Use `inheritAttributes` to specify which attributes should be inherited or `inheritAriaAttributes` to inherit all of the possible `aria` attributes.
```

core/src/components/select/select.tsx
brandyscarney, line 1: I tested this out on `main` and this branch using VoiceOver with the following:

```html
<ion-range value="50" aria-label="Banana">
  <span slot="label">Temperature</span>
</ion-range>

<ion-select value="50" aria-label="Strawberry">
  <span slot="label">Country</span>
  <ion-select-option>Option</ion-select-option>
</ion-select>
```

And on `main` it read the slotted labels for both (`Temperature` and `Country`) but on the branch it reads `Banana` and `Country`. This behavior is inconsistent, is this desired?

.github/COMPONENT-GUIDE.md
liamdebeasi: https://github.com/ionic-team/ionic-framework/pull/29145/commits/734ee1d874031419bf1887cadff038ead37a55c2

core/src/components/select/select.tsx
liamdebeasi: Good catch! This was a bug in `main` that was partially fixed in this branch.

In short, when you have both visible text and an aria label, the aria label should be announced instead of the visible text if both are provided. This was inconsistent in the select, so I fixed this in https://github.com/ionic-team/ionic-framework/pull/29145/commits/2ba4515a94d08a83852c4155c14930606c03b51b.

Range and select are the only two that had this problem because we manually manage the aria label values and conditionally render attributes (the other form controls do not)

brandyscarney: Awesome! I thought that the `aria-label` should take priority but I didn't want to assume. Thanks for adding tests too! We might want to consider this a bug fix so it shows up in the changelog.

brandyscarney (approved):
Works great! 👍

PR #29149, opened 2024-03-13
feat: rename dark/high-contrast themes to palettes

sean-perkins:
Issue number: Internal

---------

Ionic Framework currently plans to offer dark and high contrast "themes" in v8. However this naming nomenclature conflicts with a significant new feature that the team is working on towards v9+. 

## What is the new behavior?

- Migrates previous dark and high contrast "themes" to "palettes"
- Updates test infrastructure to import from the new stylesheet locations

- [x] Yes
- [ ] No

Developers that have updated to the Ionic v8 beta and have implemented the dark and high contrast themes, will need to update the import path:

```diff
-@import '@ionic/angular/css/themes/dark.always.css';
+@import '@ionic/angular/css/palettes/dark.always.css';
```

## Other information

Blocked by: https://github.com/ionic-team/ionic-framework/pull/29148. Review that first. 

Documentation PR: https://github.com/ionic-team/ionic-docs/pull/3521


core/src/themes/test/colors/index.html
```diff
@@ -320,26 +320,26 @@ <h3 class="header">Background 16%</h3>
...
-        if (theme === 'default') {
+      function togglePalette(palette) {
+        let cssFile = `/css/palettes/${palette}.always.css`;
+        if (palette === 'default') {
           // TODO FW-5862 update this to not
           // load a file when default is set. The light
-          // theme is automatically set when importing ionic.bundle.css
-          cssFile = `/src/themes/test/${theme}.css`;
+          // palette is automatically set when importing ionic.bundle.css
+          cssFile = `/src/palettes/test/${palette}.css`;
```
mapsandapps, line 336: It seems like this file is still at `/src/themes/test/default.css`

core/src/themes/test/colors/index.html
```diff
@@ -2,7 +2,7 @@
 <html lang="en" dir="ltr">
   <head>
     <meta charset="UTF-8" />
-    <title>Theme</title>
+    <title>Palette</title>
```
mapsandapps, line 5: It seems like it might make sense to rename this directory too? (`core/src/themes` -> palette)

mapsandapps (changes requested):

amandaejohnston (approved):
LGTM once Shawn's feedback is addressed 👍

core/src/themes/test/colors/index.html
sean-perkins: This would be a much more significant change and I don't believe is the right change. Currently there are global mixins and theme-specific styles in the `core/src/themes` directory (for iOS and MD themes).

core/src/themes/test/colors/index.html
sean-perkins: Good catch, I addressed that here: https://github.com/ionic-team/ionic-framework/pull/29149/commits/ff52cc22f9fc3efcfbf46fdc3b22891a9f8428d3 & simplified just saying what we are explicitly importing, since we don't really need to use the `palette` to generate the path.

mapsandapps (approved):

PR #29158, opened 2024-03-13
feat(react): add stronger typing for useIonModal and useIonPopover 

### LINKED ISSUE #28680: feat: Improved types for react useIonModal

liamdebeasi:
Issue number: resolves #28680

---------

This is a continuation of https://github.com/ionic-team/ionic-framework/pull/29058. When the `feature-7.8` was deleted (after the Ionic 7.8 release) GitHub automatically closed the PR. The following is copied from that PR.

useIonModal and useIonPopover do not type check component parameters. This can cause bugs in applications that heavily use these hooks.

## What is the new behavior?

- useIonModal and useIonPopover have new types
- If a component does not have props, second parameter can be omitted (`extends undefined`)
- Passing a JSX Element keeps old functionality (`any` for props)

- [ ] Yes
- [X] No

This may be considered a breaking change, because typescript will now complain if component props do not match up. The majority of my `useIonModal` usages required tweaks.

### DIFF
--- a/packages/react/src/hooks/useIonModal.ts
+++ b/packages/react/src/hooks/useIonModal.ts
@@ -8,14 +8,18 @@ import type { ReactComponentOrElement } from '../models/ReactComponentOrElement'
 import type { HookOverlayOptions } from './HookOverlayOptions';
 import { useOverlay } from './useOverlay';
 
-// TODO(FW-2959): types
-
 /**
  * A hook for presenting/dismissing an IonModal component
  * @param component The component that the modal will show. Can be a React Component, a functional component, or a JSX Element
  * @param componentProps The props that will be passed to the component, if required
  * @returns Returns the present and dismiss methods in an array
  */
+export function useIonModal(component: JSX.Element, componentProps?: any): UseIonModalResult;
+export function useIonModal<P extends undefined>(component: React.ComponentClass<P> | React.FC<P>): UseIonModalResult;
+export function useIonModal<P extends Record<string, never>>(
+  component: React.ComponentClass<P> | React.FC<P>
+): UseIonModalResult;
+export function useIonModal<P>(component: React.ComponentClass<P> | React.FC<P>, componentProps: P): UseIonModalResult;
 export function useIonModal(component: ReactComponentOrElement, componentProps?: any): UseIonModalResult {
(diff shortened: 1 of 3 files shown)


aeharding:
Just wondering, is it still possible to get this in for v8? :)

Here's what needs to be addressed before merge: 

- The sample app typescript issue (build failure) can be fixed by simply upgrading typescript in the sample app package.json.
     - Ionic consumers using older versions of typescript will have to manually set type `useIonModal<Type>` or simply `useIonModal<any>` for old behavior, since old typescript versions aren't smart enough to infer the prop type. But release notes should encourage simply upgrading Typescript.
- I am still looking for thoughts on this: https://github.com/ionic-team/ionic-framework/pull/29058#discussion_r1510158435

liamdebeasi:
Hey there! Sorry for the delay. The main concern we have with this change is that it requires a breaking change in usage in Ionic because developers must manually type `useIonModal` and `useIonPopover`. Is there a way to make this improvement in a backwards-compatible way? At the moment we are trying to avoid more breaking changes in Ionic 8.

aeharding:
> Is there a way to make this improvement in a backwards-compatible way?

Assuming you have passed the correct parameters (that match the existing types of your component), and your project is using a newer version of Typescript, this PR should require no code changes to projects.

To reiterate, code changes are only needed when:

 1. Passed types do not match component types. This may often happen because `useIonModal` is completely untyped, and fixing this is the main point of this PR (to ensure people are passing correct params that the component expects).
 2.  Using an old version of typescript. Newer versions of typescript will automatically infer the component type, and you do not need to explicitly pass `<Type>`. You only need to do this with old Typescript versions.

liamdebeasi:
> Using an old version of typescript. Newer versions of typescript will automatically infer the component type, and you do NOT need to explicitly define the type via useIonModal<Type>. You only need to do this with old Typescript versions.

Do you know which versions of TypeScript changes are needed for? We support TypeScript 3.7+ for Ionic 7 (https://ionicframework.com/docs/v8/reference/support#ionic-react), so depending on the version this may constitute a breaking change in the context of semantic versioning.

PR #29168, opened 2024-03-15
test(item): use themes options in config

liamdebeasi:
Issue number: Internal

---------

In https://github.com/ionic-team/ionic-framework/pull/27134 we fixed a bug for item button colors on dark mode. We added a test for this in dark mode, but it was created before we had the infrastructure to test dark mode within Playwright. As a result, the dark mode theme is hard coded into the test.

## What is the new behavior?

- This test has been simplified to only test the things we need to test
- It also relies on the built-in dark theme testing instead of hard coding the theme.

- [ ] Yes
- [x] No


core/src/components/item/test/buttons/item.e2e.ts
```diff
@@ -19,14 +19,34 @@ configs().forEach(({ title, screenshot, config }) => {
   });
 });
 
-configs({ directions: ['ltr'] }).forEach(({ title, screenshot, config }) => {
+configs({ directions: ['ltr'], themes: ['dark'] }).forEach(({ title, screenshot, config }) => {
   test.describe(title('item: buttons dark'), () => {
     test('should not have visual regressions in dark', async ({ page }) => {
-      await page.goto(`/src/components/item/test/buttons?dark=true`, config);
+      test.info().annotations.push({
```
liamdebeasi, line 25: This was missing from the original fix, so I decided to add it

brandyscarney (approved):

PR #29161, opened 2024-03-13
feat(button): add circular shape as round

thetaPC:
Issue number: internal

---------

## What is the current behavior?

The `round` shape with only an icon will display a smaller value for the border radius.

Screenshot 2024-03-13 at 4 48 05 PM
Screenshot 2024-03-13 at 4 48 00 PM

## What is the new behavior?

The `round` shape with only an icon will display as a circle.

- The changes are made to `md` and `ios`.
- The height for buttons have been updated to match [iOS](https://www.figma.com/community/file/1248375255495415511) and [MD](https://www.figma.com/community/file/1035203688168086460) designs.
   - iOS: small - 28px, regular - 34px, large - 50px
   - MD: small - 28px, regular - 40px, large - 50px
- The icon size has been updated to match [iOS](https://www.figma.com/community/file/1248375255495415511) and [MD](https://www.figma.com/community/file/1035203688168086460) designs.
   - I had to calculate the sizes manually (overlapping native and `ion-button`) since Figma did not provide them.
- The round test page has been updated to display all the possible options with `shape="round"`.
- The tests have been updated to track visual regressions for all the possible options with `shape="round"`.

Screenshot 2024-03-19 at 1 13 41 PM
Screenshot 2024-03-19 at 1 13 35 PM

## Does this introduce a breaking change?

- [ ] Yes
- [x] No

N/A

## Other information

Screenshots by components:
- Button: https://github.com/ionic-team/ionic-framework/pull/29161/commits/919301aa71a2dfebcd2d8de928c2a7cb559a983d
- Select: https://github.com/ionic-team/ionic-framework/pull/29161/commits/b58a4e28f601fa5f142fbfd06de90d33873fdc3f
- Input: https://github.com/ionic-team/ionic-framework/pull/29161/commits/73aa34e763950b1490c60586b1b1b573006c3ad3
- Item: https://github.com/ionic-team/ionic-framework/pull/29161/commits/ae8637159b22c5e0029afcd6cd50e9fd1c89f67d
- Textarea: https://github.com/ionic-team/ionic-framework/pull/29161/commits/aa02181cda22c69a853abc54d3bbd35019970b91

PR #29175, opened 2024-03-18
feat(input): add input-password-toggle component

liamdebeasi:
Issue number: Internal

---------

When given a password input it is hard to know what users are typing as the contents of the input are obscured. As a result, it is a common pattern to have a button that lets users temporarily toggle the visibility of the password so they can correct any mistakes. Ionic currently has the infrastructure for developers to implement this on their own, but this use case is so common that the team thinks it is worth having this functionality built-in to Ionic.

## What is the new behavior?

- Introduces the `ion-input-password-toggle` component. This component is a button that toggles the visibility of the text in the input it is slotted into.

- [ ] Yes
- [x] No

## Other information

⚠️ Give co-author credit to https://github.com/ionic-team/ionic-framework/pull/29141 on merge.

Docs PR: https://github.com/ionic-team/ionic-docs/pull/3541

Note: We did not do the approach listed in the other PR due to https://github.com/ionic-team/ionic-framework/pull/29141#discussion_r1523631811.


core/src/components/input-password-toggle/input-password-toggle.tsx
```diff
@@ -0,0 +1,141 @@
...
+
+  /**
+   * The icon that can be used to represent showing a password. If not set, the "eye" Ionicon will be used.
+   */
+  @Prop() showPasswordIcon?: string;
+
+  /**
+   * The icon that can be used to represent hiding a password. If not set, the "eyeOff" Ionicon will be used.
+   */
+  @Prop() hidePasswordIcon?: string;
```
liamdebeasi, line 41: The term "password" here may be redundant. What do we think of doing `showIcon` or `hideIcon` (or something similar?).

core/src/components/input-password-toggle/input-password-toggle.tsx
```diff
@@ -0,0 +1,141 @@
...
+import { eyeOff, eye } from 'ionicons/icons';
+
+import { getIonMode } from '../../global/ionic-global';
+import type { Color } from '../../interface';
+
+/**
+ * @virtualProp {"ios" | "md"} mode - The mode determines which platform styles to use.
+ */
+@Component({
+  tag: 'ion-input-password-toggle',
```
liamdebeasi, line 14: The component name is quite lengthy. I wanted to use `ion-input` in it to make it clear that it should only be used with `ion-input`. This naming also aligns with the supporting components for `ion-card`. What do we think of that name? I could also be on board with `ion-password-toggle`.

core/src/components/input/input.tsx
```diff
@@ -301,6 +305,12 @@ export class Input implements ComponentInterface {
    */
   @Event() ionStyle!: EventEmitter<StyleEventDetail>;
 
+  /**
+   * Emitted when the type property changes. Used for ion-input-password-toggle to show the correct icon.
+   * @internal
+   */
+  @Event() ionTypeChange!: EventEmitter<void>;
```
liamdebeasi, line 312: This was added so applications that change the input `type` prop outside of the password toggle component can still have the password toggle render correctly.

Sean and I discussed a couple ideas for how to solve this (the event being one of them). The other idea we discussed was having a property on `ion-input-password-toggle`. The idea here is that if developers want to change the input type they can pass the new value to the `type` property on both `ion-input` and `ion-input-password-toggle`.

I opted not to go for that because I was concerned about use cases where developers set the value of the input asynchronously and then never change it. Consider the following:

```html
<ion-input [type]="myType">...</ion-input>
```

The `type` prop defaults to `'text'`. If `myType` is `'password'` then that variable will be bound to the property sometime after the component is initially added to the DOM. In this case, developers would always need to pass `type` to `ion-input-password-toggle` component even if they never change the type themselves. While this avoids the internal event pattern, I was concerned that this would be seen as unnecessary or annoying by developers. In this case, the password toggle is slotted into the input, so we are able to reliably associate a password toggle with an input to listen for `type` property changes.

core/src/components/input-password-toggle/input-password-toggle.tsx
sean-perkins: Agree it is redundant, I'd vote for `showIcon` and `hideIcon`.

core/src/components/input-password-toggle/input-password-toggle.tsx
sean-perkins: Name is fine with me, especially since this _only_ works with `ion-input`. If it worked with any `input`, I'd vote for `ion-password-toggle`.

core/src/components/input/input.tsx
sean-perkins: Personally, I'm not a huge fan of the event emission pattern. When first getting up to speed with Ionic Framework, it was hard to reason where things were changing and why, as some very specific side effects happen only through an event being emitted. 

In this case, we are also grabbing a reference to a parent element, to attach an event listener, which feels counter to how event bubbling is designed (typically you capture events on the reference or at a parent, but here we are emitting an event for a child node). 

It's a minor change and an internal API, so I'm ok with moving forward with it here, if I'm the only one that doesn't prefer it. 

My other aside would be I'm against `forceUpdate`. It is a code smell of trying to do something outside of the typical render cycle, but requiring a re-render. I'd much prefer a local `@State` property for `type` here. 

--

My alternatives would be:
1. Bind `type` on both `ion-input` and `ion-input-password-toggle` as an implementation pattern required by developers. 
2. Have `ion-input` be responsible for querying `ion-input-password-toggle` and either calling an API or setting the `type`.

core/src/components/input-password-toggle/input-password-toggle.tsx
brandyscarney: Agree that the name should include `ion-input` since it must be used with `ion-input`.

liamdebeasi: 3 in favor of keeping the name as-is, so going to mark as resolved

core/src/components/input-password-toggle/input-password-toggle.tsx
liamdebeasi: Renamed in https://github.com/ionic-team/ionic-framework/pull/29175/commits/90e62e394fd900147affb3627c8d102fcb8310d4

core/src/components/input/input.tsx
amandaejohnston: The current implementation seems strange to me because we're bouncing between the two components a lot:

1. Toggle adds event listener to input.
2. Click toggle -> input type changed from toggle.
3. Event fired from input.
4. Toggle event listener triggered. `forceUpdate` called to detect new input type.

One way we could cut out steps is by using a more React-like approach, where the input queries for a toggle and adds a click listener to it. The listener then updates the input's type and changes the toggle's appearance accordingly. Not sure if that's the same as what Sean is proposing with alternative 2, haha. I definitely agree that having to set `type` on both the input and toggle would be annoying.

core/src/components/input-password-toggle/input-password-toggle.ios.scss
```diff
@@ -0,0 +1,18 @@
+/**
```
amandaejohnston, line 1: I noticed that this and the corresponding `md` styles are copied directly from `input.md/ios.scss`. Would it make sense to update the selectors there instead to cover the new component, or maybe pull them into a shared mixin if we're trying to avoid component coupling? My worry is that we'll have to know to update the styles in both places if we want to change things while keeping the button appearance consistent.

core/src/components/input-password-toggle/input-password-toggle.tsx
```diff
@@ -0,0 +1,140 @@
...
+   * outside of this component we also need to re-render so the correct icon is shown.
+   */
+  private onTypeChange = () => {
+    const { el, inputElRef } = this;
+
+    if (!inputElRef) return;
+
+    if (inputElRef.type !== 'text' && inputElRef.type !== 'password') {
+      printIonWarning(
+        `ion-input-password-toggle only supports inputs that accept plain text. Input of type "${inputElRef.type}" is not compatible.`,
```
amandaejohnston, line 87:
```suggestion
        `ion-input-password-toggle only supports inputs of type "text" or "password". Input of type "${inputElRef.type}" is not compatible.`,
```
"Plain text" is a little too vague to me; I could see it covering `number` type inputs, for example.

core/src/components/input-password-toggle/input-password-toggle.tsx
```diff
@@ -0,0 +1,140 @@
...
+
+    const isPasswordVisible = inputElRef?.type === 'text';
+
+    return (
+      <Host
+        class={createColorClasses(color, {
+          [mode]: true,
+        })}
+      >
+        <ion-button
```
amandaejohnston, line 113: I noticed that the button doesn't show a ripple effect in MD; was that intentional?

core/src/components/input/input.tsx
thetaPC: I would prefer to go with Sean's alternative #2.

It feels weird to have an event that would bubble down instead of the typical up. It would most likely confuse devs. I'm also not a fan of having to pass `type` to `ion-input-password-toggle`. As a user, I would expect to have ion-input to be the only element that needs type. So the query option makes sense to me.

core/src/components/input-password-toggle/input-password-toggle.tsx
liamdebeasi: Not intentional, but not 100% sure why that would be happening. I'll look into it.

core/src/components/input/input.tsx
liamdebeasi: Thanks for the feedback, everyone. In https://github.com/ionic-team/ionic-framework/pull/29175/commits/4bcc59f2e7e92ef78bdd7a2edbdf03ab534afeb8 I updated the implementation to add an internal `type` prop on the password toggle. The parent input is responsible for setting that to the correct value, and then the password toggle looks at that prop to determine which icon to render.

core/src/components/input-password-toggle/input-password-toggle.tsx
```diff
@@ -0,0 +1,151 @@
...
+
+/**
+ * @virtualProp {"ios" | "md"} mode - The mode determines which platform styles to use.
+ */
+@Component({
+  tag: 'ion-input-password-toggle',
+  styleUrls: {
+    ios: 'input-password-toggle.ios.scss',
+    md: 'input-password-toggle.md.scss',
+  },
```
liamdebeasi, line 18: I need to look closer at this tomorrow, but without these stylesheets inheriting the mode form `ion-input-password-toggle` to the inner `ion-button does not work: The mode is always "MD". This shouldn't block a review, but I did add this back for now so we can get the build passing.

core/src/components/input-password-toggle/input-password-toggle.tsx
liamdebeasi: Ok we're going to test this out in https://github.com/ionic-team/ionic-framework/pull/29192

liamdebeasi: Should be fixed in this branch now!

amandaejohnston (approved):
We discussed the possibility of using the new round button styles in https://github.com/ionic-team/ionic-framework/pull/29161 instead of needing a shadow part to style the new button, but I don't see that as a blocker 👍

core/src/components/input-password-toggle/input-password-toggle.tsx
```diff
@@ -0,0 +1,151 @@
...
+  /**
+   * Whenever the input type changes we need to re-run validation to ensure the password
+   * toggle is being used with the correct input type. If the application changes the type
+   * outside of this component we also need to re-render so the correct icon is shown.
+   */
+  @Watch('type')
+  onTypeChange(newValue: TextFieldTypes) {
+    if (newValue !== 'text' && newValue !== 'password') {
+      printIonWarning(
+        `ion-input-password-toggle only supports inputs of type "text" or "password". Input of type "${newValue}" is not compatible.`,
```
brandyscarney, line 57: Should this component render and work if the type is not supported? It still behaves like a password field after a few toggles:

https://github.com/ionic-team/ionic-framework/assets/6577830/782802d7-a2ef-45f3-9e34-fc5bf361a295

liamdebeasi: As we discussed with the team internally we are going to keep the console log and add documentation to make correct usage more clear. We will evaluate feedback from developers to see if we need to make any other changes. I've made the docs changes in https://github.com/ionic-team/ionic-docs/pull/3541/commits/d3cca0c5225b08b6de3c9471c5f13368361eeac8.

core/src/components/alert/test/a11y/alert.e2e.ts-snapshots/alert-scale-ios-ltr-Mobile-Chrome-linux.png
liamdebeasi, line 1: This is correct. Sean made this screenshot [update](https://github.com/ionic-team/ionic-framework/commit/e3755086471b8a46c3ee8f8c0ee0087f3d1748d4) in `next` but it never made its way into `feature-8.0`

liamdebeasi, line 1: This is correct. The team made this screenshot [update](https://github.com/ionic-team/ionic-framework/commit/e3755086471b8a46c3ee8f8c0ee0087f3d1748d4) in `next` but it never made its way into `feature-8.0`

brandyscarney (approved):

PR #29183, opened 2024-03-19
refactor(input, textarea) remove unused event

liamdebeasi:
This events are not used anymore as of Ionic 8. The event definition is there but they are never emitted.


thetaPC (approved):
LGTM

PR #29185, opened 2024-03-19
fix(angular): schematics account for new theme files

liamdebeasi:
Issue number: Internal

---------

While doing other work I noticed that the Angular schematics still reference the old dark theme.

## What is the new behavior?

- The Angular schematics now import the new dark theme found in Ionic v8.

- [ ] Yes
- [x] No


thetaPC (approved):
LGTM

PR #29186, opened 2024-03-19
refactor(haptics): remove cordova haptics support

liamdebeasi:
Issue number: Internal

---------

Ionic Framework relies on the haptics plugin for certain interactions such as spinning the Picker or dragging the knob on a Toggle. Up until now we have had support for both Cordova and Capacitor haptics plugins. As a result of a previous incident, the team wanted to investigate dropping support for the Cordova haptics plugin in favor of pushing developers to migrate to Capacitor.

## What is the new behavior?

- Remove support for the Cordova haptics plugin. Components that relied on this will continue to function but will no longer play haptics in Cordova environments.

- [x] Yes
- [ ] No


sean-perkins (approved):
Changes look good, should this be a `feat` vs. a `refactor`?

liamdebeasi:
According to https://github.com/ionic-team/ionic-framework/blob/main/.github/CONTRIBUTING.md#type, we use `refactor` for changes that neither fixes a bug nor adds a new feature

sean-perkins:
I guess even conventional commits doesn't have a specification around removing features. I'm ok with refactor. My main concern was visibility in the CHANGELOG for developers to know this support has been removed. It's included in the breaking changes, so I think that concern is covered.

PR #29192, opened 2024-03-20
refactor(tap-click): use pointer events api

liamdebeasi:
Issue number: Internal

---------

[Amanda pointed out that the ripple effect for the Button inside of InputPasswordToggle was not working](https://github.com/ionic-team/ionic-framework/pull/29175#discussion_r1532627841). I found out that calling `ev.preventDefault` on `pointerdown` causes `mouseup` to not get fired. On desktop, we rely on `mouseup` to know when to add the ripple effect. (`touchend` is not impacted)

Interestingly, calling `ev.preventDefault` on `pointerdown` does **not** prevent `pointerup` from being fired. The idea here is that if we migrate the tap click utility to use the PointerEvents API instead of separate mouse/touch listeners we can keep the existing tap click behavior while also fixing the bug that Amanda noted.

## What is the new behavior?

- Tap click nows listens for the Pointer Events instead of separate mouse/touch events

Impact to developers is fairly minimal. There should be no behavior change (other than the bug I noted being fixed). There should be a very small perf boost because this util now only adds 4 event listeners on the document instead of 7 previously.

- [ ] Yes
- [x] No

## Other information

Reviewers: Please manually test this on desktop devices as well as iOS and Android devices (not Chrome Dev Tools. iOS simulators are fine). Test that components such as `ion-button` correctly add the activated state (or ripple effect for MD). Also verify that the activated state is not added when tapping the button and then scrolling. For desktop, check that right clicking does not add the activated state.

### DIFF
--- a/core/src/utils/tap-click/index.ts
+++ b/core/src/utils/tap-click/index.ts
@@ -1,14 +1,13 @@
 import { doc } from '@utils/browser';
 
 import type { Config } from '../../interface';
-import { now, pointerCoord } from '../helpers';
+import { pointerCoord } from '../helpers';
 
 export const startTapClick = (config: Config) => {
   if (doc === undefined) {
     return;
   }
 
-  let lastTouch = -MOUSE_WAIT * 10;
   let lastActivated = 0;
 
   let activatableEle: HTMLElement | undefined;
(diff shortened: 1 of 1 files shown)


amandaejohnston (approved):
Verified on desktop plus my actual Android and iOS devices, looks good!

PR #29197, opened 2024-03-21
fix(searchbar): autocapitalize is initialized correctly

### LINKED ISSUE #29193: bug: autocapitalize is required in IonSearchbar types
### Prerequisites

### Ionic Framework Version

v7.x

### Current Behavior

Tried to upgrade to 7.8.1/8.0.0-beta.3, notice the following:

image

### Expected Behavior

`autocapitalize` is optional

### Steps to Reproduce

Click repro see type error

### Code Reproduction URL

https://stackblitz.com/edit/naskdu?file=src%2FApp.tsx,src%2Fmain.tsx

### Ionic Info

N/A

### Additional Information

_No response_

liamdebeasi:
Issue number: resolves #29193

---------

In an attempt to preserve backwards compatibility when adding `autocapitalize` to Searchbar, we used `!` to indicate that the prop was never undefined. The `autocapitalize` on `HTMLElement` expects this value to be a string and never undefined.

For the purposes of the property on Searchbar, setting this prop to one of the accepted values would constitute a breaking change because it would override the default browser behavior (which we previously relied upon). As a result, we used `!` to not set a default prop but inform TypeScript that this prop is always defined. This unintentionally made it so developers needed to define the `autocapitalize` property every time which is not what we want. 

## What is the new behavior?

- `autocapitalize` now defaults to the `'default'` keyword. This is an internal keyword that is used to tell Ionic to **not** set the `autocapitalize` attribute on the inner `input` element and instead rely on the default browser behavior. This satisfies the `HTMLElement` requirement that `autocapitalize` is never undefined. In Ionic 8 this `'default'` value will be replaced with `'off'`.

[Typescript currently sets the `HTMLElement` `autocapitalize` type to `string`](https://github.com/microsoft/TypeScript/blob/65812bf3ec3b9208141ef46e43d146a2eef88ae5/src/lib/dom.generated.d.ts#L10087) which is why we can add a `'default'` keyword here. There is some risk that if these type definitions change in the future that applications may encounter errors. However, changing this on the TypeScript side would itself be a breaking change. Additionally, we are moving away from `'default'` in Ionic 8, so I think this is an acceptable amount of risk.

- [ ] Yes
- [x] No

## Other information

Dev build: `7.8.2-dev.11711027016.13b2a920`

Tested in a React starter app with Searchbar, and I verified this fix works.

### DIFF
--- a/core/api.txt
+++ b/core/api.txt
@@ -1158,7 +1158,7 @@ ion-row,shadow
 
 ion-searchbar,scoped
 ion-searchbar,prop,animated,boolean,false,false,false
-ion-searchbar,prop,autocapitalize,string,undefined,true,false
+ion-searchbar,prop,autocapitalize,string,'default',false,false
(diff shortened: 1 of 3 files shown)


mapsandapps (approved):
Makes sense. I tested it & it works 👍🏻

sean-perkins (approved):

PR #29199, opened 2024-03-21
refactor(theming): remove Sass default flags

brandyscarney:
Issue number: N/A
---------

Currently, every Sass variable in Ionic has the `!default` flag added to the end.

From the [Sass variables documentation](https://sass-lang.com/documentation/variables/):

> Normally when you assign a value to a variable, if that variable already had a value, its old value is overwritten. But if you’re writing a Sass library, you might want to allow your users to configure your library’s variables before you use them to generate CSS.
> 
> To make this possible, Sass provides the `!default` flag. This assigns a value to a variable only if that variable isn’t defined or its value is [null](https://sass-lang.com/documentation/values/null). Otherwise, the existing value will be used.

In past versions of Ionic Framework, developers wrote Sass variables to rebuild Ionic Framework using their own values. In the latest versions of Ionic Framework, this is not possible. 

## What is the new behavior?
Removes the `!default` flag from all Sass variables.

- [ ] Yes
- [x] No


brandyscarney:
Closing due to conflicts. We are going to update main first here: https://github.com/ionic-team/ionic-framework/pull/29401 and then pull the changes into the `next` branch.

PR #29200, opened 2024-03-21
fix(button): use clamp for font sizes on circle shape

thetaPC:
Issue number: internal

---------

The font sizes for the icons are not clamping. This can lead to the icons getting too big if the `font-size` to the component is a large value.

## What is the new behavior?

- The font sizes for icons are now clamped for `ios` and `md`.
- Updated the small circle button's font size after reviewing iOS native.

- [ ] Yes
- [x] No

## Other information

Here are some comparisons between native and the implementation. The red icon is the implementation and the white icon is the native app:

Default:
https://github.com/ionic-team/ionic-framework/assets/13530427/b4aaedea-40ec-4378-ae26-8d7962b7ff6f

Max font size:
https://github.com/ionic-team/ionic-framework/assets/13530427/286f76f3-fd31-44b7-bba1-38eba9a6aeb3

Min font sizes:
https://github.com/ionic-team/ionic-framework/assets/13530427/8dc4b4bd-035d-44b7-863e-ba7b7947bde1


core/src/components/button/button.ios.scss
```diff
@@ -149,7 +149,7 @@
 }
 
 ::slotted(ion-icon[slot="icon-only"]) {
-  font-size: 1.125em;
+  font-size: dynamic-font-clamp(0.844, $button-ios-baseline-font-size, 2.53, 1.125em);
```
liamdebeasi, line 152: As discussed internally, we do not need to add this 4 argument.

liamdebeasi, line 152: As discussed internally, we should keep the default font sizes the same and just change the max/min values. For example, this line would become

```css
font-size: dynamic-font-clamp(0.844, 1.125em, 2.53);
```

liamdebeasi (changes requested):
I know we discussed this internally today, but I'm adding my comments here so we can keep track of changes.

core/src/components/button/button.ios.scss
thetaPC: In order to get to `em`, I would have to pass a fourth value: `dynamic-font-clamp(0.844, 1.125em, 2.53, 1em);`. Otherwise, it coverts it to `rem` as the [default](https://github.com/ionic-team/ionic-framework/blob/a313ee85edf637017a3d3d4bd5ce415bc9c0e72a/core/src/themes/ionic.functions.font.scss#L4-L5).

liamdebeasi: Yes good point! I forgot we want to keep `em`.

core/src/components/alert/test/a11y/alert.e2e.ts-snapshots/alert-scale-ios-ltr-Mobile-Chrome-linux.png
liamdebeasi, line 1: This might be a flaky screenshot

core/src/components/button/test/round/button.e2e.ts-snapshots/button-clear-round-ios-ltr-Mobile-Firefox-linux.png
liamdebeasi, line 1: Can you add a comment in the PR describing why these screenshot changes are desired?

thetaPC, line 1: The small icon did not have the correct size. The screenshot change is due to updating the value to be correct.

core/src/components/button/test/round/button.e2e.ts-snapshots/button-clear-round-md-ltr-Mobile-Chrome-linux.png
thetaPC, line 1: The small icon did not have the correct size. The screenshot change is due to updating the value to be correct.

liamdebeasi (approved):

core/src/components/button/button.ios.scss
```diff
@@ -149,19 +149,26 @@
...
-  
 }
 
 :host(.button-small) ::slotted(ion-icon[slot="icon-only"]) {
-  font-size: 1.4em;
+  /**
+    * The values were provided through a native iOS app.
+    * min font size: 13px, default font size: 17px, max font size: 38px
+    */
+  font-size: dynamic-font-clamp(0.63, 20.93px, 1.84, 1em);
```
liamdebeasi, line 171: Why is the default font size 20.93 when it says the default font size is 17 in the comments?

core/src/components/button/button.ios.scss
```diff
@@ -172,7 +179,11 @@
 }
 
 :host(.button-large) ::slotted(ion-icon[slot="icon-only"]) {
-  font-size: 1em;
+  /**
+    * The values were provided through a native iOS app.
+    * min font size: 13.5px, default font size: 20px, max font size: 40.5px
+    */
+  font-size: dynamic-font-clamp(0.85, 16px, 2.53, 1em);
```
liamdebeasi, line 186: Why is the default size 16 when it says 20 in the comments?

core/src/components/button/button.ios.scss
thetaPC: This is due to `ion-button`, the parent, having a different font size than the typical 16px.

core/src/components/button/button.ios.scss
thetaPC: This is due to `ion-button`, the parent, having a different font size than the typical 16px.

liamdebeasi (approved):
Thank you!

liamdebeasi:
Also, this is `fix` not ` refactor` because there are visual changes.

PR #29201, opened 2024-03-21
test(button): ionic theme tests

sean-perkins:
Issue number: N/A

---------

Adding new e2e tests into Ionic Framework can be difficult for new contributors when it comes to knowing where existing test cases exist and how to additionally add a new theme or option to the test. 

To avoid a lot of back and forth feedback in Github comments, I split up the test on the original PR to help demonstrate how to match existing test structures. 

## What is the new behavior?

- Removes the proposed test specific to the Ionic theme.
- Migrates the tests into the existing e2e files for button, adding specific test cases that only apply to the ionic theme.
- Added screenshot references

In this work, I tried to correct some legacy folder naming:
- `button/test/round` → `button/test/shape`
- For `fill="clear"` I updated the screenshot names, but left the folder structure. There is more work here to organize `fill="outline"` and it is not required at this time. 

- [ ] Yes
- [x] No

## Other information

I left the HTML template for the ionic theme test alone. I figure this is helpful to the developer, until the primary PR is approved. Then it could be removed if it is no longer useful.


core/src/components/button/test/shape/button.e2e.ts
brandyscarney, line 1: This folder rename is what I was going to recommend!

core/src/components/button/test/clear/button.e2e.ts-snapshots/button-fill-clear-ionic-ios-ltr-light-Mobile-Chrome-linux.png
brandyscarney, line 1: Are these being generated correctly? It seems to generate an `ionic-ios` and `ionic-md` file of the same test.

sean-perkins: The second portion is the mode, but I do believe we can restrict the modes here as well. Good catch 👍

core/src/components/button/test/shape/button.e2e.ts
```diff
@@ -0,0 +1,110 @@
+import { expect } from '@playwright/test';
+import { configs, test } from '@utils/test/playwright';
+
+/**
+ * All content takes up the full width, so RTL has no effect.
+ */
+configs({ directions: ['ltr'] }).forEach(({ title, screenshot, config }) => {
```
joselrio, line 7:
```suggestion
configs({ directions: ['ltr'], themes: ['ionic', 'md', 'ios'] }).forEach(({ title, screenshot, config }) => {
```
Think we also need ionic theme tests here, right?

PR #29206, opened 2024-03-23
chore(playwright): combine test configs for themes and modes

sean-perkins:
Issue number: N/A

---------

Developers are not able to easily add the "ionic" theme to existing test cases, without running the test against both iOS and MD mode:

```ts
configs({ themes: ['ios', 'md', 'ionic'] })
// Generates 4 test cases
// - iOS theme on iOS mode
// - MD theme on MD mode
// - Ionic theme on iOS mode
// - Ionic theme on MD mode
```

With the separation of `mode` into look and feel, the majority of test cases do not require testing the mode behavior and instead only need to test the visual theme that is applied to the component.

## What is the new behavior?

- Removes the `themes` option from the `configs` test generator object.

```ts
configs({ modes: ['ios', 'md', 'ionic-md'] })
```

- Combines `theme` and `mode` into the existing `modes` test generator object
  - The new options are `ionic-ios` and `ionic-md`, to run the Ionic theme against the respective mode.
  - This path was preferred to avoid deprecating and migrating all existing tests.

- [ ] Yes
- [x] No

## Other information

I do not have a strong preference on the semantics of `ionic-ios` vs. `ios-ionic` (theme first vs. mode first). If anyone has an opinion or alternative suggestion, please add a comment.


core/src/utils/test/playwright/generator.ts
```diff
@@ -113,21 +133,24 @@ export const configs = (testConfig: TestConfigOption = DEFAULT_TEST_CONFIG_OPTIO
...
-  const processedTheme = themes ?? DEFAULT_THEMES;
   const processedDirection = directions ?? DEFAULT_DIRECTIONS;
   const processedPalette = testConfig.palettes ?? DEFAULT_PALETTES;
 
   processedModes.forEach((mode) => {
-    processedTheme.forEach((theme) => {
-      if (!isModeValidForTheme(mode, theme)) {
-        return;
-      }
+    if (mode === 'ios' || mode === 'md') {
```
mapsandapps, line 140: I know we have a comment elsewhere, but it might be nice to note here that we're ignoring the md-ios and ios-md combos for testing.

core/src/utils/test/playwright/generator.ts
```diff
@@ -48,9 +68,9 @@ interface TestConfigOption {
  * each test title is unique.
  */
 const generateTitle = (title: string, config: TestConfig): string => {
-  const { direction, palette, theme, mode } = config;
+  const { direction, palette, mode, theme } = config;
 
-  if (theme === mode) {
+  if (theme === 'ios' || theme === 'md') {
```
liamdebeasi, line 73: Can we add a comment here too?

core/src/utils/test/playwright/generator.ts
```diff
@@ -77,9 +97,9 @@ const generateTitle = (title: string, config: TestConfig): string => {
  * and a test config.
  */
 const generateScreenshotName = (fileName: string, config: TestConfig): string => {
-  const { theme, direction, palette, mode } = config;
+  const { direction, palette, mode, theme } = config;
 
-  if (theme === mode) {
+  if (theme === 'ios' || theme === 'md') {
```
liamdebeasi, line 102: Can we add a comment here?

core/src/utils/test/playwright/generator.ts
liamdebeasi, line 140: I think we can simplify this (GitHub won't let me make a code suggestion because it would go across deleted lines)


```js
processedModes.forEach((mode) => {
    processedDirection.forEach((direction) => {
      processedPalette.forEach((palette) => {
        const [theme, modeName] = mode.split('-');
        configs.push({ direction, palette, mode: (modeName ?? theme) as Mode, theme: theme as Theme });
      });
    });
  });
```

In both cases we need to iterate over directions and palettes, so having separate code paths here doesn't really save us much work. For the `'ios'` and `'md'` cases, `modeName` will be `undefined` which is why we fall back to `theme` setting the `mode` key in the config object.

sean-perkins: I took a slightly different approach, curious what you think: https://github.com/ionic-team/ionic-framework/pull/29206/commits/eb9f5d6b8bfc60ced66dac394b02b4edbdf462bc.

I don't think we should suggest a theme can be a mode: `(modeName ?? theme) as Mode`. A mode can be a theme (for backwards compatibility), but a theme should not be a mode.

liamdebeasi: The main problem here is we don't know if `mode` contains either a theme and a mode or a combined mode string ahead of time, so the naming in my original approach is confusing.

What you have works, but it's doing an additional `indexOf` check which will iterate through the `mode` input more than we need to.

Perhaps we can have something like this:

```js
const [themeOrCombinedMode, modeName]  = mode.split('-');

const parsedTheme = themeOrCombinedMode as Theme;

// If modeName is undefined then we have a combined mode string, so the first variable is actually the mode
const parsedMode = modeName ?? themeOrCombinedMode as Mode;
```

This approach avoids the additional iteration through the string.

core/src/utils/test/playwright/generator.ts
```diff
@@ -31,14 +29,36 @@ interface TestUtilities {
...
-  palettes?: Palette[];
   /**
-   * The individual themes (iOS, Material Design and Ionic) to test
-   * against. If unspecified, defaults iOS and Material Design
+   * The color palette to test against.
+   *
+   * - "light": Test against light theme.
+   * - "dark": Test against dark theme.
+   * - "high-contrast": Test against high contrast light theme.
+   * - "high-contrast-dark": Test against high contrast dark theme.
```
liamdebeasi, line 57:
```suggestion
   * - "light": Test against light palette.
   * - "dark": Test against dark palette.
   * - "high-contrast": Test against high contrast light palette.
   * - "high-contrast-dark": Test against high contrast dark palette.
```

liamdebeasi (approved):

PR #29213, opened 2024-03-25
ROU-4822 - [feat](checkbox): Add new checkbox features under the scope of the new ionic theme.

JoaoFerreira-FrontEnd:
Issue number: **ROU-4822**

---------

## What is the new behavior?

- 4 states (Selected, Unselected, Undeterminated, Error) for ionic theme
- 5 styles (Enabled, Disabled, Hover, Pressed, Focus) for ionic theme
- Target area (minimum 48px) for ionic theme
- Implement the 2 sizes (24px and 18px) for ionic theme
- Required Checkbox
- Two orientations for the ionic theme

- [ ] Yes
- [x] No

## Preview
checkbox_IonicFramework


core/src/components/checkbox/checkbox.ionic.scss
```diff
@@ -0,0 +1,166 @@
...
+  position: absolute;
+
+  width: calc(100% + 8px);
+  height: calc(100% + 8px);
+
+  transition: border-color 0.3s;
+
+  border: 2px solid $checkbox-ionic-focus-ring-color;
+
+  /* stylelint-disable-next-line property-disallowed-list */
```
sean-perkins, line 119: Ionic Framework provides a `@include border-radius` mixin. This is why we throw the stylelint error. 

This usage should be updated to:
```
@include border-radius(var(--border-radius));
```

core/src/components/checkbox/checkbox.tsx
```diff
@@ -98,6 +98,22 @@ export class Checkbox implements ComponentInterface {
...
 
+  /**
+   * If `true`, the checkbox will be presented with an error style when it is unchecked.
+   */
+  @Prop() required = false;
+
+  /**
+   * Set to `"soft"` for a checkbox with more rounded corners.
+   */
+  @Prop({ reflect: true }) shape?: 'soft' | 'rectangular' = 'soft';
```
sean-perkins, line 109: I would recommend removing the `{ reflect: true }` for now. This applies to both usages. 

This is some history here, but Ionic Framework currently uses reflection of CSS classes on the element host, to use for the styling selectors. We should eventually move away from this, as reflecting the attributes is actually more compliant in SSR environments (mark-up provided by the developer matches the rendered output).

Since we are not using these attributes in our current style implementation, we are invoking more JS in Stencil's virtual dom to reflect the attributes to the element.

core/src/components/checkbox/checkbox.tsx
```diff
@@ -233,6 +255,7 @@ export class Checkbox implements ComponentInterface {
             <svg class="checkbox-icon" viewBox="0 0 24 24" part="container">
               {path}
             </svg>
+            {theme === 'ionic' && <div part="focus-ring" class="focus-ring"></div>}
```
sean-perkins, line 258: Be sure to document this shadow part at the top of the file. You can checkout what was done for button on the latest `next` branch. We should make it clear this shadow part is only available in the ionic theme.

core/src/components/checkbox/test/theme-ionic/checkbox.e2e.ts
```diff
@@ -0,0 +1,43 @@
+import { expect } from '@playwright/test';
```
sean-perkins, line 1: I'm working to correct an issue in our test generators: https://github.com/ionic-team/ionic-framework/pull/29206, but after that is merged we should be able to add `ionic-md` to existing test cases vs. having separate test cases specifically for the ionic theme.

The advantage is we aren't maintaining entire test suites specific to a theme, but can run the same test suite against many different themes.

JoaoFerreira-FrontEnd: Yes, that makes sense, I did just maintain a page with everything just for the stage, so the designers can see everything side by side.

PR #29216, opened 2024-03-25
fix(button): activated outline button in toolbar no longer blends into background on MD dark mode

amandaejohnston:
Issue number: N/A

---------

When using an outline-style button in a toolbar on `md` dark mode, the activated style causes the button to become invisible.

Steps to repro in `main`:
1. Update the `css-variables` themes test to use the latest dark theme styles (at least for v7) by replacing the contents of [this file](https://github.com/ionic-team/ionic-framework/blob/main/core/src/themes/test/css-variables/css/dark.css) with the [styles from the docs](https://ionicframework.com/docs/theming/dark-mode#ionic-dark-theme).
2. Host the project locally and navigate to the test at `themes/test/css-variables/index.html`.
3. Switch to dark mode within the page. Note that the activated outline button isn't visible:
image

## What is the new behavior?

- Button styles corrected. The colors were correct for `ios` mode, but had to be flipped for `md`, so I pulled the relevant styles into the mode-specific stylesheets.
- CSS variables test has also been updated to use the same dark mode CSS as in [the docs](https://ionicframework.com/docs/theming/dark-mode#ionic-dark-theme), as per the repro steps above. Let me know if you would rather this be split into a separate PR.

Changes to the toolbar test were split into a separate PR to keep this one clean and ensure the screenshot changes can be reviewed more effectively: https://github.com/ionic-team/ionic-framework/pull/29231

- [ ] Yes
- [x] No


thetaPC (changes requested):
There are a few tests that are not related to the issue. Let's keep this PR specific to it and open a new PR that introduces these test changes.

thetaPC:
@amandaejohnston Is this ready for review? It's marked as draft.

amandaejohnston:
@thetaPC No, the screenshots haven't been updated yet here. Running the job causes them to be updated off of `main` instead of the `toolbar-test-split` branch (and I can't use the Docker stuff due to it not working for Windows yet), which causes all the screenshots to be generated new instead of this PR only showing the changed ones, so I'm hoping to get https://github.com/ionic-team/ionic-framework/pull/29231 merged in so I can resolve things without having to debug all that.

amandaejohnston:
@thetaPC This is ready for review again 👍

core/src/components/button/button.md.scss
```diff
@@ -160,3 +160,12 @@
     background: #{current-color(base)};
   }
 }
+
+
+// Activated Button in Toolbar
+// --------------------------------------------------
+
+:host(.button-outline.ion-activated.in-toolbar:not(.ion-color):not(.in-toolbar-color)) .button-native {
```
thetaPC, line 168: Let's add a comment that this needs to be within this file instead of `button.scss` else it gets overridden. That way we don't end up moving it back.

core/src/components/button/button.ios.scss
```diff
@@ -212,3 +212,12 @@
 :host(.button-solid.ion-color.ion-activated) .button-native::after {
   background: #{current-color(shade)};
 }
+
+
+// Activated Button in Toolbar
+// --------------------------------------------------
+
+:host(.button-outline.ion-activated.in-toolbar:not(.ion-color):not(.in-toolbar-color)) .button-native {
```
thetaPC, line 220: Let's add a comment that this needs to be within this file instead of `button.scss` else it gets overridden. That way we don't end up moving it back.

thetaPC (approved):
LGTM, just a minor comment request

core/src/components/button/button.md.scss
amandaejohnston: Wouldn't that be the case for all the code in these mode-specific files? Is there something about these styles that seem like they'd be more likely to be moved back by accident?

Note that the MD/iOS styles are slightly different now, so putting them in the mode-specific files isn't a hack or anything.

core/src/components/button/button.ios.scss
```diff
@@ -212,3 +212,12 @@
...
+  background: var(--ion-toolbar-color, var(--color));
```
thetaPC, line 221: Why are we using color variables for background?

core/src/components/button/button.md.scss
thetaPC: Oh, I see. Then this probably ties to my question to the [iOS background style](https://github.com/ionic-team/ionic-framework/pull/29216#discussion_r1548321998).

core/src/components/button/button.ios.scss
amandaejohnston: The background and text colors reverse when an outline button is activated. Note that the iOS styles have not changed from `main`: https://github.com/ionic-team/ionic-framework/pull/29216/files#diff-5b2387a45e4270ba71b3a393ba3fe681d917cc518e6f9e059cccdcbe56dd2149L339

core/src/components/button/button.md.scss
thetaPC: I'm good without the comments, thank you!

thetaPC: You're good to merge.

PR #29218, opened 2024-03-25
chore(css): remove unit option from font scaling utils

liamdebeasi:
Maria and I were discussing this `$unit` argument and why it exists to debug some behaviors in other PRs. We discussed if it was possible to only pass the unit to convert instead of having to pass `1` and then the unit. While investigating this I realized that this argument is not used in the project anymore. Rather than refactor something that is not even being used, I'd like to remove the argument altogether to avoid confusion.
 
[Some components](https://github.com/ionic-team/ionic-framework/blob/ca0923812a89fab830c33b1e854d114c2946c988/core/src/components/item-divider/item-divider.md.scss#L54) do need to use `em` units with font scaling. In those cases they need to use a different base font size, so they can't use the `dynamic-font` functions anyways.

### DIFF
--- a/core/src/themes/ionic.functions.font.scss
+++ b/core/src/themes/ionic.functions.font.scss
@@ -1,18 +1,19 @@
 @use "sass:math";
 
 $baselineSize: 16px !default;
-$baselineUnit: 1rem !default;
 
 /**
  * Convert a font size to a dynamic font size.
  * Fonts that participate in Dynamic Type should use
  * dynamic font sizes.
  * @param size - The initial font size including the unit (i.e. px or pt)
- * @param unit (optional) - The unit to convert to. Use this if you want to
- * convert to a unit other than $baselineUnit.
  */
-@function dynamic-font($size, $unit: $baselineUnit) {
-  @return (math.div($size, $baselineSize)) * $unit;
+@function dynamic-font($size) {
+  /**
+   * Multiplying by 1rem appends the rem unit to the number while keeping the number unaltered.
+   * Example: 1.25 * 1rem = 1.25rem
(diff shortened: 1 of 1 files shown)


core/src/themes/ionic.functions.font.scss
```diff
@@ -1,30 +1,29 @@
...
  * Convert a font size to a dynamic font size.
  * Fonts that participate in Dynamic Type should use
  * dynamic font sizes.
  * @param size - The initial font size including the unit (i.e. px or pt)
- * @param unit (optional) - The unit to convert to. Use this if you want to
- * convert to a unit other than $baselineUnit.
  */
-@function dynamic-font($size, $unit: $baselineUnit) {
-  @return (math.div($size, $baselineSize)) * $unit;
+@function dynamic-font($size) {
```
brandyscarney, line 11: Couldn't we just add the `baselineSize` as an argument also to make the `baselineUnit` work with something other than `16px`? It seems like we should handle these exceptions so that we aren't saying _only_ use the `dynamic-font` functions if the base is `16px`, otherwise write the math division that is already handled by the `dynamic-font` functions. 

We could also separate them into two different functions based on the conversion unit if that would be easier to understand. If this is out of scope for now, we should add a comment to the function explaining it only works with a base of `16px` when converting to `rem` units.

liamdebeasi: 🤔 Yeah that could work. Let me play around with that a bit.

thetaPC:
With our recent conversation regarding the circular button, could we close this PR or change it since the fourth value is necessary?

PR #29222, opened 2024-03-26
fix(button): show correct background for activated button with color

brandyscarney:
Issue number: N/A

---------

Solid buttons in the `ionic` theme are using the primary background color when activated.

## What is the new behavior?
Updates the solid buttons to use the current color's shade.

| Before | After |
| ---| ---|
| next | branch |

- [ ] Yes
- [x] No

## Other information
Screenshot tests do not exist for these which is why it was not caught.


mapsandapps (approved):
Looks good. Should we add a screenshot test for this? It seems like it would be useful.

sean-perkins (approved):

brandyscarney:
@sean-perkins You were adding tests for this separately right?

core/src/components/button/test/states/button.e2e.ts
brandyscarney, line 1: I added the `md` and `ios` themes here because they also don't have coverage for this, but I can remove it if we don't want that on this PR.

sean-perkins (approved):

PR #29223, opened 2024-03-26
refactor(button): update focus ring to avoid additional API

liamdebeasi:
Issue number: Internal

---------

The team raised two concerns with the current Ionic theme focus ring implementation:

1. The border radii does not match the base button radii
2. Customizing this requires a new API to be exposed that only functions on the Ionic theme. Developers using the iOS/MD themes who try to use this will not be able to modify the focus state. This may also cause confusion because the focus ring only exists on the Ionic theme.

## What is the new behavior?

- Moved the focus ring outline to the base button element.

Developers can customize the focus ring either by using the CSS variables exposed or the `native` shadow part:

```css
ion-button.ion-focused::part(native) {
  outline: 4px solid red;
}
```

- [ ] Yes
- [x] No

## Other information

The screenshot diffs here are intentional:

1. The border radii for the focus ring now match the base button border radii
2. Because the focus ring is part of the base button (instead of a pseudo element that is absolutely positioned) the dimensions of the outline button have grown by 2px because it has to account for both the outline and the normal border on the outline button. I'd argue the original behavior was incorrect because buttons could have inconsistent sizes. The table below shows the difference using solid and outline buttons:

| `next` | branch |
| - | - |
| Screenshot 2024-03-26 at 10 49 05 AM | Screenshot 2024-03-26 at 10 50 05 AM |


core/src/components/button/button.ionic.scss
```diff
@@ -156,26 +156,9 @@
 // -------------------------------------------------------------------------------
 
 // Only show the focus ring when the button is focused
-:host(.ion-focused) {
-  --overflow: visible;
```
liamdebeasi, line 160: This isn't needed anymore because the outline is part of the normal document flow

sean-perkins (approved):
This is great, nice work 👍

BenOsodrac (approved):

thetaPC:
Is it safe to assume that moving forward that other components that are using the focus ring should follow this?

liamdebeasi:
> Is it safe to assume that moving forward that other components that are using the focus ring should follow this?

I'd expect components would use some form of this approach instead of having to add an additional element/part if possible.

PR #29224, opened 2024-03-26
fix(core): fallback detection for themes and modes

sean-perkins:
Issue number: N/A

---------

With the latest changes on `next`, the fallback behavior for detecting the themes from the Ionic config does not work when using `ion-router-outlet`. 

For example, in a React app with:

```ts
setupIonicReact({
  theme: 'ios'
});
```

and a template of:
```tsx
<IonApp>
    <IonReactRouter>
      <IonRouterOutlet>
         /* All children will apply the default mode for the mode and theme - not expected */
     </IonRouterOutlet>
  </IonReactRouter>
</IonApp>
```

All components rendered inside the `IonRouterOutlet` will apply the `mode` detection for the `theme`. This is not expected behavior, since the configuration has specified the `ionic` theme is the default theme. 

This problem occurs due to this specific logic:
```ts
const elmMode = (elm as any).mode || elm.getAttribute('mode');
```

The `ion-router-outlet` component has a property defined on the class for `mode`: https://github.com/ionic-team/ionic-framework/blob/ca0923812a89fab830c33b1e854d114c2946c988/core/src/components/router-outlet/router-outlet.tsx#L42

This means that `defaultMode` will always apply over the Ionic config's theme (`defaultTheme`).

## What is the new behavior?

- Specifying the `ionic` theme from the `initialize`/`setupIonicReact` functions will persist to children components.
- Developers can still use the `mode` to persist the changes to child nodes. However for in the case of `IonRouterOutlet`, it will require actually specifying the attribute in the mark-up:
```tsx
<IonRouterOutlet mode="md">
  /* Any children will apply the "md" mode and theme */
</IonRouterOutlet>
```

- [ ] Yes
- [x] No


core/src/global/ionic-global.ts
```diff
@@ -227,7 +227,7 @@ export const initialize = (userConfig: IonicConfig = {}) => {
      * behavior for applications that are not setting the "theme".
      */
     while (elm) {
-      const theme = (elm as any).theme || elm.getAttribute('theme');
```
liamdebeasi, line 230: Would this issue go away if we made the `mode` a virtual prop on router outlet instead?

liamdebeasi: works for me 👍

liamdebeasi (approved):

PR #29231, opened 2024-03-27
test(toolbar): add dark theme to basic test

amandaejohnston:
Issue number: N/A

---------

The toolbar `basic` test only checks light theme. Dark theme coverage is required to test against the bug fixed in https://github.com/ionic-team/ionic-framework/pull/29216.

## What is the new behavior?

Dark theme coverage added. This required converting the test to `page.setContent` instead of `page.goto`, so I went ahead and split the captured toolbars into multiple tests, including some cleanup of the test content. Since this included changes across many different tests, even ones not strictly related to the bug, this work has been split into a separate PR.

- [ ] Yes
- [x] No


core/src/components/toolbar/test/basic/toolbar.e2e.ts-snapshots/toolbar-basic-text-icon-buttons-md-ltr-dark-Mobile-Chrome-linux.png
amandaejohnston, line 1: Two of the buttons being invisible here is the bug being fixed in https://github.com/ionic-team/ionic-framework/pull/29216.

thetaPC: Then should this PR be based from FW-5721 branch?

core/src/components/toolbar/test/basic/toolbar.e2e.ts
```diff
@@ -1,14 +1,190 @@
 import { expect } from '@playwright/test';
 import { configs, test } from '@utils/test/playwright';
 
-configs().forEach(({ title, screenshot, config }) => {
+configs({ themes: ['light', 'dark'] }).forEach(({ title, screenshot, config }) => {
   test.describe(title('toolbar: basic'), () => {
-    test('should not have visual regressions', async ({ page }) => {
-      await page.goto(`/src/components/toolbar/test/basic`, config);
+    test('should not have visual regressions with text only', async ({ page }) => {
```
thetaPC, line 6: Since there are no changes between direction, then let's only test it for LTR.

core/src/components/toolbar/test/basic/toolbar.e2e.ts-snapshots/toolbar-basic-text-icon-buttons-md-ltr-dark-Mobile-Chrome-linux.png
amandaejohnston: No, because the new dark theme tests in this branch are necessary to check that the fix in the FW-5721 branch works. Doing it this way ensures that you can see the fix via the changed screenshots. Since the broken behavior already exists in `main`, I don't see the harm in temporarily adding screenshots that demonstrate it.

core/src/components/toolbar/test/basic/toolbar.e2e.ts
```diff
@@ -1,14 +1,190 @@
...
+        `
+          <ion-header>
+            <ion-toolbar>
+              <ion-buttons slot="secondary">
+                <ion-button>
+                  <ion-icon slot="icon-only" name="person-circle"></ion-icon>
+                </ion-button>
+                <ion-button>
+                  <ion-icon slot="icon-only" name="search"></ion-icon>
+                </ion-button>
```
thetaPC, line 49: Is it necessary to test a second button? Same for the other toolbar.

core/src/components/toolbar/test/basic/toolbar.e2e.ts
```diff
@@ -1,14 +1,190 @@
...
+              </ion-buttons>
+              <ion-buttons slot="primary">
+                <ion-button color="secondary">
+                  <ion-icon slot="icon-only" ios="ellipsis-horizontal" md="ellipsis-vertical"></ion-icon>
+                </ion-button>
+              </ion-buttons>
```
thetaPC, line 55: Would only one group suffice?

core/src/components/toolbar/test/basic/toolbar.e2e.ts
```diff
@@ -1,14 +1,190 @@
...
+          </ion-header>
+        `,
+        config
+      );
+
+      const header = page.locator('ion-header');
+      await expect(header).toHaveScreenshot(screenshot(`toolbar-basic-icon-buttons`));
+    });
+
+    test('should not have visual regressions with buttons with icons and text', async ({ page }) => {
```
thetaPC, line 83: Since the test is for button with icons and text. Do we need three buttons in a toolbar? It feels a bit redundant.

core/src/components/toolbar/test/basic/toolbar.e2e.ts
```diff
@@ -1,14 +1,190 @@
...
+      await page.setContent(
+        `
+          <ion-header>
+            <ion-toolbar>
+              <ion-buttons slot="secondary">
+                <ion-button>Go Back</ion-button>
+              </ion-buttons>
+              <ion-buttons slot="primary">
+                <ion-button href="#">Edit</ion-button>
+              </ion-buttons>
```
thetaPC, line 178: Could this be redundant since the first one is already testing what it looks like with a button?

core/src/components/toolbar/test/basic/toolbar.e2e.ts-snapshots/toolbar-basic-text-icon-buttons-md-ltr-dark-Mobile-Chrome-linux.png
thetaPC: Wouldn't you be able to see the fix by reviewing FW-5721 though?

amandaejohnston: Doing it this way ensures that the only screenshot updates on FW-5721 will be changes to the already-existing dark theme screenshots, showing the bug being fixed.

If we did it the other way around, FW-5721 would need to create the dark theme screenshots net-new, with the fix already present. You'd need to manually verify that the bug is actually present in the test when the fix is removed. (I did have it like that before you asked for the work to be split, but this way is better on second thought.)

I guess there's an argument for having FW-5721 _only_ convert the existing test to `page.setContent`, without breaking it up into multiple tests, and then have this PR handle breaking it up. But if we're already doing two PRs anyway, I think it makes more sense to handle all test-related updates here so that the fix is easier to verify.

thetaPC: I have no strong feelings towards this, I'm fine with this approach. Thank you for the explanation!

PR #29235, opened 2024-03-28
feat(checkbox): add ionic stylesheet and vars files

sean-perkins:
Issue number: N/A

---------

## What is the new behavior?

- Adds the new ionic theme stylesheet and sass variables file for checkbox
- Sets the ionic theme stylesheet as the default when: `<ion-checkbox theme="ionic">`

- [ ] Yes
- [x] No

## Other information

Co-authored-by credit will be applied to the merge commit.

### DIFF
--- a/core/src/components/checkbox/checkbox.ionic.scss
+++ b/core/src/components/checkbox/checkbox.ionic.scss
@@ -1,0 +1,4 @@
+@import "./checkbox.ionic.vars";
+
+// Ionic Checkbox
+// --------------------------------------------------
--- a/core/src/components/checkbox/checkbox.ionic.vars.scss
+++ b/core/src/components/checkbox/checkbox.ionic.vars.scss
@@ -1,0 +1,5 @@
+@import "../../themes/ionic.globals.ios";
+@import "../item/item.ios.vars";
+
+// Ionic Checkbox Variables
+// --------------------------------------------------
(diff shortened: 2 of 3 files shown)


core/src/components/checkbox/checkbox.ionic.vars.scss
```diff
@@ -0,0 +1,5 @@
+@import "../../themes/ionic.globals.ios";
+@import "../item/item.ios.vars";
```
amandaejohnston, line 2: Why are we importing the iOS styles here?

sean-perkins: Good question, sloppy copy & paste on my behalf with splitting up the original PR. 

I posted the question in an internal DM, but I don't believe we should be referencing any iOS/MD variables (global or component specific) in the Ionic theme. The ionic theme should be completely decoupled from the other themes. 

The value of this PR is we caught that, because the scope is small 😉

amandaejohnston (approved):

core/src/components/checkbox/checkbox.ionic.scss
```diff
@@ -0,0 +1,4 @@
+@import "./checkbox.ionic.vars";
```
brandyscarney, line 1:
```suggestion.
@import "./checkbox";
@import "./checkbox.ionic.vars";
```

brandyscarney (changes requested):

PR #29238, opened 2024-03-29
refactor(dark): use palettes through url queries in test pages

thetaPC:
Issue number: N/A

---------

If a dev wants to view a test page in dark mode, they have to manually add the styles. This can lead to a slowdown. Plus they can't use Playwright's `goto` to test both light and dark. In order to test dark mode with Playwright, the dev would need to use `setContent` instead of `goto`.

## What is the new behavior?

Dark mode can be added to any page by appending `palette=dark` to the URL.

- The param will be used to add a link tag with the correct palette file.
- Playwright will load the correct palette file when a dev uses `goto` and `{ themes: ['dark'] }`

- [ ] Yes
- [x] No

## Other information

I recommend using badge to try this out. It already has a `goto` in the basic tests.


core/scripts/testing/scripts.js
```diff
@@ -14,6 +14,26 @@
     document.head.appendChild(style);
   }
 
+  /**
+   * The term `palette` is used to as a param to match the
+   * Ionic docs, plus here is already a `ionic:theme` query being
+   * used for `md`, `ios`, and `ionic` themes.
+   */
+  const palette = window.location.search.match(/palette=([a-z]+)/);
```
brandyscarney, line 22: Why not match `ionic:palette`?

core/scripts/testing/scripts.js
```diff
@@ -14,6 +14,26 @@
...
+    const linkTag = document.createElement('link');
+    linkTag.setAttribute('rel', 'stylesheet');
+    linkTag.setAttribute('type', 'text/css');
+    /**
+     * The file must be the same as the one being used in Playwright tests
+     * when `dark` is being passed as a `theme` option.
+     * This provides consistency when Playwright is running through
+     * the test page or the `setContent` method.
+     */
+    linkTag.setAttribute('href', `/scripts/testing/themes/${palette[1]}.css`);
```
brandyscarney, line 33: Do we plan to update this for v8? https://github.com/ionic-team/ionic-framework/tree/feature-8.0/core/src/css/palettes

brandyscarney (commented):
It seems to work well! 👍 My only concern is the verbiage is a bit confusing since we are partially using `theme` and partially using `palette`, and this will need to be updated once v8 is in `main`.

core/scripts/testing/scripts.js
thetaPC: I didn't use that structure because palette is not a key within the `IonicConfig` [interface](https://github.com/ionic-team/ionic-framework/blob/c4aa090135c5fd6ea8f4e06d7faa9b982ee94ee6/core/src/utils/config.ts#L7). If I use `ionic:palette` then it might lead to confusion that it's part of the global configs.

core/scripts/testing/scripts.js
thetaPC: I've already updated it here since v8 came out first.

brandyscarney (commented):
Can we document this somewhere in `docs/`? RTL is documented here: https://github.com/ionic-team/ionic-framework/blob/2fc81ddc9b35d6909fd4b585079aedabbd659233/docs/CONTRIBUTING.md?plain=1#L146

Then I think this is good to go!

brandyscarney (approved):

PR #29240, opened 2024-03-29
chore(): create reusable test header

mapsandapps:
# Implemented

There's now a reusable header component that will allow a developer of Ionic Framework to toggle a number of things while viewing the index.html test files. This component can be reused across any of these files and includes:
- Dark mode toggle
- Theme switcher for iOS, MD, and Ionic themes
- LTR/RTL

It also has a slot so that settings can be added, e.g. settings that are  specific to the component being tested like the ones we have for some Datetime tests.

# Yet to do

I wasn't sure where to put the test-header.tsx file nor did I figure out how to prevent it from being included in Ionic Framework.

Other functionality we might want to add to the popover:
- Mode
- Locale
- Maybe more?

We still have some screenshots that include the whole page. If we don't want to have a bunch of irrelevant screenshot diffs, we could add css to make the options button invisible to playwright using [`stylePath`](https://playwright.dev/docs/api/class-pageassertions#page-assertions-to-have-screenshot-1-option-style-path). I think you can use this either in the `toHaveScreenshot` call or in the test config in playwright.config.ts:56

---------

Screenshot 2024-03-29 at 1 41 43 PM

Screenshot 2024-03-29 at 1 47 02 PM

### DIFF
--- a/core/api.txt
+++ b/core/api.txt
@@ -1678,3 +1678,6 @@ ion-toolbar,css-prop,--padding-bottom
 ion-toolbar,css-prop,--padding-end
 ion-toolbar,css-prop,--padding-start
 ion-toolbar,css-prop,--padding-top
+
+test-header,scoped
+test-header,prop,testTitle,string | undefined,undefined,false,false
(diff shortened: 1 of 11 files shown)


thetaPC:
Closing this PR as the team has decided to not add this feature. However, if we do circle back to the idea then we will reference this PR. Thank you so much!

PR #29241, opened 2024-03-29
fix(range): disable scroll when range is being dragged

thetaPC:
Issue number: internal
---------

There are a few tests that were disabled due to being flaky from gestures.

## What is the new behavior?

While fixing the tests, I found a bug that the scroll was never being disabled on scroll. Additionally, we were not taking into account that a custom scroll target could be used so it was never disabled either.

- Fixed the flaky tests.
- Content doesn't scroll when range is being dragged.
- Content can be either `ion-content` or a custom scroll target.

- [ ] Yes
- [x] No

## Other information

[Preview for `ion-content`](https://ionic-framework-git-fw-2873-ionic1.vercel.app/src/components/range/test/scroll)
[Preview for custom scroll target](https://ionic-framework-git-fw-2873-ionic1.vercel.app/src/components/range/test/scroll-target)

How to test:
1. Open either of the previews
2. Render the screen with the device simulator from the browser
3. Verify that you can scroll the page
4. Drag the range but don't let go
5. Verify that you cannot scroll the page
6. Repeat steps 2-5 with the other preview


core/src/components/range/range.tsx
```diff
@@ -325,7 +325,8 @@ export class Range implements ComponentInterface {
       this.setupGesture();
     }
 
-    this.contentEl = findClosestIonContent(this.el);
+    const ionContent = findClosestIonContent(this.el);
```
thetaPC, line 328: The scroll target test was failing because we didn't take into account that there might be a custom scroll target.

core/src/components/range/range.tsx
```diff
@@ -418,7 +419,7 @@ export class Range implements ComponentInterface {
      *
      * This only needs to be done once.
      */
-    if (contentEl && this.initialContentScrollY === undefined) {
+    if (contentEl && this.pressedKnob === undefined) {
```
thetaPC, line 422: The scroll was never being disabled which lead to flaky results.

core/src/components/range/test/range-events.e2e.ts
```diff
@@ -99,31 +107,6 @@ configs({ modes: ['ios'], directions: ['ltr'] }).forEach(({ title, config }) =>
 
         expect(rangeEndSpy.length).toBe(1);
       });
-
-      // TODO FW-2873
-      test.skip('should not scroll when the knob is swiped', async ({ page, skip }) => {
```
thetaPC, line 104: Moved this test to it's own test page to properly test the scroll since the basic page didn't have enough content to scroll on.

brandyscarney (approved):
LGTM

PR #29245, opened 2024-04-01
feat(tokens): add design tokens architecture

BenOsodrac:
## What is the new behavior?

Following the Design Doc for the new Ionic Design System, an architecture for Design Tokens was added.

1. Added new `foundations` folder inside the `src`, to hold all the tokens related files.
2. On this file, a json with all the tokens for the new Design System was added. The format followed was the one suggested on the [W3C draft](https://design-tokens.github.io/community-group/format/).
3. Added a dev dependency for [Styles Dictionary](https://amzn.github.io/style-dictionary/#/), version 3.9.2.
4. Added a tokens.js script that will take care of transforming the json tokens into the desired output. For now, three files are being generated: ionic.vars.css (:root with all CSS Variables); ionic.vars.css (scss variables for each token) and ionic.utility.css (a new utility-class for each token)
5. Added the script `npm run build.tokens` to package.json, that will generate the three files mentioned above.

For now, all these changes bring no impact to the rest of the Framework, as these variables are not yet being consumed.

The `margin`, `padding` and `border-radius` were removed from the prohibited properties on lint, to prevent lint errors with the new utility-classes. This is very open to discussion/feedback if it's seen as not ok.

The `build` command now includes the `build.tokens` script, to make sure the files are always generated, in case someone forget to run the command, after changing the json file!

- [ ] Yes
- [x] No


core/buildTokens.js
```diff
@@ -0,0 +1,147 @@
...
+StyleDictionary.registerFormat({
+  name: 'rootFormat',
+  formatter({ dictionary, file }) {
+    // Add a prefix to all variable names
+    const prefixedVariables = dictionary.allProperties.map((prop) => {
+      return `  --${prop.name}: ${prop.value};`;
+    });
+
+    return (
+      fileHeader({ file }) +
```
sean-perkins, line 19: Can we insert a comment block in the file contents that get generated that inform a user opening the file that it is auto-generated and they should not make changes in that file?

core/buildTokens.js
```diff
@@ -0,0 +1,147 @@
+// eslint-disable-next-line @typescript-eslint/no-var-requires
+const StyleDictionary = require('style-dictionary');
```
sean-perkins, line 2: Does `style-dictionary` have any provided types that we could use JSDoc annotations to import the type references for?

core/src/foundations/ionic.root.scss
```diff
@@ -0,0 +1,117 @@
+/**
+ * Ionic Design System
+ */
+
+:root {
+  --color-primary-10: #f7faff;
```
sean-perkins, line 6: Are we intentionally not prefixing these CSS variables to avoid conflicting with the existing `--ion` variables?

core/src/foundations/ionic.utility.scss
```diff
@@ -0,0 +1,115 @@
+/**
+ * Ionic Design System
+ */
+
+.color-primary-10 { color: #f7faff; }; .background-color-primary-10 {background-color: #f7faff;}
```
sean-perkins: ~ We should identify if prettier can format this output. This could be a technical debt ticket/not blocking this PR.

core/package.json
```diff
@@ -99,7 +100,8 @@
     "test.e2e.docker": "npm run docker.build && docker run -it --rm -e DISPLAY=$(cat docker-display.txt) -v $(cat docker-display-volume.txt) --ipc=host --mount=type=bind,source=./,target=/ionic ionic-playwright npm run test.e2e --",
     "test.e2e.docker.update-snapshots": "npm run test.e2e.docker -- --update-snapshots",
     "test.e2e.docker.ci": "npm run docker.build && docker run -e CI='true' --rm --ipc=host --mount=type=bind,source=./,target=/ionic ionic-playwright npm run test.e2e --",
-    "test.report": "npx playwright show-report"
+    "test.report": "npx playwright show-report",
+    "tokens": "node ./buildTokens.js"
```
sean-perkins: Question for understanding, when should a developer run this command?

core/buildTokens.js
```diff
@@ -0,0 +1,147 @@
+// eslint-disable-next-line @typescript-eslint/no-var-requires
```
sean-perkins, line 1: Existing Ionic script files are lowercase and use kebab-case if multi-word. Can we rename this file and update the script command?

core/buildTokens.js
BenOsodrac: Yes, its in fact the default behaviour 😂, I will adjust that!

core/src/foundations/ionic.root.scss
BenOsodrac: Yes, I was not sure at this moment what we would want to have going forward and I thought not using that a bit more safe. 

But that can be easily changed on the tokens script. Maybe it should be --ionic-, as this only affects the Ionic Design Systems and themes? But this is a decision that should be made on the scope of the rest of the Design Doc, as it also impacts the variables on the scope of each component for the Ionic Theme.

core/src/foundations/ionic.utility.scss
BenOsodrac: I already adjusted this on the script file and changed the `npm run tokens` script to also run `lint.sass.fix.` The generated file is different now.

Also, due to lint there were some errors that made me remove some prohibited properties (I added context on the PT description about this)

core/buildTokens.js
BenOsodrac: Ah, I actually had it named tokens.js and changed it to buildTokens.js as I saw the setupJest.js followed that approach. Will change it 👍

core/buildTokens.js
BenOsodrac: Hum, don't know, will try to check that!

core/buildTokens.js
BenOsodrac: Fixed!

core/buildTokens.js
BenOsodrac: Fixed!

core/src/foundations/ionic.root.scss
BenOsodrac: Just added a prefix to all generated variables 👍

core/src/foundations/ionic.vars.scss
```diff
@@ -0,0 +1,116 @@
+/**
+ * This is an auto-generated file, please do not change it directly.
+ * Ionic Design System
+ */
+
+$ionic-color-primary-10: #f7faff;
```
sean-perkins, line 6: What's the purpose of these sass variables? On a closer inspection they will be statically set based on the `_design-tokens.json`. If we reference these variables in our components, developers will not be able to customize the value.

I think these either need to be set to the CSS variable, with the default fallback to the static value or we don't generate them at all and reference the CSS variables directly.

core/src/foundations/ionic.root.scss
```diff
@@ -0,0 +1,118 @@
+/**
+ * This is an auto-generated file, please do not change it directly.
+ * Ionic Design System
+ */
+
+:root {
+--ionic-color-primary-10: #f7faff;
```
sean-perkins, line 7: ~ Non-blocking, we should see if we can format this or generate the output to match our indentation for CSS variables.

core/src/foundations/ionic.utility.scss
```diff
@@ -0,0 +1,587 @@
+/**
```
sean-perkins, line 1: ~ Non-blocking, same feedback here regarding formatting. If it isn't simple, we can come back and have prettier format these files.

core/src/foundations/ionic.vars.scss
BenOsodrac: The purpose was to follow a similar approach as we already have on the md/ios, where only scss variables are used internally and no global :root is defined by default. The would be used internally at component level and to set the values on the global Ionic the defaults, as suggested here: https://github.com/BenOsodrac/ionic-framework/blob/new-ionic-ds/core/src/themes/ionic/ionic.theme.default.scss

But I'm ok with using only the native CSS Variables, its in fact more inline on how I'm used to do things 👍

sean-perkins: I think it makes sense to have scss variables to use internally, but I think they need to be like `$background-color-step` variables: https://github.com/BenOsodrac/ionic-framework/blob/f3e01bb0057f6c917d582efdf4208829cc6eb14f/core/src/themes/ionic/ionic.theme.default.scss#L106.

With the current state, if I had this:
```
/* button.ionic.scss */
:host {
  color: $ionic-color-primary-10;
}
```

When the stylesheet is compiled the generated output for the developer consuming would be:
```
:host {
  color: #f7faff;
}
```

If a developer tries to customize the color palette in their application:
```
:root {
  --ionic-color-primary-10: orange;
}
```

The button stylesheet will still be `#f7faff`. This isn't desired. 

However if we have:
```
$ionic-color-primary-10: var(--ionic-color-primary-10);
```

Then a developer consuming the components can customize the color used in the component globally.

core/buildTokens.js
BenOsodrac: The have ts types: https://github.com/amzn/style-dictionary/tree/main/types, would that be useful for JSDoc?

core/src/foundations/ionic.vars.scss
BenOsodrac: Your PR is merged 👍, lets go 🚀

sean-perkins (approved):
Great work on this 👍

core/package.json
```diff
@@ -99,7 +100,8 @@
     "test.e2e.docker": "npm run docker.build && docker run -it --rm -e DISPLAY=$(cat docker-display.txt) -v $(cat docker-display-volume.txt) --ipc=host --mount=type=bind,source=./,target=/ionic ionic-playwright npm run test.e2e --",
     "test.e2e.docker.update-snapshots": "npm run test.e2e.docker -- --update-snapshots",
     "test.e2e.docker.ci": "npm run docker.build && docker run -e CI='true' --rm --ipc=host --mount=type=bind,source=./,target=/ionic ionic-playwright npm run test.e2e --",
-    "test.report": "npx playwright show-report"
+    "test.report": "npx playwright show-report",
+    "tokens": "node ./tokens.js && npm run lint.sass.fix"
```
liamdebeasi, line 104: This script should probably be run on build. My main concern here is what happens if someone modifies the JSON file but forgets to run this script? We'd potentially ship with outdated styles.

I'd recommend doing something like this:

1. Update the `build` script above to call `npm run tokens` (You could also rename this to `build.tokens` if you wanted naming consistency). This ensures that the latest tokens are generated on `npm run build`. Note that this should be run **before** `npm run build.css` since the tokens script outputs scss.
2. In the [`build-core` workflow](https://github.com/ionic-team/ionic-framework/blob/b5cb828861c568919a8a8f1090677f37dc1ce424/.github/workflows/actions/build-core/action.yml#L34)  and the [`build-core-stencil-prerelease` workflow](https://github.com/ionic-team/ionic-framework/blob/b5cb828861c568919a8a8f1090677f37dc1ce424/.github/workflows/actions/build-core-stencil-prerelease/action.yml#L32) make sure the `core/src/foundations` directory is cached.

The benefit of this is if anyone forgets to build the latest scss file the [`test-core-clean-build`](https://github.com/ionic-team/ionic-framework/blob/main/.github/workflows/actions/test-core-clean-build/action.yml) will fail in CI.

core/src/foundations/ionic.vars.scss
```diff
@@ -0,0 +1,118 @@
+/**
+ * This is an auto-generated file, please do not change it directly.
+ * Ionic Design System
+ */
+
+$ionic-color-primary-10: var(--ionic-color-primary-10, #f7faff);
```
liamdebeasi, line 6: If these variables are guaranteed to always be defined due to `ionic.root.scss` why we do we need to repeat the fallback values here?

core/tokens.js
```diff
@@ -0,0 +1,152 @@
+// eslint-disable-next-line @typescript-eslint/no-var-requires
+const StyleDictionary = require('style-dictionary');
```
liamdebeasi, line 2: Can we add a sentence or two about why we are doing this? For example, it would be good to document why we use `style-dictionary` as opposed to just hardcoding these values directly.

liamdebeasi (changes requested):
Overall looks good! I have some questions surrounding long term maintenance of this.

core/package.json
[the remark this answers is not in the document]
BenOsodrac: Yes, that makes sense, already did the change!

BenOsodrac: Hum thank you for this feedback, I think it makes total sense 👍 

1. Already updated the script name and move it to the build command.
2. Added the `core/src/foundations` to the path attribute of those actions. I don't have much context about this, is there anything more I should do?

core/tokens.js
BenOsodrac: Done!

core/src/foundations/ionic.root.scss
```diff
@@ -0,0 +1,120 @@
+/**
+ * This is an auto-generated file, please do not change it directly.
+ * Ionic Design System
+ */
+
+:root {
+  --ionic-color-primary-10: #f7faff;
```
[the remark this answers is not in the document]
BenOsodrac: Yes, we did! 

In summary, as we are introducing a new Design System, we wanted to avoid confusion and conflicts with the current one. 

As the new theme is also called Ionic, we thought the `--ionic` prefix makes total sense for these new tokens. 

This decision is also present on the [Decision Log](https://outsystemsrd.atlassian.net/wiki/spaces/RDMBVS/pages/4057366529/Native+Look+Feel+-+Decisions+Log) page 👍

core/src/foundations/ionic.vars.scss
sean-perkins: This is to ensure that the variables always have a value assigned to them. Developers should be importing a global stylesheet that includes the content of `ionic.root.scss` or assigning their own values to these variables. 

This approach is most helpful in cases where the variable assignment may fail or not be initially available at the time of render. 

It is better to have a fallback experience/value vs. a completely broken experience. 

If we decided to remove this in the future, it is pretty trivial to update the script that generates that format.

core/src/foundations/ionic.root.scss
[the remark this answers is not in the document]
liamdebeasi: sounds good, thanks!

core/package.json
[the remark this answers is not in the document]
liamdebeasi: Perfect, thanks!

liamdebeasi: Looks great! I tested this out on a separate branch, and CI correctly fails if you change the source json file without re-running the build: https://github.com/ionic-team/ionic-framework/actions/runs/8560100688/job/23458371826?pr=29276

.github/workflows/actions/build-core-stencil-prerelease/action.yml
```diff
@@ -1,12 +1,12 @@
...
+description: "Build Ionic Core with a Prerelease Build of Stencil"
 inputs:
   stencil-version:
-    description: 'The NPM tag of @stencil/core to install.'
+    description: "The NPM tag of @stencil/core to install."
     type: string
     required: true
 runs:
-  using: 'composite'
+  using: "composite"
```
liamdebeasi, line 9: Can we revert the quotation marks change?

.github/workflows/actions/build-core/action.yml
```diff
@@ -1,12 +1,12 @@
...
+description: "Build Ionic Core"
 inputs:
   ionicons-version:
-    description: 'The NPM tag of ionicons to install.'
+    description: "The NPM tag of ionicons to install."
     type: string
     required: false
 runs:
-  using: 'composite'
+  using: "composite"
```
liamdebeasi, line 9: Can we revert the quotation marks change?

liamdebeasi (approved):
Good to go on my end once the below changes are made. Great work!

.github/workflows/actions/build-core-stencil-prerelease/action.yml
BenOsodrac: Hum some stupid default config from Prettier, already changed it!

.github/workflows/actions/build-core/action.yml
BenOsodrac: Done!

sean-perkins (approved):

PR #29249, opened 2024-04-01
feat(input): add the large size for the ionic theme

brandyscarney:
Issue number: internal

---------

Input does not have a size property.

## What is the new behavior?

- Adds the `size` property with support for the `"large"` size on the `"ionic"` theme only
- Adds tests for the size property
- Note: the screenshots will not look right until the fill styles are added

- [ ] Yes
- [x] No


core/src/components/input/test/size/index.html
```diff
@@ -0,0 +1,84 @@
...
+        <div class="grid">
+          <div class="grid-item">
+            <h2>No Fill: No Size</h2>
+            <ion-input label="Email"></ion-input>
+          </div>
+
+          <div class="grid-item">
+            <h2>Solid: No Size</h2>
+            <ion-input fill="solid" label="Email"></ion-input>
+          </div>
```
thetaPC, line 59: OS doen't have a solid appearance, this can be removed along with the large version.

thetaPC (changes requested):

core/src/components/input/input.tsx
```diff
@@ -251,6 +252,12 @@ export class Input implements ComponentInterface {
    */
   @Prop() step?: string;
 
+  /**
+   * The size of the input. If "large" it will have an increased height. By default the
```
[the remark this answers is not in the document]
brandyscarney: https://github.com/ionic-team/ionic-framework/pull/29249/commits/f9727438d04c93833de88988625574fe0ce862f3

core/src/components/input/test/size/index.html
brandyscarney: Good catch, I was copy/pasting from another test! Fixed in https://github.com/ionic-team/ionic-framework/pull/29249/commits/9f13557c69d067d285429dba5e6f8c8f2a704179

thetaPC (approved):
LGTM, verified the size increase locally

sean-perkins (approved):

PR #29256, opened 2024-04-02
chore(docs): move docs to the packages directory

brandyscarney:
Move the `docs/` directory into the `packages/` directory. This will enable us to create a top level `docs/` folder for onboarding and contributing guidelines. It will also be necessary when migrating to pnpm.


docs/.npmrc
sean-perkins, line 1: This file is removed, but not tracked in the `packages/docs` folder. Is that intentional?

.github/workflows/release-ionic.yml
```diff
@@ -53,7 +53,7 @@ jobs:
       uses: ./.github/workflows/actions/download-archive
       with:
         name: ionic-docs
-        path: ./docs
+        path: .packages/docs
```
sean-perkins, line 56: Shouldn't this be:
```
path: ./packages/docs
```

sean-perkins (changes requested):

.github/workflows/release-ionic.yml
brandyscarney: oops https://github.com/ionic-team/ionic-framework/pull/29256/commits/9c6927d429ffd171286ed532c1fdf5487a2ddc31

sean-perkins (approved):

PR #29257, opened 2024-04-03
docs(contributing): using vscode on windows

sean-perkins:
Issue number: N/A

---------

Developers on a Windows environment using VS Code will experience an issue when using prettier, where files will be tracked as modified even though they have no actual diffs.

This is due to differences in how Unix based operating systems and Windows treat carriage returns.

## What is the new behavior?

- Updates the contributing guide to educate developers using VS Code on how to configure their local environment to not track additional diffs when formatting the codebase. 

- [ ] Yes
- [x] No

## Other information

This is the same documentation that exists on the ionic-docs repository for the same reason. I've applied co-authored-by credit to Netkow for his wonderful contribution to that documentation.


dotNetkow:
Thanks pal!

thetaPC (approved):
LGTM

PR #29268, opened 2024-04-03
Fw 6082

amandaejohnston:
Issue number: resolves #

---------

## What is the new behavior?

-
-
-

- [ ] Yes
- [ ] No

### DIFF
--- a/core/src/components/input/input.ionic.outline.scss
+++ b/core/src/components/input/input.ionic.outline.scss
@@ -4,46 +4,11 @@
 // ----------------------------------------------------------------
 
 :host(.input-fill-outline) {
-  --border-radius: 4px;
-  --padding-start: 16px;
-  --padding-end: 16px;
+  --border-radius: 8px;
+  --padding-start: 20px;
+  --padding-end: 20px;
 
   min-height: 56px;
-}
-
-:host(.input-fill-outline.input-shape-round) {
-  --border-radius: 28px;
-  --padding-start: 32px;
-  --padding-end: 32px;
-}
-
-/**
- * If the input has a validity state, the
- * border should reflect that as a color.
- */
-:host(.has-focus.input-fill-outline.ion-valid),
-:host(.input-fill-outline.ion-touched.ion-invalid) {
-  --border-color: var(--highlight-color);
-}
-
-/**
- * Border should be
- * slightly darker on hover.
- */
-@media (any-hover: hover) {
-  :host(.input-fill-outline:hover) {
-    --border-color: #{$background-color-step-750};
-  }
-}
-
-/**
- * The border should get thicker
- * and take on component color when
- * the input is focused.
- */
-:host(.input-fill-outline.has-focus) {
-  --border-width: var(--highlight-height);
-  --border-color: var(--highlight-color);
 }
 
 /**
@@ -80,23 +45,8 @@
  * The label should appear on top of an outline
  * container that overlaps it so it is always clickable.
  */
-:host(.input-fill-outline) .label-text-wrapper,
 :host(.input-fill-outline) .label-text-wrapper {
   position: relative;
-}
-
-/**
- * This makes the label sit above the input.
- */
-:host(.label-floating.input-fill-outline) .label-text-wrapper {
-  @include transform(translateY(-32%), scale(#{$form-control-label-stacked-scale}));
-  @include margin(0);
-
-  /**
-   * Label text should not extend
-   * beyond the bounds of the input.
-   */
-  max-width: calc((100% - var(--padding-start) - var(--padding-end) - #{$input-ionic-floating-label-padding * 2}) / #{$form-control-label-stacked-scale});
 }
 
 /**
@@ -109,8 +59,17 @@
   @include margin(6px, 0, 6px, 0);
 }
 
-// Input Fill: Outline Outline Container
+
+// Input Fill: Outline, Outline Container
 // ----------------------------------------------------------------
+
+/**
+ * Note that while the ionic theme does not make use of the notch,
+ * thus avoiding the need for separate outline container parts at all,
+ * the element's structure has been kept the same to simplify the
+ * component render function and keep selectors consistent between
+ * the ionic and md themes.
+ */
 
 :host(.input-fill-outline) .input-outline-container {
   @include position(0, 0, 0, 0);
@@ -129,9 +88,9 @@
 }
 
 /**
- * By default, each piece of the container should have
- * a top and bottom border. This gives the appearance
- * of a unified container with a border.
+ * Each piece of the container should have a top and bottom
+ * border. This gives the appearance of a unified container
+ * with a border.
  */
 :host(.input-fill-outline) .input-outline-start,
 :host(.input-fill-outline) .input-outline-notch,
@@ -149,51 +108,22 @@
 }
 
 /**
- * This element ensures that the notch used
- * the size of the scaled text so that the
- * border cut out is the correct width.
- * The text in this element should not
- * be interactive.
+ * In the md theme, this element ensures that the notch
+ * uses the size of the scaled text so that the border
+ * cut-out is the correct width. However, a notch is not
+ * shown in the ionic theme, so we only need to make sure
+ * the text in this element is not visible or interactive.
  */
 :host(.input-fill-outline) .notch-spacer {
-  /**
-   * We need $input-ionic-floating-label-padding of padding on the right.
-   * However, we also subtracted $input-ionic-floating-label-padding from
-   * the width of .input-outline-start
-   * to create space, so we need to take
-   * that into consideration here.
-   */
-  @include padding(null, #{$input-ionic-floating-label-padding * 2}, null, null);
-
-  font-size: calc(1em * #{$form-control-label-stacked-scale});
-
   opacity: 0;
   pointer-events: none;
-
-  /**
-   * The spacer currently inherits
-   * border-box sizing from the Ionic reset styles.
-   * However, we do not want to include padding in
-   * the calculation of the element dimensions.
-   * This code can be removed if input is updated
-   * to use the Shadow DOM.
-   */
-  box-sizing: content-box;
 }
 
 :host(.input-fill-outline) .input-outline-start {
   @include border-radius(var(--border-radius), 0px, 0px, var(--border-radius));
   @include border(null, null, null, var(--border-width) var(--border-style) var(--border-color));
 
-  /**
-   * There should be spacing between the translated text
-   * and .input-outline-start. However, we can't add this
-   * spacing onto the notch because it would cause the
-   * label to look like it is not aligned with the
-   * text input. Instead, we subtract a few pixels from
-   * this element.
-   */
-  width: calc(var(--padding-start) - #{$input-ionic-floating-label-padding});
+  width: calc(var(--padding-start));
 }
 
 :host(.input-fill-outline) .input-outline-end {
@@ -207,12 +137,26 @@
   flex-grow: 1;
 }
 
-/**
- * When the input either has focus or a value,
- * there should be a "cut out" at the top for
- * the floating/stacked label. We simulate this "cut out"
- * by removing the top border from the notch fragment.
- */
-:host(.label-floating.input-fill-outline) .input-outline-notch {
-  border-top: none;
+
+// Input Fill: Outline, Label Placement: Stacked
+// ----------------------------------------------------------------
+
+// Give room for the label to sit above the outline container.
+:host(.input-fill-outline.input-label-placement-stacked) {
+  // !important is needed to overwrite another !important on the host/base styles.
+  /* stylelint-disable-next-line declaration-no-important */
+  padding-top: 20px !important;
+}
+
+// This makes the label sit above the input.
+:host(.label-floating.input-fill-outline.input-label-placement-stacked:not(.input-shape-round)) .label-text-wrapper {
+  @include transform(translateY(-100%), scale(#{$form-control-label-stacked-scale}));
+  @include margin(0);
+  @include position(null, null, null, 0);
+
+  /**
+   * Label text should not extend
+   * beyond the bounds of the input.
+   */
+  max-width: calc((100% - var(--padding-start) - var(--padding-end)) / #{$form-control-label-stacked-scale});
 }
--- a/core/src/components/input/input.ionic.outline.scss
+++ b/core/src/components/input/input.ionic.outline.scss
@@ -1,0 +1,218 @@
+@import "./input.vars";
+
+// Input Fill: Outline (Ionic Theme)
+// ----------------------------------------------------------------
+
+:host(.input-fill-outline) {
+  --border-radius: 4px;
+  --padding-start: 16px;
+  --padding-end: 16px;
+
+  min-height: 56px;
+}
+
+:host(.input-fill-outline.input-shape-round) {
+  --border-radius: 28px;
+  --padding-start: 32px;
+  --padding-end: 32px;
+}
+
+/**
+ * If the input has a validity state, the
+ * border should reflect that as a color.
+ */
+:host(.has-focus.input-fill-outline.ion-valid),
+:host(.input-fill-outline.ion-touched.ion-invalid) {
+  --border-color: var(--highlight-color);
+}
+
+/**
+ * Border should be
+ * slightly darker on hover.
+ */
+@media (any-hover: hover) {
+  :host(.input-fill-outline:hover) {
+    --border-color: #{$background-color-step-750};
+  }
+}
+
+/**
+ * The border should get thicker
+ * and take on component color when
+ * the input is focused.
+ */
+:host(.input-fill-outline.has-focus) {
+  --border-width: var(--highlight-height);
+  --border-color: var(--highlight-color);
+}
+
+/**
+ * The bottom content should never have
+ * a border with the outline style.
+ */
+:host(.input-fill-outline) .input-bottom {
+  border-top: none;
+}
+
+/**
+ * Outline inputs do not have a bottom border.
+ * Instead, they have a border that wraps the
+ * input + label.
+ */
+:host(.input-fill-outline) .input-wrapper {
+  border-bottom: none;
+}
+
+:host(.input-fill-outline.input-label-placement-stacked) .label-text-wrapper,
+:host(.input-fill-outline.input-label-placement-floating) .label-text-wrapper {
+  @include transform-origin(start, top);
+
+  position: absolute;
+
+  /**
+   * Label text should not extend
+   * beyond the bounds of the input.
+   */
+  max-width: calc(100% - var(--padding-start) - var(--padding-end));
+}
+
+/**
+ * The label should appear on top of an outline
+ * container that overlaps it so it is always clickable.
+ */
+:host(.input-fill-outline) .label-text-wrapper,
+:host(.input-fill-outline) .label-text-wrapper {
+  position: relative;
+}
+
+/**
+ * This makes the label sit above the input.
+ */
+:host(.label-floating.input-fill-outline) .label-text-wrapper {
+  @include transform(translateY(-32%), scale(#{$form-control-label-stacked-scale}));
+  @include margin(0);
+
+  /**
+   * Label text should not extend
+   * beyond the bounds of the input.
+   */
+  max-width: calc((100% - var(--padding-start) - var(--padding-end) - #{$input-md-floating-label-padding * 2}) / #{$form-control-label-stacked-scale});
+}
+
+/**
+ * This ensures that the input does not
+ * overlap the floating label while still
+ * remaining visually centered.
+ */
+:host(.input-fill-outline.input-label-placement-stacked) input,
+:host(.input-fill-outline.input-label-placement-floating) input {
+  @include margin(6px, 0, 6px, 0);
+}
+
+// Input Fill: Outline Outline Container
+// ----------------------------------------------------------------
+
+:host(.input-fill-outline) .input-outline-container {
+  @include position(0, 0, 0, 0);
+
+  display: flex;
+
+  position: absolute;
+
+  width: 100%;
+  height: 100%;
+}
+
+:host(.input-fill-outline) .input-outline-start,
+:host(.input-fill-outline) .input-outline-end {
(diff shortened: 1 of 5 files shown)


core/src/components/input/input.ionic.scss
```diff
@@ -1,10 +1,14 @@
...
+@import "./input.ionic.outline.scss";
 
 // Ionic Input
 // --------------------------------------------------
 
-
+:host {
+  --border-width: 1px;
+  --border-color: #B8B8B8;
+}
```
amandaejohnston, line 11: I went ahead and put these styles here instead of the outline-specific file since they'll be needed for the underline appearance as well.

core/src/components/input/input.ionic.outline.scss
```diff
@@ -0,0 +1,162 @@
...
+// Input Fill: Outline, Outline Container
+// ----------------------------------------------------------------
+
+/**
+ * Note that while the ionic theme does not make use of the notch,
+ * thus avoiding the need for separate outline container parts at all,
+ * the element's structure has been kept the same to simplify the
+ * component render function and keep selectors consistent between
+ * the ionic and md themes.
+ */
```
amandaejohnston, line 72: I can change the render function instead if desired; this just seemed easier to follow to me.

core/src/components/input/test/size/input.e2e.ts-snapshots/input-size-large-ionic-md-ltr-light-Mobile-Chrome-linux.png
amandaejohnston, line 1: This change is due to the default label placement changing to `stacked` for this theme.

core/src/components/input/test/size/input.e2e.ts-snapshots/input-size-large-outline-ionic-md-ltr-light-Mobile-Chrome-linux.png
amandaejohnston, line 1: While this looks broken currently, I intentionally left it as-is because the test uses a `floating` label placement, which this ticket explicitly doesn't cover. When floating labels are implemented later, these screenshots will be fixed. LMK if you'd like me to change the relevant `size="large"` tests to use `stacked` instead.

core/src/components/input/input.ionic.scss
```diff
@@ -1,10 +1,14 @@
...
 @import "./input.ionic.vars";
+@import "./input.ionic.outline.scss";
 
 // Ionic Input
 // --------------------------------------------------
 
-
+:host {
+  --border-width: 1px;
+  --border-color: #B8B8B8;
```
amandaejohnston, line 10: The border color value is taken straight from the OS-provided Figma file.

core/src/components/input/input.ionic.outline.scss
```diff
@@ -0,0 +1,162 @@
...
+
+// Input Fill: Outline (Ionic Theme)
+// ----------------------------------------------------------------
+
+:host(.input-fill-outline) {
+  --border-radius: 8px;
+  --padding-start: 20px;
+  --padding-end: 20px;
+
+  min-height: 56px;
```
thetaPC, line 11: We need to reach out to the designers for validation. Their [UX requirements](https://outsystemsrd.atlassian.net/wiki/spaces/UXD/pages/4011295407/Input+-+UX+Requirements#Requirements-List) has the default as 40px but their Figma shows 56px.

amandaejohnston: Good catch; it turns out I based my styles off the extra-large size in the Figma by accident 😅 I'll get this updated to use the values for the default size.

amandaejohnston: Fixed in https://github.com/ionic-team/ionic-framework/pull/29268/commits/b31f71929b3289154020e3bb44df0d425ce87e41, screenshots updated 👍

core/src/components/input/input.ionic.outline.scss
```diff
@@ -0,0 +1,162 @@
...
+
+/**
+ * This ensures that the input does not
+ * overlap the floating label while still
+ * remaining visually centered.
+ */
+:host(.input-fill-outline.input-label-placement-stacked) input,
+:host(.input-fill-outline.input-label-placement-floating) input {
+  @include margin(6px, 0, 6px, 0);
+}
```
thetaPC, line 60: Is this snippet necessary since the stacked label is now on top of the outline? It shouldn't overlap with the input.

core/src/components/input/input.ionic.outline.scss
thetaPC: This works for me, it lessens the complexity of the structure if we only use styling. Plus, the comment notes what is going on.

core/src/components/input/input.tsx
```diff
@@ -182,7 +182,8 @@ export class Input implements ComponentInterface {
    * `"stacked"`: The label will appear smaller and above the input regardless even when the input is blurred or has no value.
    * `"fixed"`: The label has the same behavior as `"start"` except it also has a fixed width. Long text will be truncated with ellipses ("...").
    */
-  @Prop() labelPlacement: 'start' | 'end' | 'floating' | 'stacked' | 'fixed' = 'start';
+  @Prop() labelPlacement: 'start' | 'end' | 'floating' | 'stacked' | 'fixed' =
```
sean-perkins, line 185: We should explore/think on this more. Using a call expression and variable reference shows up in the generated `api.txt` and docs output. 

We could probably remove the default value and infer the label placement based on the theme (if it is not set) in the implementation. This would make the generated output clearer.

core/src/components/input/input.tsx
```diff
@@ -471,6 +472,21 @@ export class Input implements ComponentInterface {
     return typeof this.value === 'number' ? this.value.toString() : (this.value || '').toString();
   }
 
+  private getLabelPlacement() {
+    const theme = getIonTheme(this);
+    const { el, labelPlacement } = this;
+
+    if (theme === 'ionic' && labelPlacement !== 'stacked' && labelPlacement !== 'floating') {
```
sean-perkins, line 479: ~ I don't know if we need the overhead for runtime checks for every unsupported prop/prop option for a given theme. 

Open to team feedback on this, but this feels like a problem to solve closer to beta/rc/ga. We may fill these feature gaps in that time and the documentation may be clear enough for developers to navigate.

core/src/components/input/input.tsx
```diff
@@ -182,7 +182,8 @@ export class Input implements ComponentInterface {
    * `"stacked"`: The label will appear smaller and above the input regardless even when the input is blurred or has no value.
    * `"fixed"`: The label has the same behavior as `"start"` except it also has a fixed width. Long text will be truncated with ellipses ("...").
```
sean-perkins, line 183: The documentation on this property should include where it is supported (which theme, if it isn't supported on all).

core/src/components/input/input.ionic.outline.scss
```diff
@@ -0,0 +1,162 @@
...
+
+:host(.input-fill-outline.input-label-placement-stacked) .label-text-wrapper,
+:host(.input-fill-outline.input-label-placement-floating) .label-text-wrapper {
+  @include transform-origin(start, top);
+
+  position: absolute;
+
+  /**
+   * Label text should not extend
```
sean-perkins, line 38: ~ this looks to be copy and paste, we could probably remove the 40 character line wrapping since the file already extends wider than this line would be.

core/src/components/input/input.ionic.outline.scss
```diff
@@ -0,0 +1,162 @@
...
+  position: relative;
+}
+
+/**
+ * This ensures that the input does not
+ * overlap the floating label while still
+ * remaining visually centered.
+ */
+:host(.input-fill-outline.input-label-placement-stacked) input,
+:host(.input-fill-outline.input-label-placement-floating) input {
```
sean-perkins, line 58: Why are we including floating styles here? That is out of scope.

core/src/components/input/input.ionic.outline.scss
```diff
@@ -0,0 +1,162 @@
...
+
+
+// Input Fill: Outline, Label Placement: Stacked
+// ----------------------------------------------------------------
+
+// Give room for the label to sit above the outline container.
+:host(.input-fill-outline.input-label-placement-stacked) {
+  // !important is needed to overwrite another !important on the host/base styles.
+  /* stylelint-disable-next-line declaration-no-important */
+  padding-top: 20px !important;
```
sean-perkins, line 148: 1. Why are we not using our padding mixin here and disabling stylelint?
2. Should we correct the usage of `!important` in the parent file instead of continuing the practice?

core/src/components/input/input.ionic.outline.scss
```diff
@@ -0,0 +1,162 @@
...
+
+/**
+ * In the md theme, this element ensures that the notch
+ * uses the size of the scaled text so that the border
+ * cut-out is the correct width. However, a notch is not
+ * shown in the ionic theme, so we only need to make sure
+ * the text in this element is not visible or interactive.
+ */
+:host(.input-fill-outline) .notch-spacer {
+  opacity: 0;
```
sean-perkins, line 118: If the notch is not shown, why are not using `display: none` here?

sean-perkins (changes requested):
I think the implementation should adjust the rendering of the outline container to not include MD specific elements for the ionic theme. 

This would simplify the styling and the cognitive load of trying to understand why we are rendering an element and then visually hiding it for a theme.

core/src/components/input/input.ionic.outline.scss
amandaejohnston: Good call, removed in https://github.com/ionic-team/ionic-framework/pull/29268/commits/23aad82994a2651ecf83e9f695458ead049181f0.

core/src/components/input/input.tsx
amandaejohnston: Updated in https://github.com/ionic-team/ionic-framework/pull/29268/commits/8d4e8f5296966eb538b7c598380af306359e9562. I used the pattern from the `alignment` prop in `ion-popover`.

core/src/components/input/input.tsx
amandaejohnston: I'm open to removing the check and/or warning, but it was included in the design doc and is a pattern we're already using for other features. The `size` prop on `ion-input` is already implemented this way, for example.

core/src/components/input/input.tsx
amandaejohnston: Added in https://github.com/ionic-team/ionic-framework/pull/29268/commits/8d4e8f5296966eb538b7c598380af306359e9562.

core/src/components/input/input.ionic.outline.scss
amandaejohnston: True that this was copy-pasted, but it's just Liam's commenting style; this sort of thing is all over our codebase right now 😄

core/src/components/input/input.ionic.outline.scss
amandaejohnston: The styles referenced here were removed in https://github.com/ionic-team/ionic-framework/pull/29268/commits/23aad82994a2651ecf83e9f695458ead049181f0, but I removed another instance of floating label styles in https://github.com/ionic-team/ionic-framework/pull/29268/commits/4310eae9af6c1e9e6161a0141a3040e20f112fda.

core/src/components/input/input.ionic.outline.scss
amandaejohnston: 1. I didn't realize you could use `!important` with the mixin, and it also isn't required for `padding-top` (or `bottom` for that matter) since the mixin's purpose is to add RTL support. I went ahead and changed it in https://github.com/ionic-team/ionic-framework/pull/29268/commits/2288f1a3116b720be0f9f59976e3530496af8ca4 though.
2. Should we? It seemed out of scope for this PR since the base styles aren't related to the feature and I assume the original `!important` was put there for a reason. I can make a tech debt ticket if you want?

core/src/components/input/input.ionic.outline.scss
amandaejohnston: The notch was removed per other comments, so marking this as resolved.

core/src/components/input/input.ionic.outline.scss
```diff
@@ -0,0 +1,162 @@
...
+   */
+  max-width: calc(100% - var(--padding-start) - var(--padding-end));
+}
+
+/**
+ * The label should appear on top of an outline
+ * container that overlaps it so it is always clickable.
+ */
+:host(.input-fill-outline) .label-text-wrapper {
+  position: relative;
```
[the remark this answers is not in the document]
amandaejohnston: Fixed in https://github.com/ionic-team/ionic-framework/pull/29268/commits/038f2f99cd5ae74b71fd5262b127947555583ff2. This necessitated a bit of rework because the label text was previously being translated above the `<label>` element (and the whole component, for that matter), causing clicks to not be captured. I had to rearrange things so that the `<label>` bounds cover both the label text and the native input wrapper, which in turn required moving the outline since otherwise it would surround the label text as well.

core/src/components/input/input.ionic.outline.scss
amandaejohnston: I wound up removing this code in https://github.com/ionic-team/ionic-framework/pull/29268/commits/038f2f99cd5ae74b71fd5262b127947555583ff2 anyway, so marking this as resolved. See https://github.com/ionic-team/ionic-framework/pull/29268#discussion_r1552182072 for relevant discussion.

amandaejohnston:
> We should probably get clarification on implementing the start/end slots. I don't think we have a separate ticket for that. I would have assumed that it would be part of this.

Added some styles in https://github.com/ionic-team/ionic-framework/pull/29268/commits/d17c86536c4f99faafbfec4aa8875f02af598939, screenshots incoming once I fix up everything else. It looks like the default ion-icon size is already 16px as prescribed, so I only touched the margins; I figure any further sizing details should be up to the app since slots are designed for more flexibility.

core/src/components/input/input.ionic.scss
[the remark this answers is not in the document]
amandaejohnston: Colors updated in https://github.com/ionic-team/ionic-framework/pull/29268/commits/1f0a57476fcafc4e8a64f40552e1ef8da787a58e. Font size seems like more of a global typography thing? The default font size for the ionic theme is 14px, and ion-input's font size is currently set to `inherit`, so that should just work once everything is put together. At least, as far as I understand it.

core/src/components/input/test/fill/input.e2e.ts
```diff
@@ -248,3 +248,24 @@ configs({ modes: ['md'], directions: ['ltr'] }).forEach(({ title, screenshot, co
...
+            label="Email"
+            label-placement="stacked"
+            placeholder="example@ionic.io"
+          ></ion-input>
+        `,
+        config
+      );
+
+      const input = page.locator('ion-input');
+      await expect(input).toHaveScreenshot(screenshot(`input-ionic-theme-fill-outline-label-stacked`));
```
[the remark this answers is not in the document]
amandaejohnston: Fixed in https://github.com/ionic-team/ionic-framework/pull/29268/commits/08053a6062de8ed84c846b8493d9229858075007.

core/src/components/input/input.ionic.outline.scss
```diff
@@ -0,0 +1,110 @@
...
+   */
+  position: relative;
+
+  color: #373737;
+}
+
+:host(.input-fill-outline) .native-wrapper {
+  @include padding(var(--padding-top), var(--padding-end), var(--padding-bottom), var(--padding-start));
+
+  min-height: 40px;
```
thetaPC, line 64: This change causes the `large` [size styling](https://github.com/ionic-team/ionic-framework/blob/41df86257753ba45da5f8c2ce02f795807ee7726/core/src/components/input/input.ionic.scss#L19-L21) to no longer work. We'll need to adjust the large size for the outline.

core/src/components/input/input.ionic.outline.scss
```diff
@@ -0,0 +1,110 @@
+@import "./input.vars";
+
+// Input Fill: Outline (Ionic Theme)
+// ----------------------------------------------------------------
+
+:host(.input-fill-outline) {
+  --border-radius: 4px;
+  --padding-start: 12px;
+  --padding-end: 12px;
+  --placeholder-color: #535353;
```
thetaPC, line 10: This hex is an Ionic design token: `--ionic-color-neutral-600`. I'm not sure if the branch has the latest from https://github.com/ionic-team/ionic-framework/pull/29245. If it does then we should be able to do:
```suggestion
  --placeholder-color: $ionic-color-neutral-600;
```

core/src/components/input/input.ionic.outline.scss
```diff
@@ -0,0 +1,110 @@
+@import "./input.vars";
+
+// Input Fill: Outline (Ionic Theme)
+// ----------------------------------------------------------------
+
+:host(.input-fill-outline) {
+  --border-radius: 4px;
```
thetaPC, line 7: This hex is an Ionic design token: `--ionic-border-radius-rounded-small`. I'm not sure if the branch has the latest from https://github.com/ionic-team/ionic-framework/pull/29245. If it does then we should be able to do: 
```suggestion
  --border-radius: $ionic-border-radius-rounded-small;
```

core/src/components/input/input.ionic.outline.scss
```diff
@@ -0,0 +1,110 @@
...
+}
+
+:host(.input-fill-outline) .label-text-wrapper {
+  /**
+   * The label should appear on top of an outline
+   * container that overlaps it so it is always clickable.
+   */
+  position: relative;
+
+  color: #373737;
```
thetaPC, line 58: This hex is an Ionic design token:`ionic-color-neutral-700`. I'm not sure if the branch has the latest from https://github.com/ionic-team/ionic-framework/pull/29245. If it does then we should be able to do:
```suggestion
  --color: $ionic-color-neutral-700;
```

Also it should be using the `--color` to override.

core/src/components/input/input.tsx
```diff
@@ -343,6 +347,10 @@ export class Input implements ComponentInterface {
       ...inheritAriaAttributes(this.el),
       ...inheritAttributes(this.el, ['tabindex', 'title', 'data-form-type']),
     };
+
+    if (this.labelPlacement === undefined) {
+      this.labelPlacement = getIonTheme(this) === 'ionic' ? ionicThemeDefaultLabelPlacement : 'start';
```
thetaPC, line 352: Since the `ionicThemeDefaultLabelPlacement` variable was created to track the value, should another variable to created to track the value `start` for the other themes?

thetaPC (changes requested):

core/src/components/input/input.ionic.scss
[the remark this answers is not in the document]
thetaPC: Hmm, I wonder who we should check in to verify this?

core/src/components/input/input.ionic.outline.scss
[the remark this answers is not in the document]
thetaPC: Works!

core/src/components/input/input.tsx
amandaejohnston: I didn't do this because the `start` default value is only used once, right here. The ionic theme default is also used in the runtime check and printed warning up in `getLabelPlacement()`.

core/src/components/input/input.ionic.scss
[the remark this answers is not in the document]
amandaejohnston: I checked with Sean in Slack. The font used in the screenshots is hardcoded in a separate file and affects all `ionic` theme screenshots, so changing it should be out of scope for this PR. I made a ticket to track this: https://outsystemsrd.atlassian.net/browse/FW-6116

core/src/components/input/input.ionic.outline.scss
amandaejohnston: Fixed in https://github.com/ionic-team/ionic-framework/pull/29268/commits/9b482ae61602c71129b9826bbc3ead86953d3419, screenshots updated in https://github.com/ionic-team/ionic-framework/pull/29268/commits/b39e5bc75946b693c2be0db7ea32dc5f4b268b1f

core/src/components/input/input.ionic.outline.scss
amandaejohnston: Fixed in https://github.com/ionic-team/ionic-framework/pull/29268/commits/8d3bc577de78188b0839b850a6891e40e9a16929

core/src/components/input/input.ionic.outline.scss
amandaejohnston: Fixed in https://github.com/ionic-team/ionic-framework/pull/29268/commits/8d3bc577de78188b0839b850a6891e40e9a16929

core/src/components/input/input.ionic.outline.scss
amandaejohnston: Token added in https://github.com/ionic-team/ionic-framework/pull/29268/commits/8d3bc577de78188b0839b850a6891e40e9a16929. The `--color` variable is only [used on the host](https://github.com/ionic-team/ionic-framework/blob/54247d1e8c19f8af0663873dc7eebfa6295b83dd/core/src/components/input/input.scss#L66), meaning that (1) setting `--color` on the label element doesn't do anything, and (2) setting it on the host instead would also affect the color of the input value.

amandaejohnston:
There are some build differences from `ion-chip` due to an unrelated issue that's being discussed in Slack; re-requesting review anyway since I don't think that's a blocker to at least getting the code looked at.

core/src/components/input/input.ionic.scss
```diff
@@ -1,17 +1,22 @@
 @import "./input";
 @import "./input.ionic.vars";
+@import "./input.ionic.outline.scss";
 
 // Ionic Input
 // --------------------------------------------------
+
 :host {
+  --border-width: 1px;
+  --border-color: #B8B8B8;
```
thetaPC, line 10: Shoot, didn't check these. My bad.

```suggestion
  --border-width: #{$ionic-border-size-small};
  --border-color: #{$ionic-color-neutral-300};
```

thetaPC (changes requested):

PR #29273, opened 2024-04-04
feat(chip): add ionic theme and visual states

BenOsodrac:
Issue number: ROU-4840

---------

## What is the new behavior?

- Added a new Ionic Theme partial, with no dependencies to current `ios/md` architecture. This is to avoid bleeding of the current Design System into the work we are doing for the new Ionic Design System and Theme. For that reason I duplicated part of the basic styles for the Chip (display, box-sizing, etc) and created provisional scss variables, that should be replaced later on with the correct global ones from the new Design System.

- Added styles for info chip type states (default, hover, focus and disabled)

- No new tests were done on this scope, as there're still other Chip tasks being developed that would override any baselines defined now for this theme. Once the initial basic work for the Chip is complete, tests for the Ionic Theme should be implemented, under the task ROU-4837.

- A new page was created - `chip/tests/theme-ionic` - to check styles for current implementation. This will also be used and improved in future Chip Ionic Theme tasks.

- [ ] Yes
- [x] No


core/src/components/chip/chip.ionic.scss
```diff
@@ -0,0 +1,81 @@
...
+
+:host(.chip-disabled) {
+  opacity: 0.6;
+  pointer-events: none;
+}
+
+// Chip: Focus
+// ---------------------------------------------
+
+:host(:focus) {
```
sean-perkins, line 69: Focus management and styling is handled a little differently in Ionic Framework that traditional web experiences. 

We use the `.ion-focused` selector to target elements that are focusable. 

To make an element or component participate in this selector, you will need to add the `ion-focusable` CSS class to the element.

For example:
```diff
<Host
  aria-disabled={this.disabled ? 'true' : null}
  class={createColorClasses(this.color, {
    [theme]: true,
    'chip-outline': this.outline,
    'chip-disabled': this.disabled,
    'ion-activatable': true,
+   'ion-focusable': true
  })}
```

core/src/components/chip/chip.ionic.scss
```diff
@@ -0,0 +1,81 @@
...
+  color: var(--color);
+
+  font-family: $ionic-font-family;
+  font-size: $ionic-font-size-s;
+
+  cursor: pointer;
+
+  overflow: hidden;
+
+  vertical-align: middle;
```
sean-perkins, line 54: I'm unsure if this style has an effect on the visual experience with the `align-items: center;` style above. I think we can remove it.

sean-perkins (changes requested):

core/src/components/chip/chip.ionic.scss
BenOsodrac: Yeah, you probably right, I mindless copied this one from the existing chip.scss. Will remove it 👍

core/src/components/chip/chip.ionic.scss
BenOsodrac: Hum ok I wasn't sure about that actually, I thought about adding that but I saw that on current chip.scss we had this:

```
// Chip: Focus

:host(:focus) {
  outline: none;
}

:host(:focus) {
  --background: #{rgba($text-color-rgb, 0.16)};
}
```

Does that mean is it also wrong there?
What exactly does that `ion-focusable` class behind the scenes?

BenOsodrac: I already found the documentation explaining it 👍

sean-perkins: Spoke with Liam to double check my understanding as well. The existing `chip.scss` implementation is wrong/old/outdated. 

`.ion-focused` is the correct implementation 👍

BenOsodrac: Got it, thanks, already committed the code with `.ion-focused`!

sean-perkins:
Looks like we just need to run `npm run build` and push up those changes.

PR #29275, opened 2024-04-04
perf(styles): compress distributed global stylesheets

sean-perkins:
Issue number: N/A

---------

Ionic Framework developers make extensive use of multi-line comments in the Sass files. However these comments are not stripped from the generated output. 

This leads to a significant increase to the file size of CSS and can have an impact on runtime performance. 

For example: https://cdn.jsdelivr.net/npm/@ionic/core/css/core.css has 183 lines of comment blocks before a single line of CSS.

## What is the new behavior?

- Compresses the generated output from Sass → CSS to strip out multi-line comment blocks. 
- Developers can use [loud comments](https://sass-lang.com/documentation/syntax/comments/) if they want to include comments in the generated output.
- In the case of just the core.css file, the file size goes from 20kb to 10kb.
  - The entire CSS global styles goes from 758 KB to 606kb.

- [ ] Yes
- [x] No

## Other information

Dev-build: `7.8.4-dev.11712251308.1da8aede`

### DIFF
--- a/core/package.json
+++ b/core/package.json
@@ -77,7 +77,7 @@
     "build.docs.json": "stencil build --docs-json dist/docs.json",
     "clean": "node scripts/clean.js",
     "css.minify": "cleancss -O2 -o ./css/ionic.bundle.css ./css/ionic.bundle.css",
-    "css.sass": "sass --embed-sources src/css:./css",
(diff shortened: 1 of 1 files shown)


liamdebeasi (approved):
Thanks for doing this! We might want to make this a `perf` since this will reduce the bundle size for applications.

PR #29286, opened 2024-04-05
feat(checkbox): add states and tests

JoaoFerreira-FrontEnd:
Issue number: N/A

---------

## What is the new behavior?

- Adds the new states to ionic theme checkbox

- [ ] Yes
- [x] No


core/src/components/checkbox/test/states/checkbox.e2e.ts
```diff
@@ -50,5 +50,51 @@ configs({ directions: ['ltr'] }).forEach(({ title, screenshot, config }) => {
       const checkbox = page.locator('ion-checkbox');
       await expect(checkbox).toHaveScreenshot(screenshot(`checkbox-unchecked`));
     });
+
+    test('should render focus checkbox correctly', async ({ page }) => {
+      await page.setContent(
+        `
+          <div id="checkboxes">
```
sean-perkins, line 57: Similar feedback here as: https://github.com/ionic-team/ionic-framework/pull/29289#discussion_r1554065024

Can we add padding so the outline is not clipped in the generated screenshots?

sean-perkins (changes requested):

sean-perkins (approved):

PR #29289, opened 2024-04-05
feat(checkbox): required state

JoaoFerreira-FrontEnd:
Issue number: resolves N/A

---------

## What is the new behavior?

- Adds the required state to ionic theme checkbox

- [ ] Yes
- [x] No


core/src/components/checkbox/test/basic/checkbox.e2e.ts
```diff
@@ -141,4 +141,21 @@ configs({ modes: ['ionic-md'], directions: ['ltr'] }).forEach(({ title, screensh
       await expect(checkboxes).toHaveScreenshot(screenshot(`checkbox-small`));
     });
   });
+
+  test.describe(title('checkbox: basic visual tests'), () => {
+    test('should have a Invalid visual applied correctly', async ({ page }) => {
```
sean-perkins, line 146:
```suggestion
    test('should have an invalid visual applied correctly', async ({ page }) => {
```

core/src/components/checkbox/test/basic/checkbox.e2e.ts
```diff
@@ -141,4 +141,21 @@ configs({ modes: ['ionic-md'], directions: ['ltr'] }).forEach(({ title, screensh
...
+      await page.setContent(
+        `
+        <div id="checkboxes">
```
sean-perkins, line 149: We should add some padding to this container so the captured screenshot for the focus state includes the entire outline. 

For example:
```html
<div id="checkboxes" style="padding: 8px">
```

Would be good to double check the padding value to make sure it is large enough.

core/src/components/checkbox/checkbox.ionic.scss
```diff
@@ -25,4 +27,20 @@
...
+// Ionic Design Checkbox Required
+// --------------------------------------------------
+:host(.ion-invalid) {
+  .checkbox-icon {
+    border-color: #f72c2c;
+  }
+}
+
+// Focused
+:host(.ion-focused.ion-invalid) .native-wrapper {
```
sean-perkins, line 41: These styles build on the changes you have here: https://github.com/ionic-team/ionic-framework/pull/29286/files#diff-796ffc72b6171910c25d5aad085bc245095c38f104d55e253e06282702205066R77-R81.

If we wait for that PR to be reviewed first, this implementation can be simplified to:

```scss
:host(.ion-invalid) {
  --focus-ring-color: #ffafaf;
}
```

core/src/components/checkbox/test/basic/checkbox.e2e.ts
JoaoFerreira-FrontEnd: I did test this, and it should be enough, thanks for the headsup

sean-perkins (approved):

PR #29291, opened 2024-04-05
fix(input) match invalid state to ionic design

thetaPC:
Issue number: internal

---------

The invalid state for the `ionic` theme does not match design:

- error text color is wrong
- border color is wrong

## What is the new behavior?

- The error text color and border color reflects the colors in the design for invalid state.
- Added tests

- [ ] Yes
- [x] No

## Other information

N/A


core/src/components/input/input.ionic.scss
```diff
@@ -8,9 +8,8 @@
 :host {
   --border-width: #{$ionic-border-size-small};
   --border-color: #{$ionic-color-neutral-300};
-
-  // TODO(FW-6113): Verify the ionic design token is correct once it's available and remove the hardcoded value.
-  --highlight-color-invalid: var(--ionic-color-error-600, #970606);
+  --highlight-color-invalid: #{$ionic-color-error-400};
+  --text-color-invalid: #{$ionic-color-error-600};
```
thetaPC, line 12: Decided to add this new variable in order for a dev to be able to customize the text. This would follow the customization structure as md. The md theme allows a  dev to change the text color by setting `--highlight-color-invalid`. The reason why the ionic theme has separate variables is because highlight and text do not share the same colors as md is structured.

core/src/components/input/test/bottom-content/input.e2e.ts
```diff
@@ -105,12 +105,13 @@ configs({ modes: ['md', 'ionic-md'], directions: ['ltr'] }).forEach(({ title, sc
...
         await expect(errorText).toHaveText('my error');
       });
-      test('error text should change when variable is customized', async ({ page }) => {
+      test('error text and highlight should change when variable is customized', async ({ page }) => {
         await page.setContent(
           `
           <style>
             ion-input.custom-input {
               --highlight-color-invalid: purple;
+              --text-color-invalid: purple; // ionic only
```
sean-perkins, line 114: Should we be seeing a screenshot with purple text somewhere? 👀

core/src/components/input/input.scss
```diff
@@ -23,6 +23,7 @@
    * @prop --highlight-color-focused: The color of the highlight on the input when focused
    * @prop --highlight-color-valid: The color of the highlight on the input when valid
    * @prop --highlight-color-invalid: The color of the highlight on the input when invalid
+   * @prop --text-color-invalid: The color of the error text on the input when invalid. Only applies to ionic theme.
```
brandyscarney, line 26: We should create a tech debt ticket to add this to the other themes

core/src/components/input/test/bottom-content/input.e2e.ts
brandyscarney: I was wondering this too 👀

thetaPC: Added: https://github.com/ionic-team/ionic-framework/pull/29291/commits/7baf62ab0f1f89491538238abaefe8b0058cd6de and https://github.com/ionic-team/ionic-framework/pull/29291/commits/ecfc9aaf598806f344c637cab6a50abc24c63aa5

core/src/components/input/input.scss
thetaPC: Ticket created

core/src/components/input/test/bottom-content/input.e2e.ts
```diff
@@ -105,21 +105,22 @@ configs({ modes: ['md', 'ionic-md'], directions: ['ltr'] }).forEach(({ title, sc
...
```
sean-perkins, line 114:
```suggestion
              --text-color-invalid: purple; /* ionic only */
```

It does not appear to have an impact on your test behavior, but the `//` is not the syntax correct comment format for style block in HTML.

CleanShot 2024-04-11 at 00 13 17@2x

sean-perkins (approved):
One comment, non-blocking.

brandyscarney (approved):

PR #29292, opened 2024-04-07
feat(card): Add shape feature to card and style into Ionic theme

bmarcelino-fe:
Issue number: resolves #

---------

## What is the new behavior?

- Adds the new shapes to ionic theme card

- [ ] Yes
- [X] No


core/src/components/card/card.ionic.scss
```diff
@@ -0,0 +1,35 @@
+@use "../../foundations/ionic.vars.scss" as tokens;
+@import "./card";
```
BenOsodrac, line 2: We should not import the partial from ios/md card

core/src/components/card/card.ionic.scss
```diff
@@ -0,0 +1,35 @@
...
+
+// Material Design Card
```
BenOsodrac, line 4: Should be _Ionic Theme Card_

BenOsodrac (changes requested):

core/src/components/card/card.ionic.scss
bmarcelino-fe: Done

core/src/components/card/card.ionic.scss
bmarcelino-fe: Done

BenOsodrac (approved):

core/src/components/card/card.ionic.scss
```diff
@@ -0,0 +1,34 @@
...
+   * @prop --border-radius: Shape of the card
+   */
+  --background: #{tokens.$ionic-color-neutral-0};
+  --border-radius: #{tokens.$ionic-border-radius-rounded-small};
+  display: block;
+
+  padding: #{tokens.$ionic-space-base};
+
+  border: #{tokens.$ionic-border-size-small} solid #{tokens.$ionic-color-neutral-10};
+  border-radius: var(--border-radius);
```
BenOsodrac, line 18: Same as above:

`  @include mixins.border-radius(var(--border-radius));`

BenOsodrac (changes requested):

core/src/components/card/card.ionic.scss
```diff
@@ -0,0 +1,34 @@
...
+:host {
+  /**
+   * @prop --background: Background of the card
+   * @prop --border-radius: Shape of the card
+   */
+  --background: #{tokens.$ionic-color-neutral-0};
+  --border-radius: #{tokens.$ionic-border-radius-rounded-small};
+  display: block;
+
+  padding: #{tokens.$ionic-space-base};
```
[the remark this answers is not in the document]
bmarcelino-fe: Done

core/src/components/card/card.ionic.scss
bmarcelino-fe: Done

BenOsodrac (approved):

joselrio (approved):

core/src/components/card/test/theme-ionic/card.e2e.ts-snapshots/card-round-ionic-md-ltr-light-Mobile-Chrome-linux.png
brandyscarney, line 1: Can we add a background behind the card so we can see the border radius easier on the tests?

bmarcelino-fe: I can add it only for testing purpose, but with that we aren't testing the default styles for card and if for some reason the color of background will change, we can't catch it either.

bmarcelino-fe: As suggestion, we can create a div / element around the card only for testing or we can created a different test only to check the border. I think the first one is better ;)

core/src/components/card/card.tsx
```diff
@@ -82,6 +82,11 @@ export class Card implements ComponentInterface, AnchorInterface, ButtonInterfac
    */
   @Prop() routerAnimation: AnimationBuilder | undefined;
 
+  /**
+   * Set to `"soft"` for a card with more rounded corners.
+   */
+  @Prop({ reflect: true }) shape?: 'soft' | 'round' | 'rectangular';
```
brandyscarney, line 88: Do we need to add the `soft` shape? If soft is the default then we would just need to add `round` and `rectangular`:

Screenshot 2024-04-08 at 11 39 50 AM

core/src/components/card/card.tsx
```diff
@@ -82,6 +82,11 @@ export class Card implements ComponentInterface, AnchorInterface, ButtonInterfac
    */
   @Prop() routerAnimation: AnimationBuilder | undefined;
 
+  /**
+   * Set to `"soft"` for a card with more rounded corners.
```
brandyscarney, line 86:
```suggestion
   * Set to `"round"` for a card with more rounded corners, or `"rectangular"` for a card without rounded corners.
```

core/src/components/card/card.tsx
```diff
@@ -131,12 +136,14 @@ export class Card implements ComponentInterface, AnchorInterface, ButtonInterfac
...
 
   render() {
+    const { shape } = this;
     const theme = getIonTheme(this);
 
     return (
       <Host
         class={createColorClasses(this.color, {
           [theme]: true,
+          [`card-${shape}`]: shape !== undefined,
```
brandyscarney, line 146: We probably want to do something like this to only support the shape in the `"ionic"` theme:

- https://github.com/ionic-team/ionic-framework/blob/7ce115993aa2be4ba70881e5767e4ffd3fedfe21/core/src/components/button/button.tsx#L219-L235
- https://github.com/ionic-team/ionic-framework/blob/7ce115993aa2be4ba70881e5767e4ffd3fedfe21/core/src/components/button/button.tsx#L352
- https://github.com/ionic-team/ionic-framework/blob/7ce115993aa2be4ba70881e5767e4ffd3fedfe21/core/src/components/button/button.tsx#L391

core/src/components/card/test/theme-ionic/card.e2e.ts
brandyscarney, line 1: This folder should be renamed to `shape`, see how button has it: `core/src/components/button/test/shape`

core/src/components/card/test/theme-ionic/card.e2e.ts
```diff
@@ -0,0 +1,70 @@
+import { expect } from '@playwright/test';
+import { configs, test } from '@utils/test/playwright';
+
+configs({ modes: ['ionic-md'], directions: ['ltr'] }).forEach(({ title, screenshot, config }) => {
+  test.describe(title('card: Shapes'), () => {
```
brandyscarney, line 5:
```suggestion
  test.describe(title('card: shape'), () => {
```

core/src/components/card/test/theme-ionic/card.e2e.ts
```diff
@@ -0,0 +1,70 @@
...
+    test('Shape Soft: should not have visual regressions with basic card', async ({ page }) => {
+      await page.setContent(
+        `
+        <ion-card shape="soft">
+          <ion-card-header>
+            <ion-card-subtitle >Card Subtitle</ion-card-subtitle>
+            <ion-card-title >Card Title</ion-card-title>
```
brandyscarney, line 12:
```suggestion
            <ion-card-subtitle>Card Subtitle</ion-card-subtitle>
            <ion-card-title>Card Title</ion-card-title>
```

core/src/components/card/test/theme-ionic/card.e2e.ts
```diff
@@ -0,0 +1,70 @@
...
+      const card = page.locator('ion-card');
+      await expect(card).toHaveScreenshot(screenshot(`card-soft`));
+    });
+    test('Shape Round: should not have visual regressions with basic card', async ({ page }) => {
+      await page.setContent(
+        `
+        <ion-card shape="round">
+          <ion-card-header>
+            <ion-card-subtitle >Card Subtitle</ion-card-subtitle>
+            <ion-card-title >Card Title</ion-card-title>
```
brandyscarney, line 33:
```suggestion
            <ion-card-subtitle>Card Subtitle</ion-card-subtitle>
            <ion-card-title>Card Title</ion-card-title>
```

core/src/components/card/test/theme-ionic/card.e2e.ts
```diff
@@ -0,0 +1,70 @@
+import { expect } from '@playwright/test';
+import { configs, test } from '@utils/test/playwright';
+
+configs({ modes: ['ionic-md'], directions: ['ltr'] }).forEach(({ title, screenshot, config }) => {
```
brandyscarney, line 4:
```suggestion
/**
 * Shape is only available in the Ionic theme.
 */
configs({ modes: ['ionic-md'], directions: ['ltr'] }).forEach(({ title, screenshot, config }) => {
```

brandyscarney (changes requested):

core/src/components/card/test/theme-ionic/card.e2e.ts-snapshots/card-round-ionic-md-ltr-light-Mobile-Chrome-linux.png
brandyscarney: Yeah it is just really hard to see the border and our screenshot tool won't catch any differences if the radius gets removed / changed. We don't have to worry about catching things like the default background for a card, that should be caught by the `basic/` tests.

core/src/components/card/test/theme-ionic/card.e2e.ts
```diff
@@ -0,0 +1,70 @@
...
+      const card = page.locator('ion-card');
+      await expect(card).toHaveScreenshot(screenshot(`card-round`));
+    });
+    test('Shape Rectangular: should not have visual regressions with basic card', async ({ page }) => {
+      await page.setContent(
+        `
+        <ion-card shape="rectangular">
+          <ion-card-header>
+            <ion-card-subtitle >Card Subtitle</ion-card-subtitle>
+            <ion-card-title >Card Title</ion-card-title>
```
brandyscarney, line 54:
```suggestion
            <ion-card-subtitle>Card Subtitle</ion-card-subtitle>
            <ion-card-title>Card Title</ion-card-title>
```

core/src/components/card/test/theme-ionic/card.e2e.ts
bmarcelino-fe: Done ✅

core/src/components/card/test/theme-ionic/card.e2e.ts
bmarcelino-fe: Done ✅

core/src/components/card/test/theme-ionic/card.e2e.ts
bmarcelino-fe: Done ✅

core/src/components/card/test/theme-ionic/card.e2e.ts
bmarcelino-fe: Done ✅

core/src/components/card/card.tsx
bmarcelino-fe: Done ✅

core/src/components/card/card.tsx
bmarcelino-fe: Yes, we can change the behavior of card shape.

Done ✅

brandyscarney:
Did the commits get pushed? The files aren't showing any updates for me.

core/src/components/card/card.tsx
BenOsodrac: And what if in the future we want to change the default for some reason, to rectangular? Or any other value (for some strange reason 😂 ). We would need to change the public API as well. 

In my opinion the shape options and the default shape should be separate things, even if in the present they share the same value.

core/src/components/card/test/theme-ionic/card.e2e.ts
bmarcelino-fe: Done ✅

core/src/components/card/card.tsx
bmarcelino-fe: Done ✅

core/src/components/card/test/theme-ionic/card.e2e.ts-snapshots/card-round-ionic-md-ltr-light-Mobile-Chrome-linux.png
bmarcelino-fe: The tests were reviewed and added a background to identify the border-radius.

bmarcelino-fe:
> Did the commits get pushed? The files aren't showing any updates for me.

All changes are committed.

core/src/components/card/card.tsx
bmarcelino-fe: I have the same opinion and that was the rational of implemented the soft shape. @brandyscarney what is your opinion? Can we revert this to have the 3 shape, even if the default and soft has the same border-radius?

core/src/components/card/card.tsx
```diff
@@ -136,14 +150,14 @@ export class Card implements ComponentInterface, AnchorInterface, ButtonInterfac
...
-    const { shape } = this;
     const theme = getIonTheme(this);
+    const finalShape = this.getShape();
 
     return (
       <Host
         class={createColorClasses(this.color, {
           [theme]: true,
-          [`card-${shape}`]: shape !== undefined,
+          [`card-${finalShape}`]: finalShape !== undefined,
```
BenOsodrac, line 160: Isn't a method a bit overkill? Why not just do:

```
const { shape } = this;
(...)
[`card-${finalShape}`]: theme === 'ionic' && shape !== undefined
```

liamdebeasi:
Removing myself as a reviewer since Brandy is reviewing. Let me know if you'd like a review though!

core/src/components/card/card.tsx
brandyscarney: This is a great point about changing the default in the future introducing changes to the public API. I think we should stick with matching the [button implementation](https://github.com/ionic-team/ionic-framework/blob/7ce115993aa2be4ba70881e5767e4ffd3fedfe21/core/src/components/button/button.tsx#L120) for now in order to get this merged in. Then, I have created a ticket [FW-6118](https://outsystemsrd.atlassian.net/browse/FW-6118) to investigate setting the default for the components that have added a `shape` property (button, card, input, etc). 

We will need to do some investigation here to make sure that we have consistent behavior among components, and also check that it doesn't cause unexpected behaviors for developers trying to override styles. Does that work for now?

core/src/components/card/card.ionic.scss
```diff
@@ -0,0 +1,28 @@
...
+:host {
+  --background: #{tokens.$ionic-color-neutral-0};
+  --border-radius: #{tokens.$ionic-border-radius-rounded-small};
+
+  @include mixins.padding(tokens.$ionic-space-base);
+  @include mixins.border-radius(var(--border-radius));
+
+  display: block;
+
+  border: #{tokens.$ionic-border-size-small} solid #{tokens.$ionic-color-neutral-50};
```
brandyscarney, line 16: I updated this based on the Figma design: https://www.figma.com/file/Fun8j9vlydE65HixgBpA7N/%F0%9F%A5%B7-App-X---UI-Components---WIP?type=design&node-id=20317-62539&mode=design&t=z4J80RlRFDwIBeUO-4

core/src/components/card/test/theme-ionic/index.html
brandyscarney, line 1: Can we delete this test now or are we using it for something else?

brandyscarney, line 1: Can we delete this test now or are we using it for something else?

core/src/components/card/card.tsx
```diff
@@ -97,6 +102,20 @@ export class Card implements ComponentInterface, AnchorInterface, ButtonInterfac
     return this.href !== undefined || this.button;
   }
 
+  /**
+   * Disable the "xsmall" and "xlarge" sizes if the theme is "ios" or "md"
+   */
+  private getShape(): string | undefined {
```
liamdebeasi, line 108:
```suggestion
  private getShape() {
```

TypeScript is able to infer the return type here because `this.shape` is strongly typed. Adding `string | undefined` reduces the effectiveness of type checking because it allows you to compare `this.getShape()` to any string instead of the ones defined on the `shape` property.

core/src/components/card/test/shape/card.e2e.ts
```diff
@@ -0,0 +1,45 @@
...
+/**
+ * Shape is only available in the Ionic theme.
+ */
+configs({ modes: ['ionic-md'], directions: ['ltr'] }).forEach(({ title, screenshot, config }) => {
+  test.describe(title('card: shape'), () => {
+    test.describe('default', () => {
+      test('should not have visual regressions', async ({ page }) => {
+        await page.goto(`/src/components/card/test/shape`, config);
+
+        await page.setIonViewport();
```
liamdebeasi, line 13:
```suggestion
```

Since we are not taking a screenshot of the entire page we do not need this call. Same goes for all other instances of this where we do not take a screenshot of the entire page.

core/src/components/card/test/shape/card.e2e.ts
```diff
@@ -0,0 +1,45 @@
...
+import { configs, test } from '@utils/test/playwright';
+
+/**
+ * Shape is only available in the Ionic theme.
+ */
+configs({ modes: ['ionic-md'], directions: ['ltr'] }).forEach(({ title, screenshot, config }) => {
+  test.describe(title('card: shape'), () => {
+    test.describe('default', () => {
+      test('should not have visual regressions', async ({ page }) => {
+        await page.goto(`/src/components/card/test/shape`, config);
```
liamdebeasi, line 11: Optional: Since we are going to the same page for every test, we can set this goto we can set this in a [beforeEach](https://playwright.dev/docs/api/class-test#test-before-each) inside of the `chip: shape` describe block.

core/src/components/card/test/shape/card.e2e.ts
```diff
@@ -0,0 +1,45 @@
+import { expect } from '@playwright/test';
+import { configs, test } from '@utils/test/playwright';
+
+/**
+ * Shape is only available in the Ionic theme.
```
liamdebeasi, line 5: Can we add a tech debt ticket to update this? If we add the new shape variants to iOS/MD themes then these tests will become relevant.

liamdebeasi (changes requested):

core/src/components/card/test/shape/card.e2e.ts
brandyscarney: Shouldn't this just be updated when the shapes are added to ios/md along with the tests?

brandyscarney: Created [FW-6119](https://outsystemsrd.atlassian.net/browse/FW-6119) for tracking and added a comment referencing it here: https://github.com/ionic-team/ionic-framework/pull/29292/commits/648668b6363c7f7e7b7116c13755b0673a3f6471

core/src/components/card/test/shape/card.e2e.ts
brandyscarney: Removed: https://github.com/ionic-team/ionic-framework/pull/29292/commits/3c81ea03f67b5aa092a96b136829e764246adbd0

core/src/components/card/card.tsx
brandyscarney: Removed function entirely: https://github.com/ionic-team/ionic-framework/pull/29292/commits/3d209fc24546829f16d9e78a52e55acb9d38768a

core/src/components/card/card.tsx
brandyscarney: Removed this function here: https://github.com/ionic-team/ionic-framework/pull/29292/commits/3d209fc24546829f16d9e78a52e55acb9d38768a

liamdebeasi (approved):

core/src/components/card/test/theme-ionic/index.html
bmarcelino-fe: I kept this test to preview and select the shape but we can remove it. The button has the same approach.

bmarcelino-fe: I kept this test to preview and select the shape but we can remove it. The button has the same approach.

bmarcelino-fe: > Can we delete this test now or are we using it for something else?

@brandyscarney I kept this test to preview and select the shape but we can remove it. The button has the same approach. I will merge the PR and if this test doesn't make sense, we can remove it on a new feature implementation.

brandyscarney (approved):
Great work! 🙂

PR #29298, opened 2024-04-08
feat(css): global link classes for standalone and underline

sean-perkins:
Issue number: Internal

---------

N/A

## What is the new behavior?

- Adds two global classes for `.ion-link` (standalone) and `.ion-link-underline` (underline) appearances. 
- The global classes apply when directly applied to an anchor element (`a`) or when used on a parent container that renders a link internally. 

**Usage**

```html
<!-- Basic Usage -->
<a href="#" class="ion-link">Standalone</a>
<a href="#" class="ion-link-underline">Underline</a>

<!-- Nested Usage -->
<div class="ion-link">
  <a href="#">Standalone</a>
</div>
<div class="ion-link-underline">
  <a href="#">Underline</a>
</div>
```

**Focus and Activated States**

Developers should apply the `ion-focusable` and `ion-activatable` classes to the anchor elements to enable proper styling on a mobile device. For web-only usages, the fallback `:focus` and `:active` pseudo states will apply correctly.

```html
<!-- Basic Usage -->
<a href="#" class="ion-link ion-focusable ion-activatable">Standalone</a>
<a href="#" class="ion-link-underline  ion-focusable ion-activatable">Underline</a>
```

### Design Changes

This section is areas of the implementation that are not consistent with the design and why.

1. Font size: Link font sizing is inherited from its content. This is to provide visual consistency when using a link within a paragraph or existing text content. Links should not have an explicit font size, but can be customized by the developer if a specific font size is desired.
2. Color: `currentColor` was used in place of `$ionic-color-neutral-900`. This is to provide better visual consistency when a link is used within content. Text color should be set on the body or text content, which would apply neutral-900 in places where it is applied. 

These changes I need to verify with Design.

- [ ] Yes
- [x] No


core/src/css/test/link/basic/link.e2e.ts-snapshots/link-underline-ionic-md-ltr-light-Mobile-Safari-linux.png
sean-perkins, line 1: I can fix the padding on these screenshots so the focus outline isn't clipped.

core/src/css/test/link/basic/link.e2e.ts-snapshots/link-underline-nested-ionic-md-ltr-light-Mobile-Chrome-linux.png
sean-perkins, line 1: This should be underlined, looking into this.

core/src/css/link.ionic.scss
```diff
@@ -0,0 +1,120 @@
...
+  // Link: Standalone - Hover
+  // -------------------------------------------------------------------------------
+
+  @media (any-hover: hover) {
+    &:hover {
+      text-decoration: underline;
+    }
+  }
+
+  color: $ionic-color-info-400;
```
brandyscarney, line 45: Did design say to use the info color here? The design requirements state:

> 	By default, the link color is the primary color. But developers should be able to change it.

core/src/css/link.ionic.scss
```diff
@@ -0,0 +1,120 @@
...
+
+  &:active,
+  &.ion-activated {
+    color: $ionic-color-info-500;
+
+    text-decoration: underline;
+  }
+}
+
+a.ion-link,
```
brandyscarney, line 71: Do we want this to be prefixed with `ionic` instead of `ion` for now to match `typography`?

core/src/css/link.ionic.scss
```diff
@@ -0,0 +1,120 @@
...
+:not(a).ion-link > a {
```
brandyscarney, line 72: The design document states:

> Devs should be able to apply the classes to any element. (In other words, the styles should not be scoped to specific elements, like <a>.)

Should `a` be removed from the first selector? Or do we want to specify that the classes will only work on `element > a` or `a` elements?

brandyscarney, line 72: These styles will not apply if the anchor is nested twice (or more):

```html
<div class="ion-link">
  <h1><a href="#">Heading link</a></h1>
</div>
```

Is this something we want to support or no?

core/src/css/link.ionic.scss
sean-perkins: Yes, we should be using `.ionic-` here. Updated 👍

core/src/css/link.ionic.scss
sean-perkins: The link global classes should only be used to style a semantic anchor element. The design document was trying to communicate that the styles would be global CSS classes instead of only element-based styles on an anchor element, but worded it in a way that implied additional requirements.

All design requirements/visual styles are specific to an anchor element (visited styling, etc.) that are not applicable to attaching link styling to a non-anchor element. 

Refactoring the styles to only a CSS class would prevent usage at a parent element level, as the link styling would apply to all text content of the parent element and children. It would also require style resets on nested anchor tags to prevent the browser default styles. The current approach requires none of that and only applies to the anchor.

sean-perkins: Updated the selector approach and adjusted the test to render the content in an additional paragraph tag to test that case.

brandyscarney: Makes sense!

core/src/css/link.ionic.scss
sean-perkins: Implementation is based on Figma, which is marked as dev-ready and updated. Design will have a final opportunity to test the style implementation in ODC.

The UX requirements are already inconsistent/incorrect:

image

References the info colors used/in Figma, but the focus outline is wrong. I will bring this up with the larger Design team.

We do the best with the information and accuracy of the information given 🙂

brandyscarney: Sounds good 👍

joselrio (approved):

PR #29302, opened 2024-04-09
feat: add dynamic font scaling for Picker component

Kuldeepdev:
Issue number: Issue number: resolves https://github.com/ionic-team/ionic-framework/issues/24638, resolves https://github.com/ionic-team/ionic-framework/issues/18592

---------

Developers have requested that Ionic Framework support the dynamic type
feature on iOS for accessibility purposes. Ionic applications do not
respond to font scaling on iOS which can create inaccessible
applications particularly for users with low vision for Picker component. Ionic apps on
Android devices currently support the Android equivalent due to
functionality in the Chromium webview.

Developers have also requested a way of adjusting the fonts in their
Ionic Picker component consistently.

## What is the new behavior?

- Ionic components now use `rem` instead of `px` where appropriate. This
means devs can change the font size on `html` and the text in supported
Ionic components will scale up/down appropriately
- Add support for Dynamic Type on iOS (the iOS version of Dynamic Font
Scaling)
-
-
-

- [ ] Yes
- [ x] No

### DIFF
--- a/core/src/components/picker/picker.ios.vars.scss
+++ b/core/src/components/picker/picker.ios.vars.scss
@@ -31,7 +31,7 @@ $picker-ios-button-height:                        $picker-ios-toolbar-height !de
 $picker-ios-button-text-color:                    ion-color(primary, base) !default;
 
 /// @prop - Font size of the picker button
-$picker-ios-button-font-size:                     16px !default;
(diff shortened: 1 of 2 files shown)


liamdebeasi:
Hi there,

Can you please open a feature request with a use case?

PR #29303, opened 2024-04-09
feat(chip): add shape to ionic theme

BenOsodrac:
Issue number: None

---------

## What is the new behavior?

- Added new prop `shape`, only supported on Ionic Theme.
- Added scss for the different shapes.
- Added missing scss for font-weight and line-height.
- Added missing tests for focus state and new tests for shape.
- Added new font weight and line-height design tokens. This change uncovered an issue on the IOnic Typography tests that resulted on snapshots differences.
- Added & updated new snapshot baselines.

- [ ] Yes
- [x] No


core/src/components/chip/test/states/chip.e2e.ts-snapshots/chip-focused-ionic-md-ltr-light-Mobile-Firefox-linux.png
liamdebeasi, line 1: The focus ring is not showing up here

core/src/components/chip/chip.tsx
```diff
@@ -36,14 +36,21 @@ export class Chip implements ComponentInterface {
    */
   @Prop() disabled = false;
 
+  /**
+   * Define the Chip corner shape, when using the Ionic Theme.
+   */
+  @Prop({ reflect: true }) shape?: 'soft' | 'round' | 'rectangular';
```
liamdebeasi, line 42: Can we create a tech debt ticket to make sure we add these features back to the iOS/MD themes?

core/src/components/chip/test/shape/chip.e2e.ts
```diff
@@ -0,0 +1,57 @@
+import { expect } from '@playwright/test';
+import { configs, test } from '@utils/test/playwright';
+
+configs({ modes: ['ionic-md'], directions: ['ltr'] }).forEach(({ title, screenshot, config }) => {
+  /**
+   * This behavior only applies to Ionic Theme.
```
liamdebeasi, line 6: We should reference the tech debt ticket here too. When this gets added back to the iOS/MD themes these tests will apply to those themes as well and will need to be updated.

core/src/components/chip/test/states/chip.e2e.ts
```diff
@@ -41,3 +41,23 @@ configs({ modes: ['ios'], directions: ['ltr'] }).forEach(({ title, screenshot, c
     });
   });
 });
+
+configs({ modes: ['ionic-md'], directions: ['ltr'] }).forEach(({ title, screenshot, config }) => {
+  /**
+   * This behavior only applies to Ionic Theme.
+   */
+  test.describe(title('chip: focus state'), () => {
+    test('should render focus state', async ({ page }) => {
```
liamdebeasi, line 50: Should this be in a separate PR? I'm not sure I understand what this test has to do with verifying `shape` behavior. Also, I can't get the focus state using a keyboard so I'm not sure how this will work.

core/src/components/chip/test/shape/chip.e2e.ts
```diff
@@ -0,0 +1,57 @@
...
+
+        await expect(container).toHaveScreenshot(screenshot(`chip-default`));
+      });
+    });
+
+    test.describe('soft', () => {
+      test('should not have visual regressions', async ({ page }) => {
+        await page.goto(`/src/components/chip/test/shape`, config);
+
+        await page.setIonViewport();
```
liamdebeasi, line 25: Since we are not taking a screenshot of the entire page this is not needed. Same goes for all the other tests.

core/src/components/chip/test/shape/chip.e2e.ts
```diff
@@ -0,0 +1,57 @@
...
+import { configs, test } from '@utils/test/playwright';
+
+configs({ modes: ['ionic-md'], directions: ['ltr'] }).forEach(({ title, screenshot, config }) => {
+  /**
+   * This behavior only applies to Ionic Theme.
+   */
+  test.describe(title('chip: shape'), () => {
+    test.describe('default', () => {
+      test('should not have visual regressions', async ({ page }) => {
+        await page.goto(`/src/components/chip/test/shape`, config);
```
liamdebeasi, line 11: Optional: Since we are going to the same page for every test, we can set this goto we can set this in a [beforeEach](https://playwright.dev/docs/api/class-test#test-before-each) inside of the `chip: shape` describe block.

liamdebeasi (changes requested):

core/src/components/chip/test/states/chip.e2e.ts-snapshots/chip-focused-ionic-md-ltr-light-Mobile-Chrome-linux.png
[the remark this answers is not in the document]
BenOsodrac: Changed, thanks for the heads-up!

core/src/components/chip/test/states/chip.e2e.ts-snapshots/chip-focused-ionic-md-ltr-light-Mobile-Firefox-linux.png
BenOsodrac: Changed!

core/src/components/chip/chip.tsx
BenOsodrac: Sure, the ticket should be created by someone on your team I guess? @sean-perkins

core/src/components/chip/test/shape/chip.e2e.ts
BenOsodrac: Removed!

core/src/components/chip/test/shape/chip.e2e.ts
BenOsodrac: Thanks, didn't knew this, already changed to use it!

core/src/components/chip/test/states/chip.e2e.ts
BenOsodrac: @liamdebeasi About the focus, as the Chip by default has no events or interactions, the focus state is just there prepared in case a developer does it on their side, using also tabindex=0 for that effect.

core/src/components.d.ts
```diff
@@ -750,6 +750,10 @@ export namespace Components {
           * Display an outline style button.
          */
         "outline": boolean;
+        /**
```
bmarcelino-fe, line 753: Based on previous comments on Card task, the default shape of Shape can't be an option. So, to keep consistency, we can remove the round shape as option. A task was already created to investigate,  [FW-6118](https://outsystemsrd.atlassian.net/browse/FW-6118) .

I agree with @BenOsodrac related to have all the options available and supported on the component context.

bmarcelino-fe (approved):

core/src/components.d.ts
BenOsodrac: @brandyscarney should I revert this then?

core/src/components/chip/test/states/chip.e2e.ts
BenOsodrac: Or is there other way for these cases? Ate least, with :focus() or focus-visible() that would be my expectation on how this would work. Does the ion-focusable work in a different way?

core/src/components/chip/chip.tsx
brandyscarney: I created this ticket for you: https://outsystemsrd.atlassian.net/browse/FW-6120

core/src/components/chip/test/shape/chip.e2e.ts
```diff
@@ -0,0 +1,45 @@
+import { expect } from '@playwright/test';
+import { configs, test } from '@utils/test/playwright';
+
+configs({ modes: ['ionic-md'], directions: ['ltr'] }).forEach(({ title, screenshot, config }) => {
+  /**
+   * This behavior only applies to Ionic Theme.
```
brandyscarney, line 6:
```suggestion
   * This behavior only applies to Ionic Theme.
   * TODO(FW-6120): add the `ios` and `md` modes when shape support is added.
```

core/src/components/chip/test/shape/chip.e2e.ts
brandyscarney: I added a code suggestion to get this added 🙂

core/src/components/chip/chip.tsx
BenOsodrac: thank you!!

core/src/components.d.ts
brandyscarney: Yeah, let's remove the default shape for now. I added it to [FW-6118](https://outsystemsrd.atlassian.net/browse/FW-6118) at Bruno's suggestion so we can track adding it there.

brandyscarney: Are we defaulting to `"soft"` for now?

BenOsodrac: Yeah, I double-check with UX & PO and the default should always be `soft`. I had it wrong in the first place.

Already committed the changes about `soft` removal!

brandyscarney: Oh okay, thank you! I updated that ticket to reflect `soft` as the default now. 👍

core/src/components/chip/test/shape/chip.e2e.ts-snapshots/chip-soft-ionic-md-ltr-light-Mobile-Chrome-linux.png
brandyscarney, line 1: You will have to manually delete these `chip-soft-*` images since the screenshot scripts don't delete old ones

core/src/components/chip/test/shape/index.html
```diff
@@ -0,0 +1,56 @@
+<!DOCTYPE html>
+<html lang="en" dir="ltr">
+  <head>
+    <meta charset="UTF-8" />
+    <title>Chip - Ionic States</title>
```
brandyscarney, line 5:
```suggestion
    <title>Chip - Shape</title>
```

brandyscarney (approved):
Looks good to me! 🙂

core/src/components/chip/chip.tsx
```diff
@@ -36,14 +36,22 @@ export class Chip implements ComponentInterface {
    */
   @Prop() disabled = false;
 
+  /**
+   * Define the Chip corner shape, when using the Ionic Theme.
+   */
+  @Prop({ reflect: true }) shape?: 'round' | 'rectangular';
```
sean-perkins, line 42:
```suggestion
  @Prop() shape?: 'round' | 'rectangular';
```

The `reflect` option will set the attribute on the HTML element. Often used for styling concerns or if you have JS logic that queries based on an attribute. 

Since we are using `.chip-round`, etc. for styling, we don't need to reflect here from what I can tell.

core/src/components/chip/test/states/chip.e2e.ts
```diff
@@ -41,3 +41,25 @@ configs({ modes: ['ios'], directions: ['ltr'] }).forEach(({ title, screenshot, c
...
+      await page.setContent(
+        `<ion-content class="ion-padding">
+          <ion-chip class="ion-focused">
+            <ion-label>Focused</ion-label>
+          </ion-chip>
+        </ion-content>`,
+        config
+      );
+
+      const chip = page.locator('ion-content');
```
sean-perkins, line 60: This is currently targeting `ion-content` which captures a large screenshot of the entire page instead of the focused element you are wanting to verify. 

You can see that here: `core/src/components/chip/test/states/chip.e2e.ts-snapshots/chip-focused-ionic-md-ltr-light-Mobile-Chrome-linux.png`. 

From what I can gather from the previous feedback and changes here, I think you _might_ want to have `ion-padding` on a `div` around the chip and use that as a locator? This will capture a much smaller screenshot of just the chip and focus ring. Smaller screenshots = faster tests and image comparisons. 

This change will require re-generating screenshots for this test.

sean-perkins (changes requested):

sean-perkins (approved):

core/src/components/chip/test/states/chip.e2e.ts
sean-perkins: Addressed this and the above comment here: https://github.com/ionic-team/ionic-framework/pull/29303/commits/53cd9f4279e91a10a823ddcc84354acaf50a850c. 

Figured it'd be more exciting to start your day with merging an approved PR. Let me know if you have questions on the feedback/changes.

BenOsodrac: Thank you, more exciting indeed! :) 

Yeah, I went a bit overkill with the ion-content, your suggestion is better!

joselrio (approved):

core/src/components/chip/test/states/chip.e2e.ts
liamdebeasi: `ion-focusable` works similarly to `:focus-visible`, so the element needs to be eligible to receive a focus event for `ion-focused` to be applied.

Sounds good re: adding the tests.

brandyscarney (approved):

core/src/components/chip/test/states/chip.e2e.ts
BenOsodrac: Ok, that was my expectation 👍. Then on the developer side they can add the tabindex=0 if they need it (it will be the case on the ODC Widget, where we have a Click event).

liamdebeasi (approved):

PR #29307, opened 2024-04-09
feat(checkbox): target area

JoaoFerreira-FrontEnd:
Issue number: N/A

---------

## What is the new behavior?
- The target area should be 48px minimum independent from the checkbox size;
- Playwright tests to check clickable height
-

- [ ] Yes
- [x] No


core/src/components/checkbox/test/basic/checkbox.e2e.ts
```diff
@@ -141,4 +141,40 @@ configs({ modes: ['ionic-md'], directions: ['ltr'] }).forEach(({ title, screensh
       await expect(checkboxes).toHaveScreenshot(screenshot(`checkbox-small`));
     });
   });
+
+  test.describe(title('checkbox: basic visual tests'), () => {
+    test('check target area of a small size checkbox', async ({ page }) => {
+      await page.setContent(
+        `
+        <div id="checkboxes">
```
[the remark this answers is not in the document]
JoaoFerreira-FrontEnd: it makes sense, I'll clean it

core/src/components/checkbox/checkbox.ionic.vars.scss
```diff
@@ -23,6 +23,12 @@ $checkbox-ionic-border-radius: calc(var(--size) / 4 - 2px) !default;
 /// @prop - Icon size of the checkbox for the small size
 $checkbox-ionic-small-size: 16px !default;
 
+/// @prop - Padding top of the checkbox for the default size
+$checkbox-ionic-padding-top: 12px !default;
+
+/// @prop - Padding bottom of the button for the default size
```
sean-perkins, line 29: This comment refers to `button` instead of `checkbox`.

core/src/components/checkbox/checkbox.ionic.vars.scss
```diff
@@ -23,6 +23,12 @@ $checkbox-ionic-border-radius: calc(var(--size) / 4 - 2px) !default;
 /// @prop - Icon size of the checkbox for the small size
 $checkbox-ionic-small-size: 16px !default;
 
+/// @prop - Padding top of the checkbox for the default size
+$checkbox-ionic-padding-top: 12px !default;
```
sean-perkins, line 27: Can you share why the ionic theme implementation has padding on the checkbox? This isn't consistent with iOS or MD implementations. 

Checkboxes with this high of padding I would expect are rendered in a list item:
```
<ion-item>
  <ion-checkbox></ion-checkbox>
</ion-item>
```

The `ion-item` would add the respective padding.

JoaoFerreira-FrontEnd: image

I had this padding to ensure that the target area is available for every checkbox.
In the example above I remove the padding and add a background color to the second checkbox :after. 

That's why I add the padding

sean-perkins: With the padding being added for the target area, I would not expect this diff:

CleanShot 2024-04-10 at 11 54 08

With the larger checkbox size, it already had the correct minimum target area and should not apply additional padding. 

The target area should operate similar to a "minimum height" and "minimum width" where it will add padding to ensure the checkbox is at least 48px by 48px, but doesn't need to apply past that. 

For example this diff is where you would expect the padding to apply to ensure the target area:

CleanShot 2024-04-10 at 11 59 06

If I have some additional time today I can offer some implementation ideas on how to achieve this. Otherwise if you have ideas on how to alter the implementation to achieve this behavior that works too 👍

sean-perkins (approved):

PR #29313, opened 2024-04-09
feat(input): add hover state to the ionic theme

brandyscarney:
Issue number: internal

---------

The ionic input does not have a hover state.

## What is the new behavior?
Adds a hover state. 

- [ ] Yes
- [x] No

## Other information
We do not currently have a way to test hover states. See https://github.com/microsoft/playwright/issues/12077 for more information.

### DIFF
--- a/core/src/components/input/input.ionic.scss
+++ b/core/src/components/input/input.ionic.scss
@@ -1,3 +1,4 @@
+@use "../../foundations/ionic.vars.scss" as tokens;
 @import "./input";
 @import "./input.ionic.vars";
 @import "./input.ionic.outline.scss";
@@ -24,8 +25,7 @@
 // ----------------------------------------------------------------
 
 .input-bottom .helper-text {
-  // TODO(FW-6112): Verify the design token is correct once it's available and remove the hardcoded value.
-  color: var(--ionic-color-neutral-600, #535353);
+  color: #{tokens.$ionic-color-neutral-600};
 }
 
 .input-bottom .error-text {
@@ -37,7 +37,6 @@
 
 @media (any-hover: hover) {
   :host(:hover) {
-    // TODO(FW-6112): Verify the design token is correct once it's available and remove the hardcoded value.
-    --border-color: var(--ionic-color-neutral-600, #535353);
+    --border-color: #{tokens.$ionic-color-neutral-600};
   }
 }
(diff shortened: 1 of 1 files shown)


core/src/components/input/input.ionic.scss
```diff
@@ -31,3 +31,13 @@
 .input-bottom .error-text {
   color: var(--highlight-color-invalid);
 }
+
+// Input Hover
+// ----------------------------------------------------------------
+
+@media (any-hover: hover) {
+  :host(:hover) {
+    // TODO(FW-6112): Verify the design token is correct once it's available and remove the hardcoded value.
```
sean-perkins, line 40: Design tokens are available in the parent branch `ROU-4848`. Want to take care of it now or handle that later?

brandyscarney: Updated in https://github.com/ionic-team/ionic-framework/pull/29313/commits/aeebd8665c5742d59e62c626829ed31e41463d92

core/src/components/input/input.ionic.scss
```diff
@@ -24,10 +25,18 @@
 // ----------------------------------------------------------------
 
 .input-bottom .helper-text {
-  // TODO(FW-6112): Verify the design token is correct once it's available and remove the hardcoded value.
-  color: var(--ionic-color-neutral-600, #535353);
+  color: #{tokens.$ionic-color-neutral-600};
```
brandyscarney, line 28: @thetaPC I updated the TODO for FW-6112 here to use the design token, is that fine? I checked and it is the same value:

```
$ionic-color-neutral-600: var(--ionic-color-neutral-600, #535353);
```

sean-perkins, line 28:
```suggestion
  color: tokens.$ionic-color-neutral-600;
```

Optional, but we don't need to escape the Sass variables when used on a style assignment.  It will compile to:

```
color: var(--ionic-color-neutral-600, #535353);
```

Pretty sure we still need to escape when binding to a CSS variable (like you are doing below).

sean-perkins (approved):

core/src/components/input/input.ionic.scss
brandyscarney: Oh yeah, good catch. I just copy and pasted it 😂

thetaPC: That works for me. I also have it updated on my success state PR. Let's see who gets merged first? 😆

PR #29323, opened 2024-04-10
fix(input): add disabled state to the ionic theme

thetaPC:
Issue number: internal
---------

The disabled state for the `ionic` theme does not match design.

## What is the new behavior?

- Matched `ion-input` to the design for disabled: background, border, bottom, etc.
- valid/invalid/color will have a slightly different design: apply 60% opacity to border, help message, and use current disabled colors for label, text, and counter text
- Added tests

- [ ] Yes
- [x] No

## Other information

DM if you need the link for the design.

[Preview](https://ionic-framework-git-fw-6093-ionic1.vercel.app/src/components/input/test/states?ionic:theme=ionic)


core/src/components/input/test/states/input.e2e.ts-snapshots/input-disabled-no-fill-ionic-md-ltr-light-Mobile-Chrome-linux.png
brandyscarney, line 1: Shouldn't the helper text for the warning input be `#6B430C`?

core/src/components/input/input.ionic.scss
```diff
@@ -54,3 +54,39 @@
...
   }
 }
+
+// Input - Disabled
+// ----------------------------------------------------------------
+
+:host(.input-disabled) {
+  // color for the text within the input
+  --color: #{tokens.$ionic-color-neutral-400};
+  --background: #{tokens.$ionic-color-neutral-10};
```
brandyscarney, line 64:
```suggestion
  --background: #{tokens.$ionic-color-neutral-10};
  
  pointer-events: none;
```

This needs to be added to prevent the hover state on disabled inputs

brandyscarney (commented):
Looks good, just a couple things!

core/src/components/input/input.ionic.outline.scss
```diff
@@ -119,3 +119,19 @@
...
+
+:host(.input-fill-outline.input-disabled) {
+  // color for the text within the input
+  --color: #{$ionic-color-neutral-400};
+  --background: transparent;
+}
+
+:host(.input-fill-outline.input-disabled) .native-wrapper {
+  --background: #{$ionic-color-neutral-10};
+  border-radius: var(--border-radius);
```
brandyscarney, line 134: Should this rule just be applied to `.native-wrapper` or `.input-wrapper` like [`input.scss`](https://github.com/ionic-team/ionic-framework/blob/62a3e115ef87cc418df7a93e059a5d1964c24620/core/src/components/input/input.scss#L233)?

core/src/components/input/test/states/input.e2e.ts-snapshots/input-disabled-no-fill-ionic-md-ltr-light-Mobile-Chrome-linux.png
thetaPC: Yes, but I would consider the color to be added in the warning implementation ticket. The disabled work already sets the opacity to 60%. Once the warning implementation ticket is worked on, then the helper message will have the correct color with the already ready 60% opacity.

core/src/components/input/input.ionic.outline.scss
thetaPC: No, because the `input-wrapper` also includes the label. If we add it, then the styles will also show within the label area like this:

Screenshot 2024-04-15 at 11 50 46 AM

Outline only has the background and border within the native input.

brandyscarney: Oh I see. Can we use `border-radius: inherit` instead?

core/src/components/input/test/states/input.e2e.ts-snapshots/input-disabled-no-fill-ionic-md-ltr-light-Mobile-Chrome-linux.png
brandyscarney: Ah okay. Maybe we should add a note on that ticket to make sure it works?

thetaPC: Good idea! I added it under context for the color ticket, FW-6095.

core/src/components/input/input.ionic.outline.scss
thetaPC: Worked: https://github.com/ionic-team/ionic-framework/pull/29323/commits/d1f00990ae38b415e416f75fe9a635a15003104f

brandyscarney (approved):

PR #29327, opened 2024-04-11
Fw 5818

liamdebeasi:
Issue number: resolves #

---------

## What is the new behavior?

-
-
-

- [ ] Yes
- [ ] No


sean-perkins:
Observing a shift to both the button and the large title when popping the view. Doesn't happen consistently, but seems to be related to adjusting the text size. 

Could be a Safari quirk if you aren't able to reliable reproduce. I can get Safari's header UI to break/run into the content, so there could be more factors here. 

https://github.com/ionic-team/ionic-framework/assets/13732623/08c30136-8e8a-46a4-97d0-9af1d7ccfb47

liamdebeasi:
If you set `text=" "` on the BackButton (note the space) based off `main`, does the issue reproduce?

thetaPC (approved):
LGTM, worked on the original repo and on a simulator

sean-perkins (approved):
Tested against the latest changes of v7 with `text=" "` and was unable to reproduce the video I attached earlier. 

Tested again with the dev-build and was only able to get in that state once with changing the dynamic font scaling, but was unable to reproduce with the same steps consistently. Navigated 40+ times without the issue presenting itself with a consistent dynamic font scale.

I think these changes are good to merge. There is a rendering quirk in Safari's UI with changing the dynamic font scale with the address bar at the top of the screen that is having other impact (at random, but not very common). If we get issue reports we can do a deeper evaluation of what is going on there.

PR #29337, opened 2024-04-12
feat(input): add styles for clear action to ionic theme

amandaejohnston:
Issue number: Internal

---------

## What is the new behavior?

For the `ionic` theme: Adjusted the size of the clear button to match design specs, and updated the behavior so the button additionally only shows if the native input *or* any other actions within the `ion-input` are focused. (See code comments for details.)

The actual icon used has not been modified. The ticket says to "use the material design clear icon as the reference icon for this feature," and we've already added a feature allowing the icon to be customized at the app level. Let me know if the intention was to use a different default icon with the `ionic` theme.

- [ ] Yes
- [x] No


amandaejohnston:
The input UX requirements were updated to change all references to the Active state to just be when the input is focused. Updated the PR description to remove extra notes about what Active means.

thetaPC:
@amandaejohnston please update the description with the new logic for the visibility of the clear icon.

amandaejohnston:
> @amandaejohnston please update the description with the new logic for the visibility of the clear icon.

The logic used here hasn't actually changed (it was always equivalent to the input being focused), so no updates needed on that front 👍

core/src/components/input/input.ionic.scss
```diff
@@ -21,6 +21,35 @@
   min-height: 48px;
 }
 
+// Input Clear Button
+// ----------------------------------------------------------------
+
+.input-clear-icon {
+  width: 16px;
```
thetaPC, line 28: Please update FW-6100 to make sure to update the width when input size is `xlarge`. That way it doesn't slip through the cracks.

core/src/components/input/input.ionic.scss
thetaPC, line 1: The clear icon needs to have an outline when the user has focused on it like this:
image

This can also be seen in the UX reqs.

core/src/components/input/input.ionic.scss
amandaejohnston: It looks like the button is 16px across all input sizes? At least according to the Figma. Screenshot from xlarge size:
image

thetaPC (changes requested):

core/src/components/input/input.ionic.scss
thetaPC: Thank you! I re-read the UX reqs and realized that I misread it. :face_palm:

core/src/components/input/input.ionic.scss
amandaejohnston: Styles added in https://github.com/ionic-team/ionic-framework/pull/29337/commits/0e157bb7799dbd5e2415dffcff1acc29617e0611. The screenshot job is still picking up the old styles (half opacity, no blue ring); I'll troubleshoot in the morning in case it's just Git being slow.

core/src/components/input/test/basic/input.e2e.ts
```diff
@@ -131,3 +131,82 @@ configs({ modes: ['ios'], directions: ['ltr'] }).forEach(({ title, screenshot, c
...
+        config
+      );
+
+      const input = page.locator('ion-input');
+
+      await input.evaluate((el: HTMLIonInputElement) => el.setFocus());
+      await page.waitForChanges();
+
+      await page.keyboard.press('Tab');
+      await page.waitForChanges();
```
thetaPC, line 176: The tests aren't capturing the focus state because `ion-focused` is never being added. It seems that we have to manually add that class based on other examples that depend on that class.

```suggestion
      const clearButton = input.locator('.input-clear-icon');
      clearButton.evaluate((el: HTMLElement) => el.classList.add('ion-focused'));
      await page.waitForChanges();
```

thetaPC (changes requested):

core/src/components/input/input.ionic.scss
thetaPC: I looked into this briefly. [Finding](https://github.com/ionic-team/ionic-framework/pull/29337#discussion_r1566524482) is in the latest review.

core/src/components/input/test/basic/input.e2e.ts
amandaejohnston: Oh, thank you! That makes sense; the focus utility only works with `ion-app`. Saved me some stress this morning 👀

core/src/components/input/input.ionic.scss
amandaejohnston: Thank you! Correct screenshots added in https://github.com/ionic-team/ionic-framework/pull/29337/commits/5b90a08c015151fd57d0afb6ba5e63eb8e1351e9.

core/src/components/input/input.ionic.scss
```diff
@@ -21,6 +21,44 @@
...
+  height: 16px;
+
+  color: #{$ionic-color-neutral-500};
+}
+
+.input-clear-icon.ion-focused {
+  border-radius: #{$ionic-border-radius-rounded-small};
+
+  outline: 2px solid #{$ionic-color-primary-100};
```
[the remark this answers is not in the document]
amandaejohnston: Fixed in https://github.com/ionic-team/ionic-framework/pull/29337/commits/443d7023ac3204b6c4b11b853a4fbe2767906771.

core/src/components/input/test/basic/input.e2e.ts
amandaejohnston: Fixed in https://github.com/ionic-team/ionic-framework/pull/29337/commits/2e2ef4d77573c4a0a780d99297e0582e40050fed and https://github.com/ionic-team/ionic-framework/pull/29337/commits/f0a1f15fe4f594184a53950f80eff683e21adc0e 👍

thetaPC (approved):
LGTM, minor nitpick that I leave up to you to add: add padding to the tests so the focus ring doesn't get cut off.

PR #29342, opened 2024-04-15
fix(input): add focused state for ionic theme

thetaPC:
Issue number: internal
---------

The focused state for the `ionic` theme does not match design.

## What is the new behavior?

- Matched `ion-input` to the design for focused: border color and border width
- Added tests

- [ ] Yes
- [x] No

## Other information

DM if you need the link for the design.

[Preview](https://ionic-framework-git-fw-6083-ionic1.vercel.app/src/components/input/test/states?ionic:theme=ionic)


core/src/components/input/input.ionic.outline.scss
```diff
@@ -8,8 +8,6 @@
   --border-radius: #{$ionic-border-radius-rounded-small};
   --padding-start: 12px;
   --padding-end: 12px;
-  --placeholder-color: #{$ionic-color-neutral-600};
-  --placeholder-opacity: 1;
```
thetaPC, line 12: These were moved to `input.ionic.scss` since both fills use the same styles.

core/src/components/input/input.ionic.scss
```diff
@@ -11,6 +11,8 @@
   --border-color: #{$ionic-color-neutral-300};
   --highlight-color-valid: #{$ionic-color-success-400};
   --highlight-color-invalid: #{$ionic-color-error-400};
+  --placeholder-color: #{$ionic-color-neutral-500};
```
thetaPC, line 14: Updated the value to match Figma.

core/src/components/input/test/fill/input.e2e.ts-snapshots/input-fill-outline-label-stacked-ionic-md-ltr-light-Mobile-Chrome-linux.png
thetaPC, line 1: I noticed that the placeholder color value was not matching the color on Figma. These snapshots were generated with the updated color.

core/src/components/input/test/states/input.e2e.ts
```diff
@@ -142,5 +142,38 @@ configs({ modes: ['ionic-md'], directions: ['ltr'] }).forEach(({ title, screensh
         });
       });
     });
+
+    test.describe(title('focused'), () => {
+      test('should render disabled input correctly', async ({ page }) => {
```
amandaejohnston, line 147:
```suggestion
      test('should render focused input correctly', async ({ page }) => {
```

amandaejohnston (changes requested):

core/src/components/input/input.ionic.scss
thetaPC: I'm waiting on confirmation on colors but it seems like 500 is the correct color. I vote for leaving it as is. We can create a new PR if it ends up being something else.

amandaejohnston: Works for me 👍

amandaejohnston (approved):

core/src/components/input/input.ionic.scss
thetaPC: Confirmed that all fills will have the placeholder text with 500 so no changes are needed from us! Woot!

PR #29345, opened 2024-04-16
feat(ionic-theme): improve scss architecture for ionic theme.

BenOsodrac:
Issue number: None

---------

## What is the new behavior?

This PR follows the architectural changes defined on [this Design Doc](https://outsystemsrd.atlassian.net/wiki/spaces/RDMBVS/pages/3991109633/SCSS+Architecture). Due to some of the developments done in the meantime, not everything that was defined on that document was followed and some sections were simplified. 

Overall, there were two guidelines that supported the work on this PR:
- Have no impact on the current scss partials & CSS architecture for the `md/iOS` themes.
- Make everything related to the new `Ionic` Theme separated.

Based on these, here're the changes made:

**On Themes folder (private):**
- Renamed current internal scss partials on src/themes, related to md/iOS, to native.*.scss and move them to a new folder, called `native`. Updated imports on all framework.
- Remove the ionic prefix from the name of the mixins and function partials on src/themes. Updated imports on all framework.
- Created new folder, named ionic, inside src/themes. This holds the `ionic.globals.scss`, with all the mixins and functions forwarded, to be used on other scopes.
- Replaced on already created Ionic theme files, the usage of tokens and mixins with @use, to instead _@use "../../themes/ionic/ionic.globals.scss" as globals;_. This ensures an equal approach is followed everywhere and also makes it easier to change the files imported or paths, in the future, as its all in the same global file.
- Updated the foundations readMe file, with the new process for using globals.

themes

**On css folder (public):**
- Created new folder, named `ionic`, inside src/css. This holds all the files related exclusively to Ionic Theme, following the same structure as exists now for md/iOS, with a core, a bundle, utils, etc. Some files were a bit duplicated, to eliminate imports from ios or md theme partials. The only file common to both Themes is the `normalize.scss`. 
- No folder structure or renamings were done on the existing output, to prevent breaking-changes for developers already making imports from this folder.
- Updated typography and link partials to use globals instead of tokens.
- Changed `font-size` styles on typography to be on `body`, instead of `html`, to enable correct support of accessibility features on browsers and devices, related to `font-size`.

css

- [ ] Yes
- [ ] No


bmarcelino-fe (approved):

sean-perkins:
@BenOsodrac I believe we need to update this script:

https://github.com/ionic-team/ionic-framework/blob/ea1207174d0bbe95abd48a16554a07326ad7567f/core/package.json#L91

to include the `scss` extension. 

This will remove any variance between developers moving forward (hopefully).

joselrio (approved):

BenOsodrac:
> @BenOsodrac I believe we need to update this script:
> 
> https://github.com/ionic-team/ionic-framework/blob/ea1207174d0bbe95abd48a16554a07326ad7567f/core/package.json#L91
> 
> to include the `scss` extension.
> 
> This will remove any variance between developers moving forward (hopefully).

Good point, I committed that change, because it makes total sense, but still not 100% sure will be enough, need to get more context about how prettier actually works :)

sean-perkins (approved):

PR #29359, opened 2024-04-18
feat(tabs): add ionic theme for tab bar and tab button

brandyscarney:
Issue number: internal

---------

Tab bar & tab button use the `md` styles for the `ionic` theme

## What is the new behavior?
Adds the following styles for the `ionic` theme:
- Rectangular, fixed tab bar
- Box shadow with elevation z2
- Selected, activated, and focused states

- [ ] Yes
- [x] No

## Other information
- A ticket (FW-6186) has been created to track updating the hardcoded hex values to design tokens
- A ticket (FW-6189) has been created to add badge styles


core/src/components/tab-bar/tab-bar.ionic.vars.scss
brandyscarney, line 1: These variables should go into a top level theme file like this: https://github.com/ionic-team/ionic-framework/blob/ea1207174d0bbe95abd48a16554a07326ad7567f/core/src/themes/ionic.theme.default.ios.scss#L14-L18

sean-perkins (approved):
Changes look good 👍 

The new ionic theme tab bar styles do not support color tab bars with the activated/focused states well, but I would consider that out of scope for now.

PR #29371, opened 2024-04-22
Fw 1891

liamdebeasi:
Issue number: resolves #

---------

## What is the new behavior?

-
-
-

- [ ] Yes
- [ ] No

### DIFF
--- a/core/src/components/picker-column/picker-column.tsx
+++ b/core/src/components/picker-column/picker-column.tsx
@@ -601,7 +601,6 @@ export class PickerColumn implements ComponentInterface {
         <div
           aria-hidden="true"
           class="picker-opts"
-          tabindex={disabled ? undefined : 0}
           ref={(el) => {
             this.scrollEl = el;
           }}
--- a/core/src/components/picker-column/picker-column.tsx
+++ b/core/src/components/picker-column/picker-column.tsx
@@ -502,7 +502,7 @@ export class PickerColumn implements ComponentInterface {
    */
   private findNextOption = (stride: number = 1) => {
     const { activeItem } = this;
-    if (!activeItem) return undefined;
+    if (!activeItem) return null;
 
     let prevNode = activeItem;
     let node = activeItem.nextElementSibling as HTMLIonPickerColumnOptionElement | null;
@@ -533,7 +533,7 @@ export class PickerColumn implements ComponentInterface {
    */
   private findPreviousOption = (stride: number = 1) => {
     const { activeItem } = this;
-    if (!activeItem) return undefined;
+    if (!activeItem) return null;
 
     let nextNode = activeItem;
     let node = activeItem.previousElementSibling as HTMLIonPickerColumnOptionElement | null;
@@ -590,7 +590,7 @@ export class PickerColumn implements ComponentInterface {
         break;
     }
 
-    if (newOption != null) {
+    if (newOption !== null) {
       this.value = newOption.value;
 
       // This stops any default browser behavior such as scrolling
--- a/core/src/components/picker-column/picker-column.tsx
+++ b/core/src/components/picker-column/picker-column.tsx
@@ -571,10 +571,20 @@ export class PickerColumn implements ComponentInterface {
         newOption = this.findNextOption(5);
         break;
       case 'Home':
+
+        /**
+         * There is no guarantee that the first child will be an ion-picker-column-option,
+         * so we do not use firstElementChild.
+         */
         options = this.el.querySelectorAll<HTMLIonPickerColumnOptionElement>('ion-picker-column-option');
         newOption = options[0];
         break;
       case 'End':
+
+        /**
+         * There is no guarantee that the last child will be an ion-picker-column-option,
+         * so we do not use lastElementChild.
+         */
         options = this.el.querySelectorAll<HTMLIonPickerColumnOptionElement>('ion-picker-column-option');
         newOption = options[options.length - 1];
         break;
--- a/core/src/components/picker-column/picker-column.tsx
+++ b/core/src/components/picker-column/picker-column.tsx
@@ -556,7 +556,7 @@ export class PickerColumn implements ComponentInterface {
   };
 
   private onKeyDown = (ev: KeyboardEvent) => {
-    let options, newOption;
+    let newOption;
     switch (ev.key) {
       case 'ArrowDown':
         newOption = this.findNextOption();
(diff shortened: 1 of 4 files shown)


core/src/components/picker-column/picker-column.tsx
```diff
@@ -465,22 +478,161 @@ export class PickerColumn implements ComponentInterface {
...
+    let newOption = null;
+    switch (ev.key) {
+      case 'ArrowDown':
+        newOption = this.findNextOption();
+        break;
+      case 'ArrowUp':
+        newOption = this.findPreviousOption();
+        break;
+      case 'PageUp':
+        newOption = this.findPreviousOption(5);
```
liamdebeasi, line 579: A stride of 5 was based on how browsers handle PageUp/PageDown in scroll snapping containers

core/src/components/picker-column/picker-column.tsx
```diff
@@ -465,22 +478,161 @@ export class PickerColumn implements ComponentInterface {
     this.el.classList.remove('picker-column-active');
   };
 
-  get activeItem() {
+  get activeItem(): {
+    activeItem?: HTMLIonPickerColumnOptionElement;
+    activeItemIndex: number;
+  } {
```
liamdebeasi, line 484: This refactor isn't 100% necessary to ship this fix, so let me know if you'd like me to split this out. In order to correctly set `aria-valuenow` I need to now where the active options appears in a list of all options (its index) which requires an iteration. Rather that do 2 iterations, I decided to update `activeItem` to return all the information I need from a single iteration.

core/src/components/picker-column/picker-column.tsx
```diff
@@ -465,22 +478,161 @@ export class PickerColumn implements ComponentInterface {
...
+
+    for (let i = 0; i <= options.length - 1; i++) {
+      const option = options[i];
+
       /**
        * If the whole picker column is disabled, the current value should appear active
        * If the current value item is specifically disabled, it should not appear active
        */
-      if (!this.disabled && option.disabled) {
-        return false;
```
liamdebeasi, line 477: My understanding is this if statement had more checks than it needed to. According to the comment, the picker column's disabled state has no bearing on whether or not an option should be rendered as active, so we don't need to check it.

core/src/components/picker-column/picker-column.tsx
```diff
@@ -206,6 +215,10 @@ export class PickerColumn implements ComponentInterface {
     }
   }
 
+  connectedCallback() {
+    this.ariaLabel = this.el.getAttribute('aria-label') ?? 'Select a value';
```
liamdebeasi, line 219: I could be convinced to not have a default value here since the label isn't super helpful. However, I was concerned about breaking the existing experience since not using an `aria-label` on the slider role could cause tools such as Axe to start failing.

core/src/components/datetime/datetime.tsx
```diff
@@ -2045,6 +2051,7 @@ export class Datetime implements ComponentInterface {
 
     return (
       <ion-picker-column
+        aria-label="Select a day period"
```
liamdebeasi, line 2054: Open to other ideas for naming here. Wasn't sure what else to call the AM/PM picker

core/src/components/picker-column/picker-column.tsx
```diff
@@ -481,6 +511,150 @@ export class PickerColumn implements ComponentInterface {
...
+    return (
+      <div
+        ref={(el) => (this.assistiveFocusable = el)}
+        class="assistive-focusable"
+        role="slider"
+        tabindex={this.disabled ? undefined : 0}
+        aria-label={this.ariaLabel}
+        aria-valuemin={0}
+        aria-valuemax={0}
+        aria-valuenow={0}
```
liamdebeasi, line 650: I tested VoiceOver, TalkBack, and NVDA. None of them announce these values if `valuetext` is set. Setting these all to zero also avoids the WebKit bug noted.

core/src/components/picker-column/picker-column.tsx
```diff
@@ -481,6 +511,159 @@ export class PickerColumn implements ComponentInterface {
...
+    /**
+     * The below operations should be inverted when running on a mobile device.
+     * For example, swiping up will dispatch an "ArrowUp" event. On desktop,
+     * this should cause the previous option to be selected. On mobile, swiping
+     * up causes a view to scroll down. As a result, swiping up on mobile should
+     * cause the next option to be selected. The Home/End operations remain
+     * unchanged because those always represent the first/last options, respectively.
+     */
+    const mobile = isPlatform('mobile');
+    let newOption = null;
```
sean-perkins, line 587:
```suggestion
    let newOption: HTMLIonPickerColumnOptionElement | null = null;
```

The current type of this variable is `any`.

core/src/components/picker-column/picker-column.tsx
```diff
@@ -481,6 +511,159 @@ export class PickerColumn implements ComponentInterface {
...
 
+  /**
+   * Find the next enabled option after the active option.
+   * @param stride - How many options to "jump" over in order to select the next option.
+   * This can be used to implement PageUp/PageDown behaviors where pressing these keys
+   * scrolls the picker by more than 1 option. For example, a stride of 5 means select
+   * the enabled option 5 options after the active one. Note that the actual option selected
+   * may be past the stride if the option at the stride is disabled.
+   */
+  private findNextOption = (stride: number = 1) => {
```
sean-perkins, line 522: ~ `: number` type is redundant when assigning a primitive value to a property.

core/src/components/picker-column/picker-column.tsx
sean-perkins: I think a label is better than no label here. It is also incredibly straightforward for developers to customize it.

core/src/components/datetime/datetime.tsx
sean-perkins: I'm ok with the naming here.

sean-perkins (commented):
Moving into device testing next, small suggestion on a type adjustment for stricter types.

core/src/components/picker-column/picker-column.tsx
liamdebeasi: fixed in https://github.com/ionic-team/ionic-framework/pull/29371/commits/294d7712ce73d1a03cfa33d81ba604d7b2385066

sean-perkins (approved):

PR #29373, opened 2024-04-22
feat(many): expand global config for icons

thetaPC:
Issue number: internal

---------

Developers have to override icons per component instance. For example, to customize the clear input icon on Input, developers must set the `clearInputIcon` property on every instance of the component.

## What is the new behavior?

- The global config has been updated to let developers set icons at a higher level.
- Components have been updated to render the icons in the following priority: "Component property > Global config > Default"

- [ ] Yes
- [x] No

## Other information

Dev build: 8.0.0-dev.11713988726.1e82b9a0


core/src/components/fab-button/fab-button.tsx
```diff
@@ -152,8 +153,28 @@ export class FabButton implements ComponentInterface, AnchorInterface, ButtonInt
...
+  /**
+   * Get the icon to use for the close icon.
+   * If an icon is set on the component, use that.
+   * Otherwise, use the icon set in the config.
+   * If no icon is set in the config, use the default icon.
+   *
+   * @internal
+   * @returns {string} The icon to use for the close icon.
+   */
+  get fabButtonCloseIcon(): string {
```
sean-perkins, line 165: ~ This could probably be minimized to:

```ts
get fabButtonCloseIcon(): string {
  return this.closeIcon ?? config.get('fabButtonCloseIcon', close);
}
```

Since it isn't reused in multiple areas I'd recommend moving it into the `render` function:

```ts
const fabButtonCloseIcon = this.closeIcon ?? config.get('fabButtonCloseIcon', close);
```

Avoids the need for a getter, large documentation block and is 1 line of code.

core/src/components/input-password-toggle/input-password-toggle.tsx
```diff
@@ -105,14 +106,51 @@ export class InputPasswordToggle implements ComponentInterface {
...
+  /**
+   * Get the icon to use for the show icon.
+   * If an icon is set on the component, use that.
+   * Otherwise, use the icon set in the config.
+   * If no icon is set in the config, use the default icon.
+   *
+   * @internal
+   * @returns {string} The icon to use for the show icon.
+   */
+  get inputPasswordShowIcon(): string {
```
sean-perkins, line 118: ~ Similar to above, I don't think these need to be separate getters. They could be 1 line const.

sean-perkins (approved):

core/src/components/select/select.tsx
```diff
@@ -863,15 +864,13 @@ export class Select implements ComponentInterface {
...
     let icon: string;
 
-    if (isExpanded && expandedIcon !== undefined) {
-      icon = expandedIcon;
+    if (isExpanded) {
+      icon = selectExpandIcon;
     } else {
-      const defaultIcon = theme === 'ios' ? chevronExpand : caretDownSharp;
-      icon = toggleIcon ?? defaultIcon;
+      icon = selectCollapsedIcon;
```
brandyscarney, line 873: Minor, but should this be changed and combined with line 868?

```
let icon = isExpanded ? selectExpandIcon : selectCollapsedIcon;
```

core/src/components/searchbar/searchbar.tsx
```diff
@@ -100,7 +100,7 @@ export class Searchbar implements ComponentInterface {
    * Set the cancel button icon. Only available when the theme is `"md"`.
    * Defaults to `"arrow-back-sharp"`.
    */
-  @Prop() cancelButtonIcon = config.get('backButtonIcon', arrowBackSharp) as string;
```
brandyscarney, line 103: I think we should use `backButtonIcon` as a fallback if the config for `searchbarCancelIcon` is not set because this is useful for Material Design to keep them in sync: [Search specs](https://m3.material.io/components/search/specs#91980222-3d9b-4d8c-9900-df0a498c150d)

core/src/utils/config.ts
```diff
@@ -124,6 +124,108 @@ export interface IonicConfig {
...
+
+  /**
+   * Overrides the next icon for all `ion-datetime` components.
+   */
+  datetimeNextIcon?: string;
+
+  /**
+   * Overrides the previous icon for all `ion-datetime` components.
+   */
+  datetimePrevIcon?: string;
```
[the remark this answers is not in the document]
thetaPC: https://github.com/ionic-team/ionic-framework/pull/29373/commits/3830bf82e4421c83e32c21362afb97ea310815c4

core/src/components/searchbar/searchbar.tsx
thetaPC: https://github.com/ionic-team/ionic-framework/pull/29373/commits/e2a8046a33c734d24010a655b2701b1e06a71d1c

core/src/components/select/select.tsx
thetaPC: https://github.com/ionic-team/ionic-framework/pull/29373/commits/14cf2edbbc160beefdff5f0f65436cc24f7a2a32

core/src/components/toggle/toggle.tsx
```diff
@@ -243,15 +243,40 @@ export class Toggle implements ComponentInterface {
...
+  /**
+   * Get the icon to use for the checked icon.
+   * Otherwise, use the icon set in the config.
+   * If no icon is set in the config, use the default icon.
+   *
+   * @returns {string} The icon to use for the checked icon.
+   */
+  get toggleCheckedIcon(): string {
+    const theme = getIonTheme(this);
+    const defaultIcon = theme === 'md' ? checkmarkOutline : removeOutline;
```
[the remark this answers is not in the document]
thetaPC: I implemented it in this PR: https://github.com/ionic-team/ionic-framework/pull/29373/commits/bd008b763ae09863578a5ed3ba2b68fc0e924d82

core/src/components/searchbar/searchbar.tsx
```diff
@@ -609,13 +609,60 @@ export class Searchbar implements ComponentInterface {
...
     const animated = this.animated && config.getBoolean('animated', true);
     const theme = getIonTheme(this);
-    const clearIcon = this.clearIcon || (theme === 'ios' ? closeCircle : closeSharp);
-    const searchIcon = this.searchIcon || (theme === 'ios' ? searchOutline : searchSharp);
     const shouldShowCancelButton = this.shouldShowCancelButton();
+    /**
+     * The back button icon will be the default if available when
+     * the searchbar cancel icon is not set.
+     * This provides a backword compatibility since the back button
+     * icon was the default cancel button icon before.
```
brandyscarney, line 663:
```suggestion
     * The `backButtonIcon` config will be used as a fallback if the
     * `searchbarCancelIcon` config is not set. This ensures apps
     * have a way to keep the back button icon and searchbar cancel 
     * icon in sync.
```

I don't think we need to mention backwards compatibility here since this config just exists to keep the components in sync from a design perspective.

core/src/components/toggle/toggle.tsx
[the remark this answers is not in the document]
brandyscarney: Tested this out using the following config before & after your commits:

```html
<script>
  window.Ionic.config = {
    toggleCheckedIcon: 'pizza'
  }
</script>
```

The fix works great!

brandyscarney (approved):

PR #29375, opened 2024-04-22
feat(chip): set default shape to soft for ios and md theme and round for ionic theme

brandyscarney:
Issue number: internal

---------

The `shape` property defaults to `undefined` which evaluates to the "Soft" shape for the `ionic` theme and the "Round" shape for the `ios`/`md` themes.

| Property Value  | CSS Value (`ionic`) | CSS Value (`ios`) | CSS Value (`md`) |
| --------------- | ------ | ----------- | ----------- |
| `undefined`     | `4px`  | `16px`      | `16px`      |
| `"round"`       | `999px` | unsupported | unsupported |
| `"rectangular"` | `0px`  | unsupported | unsupported |

## What is the new behavior?
| Property Value  | CSS Value (`ionic`) | CSS Value (`ios`) | CSS Value (`md`) |
| --------------- | ------- | ------ | -------- |
| `"soft"`        | `4px`   | `10px`   | `8px`   |
| `"round"`       | `999px` | `999px` | `999px` |
| `"rectangular"` | `0px`   | `0px`   | `0px`   |

- Sets the default `shape` property to `"soft"` for `ios` and `md` themes and `"round"` for the ionic theme.
- Updates the `border-radius` to apply to the shape classes instead of `:host`
- Updates the `ios` theme to use `10px` for `"soft"` as this was taken from the App Store:
    ios-app-store
- Updates the `md` theme to use `8px` for `"soft"` as this is taken from the [Material Design 3 guidelines](https://m3.material.io/components/chips/specs#590903f7-2bf5-46ab-9810-d052173f41f1) & the previous value of `16px` is the equivalent of the `999px` round shape due to the height being `32px`
- Adds support for the `"round"` and `"rectangular"` shapes in `ios` and `md` themes

- [x] Yes
- [ ] No

BREAKING CHANGE:

The `border-radius` of the `ios` and `md` chip now defaults to `10px` and `8px`, respectively, instead of `16px` in accordance with the iOS and Material Design 3 guidelines. To revert to the previous appearance, set the `shape` to `"round"`, or override the `--border-radius` CSS variable to specify a different value.


core/src/components/chip/chip.scss
```diff
@@ -120,6 +120,21 @@
...
+:host(.chip-soft) {
+  --border-radius: 8px;
+}
+
+:host(.chip-round) {
+  --border-radius: 999px;
+}
+
+:host(.chip-rectangular) {
+  --border-radius: 0px;
```
brandyscarney, line 135: I hardcoded these values as I assumed we didn't want to use the design tokens in `ios` / `md` but I can update them if we do.

core/src/components/chip/chip.scss
```diff
@@ -120,6 +120,21 @@
   @include margin(-4px, -8px, -4px, 8px);
 }
 
+// Chip Shapes
+// ---------------------------------------------
+
+:host(.chip-soft) {
+  --border-radius: 8px;
```
brandyscarney, line 127: This was previously set to `16px` but I checked the [Material Design documentation](https://m3.material.io/components/chips/specs#590903f7-2bf5-46ab-9810-d052173f41f1) and they use `8px` in their latest guidelines so I decided to update it here.

core/src/components/chip/test/a11y/chip.e2e.ts-snapshots/chip-large-contents-md-ltr-Mobile-Chrome-linux.png
brandyscarney, line 1: All screenshot changes are expected because the `border-radius` has been changed to match the [Material Design 3 guidelines](https://m3.material.io/components/chips/specs#590903f7-2bf5-46ab-9810-d052173f41f1). If we wanted to continue to support the old look we could set the shape to `round` by default, but this would still scale the border radius for the increased font size screenshots. I personally think we should stick to the shape being the same default for all themes.

core/src/components/chip/chip.scss
thetaPC: Verified that it's 8px.

core/src/components/chip/test/a11y/chip.e2e.ts-snapshots/chip-large-contents-md-ltr-Mobile-Chrome-linux.png
thetaPC: I'm okay with your approach. It makes sense that the screenshots would be updated. I vote that we don't support the old look because:
1. it doesn't follow the current guidelines
2. we are targeting `next`. We would provide a heads up to devs of this change. It might be a big difference for some but at least it'll be done in `next`.

core/src/components/chip/chip.scss
thetaPC: I'm okay with this. We don't seem to use design tokens when it comes to border radius in other components for `ios` or `md`. So this approach follows the current structure.

BREAKING.md
```diff
@@ -13,3 +13,12 @@ This is a comprehensive list of the breaking changes introduced in the major ver
 - [Legacy](https://github.com/ionic-team/ionic-v3/blob/master/CHANGELOG.md)
 
 ## Version 9.x
+
+- [Components](#version-9x-components)
+  - [Chip](#version-8x-chip)
```
brandyscarney, line 18:
```suggestion
  - [Chip](#version-9x-chip)
```

core/src/components/chip/chip.tsx
```diff
@@ -37,21 +37,37 @@ export class Chip implements ComponentInterface {
...
+    const shape = this.getShape();
+
     return (
       <Host
         aria-disabled={this.disabled ? 'true' : null}
         class={createColorClasses(this.color, {
           [theme]: true,
-          // TODO(FW-6120): remove the theme==='ionic' when we add support for the `ios` and `md` modes.
-          [`chip-${shape}`]: theme === 'ionic' && shape !== undefined,
+          [`chip-${shape}`]: shape !== undefined,
```
sean-perkins, line 70: Can shape ever be `undefined` based on the returned value from `getShape()`? 
```suggestion
          [`chip-${shape}`]: true
```

brandyscarney: Fixed in https://github.com/ionic-team/ionic-framework/pull/29375/commits/dcf493be18aa2d0cf51111c4a7b20ac964756f2f

PR #29380, opened 2024-04-23
feat(input): add ionic theme styles and size property

brandyscarney:
Issue number: internal

---------

The input uses `md` styles on the `ionic` theme.

## What is the new behavior?
Adds the following for the `ionic` theme:
- The `--text-color-invalid` CSS variable
- The `size` property with support for the `large` size
- The `outline` fill styles for the input
- The `round` shape styles for the input
- The helper and error text styles
- The `focused`, `disabled` and `hover` styles
- The clear button styles

- [ ] Yes
- [x] No

## Other information

The outline fill & stacked label were the main priority here so some screenshots may look plain. Floating label will be added in future work.


thetaPC (approved):
LGTM. Might want to add to the description that outlined, stacked input was the main focus so some of the screenshots may look plain.

liamdebeasi:
Going to remove myself from reviewing since you have an approval. Lmk if you'd still like me to take a look!

core/src/components/input/input.tsx
```diff
@@ -263,9 +263,9 @@ export class Input implements ComponentInterface {
 
   /**
    * The size of the input. If "large", it will have an increased height. By default the
-   * size is unset. This property only applies to the `"ionic"` theme.
+   * size is medium. This property only applies to the `"ionic"` theme.
    */
-  @Prop() size?: 'large';
+  @Prop() size?: 'medium' | 'large' = 'medium';
```
brandyscarney, line 268: We might want to create a ticket to track adding default sizes to the other components. Right now we have them set to undefined, this will be the first component that sets a default.

thetaPC: Would it be better to change it to undefined?

brandyscarney: I think we should have a defined size like shape does, we will just have to add the support for it to the other themes and update breaking changes.

thetaPC: Okay, parent FW-6216 has been created to track all the changes.

brandyscarney: Thank you!

PR #29395, opened 2024-04-24
feat(input): add item highlight for ionic theme

sean-perkins:
Issue number: N/A

---------

The current border implementation causes layout shift. 

## What is the new behavior?

- Matches MD highlight effect for the focus border
- Border implementation does not cause layout shift

- [ ] Yes
- [x] No


sean-perkins:
Please hold, I added a small visual regression here. I'll fix the border width for outline and make sure we have screenshots capturing this.

core/src/components/input/test/fill/input.e2e.ts
```diff
@@ -111,6 +111,26 @@ configs({ modes: ['md'] }).forEach(({ title, screenshot, config }) => {
         const input = page.locator('ion-input');
         await expect(input).toHaveScreenshot(screenshot(`input-fill-outline`));
       });
+
+      test('focus state should not have visual regressions', async ({ page }) => {
```
thetaPC, line 115: This seems to be a duplicate from this existing [test](https://github.com/ionic-team/ionic-framework/blob/f81f6c241ec1b318f0a0e1d1f5eb40f36735f333/core/src/components/input/test/states/input.e2e.ts#L147-L148). What's the difference?

sean-perkins: Hmm didn't know the outline focus status was being captured already. It didn't catch the border width reducing from `2px` to `1px` (accidental earlier diff). 

Let me try a few things locally, such as splitting the images to see if it can detect the change. 

I can remove the added test from this PR.

brandyscarney (approved):
Code looks good and it works to fix the highlight. Makes sense to copy the highlight code from the `md` theme. 👍

core/src/components/input/test/fill/input.e2e.ts
thetaPC: I think you need to also remove the screenshots that were created from the tests.

Sometimes, it doesn't capture because the change is very minor. You have to delete the screenshot so it can generate a fresh one.

thetaPC (approved):
LGTM

PR #29403, opened 2024-04-25
fix(input): match medium size for ionic theme

thetaPC:
Issue number: internal

---------

There is only one size, large, for `ionic` theme. Developers can also leave it undefined and no styles will be applied.

## What is the new behavior?

- The `medium` size has been added. It's also being used as the default.
- Tests have been added.
- Target area has been added.

- [ ] Yes
- [x] No

## Other information

Verify that interactive elements can be accessed in the start and end slots along with the clear icon button.

[Preview](https://ionic-framework-git-fw-6099-ionic1.vercel.app/src/components/input/test/size?ionic:theme=ionic)


core/src/components/input/test/size/input.e2e.ts-snapshots/input-size-medium-ionic-md-ltr-light-Mobile-Chrome-linux.png
thetaPC, line 1: These look plain because underlined hasn't been implemented yet.

core/src/components/input/input.ionic.scss
```diff
@@ -19,10 +19,46 @@
...
+::slotted([slot="start"]),
+::slotted([slot="end"]),
+.input-clear-icon {
+  /**
+   * The target area has a z-index of 1, so the slotted elements 
+   * should be higher. Otherwise, the slotted elements will not 
+   * be interactable. This is especially important for the clear
+   * button, which should be clickable.
+   */
+  z-index: 10;
```
sean-perkins, line 59: What is the reasoning behind `10` vs. `2` here?

core/src/components/input/input.ionic.scss
```diff
@@ -19,10 +19,46 @@
...
+  min-height: 40px;
+}
+
 :host(.input-size-large) .native-wrapper {
   min-height: 48px;
 }
 
+// Target area
+// --------------------------------------------------
+:host .native-wrapper::after {
```
sean-perkins, line 32: ~ No changes needed here. We should consider a mixin for these common styles that we are implementing for target area in multiple components.

sean-perkins (commented):
One question around the `z-index` value, otherwise changes look good.

core/src/components/input/input.ionic.scss
thetaPC: No reasoning. I just found 10 to be visually nicer. It works with 2, updating it now.

core/src/components/input/input.ionic.scss
thetaPC: I'll add it to my todo list.

sean-perkins (approved):

liamdebeasi:
Sean approved this, so going to remove myself from review. Let me know if you'd still like me to take a look though!

PR #29407, opened 2024-04-26
docs(core): ionChange will not emit from programmatically changing value

sean-perkins:
Issue number: resolves https://github.com/ionic-team/ionic-docs/issues/3588

---------

The documentation around `ionChange` not being emitted when programmatically changing the property associated to the "value" is either inconsistent or missing from certain components.

## What is the new behavior?

- Adds the documentation to the missing components.
- Makes the documentation consistent across components.

- [ ] Yes
- [x] No

### DIFF
--- a/core/src/components/segment/segment.tsx
+++ b/core/src/components/segment/segment.tsx
@@ -94,7 +94,7 @@ export class Segment implements ComponentInterface {
   @Prop() selectOnFocus = false;
 
   /**
-   * Emitted when the value property has changed and any ragging pointer has been released from `ion-segment`.
+   * Emitted when the value property has changed and any dragging pointer has been released from `ion-segment`.
    *
    * This event will not emit when programmatically setting the `value` property.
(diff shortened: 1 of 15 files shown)


core/src/components/segment/segment.tsx
```diff
@@ -94,8 +94,9 @@ export class Segment implements ComponentInterface {
   @Prop() selectOnFocus = false;
 
   /**
-   * Emitted when the value property has changed and any
-   * dragging pointer has been released from `ion-segment`.
+   * Emitted when the value property has changed and any ragging pointer has been released from `ion-segment`.
```
thetaPC, line 97:
```suggestion
   * Emitted when the value property has changed and any dragging pointer has been released from `ion-segment`.
```

thetaPC (changes requested):

thetaPC (approved):
LGTM

Issue #28816, opened 2024-01-11
bug: Active segment button not highlighted when nested into async container


leomazza:
### Prerequisites

- [X] I have read the [Contributing Guidelines](https://github.com/ionic-team/ionic-framework/blob/main/.github/CONTRIBUTING.md#creating-an-issue).
- [X] I agree to follow the [Code of Conduct](https://ionicframework.com/code-of-conduct).
- [X] I have searched for [existing issues](https://github.com/ionic-team/ionic-framework/issues) that already report this problem, without success.

### Ionic Framework Version

v7.x

### Current Behavior

A segment with dynamic value does not highlight the selected button when nested into an `async` container.

### Expected Behavior

It should be highlighted just as it is usually the case.

### Steps to Reproduce

1. Create an `<ion-segment [value]="customValue">` with at least 2 `<ion-segment-button>`s.
2. Nest that one into a `<div *ngIf="observable$ | async">`.
3. See error

### Code Reproduction URL

https://github.com/leomazza/ionic-segment-bug

### Ionic Info

Ionic:

   Ionic CLI                     : 7.2.0 (/Users/leomazza/.nvm/versions/node/v20.10.0/lib/node_modules/@ionic/cli)
   Ionic Framework               : @ionic/angular 7.6.4
   @angular-devkit/build-angular : 17.0.10
   @angular-devkit/schematics    : 17.0.10
   @angular/cli                  : 17.0.10
   @ionic/angular-toolkit        : 9.0.0

Capacitor:

   Capacitor CLI      : 5.6.0
   @capacitor/android : not installed
   @capacitor/core    : 5.6.0
   @capacitor/ios     : not installed

Utility:

   cordova-res : 0.15.4
   native-run  : 2.0.0

System:

   NodeJS : v20.10.0 (/Users/leomazza/.nvm/versions/node/v20.10.0/bin/node)
   npm    : 10.2.3
   OS     : macOS Unknown

### Additional Information

![image](https://github.com/ionic-team/ionic-framework/assets/19129296/956ca0ef-de2a-41a3-8e13-39a37189b6cd)

NOTE: The issue is with the second segment block. The first segment button is the selected one but it is not shown as active/checked.

liamdebeasi:
Thanks, I can reproduce this. Here is a dev build if you are interested in testing a proposed fix:

```
npm install @ionic/angular@7.6.5-dev.11705415448.16878103
```

liamdebeasi:
[...]

PR #28837, opened 2024-01-16
fix(segment): setting value via binding updates button state

liamdebeasi:
Issue number: resolves #28816

---------

## What is the current behavior?

The value is set on Segment asynchronously when binding it in Angular. However, the timing works out such that the value changes after `connectedCallback` is fired but before any Stencil Watchers are configured. As a result, our `value` property watcher does not fire which causes `ionSelect` to not be emitted. Segment Buttons rely on this event to know when to update their state (if the value changes such that a segment button is now selected). This results in a checked segment button not appearing checked.

This is similar to other issues that have been fixed:

https://github.com/ionic-team/ionic-framework/pull/28510
https://github.com/ionic-team/ionic-framework/pull/28488
https://github.com/ionic-team/ionic-framework/pull/28526

## What is the new behavior?

- Segment now emits `ionSelect` on `componentDidLoad` so that any descendant segment buttons can update correctly.

## Does this introduce a breaking change?

- [ ] Yes
- [x] No

## Other information

Dev build: `7.6.5-dev.11705415448.16878103`

This is a timing issue with Stencil, so I am unable to write a reliable automated test. Reviewers should test the dev build in the repro provided in the linked issue.

PR #29418, opened 2024-04-29
fix(angular): set active segment button when dynamically changing items

### LINKED ISSUE #29414: bug: segment does not set active button when dynamically changing options in angular
### Prerequisites

### Ionic Framework Version

v7.x

### Current Behavior

When updating the nested segment buttons of the Segment control the active segment button looses it's checked state even when the value matches the value of the containing Segment control.
The issue is similar to #28816 but for segment-button, i.e. the value changes after `connectedCallback` is fired but before any Stencil Watchers are configured. As a result, the `value` property watcher does not fire which causes `updateState` not to be called.
 The fix in #28837 only fixes the problem on initial rendering, but not on subsequent updates to the nested segment buttons.

### Expected Behavior

The active segment button should be highlighted when it's `value` matches the `value` of it's containing segment control - as was the case before standalone components.

### Steps to Reproduce

1. Go to https://stackblitz.com/edit/angular-elmzeu
2. See the segment button "First" being highlighted
3. Wait 2 seconds
4. See the segment button "First" NOT being highlighted even though it's `value` matches the Segment control's `value`

### Code Reproduction URL

https://stackblitz.com/edit/angular-elmzeu

### Ionic Info

Ionic:

   Ionic CLI : 5.4.16

Utility:

   cordova-res                          : not installed
   native-run (update available: 2.0.1) : 1.7.4

System:

   NodeJS : v20.11.0
   npm    : 10.2.4
   OS     : macOS Unknown

### Additional Information

In the Stackblitz reproduction example the nested segment buttons are updated with an additional button, but keeping same value for the first 2 buttons - the problem is the same though when rendering nested buttons each with a new value while updating the `value` of the Segment control at the same time to match one of the newly rendered buttons.

sean-perkins:
Issue number: resolves #29414
---------

In Angular when the segment buttons are dynamically rendered, the segment will not set the active visual state for the selected segment after the re-render.

## What is the new behavior?

- In Angular the segment will set the active item visually when the items are dynamically changed

Previously this PR: https://github.com/ionic-team/ionic-framework/pull/28837 aimed to resolve https://github.com/ionic-team/ionic-framework/issues/28816. I have confirmed that the modified approach in the dev-build fixes #29414 and #28816 and the previous change is no longer needed. 

- [ ] Yes
- [x] No

## Other information

Dev-build: `8.0.2-dev.11714411675.10f48160`

### DIFF
--- a/core/src/components/segment/segment.tsx
+++ b/core/src/components/segment/segment.tsx
@@ -142,14 +142,6 @@ export class Segment implements ComponentInterface {
     this.setCheckedClasses();
 
     /**
-     * If the value changes before watchers
-     * are setup, then the ionSelect watch callback
-     * will not fire. As a result, we manually
-     * fire this event when Select is loaded.
-     */
-    this.ionSelect.emit({ value: this.value });
-
-    /**
      * We need to wait for the buttons to all be rendered
      * before we can scroll.
      */
@@ -495,6 +487,10 @@ export class Segment implements ComponentInterface {
     }
   };
 
+  private onSlottedItemsChange = () => {
+    this.valueChanged(this.value);
+  };
+
   private getSegmentButton = (selector: 'first' | 'last' | 'next' | 'previous'): HTMLIonSegmentButtonElement | null => {
     const buttons = this.getButtons().filter((button) => !button.disabled);
     const currIndex = buttons.findIndex((button) => button === document.activeElement);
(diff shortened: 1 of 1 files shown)


jakobe (approved):
I can confirm this fixes the mentioned issue  👍🏻

core/src/components/segment/segment.tsx
```diff
@@ -495,6 +487,10 @@ export class Segment implements ComponentInterface {
     }
   };
 
+  private onSlottedItemsChange = () => {
+    this.valueChanged(this.value);
```
thetaPC, line 491: Should there be a comment explaining why this is needed?

PR #29425, opened 2024-04-30
chore(ci): exit with correct exit code

liamdebeasi:
The docker script does not exit with the correct exit code when invoking commends in `execa`. You can verify this by changing `npm run test.e2e` in the script to `npm run foo` (a non-existent script). The underlying command will fail, but the parent script will still exit with exit code 0.

Important: The tests themselves have always errored correctly, so this does not impact test execution. Only the invocation of the test command is impacted.

This PR ensures that invoking the wrong command causes the parent script to exit with the correct code.


liamdebeasi:
This change doesn't seem to be correct (at least for flaky tests). Going to close, but I will look into this a bit more if I have time.

brandyscarney (approved):

PR #29441, opened 2024-05-01
docs(testing): add correct command

liamdebeasi:
The original command does not exist.


thetaPC (approved):
LGTM, I miss that command 😞

liamdebeasi:
We can add it back! I just removed the reference since it made the instructions incorrect 😄

PR #29446, opened 2024-05-02
fix(input): match design for xlarge size

thetaPC:
Issue number: internal

---------

There is no `xlarge` size for the ionic theme.

## What is the new behavior?

- The `xlarge` size has been added.
- Tests have been added.

- [ ] Yes
- [x] No

## Other information

[Preview](https://ionic-framework-git-fw-6099-ionic1.vercel.app/src/components/input/test/size?ionic:theme=ionic)


core/src/components/input/input.tsx
```diff
@@ -504,9 +504,11 @@ export class Input implements ComponentInterface {
   private getSize() {
     const theme = getIonTheme(this);
     const { size } = this;
-    if (theme !== 'ionic' && size === 'large') {
+    const ionicSizes = ['large', 'xlarge'];
+    if (theme !== 'ionic' && size && ionicSizes.includes(size)) {
```
sean-perkins, line 508: ~ nit: Creating an array and checking if a value exists in it is computationally slower than checking the two cases of if the size is large or xlarge. We could probably simplify that so this practice isn't copied into other areas.

sean-perkins (approved):
One minor nit, changes look good otherwise 👍

core/src/components/input/input.tsx
thetaPC: Fixed: https://github.com/ionic-team/ionic-framework/pull/29446/commits/59e79a3f1e19777d61a62e1955b953fd36a064f4

PR #29448, opened 2024-05-02
fix(angular): persist select disabled state in item

### LINKED ISSUE #29234: bug: disabled select inside item is not being fully disabled
### Prerequisites

### Ionic Framework Version

v7.x

### Current Behavior

Note: The issue only seems to appear when using Angular standalone components

When disabling a `<ion-select>` inside of a `<ion-item>`, the hover and touch / click styling still appears as enabled and pointer events are still executed.

### Expected Behavior

We would expect the styling to appear disabled as well as `point-events: none` causing no `(click)` handlers to be called. This does seem to be the case when using the classic `IonicModule.forRoot()` approach and non-standalone components.

### Steps to Reproduce

1. Click on the button within the stackblitz
2. Observe the styling when hovering  / clicking on the ion-select
3. Observe the console output when clicking

### Code Reproduction URL

https://stackblitz.com/edit/angular-vvbkmf-cz7bqx?file=src%2Fapp%2Fapp.component.html

### Ionic Info

see Stackblitz setup

### Additional Information

_No response_

sean-perkins:
Issue number: resolves #29234

---------

In certain scenarios, such as in Angular where the property binding is set a few frames after the element is rendered, the `ionStyle` event from `ion-select` can be emitted before `ion-item` has registered an event listener.

This results in situations like setting the `ion-select` as initially disabled can cause the item to not treat the element as not interactable (receives pointer events).

## What is the new behavior?

- Emits the `ionStyle` event when the `ion-select` is rendered.
- `ion-item` consistently detects the state of `ion-select` and applies the appropriate styles

- [ ] Yes
- [x] No

## Other information

Forked reproduction and dev-build available here: https://github.com/ionic-team/ionic-framework/issues/29234#issuecomment-2091866453

### DIFF
--- a/core/src/components/select/select.tsx
+++ b/core/src/components/select/select.tsx
@@ -271,6 +271,16 @@ export class Select implements ComponentInterface {
   }
 
   componentDidLoad() {
+    /**
+     * If any of the conditions that trigger the styleChanged callback
+     * are met on component load, it is possible the event emitted
+     * prior to a parent web component registering an event listener.
+     *
+     * To ensure the parent web component receives the event, we
+     * emit the style event again after the component has loaded.
+     *
+     * This is often seen in Angular with the `dist` output target.
+     */
     this.emitStyle();
   }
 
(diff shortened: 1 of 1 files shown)


thetaPC (approved):
LGTM, works great

PR #29456, opened 2024-05-03
test(angular): add angular 18 test app

sean-perkins:
Issue number: N/A

---------

Angular 18 will be releases ~ May 20th. Ionic Framework does not currently have test infrastructure to check if the RC is compatible with Ionic Framework.

## What is the new behavior?

- Adds an Angular 18 test app to the test infrastructure
  - Added tech debt tickets for additional investigation and removing the `--legacy-peer-deps` arg after v18 GA.

- [ ] Yes
- [x] No


thetaPC (approved):
LGTM, builds locally

PR #29458, opened 2024-05-03
chore(theming): remove Sass default flag & update guidelines

brandyscarney:
Remove a missed `!default` flag in the themes directory and update the Sass variable guidelines docs to eliminate the `!default` flag from the code examples, except for in the historic usage section which includes examples from previous versions where the `!default` flag was used.


sean-perkins (approved):

thetaPC (approved):
LGTM

PR #29404, opened 2024-04-25
feat(button): set default shape to soft for ios and round for md and ionic themes

brandyscarney:
Issue number: internal

---------

## What is the current behavior?
The `shape` property defaults to `"soft"` for `ios` and `"round"` for the `md` and `ionic` themes.

**Default button size**:

| Property Value  | CSS Value (`ionic`) | CSS Value (`ios`) | CSS Value (`md`) |
| --------------- | ------ | ----------- | ----------- |
| `undefined`     | `8px`  | `14px`      | `4px`      |
| `"round"`       | `999px` | unsupported | unsupported |
| `"rectangular"` | `0px`  | unsupported | unsupported |

**Large button size**:

| Property Value  | CSS Value (`ionic`) | CSS Value (`ios`) | CSS Value (`md`) |
| --------------- | ------ | ----------- | ----------- |
| `undefined`     | `8px`  | `16px`      | `4px`      |

**Small button size**:

| Property Value  | CSS Value (`ionic`) | CSS Value (`ios`) | CSS Value (`md`) |
| --------------- | ------ | ----------- | ----------- |
| `undefined`     | `4px`  | `6px`      | `4px`      |

## What is the new behavior?
The `shape` property defaults to `undefined` which evaluates to the "Soft" shape for all themes. 

**Default button size**:

| Property Value  | CSS Value (`ionic`) | CSS Value (`ios`) | CSS Value (`md`) |
| --------------- | ------- | ------ | -------- |
| `"soft"`        | `8px`   | `6px`   | `4px`   |
| `"round"`       | `999px` | `999px` | `999px` |
| `"rectangular"` | `0px`   | `0px`   | `0px`   |

**Large button size**:

| Property Value  | CSS Value (`ionic`) | CSS Value (`ios`) | CSS Value (`md`) |
| --------------- | ------ | ----------- | ----------- |
| `soft`     | `8px`  | `8px`      | `4px`      |

**Small button size**:

| Property Value  | CSS Value (`ionic`) | CSS Value (`ios`) | CSS Value (`md`) |
| --------------- | ------ | ----------- | ----------- |
| `soft`     | `4px`  | `4px`      | `4px`      |

- Adds support for the `"soft"` shape in the `ionic` theme using the existing values for `undefined`
- Adds support for the `"rectangular"` and `"round"` shapes in `ios` and `md` using `0px` and `999px` border radius
[...]

PR #29459, opened 2024-05-03
feat(button): update md button to use md3 padding

brandyscarney:
Issue number: internal

---------

Material Design and iOS buttons set different padding amounts when the `shape` is `"round"`. This was made more apparent when we changed the default shape to round for `md`, as all of the existing screenshots showed that the buttons no longer had padding top or bottom. 

## What is the new behavior?
Removes the different padding targeting `:host(.button-round)`. All button shapes should use the same padding according to guidelines. 

- [ ] Yes
- [x] No

### DIFF
--- a/core/src/components/button/button.ios.scss
+++ b/core/src/components/button/button.ios.scss
@@ -119,13 +119,8 @@
   --border-radius: 4px;
 }
 
-// TODO(FW-6202): remove the custom padding styles
 :host(.button-round) {
   --border-radius: 999px;
-  --padding-top: #{$button-ios-round-padding-top};
-  --padding-start: #{$button-ios-round-padding-start};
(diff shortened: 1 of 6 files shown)


brandyscarney:
I decided to commit this directly to FW-6192 to avoid having to update screenshots multiple times and fix conflicts: https://github.com/ionic-team/ionic-framework/pull/29404/commits/e6b7bf353951fb308a0906a26bfb6a4612226c1e

I was going to update the `md` padding to match md3 and `ios` to match iOS 17 but it is really incomplete without the text-transform, letter-spacing & height changes they need so I will do those separately. This commit will fix the screenshots that are updated due to `md`'s default shape changing to `"round"`.

PR #29476, opened 2024-05-08
feat(input): add rectangular shape for the ionic theme

brandyscarney:
Issue number: internal

---------

The input supports an undefined and a `round` shape.

## What is the new behavior?
Adds support for the `rectangular` shape for the `ionic` theme & screenshot tests for this shape with the outline fill.

- [ ] Yes
- [x] No


core/src/components/input/test/shape/input.e2e.ts
```diff
@@ -109,3 +109,60 @@ configs({ modes: ['ionic-md', 'md'] }).forEach(({ title, screenshot, config }) =
     });
   });
 });
+
+// Rectangular shape is only available in ionic theme
+// TODO(FW-6098): Add test for rectangular shape in md
+// by combining these tests with the above tests
```
thetaPC, line 115: Nit pick: Move the comment above of line 118. I also have a similar test for soft. It would be easier to read the comments if they're near their addressed test vs all at the top.

core/src/components/input/test/shape/input.e2e.ts
```diff
@@ -109,3 +109,60 @@ configs({ modes: ['ionic-md', 'md'] }).forEach(({ title, screenshot, config }) =
...
+            ></ion-input>
+            `,
+            config
+          );
+
+          const input = page.locator('ion-input');
+          await expect(input).toHaveScreenshot(screenshot(`input-shape-rectangular-fill-outline`));
+        });
+
+        test('border radius should be customizable', async ({ page }) => {
```
thetaPC, line 140: Do we need to add this test? I wonder if it's redundant to the one in outline.

thetaPC (changes requested):

core/src/components/input/test/shape/input.e2e.ts
brandyscarney: Moved in https://github.com/ionic-team/ionic-framework/pull/29476/commits/329585d9fb9a0c865bd3de38fe1b6b492fd970ee

core/src/components/input/test/shape/input.e2e.ts
brandyscarney: I don't think it's redundant since we use different selectors to style the border radius based on the shape. This ensures it still works with the rectangular shape if we end up changing the ionic theme to act like md in the future: https://github.com/ionic-team/ionic-framework/blob/704765e6203bc35c9ef78fabe6e15980bfa834ec/core/src/components/input/input.md.outline.scss#L200-L209

PR #29482, opened 2024-05-09
fix(input): only apply the shape to the outline fill in ionic theme

brandyscarney:
Issue number: internal

---------

The shape styles are being applied when the `fill` is unset in the `ionic` theme. This is incorrect per the input UX requirements as shape should only apply when `fill` is set to `"outline"`. This is made apparent when looking at the existing styles for `disabled`.

## What is the new behavior?
- Only apply the shape styles to the `"outline"` fill
- Move the use of the `--background` css variable to the native wrapper since this is what we use to style both the disabled and readonly states

| Before | After |
| ---| ---|
| before | after |

I am not sure why this wasn't caught with the shape additions, but it's probably because the minimum pixel ratio was not met. 

- [ ] Yes
- [x] No

## Other information

[Preview](https://ionic-framework-git-fix-input-shape-underline-ionic1.vercel.app/src/components/input/test/states?ionic:theme=ionic)

### DIFF
--- a/core/src/components/input/input.ionic.outline.scss
+++ b/core/src/components/input/input.ionic.outline.scss
@@ -53,6 +53,13 @@
    * input + label.
    */
   border-bottom: none;
+
+  /**
+   * Do not show a background on the input wrapper as
+   * this includes the label, instead we apply the
+   * background to the native wrapper.
+   */
+  background: transparent;
 }
 
 :host(.input-fill-outline.input-shape-round) .label-text-wrapper {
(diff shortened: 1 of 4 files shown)


core/src/components/input/input.ionic.outline.scss
```diff
@@ -128,21 +142,19 @@
   margin-inline-start: 8px;
 }
 
-// Input - Disabled
-// ----------------------------------------------------------------
+// Input Shapes
+// --------------------------------------------------
 
-:host(.input-fill-outline.input-disabled) {
```
brandyscarney, line 134: This can be removed because it is covered by the `:host(.input-disabled)` class since I moved the `--background` variable assignment to the proper element

thetaPC (approved):
LGTM

PR #29483, opened 2024-05-09
feat(input): add the readonly styles for the ionic theme

brandyscarney:
Issue number: internal

---------

There are no custom styles when readonly is added to an input.

## What is the new behavior?
- Applies an `input-readonly` class when the `readonly` property is `true`
- Styles the class in the `ionic` theme to match the design requirements for readonly inputs
- Adds screenshot tests for the readonly style for an unset `fill` and `"outline"` fill

- [ ] Yes
- [x] No

## Other information

[Preview](https://ionic-framework-git-rou-5155-ionic1.vercel.app/src/components/input/test/states?ionic:theme=ionic)


core/src/components/input/input.ionic.scss
```diff
@@ -7,6 +7,8 @@
 // --------------------------------------------------
 
 :host {
+  // TODO(ROU-5477): use design token here when it is added
+  --color: #121212;
```
thetaPC, line 11: Was it verified that it's going to be added? Else we might need to reach out to the designers to verify that this hex is the correct value.

brandyscarney: Yes I have verified that the design tokens for state have yet to be added. I actually asked about this specific hex before because I ran into it in other components. They are planned in ROU-4870 which is linked as a blocker to the ticket I added here.

thetaPC (approved):
LGTM

PR #29486, opened 2024-05-09
fix(core): malformed URIs will not throw exception

### LINKED ISSUE #29479: bug: empty screen when URI is malformed
### Prerequisites

### Ionic Framework Version

v7.x, v8.x

### Current Behavior

When the URI is malformed, Ionic throws the "URI malformed" error, resulting in an empty screen being rendered. We want to handle this error using a React error boundary and render our standard error message instead. The issue arises when Ionic attempts to get configuration from the URL: https://github.com/ionic-team/ionic-framework/blob/main/core/src/global/ionic-global.ts#L43.

Additionaly, the second issue is: How can we disable this behavior, as we don't want to allow reconfiguration of our app from the URL in production?"

### Expected Behavior

When the URI is malformed, Ionic should not drop an error. Instead, it should provide a way to catch this error or ignore it.

### Steps to Reproduce

1. Open any Ionic app in web browser
2. Add the query param `?test=%`. E.g. https://www.pandaexpress.com/location?test=%
3. If your web server doesn't handle malformed URIs, you'll encounter a white screen in production mode.

Note that in development mode, you're more likely to see the standard error message from Vite.

### Code Reproduction URL

https://www.pandaexpress.com/location?test=%

### Ionic Info

Ionic:

   Ionic CLI       : 7.2.0
   Ionic Framework : @ionic/react 7.5.8

Capacitor:

   Capacitor CLI      : 5.6.0
   @capacitor/core    : 5.6.0

Utility:

   native-run (update available: 2.0.1) : 2.0.0

System:

   NodeJS : v18.15.0
   npm    : 9.5.0
   OS     : Linux 5.15

### Additional Information

_No response_

sean-perkins:
Issue number: resolves #29479

---------

If an application includes a malformed URI, an Ionic Framework can "crash" due to an uncaught exception in parsing the URI for the Ionic config.

## What is the new behavior?

- Handle the malformed URI fallback if the config cannot be determined
- Added unit tests for this case

- [ ] Yes
- [x] No

### DIFF
--- a/core/src/global/config.ts
+++ b/core/src/global/config.ts
@@ -60,7 +60,13 @@ export const configFromURL = (win: Window) => {
     .slice(1)
     .split('&')
     .map((entry) => entry.split('='))
-    .map(([key, value]) => [decodeURIComponent(key), decodeURIComponent(value)])
+    .map(([key, value]) => {
+      try {
+        return [decodeURIComponent(key), decodeURIComponent(value)];
+      } catch (e) {
+        return ['', ''];
+      }
+    })
     .filter(([key]) => startsWith(key, IONIC_PREFIX))
     .map(([key, value]) => [key.slice(IONIC_PREFIX.length), value])
     .forEach(([key, value]) => {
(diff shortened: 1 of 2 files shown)


thetaPC (approved):
LGTM

thetaPC:
Dev build: 8.1.2-dev.11715623495.196f4eba

PR #29497, opened 2024-05-13
fix(picker): update keyboard navigation

thetaPC:
Issue number: N/A

---------

The tests for picker were failing because the focus was not set to the appropriate element and because Firefox would focus on an element it shouldn't be.

## What is the new behavior?

- Adjusted the `setFocus()`
- Skipped the `picker-opts` in the tab order
- Removed `skip()` from the tests

- [ ] Yes
- [x] No

## Other information

N/A


brandyscarney (approved):
Looks great!

PR #29502, opened 2024-05-14
test(datetime, tab-button): update screenshots after playwright update

thetaPC:
Issue number: N/A

---------

Renovate has updated [Playwright to v1.44.0](https://github.com/ionic-team/ionic-framework/pull/29470). The update is now using Chromium 125.0.6422.14, which leads to a different behavior with the datetime scale test. As for tab-button, it seems that there's a small rendering issue. Since it's only a few pixels in difference then it's okay to update the screenshots.

## What is the new behavior?

- Updated the datetime snapshot, this snapshot now aligns with Firefox and Safari. Larger font sizes leads to the weekdays to be abbreviated.
- Updated the tab-button snapshots

- [ ] Yes
- [x] No

## Other information

N/A


core/src/components/datetime/test/a11y/datetime.e2e.ts-snapshots/datetime-scale-ios-ltr-Mobile-Chrome-linux.png
brandyscarney, line 1: How was only one screenshot updated here?

thetaPC: Apparently, the other browsers already had the one letter shortened weekdays. Chrome was the only one that still showed the 3-4 character version.

brandyscarney (approved):

PR #29520, opened 2024-05-17
fix(input): adjust colors, padding, fonts & heights for ionic theme

brandyscarney:
Issue number: internal

---------

Input styles were slightly different from the ionic theme design. 

## What is the new behavior?
Updates the following:
- The font size for input (14px) and labels (12px)
- The font weight, line-height and color for the label
- The padding for the label & helper message
- Removes the label transform in order to set `font-size` directly
- Updates the Sass variables to be referenced from `tokens` where they were not

- [ ] Yes
- [x] No

## Other information

The underline (undefined) fill looks off from the design but issue ROU-5151 exists for implementing it so I assume it will be fixed there. 

- [Shapes preview](https://ionic-framework-git-fix-input-placeholder-ionic1.vercel.app/src/components/input/test/shape?ionic:theme=ionic)
- [Fills preview](https://ionic-framework-git-fix-input-placeholder-ionic1.vercel.app/src/components/input/test/fill?ionic:theme=ionic)

## Image Comparison

| Before | After | UX Design |
| ---| ---| ---|
| ionic-before | ionic-after | UX design |

> [!NOTE]
> The design is not going to match perfectly because we are not using the Inter font in our tests.


core/src/components/input/input.ionic.scss
```diff
@@ -103,25 +105,41 @@
...
+  color: $ionic-color-neutral-700;
+
+  font-size: $ionic-font-size-s;
+  font-weight: $ionic-font-weight-medium;
+
+  line-height: $ionic-font-line-height-s;
+}
+
+:host(.label-floating) .label-text-wrapper {
+  @include transform(none);
```
brandyscarney, line 121: By removing this transform we are able to set the `font-size` directly instead of scaling it down from the `:host` font size.

core/src/components/input/input.ionic.outline.scss
```diff
@@ -121,14 +121,8 @@
...
 // This makes the label sit above the input.
 :host(.label-floating.input-fill-outline.input-label-placement-stacked) .label-text-wrapper {
-  @include transform(translateY(0), scale(#{$form-control-label-stacked-scale}));
   @include margin(0);
-
-  /**
-   * Label text should not extend
-   * beyond the bounds of the input.
-   */
-  max-width: calc((100% - var(--padding-start) - var(--padding-end)) / #{$form-control-label-stacked-scale});
```
brandyscarney, line 131: This is no longer needed without the `transform` as a rule already exists to set the `max-width` here:

https://github.com/ionic-team/ionic-framework/blob/bf8dfdb5b1596e2880c6fdbc20559cbb1231649a/core/src/components/input/input.ionic.outline.scss#L76

thetaPC (approved):
LGTM, verified with Figma too

PR #29545, opened 2024-05-23
feat(tokens): add html preview based on design-tokens

BenOsodrac:
Issue number: internal

---------

## What is the new behavior?

- Added new files to scripts/tokens for the html template and css to style the generated preview
- Added Style Dictionary register for generating html elements for each main token

- [ ] Yes
- [x] No


OS-giulianasilva (approved):
Nice!

PR #29557, opened 2024-05-28
feat(ionic): ionic colors support

BenOsodrac:
Issue number: internal

---------

## What is the new behavior?

- Changed createColorClasses function to support Ionic Theme Design Systems colors. Example: `<ion-button color="red-600"></ion-button>`
- Changed colors utility-classes to manipulate css variables instead, on the components scope.
- Update snapshots for button tests that were testing ionic theme button with old ion-colors. Those are different design systems, that should not be tested together. In the future a new html test for button can be created, with all the colors available on the new Design System.

- [ ] Yes
- [x] No


core/src/utils/theme.ts
```diff
@@ -6,15 +6,20 @@ export const hostContext = (selector: string, el: HTMLElement): boolean => {
...
+ * Create the mode and color classes for the component based on the classes passed in and the theme being used
  */
 export const createColorClasses = (color: Color | undefined | null, cssClassMap: CssClassMap): CssClassMap => {
   return typeof color === 'string' && color.length > 0
-    ? {
-        'ion-color': true,
-        [`ion-color-${color}`]: true,
-        ...cssClassMap,
-      }
+    ? cssClassMap.ionic === true
```
brandyscarney, line 13: This is a pretty big divergence from the way we have generated & used colors. Was there a ticket or discussion surrounding this?

BenOsodrac: There was not, but we can certainly bring this to the rest of the team!

But with Ionic Theme we cannot use this function as it was, as it will return colors from the current md/ios Design System, not the new one. 

Besides, on the new Design System we have a much wider array of color options and the expectaction would be the developers could use the new colors we are offering with the tokens and new Design System on the color attribute on each component (only when using the new Ionic Theme).

Is there any other way you would see this being done? Or is there any impact that I'm missing for this?

thetaPC: I highly recommend that a spike is created first to determine why we need this, who is requesting this change, and any potential risks, etc. After the spike is completed, we can return to this PR if need be.

BenOsodrac: This makes sense to me, I created [this spike](https://outsystemsrd.atlassian.net/browse/ROU-10778) to tackle it. 

Thanks for the feedback 👍

PR #29561, opened 2024-05-29
refactor(button): use the new tokens

BenOsodrac:
Issue number: Internal

---------

## What is the new behavior?

- On ionic.button.scss all dependencies to button.scss were removed, to avoid the old native theme values being mixed on the new ionic theme button.
- ionic.button.vars.scss partial was removed
- All values were changed to use tokens.
- Updated snapshots that bring slight differences due to now we are totally using the ionic theme tokens.
- Tests with colors where the ionic button was being added also now have differences, as the current ion-colors aren't affecting the ionic theme button. This is expected, as the ionic theme should only work with the colors from the new Design System, which is something being tacked on this [PR](https://github.com/ionic-team/ionic-framework/pull/29557). in the future, we should remove the ionic button from these color ctests and create new ones that add the new set of colors to the ionic button.

- [ ] Yes
- [x] No


thetaPC:
@BenOsodrac this would be considered a `refactor` instead of a a `feat` since we are not introducing nothing new.

core/src/components/button/button.ionic.scss
```diff
@@ -1,32 +1,55 @@
 @use "../../themes/ionic/ionic.globals.scss" as globals;
-@import "./button";
```
thetaPC, line 2: Please revert this. We want to be consistent, we import the base styles to all components for all themes. If we want to avoid values being mixed, then we would use the variables to override any values as shown in other themes.

core/src/components/button/button.ionic.scss
```diff
@@ -1,32 +1,55 @@
...
-@import "./button.ionic.vars";
 
 // Ionic Button
 // -------------------------------------------------------------------------------
-
 :host {
+  --overflow: hidden;
```
thetaPC, line 6: Once we revert the import removal, then these duplicate variables should be removed as well.

core/src/components/button/button.ionic.vars.scss
```diff
@@ -1,125 +0,0 @@
-@import "../../themes/native/native.globals.ios";
```
thetaPC, line 1: Why was this file removed?

thetaPC (changes requested):

core/src/components/button/button.ionic.scss
BenOsodrac: Consistency is the reason I did this change. We were mixing up not only themes, but Design Systems. 

The problem here is that the current base styles (button.scss) are based on the current Design System on Ionic Framework, that supports two themes: ios/md.

We are introducing a new Design System, and a new theme (Ionic Theme) to the Framework. And I believe we should aim to keep them separated as much as possible, until its possible to join them in the future when/if we decide to update the md/ios for the new Design System (in the meantime, it might mean a bit of duplicate code on the framework, but the output for CSS for each theme will be cleaner).

I thought we were all on board with this, as it was discussed/reviewed and accepted on this [CSS Architecture Design Doc](https://outsystemsrd.atlassian.net/wiki/spaces/RDMBVS/pages/3982655893/CSS+Architecture+-+New+Design+System), particularly [this page about scss](https://outsystemsrd.atlassian.net/wiki/spaces/RDMBVS/pages/3991109633/SCSS+Architecture).

My PR goes in the direction of meeting the architecture represented on this diagram:
image

But if you're not comfortable with it and think we are doing something wrong here, I would say we can schedule a team meeting to discuss this general architecture stuff., so that everyone as a more active way to give their insights about it. After that we can come back to this PR, and change it, accept it or just close it 👍

core/src/components/button/button.ionic.vars.scss
BenOsodrac: Besides, is importing stuff from a different Design System and Theme (ios).

thetaPC (approved):
LGTM

PR #29564, opened 2024-05-29
refactor(checkbox): use the new tokens

BenOsodrac:
Issue number: internal

---------

## What is the new behavior?

- updated ionic.checkbox.scss with ionic tokens and removed references to checkbox.scss partial
- removed ionic.checkbox.vars.scss file
- update snapshots resulted from slight differences due to the use of the updated tokens.
- added theme-ionic html file to test basic use-cases with ionic theme.
- border-radius now won't auto scale with `--size`. This was not possible to do using the gloabls tokens variables and the css calc(), as it was before. Besides, this is not something we are very interested to offer on Ionic theme. It should be on developer to also update the `--border-radius` variable if they want, after changing `--size`.

- [ ] Yes
- [x] No


thetaPC:
@BenOsodrac this would be considered a `refactor` instead of a a `feat` since we are not introducing nothing new. I also recommend to update the PR title to `refactor(checkbox): use the new tokens`

core/src/components/checkbox/checkbox.ionic.scss
```diff
@@ -1,34 +1,46 @@
-@import "./checkbox";
```
thetaPC, line 1: Please revert this. We want to be consistent, we import the base styles to all components for all themes. If we want to avoid values being mixed, then we would use the variables to override any values as shown in other themes.

core/src/components/checkbox/checkbox.ionic.vars.scss
thetaPC, line 1: Why was this file removed?

thetaPC (changes requested):

core/src/components/checkbox/checkbox.ionic.scss
BenOsodrac: See my answer [here](https://github.com/ionic-team/ionic-framework/pull/29561#discussion_r1622030340), please.

We can leave this PR in stand by for now :)

core/src/components/checkbox/checkbox.ionic.vars.scss
BenOsodrac: As all values for the Ionic Theme come from the global tokens, I see no reason to have a separate partial with some variables, when even some of them are not reused more than once.

BenOsodrac: Besides, is importing stuff from a different Design System and Theme (ios)

thetaPC (changes requested):

core/src/components/checkbox/checkbox.ionic.scss
```diff
@@ -44,94 +56,336 @@
...
-  /* TODO(FW-6183): Use design token variables */
-  background-color: #105ed1; // mix of '#1068eb', '#121212', 0.12, 'rgb'
+  background-color: globals.$ionic-color-primary-900;
 }
 
 // Ionic Design Checkbox Shapes
 // --------------------------------------------------
-:host(.checkbox-shape-soft) {
-  --border-radius: #{$checkbox-ionic-border-radius};
-}
```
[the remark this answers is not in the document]
BenOsodrac: Commited!

BenOsodrac:
@thetaPC For some reason I'm not able to update snapshots today using Docker, it seems to stuck at the end and the e2e tests locally detect no differences 🤔. Can you check on your side please?

core/src/components/checkbox/checkbox.ionic.scss
```diff
@@ -1,34 +1,46 @@
...
 // Ionic Checkbox
 // --------------------------------------------------
 
 :host {
   // Border
-  --border-radius: #{$checkbox-ionic-border-radius};
-  --border-width: #{$checkbox-ionic-border-width};
-  --border-style: #{$checkbox-ionic-border-style};
-  --border-color: #{$checkbox-ionic-background-color-off};
+  --border-radius: #{globals.$ionic-border-radius-100};
```
thetaPC, line 8: Let's remove this since the default soft has been added back.

thetaPC (changes requested):
One last thing that I missed. I'll look into getting the snapshots updated.

core/src/components/checkbox/checkbox.ionic.scss
```diff
@@ -44,94 +56,340 @@
...
 :host(.checkbox-size-small) {
-  // Size
-  --size: #{$checkbox-ionic-small-size};
+  --border-radius: #{globals.$ionic-border-radius-050};
+  --size: #{globals.$ionic-scale-400};
+}
+
+:host(.checkbox-size-small.checkbox-shape-rectangular) {
+  --border-radius: #{globals.$ionic-border-radius-0};
+}
```
thetaPC, line 295: This is not necessary because the border radius will always be 0 regardless of the size. So the border will be set correctly with just `.checkbox-shape-rectangular`.

```suggestion
```

core/src/components/checkbox/checkbox.ionic.scss
```diff
@@ -44,94 +56,340 @@
...
+// :host(.checkbox-label-placement-stacked.checkbox-alignment-center) .label-text-wrapper {
+//   @include globals.transform-origin(center, top);
+// }
+
 // Ionic Design Checkbox Sizes
 // --------------------------------------------------
 :host(.checkbox-size-small) {
-  // Size
-  --size: #{$checkbox-ionic-small-size};
+  --border-radius: #{globals.$ionic-border-radius-050};
```
thetaPC, line 289: Let's remove `border-radius` line here and move it below. This would follow the structure that's been set, good example is `badge.ionic.scss` with how it sets shape soft and sizes.

```suggestion
```

core/src/components/checkbox/checkbox.ionic.scss
```diff
@@ -44,94 +56,340 @@
...
-  background-color: #105ed1; // mix of '#1068eb', '#121212', 0.12, 'rgb'
+  background-color: globals.$ionic-color-primary-900;
 }
 
 // Ionic Design Checkbox Shapes
 // --------------------------------------------------
 :host(.checkbox-shape-soft) {
-  --border-radius: #{$checkbox-ionic-border-radius};
+  --border-radius: #{globals.$ionic-border-radius-100};
 }
```
thetaPC, line 388: This is where we would override the border radius for soft and small.

```suggestion

:host(.checkbox-size-small.checkbox-shape-soft) {
  --border-radius: #{globals.$ionic-border-radius-050};
}

```

thetaPC (changes requested):
When running Docker tests locally, I noticed some inconsistencies for border radius. The requested changes should align the code to the current structure and fix those inconsistencies. 

Let me know if Docker is still not working for you after adding these changes.

core/src/components/checkbox/checkbox.ionic.scss
BenOsodrac: removed!

BenOsodrac:
@thetaPC all passed now 👍, but locally docker is still weird for me for some reason.

thetaPC (approved):
LGTM

PR #29565, opened 2024-05-29
refactor(chip): use the new tokens

BenOsodrac:
Issue number: internal

---------

## What is the new behavior?

- Replaced old fixed values with new tokens.
- Added theme="ionic" to ionic test page.

- [ ] Yes
- [x] No


thetaPC:
@BenOsodrac this would be considered a `refactor` instead of a a `feat` since we are not introducing nothing new. I also recommend to update the PR title to `refactor(chip): use the new tokens`

thetaPC (approved):
LGTM

PR #29583, opened 2024-06-04
feat(item): add styles for disabled state in ionic theme

rugoncalves:
Issue number: internal

---------

## What is the new behavior?
- Added scss for the disabled state of `ion-item` for the Ionic theme
- Added disabled elements to test page

- [ ] Yes
- [x] No


core/src/components/item/item.ionic.scss
```diff
@@ -88,3 +88,22 @@
...
+
+:host(.item-disabled) {
+  cursor: default;
+  pointer-events: none;
+
+  &::after {
+    @include globals.position-horizontal(0, 0);
+    position: absolute;
+    top: 0;
+    bottom: 0;
```
brandyscarney, line 103:
```suggestion
    @include globals.position(0, 0, 0, 0);

    position: absolute;
```

We can use the `position` mixin here since we are also setting top & bottom. 🙂

core/src/components/item/test/theme-ionic/index.html
brandyscarney, line 1: Do we plan to add e2e tests for this component later? We could probably just add the `ionic` theme here: https://github.com/ionic-team/ionic-framework/blob/next/core/src/components/item/test/disabled/item.e2e.ts

rugoncalves: @gnbm thoughts?

brandyscarney (approved):

core/src/components/item/test/theme-ionic/index.html
rugoncalves: Thank you @brandyscarney for the help on this. @gnbm automated screenshot tests are running on the ionic theme for the disabled state.

PR #29622, opened 2024-06-13
feat(list): add shapes

thetaPC:
Issue number: internal

---------

List does not have shapes.

## What is the new behavior?

- Added shapes (round, soft, and rectangular) for the ionic theme
- Set round as the default shape
- Added tests

- [ ] Yes
- [x] No

## Other information

[Preview](https://ionic-framework-git-rou-10781-ionic1.vercel.app/src/components/list/test/shape?ionic:theme=ionic)


joselrio (approved):

OS-giulianasilva (approved):

core/src/components/list/list.tsx
```diff
@@ -43,9 +54,27 @@ export class List implements ComponentInterface {
...
+    const theme = getIonTheme(this);
+    const { shape } = this;
+
+    // TODO(ROU-10831): Remove theme check when shapes are defined for all themes.
+    if (theme !== 'ionic') {
+      return undefined;
+    }
+
+    if (shape === undefined) {
+      return 'round';
```
OS-giulianasilva, line 67: The task's AC says that:

> The default value changes based on the property inset
> inset=false (default) → default value: Rectangular
> inset=true → default value: Round

Shouldn't this be tackled in this method?

thetaPC: Good catch! https://github.com/ionic-team/ionic-framework/pull/29622/commits/ede0fcecc189ce74cbd8d2a2597131607d53f1bc

thetaPC: @OS-giulianasilva ready for review

OS-giulianasilva (approved):

PR #29623, opened 2024-06-14
fix(checkbox): separate background from svg in order to use proper sizes for ionic theme

brandyscarney:
Issue number: internal

---------

The checkbox background color and svg are both set to the same size. 

## What is the new behavior?
Moves the background and border to the native wrapper so that we can change the svg size separately.

- [ ] Yes
- [x] No

## Other information

[Preview](https://ionic-framework-git-fix-checkbox-design-ionic1.vercel.app/src/components/checkbox/test/basic?ionic:theme=ionic)


core/src/components/checkbox/test/basic/checkbox.e2e.ts-snapshots/checkbox-size-ionic-md-ltr-light-Mobile-Chrome-linux.png
brandyscarney, line 1: This change is expected because `--size` only applies to the width of the background after separating the svg from it.

core/src/components/checkbox/checkbox.ionic.scss
```diff
@@ -319,17 +326,21 @@ input {
 
 :host(.checkbox-disabled) .native-wrapper:after {
   @include globals.disabled-state();
+  @include globals.position(unset, unset, unset, unset);
+
+  width: inherit;
+  height: inherit;
```
brandyscarney, line 332: These styles are required in order to properly cover the border of the checkbox element. When `top` and `left` are set, it is calculated from the [beginning of the content](https://www.w3.org/TR/CSS21/visudet.html#containing-block-details), inside of the border. The alternative solution I found was calculating the position based on the `--border-width` CSS variable:

```scss
@include globals.position(
    calc(-1 * var(--border-width)), 
    calc(-1 * var(--border-width)), 
    calc(-1 * var(--border-width)), 
    calc(-1 * var(--border-width))
);
```

Inheriting the width and height is a cleaner solution, in my opinion. I could update the `disabled-state` mixin to remove `position` and inherit `width` and `height` as well, but I wasn't sure if this would have any negative consequences on other components.

BenOsodrac: I think the inherit solution is much more clean yes 💪.

About the mixin, I'm not so sure, I think we will need the position on some cases.

For now is only being used on Checkbox, as other components were made before the mixin, but I did a quick test on the Chip and it works correctly by adding the mixin as it is now, but when I replaced the position with the width/height inherit, it didn't work, as the size was 0 🫠

core/src/components/checkbox/test/basic/checkbox.e2e.ts-snapshots/checkbox-size-ionic-md-ltr-light-Mobile-Chrome-linux.png
BenOsodrac, line 1: @brandyscarney I think these differences are not intended right? And a side-effect o my other comment about the checkmark-width

core/src/components/checkbox/checkbox.ionic.scss
```diff
@@ -8,7 +8,8 @@
   --border-width: #{globals.$ionic-border-size-025};
   --border-style: #{globals.$ionic-border-style-solid};
   --border-color: #{globals.$ionic-color-neutral-800};
-  --checkmark-width: 3;
+  --checkmark-width: #{globals.$ionic-scale-400};
```
BenOsodrac, line 11: I think as these variables are in px they won't work for the --checkmark-width. 

Maybe we can add something like [this](https://css-tricks.com/snippets/sass/strip-unit-function/)?

BenOsodrac (changes requested):

core/src/components/checkbox/checkbox.ionic.scss
brandyscarney: I moved it to apply to the svg `width` instead of the path here: https://github.com/ionic-team/ionic-framework/pull/29623/files#diff-796ffc72b6171910c25d5aad085bc245095c38f104d55e253e06282702205066R154

Before:

```scss
.checkbox-icon path {
  fill: none;
  stroke: var(--checkmark-color);
  stroke-width: var(--checkmark-width);

  opacity: 0;
}
```

After:

```scss
.checkbox-icon {
  width: var(--checkmark-width);
  height: var(--checkmark-height);
}

.checkbox-icon path {
  fill: var(--checkmark-color);

  opacity: 0;
}
```

Since the Phosphor Icon doesn't use `stroke` it needs to be applied to the svg itself. This is different than the `ios` and `md` themes, but I think this makes more sense. What do you think?

core/src/components/checkbox/checkbox.ionic.scss
brandyscarney: Hmm okay. We can just add this only for checkbox. I can add some comments as to why it is different than the mixin. Then we can look into if we need to change it later.

core/src/components/checkbox/checkbox.ionic.scss
BenOsodrac: That makes sense yes! 

But it still is not related to the `px` issue, right? Or am I missing something? 🤔 

Its causing this change in snapshot here: https://github.com/ionic-team/ionic-framework/pull/29623#discussion_r1662805252

core/src/components/checkbox/test/basic/checkbox.e2e.ts-snapshots/checkbox-size-ionic-md-ltr-light-Mobile-Chrome-linux.png
brandyscarney: Oh no this is intentional because before the `--size` css variable was changing both the icon and the background on the same element but now they are separate elements. So the `--size` only sets the background and you have to set `--checkmark-width` to change the size of the icon. 

Before:

```scss
.checkbox-icon {
  width: var(--size);
  height: var(--size);
  
  background: var(--checkbox-background);
}

.checkbox-icon path {
  fill: none;
  stroke: var(--checkmark-color);
  stroke-width: var(--checkmark-width);
}
```

After:

```scss
.native-wrapper {
  width: var(--size);
  height: var(--size);

  background: var(--checkbox-background);
}

.checkbox-icon {
  position: relative;

  width: var(--checkmark-width);
  height: var(--checkmark-height);
}

.checkbox-icon path {
  fill: var(--checkmark-color);
}
```

We could update this though to something that makes more sense but I think we should keep the widths separate since the design tokens used in Figma are different:

```scss
--size: #{globals.$ionic-scale-600};
--checkmark-width: #{globals.$ionic-scale-400};
--checkmark-height: var(--checkmark-width);
```

BenOsodrac: Ah got it, sorry, that makes sense yes 👍

BenOsodrac (approved):

JoaoFerreira-FrontEnd (approved):

core/src/components/checkbox/checkbox.ionic.scss
brandyscarney: Added a comment explaining why we override the mixin: https://github.com/ionic-team/ionic-framework/pull/29623/commits/3bf7ea3f22dd6fbe98f942c21773d6ef6acc1d54

PR #29628, opened 2024-06-17
fix(typography): improve ionic global typography 

BenOsodrac:
Issue number: internal

---------

## What is the new behavior?

- Updated tokens utilities to no longer add font-family to utillity-classes. That is something that should only be set at a global layer, no sense to have that on a utility-class.
- Updated text token JSON as received by UX/UI Team.
- Improved tokens utilities to correctly set font-style to italic with italic font tokens.
- Adjusted typography Ionic scss to use the new typography classes and mixins based on tokens.
- Removed old basic typography test, as it was testing utility-classes that do not exist anymore.
- Added new typography test for ionic display tokens.
- Added a TODO to tackle the Inter font loading scenario, as it envolves research that impact other contexts as well.

- [x] Yes
- [ ] No

## Other information

[Sample Typography Screen](https://ionic-framework-e2le05h3a-ionic1.vercel.app/src/css/test/typography/basic)


OS-giulianasilva (approved):

core/scripts/tokens/utilities.js
thetaPC, line 1: In the future, we should consider unifying`generateTypographyValue()` and `generateTypographyUtilityClass()` since they are the same outcome. That way we don't have to make changes to both areas everytime.

thetaPC:
@BenOsodrac when possible please also add the sample screen in the PR description. I recommend putting it under "Other Information." If it's added, then it would speed up the review process since we don't need to look it up manually or through the Jira ticket.

packages/vue/test/base/tests/e2e/specs/inputs.cy.js
```diff
@@ -1,6 +1,6 @@
 describe('Inputs', () => {
   beforeEach(() => {
-    cy.visit('/inputs')
+    cy.visit('/inputs');
```
thetaPC, line 3: Revert this since this change is not part of the tokens work.
```suggestion
    cy.visit('/inputs')
```

thetaPC (changes requested):
Minor change request

core/scripts/tokens/utilities.js
BenOsodrac: Yes, I thought about that, but as they have slight differences on the outcome, the function might get a bit complex with all the exceptions

thetaPC: Understood, then maybe we should add a comment to remind devs that if a change happens in one function then the other function needs to be updated too.

thetaPC (approved):
LGTM

PR #29629, opened 2024-06-17
feat(ion-header): add basic UI for ionic theme

BenOsodrac:
Issue number: internal

---------

## What is the new behavior?

- Added new common header partial, to be used by ionic theme and ios/md
- Renamed header.scss to header.native.scss. This inherits the common created above and its used only on ios and md partial
- Created new partial for ionic theme.
- Added ionic theme to basic e2e tests.

- [ ] Yes
- [x] No

## Other information

[Sample Header Screen](https://ionic-framework-rcfex2fep-ionic1.vercel.app/src/components/header/test/basic?ionic:theme=ionic)


OS-giulianasilva (approved):

core/src/components/header/header.ionic.scss
```diff
@@ -0,0 +1,13 @@
+@use "./header.common";
+@use "../../themes/ionic/ionic.globals.scss" as globals;
+
+// Ionic Header
+// --------------------------------------------------
+
+ion-header {
+  --box-shadow: #{globals.$ionic-elevation-200};
+
+  box-shadow: var(--box-shadow);
```
thetaPC, line 10: Are we expecting the box shadow to change? If not, then let's follow the structure that is established for `md`:

```suggestion
  .header-ionic {
     box-shadow: #{globals.$ionic-elevation-200};
  }
```

core/src/components/header/test/basic/header.e2e.ts
```diff
@@ -1,7 +1,7 @@
 import { expect } from '@playwright/test';
 import { configs, test } from '@utils/test/playwright';
 
-configs().forEach(({ title, screenshot, config }) => {
+configs({ modes: ['ionic-md', 'md', 'ios'], directions: ['ltr'] }).forEach(({ title, screenshot, config }) => {
```
thetaPC, line 4: The box shadow isn't being shown. Let's update this test to capture that style:

```
await page.setContent(
  `
  <style>
    .container {
      background-color: #eaeaea;
      padding: 10px;
    }
  </style>
v
  <div class="container">
    <ion-header>
      <ion-toolbar>
        <ion-title>Header - Default</ion-title>
      </ion-toolbar>
    </ion-header>
  </div>
`,
  config
);

const container = page.locator('.container');
await expect(container).toHaveScreenshot(screenshot(`header-diff`));
```

thetaPC (changes requested):

core/src/components/header/header.ionic.scss
```diff
@@ -0,0 +1,13 @@
...
+
+// Ionic Header
+// --------------------------------------------------
+
+ion-header {
+  z-index: 10;
+
+  .header-ionic {
+    box-shadow: #{globals.$ionic-elevation-200};
+  }
```
thetaPC, line 12: This needs to be moved outside of `ion-header` for it to work. By placing it inside, it targets `ion-header > .header-ionic`.

thetaPC (changes requested):

core/src/components/header/header.ionic.scss
BenOsodrac: good catch, thanks 👍

thetaPC (approved):
LGTM, also thank you for separating the files into native and common!!!

The snapshots for the ionic theme had to be regenerated. There are moments that the pixel difference is so small that Playwright doesn't recognize the change. In this case, it was the box shadow. I regenerated the snapshots for you so you're not waiting on an approval while I'm out.

BenOsodrac:
> LGTM, also thank you for separating the files into native and common!!!
> 
> The snapshots for the ionic theme had to be regenerated. There are moments that the pixel difference is so small that Playwright doesn't recognize the change. In this case, it was the box shadow. I regenerated the snapshots for you so you're not waiting on an approval while I'm out.

Ah thank you! 🖤 

No prob, separating the files does result in cleaner code no doubt :)

PR #29642, opened 2024-06-20
feat(item-option): add disabled styles

thetaPC:
Issue number: internal

---------

The ionic theme does not have styles for a disabled `item-option`.

## What is the new behavior?

- Added the disable styles for `item-option` when using the ionic theme
- Added tests

- [ ] Yes
- [x] No

## Other information

[Preview](https://ionic-framework-git-rou-10784-ionic1.vercel.app/src/components/item-sliding/test/states?ionic:theme=ionic)


core/src/components/item-option/item-option.ionic.scss
```diff
@@ -63,3 +63,15 @@
 :host(.item-option-rectangular) {
   @include globals.border-radius(globals.$ionic-border-radius-0);
 }
+
+// Item Option States
+// --------------------------------------------------
+
+/* Disabled */
+:host(.item-option-disabled) {
+  opacity: 0.6;
```
BenOsodrac, line 72: We have a mixin for the disabled status, I think it works for this use-case as well: https://github.com/ionic-team/ionic-framework/blob/3d94b234a0c43e8bd68db0cbcd2b39ebf6f0fb79/core/src/themes/ionic/ionic.mixins.scss#L15

BenOsodrac (changes requested):

core/src/components/item-option/item-option.ionic.scss
thetaPC: @BenOsodrac this mixin wouldn't work in this case. Figma desisns are showing that the disabled state for the `item-option` only have a `opacity` style change. If we use `disabled-state()` then it won't match Figma because this mixin also changes the background color.

PR #29666, opened 2024-06-26
chore(deps): update dependency @stencil/core to v4.19.0

brandyscarney:
### Release Notes

<details>
<summary>ionic-team/stencil (@&#8203;stencil/core)</summary>

### [`v4.19.0`](https://togithub.com/ionic-team/stencil/blob/HEAD/CHANGELOG.md#-4190-2024-06-26)

[Compare Source](https://togithub.com/ionic-team/stencil/compare/v4.18.3...v4.19.0)

### Bug Fixes

* **compiler:** support rollup's external input option ([#3227](https://github.com/ionic-team/stencil/issues/3227)) ([2c68849](https://github.com/ionic-team/stencil/commit/2c6884970baf9f01f36d0843ce4ad59745e5a1f0)), fixes [#3226](https://github.com/ionic-team/stencil/issues/3226)
* **emit:** don't emit test files ([#5789](https://github.com/ionic-team/stencil/issues/5789)) ([50892f1](https://github.com/ionic-team/stencil/commit/50892f153c4c95e2728ecc460c87582fcd763a1e)), fixes [#5788](https://github.com/ionic-team/stencil/issues/5788)
* **hyrdate:** support vdom annotation in nested dsd structures ([#5856](https://github.com/ionic-team/stencil/issues/5856)) ([61bb5e3](https://github.com/ionic-team/stencil/commit/61bb5e3a080c011fb3242c0428cad9238b43149d))
* label attribute not toggling input ([#3474](https://github.com/ionic-team/stencil/issues/3474)) ([13db920](https://github.com/ionic-team/stencil/commit/13db92075b8dec53f5226761cec5ace5edb73d0c)), fixes [#3473](https://github.com/ionic-team/stencil/issues/3473)
* **mock-doc:** expose ShadowRoot and DocumentFragment globals ([#5827](https://github.com/ionic-team/stencil/issues/5827)) ([98bbd7c](https://github.com/ionic-team/stencil/commit/98bbd7c0d6fb67f085aa9ce0c3013e942c882be2)), fixes [#3260](https://github.com/ionic-team/stencil/issues/3260)
* **runtime:** allow watchers to fire w/ no Stencil members ([#5855](https://github.com/ionic-team/stencil/issues/5855)) ([850ad4f](https://github.com/ionic-team/stencil/commit/850ad4f4dd7c2349109be987af1e6f5df8c39608)), fixes [#5854](https://github.com/ionic-team/stencil/issues/5854)
[...]

PR #29670, opened 2024-06-27
chore(angular-server): add sync github action and update @stencil/core to 4.19.1 in core

brandyscarney:
Issue number: internal

---------

This PR does a couple things:

**1 - Adds a `sync` command to get the latest `core` build in `angular-server`**

The release process for `angular-server` failed [here](https://github.com/ionic-team/ionic-framework/actions/runs/9686982182/job/26730689874). This failure was only made apparent because the Lerna command [here](https://github.com/ionic-team/ionic-framework/blob/52ff0505e86dc204cb2fd2fdaf67229e6eeb36f8/.github/workflows/actions/publish-npm/action.yml#L35) runs prior to building each package. 

This should have been caught by CI on the [update to @stencil/core to v4.19.0](https://github.com/ionic-team/ionic-framework/pull/29666), but the `angular-server` package is the only package that doesn't sync `core` before it builds. This PR adds a `sync` command so that `angular-server` will use the latest core build during the normal build action. 

**2 - Resolving types errors in `angular-server`**

After properly syncing `core` in `angular-server` using the command added, running build fails with the following:

Screenshot 2024-06-27 at 1 16 15 PM

This was a regression in Stencil. `@stencil/core` has been updated to resolve these errors so the build passes.

### DIFF
--- a/.github/workflows/actions/build-angular-server/action.yml
+++ b/.github/workflows/actions/build-angular-server/action.yml
@@ -11,8 +11,8 @@ runs:
       run: npm ci
       shell: bash
       working-directory: ./packages/angular-server
-    - name: Install Dependencies
-      run: npx lerna@5 bootstrap --include-dependencies --ignore-scripts -- --legacy-peer-deps
+    - name: Sync
+      run: npm run sync
       shell: bash
       working-directory: ./packages/angular-server
     - name: Build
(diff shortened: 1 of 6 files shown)


sean-perkins (approved):
Looks good, tested against https://github.com/sean-perkins/ionic-angular-universal-example

PR #29675, opened 2024-06-28
feat(header): add new divider property

BenOsodrac:
Issue number: internal

---------

## What is the new behavior?

- Add new divider property on header's Ionic theme
- Added new divider test.
- Added TODO for support on ios/ md in future.

- [ ] Yes
- [x] No

## Other information

[Header divider sample](https://ionic-framework-a13vvxiyy-ionic1.vercel.app/src/components/header/test/divider?ionic:theme=ionic)


core/src/components/header/test/divider/index.html
```diff
@@ -0,0 +1,28 @@
+<!DOCTYPE html>
+<html lang="en" dir="ltr">
+  <head>
+    <meta charset="UTF-8" />
+    <title>Header - Basic</title>
```
OS-giulianasilva, line 5:
```suggestion
    <title>Header - Divider</title>
```

OS-giulianasilva (approved):

core/src/components/header/header.tsx
```diff
@@ -46,6 +46,12 @@ export class Header implements ComponentInterface {
    */
   @Prop() collapse?: 'condense' | 'fade';
 
+  /**
+   * If `true`, the header will have a line at the bottom.
+   * TODO(ROU-10855): add support for this prop on ios/md themes
+   */
+  @Prop({ reflect: true }) divider = false;
```
brandyscarney, line 53: Do we need this as a property? We have support for a border in the other themes by applying it by default in CSS. If a user wants to remove it, they can add the `ion-no-border` CSS class to `ion-header`. Here is how it looks for `ios` (it only applies the border to the last toolbar in a header):

```scss
// header.ios.scss

.header-ios ion-toolbar:last-of-type {
  --border-width: 0 0 #{$hairlines-width};
}

.header-ios.ion-no-border ion-toolbar:last-of-type {
  --border-width: 0;
}
```

BenOsodrac: I understand we can do it with CSS + utilities, but it was followed what was defined by the UX/UI team to have this property. And as this has small impact I think its a good opportunity to make the experience more similar between our different contexts.

But we can bring this to the rest of the team, if you feel this isn't really the best approach!

core/src/components/header/test/divider/header.e2e.ts
```diff
@@ -0,0 +1,14 @@
+import { expect } from '@playwright/test';
+import { configs, test } from '@utils/test/playwright';
+
+configs({ modes: ['ionic-md'], directions: ['ltr'] }).forEach(({ title, screenshot, config }) => {
+  test.describe(title('header: divider'), () => {
+    test('should not have visual regressions with divider header', async ({ page }) => {
+      await page.goto(`/src/components/header/test/divider`, config);
+
+      await page.setIonViewport();
```
sean-perkins, line 9: Capturing full page screenshots will slow CI performance over time. Since the screenshot is only for the header, you could use `.setContent` of the header with a bit of padding to include the divider/shadow.

core/src/components/header/header.tsx
```diff
@@ -46,6 +46,12 @@ export class Header implements ComponentInterface {
    */
   @Prop() collapse?: 'condense' | 'fade';
 
+  /**
+   * If `true`, the header will have a line at the bottom.
+   * TODO(ROU-10855): add support for this prop on ios/md themes
+   */
```
sean-perkins, line 52: Since you are reflecting a class based on the prop and using that as the CSS selector, the attribute does not need to be reflected.

core/src/components/header/test/divider/header.e2e.ts
BenOsodrac: Thanks, always forget about that 😂

brandyscarney (approved):
Looks good! 🙂

sean-perkins (approved):

PR #29684, opened 2024-07-01
feat(tokens): update primitive json

BenOsodrac:
Issue number: internal

---------

## What is the new behavior?

- Update primitive json tokens
- Update generated files on foundations folder.

- [ ] Yes
- [x] No

## Other Information
[Tokens Preview Sample](https://ionic-framework-fugbxhshx-ionic1.vercel.app/src/foundations/tokens.preview.html)


core/src/foundations/ionic.root.scss
```diff
@@ -59,6 +59,7 @@
   --ionic-space-600: 24px;
   --ionic-space-700: 28px;
   --ionic-space-800: 32px;
+  --ionic-space-850: 34px;
```
brandyscarney, line 62: Is there a reason we are adding 850 but not 750, 650, etc? Are we just adding tokens as we need them?

brandyscarney: I spoke with Beth and she said we can review all of the tokens once they are implemented. 👍

core/src/foundations/tokens.preview.html
```diff
@@ -1725,6 +1727,10 @@ <h1>Space Tokens</h1>
             <div class="space-token" style="margin: 32px">-space-800 (32px)</div>
           </div>
 
+          <div class="spacing-wrapper">
+            <div class="space-token" style="margin: 34px">-space-850 (34px)</div>
```
brandyscarney, line 1731: Why don't these use the classes like `.ionic-margin-space-850`?

brandyscarney (approved):

core/src/foundations/tokens.preview.html
BenOsodrac: At the time I wanted to have this file abstract of the other ones, to be as `pure` as possible. 
But it makes sense, I think we can do that in a future improvement to this 👍

brandyscarney: That's fine with me!

PR #29691, opened 2024-07-05
refactor(item): use correct button sizes

thetaPC:
Issue number: internal

---------

Ionic theme: Buttons within a list do not respect their given sizes.

## What is the new behavior?

- Split the styles into files
- Updated the disabled state to match Figma
- Buttons render with the correct sizes

- [ ] Yes
- [x] No

## Other information

[Preview](https://ionic-framework-git-rou-10850-ionic1.vercel.app/src/components/item/test/buttons?ionic:theme=ionic)


OS-giulianasilva (approved):

core/src/components/item/item.ionic.scss
```diff
@@ -1,121 +1,66 @@
...
-    @include globals.position(0, 0, 0, 0);
-
-    position: absolute;
 
-    background: #{globals.$ionic-state-disabled};
-
-    content: "";
-  }
+:host(.item-disabled) {
+  opacity: 0.6;
```
BenOsodrac, line 46: Can't we use `:host(.item-disabled) {@include globals.disabled-state();}` here?

BenOsodrac (changes requested):

core/src/components/item/item.ionic.scss
thetaPC: @BenOsodrac the disabled state on Figma only has a opacity change. `disabled-state()` changes the background color, which doesn't match Figma.

BenOsodrac: But it is a rgba one from the tokens: `rgba(255, 255, 255, 0.6))` which matches the #ffffff hex color on figma that is also using the disabled token:

Screenshot 2024-07-09 at 09 06 47

thetaPC: I looked into this more. The `disabled-state()` mixin won’t work because it [targets](https://github.com/ionic-team/ionic-framework/blob/c48b33b360c57f32d756a4cd496382a441170653/core/src/themes/ionic/ionic.mixins.scss#L20) `background-color` but `item` uses a variable, `--background`, to [handle](https://github.com/ionic-team/ionic-framework/blob/70bd41f65d2fd2bcc6e0671c2b71106edd98756a/core/src/components/item/item.common.scss#L192) it. By setting, `background-color`, it never applies the background color correctly.

We could set `--background: rgba(255, 255, 255, 0.6)` but this also doesn’t work because it changes the background color for the `.item-native` [child element](https://github.com/ionic-team/ionic-framework/blob/70bd41f65d2fd2bcc6e0671c2b71106edd98756a/core/src/components/item/item.common.scss#L157). The opacity needs to be set on the `ion-item`. The only reason that it looks like it’s working is because opacity is already being [set](https://github.com/ionic-team/ionic-framework/blob/70bd41f65d2fd2bcc6e0671c2b71106edd98756a/core/src/components/item/item.common.scss#L150) through `item.common`.

Screenshot 2024-07-09 at 8 39 21 AM

BenOsodrac: Alright, I'm convinced! :)

Thanks for looking into, maybe once we gather more different implementations of disabled states we can see it better how to improve the mixin or create a new one 👍

BenOsodrac (approved):

PR #29702, opened 2024-07-09
test(button): add ionic mode to outline

thetaPC:
Issue number: internal

---------

The outline tests for button does not include the ionic theme.

## What is the new behavior?

- Added ionic theme to the outline test

- [ ] Yes
- [x] No


core/src/components/button/test/outline/button.e2e.ts-snapshots/button-outline-ionic-md-ltr-light-Mobile-Chrome-linux.png
thetaPC, line 1: The test looks plain because the colors aren't implemented. This will be done in another PR.

Issue #28574, opened 2023-11-23
bug: All Angular standalone components are not tree shaken from the main JS chunk


StefanNedelchev:
### Prerequisites

- [X] I have read the [Contributing Guidelines](https://github.com/ionic-team/ionic-framework/blob/main/.github/CONTRIBUTING.md#creating-an-issue).
- [X] I agree to follow the [Code of Conduct](https://ionicframework.com/code-of-conduct).
- [X] I have searched for [existing issues](https://github.com/ionic-team/ionic-framework/issues) that already report this problem, without success.

### Ionic Framework Version

v7.x

### Current Behavior

I tried migrating my current app to Angular 17 and then migrating to Standalone Ionic components. I followed the official guideline and used the automated tool for help. Then I found out that the main JS chunk suddenly became ~1.3MB (compared to the previous ~600kb). 

All my modules and pages are lazy loaded and this was a surprise for me. I double checked that all instances of `IonicModule` are gone and all imports come from the `/standalone` path. I also double checked that I only import what is needed and where it's needed. Everything seems to be fine. The app is a mixture between modules and standalone components (if this matters).

After checking the bundle that's what I found:

![Screenshot from 2023-11-23 21-41-39](https://github.com/ionic-team/ionic-framework/assets/15238282/37ef7617-5677-4dea-9c36-3071863252e0)

As you can see - the main chink is full of Ionic components which are not tree-shaken.

Actually I could also reproduce this in a clean Ionic standalone project with the latest CLI:

![Screenshot from 2023-11-23 21-58-55](https://github.com/ionic-team/ionic-framework/assets/15238282/27895496-3303-445e-89a6-e4f2aa7a56ff)

The issue happens with both webpack and esbuild.

### Expected Behavior

I would expect all components to be tree-shaken and bundled in their respective places away from main JS.

### Steps to Reproduce

1. Generate a fresh Ionic project with the latest CLI - Chose Angular and Standalone. For template chose the one with the side menu (because it has an additional lazy loaded module).
2. Edit the App component by removing all Ionic components from the template, leaving the router outlet only
3. Remove the imports of these components from the TS file too
4. Do a production build

### Ionic Info

Ionic:

   Ionic CLI                     : 7.1.5 (/usr/lib/node_modules/@ionic/cli)
   Ionic Framework               : @ionic/angular 7.5.6
   @angular-devkit/build-angular : 17.0.3
   @angular-devkit/schematics    : 17.0.3
[...]

PR #29717, opened 2024-07-16
fix(core): add an export map to import individual components

### LINKED ISSUE #29716: feat: update `moduleResolution` from `node` to `node16` or `bundler`
### Prerequisites

Currently Ionic framework (`@ionic/core`) uses `node` as `moduleResolution` which is meant to be used with Node v10 or older. Modern code bases should use something like `node16` or `bundler` to support export maps and other ESM features.

### Describe the Use Case

I've started to work on #29715 which requires `node16` due to the fact that we modernized the React Output Target code base.

- set `"moduleResolution": "bundler"` in `/core/tsconfig.json`
- create an export map to export each component in `/core/package.json`

### Describe Alternatives

n/a

### Related Code

n/a

### Additional Information

n/a

christian-bromann:
The core package doesn't provide an export map, which prevents modern code bases to import specific components.

fixes #29716

## What is the new behavior?
This patch contains:

- a switch from `node` to `bundler` in the `moduleResolution field, both resolutions are very similar - changing this has no impact on the user as it is only for internal development
- adding export entries for individual components to allow packages using modern module resolutions to import individual components

- [ ] Yes
- [x] No

This should not impact users at all but enable them to use a different module resolution. However, I may not be too familiar how Ionic components are imported, therefor it would be good to do some thorough testing.

## Other information

n/a

### DIFF
--- a/core/package.json
+++ b/core/package.json
@@ -23,6 +23,285 @@
   "collection:main": "dist/collection/index.js",
   "collection": "dist/collection/collection-manifest.json",
   "types": "dist/types/interface.d.ts",
+  "exports": {
+    ".": {
+      "import": "./dist/index.js",
+      "require": "./dist/index.cjs.js",
+      "types": "./dist/types/interface.d.ts"
+    },
+    "./ion-accordion-group": {
+      "import": "./components/ion-accordion-group.js",
+      "types": "./components/ion-accordion-group.d.ts"
+    },
+    "./ion-accordion": {
+      "import": "./components/ion-accordion.js",
+      "types": "./components/ion-accordion.d.ts"
+    },
+    "./ion-avatar": {
+      "import": "./components/ion-avatar.js",
+      "types": "./components/ion-avatar.d.ts"
+    },
+    "./ion-backdrop": {
+      "import": "./components/ion-backdrop.js",
+      "types": "./components/ion-backdrop.d.ts"
+    },
+    "./ion-badge": {
+      "import": "./components/ion-badge.js",
+      "types": "./components/ion-badge.d.ts"
+    },
+    "./ion-breadcrumbs": {
+      "import": "./components/ion-breadcrumbs.js",
+      "types": "./components/ion-breadcrumbs.d.ts"
+    },
+    "./ion-buttons": {
+      "import": "./components/ion-buttons.js",
+      "types": "./components/ion-buttons.d.ts"
+    },
+    "./ion-card-content": {
+      "import": "./components/ion-card-content.js",
+      "types": "./components/ion-card-content.d.ts"
+    },
+    "./ion-card-header": {
+      "import": "./components/ion-card-header.js",
+      "types": "./components/ion-card-header.d.ts"
+    },
+    "./ion-card-subtitle": {
+      "import": "./components/ion-card-subtitle.js",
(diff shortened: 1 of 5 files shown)


core/package.json
```diff
@@ -23,6 +23,285 @@
   "collection:main": "dist/collection/index.js",
   "collection": "dist/collection/collection-manifest.json",
   "types": "dist/types/interface.d.ts",
+  "exports": {
+    ".": {
+      "import": "./dist/index.js",
+      "require": "./dist/index.cjs.js",
+      "types": "./dist/types/interface.d.ts"
+    },
+    "./ion-accordion-group": {
```
sean-perkins, line 32: I'd recommend adding `/components/` to the path to minimize the changes to the Vue and React packages (and anyone consuming the custom elements bundles from `@ionic/core`).

core/package.json
```diff
@@ -23,6 +23,285 @@
   "collection:main": "dist/collection/index.js",
   "collection": "dist/collection/collection-manifest.json",
   "types": "dist/types/interface.d.ts",
+  "exports": {
+    ".": {
```
sean-perkins, line 27: We'll need to additionally export the `./dist/docs.json`.

e.g.:
```json
"./docs.json": {
      "import": "./dist/docs.json",
      "require": "./dist/docs.json"
    },
```

There are 3 scripts in the vue project that reference the `dist/docs.json` path that will need to be updated to remove `dist` from the path.

There is an error you will see locally with Vue that you can ignore for now. I am working with the team to make them aware (exists in `main`):
```
(!) Plugin typescript: @rollup/plugin-typescript TS2315: Type 'Plugin_2' is not generic.
src/ionic-vue.ts: (24:24)

24 export const IonicVue: Plugin<[IonicConfig?]> = {
```

christian-bromann:
Thanks @sean-perkins for the feedback, I applied suggested improvements.

sean-perkins:
@christian-bromann it looks like the `docs.json` still needs to be added to the `package.json` `exports` to resolve the Vue related build issues.

For the angular build issues/test issues, you should see this error:
```
Error: Module not found: Error: Package path ./loader
```

we should export the loader path as well for the lazy loaded distribution.

sean-perkins (approved):
Looks good to me 👍 

We should definitely get someone from the Ionic Framework team to look at this for visibility to the change/impacts. 

There is some risk that we are missing exporting an obscure build file that is not leveraged in Ionic Framework's packages/test apps, but a developer is making use of. I still think we could publish this as a patch or minor and if those cases are reported, we add the missing exports paths/evaluate what files are being consumed.

muuvmuuv:
Will this also fix the described code-splitting issue here https://github.com/ionic-team/ionic-framework/issues/28574 ? Sounds like with that change I am able to just use a component from one exported path instead of just `@ionic/angular/standalone`, right?

thetaPC:
Thanks for kicking this off, @christian-bromann. We're going to close this one, as both parts of #29716 are now covered by other work:

- The package exports map for importing individual components was added in #31180 (already merged).
- The moduleResolution switch from node to bundler is handled in #31280, alongside a broader TypeScript 6 modernization across all packages.

Both target the v9 release, so this PR is superseded by them. Closing here to consolidate the effort. Thanks again for getting this started!

Issue #29712, opened 2024-07-15
bug: Overlays override id assigned through htmlAttributes


sean-perkins:
### Prerequisites

- [X] I have read the [Contributing Guidelines](https://github.com/ionic-team/ionic-framework/blob/main/docs/CONTRIBUTING.md#creating-an-issue).
- [X] I agree to follow the [Code of Conduct](https://ionicframework.com/code-of-conduct).
- [X] I have searched for [existing issues](https://github.com/ionic-team/ionic-framework/issues) that already report this problem, without success.

### Ionic Framework Version

v8.x

### Current Behavior

Split from: https://github.com/ionic-team/ionic-framework/issues/29704

In Ionic v8, overlays are overriding the `id` specified in the `htmlAttributes` with the incremental overlay ID assigned by Ionic Framework. 

Applies to: modal, popover, action-sheet, loading, picker-legacy and toast.

### Expected Behavior

Ionic's overlay should not assign the incremental overlay ID if the overlay already has an ID assigned to it by either the `id` attribute directly or the `htmlAttributes`. 

### Steps to Reproduce

1. Create a new overlay, giving it a custom ID through `htmlAttributes`
2. Observe that the overlay does not have the custom ID

This applies to all overlays **except** `ion-alert`, which has been resolved: https://github.com/ionic-team/ionic-framework/pull/29708

For the reproduction URL we can either use the original reproduction and replace the alert with another overlay or assign `htmlAttributes` with an `id` property to any other overlay to observe the problem. 

### Code Reproduction URL

N/A

### Ionic Info

N/A - reproduces in Ionic v8+.

mikelhamer:
I'm going to pick this up and start work!

PR #29722, opened 2024-07-19
fix(overlays): do not overwrite id set in htmlAttributes

mikelhamer:
Issue number: resolves # 29712

---------

In every type of overlay,  the auto incremented overlay id is overwriting any id set in htmlAttributes.

## What is the new behavior?
The id in htmlAttributes now takes precedence. 

-
-
-

- [ ] Yes
- [x] No


mikelhamer:
@sean-perkins I've created this PR to fix #29712 , but something I noticed is that the overlay ID is being used in the picker-legacy component, but not in the [the new picker component](https://github.com/ionic-team/ionic-framework/blob/2bc3b1feae45b4b10beb343699947541da9dfbfe/core/src/components/picker/picker.tsx). 
Should overlay ID be added there as well, along with making sure it doesn't overwrite the id in htmlAttributes?

sean-perkins:
@mikelhamer great observation, you are correct in your current implementation pattern. `ion-picker-legacy` requires this fix, but `ion-picker` is not an overlay, it is optionally presented in one if a developer configures it with a popover or modal. I've updated the original issue description to correct this.

mikelhamer:
Awesome! The more you know... 🌈 ⭐️

sean-perkins (approved):
Awesome 👏 thanks for knocking these out!

brandyscarney (approved):
Thank you for the PR! Great work. 🎉

mikelhamer:
Thank you! 🚀

PR #29740, opened 2024-07-30
chore(playwright): add click method to promise assertion lint checks

sean-perkins:
Issue number: N/A

---------

Ionic Framework has a set of custom lint assertions to avoid creating flaky Playwright tests by forgetting to await a promise. However the `click` method was not included in the original list of methods to check. 

## What is the new behavior?

- Developers will receive a lint error when forgetting to await click methods from Playwright
- Resolves existing tests where this lint issue was present

- [ ] Yes
- [x] No


thetaPC (approved):
LGTM

PR #29744, opened 2024-08-01
fix(angular): remove the tabindex set by routerLink from Ionic components

### LINKED ISSUE #20632: bug: ion-button with [routerLink] will display a border if being activated
# Bug Report

**Ionic version:**

[x] **4.11.9**

**Current behavior:**
When activating an `<ion-button>` element with a `[routerLink]` directive being set, the button will display show a border.

image

**Expected behavior:**
No border should be displayed when a button gets activated.

**Steps to reproduce:**
- Use an empty template to create an `ion-button` like below.
- Either click the element to briefly see the border or cycle through the elements with tab

**Related code:**

```html
<ion-button expand="block" fill="clear" [routerLink]="['/tabs/tab2']">Some button</ion-button>
```

https://stackblitz.com/edit/ionic-v4-angular-tabs-9a3req

**Other information:**
I'm not sure whether buttons are supposed to work like this, but this does not happen with a `href` being set.
I would only expect such behaviour in a real browser app, where anchors are being colored accordingly - but not within a Hybrid App on iOS :)

**Ionic info:**

```

Ionic:

   Ionic CLI                     : 6.1.0 (/Users/Wilk/.nvm/versions/node/v10.16.3/lib/node_modules/@ionic/cli)
   Ionic Framework               : @ionic/angular 4.11.9
   @ionic/angular-toolkit        : 2.1.2

Cordova:

   Cordova Platforms : ios 5.1.1
   Cordova Plugins   : cordova-plugin-ionic-keyboard 2.2.0, cordova-plugin-ionic-webview 4.1.3, (and 4 other plugins)

Utility:

   cordova-res (update available: 0.9.0) : 0.8.1
   native-run                            : 0.3.0

System:

   ios-deploy : 1.9.4
   ios-sim    : 8.0.2
   NodeJS     : v10.16.3 (/Users/Wilk/.nvm/versions/node/v10.16.3/bin/node)
   npm        : 6.9.0
   OS         : macOS Catalina
   Xcode      : Xcode 11.3.1 Build version 11C504
```

brandyscarney:
Issue number: resolves #20632

---------

When using the `routerLink` directive in Angular, it automatically adds `tabindex="0"` to the element. This creates issues with Ionic components that render native button or anchor elements, as they have their own focus management. As a result, when navigating between list items with `routerLink` using the `Tab` key, you need to press the `Tab` key twice to move to the next item. This problem is illustrated in the following demo:

[Open in StackBlitz](https://stackblitz.com/edit/angular-blfa7h?file=src%2Fapp%2Fexample.component.html)

Related Angular issue: https://github.com/angular/angular/issues/28345

## What is the new behavior?
Updated our `RouterLinkDelegateDirective` to check if the element using `routerLink` is one of the following Ionic components: `ion-back-button`, `ion-breadcrumb`, `ion-button`, `ion-card`, `ion-fab-button`, `ion-item`, `ion-item-option`, `ion-menu-button`, `ion-segment-button`, or `ion-tab-button`. If so, it removes the `tabindex` attribute from the element. This allows these Ionic components to let the native button or anchor element handle the focus.

This solution is demonstrated in the following demo:

[Open in StackBlitz](https://stackblitz.com/edit/angular-blfa7h-svmguh?file=src%2Fapp%2Fexample.component.html)

> [!NOTE]
> I did not include the `ion-router-link` component in the list to remove `tabindex` because [the router link documentation](https://ionicframework.com/docs/api/router-link) does not recommend using it with Angular:
>> Note: this component should only be used with vanilla and Stencil JavaScript projects. For Angular projects, use an `<a>` and `routerLink` with the Angular router.

- [ ] Yes
- [x] No

## Other information

Dev build: `8.2.7-dev.11722448707.1e8c66e6`

### DIFF
--- a/packages/angular/common/src/directives/navigation/router-link-delegate.ts
+++ b/packages/angular/common/src/directives/navigation/router-link-delegate.ts
@@ -31,10 +31,43 @@ export class RouterLinkDelegateDirective implements OnInit, OnChanges {
 
   ngOnInit(): void {
     this.updateTargetUrlAndHref();
+    this.updateTabindex();
   }
 
   ngOnChanges(): void {
     this.updateTargetUrlAndHref();
+  }
+
+  /**
+   * The `tabindex` is set to `0` by default on the host element when
+   * the `routerLink` directive is used. This causes issues with Ionic
+   * components that wrap an `a` or `button` element, such as `ion-item`.
+   * See issue https://github.com/angular/angular/issues/28345
+   *
+   * This method removes the `tabindex` attribute from the host element
+   * to allow the Ionic component to manage the focus state correctly.
+   */
+  private updateTabindex() {
+    // Ionic components that render a native anchor or button element
+    const ionicComponents = [
+      'ION-BACK-BUTTON',
+      'ION-BREADCRUMB',
+      'ION-BUTTON',
+      'ION-CARD',
+      'ION-FAB-BUTTON',
+      'ION-ITEM',
+      'ION-ITEM-OPTION',
+      'ION-MENU-BUTTON',
+      'ION-SEGMENT-BUTTON',
+      'ION-TAB-BUTTON',
+    ];
+    const hostElement = this.elementRef.nativeElement;
+
+    if (ionicComponents.includes(hostElement.tagName)) {
+      if (hostElement.getAttribute('tabindex') === '0') {
+        hostElement.removeAttribute('tabindex');
+      }
+    }
   }
 
   private updateTargetUrlAndHref() {
--- a/packages/angular/test/base/e2e/src/standalone/router-link.spec.ts
+++ b/packages/angular/test/base/e2e/src/standalone/router-link.spec.ts
@@ -2,7 +2,7 @@ describe('RouterLink', () => {
   beforeEach(() => {
     cy.visit('/standalone/router-link');
   });
-  
+
   it('should mount the root component', () => {
     cy.ionPageVisible('app-router-link');
 
@@ -20,5 +20,13 @@ describe('RouterLink', () => {
     cy.get('button').contains('I\'m a button').click();
 
     cy.url().should('include', '/standalone/popover');
+  });
+
+  it('should have tabindex="0" with a native span', () => {
+    cy.get('span').should('have.attr', 'tabindex', '0');
+  });
+
+  it('should not have tabindex set with an ionic button', () => {
+    cy.get('ion-button').should('not.have.attr', 'tabindex');
   });
 });
--- a/packages/angular/test/base/e2e/src/standalone/router-link.spec.ts
+++ b/packages/angular/test/base/e2e/src/standalone/router-link.spec.ts
@@ -22,6 +22,11 @@ describe('RouterLink', () => {
     cy.url().should('include', '/standalone/popover');
   });
 
+  // Angular sets the `tabindex` to `"0"` on any element that uses
+  // the `routerLink` directive. Ionic removes the `tabindex` from
+  // components that wrap an `a` or `button` element, so we are
+  // checking here that it is only removed from Ionic components.
+  // https://github.com/ionic-team/ionic-framework/issues/20632
   it('should have tabindex="0" with a native span', () => {
     cy.get('span').should('have.attr', 'tabindex', '0');
   });
--- a/packages/angular/test/base/e2e/src/lazy/router-link.spec.ts
+++ b/packages/angular/test/base/e2e/src/lazy/router-link.spec.ts
@@ -123,6 +123,16 @@ describe('Router Link', () => {
       testBack();
     });
   });
+
+  describe('tabindex', () => {
+    it('should have tabindex="0" with a native span', () => {
+      cy.get('#span').should('have.attr', 'tabindex', '0');
+    });
+
+    it('should not have tabindex set with an ionic button', () => {
+      cy.get('#routerLink').should('not.have.attr', 'tabindex');
+    });
+  });
 });
 
 function testForward() {
--- a/packages/angular/test/base/e2e/src/lazy/router-link.spec.ts
+++ b/packages/angular/test/base/e2e/src/lazy/router-link.spec.ts
@@ -124,6 +124,11 @@ describe('Router Link', () => {
     });
   });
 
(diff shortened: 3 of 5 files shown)


packages/angular/common/src/directives/navigation/router-link-delegate.ts
```diff
@@ -31,12 +31,45 @@ export class RouterLinkDelegateDirective implements OnInit, OnChanges {
...
+   * components that wrap an `a` or `button` element, such as `ion-item`.
+   * See issue https://github.com/angular/angular/issues/28345
+   *
+   * This method removes the `tabindex` attribute from the host element
+   * to allow the Ionic component to manage the focus state correctly.
+   */
+  private updateTabindex() {
+    // Ionic components that render a native anchor or button element
+    const ionicComponents = [
+      'ION-BACK-BUTTON',
```
thetaPC, line 53: How should I test this one? It doesn't display when replacing `defaultHref` with `routerLink`.

packages/angular/common/src/directives/navigation/router-link-delegate.ts
```diff
@@ -31,12 +31,45 @@ export class RouterLinkDelegateDirective implements OnInit, OnChanges {
...
+      'ION-BREADCRUMB',
```
thetaPC, line 54: When testing it on standalone, the `ion-breadcrumb` becomes a string when using `routerLink` instead. This happens before the fix so I don't think it's related. If this isn't supposed to be happening, then we should make a new ticket to investigate this.

packages/angular/common/src/directives/navigation/router-link-delegate.ts
```diff
@@ -31,12 +31,45 @@ export class RouterLinkDelegateDirective implements OnInit, OnChanges {
...
+      'ION-BUTTON',
+      'ION-CARD',
```
thetaPC, line 56: When testing it on standalone, the `ion-card` does not become a link when using `routerLink` instead. This happens before the fix so I don't think it's related. If this isn't supposed to be happening, then we should make a new ticket to investigate this.

thetaPC (changes requested):

packages/angular/test/base/e2e/src/standalone/router-link.spec.ts
```diff
@@ -21,4 +21,12 @@ describe('RouterLink', () => {
...
     cy.url().should('include', '/standalone/popover');
   });
+
+  it('should have tabindex="0" with a native span', () => {
+    cy.get('span').should('have.attr', 'tabindex', '0');
+  });
+
+  it('should not have tabindex set with an ionic button', () => {
+    cy.get('ion-button').should('not.have.attr', 'tabindex');
+  });
```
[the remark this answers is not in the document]
brandyscarney: Added in https://github.com/ionic-team/ionic-framework/pull/29744/commits/aa31b62230115b5186c1e16a4f486768e0e22c67

packages/angular/common/src/directives/navigation/router-link-delegate.ts
```diff
@@ -31,12 +31,45 @@ export class RouterLinkDelegateDirective implements OnInit, OnChanges {
...
+      'ION-FAB-BUTTON',
+      'ION-ITEM',
+      'ION-ITEM-OPTION',
+      'ION-MENU-BUTTON',
+      'ION-SEGMENT-BUTTON',
+      'ION-TAB-BUTTON',
```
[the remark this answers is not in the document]
brandyscarney: Is this happening in `lazy` or `standalone` or both? I am not able to reproduce on `lazy` by adding the following to `packages/angular/test/base/src/app/lazy/router-link/router-link.component.html`:

```html
<ion-tab-bar>
  <ion-tab-button routerLink="/lazy/router-link-page">Tab</ion-tab-button>
  <ion-tab-button routerLink="/lazy/router-link-page">Tab</ion-tab-button>
  <ion-tab-button routerLink="/lazy/router-link-page">Tab</ion-tab-button>
</ion-tab-bar>
```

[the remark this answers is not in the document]
brandyscarney: Disregard, I was able to reproduce this in standalone.

packages/angular/common/src/directives/navigation/router-link-delegate.ts
brandyscarney: I created ticket `ROU-10993` to address this. I tested it out on `main` and the same issue occurs so I don't think it should block this PR.

packages/angular/common/src/directives/navigation/router-link-delegate.ts
brandyscarney: I created ticket `ROU-10993` to address this. I tested it out on `main` and the same issue occurs so I don't think it should block this PR.

packages/angular/common/src/directives/navigation/router-link-delegate.ts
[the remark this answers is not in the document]
brandyscarney: I created ticket `ROU-10993` to address this. I tested it out on `main` and the same issue occurs so I don't think it should block this PR.

packages/angular/common/src/directives/navigation/router-link-delegate.ts
brandyscarney: For `lazy`:

1. Update `packages/angular/test/base/src/app/lazy/router-link-page/router-link-page.component.html` to include `routerLink` on `ion-back-button`:
    ```html
    <ion-back-button routerLink="/lazy/router-link-page"></ion-back-button>
    ```
1. Rebuild / sync the app
1. Navigate to http://localhost:4200/lazy/router-link
1. Click on the first button `ion-button[routerLink]`
1. Attempt to tab to the `ion-back-button`
1. With my fix you will tab once to it, without it you will see a blue focus ring and have to double tab

For `standalone` you can add `routerLink` onto the back button at `packages/angular/test/base/src/app/standalone/back-button/back-button.component.html` and then navigate to http://localhost:4200/standalone/back-button but it doesn't seem to have `tabindex="0"` with or without the fix.

thetaPC (approved):
LGTM

PR #29748, opened 2024-08-05
fix(list): update styles causing memory leak

tanner-reits:
Issue number: resolves #

---------

## What is the new behavior?

-
-
-

- [ ] Yes
- [ ] No

### DIFF
--- a/core/src/components/list/list.ios.scss
+++ b/core/src/components/list/list.ios.scss
@@ -38,11 +38,6 @@
   --inner-border-width: 0;
 }
 
-.list-ios.list-inset + ion-list.list-inset {
-  @include margin(0, null, null, null);
-}
-
-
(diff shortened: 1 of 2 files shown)


brandyscarney (approved):
I verified that removing this selector does not affect inset lists due to the top and bottom margins overlapping. I reviewed the [original commit](https://github.com/ionic-team/ionic-framework/commit/c878289ed4486d5e06eabc42fb01f8ab198534a9#diff-bdc7d05ddb9690adad6a439216da72fa9332d183098f40e01aaac91d00815610R64) that added this and tested inset lists in a card, but I can't see any reason why this style would be needed. There are no screenshot diffs, so no tests were ever added for it. Therefore, I am good with removing this.

PR #29753, opened 2024-08-06
fix(content): allow custom roles and aria attributes to be set on content

brandyscarney:
Issue number: N/A

---------

Setting a custom `role` on the `ion-content` element does not work.

## What is the new behavior?
- Inherit attributes for the content element which allows a custom `role` property to be set
- Adds e2e tests for content, header, and footer verifying that the proper roles are assigned

- [ ] Yes
- [x] No

## Other information
To test this PR:

1. Switch to the branch and navigate to the `core/` directory
1. Make sure to run `npx playwright install` if it has not been updated recenly
1. Run `npm run test.e2e src/components/content/test/a11y/`
1. Verify that the tests pass
1. Remove my fix in `core/src/components/content/content.tsx` and run the test again
1. Verify that the `should allow for custom role` tests fail

### DIFF
--- a/core/src/components/content/content.tsx
+++ b/core/src/components/content/content.tsx
@@ -1,6 +1,7 @@
 import type { ComponentInterface, EventEmitter } from '@stencil/core';
 import { Build, Component, Element, Event, Host, Listen, Method, Prop, forceUpdate, h, readTask } from '@stencil/core';
-import { componentOnReady, hasLazyBuild } from '@utils/helpers';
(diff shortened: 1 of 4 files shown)


thetaPC (changes requested):

thetaPC (approved):
LGTM, just realized that there's no issue related to this. So disregard my previous request.

BenOsodrac (approved):

Issue #25184, opened 2022-04-25
feat: allow IonTabs to be used without a router outlet


sean-perkins:
### Prerequisites

- [X] I have read the [Contributing Guidelines](https://github.com/ionic-team/ionic-framework/blob/main/.github/CONTRIBUTING.md#creating-an-issue).
- [X] I agree to follow the [Code of Conduct](https://ionicframework.com/code-of-conduct).
- [X] I have searched for [existing issues](https://github.com/ionic-team/ionic-framework/issues) that already include this feature request, without success.

### Describe the Feature Request

Each framework wrapper couples the `ion-router-outlet` within `ion-tabs`, prevent developers from using `ion-tabs` in a standalone context or passing in another router solution.

`ion-tab` implementations through our framework wrappers, should support a new flag/option to disable coupling the `ion-router-outlet`.

The web component implementation allows for this behavior, the framework wrappers do not; there shouldn't be a forced opinionation for using the base component.

### Describe the Use Case

Applications that want to display content within their app, in a tabbed experience, without the complexities of routing or the need for routing; should be able to do so. 

If I wanted to create an application that:
1. Displays a tab-like experience within a modal (separate from URL navigation)
2. Creates a learning module that uses tabs to separate the content, without navigation events when clicking through the tab

### Describe Preferred Solution

```html
<ion-tabs no-outlet>

  <ion-tab tab="tab-one">
    Tab 1 content
  </ion-tab>

  <ion-tab tab="tab-two">
    Tab 2 content
  </ion-tab>

  <ion-tab-bar slot="bottom">
    <ion-tab-button tab="tab-one">
      <ion-label>Tab One</ion-label>
    </ion-tab-button>

    <ion-tab-button tab="tab-two">
      <ion-label>Tab Two</ion-label>
    </ion-tab-button>

  </ion-tab-bar>

</ion-tabs>
```

### Describe Alternatives

Alternatives would include creating your own tabs component, that brings in `ion-tab-bar` and sets up all the opinionated styling of `ion-tabs`.

### Additional Information

Similar issues/requests: #24991

riderx:
@sean-perkins it seems the `IonTabBar` also need the router to get the active tab.
Should both have this option?
And that mean they need a way to `init` to the right tab ?

riderx:
@sean-perkins i made a PR for changes in Vue

sean-perkins:
[...]

PR #29772, opened 2024-08-14
feat(react): use tabs without router

thetaPC:
Issue number: part of #25184 

This PR does not completely close the issue since it only addresses the React portion. The other frameworks will be done through other PRs.

---------

React: Developers have to provide `IonRouterOutlet` to use `IonTabs`. This means that developers are forced to tie their tabs to a router.

## What is the new behavior?

- `IonTabs` can be used without a router outlet as long as `IonTab` is a child and `IonRouterOutlet` is not.
- Added a test.
- Added a test page: `/tabs-basic`

- [ ] Yes
- [x] No

Developers who are still using `IonTabs` with `IonRouterOutlet` will not experience any changes. This feature PR introduces another option to use `IonTabs`, no changes for the current implementation.

## Other information

Dev build: 8.2.8-dev.11724280606.1a8c11d4

### DIFF
--- a/core/src/components/tabs/tabs.tsx
+++ b/core/src/components/tabs/tabs.tsx
@@ -43,16 +43,7 @@ export class Tabs implements NavOutlet {
 
   async componentWillLoad() {
     if (!this.useRouter) {
-      /**
-       * Router is used if it exists and there is no `no-router` attribute
-       * and the `ion-tab` is not being used.
-       *
-       * If the router is not being used, then the tabs will behave as a
-       * basic tab-based navigation without the history stack or URL updates
-       * associated with the router.
-       */
-      this.useRouter =
-        !!document.querySelector('ion-router') && !this.el.closest('[no-router]') && !this.el.querySelector('ion-tab');
+      this.useRouter = !!document.querySelector('ion-router') && !this.el.closest('[no-router]');
     }
     if (!this.useRouter) {
       const tabs = this.tabs;
--- a/core/src/components/tabs/tabs.tsx
+++ b/core/src/components/tabs/tabs.tsx
@@ -43,7 +43,16 @@ export class Tabs implements NavOutlet {
 
   async componentWillLoad() {
     if (!this.useRouter) {
-      this.useRouter = !!document.querySelector('ion-router') && !this.el.closest('[no-router]');
+      /**
+       * Router is used if it exists and there is no `no-router` attribute
+       * and the `ion-tab` is not being used.
+       *
+       * If the router is not being used, then the tabs will behave as a
+       * basic tab-based navigation without the history stack or URL updates
+       * associated with the router.
+       */
+      this.useRouter =
+        !!document.querySelector('ion-router') && !this.el.closest('[no-router]') && !this.el.querySelector('ion-tab');
     }
     if (!this.useRouter) {
       const tabs = this.tabs;
--- a/core/src/components/tabs/tabs.tsx
+++ b/core/src/components/tabs/tabs.tsx
@@ -43,7 +43,17 @@ export class Tabs implements NavOutlet {
 
   async componentWillLoad() {
     if (!this.useRouter) {
-      this.useRouter = !!document.querySelector('ion-router') && !this.el.closest('[no-router]');
+      /**
+       * JavaScript andd StencilJS use `ion-router`, while
+       * the other frameworks use `ion-router-outlet`.
+       *
+       * If either component is present then tabs will not use
+       * a basic tab-based navigation. It will use the history
+       * stack or URL updates associated with the router.
+       */
+      this.useRouter =
+        (!!document.querySelector('ion-router') || !!this.el.querySelector('ion-router-outlet')) &&
+        !this.el.closest('[no-router]');
     }
     if (!this.useRouter) {
       const tabs = this.tabs;
--- a/core/src/components/tabs/tabs.tsx
+++ b/core/src/components/tabs/tabs.tsx
@@ -52,7 +52,7 @@ export class Tabs implements NavOutlet {
        * stack or URL updates associated with the router.
        */
       this.useRouter =
-        (!!document.querySelector('ion-router') || !!this.el.querySelector('ion-router-outlet')) &&
+        (!!this.el.querySelector('ion-router-outlet') || !!document.querySelector('ion-router')) &&
         !this.el.closest('[no-router]');
     }
     if (!this.useRouter) {
--- a/packages/react/src/components/navigation/IonTabs.tsx
+++ b/packages/react/src/components/navigation/IonTabs.tsx
@@ -155,7 +155,7 @@ export const IonTabs = /*@__PURE__*/ (() =>
       });
 
       if (!outlet && !hasTab) {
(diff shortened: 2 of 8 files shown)


dev-charles15531:
I think i can do a vue PR for this. What do you think?

thetaPC:
@dev-charles15531 We greatly appreciate the offer! However, we will be working on the Vue portion very soon. I would recommend looking through the other existing issues. There are some related to router if you're interested in those.

dev-charles15531:
Okay, thank you!

core/src/components/tabs/tabs.tsx
```diff
@@ -43,7 +43,17 @@ export class Tabs implements NavOutlet {
...
-      this.useRouter = !!document.querySelector('ion-router') && !this.el.closest('[no-router]');
+      /**
+       * JavaScript andd StencilJS use `ion-router`, while
+       * the other frameworks use `ion-router-outlet`.
+       *
+       * If either component is present then tabs will not use
+       * a basic tab-based navigation. It will use the history
+       * stack or URL updates associated with the router.
+       */
+      this.useRouter =
```
sean-perkins, line 54: ~ I'd order the querying in the highest expected usage (left to right). For example there is more developers using the JS framework output of Ionic Framework and using `ion-router-outlet` as a result. If we query it first, we skip the operation of querying for `ion-router` and result in better runtime performance for a majority of applications.

packages/react/src/components/navigation/IonTabs.tsx
```diff
@@ -144,24 +153,36 @@ export const IonTabs = /*@__PURE__*/ (() =>
...
             <PageManager className={className ? `${className}` : ''} routeInfo={this.context.routeInfo} {...props}>
-              <ion-tabs className="ion-tabs" style={hostStyles}>
-                {tabBar.props.slot === 'top' ? tabBar : null}
-                <div style={tabsInner} className="tabs-inner">
-                  {outlet}
-                </div>
-                {tabBar.props.slot === 'bottom' ? tabBar : null}
-              </ion-tabs>
+              <IonTabsInner {...this.props}>
+                {React.Children.map(children, (child: any) => {
```
sean-perkins, line 175: Was there a type issue we ran into that resulted in needing `any` here?

core/src/components/tabs/tabs.tsx
thetaPC: https://github.com/ionic-team/ionic-framework/pull/29772/commits/f0ba2d7563d7da778843c4e6fe22210187269147

packages/react/src/components/navigation/IonTabs.tsx
thetaPC: It complains with `child.type`. That's most likely the reason that Liam used `any` in a [similar scenario](https://github.com/ionic-team/ionic-framework/blob/f0ba2d7563d7da778843c4e6fe22210187269147/packages/react/src/components/navigation/IonTabs.tsx#L104) in the same file.

sean-perkins: In my opinion that's not a good reason to disable type checking. Enabling it highlights that a child can be `null` or `undefined` (a case not handled in the implementation) and can also be a `string` or `number`. 

The `type` property comes from expecting that the child is a `React.ReactElement` type. This is one of a few types that could exist. 

The appropriate implementation should be:
```tsx
{React.Children.map(children, (child: React.ReactNode) => {
  if (React.isValidElement(child)) {
    if (child.type === IonTabBar || (child.type as any).isTabBar) {
      /**
       * The modified tabBar needs to be returned to include
       * the context and the overridden methods.
       */
      return tabBar;
    }
  }
  return child;
})}
```

The `as any` could be substituted with a custom type we have for the `isTabBar` type signature. However I think today it is hacked into implementations and doesn't have a type signature.

thetaPC: https://github.com/ionic-team/ionic-framework/pull/29772/commits/9c6d2a44772a27d39249aa0d486338a6d4052268

BenOsodrac (approved):

sean-perkins (approved):
Woo, nice work on this!

core/src/components/tabs/tabs.tsx
brandyscarney, line 1: I found an issue with the `IonTabButton` component when it includes an `href` attribute, which we might want to address. Here is the relevant code:

```tsx
<IonApp>
  <IonReactRouter>
    <IonTabs>
      <IonTab tab="tab1">
        <IonContent>
          Tab 1
        </IonContent>
      </IonTab>
      <IonTab tab="tab2">
        <IonContent>
        Tab 1
        </IonContent>
      </IonTab>
      <IonTab tab="tab3">
        <IonContent>
          Tab 3
        </IonContent>
      </IonTab>

      <IonTabBar slot="bottom">
        <IonTabButton tab="tab1" href="/tab1">
          <IonIcon aria-hidden="true" icon={triangle} />
          <IonLabel>Tab 1</IonLabel>
        </IonTabButton>
        <IonTabButton tab="tab2" href="/tab2">
          <IonIcon aria-hidden="true" icon={ellipse} />
          <IonLabel>Tab 2</IonLabel>
        </IonTabButton>
        <IonTabButton tab="tab3" href="/tab3">
          <IonIcon aria-hidden="true" icon={square} />
          <IonLabel>Tab 3</IonLabel>
        </IonTabButton>
      </IonTabBar>
    </IonTabs>
  </IonReactRouter>
</IonApp>
```

Follow these steps to reproduce the problem:
1. Create a new Ionic React tabs app: `ionic start myApp tabs --type=react`
2. Navigate to the app directory: `cd ./myApp`
3. Install the dev build: `npm i @ionic/react@8.2.8-dev.11723747334.19c4ecce @ionic/react-router@8.2.8-dev.11723747334.19c4ecce`
4. Update `src/App.tsx` with the above code
5. Serve the app: `ionic serve` 
6. Navigate to http://localhost:8100/ in the browser
7. Click on the 3rd tab "Tab 3"
8. See that it switches to the 3rd tab correctly
9. Refresh the browser
10. See that "Tab 3" is still selected but the 1st tab "Tab 1" content is displayed

> I can also upload my repository to GitHub and share it if that is easier.

If we remove the `href` attribute from the `IonTabButton`, the issue is resolved, and it behaves as expected. I am unsure if this is something we want to address directly in the code or perhaps provide a warning for.

thetaPC: https://github.com/ionic-team/ionic-framework/pull/29772/commits/a8c11d45ffe8d616b893955a8b68085ac75614d8

thetaPC: Issue 1: https://github.com/ionic-team/ionic-framework/pull/29772/commits/430aa0455c65d3835346b97ccdb09a36b2847132

brandyscarney (approved):
Tested the latest dev build in:
- A new React tabs starter app
- A new React tabs starter app with `IonRouterOutlet` removed and `IonTab` components containing only text content
- The [Ionic React Conference App](https://github.com/ionic-team/ionic-react-conference-app)

Everything is working correctly. Great job on this! 🎉

PR #29776, opened 2024-08-16
feat(textarea): add ionic theme design with md size

brandyscarney:
Issue number: internal

---------

Textarea does not have any styles for the ionic theme.

## What is the new behavior?
- Adds the typography styles for the default size (md)
- Adds the styles for the outline fill 
- Adds the styles for the label & helper text
- Adds the spacing for the label, textarea, counter and helper text

- [ ] Yes
- [x] No

## Other information

| Figma Design | Implemented Textarea |
| --- | --- | 
| Figma Design | Implemented Textarea | 

- [Basic Preview](https://ionic-framework-git-rou-10916-ionic1.vercel.app/src/components/textarea/test/basic?ionic:theme=ionic)
- [Size Preview](https://ionic-framework-git-rou-10916-ionic1.vercel.app/src/components/textarea/test/size?ionic:theme=ionic)


core/src/components/textarea/textarea.ionic.vars.scss
BenOsodrac, line 1: @brandyscarney Why not delete this file as is empty? 

I don't see the need in general for these vars partials for the Ionic Theme. They make sense for ios/md, as its a way to have the values documented, but for the new theme we have the tokens files already

core/src/components/textarea/textarea.ionic.scss
```diff
@@ -0,0 +1,80 @@
+@use "../../themes/ionic/ionic.globals.scss" as globals;
+@import "./textarea.common";
+@import "./textarea.ionic.vars";
```
BenOsodrac, line 3: This import can be removed

core/src/components/textarea/textarea.ionic.outline.scss
```diff
@@ -0,0 +1,72 @@
+@use "../../themes/ionic/ionic.globals.scss" as globals;
+@import "./textarea.vars";
```
BenOsodrac, line 2: This is not used I think and it shouldn't be imported here in my opinion, as its a partial that holds a value that doesn't come from a token.

core/src/components/textarea/textarea.ionic.outline.scss
```diff
@@ -0,0 +1,72 @@
...
+ */
+:host(.textarea-fill-outline) .textarea-bottom {
+  border-top: none;
+}
+
+// Textarea Fill: Outline, Native Textarea
+// ----------------------------------------------------------------
+
+:host(.textarea-fill-outline) textarea {
+  margin-top: 4px;
```
BenOsodrac, line 55:
```suggestion
  margin-top: globals.$ionic-space-100;
```

core/src/components/textarea/textarea.ionic.outline.scss
```diff
@@ -0,0 +1,72 @@
...
+}
+
+// Textarea Fill: Outline, Start / End Slots
+// ----------------------------------------------------------------
+
+:host(.textarea-fill-outline) .start-slot-wrapper,
+:host(.textarea-fill-outline) .end-slot-wrapper {
+  margin-top: 3px;
```
BenOsodrac, line 63: Can we use the DS values here, maybe the border token `ionic-border-size-075`? Or maybe we should talk with UX to have outline tokens if we think its not correct to use the ones from border.

core/src/components/textarea/textarea.ionic.outline.scss
```diff
@@ -0,0 +1,72 @@
...
+}
+
+:host(.textarea-fill-outline) .start-slot-wrapper {
+  margin-inline-start: 1px;
```
BenOsodrac, line 67: Same as above, but `ionic-border-size-025`

core/src/components/textarea/textarea.ionic.outline.scss
```diff
@@ -0,0 +1,72 @@
...
+}
+
+:host(.textarea-fill-outline) .end-slot-wrapper {
+  margin-inline-end: 1px;
```
BenOsodrac, line 71: Same as above, but `ionic-border-size-025`

BenOsodrac (changes requested):

core/src/components/textarea/textarea.ionic.scss
```diff
@@ -0,0 +1,80 @@
...
+@import "./textarea.ionic.outline.scss";
```
thetaPC, line 4: We should be using `@use` because:
1. That's how we've been setting it up so we would stay consistent with other `ionic` theme files
2. We would have less files to convert once we move from `@import`

core/src/components/textarea/textarea.ionic.outline.scss
thetaPC: In my opinion, it can be confusing to use a token that isn't related.

core/src/components/textarea/textarea.ionic.outline.scss
thetaPC, line 67: Is this value supposed to match the border width? If it is, then I agree using the token but with a comment so avoid confusion.

core/src/components/textarea/textarea.ionic.outline.scss
thetaPC, line 71: Is this value supposed to match the border width? If it is, then I agree using the token but with a comment so avoid confusion.

core/src/components/textarea/textarea.ionic.scss
```diff
@@ -0,0 +1,80 @@
...
+
+// Ionic Textarea
+// --------------------------------------------------
+
+:host {
+  --border-radius: #{globals.$ionic-border-radius-400};
+  --color: #{globals.$ionic-color-neutral-1200};
+  --highlight-color-valid: #{globals.$ionic-color-success-base};
+  --highlight-color-invalid: #{globals.$ionic-color-danger-base};
```
thetaPC, line 13: We should add `ionic-md` to the textarea highlight tests so we have snapshots for valid and invalid.

core/src/components/textarea/textarea.ionic.scss
thetaPC, line 1: The text goes outside of the textarea at the bottom:
 
Screenshot 2024-08-20 at 9 28 11 AM

core/src/components/textarea/test/size/textarea.e2e.ts
```diff
@@ -0,0 +1,74 @@
...
+            value="hi@ionic.io"
+          ></ion-textarea>
+        `,
+          config
+        );
+
+        const textarea = page.locator('ion-textarea');
+        await expect(textarea).toHaveScreenshot(screenshot(`textarea-size-medium-outline`));
+      });
+      test('should not have visual regressions with fill outline and round shape', async ({ page }) => {
```
thetaPC, line 55: This test seems redundant with `should not have visual regressions with fill outline` since the default shape is round. I would vote to remove this one.

thetaPC (changes requested):

core/src/components/textarea/textarea.ionic.vars.scss
brandyscarney: https://github.com/ionic-team/ionic-framework/pull/29776/commits/ea98cf34619813eebbd090272c124cdb515ba27b

core/src/components/textarea/textarea.ionic.scss
brandyscarney: https://github.com/ionic-team/ionic-framework/pull/29776/commits/ea98cf34619813eebbd090272c124cdb515ba27b

core/src/components/textarea/textarea.ionic.outline.scss
brandyscarney: https://github.com/ionic-team/ionic-framework/pull/29776/commits/ea98cf34619813eebbd090272c124cdb515ba27b

core/src/components/textarea/textarea.ionic.scss
brandyscarney: Good call, I copied this over from the `input` file which is an older implementation. Fixed in https://github.com/ionic-team/ionic-framework/pull/29776/commits/ea98cf34619813eebbd090272c124cdb515ba27b

core/src/components/textarea/textarea.ionic.scss
brandyscarney: I think we should wait to add the snapshots until we’ve implemented these properly, as they are currently just copied from input which is outdated.

core/src/components/textarea/test/size/textarea.e2e.ts
brandyscarney: Hmm yeah I copied it over from the input tests, but I think input was implemented when `soft` was the default shape. I removed it here: https://github.com/ionic-team/ionic-framework/pull/29776/commits/9e3ff8b4224733ac9f6a2fb8b45f0e5ac40ab157

core/src/components/textarea/textarea.ionic.outline.scss
brandyscarney: https://github.com/ionic-team/ionic-framework/pull/29776/commits/559a6f314d00dad78999a0b9ad14934dd74d00f9

core/src/components/textarea/textarea.ionic.outline.scss
brandyscarney: I spoke with Beth about this to determine the best token and realized I wasn’t setting the `ion-icon` to the correct scale. Adjusting it to the right size allowed me to remove the start/end margin, and then only `margin-top` of `2px` was needed, which I added using the `$ionic-space-050` token. This has been resolved in [this commit](https://github.com/ionic-team/ionic-framework/pull/29776/commits/e6d53fbb6a3e572082b1999f79f8eb696b14e6cd).

core/src/components/textarea/textarea.ionic.outline.scss
brandyscarney: I removed this per my comment here: https://github.com/ionic-team/ionic-framework/pull/29776#discussion_r1724010343

core/src/components/textarea/textarea.ionic.outline.scss
brandyscarney: I removed this per my comment here: https://github.com/ionic-team/ionic-framework/pull/29776#discussion_r1724010343

core/src/components/textarea/textarea.ionic.outline.scss
brandyscarney: No the value was just me adjusting it to match the Figma design, but it turns out this wasn't needed when I increased the size of `ion-icon` to the proper scale as mentioned here: https://github.com/ionic-team/ionic-framework/pull/29776#discussion_r1724010343

core/src/components/textarea/textarea.ionic.outline.scss
brandyscarney: Same as my comment here: https://github.com/ionic-team/ionic-framework/pull/29776#discussion_r1724011028

core/src/components/textarea/textarea.ionic.scss
brandyscarney: Fixed by https://github.com/ionic-team/ionic-framework/pull/29776/commits/303c13406cd8dffec610555e0981d67b8f6e59c3

BenOsodrac (approved):

thetaPC (approved):
LGTM

Issue #5926, opened 2016-03-24
ion-refresher pull down is blank in v2


palidanx:
#### Short description of the problem:
When using ion-refresher, the pull to refresh icons are not appearing.

#### What behavior are you expecting?
When the ion-refresher is used, the pull to refresh icons are displayed like
http://ionicframework.com/docs/v2/api/components/scroll/Refresher/

**Steps to reproduce:**
1)
In a .js page put in
```
  <ion-refresher (start)="doStart($event)"
                   (refresh)="doRefresh($event)"
                   (pulling)="doPulling($event)"
                   >
  </ion-refresher>
``` 
2) Attempt to pull the content down
3) Notice that the content the methods are called, but the pull to refresh icons are not there.

**Other information:** 

Using
```
  <ion-refresher (start)="doStart($event)"
                   (refresh)="doRefresh($event)"
                   (pulling)="doPulling($event)"
                   pullingtext="Pull to Refresh"
                   refreshingtext="Refreshing ..."
                   refreshingicon="refresher-pulling-icon"
                   >
  </ion-refresher>
```
doesn't work either.  I noticed in this page
http://ionicframework.com/docs/v2/demos/refresher/?ionicplatform=ios

It has this content
```
<ion-refresher-content pullingtext="Pull to refresh..." refreshingtext="Refreshing..." state="inactive">  <div class="refresher-pulling">
<div class="refresher-refreshing">
</ion-refresher-content>
```
 
But on my app, it is missing the `refresher-pulling` and `refresh-refreshing` divs.

Here is a copy of my ion-refresher which is generated
````
<ion-refresher pullingtext="Pull to Refresh" refreshingicon="refresher-pulling-icon" refreshingtext="Refreshing ..." class="">
  </ion-refresher>
```

**Ionic Version
```
Cordova CLI: 6.0.0
Ionic Version: 2.0.0-beta.3
Ionic CLI Version: 2.0.0-beta.19
Ionic App Lib Version: 2.0.0-beta.9
OS: Distributor ID:	Ubuntu Description:	Ubuntu 14.04.3 LTS 
Node Version: v5.4.1
```

palidanx:
Ah nvm.  I found that there was another bug that solved this.  The proper way is to put the text in `ion-refresher-content`

```
<ion-refresher (start)="doStart($event)"
                   (refresh)="doRefresh($event)"
                   (pulling)="doPulling($event)">
  <ion-refresher-content
[...]

PR #29780, opened 2024-08-19
fix(test): fix Stencil Nightly build

christian-bromann:
The Playwright test for `core/src/components/menu/test/safe-area/menu.e2e.ts` started to fail after introducing the following patch to Stencil: [#5926](https://github.com/ionic-team/stencil/pull/5926). After debugging the situation it turns out that the test overwrites the first style in the `<head />` tag which turns out to be a component style that caused all screenshot test to fail.

## What is the new behavior?
Overwrite the existing style by adding a new style tag at the bottom of the page.

- [ ] Yes
- [x] No

## Other information

n/a


brandyscarney (approved):

thetaPC (approved):
LGTM

PR #29782, opened 2024-08-19
fix(react): intellisense works with IntelliJ

### LINKED ISSUE #29755: bug: Breaking of react types with ionic elements
### Prerequisites

### Ionic Framework Version

v8.x

### Current Behavior

IN INTELLIJ

Screenshot 2024-08-06 at 2 02 31 PM

In React 8, the types do not generate for Ion Elements

### Expected Behavior

The type to not be unknown, and have type interface help while coding.

### Steps to Reproduce

IDE:INTELLIJ

1.ionic start myapp --type=react
2. Add <Ion${anything ionic element}>
3. Hover 

### Code Reproduction URL

https://github.com/ionic-team/starters/tree/main/react/official/blank

### Ionic Info

Ionic:

   Ionic CLI       : 7.2.0 (/opt/homebrew/lib/node_modules/@ionic/cli)
   Ionic Framework : @ionic/react 8.2.6

Capacitor:

   Capacitor CLI      : 6.1.1
   @capacitor/android : 6.1.1
   @capacitor/core    : 6.1.1
   @capacitor/ios     : 6.1.1

Utility:

   native-run  : 2.0.1

System:

   NodeJS : v21.7.2 (/opt/homebrew/Cellar/node/21.7.2/bin/node)
   npm    : 10.5.0
   OS     : macOS Unknown

### Additional Information

_No response_

thetaPC:
Issue number: resolves #29755 

---------

Types do not generate when a React app is opened in IntelliJ IDE.

## What is the new behavior?

- Updated `stencil/react-output-target` to the latest
- Updated the stencil config file
- Updated `jest`, `ts-jest`, and `typescript` because of the changes in `packages/react/tsconfig.json`
- Installed `jest-environment-jsdom` because Jest 26 no longer ships it by default since the test environment is now node by default. The test environment needs to be changed to `jsdom` when [building a web app](https://jestjs.io/docs/configuration#testenvironment-string).

Screenshot 2024-08-19 at 3 44 56 PM
Screenshot 2024-08-19 at 3 45 11 PM

- [ ] Yes
- [x] No

No visual changes are introduced and there will be no changes needed by developers. Their apps will continue to work as is.

## Other information

Dev build: 8.2.8-dev.11724100788.103018f2


packages/react/package.json
```diff
@@ -62,17 +63,19 @@
     "@typescript-eslint/parser": "^5.48.2",
     "eslint": "^7.32.0",
     "fs-extra": "^9.0.1",
-    "jest": "^26.6.3",
+    "jest": "^29.7.0",
+    "jest-environment-jsdom": "^29.7.0",
```
thetaPC, line 67: `jsdom` is needed to be installed since Jest 26 no longer ships it by default since the test environment is now `node` by default. The test environment needs to be changed to `jsdom` [when building a web app](https://jestjs.io/docs/configuration#testenvironment-string).

sean-perkins (approved):
Great work on this 👏 

I think this should be a `fix` or `feat`. Either: `fix(react): intellisense works with IntelliJ` or `feat(react): generate functional components` (as examples). 

One thing to be aware with these changes is that it only fixes the problem in the generated component wrappers. All of the legacy components that use class components likely have the same issue. We need to make the migration at some point in the future to support React Router 6, so it might be worth migrating those manual components over one-by-one over future releases. It shouldn't be a breaking change since the APIs are remaining the same.

brandyscarney (approved):

PR #29783, opened 2024-08-20
feat(checkbox): display as block when justify or alignment properties are defined

brandyscarney:
Issue number: internal

---------

## What is the current behavior?
When adding the `justify` or `alignment` property to an `ion-checkbox` it does not change the design because the checkbox is displayed inline.

## What is the new behavior?
- Change the checkbox's `display` property to `block` when the `justify` or `alignment` property is set.
- Set the default `justify-content` style to `space-between` so that a checkbox with `width: 100%` set **without** `justify` or `alignment` set will still have the same default
- Modifies the `label` e2e test to remove the explicit width as setting the property will make them full-width
- Adds two examples to the `label` e2e test of labels that do not have `justify` set but use `width: 100%` to ensure they are working as expected without it

## Does this introduce a breaking change?
- [ ] Yes
- [x] No

## Other information
- [Basic Preview](https://ionic-framework-git-rou-10906-ionic1.vercel.app/src/components/checkbox/test/basic)
- [Label Preview](https://ionic-framework-git-rou-10906-ionic1.vercel.app/src/components/checkbox/test/label)
- [Item Preview](https://ionic-framework-git-rou-10906-ionic1.vercel.app/src/components/checkbox/test/item)

PR #29801, opened 2024-08-26
feat(radio): display as block when justify or alignment properties are defined

brandyscarney:
Issue number: internal

---------

When adding the `justify` or `alignment` property to an `ion-radio` it does not change the design because the radio is displayed inline.

## What is the new behavior?
This is doing the same thing as https://github.com/ionic-team/ionic-framework/pull/29783

- Change the radio's `display` property to `block` when the `justify` or `alignment` property is set.
- Set the default `justify-content` style to `space-between` so that a radio with `width: 100%` set **without** `justify` or `alignment` set will still have the same default
- Modifies the `label-placement` e2e test to remove the explicit width as setting the property will make them full-width
- Adds two examples to the `label-placement` e2e test of labels that do not have `justify` set but use `width: 100%` to ensure they are working as expected without it
- Adds one example to the `label-placement` e2e test of a long label that uses `justify` to ensure it still wraps properly (this existed in the checkbox test but not radio)

- [ ] Yes
- [x] No

## Other information
- [Basic Preview](https://ionic-framework-git-rou-11044-ionic1.vercel.app/src/components/radio/test/basic)
- [Label Preview](https://ionic-framework-git-rou-11044-ionic1.vercel.app/src/components/radio/test/label-placement)
- [Item Preview](https://ionic-framework-git-rou-11044-ionic1.vercel.app/src/components/radio/test/item)


core/src/components/radio/test/basic/index.html
thetaPC, line 1: The tags aren't wrapping as expected. For example, `radio-group` and `radio` are on the same line. Let's make sure the tags are on new lines as the other test pages.

thetaPC (changes requested):

core/src/components/radio/test/basic/index.html
brandyscarney: This was done by the linter. I don't know how to fix it in prettier.

thetaPC: Got it!

BenOsodrac (approved):

PR #29820, opened 2024-08-30
fix(range): emit correct value when knob is at start of bar

### LINKED ISSUE #29792: bug: ion-range emit value 'NaN' when dragging to the far left end of the screen when the browser is full screen
### Prerequisites

### Ionic Framework Version

v7.x

### Current Behavior

[Peek 2024-08-23 14-31.webm](https://github.com/user-attachments/assets/7608c710-cfe9-470d-9721-632f10ade866)
Please refer to the attached video.

In event handlers such as ionInput or ionKnobMoveEnd on an ion-range, it can happen that the event.detail.value is NaN.
When specifying currentX in the onEnd function inside range.js, detail.currentX can be 0. 
But, If it is 0, it tries to assign clientX to currentX by || even though it is a GestureDetail. However, in this case, clientX is undefined, which causes a NaN to occur later.

The problematic part of the code is attached as the image below.
image (18)
The bug can be fixed if it is modified to pass 0 when detail.currentX is 0.

### Expected Behavior

When detail.currentX is 0, the value should be passed as is.

### Steps to Reproduce

1. clone reproduction code.
2. start project
3. set browser to full screen
4. drag knob  to the far left end of the screen

### Code Reproduction URL

https://github.com/dodo-Riley/ionic-range-test

### Ionic Info

Ionic:

   Ionic CLI                     : 7.2.0 (/home/toonsquare/.nvm/versions/node/v18.18.1/lib/node_modules/@ionic/cli)
   Ionic Framework               : @ionic/angular 7.8.6
   @angular-devkit/build-angular : 15.2.11
   @angular-devkit/schematics    : 15.2.11
   @angular/cli                  : 15.2.11

Utility:

   cordova-res                          : 0.15.4
   native-run (update available: 2.0.1) : 2.0.0

System:

   NodeJS : v18.18.1 (/home/toonsquare/.nvm/versions/node/v18.18.1/bin/node)
   npm    : 9.8.1
   OS     : Linux 6.5

### Additional Information

_No response_

sean-perkins:
Issue number: resolves #29792

---------

When the user drags the range knob (most easily reproduced in fullscreen mode) and the gesture emits a current x position of `0`, the range incorrectly emits a value of `NaN`.

## What is the new behavior?

- `ion-range` does not emit `NaN` and instead emits the correct range value for the knob

- [ ] Yes
- [ ] No

### DIFF
--- a/core/src/components/range/range.tsx
+++ b/core/src/components/range/range.tsx
@@ -449,7 +449,7 @@ export class Range implements ComponentInterface {
    */
   private onEnd(detail: GestureDetail | MouseEvent) {
     const { contentEl, initialContentScrollY } = this;
-    const currentX = (detail as GestureDetail).currentX || (detail as MouseEvent).clientX;
+    const currentX = (detail as GestureDetail).currentX ?? (detail as MouseEvent).clientX;
 
     /**
      * The `pressedKnob` can be undefined if the user never


core/src/components/range/range.tsx
```diff
@@ -449,7 +449,7 @@ export class Range implements ComponentInterface {
    */
   private onEnd(detail: GestureDetail | MouseEvent) {
     const { contentEl, initialContentScrollY } = this;
-    const currentX = (detail as GestureDetail).currentX || (detail as MouseEvent).clientX;
+    const currentX = (detail as GestureDetail).currentX ?? (detail as MouseEvent).clientX;
```
sean-perkins, line 452: The problem here was that `detail.currentX` can evaluate to `0` and `0 || detail.clientX` will treat the `0` as falsy and return the `detail.clientX` which is `undefined`. 

Instead we wanted to take the `currentX` or take the `clientX` if the value is null/undefined.

PR #29853, opened 2024-09-10
feat(modal): add shape prop & styling for ionic theme

tanner-reits:
Issue number: internal

---------

No shape prop or ionic theme styling

## What is the new behavior?

- Adds `shape` prop to `ion-modal`
- Adds styling for shape in ionic theme

- [ ] Yes
- [x] No


core/src/components/modal/test/shape/index.html
```diff
@@ -0,0 +1,176 @@
...
+    <link href="../../../../../css/ionic.bundle.css" rel="stylesheet" />
+    <link href="../../../../../scripts/testing/styles.css" rel="stylesheet" />
+    <script src="../../../../../scripts/testing/scripts.js"></script>
+    <script nomodule src="../../../../../dist/ionic/ionic.js"></script>
+    <script type="module" src="../../../../../dist/ionic/ionic.esm.js"></script>
+
+    <style>
+      ion-modal {
+        --box-shadow: 0px 0px 30px 10px rgba(0, 0, 0, 0.1);
+      }
```
tanner-reits, line 19: This will go away once we have the rest of the ionic theme styles added for modal. It's just hear so the border radius is easier to see in the screenshots

core/src/components/modal/modal.ionic.scss
```diff
@@ -0,0 +1,19 @@
+@use "../../themes/ionic/ionic.globals.scss" as globals;
+@import "./modal";
```
brandyscarney, line 2: Typically we have been splitting the common file (`modal.scss`) into two files `modal.native.scss` and `modal.common.scss` and then importing the common one instead:

```diff
-@import "./modal";
+@use "./modal.common";
```

However, since this PR is only adding shape, I don't think it makes sense to try to figure out what the values in `core/src/components/modal/modal.vars.scss` will be for the Ionic theme. We have a ticket (ROU-11153) for adding the default styling, so I think we can handle that there. OPTIONAL suggestion to keep track of this:

```suggestion
// TODO(ROU-11153): separate modal styles to common and native
@import "./modal";
```

brandyscarney (approved):

PR #29865, opened 2024-09-11
fix(react, vue): add default value for navManager on tabs

thetaPC:
Issue number: N/A

---------

React: Refreshes the app upon a tab button press and returns to the first tab content.
Vue: Does not render the tabs when the app doesn't use a router at all because of `ionRouter` not being defined.

## What is the new behavior?

- React: Users can press on the tab button without having the app refresh and displays the associated tab content.
- Vue: Renders the tabs when router is not present.

- [ ] Yes
- [x] No

## Other information

Dev build: 8.3.1-dev.11726094383.1cdbab6f

How to test:
I recommend using the playgrounds in the Ionic Framework docs repo since those playgrounds are setup to not have any routers at a high level.

1. Navigate to the [basic usage section](https://ionic-docs-git-rou-11138-ionic1.vercel.app/docs/api/tabs#basic-usage)
2. Select React
3. Open the StackBlitz example
4. Click on any tab button
5. Notice that the app refreshes and returns to the first tab content
6. Install the dev build: `npm install @ionic/react@8.3.1-dev.11726094383.1cdbab6f`
7. Verify that the app doesn't refresh and reset the content after clicking on any tab button
8. Navigate to the [basic usage section](https://ionic-docs-git-rou-11138-ionic1.vercel.app/docs/api/tabs#basic-usage)
9. Select Vue
10. Open the StackBlitz example
11. Notice that tabs doesn't render
12. Open the console and notice that there's an error `ionRouter is undefined`
13. Install the dev build: `npm install @ionic/vue@8.3.1-dev.11726094383.1cdbab6f`
14. Verify that it renders
15. Verify that the console no longer shows the error

### DIFF
--- a/packages/react/src/components/navigation/IonTabBar.tsx
+++ b/packages/react/src/components/navigation/IonTabBar.tsx
@@ -212,8 +212,10 @@ class IonTabBarUnwrapped extends React.PureComponent<InternalProps, IonTabBarSta
       if (this.props.onIonTabsDidChange) {
         this.props.onIonTabsDidChange(new CustomEvent('ionTabDidChange', { detail: { tab: e.detail.tab } }));
       }
-      this.setActiveTabOnContext(e.detail.tab);
-      this.context.changeTab(e.detail.tab, currentHref, e.detail.routeOptions);
+      if (this.props.routerOutletRef?.current) {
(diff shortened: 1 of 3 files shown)


tanner-reits (approved):
LGTM! Tested reproduction steps and both cases look fixed!

PR #29873, opened 2024-09-13
feat(backdrop): ionic theme styles

tanner-reits:
Issue number: internal

---------

No Ionic theme styles for the `ion-backdrop` element

## What is the new behavior?

Adds Ionic theme styles for `ion-backdrop` and updates snapshots.

- [ ] Yes
- [x] No

## Other information

The backdrop changes were only applied to `ion-modal` for now as all other components where `ion-backdrop` is used do not yet have dedicated Ionic theme styles. This is relevant as those components will need to set `--backdrop-opacity: <value>` for their own implementations.


brandyscarney (approved):

core/src/components/backdrop/backdrop.ionic.scss
```diff
@@ -0,0 +1,7 @@
+@use "../../themes/ionic/ionic.globals.scss" as globals;
+@import "./backdrop";
+
+:host {
+  background-color: globals.$ionic-color-base-black;
+  backdrop-filter: blur(2px);
```
BenOsodrac, line 6: @tanner-reits I didn't found this blur on the Figma? Was it required by UX?

tanner-reits: @BenOsodrac It was in the styles listed in dev mode:
```
opacity: 0.7;
background: var(--color-base-black, #111);
backdrop-filter: blur(2px);
```

BenOsodrac: Its just that looking at the actual UI doesn't look it has blur. Also, from the snapshots, it seems its only working on Chrome.

Maybe it might need the `-webkit-` for it to work, as mentioned in this [this source](https://www.reddit.com/r/css/comments/12jhh10/solution_backdrop_filter_blur_doesnt_work_in/).

tanner-reits: @BenOsodrac So the issue with the snapshots is that Playwright doesn't seem to correctly handle the `backdrop-filter` property (https://github.com/microsoft/playwright/issues/28363). If you open the [preview for the sheet modal](https://ionic-framework-ogvfcq66v-ionic1.vercel.app/src/components/modal/test/sheet?ionic:theme=ionic) in Safari, Chrome, or Firefox, the blur effect is there. 

Seems like this is something we've ran into before with other components, and not really anything we can do about it. Might get resolved with new binaries a part of the iOS 18 release, but can't be certain.

tanner-reits:
@brandyscarney Also re-request review from you since I made a change in my latest commit. The blur effect wasn't apply on desktop browsers, so I changed how the background color and opacity are set so it's visible in the preview links now

brandyscarney (approved):

core/src/components/backdrop/backdrop.ionic.scss
BenOsodrac: Got it, thanks, great work 🚀

BenOsodrac (approved):

PR #29876, opened 2024-09-16
feat(modal): add default ionic theme styles

tanner-reits:

---------

No default modal styles for the Ionic theme

## What is the new behavior?

- Splits common and native styles into separate stylesheets
- Adds default styles to modals for Ionic theme

- [ ] Yes
- [x] No

## Other information

I only applied some of the global styles to the sheet modal vairation since it is unclear whether the card variation will be used in the widgets


core/src/css/ionic/core.ionic.scss
```diff
@@ -36,45 +36,76 @@ body.backdrop-no-scroll {
...
+  @include globals.position(0, null, null, 0);
+
+  position: absolute;
+
+  width: 100%;
+  height: 100%;
+
+  transform: translateZ(0);
+
+  text-align: center;
```
tanner-reits, line 90: I had to override the styles for `ion-title` because the ionic theme styles for that component use `:host-context()` which is supported in Safari and Firefox. I can make this change in the title styles themselves, but it does result in some screenshots changing and I wasn't exactly sure what the correct styles were for that component

thetaPC (approved):
LGTM, just minor requests

brandyscarney (approved):

PR #29847, opened 2024-09-07
fix(vue): incorrect view rendered when using router.go(-n) 

xxllxhdj:
Issue number: resolves #28201

---------

## What is the current behavior?

Say you have the following application history:

`/pageA` --> `/pageB` --> `/pageC` --> `/pageD` -->  back to `/pageC`

If you were to call router.go(-2) on `/pageC`,  you would be brought back to `/pageB`. But the address in the browser address bar is correct.

## What is the new behavior?

In above case, if you were to call router.go(-2) on `/pageC`,  you would be brought back to `/pageA`.

## Does this introduce a breaking change?

- [ ] Yes
- [x] No

Issue #28201, opened 2023-09-20
bug: router.go() not working as expected in ionic vue app


TaraWei-twobulls:
### Prerequisites

- [X] I have read the [Contributing Guidelines](https://github.com/ionic-team/ionic-framework/blob/main/.github/CONTRIBUTING.md#creating-an-issue).
- [X] I agree to follow the [Code of Conduct](https://ionicframework.com/code-of-conduct).
- [X] I have searched for [existing issues](https://github.com/ionic-team/ionic-framework/issues) that already report this problem, without success.

### Ionic Framework Version

v7.x

### Current Behavior

Actual behaviour: app view is in Page2 but url is Page1's.

### Expected Behavior

Expected to be in Page1 for both url and view.

### Steps to Reproduce

1. Home page > Page 1 > Page2 > Page3 using ionRouter.push()
2. Page3 > Page2 using ionRouter.back()
3. Trigger history.go(-2) or router.go(-2) in Page2
[4f603fb4-1c47-478f-9229-391b3fcef05c.webm](https://github.com/ionic-team/ionic-framework/assets/107837804/9e600f49-ed7f-42d6-bc18-72839b5c0fc9)

### Code Reproduction URL

https://github.com/TaraWei-twobulls/Ionic-vue-app-router-issue-reproduction/tree/main

### Ionic Info

Ionic:

   Ionic CLI       : 7.1.1 (/usr/local/lib/node_modules/@ionic/cli)
   Ionic Framework : @ionic/vue 7.1.2

Capacitor:

   Capacitor CLI      : 5.1.1
   @capacitor/android : 5.1.1
   @capacitor/core    : 5.1.1
   @capacitor/ios     : 5.1.1

### Additional Information

  Seems  ionic outlet view does not know where the current view is, and it always go back from the last view in stack though user has been back to the previous view

TaraWei-twobulls:
Reproduction added. Please find the example codebase and the issue recording.

liamdebeasi:
Did you push the complete code? I see `myApp` is added as a submodule, but there's no information in there so I can't initialize it.

TaraWei-twobulls:
Sorry just noticed that and updated it.
Please run `npm i` and` ionice serve` to run it locally.

thetaPC:
Thank you for reporting the issue. I can confirm that the `router.go()` is not navigating to the correct view. This issue has been categorized as a bug to prioritize in our backlog.

bentleyo:
[...]

PR #29877, opened 2024-09-17
fix(vue): incorrect view rendered when using router.go(-n)

bentleyo:
This PR fixes the navigation issue related to `router.go` that was identified in issue #28201. After working on this issue I realised that @xxllxhdj has already created a PR for this in #29847. While their fix is great, I have added tests that replicate the issue and approached the problem using the existing code.

The core problem was that when using `findLastLocation` with deltas less than -1, the code no longer checked relative to the specified route.

If a user navigates from `/home` -> `/pageA` -> `/pageB` -> `/pageC` -> `/pageB` (by navigating backwards) then `router.go(-2)` is called the URL will be updated to `/home` correctly, but the app will try to render `/pageA`.

## What is the new behavior?

The app will correctly render `/pageA`, which matches the URL.

- [ ] Yes
- [X] No

### DIFF
--- a/packages/vue-router/src/locationHistory.ts
+++ b/packages/vue-router/src/locationHistory.ts
@@ -239,15 +239,14 @@ export const createLocationHistory = () => {
         }
       }
     }
-    if (delta < -1) {
-      return locationHistory[locationHistory.length - 1 + delta];
-    } else {
-      for (let i = locationHistory.length - 2; i >= 0; i--) {
-        const ri = locationHistory[i];
-        if (ri) {
-          if (ri.pathname === routeInfo.pushedByRoute) {
+    for (let i = locationHistory.length - 2; i >= 0; i--) {
+      const ri = locationHistory[i];
+      if (ri) {
+        if (ri.pathname === routeInfo.pushedByRoute) {
+          if (delta >= -1) {
             return ri;
           }
+          return locationHistory[i + 1 + delta]
         }
       }
     }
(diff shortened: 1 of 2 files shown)


bentleyo:
The main problem was that when using `findLastLocation` with deltas less than -1, the code no longer checked relative to the specified route. My code change fixes this 😄

bentleyo:
@brandyscarney just a heads up, this PR relates to one assigned to you here: https://github.com/ionic-team/ionic-framework/pull/29877

bentleyo:
@thetaPC are you able to check this one for me and let me know if any changes required on my end to get this one through? 🙏

packages/vue-router/src/locationHistory.ts
thetaPC, line 1: Thank you for submitting a PR!

I noticed you removed the if statement from the code. I'd love to understand your reasoning behind that change.

bentleyo: Hi @thetaPC thanks for the review!

The `if` statement I've removed was originally added to handle the situation where `delta` is less than `-1`. In that situation what it does is return the location `delta` many steps back from the end of the history stack.

```
if (delta < -1) {
  return locationHistory[locationHistory.length - 1 + delta];
}
```

However, this wasn't correct behaviour. There is a `routeInfo` passed into the function and the returned location should be `delta` steps back from that. Not back from the end of `locationHistory`.

This is why the bug reported in the issue was happening.

I've made it so it handles both situations using that same code. This is mostly reusing the code that was in the `else` statement.

That code essentially searches backwards through the history stack until it finds `routeInfo` and returns its parent (handling the `delta` `-1` situation). I realised I could use index of that location in the array + delta to return the correct location.

I've added tests to handle the situations and ensure the behaviour works as expected. Let me know if you need any more details :)

thetaPC (approved):
LGTM, thank you!

thetaPC:
Gave @xxllxhdj co author credit as the team decided to go with this approach. Thank you everyone!

PR #29884, opened 2024-09-19
fix(segment): prevent flickering for scrollable on iOS

### LINKED ISSUE #29523: bug: ion-segment scrollable IOS animation bug
### Prerequisites

### Ionic Framework Version

v7.x, v8.x

### Current Behavior

When you use multiple segment items there is strange animation bug on ios devices when using scrollable option.  (Android, Web - even Safari works fine)

```tsx
<IonSegment scrollable={true} value="heart">
  <IonSegmentButton value="home">
    <IonIcon icon={home}></IonIcon>
  </IonSegmentButton>
  <IonSegmentButton value="heart">
    <IonIcon icon={heart}></IonIcon>
  </IonSegmentButton>
  <IonSegmentButton value="pin">
    <IonIcon icon={pin}></IonIcon>
  </IonSegmentButton>
  <IonSegmentButton value="star">
    <IonIcon icon={star}></IonIcon>
  </IonSegmentButton>
</IonSegment>
```

Here you can see the result on real device, i did a little CSS changes but tried also with original styles and there is the same result. I also did a little digging in your code and managed to solve the problem:

In Segment class there is method

```ts
private scrollActiveButtonIntoView(smoothScroll = true) {
  const { scrollable, value, el } = this;

  if (scrollable) {
    const buttons = this.getButtons();
    const activeButton = buttons.find((button) => button.value === value);
    if (activeButton !== undefined) {
      const scrollContainerBox = el.getBoundingClientRect();
      const activeButtonBox = activeButton.getBoundingClientRect();

      /**
        * Subtract the active button x position from the scroll
        * container x position. This will give us the x position
        * of the active button within the scroll container.
        */
      const activeButtonLeft = activeButtonBox.x - scrollContainerBox.x;

      /**
        * If we just used activeButtonLeft, then the active button
        * would be aligned with the left edge of the scroll container.
        * Instead, we want the segment button to be centered. As a result,
        * we subtract half of the scroll container width. This will position
        * the left edge of the active button at the midpoint of the scroll container.
        * We then add half of the active button width. This will position the active
        * button such that the midpoint of the active button is at the midpoint of the
        * scroll container.
        */
      const centeredX = activeButtonLeft - scrollContainerBox.width / 2 + activeButtonBox.width / 2;

      /**
        * We intentionally use scrollBy here instead of scrollIntoView
        * to avoid a WebKit bug where accelerated animations break
        * when using scrollIntoView. Using scrollIntoView will cause the
        * segment container to jump during the transition and then snap into place.
        * This is because scrollIntoView can potentially cause parent element
        * containers to also scroll. scrollBy does not have this same behavior, so
        * we use this API instead.
        *
        * Note that if there is not enough scrolling space to center the element
        * within the scroll container, the browser will attempt
        * to center by as much as it can.
        */
      el.scrollBy({
        top: 0,
        left: centeredX,
        behavior: smoothScroll ? 'smooth' : 'instant',
      });
    }
  }
}
```

Instead of `el.scrollBy` if I call `el.scrollTo`, the problem dissapears and it works as it should

https://github.com/ionic-team/ionic-framework/assets/58735164/21208d4d-bfd9-47ed-a74a-dcc45165b2e0

### Expected Behavior

The same result when you just change `el.scrollBy` to `el.scrollTo`

https://github.com/ionic-team/ionic-framework/assets/58735164/1a0af05f-a8f2-440e-8811-03804e6893eb

### Steps to Reproduce

Run any basic Ionic/Capacitor project on ios device and use scrollable ion-segment with multiple items, that will has width more than 100% of device width

### Code Reproduction URL

https://stackblitz.com/edit/dw3jre?file=package.json

### Ionic Info

Ionic:

   Ionic CLI                     : 7.0.1 (/usr/local/lib/node_modules/@ionic/cli)
   Ionic Framework               : @ionic/angular 8.0.0
   @angular-devkit/build-angular : 17.3.5
   @angular-devkit/schematics    : 17.3.5
   @angular/cli                  : 17.3.5

Capacitor:

   Capacitor CLI      : 6.0.0
   @capacitor/android : 6.0.0
   @capacitor/core    : 6.0.0
   @capacitor/ios     : 6.0.0

Utility:

   cordova-res : 0.15.4
   native-run  : 2.0.1

### Additional Information

_No response_

thetaPC:
Issue number: resolves #29523

---------

The scrollable segment flickers on iOS physical devices or simulators when the active button is near the edge of the screen. The jump is due to the button being scrolled to the center and snaps back to the edge since the button was scrolled past the container.

## What is the new behavior?

- Switched to `scrollTo` provides for a smoother transition.
- Gave co author credit to the original reporter since they provided part of the solution
- No new tests were created since functionality stays the same and testing on Playwright would be impossible to recreate

- [ ] Yes
- [x] No

## Other information

Dev build: 8.3.2-dev.11726779768.16e1f1d2

How to test:
1. Create a new app through any starter
2. Add a scrollable segment with at least 6 buttons (code snippet example below)
3. Recommended to change the segment mode to `md` since it's easier to see the flicker
4. Build the app and open it in an iOS or simulator (if more instructions on how to do this is needed, reach out to me)
5. Click on the third button
6. Click on the first button
7. Notice the flicker
8. Click over to the third to last button
9. Click on either the last two buttons
10. Notice the flicker
11. Install the dev build
12. Verify the load does not flicker
13. Repeat steps 4 and 5
14. Verify the flicker is no longer there
15. Repeat steps 7 and 8
16. Verify the flicker is no longer there

```js
<ion-segment value="2" scrollable="true" mode="md">
  <ion-segment-button value="1">
    <ion-label>Button 1</ion-label>
  </ion-segment-button>
  <ion-segment-button value="2">
    <ion-label>Button 2</ion-label>
  </ion-segment-button>
  <ion-segment-button value="3">
    <ion-label>Button 3</ion-label>
  </ion-segment-button>
  <ion-segment-button value="4">
    <ion-label>Button 4</ion-label>
  </ion-segment-button>
  <ion-segment-button value="5">
    <ion-label>Button 5</ion-label>
  </ion-segment-button>
  <ion-segment-button value="6">
    <ion-label>Button 6</ion-label>
  </ion-segment-button>
</ion-segment>
```

### DIFF
--- a/core/src/components/segment/segment.tsx
+++ b/core/src/components/segment/segment.tsx
@@ -340,39 +340,38 @@ export class Segment implements ComponentInterface {
          * scroll container.
          */
         const centeredX = activeButtonLeft - scrollContainerBox.width / 2 + activeButtonBox.width / 2;
-        console.log(centeredX);
 
         /**
-         * We intentionally use scrollBy here instead of scrollIntoView
+         * newScrollPosition is the absolute scroll position that the
+         * container needs to move to in order to center the active button.
+         * It is calculated by adding the current scroll position
+         * (scrollLeft) to the offset needed to center the button
+         * (centeredX).
+         */
+        const newScrollPosition = el.scrollLeft + centeredX;
+
+        /**
+         * We intentionally use scrollTo here instead of scrollIntoView
          * to avoid a WebKit bug where accelerated animations break
          * when using scrollIntoView. Using scrollIntoView will cause the
          * segment container to jump during the transition and then snap into place.
          * This is because scrollIntoView can potentially cause parent element
-         * containers to also scroll. scrollBy does not have this same behavior, so
+         * containers to also scroll. scrollTo does not have this same behavior, so
          * we use this API instead.
+         *
+         * scrollTo is used instead of scrollBy because there is a
+         * Webkit bug that causes scrollBy to not work smoothly when
+         * the active button is near the edge of the scroll container.
+         * This leads to the buttons to jump around during the transition.
          *
          * Note that if there is not enough scrolling space to center the element
          * within the scroll container, the browser will attempt
          * to center by as much as it can.
          */
-        // el.scrollBy({
-        //   top: 0,
-        //   left: centeredX,
-        //   behavior: smoothScroll ? 'smooth' : 'instant',
-        // });
-
-        activeButton.scrollIntoView({
+        el.scrollTo({
+          top: 0,
+          left: newScrollPosition,
           behavior: smoothScroll ? 'smooth' : 'instant',
-          inline: 'center',
-
-          /**
-           * Segment should scroll on the
-           * horizontal axis. `block: 'nearest'`
-           * ensures that the vertical axis
-           * does not scroll if the segment
-           * as a whole is already in view.
-           */
-          block: 'nearest',
         });
       }
     }
--- a/core/src/components/segment/segment.tsx
+++ b/core/src/components/segment/segment.tsx
@@ -340,6 +340,7 @@ export class Segment implements ComponentInterface {
          * scroll container.
          */
         const centeredX = activeButtonLeft - scrollContainerBox.width / 2 + activeButtonBox.width / 2;
+        console.log(centeredX);
 
         /**
          * We intentionally use scrollBy here instead of scrollIntoView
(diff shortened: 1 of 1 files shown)


brandyscarney (approved):
Fix looks good! 👍

tanner-reits (approved):
Fix looks good!

ZanogeaDumitru:
Solving this problem itself caused another problem. 
Now the ion-segment no longer flicks but if a button is physically located halfway outside the view it is no longer scrolled to the center, or inside the view (when selected) if it is the last element. With a few clicks on the elements just nearby it resumes but if you try the click again when it is outside it presents the same problem again.

thetaPC:
Hello @ZanogeaDumitru! Please open a new issue with a repro so we can look at it as soon as possible. Thanks!

PR #29891, opened 2024-09-24
fix(ionic): update colors to import function

thetaPC:
Issue number: resolves #

---------

## What is the new behavior?

-
-
-

- [ ] Yes
- [ ] No


core/src/components/input/input.ionic.scss
```diff
@@ -140,7 +140,7 @@
 // ----------------------------------------------------------------
 
 .input-bottom {
-  @include padding(7px, 0);
+  @include padding(globals.$ionic-space-100, 0);
```
thetaPC, line 143: After adding the ionic theme file to the `setContent`, the snapshots were trying to add more spacing than expected. After investigating, I noticed that this no longer matched Figma.

core/src/components/item/item.ionic.scss
thetaPC, line 1: With the new changes to Playwright, I realized that the color and font size were not being displayed correctly when using `setContent`. This was due to the lack of `ion-content`, `ion-content` would set these styles.

core/src/components/item/item.ionic.scss
```diff
@@ -19,8 +19,10 @@
   --padding-end: #{globals.$ionic-space-400};
   --padding-bottom: #{globals.$ionic-space-200};
   --padding-start: #{globals.$ionic-space-400};
+  --color: #{globals.$ionic-color-neutral-1200};
```
thetaPC, line 22: Since this color is close to black, the snapshots are going to look very similar.

core/src/components/list-header/list-header.ionic.scss
```diff
@@ -4,10 +4,12 @@
 
 :host {
   --background: transparent;
+  --color: #{globals.$ionic-color-neutral-1200};
```
thetaPC, line 7: Since this color is close to black, the snapshots are going to look very similar.

core/src/components/spinner/spinner.ionic.scss
```diff
@@ -4,6 +4,10 @@
 // Ionic Spinner
 // --------------------------------------------------
 
+:host {
+  --color: #{globals.$ionic-color-neutral-800};
```
thetaPC, line 8: Updated this to match Figma else it was trying to re-generate with black being the default value.

core/src/utils/test/playwright/page/utils/set-content.ts
```diff
@@ -44,6 +44,7 @@ export const setContent = async (page: Page, html: string, testInfo: TestInfo, o
         <meta charset="UTF-8" />
         <meta name="viewport" content="width=device-width, initial-scale=1.0, minimum-scale=1.0" />
         <link href="${baseUrl}/css/ionic.bundle.css" rel="stylesheet" />
+        ${theme === 'ionic' ? `<link href="${baseUrl}/css/ionic/bundle.ionic.css" rel="stylesheet" />` : ''}
```
thetaPC, line 47: This matches the [script](https://github.com/ionic-team/ionic-framework/blob/21496fb5ee45b99c4878aec5cb1c4f6320f6aa0f/core/scripts/testing/scripts.js#L40-L50) that is being used in the test pages.

thetaPC:
Closing in favor of https://github.com/ionic-team/ionic-framework/pull/29898

PR #29903, opened 2024-09-26
test(input): update snapshots

thetaPC:
Issue number: internal

---------

The snapshots for input are not accurate. This can be verified by comparing it to the test pages, which do import the right bundle file.

## What is the new behavior?

- Updated snapshots
- Updated padding for the bottom content to match Figma

- [ ] Yes
- [x] No

## Other information


core/src/components/input/input.ionic.scss
```diff
@@ -140,7 +140,7 @@
 // ----------------------------------------------------------------
 
 .input-bottom {
-  @include padding(7px, 0);
+  @include padding(globals.$ionic-space-100, 0);
```
brandyscarney, line 143: The Figma design looks like this should only apply padding to the top. Is this correct?

thetaPC: Can confirm: https://github.com/ionic-team/ionic-framework/pull/29903/commits/4e3bd15d0bcfab65dcfc47be779b4e71e74cbc58

brandyscarney (approved):

PR #29905, opened 2024-09-27
refactor(item): update color and font size

thetaPC:
Issue number: internal

---------

## What is the current behavior?

Playwright was not importing the correct ionic bundle file, which has caused some snapshots to not be correct. This is apparent when a test uses `setContent`. There are several components that are needing their snapshots to be re-generated. In order to not have a PR with an immense amount of snapshots, several PRs will be created to split off the work to make it easier on reviewers.

The snapshots for item are not accurate. This can be verified by comparing it to the test pages, which do import the right bundle file.

## What is the new behavior?

- Updated snapshots
- Updated color and font size to match design

## Does this introduce a breaking change?

- [ ] Yes
- [x] No

## Other information

[Base branch](https://github.com/ionic-team/ionic-framework/pull/29898)

PR #29907, opened 2024-09-27
refactor(list-header): update color and font size

thetaPC:
Issue number: internal

---------

The snapshots for list header are not accurate. This can be verified by comparing it to the test pages, which do import the right bundle file.

Since the `ion-item` was updated, then the snapshots for `ion-list-header` would also need to be updated due to `ion-list-header` using `ion-item`.

## What is the new behavior?

- Updated snapshots
- Updated color and text size to match design

- [ ] Yes
- [x] No

## Other information

[Item branch](https://github.com/ionic-team/ionic-framework/pull/29905)


core/src/components/list-header/list-header.ionic.scss
```diff
@@ -4,10 +4,12 @@
 
 :host {
   --background: transparent;
+  --color: #{globals.$ionic-color-neutral-1200};
 
   min-height: 58px;
 
   font-size: globals.$ionic-font-size-350;
+  font-weight: globals.$ionic-font-weight-medium;
```
brandyscarney, line 12: Should this use `body/md/medium` instead?

thetaPC: Good catch! https://github.com/ionic-team/ionic-framework/pull/29907/commits/668fe1b6ee331ae863f604caf5ce23d7f17e9f15

brandyscarney (approved):

PR #29909, opened 2024-09-27
test(list): update snapshots

thetaPC:
Issue number: internal

---------

The snapshots for list are not accurate. This can be verified by comparing it to the test pages, which do import the right bundle file.

Since the `ion-item` and `ion-list-header` were updated, then the snapshots for `ion-list` would also need to be updated due to `ion-list` using `ion-item` and `ion-list-header`.

## What is the new behavior?

- Updated snapshots

- [ ] Yes
- [x] No

## Other information

[List header branch](https://github.com/ionic-team/ionic-framework/pull/29907)
[Item branch](https://github.com/ionic-team/ionic-framework/pull/29905)


core/src/components/list/test/shape/list.e2e.ts-snapshots/list-shape-rectangular-with-header-ionic-md-ltr-light-Mobile-Chrome-linux.png
thetaPC, line 1: It seems that there was a regression at some point because the header should not have a white background, should be transparent. We are now finding about this because the correct bundle file is being used. A different ticket will be in charge of the fix.

PR #29913, opened 2024-09-27
test(toggle): update snapshots

thetaPC:
Issue number: internal

---------

The snapshots for toggle are not accurate. This can be verified by comparing it to the test pages, which do import the right bundle file.

## What is the new behavior?

- Updated snapshots

- [ ] Yes
- [x] No

## Other information


core/src/components/toggle/test/enable-on-off-labels/toggle.e2e.ts-snapshots/toggle-on-off-labels-color-ionic-md-ltr-dark-Mobile-Chrome-linux.png
brandyscarney, line 1: Is dark mode not expected to work due to the new file imports?

thetaPC: Honestly not sure why it had a black background in the first place because when you [comment out the bundle files](https://github.com/ionic-team/ionic-framework/blob/85a45ff3077b2ed1f9050fd0183269b65ab9bf0b/core/scripts/testing/scripts.js#L43), the background is still white in the test pages.

`ionic` theme hasn't been set up for dark mode. For example, the list background will always be [white](https://github.com/ionic-team/ionic-framework/blob/85a45ff3077b2ed1f9050fd0183269b65ab9bf0b/core/src/components/list/list.ionic.scss#L14) regardless of palette. If you compare this to `md`, that theme is changing the background based on a [global variable](https://github.com/ionic-team/ionic-framework/blob/85a45ff3077b2ed1f9050fd0183269b65ab9bf0b/core/src/themes/native/native.theme.default.md.scss#L26). The value for this variable changes based on the palette file.

brandyscarney (approved):

PR #29922, opened 2024-10-02
feat(toast): add the default styles for ionic theme

thetaPC:
Issue number: internal

---------

Toast does not have the `ionic` theme styles.

## What is the new behavior?

- Added `ionic` default styles for toast
- Added snapshots

- [ ] Yes
- [x] No

## Other information

[Preview](https://ionic-framework-git-rou-11198-ionic1.vercel.app/src/components/toast/test/basic?ionic:theme=ionic)


core/src/components/toast/toast.ionic.scss
```diff
@@ -0,0 +1,76 @@
...
+
+.toast-content {
+  @include globals.padding(globals.$ionic-space-300, globals.$ionic-space-400);
+}
+
+// Toast Button
+// --------------------------------------------------
+
+.toast-layout-baseline .toast-button-group-start {
+  @include globals.margin(null, null, null, globals.$ionic-space-400);
```
brandyscarney, line 43: Should we do `gap` here instead?

core/src/components/toast/toast.ionic.scss
```diff
@@ -0,0 +1,76 @@
...
+
+  overflow: hidden;
+}
+
+// Toast Icon
+// --------------------------------------------------
+
+.toast-icon,
+.toast-button-icon {
+  font-size: globals.$ionic-font-size-600;
```
brandyscarney, line 75: Should this be `scale` instead?

thetaPC: https://github.com/ionic-team/ionic-framework/pull/29922/commits/d931e275743196e1602ab66f82fdaa5952f23f52

core/src/components/toast/toast.ionic.scss
thetaPC: No, if we replace margin with gap then there wouldn't be any spacing at the beginning. Here's an example:

Screenshot 2024-10-03 at 4 42 32 PM

This is because gap only adds spacing between and not at the ends. I did realize that gap should go within the button groups though: https://github.com/ionic-team/ionic-framework/pull/29922/commits/f0bc6f0b53c4dc4bdcaf085c18cf3161d4c6a70f

brandyscarney: Ahh yeah sorry I commented on the wrong style! I meant the gap between the button and the text.

brandyscarney (approved):

core/src/components/toast/toast.ionic.scss
thetaPC: lol, you were just testing me

tanner-reits (changes requested):
Think I found one discrepancy with the designs. If there is an icon the left (in LTR mode, before the text) the distance between the icon and text should be 8px but right now it's 16px. Otherwise, looks good!

tanner-reits (approved):

Issue #29885, opened 2024-09-20
bug: ion-tab-bar is not working when used as a standalone Vue component after the v8.3.0/v8.3.1 update


ozgurg:
### Prerequisites

- [X] I have read the [Contributing Guidelines](https://github.com/ionic-team/ionic-framework/blob/main/docs/CONTRIBUTING.md#creating-an-issue).
- [X] I agree to follow the [Code of Conduct](https://ionicframework.com/code-of-conduct).
- [X] I have searched for [existing issues](https://github.com/ionic-team/ionic-framework/issues) that already report this problem, without success.

### Ionic Framework Version

v8.x

### Current Behavior

If the `ion-tab-bar` component is extracted as a standalone Vue component, it does not change the route and is basically broken after the v8.3.0/v8.3.1 update.

### Expected Behavior

Works as expected, even when used as a standalone component, like before version v8.3.0/v8.3.1.

### Steps to Reproduce

**Folder Information:**

- ✅ **[8.2.9](https://github.com/ozgurg/ionic-8.3.0-ion-tab-bar-vue-component-reproduce/tree/main/8.2.9)**: The `ion-tab-bar` is in [views/TabsPage.vue](https://github.com/ozgurg/ionic-8.3.0-ion-tab-bar-vue-component-reproduce/blob/main/8.2.9/src/views/TabsPage.vue) and it works.

- ✅ **[8.2.9-standalone](https://github.com/ozgurg/ionic-8.3.0-ion-tab-bar-vue-component-reproduce/tree/main/8.2.9-standalone)**: The `ion-tab-bar` is extracted as-is to [components/TabBar.vue](https://github.com/ozgurg/ionic-8.3.0-ion-tab-bar-vue-component-reproduce/blob/main/8.2.9-standalone/src/components/TabBar.vue) and it works.

- ✅ **[8.3.0](https://github.com/ozgurg/ionic-8.3.0-ion-tab-bar-vue-component-reproduce/tree/main/8.3.0)**: The `ion-tab-bar` is in [views/TabsPage.vue](https://github.com/ozgurg/ionic-8.3.0-ion-tab-bar-vue-component-reproduce/blob/main/8.3.0/src/views/TabsPage.vue) and it works.

[...]

PR #29925, opened 2024-10-07
fix(tabs, tab-bar): use standalone tab bar in Vue

thetaPC:
Issue number: resolves part of #29885

---------

Vue: Tab bar could be a standalone element within `ion-tabs` and would navigate without issues with a router outlet before v8.3:

```vue
<ion-tabs>
  <ion-router-outlet></ion-router-outlet>

  <TabBar></TabBar>
</ion-tabs>
```

It would work as if it was written as:

```vue
<ion-tabs>
  <ion-router-outlet></ion-router-outlet>

  <ion-tab-bar slot="bottom">
    <!-- Buttons -->
  </ion-tab-bar>
</ion-tabs>
```

After v8.3, any `ion-tab-bar` that was not a direct child of `ion-tabs` would lose it's expected behavior when used with a router outlet. If a user clicked on a tab button, then the content would not be redirected to that expected view.

## What is the new behavior?

The Vue tab bar's `_hasRouterOutlet` prop has been modified to accept both Boolean and `undefined` values. When the value is `undefined`, it indicates that the `ion-tab-bar` component was not found within the slotted content of `ion-tabs`. This typically occurs when the tab bar is not passed as a direct child of `ion-tabs`.

In such cases, the tab bar will perform its own check to determine if an `ion-router-outlet` is present. This check cannot be handled by `tabs`, as it would need to wait until after mounting to access the DOM, which would be too late since the tab bar would have already executed the logic that depends on `_hasRouterOutlet`.

- [ ] Yes
- [x] No

## Other information

Dev build: `8.3.3-dev.11728341918.1d3d4b3f`

How to test:

1. Clone the issue repro
2. Navigate to folder `8.3.0-standalone`
3. Install deps
4. Start app
5. Click on the tab buttons
6. Notice the content and URL doesn't change
7. Install dev build: `npm install @ionic/vue@8.3.3-dev.11728341918.1d3d4b3f`
8. Start app
9. Click on the tab buttons
10. Verify that the correct content and URL displays

### DIFF
--- a/packages/vue/src/components/IonTabBar.ts
+++ b/packages/vue/src/components/IonTabBar.ts
@@ -259,9 +259,8 @@ export const IonTabBar = defineComponent({
      * the provide/inject.
      */
     const tabBarData = inject<Ref<TabBarData>>("tabBarData");
-    const hasRouterOutlet = tabBarData.value.hasRouterOutlet;
-    this.$data.tabState.hasRouterOutlet = hasRouterOutlet;
 
+    this.$data.tabState.hasRouterOutlet = tabBarData.value.hasRouterOutlet;
     this.$data._tabsWillChange = tabBarData.value._tabsWillChange;
     this.$data._tabsDidChange = tabBarData.value._tabsDidChange;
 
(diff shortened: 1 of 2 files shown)


manhtruongwang:
Do this also fix react version?

thetaPC:
Hello @manhtruongwang! This PR only addresses the issue for Vue. A separate PR will be created for React.

tanner-reits (approved):
LGTM, confirmed it fixes the issue repro

tanner-reits (approved):
Don't have a lot of Vue experience, but provide/inject seems like a good approach. Confirmed this fixes the standalone reproduction and that the router outlet approach still works as well. Nice work 👍

PR #29931, opened 2024-10-10
test(styles): update button styles for test pages

thetaPC:
Issue number: internal

---------

Test styles causes native buttons to have [certain styling](https://github.com/ionic-team/ionic-framework/blob/main/core/scripts/testing/styles.css#L52-L64). This was done to spruce up the buttons used for testing purposes only. However, this ended up adding styles to native buttons within Ionic components.

## What is the new behavior?

- Test styles for native buttons are only applied to buttons that are not part of a Ionic component

- [ ] Yes
- [x] No

## Other information

How to test:

1. Run the project locally from the `main` branch
2. [Comment](https://github.com/ionic-team/ionic-framework/blob/5d208e9daacad63294489e0b1bd6ec2b2645babd/core/src/components/searchbar/searchbar.md.scss#L91) out `border: 0` from `.searchbar-clear-button` from the searchbar `md` theme file (`ios` also works)
3. Navigate to the basic test page: `/src/components/searchbar/test/basic`
4. Notice a teal border around the clear buttons
5. Checkout to this PR's branch
6. Make sure steps 2-3 are done
7. Verify that the teal border is not being applied to the clear buttons
8. Verify that only native buttons outside of the Ionic components have a teal appearance: `/src/components/loading/test/standalone` and `/src/components/action-sheet/test/is-open`

### DIFF
--- a/core/src/components/popover/test/size/index.html
+++ b/core/src/components/popover/test/size/index.html
@@ -16,7 +16,7 @@
       ion-app > ion-content {
         --background: #dddddd;
       }
-      ion-content button {
+      ion-content button.trigger {
         padding: 12px 16px;
       }
       .grid {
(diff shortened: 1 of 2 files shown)


core/src/components/popover/test/size/index.html
```diff
@@ -16,7 +16,7 @@
       ion-app > ion-content {
         --background: #dddddd;
       }
-      ion-content button {
+      ion-content button.trigger {
```
thetaPC, line 19: This change was needed to override the padding from `styles.css`.

BenOsodrac (approved):

tanner-reits (approved):
Tested changes and looks good.

Might be worth noting that this only affected `scoped` components. The global test styles cannot seep into the Shadow DOM of components so they were not being affected. And, the selector used in the fixed (`sc-ion-*`) is only added to elements that are a part of components with scoped encapsulation.

PR #29932, opened 2024-10-11
feat(modal): add static backdrop behaviour when using ionic theme

OS-pedrolourenco:
Issue number: internal

---------

The modal's backdrop opacity changes dynamically according to the current breakpoint, independently of what theme is being used.

## What is the new behavior?
- The modal's backdrop opacity is now static once it has finished entering the screen and while dragging the modal's handle.

- [ ] Yes
- [x] No


thetaPC:
@OS-pedrolourenco please fix the issues related to snapshots. Here's the [doc](https://github.com/ionic-team/ionic-framework/blob/main/docs/core/testing/usage-instructions.md) that goes over how to run the snapshot tests and how to update them.

core/src/components/modal/animations/sheet.ts
```diff
@@ -12,7 +12,13 @@ export const createSheetEnterAnimation = (opts: ModalAnimationOptions) => {
...
    */
   const shouldShowBackdrop = backdropBreakpoint === undefined || backdropBreakpoint < currentBreakpoint!;
-  const initialBackdrop = shouldShowBackdrop ? `calc(var(--backdrop-opacity) * ${currentBreakpoint!})` : '0';
+
+  const initialBackdrop =
+    opts.theme === 'ionic'
+      ? staticBackdropOpacity
+      : shouldShowBackdrop
+      ? `calc(var(--backdrop-opacity) * ${currentBreakpoint!})`
+      : '0';
```
thetaPC, line 21: Let's expand this to make it easier to read. Nested ternary can be hard to read.

Example:
```suggestion
  let initialBackdrop = '0';
  if (opts.theme === 'ionic') {
     initialBackdrop = staticBackdropOpacity;
  } else if (shouldShowBackdrop) {
     initialBackdrop = `calc(var(--backdrop-opacity) * ${currentBreakpoint!})`;
  }
```

core/src/components/modal/gestures/sheet.ts
```diff
@@ -65,12 +67,18 @@ export const createSheetGesture = (
...
+  ];
+
   const SheetDefaults = {
     WRAPPER_KEYFRAMES: [
       { offset: 0, transform: 'translateY(0%)' },
       { offset: 1, transform: 'translateY(100%)' },
     ],
-    BACKDROP_KEYFRAMES: backdropBreakpoint !== 0 ? customBackdrop : defaultBackdrop,
+    BACKDROP_KEYFRAMES:
+      theme === 'ionic' ? ionicThemeBackdrop : backdropBreakpoint !== 0 ? customBackdrop : defaultBackdrop,
```
thetaPC, line 81: Let's also expand this. Nested ternary can be hard to read.

core/src/components/modal/modal.tsx
```diff
@@ -1048,6 +1052,10 @@ interface ModalOverlayOptions {
    * to fade in when using a sheet modal.
    */
   backdropBreakpoint: number;
+  /**
+   * The theme used by the sheet modal
+   */
+  theme: Theme;
```
thetaPC, line 1058: Instead of passing `theme`, it might be best to pass `initialBackdrop` and `backdropValue`. These two new values would allow any theme to be customizable instead of relying on checking for specific themes.

core/src/components/modal/animations/sheet.ts
```diff
@@ -12,7 +12,13 @@ export const createSheetEnterAnimation = (opts: ModalAnimationOptions) => {
    * current breakpoint, then the backdrop should be fading in.
    */
   const shouldShowBackdrop = backdropBreakpoint === undefined || backdropBreakpoint < currentBreakpoint!;
-  const initialBackdrop = shouldShowBackdrop ? `calc(var(--backdrop-opacity) * ${currentBreakpoint!})` : '0';
+
+  const initialBackdrop =
+    opts.theme === 'ionic'
```
thetaPC, line 17: Let's not using theme to check. I would use `opts.initialBackdrop`. Read the comment below regarding this.

core/src/components/modal/animations/sheet.ts
```diff
@@ -40,10 +46,11 @@ export const createSheetLeaveAnimation = (opts: ModalAnimationOptions) => {
    * is defined, so we need to account for that offset by figuring out
    * what the current backdrop value should be.
    */
-  const backdropValue = `calc(var(--backdrop-opacity) * ${getBackdropValueForSheet(
-    currentBreakpoint!,
-    backdropBreakpoint!
-  )})`;
+  const backdropValue =
+    opts.theme === 'ionic'
```
thetaPC, line 50: Let's not using theme to check. I would use `opts.backdropValue`. Read the comment below regarding this.

thetaPC (changes requested):

thetaPC:
Updated PR title to remove the redundant `modal` mention since scope already let's us know which component the work was done on.

core/src/components/modal/gestures/sheet.ts
thetaPC, line 1: The changes shouldn't be relying on a theme. `createSheetGesture` should also accept `initialBackdropOpacity` and `backdropOpacityValue` so it follows the same setup as `modal/animations/sheet.ts`. Make sure to also add comments of what these values are doing.

```diff
const { gesture, moveSheetToBreakpoint } = createSheetGesture(
  this.el,
  this.backdropEl!,
  wrapperEl,
  initialBreakpoint,
  backdropBreakpoint,
  ani,
  this.sortedBreakpoints,
+  initialBackdropOpacity: this.theme === 'ionic' ? staticBackdropOpacity : undefined,
+  backdropOpacityValue: this.theme === 'ionic' ? staticBackdropOpacity : undefined,
  () => this.currentBreakpoint ?? 0,
  () => this.sheetOnDismiss(),
  (breakpoint: number) => {
    if (this.currentBreakpoint !== breakpoint) {
      this.currentBreakpoint = breakpoint;
      this.ionBreakpointDidChange.emit({ breakpoint });
    }
  }
);
```

thetaPC (changes requested):

thetaPC (changes requested):
Still waiting on the changes I requested, once those are applied, please request me for another review. Thanks.

core/src/components/modal/modal.tsx
```diff
@@ -1044,10 +1047,15 @@ interface ModalOverlayOptions {
...
   /**
-   * The point after which the backdrop will being
+   * The point after which the backdrop will begin
    * to fade in when using a sheet modal.
    */
   backdropBreakpoint: number;
+  /**
+   * Defines whether the backdrop should have a
+   * static opacity. (true for ionic theme only)
+   */
```
thetaPC, line 1057: Suggestion: Maybe the comment should mention that `staticBackdropOpacity` relies on the `--backdrop-opacity` variable.

thetaPC (approved):
LGTM, left one comment but it's not a mandatory change, simply a suggestion.

PR #29933, opened 2024-10-11
feat(accordion): add default styles for ionic theme

brandyscarney:
Issue number: internal

---------

The Accordion uses the Material Design styles for the Ionic theme.

## What is the new behavior?
- Updates Accordion to use the Ionic theme styles defined in Figma
- Adds an e2e test to the basic test to check for text content in the `content` slot
- Adds the `ionic` theme to the `multiple` e2e test
- Adds an e2e test for `states` which screenshots disabled, activated & focused
- Updates Item to accommodate the Accordion styles
- Updates the global core scss file to override the Toggle Icon and Items in an Accordion

- [ ] Yes
- [x] No

## Other information
This includes the changes on #29938, so that needs to be merged first. 

- [Basic Preview](https://ionic-framework-git-rou-11257-ionic1.vercel.app/src/components/accordion/test/basic?ionic:theme=ionic)
- [States Preview](https://ionic-framework-git-rou-11257-ionic1.vercel.app/src/components/accordion/test/states?ionic:theme=ionic)
- [Expand Preview](https://ionic-framework-git-rou-11257-ionic1.vercel.app/src/components/accordion-group/test/expand?ionic:theme=ionic)


core/src/components/accordion/accordion.ionic.scss
```diff
@@ -0,0 +1,83 @@
+@use "../../themes/ionic/ionic.globals.scss" as globals;
+@use "./accordion.common";
+
+// Ionic Accordion
+// --------------------------------------------------
+
+:host {
+  position: relative;
+
+  min-width: 272px;
```
BenOsodrac, line 10: Why so specific? Couldn't it be one of the larger scale values?

BenOsodrac (approved):

core/src/components/accordion/accordion.ionic.scss
brandyscarney: I asked design if we should use a token instead and they said: "in my opinion it should not be a token because it's a random number so I would move forward with hardcoding it as px" 🤷‍♀️

BenOsodrac: _"because it's a random number"_  sure it is, that's what I feared 😂.

Alright then 👍

brandyscarney: No problem I had the same question when I implemented it. 😆

tanner-reits (approved):
LGTM!

PR #29936, opened 2024-10-14
feat(toast): implement new shapes

OS-pedrolourenco:
Issue number: internal

---------

The shape of the toast component could not be customized.

## What is the new behavior?
- I have introduced a new shape property which determines whether the toast widget has soft, round or rectangular corners.

- [ ] Yes
- [x] No


core/src/components/toast/toast.tsx
```diff
@@ -484,6 +493,16 @@ export class Toast implements ComponentInterface, OverlayInterface {
     return buttons;
   }
 
+  private getShape(theme: Theme): string | undefined {
+    const { shape } = this;
+
```
tanner-reits, line 498: Add a check here for if `theme !== 'ionic'` to return `undefined` as we don't support shapes in `ios` or `md` themes yet. Example: https://github.com/ionic-team/ionic-framework/blob/next/core/src/components/modal/modal.tsx#L904-L907

tanner-reits (approved):
Looks good! Nice work 💪 Just one small request, but doesn't impact functionality so won't make it a blocker

thetaPC:
@OS-pedrolourenco I updated the PR title. It had `feat(modal)` which is incorrect since the PR is for toast. Plus I removed the redundant mention of toast, it's not needed since the target component is already in the scope.

core/src/components/toast/test/shape/index.html
```diff
@@ -0,0 +1,62 @@
...
+    <ion-app>
+      <ion-header>
+        <ion-toolbar>
+          <ion-title>Toast - Shape</ion-title>
+        </ion-toolbar>
+      </ion-header>
+
+      <ion-content class="ion-padding" id="content">
+        (Only available for ionic theme)
+        <button
```
thetaPC, line 30: Let's also add a default toast. When it comes to the shapes test pages, we have the default and then the shapes. You can see an example on the Avatar shape page.

core/src/components/toast/test/shape/toast.e2e.ts
```diff
@@ -0,0 +1,73 @@
...
+    });
+
+    test('should render a round toast', async () => {
+      await toastFixture.openToast('#round-shape-toast');
+      await toastFixture.screenshot('round', screenshot);
+    });
+
+    test('should render a rectangular toast', async () => {
+      await toastFixture.openToast('#rect-shape-toast');
+      await toastFixture.screenshot('rect', screenshot);
```
thetaPC, line 70:
```suggestion
      await toastFixture.screenshot('shape-rectangular', screenshot);
```

This should follow the same file structure as the other shapes. Since the file names will be changes, make sure to delete any files with the old structure: `toast-rect-*`, because updating snapshots will not delete them for you.

core/src/components/toast/test/shape/toast.e2e.ts
```diff
@@ -0,0 +1,73 @@
...
+    });
+
+    test('should render a soft toast', async () => {
+      await toastFixture.openToast('#soft-shape-toast');
+      await toastFixture.screenshot('soft', screenshot);
+    });
+
+    test('should render a round toast', async () => {
+      await toastFixture.openToast('#round-shape-toast');
+      await toastFixture.screenshot('round', screenshot);
```
thetaPC, line 65:
```suggestion
      await toastFixture.screenshot('shape-round', screenshot);
```
```suggestion
      await toastFixture.screenshot('shape-rectangular', screenshot);
```

This should follow the same file structure as the other shapes. Since the file names will be changes, make sure to delete any files with the old structure: `toast-round-*`, because updating snapshots will not delete them for you.

core/src/components/toast/test/shape/toast.e2e.ts
```diff
@@ -0,0 +1,73 @@
...
+    let toastFixture: ToastFixture;
+
+    test.beforeEach(async ({ page }) => {
+      toastFixture = new ToastFixture(page);
+      await toastFixture.goto(config);
+    });
+
+    test('should render a soft toast', async () => {
+      await toastFixture.openToast('#soft-shape-toast');
+      await toastFixture.screenshot('soft', screenshot);
```
thetaPC, line 60:
```suggestion
      await toastFixture.screenshot('shape-soft', screenshot);
```
```suggestion
      await toastFixture.screenshot('shape-rectangular', screenshot);
```

This should follow the same file structure as the other shapes. Since the file names will be changes, make sure to delete any files with the old structure: `toast-soft-*`, because updating snapshots will not delete them for you.

core/src/components/toast/toast.tsx
thetaPC: I agree with Tanner. Make sure to add this check and also create the ticket.

core/src/components/toast/toast.tsx
```diff
@@ -484,6 +493,16 @@ export class Toast implements ComponentInterface, OverlayInterface {
     return buttons;
   }
 
+  private getShape(theme: Theme): string | undefined {
```
thetaPC, line 496: Theme does not need to be passed. It can be done through `const theme = getIonTheme(this);`

thetaPC (changes requested):

core/src/components/toast/toast.tsx
```diff
@@ -71,6 +71,9 @@ export class Toast implements ComponentInterface, OverlayInterface {
   private customHTMLEnabled = config.get('innerHTMLTemplatesEnabled', ENABLE_HTML_CONTENT_DEFAULT);
   private durationTimeout?: ReturnType<typeof setTimeout>;
   private gesture?: Gesture;
+  private get theme(): Theme {
+    return getIonTheme(this);
+  }
```
thetaPC, line 76: Let's remove this and just use `getIonTheme` within `getShape` because the structure would be consistent and because this function is redundant to `getIonTheme`.

```suggestion
```

core/src/components/toast/toast.tsx
```diff
@@ -484,6 +496,21 @@ export class Toast implements ComponentInterface, OverlayInterface {
...
 
+  private getShape(): string | undefined {
+    const { theme, shape } = this;
+
+    // TODO(ROU-11300): Remove theme check when shapes are defined for all themes.
+    if (theme !== 'ionic') {
+      return undefined;
+    }
+
+    if (theme === 'ionic' && shape === undefined) {
```
thetaPC, line 507:
```suggestion
    if (shape === undefined) {
```
There's no need to do this extra check when the first if statement verifies that only `ionic` would be the theme at this point.

core/src/components/toast/toast.tsx
```diff
@@ -609,11 +636,12 @@ export class Toast implements ComponentInterface, OverlayInterface {
   };
 
   renderButtons(buttons: ToastButton[], side: 'start' | 'end') {
+    const { theme } = this;
```
thetaPC, line 639: Read comment above.

```suggestion
    const theme = getIonTheme(this);
```

core/src/components/toast/toast.tsx
```diff
@@ -691,11 +719,11 @@ export class Toast implements ComponentInterface, OverlayInterface {
   }
 
   render() {
-    const { layout, el, revealContentToScreenReader, header, message } = this;
+    const { layout, el, revealContentToScreenReader, theme, header, message } = this;
```
thetaPC, line 722: Read comment above.

```suggestion
    const { layout, el, revealContentToScreenReader, header, message } = this;
```

core/src/components/toast/toast.tsx
```diff
@@ -691,11 +719,11 @@ export class Toast implements ComponentInterface, OverlayInterface {
...
     const allButtons = this.getButtons();
     const startButtons = allButtons.filter((b) => b.side === 'start');
     const endButtons = allButtons.filter((b) => b.side !== 'start');
-    const theme = getIonTheme(this);
+    const shape = this.getShape();
```
thetaPC, line 726: Read comment above regarding theme.

```suggestion
    const theme = getIonTheme(this);
    const shape = this.getShape();
```

thetaPC (changes requested):

thetaPC (changes requested):
Still waiting on the changes I requested, once those are applied, please request me for another review. Thanks.

core/src/components/toast/toast.tsx
OS-pedrolourenco: The reason for this method is because I was concerned that the theme could change in runtime (hence I did not create a simple field) and to avoid code repetition. However, I get your feedback and will follow your suggestion.

thetaPC (approved):
LGTM

PR #29941, opened 2024-10-16
fix(SafeAreas): fix iOS fallback

JoaoFerreira-FrontEnd:
Issue number: internal

---------
- fix ion-statusbar css var fallback;

- CSS var --ion-statusbar-padding was not working because the Sass variable was not escaped.

- [ ] Yes
- [x] No


BenOsodrac (approved):

gnbm (approved):
LGTM 📱

PR #29945, opened 2024-10-17
feat(toggle): add iOS 18 haptic feedback

### LINKED ISSUE #29942: feat: add haptic feedback to Toggle in Safari on iOS 18+
### Prerequisites

Safari since iOS 18 [supports](https://webkit.org/blog/15865/webkit-features-in-safari-18-0/#html) haptic feedback on checkbox elements via a non-standard `switch` attribute. We could use it to provide haptic feedback when switching Toggle  in non-hybrid environments. This could be a workaround for lack of support for Vibration API in Safari.

### Describe the Use Case

Adding haptic feedback to Toggle component in non-hybrid environments would improve native feel of web and PWA apps built with Ionic.

To trigger haptic feedback programatically we need an `<input type=checkbox switch>` element with an associated `<label>` element. With this setup, when we programatically trigger click() on the label, safari will emit the haptic feedback. 

### Describe Alternatives

Keep current implementation and wait for Safari to implement Vibration API (if it ever does).

### Related Code

Codepen provided in WebKit release notes, demonstrating haptic feedback in checkbox of type switch: https://codepen.io/jensimmons/pen/GReLKWg

### Additional Information

_No response_

jedlikowski:
Issue number: resolves #29942

---------

[iOS 18 has haptics on toggles](https://webkit.org/blog/15865/webkit-features-in-safari-18-0/#html), making a vibration when the user taps on a toggle. However, Ionic Framework does not match the implementation.

## What is the new behavior?

- Added haptics to the `onClick` function for iOS mode only. This works when the iOS device has haptics enabled.

- [ ] Yes
- [x] No

## Other information

Dev build: `8.5.3-dev.11743549985.1c03dfea`


core/src/components/toggle/toggle.tsx
```diff
@@ -312,7 +343,8 @@ export class Toggle implements ComponentInterface {
           [`toggle-${rtl}`]: true,
         })}
       >
-        <label class="toggle-wrapper">
+        {this.renderFallbackHapticElements()}
+        <label class="toggle-wrapper" onClick={this.onClick}>
```
jedlikowski, line 347: `onClick listener` moved to label, without this change `Host` was capturing clicks on the haptic `<label>` too, not only on `.toggle-wrapper`

core/src/components/toggle/toggle.tsx
```diff
@@ -285,6 +290,33 @@ export class Toggle implements ComponentInterface {
...
+   * On Safari (iOS 18+) we can trigger haptic feedback programatically
+   * by rendering <input type="checkbox" switch> element
+   * with an associated <label>  and triggering click()
+   * on the <label> element.
+   */
+  private renderFallbackHapticElements() {
+    const { inputId } = this;
+    const mode = getIonMode(this);
+
+    if (hapticAvailable() || mode !== 'ios') {
```
jedlikowski, line 303: If I understand correctly, we don't need these elements if Vibration API is available, because then `hapticSelection()` will work.

thetaPC:
Hello! I've [responded](https://github.com/ionic-team/ionic-framework/issues/29942#issuecomment-2734698612) to the [feature request](https://github.com/ionic-team/ionic-framework/issues/29942) for more information.

core/src/components/toggle/toggle.tsx
```diff
@@ -247,6 +248,14 @@ export class Toggle implements ComponentInterface {
   }
 
   private onClick = (ev: MouseEvent) => {
+    /**
+     * The haptics for the toggle is
+     * an iOS-only feature when tapped.
+     * As a result, it should be
+     * disabled on Android.
```
ShaneK, line 255:
```suggestion
     * The haptics for the toggle on tap is
     * an iOS-only feature. As such, it should
     * only trigger on iOS.
```

Just trying to clear up the grammar here a little bit to make the intention clearer.

ShaneK (approved):
Minor grammar nit you can take or leave, but looks good to me!

brandyscarney (approved):

thetaPC:
@jedlikowski Thank you for the PR! Also congrats on your first PR with Ionic Framework!!

thetaPC:
@jedlikowski I used `hapticSelection()` to:
- keep consistency with how we handle haptics.
- lessen the potential risk of adding a hidden input and depending on the click. Would it cause accessibility issues? Could the event trigger unexpected issues? Etc.
- not cause a regression. Using the hidden input lead to toggles within an item to not click correctly when a user wants to click on the item to change the toggle's state.

You will need to test the haptics on a real physical device, especially since it needs to check if the device has haptics enabled. This follows how native iOS handles the toggle vibration.

If you believe that this solution should be altered to handle other cases, then please open a new issue with use cases.

Issue #29449, opened 2024-05-03
bug: ion-input-password-toggle causes label to float even when input is empty


bodinaren:
### Prerequisites

- [X] I have read the [Contributing Guidelines](https://github.com/ionic-team/ionic-framework/blob/main/.github/CONTRIBUTING.md#creating-an-issue).
- [X] I agree to follow the [Code of Conduct](https://ionicframework.com/code-of-conduct).
- [X] I have searched for [existing issues](https://github.com/ionic-team/ionic-framework/issues) that already report this problem, without success.

### Ionic Framework Version

v8.x

### Current Behavior

When an `<ion-input type="password" label-placement="floating">` is empty but has a `ion-input-password-toggle` the label is floating, something which normally only happens when the input has a value.

### Expected Behavior

I expect the label to be placed the same as if the `ion-input-password-toggle` wasn't used so that all empty inputs looks the same.

### Steps to Reproduce

The problem is immediately visible in the 3rd input in the stackblitz reproduction.

### Code Reproduction URL

https://stackblitz.com/edit/angular-k1wdev?file=src%2Fapp%2Fexample.component.html

### Ionic Info

Ionic:

   Ionic CLI       : 7.2.0 (C:\Users\bodin\AppData\Roaming\npm\node_modules\@ionic\cli)
   Ionic Framework : @ionic/react 7.5.3

Capacitor:

   Capacitor CLI      : 5.5.1
   @capacitor/android : 5.5.1
   @capacitor/core    : 5.5.1
   @capacitor/ios     : 5.5.1

Utility:

   cordova-res : not installed globally
   native-run  : 1.7.3

System:

   NodeJS : v20.11.1 (C:\Program Files\nodejs\node.exe)
   npm    : 10.7.0
   OS     : Windows 10 (actually 11, Ionic info reports wrong)

### Additional Information

Given the comment here the error appears to be a known one, or at least an expected one, but has remained untouched for 6 months.

https://github.com/ionic-team/ionic-framework/blob/main/core/src/components/input/input.tsx#L718-L721

thetaPC:
Thank you for submitting the issue!

I was able to confirm that this is an issue. We plan on allowing labels to float when slots are present. We'll update this when there's more information.

yannisvanlaerhoven3:
Would love a fix 🙏

ErikOnBike:
Pretty please?
[...]

PR #29948, opened 2024-10-18
fix(input): do not float label because of slotted items

### LINKED ISSUE #28665: bug: input floating label is always floating when using start/end slots
### Prerequisites

### Ionic Framework Version

v7.x

### Current Behavior

`ion-input` with `labelPlacement="floating"` not working as intended with slotted `ion-button` inside. The label is fixed on top even with pristine input.

input bug

The label of the first `ion-input` without slotted button is correct.

### Expected Behavior

When `ion-input` has or loses focus the label should change its position.

### Steps to Reproduce

1. create an `ion-input` wit `labelPlacement="floating"`
2. create an `ion-button` with `slot="end"` inside

### Code Reproduction URL

_No response_

### Ionic Info

Ionic:
   Ionic CLI                     : 7.1.5
   Ionic Framework               : @ionic/angular 7.6.0
   @angular-devkit/build-angular : 16.2.8
   @angular-devkit/schematics    : 16.2.8
   @angular/cli                  : 16.2.8

Capacitor:
   Capacitor CLI      : 5.5.1
   @capacitor/android : 5.5.1
   @capacitor/core    : 5.5.1
   @capacitor/ios     : 5.5.1

Utility:
   native-run  : 1.7.4

System:
   NodeJS : v20.10.0
   npm    : 10.2.3
   OS     : Windows 10

### Additional Information

_No response_

ErikOnBike:
Issue number: resolves #28665

---------

The input label is floating if an element is slotted. This is especially annoying if a password toggle is present, since it will start floating the label (even when no input is available).

## What is the new behavior?

A label will only float (if specified and) if it has a value or received focus.

- [ ] Yes
- [X] No

## Other information

Screenshot before:
image

Screenshot after:
image

This issue also seems to fix issue #29449

### DIFF
--- a/core/package-lock.json
+++ b/core/package-lock.json
@@ -672,31 +672,28 @@
       }
     },
     "node_modules/@capacitor/haptics": {
-      "version": "7.0.0",
-      "resolved": "https://registry.npmjs.org/@capacitor/haptics/-/haptics-7.0.0.tgz",
-      "integrity": "sha512-8uI8rWyAbq8EzkjS+sHZSncyzujHzVbuLKgj8J5H0yUL6+r26F16gVA2iuQuIBvzbSMy7Y0/pUuWlwZr/H8AKg==",
+      "version": "7.0.1",
+      "resolved": "https://registry.npmjs.org/@capacitor/haptics/-/haptics-7.0.1.tgz",
+      "integrity": "sha512-ewZmspE5krgDUj5ZvUDcfNZvgerAIr+3bDSk6DLzyvBZ/dYmr/tMLu5H6WtYaaKYZJ32aZAudGpIal5epDyBYA==",
       "dev": true,
-      "license": "MIT",
       "peerDependencies": {
         "@capacitor/core": ">=7.0.0"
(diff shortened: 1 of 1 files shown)


ErikOnBike:
Changed the visual removal of slotted items to use 'visibility: hidden' to prevent changing the input component's size where 'display: none' could change the size.

ErikOnBike:
> Hey @ErikOnBike
> 
> Thanks for taking a stab at this! Unfortunately, hiding the slotted content until focus doesn't match the designs for Material Design (see demo [here](https://material-components.github.io/material-components-web-catalog/#/component/text-field?icons=leadingIcon%2CtrailingIcon&type=outlined)). We have a task to look into this internally, but I'm not sure when we'll be able to do so if you want to try to match that behavior.

Hi @tanner-reits,

You're completely right. I think I mixed up two issues and tried to solve both in one go. (Not very sensible ;-)
Okay if I only fix the real issue here: prevent label from floating if no value is entered in the input field? So the icons remain.

PR #29951, opened 2024-10-18
fix(overlay): hide from screen readers while animating

### LINKED ISSUE #29857: bug: Focus state at ion-select with interface of action-sheet using TalkBack
### Prerequisites

### Ionic Framework Version

v7.x

### Current Behavior

We can clearly see that when **TalkBack** is turned on and start to read options in **ion-select**, focus state of HTML element is out of bounds.

**Important to note** 
This is not happening always, the best (to reproduce this) will be to test it when you scroll page a bit, and then open a **ion-select**

https://github.com/user-attachments/assets/39ae8f9a-6c12-43bd-95db-2ff2ac52094b

### Expected Behavior

focus of HTML element to be around correctly focused option

### Steps to Reproduce

1. Go to https://ionicframework.com/docs/api/select
2. Find example with 'action-sheet' example
3. Test it

### Code Reproduction URL

https://angular-kzrikb.stackblitz.io/

### Ionic Info

Ionic:

   Ionic CLI                     : 7.0.1 
   Ionic Framework               : @ionic/angular 7.6.2
   @angular-devkit/build-angular : 17.0.8
   @angular-devkit/schematics    : 17.0.8
   @angular/cli                  : 17.0.8

Capacitor:

   Capacitor CLI      : 5.6.0
   @capacitor/android : 5.6.0
   @capacitor/core    : 5.6.0
   @capacitor/ios     : 5.6.0

Cordova:

Utility:

   native-run  : 2.0.0

System:

   NodeJS : v18.20.0 
   npm    : 10.5.0
   OS     : Windows 10

### Additional Information

_No response_

thetaPC:
Issue number: resolves #29857 

---------

Screen readers like Android Talkback would not have the focus ring on the correct element. For example, Talkback would announce the buttons correctly within action sheet but the focus ring was no where to be seen.

After digging around, the focus rings are located out of screen because the action sheet is mounted to the DOM out of the screen first then transitions into the screen. There are some screen readers that do not behave as expected when an element uses `transform` styles like action sheet.

https://github.com/user-attachments/assets/5a700bcc-3149-47a9-96ff-0aef99dd2bb3

## What is the new behavior?

- When an overlay is being animated (presenting or dismissing), the overlay will hide from screen readers. This allows the element to navigate to its correct destination for screen readers to interact with. Plus, we shouldn't allow screen readers to interact with content in the middle of an animation. It could lead to some confusion.

- [ ] Yes
- [x] No

## Other information

Dev build: 8.3.3-dev.11729276019.194c165c

**A physical Android device will be needed, the issue does not appear in simulators**

Components that need to be tested because they use overlays:
- Action sheet
- Alert
- Loading
- Modal
- Popover
- Select w/ action sheet interface
- Select w/ alert interface
- Select w/ popover interface
- Toast

How to test:
1. Create a starter app (any framework will do)
2. Add an action sheet
3. Build app for mobile devices
```
ionic build

ionic cap add ios
ionic cap add android

ionic cap copy && ionic cap sync
```
4. Open the app in Android Studio: `ionic cap open android`
5. Connect the Android device to Android Studio
6. Open app in Android device
7. Launch Talkback
8. Navigate back to the app
9. Open action sheet 
10. Swipe over to the action sheet's buttons
11. Notice that the buttons don't have a focus ring
12. Go back to the starter
4. Install the dev build
5. Add the components to the app
6. Sync app: `ionic cap copy && ionic cap sync`
13. Relaunch the app on the Android device
14. Verify that the focus ring appears on the action sheet's buttons
15. Verify that the other overlays are working as intended

### DIFF
--- a/core/src/utils/overlays.ts
+++ b/core/src/utils/overlays.ts
@@ -958,6 +958,10 @@ export const createTriggerController = () => {
 const hideAnimatingOverlayFromScreenReaders = (overlay: HTMLIonOverlayElement) => {
   if (doc === undefined) return;
 
+  /**
+   * Once the animation is complete, this attribute will be removed.
+   * This is done at the end of the `present` method.
+   */
   overlay.setAttribute('aria-hidden', 'true');
 };
 
--- a/core/src/utils/overlays.ts
+++ b/core/src/utils/overlays.ts
@@ -514,7 +514,20 @@ export const present = async <OverlayPresentOptions>(
 
   document.body.classList.add(BACKDROP_NO_SCROLL);
 
-  hideOverlaysFromScreenReaders(overlay.el);
+  hideUnderlyingOverlaysFromScreenReaders(overlay.el);
+
+  /**
+   * Set aria-hidden="true" to hide the overlay from screen readers
+   * during the animation. This ensures that assistive technologies
+   * like TalkBack do not announce or interact with the content until
+   * the animation is complete, avoiding confusion for users.
+   *
+   * Additionally, it prevents focus rings from appearing in incorrect
+   * positions due to the transition (specifically `transform` styles),
+   * ensuring that when aria-hidden is removed, the focus rings are
+   * correctly displayed in the final location of the elements.
+   */
+  overlay.el.setAttribute('aria-hidden', 'true');
 
   overlay.presented = true;
   overlay.willPresent.emit();
@@ -939,7 +952,7 @@ export const createTriggerController = () => {
  * @param newTopMostOverlay - The overlay that is being presented. Since the overlay has not been
  * fully presented yet at the time this function is called it will not be included in the getPresentedOverlays result.
  */
-const hideOverlaysFromScreenReaders = (newTopMostOverlay: HTMLIonOverlayElement) => {
+const hideUnderlyingOverlaysFromScreenReaders = (newTopMostOverlay: HTMLIonOverlayElement) => {
   if (doc === undefined) return;
 
   const overlays = getPresentedOverlays(doc);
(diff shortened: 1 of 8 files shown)


core/src/utils/overlays.ts
```diff
@@ -514,7 +514,8 @@ export const present = async <OverlayPresentOptions>(
 
   document.body.classList.add(BACKDROP_NO_SCROLL);
 
-  hideOverlaysFromScreenReaders(overlay.el);
+  hideUnderlyingOverlaysFromScreenReaders(overlay.el);
```
thetaPC, line 517: Just changed the name to distinguish between this function and the next.

core/src/utils/overlays.ts
```diff
@@ -929,6 +942,25 @@ export const createTriggerController = () => {
...
+ * in incorrect positions due to the transition (specifically `transform`
+ * styles), ensuring that when aria-hidden is removed, the focus rings are
+ * correctly displayed in the final location of the elements.
+ *
+ * @param overlay - The overlay that is being animated.
+ */
+const hideAnimatingOverlayFromScreenReaders = (overlay: HTMLIonOverlayElement) => {
+  if (doc === undefined) return;
+
+  overlay.setAttribute('aria-hidden', 'true');
```
christian-bromann, line 961: Optional: maybe add a comment referencing where the attribute is being added. My initial thought was: when we set it here, will it ever be removed somewhere else?

thetaPC: https://github.com/ionic-team/ionic-framework/pull/29951/commits/bc2c086f1d73300a8cef3650482a1a36a9700734

christian-bromann (approved):
LGTM 👍

PR #29954, opened 2024-10-21
feat(menu): add support for passing role to close events

brandyscarney:
Issue number: internal

---------

The menu `ionWillClose` and `ionDidClose` events do not emit any details.

## What is the new behavior?
- Adds the `MenuCloseEventDetail` interface which includes an optional `role` property
- The `ionWillClose` and `ionDidClose` emit the `role` property for the following scenarios:
  - A role of `'gesture'` when dragging the menu closed
  - A role of `'backdrop'` when clicking on the backdrop to close the menu
  - A role of `'backdrop'` when the the menu is closed using the escape key
  - A role of `undefined` when the menu is closed from a button inside of the menu

- [ ] Yes
- [x] No

## Other information

[Preview](https://ionic-framework-git-rou-11148-ionic1.vercel.app/src/components/menu/test/basic)

### DIFF
--- a/core/src/components/menu/menu.tsx
+++ b/core/src/components/menu/menu.tsx
@@ -322,6 +322,20 @@ export class Menu implements ComponentInterface, MenuI {
     }
   }
 
+  @Listen('click', { capture: true })
+  onBackdropClick(ev: any) {
+    // TODO(FW-2832): type (CustomEvent triggers errors which should be sorted)
+    if (this._isOpen && this.lastOnEnd < ev.timeStamp - 100) {
+      const shouldClose = ev.composedPath ? !ev.composedPath().includes(this.menuInnerEl) : false;
+
+      if (shouldClose) {
+        ev.preventDefault();
+        ev.stopPropagation();
+        this.close(undefined, BACKDROP);
+      }
+    }
+  }
+
   onKeydown(ev: KeyboardEvent) {
     if (ev.key === 'Escape') {
       this.close(undefined, BACKDROP);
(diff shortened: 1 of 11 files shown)


core/src/components/menu/menu.tsx
```diff
@@ -322,23 +322,9 @@ export class Menu implements ComponentInterface, MenuI {
...
-        ev.stopPropagation();
-        this.close();
-      }
-    }
-  }
-
   onKeydown(ev: KeyboardEvent) {
     if (ev.key === 'Escape') {
-      this.close();
+      this.close(undefined, ESCAPE);
```
brandyscarney, line 327: This is different from how the ion-modal works. When the modal is closed using the escape key, it emits a role of 'backdrop'. Should we match the modal behavior here?

core/src/components/menu/menu.tsx
```diff
@@ -731,13 +717,26 @@ export class Menu implements ComponentInterface, MenuI {
...
       // emit close event
-      this.ionDidClose.emit();
+      this.ionDidClose.emit({ role });
 
       // undo focus trapping so multiple menus don't collide
       document.removeEventListener('focus', this.handleFocus, true);
     }
   }
 
+  private onBackdropTap = (ev: any) => {
```
brandyscarney, line 727: I’m not sure if there are any consequences to moving this from the `click` listener, but all the other overlays use `onIonBackdropTap`.

core/src/components/menu/menu.tsx
brandyscarney: Changed it to match the other overlays for consistency: https://github.com/ionic-team/ionic-framework/pull/29954/commits/721bc45f9562acb32b8c9c5fbf156995ab480ca8

core/src/components/menu/menu.tsx
brandyscarney: I was able to revert this change: https://github.com/ionic-team/ionic-framework/pull/29954/commits/f5179d9b729c91762bae12c5aa7f722b39695be2

When I implemented this feature in a different way this change was necessary. However, since I now pass the `role` in the close method, it no longer needs to be called in `onIonBackdropTap`.

thetaPC (approved):
LGTM

tanner-reits (approved):
Looks good!

PR #29955, opened 2024-10-21
fix(modal): fix enter animation when the initial breakpoint value is less than the backdrop breakpoint

OS-pedrolourenco:
Issue number: internal

---------

This PR aims to fix an issue introduced in https://github.com/ionic-team/ionic-framework/pull/29932 where the enter animation for ionic theme was in a broken state when the modal's initial breakpoint value was lesser than the specified backdrop breakpoint. 

## What is the new behavior?
- The erroneous behaviour happened because the shouldShowBackdrop constant had a false value which was being taken into consideration when computing the backdrop opacity for the ionic theme. This PR fixes this by always transitioning the backdrop opacity from 0 to 0.7 once the modal appears, independently of the initial, current or activation breakpoint.

- [ ] Yes
- [x] No

### DIFF
--- a/core/src/components/modal/animations/sheet.ts
+++ b/core/src/components/modal/animations/sheet.ts
@@ -12,9 +12,12 @@ export const createSheetEnterAnimation = (opts: ModalAnimationOptions) => {
    * current breakpoint, then the backdrop should be fading in.
    */
   const shouldShowBackdrop = backdropBreakpoint === undefined || backdropBreakpoint < currentBreakpoint!;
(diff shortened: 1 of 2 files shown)


tanner-reits (approved):
Confirmed the fix and update the PR title.

Only small feedback would be to add a comment to the `backdropBreakpoint` prop on the modal calling out that it is ignored for the `ionic` theme.

thetaPC (approved):
LGTM

Issue #29773, opened 2024-08-15
bug: Accessibility issue: Backdrop in ion-popover component has invalid aria-hidden value 


annamelnyk:
### Prerequisites

- [X] I have read the [Contributing Guidelines](https://github.com/ionic-team/ionic-framework/blob/main/docs/CONTRIBUTING.md#creating-an-issue).
- [X] I agree to follow the [Code of Conduct](https://ionicframework.com/code-of-conduct).
- [X] I have searched for [existing issues](https://github.com/ionic-team/ionic-framework/issues) that already report this problem, without success.

### Ionic Framework Version

v7.x

### Current Behavior

Backdrop (shadow part of ion-popover component) contains attribute **aria-hidden=true**, which is not correct from accessibility purposes.
Console shows error below:
`Blocked aria-hidden on a <ion-backdrop> element because the element that just received focus must not be hidden from assistive technology users. Avoid using aria-hidden on a focused element or its ancestor. Consider using the inert attribute instead, which will also prevent focus. For more details, see the aria-hidden section of the WAI-ARIA specification at ` 

### Expected Behavior

Backdrop (shadow part of ion-popover component) NOT contains attribute **aria-hidden**

### Steps to Reproduce

1. Open the <ion-popover> element
2. Click on backdrop
3. Popover will be dismissed and console error will appear

### Code Reproduction URL

https://stackblitz.com/edit/angular-e2jj3l-tpkxhe?file=src%2Fapp%2Fapp.component.html

### Ionic Info

**Ionic:**

   Ionic CLI                     : 7.1.1
   Ionic Framework               : @ionic/angular 7.5.5
   @angular-devkit/build-angular : 16.0.6
   @angular-devkit/schematics    : 16.0.6
   @angular/cli                  : 16.0.6
   @ionic/angular-toolkit        : 9.0.0

**Capacitor:**
   Capacitor CLI      : 5.0.5
   @capacitor/android : 5.0.5
   @capacitor/core    : 5.0.5
   @capacitor/ios     : 5.0.5

**Utility:**
   cordova-res : not installed globally
   native-run  : 1.7.4

**System:**
   NodeJS : v20.13.1
   npm    : 10.5.2
   OS     : Windows 10

dev-charles15531:
I am not able to see any sort of warning or error in my console when I click the backdrop.
I am using the issue reproduction url you sent.

annamelnyk:
@dev-charles15531 thank you for quick reply.
[...]

PR #29962, opened 2024-10-23
fix(overlays): do not hide root when toast appears

thetaPC:
Issue number: stemmed from #29773

---------

Certain Chrome and Edge versions (confirmed: Chrome v127, v128, v129 and Edge v127) would indicate that certain elements have an accessibility violation:

```
Blocked aria-hidden on a "ELEMENT NAME" element because the element that just received
focus must not be hidden from assistive technology users. Avoid using aria-hidden on a
focused element or its ancestor. Consider using the inert attribute instead, which will also
prevent focus. For more details, see the aria-hidden section of the WAI-ARIA specification at
```

This issue happens when a toast appears and the users clicks on any element that is not related to toast. This is due to the main content having an `aria-hidden` so users should not to be able to interact with any of those elements. This isn't an issue when an overlay uses a backdrop, like `ion-alert` because the backdrop prevents a user from interacting with those elements.

## What is the new behavior?

- When toast is present, the main content no longer has an `aria-hidden`. This aligns with accessibility guidelines. I also verified with other Framework, MD states "Don't trap focus in the snackbar. Users should be able to freely navigate in and out."

- [ ] Yes
- [x] No

## Other information

Dev build: `8.3.4-dev.11729879684.1ea28919`

1. Clone this [repo](https://github.com/brandyscarney/test-angular-overlays)
2. Install deps
3. Run the app on a private browser (Chrome v127, v128, v129 or Edge v127)
4. Open browser console
5. Click on "Open Toast" button
6. Click on any element other than "Open Toast" button, like "Open Popover"
7. Notice the error message
8. Close private browser
9. Install dev build: `npm install @ionic/angular@8.3.4-dev.11729879684.1ea28919`
10. Repeat steps 4-7
11. Verify that the error message doesn't occur

### DIFF
--- a/core/src/utils/overlays.ts
+++ b/core/src/utils/overlays.ts
@@ -510,7 +510,17 @@ export const present = async <OverlayPresentOptions>(
     return;
   }
 
-  setRootAriaHidden(true);
+  
+  /**
+   * Due to accessibility guidelines, toasts do not have
+   * focus traps.
+   * 
+   * All other overlays should have focus traps to prevent
+   * the keyboard focus from leaving the overlay.
+   */
+  if (overlay.el.tagName !== 'ION-TOAST') {
+    setRootAriaHidden(true);
+  }
 
   document.body.classList.add(BACKDROP_NO_SCROLL);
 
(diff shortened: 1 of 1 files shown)


BenOsodrac (approved):

brandyscarney (approved):
Looks good!

PR #29964, opened 2024-10-23
fix(alert): use correct heading structure for subHeader when header exists

brandyscarney:
Issue number: internal

---------

Both the `header` and `subHeader` in an Alert render as an `<h2>` element. 

## What is the new behavior?
- The `header` will always render as an `<h2>` element.
- If there is no `header` defined, the `subHeader` will render as an `<h2>` element.
- If there is a `header` defined, the `subHeader` will render as an `<h3>` element.
- Updates `ariaLabelledBy` to include both `header` and `subHeader` ids when both are defined.
- Updates the `a11y` e2e test to use new values & check for tag names. 

- [ ] Yes
- [x] No

## Other information
[Preview](https://ionic-framework-git-rou-11270-ionic1.vercel.app/src/components/alert/test/a11y)


core/src/components/alert/alert.tsx
thetaPC, line 1: Somewhat related but not a blocker: should `.alert-head` div still render if no header or subheader exists? I noticed that it does when looking at test page.

thetaPC (approved):
LGTM

core/src/components/alert/alert.tsx
brandyscarney: Good question! It has some padding that is used in combination with the message. Removing it would result in the message having no padding on the top.

Screenshot 2024-10-24 at 12 54 50 PM

thetaPC: Oh good point! I wonder if we should have that padding on the container instead? Again not a blocker.

brandyscarney: We could probably re-evaluate it for md3 styles.

core/src/components/alert/alert.tsx
```diff
@@ -730,10 +730,12 @@ export class Alert implements ComponentInterface, OverlayInterface {
...
     /**
-     * If the header is defined, use that. Otherwise, fall back to the subHeader.
-     * If neither is defined, don't set aria-labelledby.
+     * Use both the header and subHeader ids if they are defined.
+     * If only the header is defined, use the header id.
+     * If only the subHeader is defined, use the subHeader id.
+     * If neither are defined, do not set aria-labelledby.
      */
-    const ariaLabelledBy = header ? hdrId : subHeader ? subHdrId : null;
+    const ariaLabelledBy = header && subHeader ? `${hdrId} ${subHdrId}` : header ? hdrId : subHeader ? subHdrId : null;
```
tanner-reits, line 738: Chained ternary expressions like this are so gross 🤮

tanner-reits (approved):

PR #29973, opened 2024-10-28
feat(accordion): add shape styles for the ionic theme

brandyscarney:
Issue number: internal

---------

Accordions inside of an inset accordion group add a round border radius regardless of the shape applied.

## What is the new behavior?
- Applies border radius based on the `shape` defined on the parent accordion group
- Adds an e2e test for shape with focused items to show the border radius
- Adds screenshots for the e2e test

- [ ] Yes
- [x] No

## Other information

[Preview](https://ionic-framework-git-rou-11260-ionic1.vercel.app/src/components/accordion/test/shape?ionic:theme=ionic)


core/src/utils/test/playwright/page/utils/set-content.ts
brandyscarney, line 1: This is required so that the tests that use `setContent` will not include the default bundle either.

PR #29974, opened 2024-10-28
feat(css): add CSS utility classes to the ionic theme bundle to match other themes

brandyscarney:
Issue number: internal

---------

The `ionic` theme includes the default (iOS/MD) bundle in the tests. Additionally, the `ionic` theme bundle is missing CSS utility classes provided by the default bundle, so removing that bundle causes issues in several tests.

## What is the new behavior?
The default bundle is removed from the tests for the `ionic` theme because it adds global component styles that the `ionic` theme does not need. The missing utility files are imported, and padding/margin classes are generated from the design tokens, as many tests rely on `ion-padding` and `ion-text-center` being available. This update ensures the `ionic` theme includes the same classes offered in our documentation: https://ionicframework.com/docs/layout/css-utilities.

- [ ] Yes
- [x] No


core/src/css/ionic/utils.bundle.ionic.scss
brandyscarney, line 1: This file includes the CSS utilities documented here: https://ionicframework.com/docs/layout/css-utilities

core/scripts/testing/scripts.js
```diff
@@ -47,6 +47,11 @@
       linkTag.setAttribute('href', '/css/ionic/bundle.ionic.css');
       document.head.appendChild(linkTag);
     }
+
+    const defaultThemeLinkTag = document.querySelector('link[href*="css/ionic.bundle.css"]');
```
BenOsodrac, line 51: @brandyscarney I think this file doesn't exist 🤔, or am I missing something? Maybe that's the problem, and the link on the tests pages need to be updated or removed? 

At least for Ionic Theme tests, all its needed should be the one added on line 47

core/src/css/ionic/utils.bundle.ionic.scss
```diff
@@ -1 +1,4 @@
 @forward "../../foundations/ionic.utility";
+
+// CSS Utils
+@import "../utils.bundle";
```
BenOsodrac, line 4: I'm ok with this, as mostly this brings agnostic classes. But this also brings the padding utility-classes file, that bring padding values that might not match the ones from the new Design System and we are already creating those in the _/foundations/ionic.utility_

I would instead individually import the others, like:
@import "../float-elements";
@import "../text-alignment";
@import "../text-transformation";
@import "../flex-utils";
@import "../display";

BenOsodrac, line 4:
```suggestion
@import "../float-elements";
@import "../text-alignment";
@import "../text-transformation";
@import "../flex-utils";
@import "../display";
```

BenOsodrac (changes requested):

BenOsodrac (approved):

tanner-reits (approved):
I think this is g2g

core/scripts/tokens/index.js
```diff
@@ -23,136 +24,138 @@
...
+            utilityClass = generateShadowUtilityClasses(propName);
+            break;
+          default:
+            utilityClass = generateUtilityClasses(tokenCategory, propName);
+        }
+
+        return otherUtilityClasses.push(utilityClass);
+      });
+
+      const defaultSpaceUtilityClasses = generateDefaultSpaceUtilityClasses();
```
brandyscarney, line 146: These were the only lines added - the rest is formatting

core/scripts/tokens/index.js
```diff
@@ -23,136 +24,138 @@
...
+      otherUtilityClasses.push(defaultSpaceUtilityClasses);
```
brandyscarney, line 147: These were the only lines added - the rest is formatting

brandyscarney:
Requesting re-review because I changed the scope of this PR.

BenOsodrac (approved):

BenOsodrac:
LGTM 🚀

tanner-reits (approved):
LGTM

PR #29977, opened 2024-10-29
fix(theme): fix `rtl` mixin for ionic theme

tanner-reits:

---------

Unable to use the `rtl` mixin from the ionic theme global import:

```scss
@use "../../themes/ionic/ionic.globals.scss" as globals;

:host {
  @include globals.rtl() {
    ...
  }
}
```

## What is the new behavior?

Allows the `rtl` mixin to be accessed from the global sass import

- [ ] Yes
- [x] No


brandyscarney (approved):
LGTM!

thetaPC (approved):
LGTM, thanks!

PR #29983, opened 2024-10-30
refactor(many): remove *.ionic.vars files

brandyscarney:
Issue number: internal

---------

Some components have `component.ionic.vars.scss` files defined for the Ionic theme.

## What is the new behavior?
- Removes the `*.ionic.vars.scss` files for item, list and input and uses the design tokens directly
- Removes the `tab-bar.ionic.vars.scss` file and moves the variables to the default ionic theme file since this uses global css variables

- [ ] Yes
- [x] No


BenOsodrac (approved):
💙

core/src/themes/ionic/ionic.theme.default.scss
```diff
@@ -89,3 +90,13 @@ $ionic-colors: (
...
+
+// Ionic Tabs & Tab Bar
+// --------------------------------------------------
+
+$tabbar-ionic-background: var(--ion-tab-bar-background, globals.$ion-primitives-base-white);
+$tabbar-ionic-background-activated: var(--ion-tab-bar-background-activated, globals.$ion-primitives-neutral-100);
+$tabbar-ionic-background-focused: var(--ion-tab-bar-background-focused, transparent);
+$tabbar-ionic-color: var(--ion-tab-bar-color, globals.$ion-primitives-neutral-800);
+$tabbar-ionic-color-selected: var(--ion-tab-bar-color-selected, globals.$ion-semantics-primary-base);
+$tabbar-ionic-border-color: var(--ion-tab-bar-border-color, transparent);
```
tanner-reits, line 102: Why are we creating special variables for these instead of just referencing the existing variables like we do for everything else?

brandyscarney: These are created separately because they use the global application color variables defined here: https://ionicframework.com/docs/theming/themes#application-colors

This matches how `ios` and `md` define them: https://github.com/ionic-team/ionic-framework/blob/ca091e277375815b9a24700bc782bd0fd9a20d1c/core/src/themes/native/native.theme.default.md.scss#L15-L22

Ideally we would want to add more of these to the ionic theme to match the available colors offered for toolbar, item, etc.

tanner-reits: Gotcha. Thanks for clarifying!

tanner-reits (approved):

PR #29979, opened 2024-10-30
feat(range): adding custom styling for ionic theme

OS-pedrolourenco:
Issue number: internal

---------

## What is the current behavior?
No ionic theme custom implementation existed for the ion-range component.

## What is the new behavior?

- Added ionic theme specific styling (based on the ios styling) for the range according to UX requirements.

## Does this introduce a breaking change?

- [ ] Yes
- [x] No

## Other information

image

PR #29988, opened 2024-10-31
feat(range):  updating the range pin to always be visible when enabled for ionic theme

OS-pedrolourenco:
Issue number: internal

---------

The ion-range pin is only shown we the range is being dragged for all themes.

## What is the new behavior?

- This PR updates the range pin so that it is alway visible when it is enabled for the ionic theme.

- [ ] Yes
- [x] No


core/src/components/range/range.ionic.scss
brandyscarney, line 1: Line 66 should be:

```
@include padding(calc(globals.$ion-space-100 + globals.$ion-scale-600), null, null, null);
```

This will make it so the padding top is the height of the text + the space between the knob and text

core/src/components/range/range.ionic.scss
```diff
@@ -107,12 +107,11 @@
 }
 
 .range-pin {
-  @include transform(translate3d(0, 100%, 0), scale(0.01));
-  @include padding(globals.$ion-space-200, globals.$ion-space-200, globals.$ion-space-200, globals.$ion-space-200);
+  @include padding(globals.$ion-space-300, globals.$ion-space-300, globals.$ion-space-300, globals.$ion-space-300);
```
brandyscarney, line 110:
```suggestion
  @include padding(null, null, globals.$ion-space-100, null);
```

We only need the bottom padding to separate the knob from the text

core/src/components/range/range.ionic.scss
```diff
@@ -107,12 +107,11 @@
...
 
   min-width: 28px;
 
-  transition: transform 120ms ease;
+  transform: translate3d(0, calc(-100% + 11px), 0) scale(1);
```
brandyscarney, line 114:
```suggestion
  transform: translate3d(0, calc(-100%), 0);
```

We can transform by the pin height by itself since the padding on it handles the space between the knob and text

brandyscarney (commented):
All of my requested changes are dependent on the `--height` variable being set to `40px` as requested on https://github.com/ionic-team/ionic-framework/pull/29979. This will make the total range height equal to `68px`, matching Figma.

core/src/components/range/range.tsx
```diff
@@ -928,7 +946,7 @@ const renderKnob = (
     >
       {pin && (
         <div class="range-pin" role="presentation" part="pin">
-          {pinFormatter(value)}
+          {pinFormatter(value) + (theme === 'ionic' ? '%' : '')}
```
tanner-reits, line 949: What do you think about changing the `pinFormatter` prop to return `number | string`. That way, we don't have to add this theme specific logic, and the OS widget could provide its own formatter that would append the `%`

tanner-reits (changes requested):
One suggestion to eliminate changes in the component

core/src/components/range/range.tsx
OS-pedrolourenco: Sounds great! I will do that

tanner-reits:
LGTM!

tanner-reits (approved):

PR #30018, opened 2024-11-14
feat(select): add sizes for ionic theme

thetaPC:
Issue number: internal

---------

The `ionic` theme does not have the styling for sizes.

## What is the new behavior?

- Added the size prop.
- Defaults medium for `ionic`, undefined for the others.
- Added tests.

- [ ] Yes
- [x] No

## Other information

[Preview](https://ionic-framework-git-rou-11318-ionic1.vercel.app/src/components/select/test/size?ionic:theme=ionic)


core/src/components/select/select.tsx
```diff
@@ -204,6 +204,15 @@ export class Select implements ComponentInterface {
    */
   @Prop({ mutable: true }) value?: any | null;
 
+  /**
+   * Set to `"small"` for a select with less height,
+   * `"medium"` for the default select height,
+   * or `"large"` for a select with more height.
+   *
+   * Defaults to `"medium"` for the ionic theme, and  undefined for all other themes.
+   */
```
brandyscarney, line 213:
```suggestion
  /**
   * The size of the select. If "large" it will increase the height of the select, while
   * "small" and "medium" provide progressively smaller heights. The default size is "medium".
   *
   * Defaults to `"medium"` for the ionic theme, and  undefined for all other themes.
   */
```

This matches the documentation in textarea more closely: https://github.com/ionic-team/ionic-framework/blob/bc3d30c3d28e617d76e6333f5593175d54060aad/core/src/components/textarea/textarea.tsx#L258-L262

brandyscarney (approved):
Looks good! Non-blocking suggestion to update the size description.

core/src/components/select/select.tsx
brandyscarney: Good catch!

Issue #30019, opened 2024-11-14
bug: open keyboard is not dismissed when card modal is dragged


hedinasr:
### Prerequisites

- [X] I have read the [Contributing Guidelines](https://github.com/ionic-team/ionic-framework/blob/main/docs/CONTRIBUTING.md#creating-an-issue).
- [X] I agree to follow the [Code of Conduct](https://ionicframework.com/code-of-conduct).
- [X] I have searched for [existing issues](https://github.com/ionic-team/ionic-framework/issues) that already report this problem, without success.

### Ionic Framework Version

v8.x

### Current Behavior

When keyboard is open in a card modal, dragging the modal down does not close the keyboard, causing a weird UI behavior.

https://github.com/user-attachments/assets/62f110ab-bd3e-4f08-a168-37ff8bb3a667

### Expected Behavior

When keyboard is open in a card modal, dragging the modal down should close the keyboard.

### Steps to Reproduce

1. Open this stackblitz https://stackblitz.com/~/github.com/hedinasr/ion-card-modal-keyboard-bug with Safari (iOS)
2. Open the modal
3. Focus on the input to open the keyboard
4. Drag the modal down

### Code Reproduction URL

https://github.com/hedinasr/ion-card-modal-keyboard-bug

### Ionic Info

Ionic:

   Ionic CLI       : 7.2.0 (/usr/local/lib/node_modules/@ionic/cli)
   Ionic Framework : @ionic/react 8.3.1

Capacitor:

   Capacitor CLI      : 6.1.2
   @capacitor/android : 6.1.2
   @capacitor/core    : 6.1.2
   @capacitor/ios     : 6.1.2

Utility:

   cordova-res : not installed globally
   native-run  : 2.0.1

System:

   NodeJS : v22.9.0 (/usr/local/Cellar/node/22.9.0_1/bin/node)
   npm    : 10.8.3
   OS     : macOS Unknown

aeharding:
I believe this is a duplicate of https://github.com/ionic-team/ionic-framework/issues/23955#issuecomment-1821377765 - an `onDragStart` event for modal would allow developer to customize the behavior.

aeharding:
This issue can probably be closed now that https://github.com/ionic-team/ionic-framework/issues/23955#issuecomment-1821377765 has been solved, the developer can now configure dismissing the keyboard on the `onIonDrag` event!

ShaneK:
[...]

Issue #23955, opened 2021-09-20
feat: dig into ionic animations / gestures


EinfachHans:
### Prequisites

- [X] I have read the [Contributing Guidelines](https://github.com/ionic-team/ionic-framework/blob/main/.github/CONTRIBUTING.md#creating-an-issue).
- [X] I agree to follow the [Code of Conduct](https://ionicframework.com/code-of-conduct).
- [X] I have searched for [existing issues](https://github.com/ionic-team/ionic-framework/issues) that already include this feature request, without success.

### Describe the Feature Request

If a FR like this already exist, please close, but i didn't found one.

It would be great to be able to "dig" into Animations and Gestures that are used in the Ionic Framework.

### Describe the Use Case

It would allow us to show, hide or transform & animate stuff in the app while the user performs something. For example with the new sheet style modal we could fade content in or out if the modal is dragged up or dragged down.

Also this can be used for navigation animations.

### Describe Preferred Solution

I guess this is kinda tricky, as this can be either defined globally or per component. 

For example if i want to implement this for a Modal it must be possible for me to add this in the Component that the Modal uses like 

```
this.modal.animation.subscribe()
```

### Describe Alternatives

Currently i think something like this is possible if we copy/paste all the animations/gestures from ionic into our app and modify to call a handler, but this is not very good, as it doesn't update the animations if you update it internally 🤔

liamdebeasi:
Can you please clarify what it means to "dig into an animation" in Ionic? It sounds like you are looking to get drag events emitted from the modal with the sheet gesture?

EinfachHans:
Yes exactly. But not only the *Drag* Gesture and not only the Sheet Modal. If for example (when possible in the future) the sheet modal is changed via a function it should also executy my listener to perform my animations. 

Or if we talk about Page Transistions this should work on the first navigate, as well on the navigate back or swipe to go back.

liamdebeasi:
Some of these animations/transitions are not meant to be extended/customized, so I would prefer not to add this feature to every animation unless we have use cases for each one.

The sheet modal use case makes sense, so I think we can add that one.

EinfachHans:
Yeah maybe every animation was too much 😃 But i wouldn't limit this on sheet modals, we could start with modals in generall?

liamdebeasi:
How would regular modals benefit from the drag event?

EinfachHans:
[...]

PR #30020, opened 2024-11-14
fix(modal): dismiss keyboard on drag card modal

hedinasr:
Issue number: resolves https://github.com/ionic-team/ionic-framework/issues/30019

---------

Dragging a card modal on iOS while the keyboard is open does not close the keyboard.

## What is the new behavior?
Dragging a card modal closes the keyboard.

- [ ] Yes
- [x] No

## Other information

|Before|After|
|------|------|
| <video src="https://github.com/user-attachments/assets/7f6071dc-dfd2-4f38-97fa-a757b971e1d3"/>|<video src="https://github.com/user-attachments/assets/2f63a6c2-8268-40cd-8cde-fa621a8f030b" />|

### DIFF
--- a/core/src/components/modal/gestures/swipe-to-close.ts
+++ b/core/src/components/modal/gestures/swipe-to-close.ts
@@ -1,7 +1,7 @@
 import { getTimeGivenProgression } from '@utils/animation/cubic-bezier';
 import { isIonContent, findClosestIonContent, disableContentScrollY, resetContentScrollY } from '@utils/content';
 import { createGesture } from '@utils/gesture';
-import { clamp, getElementRoot } from '@utils/helpers';
+import { clamp, getElementRoot, raf } from '@utils/helpers';
 import { OVERLAY_GESTURE_PRIORITY } from '@utils/overlays';
 
 import type { Animation } from '../../../interface';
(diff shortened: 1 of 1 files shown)


aeharding:
This is a breaking change for my use case, where users desire the keyboard to remain present.

A better solution would be adding onDragStart event to modal so developers can customize behavior - see: https://github.com/ionic-team/ionic-framework/issues/23955#issuecomment-1821377765

hedinasr:
@aeharding actually, this is exactly the same behavior as the sheet modal; which closes the keyboard on drag. Can you show me an example of your use case?

aeharding:
> @aeharding actually, this is exactly the same behavior as the sheet modal; which closes the keyboard on drag. Can you show me an example of your use case?

For example, iOS mail, if you pull down an email slightly the keyboard stays visible.

hedinasr:
@aeharding do you think that we can add some prop to conditionally activate this feature? Like having a `dimissKeyboard` boolean prop? I can implement the feature here.

ShaneK:
Hey! Thanks for the PR! I think I do agree with @aeharding here that this may not be desired behavior. As he suggested on the linked issue (and here), I think we'd prefer an event be emitted on drag start/end so developers can rig up their own event handling to do what they want - in your case, dismiss the keyboard.

I wouldn't even be opposed to this just being an optional property to enable keyboard dismiss on drag, but I still think an event being emitted on drag events would be a much nicer way to go

aeharding:
https://github.com/ionic-team/ionic-framework/issues/23955#issuecomment-1821377765 was solved so this MR can probably be closed! :tada:

ShaneK:
Thanks for bringing this back up @aeharding! This should be possible now in a way that should work for everyone with the drag events in 8.8.0, so I'm going to close this now

PR #30025, opened 2024-11-19
fix(overlays): announce info after opening based on platform

thetaPC:
Issue number: internal

---------

PR https://github.com/ionic-team/ionic-framework/pull/29951 would hide the overlays from screen readers while animating. This allows the element to navigate to its correct destination for screen readers to interact with. Otherwise, the focus rings would never appear. However, this ended up breaking the interaction for iOS.

## What is the new behavior?

- Overlays are hidden from screen readers while animating only if the platform is `android`. Since the original issue only applied to Android devices.

- [ ] Yes
- [x] No

## Other information

Dev build: `8.4.1-dev.11732033492.170e160c`

Test on iOS and Android devices.

### DIFF
--- a/core/src/utils/overlays.ts
+++ b/core/src/utils/overlays.ts
@@ -32,6 +32,7 @@ import {
   removeEventListener,
 } from './helpers';
 import { printIonWarning } from './logging';
+import { isPlatform } from './platform';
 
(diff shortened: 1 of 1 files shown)


brandyscarney (approved):
LGTM!

PR #30032, opened 2024-11-21
fix(accordion): move padding to inner element to use proper animation

brandyscarney:
Issue number: internal

---------

The animation is off for the `ionic` theme when using an accordion with text content.

## What is the new behavior?
- Moves the padding to the inner content element so the `offsetHeight` will be calculated properly and the animation will be smooth

- [ ] Yes
- [x] No

## Other information

[Preview](https://ionic-framework-git-rou-11362-ionic1.vercel.app/src/components/accordion/test/shape?ionic:theme=ionic)

### DIFF
--- a/core/src/components/accordion/accordion.ionic.scss
+++ b/core/src/components/accordion/accordion.ionic.scss
@@ -58,7 +58,7 @@
 // Accordion Content
 // --------------------------------------------------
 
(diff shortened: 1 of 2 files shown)


thetaPC (approved):
LGTM

PR #30035, opened 2024-11-22
fix(menu): hide from screen readers while animating

thetaPC:
Issue number: internal

---------

When the menu is presented on an Android device, TalkBack's focus rings may appear in the wrong position due to the transition (specifically `transform` styles). This occurs because the focus rings are initially displayed at the starting position of the elements before the transition begins.

## What is the new behavior?

- When an overlay is being animated (presenting or dismissing), the overlay will hide from screen readers. This allows Talkback to display the focus rings on the correct position.

- [ ] Yes
- [x] No

## Other information

Dev build: `8.4.1-dev.11732305980.19d90e1c`

Related to https://github.com/ionic-team/ionic-framework/pull/29951

### DIFF
--- a/core/src/components.d.ts
+++ b/core/src/components.d.ts
@@ -2303,6 +2303,7 @@ export namespace Components {
           * The name of the control, which is submitted with the form data.
          */
         "name": string;
+        "setFocus": () => Promise<void>;
         /**
           * the value of the radio group.
          */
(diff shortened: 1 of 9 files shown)


thetaPC:
Closed since the branch name is wrong. Opened the [correct one](https://github.com/ionic-team/ionic-framework/pull/30036).

PR #30041, opened 2024-11-25
test(angular): add ng19 test app

brandyscarney:
Issue number: internal

---------

There are tests apps for Angular 16, 17 and 18

## What is the new behavior?
Adds a test app for Angular 19

- [ ] Yes
- [x] No


packages/angular/test/apps/ng17/e2e/src/lazy/bind-component-inputs.spec.ts
brandyscarney, line 1: These files were moved to `base/` as they are all the same between versions

thetaPC (approved):
LGTM

PR #30053, opened 2024-11-29
fix(input): change focus token used on ionic theme

BenOsodrac:
Issue number: internal

---------

- Changed input focus token.
- Replaced some hard values with equivalent token.
- Updated snapshots.

- [ ] Yes
- [ ] No


brandyscarney (approved):

thetaPC (approved):
LGTM

PR #29950, opened 2024-10-18
feat(tokens): update tokens architecture and usage

BenOsodrac:
Issue number: internal

---------

## What is the new behavior?

- Removed old tokens stored on foundations folder. Now the jsons with the tokens come from the new dependency: https://github.com/OutSystems/outsystems-design-tokens.
- The tokens from UX Team completely changed the structure and in some cases the values. Everything was replaced on the project to use the new tokens.
- Snapshots updated and differences reviewed with UX Team. A fix was done on the testing css, to make sure the correct _Inter_ font-family was actually used. This resulted in some differences on the text for some componentes, that had snapshots with the native fonts.
- Removed generation of scss file on tokens script, that contained the :root selector with the custom CSS properties, as they were not used on the Ionic context.
- Removed generation of html file with tokens preview, as there wasn't a great value on this (we have storybook on other contexts) and it allowed to reduce a lot of code and complexity from the tokens script.
- The token command was adapted to use the command available from the https://github.com/OutSystems/outsystems-design-tokens package, using the `--config` prop, where we pass the path to our Ionic token script, where we generate the needed scss variables and utility-classes (by default the tokens repo tries to be as agnostic as possible, and only generates the css variables, without the prefix and added details we need on the Ionic side).
- Removed the token command from the npm run build, as it unnecessarily added time on that command to run. Besides, it should not be common that we need to run this command in the future.
- Updated reference to latest version of [Style Dictionary 4.1.3](https://v4.styledictionary.com/version-4/statement/). Version 4 comes with a lot of improvements, but also breaking-changes, so it was needed to adapt the tokens script. The code on the script is also now cleaner.
- Changed prefix from ionic to ion. The term ionic was used initially, to help differentiate from the old ios/md stuff, but I feel with the current scss architecture, that is no longer needed, and we can use the same prefix across themes.

This requires to run `npm install ` again.

## Does this introduce a breaking change?

- [x] Yes
- [ ] No

PR #30057, opened 2024-12-02
feat(tab-button): add the shape property and styles for the ionic theme

OS-pedrolourenco:
Issue number: internal

---------

Tab button does not have a shape property.

## What is the new behavior?
- Adds support for the shape property in tab button.
- Adds styles for the "soft", "round" and "rectangular" shapes in the ionic theme
- Defaults the shape to "round" for the ionic theme
- Adds an e2e test for shape with screenshots of all shapes

- [ ] Yes
- [x] No


core/src/components/tab-button/test/shape/index.html
```diff
@@ -0,0 +1,61 @@
+<!DOCTYPE html>
+<html lang="en" dir="ltr">
+  <head>
+    <meta charset="UTF-8" />
+    <title>Tab Button - Shape</title>
+    <meta name="viewport" content="width=device-width, initial-scale=1.0, minimum-scale=1.0" />
+    <link href="../../../../../css/core.css" rel="stylesheet" />
```
thetaPC, line 7:
```suggestion
    <link rel="stylesheet" href="../../../../../css/ionic.bundle.css" />
```

core/src/components/tab-button/test/shape/index.html
```diff
@@ -0,0 +1,61 @@
...
+          <ion-text style="flex: 1">
+            Soft Shape
+          </ion-text>
+          <ion-text style="flex: 1">
+            Round Shape
+          </ion-text>
+          <ion-text style="flex: 1">
+            Rectangular Shape
+          </ion-text>
+        </ion-tab-bar>
```
thetaPC, line 39: Remove this

core/src/components/tab-button/test/shape/index.html
```diff
@@ -0,0 +1,61 @@
...
+
+        <ion-tab-bar>
+          <ion-tab-button id="default-tab-button" style="background: #ff9e99">
+            <ion-icon aria-hidden="true" name="stop-circle"></ion-icon>
+          </ion-tab-button>
```
thetaPC, line 44:
```suggestion
          <ion-tab-button id="default-tab-button" style="background: #ff9e99">
            <ion-icon aria-hidden="true" name="stop-circle"></ion-icon>
            <ion-label>Default</ion-label>
          </ion-tab-button>
```

There's no need to add a second tab bar just to show which shape it is. You can easily do this with one tab bar and labels.

core/src/components/tab-button/test/shape/index.html
```diff
@@ -0,0 +1,61 @@
...
+            <ion-icon aria-hidden="true" name="cloud"></ion-icon>
+          </ion-tab-button>
+
+          <ion-tab-button id="round-tab-button" shape="round" style="background: #ff9e99">
+            <ion-icon aria-hidden="true" name="stop-circle"></ion-icon>
+          </ion-tab-button>
+
+          <ion-tab-button id="rect-tab-button" shape="rectangular" style="background: #a3ff85">
+            <ion-icon aria-hidden="true" name="square"></ion-icon>
+          </ion-tab-button>
```
thetaPC, line 56: Add the labels to each button with the label being the name of shape as shown in my previous comment.

core/src/components/tab-button/test/shape/index.html
```diff
@@ -0,0 +1,61 @@
...
+          <ion-text style="flex: 1">
+            Round Shape
+          </ion-text>
+          <ion-text style="flex: 1">
+            Rectangular Shape
+          </ion-text>
+        </ion-tab-bar>
+
+        <ion-tab-bar>
+          <ion-tab-button id="default-tab-button" style="background: #ff9e99">
```
thetaPC, line 42:
```suggestion
          <ion-tab-button id="default-tab-button">
```

Remove all the styles to the tab button. Use `--background: #f6f6f6;` on tab bar to accomplish the same scenario:

```html
ion-tab-bar {
   --background: #f6f6f6;
}
```

core/src/components/tab-button/test/shape/index.html
```diff
@@ -0,0 +1,61 @@
...
+            <ion-icon aria-hidden="true" name="stop-circle"></ion-icon>
```
thetaPC, line 43: Use triangle icons for all the tab buttons. This matches Figma designs and it makes it easier to debug when we stick as close to Figma as possible.

```suggestion
            <ion-icon aria-hidden="true" name="triangle-outline"></ion-icon>
```

core/src/components/tab-button/test/shape/index.html
```diff
@@ -0,0 +1,61 @@
...
+
+  <body>
+    <ion-app>
+      <ion-header>
+        <ion-toolbar>
+          <ion-title>Tab Button - Shape</ion-title>
+        </ion-toolbar>
+      </ion-header>
+
+      <ion-content id="content" style="text-align: center">
```
thetaPC, line 22: Since we want to render a simple and pure version of tab button, remove the style.
```suggestion
      <ion-content id="content">
```

thetaPC, line 22: Is there a reason that an `id` attribute was added? If not, then remove.

core/src/components/tab-button/test/shape/tab-button.e2e.ts
```diff
@@ -0,0 +1,55 @@
+import { expect } from '@playwright/test';
+import { configs, test } from '@utils/test/playwright';
+import type { E2EPage, E2EPageOptions, ScreenshotFn } from '@utils/test/playwright';
+
+class TabButtonFixture {
```
thetaPC, line 5: Why was a fixture created?

core/src/components/tab-button/test/shape/tab-button.e2e.ts
```diff
@@ -0,0 +1,55 @@
...
+    let tabButtonFixture: TabButtonFixture;
+
+    test.beforeEach(async ({ page }) => {
+      tabButtonFixture = new TabButtonFixture(page);
+      await tabButtonFixture.goto(config);
+    });
+
+    test('should render the default tab button', async () => {
+      await tabButtonFixture.screenshot('shape-round', screenshot, '#default-tab-button');
+    });
```
thetaPC, line 41: Remove, defaults are needed to be tested since those are already tested in a default tab bar test.

thetaPC (changes requested):

core/src/components/tab-button/test/shape/index.html
OS-pedrolourenco: The whole tab-bar? In my opinion, this caption helps anyone who visits this page on the browser understand better which understand what is being showcased.

core/src/components/tab-button/test/shape/index.html
OS-pedrolourenco: My goal was to set a unique background color and icon to distinguish each button and its shape. Setting a background color to the tab-bar will not make the button shapes visible.

core/src/components/tab-button/test/shape/index.html
OS-pedrolourenco: After seeing your comment below and considering that the default shape will not be tested on the same file anymore, I will follow your suggestion and use labels instead of this caption.

core/src/components/tab-button/test/shape/tab-button.e2e.ts
OS-pedrolourenco: This file was created based on toast/test/shape/toast.e2e.ts which also has a fixture and it made sense for me to have one here in order to access the same page and elements on every test.

core/src/components/tab-button/tab-button.tsx
```diff
@@ -107,6 +116,22 @@ export class TabButton implements ComponentInterface, AnchorInterface {
     }
   }
 
+  private getShape(): string | undefined {
+    const theme = getIonTheme(this);
+    const { shape } = this;
+
+    // TODO(ROU-11300): Remove theme check when shapes are defined for all themes.
```
thetaPC, line 123: When copying over code, let's verify what exactly is being copied. This ticket is not associated with tab-button. Make a Jira ticket for tab-button.

core/src/components/tab-button/test/shape/index.html
```diff
@@ -0,0 +1,55 @@
...
+        <ion-toolbar>
+          <ion-title>Tab Button - Shape</ion-title>
+        </ion-toolbar>
+      </ion-header>
+
+      <ion-content>
+        <ion-text> (Only available for ionic theme) </ion-text>
+
+        <ion-tab-bar>
+          <ion-tab-button id="default-tab-button">
```
thetaPC, line 26: Since default is not being tested, remove the id.
```suggestion
          <ion-tab-button>
```

core/src/components/tab-button/test/shape/tab-button.e2e.ts
```diff
@@ -0,0 +1,51 @@
...
+    const tabButton = page.locator(buttonId);
+
+    await expect(tabButton).toHaveScreenshot(screenshotString);
+  }
+}
+
+/**
+ * This behavior does not vary across directions.
+ */
+configs({ modes: ['ionic-md'], directions: ['ltr'] }).forEach(({ config, screenshot, title }) => {
```
thetaPC, line 30: All the screenshots are wrong. They do not match the HTML test page. Please fix.

thetaPC (changes requested):
Please make sure that tests are passing before requesting reviews.

core/src/components/tab-button/test/shape/tab-button.e2e.ts
OS-pedrolourenco: I have since fixed those screenshots. However, the basic tab-bar tests were also failing. On the one hand, it makes sense since the default button corners have changed. On the other hand, I do not understand how the border color could have changed:
image
Perhaps it's related to design tokens?

OS-pedrolourenco: I also just noticed that the snapshots for the "darwin" OS for the tab-button are also wrong but somehow the tests pass locally 😕

thetaPC: `darwin` snapshots are from tests that were run on your local computer. Those snapshots may vary based on the computer, for example snapshots on a Windows device will most likely not be the same as a Mac device. So they're not considered the true source and are not allowed to be pushed to the repro.

In order to make sure [snapshots are always being generated correctly](https://github.com/ionic-team/ionic-framework/blob/000f55303e459c583e642337fb1894f419f37d48/docs/core/testing/usage-instructions.md#running-tests-inside-docker) to the same environment as the CI, we run the tests within Docker: `npm run test.e2e.docker`. You'll know that the snapshots are correct when they have the `linux` keyword within the file.

Make sure you are using that command to run the tests. You can also use `npm run test.e2e.docker.update-snapshots` to [update the true snapshots](https://github.com/ionic-team/ionic-framework/blob/000f55303e459c583e642337fb1894f419f37d48/docs/core/testing/usage-instructions.md#generating-or-updating-ground-truths-with-docker-local-development) locally.

In order to fix this PR, I suggest running:
1. `npm run test.e2e.docker tab-bar tab-button`. This command will run all the tests that have the keywords tab-bar or tab-bar.
2. Review the failing tests: `npx playwright show-report`
3. It's extremely important that you understand why the tests are failing. They should only be failing for `ionic` theme and anything related to the tab button radius.
4. If you feel confident, then update the snapshots: `npm run test.e2e.docker.update-snapshots tab-button tab-bar`. You might not need to update both components so adjust the command as needed.

thetaPC (changes requested):
Snapshots are still wrong. Tests are still failing.

core/src/components/tab-button/test/shape/tab-button.e2e.ts
OS-pedrolourenco: I see! That makes a more sense! Thanks for the context!
I noticed yesterday or the day before that I was running snapshot tests incorrectly and started using Docker like you mentioned. What I wasn't aware was that those "darwin" files were leftovers from running the tests incorrectly. I will remove them then.

By the way, did you get a chance to look at my first comment regarding the unexpected change in border color for the tab-button?

thetaPC: I did, that's why I said to make sure you review the failing tests. I'm not seeing a change of color. The tests are failing for another reason.

OS-pedrolourenco: As I said before, the remaining tab-bar tests are failing due to the default border corners now being different. But, besides that, I also find it odd that the border color for the tab-button in the image I sent [here](https://github.com/ionic-team/ionic-framework/pull/30057#discussion_r1869223573) has also changed to a different shade of blue. As such, I am raising a flag before updating the snapshots that might be incorrect.

OS-pedrolourenco: After looking into a bit mor. I can indeed confirm that the color change was caused by the change in design tokens [here](https://github.com/ionic-team/ionic-framework/pull/29950. Thus, I conclude that snapshot tests do not check for color changes since these tests did not break with Bernardo's PR. I will then proceed with updating the snapshots

thetaPC (approved):
LGTM

PR #30058, opened 2024-12-03
fix(list-header): match background with designs

thetaPC:
Issue number: internal

---------

The background of the list header is white because it's within an `ion-list`, which has a white background.

## What is the new behavior?

- The background of the list header is transparent, matching design.
- Updated snapshots.

- [ ] Yes
- [x] No

## Other information

[Preview](https://ionic-framework-git-rou-11243-ionic1.vercel.app/src/components/list/test/shape?ionic:theme=ionic)


core/src/components/item-options/item-options.ionic.scss
```diff
@@ -7,6 +7,8 @@
 .item-options-ionic {
   align-items: center;
 
+  background: #{globals.$ion-primitives-base-white};
```
thetaPC, line 10: Necessary since the same style was removed from `ion-list`.

tanner-reits (approved):
LGTM

PR #30067, opened 2024-12-09
fix(tokens): fixed border generated utilityclasses

BenOsodrac:
Issue number: internal

---------

## What is the new behavior?

- Now the different border tokens will generate the utility-classes with the expected CSS attribute for each one!

- [ ] Yes
- [x] No


thetaPC (approved):
LGTM

PR #30070, opened 2024-12-10
fix(range): handle unsupported values for range min, max, and step

### LINKED ISSUE #29667: bug: ion-range throws: Cannot read properties of undefined.
### Prerequisites

### Ionic Framework Version

v8.x

### Current Behavior

Since the latest Ionic version v8 (coming from Ionic 6), the ion-range throws errors.

```
TypeError: Cannot read properties of undefined (reading 'toString')
  at getDecimalPlaces (ion-range.entry.js:15:14)
  at ion-range.entry.js:50:57
  at Array.map (<anonymous>)
  at roundToMaxDecimalPlaces (ion-range.entry.js:50:46)
  at ratioToValue (ion-range.entry.js:666:12)
  at get valA (ion-range.entry.js:394:16)
  at Range.renderRangeSlider (ion-range.entry.js:479:22)
  at Range.render (ion-range.entry.js:615:300)
```

This error occurs at least on Angular. I think it's because following:

- Set value: Call `updateRatio`: min/max possible undefined
- Set min: Call `updateRatio`: max/value possible undefined
- Set max: Call `updateRatio`: min/value possible undefined

This results in runtime error and invisible range component. It's maybe because a framework like Angular creates the attibutes for binding even if the value is not yet ready. In this case, it's undefined first. But idk. Anyway, it would be a good practice to treat undefined, wouldn't it?

### Expected Behavior

Handle `undefined`. It's possible that one value can be undefined.

### Steps to Reproduce

```html
<ion-item *ngIf="hasRange()">
  <ion-range [(ngModel)]="value" [min]="min" [max]="max"></ion-range>
</ion-item>
```

All values are inputs (`Input` Angular decorator). The `hasRange` method returns true if both values (min, max) are available (not undefined) and not equal (min !== max).

In my case I created a modal with componentProps which goes to Angular component inputs.

NOTE: I already catch undefined values. It seems that this problem comes from Ionic framework.

### Code Reproduction URL

https://codepen.io/fudom/pen/ZENZxdq
But not really good reproducible in a playground. Just handle undefined values!

### Ionic Info

[Ionic Angular](https://www.npmjs.com/package/@ionic/angular) 8.2.3

### Additional Information

- [docs](https://ionicframework.com/docs/api/range)
- [range.tsx](https://github.com/ionic-team/ionic-framework/blob/52ff0505e86dc204cb2fd2fdaf67229e6eeb36f8/core/src/components/range/range.tsx)
- [utils/floating-point](https://github.com/ionic-team/ionic-framework/blob/52ff0505e86dc204cb2fd2fdaf67229e6eeb36f8/core/src/utils/floating-point/index.ts)

fudom:
Issue number: resolves #29667

---------

Currently, if min/max are set to undefined on `IonRange` (which is an accepted value), it breaks the DOM.

## What is the new behavior?
After these changes, if min/max are set to undefined or any unsupported value (such as infinity or a NaN), it will fall back to the default values for min and max (currently, 1 and 100 respectively).

- [ ] Yes
- [x] No


core/src/utils/type-guards.ts
Em-5455, line 1: Can SafeNumber be defined by a fixed value or a fixed range of values, 

For example, 
>0 and <0.00005
any input is number between 0 and 0.00005 to be defined and therefore handled as a "nullish value" and enable the rule to be applied  if (!isSafeNumber(n)) return 0;

core/src/utils/helpers.ts
```diff
@@ -424,3 +424,10 @@ export const getNextSiblingOfType = <T extends Element>(element: Element): T | n
   }
   return null;
 };
+
+/**
+ * Checks input for usable number. Not NaN and not Infinite.
+ */
+export const isSafeNumber = (input: unknown): input is number => {
```
thetaPC, line 431: Should this be `as` not `is`?

```suggestion
export const isSafeNumber = (input: unknown): input as number => {
```

ShaneK: No, this is a typescript type predicate for return types. It tells typescript that if the result is true, it can safely assume the input is a number from then on out. So if you do `if (isSafeNumber(abc)) abc++;` then typescript knows that's fine because you checked and told typescript it's safe to assume it's a number now.

thetaPC (approved):
LGTM

thetaPC:
@fudom Congrats on your first PR!

core/src/components/range/range.tsx
```diff
@@ -300,6 +314,11 @@ export class Range implements ComponentInterface {
     }
 
     this.inheritedAttributes = inheritAriaAttributes(this.el);
+    // If the min or max is not safe, set it to 0 or 100 respectively.
+    // Our watch does this, but not before the initial load.
```
brandyscarney, line 318:
```suggestion
    // If min, max, or step are not safe, set them to 0, 100, and 1, respectively.    
    // Each watch does this, but not before the initial load.
```

brandyscarney (approved):
Looks great 🙂

ShaneK:
Thank you for your PR! It was merged into main and will go out with the next release.

PR #30076, opened 2024-12-11
feat(select): add focus on tab

BenOsodrac:
Issue number: internal

---------

## What is the new behavior?

- Added focus styles on tab for the ion-select, using the --highlight variable.
- Adjusted highlight tests to work for ionic as well, just for the outline type.
- Fixed select border-color token used and updated snapshots (this affected all ionic theme snapshots for select). Although not directly related to focus task, the incorrect color was making it very hard to see the differences between the focus colors.

- [ ] Yes
- [x] No

## Other information

- [Select focus](https://ionic-framework-git-rou-11445-select-ionic1.vercel.app/src/components/select/test/highlight?ionic:theme=ionic)


brandyscarney (approved):
Should we add tests for this?

BenOsodrac:
> Should we add tests for this?

Yes, good point, just added them :)

core/src/components/select/select.ionic.scss
```diff
@@ -181,6 +182,18 @@
   color: globals.$ion-primitives-neutral-500;
 }
 
+// Focus
+// ---------------------------------------------
+:host(.ion-focused) .select-wrapper-inner {
```
thetaPC, line 187: Based on the designs, there shouldn't be an offset for any of the selects. But I'm seeing an offset for the outline variant.

core/src/components/select/select.ionic.scss
```diff
@@ -181,6 +182,18 @@
...
+  @include globals.focused-state(null, null, var(--focus-ring-color));
+}
+
+// Invalid
+// ---------------------------------------------
+:host(.ion-invalid) {
+  --focus-ring-color: #{globals.$ion-border-focus-error};
+}
```
thetaPC, line 195: This needs to follow the structure of `md` that uses `--highlight-color-invalid`:
- https://github.com/ionic-team/ionic-framework/blob/ef9a8ef550b771c5c8154f1db95343e85396d5e0/core/src/components/select/select.common.scss#L238-L240
- https://github.com/ionic-team/ionic-framework/blob/ef9a8ef550b771c5c8154f1db95343e85396d5e0/core/src/components/select/select.md.outline.scss#L25-L28
- https://github.com/ionic-team/ionic-framework/blob/ef9a8ef550b771c5c8154f1db95343e85396d5e0/core/src/components/select/select.md.solid.scss#L30-L33

core/src/components/select/select.ionic.scss
```diff
@@ -11,6 +11,7 @@
   // ios and md. This will allow us to have a single color map.
   --border-color: #{globals.current-color(neutral)};
   --border-width: #{globals.$ion-border-size-025};
+  --focus-ring-color: #{globals.$ion-border-focus-default};
```
thetaPC, line 14: There's already a focus variable: `--highlight-color-focused`. I recommend reviewing the [docs](https://ionicframework.com/docs/api/select#css-custom-properties).
- https://github.com/ionic-team/ionic-framework/blob/ef9a8ef550b771c5c8154f1db95343e85396d5e0/core/src/components/select/select.common.scss#L37

core/src/components/select/test/states/select.e2e.ts
thetaPC, line 1: This file does not need to be added since focus and invalid states are being tested within the [highlight test file](https://github.com/ionic-team/ionic-framework/blob/ef9a8ef550b771c5c8154f1db95343e85396d5e0/core/src/components/select/test/highlight/select.e2e.ts#L20-L45). Please review the test files to determine if a new one is needed.

Since this file is not needed, update the highlight tests to include the `ionic-md` and delete the snapshots that were already submitted to this PR.

core/src/components/select/select.ionic.scss
BenOsodrac: Thanks for the heads-up, will change it!

core/src/components/select/select.ionic.scss
BenOsodrac: Ah silly me, I saw that and my brain instantly thought about text-highlight for some reason :). Will make the needed changes!

core/src/components/select/select.ionic.scss
BenOsodrac: Done!

core/src/components/select/select.ionic.scss
BenOsodrac: Implemented the changes! Added only for the outline, as its the only type used on Ionic theme.

Also took the opportunity to fix the border color token, thats why we have so many snapshots updated!

core/src/components/select/select.ionic.scss
BenOsodrac: This is done ✅

core/src/components/select/select.ionic.outline.scss
```diff
@@ -50,3 +50,10 @@
 
   box-sizing: border-box;
 }
+
+// Focus
+// ---------------------------------------------
+
+:host(.ion-focused.select-fill-outline) {
+  --border-color: var(--highlight-color);
```
thetaPC, line 58:
```suggestion
  --border-color: var(--highlight-color);
  --border-width: #{globals.$ion-border-size-050};
```
When focused, the border width also changes.

core/src/components/select/select.ionic.scss
```diff
@@ -9,14 +9,18 @@
   --background: #{globals.$ion-primitives-base-white};
   // TODO(ROU-10778, ROU-10875): Sync the color names to the design system of
   // ios and md. This will allow us to have a single color map.
-  --border-color: #{globals.current-color(neutral)};
+  --border-color: #{globals.$ion-primitives-neutral-500};
```
thetaPC, line 12: Why was this value changed? The ticket is supposed to be for focus only. I would recommend making a separate PR.

core/src/components/select/select.ionic.scss
```diff
@@ -9,14 +9,18 @@
...
   --border-width: #{globals.$ion-border-size-025};
+  --focus-ring-color: #{globals.$ion-border-focus-default};
```
thetaPC, line 14:
```suggestion
```

This variable is no longer being used so it shouldn't be included.

core/src/components/select/select.ionic.scss
```diff
@@ -9,14 +9,18 @@
...
   --padding-start: #{globals.$ion-space-400};
   --padding-end: #{globals.$ion-space-400};
   --padding-top: #{globals.$ion-space-300};
   --padding-bottom: #{globals.$ion-space-300};
   --placeholder-color: #{globals.$ion-primitives-neutral-800};
   --placeholder-opacity: 1;
+  --highlight-color-focused: #{globals.$ion-border-focus-default};
+  --highlight-color-valid: #{globals.$ion-semantics-success-900};
+  --highlight-color-invalid: #{globals.$ion-semantics-danger-800};
```
thetaPC, line 23: Why was this value changed? The ticket is supposed to be for focus only. I would recommend making a separate PR.

core/src/components/select/test/highlight/select.e2e.ts
```diff
@@ -149,6 +149,53 @@ configs({ modes: ['md'], directions: ['ltr'] }).forEach(({ title, screenshot, co
...
+
+        const select = page.locator('ion-select');
+        await expect(select).toHaveScreenshot(screenshot(`select-outline-highlight`));
+      });
+    });
+  });
+});
+
+configs({ modes: ['md', 'ionic-md'], directions: ['ltr'] }).forEach(({ title, screenshot, config }) => {
+  test.describe(title('select: highlights'), () => {
```
thetaPC, line 198: As mentioned previously, we should open a new PR that addresses valid and invalid states. This PR should only be for focus.

thetaPC (changes requested):

core/src/components/select/select.ionic.scss
BenOsodrac: But it is related to focus as well. When the user focus on a valid or invalid select, the focus color should have the expected semantic color.

core/src/components/select/test/highlight/select.e2e.ts
BenOsodrac: But the md highlight implementation seems to take this under the same feature, using the highlight variable mechanism. Focus can be normal, valid or invalid.

core/src/components/select/select.ionic.scss
BenOsodrac: I explain my reasons on the PR description. As this was such a simple fix for a obvious bug and with impacts on focus visibility on snapshots, I thought we could include it here. Its all related related to changes on select border color, after all. 

Can we keep the changes? If you still feel it makes no sense, I can do it separately yes 👍

core/src/components/select/select.ionic.outline.scss
```diff
@@ -50,3 +50,11 @@
...
+  --border-width: #{globals.$ion-border-size-050};
+}
```
thetaPC, line 60: Apologies, I just noticed that on the designs, only the focus has a greater width, not the valid or invalid. So let's do something like:

```suggestion
:host(.ion-focused.select-fill-outline) {
  --border-color: var(--highlight-color);
}

:host(.ion-focused.select-fill-outline:not(.ion-invalid):not(.ion-valid)) {
  --border-width: #{globals.$ion-border-size-050};
}
```

core/src/components/select/select.ionic.scss
```diff
@@ -9,14 +9,17 @@
   --background: #{globals.$ion-primitives-base-white};
   // TODO(ROU-10778, ROU-10875): Sync the color names to the design system of
   // ios and md. This will allow us to have a single color map.
-  --border-color: #{globals.current-color(neutral)};
+  --border-color: #{globals.$ion-primitives-neutral-500};
```
thetaPC, line 12: Based on the designs, it seems that they're using neutral-800.

Screenshot 2024-12-16 at 10 44 37 AM

core/src/components/select/select.ionic.outline.scss
```diff
@@ -50,3 +50,11 @@
 
   box-sizing: border-box;
 }
+
+// Focus
```
thetaPC, line 54: I also noticed that the invalid example on the highlight test page wasn't rendering correctly. It should have a red border. This can be seen when comparing it to md.

In order to fix that, the following styles need to be added:
```scss
/**
 * If the select has a validity state, the
 * border should reflect that as a color.
 */
:host(.has-focus.select-fill-outline.ion-valid),
:host(.select-fill-outline.ion-touched.ion-invalid) {
  --border-color: var(--highlight-color);
}
```

thetaPC (changes requested):

core/src/components/select/select.ionic.scss
BenOsodrac: That's true, no idea where I got the 500 when inspecting the Figma file 🤔 .

Will change, thanks for the heads-up!

core/src/components/select/select.ionic.outline.scss
BenOsodrac: Thanks, no problem, I should have noticed that myself!

core/src/components/select/select.ionic.outline.scss
BenOsodrac: Hum, didn't notice as it seems to work well when receiving the actual focus

BenOsodrac: Fixed!

core/src/components/select/select.ionic.scss
thetaPC: Perfect! That makes sense to me. Thank you for the double check!!

thetaPC (approved):
LGTM

brandyscarney (approved):

PR #30087, opened 2024-12-16
fix(select):  ensure select sizes are respected when the label is empty

JoaoFerreira-FrontEnd:
Issue number: internal

---------

Currently, the experience planned for ionic select has the sizes fixed and should be respected even when the labels are not presented. 
https://stackblitz.com/edit/ku3fqz?file=src%2Fmain.tsx

## What is the new behavior?

- When select is used with Ionic-theme, the select should respect the height sizes established.
- https://stackblitz.com/edit/ku3fqz-2vnasdmx?file=package-lock.json,package.json
- https://ionic-framework-cdtf7hsqy-ionic1.vercel.app/src/components/select/test/size?ionic:theme=ionic

- [ ] Yes
- [x] No


core/src/components/select/select.ionic.scss
```diff
@@ -96,6 +96,10 @@
   width: 100%;
 }
 
+:host(.select-label-placement-stacked) .select-wrapper-inner{
+  flex-grow: 0;
+}
```
thetaPC, line 101: This approach does not align with the styling structure. `flex-grow` is being applied to all the themes because it exists within the common file. Since `ionic` theme does not require this style, then we need to move the following styles to the `select.native.scss` file: https://github.com/ionic-team/ionic-framework/blob/6e58dbd78321b8013bb637fb423b5fa10b4b3ffd/core/src/components/select/select.common.scss#L224-L233

Once that is moved then the styles that were added to the `ionic` theme can be removed. Plus, I would recommend moving the following styles to `select.native.scss` file as well:
https://github.com/ionic-team/ionic-framework/blob/6e58dbd78321b8013bb637fb423b5fa10b4b3ffd/core/src/components/select/select.common.scss#L60-L68

I would recommend reviewing how the file structure are set for the themes. That way we're not having to overwrite styles when we can move them to the appropriate locations.

thetaPC (changes requested):

thetaPC:
@JoaoFerreira-FrontEnd

Make sure to have the checks passing before requesting a review.

Please review how we write the PR descriptions. This one is missing information on the current behavior. It helps the reviewer understand the issue if there's text vs having to figure out the issue through a repro. It would have taken me some time to understand the issue if I hadn't already known about it prior. But it might have taken some time for someone else since there was no mention of sizes being affected by the lack of a label.

Please review how we write the PR titles. We don't end the titles with a period and there's no need to repeat the component name within the title since it's already mentioned in the scope. I would recommend something like: `fix(select): ensure select sizes are respected when the label is empty`

thetaPC (approved):
LGTM

core/src/components/select/test/fill/select.e2e.ts-snapshots/select-fill-outline-ionic-md-ltr-light-Mobile-Chrome-linux.png
thetaPC, line 1: These snapshots shouldn't be re-generating. Make sure that this branch has the latest from next.

JoaoFerreira-FrontEnd: It has

brandyscarney (approved):

PR #30097, opened 2024-12-22
feat(modal): add expandToScroll property to allow scrolling at all breakpoints

### LINKED ISSUE #24631: feat: sheet modal, add option to prefer scrolling when not fully expanded
### Prerequisites

### Ionic Framework Version

- [ ] v4.x
- [ ] v5.x
- [X] v6.x

### Current Behavior

Hi, in the Sheet Modal the scroll of the content doesn't appear unless it's at 100% size, in other breakpoints the scroll doesn't appear.

### Expected Behavior

The scroll must adjust to the content at all times, regardless of the breakpoint.

### Steps to Reproduce

```
<ion-modal [isOpen]="true" [breakpoints]="[0.1, 0.5, 1]" [initialBreakpoint]="0.5">
    <ng-template>
      <ion-content>
        <ion-list>
          <ion-item>
            <ion-label>Text</ion-label>
          </ion-item>
          <ion-item>
            <ion-label>Text</ion-label>
          </ion-item>
          <ion-item>
            <ion-label>Text</ion-label>
          </ion-item>
          ...
          <ion-item>
            <ion-label>Text</ion-label>
          </ion-item>
          <ion-item>
            <ion-label>Text</ion-label>
          </ion-item>
          <ion-item>
            <ion-label>Text</ion-label>
          </ion-item>
        </ion-list>
      </ion-content>
    </ng-template>
  </ion-modal>
```

### Code Reproduction URL

_No response_

### Ionic Info

Ionic:

Ionic CLI : 6.18.1

Utility:

native-run : 1.5.0

System:

NodeJS : v17.3.0
npm : 8.3.0
OS : Windows 10

### Additional Information

_No response_

kumibrr:
Issue number: resolves #24631

---------

Sheet modals do not allow content scrolling unless its breakpoint is 1.

## What is the new behavior?
Sheet modals now have an additional property `expandToScroll` which allows the sheet modal to scroll content at all breakpoints when it is set to `false`.

- [ ] Yes
- [x] No

## Other information

https://github.com/user-attachments/assets/97f51a89-096e-44d7-a8cf-66363133e223


davidalejandroaguilar:
I'm new to Ionic and I'm excited to see this closing a much needed 3 year old issue. Happy to be here just in time! 🥳 

Looking forward to implementing these awesome UX patterns https://sarunw.com/posts/bottom-sheet-in-ios-15-with-uisheetpresentationcontroller/

thetaPC:
@kumibrr Thank you for submitting this PR!

Could you explain why a new array, `snapBreakpoints`, is being created instead of just using the existing `breakpoints` array? If we continue using `breakpoints` then it would follow the setup on native iOS.

kumibrr:
> @kumibrr Thank you for submitting this PR!
> 
> Could you explain why a new array, `snapBreakpoints`, is being created instead of just using the existing `breakpoints` array? If we continue using `breakpoints` then it would follow the setup on native iOS.

Thank you for reviewing and asking!

There's two main reasons I made a new array, the first is to avoid breaking changes with any current setup.
The second one is that there's some exceptions in the native iOS setup, like on Maps, in which there's some breakpoints that do not allow content scroll despite having scrollable content.


https://github.com/user-attachments/assets/ccdb8ae4-c92f-4814-bd5f-15d5aa382618

thetaPC:
@kumibrr That makes sense to me. Could you try a video that showcases your proposed behavior? Or provide steps on how I can view it on my end? It seems that the one you sent in the last comment is not rendering.

kumibrr:
@thetaPC Sure, sorry about that.
Here's two videos, one is the current behavior in the Maps app, where I showcase the current behavior, and the proposed one, in which `breakpoints = [0,0.25, 0.5, 0.75]` and `snapBreakpoints = [0.5, 0.75]`.

If videos do not work, you can try the proposed behavior in the following link, then clicking on "Present Sheet Modal (snapBreakpoints)", it has `breakpoints = [0,0.25, 0.5, 0.75]` and `snapBreakpoints = [0.5, 0.75]` https://ionic-framework-git-fork-kumibrr-main-ionic1.vercel.app/src/components/modal/test/sheet




https://github.com/user-attachments/assets/a9bc9f45-16aa-4e4b-a323-41aabdf5bd1a

https://github.com/user-attachments/assets/aaad87e8-bbdd-4f26-bbe5-5ac8d704b2ea

kumibrr:
@thetaPC That's something I thought when making it, as you said, I could not find any native app that matches this.

I also had in mind `scrollThreshold` instead of `snapBreakpoints`, which would receive a single breakpoint, and would make any breakpoints above it scrollable.
ex. `scrollThreshold="0.5"`.
It matches native behavior much better, though not as flexible.

If that sounds good, should I make a new PR with the alternative `scrollThreshold` and close this one? Or should I push new commits to this one, and change the title?

thetaPC:
Understood! I'll have to talk to the team to decide on the approach. The goal is to stay as close to native as possible so that might lead to sticking with an all or nothing concept like iOS: either scrolling only happens on max height or scrolling happens in all heights.

I'll aim to get back to you within the next week since this PR is very promising!

thetaPC:
@kumibrr hope your weekend went well!

I've discussed it with the team and we want to stick with how native iOS does it, meaning that either the user can scroll when the sheet is fully expanded or at all the breakpoints. This can be done with a boolean. Here's an [example](https://github.com/ionic-team/ionic-framework/compare/ROU-5153?expand=1) of the direction that we want to go. We would like for this PR to use the boolean and also add/update comments (again examples can be shown via the link). We are looking forward to seeing those requested changes!

core/src/components/modal/gestures/sheet.ts
```diff
@@ -323,6 +346,13 @@ export const createSheetGesture = (
         },
       ]);
 
+      if (contentAnimation) {
+        contentAnimation.keyframes([
+          { offset: 0, maxHeight: `${(1 - breakpointOffset) * 100}%` },
+          { offset: 1, maxHeight: `${snapToBreakpoint * 100}%` },
+        ]);
```
thetaPC, line 353: Let's add a comment of what this block of code is doing. [Example](https://github.com/ionic-team/ionic-framework/blob/9794a63f85ca81103d3bbb1a166a81550ba442f4/core/src/components/modal/gestures/sheet.ts#L345-L350) can be found.

thetaPC (changes requested):
As mentioned in our previous conversation, let's use a boolean to determine when the scrolling will happen.

kumibrr:
@thetaPC Fine by me!
I've just submitted the requested changes, but the added footer to the content appears abruptly after the modal has opened.

kumibrr:
@thetaPC I've done some more changes to fix the footer animation on creation. Now the animation is smooth with no jittering.

I have also moved the contentAnimation initialization from `modal/gestures/sheet.ts` to `modal/animations/sheet.ts` where it belongs.

And I've added a new property `animateContentHeight` in `ModalAnimationOptions` that will enable the contentAnimation when `scrollAtEdge === false`

kumibrr:
Alright. I've spent quite some time but couldn't figure out how to fix it. This also happens in mobile Safari, mobile Chrome (using an Android device) but Chrome desktop masked it (it changes so fast I didn't caught it)

As seen in the attached gif, the problem is that the footer jumps when breakpoint changes. I have my guesses, but I don't want to bias any guesses.

ScreenRecording2025-01-14at20 10 42-ezgif com-video-to-gif-converter.

thetaPC:
@kumibrr I do see the flicker of the footer on my end. This has been one of the major issues that we encounter when trying to introduce the feature. What are your guesses of why it's happening?

kumibrr:
@thetaPC The `wrapperAnimation` moves the wrapper downwards, and the `contentAnimation` changes the height of the element. The wrapper animation is GPU accelerated, but the contentAnimation is not. In slower CPUs or using a browser that doesn't perform as well in layout shifts makes the flickering more noticeable.

I'll try reimplementing the animation with `transform: scaleY`. I'll reach back.

thetaPC:
@kumibrr were you able to make progress on this?

kumibrr:
Hello @thetaPC I had a busy week, but I'm making some progress!
Can see it here: https://github.com/kumibrr/ionic-framework/tree/scale-y

As I said, I'm changing the implementation to use `transform: scaleY` instead of `height` property, but came with another set of issues and challenges.

The biggest problems I'm currently facing are the following:
1. Whenever animation finishes there's a flicker the content of the modal.
This is caused by changing between the `moveSheetToBreakpoint` animation and the default animation. The content animation has a contentScale = 1 / containerScale, and this has been problematic as the ratio leads to Infinity in the highest values.
2. The footer problem persists.
3. The scrollbars do not behave correctly: they appear way after they're required
4. It's more complex than the height implementation
 
And a gif showing the current `transform: scaleY()` implementation
Screen recording

I've been also thinking in a workaround for the footer, manually querying it, and moving it outside of the ion-page with a `position: fixed` and adding it to the  `wrapperAnimation`. But I hate this workaround, that's the reason I keep trying with the scale implementation, and given the performance improvement over the height transform I think it's worth it.

I hope to spend more time on this over the weekend.

thetaPC:
@kumibrr Thanks for the update! When I checked out your branch, I noticed that the content top and bottom was being cut off when using Firefox. It might be best to stick with `maxHeight` since that provided the least amount of issues. The footer issue could be solved with your suggestion to manually query it. I've been working on that approach a bit today and it seems that it might be the right direction.

kumibrr:
Hello again @thetaPC!
Today I bring you some updates!
I've followed your suggestion to stick with `maxHeight` with great success.

As you might see in the example, I've added a footer with dynamic height.
At the moment it won't increase the height of it and the content gets under the footer.

I'll fix them right away, and I also wanted to ask you for any possible improvements.

thetaPC:
@kumibrr lol seems that we had the same idea on implementation.

When I was testing the new changes, I noticed that the modal can no longer close when dragging.

I was also able to get the padding to work by moving the logic to `md.enter.ts`:

```ts
const baseAnimation = createAnimation()
  .addElement(baseEl)
  .easing('cubic-bezier(0.36,0.66,0.04,1)')
  .duration(280)
  .addAnimation([backdropAnimation, wrapperAnimation])
  .beforeAddRead(() => {
    console.log('beforeAddRead');
    const ionFooter = baseEl.querySelector('ion-footer');
    if (ionFooter) {
      const footerHeight = ionFooter.clientHeight;
      // clone the footer element
      const clonedFooter = ionFooter.cloneNode(true) as HTMLElement;
      // move the cloned footer to the end of base el shadow root
      baseEl.shadowRoot!.appendChild(clonedFooter);
      // delete the original footer
      ionFooter.remove();
      // add padding bottom to the .ion-page element to be
      // the same as the cloned footer height
      const page = baseEl.querySelector('.ion-page') as HTMLElement;
      page.style.setProperty('padding-bottom', `${footerHeight}px`);
    }
  });
```

and on modal.scss

```scss
:host(.modal-sheet) ion-footer {
  position: absolute;
  bottom: 0;

  width: var(--width);
}
```

This allows the modal to close on drag. However, I'm still trying to figure out how to get the footer to slide out on dismiss.

kumibrr:
@thetaPC I like waay more the place you suggested to perform the footer DOM operations.

I made the changes you made, and removed the animations from the `gestures/sheet.ts` as they're no longer needed.
I've cleaned up a bit and I also made the animation work for both enter and exit.

> When I was testing the new changes, I noticed that the modal can no longer close when dragging.
Ah yeah, I forgot that I removed that, I had changed it while I was working on it.

> @kumibrr lol seems that we had the same idea on implementation.
> That's pretty cool 😁

thetaPC:
@kumibrr We're getting really close to finishing this. The last issue is that the footer should also slide out as the content slides out. Notice in this video that the footer is still showing even though the content is gone:

https://github.com/user-attachments/assets/3542a25d-fe70-408d-b944-55482858e8fa

It should be slide out like:


https://github.com/user-attachments/assets/a8be3336-496c-4801-87fc-5a170cb02db4

diegosystem32:
Has this request already been made? Do we already have this functionality?

thetaPC:
@kumibrr I was able to simplify the approach for the footer and fixed a few minor details that we missed. Here's my [PR](https://github.com/kumibrr/ionic-framework/pull/1) that addresses it. Feel free to merge it into your branch. If you prefer a copy and paste, I would appreciate co-author be given.

Thank you so much for everything! We're so close!

core/src/components/modal/gestures/sheet.ts
```diff
@@ -110,6 +116,33 @@ export const createSheetGesture = (
...
+   * Changes the footer that is currently visible
+   * @param footer - The footer to show
+   */
+  const swapFooterVisibility = (footer: 'original' | 'cloned') => {
+    const footerToHide = footer === 'original' ?
+      wrapperEl.nextElementSibling as HTMLIonFooterElement :
+      baseEl.querySelector('ion-footer') as HTMLIonFooterElement;
+    const footerToShow = footer === 'original' ?
+      baseEl.querySelector('ion-footer') as HTMLIonFooterElement :
+      wrapperEl.nextElementSibling as HTMLIonFooterElement;
```
thetaPC, line 131:
```suggestion
    const originalFooter = baseEl.querySelector('ion-footer') as HTMLIonFooterElement;
    const clonedFooter = wrapperEl.nextElementSibling as HTMLIonFooterElement;
    const footerToHide = footer === 'original' ? clonedFooter : originalFooter;
    const footerToShow = footer === 'original' ? originalFooter : clonedFooter;
```

core/src/components/modal/gestures/sheet.ts
```diff
@@ -110,6 +116,33 @@ export const createSheetGesture = (
...
+
+    footerToShow.style.removeProperty('display');
+    footerToShow.removeAttribute('aria-hidden');
+
+    const pagePadding =
+      footer === 'original' ? 0 : footerToShow.clientHeight;
```
thetaPC, line 137:
```suggestion
```
Let's use the `removeProperty` if the footer is original. See next suggestion.

core/src/components/modal/gestures/sheet.ts
```diff
@@ -110,6 +116,33 @@ export const createSheetGesture = (
...
+
+    const page = baseEl.querySelector('.ion-page') as HTMLElement;
+    page.style.setProperty('padding-bottom', `${pagePadding}px`);
```
thetaPC, line 140:
```suggestion
    if (footer === 'original') {
      page.style.removeProperty('padding-bottom');
    } else {
      const pagePadding = footerToShow.clientHeight;
      page.style.setProperty('padding-bottom', `${pagePadding}px`);
    }
```

Since we added the inline padding bottom, then we can safely remove it so it reverts to the original inline styles.

thetaPC (changes requested):
Minor changes, also make sure to run `npm run lint` locally and push the lint changes. Otherwise, the checks are going to fail and the PR won't be able to merge.

kumibrr:
@thetaPC Alright, done!

A really minor thing I just noticed is that the cloned footer has a different height for iOS styles.
Here's the video:

https://github.com/user-attachments/assets/be6717dd-05cf-4054-9b17-da046d36cde2

thetaPC:
@kumibrr Yes, they have different heights but they still work great. Tested both on my end. Really happy to hear that it's working on an older phone too!

Just getting the checks to pass and we can have another member of the team to the next round of reviews. 🎉

kumibrr:
@thetaPC Oh, I had pushed the fix to the height while the actions were running, sorry!
The footer's toolbar did had some styles applied by core.scss.

Anyway, thank you for all your help! 🫂

core/src/components/modal/modal.scss
```diff
@@ -166,3 +166,14 @@ ion-backdrop {
...
+:host(.modal-sheet.modal-scroll-all) ion-footer {
+  position: absolute;
+  bottom: 0;
+
+  width: var(--width);
+}
+
+:host(.modal-sheet.modal-scroll-all) ion-footer ion-toolbar:first-of-type {
+  padding-top: 6px;
+}
```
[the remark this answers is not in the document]
kumibrr: Sure, I've pasted the following snippet found in `core.scss`
```scss
/**
 * Card style modal needs additional padding on the
 * top of the header. We accomplish this by targeting
 * the first toolbar in the header.
 * Footer also needs this. We do not adjust the bottom
 * padding though because of the safe area.
 */
html.ios ion-modal.modal-card ion-header ion-toolbar:first-of-type,
html.ios ion-modal.modal-sheet ion-header ion-toolbar:first-of-type,
html.ios ion-modal ion-footer ion-toolbar:first-of-type {
  padding-top: 6px;
}
```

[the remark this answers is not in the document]
kumibrr: Oh, I now see the issue. I was applying it to both md and ios.

I've moved it to `modal.ios.scss`

core/src/components/modal/modal.ios.scss
```diff
@@ -87,3 +87,7 @@
 :host(.modal-sheet) .modal-wrapper {
   @include border-radius(var(--border-radius), var(--border-radius), 0, 0);
 }
+
+:host(.modal-sheet.modal-scroll-all) ion-footer ion-toolbar.ios:first-of-type {
+  padding-top: 6px;
+}
```
thetaPC, line 93:
```suggestion
// iOS Sheet Modal - Scroll at all breakpoints
// --------------------------------------------------

/*
 * Sheet modals require an additional padding as mentioned in the
 * `core.scss` file. However, there's a workaround that requires
 * a cloned footer to be added to the modal. This is only necessary
 * because the core styles are not being applied to the cloned footer.
 * It's important to update the padding value when the core styles are
 * updated.
 */
:host(.modal-sheet.modal-scroll-all) ion-footer ion-toolbar:first-of-type {
  padding-top: 6px;
}
```
No need to include `.ios` since the styles are already within the iOS file. Let's also add the comment.

core/src/components/modal/modal.scss
```diff
@@ -166,3 +166,10 @@ ion-backdrop {
   position: absolute;
   bottom: 0;
 }
+
+:host(.modal-sheet.modal-scroll-all) ion-footer {
```
thetaPC, line 170:
```suggestion

// Sheet Modal - Scroll at all breakpoints
// --------------------------------------------------
:host(.modal-sheet.modal-scroll-all) ion-footer {
```

thetaPC (changes requested):
Also update the comment on `core.scss` to this:
```
/*
 * Sheet modals require an additional padding as mentioned in the
 * `core.scss` file. However, there's a workaround that requires
 * a cloned footer to be added to the modal. This is only necessary
 * because the core styles are not being applied to the cloned footer.
 * It's important to update the padding value when the core styles are
 * updated.
 */
```

core/src/components/modal/modal.scss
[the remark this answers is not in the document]
thetaPC: Good catch on the padding!!

packages/vue/src/components/Overlays.ts
thetaPC, line 1: Recommended by the Stencil team to delete this file in order to get the build to pass.

thetaPC (approved):
LGTM, requesting another team member to review

core/src/css/core.scss
```diff
@@ -51,6 +51,13 @@ body.backdrop-no-scroll {
  * the first toolbar in the header.
  * Footer also needs this. We do not adjust the bottom
  * padding though because of the safe area.
+ *
+ * There is a workaround to allow sheet modals to scroll
+ * at any breakpoint. This requires the footer to be
+ * cloned. However, the following styles are not applied.
+ * This is fixed by adding the styles directly on the
+ * modal iOS file. If this padding value is updated, then
+ * the value also needs to be updated in the modal iOS file.
```
brandyscarney, line 60:
```suggestion
```

Instead of adding a comment to indicate that this value needs to be updated in multiple places, we should define a variable in the following file: https://github.com/ionic-team/ionic-framework/blob/main/core/src/components/modal/modal.vars.scss

```scss
/// @prop - Padding top of the sheet modal
$modal-sheet-padding-top:        6px;
```
Then, import the file at the top of this file: 

```scss
@import "../components/modal/modal.vars";
```

And use the variable:

```scss
html.ios ion-modal.modal-card ion-header ion-toolbar:first-of-type,
html.ios ion-modal.modal-sheet ion-header ion-toolbar:first-of-type,
html.ios ion-modal ion-footer ion-toolbar:first-of-type {
  padding-top: $modal-sheet-padding-top;
}
```

Since `modal.vars.scss` is already available in `modal.ios.scss`, the variable can replace `6px` there as well without requiring additional imports.

brandyscarney (changes requested):
Great work on this! I am requesting changes but I am going to talk with @thetaPC offline to see if we can finalize this so don't worry about accepting them. 🙂

core/src/components/modal/animations/md.leave.ts
```diff
@@ -21,16 +21,44 @@ const createLeaveAnimation = () => {
...
+      if (expandToScroll) {
+        // Scroll can only be done when the modal is fully expanded.
+        return;
+      }
+
+      /**
+       * If expandToScroll is disabled, we need to swap
+       * the visibility to the original, so the footer
+       *  dismisses with the modal and doesn't stay
+       *  until the modal is removed from the DOM.
```
brandyscarney, line 46:
```suggestion
       * dismisses with the modal and doesn't stay
       * until the modal is removed from the DOM.
```

core/src/components/modal/animations/ios.leave.ts
```diff
@@ -32,7 +32,33 @@ export const iosLeaveAnimation = (baseEl: HTMLElement, opts: ModalAnimationOptio
...
+      if (expandToScroll) {
+        // Scroll can only be done when the modal is fully expanded.
+        return;
+      }
+
+      /**
+       * If expandToScroll is disabled, we need to swap
+       * the visibility to the original, so the footer
+       *  dismisses with the modal and doesn't stay
+       *  until the modal is removed from the DOM.
```
brandyscarney, line 46:
```suggestion
       * dismisses with the modal and doesn't stay
       * until the modal is removed from the DOM.
```

core/src/components/modal/gestures/sheet.ts
```diff
@@ -110,6 +116,37 @@ export const createSheetGesture = (
     baseEl.classList.add(FOCUS_TRAP_DISABLE_CLASS);
   };
 
+  /**
+   * Used when `expandToScroll` is disabled.
+   * Changes the footer that is currently visible
+   * @param footer - The footer to show
+   */
```
brandyscarney, line 123:
```suggestion
  /**
   * Toggles the visible modal footer when `expandToScroll` is disabled.
   * @param footer The footer to show.
   */
```

core/src/components/modal/gestures/sheet.ts
```diff
@@ -187,6 +233,16 @@ export const createSheetGesture = (
      */
     canDismissBlocksGesture = baseEl.canDismiss !== undefined && baseEl.canDismiss !== true && minBreakpoint === 0;
 
+    /**
+     * If expandToScroll is disabled, we need to swap
+     * the visibility to the original, so if the modal
```
brandyscarney, line 238:
```suggestion
     * the footer visibility to the original, so if the modal
```

core/src/components/modal/gestures/sheet.ts
```diff
@@ -323,6 +379,20 @@ export const createSheetGesture = (
         },
       ]);
 
+      if (contentAnimation) {
+        /**
+         * The modal content should scroll at any breakpoint when expandToScroll
+         * is disabled. In order to do this, the content needs to be completely
+         * viewable so scrolling can access everything. Othewise, the default
```
brandyscarney, line 386:
```suggestion
         * viewable so scrolling can access everything. Otherwise, the default
```

core/src/components/modal/gestures/sheet.ts
```diff
@@ -332,20 +402,30 @@ export const createSheetGesture = (
      */
     gesture.enable(false);
 
+    /**
+     * If expandToScroll is disabled, we need to swap
+     * the visibility to the cloned one so the footer
```
brandyscarney, line 407:
```suggestion
     * the footer visibility to the cloned one so the footer
```

core/src/components/modal/gestures/sheet.ts
```diff
@@ -332,20 +402,30 @@ export const createSheetGesture = (
...
-     * scrolling immediately. The sheet modal animation takes ~500ms to finish
-     * so if we wait until then there is a visible delay for when scrolling is
-     * re-enabled. Native iOS allows for scrolling on the sheet modal as soon
-     * as the gesture is released, so we align with that.
+     * If the sheet is going to be fully expanded or if the sheet has toggled
+     * to scroll at any breakpoint then we should enable scrolling immediately.
+     * then we should enable scrolling immediately. The sheet modal animation
+     * takes ~500ms to finish so if we wait until then there is a visible delay
+     * for when scrolling is re-enabled. Native iOS allows for scrolling on the
+     * sheet modal as soon as the gesture is released, so we align with that.
```
brandyscarney, line 426:
```suggestion
     * Enables scrolling immediately if the sheet is about to fully expand
     * or if it allows scrolling at any breakpoint. Without this, there would
     * be a ~500ms delay while the modal animation completes, causing a
     * noticeable lag. Native iOS allows scrolling as soon as the gesture is
     * released, so we align with that behavior.
```

brandyscarney (approved):

OS-giulianasilva (approved):

thetaPC:
@rgoldiez Hi there! Version 8.5 doesn’t have an official release date yet, but definitely keep an eye out as it will be available soon. Thanks for your patience!

thetaPC:
@a7medm7med Thank you for letting us know! We've confirmed that native iOS does allow this and we are working on [adding](https://github.com/ionic-team/ionic-framework/pull/30235) this.

PR #30117, opened 2025-01-08
feat(tab-bar): make box shadow independent of expand mode

OS-pedrolourenco:
Issue number: internal
---------

When the tab bar's expand mode is 'full' on the ionic theme, it does not have a box shadow to separate it from the rest of the page.

## What is the new behavior?

- The tab bar on the ionic theme now has a box shadow independently of its expand mode.

- [ ] Yes
- [x] No


thetaPC:
@OS-pedrolourenco please make sure the PR checks are passing before marking it ready for review. It speeds up the process for the reviewer.

thetaPC (changes requested):
One small nitpick, the PR title. This is not a feat, a feat is when we're adding something new. In this case, we're making a change to an existing style, so it's a refactor. Also we try to start the sentence with a verb.

`refactor(tab-bar): make box shadow independent of expand for ionic`

OS-pedrolourenco:
@thetaPC I have **changed the behaviour** of the tab-bar so that it now has a shadow when its expand mode is 'full' and not just 'compact', as **requested by UX**. Wouldn't you say that that is feature and not a refactor?
I will correct the verb usage right away though. Just wanted to align with you on this for future reference.

thetaPC:
> @thetaPC I have **changed the behaviour** of the tab-bar so that it now has a shadow when its expand mode is 'full' and not just 'compact', as **requested by UX**. Wouldn't you say that that is feature and not a refactor? I will correct the verb usage right away though. Just wanted to align with you on this for future reference.

As you mentioned, you did a code change so that's a refactor. Here's the definitions for:
- **refactor**: A code change that neither fixes a bug nor adds a feature
- **feat**: A new feature
More information on these can be found at the [Conventional Commits](https://www.conventionalcommits.org/en/v1.0.0/) site.

PR #30133, opened 2025-01-14
fix(segment-button): protect connectedCallback for when segment-content has not yet been created

OS-pedrolourenco:
Issue number: internal

---------

When the `connectedCallback` method is called for a segment-button and its corresponding segment-content has not been created in that instant, a console error is thrown and the method returns.

## What is the new behavior?

- `connectedCallback` will now wait, at most 1 second, for the corresponding segment-content to be created.
- The new behaviour can be tested in segment-view/basic.

- [ ] Yes
- [x] No

### DIFF
--- a/core/src/components/segment-button/segment-button.tsx
+++ b/core/src/components/segment-button/segment-button.tsx
@@ -80,15 +80,16 @@ export class SegmentButton implements ComponentInterface, ButtonInterface {
 
   private waitForSegmentContent(ionSegment: HTMLIonSegmentElement | null, contentId: string): Promise<HTMLElement> {
     return new Promise((resolve, reject) => {
-      if (!ionSegment) {
-        reject(new Error(`Segment not found when looking for Segment Content`));
-      }
-
-      let timeoutId: any = null;
+      let timeoutId: NodeJS.Timeout | undefined = undefined;
       let animationFrameId: number;
 
       const check = () => {
-        const segmentView = this.getNextSiblingOfType<HTMLIonSegmentViewElement>(ionSegment!); // Skip the text nodes
+        if (!ionSegment) {
+          reject(new Error(`Segment not found when looking for Segment Content`));
+          return;
+        }
+
+        const segmentView = this.getNextSiblingOfType<HTMLIonSegmentViewElement>(ionSegment); // Skip the text nodes
         const segmentContent = segmentView?.querySelector(
           `ion-segment-content[id="${contentId}"]`
         ) as HTMLIonSegmentContentElement | null;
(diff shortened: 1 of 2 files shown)


core/src/components/segment-button/segment-button.tsx
```diff
@@ -67,7 +67,51 @@ export class SegmentButton implements ComponentInterface, ButtonInterface {
...
+    return new Promise((resolve, reject) => {
+      if (!ionSegment) {
+        reject(new Error(`Segment not found when looking for Segment Content`));
+      }
+
+      let timeoutId: any = null;
+      let animationFrameId: number;
+
+      const check = () => {
+        const segmentView = this.getNextSiblingOfType<HTMLIonSegmentViewElement>(ionSegment!); // Skip the text nodes
```
christian-bromann, line 91: How about moving the `if (!ionSegment) {` into the `check` method instead so we can avoid using `!`?

core/src/components/segment-button/segment-button.tsx
```diff
@@ -67,7 +67,51 @@ export class SegmentButton implements ComponentInterface, ButtonInterface {
...
+    return null;
+  }
+
+  private waitForSegmentContent(ionSegment: HTMLIonSegmentElement | null, contentId: string): Promise<HTMLElement> {
+    return new Promise((resolve, reject) => {
+      if (!ionSegment) {
+        reject(new Error(`Segment not found when looking for Segment Content`));
+      }
+
+      let timeoutId: any = null;
```
christian-bromann, line 87: let's properly type this
```suggestion
      let timeoutId: number | undefined = undefined;
```

core/src/components/segment-button/segment-button.tsx
```diff
@@ -67,7 +67,51 @@ export class SegmentButton implements ComponentInterface, ButtonInterface {
...
+      let animationFrameId: number;
+
+      const check = () => {
+        const segmentView = this.getNextSiblingOfType<HTMLIonSegmentViewElement>(ionSegment!); // Skip the text nodes
+        const segmentContent = segmentView?.querySelector(
```
christian-bromann, line 92: if we have the `if (!ionSegment) {` check in this function we can just do
```suggestion
        const segmentContent = segmentView.querySelector(
```

christian-bromann (commented):
Some minor comments

core/src/components/segment-button/segment-button.tsx
OS-pedrolourenco: I'm sorry but I don't see how that is the case seen as though getNextSiblingOfType could still return null

christian-bromann: True, disregard this then.

christian-bromann (approved):
LGTM 👍

OS-giulianasilva (approved):
Nice!

PR #30134, opened 2025-01-14
test(tab-bar): add snapshot for box shadow

OS-pedrolourenco:
Issue number: internal

---------

This PR is a follow-up of this one: https://github.com/ionic-team/ionic-framework/pull/30117

## What is the new behavior?

-  This PR adds a snapshot test and modifies the existing one in order to test the change done in the previous PR.

- [ ] Yes
- [x] No


core/src/components/tab-bar/test/expand/tab-bar.e2e.ts
thetaPC, line 1: Snapshots should always aim to be as small as possible. That's why the original test was using `tabBar` instead of content. Since we want to capture the box shadow, then create a container that has padding and use the container for the snapshot instead for both tests. Here's some pseudo code:

```ts
.container {
   padding: 20px 10px;
}

<ion-content>
   <div class="container">
      <ion-tab-bar>
   </div>
</ion-content>

const container = page.locator('.container');

await expect(container).toHaveScreenshot...
```

thetaPC (changes requested):
The PR also needs:
- a test test, not feat, since you're adding a test
- not be redundant, the component is already mentioned in the scope so no need to mention it again. The same goes for the type.

`test(tab-bar): add snapshot for box shadow`

core/src/components/tab-bar/test/expand/tab-bar.e2e.ts
OS-pedrolourenco: Done. The compact snapshots look a bit strange though.
By the way, I ended up not using the container in your suggestion seen as though it had no effect on the snapshots.

thetaPC: This is because I used pseudo code as in my code is only semi complete so:
1. Please use the proper syntax
2. Add the suggested code to the correct locations
3. use the container

```ts
.container {
   padding: 20px 10px;
}

<ion-content>
   <div class="container">
      <ion-tab-bar> // not complete, use proper syntax
   </div>
</ion-content>

const container = page.locator('.container');

await expect(container).toHaveScreenshot... // not complete, use proper syntax
```

thetaPC (changes requested):

core/src/components/tab-bar/test/expand/tab-bar.e2e.ts
OS-pedrolourenco: I hope I have addressed your concern with my latest commit. The only adjustment I had to do was to manually set the container dimensions on the compact tab bar test since, with only the padding attribute set up, the container shown in the snapshot did not fit the tab bar for some reason and was hiding it instead:
image
If you have any suggestion for better working around this, please let me know.

thetaPC (approved):
LGTM, assigning to Brandy for a second check. Once she gives an approval, then this PR can be merged.

core/src/components/tab-bar/test/expand/tab-bar.e2e.ts
```diff
@@ -14,31 +58,42 @@ configs({ directions: ['ltr'], modes: ['ionic-md'] }).forEach(({ title, screensh
...
-              <ion-label>Label</ion-label>
-            </ion-tab-button>
-  
-            <ion-tab-button tab="3">
-              <ion-icon name="triangle-outline"></ion-icon>
-              <ion-label>Label</ion-label>
-            </ion-tab-button>
-          </ion-tab-bar>
-        `,
+          <ion-content>
```
brandyscarney, line 70: We should remove `ion-content` as this is not needed

brandyscarney (approved):
Approving with the removal of `ion-content` 🙂

PR #30135, opened 2025-01-15
[Optimization] The useXXX method can be directly called in the vue component, and it can be automatically uninstalled

mchao123:
Optimized useBackButton and useKeyboard, now direct reference in vue component will support auto uninstall.

## What is the new behavior?

- Added automatic uninstallation for useBackButton.
- Added automatic uninstallation for useKeyboard.

- [ ] Yes
- [ + ] No

## Other information

This is a default specification for vue users that the useXXX function should be called directly from within the vue component and not subsequently care about uninstallation.

### DIFF
--- a/packages/vue/src/hooks/back-button.ts
+++ b/packages/vue/src/hooks/back-button.ts
@@ -1,4 +1,5 @@
 import type { BackButtonEvent } from "@ionic/core/components";
+import { onBeforeUnmount, getCurrentInstance } from 'vue'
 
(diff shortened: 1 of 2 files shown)


brandyscarney:
Hello! Thank you for the PR! Is there an associated issue for this?

mchao123:
Sorry for the late reply, this commit mainly uses vue's hook to automatically unload bound events, which should normally be properly logged out when the component is unloaded

thetaPC:
@mchao123 Thanks for the PR! Please create an issue so the team can understand the need for this optimization.

PR #30138, opened 2025-01-17
fix(segment-button): protect connectedCallback for when segment-content has not yet been created (cherry-pick)

OS-pedrolourenco:
Issue number: internal

---------

 When the `connectedCallback` method is called for a segment-button and its corresponding segment-content has not been created in that instant, a console error is thrown and the method returns.

## What is the new behavior?

- `connectedCallback` will now wait, at most 1 second, for the corresponding segment-content to be created.
- The new behaviour can be tested in segment-view/basic.

- [ ] Yes
- [x] No

### DIFF
--- a/core/src/components/segment-button/segment-button.tsx
+++ b/core/src/components/segment-button/segment-button.tsx
@@ -91,7 +91,7 @@ export class SegmentButton implements ComponentInterface, ButtonInterface {
         const segmentContent = segmentView?.querySelector(
           `ion-segment-content[id="${contentId}"]`
         ) as HTMLIonSegmentContentElement | null;
-        if (segmentContent) {
+        if (segmentContent && timeoutId) {
           clearTimeout(timeoutId); // Clear the timeout if the segmentContent is found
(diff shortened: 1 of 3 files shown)


core/src/components/segment-button/segment-button.tsx
```diff
@@ -65,7 +65,52 @@ export class SegmentButton implements ComponentInterface, ButtonInterface {
     this.updateState();
   }
 
-  connectedCallback() {
+  private getNextSiblingOfType<T extends Element>(element: Element): T | null {
```
christian-bromann, line 68: This function does not need to be connected to the component class as it doesn't uses any component state.

christian-bromann (approved):
One minor nit but overall LGTM 👍

core/src/components/segment-button/segment-button.tsx
OS-pedrolourenco: Moved the method to helpers.tsx

christian-bromann (approved):
👍

PR #30140, opened 2025-01-20
feat(checkbox): add helperText and errorText properties

### LINKED ISSUE #29810: feat: add helperText and errorText properties for ion-checkbox
### Prerequisites

Checkbox are often used in an ion-items in ion-lists. Since migrating to the new label/helper/error syntax, it is not possible anymore to easily add a helper or error text to ion-checkbox.

### Describe the Use Case

See above.

_No response_

### Describe Alternatives

_No response_

### Related Code

_No response_

### Additional Information

_No response_

brandyscarney:
Issue number: resolves #29810

---------

Checkbox does not support helper and error text.

## What is the new behavior?
Adds support for helper and error text, similar to input and textarea.

- [ ] Yes
- [x] No

## Other information

- [Bottom Content: Preview](https://ionic-framework-git-rou-11141-ionic1.vercel.app/src/components/checkbox/test/bottom-content)
- [Item: Preview](https://ionic-framework-git-rou-11141-ionic1.vercel.app/src/components/checkbox/test/item)


core/src/components/checkbox/checkbox.scss
```diff
@@ -148,6 +148,53 @@ input {
...
+
+  justify-content: space-between;
+
+  font-size: dynamic-font(12px);
+
+  white-space: normal;
+}
+
+:host(.checkbox-label-placement-stacked) .checkbox-bottom {
+  font-size: dynamic-font(16px);
```
brandyscarney, line 167: I added this because without it the text is really small.

core/src/components/checkbox/checkbox.scss
```diff
@@ -260,6 +309,8 @@ input {
  */
 :host(.checkbox-label-placement-stacked) .checkbox-wrapper {
   flex-direction: column;
+
+  text-align: center;
```
brandyscarney, line 313: These text alignment changes are necessary because without it the label is left aligned and when you change between helper/error text it jumps:


https://github.com/user-attachments/assets/c396a00a-d25c-4970-8e31-1176d94ea22c

After my changes:

https://github.com/user-attachments/assets/16fa1e52-2d73-4093-99f4-05faa8961810

This issue does not exist without helper/error text.

core/src/components/checkbox/checkbox.tsx
```diff
@@ -167,6 +182,48 @@ export class Checkbox implements ComponentInterface {
...
+     * undefined and empty string values should
+     * be treated as not having helper/error text.
+     */
+    const hasHintText = !!helperText || !!errorText;
+    if (!hasHintText) {
+      return;
+    }
+
+    return (
+      <div class="checkbox-bottom">
```
brandyscarney, line 216: The class `checkbox-bottom` doesn't make as much sense here since it's not on the bottom of the entire checkbox but it is under the label still and I named it this to remain consistent with input, textarea, etc.

core/src/components/checkbox/test/bottom-content/checkbox.e2e.ts-snapshots/checkbox-helper-text-end-wrapping-ios-ltr-Mobile-Chrome-linux.png
brandyscarney, line 1: The helper text does look a bit odd here with the text alignment. If we left it aligned to the left we would need to change the justify to make it look okay in cases where the helper text is short.

core/src/components/checkbox/checkbox.scss
```diff
@@ -217,6 +224,8 @@ input {
  */
 :host(.checkbox-label-placement-end) .checkbox-wrapper {
   flex-direction: row-reverse;
+
+  justify-content: start;
```
brandyscarney, line 228: This might cause unwanted changes in some use cases, but this matches what is expected in the Figma design. Here is how it looks before and after this change with `width: 100%` on the checkbox:

| Before | After |
| ---| ---|
| helper-text-label-end-before | helper-text-label-end-after |
| label-placement-end-before | label-placement-end-after |

Because this changes the default behavior in an item, maybe we want to change this later and just recommend they set the `justify` property when using it with helper text? I am not sure, but it jumps around with the `space-between` default:

https://github.com/user-attachments/assets/12278076-378c-4b56-bbf9-0d07fd8fd4f4

core/src/components/checkbox/checkbox.scss
```diff
@@ -287,6 +298,46 @@ input {
   @include transform-origin(center, top);
 }
 
+// Justify Content
```
brandyscarney, line 301: I didn't make changes to any of this code I just moved it lower so it would take priority over the `justify-content` I specified in `label-placement-end`.

core/src/components/checkbox/checkbox.scss
thetaPC: This makes sense to me. Everything is centered already so the supporting text should follow that.

core/src/components/checkbox/checkbox.scss
thetaPC: @brandyscarney which Figma design are you referring to?

core/src/components/checkbox/checkbox.tsx
```diff
@@ -228,6 +287,7 @@ export class Checkbox implements ComponentInterface {
             part="label"
           >
             <slot></slot>
+            {this.renderHintText()}
```
thetaPC, line 290: The text (helper or error) is being read twice by VoiceOver on Safari, Firefox, and Chrome. After some quick debugging, it seems that this happening because the text is inside of label. `ion-input` has it after the `<label>` which mimics how the native input states it.

https://github.com/user-attachments/assets/d72b0a15-9266-45b7-9872-70c99b3f2562

thetaPC, line 290: I noticed that the focus with VoiceOver does not highlight correctly. This is happening on the `main` branch. We should consider creating a spike to fix it.

core/src/components/checkbox/test/bottom-content/checkbox.e2e.ts
```diff
@@ -0,0 +1,205 @@
...
+  });
+});
+
+/**
+ * Rendering is different across modes
+ */
+configs({ modes: ['ios', 'md'], directions: ['ltr'] }).forEach(({ title, screenshot, config }) => {
+  test.describe(title('checkbox: helper text rendering'), () => {
+    // Check the default label placement, end, and stacked
+    [undefined, 'end', 'stacked'].forEach((labelPlacement) => {
```
thetaPC, line 102: What was the reasoning to only checking for these placements?

core/src/components/checkbox/test/bottom-content/checkbox.e2e.ts
```diff
@@ -0,0 +1,205 @@
...
+
+        const bottomEl = page.locator('ion-checkbox');
+        await expect(bottomEl).toHaveScreenshot(
+          screenshot(`checkbox-helper-text${labelPlacement ? `-${labelPlacement}` : ''}-wrapping`)
+        );
+      });
+    });
+  });
+
+  test.describe(title('checkbox: error text rendering'), () => {
```
thetaPC, line 137: Why does error text not get a check for the end placement?

thetaPC (changes requested):
The supporting text being read twice is the main issue. The rest are just questions.

core/src/components/checkbox/checkbox.scss
brandyscarney: @thetaPC I will send you a link

thetaPC: After some review, I agree with these changes. It also seems like a fix so it's a win win.

core/src/components/checkbox/test/bottom-content/checkbox.e2e.ts
brandyscarney: The layout of the helper text wasn't different for `'fixed'` compared to `undefined` so I left it off, but I can add it if you think we should check it:

| `undefined` | `'fixed'` |
| --- | --- |
| checkbox-default | checkbox-fixed |

core/src/components/checkbox/test/bottom-content/checkbox.e2e.ts
brandyscarney: The error text checks are mainly making sure it shows up and the colors look right. The layout for the error text is the same as the helper text, but in the case of stacked it needs to still be centered so I checked that too. Honestly we could probably remove stacked for it, but I don't feel strongly either way if you want me to check all of the label placements.

core/src/components/checkbox/checkbox.tsx
brandyscarney: Hey! Are you referring to macOS or iOS?

Here's what I’m seeing:

**Chrome macOS** 



**Firefox macOS**



**iOS**



Could you clarify the issue you're seeing? Thanks!

thetaPC: I see it on Safari macOS.

brandyscarney: Ah the one browser I didn't check 😂 Okay I see it, I will make a ticket!

brandyscarney: ROU-11658 created

thetaPC: Thank you!

brandyscarney: Tested in macOS, iOS and Android talkback and it is announcing properly after this fix: https://github.com/ionic-team/ionic-framework/pull/30140/commits/bac73833ca401338766cafcfd7480208b49b1d64

core/src/components/checkbox/test/bottom-content/checkbox.e2e.ts
thetaPC: I'm fine with this. Thanks!

core/src/components/checkbox/test/bottom-content/checkbox.e2e.ts
thetaPC: I'm fine with this. Thanks!

thetaPC (approved):
LGTM

core/src/components/checkbox/checkbox.tsx
```diff
@@ -174,6 +189,48 @@ export class Checkbox implements ComponentInterface {
...
+   */
+  private renderHintText() {
+    const { helperText, errorText, helperTextId, errorTextId } = this;
+
+    /**
+     * undefined and empty string values should
+     * be treated as not having helper/error text.
+     */
+    const hasHintText = !!helperText || !!errorText;
+    if (!hasHintText) {
```
BenOsodrac, line 218: why create the const and not make the validation on the if directly?

BenOsodrac (approved):

core/src/components/checkbox/checkbox.scss
```diff
@@ -148,46 +148,53 @@ input {
...
+
+  display: flex;
 
-:host(.checkbox-justify-space-between) .checkbox-wrapper {
   justify-content: space-between;
-}
 
-:host(.checkbox-justify-start) .checkbox-wrapper {
-  justify-content: start;
+  font-size: dynamic-font(12px);
```
BenOsodrac, line 161: Same as above. This should be separated and on Ionic theme use the 
body.md.regular token.

core/src/components/checkbox/checkbox.scss
BenOsodrac: Same as above* :)

core/src/components/checkbox/checkbox.scss
```diff
@@ -148,46 +148,53 @@ input {
   opacity: 0;
 }
 
-// Justify Content
-// ---------------------------------------------
+// Checkbox Bottom Content
+// ----------------------------------------------------------------
+
+.checkbox-bottom {
+  @include padding(4px, null, null, null);
```
BenOsodrac, line 155: This should not be on the common scope, as each theme will use its proper token (even if the value is the same). 

Maybe we can create now the checkbox.native.scss and divide the Design System decision there (mostly this padding and the font-sizes I think)

BenOsodrac (changes requested):

core/src/components/checkbox/checkbox.tsx
brandyscarney: I just copied this implementation from input: https://github.com/ionic-team/ionic-framework/blob/cd5c27a12a03cc11da7ae86d89ad402343fed1c0/core/src/components/input/input.tsx#L625

BenOsodrac (approved):

core/src/components/checkbox/checkbox.scss
brandyscarney: Spoke with Bernardo offline - this will be accomplished in another task

core/src/components/checkbox/checkbox.scss
brandyscarney: https://github.com/ionic-team/ionic-framework/pull/30140#discussion_r1969963015

PR #30143, opened 2025-01-21
feat(select): add helperText and errorText properties 

### LINKED ISSUE #29205: feat: select, add errorText and helperText props
### Prerequisites

ion-select needs the errorText attribute to display e.g. if I want to make the selection mandatory.

Please add this or provide a proper workaround to display an error message below the ion-select

### Describe the Use Case

When I use Angular FormControls, I can smoothly add errorText to Input etc. to display that something is wrong while validating the form.

_No response_

### Describe Alternatives

_No response_

### Related Code

_No response_

### Additional Information

_No response_

brandyscarney:
Issue number: resolves #29205

---------

Select does not support helper and error text.

## What is the new behavior?
- Adds support for `helperText` and `errorText`
- Adds parts for `helper-text`, `error-text` and `supporting-text`
- Adds an e2e test for helper and error text with functional tests and screenshot tests

- [ ] Yes
- [x] No

## Other information

[Preview](https://ionic-framework-git-rou-11551-ionic1.vercel.app/src/components/select/test/bottom-content)

---------

Co-authored-by: swimer11 <65334157+swimer11@users.noreply.github.com>


core/src/components/select/select.tsx
```diff
@@ -983,13 +998,64 @@ export class Select implements ComponentInterface {
...
+  private renderHintText() {
+    const { helperText, errorText, helperTextId, errorTextId } = this;
+
+    return [
+      <div id={helperTextId} class="helper-text" part="supporting-text helper-text">
+        {helperText}
+      </div>,
+      <div id={errorTextId} class="error-text" part="supporting-text error-text">
+        {errorText}
+      </div>,
```
brandyscarney, line 1036: I added the `supporting-text` part so users could change them at the same time using: 

```css
ion-select::part(supporting-text) {
  font-size: 20px;
}
```

but if we think it should only be the one part they could write this without it:

``` css
ion-select::part(helper-text),
ion-select::part(error-text) {
  font-size: 20px;
}
```

I am copying [Material Design 2](https://m2.material.io/components/text-fields#interactive-demo) and [Material Design 3](https://m3.material.io/components/text-fields/overview) naming here, but let me know if there are other ideas.

core/src/components/select/select.scss
thetaPC, line 1: Would it be beneficial to update `@prop --highlight-color-invalid` definition to also mention that it will also change error text?

https://github.com/ionic-team/ionic-framework/blob/66d6b07ece162450ffed822b4807d1bfb10c632b/core/src/components/select/select.scss#L19

thetaPC (approved):
LGTM, potential suggestion. No blockers.

core/src/components/select/select.scss
brandyscarney: I don't think so, since we don't document similar cases elsewhere. For example, changing the `--color-checked` variable also updates the `--ripple-color` variable (changing the ripple effect color) in segment button but we don't explicitly document this: https://github.com/ionic-team/ionic-framework/blob/cd5c27a12a03cc11da7ae86d89ad402343fed1c0/core/src/components/segment-button/segment-button.scss#L267

It also updates the `--indicator-color`: https://github.com/ionic-team/ionic-framework/blob/cd5c27a12a03cc11da7ae86d89ad402343fed1c0/core/src/components/segment-button/segment-button.md.scss#L106

If you feel like we should though I can make the changes.

PR #30155, opened 2025-01-24
feat(select): add required prop

OS-giulianasilva:

---------

- Currently, the screen reader do not announce the component as required when `required={true}`.

## What is the new behavior?
- Added a new `required` prop to be used for accessibility purposes that adds the `aria-required` attribute to select's inner native button.

- [ ] Yes
- [x] No


joselrio (approved):

core/src/components/select/test/select.spec.tsx
```diff
@@ -125,3 +125,35 @@ describe('select: slot interactivity', () => {
     expect(divSpy).toHaveBeenCalled();
   });
 });
+
+describe('ion-select: required', () => {
+  it('should have a aria-required attribute as true in inner button', async () => {
```
gnbm, line 130:
```suggestion
  it('the inner button should have the aria-required attribute set to true', async () => {
```

core/src/components/select/test/select.spec.tsx
```diff
@@ -125,3 +125,35 @@ describe('select: slot interactivity', () => {
...
+    });
+
+    const select = page.body.querySelector('ion-select')!;
+
+    const nativeButton = select.shadowRoot!.querySelector('button')!;
+
+    expect(nativeButton.getAttribute('aria-required')).toBe('true');
+  });
+
+  it('should not have a aria-required attribute as false in inner button', async () => {
```
gnbm, line 145:
```suggestion
  it('the inner button should have the aria-required attribute set to false', async () => {
```

gnbm (approved):

core/src/components/select/test/select.spec.tsx
thetaPC: The current version aligns with the testing names standards for Ionic. Please leave it as is.

core/src/components/select/test/select.spec.tsx
thetaPC: The current version aligns with the testing names standards for Ionic. Please leave it as is.

thetaPC (approved):
LGTM, tested with VoiceOver

PR #30156, opened 2025-01-24
feat(toggle): add required prop

OS-giulianasilva:

---------

- Currently, the screen reader do not announce the component as required when `required={true}`.

## What is the new behavior?
- Added a new `required` prop to be used for accessibility purposes that adds the `required` attribute to toggle's inner native input.

- [ ] Yes
- [x] No


joselrio (approved):

core/src/components/toggle/toggle.tsx
```diff
@@ -290,7 +297,8 @@ export class Toggle implements ComponentInterface {
   }
 
   render() {
-    const { activated, color, checked, disabled, el, justify, labelPlacement, inputId, name, alignment } = this;
+    const { activated, color, checked, disabled, el, justify, labelPlacement, inputId, name, alignment, required } =
+      this;
```
gnbm, line 301: Any reason why we have this extra line-break before `this`?

JoaoFerreira-FrontEnd (approved):

core/src/components/toggle/toggle.tsx
OS-giulianasilva: Just `eslint` reasons haha

thetaPC (approved):
LGTM, tested with VoiceOver

PR #30158, opened 2025-01-24
fix(vue): make necessary updates for Vue update

christian-bromann:
This patch includes some necessary updates for `@stencil/vue-output-target@v0.9.0`:

- we started to export Stencils helpers as runtime via `@stencil/vue-output-target/runtime` similar to what we did in React
- this version requires some updates to Vue and TypeScript as well
- adjustments related to that update


brandyscarney:
Closing for https://github.com/ionic-team/ionic-framework/pull/30159

PR #30159, opened 2025-01-25
fix(vue): update Stencil Vue output target

christian-bromann:
This patch includes some necessary updates for `@stencil/vue-output-target@v0.9.0`:

- we started to export Stencils helpers as runtime via `@stencil/vue-output-target/runtime` similar to what we did in React
- this version requires some updates to Vue and TypeScript as well
- adjustments related to that update


gnbm (approved):

packages/vue/package.json
```diff
@@ -52,7 +52,9 @@
     "@babel/types": "^7.18.4",
     "@ionic/eslint-config": "^0.3.0",
     "@ionic/prettier-config": "^2.0.0",
+    "@rollup/plugin-node-resolve": "^16.0.0",
     "@rollup/plugin-typescript": "^11.1.5",
+    "@stencil/vue-output-target": "0.9.4",
```
brandyscarney, line 57: Does this need to be here when we have it here: https://github.com/ionic-team/ionic-framework/blob/0030be8195798fb097d931efd5c9913c99545780/core/package.json#L53

We don't do this for any of the other frameworks: https://github.com/ionic-team/ionic-framework/blob/0030be8195798fb097d931efd5c9913c99545780/packages/angular/package.json#L62

christian-bromann: > Does this need to be here when we have it here:

Yes, starting with `v0.9.0` we are shipping the Vue runtime as part of the package (see `import { defineContainer } from '@stencil/vue-output-target/runtime';` in [here](https://github.com/ionic-team/ionic-framework/blob/746b3380f8be87b6d01280217e0fc744b72df49e/packages/vue/src/proxies.ts#L3)). Since this code will be bundled with Rollup, we are ok keeping it as dev dependencies so for the Ionic user, nothing changes.

> We don't do this for any of the other frameworks:

We will have to do the same once we update React. Angular doesn't ship a runtime export.

brandyscarney: Thanks for the info! Sounds good.

brandyscarney (approved):

OS-giulianasilva (approved):

robkorv:
Just an FYI as I am reaching a deadline and do not have time to provide an example. (I secretly hope someone who reads this has an aha! moment, that I can benefit from ;)

I'm loosing all `ion*` events in vue using tabs with v8.4.3.
The ion-select does emit `update:modelValue` but not the `ionChange`.
The same goes for `ionInfinite` emit, it does not get triggered anymore on ion-infinite-scroll.

Reverting back to v8.4.2 resolves this.

brandyscarney:
Thanks @robkorv we are aware of this issue and looking into it!

robkorv:
Awesome! I see someone submitted an issue. I will track it from there. https://github.com/ionic-team/ionic-framework/issues/30206

PR #30186, opened 2025-02-11
feat(css): add new utility-classes

BenOsodrac:
Issue number: internal

---------

## What is the new behavior?

Added new utility-classes related to:
- flex
- position
- size

- [ ] Yes
- [x] No


thetaPC (commented):
Since these appear to be more directed towards the `ionic` theme, it might be best to move them to the `css/ionic` folder. Why are they being added to core?

BenOsodrac:
> Since these appear to be more directed towards the `ionic` theme, it might be best to move them to the `css/ionic` folder. Why are they being added to core?

The request come from the Ionic theme context yes, but I its all classes related to agnostic/normal css rules, that I see no problem adding to the common context, as some of them, like the flex ones, are just an extension of the existing offering.

PR #30200, opened 2025-02-19
fix(select): fix auto scroll to selected item in ion-select

JoaoFerreira-FrontEnd:
Issue number: internal
---------

- when the ion-select is with the interface action-sheet or alert is not scrolling to the selected item on open

## What is the new behavior?

- guarantee focusVisibleElement is called on all interfaces;
- change the test page so all select have scroll;

- [preview page ](https://ionic-framework-git-rou-11568-ionic1.vercel.app/src/components/select/test/basic)

- [ ] Yes
- [x] No


thetaPC:
@JoaoFerreira-FrontEnd please fix the checks before marking as ready for review

PR #30205, opened 2025-02-21
feat(ion-datetime): add hidden-days when theme:ionic

JoaoFerreira-FrontEnd:
Issue number: internal

---------

## What is the new behavior?

- add logic to generate more days per month if the displayHiddenDays flag is true;
- change datetime component to use this attribute when in theme: ionic;
- document changes;

- [ ] Yes
- [x] No

### DIFF
--- a/core/src/components/datetime/datetime.tsx
+++ b/core/src/components/datetime/datetime.tsx
@@ -2193,7 +2193,7 @@ export class Datetime implements ComponentInterface {
       </div>
     );
   }
-  private renderMonth(month: number, year: number) {
+  private renderMonth(month: number, year: number, theme: Theme) {
     const { disabled, readonly } = this;
 
     const yearAllowed = this.parsedYearValues === undefined || this.parsedYearValues.includes(year);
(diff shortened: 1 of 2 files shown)


core/src/components/datetime/utils/data.ts
```diff
@@ -127,14 +134,38 @@ export const getDaysOfMonth = (month: number, year: number, firstDayOfWeek: numb
...
+    const positiveOffset = 41 - (numDays + offset);
+    for (let i = 0; i < positiveOffset; i++) {
+      days.push( {day:i+1, dayOfWeek:((numDays + offset) + i) % 7, hiddenDay: true} )
+    }
 
-  for (let i = 0; i <= offset; i++) {
-    days = [{ day: null, dayOfWeek: null }, ...days];
+  } else {
+    for (let i = 0; i <= offset; i++) {
+      days = [{ day: null, dayOfWeek: null, hiddenDay:true }, ...days];
```
thetaPC, line 167: Wouldn't `hiddenDay` be false?

core/src/components/datetime/datetime.tsx
```diff
@@ -2234,10 +2234,35 @@ export class Datetime implements ComponentInterface {
...
+              } else if(hiddenDay && day !== null && day < 15) {
+                // Leading with the hidden day from the next month
+                // if its a hidden day and is lower than '15' (first two weeks)
+                if(month === 12) {
+                  _year = year + 1;
+                  _month = 1;
+                } else {
+                  _month = month + 1;
+                }
+              }
```
thetaPC, line 2262: Wouldn't this calculation be done through `getDaysOfMonth`?

thetaPC (commented):
Please add more comments of what the code is doing. Datetime is already complex, this new logic should be documented well so the devs aren't spending too much time trying to figure out what's going on.

PR #30213, opened 2025-02-25
feat(badge): add support for hint feature

BenOsodrac:
Issue number: internal

---------

## What is the new behavior?

- Added new position prop.
- Based on the prop, a new badge-position css class is added on the host.
- Removed css rule that prevented the badge to be rendered when empty.
- Added common and themes styles to support the position prop and new scale size (on ios/md defaults to min-width variable, as size is not yet supported on native themes).
- Added new tests specific for this new feature and added new screenshots.
- Support to properly work inside Avatar, Button and TabButton will be added on different tasks (in the meantime the snapshots will appear wrong for these use-cases)

- [ ] Yes
- [x] No

## Other information

- [Ionic Theme Sample](https://ionic-framework-6n0cn175k-ionic1.vercel.app/src/components/badge/test/hint?ionic:theme=ionic)
- [MD Theme Sample](https://ionic-framework-6n0cn175k-ionic1.vercel.app/src/components/badge/test/hint?ionic:theme=md)
- [iOS Theme Sample](https://ionic-framework-6n0cn175k-ionic1.vercel.app/src/components/badge/test/hint?ionic:theme=ios)


core/src/components/tab-button/test/basic/tab-button.e2e.ts-snapshots/tab-button-badge-ios-ltr-Mobile-Firefox-linux.png
BenOsodrac, line 1: This is temporary until we add support for the new feature on ionTabs on the following PR's.

core/src/components/badge/test/hint/badge.e2e.ts
```diff
@@ -0,0 +1,14 @@
+import { expect } from '@playwright/test';
+import { configs, test } from '@utils/test/playwright';
+
+configs({ modes: ['md', 'ios', 'ionic-md'] }).forEach(({ config, screenshot, title }) => {
```
brandyscarney, line 4:
```suggestion
configs({ directions: ['ltr'], modes: ['md', 'ios', 'ionic-md'] }).forEach(({ config, screenshot, title }) => {
```

We should exclude `rtl` here since we aren't checking for direction

core/src/components/badge/badge.tsx
```diff
@@ -47,6 +47,14 @@ export class Badge implements ComponentInterface {
    */
   @Prop() size?: 'xxsmall' | 'xsmall' | 'small' | 'medium' | 'large' | 'xlarge';
 
+  /**
+   * Set to `"top-right"` to position the badge on top right absolute position of the parent element.
+   * Set to `"bottom-right"` to position the badge on bottom right absolute position of the parent element.
+   *
+   * Defaults to `"static"`.
+   */
+  @Prop() position: 'top-right' | 'bottom-right' | 'static' = 'static';
```
brandyscarney, line 56:
```suggestion
  @Prop() position?: 'top-right' | 'bottom-right';
```

Can we treat `undefined` as the static option?

brandyscarney, line 56: I'm not sure about the property name and values. Here are some of the other components for reference:

| Component             | Property     | Options                                  |
| ----------------------| -------------| -----------------------------------------|
| `ion-infinite-scroll` | `position`   | `"bottom"` \| `"top"`                    |
| `ion-label`           | `position`   | `"fixed"` \| `"floating"` \| `"stacked"` |
| `ion-toast`           | `position`   | `"bottom"` \| `"middle"` \| `"top"`      |
| `ion-fab`             | `horizontal` | `"center"` \| `"end"` \| `"start"`       |
| `ion-fab`             | `vertical`   | `"bottom"` \| `"center"` \| `"top"`      |

Do we plan to eventually support a `left` property? Maybe we should follow the approach used by `ion-fab` and offer `horizontal` and `vertical` options instead? This is how MUI handles it: https://mui.com/material-ui/react-badge/#badge-alignment

Also, we should use `end` instead of `right` if it flips in RTL.

core/src/components/tab-button/test/basic/tab-button.e2e.ts-snapshots/tab-button-badge-ios-ltr-Mobile-Firefox-linux.png
brandyscarney: What is causing this?

brandyscarney (changes requested):

core/src/components/badge/badge.tsx
BenOsodrac: Not planned, but I thought this would scale better if that changes in the future. But I agree with you with the start/end stuff, will change that. Maybe will make it bottom | top, to make it similar to other components.

I think for this horizontal and vertical would be too much just for two positions.

BenOsodrac: I think so, I added the static to create a selector to easily make CSS styles, in this case we just add position: absolute using :not(.position-static). But I can make a more complex :not without this, will make the change.

core/src/components/tab-button/test/basic/tab-button.e2e.ts-snapshots/tab-button-badge-ios-ltr-Mobile-Firefox-linux.png
BenOsodrac: I believe is the removal of the :empty styles with display: none. Not sure if that was a forgotten empty badge or placed to test precisely that. 

I was intending to fix it on the tab-bar related task, as it should look as expected when the PR is worked on!

What do you think?

core/src/components/badge/badge.tsx
christian-bromann: What would `static` stand for? Should this be described as well? Any reason we don't include `top-left` and `bottom-left`?

BenOsodrac: My original idea for static, was to stand for the actual css property used, position: static, as in no absolute position is being used.

But after @brandyscarney feedback I changed to just _top_ and _bottom._ As for other positions, according to UX team only the right/end side makes sense for them.

thetaPC: My vote is to follow what `fab` does with `horizontal` and `vertical`. It would stay consistent with what we have, leading for an easier introduction for the community.

BenOsodrac: But there's also _position_ in ion-infinite-scroll, ion-label and ion-toast. Personally, when I see horizontal or vertical props, I think about alignment, not absolute positioning 🤔

BenOsodrac: @brandyscarney @thetaPC @christian-bromann do you all vote for the ion-fab approach and instead have a `vertical` prop that accepts `top | bottom` ?

thetaPC: Yes for me

christian-bromann: +1

BenOsodrac: Made the change 👍

brandyscarney: Sure, it should be a pretty simple addition.

core/src/components/badge/badge.tsx
```diff
@@ -89,6 +95,7 @@ export class Badge implements ComponentInterface {
           [theme]: true,
           [`badge-${shape}`]: shape !== undefined,
           [`badge-${size}`]: size !== undefined,
+          [`badge-${this.vertical}`]: this.vertical !== undefined,
```
brandyscarney, line 98:
```suggestion
          [`badge-vertical-${this.vertical}`]: this.vertical !== undefined,
```

Update to match the [fab naming convention](https://github.com/ionic-team/ionic-framework/blob/cd5c27a12a03cc11da7ae86d89ad402343fed1c0/core/src/components/fab/fab.tsx#L84-L85) OR we can handle this as part of `ROU-11329` if you want to add a TODO here since shape & size are also different.

brandyscarney (approved):
Looks good! Just the one question I sent you offline & a suggestion for the naming of the class. 👍

core/src/components/badge/badge.tsx
BenOsodrac: I will do a commit, as it needs update on the scss as well :)

brandyscarney: Yeah I was too lazy to request changes on all of them 😝

core/src/components/badge/test/hint/index.html
```diff
@@ -0,0 +1,46 @@
...
+
+  <body>
+    <ion-app>
+      <ion-header>
+        <ion-toolbar>
+          <ion-title>Badge - Hint</ion-title>
+        </ion-toolbar>
+      </ion-header>
+
+      <ion-content id="content">
```
thetaPC, line 25: Since the ID is not being used, then it should be removed.

```suggestion
      <ion-content>
```

core/src/components/tab-button/test/basic/tab-button.e2e.ts-snapshots/tab-button-badge-ios-ltr-Mobile-Firefox-linux.png
thetaPC: Can we verify that the task mentions to fix it? That way it doesn't get forgotten. Also should include that `md` hint is rendering squished and needs to be fixed.

thetaPC (approved):
LGTM, minor suggestion but not a blocker

core/src/components/badge/badge.ionic.scss
```diff
@@ -145,3 +145,36 @@
...
+// Badge Empty
+// --------------------------------------------------
+
+:host(:empty) {
+  --padding-start: 0;
+  --padding-end: 0;
+}
+
+:host(:empty[vertical]) {
+  border: globals.$ion-border-size-025 globals.$ion-border-style-solid globals.$ion-bg-surface-inverse;
```
thetaPC, line 158: I don't see this border on badge when it's a status/hint. This should be removed to match design. This is also shown in the Avatar designs.

thetaPC (changes requested):

core/src/components/badge/test/hint/index.html
BenOsodrac: fixed!

core/src/components/badge/badge.ionic.scss
BenOsodrac: The border is present when the badge is used inside other components, thats why I added this only when the [vertical] attribute is being used, as it means that is not being used isolated. 

Its true that for now, on the scope of this PR, this wasn't probably needed but it will for all the following PR that add support for usage in other components (avatar, tab-bar, button). As this is a common style for badge, I thought it made sense to add it already. 

Let me know if I understood correctly or might be missing something :)

BenOsodrac: Double-checked with UX team and you're right. In fact, the border was removed from all use-cases. 

Commited that change.

Also change the CSS position rule to work not only when its empty, but just when the vertical prop is being used (also checked with UX).

thetaPC (approved):
LGTM

PR #30214, opened 2025-02-25
fix(select): scroll to the selected option before the overlay is visible

brandyscarney:


core/src/components/select/select.tsx
```diff
@@ -310,66 +310,89 @@ export class Select implements ComponentInterface {
...
+        overlay.addEventListener('ionAlertWillPresent', scrollAfterRender, { once: true });
+      } else if (this.interface === 'action-sheet') {
+        overlay.addEventListener('ionActionSheetWillPresent', scrollAfterRender, { once: true });
       }
     }
 
+    overlay.onDidDismiss().then(() => {
+      this.overlay = undefined;
+      this.isExpanded = false;
+      this.ionDismiss.emit();
```
BenOsodrac, line 391: Shouldn't we remove the listeners added above in some place?

BenOsodrac (approved):

core/src/components/select/select.tsx
brandyscarney: No, because they will be removed after their first execution due to [once](https://developer.mozilla.org/en-US/docs/Web/API/EventTarget/addEventListener#once) being set to `true`. Additionally, we clean up the `overlay` reference in the `onDidDismiss` handler.

thetaPC (approved):
LGTM, thank you!! Just a tiny suggestion but not a blocker.

core/src/components/select/select.tsx
```diff
@@ -310,66 +310,89 @@ export class Select implements ComponentInterface {
...
+    };
 
-        focusVisibleElement(firstEnabledOption.closest('ion-item')!);
+    // For modals and popovers, we can scroll before they're visible
+    if (this.interface === 'modal') {
+      overlay.addEventListener('ionModalWillPresent', scrollSelectedIntoView, { once: true });
+    } else if (this.interface === 'popover') {
+      overlay.addEventListener('ionPopoverWillPresent', scrollSelectedIntoView, { once: true });
+    } else {
+      // For alerts and action sheets, we need to wait a frame after willPresent
```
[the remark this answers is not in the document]
brandyscarney: Is this good? https://github.com/ionic-team/ionic-framework/pull/30214/commits/665f908e0fb8aa88a6ab42c6eebe17621a95018b

thetaPC (approved):
LGTM, thank you!

thetaPC (approved):
LGTM

PR #30217, opened 2025-02-26
feat(react): fixing support for react 19, adding test app for react 19

### LINKED ISSUE #29991: bug: Issues with React 19 types
### Prerequisites

### Ionic Framework Version

v8.x

### Current Behavior

I've been playing with React 19. I've noticed that Ionic has issues with some of the React 19 types. (This PR is proposed by core React team.)

https://github.com/DefinitelyTyped/DefinitelyTyped/pull/69022

Some components do work, like `IonButton`. But lots don't, like `IonBackButton` and `IonApp`:

```
'IonBackButton' cannot be used as a JSX component.
  Its type '{ new (props: Props | Readonly<Props>): { context: NavContextState; clickButton: (e: MouseEvent<Element, MouseEvent>) => void; ... 17 more ...; UNSAFE_componentWillUpdate?(nextProps: Readonly<...>, nextState: Readonly<...>, nextContext: any): void; }; new (props: Props, context: any): { ...; }; readonly displayName:...' is not a valid JSX element type.
```

### Expected Behavior

All components work with React 19.

Given that React 19 will be released imminently, I think it would be smart to get ahead of this :)

### Steps to Reproduce

1. Open https://stackblitz.com/edit/i9zdvi-adpda9
2. Observe type issues with 

### Code Reproduction URL

https://stackblitz.com/edit/i9zdvi-adpda9

### Ionic Info

Running stackblitz

### Additional Information

Good news: Ionic works great with 19 so far in my testing, just issues with types :)

ShaneK:
Issue number: resolves #29991

---------

Currently, there is no test app for React 19. Also, the output test in the react 18 test app is broken because it uses vitest but that file is trying to use jest still.

Additionally, there's issues with using Ionic Framework with React 19 due to several typescript changes. E.g. Issue #29991 

## What is the new behavior?

Now, there is a test app for React 19 and React 19 should work with Ionic Framework
- Additionally, fixed the output test for react 18 by moving to vitest
- Going forward, react test apps will default to the vitest test app

- [ ] Yes
- [X] No


packages/react/test/apps/react17/src/OutputTarget.test.tsx
thetaPC, line 1: Why was this file only added to React 17?

thetaPC (commented):
There are extra files that are not related to React 19. It seems that they're coming from the fact that `feature-8.5` is behind on `main`. I recommend syncing `feature-8.5` to make sure this PR only has the files related to 19.

Other than that, I have one other comment.

packages/react/test/apps/react17/src/OutputTarget.test.tsx
ShaneK: Previously, the react base was using jest even though 18 and 19 use vitest. 17 uses jest, though, so I left this file to override the test for 17 and updated base to use vitest for 18 and 19 and all reacts going forward, which will presumably use vitest.

thetaPC (approved):
LGTM

PR #30218, opened 2025-02-26
fix(focusVisibleElement): set focus on custom appRootSelector

thetaPC:
Issue number: internal

---------

The `focusVisibleElement` function does not take into account that a user might use a custom `appRootSelector`. If they do, the the `setFocus` function will throw an error that it does not exist leading to focus not being established on the elements.

This error can be seen when opening a `ion-select` with a modal interface.

Screenshot 2025-02-26 at 1 24 18 PM

## What is the new behavior?

- Created a new function to reuse the focus logic within `ion-app` and for the custom app root selector. This will make sure that there will be consistency between both if that logic needs to change in the future.

- [ ] Yes
- [x] No

### DIFF
--- a/core/src/components/app/app.tsx
+++ b/core/src/components/app/app.tsx
@@ -17,7 +17,13 @@ export class App implements ComponentInterface {
   @Element() el!: HTMLElement;
 
   /**
-   * Sets focus on elements that use `ion-focusable`.
+   * Used to set focus on an element that uses `ion-focusable`.
+   * Do not use this if focusing the element as a result of a keyboard
+   * event as the focus utility should handle this for us. This method
+   * should be used when we want to programmatically focus an element as
+   * a result of another user action. (Ex: We focus the first element
+   * inside of a popover when the user presents it, but the popover is not always
+   * presented as a result of keyboard action.)
    *
    * @param elements - The elements to set focus on.
    */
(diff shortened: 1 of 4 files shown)


core/src/components/app/app.tsx
```diff
@@ -17,18 +17,17 @@ export class App implements ComponentInterface {
   @Element() el!: HTMLElement;
 
   /**
-   * Used to set focus on an element that uses `ion-focusable`.
-   * Do not use this if focusing the element as a result of a keyboard
-   * event as the focus utility should handle this for us. This method
-   * should be used when we want to programmatically focus an element as
-   * a result of another user action. (Ex: We focus the first element
-   * inside of a popover when the user presents it, but the popover is not always
-   * presented as a result of keyboard action.)
```
brandyscarney, line 26: Why did this comment get removed here? This is what ends up in the documentation here: https://ionicframework.com/docs/api/app#setfocus

brandyscarney (approved):
Code changes look good, one suggestion on the comment. 👍

core/src/components/app/app.tsx
thetaPC: It made sense to me to remove it since `focusElements` has it but I'm not committed to it. I added it back: https://github.com/ionic-team/ionic-framework/pull/30218/commits/2edbdaffdcc5f80c162a52b0687a2861acf9ba6c

PR #30195, opened 2025-02-17
fix(capacitor): replace deprecated platform check method

r-yanyo:
Issue number: resolves internal

ref: https://github.com/ionic-team/capacitor/issues/7884

---------

## What is the current behavior?
`this.platform.is('capacitor')` returns `false` in Capacitor App.

## What is the new behavior?
`this.platform.is('capacitor')` returns as expected.

## Does this introduce a breaking change?

- [ ] Yes
- [x] No

## Other information
It might also need to be fixed.

https://github.com/ionic-team/ionic-framework/blob/14b6538d98303cb753d881ec6978fb98f53ed54c/core/src/utils/test/platform.utils.ts#L32

PR #30228, opened 2025-03-03
fix(capacitor): use proper types for capacitor v7 support

ShaneK:

---------

Currently, Capacitor types are outdated in Ionic Framework and we're accessing a type property that no longer exists in Capacitor 7.0.0+

## What is the new behavior?

This PR updates the capacitor version and addresses removal of `.Plugins` from `@capacitor/core`'s `CapacitorGlobal`, which we rely on to dynamically access plugins that the user may or may not have installed.

The fix for this was creating a custom type definition to support accessing `Plugins`. While `Plugins` was removed from Capacitor if we were accessing it directly from core, we're pulling it from the window in the browser, where it's still exposed, so we just needed to make our type reflect that.

- [ ] Yes
- [X] No

## Other information

Technically, this issue does not prevent Framework from working with Capacitor 7 because it's only a typing issue, but it's still a minor issue that should be addressed in our effort to support Capacitor 7. This PR, along with #30195, should make it ready for that.


brandyscarney (approved):

thetaPC (approved):
LGTM

PR #30245, opened 2025-03-13
feat(searchbar): update according to design

OS-pedrolourenco:
Issue number: internal

---------

The searchbar component does not currently match our UX designs for the ionic theme.

## What is the new behavior?
Updated the searchbar so that it does:
- Adjusted the searchbar input paddings when no leading or trailing icons are rendered;
- Added an ellipsis text overflow to the searchbar input;
- Updated the background, text and icon colors so that they are reactive to state changes like having inserted text and being disabled (some design tokens are still TBD);
- The clear button is now hidden when the searchbar is disabled.

- [ ] Yes
- [x] No


core/src/components/searchbar/searchbar.ionic.scss
```diff
@@ -9,16 +9,18 @@
    * @prop --focus-ring-color: The color of the ring around the focused element.
    * @prop --focus-ring-width: The width of the ring around the focused element.
    */
-  --background: #{globals.$ion-primitives-neutral-100};
+  --background: #{globals.$ion-bg-neutral-subtle-default};
```
thetaPC, line 12: Design shows this to be `bg/input/bold/default`. Were you able to with the design team of which token to use?

core/src/components/searchbar/searchbar.ionic.scss
```diff
@@ -9,16 +9,18 @@
...
   --clear-button-color: #{globals.$ion-primitives-neutral-1000};
   --color: #{globals.$ion-primitives-neutral-1200};
-  --icon-color: #{globals.$ion-primitives-neutral-800};
-  --placeholder-color: #{globals.$ion-primitives-neutral-800};
+  --icon-color: #{globals.$ion-icon-default};
+  --placeholder-color: #{globals.$ion-text-subtlest};
   --focus-ring-color: #{globals.$ion-border-focus-default};
   --focus-ring-width: #{globals.$ion-border-size-050};
+  --leading-icon-padding: #{calc(globals.$ion-scale-400 + globals.$ion-space-200)};
+  --clear-button-padding: #{calc(globals.$ion-scale-400 + globals.$ion-space-200)};
```
thetaPC, line 23: Why were these new variables added? We need to be careful when adding new ones because it introduces them to the public so we try to make sure that they're necessary.

If they're to make it easier to keep track of then it might be best to make Sass variables.

core/src/components/searchbar/searchbar.ionic.scss
```diff
@@ -38,6 +40,8 @@
 .searchbar-input {
   @include globals.padding(globals.$ion-space-300, null);
 
+  text-overflow: ellipsis;
```
thetaPC, line 43: Let's make a test to prevent a regression on this. I would place it within basic and only for `ionic-md` and `ltr` since functionality doesn't change.

core/src/components/searchbar/searchbar.ionic.scss
```diff
@@ -92,6 +100,16 @@
...
 
+// Searchbar Search Icon & Cancel Icon
+// -----------------------------------------
+:host(:not(.searchbar-has-leading-icons)) {
+  --leading-icon-padding: #{globals.$ion-scale-0};
+}
+
+:host(:not(.searchbar-has-value)) {
+  --icon-color: #{globals.$ion-icon-subtlest};
+}
```
thetaPC, line 111: As one of our standards, we try to not use `:not` when possible. I would recommend having the default under `:host` as `--icon-color: #{globals.$ion-icon-subtlest};` as I mentioned in my earlier comment.

And then update this to be:
```suggestion
:host(.searchbar-has-value) {
  --icon-color: #{globals.$ion-icon-default};
}
```

thetaPC (changes requested):
Also the PR title should use `refactor` since we are updating the styles, not adding them in.

core/src/components/searchbar/searchbar.tsx
```diff
@@ -816,25 +822,39 @@ export class Searchbar implements ComponentInterface {
...
-            <ion-icon aria-hidden="true" icon={searchbarClearIcon} lazy={false} class="searchbar-clear-icon"></ion-icon>
-          </button>
+          {theme !== 'ionic' || this.searchIcon !== undefined ? (
+            <ion-icon
+              aria-hidden="true"
+              icon={searchbarSearchIcon}
+              lazy={false}
+              class="searchbar-search-icon"
+            ></ion-icon>
+          ) : null}
```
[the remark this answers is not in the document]
OS-pedrolourenco: What I was doing with this code is adding the ability to not render a searchIcon. The default icon is still the magnifying glass but that is defined on the OutSystems side. On the other hand, it might be a bit cleaner to just had a new prop called `showSearchIcon`. What do you think?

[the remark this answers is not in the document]
thetaPC: I would leave it how it was originally. If OS is going to set it on their end, then it wouldn't cause any issues to leave in the default.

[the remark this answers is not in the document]
OS-pedrolourenco: Hmm but the original always rendered a searchIcon. I want to add the ability to not have one. Either this way where I was checking for an undefined searchIcon prop or by creating a new prop to determine whether a searchIcon should be rendered.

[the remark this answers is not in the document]
thetaPC: Discussed offline and decided with this concept:
```
@Prop() searchIcon?: string | boolean;

{this.searchIcon !== false ? (
  <ion-icon
    aria-hidden="true"
    icon={searchbarSearchIcon}
    lazy={false}
    class="searchbar-search-icon"
  ></ion-icon>
) : null}
```

core/src/components/searchbar/searchbar.ionic.scss
OS-pedrolourenco: I think the change I did in `searchbar/test/basic/searchbar.e2e.ts` should cover this no?

core/src/components/searchbar/searchbar.ionic.scss
OS-pedrolourenco: These variables were created with the purpose of making the searchbar's start and end paddings be dynamic according to whether it is rendering leading and/or trailing icons. That being said, I don't see a reason why they have to be CSS variables instead of SCSS variables so I'll make that change 👍🏼

core/src/components/searchbar/searchbar.ionic.scss
thetaPC: Good catch! Yes, please disregard this request. Thanks!

core/src/components/searchbar/searchbar.tsx
```diff
@@ -620,12 +629,19 @@ export class Searchbar implements ComponentInterface {
    * Clear button should be shown if one of two conditions applies:
    * 1. `showClearButton` is set to `always`.
    * 2. `showClearButton` is set to `focus`, and the searchbar has been focused.
+   * 3. `theme` is not set `ionic`, or the searchbar has not been disabled.
```
thetaPC, line 632:
```suggestion
   * Unless the `theme` is `ionic` and the searchbar is disabled.
```
Thinking more about this, this suggestion sounds easier to understand.

core/src/components/searchbar/searchbar.ionic.scss
```diff
@@ -237,19 +354,47 @@
...
    * the size of the leading icon (search or cancel),
    * and the gap between the icon and the input.
+   *
+   * Padding end is based on
+   * desired padding from design,
+   * the size of the trailing icon (clear),
+   * and the gap between the icon and the input.
    */
-  padding-inline-start: calc(9px + globals.$ion-scale-500 + globals.$ion-space-200);
+  @include globals.padding-horizontal(globals.$ion-space-500, globals.$ion-space-500);
```
thetaPC, line 371:
```suggestion
  @include globals.padding-horizontal(globals.$ion-space-500);
```

core/src/components/searchbar/searchbar.tsx
```diff
@@ -783,10 +800,12 @@ export class Searchbar implements ComponentInterface {
           'searchbar-has-value': this.hasValue(),
           'searchbar-left-aligned': this.shouldAlignLeft,
           'searchbar-has-focus': this.focused,
+          'searchbar-should-show-search-icon': this.shouldShowSearchIcon(),
           'searchbar-should-show-clear': this.shouldShowClearButton(),
           'searchbar-should-show-cancel': this.shouldShowCancelButton(),
           [`searchbar-shape-${shape}`]: shape !== undefined,
           [`searchbar-size-${size}`]: size !== undefined,
+          'in-toolbar': hostContext('ion-toolbar', this.el),
```
thetaPC, line 808: Why was this added? We already were applying the styles properly without this class `:host-context(ion-toolbar)`. If possible, let's revert this back to only use the styles.

thetaPC (changes requested):

core/src/components/searchbar/searchbar.tsx
OS-pedrolourenco: This was added in order to get higher specificity on the toolbar specific padding styles so that they could prevail over the other padding styles we introduced.

PR #30235, opened 2025-03-07
refactor(modal): allow dragging when expandToScroll is false

thetaPC:
Issue number: internal

---------

## What is the current behavior?

The community has [pointed](https://github.com/ionic-team/ionic-framework/pull/30097#issuecomment-2660473695) out a missing aspect to the new feature of being able to scroll content on all breakpoints for the sheet modal.  It's been verified (by comparing native iOS) that the swipe to close or minimize on content (not handle) is indeed missing but only on a certain conditions:

User can swipe on the content to close when:
- The sheet modal is on the smallest breakpoint
- The content has been scrolled to the top

User can swipe down to collapse to the next smallest breakpoint
- The content has been scrolled to the top

## What is the new behavior?

User can swipe on the content to close when:
- The sheet modal is on the smallest breakpoint
- The content has been scrolled to the top

User can swipe down to collapse to the next smallest breakpoint
- The content has been scrolled to the top

## Does this introduce a breaking change?

- [ ] Yes
- [x] No

## Other information

[Preview](https://ionic-framework-git-fw-6283-ionic1.vercel.app/src/components/modal/test/sheet)

PR #30257, opened 2025-03-15
fix(modal): consider scrollable content while dragging when expandToScroll is false

kumibrr:
Issue number: **no error submited yet**

---------
Changes introduced by #30235 caused two major issues:
- Animations were not being played when increasing breakpoints.
- When the scrollable content was at the top, and a big scroll down the list was made, the modal would jump to another breakpoint. 

## What is the new behavior?
- When `expandToScroll` is false, the swipe gesture is allowed unless it's a pull down within scrollable content.
- When `expandToScroll` is false, we cancel the `moveSheetToBreakpoint` when a scroll to top is being done within the scrollable content

- [ ] Yes
- [ x] No

## Other information
| Before | After |
|--------|-------|
| https://github.com/user-attachments/assets/e1c22f48-f990-45cf-a6c4-1aec0d019c6d     |  https://github.com/user-attachments/assets/f01e0480-748f-40af-ac11-94f790f0e197     |

### DIFF
--- a/core/src/components/modal/gestures/sheet.ts
+++ b/core/src/components/modal/gestures/sheet.ts
@@ -265,10 +265,14 @@ export const createSheetGesture = (
   const onMove = (detail: GestureDetail) => {
     /**
      * If `expandToScroll` is disabled, we should not allow the swipe gesture
-     * to continue if the gesture is not pulling down.
+     * to continue if the gesture is not pulling down within scrollable content.
      */
     if (!expandToScroll && detail.deltaY <= 0) {
-      return;
+      const contentEl = findClosestIonContent(detail.event.target! as HTMLElement)
+      const scrollEl = contentEl && isIonContent(contentEl) ? getElementRoot(contentEl).querySelector('.inner-scroll') : contentEl;
+      if (scrollEl) {
+        return;
+      }
     }
 
     /**
--- a/core/src/components/modal/gestures/sheet.ts
+++ b/core/src/components/modal/gestures/sheet.ts
@@ -340,6 +340,19 @@ export const createSheetGesture = (
       return Math.abs(b - diff) < Math.abs(a - diff) ? b : a;
     });
 
+    /**
+     * If expandToScroll is disabled, we should not allow the moveSheetToBreakpoint
+     * function to be called if the user is trying to swipe upwards and the content
+     * is not scrolled to the top.
+     */
+    if (!expandToScroll && detail.deltaY <= 0 && findClosestIonContent(detail.event.target! as HTMLElement)) {
+      const contentEl = findClosestIonContent(detail.event.target! as HTMLElement)!;
+      const scrollEl = isIonContent(contentEl) ? getElementRoot(contentEl).querySelector('.inner-scroll') : contentEl;
+      if (scrollEl!.scrollTop > 0) {
+        return;
+      }
+    }
+
     moveSheetToBreakpoint({
       breakpoint: closest,
       breakpointOffset: offset,


brandyscarney:
Thank you for the PR! We have this on our list to review soon. Please ignore the failing screenshot tests. We are looking into some issues with a recent Playwright update.

ShaneK:
@kumibrr if you could pull main into your branch and push again, the workflow should work this time!

ShaneK:
This has been merged and will deploy with the next patch version, thank you for your PR!

kumibrr:
@ShaneK Thank you guys for moving this so quickly and thank you for the feedback!

❤️

aeharding:
Hi! I am concerned about the logic in onMove causing performance regressions for existing Ionic apps. onMove is caused a lot and onMove was designed to be optimized and not call the DOM. This is a high touchpoint part of the code for every Ionic app that uses swipe.

Is it possible to move DOM query logic outside of onMove? Or could we please be able to disable this logic for performance optimized apps? Thanks.

kumibrr:
> Hi! I am concerned about the logic in onMove causing performance regressions for existing Ionic apps. onMove is caused a lot and onMove was designed to be optimized and not call the DOM. This is a high touchpoint part of the code for every Ionic app that uses swipe (edit: with expandToScroll).
> 
> Is it possible to move DOM query logic outside of onMove? Or could we please be able to disable this logic for performance optimized apps? Thanks.

Hi! Thank you for sharing your thoughts, I had the same while making this PR, I'll add the section of the code that worries you and will comment on it on why it should not be problematic.
```js
const onMove = (detail: GestureDetail) => {
     if (!expandToScroll && detail.deltaY <= 0) {
       // If expandToScroll is true, it won't check anything else, I'll just skip all the code bellow and no DOM queries will be executed.
       // If expandToScroll is false, it will check if the deltaY is less or equal to zero. If it's not, same as before, it'll skip all the code bellow and no DOM queries will be executed.

       // the DOM query will only fire if the followings conditions are true: expandToScroll is false, deltaY is zero or negative
       const contentEl = findClosestIonContent(detail.event.target! as HTMLElement);
       // DOM query count is now 1.
       const scrollEl =
         contentEl && 
         // If contentEl is null, it'll skip the code bellow. and one DOM Query count would've been executed.
         isIonContent(contentEl) 
         // If contentEl is not an ionContent, it'll skip the next line.
         ? getElementRoot(contentEl).querySelector('.inner-scroll')
         // If contentEl is an ionContent, it'll query it's .inner-scroll
         // DOM query count is now 2.
         : contentEl;
       if (scrollEl) {
         return;
       }
     }
```

I'll check later if further improvements can be made without compromising on stability within the component.
Do you @aeharding have any suggestions?

kumibrr:
Alright, I've refactored the check to only do a contentEl.contains and seems to work fine. Does anyone have any thoughts on the possible downsides of this implementation from the current?

```js
  const onMove = (detail: GestureDetail) => {
    /**
     * If `expandToScroll` is disabled, and an upwards swipe gesture is done within
     * the scrollable content, we should not allow the swipe gesture to continue.
     */
    if (!expandToScroll && detail.deltaY <= 0 && contentEl?.contains(detail.event.target as HTMLElement)) {
      return;
    }
```
```js
  const onEnd = (detail: GestureDetail) => {
    /**
     * If expandToScroll is disabled, we should not allow the moveSheetToBreakpoint
     * function to be called if the user is trying to swipe content upwards and the content
     * is not scrolled to the top.
     */
    if (!expandToScroll && detail.deltaY <= 0 && contentEl?.contains(detail.event.target as HTMLElement)) {
      return;
    }
```

ShaneK:
I'm considering a potential alternative solution where we cache the targetted element and if it's a scrollable in `onStart` and reference the cached value in `onMove` and `onEnd`. I think this might be a solid compromise with only a minor potential issue where the drag target changes without triggering `onStart`, but I need testing to see if that will be a problem.

ShaneK:
@kumibrr, @aeharding: after testing, I believe the caching solution should work great. I have a PR open here and I hope it'll be merged before the end of the day #30267

ShaneK:
BTW, these changes and the related regression fix got deployed with 8.5.1 yesterday 🚀

PR #30261, opened 2025-03-17
fix(input): remove z-index target-area for ionic theme

OS-martacarlos:
Issue number: internal
---------

Currently,  operations associated with gestures aren’t working, for the ionic theme:
- can’t select the text using double tap
- can’t longpress
- can’t move the cursor to any positions

## What is the new behavior?
- It's now possible to select text using double tap, long press and move the cursor to different positions
- Adds new tests for double clicks

| Before | Now |
|--------|-----|
|prev |  now |

- [ ] Yes
- [x] No

## Notes
Currently, the double click text selection is not working (the text isn't selected) on the automated tests for:
- ionic theme, on firefox and safari
- md/ios themes, on firefox

as manual tests do not have this problem, we'll be skipping these tests for now and come back to them at a later date


core/src/components/input/test/basic/input.e2e.ts
```diff
@@ -63,6 +62,45 @@ configs().forEach(({ title, screenshot, config }) => {
...
+            label="Label"
+            label-placement="stacked"
+            clear-input="true"
+            value="Text"
+          ></ion-input>
+        `,
+          config
+        );
+        const input = page.locator('ion-input');
+        input.dblclick();
```
thetaPC, line 99: I think the reason that the native input should be used to double click on is because we verify that the mouse actually is on top of the input. If we use just `ion-input` then the Playwright mouse might be clicking on the right area to trigger a focus when using specific browsers.

Also it helps to add the missing `await` to `dbclick`.

```suggestion
        const input = page.locator('ion-input');
        const nativeInput = input.locator('input');
        await nativeInput.dblclick();
```

core/src/components/input/test/basic/input.e2e.ts
```diff
@@ -185,6 +223,28 @@ configs({ modes: ['ionic-md'] }).forEach(({ title, screenshot, config }) => {
...
+            label="Label"
+            label-placement="stacked"
+            clear-input="true"
+            value="Text"
+          ></ion-input>
+      `,
+        config
+      );
+      const input = page.locator('ion-input');
+      await input.dblclick();
```
thetaPC, line 244: I think the reason that the native input should be used to double click on is because we verify that the mouse actually is on top of the input. If we use just `ion-input` then the Playwright mouse might be clicking on the right area to trigger a focus when using specific browsers.

```suggestion
      const input = page.locator('ion-input');
      const nativeInput = input.locator('input');
      await nativeInput.dblclick();
```

core/src/components/input/test/basic/input.e2e.ts
```diff
@@ -185,6 +223,28 @@ configs({ modes: ['ionic-md'] }).forEach(({ title, screenshot, config }) => {
...
+      await page.waitForChanges();
```
thetaPC, line 245: This isn't needed because we have the `await` on `dbclick`.

```suggestion
```

core/src/components/input/test/basic/input.e2e.ts
```diff
@@ -185,6 +223,28 @@ configs({ modes: ['ionic-md'] }).forEach(({ title, screenshot, config }) => {
       const container = page.locator('#container');
       await expect(container).toHaveScreenshot(screenshot(`input-clear-button-focused`));
     });
+
+    test('should not have visual regressions when user dbclicks', async ({ page, browserName }) => {
+      if (browserName === 'firefox' || browserName === 'webkit') {
+        //TODO: Figure out why it's not working for webkit & firefox
+        test.skip();
+      }
```
thetaPC, line 231: Once I used the native input to click on then the Firefox and Safari tests were passing as expected.

```suggestion
```

core/src/components/input/test/basic/input.e2e.ts
```diff
@@ -63,6 +62,45 @@ configs().forEach(({ title, screenshot, config }) => {
...
+        const input = page.locator('ion-input');
+        input.dblclick();
+        // Validates the display of an input with a clear button.
+        await expect(input).toHaveScreenshot(screenshot(`input-dbclick`));
+      });
+      test('should not have visual regressions with stacked label', async ({ page, browserName }) => {
+        if (browserName === 'firefox') {
+          //TODO: Figure out why it's not working for firefox
+          test.skip();
+        }
```
thetaPC, line 86: Once I used the native input to click on then the Firefox tests were passing as expected.

```suggestion
```

core/src/components/input/test/basic/input.e2e.ts
```diff
@@ -185,6 +223,28 @@ configs({ modes: ['ionic-md'] }).forEach(({ title, screenshot, config }) => {
       const container = page.locator('#container');
       await expect(container).toHaveScreenshot(screenshot(`input-clear-button-focused`));
     });
+
+    test('should not have visual regressions when user dbclicks', async ({ page, browserName }) => {
```
thetaPC, line 227:
```suggestion
    test('should not have visual regressions when user double clicks', async ({ page, browserName }) => {
```

core/src/components/input/test/basic/input.e2e.ts
```diff
@@ -63,6 +62,45 @@ configs().forEach(({ title, screenshot, config }) => {
...
+          ></ion-input>
+        `,
+          config
+        );
+        const input = page.locator('ion-input');
+        input.dblclick();
+        // Validates the display of an input with a clear button.
+        await expect(input).toHaveScreenshot(screenshot(`input-dbclick`));
+      });
+      test('should not have visual regressions with stacked label', async ({ page, browserName }) => {
```
thetaPC, line 82:
```suggestion
      test('should not have visual regressions with stacked label when user double clicks', async ({ page, browserName }) => {
```

core/src/components/input/test/basic/input.e2e.ts
```diff
@@ -63,6 +62,45 @@ configs().forEach(({ title, screenshot, config }) => {
         await expect(input).toHaveScreenshot(screenshot(`input-with-clear-button-stacked`));
       });
     });
+    test.describe('input click behaviors', () => {
+      test('should not have visual regressions', async ({ page }) => {
```
thetaPC, line 66:
```suggestion
      test('should not have visual regressions when user double clicks', async ({ page }) => {
```

core/src/components/input/test/basic/input.e2e.ts
```diff
@@ -63,6 +62,45 @@ configs().forEach(({ title, screenshot, config }) => {
...
+          <ion-input
+            label="Label"
+            clear-input="true"
+            value="Text"
+          ></ion-input>
+        `,
+          config
+        );
+        const input = page.locator('ion-input');
+        input.dblclick();
```
thetaPC, line 78: I think the reason that the native input should be used to double click on is because we verify that the mouse actually is on top of the input. If we use just `ion-input` then the Playwright mouse might be clicking on the right area to trigger a focus when using specific browsers.

Also it helps to add the missing `await` to `dbclick`.

```suggestion
        const input = page.locator('ion-input');
        const nativeInput = input.locator('input');
        await nativeInput.dblclick();
```

thetaPC (changes requested):

core/src/components/input/input.ionic.scss
```diff
@@ -65,8 +65,6 @@
 
   // Cursor should match the native input when hovering over the target area.
   cursor: text;
-
-  z-index: 1;
```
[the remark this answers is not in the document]
OS-martacarlos: No, it's still a question with no answer

core/src/components/input/test/basic/input.e2e.ts
OS-martacarlos: aah I thought we wanted to test it specifically with the ion-input but you're right, targeting the input itself makes more sense

thetaPC: Sometimes we have to do stuff like this because Playwright ends up clicking the wrong area and it's a bit hard to tell where it's clicking on. By using the native input, we are removing any doubt of its location.

thetaPC (approved):
LGTM

core/src/components/input/input.ionic.scss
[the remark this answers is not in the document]
thetaPC: We'll just have to keep this in mind if we run into issues later. Have we verified if textarea is behaving correctly? It follows the same structure as input that it might also lead to the same issue.

BenOsodrac (approved):

core/src/components/input/input.ionic.scss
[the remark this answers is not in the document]
OS-martacarlos: I've done a simple / quick test on the textarea and haven't found any issue so far

PR #30262, opened 2025-03-17
feat(datetime): add showAdjacentDays to display days from the previous and next months

JoaoFerreira-FrontEnd:
Issue number: Internal

## What is the new behavior?

Adds a new property to datetime, showAdjacentDays, that when true will show the last days of the previous month and the first days of the next month. This will just occupy empty "cells" at the beginning of the month "table" and add rows to the table until a maximum of 6 rows are displayed.

## Changes
- add styles for adjacent day button
- add `showAdjacentDays` property to datetime component
- change month generation to respect new property
- add visual tests to new feature

- [ ] Yes
- [x] No

## Other information

[Preview](https://ionic-framework-git-rou-11118v2-ionic1.vercel.app/src/components/datetime/test/show-adjacent-days)


core/package-lock.json
brandyscarney, line 1: Can we revert this file since you didn't change any dependencies?

brandyscarney (commented):
Looks good just requested a couple things!

brandyscarney (approved):

core/src/components/datetime/utils/data.ts
```diff
@@ -102,8 +102,16 @@ export const getDaysOfWeek = (locale: string, mode: Mode, firstDayOfWeek = 0) =>
  * the firstDayOfWeek value (Sunday by default)
  * using null values.
  */
-export const getDaysOfMonth = (month: number, year: number, firstDayOfWeek: number) => {
+export const getDaysOfMonth = (month: number, year: number, firstDayOfWeek: number, showAdjacentDays = false) => {
   const numDays = getNumDaysInMonth(month, year);
+  let previousNumDays: number; //previous month number of days
+  if (month === 1) {
+    //If january the previous month should be january and the last year
```
thetaPC, line 109:
```suggestion
    // If the current month is January, the previous month should be December of the previous year.
```

core/src/components/datetime/utils/data.ts
```diff
@@ -102,8 +102,16 @@ export const getDaysOfWeek = (locale: string, mode: Mode, firstDayOfWeek = 0) =>
...
+    previousNumDays = getNumDaysInMonth(12, year - 1);
+  } else {
+    //If not the previous month should be month -1 and the current year
```
thetaPC, line 112:
```suggestion
    // Otherwise, the previous month should be the current month - 1 of the same year.
```

thetaPC (changes requested):

core/src/components/datetime/datetime-interface.ts
```diff
@@ -15,6 +15,7 @@ export interface DatetimeParts {
   hour?: number;
   minute?: number;
   ampm?: 'am' | 'pm';
+  adjacentDay?: boolean;
```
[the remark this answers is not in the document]
gnbm: Maybe something more like `showAdjacentDays`? Even though we already have it on another place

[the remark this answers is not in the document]
thetaPC: We should try to minimize duplicate variable names to avoid confusion. Plus I believe this variable has a different behavior. `showAdjacentDays` is to show days from prev and next months. While `adjacentDay` is to determine if the current day is an adjacent day instead of the current month.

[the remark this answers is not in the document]
brandyscarney: I think this is on the individual days so it would be like whether or not that current day is an adjacent day or not.

core/src/components/datetime/datetime.tsx
```diff
@@ -2340,29 +2380,37 @@ export class Datetime implements ComponentInterface {
...
+                    }
+
                     this.setWorkingParts({
                       ...this.workingParts,
-                      month,
+                      month: _month,
                       day,
-                      year,
+                      year: _year,
+                      isAdjacentDay: isAdjacentDay,
```
thetaPC, line 2393:
```suggestion
                      isAdjacentDay,
```

core/src/components/datetime/datetime.tsx
```diff
@@ -2340,29 +2380,37 @@ export class Datetime implements ComponentInterface {
...
                     // multiple only needs date info, so we can wipe out other fields like time
                     if (multiple) {
                       this.setActiveParts(
                         {
-                          month,
+                          month: _month,
                           day,
-                          year,
+                          year: _year,
+                          isAdjacentDay: isAdjacentDay,
```
thetaPC, line 2403:
```suggestion
                          isAdjacentDay,
```

core/src/components/datetime/datetime.tsx
```diff
@@ -2340,29 +2380,37 @@ export class Datetime implements ComponentInterface {
...
                       );
                     } else {
                       this.setActiveParts({
                         ...activePart,
-                        month,
+                        month: _month,
                         day,
-                        year,
+                        year: _year,
+                        isAdjacentDay: isAdjacentDay,
```
thetaPC, line 2413:
```suggestion
                        isAdjacentDay,
```

thetaPC (changes requested):

PR #30263, opened 2025-03-17
fix(snapshots): use default value for cmd line argument

thetaPC:
Issue number: resolves #

---------

## What is the new behavior?

-
-
-

- [ ] Yes
- [ ] No


brandyscarney (approved):
Confirmed that this correctly adds or updates screenshots when a component path is passed for:
- A new e2e test
- A new e2e test with updates made after creating the screenshots
- An existing e2e test with **significant** changes to the component (e.g. changing background color from blue to red)

However, this does not work for:
- An existing e2e test with **minor** changes to the component (e.g. adjusting a stepped color to the next closest shade):
  - I changed the input helper text color from `$text-color-step-300` to `$text-color-step-400`

The only way to detect this type of change would be by using `all`, but doing so would re-generate screenshots even when no actual visual differences exist for those components. Given this, I'm okay with moving forward with this fix as is. 👍

PR #30267, opened 2025-03-18
perf(sheet): fixing performance regression on modal sheets when expandToScroll is false

ShaneK:
Issue number: internal

---------

Currently, when a sheet is moved while `expandToScroll` is disabled, the DOM is queried excessively causing performance degradation.

## What is the new behavior?

We now cache the targeted element in `onStart` and refer to it in `onMove` and `onEnd`, preventing over-querying the DOM

- [ ] Yes
- [X] No

## Other information

This regression was introduced in #30257 and quickly highlighted by a member of the community

### DIFF
--- a/core/src/components/modal/gestures/sheet.ts
+++ b/core/src/components/modal/gestures/sheet.ts
@@ -1,6 +1,6 @@
-import { isIonContent, findClosestIonContent } from '@utils/content';
+import { findClosestIonContent, isIonContent } from '@utils/content';
 import { createGesture } from '@utils/gesture';
-import { clamp, raf, getElementRoot } from '@utils/helpers';
+import { clamp, getElementRoot, raf } from '@utils/helpers';
 import { FOCUS_TRAP_DISABLE_CLASS } from '@utils/overlays';
 
 import type { Animation } from '../../../interface';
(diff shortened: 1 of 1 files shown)


core/src/components/modal/gestures/sheet.ts
```diff
@@ -233,6 +234,15 @@ export const createSheetGesture = (
      */
     canDismissBlocksGesture = baseEl.canDismiss !== undefined && baseEl.canDismiss !== true && minBreakpoint === 0;
 
+    /**
+     * Cache the scroll element reference when the gesture starts,
+     * this allows us to avoid querying the DOM for the target in onMove,
+     * which would impact performance significantly.
+     */
+    const targetEl = findClosestIonContent(detail.event.target! as HTMLElement);
```
aeharding, line 242: Maybe put this logic in an `if (!expandToScroll)`? To skip during gesture start for most scenarios that don't use it.

ShaneK: Excellent catch, done!

brandyscarney:
I created a dev build for device testing: `8.5.1-dev.11742393301.11cf5904`

brandyscarney (approved):

thetaPC (approved):
LGTM, tested on iOS as well

PR #30270, opened 2025-03-19
feat(divider): add new ion-divider component

OS-martacarlos:
Issue number: internal
---------

## What is the new behavior?
- Introduces a new component, `ion-divider` which has two props, `spacing` and `inset`. 
- Adds e2e tests to both `spacing` and `inset` props

- [ ] Yes
- [x] No

## Other information
- As discussed with the PO, using this new component inside an `ion-item` is out of scope , for now.


core/src/components/divider/divider.tsx
```diff
@@ -0,0 +1,58 @@
...
+export class Divider implements ComponentInterface {
+  /**
+   * Set to `"xxsmall"` for the smallest spacing.
+   * Set to "xsmall" for a very small spacing.
+   * Set to `"small"` for a small spacing.
+   * Set to "medium" for a medium spacing.
+   * Set to "large" for a large spacing.
+   * Set to `"xlarge"` for the largest spacing.
+   *
+   * Defaults to `"small"` for the `ionic` theme, undefined for all other themes.
```
brandyscarney, line 23:
```suggestion
   * Defaults to `"xxsmall"`.
```

The Figma design shows default as xxsmall, is this right?

core/src/components/divider/divider.tsx
```diff
@@ -0,0 +1,58 @@
...
+  shadow: true,
+})
+export class Divider implements ComponentInterface {
+  /**
+   * Set to `"xxsmall"` for the smallest spacing.
+   * Set to "xsmall" for a very small spacing.
+   * Set to `"small"` for a small spacing.
+   * Set to "medium" for a medium spacing.
+   * Set to "large" for a large spacing.
+   * Set to `"xlarge"` for the largest spacing.
```
brandyscarney, line 21:
```suggestion
   * Set to `"xxsmall"` for the smallest spacing.
   * Set to "xsmall" for very small spacing.
   * Set to `"small"` for small spacing.
   * Set to "medium" for medium spacing.
   * Set to "large" for large spacing.
   * Set to `"xlarge"` for the largest spacing.
```

core/src/components/divider/divider.scss
```diff
@@ -0,0 +1,78 @@
+@use "../../themes/ionic/ionic.globals.scss" as globals;
+@import "../../themes/native/native.globals";
```
brandyscarney, line 2: Is this file used?

core/src/components/divider/test/basic/divider.e2e.ts
```diff
@@ -0,0 +1,14 @@
+import { expect } from '@playwright/test';
+import { configs, test } from '@utils/test/playwright';
+
+configs().forEach(({ config, screenshot, title }) => {
+  test.describe(title('divider: rendering'), () => {
+    test('should not have visual regressions', async ({ page }) => {
+      await page.goto('/src/components/divider/test/basic', config);
```
brandyscarney, line 7: Instead of taking a screenshot of the entire page, we should use `setContent` to take screenshots of just the divider in different scenarios. You can find an example here: https://github.com/ionic-team/ionic-framework/blob/40932d7b2d9f6650ed621e0183b9c70b74debff8/core/src/components/avatar/test/shape/avatar.e2e.ts#L11-L16

core/src/components/divider/test/basic/index.html
```diff
@@ -0,0 +1,41 @@
...
+        <ion-toolbar>
+          <ion-title>Divider - Basic</ion-title>
+        </ion-toolbar>
+      </ion-header>
+
+      <ion-content>
+        <div>
+          <h3>Default</h3>
+          Top
+          <ion-divider ion-divider></ion-divider>
```
brandyscarney, line 29:
```suggestion
          <ion-divider></ion-divider>
```

core/src/components/divider/test/spacing/divider.e2e.ts
```diff
@@ -0,0 +1,31 @@
...
+
+configs({ directions: ['ltr'], modes: ['ionic-md'] }).forEach(({ config, screenshot, title }) => {
+  test.describe(title('divider: spacing'), () => {
+    ['xxsmall', 'xsmall', 'small', 'medium', 'large', 'xlarge', 'xxlarge'].forEach((spacing) => {
+      test(`should render divider with ${spacing} spacing`, async ({ page }) => {
+        await page.setContent(
+          `
+        <style>
+            #container {
+              padding: 10px;
```
brandyscarney, line 12: Do we need this padding here? We should try to capture the divider as it is without any extra css.

core/src/components/divider/divider.scss
```diff
@@ -0,0 +1,78 @@
...
+  --margin-top: #{globals.$ion-space-1000};
+  --margin-bottom: #{globals.$ion-space-1000};
+}
+
+// Divider Inset
+// --------------------------------------------------
+
+:host(.divider-inset) {
+  --margin-top: #{globals.$ion-space-600};
+  --margin-bottom: #{globals.$ion-space-600};
```
brandyscarney, line 75: This margin is overriding the margin defined by the spacing. I think this is incorrect. You might want to move this rule above the spacing.

core/src/components/divider/test/spacing/index.html
brandyscarney, line 1: Could this be updated to a grid layout instead of using dropdowns? For example: https://ionic-framework-git-next-ionic1.vercel.app/src/components/input/test/bottom-content?ionic:theme=ionic

Maybe the left column could be all of the spacing from xxsmall to xxlarge and the right column inset + spacing?

brandyscarney (changes requested):
Looking good so far, just found a few concerns. 🙂

core/src/components/divider/test/spacing/divider.e2e.ts
OS-martacarlos: it is indeed a leftover for when I was testing . I'll remove it

core/src/components/divider/divider.scss
OS-martacarlos: In the figma design, if the `inset = true` , the margins should be same for all sizes (which is `#{globals.$ion-space-600}`)

brandyscarney: Can we ask design about this? Under inset it just says: `Sets if the list has horizontal margin or not. By default, false.` I read that to mean that it only changes the horizontal margin.

core/src/components/divider/divider.common.scss
```diff
@@ -0,0 +1,41 @@
...
+  --padding-start: 0px;
+  --padding-end: 0px;
+
+  display: block;
+
+  width: 100%;
+
+  /* stylelint-disable */
+  @include ltr() {
+    padding-right: var(--padding-end);
```
brandyscarney, line 21:
```suggestion
    padding-right: calc(var(--padding-end) + var(--ion-safe-area-right, 0px));
```

I realized that I copied the original style from item, which is slightly different because it has inner padding that accounts for this. However, we also need to account for the right safe area here.

core/src/components/divider/divider.common.scss
```diff
@@ -0,0 +1,41 @@
...
+    padding-left: calc(var(--padding-start) + var(--ion-safe-area-left, 0px));
+  }
+
+  @include rtl() {
+    padding-right: calc(var(--padding-start) + var(--ion-safe-area-right, 0px));
+    padding-left: var(--padding-end);
```
brandyscarney, line 27:
```suggestion
    padding-left: calc(var(--padding-end) + var(--ion-safe-area-left, 0px));
```

I realized that I copied the original style from item, which is slightly different because it has inner padding that accounts for this. However, we also need to account for the left safe area here.

core/src/components/divider/test/basic/divider.e2e.ts
```diff
@@ -0,0 +1,40 @@
+import { expect } from '@playwright/test';
+import { configs, test } from '@utils/test/playwright';
+
+configs({ directions: ['ltr', 'rtl'], modes: ['ios', 'md', 'ionic-md'] }).forEach(({ config, screenshot, title }) => {
```
brandyscarney, line 4:
```suggestion
configs({ directions: ['ltr'], modes: ['ios', 'md', 'ionic-md'] }).forEach(({ config, screenshot, title }) => {
```

I think we can skip the rtl screenshots here since nothing on the basic divider changes

core/src/components/divider/test/basic/divider.e2e.ts-snapshots/divider-basic-default-ionic-md-rtl-light-Mobile-Chrome-linux.png
brandyscarney, line 1: Make sure to delete these screenshots if you skip the rtl tests

core/src/components/divider/test/spacing/divider.e2e.ts
```diff
@@ -0,0 +1,25 @@
+import { expect } from '@playwright/test';
+import { configs, test } from '@utils/test/playwright';
+
+configs({ directions: ['ltr', 'rtl'], modes: ['ios', 'md', 'ionic-md'] }).forEach(({ config, screenshot, title }) => {
```
brandyscarney, line 4:
```suggestion
configs({ directions: ['ltr'], modes: ['ios', 'md', 'ionic-md'] }).forEach(({ config, screenshot, title }) => {
```

We can skip rtl here since the spacing only applies to top/bottom

core/src/components/divider/test/spacing/divider.e2e.ts-snapshots/divider-spacing-large-ionic-md-rtl-light-Mobile-Chrome-linux.png
brandyscarney, line 1: Make sure to delete these screenshots if you skip the rtl tests

core/src/components/divider/test/basic/divider.e2e.ts
```diff
@@ -0,0 +1,40 @@
...
+        `,
+        config
+      );
+
+      const container = page.locator('#container');
+
+      await expect(container).toHaveScreenshot(screenshot(`divider-basic-default`));
+    });
+
+    test('should not have visual regressions when inset is enabled', async ({ page }) => {
```
brandyscarney, line 23: If you want to keep the rtl tests for inset since it applies to left/right then you can move it to a separate config like this:

```ts
configs({ modes: ['ios', 'md', 'ionic-md'] }).forEach(({ config, screenshot, title }) => {
```

If you leave `directions` out of the config it will run for both `ltr` and `rtl` 🙂

core/src/components/divider/test/basic/divider.e2e.ts
brandyscarney, line 4:
```suggestion
configs({ directions: ['ltr'], modes: ['md', 'ionic-md'] }).forEach(({ config, screenshot, title }) => {
```

Optional: we could actually skip `ios` for all of these tests since ios and md use the same values

brandyscarney (commented):
Thank you for making those changes! The only blocking change I'm requesting is that I noticed the LTR/RTL padding is missing on one side because I copied the styles from item. Sorry I just realized this! The rest of my requests are just to limit the number of screenshots generated by the E2E tests to only what we really need. 🙂

brandyscarney (approved):
LGTM! 🎉

core/src/components/divider/divider.common.scss
```diff
@@ -0,0 +1,41 @@
+@import "../../themes/native/native.globals";
```
thetaPC, line 1: Since this is the common file then we import the exact path that is needed. Especially since `native.globals` is for the native (iOS and md) only. It shouldn't be used with ionic.

```suggestion
@import "../../themes/mixins";
```

core/src/components/divider/divider.ionic.scss
```diff
@@ -0,0 +1,47 @@
+@use "../../themes/ionic/ionic.globals.scss" as globals;
+@import "./divider.common.scss";
```
thetaPC, line 2: We don't want to use `@import` anymore because it's deprecated. We only do it with the ionic theme. At some point, we'll make the shift for iOS and md to no longer use `@import`.

```suggestion
@use "./divider.common.scss";
```

thetaPC (changes requested):

thetaPC (approved):
LGTM

PR #30282, opened 2025-03-20
feat(select): allow to pass custom close button text

niconaso:
Issue number: internal

---------

## What is the new behavior?

- Introduces a new property to Modal interface to customize the close button.
- Adds e2e tests to for the new prop.

- [ ] Yes
- [x] No


thetaPC:
Thank you for submitting the PR!

Please create an issue for this. We'd like to understand the use case for this feature before we expose this API. Thanks!

niconaso:
Thanks @thetaPC. I was going to create an issue for this PR but I've found an existing one -> https://github.com/ionic-team/ionic-framework/issues/30295

luisbytes:
@thetaPC @niconaso I consider better to use the existing cancelText property, instead of creating a new one.

core/src/components/select/select.tsx
```diff
@@ -724,6 +724,7 @@ export class Select implements ComponentInterface {
       component: 'ion-select-modal',
       componentProps: {
         header: interfaceOptions.header,
+        closeText: interfaceOptions.closeText,
```
luisbytes, line 727: cancelText: this.cancelText,

I consider better to use the existing property cancelText, instead of creating a new one, so we avoid having two properties for the same thing.

niconaso: I agree with you @luisbytes. Thanks for letting me know. 
I'll change it and push the changes today.

niconaso: Done @luisbytes 😄

mae-financial-com:
Is there any date when this feature might get released?

thetaPC:
We appreciate your interest in this feature! While the team does plan to address it at some point, there's no set timeframe for its release. We recommend subscribing to the issue for any future updates.

core/src/components/select/test/basic/select.e2e.ts-snapshots/select-basic-modal-scroll-to-selected-ios-ltr-Mobile-Chrome-linux.png
thetaPC, line 1: The difference is the cancel button text. Since we are passing the cancel button text from `select`, it now uses the [default value](https://github.com/niconaso/ionic-framework/blob/e30f907cb446af9720125d909d8497d8c467dfe5/core/src/components/select/select.tsx#L95) that `select` sets for cancel button. I believe this is actually a fix because it lines up  with the other interfaces.

thetaPC:
Thank you for the PR! The feature will be released in v8.8!!

PR #30284, opened 2025-03-20
feat(avatar):  add disabled property

JoaoFerreira-FrontEnd:
Issue number: internal

---------

Avatar does not have a disabled state for the ionic theme.

## What is the new behavior?
- Added styles for ionic theme disabled state
- Added states e2e test & snapshots

- [ ] Yes
- [x] No

## Other information

[Preview](https://ionic-framework-git-rou-11728-ionic1.vercel.app/src/components/avatar/test/states?ionic:theme=ionic)


core/src/components/avatar/avatar.common.scss
```diff
@@ -25,3 +25,10 @@
 
   overflow: hidden;
 }
+
+// Disabled Checkbox
+// ---------------------------------------------
+
+:host(.avatar-disabled) {
```
brandyscarney, line 32: Why do we add this here when there are no styles for the `md` and `ios` themes?

core/src/components/avatar/avatar.ionic.scss
```diff
@@ -171,3 +171,9 @@
 :host(.avatar-xxsmall) ::slotted(ion-badge.badge-vertical-bottom:empty) {
   transform: translate(globals.$ion-scale-100, calc(globals.$ion-scale-100));
 }
+
+// Avatar Disabled
+// --------------------------------------------------
+:host(.avatar-disabled) {
+  opacity: 0.6;
```
brandyscarney, line 178: We should be using the design token here: `$ion-state-disabled`

core/src/components/avatar/test/shape/avatar.e2e.ts
```diff
@@ -188,5 +188,188 @@ configs({ directions: ['ltr'], modes: ['ionic-md'] }).forEach(({ config, screens
         await expect(container).toHaveScreenshot(screenshot(`avatar-shape-soft-image`));
       });
     });
+
+    test.describe('round-disabled', () => {
```
brandyscarney, line 192: Do we need to add screenshots for all of the shapes?

core/src/components/avatar/test/basic/avatar.e2e.ts
```diff
@@ -14,11 +14,15 @@ configs({ directions: ['ltr'] }).forEach(({ config, screenshot, title }) => {
       const avatarChip = page.locator('#avatar-chip');
       const avatarItemStart = page.locator('#avatar-item-start');
       const avatarItemEnd = page.locator('#avatar-item-end');
+      const avatarCharactersDisabled = page.locator('#avatar-characters-disabled');
+      const avatarDisabled = page.locator('#avatar-disabled');
```
brandyscarney, line 18: Can we move these tests to only take screenshots for the `ionic` theme and also use `setContent` in order to capture just the avatar. Here is an example: https://github.com/ionic-team/ionic-framework/blob/40932d7b2d9f6650ed621e0183b9c70b74debff8/core/src/components/badge/test/shape/badge.e2e.ts#L7-L24

Please make sure not to include the `ion-item` in the screenshots.

brandyscarney (changes requested):
Looks good just a few requested changes!

core/src/components/avatar/avatar.tsx
```diff
@@ -93,6 +98,7 @@ export class Avatar implements ComponentInterface {
           [`avatar-${shape}`]: shape !== undefined,
           [`avatar-image`]: this.hasImage,
           [`avatar-icon`]: this.hasIcon,
+          [`avatar-disabled`]: this.disabled,
```
ShaneK, line 101: Nit: Might be worth doing something like
```typescript
const { disabled } = this;
[...]
[`avatar-disabled`]: disabled
```
You could do that with `hasImage` and `hasIcon` too and things might look a little more clean. This is just a nit though, no worries if you don't want to do that.

core/src/components/avatar/avatar.common.scss
JoaoFerreira-FrontEnd: I thought it was so generic it makes sense there, but I'll put it on ionic.scss

core/src/components/avatar/avatar.ionic.scss
JoaoFerreira-FrontEnd: I'll use it

core/src/components/avatar/test/states/avatar.e2e.ts
```diff
@@ -0,0 +1,28 @@
...
+            <img src="/src/components/avatar/test/avatar.svg" />
+          </ion-avatar>
+        </div> 
+      `,
+        config
+      );
+      const avatarCharactersDisabled = page.locator('#avatar-characters-disabled');
+      const avatarDisabled = page.locator('#avatar-disabled');
+      await expect(avatarCharactersDisabled).toHaveScreenshot(screenshot(`avatar-characters-disabled-diff`));
+      await expect(avatarDisabled).toHaveScreenshot(screenshot(`avatar-disabled-diff`));
```
brandyscarney, line 25:
```suggestion
      const container = page.locator('#container');

      await expect(container).toHaveScreenshot(screenshot(`avatar-disabled`));
```

Instead of querying for all of these we could just take one screenshot of the container

core/src/components/avatar/test/states/index.html
```diff
@@ -0,0 +1,34 @@
...
+      <ion-header>
+        <ion-toolbar>
+          <ion-title>Avatar - States</ion-title>
+        </ion-toolbar>
+      </ion-header>
+
+      <ion-content>
+        <h3>Disabled</h3>
+        <ion-avatar id="avatar-characters-disabled" disabled> AV </ion-avatar>
+        <ion-avatar id="avatar-disabled" disabled>
```
brandyscarney, line 28:
```suggestion
        <ion-avatar disabled>AV</ion-avatar>
        <ion-avatar disabled>
```

core/src/components/avatar/test/states/avatar.e2e.ts
```diff
@@ -0,0 +1,27 @@
...
+ * Usages of Avatar in slots are tested in components that use Avatar.
+ */
+configs({ directions: ['ltr'], modes: ['ionic-md'] }).forEach(({ config, screenshot, title }) => {
+  test.describe(title('avatar: states'), () => {
+    test('should not have visual regressions', async ({ page }) => {
+      await page.setContent(
+        ` 
+        <div id="container"> 
+          <ion-avatar id="avatar-characters-disabled" disabled> AV </ion-avatar>
+          <ion-avatar id="avatar-characters-disabled" disabled>
```
brandyscarney, line 15:
```suggestion
          <ion-avatar disabled> AV </ion-avatar>
          <ion-avatar disabled>
```

You can remove the ids here too but shouldn't need to generate screenshots again after

brandyscarney (approved):
Looks good! 👍

ShaneK (approved):
Looks good to me! I do agree with Brandy's comment though, mostly because if it doesn't get cleaned up now it'll probably never be and may end up being copied/pasted around. But it's pretty minor, so 🤷

PR #30285, opened 2025-03-20
test(angular): fixing over-import of RouterModule in the standalone router-link test

ShaneK:

---------

Currently the standalone test app for router-link in Angular imports all of `RouterModule` when it only needs to import `RouterLink` from `@angular/router`

## What is the new behavior?

Now, the Angular test app will import only what it needs from router instead of over-importing.

- [ ] Yes
- [X] No

## Other information

This PR helps align our Angular test apps with a more real-world implementation expectation


thetaPC (approved):
LGTM

PR #30286, opened 2025-03-20
test(angular): add a landing page to go between lazy and standalone apps

brandyscarney:
## What is the new behavior?
Adds a landing page to the Angular test app to make it easier to get to and debug standalone.

- [ ] Yes
- [x] No


packages/angular/test/apps/ng19/src/app/app.module.ts
thetaPC, line 1: Why was this deleted?

thetaPC (approved):
LGTM, thank you so much for this! It's been annoying to test standalone. I also prefer the standalone landing page with the sections. Maybe we can do the same for lazy in the future.

packages/angular/test/apps/ng19/src/app/app.module.ts
brandyscarney: Sorry I had this set to merge on approval. This file was specific to the `ng19` app but I updated the file in the `base` folder to match this file: https://github.com/ionic-team/ionic-framework/pull/30286/files#diff-174033d040c9d2279d515398002826936441e7892b50ecde6a815fb719c48b9e

Since they are the same now and what's in the base folder works for all versions, this isn't needed.

PR #30292, opened 2025-03-21
feat(accordion): add skip-accordion-toggle attribute

### LINKED ISSUE #30291: feat: add skip-accordion-toggle attribute
### Prerequisites

For Accordion and Accordion Groups, you can use the `event.stopPropagation()` method to avoid toggling the accordion if something is clicked inside the `header` slot, but this will break anything that depends on the event bubbling up (e.g. having a Radio Group parent component). Another option is required.

### Describe the Use Case

Creating a tree view with Accordions and using a Radio Group so only one item is selected. If you click on the Radio of an accordion header (which would be a folder) it will toggle the accordion, which is not exactly the expected behavior.

By having a `skip-accordion-toggle` attribute, you can skip the Accordion toggle for any component inside the Accordion header, while letting the event bubble up.

### Describe Alternatives

_No response_

### Related Code

_No response_

### Additional Information

_No response_

eduardoRoth:
Issue number: resolves #30291

---------

Any component that is clicked inside the `slot="header"` of an Accordion component will trigger the `ionChange` event if it's inside an Accordion Group.

You can stop this by using `event.stopPropagation()` on the component inside the Accordion header, but sometimes you require the event to bubble up, for example when using a Radio Group. In those cases, if you stop the event from bubbling, the Radio Group won't update it's value. 

## What is the new behavior?

By adding a `skip-accordion-toggle` attribute to any element inside the `slot="header"` it will allow the event to bubble up while skipping the accordion from being toggled.

- [ ] Yes
- [X] No

## Other information
N/A

### DIFF
--- a/core/src/components/accordion/accordion.tsx
+++ b/core/src/components/accordion/accordion.tsx
@@ -381,12 +381,22 @@ export class Accordion implements ComponentInterface {
     return previousSibling as HTMLIonAccordionElement;
   };
 
-  private toggleExpanded() {
+  private toggleExpanded(event: MouseEvent) {
     const { accordionGroupEl, disabled, readonly, value, state } = this;
 
     if (disabled || readonly) return;
 
     if (accordionGroupEl) {
+      /**
+       * We check if the element that was clicked has the attribute
+       * skip-accordion-toggle. In case the element has it, then the accordion
+       * won't request an accordion toggle if it's inside an accordion group
+       */
+      const target = event.target as HTMLInputElement;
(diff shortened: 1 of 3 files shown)


thetaPC:
Thank you for submitting the PR!

After discussing, we decided to not implement it due to the internal guard clauses that the components have and native UX standards. Two workarounds and an in depth explanation were [provided](https://github.com/ionic-team/ionic-framework/issues/30291#issuecomment-3781377566) at the [issue](https://github.com/ionic-team/ionic-framework/issues/30291). I recommend reviewing them.

PR #30293, opened 2025-03-21
fix(range): emit ionInput when value changes

### LINKED ISSUE #29619: bug: range ionInput event behavior inconsistent with documentation
### Prerequisites

### Ionic Framework Version

v7.x, v8.x

### Current Behavior

The `onIonInput` event behavior isn't consistent with the documentation for the event. This is the current description:

> The `ionInput` event is fired for `<ion-range>` elements when the value is modified. Unlike `ionChange`, `ionInput` is fired continuously while the user is dragging the knob.

The actual behavior is that the event is triggered whenever the knob is moved a single pixel (after an initial "sticky" behavior) instead of it triggering whenever the actual value is modified, causing a lot of events being triggered for absolutely no reason. The second part "_is fired continously_" is ambivalent and could match either expectation.

### Expected Behavior

I would expect the event to trigger only when the value is modified rather than when the knob is moved.

### Steps to Reproduce

1. Open the reproduction [Stackblitz link](https://stackblitz.com/edit/umlmf6?file=src%2Fmain.tsx).
2. Open the browsers developer console.
3. Drag the slider a short distance until the first `ionInput` event is triggered, then keep moving the knob a short distance and see how the event continues to be logged even without the value changing.

### Code Reproduction URL

https://stackblitz.com/edit/umlmf6?file=src%2Fmain.tsx

### Ionic Info

Ionic:

   Ionic CLI : 7.2.0.

Utility:

System:

   NodeJS : v20.9.0
   npm    : 10.1.0
   OS     : Windows 10 (actual version 11)

### Additional Information

As I see it there's 3 possible solutions:

1. Only trigger the event when the value has changed from it's previous value as the documentation says.
2. Same as 1 but only when `snaps: true`.
3. Update the documentation to not specify `when the value is modified` but rather something like `when the knob is moved`.

I believe 1 is the preferable solution, but it might be considered a breaking change and so 2 is a reasonable compromise as there's no practical nor visual updates between value-changes when `snaps: true` making it a non-breaking change.

3 is by far the least preferable solution in my opinion as this issue is not only a documentation error but the current implementation also carries an unnecessary performance impact (again, at least when `snaps: true`).

thetaPC:
Issue number: resolves #29619 

---------

The `ionInput` emits for range even when the value hasn't changed. This does not match our documentation. It should only emit when the value changes (and continuously while the user is dragging the knob).

## What is the new behavior?

- Moved the emitter to the value watch function, to determine if the value has changed.
- Added a test
-

- [ ] Yes
- [x] No

## Other information

How to test:
1. Navigate to the [range basic HTML file](https://github.com/ionic-team/ionic-framework/blob/8ed08fcba59c9fde5c6d08d721fd1d00642de4f9/core/src/components/range/test/basic/index.html#L77)
2. Add the following script
```
const ionicRanges = document.querySelectorAll('ion-range');

ionicRanges.forEach(range => {  
  range.addEventListener('ionInput', function(ev) {
    console.log('ionInput', ev.currentTarget.value);
  });
});
```
3. Navigate to the [range test page](http://localhost:3333/src/components/range/test/basic)
4. Open the console
5. Move the single knob range (let go when you're done)
6. Verify that the value is shown in the console
7. Tap as close to the middle of the knob. The goal is to tap it without the value moving.
8. Verify that the value does not show in the console

### DIFF
--- a/core/src/components/range/range.tsx
+++ b/core/src/components/range/range.tsx
@@ -231,7 +231,7 @@ export class Range implements ComponentInterface {
    * @param oldVal - The old value.
    * @returns `true` if the values are different, `false` otherwise.
    */
-  private areValuesDifferent = (newVal: RangeValue, oldVal: RangeValue) => {
+  private compareValues = (newVal: RangeValue, oldVal: RangeValue) => {
     if (typeof newVal === 'object' && typeof oldVal === 'object') {
       return newVal.lower !== oldVal.lower || newVal.upper !== oldVal.upper;
     }
--- a/core/src/components/range/range.tsx
+++ b/core/src/components/range/range.tsx
@@ -213,7 +213,12 @@ export class Range implements ComponentInterface {
    */
   @Prop({ mutable: true }) value: RangeValue = 0;
   @Watch('value')
-  protected valueChanged() {
+  protected valueChanged(newVlue: RangeValue, oldValue: RangeValue) {
+    // The check is necessary because the value can be an object.
+    if (JSON.stringify(newVlue) !== JSON.stringify(oldValue)) {
+      this.ionInput.emit({ value: this.value });
+    }
+
     if (!this.noUpdate) {
       this.updateRatio();
     }
@@ -590,8 +595,6 @@ export class Range implements ComponentInterface {
           lower: Math.min(valA, valB),
           upper: Math.max(valA, valB),
         };
-
-    this.ionInput.emit({ value: this.value });
 
     this.noUpdate = false;
   }
--- a/core/src/components/range/range.tsx
+++ b/core/src/components/range/range.tsx
@@ -213,9 +213,9 @@ export class Range implements ComponentInterface {
    */
   @Prop({ mutable: true }) value: RangeValue = 0;
   @Watch('value')
-  protected valueChanged(newVlue: RangeValue, oldValue: RangeValue) {
-    // The check is necessary because the value can be an object.
-    if (JSON.stringify(newVlue) !== JSON.stringify(oldValue)) {
+  protected valueChanged(newValue: RangeValue, oldValue: RangeValue) {
+    const valuesChanged = this.areValuesDifferent(newValue, oldValue);
+    if (valuesChanged) {
       this.ionInput.emit({ value: this.value });
     }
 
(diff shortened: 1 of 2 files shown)


core/src/components/range/range.tsx
```diff
@@ -213,7 +213,12 @@ export class Range implements ComponentInterface {
    */
   @Prop({ mutable: true }) value: RangeValue = 0;
   @Watch('value')
-  protected valueChanged() {
+  protected valueChanged(newVlue: RangeValue, oldValue: RangeValue) {
+    // The check is necessary because the value can be an object.
+    if (JSON.stringify(newVlue) !== JSON.stringify(oldValue)) {
```
ShaneK, line 218: JSON.stringify is a bit slow. I don't know that it super matters in the long run, but it would be significantly more efficient if you compared the properties without stringifying them. I know it's a bit annoying because they're a union type of a number and object, but worth considering I guess.

ShaneK (approved):
Generally looks good, I left one comment but it's probably just a nit. Nice work 🚀

core/src/components/range/range.tsx
thetaPC: [Done](https://github.com/ionic-team/ionic-framework/pull/30293/commits/e8b57c838b5c7d9658f86530a886c16236fd1c7e)!

core/src/components/range/range.tsx
```diff
@@ -213,12 +213,32 @@ export class Range implements ComponentInterface {
...
   }
 
+  /**
+   * Compares two RangeValue inputs to determine if they are different.
+   *
+   * @param newVal - The new value.
+   * @param oldVal - The old value.
+   * @returns `true` if the values are different, `false` otherwise.
+   */
+  private areValuesDifferent = (newVal: RangeValue, oldVal: RangeValue) => {
```
brandyscarney, line 234:
```suggestion
  private compareValues = (newVal: RangeValue, oldVal: RangeValue) => {
```

Nit: can we name this something else? We don't have any other functions that begin with `are`.

Select uses a similar function here: https://github.com/ionic-team/ionic-framework/blob/9e11fcb2bb504916a1ef910a011f80b93be36109/core/src/utils/forms/compare-with-utils.ts#L10

brandyscarney (approved):
Non-blocking suggestion but otherwise looks good! 🎉

core/src/components/range/range.tsx
thetaPC: Done: https://github.com/ionic-team/ionic-framework/pull/30293/commits/f7ad2ba7b76fd47213f38a257dfa4c516ae3d854 and https://github.com/ionic-team/ionic-framework/pull/30293/commits/812d7674efe63983441e39e0b613a864bfcd0f56

PR #30300, opened 2025-03-24
fix(checkbox): ensure proper visual selection when navigating via VoiceOver in Safari

ShaneK:

---------

Currently, MacOS voice over on Safari does not recognize ion-checkbox correctly and fails to highlight the element properly

## What is the new behavior?

By adding the role property to the host element, we're correctly identifying ion-checkbox as a checkbox so Safari knows how to handle it.

- [ ] Yes
- [X] No


core/src/components/checkbox/checkbox.scss
```diff
@@ -102,6 +102,14 @@
   @include margin(null, null, $checkbox-item-label-margin-bottom, null);
 }
 
+/**
+  * The native control it must be hidden with display:none instead of aria-hidden="true"
+  * to avoid nested interactive elements accessibility issues
+ */
+.checkbox-native-control {
```
brandyscarney, line 109: You should be able to target the `input` here directly without adding a class

ShaneK: 🫡

core/src/components/checkbox/checkbox.scss
```diff
@@ -111,8 +111,12 @@
   display: none;
 }
 
+/**
+  * The native control it must be hidden with display:none instead of aria-hidden="true"
+  * to avoid nested interactive elements accessibility issues
```
brandyscarney, line 116:
```suggestion
  * The native input must be hidden with display instead of visibility or 
  * aria-hidden to avoid accessibility issues with nested interactive elements
```

core/src/components/checkbox/checkbox.tsx
```diff
@@ -253,11 +262,22 @@ export class Checkbox implements ComponentInterface {
 
     renderHiddenInput(true, el, name, checked ? value : '', disabled);
 
+    // Determine appropriate accessible name
+    const hasLabelContent = el.textContent !== '';
+    const inputLabelId = inputId + '-lbl';
+
+    // The host must be a checkbox role to support Safari's Voice Over accessibility
```
brandyscarney, line 269:
```suggestion
    // The host element must have a checkbox role to ensure proper VoiceOver
    // support in Safari for accessibility
```

brandyscarney (commented):
A few nit comments but other than that it looks good.

brandyscarney (approved):

core/src/components/checkbox/checkbox.scss
```diff
@@ -111,8 +111,12 @@
   display: none;
 }
 
+/**
+  * The native input must be hidden with display instead of visibility or 
+  * aria-hidden to avoid accessibility issues with nested interactive elements
+ */
```
thetaPC, line 117:
```suggestion
/**
 * The native input must be hidden with display instead of visibility or 
 * aria-hidden to avoid accessibility issues with nested interactive elements.
 */
```

core/src/components/checkbox/checkbox.tsx
```diff
@@ -181,6 +181,15 @@ export class Checkbox implements ComponentInterface {
     this.ionBlur.emit();
   };
 
+  private onKeyDown = (ev: KeyboardEvent) => {
+    if (ev.key === ' ' || ev.key === 'Enter') {
```
thetaPC, line 185: Based on [W3C](https://www.w3.org/WAI/ARIA/apg/patterns/checkbox/examples/checkbox/#kbd_label), only the `SPACE` key should toggle. I also verified with Chakra UI, a library known for it's accessibility.

```suggestion
    if (ev.key === ' ') {
```

core/src/components/checkbox/checkbox.tsx
```diff
@@ -253,11 +262,23 @@ export class Checkbox implements ComponentInterface {
 
     renderHiddenInput(true, el, name, checked ? value : '', disabled);
 
+    // Determine appropriate accessible name
```
thetaPC, line 265:
```suggestion
    // Determine appropriate accessible name.
```

core/src/components/checkbox/checkbox.tsx
```diff
@@ -253,11 +262,23 @@ export class Checkbox implements ComponentInterface {
...
+    const hasLabelContent = el.textContent !== '';
+    const inputLabelId = inputId + '-lbl';
+
+    // The host element must have a checkbox role to ensure proper VoiceOver
+    // support in Safari for accessibility
```
thetaPC, line 270:
```suggestion
    // support in Safari for accessibility.
```

core/src/components/checkbox/checkbox.tsx
```diff
@@ -253,11 +262,23 @@ export class Checkbox implements ComponentInterface {
 
     renderHiddenInput(true, el, name, checked ? value : '', disabled);
 
+    // Determine appropriate accessible name
+    const hasLabelContent = el.textContent !== '';
+    const inputLabelId = inputId + '-lbl';
```
thetaPC, line 267: I wonder if there's a benefit to having this done in `connectedCallback` like [how](https://github.com/ionic-team/ionic-framework/blob/bbeff5b6f2356b69809c56924a1a803dcc95eefc/core/src/components/radio-group/radio-group.tsx#L120-L122) radio group does it. Thoughts?

core/src/components/checkbox/checkbox.tsx
```diff
@@ -253,11 +262,23 @@ export class Checkbox implements ComponentInterface {
...
+    // The host element must have a checkbox role to ensure proper VoiceOver
+    // support in Safari for accessibility
     return (
       <Host
+        role="checkbox"
         aria-checked={indeterminate ? 'mixed' : `${checked}`}
         aria-describedby={this.getHintTextID()}
         aria-invalid={this.getHintTextID() === this.errorTextId}
+        aria-labelledby={hasLabelContent ? inputLabelId : null}
+        aria-label={!hasLabelContent ? inheritedAttributes['aria-label'] || 'checkbox' : null}
```
thetaPC, line 278: It would be best to not give it a default value like `checkbox` since apps can be in a different language and the dev might not pass an `aria-label`. What about?

```suggestion
        aria-label={inheritedAttributes['aria-label'] || null}
```

thetaPC, line 278: By using `inheritedAttributes['aria-label']` here, do we still need to [include](https://github.com/ionic-team/ionic-framework/blob/bbeff5b6f2356b69809c56924a1a803dcc95eefc/core/src/components/input/input.tsx#L821) it on `native-input`?

thetaPC (changes requested):
Tested on Windows (Firefox, Chrome, Edge) and MacOS (Firefox, Chrome, Safari), sounds as expected.

Just some code requests.

core/src/components/checkbox/checkbox.tsx
ShaneK: You're required to provide a default label or it fails accessibility validation

core/src/components/checkbox/checkbox.tsx
ShaneK: No need for that, it doesn't need to dynamically change, but I will lift it out of render and make it a private variable

core/src/components/checkbox/checkbox.tsx
ShaneK: I don't understand what you mean, I'm sorry

core/src/components/checkbox/checkbox.tsx
ShaneK: hasLabelContent should stay in render though since it can change dynamically

core/src/components/checkbox/checkbox.tsx
thetaPC: Discussed offline, there shouldn't be any issues with leaving it as is. Plus, by removing it from `.native-input` then it might lead to changes that we didn't expect down the line.

thetaPC (approved):
LGTM

PR #30303, opened 2025-03-26
feat(button): add support for badge

OS-pedrolourenco:
Issue number: internal

---------

The ion-badge is not currently supported when inside the ion-button.

## What is the new behavior?
- The experience of the ion-badge inside the ion-button has been implemented according to our designs.

- [ ] Yes
- [x] No


core/src/components/button/button.common.scss
```diff
@@ -306,3 +307,25 @@ ion-ripple-effect {
...
+
+// This rule only works in Chrome
+:has(ion-badge) .button-native {
+  overflow: visible;
+}
+
+// This rule only works in Safari
+:host(:has(ion-badge)) .button-native {
+  overflow: visible;
+}
```
OS-pedrolourenco, line 322: I am not quite sure why this happens but I would appreciate some feedback on this.

core/src/components/badge/badge.common.scss
```diff
@@ -48,6 +49,14 @@
   color: #{color.current-color(contrast)};
 }
 
+:host(:not(.in-tab-button).badge-vertical-top) {
+  top: 0;
+}
+
+:host(:not(.in-tab-button).badge-vertical-bottom) {
+  bottom: 0;
+}
```
thetaPC, line 58: Why was this removed from empty?

core/src/components/badge/badge.common.scss
```diff
@@ -1,3 +1,4 @@
+@use "../../themes/ionic/ionic.globals.scss" as globals;
```
thetaPC, line 1: This file is for all themes so we do not use globals here since globals is for ionic only.

```suggestion
```

core/src/components/badge/badge.common.scss
```diff
@@ -56,10 +65,24 @@
...
+  height: globals.$ion-scale-400;
+
+  font-size: globals.$ion-font-size-300;
+
+  line-height: globals.$ion-font-line-height-500;
+
+  ::slotted(ion-icon) {
+    width: globals.$ion-scale-300;
+    height: globals.$ion-scale-300;
+  }
```
thetaPC, line 87: Tokens are only meant for ionic them only. Move these to the ionic file and use vars for native. Review how tab button implements this.

core/src/components/badge/badge.ionic.scss
```diff
@@ -190,3 +190,11 @@
 
   transform: translate(-50%, -50%);
 }
+
+// Badge in Button
+// --------------------------------------------------
+
+:host([vertical]:not(.in-tab-button).in-button.badge-small) {
+  // stylelint-disable-next-line property-disallowed-list
+  right: calc(-1 * globals.$ion-space-050);
```
thetaPC, line 199: We need to be using the `position` mixin to allow apps to have the correct style regardless of their direction (LTR or RTL).

core/src/components/button/button.common.scss
thetaPC, line 322: Is this needed for native or only ionic? If it's ionic then I would look at how tab-button implements this.

thetaPC: If you're unsure on why things happen then we shouldn't implement it until completely understanding. Otherwise, this can lead to many issues down the line. It's best to ask for guidance before submitting a PR for review.

thetaPC (changes requested):

core/src/components/button/button.common.scss
OS-pedrolourenco: It is needed for every theme so that the badge isn't cut out by the native button's border radius.

core/src/components/badge/badge.common.scss
OS-pedrolourenco: I removed this from empty since it is now applicable also when badge is not empty.

core/src/components/button/button.common.scss
```diff
@@ -306,3 +307,29 @@ ion-ripple-effect {
...
+
+.button-native {
+  overflow: visible;
+
+  ion-ripple-effect {
+    // stylelint-disable-next-line property-disallowed-list
+    border-radius: inherit;
+  }
+
+  &::after {
```
thetaPC, line 322: This can be re-written for easier readability:

```suggestion
.button-native  ion-ripple-effect,
.button-native::after {
```

thetaPC (changes requested):

core/src/components/button/button.common.scss
```diff
@@ -306,3 +307,30 @@ ion-ripple-effect {
...
+  @include border-radius(inherit);
+}
+
+// This rule works for Chrome.
+:has(ion-badge) .button-native {
+  --overflow: visible;
+}
+
+// This rule works for the rest of the browsers.
+:host(:has(ion-badge)) .button-native {
```
thetaPC, line 325:
```suggestion
// This rule works for Chrome.
:has(ion-badge) .button-native,
// This rule works for the rest of the browsers.
:host(:has(ion-badge)) .button-native {
```

thetaPC (changes requested):

core/src/components/button/button.common.scss
thetaPC: Discussed offline: Selectors must be split into individual selectors in case the browser doesn't support a specific selector. For example, Firefox and Safari doesn't support `:host-context()`. If an invalid selector is used, then the entire [group](tps://www.w3.org/TR/selectors-3/#grouping) of selectors will be ignored.

So we'll leave this as is.

thetaPC (approved):
LGTM

core/src/components/badge/badge.common.scss
```diff
@@ -63,3 +63,20 @@
 :host(:not(.in-tab-button)[vertical].badge-vertical-bottom) {
   bottom: 0;
 }
+
+// Badge in Button
+// --------------------------------------------------
+
+:host(:not(:empty).in-button) {
```
BenOsodrac, line 70: This CSS should be moved to badge.native.scss, along with the vars to badge.native.vars.scss.
Then this css should be repeated on ionic.theme.scss but using the design tokens directly. 

This way, wee keep the consistency with the scss architecture of other patterns, between design systems, and if in the future the tokens change values, we won't need to manually change this CSS for the Ionic Theme.

BenOsodrac (changes requested):

BenOsodrac (approved):

PR #30306, opened 2025-03-26
feat(css): add support for the subtle color map

brandyscarney:
Issue number: internal

---------

The colors are only available in bold colors.

## What is the new behavior?
Adds support for a `subtle` hue to the colors map, generating all of the same variants but in more subtle colors. This only applies for the `ionic` theme. Adds an additional variant, `foreground` to be used by components when the color is being used for text.

The `ios` and `md` themes do not add the `foreground` variant because this would be a breaking change, requiring all color overrides, palettes, and new colors to include this variant. I have added tests to make sure they still work without it.

- [ ] Yes
- [x] No

No changes are made to the existing colors.

## Other information

[Preview](https://ionic-framework-git-fw-6252-ionic1.vercel.app/src/themes/ionic/test/colors?ionic:theme=ionic)


core/src/themes/ionic/ionic.theme.default.scss
```diff
@@ -10,85 +10,211 @@
...
+      contrast: globals.$ion-text-danger,
+      contrast-rgb: globals.$ion-text-danger-rgb,
+      shade: globals.$ion-bg-danger-subtle-press,
+      // TODO this is not a verified value
+      tint: globals.$ion-semantics-danger-200,
+      foreground: globals.$ion-text-danger,
+    ),
   ),
+  // neutral/base equivalent
+  light:
```
brandyscarney, line 144: I think we should probably make light use a lighter version but I am not sure

core/src/themes/ionic/ionic.theme.default.scss
```diff
@@ -10,85 +10,211 @@
...
+    (
+      bold: (
+        base: globals.$ion-bg-neutral-base-default,
+        base-rgb: globals.$ion-bg-neutral-base-default-rgb,
+        contrast: globals.$ion-text-default,
+        contrast-rgb: globals.$ion-text-default-rgb,
+        // TODO this fails AA testing
+        // shade: globals.$ion-bg-neutral-base-press,
+        shade: globals.$ion-primitives-neutral-600,
```
brandyscarney, line 153: The 600 token is one step lighter but it fails AA with default text

core/src/themes/ionic/ionic.theme.default.scss
```diff
@@ -10,85 +10,211 @@
...
 // Color map should provide
-// - base: main color to be used.
-// - contrast: Color that will provide readable text against the base color
-// - shade: 12% darker version of the base color (mix with black)
-// - tint: 10% lighter version of the base color (mix with white)
+// - bold: a map of the bold color variations
+// - subtle: a map of the subtle color variations
+//
+// Each hue color map should provide
+// TODO: rename base to background?
```
brandyscarney, line 17: I think renaming it to background would make a lot more sense with how the design tokens are used but this would be a breaking change if we also updated the css variables: https://ionicframework.com/docs/theming/colors#new-color-creator

core/src/themes/ionic/ionic.theme.default.scss
```diff
@@ -10,85 +10,211 @@
...
-    shade: color.get-color-shade($success),
-    tint: color.get-color-tint($success),
+    bold: (
+      base: globals.$ion-bg-primary-base-default,
+      base-rgb: globals.$ion-bg-primary-base-default-rgb,
+      contrast: globals.$ion-text-inverse,
+      contrast-rgb: globals.$ion-text-inverse-rgb,
+      shade: globals.$ion-bg-primary-base-press,
+      // TODO this is not a verified value
+      tint: globals.$ion-semantics-primary-600,
```
brandyscarney, line 37: I used the token one step lighter than the base for all tint values

core/src/themes/ionic/ionic.theme.default.scss
```diff
@@ -10,85 +10,211 @@
...
+        base-rgb: globals.$ion-bg-info-subtle-default-rgb,
+        contrast: globals.$ion-text-info,
+        contrast-rgb: globals.$ion-text-info-rgb,
+        shade: globals.$ion-bg-info-subtle-press,
+        // TODO this is not a verified value
+        tint: globals.$ion-semantics-info-200,
+        foreground: globals.$ion-text-info,
+      ),
+    ),
+  // TODO - no equivalent for tertiary, can we add purple as a color?
```
brandyscarney, line 75: Tertiary is the only one missing

BenOsodrac: If we need to add something dummy for now, UX suggested Violet

core/src/themes/ionic/ionic.theme.default.scss
BenOsodrac: Sounds good to me for now, the Ionic Design System doesn't have these exact concepts (at least from UX Team), so this does the job I think

core/src/themes/ionic/ionic.theme.default.scss
BenOsodrac: Maybe use `--ion-color-dark-subtle-contrast` or `--ion-bg-neutral-bold-press?`

core/src/themes/ionic/ionic.theme.default.scss
BenOsodrac: Looking at the Colors page, I think is already light enough and different form Medium and Dark, I'm ok with this :)

core/src/themes/ionic/ionic.theme.default.scss
brandyscarney: I sent the colors to design and they are going to take a look at them so I am sure they will get back to me on this one.

core/src/themes/native/test/colors/theme.e2e.ts
```diff
@@ -31,7 +31,7 @@ const styleTestHelpers = `
       background: rgba(var(--ion-color-base-rgb), 0.16);
     }
     .ion-color {
-      color: var(--ion-color-base);
+      color: var(--ion-color-foreground);
```
thetaPC, line 34: Should we also update the native colors page? Like this [line](https://github.com/ionic-team/ionic-framework/blob/ad885c1c222dcfa704a4548ed84ba91d01145afa/core/src/themes/native/test/colors/index.html#L49)?

core/src/themes/functions.color.scss
```diff
@@ -25,17 +29,32 @@
...
+@function ion-color($name, $variation, $alpha: null, $rgb: null, $subtle: false) {
   @if not($theme-colors) {
     @error 'No theme colors set. Please make sure to call set-theme-colors($colorsMap) before using ion-color()';
   }
 
-  $values: map-get($theme-colors, $name);
-  $value: map-get($values, $variation);
-  $variable: --ion-color-#{$name}-#{$variation};
+  $values: map.get($theme-colors, $name);
+  $values: map.get($values, if($subtle, subtle, bold));
```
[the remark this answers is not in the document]
brandyscarney: Yeah I mean I can change it if it makes it clearer.

core/src/themes/ionic/ionic.theme.default.scss
```diff
@@ -10,84 +10,200 @@
...
+// - bold: a map of the bold color variations
+// - subtle: a map of the subtle color variations
+//
+// Each hue color map should provide
+// - base: The main color used for backgrounds
+// - base-rgb: The base color in RGB format
+// - contrast: A color that ensures readable text on the base color
+// - contrast-rgb: The contrast color in RGB format
+// - shade: A darker variant of the base color, used for pressed/active states
+// - tint: A lighter variant of the base color, used for ?
```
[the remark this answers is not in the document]
brandyscarney: We don't actually use `tint` anywhere for the theme right now so yeah I don't know what to put there. 😂

core/src/themes/native/test/colors/theme.e2e.ts
brandyscarney: We can just as another way to make sure the fallback is working. Updated: https://github.com/ionic-team/ionic-framework/pull/30306/commits/2a92d81ed9f40aeb5b3af484fdbcf0ec584eb540

thetaPC: There's a few other instances that should be changed in this file. I didn't link them so you wouldn't get spammed with links.

brandyscarney: Where? I only see `.ion-color` used once.

core/src/themes/functions.color.scss
```diff
@@ -79,43 +98,157 @@
...
   }
 
   @each $color-name, $value in $theme-colors {
-    --ion-color-#{$color-name}: #{map-get($value, base)};
-    --ion-color-#{$color-name}-rgb: #{color-to-rgb-list(map-get($value, base))};
-    --ion-color-#{$color-name}-contrast: #{map-get($value, contrast)};
-    --ion-color-#{$color-name}-contrast-rgb: #{color-to-rgb-list(map-get($value, contrast))};
-    --ion-color-#{$color-name}-shade: #{map-get($value, shade)};
-    --ion-color-#{$color-name}-tint: #{map-get($value, tint)};
+    @each $hue in (bold, subtle) {
```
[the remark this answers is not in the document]
brandyscarney: The `generate-color-variables` function is only called by the `core.scss` and `core.ionic.scss` files. The other palettes generate their variables like this:

https://github.com/ionic-team/ionic-framework/blob/97f96b4528e968e68c4977c9b31489ac0b5642c7/core/src/css/palettes/dark.scss#L71-L82

core/src/themes/ionic/ionic.theme.default.scss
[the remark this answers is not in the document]
thetaPC: I say we leave as is. It clearly tells us that it's used for the unknown.

core/src/themes/functions.color.scss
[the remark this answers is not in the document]
thetaPC: That's neat! I thought it was all from the same mixin file. Would not including bold in these palettes cause any issues in your opinion?

core/src/themes/functions.color.scss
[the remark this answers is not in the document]
thetaPC: Personally, I would prefer changing it to make it clear but this is not the hill I want to die on. 😆

core/src/themes/functions.color.scss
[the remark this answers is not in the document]
brandyscarney: The `bold` color map in the default colors file generates the same exact variables as above. The `subtle` map is the only one that prefixes them with `subtle`, as seen here in `@mixin generate-color-variables()`: 

```scss
$prefix: if($hue == subtle, "-subtle", "");

--ion-color-#{$color-name}#{$prefix}: #{map.get($colors, base)};
```

For bold it will generate the `--ion-color-primary` variable, but subtle will generate `--ion-color-primary-subtle`.

The mixin for the dark palette is also generating `--ion-color-primary`, so ultimately it doesn't matter that the `bold` map isn't included. In the future we should probably have all palettes use the same mixin, but for now I was mainly focused on updating the main color palette to support `subtle`.

core/src/themes/native/test/colors/theme.e2e.ts
brandyscarney: I don't know actually how we will be using those yet, since the `foreground` was created for text-only color components without a background and those are all using "subtle" backgrounds. Right now we don't have native using a different `foreground` at all, so changing them won't do anything. I can update it though in case it catches something in the future.

core/src/themes/functions.color.scss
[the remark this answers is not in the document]
thetaPC: That's right, thank you for the reminder!!

thetaPC (approved):
LGTM

core/src/themes/functions.color.scss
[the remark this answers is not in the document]
brandyscarney: Updated https://github.com/ionic-team/ionic-framework/pull/30306/commits/7174092fdb41e21fa62219754b46a4232559afbb

core/src/themes/native/test/colors/theme.e2e.ts
brandyscarney: Updated https://github.com/ionic-team/ionic-framework/pull/30306/commits/7174092fdb41e21fa62219754b46a4232559afbb

PR #30307, opened 2025-03-26
feat(badge): add subtle hue for the ionic theme

brandyscarney:
Issue number: resolves #

---------

## What is the new behavior?

-
-
-

- [ ] Yes
- [ ] No

### DIFF
--- a/core/src/components/badge/badge.ionic.scss
+++ b/core/src/components/badge/badge.ionic.scss
@@ -20,6 +20,20 @@
   font-weight: globals.$ion-font-weight-medium;
 }
 
+// Subtle Badge
+// --------------------------------------------------
+
+:host(.badge-subtle) {
+  --background: #{globals.ion-color(primary, base, $subtle: true)};
+  --background-activated: #{globals.ion-color(primary, shade, $subtle: true)};
+  --color: #{globals.ion-color(primary, contrast, $subtle: true)};
+}
+
+:host(.badge-subtle.ion-color) {
+  background: globals.current-color(base, $subtle: true);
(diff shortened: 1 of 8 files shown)


core/src/components/badge/test/hint/badge.e2e.ts-snapshots/badge-hint-avatar-ionic-md-ltr-light-Mobile-Chrome-linux.png
brandyscarney, line 1: The design shows Avatar as allowing subtle which is why I left this - need to check with design

BenOsodrac (approved):

core/src/components/badge/badge.ionic.scss
```diff
@@ -20,6 +18,32 @@
...
+:host(.badge-bold.ion-color) {
+  background: globals.current-color(base);
+  color: globals.current-color(contrast);
+}
+
+// Subtle Badge
+// --------------------------------------------------
+
+:host(.badge-subtle) {
+  --background: #{globals.ion-color(primary, base, $subtle: true)};
```
thetaPC, line 38: Do we need to pass `$subtle` or can it just be `true`?

brandyscarney: The function accepts a few values before `$subtle`:

```scss
@function ion-color($name, $variation, $alpha: null, $rgb: null, $subtle: false) {
```

So you need to pass `$subtle` or you would have to change it to this:

```scss
--background: #{globals.ion-color(primary, base, null, null, true)};
```

I thought it was more clear what the code is doing by passing the variable.

PR #30310, opened 2025-03-26
fix(modal): change default visibility to original footer when expandToScroll is false

### LINKED ISSUE #30315: bug: sheet modal. EventListeners don't work on ion-footer when expandToScroll is false
### Prerequisites

### Ionic Framework Version

v8.x

### Current Behavior

When a EventListener is attached to a ion-footer (or any of its children) that's in a sheet modal with expandToScroll = false, the event listener does not work

### Expected Behavior

The EventListener should work for the ion-footer and its children.

### Steps to Reproduce

1. Create a sheet modal with `expandToScroll = false`
2. Add a ion-footer with a button and a listener
3. Click and it won't do anything.

### Code Reproduction URL

https://stackblitz.com/edit/yrmrkw2f?file=src%2Fapp%2Fexample.component.html

### Ionic Info

Ionic:

   Ionic CLI: 7.2.1
   Ionic Framework: @ionic/angular 8.5.1
   @angular-devkit/build-angular: 19.2.4
   @angular-devkit/schematics: not installed
   @angular/cli: 19.2.4
   @ionic/angular-toolkit: 12.1.1

Utility:

System:
   NodeJS : v20.18.1
   npm    : 10.8.2
   OS     : macOS Unknown

### Additional Information

PR to fix it: https://github.com/ionic-team/ionic-framework/pull/30310

kumibrr:
Issue number: resolves #30315

---------

When `expandToScroll = false`, the original footer is swapped by a cloned one to avoid animation issues.
But if the original footer has some eventListeners, they won't work on the cloned footer.

## What is the new behavior?
Footer visibility is the original by default, and the visibility will swap to the cloned one when the animation starts, and will be swapped back to the original when the animation ends

- [ ] Yes
- [x] No

## Other information

| Before | After |
|--------|-------|
|  <video src="https://github.com/user-attachments/assets/d811936e-e7ac-40c5-97bc-fd5d6583ad3a"> |  <video src="https://github.com/user-attachments/assets/c322288f-665b-41f2-9516-6c99224c4c58">   |

EDIT: The button in the videos "jump" from left to right, this is caused by another issue I discovered yesterday (stylesheets not being applied), but it's quite tricky to solve. I've opened an issue for that https://github.com/ionic-team/ionic-framework/issues/30312

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,17 @@
 All notable changes to this project will be documented in this file.
 See [Conventional Commits](https://conventionalcommits.org) for commit guidelines.
 
+## [8.5.8](https://github.com/ionic-team/ionic-framework/compare/v8.5.7...v8.5.8) (2025-05-28)
+
+
+### Bug Fixes
+
+* **input-password-toggle, button:** force update aria attributes ([#30411](https://github.com/ionic-team/ionic-framework/issues/30411)) ([4e38700](https://github.com/ionic-team/ionic-framework/commit/4e387005663b4e8425cb28e41608bb4f924b3864))
+
+
+
+
+
 ## [8.5.7](https://github.com/ionic-team/ionic-framework/compare/v8.5.6...v8.5.7) (2025-05-07)
 
 
--- a/core/CHANGELOG.md
+++ b/core/CHANGELOG.md
@@ -3,6 +3,17 @@
 All notable changes to this project will be documented in this file.
 See [Conventional Commits](https://conventionalcommits.org) for commit guidelines.
 
+## [8.5.8](https://github.com/ionic-team/ionic-framework/compare/v8.5.7...v8.5.8) (2025-05-28)
+
+
+### Bug Fixes
+
+* **input-password-toggle, button:** force update aria attributes ([#30411](https://github.com/ionic-team/ionic-framework/issues/30411)) ([4e38700](https://github.com/ionic-team/ionic-framework/commit/4e387005663b4e8425cb28e41608bb4f924b3864))
+
+
+
+
+
 ## [8.5.7](https://github.com/ionic-team/ionic-framework/compare/v8.5.6...v8.5.7) (2025-05-07)
 
 
--- a/core/package-lock.json
+++ b/core/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "@ionic/core",
-  "version": "8.5.7",
+  "version": "8.5.8",
   "lockfileVersion": 2,
   "requires": true,
   "packages": {
     "": {
       "name": "@ionic/core",
-      "version": "8.5.7",
+      "version": "8.5.8",
       "license": "MIT",
       "dependencies": {
         "@stencil/core": "4.20.0",
--- a/core/package.json
+++ b/core/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@ionic/core",
-  "version": "8.5.7",
+  "version": "8.5.8",
   "description": "Base components for Ionic",
   "keywords": [
     "ionic",
--- a/lerna.json
+++ b/lerna.json
@@ -3,5 +3,5 @@
     "core",
     "packages/*"
   ],
-  "version": "8.5.7"
+  "version": "8.5.8"
 }
--- a/packages/angular-server/CHANGELOG.md
+++ b/packages/angular-server/CHANGELOG.md
@@ -3,6 +3,14 @@
 All notable changes to this project will be documented in this file.
 See [Conventional Commits](https://conventionalcommits.org) for commit guidelines.
 
+## [8.5.8](https://github.com/ionic-team/ionic-framework/compare/v8.5.7...v8.5.8) (2025-05-28)
+
+**Note:** Version bump only for package @ionic/angular-server
+
+
+
+
+
 ## [8.5.7](https://github.com/ionic-team/ionic-framework/compare/v8.5.6...v8.5.7) (2025-05-07)
 
 **Note:** Version bump only for package @ionic/angular-server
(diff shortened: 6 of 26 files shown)


thetaPC:
Thank you for submitting the PR! However, we require issues to be submitted first and then tied to the PR for us to proceed with reviewing.

kumibrr:
> Thank you for submitting the PR! However, we require issues to be submitted first and then tied to the PR for us to proceed with reviewing.

Done!

core/src/components/modal/animations/md.enter.ts
thetaPC, line 1: We need to update this [comment](https://github.com/ionic-team/ionic-framework/blob/main/core/src/components/modal/animations/md.enter.ts#L62-L63) to reflect the changes. Maybe something like:

```diff
- * flickering. The original footer is hidden until the modal
- * is dismissed. This maintains the animation of the footer
+ * flickering. The cloned footer is hidden until the modal
+ * is moving. This maintains the animation of the footer
```

core/src/components/modal/animations/ios.enter.ts
thetaPC, line 1: We need to update this [comment](https://github.com/ionic-team/ionic-framework/blob/main/core/src/components/modal/animations/ios.enter.ts#L60-L61) to reflect the changes. Maybe something like:

```diff
- * flickering. The original footer is hidden until the modal
- * is dismissed. This maintains the animation of the footer
+ * flickering. The cloned footer is hidden until the modal
+ * is moving. This maintains the animation of the footer
```

core/src/components/modal/gestures/sheet.ts
```diff
@@ -462,6 +452,7 @@ export const createSheetGesture = (
         .onFinish(
           () => {
             if (shouldRemainOpen) {
+              swapFooterVisibility('original');
```
thetaPC, line 455: Let's make sure to keep the comment and the check.

```suggestion
              /**
               * If expandToScroll is disabled, we need to swap
               * the footer visibility to the original, so if the modal
               * is dismissed, the footer dismisses with the modal
               * and doesn't stay on the screen after the modal is gone.
               */
              if (!expandToScroll) {
                swapFooterVisibility('original');
              }
```

thetaPC (changes requested):
Let's also add a test to prevent a regression under `modal/test/sheet/modal.e2e.ts`. I recommend using this [test](https://github.com/ionic-team/ionic-framework/blob/4317da080ca26a27575a7f7905d2703c89b685cc/core/src/components/modal/test/sheet/modal.e2e.ts#L149) as a base. Place the new test under the [rendering group](https://github.com/ionic-team/ionic-framework/blob/4317da080ca26a27575a7f7905d2703c89b685cc/core/src/components/modal/test/sheet/modal.e2e.ts#L5).

core/src/components/modal/test/sheet/modal.e2e.ts
```diff
@@ -22,6 +22,46 @@ configs({ directions: ['ltr'] }).forEach(({ title, screenshot, config }) => {
         animations: 'allow',
       });
     });
+    test('should have working eventListener in ion-footer when expandToScroll is false', async ({ page }) => {
```
thetaPC, line 25:
```suggestion
    test('should have working event listener in ion-footer when expandToScroll is false', async ({ page }) => {
```

core/src/components/modal/test/sheet/modal.e2e.ts
```diff
@@ -22,6 +22,46 @@ configs({ directions: ['ltr'] }).forEach(({ title, screenshot, config }) => {
...
+      await page.goto('/src/components/modal/test/sheet', config);
```
thetaPC, line 26: We don't need to use the `goto` function since `setContent` is being used.

```suggestion
```

core/src/components/modal/test/sheet/modal.e2e.ts
```diff
@@ -22,6 +22,46 @@ configs({ directions: ['ltr'] }).forEach(({ title, screenshot, config }) => {
...
+        config
+      );
+
+      const ionModalDidPresent = await page.spyOnEvent('ionModalDidPresent');
+      const ionModalDidDismiss = await page.spyOnEvent('ionModalDidDismiss');
+
+      await page.click('#open-modal');
+      await ionModalDidPresent.next();
+
+      await page.click('#dismiss');
```
thetaPC, line 62: Let's keep the consistency of this file by using variables.

```suggestion
      const dismissButton = page.locator('#dismiss');
      const openButton = page.locator('#open-modal');

      const ionModalDidPresent = await page.spyOnEvent('ionModalDidPresent');
      const ionModalDidDismiss = await page.spyOnEvent('ionModalDidDismiss');

      await openButton.click();
      await ionModalDidPresent.next();

      await dismissButton.click();
```

thetaPC (changes requested):
Almost, just some minor requests!

thetaPC (approved):
LGTM, requesting more approvals

brandyscarney (changes requested):
Thank you for the PR and your continued work on this, we really appreciate it. 🙂 

I noticed an issue on `md` mode that was introduced with this fix. The footer is not visible and then becomes visible after the modal is fully presented. Here are comparison videos of the modal on `main` vs this PR:

| `main` | PR branch |
| ---| ---|
| <video src="https://github.com/user-attachments/assets/aaac9df8-46ba-4956-9ff9-5372ba2da97d"></video> | <video src="https://github.com/user-attachments/assets/cc68b366-bd09-44bf-8616-3383326a4016"></video> |

This can be reproduced on Chrome, Firefox and Safari without emulating a device. It does not happen in `ios` mode.

kumibrr:
@brandyscarney Thank you for pointing that issue. I'll work on it later this weekend.

thetaPC:
@kumibrr Let's place this PR on hold. Due to the increase number of issues submitted due to the `ion-footer`, we are going go back to the drawing board on how we can render the original footer instead of making a copy.

PR #30311, opened 2025-03-26
fix(divider): remove safe areas from horizontal padding

OS-martacarlos:
Issue number: internal

---------
The horizontal padding of the divider component was being calculated taking into account safe areas. 
However, since this component can be placed inside other components that already take these areas into account, this is not necessary.

## What is the new behavior?
- Removes the safe areas

- [ ] Yes
- [x] No

## Other information

Confirmed with design: the new divider component’s main use case is to be used inside ion-content (or one of it’s children).  As such, the responsibility of taking into account safe areas will fall on the developer (as it’s already stated in the [documentation](https://ionicframework.com/docs/api/content#safe-area-padding)). This means, the divider component itself will not account for the safe areas for its' specific padding.

The default spacing also changed from `xxsmall` to `medium`


core/src/components/divider/divider.common.scss
```diff
@@ -18,13 +18,13 @@
 
   /* stylelint-disable */
```
brandyscarney, line 19: Since we don't need to account for the safe area here, we can use the regular padding mixin:

```scss
@include padding-horizontal(var(--padding-start), var(--padding-end));
```

This mixin will automatically handle the `ltr`/`rtl` flip.

Also make sure to remove the stylelint comments

brandyscarney (approved):

core/src/components/divider/divider.common.scss
OS-martacarlos: will do, thank you!

BenOsodrac (approved):

PR #30328, opened 2025-04-01
fix(segment-button): ensure ion-text inside segment button will have the same visuals as ion-label

JoaoFerreira-FrontEnd:
Issue number: internal

---------

- The segment button is expecting an ion-label, however on OutSystems we were using the ion-text, since the experience is using the Mobile Text widget.

## What is the new behavior?

- add styles for ion-text in every instance we were customize ion-label;

- [ ] Yes
- [x] No

### DIFF
--- a/core/src/components/segment-button/segment-button.ionic.scss
+++ b/core/src/components/segment-button/segment-button.ionic.scss
@@ -44,7 +44,7 @@
   gap: globals.$ion-space-100;
 }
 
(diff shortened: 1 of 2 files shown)


brandyscarney:
Can we close this since it isn't being implemented this way now?

brandyscarney:
Closing since the related issue was discarded

PR #30337, opened 2025-04-04
fix(input): apply position to .cloned-input only for native themes

OS-martacarlos:

---------

Using the ionic theme, `ion-input` inside an `ion-accordion` , that require a scroll to focus, when clicked , it's text would shift

Screenshot 2025-04-04 at 12 32 22

After some troubleshooting, we identified this visual bug came from logic of `addClone`, which creates a clone of the input , in order for it to be focusable and working on iOS/Android (code [here](https://github.com/ionic-team/ionic-framework/blob/4317da080ca26a27575a7f7905d2703c89b685cc/core/src/utils/input-shims/hacks/common.ts#L34) )

The cloned input has an additional class, `cloned-input`  and, for the ionic theme, the `@position` mixin was causing the cloned input `inset-inline-start` to be 0, which meant it was not being placed _exactly_ on top of the "OG" input
 
## What is the new behavior?
There's no longer a visual bug on inputs inside accordions, when using the ionic theme

To fix it, we split the `cloned-input` class between the `input.native` and `input.common` scss files, where the `@position` mixin is only used in `input.native`

- [ ] Yes
- [x] No

## Other information

- Bug reproduction [here](https://github.com/OutSystems/ui-comp-ionic-sandboxes/tree/ROU-11748-bug)
- With this fix [here](https://github.com/OutSystems/ui-comp-ionic-sandboxes/tree/ROU-11748)


thetaPC (approved):
LGTM

OS-giulianasilva (approved):

PR #30349, opened 2025-04-10
fix(toggle): ensure proper visual selection when navigating via VoiceOver in Safari

ShaneK:

---------

Currently, MacOS voice over on Safari does not recognize ion-toggle correctly and fails to highlight the element properly

## What is the new behavior?

By adding the role property to the host element, we're correctly identifying the toggle so Safari knows how to handle it.

- [ ] Yes
- [X] No

### DIFF
--- a/core/src/components/toggle/toggle.scss
+++ b/core/src/components/toggle/toggle.scss
@@ -31,8 +31,6 @@
 
   max-width: 100%;
 
-  outline: none;
-
   cursor: pointer;
   user-select: none;
   z-index: $z-index-item-input;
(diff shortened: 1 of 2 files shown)


thetaPC (commented):
Seems to be working. There is no focused state on toggle like checkbox has, which makes it hard to determine if I'm truly on the the component. How much work would it take to try to add the focus ring that checkbox has?

Screenshot 2025-04-10 at 1 41 15 PM

ShaneK:
Turns out, it was pretty easy. Was worried that I'd need to update screenshots, but nope.

thetaPC (approved):
LGTM

brandyscarney (approved):

PR #30371, opened 2025-04-24
refactor(toolbar): improve slot layout and visibility handling

ShaneK:

---------

Currently, the Ionic theme toolbar uses `position: absolute` in modals to center its text or just `text-align: center` in other scenarios. This can lead to issues with alignment that can cause overlapping elements.

image

## What is the new behavior?

This new behavior reworks how slots get rendered and the content gets placed in the toolbar. It no longer relies on `host-context` at all (which doesn't work in Firefox) and instead uses `flex` to lay out the toolbar and responsively react to the content around it.

This does, however, have the drawback of preventing us from properly centering the content of the toolbar with just CSS, so there's new JavaScript that checks for content in certain slots and shows/hides slots based on whether or not there's anything in them, to allow the toolbar content to properly center itself. 

https://github.com/user-attachments/assets/5097307f-e11c-478d-af93-f3d998958c71

All of this combined leads to a much nicer toolbar experience that's actually responsive and works for any type of content - not just ion-titles.

### Screenshots
Many of the screenshots had to be updated because the previous version of the toolbar was absolutely positioned in most cases and had a 3px top and bottom padding. This change removes that padding and makes it more closely align with the Figma mocks.

- [ ] Yes
- [X] No

### Preview:
[Toolbar](https://ionic-framework-git-next-branch-fw-6439-ionic1.vercel.app/src/components/toolbar/test/basic?ionic:theme=ionic)


core/src/components/segment/test/toolbar/index.html
```diff
@@ -0,0 +1,160 @@
+<!DOCTYPE html>
+<html lang="en" dir="ltr">
+  <head>
+    <meta charset="UTF-8" />
+    <title>Segment - Basic</title>
```
brandyscarney, line 5:
```suggestion
    <title>Segment - Toolbar</title>
```

core/src/components/toolbar/toolbar.ionic.scss
brandyscarney, line 1: Padding start & end are different:

Screenshot 2025-04-28 at 10 45 22 AM

Screenshot 2025-04-28 at 10 46 51 AM

core/src/components/toolbar/toolbar.tsx
```diff
@@ -88,19 +147,20 @@ export class Toolbar implements ComponentInterface {
   }
 
   render() {
-    const theme = getIonTheme(this);
+    const theme = getIonTheme(this.el);
```
brandyscarney, line 150:
```suggestion
    const theme = getIonTheme(this);
```

Is this change necessary? I see the same theme logged checking `this`.

core/src/components/toolbar/toolbar.tsx
```diff
@@ -58,6 +58,65 @@ export class Toolbar implements ComponentInterface {
...
+    this.el.classList.remove(...classesToRemove);
+  }
+
+  private hasSlotContent(slotName: string): boolean {
+    const slotNode = this.el.shadowRoot?.querySelector(`slot[name="${slotName}"]`) as HTMLSlotElement | null;
+    if (slotNode) {
+      const assigned = slotNode.assignedNodes();
+      return assigned.length > 0;
+    }
+    return false;
```
brandyscarney, line 119:
```suggestion
    return !!slotNode && slotNode.assignedNodes().length > 0;
```

core/src/components/toolbar/toolbar.tsx
```diff
@@ -58,6 +58,61 @@ export class Toolbar implements ComponentInterface {
...
+    // of slots on either side of the toolbar.
+    const conditions = [
+      { name: 'end', requiredSlots: ['start'], excludeSlots: ['end', 'primary'] },
+      { name: 'start', requiredSlots: ['end'], excludeSlots: ['start', 'secondary'] },
+      { name: 'secondary', requiredSlots: ['primary'], excludeSlots: ['secondary', 'start'] },
+      { name: 'primary', requiredSlots: ['secondary'], excludeSlots: ['primary', 'end'] },
+    ];
+    conditions.forEach((condition) => {
+      const hasRequiredSlots = condition.requiredSlots.every((slot) => classesToAdd.includes(`has-${slot}-content`));
+      const hasExcludedSlots = condition.excludeSlots.some((slot) => classesToAdd.includes(`has-${slot}-content`));
```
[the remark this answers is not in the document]
ShaneK: I intentionally didn't want to call `hasSlotContent` again, since it has to query the DOM it's a bit expensive. Checking for a string in an array is much quicker.

core/src/components/toolbar/toolbar.tsx
```diff
@@ -58,6 +58,61 @@ export class Toolbar implements ComponentInterface {
...
+      const className = `show-${condition.name}`;
+
+      if (hasRequiredSlots && !hasExcludedSlots) {
+        classesToAdd.push(className);
+      } else {
+        this.el.classList.remove(className);
```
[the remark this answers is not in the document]
ShaneK: Good catch, looks like I missed a spot in refactoring

core/src/components/toolbar/toolbar.ionic.scss
```diff
@@ -62,31 +53,42 @@ $toolbar-order-ionic: (
 // Ionic Toolbar Slot Placement
 // --------------------------------------------------
 
-::slotted([slot="secondary"]) {
-  order: map-get($toolbar-order-ionic, slot-secondary);
-}
+:host(.has-end-content) slot[name="end"],
+:host(.show-end) slot[name="end"] {
```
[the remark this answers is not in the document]
brandyscarney: Oh okay I see. I was trying to follow the logic but it's super confusing. 😂 Looks good then!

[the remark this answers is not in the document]
ShaneK: You could, in theory, consolidate these, but for me it's much easier to wrap my head around it with them being separate, which should mean it's a lot more readable as it is now than if we tried to combine their purposes. There's no real benefit to having them combined other than one less selector, but the code would become more difficult to understand in a way that isn't necessarily worth the trade off, in my opinion.

core/src/components/modal/test/shape/modal.e2e.ts-snapshots/modal-shape-sheet-default-ionic-md-ltr-light-Mobile-Chrome-linux.png
brandyscarney, line 1: It looks like the title here isn't centered anymore.

ShaneK: I reworked the way things get centered to better enforce centering contents by making the hidden slots take up at least as much width as their opposite slots to fix this. It seems to have addressed it, but if you just click the link above you'll be taken to the old commit, you'll need to go find it by name to see it now, or click here:
https://github.com/ionic-team/ionic-framework/pull/30371/files#diff-6ba5b71b8e40b5f087c012f294565bf487d1f1b5fd50ff4f18c3c12ad8e973bf

brandyscarney, line 1: Good fix!

title-center

core/src/components/toolbar/toolbar.ionic.scss
```diff
@@ -61,32 +52,53 @@ $toolbar-order-ionic: (
...
+  display: flex;
 
-::slotted([slot="primary"]) {
-  order: map-get($toolbar-order-ionic, slot-primary);
+  flex: 1 0 var(--start-end-size, 0);
+  justify-content: flex-end;
 
   text-align: end;
+  gap: globals.$ion-space-400;
```
brandyscarney, line 76:
```suggestion

```

This gap does nothing, you need to apply it to `ion-buttons`

core/src/components/toolbar/toolbar.ionic.scss
```diff
@@ -61,32 +52,53 @@ $toolbar-order-ionic: (
...
 }
 
-::slotted([slot="start"]) {
-  order: map-get($toolbar-order-ionic, slot-start);
+:host(.has-start-content) slot[name="start"],
+:host(.show-start) slot[name="start"] {
+  display: flex;
+
+  flex: 1 0 var(--start-end-size, 0);
+  gap: globals.$ion-space-400;
```
brandyscarney, line 84:
```suggestion

```

This gap does nothing, you need to apply it to `ion-buttons`

core/src/components/toolbar/toolbar.ionic.scss
```diff
@@ -61,32 +52,53 @@ $toolbar-order-ionic: (
...
+:host(.show-primary) slot[name="primary"] {
   display: flex;
 
-  flex: 1;
-  gap: globals.$ion-space-200;
+  flex: 1 0 var(--primary-secondary-size, 0);
+  justify-content: flex-end;
+
+  text-align: end;
+  gap: globals.$ion-space-400;
```
brandyscarney, line 95:
```suggestion

```

This gap does nothing, you need to apply it to `ion-buttons`

core/src/components/toolbar/toolbar.ionic.scss
```diff
@@ -61,32 +52,53 @@ $toolbar-order-ionic: (
...
-::slotted([slot="end"]) {
-  justify-content: end;
-  order: map-get($toolbar-order-ionic, slot-end);
+:host(.has-secondary-content) slot[name="secondary"],
+:host(.show-secondary) slot[name="secondary"] {
+  display: flex;
 
-  font-size: globals.$ion-font-size-600;
+  flex: 1 0 var(--primary-secondary-size, 0);
+  gap: globals.$ion-space-400;
```
brandyscarney, line 103:
```suggestion

```

This gap does nothing, you need to apply it to `ion-buttons`

core/src/components/toolbar/toolbar.ionic.scss
```diff
@@ -31,8 +23,7 @@ $toolbar-order-ionic: (
 // --------------------------------------------------
 
 .toolbar-content {
-  flex: 2;
-  order: map-get($toolbar-order-ionic, content);
+  flex: 1 1 auto;
 
   min-width: 0;
 }
```
brandyscarney, line 29:
```suggestion
}

::slotted(ion-buttons) {
  gap: globals.$ion-space-200;
}
```

This will give us the gap between the buttons:

Screenshot 2025-04-29 at 3 27 58 PM

Alternatively we can add this to `buttons.ionic.scss`.

core/src/components/toolbar/toolbar.ionic.scss
brandyscarney, line 1: Why is there so much unused space here?

Screenshot 2025-04-29 at 3 30 32 PM
Screenshot 2025-04-29 at 3 31 13 PM

brandyscarney (changes requested):

core/src/components/toolbar/toolbar.ionic.scss
ShaneK: These are an artifact from when slotted content was being forced to flex, I've removed that. I don't think changing ion-buttons is in the scope of this PR

core/src/components/toolbar/toolbar.ionic.scss
ShaneK: Done through the toolbar.ionic.scss

core/src/components/toolbar/toolbar.ionic.scss
ShaneK: flex-grow was allowing the ion-buttons to grow before their size was being measured. I've fixed that by disabling flex-grow.

brandyscarney (approved):
This looks and works great! 🎉 Smart solution to a complex problem! In the future, we might want to explore relying more on Stencil for slot updates and width detection, but I don't think we need to hold this up. 👍

thetaPC (approved):
LGTM, great job!!!!

We really should consider not using `ion-header` in our test pages to avoid so many screenshot changes.

ShaneK:
@thetaPC we have a card to crop it out of the screenshots, it's entirely possible to make the screenshots the relevant components that are actually being tested so this doesn't happen again. Check out FW-6505

PR #30386, opened 2025-04-30
feat(input-otp): add new input-otp component

brandyscarney:
Issue number: resolves #

---------

## What is the new behavior?

-
-
-

- [ ] Yes
- [ ] No


exorex:
Thank you, @brandyscarney, for your work on this component.

One feature suggestion: pressing CTRL+A (or CMD+A on macOS) inside the component could select and clear all input fields at once. This would align with common input behavior and improve UX for users needing to re-enter their code quickly.

Example: https://input-otp.rodz.dev/

core/src/components/input-otp/input-otp.tsx
```diff
@@ -0,0 +1,451 @@
...
+
+import type { InputOtpChangeEventDetail, InputOtpCompleteEventDetail } from './input-otp-interface';
+
+@Component({
+  tag: 'ion-input-otp',
+  styleUrls: {
+    ios: 'input-otp.ios.scss',
+    md: 'input-otp.md.scss',
+  },
+  scoped: true,
```
christian-bromann, line 16: What is the reason for making this a scoped component vs a shadow one?

core/src/components/input-otp/input-otp.tsx
```diff
@@ -0,0 +1,451 @@
...
+  }
+
+  private handlePaste(event: ClipboardEvent) {
+    const { inputRefs, length, validKeys } = this;
+
+    event.preventDefault();
+    const pastedText = event.clipboardData?.getData('text') || '';
+    const validChars = pastedText
+      .split('')
+      .filter((char) => validKeys.test(char.toLowerCase()))
```
christian-bromann, line 250: Why using `toLowerCase` here? It seems like `validKeys` considers lower and upper casing via `/^[a-zA-Z0-9]$/i;`, when using a custom pattern, this may be confusing, no?

core/src/components/input-otp/input-otp.tsx
```diff
@@ -0,0 +1,451 @@
...
+      // Let all tab events proceed normally
+      return;
+    }
+  }
+
+  private handlePaste(event: ClipboardEvent) {
+    const { inputRefs, length, validKeys } = this;
+
+    event.preventDefault();
+    const pastedText = event.clipboardData?.getData('text') || '';
```
christian-bromann, line 247: How about returning early here if nothing is in the clipboard?
```suggestion
    const pastedText = event.clipboardData?.getData('text');
    if (typeof pastedText === 'undefined') {
       return
    }
```

christian-bromann (approved):
Some minor nits

core/src/components/input-otp/input-otp.tsx
brandyscarney: https://github.com/ionic-team/ionic-framework/pull/30386/commits/46a3fb7db9606b417182577781169ce6555c4b08

core/src/components/input-otp/input-otp.tsx
brandyscarney: This was a bug, it shouldn't have been case insensitive, fixed: https://github.com/ionic-team/ionic-framework/pull/30386/commits/97649a006b35a7cde236857473961d96c503fa8b

core/src/components/input-otp/input-otp.tsx
brandyscarney: I chose to make it a `scoped` component to match [input](https://github.com/ionic-team/ionic-framework/blob/7d639b0412120523f758942c855cb69f9a52e9d9/core/src/components/input/input.tsx#L44). In our [latest research](https://github.com/ionic-team/ionic-framework-design-documents/blob/main/projects/ionic-framework/components/input/0005-shadow-dom.md) we found that we couldn't convert input to shadow because there are a couple of properties that don't work with shadow. One of them being `autocomplete` which is used [here](https://github.com/ionic-team/ionic-framework/blob/97649a006b35a7cde236857473961d96c503fa8b/core/src/components/input-otp/input-otp.tsx#L473). I don't think we are able to move components using native `input` elements to shadow yet.

core/.eslintrc.js
brandyscarney, line 1: This was required in order to use `<>` in `input-otp.tsx1`

core/tsconfig.json
brandyscarney, line 1: This was required in order to use `<>` in `input-otp.tsx1`

packages/angular/src/index.ts
brandyscarney, line 1: I don't know why input only emits `InputCustomEvent` and `InputChangeEventDetail` here when it also has `InputInputEventDetail`. Something to look into.

packages/angular/standalone/src/index.ts
brandyscarney, line 1: I don't know why input only emits `InputCustomEvent` and `InputChangeEventDetail` here when it also has `InputInputEventDetail`. Something to look into.

packages/vue/src/index.ts
brandyscarney, line 1: I don't know why input only emits `InputCustomEvent` and `InputChangeEventDetail` here when it also has `InputInputEventDetail`. Something to look into.

packages/react/src/components/index.ts
brandyscarney, line 1: I don't know why input emits `InputCustomEvent` and `InputChangeEventDetail` and `InputInputEventDetail` here when the other frameworks don't emit `InputInputEventDetail`. Something to look into.

core/src/components/input-otp/input-otp.ios.scss
```diff
@@ -0,0 +1,20 @@
...
+// iOS Input OTP
+// --------------------------------------------------
+
+:host {
+  --border-width: #{$hairlines-width};
+}
+
+:host(.has-focus) .native-input:focus {
+  --border-width: 1px;
+}
```
brandyscarney, line 13: These borders will only be visible on devices that support hairline (half pixel) border widths

core/src/components/input-otp/test/size/input-otp.e2e.ts-snapshots/input-otp-large-collapsed-ios-ltr-Mobile-Chrome-linux.png
brandyscarney, line 1: The cut off boxes in the "collapsed" tests are intentional - this is testing that the input width collapses to 40px when there is not enough room

core/src/components/input-otp/test/shape/input-otp.e2e.ts-snapshots/input-otp-rectangular-large-ios-ltr-Mobile-Safari-linux.png
brandyscarney, line 1: The issue with the borders not showing up in Safari seems to be a Playwright issue. Notice this only happens for the rectangular shape which has a `0px` `border-radius` - the other shapes show up fine. I tried messing around with the device we use with different but no perfect results. If you open the test locally in Safari the borders do show up.

core/src/components/input-otp/input-otp.tsx
```diff
@@ -0,0 +1,786 @@
...
+  },
+  scoped: true,
+})
+export class InputOTP implements ComponentInterface {
+  private inputRefs: HTMLInputElement[] = [];
+  private inputId = `ion-input-otp-${inputIds++}`;
+  private parsedSeparators: number[] = [];
+
+  /**
+   * The value of the OTP input when it is focused.
```
ShaneK, line 31:
```suggestion
   * Stores the initial value of the input when it receives focus.
   * Used to determine if the value changed during the focus session
   * to avoid emitting unnecessary change events on blur.
```

ShaneK (changes requested):
Looks really good so far! I requested some changes and I think VoiceOver has some issues. I didn't test with any other a11y systems though since I don't really have an easy way to do that, but I imagine they'll act pretty similarly.

core/src/components/input-otp/input-otp.tsx
```diff
@@ -0,0 +1,786 @@
...
+      this.inputRefs[tabbableIndex]?.focus();
+    }
+  }
+
+  /**
+   * Get the regex pattern for allowed characters.
+   * If a pattern is provided, use it to create a regex pattern
+   * Otherwise, use the default regex pattern based on type
+   */
+  private get validKeys(): RegExp {
```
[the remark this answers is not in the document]
brandyscarney: https://github.com/ionic-team/ionic-framework/pull/30386/commits/26349c555ad5be1a0d103171d907729773adfddf

core/src/components/input-otp/input-otp.tsx
```diff
@@ -0,0 +1,786 @@
...
+  @Element() el!: HTMLIonInputOtpElement;
+
+  @State() private inputValues: string[] = [];
+  @State() hasFocus = false;
+
+  /**
+   * Indicates whether and how the text value should be automatically capitalized as it is entered/edited by the user.
+   * Available options: `"off"`, `"none"`, `"on"`, `"sentences"`, `"words"`, `"characters"`.
+   */
+  @Prop() autocapitalize = 'off';
```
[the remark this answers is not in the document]
brandyscarney: I was actually going to do this but I noticed that input [does not](https://github.com/ionic-team/ionic-framework/blob/00beab4af8cfa9a2cfea833eeed6b28dba53b311/core/src/components/input/input.tsx#L95). I found this strange considering we do set the accepted list for [autocomplete](https://github.com/ionic-team/ionic-framework/blob/00beab4af8cfa9a2cfea833eeed6b28dba53b311/core/src/components/input/input.tsx#L100C11-L100C23) and [autocorrect](https://github.com/ionic-team/ionic-framework/blob/00beab4af8cfa9a2cfea833eeed6b28dba53b311/core/src/components/input/input.tsx#L105). I didn't want to make input and input-otp different though since it might be confusing if people are using both. Maybe we can make an issue to update the type for both in a major release?

core/src/components/input-otp/input-otp.tsx
```diff
@@ -0,0 +1,786 @@
...
+    // text is selected and let the onPaste / onInput handler manage it
+    if (isPasteShortcut || isTextSelection) {
+      return;
+    }
+
+    if (event.key === 'Backspace') {
+      if (this.inputValues[index]) {
+        // Remove the value at the current index
+        this.inputValues[index] = '';
+
```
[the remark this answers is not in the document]
brandyscarney: https://github.com/ionic-team/ionic-framework/pull/30386/commits/eb7f8382b8accbd4609273e5932d4f324fa0d8d0

core/src/components/input-otp/input-otp.tsx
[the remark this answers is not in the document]
ShaneK: I don't know that we need to tie this component's autocomplete to the autocomplete of input, I can't see the value in that. I do agree we should probably make a ticket to update the one for input in a major though, if there's no value in arbitrary string input for this prop.

core/src/components/input-otp/input-otp.tsx
```diff
@@ -0,0 +1,786 @@
...
+  /**
+   * Gets the string pattern to pass to the input element
+   * and use in the regex for allowed characters.
+   */
+  private getPattern(): string {
+    const { pattern, type } = this;
+    if (pattern) {
+      return pattern;
+    }
+    return type === 'number' ? '[0-9]' : '[a-zA-Z0-9]';
```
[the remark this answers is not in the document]
brandyscarney: Great catch! Could you take a look at this change and the tests to see if there are any use cases I missed: https://github.com/ionic-team/ionic-framework/pull/30386/commits/2a9c274821463e3e51886cabb5b15741005abb03

core/src/components/input-otp/input-otp.tsx
```diff
@@ -0,0 +1,786 @@
...
+   */
+  private getFirstEmptyIndex() {
+    const { inputValues, length } = this;
+
+    for (let i = 0; i < length; i++) {
+      if (!inputValues[i] || inputValues[i] === '') {
+        return i;
+      }
+    }
+    return -1;
```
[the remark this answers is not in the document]
brandyscarney: I played around with this but this code breaks the automatic focus changing when typing in the input box. I can change it to use `findIndex` though but I had to create a new array to pad the empty values with `''`. This seems to be passing all tests. Are you okay with this? https://github.com/ionic-team/ionic-framework/pull/30386/commits/ec35637753589426f4283bef707e2a7383be8d4d

core/src/components/input-otp/input-otp.tsx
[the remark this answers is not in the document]
brandyscarney: Took a look at this again and it appears it's because [autocapitalize](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Global_attributes/autocapitalize) is a global attribute so trying to set it in a different type results in this error:

```
[ ERROR ]  TypeScript: src/components.d.ts:4012:15
           Interface 'HTMLIonInputOtpElement' cannot simultaneously
           extend types 'IonInputOtp' and 'HTMLStencilElement'.Named
           property 'autocapitalize' of types 'IonInputOtp' and
           'HTMLStencilElement' are not identical.
```

core/src/components/input-otp/input-otp.tsx
```diff
@@ -0,0 +1,782 @@
...
+  /**
+   * Gets the string pattern to pass to the input element
+   * and use in the regex for allowed characters.
+   */
+  private getPattern(): string {
+    const { pattern, type } = this;
+    if (pattern) {
+      return pattern;
+    }
+    return type === 'number' ? '[0-9]' : '[\\p{L}\\p{N}]';
```
ShaneK, line 321:
```suggestion
    return type === 'number' ? '[\p{N}]' : '[\\p{L}\\p{N}]';
```
Very close! Missed this one though, which should allow for non-latic numbers with the number filter.

ShaneK (changes requested):
Very close! Just missed one thing

core/src/components/input-otp/input-otp.tsx
brandyscarney: https://github.com/ionic-team/ionic-framework/pull/30386/commits/1c9c939eef9a2c522f82d77cf98543b2fc45dc5d

ShaneK (approved):

packages/angular/test/base/src/app/lazy/inputs/inputs.component.html
```diff
@@ -71,6 +71,16 @@
       <ion-note slot="end">{{input}}</ion-note>
     </ion-item>
 
+    <ion-item>
+      <ion-input-otp [(ngModel)]="inputOtp">Input OTP</ion-input-otp>
+      <ion-note slot="end">{{inputOtp}}</ion-note>
+    </ion-item>
+
+    <ion-item color="dark">
+      <ion-input-otp [(ngModel)]="inputOtp">Input OTP Mirror</ion-input-otp>
```
thetaPC, line 80: Should this also mirror the backspace? The value updates when I enter a value but when I delete a value, the mirror doesn't reflect.

packages/angular/test/base/src/app/lazy/inputs/inputs.component.html
```diff
@@ -71,6 +71,16 @@
...
+      <ion-note slot="end">{{inputOtp}}</ion-note>
+    </ion-item>
```
thetaPC, line 82: The reset button doesn't seem to do anything with these otp fields. I would expect the fields to be cleared.

packages/angular/test/base/src/app/lazy/form/form.component.html
```diff
@@ -44,6 +44,20 @@
...
 
+      <ion-item>
+        <ion-input-otp id="touched-input-otp-number-test" formControlName="inputOtp" class="required">Input OTP (required)</ion-input-otp>
+      </ion-item>
+
+      <ion-item>
+        <ion-input-otp id="touched-input-otp-text-test" type="text" formControlName="inputOtpText" class="required">Input OTP Text (required)</ion-input-otp>
+      </ion-item>
+
+      <ion-button id="input-otp-touched" (click)="setOtpTouched()">Set Input OTP Touched</ion-button>
```
thetaPC, line 55: Let's move this button to be under the first otp for clarity. I thought it was going to target the otp with the form control name of `inputOtpText`.

thetaPC (changes requested):

core/.eslintrc.js
```diff
@@ -26,7 +26,7 @@ module.exports = {
     "@typescript-eslint/no-unused-vars": [
       "warn",
       {
-        "varsIgnorePattern": "^h$"
+        "varsIgnorePattern": "^(h|Fragment)$"
```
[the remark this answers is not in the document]
brandyscarney: Yes my comments were a bit lost in the feedback: https://github.com/ionic-team/ionic-framework/pull/30386/files/a46322bacf9cc4d838d741f00b2b43eec5fb87af#diff-04255a712cddfa496ba085165b80d1e5a7dc244bce96e1873982b25667e9c5d6

core/src/components/input-otp/input-otp.tsx
```diff
@@ -0,0 +1,782 @@
...
+                  onFocus={this.onFocus(index)}
+                  onKeyDown={this.onKeyDown(index)}
+                  onPaste={this.onPaste}
+                />
+              </div>
+              {this.showSeparator(index) && <div class="input-otp-separator" />}
+            </>
+          ))}
+        </div>
+        <div class="input-otp-description">
```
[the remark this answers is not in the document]
brandyscarney: https://github.com/ionic-team/ionic-framework/pull/30386#discussion_r2085700793

core/.eslintrc.js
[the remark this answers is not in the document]
thetaPC: 🤦‍♀️

core/src/components/input-otp/input-otp.tsx
```diff
@@ -0,0 +1,782 @@
...
+   * - Tab: Allows normal tab navigation between components
+   *
+   * Input Behavior:
+   * - Validates input against the allowed pattern
+   * - When entering a key in a filled box:
+   *   - Shifts existing values right if there is room
+   *   - Updates the value of the input group
+   *   - Prevents default behavior to avoid automatic focus shift
+   */
+  private onKeyDown = (index: number) => (event: KeyboardEvent) => {
```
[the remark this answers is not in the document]
brandyscarney: These were all decided as nice to have but not requirements to release in the initial rollout:

**Nice to Have:**
- Support a placeholder
- Support selecting all (CMD/CTRL + A)
  - Allow backspace to clear or entering text to erase and start over
- Support hover state

core/tsconfig.json
```diff
@@ -10,6 +10,7 @@
     "assumeChangesOnlyAffectDirectDependencies": true,
     "jsx": "react",
     "jsxFactory": "h",
+    "jsxFragmentFactory": "Fragment",
```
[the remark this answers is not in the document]
brandyscarney: Yes my comments were a bit lost in the feedback: https://github.com/ionic-team/ionic-framework/pull/30386/files/a46322bacf9cc4d838d741f00b2b43eec5fb87af#r2085383481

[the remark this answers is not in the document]
thetaPC: 🤦‍♀️ x2

core/src/components/input-otp/test/shape/input-otp.e2e.ts-snapshots/input-otp-rectangular-large-ios-ltr-Mobile-Safari-linux.png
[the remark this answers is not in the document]
thetaPC: 🤦‍♀️ x3. Worst part is that I read all the comments before reviewing... I blame it on sleepy brain.

core/src/components/input-otp/input-otp.tsx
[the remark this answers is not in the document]
brandyscarney: When I first spoke with design they told me this component doesn't have helper and error text and the developers have to add it themselves in the description. However, it does look like they might have updated it so it's possible they are expecting it to be automatically handled now. I will reach out to them to confirm the component does not require helper/error text.

core/src/components/input-otp/input-otp.tsx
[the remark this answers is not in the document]
brandyscarney: So we can't change it to where it doesn't render (at least not easily) due to it wrapping the default `<slot>` but I took a look at how other components handle it not being passed (toggle, checkbox, etc.) and they just add `display: none` when it is empty. I implemented that here: https://github.com/ionic-team/ionic-framework/pull/30386/commits/9d905b2b943453b28d79d9f33df99038534bc125

core/src/components/input-otp/input-otp.tsx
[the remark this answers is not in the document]
brandyscarney: Design confirmed that this component does not flip between helper & error text like the other components:

> When it's set to “Mandatory” and there's an error, the error message appears below the input—but it doesn’t replace the helper text; both are shown at the same time.
This behavior is specific to this component—we want to display both the helper message and the error message together, rather than having one replace the other like in other inputs.

core/src/components/input-otp/test/color/input-otp.e2e.ts-snapshots/input-otp-color-solid-ios-ltr-Mobile-Chrome-linux.png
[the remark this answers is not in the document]
brandyscarney: Unfortunately this isn't possible with the current implementation. The `ion-input-otp` has to have the `.has-focus` class which is easy to set on load, but it also depends on the native input having `:focus`. Since only one element can be focused at a time, this can't be applied to multiple inputs at once.

core/src/components/input-otp/input-otp.tsx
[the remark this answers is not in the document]
brandyscarney: Yeah I think their thought process is that if you have the following you wouldn't want it to change to an error message because they still need the link:

Screenshot 2025-05-23 at 1 19 04 PM

packages/angular/test/base/src/app/lazy/inputs/inputs.component.html
brandyscarney: Great catch! This was a bug in how the values were being handled. Added a fix & tests here: https://github.com/ionic-team/ionic-framework/pull/30386/commits/4015ee4fea8cb6d6b17ef935aecfb7919137a09d

packages/angular/test/base/src/app/lazy/form/form.component.html
brandyscarney: Moved in https://github.com/ionic-team/ionic-framework/pull/30386/commits/4015ee4fea8cb6d6b17ef935aecfb7919137a09d

core/src/components/input-otp/test/basic/index.html
```diff
@@ -0,0 +1,125 @@
...
+        <ion-toolbar>
+          <ion-title>Input OTP - Basic</ion-title>
+        </ion-toolbar>
+      </ion-header>
+
+      <ion-content class="ion-padding">
+        <div class="grid">
+          <div class="grid-item">
+            <h2>Default</h2>
+            <ion-input-otp> Didn't get a code? <a href="#">Resend the code</a> </ion-input-otp>
```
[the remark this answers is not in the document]
brandyscarney: Can you tell me what browser you used or how you reproduced this? I am not able to reproduce in Chrome or Firefox:

https://github.com/user-attachments/assets/12df3bac-3ae3-4f33-a56f-78bea2fe293e

core/src/components/input-otp/test/basic/index.html
```diff
@@ -0,0 +1,125 @@
...
+            <ion-input-otp length="2"> Didn't get a code? <a href="#">Resend the code</a> </ion-input-otp>
+            <ion-input-otp id="numberValue" length="6">
```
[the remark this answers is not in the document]
brandyscarney: I took a look at what some of the other libraries do, and it seems that I am not handling this the right way. Other libraries always paste from the 1st input, no matter how many input boxes have text in them. I have updated this behavior & the tests and will update it in the documentation. Let me know if you disagree with this approach or if you'd like me to send you some examples from the other libraries!

The fix: https://github.com/ionic-team/ionic-framework/pull/30386/commits/8c29200bc97d0ce651f5d993636b46bfdf893e22

core/src/components/input-otp/test/basic/index.html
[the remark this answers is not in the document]
thetaPC: Weird, I can't replicate it anymore. 🤔 It was Firefox, maybe there was a patch on the browser.

docs/component-guide.md
thetaPC, line 1: Thank you so much!

thetaPC (approved):
LGTM 🎊

PR #30392, opened 2025-05-05
feat(badge): remove xxsmall, xsmall and xlarge sizes

OS-giulianasilva:

---------

The badge presents the `xxsmall`, `xsmall`, and `xlarge`. The user experience design expects only the `small` and `medium` sizes for badges with content (number or icon), and `small`, `medium`, and `large` sizes for empty badges.

## What is the new behavior?
- The badge should support only the sizes mentioned above.
- The `small` and `medium` sizes must be adjusted to the new requirements for badges with content.
- The size property descriptions must inform that the `large` size is only available for empty badges.

- [x] Yes
- [ ] No


core/src/components/badge/badge.ionic.scss
```diff
@@ -5,8 +5,8 @@
 // --------------------------------------------------
 
 :host {
-  --padding-start: #{globals.$ion-space-200};
-  --padding-end: #{globals.$ion-space-200};
+  --padding-start: #{globals.$ion-space-050};
+  --padding-end: #{globals.$ion-space-050};
```
ShaneK, line 9: We should remove the padding defaults and set these in the small size badge specifically so that if the defaults change in the future we don't need to do CSS updates.

core/src/components/badge/test/shape/badge.e2e.ts
```diff
@@ -69,7 +59,6 @@ configs({ directions: ['ltr'], modes: ['ionic-md'] }).forEach(({ config, screens
         <div id="container">
           <ion-badge shape="soft" size="medium">1</ion-badge>
           <ion-badge shape="soft" size="large">1</ion-badge>
```
ShaneK, line 61:
```suggestion
```
I think you meant to get rid of this, there are no styles for it and it looks broken in [the screenshots](https://github.com/ionic-team/ionic-framework/pull/30392/files#diff-3209f20c6b8bb4ff64e342bf437c92ba8e819932945a8de290e88a1bd7df7e5a). However, I do think you should still have styles for the sizes that aren't supported for a type but are choices so they don't just break like this. I think you should copy the styles for the closest one that is supported. For example, in this case, I think the `large` size should have taken on the medium styles.

ShaneK (changes requested):
Generally, we're moving towards a better way of doing defaults that makes it easier to change in the future if we need to. I've left a few change requests in specific areas, but [avatar](https://github.com/ionic-team/ionic-framework/blob/next/core/src/components/avatar/avatar.tsx) is a really good example of this paradigm in use for size.

core/src/components/badge/badge.tsx
```diff
@@ -46,16 +46,13 @@ export class Badge implements ComponentInterface {
...
-   * Set to "medium" for a medium badge.
-   * Set to "large" for a large badge.
-   * Set to `"xlarge"` for the largest badge.
+   * Set to `"medium"` for a medium badge.
+   * Set to `"large"` for a large badge, when it is empty (no text or icon).
    *
    * Defaults to `"small"` for the `ionic` theme, undefined for all other themes.
    */
-  @Prop() size?: 'xxsmall' | 'xsmall' | 'small' | 'medium' | 'large' | 'xlarge';
+  @Prop() size?: 'small' | 'medium' | 'large';
```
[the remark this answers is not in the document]
OS-giulianasilva: Do you think I should apply this default even if the [badge's `getSize()`](https://github.com/ionic-team/ionic-framework/blob/1885088f15e1f202efbcfa84a15c72a583870178/core/src/components/badge/badge.tsx#L79-L93) returns `small` when `undefined`?

core/src/components/badge/badge.ionic.scss
OS-giulianasilva: Thanks for the explanation! They were removed.

core/src/components/badge/test/shape/badge.e2e.ts
OS-giulianasilva: I agree with you! I've fixed the test and [added the `large` size](https://github.com/ionic-team/ionic-framework/blob/d62ddbdc85e0cb53f0146bb87f4fb8b8320c5768/core/src/components/badge/badge.ionic.scss#L88-L105) to have the same styles as the `medium` size.

core/src/components/badge/badge.tsx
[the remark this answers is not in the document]
ShaneK: No, I guess not. I mean, I'd rather see it defined where the prop is set than as a fallback in `getSize()`, but it needs to be an optional property for other themes to work, so there's not really a way to avoid having the undefined check

ShaneK (approved):
Looks really good, great work!

ShaneK:
Weirdly your screenshot tests seem to have failed even though you updated them after pushing 🤔 I guess maybe try updating them again? If nothing changes then run your tests again? Very odd!

BenOsodrac (approved):

PR #30393, opened 2025-05-06
feat(radio): make radio buttons full width for ionic theme

OS-martacarlos:

---------
`ion-radio`s would only use the necessary width

## What is the new behavior?
For the ionic theme:
- `ion-radio` now take up the full width, following the new design specifications ([here](https://www.figma.com/design/xSk9vrYWTlLO1uLn12Wf4Q/-GA--Mobile-UI---Dev-Experience?node-id=1236-15647&t=WuzvgKB9MCMm8VUO-4))
- `ion-radio` and `ion-radio-group` test snapshots were updated accordingly

- [ ] Yes
- [x] No


OS-martacarlos:
closing this PR as it's not necessary, the same look can be achieved using `justify`

PR #30404, opened 2025-05-09
feat(tokens): create variables fallbacks based on its alias token

BenOsodrac:
Issue number:  internal

---------

Currently, even if a token used an alias token as its value, when we're generating the variables we wouldn't respect this connection and were simply adding the fixed final value.

## What is the new behavior?

- Now when a token uses an alias token, we keep that relationship by making sure the alias variable is used as fallback.

- [ ] Yes
- [ ] No


thetaPC (approved):
LGTM, one minor suggestion but not a blocker

core/scripts/tokens/utils.js
```diff
@@ -48,6 +48,18 @@ function removeConsecutiveRepeatedWords(str) {
     return str.replace(/(\b\w+\b)(-\1)+/g, '$1');
 }
 
+// Generates a reference variable for an alias token type
```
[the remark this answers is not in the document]
BenOsodrac: Done!

thetaPC (approved):
LGTM

core/src/foundations/ionic.vars.scss
```diff
@@ -939,6 +367,578 @@ $ion-primitives-pink-1100: var(--ion-primitives-pink-1100, #8b2248);
 $ion-primitives-pink-1100-rgb: var(--ion-primitives-pink-1100-rgb, 139, 34, 72);
 $ion-primitives-pink-1200: var(--ion-primitives-pink-1200, #621833);
 $ion-primitives-pink-1200-rgb: var(--ion-primitives-pink-1200-rgb, 98, 24, 51);
+$ion-semantics-primary-100: var(--ion-semantics-primary-100, $ion-primitives-blue-100);
+$ion-semantics-primary-100-rgb: var(--ion-semantics-primary-100-rgb, 242, 244, 253);
```
brandyscarney, line 371:
```suggestion
$ion-semantics-primary-100-rgb: var(--ion-semantics-primary-100-rgb, $ion-primitives-blue-100-rgb);
```

Should this fallback to `$ion-primitives-blue-100-rgb` (and the same for all other semantic rgb vars)?

core/src/foundations/ionic.vars.scss
```diff
@@ -939,6 +367,578 @@ $ion-primitives-pink-1100: var(--ion-primitives-pink-1100, #8b2248);
...
+$ion-scale-4000: var(--ion-scale-4000, 160px);
+$ion-scale-5000: var(--ion-scale-5000, 200px);
+$ion-scale-6200: var(--ion-scale-6200, 248px);
+$ion-scale-7400: var(--ion-scale-7400, 296px);
+$ion-scale-9000: var(--ion-scale-9000, 360px);
+$ion-scale-025: var(--ion-scale-025, 1px);
+$ion-scale-050: var(--ion-scale-050, 2px);
+$ion-scale-075: var(--ion-scale-075, 3px);
+$ion-border-default: var(--ion-border-default, $ion-primitives-neutral-400);
+$ion-border-default-rgb: var(--ion-border-default-rgb, 213, 213, 213);
```
brandyscarney, line 533:
```suggestion
$ion-border-default-rgb: var(--ion-border-default-rgb, $ion-primitives-neutral-400-rgb);
```

Should this fallback to `$ion-primitives-neutral-400-rgb` (and the same for all other border rgb vars)?

core/src/foundations/ionic.vars.scss
```diff
@@ -939,6 +367,578 @@ $ion-primitives-pink-1100: var(--ion-primitives-pink-1100, #8b2248);
...
+$ion-border-radius-0: var(--ion-border-radius-0, $ion-scale-0);
+$ion-border-radius-100: var(--ion-border-radius-100, $ion-scale-100);
+$ion-border-radius-200: var(--ion-border-radius-200, $ion-scale-200);
+$ion-border-radius-300: var(--ion-border-radius-300, $ion-scale-300);
+$ion-border-radius-400: var(--ion-border-radius-400, $ion-scale-400);
+$ion-border-radius-500: var(--ion-border-radius-500, $ion-scale-500);
+$ion-border-radius-800: var(--ion-border-radius-800, $ion-scale-800);
+$ion-border-radius-1000: var(--ion-border-radius-1000, $ion-scale-1000);
+$ion-border-radius-full: var(--ion-border-radius-full, 999px);
+$ion-border-radius-025: var(--ion-border-radius-025, $ion-scale-050);
```
brandyscarney, line 580: Kind of unrelated to the PR but why is this one not after `border-radius-0`? And also why does it jump from `0` -> `25` -> `100`

core/src/foundations/ionic.vars.scss
```diff
@@ -939,6 +367,578 @@ $ion-primitives-pink-1100: var(--ion-primitives-pink-1100, #8b2248);
...
+$ion-bg-primary-base-default: var(--ion-bg-primary-base-default, $ion-semantics-primary-base);
+$ion-bg-primary-base-default-rgb: var(--ion-bg-primary-base-default-rgb, 16, 92, 239);
```
brandyscarney, line 582:
```suggestion
$ion-bg-primary-base-default-rgb: var(--ion-bg-primary-base-default-rgb, $ion-semantics-primary-base-rgb);
```

Should this fallback to `$ion-semantics-primary-base-rgb` (and the same for all other bg rgb vars)?

core/src/foundations/ionic.vars.scss
```diff
@@ -939,6 +367,578 @@ $ion-primitives-pink-1100: var(--ion-primitives-pink-1100, #8b2248);
...
+$ion-bg-extended-pink-base-press: var(--ion-bg-extended-pink-base-press, $ion-primitives-pink-1100);
+$ion-bg-extended-pink-base-press-rgb: var(--ion-bg-extended-pink-base-press-rgb, 139, 34, 72);
+$ion-bg-extended-pink-subtle-default: var(--ion-bg-extended-pink-subtle-default, $ion-primitives-pink-100);
+$ion-bg-extended-pink-subtle-default-rgb: var(--ion-bg-extended-pink-subtle-default-rgb, 254, 243, 245);
+$ion-bg-extended-pink-subtle-press: var(--ion-bg-extended-pink-subtle-press, $ion-primitives-pink-300);
+$ion-bg-extended-pink-subtle-press-rgb: var(--ion-bg-extended-pink-subtle-press-rgb, 250, 211, 220);
+$ion-text-default: var(--ion-text-default, $ion-primitives-neutral-1200);
+$ion-text-default-rgb: var(--ion-text-default-rgb, 36, 36, 36);
+$ion-text-subtle: var(--ion-text-subtle, $ion-primitives-neutral-1000);
+$ion-text-subtle-rgb: var(--ion-text-subtle-rgb, 59, 59, 59);
```
brandyscarney, line 750:
```suggestion
$ion-text-subtle-rgb: var(--ion-text-subtle-rgb, $ion-primitives-neutral-1000-rgb);
```

Should this fallback to `$ion-primitives-neutral-1000-rgb` (and the same for all other text rgb vars)?

core/src/foundations/ionic.vars.scss
```diff
@@ -939,6 +367,578 @@ $ion-primitives-pink-1100: var(--ion-primitives-pink-1100, #8b2248);
...
+$ion-text-extended-violet: var(--ion-text-extended-violet, $ion-primitives-violet-900);
+$ion-text-extended-violet-rgb: var(--ion-text-extended-violet-rgb, 101, 26, 197);
+$ion-text-extended-purple: var(--ion-text-extended-purple, $ion-primitives-purple-1000);
+$ion-text-extended-purple-rgb: var(--ion-text-extended-purple-rgb, 129, 42, 170);
+$ion-text-extended-magenta: var(--ion-text-extended-magenta, $ion-primitives-magenta-1100);
+$ion-text-extended-magenta-rgb: var(--ion-text-extended-magenta-rgb, 137, 43, 122);
+$ion-text-extended-pink: var(--ion-text-extended-pink, $ion-primitives-pink-1000);
+$ion-text-extended-pink-rgb: var(--ion-text-extended-pink-rgb, 170, 42, 88);
+$ion-icon-default: var(--ion-icon-default, $ion-primitives-neutral-1200);
+$ion-icon-default-rgb: var(--ion-icon-default-rgb, 36, 36, 36);
```
brandyscarney, line 796:
```suggestion
$ion-icon-default-rgb: var(--ion-icon-default-rgb, $ion-primitives-neutral-1200-rgb);
```

Should this fallback to `$ion-primitives-neutral-1200-rgb` (and the same for all other icon rgb vars)?

BenOsodrac:
> I just have a question (repeatedly asked but the same essential question) about the fallback for the rgb variables.

I'm preventing the generation of fallbacks for rgb variables. Mainly as it would add a little bit more code to achieve that and also rgb are only generated for internal use, they don't come from actual rgb tokens, so I thought it wouldn't bring as much value, as its lees likely devs will customize these.

But we can always return to this in the future, if we see it makes sense to add.

core/src/foundations/ionic.vars.scss
BenOsodrac: I believe its due to the order on the actual JSON, as these were added later by the UX team

brandyscarney (approved):

PR #30409, opened 2025-05-13
fix(scroll-assist): allow focus on input's siblings

thetaPC:
Issue number: internal

---------

If scroll assist is on (by default it is on for iOS mobile devices), then focus will be redirected to the input if a `focusIn` event fires within the component. This leads to some elements to not being accessible with screen readers.

For example: When the clear button is focused, focusin is dispatched and bubbles up to the ion-input. Our [scroll assist utility listens for focusin](https://github.com/ionic-team/ionic-framework/blob/2fc81ddc9b35d6909fd4b585079aedabbd659233/core/src/utils/input-shims/hacks/scroll-assist.ts#L135) to adjust the scroll position. It also causes the input to be re-focused. As a result, when swiping to the clear button with a screen reader, focus will be forcibly moved back to the input. This means that users cannot swipe away from the input to the right when using a screen reader.

This issue was resolved via PR https://github.com/ionic-team/ionic-framework/pull/29366. However, this only fixed the issue with the clear button. Since then, `ion-input` has implemented `start` and `end` slots. This has opened up the possibility of other focusable elements to being present within `ion-input`. The same issue applies to them. A great example of this is the `input-password-toggle`. The password toggle is required to have a `slot="end"` to render appropriately within `ion-input`. But the users can never access the toggle when using a screen reader.

## What is the new behavior?

Since we have no control of what is being passed within the the slots, the fix was done through the scroll assist instead of `ion-input`. It checks if any siblings of the native input have been focused, if they have then leave the focus on them instead of re-directing back to the native input.

- Updated the scroll assist logic to check for focusable siblings of the native input. If found, then allow focus on them.
- Removed the `focusIn` code from the clear button since it's being handle at the scroll assist level.

This fix will be introduced in a feature to lessen any possible issues in case of a regression.

- [ ] Yes
- [x] No

## Other information

N/A

### DIFF
--- a/core/src/utils/input-shims/hacks/scroll-assist.ts
+++ b/core/src/utils/input-shims/hacks/scroll-assist.ts
@@ -190,7 +190,7 @@ const setManualFocus = (el: HTMLElement) => {
    * the attribute here, screen readers would be unable to navigate to and interact
    * with these sibling elements.
    *
-   * Without this check, we would need to call `ev.stopPropagation()` on the 
+   * Without this check, we would need to call `ev.stopPropagation()` on the
    * 'focusin' event of each focusable sibling to prevent the scroll assist
    * listener from incorrectly moving focus back to the input. This approach
    * is less maintainable and more error-prone.
--- a/core/src/utils/input-shims/hacks/scroll-assist.ts
+++ b/core/src/utils/input-shims/hacks/scroll-assist.ts
@@ -181,6 +181,24 @@ const setManualFocus = (el: HTMLElement) => {
     return;
   }
 
+  /**
+   * Optimization for scenarios where the currently focused element is a sibling
+   * of the target element. In such cases, we avoid setting `SKIP_SCROLL_ASSIST`.
+   *
+   * This is crucial for accessibility: input elements can now contain focusable
+   * siblings (e.g., clear buttons, slotted elements). If we didn't skip setting
+   * the attribute here, screen readers would be unable to navigate to and interact
+   * with these sibling elements.
+   *
+   * Without this check, we would need to call `ev.stopPropagation()` on the 
+   * 'focusin' event of each focusable sibling to prevent the scroll assist
+   * listener from incorrectly moving focus back to the input. This approach
+   * is less maintainable and more error-prone.
+   */
+  if (document.activeElement?.parentNode === el.parentNode) {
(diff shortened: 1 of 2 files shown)


ShaneK (approved):
This looks fine to me and seems to work (assuming I'm testing correctly haha) but is there a reason it's going into the feature branch instead of main?

core/src/utils/input-shims/hacks/scroll-assist.ts
```diff
@@ -181,6 +181,24 @@ const setManualFocus = (el: HTMLElement) => {
...
+   *
+   * This is crucial for accessibility: input elements can now contain focusable
+   * siblings (e.g., clear buttons, slotted elements). If we didn't skip setting
+   * the attribute here, screen readers would be unable to navigate to and interact
+   * with these sibling elements.
+   *
+   * Without this check, we would need to call `ev.stopPropagation()` on the
+   * 'focusin' event of each focusable sibling to prevent the scroll assist
+   * listener from incorrectly moving focus back to the input. This approach
+   * is less maintainable and more error-prone.
```
ShaneK, line 196:
```suggestion
   * listener from incorrectly moving focus back to the input. That approach
   * would be maintainable and more error-prone.
```

Technically I think the way it is now is fine, just a little confusing. I tried to clear up confusion, but feel free to disregard this suggestion

thetaPC:
@ShaneK I'm not too familiar how the scroll assist works as a whole within Ionic Framework. How many components does this effect? Was there a reason that the fix for the clear button was done through the input instead at the scroll assist? etc. I figured that it would be safer to introduce it in a feature in case the fix ends up needing a regression. Thoughts?

ShaneK:
> @ShaneK I'm not too familiar how the scroll assist works as a whole within Ionic Framework. How many components does this effect? Was there a reason that the fix for the clear button was done through the input instead at the scroll assist? etc. I figured that it would be safer to introduce it in a feature in case the fix ends up needing a regression. Thoughts?

Sorry, I really don't have the context to answer this right now. @brandyscarney might when she's back?

brandyscarney (approved):

ShaneK (approved):

brandyscarney (approved):
LGTM! Nice job!

PR #30411, opened 2025-05-14
fix(input-password-toggle, button): force update aria attributes

thetaPC:
Issue number: internal

---------

The `ion-input-password-toggle` has aria attributes that are updated based on the value visibility. However, those values do not reflect on the native button. This leads to the screen readers to not announce correctly.

## What is the new behavior?

- The aria attributes now reflects correctly within the native button.
- The `aria-label` has been updated to indicate the state of visibility.
- Added tests.

- [ ] Yes
- [x] No

## Other information

Preview


brandyscarney (approved):
LGTM!

PR #30414, opened 2025-05-16
fix(button): update styling for badge

OS-pedrolourenco:
Issue number: internal
---------

Some stylings for the ion-badge when slotted in an ion-button were missing.

## What is the new behavior?
Updated the styling for the ion-badge when slotted in an ion-button according to the designs:
- Adjusted badge padding;
- Abandoned the `:has` pseudo-selector for a simple 'button-had-badge' class;
- Adjusted font-size for icons inside the button;
- Adjusted badge positioning according to button size and badge length.

- [ ] Yes
- [x] No


core/src/components/button/button.ionic.scss
```diff
@@ -131,15 +131,23 @@
...
 }
 
 :host(.button-small),
+:host(.button-medium) {
+  ::slotted(ion-icon[slot="start"]),
+  ::slotted(ion-icon[slot="end"]),
+  ::slotted(ion-icon[slot="icon-only"]) {
+    font-size: globals.$ion-font-size-500;
+  }
+}
```
thetaPC, line 144: Why are these needed when the [default](https://github.com/ionic-team/ionic-framework/pull/30414/files?file-filters%5B%5D=.html&file-filters%5B%5D=.scss&file-filters%5B%5D=.ts&file-filters%5B%5D=.tsx&show-viewed-files=true#diff-e7130eac638157a72b40aacb382be024fcd67337af1ab1d88b2c5a5f2f28912dR134) (line 134) is already 500?

core/src/components/button/button.ionic.scss
```diff
@@ -264,8 +272,47 @@
...
+  --padding-top: #{globals.$ion-space-0};
+  --padding-bottom: #{globals.$ion-space-0};
+}
+
+:host(.button-small) ::slotted(ion-badge) {
+  @include globals.position(null, calc(-1 * globals.$ion-space-050), null, null);
+}
+
+:host(.button-medium),
+:host:not(.button-small):not(.button-large) {
```
thetaPC, line 288: Why do we need this check? We are already targeting `medium`? Plus we try to minimize the use of `:not` when possible.

core/src/components/button/button.ionic.scss
```diff
@@ -264,8 +272,47 @@
...
+  }
+  ::slotted(ion-badge.long-badge.badge-vertical-bottom) {
+    @include globals.position($bottom: globals.$ion-space-100);
+  }
+
+  ::slotted(ion-badge:not(.long-badge).badge-vertical-top) {
+    @include globals.position(globals.$ion-space-100, globals.$ion-space-150, null, null);
+  }
+  ::slotted(ion-badge:not(.long-badge).badge-vertical-bottom) {
+    @include globals.position(null, globals.$ion-space-150, globals.$ion-space-100, null);
```
thetaPC, line 300: Why are do longer badges have different position values? I'm not seeing those differences in the designs.

thetaPC (changes requested):
Please update the PR title to be a fix since the PR is aiming to change the outcome to reflect the true styles.

core/src/components/button/button.ionic.scss
OS-pedrolourenco: These are the subtle differences I noticed in the designs:
image

OS-pedrolourenco: After discussing with a UX team member, their final decision was to move on with the current implemented positioning of the badge and to leave the character number limitation as a responsibility of the developer.

core/src/components/badge/test/hint/badge.e2e.ts-snapshots/badge-hint-button-size-md-ltr-Mobile-Chrome-linux.png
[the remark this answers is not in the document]
thetaPC: We can leave them as is.

thetaPC (approved):
LGTM

PR #30416, opened 2025-05-19
feat(datetime): fix height when presentation='time'

OS-giulianasilva:
Issue number: resolves #internal

---------

The `datetime` component had a broken display for the `ionic` theme, because the `height` defined for that component was too small.
image

## What is the new behavior?
The CSS code limiting the `height` was removed. Now, the height is set automatically.
image

- [ ] Yes
- [x] No


core/src/components/datetime/datetime.ionic.scss
```diff
@@ -178,8 +178,6 @@
 // -----------------------------------
 :host .datetime-time {
   @include globals.padding(globals.$ion-space-100, globals.$ion-space-400);
-
-  height: globals.$ion-scale-800;
```
ShaneK, line 182: I don't think this is the right approach. This is the desired height in the figma, removing it will make it no longer align with the designs. You may have to try something else to resolve this one, or get a designer to agree with this change.

ShaneK (changes requested):

core/src/components/datetime/datetime.ionic.scss
OS-giulianasilva: Thanks for noticing that! As a fix, I added the `height` to the time `button` instead of the datetime `container`. Let me know if you agree with that.

ShaneK (approved):
Solid fix! Looks good to me 🚀

PR #30417, opened 2025-05-19
feat(input): add styles for ionic theme for fill=solid

JoaoFerreira-FrontEnd:
Issue number: Internal

---------

## What is the new behavior?

- add fill solid to ionic theme input;
- move input.outline.scss to input.ionic.scss;
- add fill solid to states page;
- remove highlight styles for ionic theme;
- add tests for ionic solid input;

- [ ] Yes
- [x] No

## Other information

- fill: https://ionic-framework-git-rou-11889-ionic1.vercel.app/src/components/input/test/fill?ionic:theme=ionic
- states: https://ionic-framework-git-rou-11889-ionic1.vercel.app/src/components/input/test/states?ionic:theme=ionic


BenOsodrac (approved):

core/src/components/input/input.ionic.outline.scss
brandyscarney, line 1: Why are we deleting this file but adding a solid one? Can't we keep both and default the fill styles to `input-fill-outline`?

core/src/components/input/input.ionic.scss
```diff
@@ -1,26 +1,127 @@
...
 @use "./input.common";
-@forward "./input.ionic.outline.scss";
-
+@forward "./input.ionic.solid.scss";
 // Ionic Input
 // --------------------------------------------------
 
 :host {
   --color: #{globals.$ion-primitives-neutral-1200};
   --border-width: #{globals.$ion-border-size-025};
```
brandyscarney, line 9: We should move all styles specific to `fill` to the `input-fill-outline` class like we do for button: 

- [button-solid](https://github.com/ionic-team/ionic-framework/blob/1877ac5dd13ddc0a969c75147d08558b3093f117/core/src/components/button/button.ionic.scss#L45)
- [button-outline](https://github.com/ionic-team/ionic-framework/blob/1877ac5dd13ddc0a969c75147d08558b3093f117/core/src/components/button/button.ionic.scss#L59)

This will make it so that in the future if we need to change the default fill we don't have to move the styles around.

core/src/components/input/input.ionic.scss
```diff
@@ -1,26 +1,127 @@
...
   --border-color: #{globals.$ion-primitives-neutral-500};
+  --border-radius: #{globals.$ion-border-radius-100};
```
brandyscarney, line 11: Similarly, can we move all styles specific to `shape` to the shape class such as `input-fill-round` so that if we change the default we don't have to move styles?

brandyscarney, line 11: Where is this value coming from? In Figma I see the following:

`round` -> `round.xl` -> `16px` -> `$ion-round-xl` (`$ion-border-radius-400` -> `$ion-scale-400: var(--ion-scale-400, 16px);`)
`soft` -> `soft.xl` -> `8px` -> `$ion-soft-xl` (`$ion-border-radius-200` -> `$ion-scale-200: var(--ion-scale-200, 8px);`)
`rectangular` -> `rectangular.xl` -> `0px` -> `$ion-rectangular-xl` (`$ion-border-radius-0` -> `$ion-scale-0: var(--ion-scale-0, 0px);`)

core/src/components/input/input.ionic.scss
```diff
@@ -1,26 +1,127 @@
...
   --highlight-color-valid: #{globals.$ion-semantics-success-900};
-  --highlight-color-invalid: #{globals.$ion-semantics-danger-800};
+  --highlight-color-invalid: #{globals.$ion-border-danger-default};
+  --padding-start: #{globals.$ion-space-300};
+  --padding-end: #{globals.$ion-space-300};
```
brandyscarney, line 15: This seems to only apply for the `small` size - can we move this to `.input-size-small`?

core/src/components/input/input.ionic.scss
```diff
@@ -33,6 +134,21 @@
   min-height: globals.$ion-scale-1400;
 }
 
+// Input Shapes
+// --------------------------------------------------
+
+:host(.input-shape-soft) {
+  --border-radius: #{globals.$ion-border-radius-200};
```
brandyscarney, line 141:
```suggestion
  --border-radius: #{globals.$ion-soft-xl};
```

This would match Figma

core/src/components/input/input.ionic.scss
```diff
@@ -33,6 +134,21 @@
...
+}
+
+:host(.input-shape-round) {
+  --border-radius: #{globals.$ion-border-radius-400};
```
brandyscarney, line 145:
```suggestion
  --border-radius: #{globals.$ion-round-xl};
```

This would match Figma

core/src/components/input/input.ionic.scss
```diff
@@ -33,6 +134,21 @@
...
+}
+
+:host(.input-shape-rectangular) {
+  --border-radius: #{globals.$ion-border-radius-0};
```
brandyscarney, line 149:
```suggestion
  --border-radius: #{globals.$ion-rectangular-xl};
```

This would match Figma

brandyscarney (changes requested):
Looks good just a few requested changes on where styles are applied!

core/src/components/input/input.ionic.scss
JoaoFerreira-FrontEnd: It was from [here](https://github.com/ionic-team/ionic-framework/pull/30417/files#diff-cbf91c9b97e425d969df4730e12b3424dd54de7664b9e629d455d2e63539c84eL7), I can delete it, but it's possible to have an undefined shape when in ionic theme [here](https://github.com/ionic-team/ionic-framework/pull/30417/files#diff-a39147111a459c7d50e285f51eddf2d3fee983bda67654c842a2c645565af6bcR566)

brandyscarney: The `undefined` shape is only when it isn't the `ionic` theme:

Screenshot 2025-05-22 at 2 04 01 PM

brandyscarney (approved):

PR #30424, opened 2025-05-25
chore(git): improve Ionitron to stop closing issues that have the 'needs: investigation' label

gnbm:
Issue number: internal

---------

- Currently, `Ionitron` is closing issues when they have the `needs: investigation` label.
- We might want to use the duplicate ticket as our source of truth since a community member has made a [suggestion](https://github.com/ionic-team/ionic-framework/issues/30380#issuecomment-2848626887) that could fix it.

## What is the new behavior?

- The `ionic-issue-bot.yml` file was changed to stop closing issues that have the `needs: investigation` label
- To prevent `Ionitron` from closing issues with the `needs: investigation` label, we've added this label to the `exemptLabels` list under the `stale`, `noReply`, and `noReproduction` sections. 
- This tells the bot not to close issues with that label, even if they meet the criteria for being stale or lacking a reply/reproduction.

- [ ] Yes
- [X] No

## Other information

- _N.A._

### DIFF
--- a/.github/ionic-issue-bot.yml
+++ b/.github/ionic-issue-bot.yml
@@ -132,8 +132,6 @@ noReproduction:
   maxIssuesPerRun: 100
   label: "ionitron: needs reproduction"
   responseLabel: triage
-  exemptLabels:
-    - "needs: investigation"
   exemptProjects: true
   exemptMilestones: true
   message: >


.github/ionic-issue-bot.yml
```diff
@@ -113,6 +114,8 @@ noReply:
   maxIssuesPerRun: 100
   label: "needs: reply"
   responseLabel: triage
+  exemptLabels:
+    - "needs: investigation"
```
brandyscarney, line 118:
```suggestion
```

This isn't needed because the issue is only closed if it has `needs: reply` and we don't add that to investigative tickets

.github/ionic-issue-bot.yml
```diff
@@ -131,6 +134,8 @@ noReproduction:
   maxIssuesPerRun: 100
   label: "ionitron: needs reproduction"
   responseLabel: triage
+  exemptLabels:
+    - "needs: investigation"
```
brandyscarney, line 138:
```suggestion
```

This isn't needed because the issue is only closed if it has `ionitron: needs reproduction` and we don't add that to investigative tickets

brandyscarney (approved):

PR #30444, opened 2025-05-30
fix(input-otp): correctly handle autofill by splitting the values into all inputs

brandyscarney:
Issue number: N/A

---------

When filling the Input OTP from inside of a Capacitor app or using autofill in the browser, the entire value is being inserted into the 1st input box so it is only getting the 1st character.

## What is the new behavior?
Detect when the value is longer than one in the `onInput` handler and split it up across all inputs.

- [ ] Yes
- [x] No

## Other information

Dev build: `8.5.7-dev.11748895007.1c98a49b`


thetaPC (approved):
LGTM, thanks!

PR #29845, opened 2024-09-06
fix(item-sliding): check for side attribute to avoid an `undefined` value

brandyscarney:
Issue number: resolves #29499

---------

## What is the current behavior?
Using the bundled version of Ionic results in an error `Uncaught (in promise) Error: "undefined" is not a valid value for [side]. Use "start" or "end" instead.` with `<ion-item-sliding>`.

Reproduction 1: 
- https://stackblitz.com/edit/stackblitz-starters-mezoy6?file=index.html
- Press the Add (`+`) button several times

It is also reproducible in an Ionic Angular app when installed via npm.

Reproduction 2: 
- https://stackblitz.com/edit/angular-exwgke?file=src%2Fapp%2Fexample.component.ts
- Press the `Add Items` button several times

## What is the new behavior?
Check for the side attribute to avoid `side` being `undefined`.

## Does this introduce a breaking change?
- [ ] Yes
- [x] No

## Other information

Dev build: `8.6.1-dev.11749562115.1e681558`

Test PR: https://github.com/ionic-team/ionic-framework/pull/30469

The tests must be merged separately, after this is released, in order to use the fix from the CDN.

PR #30469, opened 2025-06-10
test(item-sliding): add test for side issue with CDN

brandyscarney:
Adds test for https://github.com/ionic-team/ionic-framework/pull/29845

Must be merged after the fix in that PR is released. The test will fail while using the older version of the CDN.


core/src/utils/test/playwright/page/utils/set-content.ts
```diff
@@ -33,18 +33,33 @@ export const setContent = async (page: Page, html: string, testInfo: TestInfo, o
...
     <!DOCTYPE html>
     <html dir="${direction}" lang="en">
       <head>
         <title>Ionic Playwright Test</title>
         <meta charset="UTF-8" />
         <meta name="viewport" content="width=device-width, initial-scale=1.0, minimum-scale=1.0" />
-        <link href="${baseUrl}/css/ionic.bundle.css" rel="stylesheet" />
         <link href="${baseUrl}/scripts/testing/styles.css" rel="stylesheet" />
         ${palette !== 'light' ? `<link href="${baseUrl}/css/palettes/${palette}.always.css" rel="stylesheet" />` : ''}
         <script src="${baseUrl}/scripts/testing/scripts.js"></script>
```
christian-bromann, line 66: Why not these as well?

christian-bromann (approved):
LGTM 👍 

Had one question to better understand the approach but non blocking.

core/src/utils/test/playwright/page/utils/set-content.ts
brandyscarney: The testing scripts file is required for Playwright. Without it we would have to set all of the stuff for telling Playwright the app is loaded and determining the document direction, palette, etc: https://github.com/ionic-team/ionic-framework/blob/main/core/scripts/testing/scripts.js

I actually accidentally removed this at first and ran into an issue where none of the test after `setContent` was executed because `window.testAppLoaded = true` was never set.

ShaneK (approved):

PR #30482, opened 2025-06-12
fix(range): improve focus and blur handling for dual knobs

ShaneK:
Issue number: resolves #internal

---------

Currently, if you use keyboard navigation to move between dual range slider knobs, only the first knob you navigate to is highlighted. This is because both elements in the same component are marked as focusable and the code that manages focusable doesn't take into account multiple elements in the same component.

https://github.com/user-attachments/assets/36d84eed-6928-446e-becd-ffa2a97e3cc2

## What is the new behavior?

After these changes, we manage focusing on dual knob range sliders manually, so using tab navigation through dual knob range sliders focuses knobs as expected.

- [ ] Yes
- [X] No

## Other information

[Test - Range - Basic screen](https://ionic-framework-git-fw-6401-ionic1.vercel.app/src/components/range/test/basic)

### DIFF
--- a/core/src/components/range/test/basic/range.spec.ts
+++ b/core/src/components/range/test/basic/range.spec.ts
@@ -156,46 +156,89 @@ describe('range: dual knobs focus management', () => {
     expect(blurEventFired).toBe(true);
   });
 
-  it('should correctly handle Tab navigation between knobs', async () => {
+  it('should correctly handle Tab navigation between knobs using KeyboardEvent', async () => {
+    // Using KeyboardEvent to simulate Tab key is more realistic than just firing focus events
+    // because it tests the actual keyboard navigation behavior users would experience
     const page = await newSpecPage({
       components: [Range],
       html: `
+        <button id="before">Before</button>
         <ion-range dual-knobs="true" min="0" max="100" value='{"lower": 25, "upper": 75}' aria-label="Dual range">
         </ion-range>
+        <button id="after">After</button>
       `,
     });
 
     const range = page.body.querySelector('ion-range')!;
+    const beforeButton = page.body.querySelector('#before') as HTMLElement;
     await page.waitForChanges();
 
     const knobA = range.shadowRoot!.querySelector('.range-knob-a') as HTMLElement;
     const knobB = range.shadowRoot!.querySelector('.range-knob-b') as HTMLElement;
 
-    // Simulate Tab to first knob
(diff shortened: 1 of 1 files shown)


core/src/components/range/test/basic/range.spec.ts
```diff
@@ -0,0 +1,199 @@
...
+      focusEventFired = true;
+    });
+
+    const knobA = range.shadowRoot!.querySelector('.range-knob-a') as HTMLElement;
+
+    // Focus knob A
+    knobA.dispatchEvent(new Event('focus'));
+    await page.waitForChanges();
+
+    expect(focusEventFired).toBe(true);
```
brandyscarney: Optional: should we check that `ionFocus` is not emitted again after changing focus from knob A to knob B?

brandyscarney (approved):
Looks good and works well! One optional comment on an additional test we could add.

core/src/components/range/test/basic/range.spec.ts
ShaneK: Great idea! Done! [40db335](https://github.com/ionic-team/ionic-framework/pull/30482/commits/40db33569c5d38c0c81b6233163e25b7ae2155ee)

core/src/components/range/test/basic/range.spec.ts
```diff
@@ -0,0 +1,201 @@
...
+      `,
+    });
+
+    const range = page.body.querySelector('ion-range')!;
+    await page.waitForChanges();
+
+    const knobA = range.shadowRoot!.querySelector('.range-knob-a') as HTMLElement;
+    const knobB = range.shadowRoot!.querySelector('.range-knob-b') as HTMLElement;
+
+    // Simulate Tab to first knob
```
thetaPC: Couldn't we use `KeyboardEvent` to simulate the `TAB`?

thetaPC (approved):
LGTM, left a question but it's not a blocker

core/src/components/range/test/basic/range.spec.ts
ShaneK: Great idea! I wasn't sure how to do that before, but this was a solid lead. Done here: [3f514ff](https://github.com/ionic-team/ionic-framework/pull/30482/commits/3f514ff54c9780a5476e21f525d57df3adfd61fe)

PR #30496, opened 2025-06-17
fix(sheet): fix toolbar not showing contents in ios for ionic theme

OS-giulianasilva:

---------

The `ion-toolbar` contents inside the `sheet` were not being displayed correctly in `iOS`.

## What is the new behavior?
This PR add `z-index` values of the `ion-toolbar`'s `content` and `background` containers to fix the issue.

- [ ] Yes
- [x] No

## Other information
Before:
Sample Image

After: 
Sample Image


BenOsodrac (approved):

BenOsodrac (approved):

core/src/components/modal/test/sheet/modal.e2e.ts
```diff
@@ -175,7 +177,7 @@ configs({ modes: ['ios'], directions: ['ltr'] }).forEach(({ title, config }) =>
         const ionBreakpointDidChange = await page.spyOnEvent('ionBreakpointDidChange');
         const header = page.locator('.modal-sheet ion-header');
 
-        await dragElementBy(header, page, 0, 125);
+        await dragElementBy(header, page, 0, 110);
```
OS-giulianasilva, line 180: Change required due to [errors when updating the snapshots](https://github.com/ionic-team/ionic-framework/actions/runs/15729294792).

> Error: The element is being dragged past the bottom of the viewport. Update the dragByY value to prevent going out of bounds. A recommended value is 110.5.

ShaneK (approved):
This looks fine to me. I'm not super happy with using z-indexes to fix this issue, but it's probably caused by other z-indexes at this point. This sort of z-index hell happens with manual manipulation like this 🫠

PR #30507, opened 2025-06-23
test(angular): add ng20 test app

thetaPC:
Issue number: internal

---------

There are tests apps for Angular 16, 17, 18, and 19 only. Angular 20 has been released and should be included.

## What is the new behavior?

- Adds a test app for Angular 20

- [ ] Yes
- [x] No

## Other information

N/A


.github/workflows/actions/test-angular-e2e/action.yml
```diff
@@ -8,7 +8,7 @@ runs:
   steps:
     - uses: actions/setup-node@v4
       with:
-        node-version: 18
+        node-version: 22.x
```
thetaPC, line 11: Angular 20 requires a minimal node version of 20. I verified that version 22 works with all the Angular test apps that we support. By updating it to `22.x`, renovate will be able to update it when necessary instead of keeping it pinned.

packages/angular/test/apps/ng20/e2e/src/.gitkeep
brandyscarney, line 1: Do we need this file?

thetaPC: I don't think so. I noticed that the base folder isn't up to date with the latest implementation. I plan to update the files in a different PR and also look into which files are needed under the version apps. So you actually caught me in before stages 😆

PR #30508, opened 2025-06-23
fix(toolbar): allow slot to size to content to account for images

brandyscarney:
Issue number: internal

---------

Images do not appear because their width is calculated as `0` when determining the widths of slotted content.

## What is the new behavior?
Sets the `flex-basis` variable to `auto` to allow it to calculate the width using the image's intrinsic width and then sets it back to the width of the image.

- [ ] Yes
- [x] No

## Other information

Dev build: `8.6.2-dev.11750780384.13227a56`


thetaPC (approved):
LGTM

PR #30509, opened 2025-06-23
feat(datetime): add divider to datetime-action-buttons

JoaoFerreira-FrontEnd:
Issue number: internal

---------
## What is the new behavior?

Now, the action buttons have a divider for improved readability.

- [ ] Yes
- [x] No

### DIFF
--- a/core/src/components/datetime/datetime.ionic.scss
+++ b/core/src/components/datetime/datetime.ionic.scss
@@ -203,6 +203,17 @@
 
 // Calendar / Footer / Action Buttons
 // -----------------------------------
+:host .datetime-action-buttons {
+  @include globals.padding(
+    globals.$ion-space-200,
+    globals.$ion-space-400,
+    globals.$ion-space-200,
+    globals.$ion-space-400
+  );
+
+  margin-top: globals.$ion-space-200;
+  border-top: globals.$ion-border-style-solid globals.$ion-border-size-025 globals.$ion-primitives-neutral-300;
+}
 
 :host .datetime-buttons ion-buttons,
 .datetime-action-buttons .datetime-action-buttons-container {


core/src/components/datetime/datetime.ionic.scss
```diff
@@ -216,6 +216,13 @@
 
 // Calendar / Footer / Action Buttons
 // -----------------------------------
+:host .datetime-footer {
+  @include globals.padding(globals.$ion-space-200, globals.$ion-space-0, globals.$ion-space-200, globals.$ion-space-0);
+  margin-top: globals.$ion-space-200;
+
+  border-top: globals.$ion-border-style-solid globals.$ion-border-size-025 globals.$ion-primitives-neutral-300;
```
BenOsodrac: I believe you're supposed to use the margin and border mixins here

ShaneK: In addition to Ben's comment, where are these design details being pulled from? From the Figma, there is no border on the datepicker, unless it's been updated in a place I'm not aware of?

JoaoFerreira-FrontEnd: I've updated the code to use the mixins. The design I sent it to you via JIRA comment.

ShaneK: Awesome! Looks like you also need to update your screenshots

core/src/components/datetime/datetime.ionic.scss
```diff
@@ -216,6 +216,17 @@
 
 // Calendar / Footer / Action Buttons
 // -----------------------------------
+:host .datetime-footer {
+  @include globals.padding(globals.$ion-space-200, globals.$ion-space-0, globals.$ion-space-200, globals.$ion-space-0);
+  @include globals.margin(globals.$ion-space-200, 0, 0, 0);
```
ShaneK, line 221: So there's a problem with this, I don't think the bottom padding should be here or - if it should - then there should be padding on the sides too. The path you've taken here makes it so the padding on the sides has to be handled by the containing component while the padding on the bottom is handled by this one.

I think you need to make a choice, either:
- A: This shouldn't have bottom padding and that should be done by whatever is containing the element (probably the best choice), spacing around this element shouldn't be handled by it directly IMO
- B: Add side padding so it matches the figma designs and padding is handled consistently. I don't think you can do this without  messing up the spacing on the rest of the datepicker though.

Additionally, these numbers do not align with the values in figma - space 300, for the padding and margin.

core/src/components/datetime/datetime.ionic.scss
```diff
@@ -216,6 +216,17 @@
...
+  @include globals.border(
+    globals.$ion-border-style-solid globals.$ion-border-size-025 globals.$ion-primitives-neutral-300,
```
ShaneK: This color doesn't seem to align with the Figma color. It seems to be `#E0E0E0` while the Figma design uses `#⁨⁨⁨⁨D5D5D5`.

ShaneK (changes requested):
This doesn't seem quite right. Let me know if I'm missing something!

core/src/components/datetime/datetime.ionic.scss
JoaoFerreira-FrontEnd: Agreed, I decide to remove the padding bottom as suggested.

core/src/components/datetime/datetime.ionic.scss
JoaoFerreira-FrontEnd: You're right dividers color updated

ShaneK (approved):
This looks right to me, assuming you get whatever's going on in the screenshot tests ironed out 💪

PR #30510, opened 2025-06-24
fix(modal): add conditional tabIndex for handle cycling

ShaneK:

---------

Currently, you cannot tab into a sheet modal from outside of it (the background), even with `handleBehavior` set to `cycle`. This destroys the accessibility of moving from the background behind a sheet modal to the contents of a sheet modal/the drag bar to be able to cycle the size.

## What is the new behavior?

Now you can get into a sheet modal from outside of it and interact with its contents/drag handle when `handleBehavior` is set to `cycle`. This opens up the accessibility of the sheet modal and allows for interacting with background elements with sheet modals open using accessibility tools like VoiceOver and TalkBack.

- [ ] Yes
- [X] No

## Other information

[Relevant test screen](https://ionic-framework-git-fw-6523-ionic1.vercel.app/src/components/modal/test/sheet)

Dev build: `8.6.3-dev.11750971489.140836b0`


core/src/components/modal/test/sheet/index.html
```diff
@@ -106,6 +106,12 @@
           >
             Present Sheet Modal (Scroll at any breakpoint)
           </button>
+          <button
+            id="scroll-at-edge-modal"
```
thetaPC: Let's update the `id` value so there aren't duplicates.

thetaPC (commented):
When testing the dev build on the minimal repro, I noticed that I have to tab twice from the increment button to get to the first interactable element within the modal. Is that expected? I would have thought that I only need one tab.

core/src/components/modal/test/sheet/index.html
ShaneK: Good catch 😅 Fixed in [0b1c67d](https://github.com/ionic-team/ionic-framework/pull/30510/commits/0b1c67d95ef01aa09959cba6ad857d2a948c8bd5)

thetaPC (dismissed):
Let's try to add a test to prevent regression.

PR #30517, opened 2025-06-27
chore(angular): upgrade test apps to latest

thetaPC:
Issue number: resolves #

---------

## What is the new behavior?

-
-
-

- [ ] Yes
- [ ] No


packages/angular/test/apps/ng16/angular.json
thetaPC, line 1: Not sure why this was considered a rename when I manually deleted the file from `ng19` and added a new file within `ng16`. The same applies for the other renames.

brandyscarney (approved):

brandyscarney (approved):
Reviewed code, tests all versions of the test app. Looks good! 👍

packages/angular/test/apps/ng16/angular.json
ShaneK: Git doesn't have a way to identify when a file was deleted or renamed, so if a file is removed and another one is added and they're similar enough it just considers it a rename. If a file just disappears then it considers it deleted. Sometimes it has false positives like this haha

PR #30518, opened 2025-06-27
test(breadcrumbs): update Playwright tests to use setContent

brandyscarney:
Blocks https://github.com/ionic-team/ionic-framework/pull/30513

----

The latest Playwright update in #30513 exposed an issue in Safari where dynamically setting the RTL direction on the document no longer works. This causes the Breadcrumbs screenshots to fail because they are not flipping the arrow icon. 

This PR converts the Breadcrumbs e2e tests to use `setContent` which works properly with RTL and also breaks the tests up more so they aren't one giant screenshot.

- The [Basic](https://ionic-framework-git-bc-test-breadcrumbs-ionic1.vercel.app/src/components/breadcrumbs/test/basic) test has been updated to focus mainly on the default breadcrumbs without any colors or collapsed breadcrumbs.
- The [Collapsed](https://ionic-framework-git-bc-test-breadcrumbs-ionic1.vercel.app/src/components/breadcrumbs/test/collapsed) test has been updated to remove some of the duplicated examples and ported over one from the basic test.
- The [Color](https://ionic-framework-git-bc-test-breadcrumbs-ionic1.vercel.app/src/components/breadcrumbs/test/color) test was added which focuses on setting different colors on the Breadcrumbs / Breadcrumb and moves this out of the basic test.


core/src/components/breadcrumbs/test/collapsed/breadcrumbs.e2e.ts
brandyscarney, line 1: Some of these could probably be removed as screenshots and just check the count on the visible breadcrumbs. LMK if you think there are too many screenshots here.

ShaneK: I think a lot of these could just be programmatic tests, but also they're probably fine as visual regression tests like this. I'm fine either way. I do like this much more granular approach to visually testing them though!

PR #30519, opened 2025-06-30
fix(item): allow nested content to be conditionally interactive

### LINKED ISSUE #29763: bug: conditional appearance of the content of an ion-item prevent click on right and left edge of the ion-item (with checkbox)
### Prerequisites

### Ionic Framework Version

v8.x

### Current Behavior

When putting a ion-checkbox inside of an ion-item, the click on the right and left edge of the ion-item only works at the beginning. Once I hide and show again the checkbox, the click on the right and left edge of the ion-item does not work anymore.

https://github.com/user-attachments/assets/2d3c7e10-3c44-4879-a9c8-e5a5bc6fab0f

### Expected Behavior

The ion-item should be clickable after hiding/showing the checkbox.

### Steps to Reproduce

1. Paste the following code in a page:

```html
  <ion-item>
    @if (showCheckbox()) {
      <ion-checkbox (ionChange)="stateChange()">Test checkbox</ion-checkbox>
    } @else {
    <ion-label>
      Test checkbox
    </ion-label>

    <ion-spinner
      slot="end"
      name="lines"
      class="item-spinner">
    </ion-spinner>
    }
  </ion-item>

  <ion-button (click)="toggleDisplay()">Toggle</ion-button>
```

```ts
  showCheckbox = signal(true)

  stateChange() {
    console.log('Checkbox state changed.')
  }

  toggleDisplay() {
    this.showCheckbox.set(!this.showCheckbox())
  }
```

2. Try to click on the left or right edge of the ion-item after toggling the display of the checkbox once.
3. The click should not work after toggling the display.

### Code Reproduction URL

https://github.com/JulienLecoq/ion-item_click_bug_after_toggling_display/tree/main

### Ionic Info

Ionic:

   Ionic CLI                     : 7.2.0 (/Users/julien_lecoq/.nvm/versions/node/v20.14.0/lib/node_modules/@ionic/cli)
   Ionic Framework               : @ionic/angular 8.2.6
   @angular-devkit/build-angular : 18.1.4
   @angular-devkit/schematics    : 18.1.4
   @angular/cli                  : 18.1.4

Capacitor:

   Capacitor CLI      : 6.1.2
   @capacitor/core    : 6.1.2

Utility:

   native-run  : 2.0.1

System:

   NodeJS : v20.14.0 (/Users/julien_lecoq/.nvm/versions/node/v20.14.0/bin/node)
   npm    : 10.7.0
   OS     : macOS Unknown

### Additional Information

_No response_

soundproofboot:
Issue number: resolves #29763 

---------

If the nested content of an ion-item is conditionally rendered and goes from containing zero interactive elements to one or more, a render is not triggered in Angular and the item does not behave correctly for one or more nested inputs.

## What is the new behavior?

- A mutation observer is created in `connectedCallback()` to watch for changes in item's child list.  When the `childList` changes, two pieces of state that track whether the item needs to be interactive and whether there are multiple interactive elements are updated.
- Add `disconnectedCallback()` and logic to disconnect Mutation Observer.
- Create new function `totalNestedInputs()` with logic from existing `setMultipleInputs` function to be used for both `setMultipleInputs` and new function `setIsInteractive`.

- [ ] Yes
- [x] No

## Other information

 I opted for the MutationObserver over a `slotchange` listener because the `slotchange` fires synchronously on any slot within the shadowRoot, and the MutationObserver fires once on the item element itself. I attempted to add the minimum amount of logic to achieve this but there may be a more elegant solution that combines what `multipleInputs` and `isInteractive` are doing but requires more changes to existing code.

### DIFF
--- a/core/src/components/item/item.tsx
+++ b/core/src/components/item/item.tsx
@@ -32,7 +32,6 @@ export class Item implements ComponentInterface, AnchorInterface, ButtonInterfac
   private labelColorStyles = {};
   private itemStyles = new Map<string, CssClassMap>();
   private inheritedAriaAttributes: Attributes = {};
-  private observer?: MutationObserver;
 
   @Element() el!: HTMLIonItemElement;
 
@@ -165,19 +164,6 @@ export class Item implements ComponentInterface, AnchorInterface, ButtonInterfac
 
   connectedCallback() {
     this.hasStartEl();
-
-    // create MutationObserver to watch for changes in nested content
-    // and update state to reflect change in behavior and rerender
-    if (typeof MutationObserver !== 'undefined' && !this.observer) {
-      this.observer = new MutationObserver(() => {
-        this.setIsInteractive();
-        this.setMultipleInputs();
-      });
-
-      this.observer.observe(this.el, {
-        childList: true,
-      });
-    }
   }
 
   componentWillLoad() {
@@ -187,15 +173,9 @@ export class Item implements ComponentInterface, AnchorInterface, ButtonInterfac
   componentDidLoad() {
     raf(() => {
       this.setMultipleInputs();
+      this.setIsInteractive();
       this.focusable = this.isFocusable();
     });
-  }
-
-  disconnectedCallback(): void {
-    // disconnect MutationObserver when component is disconnected from the DOM
-    if (this.observer) {
-      this.observer.disconnect();
-    }
   }
 
   private totalNestedInputs() {
@@ -237,8 +217,14 @@ export class Item implements ComponentInterface, AnchorInterface, ButtonInterfac
     // If item contains any interactive children, set isInteractive to `true`
     const { covers, inputs, clickables } = this.totalNestedInputs();
 
-    this.isInteractive = !!(covers.length + inputs.length + clickables.length);
+    this.isInteractive = covers.length > 0 || inputs.length > 0 || clickables.length > 0;
   }
+
+  // slot change listener updates state to reflect how/if item should be interactive
+  private updateInteractivityOnSlotChange = () => {
+    this.setIsInteractive();
+    this.setMultipleInputs();
+  };
 
   // If the item contains an input including a checkbox, datetime, select, or radio
   // then the item will have a clickable input cover that covers the item
@@ -403,12 +389,12 @@ export class Item implements ComponentInterface, AnchorInterface, ButtonInterfac
           disabled={disabled}
           {...clickFn}
         >
-          <slot name="start"></slot>
+          <slot name="start" onSlotchange={this.updateInteractivityOnSlotChange}></slot>
           <div class="item-inner">
             <div class="input-wrapper">
-              <slot></slot>
+              <slot onSlotchange={this.updateInteractivityOnSlotChange}></slot>
             </div>
-            <slot name="end"></slot>
+            <slot name="end" onSlotchange={this.updateInteractivityOnSlotChange}></slot>
             {showDetail && (
               <ion-icon
                 icon={detailIcon}
--- a/core/src/components/item/item.tsx
+++ b/core/src/components/item/item.tsx
@@ -32,11 +32,13 @@ export class Item implements ComponentInterface, AnchorInterface, ButtonInterfac
   private labelColorStyles = {};
   private itemStyles = new Map<string, CssClassMap>();
   private inheritedAriaAttributes: Attributes = {};
+  private observer?: MutationObserver;
 
   @Element() el!: HTMLIonItemElement;
 
   @State() multipleInputs = false;
   @State() focusable = true;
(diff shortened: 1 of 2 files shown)


core/src/components/item/item.tsx
```diff
@@ -201,6 +233,13 @@ export class Item implements ComponentInterface, AnchorInterface, ButtonInterfac
       (covers.length > 0 && this.isClickable());
   }
 
+  private setIsInteractive() {
+    // If item contains any interactive children, set isInteractive to `true`
+    const { covers, inputs, clickables } = this.totalNestedInputs();
+
+    this.isInteractive = !!(covers.length + inputs.length + clickables.length);
```
ShaneK:
```suggestion
    this.isInteractive = covers.length || inputs.length || clickables.length;
```
This change request is very minor, but it's more obvious what you're doing at a glance and it allows short circuiting from the engine as soon as any of the checks pass. I don't know if the compiler is smart enough to see what you're doing here and break it out the same way when you're adding lengths like that, but even if it is I still think the readability would be improved. You can even omit the explicit boolean casting (`!!`) since that happens from the `||` operator.

core/src/components/item/item.tsx
```diff
@@ -32,11 +32,13 @@ export class Item implements ComponentInterface, AnchorInterface, ButtonInterfac
   private labelColorStyles = {};
   private itemStyles = new Map<string, CssClassMap>();
   private inheritedAriaAttributes: Attributes = {};
+  private observer?: MutationObserver;
```
ShaneK: I think we should rename this so it's more obvious what its function is from just looking at it. In these components, a lot tends to happen, so clarity in variable naming - especially at this scope - is important. Probably something more like `interactivityObserver` or `contentObserver` if we think someday we'll do something else when the child content changes.

soundproofboot:
> Overall this looks alright and I requested some minor changes for clarity. However, I am worried about this approach because now we're creating an observable for every item in a list, which is probably fine for small lists but may impact performance in larger lists - especially those with content that changes.
> 
> I think you should try to see if you can do the same thing with existing [Stencil lifecycle hooks](https://stenciljs.com/docs/component-lifecycle), such as [`componentDidRender`](https://stenciljs.com/docs/component-lifecycle#componentdidrender), which _should_, from my understanding, already hook into the changes you're looking for. Possibly even `componentDidUpdate` - but be warned that only fires after the first render. If you went this path you'd need to do a check after first render, too. I'd encourage exploring those paths before continuing down this one.

Awesome, thank you for the feedback.  I'll look back at the lifecycle hooks again and see if I can come up with a more performant solution.

soundproofboot:
@ShaneK I replaced the MutationObserver with a `slotchange` handler that should effectively be doing the same thing but still might feel heavy handed. For this issue the parent needs to potentially render again when the nested content changes and I can't find [anything built into Stencil](https://github.com/stenciljs/core/issues/2770) like that - since the parent component itself isn't updating, none of the lifecycle hooks are firing. I think attaching three handlers per component may be slightly more performant than a mutation observer. My understanding is that the slots themselves are preserved between renders so they are not being recreated.

ShaneK (approved):
I think this solution looks a lot more clean than the previous one, thanks for trying again!

core/src/components/item/item.tsx
brandyscarney, line 1: The fix looks good! Let's add a test for this to avoid a regression. I would throw it in `core/src/components/item/test/inputs/item.e2e.ts` in the last `test.describe(title('item: inputs'), () => {` block:

```ts
test('should update interactivity state when elements are conditionally rendered', async ({ page }, testInfo) => {
  testInfo.annotations.push({
    type: 'issue',
    description: 'https://github.com/ionic-team/ionic-framework/issues/29763',
  });

  await page.setContent(
    `
    <ion-list>
      <ion-item>
        <ion-label>Conditional Checkbox</ion-label>
      </ion-item>
    </ion-list>
    `,
    config
  );

  const item = page.locator('ion-item');

  await page.evaluate(() => {
    const item = document.querySelector('ion-item');
    const checkbox = document.createElement('ion-checkbox');
    item?.appendChild(checkbox);
  });

  await page.waitForChanges();

  const checkbox = page.locator('ion-checkbox');
  await expect(checkbox).not.toBeChecked();

  // Test that clicking on the left edge of the item toggles the checkbox
  await item.click({
    position: {
      x: 5,
      y: 5,
    },
  });

  await expect(checkbox).toBeChecked();
});
```

soundproofboot: Thank you, and thank you for writing the test, I'll get it in today.

soundproofboot: I added the test and verified it passed, removed the changes to `item` and verified it failed. Thanks for your help.

PR #30520, opened 2025-06-30
fix(modal): support iOS card view transitions for viewport changes

### LINKED ISSUE #30296: bug: Switching between card modal & "regular" modal when modal is still open causes page behind it to remain in card modal state
### Prerequisites

### Ionic Framework Version

v8.x

### Current Behavior

Currently I have an implementation of the modals where we use card modals in our iOS app. For accessibility reasons, we've recently enabled landscape mode for our app. Since card modals are very small in landscape mode, I want to only include the presentingElement on the modal when the screen is in portrait mode. I am doing this by setting a presentingElement state variable to null, which in turn sets the presentingElement property on undefined in the inline modal. That last part is only necessary since presentingElement only accepts an HTML element or undefined and not null.

When the modal is opened, it regularly opens like a card modal as expected. When you turn your device, the modal changes into a full screen "regular" modal. When you close the full screen modal, however, the page behind it (which was used as the presentingElement) is still "boxed" in because of the remnants of the card modal animation in the background. This can only be reversed by opening the modal as a card modal again and closing it. The result looks like this:

Image

While I expect the page to look like this:

Image

### Expected Behavior

I would expect the page behind it to return to its regular state when the presentingElement is no longer set on the inline modal and the page to return to the full width/length of the device.

The flipping of modals from card modals to full screen modals is a pattern that can be observed in more iOS apps, so should not be very niche. Whatsapp for instance behaves this way in every "New" dialog (such as creating a new chat), when opening the new chat card modal and flipping your device, it will turn into a full screen modal.

### Steps to Reproduce

1. Clone the repo, install the dependencies and open the simulator (or connect your own device and deploy it on there)
2. In Portrait mode, click the blue OPEN MODAL button.
3. See the card modal open
4. Now flip your device to landscape mode (any side is fine)
5. See the full screen modal replace the card modal
6. Now close the modal using the blue "CLOSE MODAL" button
7. See the bugged screen behind it

To revert it back, do the following:

1. Flip your phone back to portrait mode (upside down is turned off, so only "regular" portrait works)
2. Notice that the page is still looking bugged (!!!)
3. Now open the modal again using the button
4. See the card modal appear
5. Click the CLOSE MODAL button
6. Notice that the page returns to normal
7. Flip the phone again
8. Notice that the page remains full screen.

### Code Reproduction URL

https://github.com/ijsje71/bug-switching-card-modal-and-regular-modal#

### Ionic Info

Ionic:

   Ionic CLI       : 7.2.0 (/usr/local/lib/node_modules/@ionic/cli)
   Ionic Framework : @ionic/react 8.5.1

Capacitor:

   Capacitor CLI      : 7.1.0
   @capacitor/core    : 7.1.0
   @capacitor/ios     : 7.1.0

Utility:

   native-run  : 2.0.1

System:

   NodeJS : v22.14.0
   npm    : 10.9.2
   OS     : macOS Unknown

### Additional Information

Since the buggy screen still appears when the modal is closed & the phone is flipped and it's fixed when the modal is opened and closed, it seems that it's related to something not reverting the presentingElement when the dismiss is not called on the modal while the presentingElement is still there. 

Hopefully a fix can be implemented that allows us to keep the modal open, as seen on other applications like the one I mentioned in the expected behaviour section.

ShaneK:
Issue number: resolves #30296

---------

Currently, there is no support for moving between an iOS card view (mobile, portrait modal with presenting element) to a non-card view when the resolution changes (e.g., the device goes from a portrait layout to landscape). This causes issues both way because modals that should be card modals when the user transitions to a portrait view stay as non-card modals and modals that were card modals when they were opened but the user goes to landscape view end up with a black box stuck around the edges of the screen. 

## What is the new behavior?

With this change, we now fully support transitioning between the two modal views when the resolution changes. This should fix the issue where the background could become stuck and should be a nicer experience for users switching between the two orientations while using modals.

- [ ] Yes
- [X] No


mikaelkaron:
Just happened to scroll passed this issue and tried your example in firefox/desktop. I'm seeing a weird error that could (probably is) entirely unrelated to this fix but I felt that it would be worth mentioning.

- open the [test screen](https://ionic-framework-git-fw-6596-ionic1.vercel.app/src/components/modal/test/card?ionic:mode=ios) in full screen mode
- click "card modal" button
- resize screen to trigger IOS card-view
- pull the modal card down (without releasing) and then bring it back up to the top
- resize screen back to full-screen
- modal is now stuck in the bottom (but with a ghost shadow where it _should_ have been) 
attached
- resize back to mobile view
- modal is now hidden off-screen attached

since I've never really tried the IOS stuff on a desktop before I don't know if this is existing or a regression.

ShaneK:
@mikaelkaron That's weird! This issue is still a work in progress, turns out there's a lot of issues here with nested cards I need to hammer out. This is far more complex than I was initially thinking it was, unfortunately. I'll try to see what's up with that too, hopefully it's part of some of the weirdness already ongoing. Thanks for trying it!

core/src/components/modal/animations/ios.transition.ts
```diff
@@ -0,0 +1,198 @@
...
+    const wrapperAnimation = createAnimation()
+      .addElement(root.querySelectorAll('.modal-wrapper, .modal-shadow')!)
+      .fromTo('opacity', '1', '1'); // Keep wrapper visible
+
+    const backdropAnimation = createAnimation()
+      .addElement(root.querySelector('ion-backdrop')!)
+      .fromTo('opacity', 'var(--backdrop-opacity)', 'var(--backdrop-opacity)'); // Keep backdrop visible
+
+    // Animate presentingEl from normal state to portrait state
+    const transformOffset = !CSS.supports('width', 'max(0px, 1px)') ? '30px' : 'max(30px, var(--ion-safe-area-top))';
```
brandyscarney, line 145: Where is this 30px coming from?

core/src/components/modal/modal.tsx
```diff
@@ -963,6 +975,149 @@ export class Modal implements ComponentInterface, OverlayInterface {
...
+    // Only enable for iOS card modals when no custom animations are provided
+    if (getIonMode(this) !== 'ios' || !this.presentingElement || this.enterAnimation || this.leaveAnimation) {
+      return;
+    }
+
+    // Set initial view state
+    this.currentViewIsPortrait = window.innerWidth < 768;
+
+    // Create debounced resize handler
+    this.resizeListener = () => {
```
brandyscarney: Why did we go with this over:

```
@Listen('resize', { target: 'window' })
```

See content & col components

brandyscarney (approved):
Non-blocking questions on the implementation, but looks great!

core/src/components/modal/animations/ios.transition.ts
ShaneK: Mirroring functionality from leave/enter:
https://github.com/ionic-team/ionic-framework/blob/d3b968b4439d0cf75a2d435391b6491730986f80/core/src/components/modal/animations/ios.leave.ts#L65

core/src/components/modal/modal.tsx
ShaneK: Good question, let me fix it!

ShaneK: Fixed in [278c2d7](https://github.com/ionic-team/ionic-framework/pull/30520/commits/278c2d7968323772e47abcddb9ad636536ccf5d2)

core/src/components/modal/animations/ios.transition.ts
ShaneK: I did try to refactor ios.transition.ts and leave/enter to reference the same base functions or variables but the refactor was turning out worse than the current implementation and far more confusing. Maybe in the future we can try again so it's less repeated magic variables like this one

PR #30522, opened 2025-06-30
docs(angular): update testing ng-add

soundproofboot:
Issue number: resolves [2091 in docs](https://github.com/ionic-team/ionic-docs/issues/2091)

---------

Documentation to test `ng add` schematic for @ionic/angular is outdated.

## What is the new behavior?

- More steps added to install and build libraries, change using `npx schematics` for `ng add`.

- [ ] Yes
- [x] No

### DIFF
--- a/packages/angular/README.md
+++ b/packages/angular/README.md
@@ -21,35 +21,47 @@ Ionic Angular specific building blocks on top of [@ionic/core](https://www.npmjs
 ## Testing ng-add in ionic
 
 1. Pull the latest from `main`
-2. Build ionic/angular: `npm run build`
-3. Run `npm link` from `ionic/angular/dist` directory
-4. Create a blank angular project
+2. Install dependencies and build `core` package:
 
 ```
-ng new add-test
-// Say yes to including the router, we need it
-cd add-test
+$ cd core
+$ npm install
+$ npm run build
 ```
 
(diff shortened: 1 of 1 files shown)


brandyscarney:
I ran into some errors while testing out these steps and found that we needed to add a fix to support the latest Angular build system. I created a PR [here](https://github.com/ionic-team/ionic-framework/pull/30525) with updated steps & a fix. I will add you as a co-author on that PR, but could you let me know if you see any issues? Thank you! 🙂

soundproofboot:
> I ran into some errors while testing out these steps and found that we needed to add a fix to support the latest Angular build system. I created a PR [here](https://github.com/ionic-team/ionic-framework/pull/30525) with updated steps & a fix. I will add you as a co-author on that PR, but could you let me know if you see any issues? Thank you! 🙂

@brandyscarney Yes, I have not gotten the schematic to work any which way with Angular 20 but assumed that was a separate issue and wasn't sure how to approach that. I will test out the steps in that PR tomorrow and comment back here with the results.

soundproofboot:
@brandyscarney I verified I was getting that same error with Angular 20.0.5 and then followed the new steps on the fix branch. Everything was easy to follow and installed with no issues. I also tried with 19 and works the same. Thank you, I had tried to debug the schematic but did not get very far.

PR #30544, opened 2025-07-12
fix(modal): dismiss modal when parent element is removed from DOM

### LINKED ISSUE #30389: bug: [Angular] inline ion-modal is not and cannot be dismissed when the parent component is destroyed
### Prerequisites

### Ionic Framework Version

v8.x

### Current Behavior

When opening an ion-modal inline, if the parent component gets destroyed when the modal is open, the modal is not dismissed nor destroyed

### Expected Behavior

The modal should close when the parent component is destroyed

### Steps to Reproduce

1. Create an inline `ion-modal` inside a child-component
2. Destroy the Child component (`ngIf` or similar)

### Code Reproduction URL

https://stackblitz.com/edit/tmgkbwps

### Ionic Info

Ionic:

   Ionic CLI                     : 7.2.0
   Ionic Framework               : @ionic/angular 8.5.6
   @angular-devkit/build-angular : 19.2.6
   @angular-devkit/schematics    : 19.2.6
   @angular/cli                  : 19.2.6

Capacitor:

   Capacitor CLI      : 7.2.0
   @capacitor/android : 7.2.0
   @capacitor/core    : 7.2.0

Cordova:

Utility:

   native-run  : 2.0.1

System:

   NodeJS : v20.18.3 (/Users/me/.volta/tools/image/node/20.18.3/bin/node)
   npm    : 10.8.2
   OS     : macOS Unknown

### Additional Information

I have not found any workaround for this besides use some flag inside the parent component to "render the content", which is dangerous since we can not prevent a regular Angular `if` usage in the future.

ShaneK:
Issue number: resolves #30389

---------

Currently, when the element an ion-modal was presented from is removed, the modal stays presented and can be broken depending on the framework. This is unlike #30540, where children of open modals were being kept open. In this case, specifically the DOM element is being removed for whatever reason and the modal is staying open.

## What is the new behavior?

We're now identifying our parent component on load and watching it with a mutation observer to determine if it gets removed from the DOM. If it does, we trigger a dismiss. This, conveniently, works nicely with #30540 and will dismiss all children and grandchildren as well. 

- [ ] Yes
- [X] No

## Other information

The issue this resolves was already marked closed, but on closer inspection I determined that was a mistake. I believed this issue was related to another one I was dealing with and it is, but it wasn't quite the same.

After this issue is merged, I believe we will have handled all avenues of possibly ending up with broken modals because of parent elements or modals being removed.

### DIFF
--- a/core/src/components/modal/modal.tsx
+++ b/core/src/components/modal/modal.tsx
@@ -96,6 +96,11 @@ export class Modal implements ComponentInterface, OverlayInterface {
   private viewTransitionAnimation?: Animation;
   private resizeTimeout?: any;
 
+  // Mutation observer to watch for parent removal
+  private parentRemovalObserver?: MutationObserver;
+  // Cached original parent from before modal is moved to body during presentation
+  private cachedOriginalParent?: HTMLElement;
+
   lastFocus?: HTMLElement;
   animation?: Animation;
 
@@ -398,6 +403,7 @@ export class Modal implements ComponentInterface, OverlayInterface {
   disconnectedCallback() {
     this.triggerController.removeClickListener();
     this.cleanupViewTransitionListener();
+    this.cleanupParentRemovalObserver();
   }
 
   componentWillLoad() {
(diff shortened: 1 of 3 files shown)


brandyscarney (approved):
Fix looks good! Great work! 🤖

PR #30547, opened 2025-07-14
fix(segment-view): scroll to correct content when height is not set

### LINKED ISSUE #30543: bug: iOS: Swipeable second segment view not loading as default despite being selected
### Prerequisites

### Ionic Framework Version

v8.x

### Current Behavior

I'm trying to load the second segment view as the default when the screen appears. However, on iOS, the first segment view keeps showing even though the second segment button is already selected.

I'm attaching the relevant code below. I believe this might be a bug.
Has anyone else encountered the same issue?

Image

### Expected Behavior

It should show the second segment by default when loading the page. 

### Steps to Reproduce

Write this code in ngModule based project and simply run it. 
already have assigned the ### value="second"

` <ion-segment value="second">
  <ion-segment-button value="first" content-id="first">
    <ion-label>First</ion-label>
  </ion-segment-button>
  <ion-segment-button value="second" content-id="second">
    <ion-label>Second</ion-label>
  </ion-segment-button>
  <ion-segment-button value="third" content-id="third">
    <ion-label>Third</ion-label>
  </ion-segment-button>
</ion-segment>
<ion-segment-view>
  <ion-segment-content id="first">First</ion-segment-content>
  <ion-segment-content id="second">Second</ion-segment-content>
  <ion-segment-content id="third">Third</ion-segment-content>
</ion-segment-view>`

### Code Reproduction URL

https://ionicframework.com/docs/api/segment

### Ionic Info

Ionic:

   Ionic CLI                     : 7.2.0 (/Users/macos/.nvm/versions/node/v23.1.0/lib/node_modules/@ionic/cli)
   Ionic Framework               : @ionic/angular 8.4.3
   @angular-devkit/build-angular : 19.1.8
   @angular-devkit/schematics    : 19.1.8
   @angular/cli                  : 19.1.8
   @ionic/angular-toolkit        : 12.1.1

Capacitor:

   Capacitor CLI      : 7.0.1
   @capacitor/android : 7.0.1
   @capacitor/core    : 7.0.1

Utility:

   cordova-res : 0.15.4
   native-run  : 2.0.1

System:

   NodeJS : v23.1.0 (/Users/macos/.nvm/versions/node/v23.1.0/bin/node)
   npm    : 10.9.0
   OS     : macOS Unknown

### Additional Information

_No response_

brandyscarney:
Issue number: resolves #30543

---------

On desktop Safari and Android the segment view is not switching to the right segment content when the height is dynamically set or read in as 0. This can be reproduced in the following StackBlitz on Android: https://6dhropnr-aheyyprl.stackblitz.io/

## What is the new behavior?
- Sets `min-height` to `1px` on `ion-segment-content` so that it continues to work with a dynamically set height
- Adds an e2e test for Safari to verify the correct content is displayed when it contains a dynamically loaded image

- [ ] Yes
- [x] No

## Other information

Dev build: `8.6.2-dev.11752524287.1d61cd78`

[Preview](https://ionic-framework-git-fw-6586-ionic1.vercel.app/src/components/segment-view/test/dynamic-height)

### DIFF
--- a/core/src/components/segment-view/test/dynamic-height/segment-view.e2e.ts
+++ b/core/src/components/segment-view/test/dynamic-height/segment-view.e2e.ts
@@ -59,6 +59,13 @@ configs({ modes: ['md'], directions: ['ltr'] }).forEach(({ title, config }) => {
       // Wait for the content to be scrolled
       await page.waitForChanges();
 
+      // Wait for the image to load and be visible
+      const imgLocator = page.locator('ion-segment-content#third ion-img');
+      await imgLocator.waitFor({ state: 'visible', timeout: 10000 });
+
+      // Wait for any layout adjustments
+      await page.waitForChanges();
+
       // Check that the third content is visible
       const segmentView = page.locator('ion-segment-view');
       const thirdContent = page.locator('ion-segment-content#third');
--- a/core/src/components/segment-view/test/dynamic-height/segment-view.e2e.ts
+++ b/core/src/components/segment-view/test/dynamic-height/segment-view.e2e.ts
@@ -1,0 +1,78 @@
+import { expect } from '@playwright/test';
+import { configs, test } from '@utils/test/playwright';
+
+/**
+ * This behavior does not vary across modes/directions
+ */
+configs({ modes: ['md'], directions: ['ltr'] }).forEach(({ title, config }) => {
+  test.describe(title('segment-view: dynamic height'), () => {
+    test('should show the third content when clicking the third button', async ({ page, skip }) => {
+      // Skip this test on Chrome and Firefox
+      skip.browser('firefox', 'Original issue only happens on Safari.');
+      skip.browser('chromium', 'Original issue only happens on Safari.');
+
+      await page.setContent(
+        `
+        <style>
+          ion-segment-content {
+            display: flex;
+            align-items: center;
+            justify-content: center;
+            height: fit-content;
+          }
+        </style>
+
+        <ion-segment>
+          <ion-segment-button value="first" content-id="first">
+            <ion-label>First</ion-label>
+          </ion-segment-button>
+          <ion-segment-button value="second" content-id="second">
+            <ion-label>Second</ion-label>
+          </ion-segment-button>
+          <ion-segment-button value="third" content-id="third">
+            <ion-label>Third</ion-label>
+          </ion-segment-button>
+        </ion-segment>
+        <ion-segment-view>
+          <ion-segment-content id="first">
+            Zombie ipsum reversus ab viral inferno, nam rick grimes malum cerebro. De carne lumbering animata corpora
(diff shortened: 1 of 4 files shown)


core/src/components/segment-view/test/dynamic-height/segment-view.e2e.ts
```diff
@@ -0,0 +1,78 @@
...
+            mauris. Hi mindless mortuis soulless creaturas, imo evil stalking monstra adventus resi dentevil vultus
+            comedat cerebella viventium. Qui animated corpse, cricket bat max brucks terribilem incessu zomby. The
+            voodoo sacerdos flesh eater, suscitat mortuos comedere carnem virus. Zonbi tattered for solum oculi eorum
+            defunctis go lum cerebro. Nescio brains an Undead zombies. Sicut malus putrid voodoo horror. Nigh tofth eliv
+          </ion-segment-content>
+          <ion-segment-content id="second">
+            <ion-input value="" label="Email"></ion-input>
+          </ion-segment-content>
+          <ion-segment-content id="third">
+            <ion-img class="img-part" src="https://picsum.photos/200/300"></ion-img>
```
ShaneK, line 49: I can verify this works, however this test is going to be quite annoying for us at some point in the future, if not immediately. The problem is, it doesn't wait for the image to load before verifying the height of the third tab, which means we're depending on the speed of our internet and this image hosts' bandwidth.

I'd recommend either changing the way this test works, or adding something that waits for the image to load like this:

```typescript
      await page.waitForFunction(() => {
        const ionImg = document.querySelector('ion-segment-content#third ion-img');
        if (!ionImg) return false;

        // Check if the shadow DOM img element is loaded
        const shadowRoot = ionImg.shadowRoot;
        if (!shadowRoot) return false;

        const img = shadowRoot.querySelector('img');
        return img && img.complete && img.naturalHeight !== 0;
      });
```

This happened to me locally while testing, BTW. That's how I found the issue 😅 
Screenshot-2025-07-16 at 06 26 26

ShaneK (approved):
I've verified this fix works in Safari and generally approve. However, I do think the test may cause us grief in the future and recommend changing it.

core/src/components/segment-view/test/dynamic-height/segment-view.e2e.ts
brandyscarney: Good catch! Can you try it now? https://github.com/ionic-team/ionic-framework/pull/30547/commits/424f62a004b1c2ee67b3506383db6aa71abbc04b

ShaneK (approved):

core/src/components/segment-view/test/dynamic-height/segment-view.e2e.ts
ShaneK: Works great! Thanks!

PR #30567, opened 2025-07-23
feat(css): add new css utility classes for display and flex utils

### LINKED ISSUE #22469: feat: Complete CSS flex utilities by display and flex properties.
# Feature Request

**Ionic version:**
[x] **5.x** (5.4.3)

It's a great idea to provide these css utilities. Every great UI framework should have it. But some essential css utilities are missing. 
For example a class for `display: flex` and `flex: 1` etc. Nothing special, only the very common flex attributes.

A good example is [Vuetify](https://vuetifyjs.com/en/styles/flex/#enabling-flexbox) (and [here](https://vuetifyjs.com/en/styles/display/#display)). I thinking about classes like: 

```scss
.ion-flex-container {
  display: flex;
}

.ion-flex-column {
  flex-direction: column;
}

.ion-flex {
  flex: 1;
}

/* And utilities for display attr. like: (or .ion-d-block) */
.ion-display-block {
  display: block;
}

.ion-display-flex {
  display: flex;
}
```

**Describe Alternatives**
I could create my own global classes. But it would be more consistent when these is provided by the UI framework itself. And it's about avoiding global styles.

**Related Code**
See the Ionic [source code](https://github.com/ionic-team/ionic-framework/blob/v5.4.3/core/src/css/flex-utils.scss) and [docs](https://ionicframework.com/docs/layout/css-utilities#flex-properties). 

**Additional Context**
I understand that the current flex classnames are only for modification of alignment in a flex container (e.g. from an Ionic element.) But it would be great if Ionic provides such CSS utilities (like Vuetify) to allow e.g. flex layouts with no need of custom CSS. It's not a question what the developer can or should create. In this case, I could create the whole UI framework like Ionic. It's more a philosophical question about what can a UI framework provide for UI stuff to make the life of developers easier. A build-in solution with no need to add more dependencies or custom css. Providing the power to make something awesome. See it as an idea and decide whether to close it or do it. ... I don't want to compare UI frameworks. So sorry that I notice [Vuetify](https://vuetifyjs.com/en/) again. But I like the way to break imaginary blockages of the strict material design. In my opinion we should not follow incomprehensible restrictions when it's restricts our creativity and freedom. But maybe another topic. Back to this css request... :)

brandyscarney:
Issue number: resolves #22469

---------

CSS classes to change the `display` property and various `flex` properties do not exist.

## What is the new behavior?
- Adds new responsive display classes with the following values: `none`, `inline`, `inline-block`, `block`, `flex`, `inline-flex`, `grid`, `inline-grid`, `table`, `table-cell`, `table-row`
- Adds new responsive flex util classes for the following properties: `align-content`, `align-items`, `align-self`, `justify-content`, `flex-direction`, `flex-wrap`, `flex`, `flex-grow` , `flex-shrink`, `order`
- Adds e2e tests to verify the correct classes are in the CSS files

- [ ] Yes
- [x] No

## Other information

Dev build: `8.6.3-dev.11753378718.15aa6551`

### DIFF
--- a/core/src/components/back-button/test/basic/index.html
+++ b/core/src/components/back-button/test/basic/index.html
@@ -125,7 +125,7 @@
 
         <ion-toolbar color="dark">
           <ion-buttons slot="start">
-            <ion-back-button class="ion-hide"></ion-back-button>
+            <ion-back-button class="ion-display-none"></ion-back-button>
           </ion-buttons>
           <ion-title>Hidden</ion-title>
         </ion-toolbar>
--- a/core/src/components/progress-bar/progress-bar.tsx
+++ b/core/src/components/progress-bar/progress-bar.tsx
@@ -107,6 +107,7 @@ const renderProgress = (value: number, buffer: number) => {
      * When finalBuffer === 1, we use display: none
      * instead of removing the element to avoid flickering.
      */
+    // TODO(FW-6697): change `ion-hide` class to `ion-display-none` or another class
     <div
       class={{ 'buffer-circles-container': true, 'ion-hide': finalBuffer === 1 }}
       style={{ transform: `translateX(${finalBuffer * 100}%)` }}


ShaneK (approved):
Great work, looks good to me! 🎉

PR #30569, opened 2025-07-24
feat(content): inherit display

brandyscarney:
Issue number: resolves #

---------

## What is the new behavior?

-
-
-

- [ ] Yes
- [ ] No


brandyscarney:
This would cause a breaking change for popovers or we would have to set the display back to `block` for them, in which case the inheriting wouldn't work. This needs design work.

PR #30587, opened 2025-07-31
chore(deps): update @stencil/core to v4.36.2

brandyscarney:
Issue number: resolves #

---------

## What is the new behavior?

-
-
-

- [ ] Yes
- [ ] No

### DIFF
--- a/core/package-lock.json
+++ b/core/package-lock.json
@@ -9,7 +9,7 @@
       "version": "8.7.0",
       "license": "MIT",
       "dependencies": {
-        "@stencil/core": "4.33.1",
+        "@stencil/core": "4.36.2",
         "ionicons": "^8.0.13",
         "tslib": "^2.1.0"
       },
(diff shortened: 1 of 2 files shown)


ShaneK:
Also I was curious why this didn't impact split-pane. I can't see anything bad happening there, but I guess I'm not sure why exactly this happened in the first place

brandyscarney:
> Looks good to me. Mildly worried about the `any` declarations on `el.__children` but not sure what to do better there at the moment.

Those were there prior to my change so I left it alone.

> Also I was curious why this didn't impact split-pane. I can't see anything bad happening there, but I guess I'm not sure why exactly this happened in the first place

Split pane is a [shadow component](https://github.com/ionic-team/ionic-framework/blob/fdb77960ad00efd9955025172696c8c5a1fa082b/core/src/components/split-pane/split-pane.tsx#L26) and this only needs to be applied to non-shadow components.

PR #30599, opened 2025-08-04
fix(ion-tab): add fallback to select tab if router integration fails

### LINKED ISSUE #30552: bug: IonTab not selected on dynamic routes (introduced in 8.6.0)
### Prerequisites

### Ionic Framework Version

v8.x

### Current Behavior

When navigating directly to a tab in a dynamic route that includes a Tab Bar component (eg: page/1/home), the tab is not selected.

This issue is introduced in 8.6.0. In the provided example it can be reproduced changing the dependencies version.

### Expected Behavior

When navigating directly to a tab in a dynamic route that includes a Tab Bar component (eg: page/1/home), the tab is selected

### Steps to Reproduce

1. Click on any of the dynamic pages in the initial page
2. Observe that the home tab is not selected

### Code Reproduction URL

https://stackblitz.com/edit/sb1-stbfygxm?file=package.json

### Ionic Info

Ionic:

   Ionic CLI : 5.4.16

Utility:

   cordova-res : not installed
   native-run  : not installed

System:

   NodeJS : v20.19.1
   npm    : 10.8.2
   OS     : Linux 5.0

### Additional Information

Version 8.5.9 works as expected. After 8.6.0 it does not (any posterior version).
I've been looking into the commits of that version but could not identify the issue.

Thanks for you help ❤️

ShaneK:
Issue number: resolves #30552

---------

Something cause a timing shift in v8.6.0 that messed up the timing required for react router to set the active tab ID. Currently, when the router goes to set the tab ID, it's possibly too early and the tab may not exist yet, causing it to go unset. 

## What is the new behavior?

This PR is a workaround that allows tabs to check when they're rendered if a tab should be selected as a fallback for the router not setting them. I don't think the tabs, in the long run, should be responsible for this, but I think this is a good intermediate step until the upcoming react router upgrade, when we can look into a better solution for react router that may require less timing precision.

- [ ] Yes
- [X] No


core/src/components/tabs/tabs.tsx
```diff
@@ -68,7 +68,26 @@ export class Tabs implements NavOutlet {
...
     const tabBar = this.el.querySelector('ion-tab-bar');
     if (tabBar) {
-      const tab = this.selectedTab ? this.selectedTab.tab : undefined;
+      let tab = this.selectedTab ? this.selectedTab.tab : undefined;
+
+      // Fallback: if no selectedTab is set but we're using router mode,
+      // determine the active tab from the current URL. This works around
+      // timing issues in React Router integration where setRouteId may not
+      // be called in time for the initial render.
+      if (!tab && this.useRouter && typeof window !== 'undefined') {
```
brandyscarney, line 78: Can we create a ticket to investigate removing this for React Router 6 and add a TODO comment above this to remove it like:

```
// TODO(FW-1234): Remove this with React Router upgrade
```

ShaneK: Done, but I stupidly amended it so I don't have a commit to point you to. [Here's the line, though.](https://github.com/ionic-team/ionic-framework/pull/30599/files#diff-c81e7421844580393760b9b88460d44537261b56061b496988ff6d39647018bbR77)

packages/react/test/base/src/App.tsx
```diff
@@ -63,6 +64,7 @@ const App: React.FC = () => (
         <Route path="/navigation" component={NavComponent} />
         <Route path="/tabs" component={Tabs} />
         <Route path="/tabs-basic" component={TabsBasic} />
+        <Route path="/tabs-direct-navigation" component={TabsDirectNavigation} />
```
brandyscarney, line 67: Can we add this to `Main.tsx`:

```tsx
<IonItem routerLink="/tabs-direct-navigation">
  <IonLabel>Tabs with Direct Navigation</IonLabel>
</IonItem> 
```

brandyscarney (approved):
LGTM!

packages/react/test/base/src/App.tsx
ShaneK: https://github.com/ionic-team/ionic-framework/pull/30602

PR #30602, opened 2025-08-05
test(react): general navigation improvement

ShaneK:

---------

Manual navigation in the react tests is very finnicky

## What is the new behavior?

Improved react test navigation so it more consistently moves from page to page if you're manually reviewing react tests. This was done mostly by adding missing `ion-page` components, but also by adding an exact match to the main component. They're still pretty shaky and not great, but better than before.

- [ ] Yes
- [X] No

### DIFF
--- a/packages/react/test/base/src/pages/TabsDirectNavigation.tsx
+++ b/packages/react/test/base/src/pages/TabsDirectNavigation.tsx
@@ -78,7 +78,7 @@ const TabsDirectNavigation: React.FC = () => {
         </IonTabButton>
 
         <IonTabButton tab="library" href="/tabs-direct-navigation/library" data-testid="library-tab">
-          <IonIcon icon={libraryOutline}></IonIcon>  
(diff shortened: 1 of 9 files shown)


packages/react/test/base/src/pages/TabsDirectNavigation.tsx
```diff
@@ -78,17 +78,17 @@ const TabsDirectNavigation: React.FC = () => {
...
         </IonTabButton>
 
         <IonTabButton tab="search" href="/tabs-direct-navigation/search" data-testid="search-tab">
           <IonIcon icon={searchOutline}></IonIcon>
           <IonLabel>Search</IonLabel>
         </IonTabButton>
-      </IonTabBar>
-    </IonTabs>
+        </IonTabBar>
+      </IonTabs>
```
brandyscarney:
```suggestion
      </IonTabBar>
    </IonTabs>
```

brandyscarney:
Great improvements here! ⭐

PR #30622, opened 2025-08-11
fix(select): add Enter keypress to dismiss

### LINKED ISSUE #30561: bug: Select-Popover does not select the element when enter is pressed
### Prerequisites

### Ionic Framework Version

v8.x

### Current Behavior

Image

Pressing enter when selecting an item does nothing. Only space button works.

### Expected Behavior

It should select it, like when pressing the Space button.

### Steps to Reproduce

1. Create and open ion-select with interface='popover'
2. Navigate to it with keyboard
3. Select an option and press enter

### Code Reproduction URL

-

### Ionic Info

    "@ionic/angular": "8.5.6",

### Additional Information

This is the exact cause in the code:

From https://github.com/ionic-team/ionic-framework/blob/5f12cf8c0428ed347bb9753a1dd22088bf46b3b4/core/src/components/select-popover/select-popover.tsx#L163

```html
            <ion-radio
              value={option.value}
              disabled={option.disabled}
              onClick={() => this.dismissParentPopover()}
              onKeyUp={(ev) => {
                if (ev.key === ' ') {
                  /**
                   * Selecting a radio option with keyboard navigation,
                   * either through the Enter or Space keys, should
                   * dismiss the popover.
                   */
                  this.dismissParentPopover();
                }
              }}
            >
```

The comment said it should work with Enter key, but the code only checks for space key. This is created from PR https://github.com/ionic-team/ionic-framework/commit/7578aa3c598451b6df1cf1eca26ad288c0526121 . It has been unchanged since then.

The fix is simply add enter checking there:

```js
if (ev.key === ' ' || ev.key === 'Enter') {
```

Let me know if the Ionic team is online with this suggestion and I'm happy to create the PR to contribute.

Thank you!

haryopba:
Issue number: resolves #30561

---------

Pressing `Enter` does not dismiss modal or popover in `select-modal` or `select-popover`

## What is the new behavior?

- Pressing `Enter` dismisses modal or popover in `select-modal` or `select-popover`

- [ ] Yes
- [x] No

## Other information
N/A

### DIFF
--- a/core/src/components/radio-group/radio-group.tsx
+++ b/core/src/components/radio-group/radio-group.tsx
@@ -212,8 +212,8 @@ export class RadioGroup implements ComponentInterface {
       }
 
       // Update the radio group value when a user presses the
-      // space bar on top of a selected radio
-      if ([' '].includes(ev.key)) {
+      // enter or space bar on top of a selected radio
+      if (['Enter', ' '].includes(ev.key)) {
         const previousValue = this.value;
         this.value = this.allowEmptySelection && this.value !== undefined ? undefined : current.value;
         if (previousValue !== this.value || this.allowEmptySelection) {
--- a/core/src/components/radio-group/test/basic/radio-group.e2e.ts
+++ b/core/src/components/radio-group/test/basic/radio-group.e2e.ts
@@ -26,7 +26,7 @@ configs({ modes: ['ios'], directions: ['ltr'] }).forEach(({ title, config }) => 
         config
       );
 
-      await radioFixture.checkRadio('keyboard');
+      await radioFixture.checkRadio({ method: 'keyboard', key: 'Space' });
       await radioFixture.expectChecked(true);
     });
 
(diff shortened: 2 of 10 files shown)


brandyscarney:
Thank you for the PR and tests! It looks like this doesn't work to select the value using `Enter` and there is also an issue where the modal immediately closes when you use `Enter` to open it after selecting something:

https://github.com/user-attachments/assets/d6fe7ba5-59f3-48b7-a8be-89fe1798183e

This also happens on the Popover example. 

I think we need to add something like this to `onKeyDown` in `radio-group` to capture the value:

```tsx
// Handle Enter key for select interfaces
if (['Enter'].includes(ev.key) && inSelectInterface) {
  const previousValue = this.value;
  this.value = current.value;
  if (previousValue !== this.value) {
    this.emitValueChange(ev);
  }
  ev.preventDefault();
}
```

But I'm not sure how to fix the other problem right now.

brandyscarney (changes requested):
Thank you for the fix and tests! It looks like this doesn't work to select the value using `Enter` and there is also an issue where the modal immediately closes when you use `Enter` to open it after selecting something:

https://github.com/user-attachments/assets/d6fe7ba5-59f3-48b7-a8be-89fe1798183e

This also happens on the Popover example. 

I think we need to add something like this to `onKeyDown` in `radio-group` to capture the value:

```tsx
// Handle Enter key for select interfaces
if (['Enter'].includes(ev.key) && inSelectInterface) {
  const previousValue = this.value;
  this.value = current.value;
  if (previousValue !== this.value) {
    this.emitValueChange(ev);
  }
  ev.preventDefault();
}
```

But I'm not sure how to fix the other problem right now.

haryopba:
@brandyscarney, sorry for the late response. The immediate close happens because the modal opens on keydown with `Enter`, instead of on keyup like it does with `Space`. It's a behaviour from `ion-item` and native button, to fix this I use `onKeyDown` attribute instead of `onKeyUp` in `select-modal` and `select-popover`.

brandyscarney:
@haryopba Thank you for taking another look! I considered using `onKeyDown` as well, but if you hold the `Enter` key down you end up with this:

https://github.com/user-attachments/assets/82424917-5594-493f-bfd8-4cd0386e0d3e

I think we'll need to find another solution. 🤔

ShaneK:
Hey, thanks for your PR! [We've gone with a different approach](https://github.com/ionic-team/ionic-framework/pull/31093), but we appreciate your contribution!

The fix for the original issue will be released soon. Thanks again!

Edit: Has been released in [v8.8.5](https://github.com/ionic-team/ionic-framework/releases/tag/v8.8.5)

PR #30633, opened 2025-08-14
fix(checkbox): add aria attributes to ignore checkbox icon

### LINKED ISSUE #30231: bug: ion-checkbox svg visible to assistive technology
### Prerequisites

### Ionic Framework Version

v8.x

### Current Behavior

`ion-checkbox` has an svg checkbox icon that currently appears in the accessibility tree as having role 'image', and can be targeted by screen readers, but with no other contextual information about the svgs, making it difficult to understand.

Using VoiceOver, I get the following experience when stepping into a checkbox and past the label: 
> Image

Scanning the page with analysis tools such as axe devTools will also result in "SVG element missing accessible name" types of errors.

### Expected Behavior

I think the svg should be hidden from AT with `aria-hidden="true"` or similar, so only the entire checkbox control and any label inside is reachable. 

### Steps to Reproduce

1. Navigate to a page with an `ion-checkbox`, e.g. https://stackblitz.com/edit/r1exi21m
2. Browse the page with a screen-reader, and step into the checkbox control.
3. See that svg inside ion-checkbox is reachable

### Code Reproduction URL

https://stackblitz.com/edit/r1exi21m

### Ionic Info

Reproduced in stackblitz.

### Additional Information

_No response_

ShaneK:
Issue number: resolves #30231

---------

Currently, if you're using voiceover on Chrome or Firefox, you can navigate the checkbox in such a way that it announces the checkbox image, which provides no context to what you're actually selecting.

## What is the new behavior?

With this change, you can no longer select the checkbox image.

- [ ] Yes
- [X] No

## Other information

[Screen](https://ionic-framework-git-ionic-56-ionic1.vercel.app/src/components/checkbox/test/basic)

### DIFF
--- a/core/src/components/checkbox/checkbox.tsx
+++ b/core/src/components/checkbox/checkbox.tsx
@@ -328,7 +328,7 @@ export class Checkbox implements ComponentInterface {
             {this.renderHintText()}
           </div>
           <div class="native-wrapper">
-            <svg class="checkbox-icon" viewBox="0 0 24 24" part="container" aria-hidden="true" role="presentation">
+            <svg class="checkbox-icon" viewBox="0 0 24 24" part="container" aria-hidden="true">
               {path}
             </svg>
           </div>


core/src/components/checkbox/checkbox.tsx
```diff
@@ -328,7 +328,7 @@ export class Checkbox implements ComponentInterface {
             {this.renderHintText()}
           </div>
           <div class="native-wrapper">
-            <svg class="checkbox-icon" viewBox="0 0 24 24" part="container">
+            <svg class="checkbox-icon" viewBox="0 0 24 24" part="container" aria-hidden="true" role="presentation">
```
brandyscarney:
```suggestion
            <svg class="checkbox-icon" viewBox="0 0 24 24" part="container" aria-hidden="true">
```

We shouldn't need `role` since it is removed from the accessibility tree by using `aria-hidden`

brandyscarney (changes requested):

core/src/components/checkbox/checkbox.tsx
ShaneK: Good catch! Accidentally left it around from initial testing 😅

PR #12278, opened 2017-07-06
Add import information

b264:
Cannot use ToastController without knowing from where to import it

#### Short description of what this resolves:
The doc is incomplete because this code won't work without first importing the class.  This is a common occurrence across virtually all of the Ionic docs.  This is the one that got me today, so I thought I'd fix it.

#### Changes proposed in this pull request:

- add import to code example in doc

PR #30652, opened 2025-08-23
TypeScript cannot locate the module you're trying to import

KristionB:
Issue number: resolves #12278 

---------

This pull request addresses the TypeScript error "Cannot find module './index' or its corresponding type declarations."

## What is the new behavior?

- module reference is correct
- type declarations are properly resolved
-  improving compatibility and reducing build-time errors.

- [X] No

### DIFF
--- a/core/scripts/custom-elements/custom-elements.d.ts
+++ b/core/scripts/custom-elements/custom-elements.d.ts
@@ -1,2 +1,3 @@
-export * from './index';
-export * from '../dist/types/interface';
+// Check if './index' file exists and is correctly exported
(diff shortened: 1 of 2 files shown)


brandyscarney:
Thanks for the PR! It looks like you have linked to another PR that this fixes instead of an issue. Is there an open issue this fixes?

KristionB:
> Thanks for the PR! It looks like you have linked to another PR that this fixes instead of an issue. Is there an open issue this fixes?

I looked back at it there's no open issue this fixes so my mistake, you're right it fixes an issue.

PR #30655, opened 2025-08-27
fix(tab-button): improve badge position and button states;

JoaoFerreira-FrontEnd:
Issue number: internal
---------

## What is the new behavior?

- define padding space for ionic tab-button;
- guarantee hover and focus occupies the height;
- when has a status-badge the top position should be always the same;
- add example with tab-buttons without label;

- [ ] Yes
- [x] No

## Other information

[preview](https://ionic-framework-git-rou-11804-ionic1.vercel.app/src/components/badge/test/hint?ionic:theme=ionic)

### DIFF
--- a/core/src/components/badge/test/hint/index.html
+++ b/core/src/components/badge/test/hint/index.html
@@ -396,6 +396,31 @@
               </ion-tab-button>
             </ion-tab-bar>
           </div>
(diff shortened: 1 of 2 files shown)


brandyscarney:
Can we close this? @JoaoFerreira-FrontEnd

PR #30658, opened 2025-08-28
feat(col): modernizing Grid component

rugoncalves:

The existing grid component (`ion-grid`, `ion-row`, and `ion-col`) in Ionic was developed several years ago and has not received significant updates since then. As a result, it does not leverage modern CSS features. For example, the gutter (spacing) between columns is implemented using the border property, which is an outdated technique.

## What is the new behavior?

- `--ion-grid-gap`: this new CSS variable, will indicate the gap size in the grid. Defaults to `0px` - the current value.
- `ion-col`: has a new way to calculate the width of the column. Additionally a new property `order` (and responsive variants) was added, and will control the order of the column.
- `ion-row`: uses the newly introduced custom property `--ion-grid-gap`. This property will indicate the gap size in the grid.

- [x] Yes
- [ ] No

The properties `pull` and `push` from `ion-col`, have been removed. The functionality achieved with them, is now achieved with the new property `order` and the existing `size`. More information and migration examples can be read in `BREAKING.md` file.


core/src/components/col/col.scss
```diff
@@ -4,6 +4,11 @@
 // --------------------------------------------------
 
 :host {
+  /**
+   * @prop --col-unit-size: The size of each Column unit.
+   */
+  --col-unit-size: calc((100% - (var(--ion-grid-columns, 12) - 1) * var(--ion-grid-gap)) / var(--ion-grid-columns, 12));
```
BenOsodrac: Would it make sense to instead of 12, add a --grid-type variable here, that by default would be 12?

core/src/components/col/col.tsx
```diff
@@ -55,73 +53,41 @@ export class Col implements ComponentInterface {
...
-  @Prop() pullXl?: string;
-
-  /**
-   * The amount to push the column, in terms of how many columns it should shift to the end
-   * of the total available.
+/**
+   * The order of the column, in terms of where the column should position itself in the columns renderer.
+   * If no value is passed, the column order implicit value will be the order in the html structure.
    */
-  @Prop() push?: string;
```
brandyscarney: Can we deprecate the `push` properties instead of removing them?

core/src/components/col/col.scss
rugoncalves: But what where would `--grid-type` variable come from?
I can also simply delete the the constant `12`, as the variable `--ion-grid-columns` will be initialized with 12.

core/src/components/col/col.tsx
rugoncalves: Let me circle that back and reply to you.

core/src/components/col/col.scss
BenOsodrac: Yes I was suggesting something like that, just create a variable to leave it open for the future, with default value set to 12. If anyone wants they can change that variable to also change the grid columns

core/src/components/col/col.tsx
rugoncalves: As discussed, the properties were re-added, and deprecated. Additionally an warning was added when one of those properties are used:
warning-message

core/src/components/col/col.scss
rugoncalves: As discussed, this is an existing ionic variable `--ion-grid-columns`. We are just using it. The default 12, was **removed** from the implementation. Thank you for the feedback.

BREAKING.md
```diff
@@ -32,3 +33,51 @@ This is a comprehensive list of the breaking changes introduced in the major ver
 <h4 id="version-9x-chip">Chip</h4>
 
 - The `border-radius` of the `ios` and `md` chip now defaults to `10px` and `8px`, respectively, instead of `16px` in accordance with the iOS and Material Design 3 guidelines. To revert to the previous appearance, set the `shape` to `"round"`, or override the `--border-radius` CSS variable to specify a different value.
+
+<h4 id="version-9x-grid">Grid</h4>
+- The properties `pull` and `push`, have been removed. The similar look can be achieved with the newly added property `order`.
```
```suggestion

- The properties `pull` and `push` have been deprecated and no longer work. A similar look can be achieved with the newly added property `order`.
```

BREAKING.md
```diff
@@ -32,3 +33,51 @@ This is a comprehensive list of the breaking changes introduced in the major ver
...
+
+<h5>Example 1: Swap two columns</h5>
```
brandyscarney, line 41: Can we add an example for migrating this:

```html
<ion-grid>
  <ion-row>
    <ion-col size="auto" push="1">
      <div>ion-col push 1</div>
    </ion-col>
    <ion-col size="auto" push="1">
      <div>ion-col push 1</div>
    </ion-col>
  </ion-row>
</ion-grid>
```

BREAKING.md
```diff
@@ -32,3 +33,51 @@ This is a comprehensive list of the breaking changes introduced in the major ver
...
+```
+**Version 9.x+**
+```html
+<ion-grid>
+  <ion-row>
+    <ion-col order="2">1</ion-col>
+    <ion-col size="3" order="1">2</ion-col>
+  </ion-row>
+</ion-grid>
+```
```
brandyscarney, line 137: Can we add an example for migrating size & push/pull with breakpoints:

```html
<ion-grid>
  <ion-row>
    <ion-col size="3" size-md="6" push="9" push-md="6">
      <div>ion-col size="3" size-md="6" push="9" push-md="6"</div>
    </ion-col>
    <ion-col size="9" size-md="6" pull="3" pull-md="6">
      <div>ion-col size="9" size-md="6" pull="3" pull-md="6"</div>
    </ion-col>
  </ion-row>
</ion-grid>
```

brandyscarney (changes requested):
I left a comment on the Jira ticket with more info. 🙂

BREAKING.md
rugoncalves: done

BREAKING.md
rugoncalves: done

core/src/components/grid/test/basic/grid.e2e.ts-snapshots/grid-basic-md-ltr-Mobile-Chrome-linux.png
ShaneK, line 1: These screenshot diffs indicate there's a regression happening here that will mess up the way columns are supposed to be working. I'd need to see this fixed before I could approve this PR, as it indicates there's some kind of hidden width taking up space that shouldn't be, forcing what should be an inline column into the next row.

rugoncalves: Fixed, appreciate the attentive review!

core/src/components/col/col.scss
```diff
@@ -20,10 +27,46 @@
...
+:host([class^="ion-grid-offset-col--"]),
+:host([class*=" ion-grid-offset-col--"]) {
+  --margin-calc: calc(
+    var(--col-unit-size) * var(--ion-grid-col-margin) + (var(--ion-grid-gap, 0px) * var(--ion-grid-col-margin))
+  );
+
+  @include margin-horizontal(var(--margin-calc), 0);
+}
+
+$grid-col-number: 12;
```
ShaneK, line 67: This caps generated classes at 12, but `--ion-grid-columns` is configurable. If somebody sets a higher number of columns in `--ion-grid-columns`, then their width will be calculated for that number of columns, but the number of generated columns will be incorrect. 

Either change how you get this ceiling so it comes from the configured variable, or document that spans/orders/offsets > 12 require custom CSS.

core/src/components/grid/test/offsets-pull-push/grid.e2e.ts
```diff
@@ -0,0 +1,17 @@
+import { expect } from '@playwright/test';
+import { configs, test } from '@utils/test/playwright';
+
+/**
+ * ion-grid does not have different styling per-mode
+ */
+configs({ modes: ['md'] }).forEach(({ title, screenshot, config }) => {
+  test.describe(title('grid: offsets'), () => {
+    test('should not have visual regressions', async ({ page }) => {
+      await page.goto(`/src/components/grid/test/offsets`, config);
```
ShaneK: Is this the right path? It's kind of odd that this test exists under `offsets-pull-push`, but is testing against `offsets`. Maybe this test should be in the offsets e2e and there should be tests here for `offsets-pull-push`? Or maybe you just accidentally used the wrong path? Not sure.

BREAKING.md
```diff
@@ -32,3 +33,105 @@ This is a comprehensive list of the breaking changes introduced in the major ver
...
+    </ion-col>
+  </ion-row>
+</ion-grid>
+```
+**Version 9.x+**
+```html
+<ion-grid>
+  <ion-row>
+    <ion-col size="auto" order="2" order-md="2">
+      <div>ion-col size="3" size-md="6" order="9" order-md="6"</div>
```
ShaneK:
```suggestion
      <div>ion-col size="auto" order="2" order-md="2"</div>
```

The text doesn't reflect what's set here

BREAKING.md
```diff
@@ -32,3 +33,105 @@ This is a comprehensive list of the breaking changes introduced in the major ver
...
+    </ion-col>
+    <ion-col size="auto" order="1" order-md="1">
+      <div>ion-col size="9" size-md="6" order="3" order-md="6"</div>
```
ShaneK:
```suggestion
      <div>ion-col size="auto" order="1" order-md="1"</div>
```
The text doesn't reflect what's set here

BREAKING.md
```diff
@@ -32,3 +33,105 @@ This is a comprehensive list of the breaking changes introduced in the major ver
...
+    </ion-col>
+  </ion-row>
+</ion-grid>
+```
+**Version 9.x+**
+```html
+<ion-grid>
+  <ion-row>
+    <ion-col size="auto" offset="1">
+      <div>ion-col push 1</div>
```
ShaneK:
```suggestion
      <div>ion-col size="auto" offset="1"</div>
```
The text doesn't reflect what's set here

BREAKING.md
```diff
@@ -32,3 +33,105 @@ This is a comprehensive list of the breaking changes introduced in the major ver
...
+    </ion-col>
+    <ion-col size="auto">
+      <div>ion-col push 1</div>
```
ShaneK:
```suggestion
      <div>ion-col size="auto"</div>
```
The text doesn't reflect what's set here

core/src/components/row/row.scss
```diff
@@ -2,9 +2,12 @@
...
 // --------------------------------------------------
-
 :host {
+  --ion-grid-gap: 0px;
+
   display: flex;
 
   flex-wrap: wrap;
+
+  gap: var(--ion-grid-gap, 0px);
```
ShaneK, line 13: This will always be `0px`, I believe, since you're setting this var and using it in the same block. Is this intended? Seems redundant.

ShaneK (changes requested):
This is looking pretty good! Just a few change requests

core/src/components/row/row.scss
rugoncalves: Agree and changed.

core/src/components/grid/test/offsets-pull-push/grid.e2e.ts
rugoncalves: It was a lapse. These tests for now will have show case the deprecated features Pull & Push, not being applied, as they are deprecated. It might be useful for the future. If you think otherwise, we can remove it completely.

core/src/components/col/col.scss
rugoncalves: The configuration `--ion-grid-columns` will be set after the framework is compiled, and while is being used by the developer. And what you're mentioning will be handled by the future Ionic modular approach as talked with @brandyscarney - hence this was left like this. Let me know, if I should leave any special comment regarding the matter.

ShaneK: Cool! It might be worth adding a comment to document that then, currently it just looks like a possibly-unintended issue and somebody reviewing this in the future might not have this context.

core/src/components/row/row.scss
ShaneK: The way you resolved this is sort of the opposite of what I expected. I think you should leave the default and remove the `--ion-grid-gap` definition. Defining it in this block means it has to be overwritten here (specifically, on each row individually) and can't be inherited from anywhere else in the app, which sorta subverts the cascading part of cascading style sheets. I think you should remove the initial definition and leave the 0px fallback.

core/src/components/grid/test/offsets/grid.e2e.ts-snapshots/grid-offsets-md-ltr-Mobile-Chrome-linux.png
rugoncalves, line 1: @ShaneK I see differences in the offset part, but the original, looks incorrect how it was working... but not 100% sure.

ShaneK: Yeah, I agree. I'm not sure why the original was like that, I don't see any reasoning for it in the code. May be missing something, but I think it's fine for now

ShaneK (approved):
Great work! 🎉

PR #30674, opened 2025-09-09
fix(datetime): re-emit wheel ionScrollStart/ionScrollEnd from picker columns

### LINKED ISSUE #30449: bug: ion-datetime and ion-picker-column do not emit reliable change or scroll events, causing stale values on Save
### Prerequisites

### Ionic Framework Version

v8.x

### Current Behavior

When using ion-datetime or ion-picker-column in wheel mode, fast scrolling followed by clicking a Save button leads to saving stale values. This occurs because ionChange is not triggered when the scroll is still in motion or finishes after the click.

### Expected Behavior

We expect events like ionScrollStart and ionScrollEnd to be emitted to detect when scrolling begins and finishes. Alternatively, ionChange should reliably trigger when the final selection is settled, even if the user taps outside the picker or interacts before the scroll ends.

### Steps to Reproduce

1. Add <ion-datetime presentation="date" preferWheel="true"> to a page
2. Scroll one of the columns (e.g. months) quickly
3. Immediately tap a Save button that reads the value
6. Observe that the saved value is incorrect—the scroll hadn’t finished, and ionChange did not fired in time

Make sure to open dev tools and open it as if on mobile

### Code Reproduction URL

https://angular-oq2qea-rsrvuccx.stackblitz.io

### Ionic Info

 @ionic/angular@8.5.8

   @angular-devkit/build-angular : 19.2.14
   @angular-devkit/schematics    : 19.2.14
   @angular/cli                  : 19.2.14

### Additional Information

We need to disable the Save button during wheel scrolling in both `<ion-datetime>` and `<ion-picker-column>`. Currently, there’s no event to detect when scrolling starts or ends. ionChange is not reliably triggered—especially when the wheel is dragged outside of the visual bounds (e.g., fast swipe to top/bottom on mobile).

vunizhona:
Issue number: resolves #30449

---------

Wheel scrolling inside ion-datetime does not provide host-level lifecycle events for app code to react to start/end of scrolling.

## What is the new behavior?

- ion-datetime re-emits wheel scroll lifecycle events from inner columns:
  - ionScrollStart when any wheel column begins scrolling.
  - ionScrollEnd once all active columns settle, coalesced with a 300ms debounce.
  - Events do not bubble to avoid duplicate handling; they are emitted on the host.
- ion-picker-column emits ionScrollStart and ionScrollEnd at scroll start/end.
- API/types updated to expose the new events across frameworks (Angular/Vue proxies).

- [ ] Yes
- [x] No

## Other information

- core/src/components/datetime/datetime.tsx: capture-phase listeners, coalescing logic, new events
- core/src/components/picker-column/picker-column.tsx: emits start/end at scroll boundaries
- core/src/components/datetime/test/prefer-wheel/datetime.spec.ts: replace fake timers with explicit waits
- core/api.txt, core/src/components.d.ts, Angular/Vue proxies: API surface/types for new events
- Proxy changes are generated to reflect the new events(updated during the build).

Verification

- Targeted specs: cd core && npm ci && npm run test.spec -- src/components/datetime/test/prefer-wheel/datetime.spec.ts (expect 3 passed).

- Manual: cd core && npm start, open /src/components/datetime/test/prefer-wheel/index.html.

Scroll a wheel column: app receives one ionScrollStart at the beginning, one ionScrollEnd after settling.


ShaneK:
Hey, thank you so much for this attempt to resolve this issue! I ended up going a different route in #31397 because my main problem with this one is it changes the API surface making it require at least a minor change and an API design discussion. The solution I went with in my PR fixes the issue without making any feature changes.

But we really appreciate you making this PR! It definitely highlighted the issue for us. Thank you!

Issue #30079, opened 2024-12-11
bug: Swipeable Segments wont work with dir=RTL


emnawer:
### Prerequisites

- [X] I have read the [Contributing Guidelines](https://github.com/ionic-team/ionic-framework/blob/main/docs/CONTRIBUTING.md#creating-an-issue).
- [X] I agree to follow the [Code of Conduct](https://ionicframework.com/code-of-conduct).
- [X] I have searched for [existing issues](https://github.com/ionic-team/ionic-framework/issues) that already report this problem, without success.

### Ionic Framework Version

v8.x

### Current Behavior

- Swipeable Segments (`IonSegment` and `IonSegmentView`) do not work if dir="rtl" add to `body` tag. If click on button it wont slide content of the next segment.
- Under Firefox's Responsive Design Mode (mobile): if try to swipe instead of clicking the button it will through an error:_

```
TypeError: button is undefined
    getIndicator ion-segment.js:262
    checkButton ion-segment.js:267
    handleSegmentViewScroll ion-segment.js:333
    hostListenerProxy index.js:2925
    emitEvent index.js:804
    emit index.js:793
    handleScroll ion-segment-view.js:27
    hostListenerProxy index.js:2925
    ael index.js:131
 undefined [index.js:52:61](http://localhost:8100/node_modules/@stencil/core/internal/client/index.js)
    consoleError index.js:52
    hostListenerProxy index.js:2928
    emitEvent index.js:804
    emit index.js:793
    handleScroll ion-segment-view.js:27
    hostListenerProxy index.js:2925
    (Async: EventListener.handleEvent)
    ael index.js:131
    addHostEventListeners index.js:2910
    addHostEventListeners index.js:2906
    connectedCallback index.js:2647
    React 105
    (Async: VoidFunction)
        ensureRootIsScheduled
        scheduleUpdateOnFiber
        enqueueSetState
        setState
    handleHistoryChange IonRouter.tsx:204
    handleHistoryChange IonReactRouter.tsx:37
    listener history.js:155
    notifyListeners history.js:173
    notifyListeners history.js:172
    setState history.js:288
    replace history.js:402
    confirmTransitionTo history.js:145
    replace history.js:385
    handleNavigate IonRouter.tsx:243
    handleNavigateBack IonRouter.tsx:292
    goBack NavManager.tsx:96
    clickButton IonBackButton.tsx:37
    React 23
    (Async: EventListener.handleEvent)
        addEventBubbleListener
[...]

PR #30675, opened 2025-09-09
fix(segment-view): scroll and select the right item when the component is in RTL context;

JoaoFerreira-FrontEnd:
Issue number: resolves [#30079](https://github.com/ionic-team/ionic-framework/issues/30079)

---------

Segments (IonSegment and IonSegmentView) do not work if placed on a dir="rtl" context. If click on button, it won't slide content of the next segment.

## What is the new behavior?

- calculate scroll value having into consideration the dir value;

- [ ] Yes
- [x] No

## Other information

[Preview](https://ionic-framework-git-fw-6768-ionic1.vercel.app/src/components/segment-view/test/rtl)

### DIFF
--- a/core/src/components/segment-view/segment-view.tsx
+++ b/core/src/components/segment-view/segment-view.tsx
@@ -43,7 +43,6 @@ export class SegmentView implements ComponentInterface {
     const max = scrollWidth - clientWidth;
     const scrollRatio = (isRTL(this.el) ? -1 : 1) * (scrollLeft / max);
 
-
     this.ionSegmentViewScroll.emit({
       scrollRatio,
       isManualScroll: this.isManualScroll ?? true,


core/src/components/segment-view/segment-view.tsx
```diff
@@ -39,7 +39,8 @@ export class SegmentView implements ComponentInterface {
   @Listen('scroll')
   handleScroll(ev: Event) {
     const { scrollLeft, scrollWidth, clientWidth } = ev.target as HTMLElement;
-    const scrollRatio = scrollLeft / (scrollWidth - clientWidth);
+    const isRTL = window.getComputedStyle(this.el).direction === 'rtl';
+    const scrollRatio = (isRTL ? -1 : 1) * (scrollLeft / (scrollWidth - clientWidth));
```
ShaneK: Please normalize the scroll position using our normal conventions. 

Example: 
```typescript
const max = scrollWidth - clientWidth; 
const logical = isRTL(this.el) ? (max - scrollLeft) : scrollLeft; 
const scrollRatio = max > 0 ? clamp(0, logical / max, 1) : 0;
```

 This avoids negatives, handles RTL consistently, and keeps `nextIndex` in-bounds. You can import isRTL from `@utils/rtl` and clamp from `@utils/helpers` for consistency with other components.

core/src/components/segment-view/segment-view.tsx
```diff
@@ -127,7 +129,7 @@ export class SegmentView implements ComponentInterface {
     const contentWidth = this.el.offsetWidth;
     this.el.scrollTo({
       top: 0,
-      left: index * contentWidth,
+      left: (isRTL ? -1 : 1) * index * contentWidth,
```
ShaneK: This current approach may not work in certain browsers. Instead of negating the left offset, you may want to do something like:

```typescript
const offset = index * contentWidth; 
const max = this.el.scrollWidth - this.el.clientWidth;
```
 and set `left: isRTL(this.el) ? (max - offset) : offset`. 

I think that should work better, anyway, you may need to play with it some. We should be using `isRTL` here at the very least.

core/src/components/segment-view/test/basic/segment-view.e2e.ts
```diff
@@ -88,86 +88,86 @@ configs({ modes: ['md'], directions: ['ltr'] }).forEach(({ title, config }) => {
...
-      </ion-segment-view>
-    `,
-      config
-    );
-
-    await page
-      .locator('ion-segment-view')
-      .evaluate(
-        (segmentView: HTMLIonSegmentViewElement) => !segmentView.classList.contains('segment-view-scroll-disabled')
+    test('should set correct segment button as checked and show correct content when programmatically setting the segment vale', async ({
```
ShaneK:
```suggestion
    test('should set correct segment button as checked and show correct content when programmatically setting the segment value', async ({
```

core/src/components/segment-view/segment-view.tsx
JoaoFerreira-FrontEnd: I've tried to implement this as proposed, but it doesn't work. Why don't we use the isRTL just to define the signal of the scroll?  like: 
```typescript
const max = scrollWidth - clientWidth; 
const scrollRatio = (isRTL(this.el) ? -1 : 1) * (scrollLeft / max);
```

core/src/components/segment-view/segment-view.tsx
JoaoFerreira-FrontEnd: After testing/debbunging it, `max -offset` when in RTL, sometimes return 0 when it should return an interval. 

It does work fine like: 
```typescript
    const offset = index * contentWidth; 

    this.el.scrollTo({
      top: 0,
      left: (isRTL(this.el) ? -1 : 1 ) * offset,
      behavior: smoothScroll ? 'smooth' : 'instant',
    });
```

core/src/components/segment-view/segment-view.tsx
ShaneK: I was worried about something that I think may not actually be an issue. I'm okay with this, thanks for updating to use the helper method.

Issue #222, opened 2013-11-26
Use sufix or prefix for main classes in ionic.css


sarath2:
First thank for your great library.

eg: 
.io-modal{}  instead of .modal{}  

beocz I used bootstrap for large screens.
 ionic replace bootstrap features.

Otherwise it work nicely.

priand:
FYI, we just faced the problem today. We are assembling a series of components under an ionic app, and some of these components are already built using bootstrap. And we are having conflict, like the definition of .modal. Evan is right, and the burden of having a prefix is out-beaten by a better integration story.

ewinslow:
:) thanks for the vote of confidence. Doubtful we will see the issue fixed though.

Hopefully this problem will just quietly go away as folks move over to shadow Dom. (I.e. angular 2)

yyl000000:
A similar discussion thread in Twitter Bootstrap suggests the _variables.scss file to implement an additional variable such as $prefix. When it is left blank, SASS will generate the clean CSS class names exactly as they are now. When $prefix has a value, SASS will generate prefixed class names. That way, it doesn't affect existing user base, while different teams can select different prefixes to avoid collisions with the UI frameworks that they are currently using. The best of both worlds.

yelhouti:
Having the same problem when trying to mix clarity with ionic, just saying

PR #30678, opened 2025-09-10
feat(tokens): update tokens usage to reflect UI architecture

BenOsodrac:
Issue number: internal

---------

## What is the new behavior?

This pull request updates the color and background variables across many Ionic component styles to use new design system tokens, improving consistency and maintainability. It also temporarily skips several visual regression tests for the `datetime` component due to a known issue (FW-6769).

**Design System Token Migration:**

* Updated `--color` and `--background` variables in many components (e.g., `card`, `card-title`, `list-header`, `item`, `segment`, `radio`, `select`, `textarea`, `input`, `checkbox`, `chip`, `datetime`, `item-options`, `searchbar`) to use the new `globals.$ion-text-default` and `globals.$ion-bg-surface-default` tokens instead of older primitives. This change ensures consistent theming and easier future maintenance. [[1]](diffhunk://#diff-efdc0d0b924ff22b22228f52baebe09c69cce416ab33ba4a572c678ba3aa38f8L16-R16) [[2]](diffhunk://#diff-c213fde8dd24fa2839aa2744fc6d06635c2a8e5f50f93996afa63592299eae84L8-R8) [[3]](diffhunk://#diff-ff9d7e4249d9ff015206ca8f6ce3553d6217ae196bccdb0e4fbd26b93b45489dL8-R13) [[4]](diffhunk://#diff-796ffc72b6171910c25d5aad085bc245095c38f104d55e253e06282702205066L33-R33) [[5]](diffhunk://#diff-796ffc72b6171910c25d5aad085bc245095c38f104d55e253e06282702205066L63-R63) [[6]](diffhunk://#diff-3c751a473bd2fa61cfd77529c7b9e496f34cf198a4aed0bea0b57d1a1f7fe40aL120-R126) [[7]](diffhunk://#diff-5df43981fd8a3a93e815ade41875432d0620892c30ae86858821a1270087fef6L12-R12) [[8]](diffhunk://#diff-5df43981fd8a3a93e815ade41875432d0620892c30ae86858821a1270087fef6L143-R143) [[9]](diffhunk://#diff-5df43981fd8a3a93e815ade41875432d0620892c30ae86858821a1270087fef6L214-R214) [[10]](diffhunk://#diff-cbf91c9b97e425d969df4730e12b3424dd54de7664b9e629d455d2e63539c84eL4-R4) [[11]](diffhunk://#diff-4d89e41650632cecdd15dbe6b2ef5ad30d100788ecd786dd97a990aaf1ab08d9L9-R9) [[12]](diffhunk://#diff-690dbcc4c16a68fc142652cfda82fe7c2008b27a7fcb63dac7e2926598edced2L14-R14) [[13]](diffhunk://#diff-00b462d69386f16806ec2f51ec487d3c47faadfbe68f3b718d336642c80379beL8-R13) [[14]](diffhunk://#diff-bb3e65e73d30ca817f259cc0f805118bfbdee6ce562dd7171afddbca03b7265aL6-R6) [[15]](diffhunk://#diff-c55248a8c9a55b4714cca5b2daa538f3bff0d1260b9249af1b20f7e845a72741L20-R20) [[16]](diffhunk://#diff-c55248a8c9a55b4714cca5b2daa538f3bff0d1260b9249af1b20f7e845a72741L58-R58) [[17]](diffhunk://#diff-c55248a8c9a55b4714cca5b2daa538f3bff0d1260b9249af1b20f7e845a72741L70-R70) [[18]](diffhunk://#diff-36e6af13208ea67992ffd6ac4cb84d0dceea7fb8ca0bcfaccb578d9ac296c9aaL17-R17) [[19]](diffhunk://#diff-8409cb5f296c6ba6c9e0747c7eb6d6f837935bc158bb42406631669d6fcc12b3L8-R8) [[20]](diffhunk://#diff-030647f0cfb9ca6f042e9d7114611c7076a7c01af73076d1db05cc881b40ddf7L9-R9) [[21]](diffhunk://#diff-030647f0cfb9ca6f042e9d7114611c7076a7c01af73076d1db05cc881b40ddf7L165-R165) [[22]](diffhunk://#diff-45914a3d091d2c33732c1bdfbd74ce4375d12c6b23811c211a038d829e110886L10-R16)

**Test Suite Maintenance:**

* Temporarily skipped several E2E visual regression tests for datetime and item components, referencing TODO FW-6769, to prevent test failures while an upstream issue is addressed. [[1]](diffhunk://#diff-28d032b5dd4c52cbf575dd85894b936c883d43093a03b1e5fca940ee3a626cb3L31-R32) [[2]](diffhunk://#diff-6bcce71748a9d3fff9d03f4829f9ee2f058ff2ff419e5a0168f6b1200e29ba0fL33-R34) [[3]](diffhunk://#diff-6bcce71748a9d3fff9d03f4829f9ee2f058ff2ff419e5a0168f6b1200e29ba0fL87-R89) [[4]](diffhunk://#diff-5c6fab9ab164e203679ae4adfafbc0c4c84c569e8fc37f788f5688707cdd283bL16-R17) [[5]](diffhunk://#diff-eac254ce4208c500e31c420e6919b2141fb67c875b062b6d8cbbe3b262216cedL32-R33)

There're some snapshots updates that seem to have no visible differences, so I would say its ok to accept the new ones.

- [ ] Yes
- [x] No


core/src/components/item/item.ionic.scss
```diff
@@ -5,13 +5,12 @@
 // --------------------------------------------------
 
 :host {
-  --background: #{globals.$ion-primitives-base-white};
-  --background-activated: #{globals.$ion-primitives-neutral-200};
+  --background: #{globals.$ion-bg-surface-default};
+  --background-activated: #{globals.$ion-bg-select-default};
```
brandyscarney, line 9: Is this value right? When pressing an item it is now primary instead of gray, see [item/buttons](https://ionic-framework-git-rou-11901-v2-ionic1.vercel.app/src/components/item/test/buttons?ionic:theme=ionic)

Figma seems to use this value: `bg/neutral/subtlest/press` which evaluates to `primitives/neutral/200`

core/src/components/item/item.ionic.scss
```diff
@@ -5,13 +5,12 @@
...
   --border-color: #{globals.$ion-primitives-neutral-300};
   --border-style: #{globals.$ion-border-style-solid};
   --border-width: #{0 0 globals.$ion-border-size-025 0};
-  --color: #{globals.$ion-primitives-neutral-1200};
-  --detail-icon-color: #{globals.$ion-primitives-neutral-800};
```
brandyscarney, line 14: I think this should have been left in, according to Figma:

```scss
  --detail-icon-color: #{globals.$ion-icon-subtlest};
```

core/src/css/ionic/core.ionic.scss
```diff
@@ -16,6 +16,9 @@
       @include globals.generate-color($color-name);
     }
   }
+
+  /* Default background color of all components to default background surface token */
+  --background: #{globals.$ion-bg-surface-default};
```
brandyscarney, line 21: Why is this needed?

core/src/themes/ionic/ionic.theme.default.scss
```diff
@@ -210,8 +210,8 @@ $ionic-colors: (
 // Ionic Tabs & Tab Bar
 // --------------------------------------------------
 
-$tabbar-ionic-background: var(--ion-tab-bar-background, globals.$ion-primitives-base-white);
-$tabbar-ionic-background-activated: var(--ion-tab-bar-background-activated, globals.$ion-primitives-neutral-100);
+$tabbar-ionic-background: var(--ion-tab-bar-background, globals.$ion-bg-surface-default);
+$tabbar-ionic-background-activated: var(--ion-tab-bar-background-activated, globals.$ion-bg-select-default);
```
brandyscarney, line 214: This is using `$ion-semantics-primary-100` instead of the grey primitive. It looks like tabs are using `bg/neutral/subtlest/press` for their pressed state, however this variable `$tabbar-ionic-background-activated` isn't being used at all.

core/src/css/ionic/core.ionic.scss
BenOsodrac: To make it easier to change the background across the Framework, just by changing the bg-surface-default token. For example, something like a Dark Theme, now overriding a couple global tokens does a nice 80/20 on that:

` 
--token-bg-surface-default: #222;
--token-text-default: #fff;
`

core/src/components/item/item.ionic.scss
BenOsodrac: I spoke before with Mariana, and she concluded that the neutral value was not correct for these (probably need to update the Figma now).

The ideia is that we try to use as much as possible the global/default tokens, instead of neutral/white/etc, to make it easier to customize, by doing less css and tokens overrides.

core/src/components/item/item.ionic.scss
BenOsodrac: Same principle as https://github.com/ionic-team/ionic-framework/pull/30678#discussion_r2369897918

core/src/themes/ionic/ionic.theme.default.scss
BenOsodrac: Same principle as https://github.com/ionic-team/ionic-framework/pull/30678#discussion_r2369897918, these activated states should react to primary color changes, instead of using a specific neutral one.
 
And its being used here  😄 : https://github.com/ionic-team/ionic-framework/blob/d4d8d8085af8fc563720b18ab897a8b76ad8c685/core/src/components/tab-bar/tab-bar.ionic.scss#L6

brandyscarney: Oh I had the `components/` directory excluded from my search from a previous search. 😂 My bad.

core/src/css/ionic/core.ionic.scss
brandyscarney: We should use the global theme variables then like this:

```scss
$background-color-value: $ion-bg-surface-default;
$background-color: var(--ion-background-color, $background-color-value);
```

Then the component:

```scss
:host {
  --background: #{$background-color};
}
```

See how our dark theme works: [dark.scss](https://github.com/ionic-team/ionic-framework/blob/2dd212dad222c013c087d2c998ead8413cad0ee7/core/src/css/palettes/dark.scss#L95-L96)

brandyscarney (approved):
Added a small suggestion to tie in with the global theming but since this will likely be re-worked with the modular work it is not a blocker. 👍

core/src/css/ionic/core.ionic.scss
BenOsodrac: But we don't have that for Ionic Theme, like for native: https://github.com/ionic-team/ionic-framework/blob/2dd212dad222c013c087d2c998ead8413cad0ee7/core/src/themes/native/native.theme.default.scss#L124. We would need to have something like this right?

And we would need to adjust all the ionic theme components background rules to have that rule as well?

As you said, we can revisit this then when the modular work is done. For now, this unblocks the OutSystems use-case for easy primary color customization using :root and some token variables.

brandyscarney: Yeah the `$background-color-value` just exists to generate the stepped colors, but yeah let's revisit when we do the modular work!

PR #30685, opened 2025-09-18
feat(textarea): add styles for textarea solid fill

JoaoFerreira-FrontEnd:
Issue number: internal

---------

## What is the new behavior?

- highlight should only appear on md;
- fix some textarea ionic css vars usage;
- create sheet for ionic.solid;

- [ ] Yes
- [x] No

### DIFF
--- a/core/src/components/textarea/textarea.ionic.scss
+++ b/core/src/components/textarea/textarea.ionic.scss
@@ -1,6 +1,7 @@
 @use "../../themes/ionic/ionic.globals.scss" as globals;
 @use "./textarea.common";
 @use "./textarea.ionic.outline.scss" as outline;
+@use "./textarea.ionic.solid.scss" as solid;
 
 // Ionic Textarea
 // --------------------------------------------------
@@ -122,7 +123,7 @@
 // ----------------------------------------------------------------
 
 ion-icon {
-  color: globals.$ion-primitives-neutral-800;
+  --color: globals.$ion-icon-subtlest;
 
   font-size: globals.$ion-scale-400;
 }
@@ -188,7 +189,7 @@ ion-icon {
 
 @media (any-hover: hover) {
   :host(:hover) {
-    --border-color: #{globals.$ion-primitives-neutral-600};
+    --border-color: #{globals.$ion-border-focus-default};
   }
 }
 
@@ -196,10 +197,10 @@ ion-icon {
 // ----------------------------------------------------------------
 
 :host(.textarea-disabled) {
-  --color: #{globals.$ion-primitives-neutral-500};
-  --background: #{globals.$ion-primitives-neutral-100};
-  --border-color: #{globals.$ion-primitives-neutral-300};
-  --placeholder-color: #{globals.$ion-primitives-neutral-500};
+  --color: #{globals.$ion-text-disabled};
+  --background: #{globals.$ion-bg-input-disabled};
+  --border-color: #{globals.$ion-border-disabled};
+  --placeholder-color: #{globals.$ion-text-disabled};
 }
 
 :host(.textarea-disabled:not(.ion-valid)) .textarea-bottom .helper-text,
@@ -229,7 +230,7 @@ ion-icon {
 // ----------------------------------------------------------------
 
 :host(.textarea-readonly) {
-  --background: #{globals.$ion-primitives-neutral-100};
+  --background: #{globals.$ion-border-input-read-only};
 }
 
 /**
--- a/core/src/components/textarea/textarea.ionic.solid.scss
+++ b/core/src/components/textarea/textarea.ionic.solid.scss
@@ -1,0 +1,58 @@
+@use "../../themes/ionic/ionic.globals.scss" as globals;
+
+// Ionic Textarea Fill: Solid
+// ----------------------------------------------------------------
+
+:host(.textarea-fill-solid) {
+  --border-color: #{globals.$ion-bg-input-bold-default};
+  --background: #{globals.$ion-bg-input-bold-default};
+}
+
+:host(.textarea-fill-solid) .textarea-wrapper {
+  border-bottom: none;
+  background: none;
+}
+
+:host(.textarea-fill-solid) .textarea-wrapper-inner {
+  --border-width: #{globals.$ion-border-size-050};
+  @include globals.border-radius(var(--border-radius));
+  position: relative;
+  background: var(--background);
+  border: var(--border-width) var(--border-style) var(--border-color);
+}
+
+:host(.textarea-fill-solid) .textarea-bottom {
+    --border-width: #{globals.$ion-scale-0};
+}
+
+// Focus
+// ----------------------------------------------------------------
+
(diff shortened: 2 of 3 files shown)


core/src/components/textarea/textarea.ionic.scss
```diff
@@ -229,7 +230,7 @@ ion-icon {
 // ----------------------------------------------------------------
 
 :host(.textarea-readonly) {
-  --background: #{globals.$ion-primitives-neutral-100};
+  --background: #{globals.$ion-border-input-read-only};
```
brandyscarney:
```suggestion
  --background: #{globals.$ion-bg-input-read-only};
```

core/src/components/textarea/textarea.ionic.solid.scss
```diff
@@ -0,0 +1,60 @@
...
+  position: relative;
+
+  border: var(--border-width) var(--border-style) var(--border-color);
+
+  background: var(--background);
+}
+
+:host(.textarea-fill-solid) .textarea-bottom {
+  --border-width: #{globals.$ion-scale-0};
+}
```
brandyscarney, line 29:
```suggestion
```

This doesn't appear to do anything

core/src/components/textarea/textarea.ionic.solid.scss
```diff
@@ -0,0 +1,60 @@
...
+
+:host(.textarea-fill-solid) .textarea-wrapper {
+  border-bottom: none;
+
+  background: none;
+}
+
+:host(.textarea-fill-solid) .textarea-wrapper-inner {
+  --border-width: #{globals.$ion-border-size-050};
```
brandyscarney: Can we move this `--border-width` to the host element?

core/src/components/textarea/textarea.ionic.solid.scss
```diff
@@ -0,0 +1,60 @@
...
+
+:host(.textarea-fill-solid) .textarea-bottom {
+  --border-width: #{globals.$ion-scale-0};
+}
+
+// Focus
+// ----------------------------------------------------------------
+
+:host(.textarea-fill-solid.has-focus) {
+  --border-width: #{globals.$ion-border-size-050};
```
brandyscarney: Do we need to define `--border-width` here if it is defined on the host?

core/src/components/textarea/textarea.ionic.solid.scss
```diff
@@ -0,0 +1,60 @@
...
+}
+
+// Textarea - Disabled
+// ----------------------------------------------------------------
+
+:host(.textarea-fill-solid.textarea-disabled) {
+  --color: #{globals.$ion-text-disabled};
+  --background: #{globals.$ion-bg-input-bold-disabled};
+  --border-color: #{globals.$ion-bg-input-bold-disabled};
+  --placeholder-color: #{globals.$ion-text-disabled};
```
brandyscarney: Shouldn't this be defined in `textarea.ionic.scss` since it is shared by both fills?

brandyscarney (changes requested):
Great work! A few questions on some styles & a requested change for a variable that appears different in Figma. 👍

core/src/components/textarea/textarea.ionic.solid.scss
JoaoFerreira-FrontEnd: after remove it: this start to happen: 
image

core/src/components/textarea/textarea.ionic.solid.scss
JoaoFerreira-FrontEnd: The difference between this one and the outline is the color of the border which should be the same as the background on this fill. 
image

JoaoFerreira-FrontEnd: And Figma says it should be the bold variation.

core/src/components/textarea/textarea.ionic.solid.scss
brandyscarney: Oh okay I missed that! Leave this in then.

core/src/components/textarea/textarea.ionic.solid.scss
brandyscarney: Sorry my comment was only in reference to `--placeholder-color`.

core/src/components/textarea/textarea.ionic.solid.scss
```diff
@@ -0,0 +1,60 @@
...
+// ----------------------------------------------------------------
+
+:host(.textarea-fill-solid.has-focus) {
+  --border-color: #{globals.$ion-border-focus-default};
+}
+
+:host(.textarea-fill-solid.has-focus.ion-valid) {
+  --border-width: #{globals.$ion-border-size-050};
+}
+
```
brandyscarney:
```suggestion
```

We shouldn't need this since it is defined on `.has-focus`: https://github.com/ionic-team/ionic-framework/pull/30685/files#diff-45914a3d091d2c33732c1bdfbd74ce4375d12c6b23811c211a038d829e110886R179

core/src/components/textarea/textarea.ionic.outline.scss
```diff
@@ -60,9 +60,10 @@
   margin-top: globals.$ion-space-100;
 }
 
-// Input Focus
+// Ionic Textarea - Readonly
 // ----------------------------------------------------------------
 
-:host(.textarea-fill-outline.has-focus) {
-  --border-width: #{globals.$ion-border-size-050};
+:host(.textarea-fill-outline.textarea-readonly) {
```
brandyscarney, line 74: The valid and invalid border colors are not being displayed because they are overridden by this rule.

Image

Image

Preview here: [states](https://ionic-framework-git-rou-12225-ionic1.vercel.app/src/components/textarea/test/states?ionic:theme=ionic)

core/src/components/textarea/textarea.ionic.solid.scss
JoaoFerreira-FrontEnd: here: https://github.com/ionic-team/ionic-framework/pull/30685/files#diff-45914a3d091d2c33732c1bdfbd74ce4375d12c6b23811c211a038d829e110886L154

brandyscarney: We should really move styles out that only apply to certain fills instead of overriding them. Maybe we can move that to the `outline` file instead.

core/src/components/textarea/textarea.ionic.solid.scss
```diff
@@ -0,0 +1,62 @@
...
+// Ionic Textarea - Readonly
+// ----------------------------------------------------------------
+
+:host(.textarea-fill-solid.textarea-readonly) {
+  --background: #{globals.$ion-bg-input-bold-read-only};
+  --border-color: #{globals.$ion-bg-input-bold-read-only};
+}
+
+:host(.textarea-fill-solid.textarea-readonly.ion-invalid) {
```
ShaneK, line 52: It looks like there's no visual test for this particular combination. I'd suggest creating one so we can make sure it doesn't regress in the future.

PR #30227, opened 2025-03-03
fix(vue): update output target and properly emit events

christian-bromann:
Issue number: resolves #30206 resolves #30178 resolves #30177 resolves #30175 resolves #30170

---------

## What is the current behavior?
There have been plenty of issues reported in regards to Vue components failing to propagate events. It seems like when we updated the Vue output target and started to use the provided runtime code from the output target, we have changed the way how event names are computed. Ionic has used a custom wrapper for handling events that would kebab case event names. That is no longer needed and removing it fixes observed issues.

Reproduction case working: https://stackblitz.com/edit/vj18czas-wdhzxjom?file=package.json

## What is the new behavior?
We have received a fix for this in https://github.com/stenciljs/output-targets/pull/617 which I hope will resolve this issue by updating the dependency.

## Does this introduce a breaking change?

- [ ] Yes
- [x] No

## Other information

Dev build: `8.4.4-dev.11741193800.14916f6f`

PR #30688, opened 2025-09-19
fix(vue): add component-specific overlay events

### LINKED ISSUE #30641: bug: Vue ion-modal `ionModalDidPresent` not emitted since 8.4.4 to latest (8.7.2)
### Prerequisites

### Ionic Framework Version

v8.x

### Current Behavior

Before `@ionic/vue` `8.4.4`, both long forms and short forms of events worked; now only short forms work:

So, for example, `didPresent` works(gets triggered), but `ionModalDidPresent` doesn't.

### Expected Behavior

In the expected scenario(as per docs), both short and long versions of events should be emitted, but it is probably best to remove the long-form events.

### Steps to Reproduce

Go to: https://stackblitz.com/~/github.com/andrei0x309/ionic-vue-modal-event-bug
Observe `ionModalDidPresent` is not emitted when @ionic/vue > 8.4.3

### Code Reproduction URL

https://github.com/andrei0x309/ionic-vue-modal-event-bug

### Ionic Info

Ionic:

   Ionic CLI       : 7.2.1 (C:\Users\andrei0x309\AppData\Roaming\npm\node_modules\@ionic\cli)
   Ionic Framework : @ionic/vue 8.7.2

Capacitor:

   Capacitor CLI      : 7.4.2
   @capacitor/core    : 7.4.2

Utility:

   cordova-res : 0.15.4
   native-run  : 2.0.1

System:

   NodeJS : v22.13.1 (C:\Program Files\nodejs\node.exe)
   npm    : 10.8.1
   OS     : Windows 10

### Additional Information

If deleting long forms of events is acceptable, I can make a PR to update that, it would be a breaking change but it's far better than having some events that don't trigger, because IDEs will indicate that those events don't exist, and devs can fix that very easy by using the "short form" which should become the only form. ( Docs will also be updated)

The problem started with this PR: https://github.com/ionic-team/ionic-framework/commit/11554a5d3590c660dbf609931dcb63cc2daf79cb

But I don't see anything wrong with it, might be because of stenciljs vue output pkg

ShaneK:
Issue number: resolves #30641

---------

Currently, Vue modals do not emit `ionModal...` events. This happened due to a change in the way the stencil output targets for Vue changed. Christian [updated the overlays](https://github.com/ionic-team/ionic-framework/pull/30227/files#diff-7e46aba01094c4917cd55e8eebd263fc4a297a2d62143f1ae30959ec4e023b6f) to support the base events, but not the component-specific events. 

## What is the new behavior?

With this change, you'll be able to bind to the events as described in the Ionic documentation.

- [ ] Yes
- [X] No

### DIFF
--- a/packages/vue/src/utils/overlays.ts
+++ b/packages/vue/src/utils/overlays.ts
@@ -1,5 +1,5 @@
-import type { VNode, ComponentOptions } from "vue";
-import { defineComponent, h, ref, onMounted } from "vue";
+import type { ComponentOptions, VNode } from "vue";
+import { defineComponent, h, onMounted, ref } from "vue";
 
 // TODO(FW-2969): types
 
@@ -147,23 +147,30 @@ export const defineOverlayContainer = <Props extends object>(
       const elementRef = ref();
 
       onMounted(() => {
+        // Convert name from kebab-case to camelCase
+        const componentName = name.replace(/-([a-z])/g, (_, letter) => letter.toUpperCase());
         elementRef.value.addEventListener("ionMount", (ev: Event) => {
           emit("ionMount", ev);
+          emit(componentName + "IonMount", ev);
           isOpen.value = true;
         });
         elementRef.value.addEventListener("willPresent", (ev: Event) => {
           emit("willPresent", ev);
+          emit(componentName + "WillPresent", ev);
           isOpen.value = true;
         });
         elementRef.value.addEventListener("didDismiss", (ev: Event) => {
           emit("didDismiss", ev);
+          emit(componentName + "DidDismiss", ev);
           isOpen.value = false;
         });
         elementRef.value.addEventListener("willDismiss", (ev: Event) => {
           emit("willDismiss", ev);
+          emit(componentName + "WillDismiss", ev);
         });
         elementRef.value.addEventListener("didPresent", (ev: Event) => {
           emit("didPresent", ev);
+          emit(componentName + "DidPresent", ev);
         });
       });
 


packages/vue/test/base/tests/e2e/specs/overlays.cy.js
```diff
@@ -214,6 +226,49 @@ describe('Overlays', () => {
     });
   });
 
+  it('should fire long-form lifecycle events on overlays', () => {
+    cy.get('ion-radio#ion-modal').click();
+    cy.get('ion-radio#component').click();
```
brandyscarney, line 231: Is there a reason why we don't increment & test the controller overlays?

packages/vue/test/base/src/views/Overlays.vue
```diff
@@ -335,21 +343,37 @@ export default defineComponent({
...
+    const ionModalDidDismiss = ref(0);
 
     const onModalWillPresent = () => willPresent.value += 1;
     const onModalDidPresent = () => { didPresent.value += 1; setModalRef(true); }
     const onModalWillDismiss = () => willDismiss.value += 1;
     const onModalDidDismiss = () => { didDismiss.value += 1; setModalRef(false); }
+    const onModalIonWillPresent = () => ionModalWillPresent.value += 1;
+    const onModalIonDidPresent = () => ionModalDidPresent.value += 1;
+    const onModalIonWillDismiss = () => ionModalWillDismiss.value += 1;
+    const onModalIonDidDismiss = () => ionModalDidDismiss.value += 1;
```
brandyscarney:
```suggestion
    const onIonModalWillPresent = () => ionModalWillPresent.value += 1;
    const onIonModalDidPresent = () => ionModalDidPresent.value += 1;
    const onIonModalWillDismiss = () => ionModalWillDismiss.value += 1;
    const onIonModalDidDismiss = () => ionModalDidDismiss.value += 1;
```

Nit

packages/vue/test/base/src/views/Overlays.vue
```diff
@@ -335,21 +343,37 @@ export default defineComponent({
...
 
     return {
       onModalWillPresent,
       onModalDidPresent,
       onModalWillDismiss,
       onModalDidDismiss,
+      onModalIonWillPresent,
+      onModalIonDidPresent,
+      onModalIonWillDismiss,
+      onModalIonDidDismiss,
```
brandyscarney:
```suggestion
      onIonModalWillPresent,
      onIonModalDidPresent,
      onIonModalWillDismiss,
      onIonModalDidDismiss,
```

Nit

packages/vue/test/base/src/views/Overlays.vue
```diff
@@ -98,6 +102,10 @@
         @didPresent="onModalDidPresent"
         @willDismiss="onModalWillDismiss"
         @didDismiss="onModalDidDismiss"
+        @ionModalWillPresent="onModalIonWillPresent"
+        @ionModalDidPresent="onModalIonDidPresent"
+        @ionModalWillDismiss="onModalIonWillDismiss"
+        @ionModalDidDismiss="onModalIonDidDismiss"
```
brandyscarney:
```suggestion
        @ionModalWillPresent="onIonModalWillPresent"
        @ionModalDidPresent="onIonModalDidPresent"
        @ionModalWillDismiss="onIonModalWillDismiss"
        @ionModalDidDismiss="onIonModalDidDismiss"
```

Nit

packages/vue/test/base/src/views/Overlays.vue
brandyscarney, line 1: Should we update this test & the cypress test to include a test for one or all of the other overlays that have these events: Alert, Action Sheet, Loading, Toast, or Popover?

brandyscarney (approved):
The fix looks good! I mostly have minor suggestions and questions about whether we should add more tests or not. I will defer to you on if you think we need them. 👍

packages/vue/test/base/tests/e2e/specs/overlays.cy.js
ShaneK: I was doing minimal changes necessary to test this specific use case and component overlays already had the wiring for this test, so I just took advantage of that here. I was trying to get this PR done quick enough to get it deployed in the patch tomorrow, basically.

packages/vue/test/base/src/views/Overlays.vue
ShaneK: Since they all come from the same place, I think this is a fine enough canary for now. Probably not a bad card for later though

packages/vue/test/base/src/views/Overlays.vue
ShaneK: [aba8025](https://github.com/ionic-team/ionic-framework/pull/30688/commits/aba8025d69a83d7cb6e6bc1ab8a42ff8f210ad47)

packages/vue/test/base/tests/e2e/specs/overlays.cy.js
ShaneK: [e585c66](https://github.com/ionic-team/ionic-framework/pull/30688/commits/e585c66adab0482ef5486cc55110de94f8df562e)

PR #30690, opened 2025-09-23
fix(header): show iOS condense header when app is in MD mode

### LINKED ISSUE #29929: bug: Ionic mode detection on collapsible ion-header not working consistently
### Prerequisites

### Ionic Framework Version

v8.x

### Current Behavior

In Ionic 6 I was able to use ion-header with `collapse="condense"` even when my App was in "md" mode by setting the header mode to "ios":

```html
<ion-header translucent="true" mode="ios">
  <ion-toolbar mode="md">
    <ion-title>Headline</ion-title>
  </ion-toolbar>
</ion-header>

<ion-content>
  <ion-header collapse="condense" mode="ios">
    <ion-toolbar mode="md">
      <ion-title size="large">Headline</ion-title>
    </ion-toolbar>
  </ion-header>
</ion-content>
```

This is still working with Ionic 8 when the current page is directly accessed via refresh in the browser.

But once you navigate away within the app and navigate back to the affected page, the collapsible large header is gone and the original header on top is empty.

### Expected Behavior

I am not happy that I read in some bug report that changing the mode on the fly is not working anymore.

Even when not changing, but just setting the mode via variable on element creation, it is not working:

This works:
```
<ion-header translucent="true" mode="ios">
```

These don't:
```
<ion-header translucent="true" mode="{{'ios'}}">
<ion-header translucent="true" [mode]="'ios'">
<ion-header translucent="true" [attr.mode]="'ios'">
```

It would be wonderful to have the original bevaiour back, but at least I am expecting the behaviour to be consistent.

### Steps to Reproduce

1.) Use the conference app:

```
ionic start "Conference App" https://github.com/ionic-team/ionic-conference-app
cd conference-app
ionic serve
```

2.) Edit `src/app/pages/schedule/schedule.html` and just add the modes:
```html
<ion-header translucent="true" mode="ios">
  <ion-toolbar mode="md">
...

<ion-content fullscreen="true">
  <ion-header collapse="condense" mode="ios">
    <ion-toolbar mode="md">
...
```

3.) Directly visit `http://localhost:8100/app/tabs/schedule` in your browser (the app should automatically be in "md" mode) still the iOS collapsible header is working
image

4.) Now navigate via the App sidemenu to some other page like "About", then back to "Schedule" and now the header is broken:
image

### Code Reproduction URL

See steps above in blank conference starter

### Ionic Info

Ionic:

   Ionic CLI                     : 7.2.0
   Ionic Framework               : @ionic/angular 8.2.0
   @angular-devkit/build-angular : 18.0.2
   @angular-devkit/schematics    : 18.0.2
   @angular/cli                  : 18.0.2

Capacitor:

   Capacitor CLI      : 4.6.3
   @capacitor/android : 4.6.3
   @capacitor/core    : 4.6.3
   @capacitor/ios     : 4.6.3

Utility:

   cordova-res                          : 0.15.4
   native-run (update available: 2.0.1) : 1.7.4

System:

   NodeJS : v18.20.3
   npm    : 10.7.0
   OS     : macOS Unknown

### Additional Information

While inspecting the `checkCollapsibleHeader()` function in

https://github.com/ionic-team/ionic-framework/blob/main/core/src/components/header/header.tsx#L73

I noticed that it checks the current mode by calling `getIonMode(this)`
```ts
private async checkCollapsibleHeader() {
    if (mode !== 'ios') {
      return;
    }
```

which in turn calls stencil's `getMode(ref)` 
```ts
export const getIonMode = (ref?: any): Mode => {
  return (ref && getMode(ref)) || defaultMode;
};
```

which I suspect to not reliably detected the hostRef's mode
```ts
export const getMode = (ref: d.RuntimeRef) => getHostRef(ref).$modeName$;
```
https://github.com/ionic-team/stencil/blob/5c4c969d2ee8e349891e5d41730512d9d257d893/src/runtime/mode.ts#L10

kumibrr:
Issue number: resolves #29929

---------

When forcing `mode=ios` in a collapsible header, `.header-collapse-condense` would still be applied from the `header.md.scss` file, leaving the collapsible header always hidden.

## What is the new behavior?
When forcing `mode=ios` in a collapsible header, the `.header-collapse-condense` styles from the `header.md.scss` file won't be applied, and the collapsible header will be visible.

- [ ] Yes
- [x] No

## Other information
Something worth mentioning is that this behavior only appears after initial load: if the route is loaded refreshing the page, the header will appear and work correctly, but navigating forth and back will apply both the .ios and .md style files.

I showcase this with a modal because It'll always display the broken hehavior.

| Before | After |
|--------|-------|
| <video src="https://github.com/user-attachments/assets/1307ee9f-452a-4b00-877d-0b8e360d3bf7">     | <video src="https://github.com/user-attachments/assets/f9ee3851-ce94-4a27-9947-37aa1f5433b9">  |


ShaneK:
Hey @kumibrr! Thanks for this request! I hope you don't mind, I've modified your PR to remove the HTML file as we usually only use them for things that change a lot and we'll be checking on often. I've replaced it with a regression test that uses `setContent` instead, so we'll be alerted if this becomes a problem again in the future.

Hopefully we can get this merged and deployed soon. Thanks!

PR #30692, opened 2025-09-23
feat(themes): add base tokens and update colors to use new token system

brandyscarney:
Issue number: internal

---------

The colors are generated using the Sass functions and a `$colors` map and the design tokens are undefined.

## What is the new behavior?
- Moves the helper & theme `spec` tests into the `test` directory with the other ones
- Defines base tokens for the default, light & dark tokens
- Defines the interfaces for the different theme objects
- Adds new theme utility functions to mix colors together (used for shade & tint) and convert hex to rgb
- Updates the functions that generate the design token variables to:
  - Automatically generate the rgb variables without needing to set them manually
  - Create the color classes (`.ion-color-primary`, `.ion-color-secondary`, etc.) which are used by the components to change their color via the `color` property
- Adds 3 e2e tests to verify visually that the design tokens are generating properly: Basic, Color & Typography
- Removes the old Sass mixins/functions & updates the ones we still need to pull from the design tokens instead of the `$colors` map

- [x] Yes
- [ ] No

This adds `foreground` as a required variant to the colors now. Currently foreground is only used by clear buttons but it should be used by all text-only components. This will break for any apps overriding `--ion-color-primary` without `--ion-color-primary-foreground`. We will have to look into improving the migration path on this later.

## Other information

- [Themes / Basic](https://ionic-framework-git-fw-6748-ionic1.vercel.app/src/themes/test/basic)
- [Themes / Color](https://ionic-framework-git-fw-6748-ionic1.vercel.app/src/themes/test/color)
- [Themes / Typography](https://ionic-framework-git-fw-6748-ionic1.vercel.app/src/themes/test/typography)


core/scripts/testing/styles.css
```diff
@@ -56,6 +56,7 @@ html.ios.ios {
 html.ionic,
 html.ionic.ios,
 html.ionic.md {
+  /* TODO: remove this with the ionic theme updates */
```
thetaPC, line 59: Is this being tracked through a ticket? My concern is that we'll forget about it later. Same question for the other TODOs.

core/src/utils/theme.ts
```diff
@@ -114,6 +139,96 @@ export const generateCSSVars = (theme: any, prefix: string = CSS_PROPS_PREFIX):
...
+ * @param theme The theme object containing color definitions
+ * @returns CSS string with .ion-color-{colorName} utility classes
+ */
+export const generateColorClasses = (theme: any): string => {
+  // Look for colors in the light palette first, then fallback to the
+  // direct color property if there is no light palette
+  const colors = theme?.palette?.light?.color || theme?.color;
+
+  if (!colors || typeof colors !== 'object') {
+    return '';
```
thetaPC, line 190: Should we show a warning?

core/src/utils/theme.ts
```diff
@@ -254,3 +387,73 @@ export const applyComponentTheme = (element: HTMLElement): void => {
...
+    // Short hex format like 'fff' → expand to 'ffffff'
+    r = parseInt(cleanHex[0] + cleanHex[0], 16);
+    g = parseInt(cleanHex[1] + cleanHex[1], 16);
+    b = parseInt(cleanHex[2] + cleanHex[2], 16);
+  } else {
+    // Full hex format like 'ffffff'
+    r = parseInt(cleanHex.substr(0, 2), 16);
+    g = parseInt(cleanHex.substr(2, 2), 16);
+    b = parseInt(cleanHex.substr(4, 2), 16);
+  }
```
thetaPC, line 427: Might be best to create a function for this since it's also being used in `mix()`.

core/src/utils/test/theme.spec.ts
```diff
@@ -62,3 +74,489 @@ describe('getClassMap()', () => {
...
+
+    const target = page.body.querySelector('ion-card-content')!;
+
+    const css = ':host { background-color: red; }';
+    injectCSS(css, target);
+
+    expect(target.innerHTML).toContain(`<style>${css}</style>`);
+  });
+
+  it('should inject CSS into an element with a shadow root', async () => {
```
thetaPC, line 273: Do we need to check for scoped elements too?

core/src/utils/test/theme.spec.ts
```diff
@@ -62,3 +74,489 @@ describe('getClassMap()', () => {
...
+    const expectedCSS = `
+      :root {
+        --ion-border-width-sm: 4px;
+        --ion-spacing-md: 12px;
+        --ion-dynamic-font: -apple-system-body;
+      }
+
+      @media(prefers-color-scheme: dark) {
+        :root {
+          --ion-enabled: system;
```
thetaPC, line 524: How is this variable being used?

core/scripts/testing/styles.css
brandyscarney: The Ionic theme updates is the next ticket I am working on

core/src/utils/test/theme.spec.ts
brandyscarney: It isn't but I didn't want to add custom scripts to remove it until we figured out how we would use it.

core/src/utils/theme.ts
brandyscarney: https://github.com/ionic-team/ionic-framework/pull/30692/commits/768cfa8a1625225a7935af6abf54771c61174dee

core/src/utils/theme.ts
brandyscarney: https://github.com/ionic-team/ionic-framework/pull/30692/commits/88ff27540be24c7935af47e5e269f89cb4df74bb https://github.com/ionic-team/ionic-framework/pull/30692/commits/dab8950da8e6e11029968b912d713855c9d56d94

core/src/utils/test/theme.spec.ts
brandyscarney: It isn't different than normal elements but I added it: https://github.com/ionic-team/ionic-framework/pull/30692/commits/fbe7afb1887eb23c21c963d78fa519ae0e57e42b

core/src/utils/theme.ts
```diff
@@ -114,6 +141,106 @@ export const generateCSSVars = (theme: any, prefix: string = CSS_PROPS_PREFIX):
...
+ * @param theme The theme object containing color definitions
+ * @returns CSS string with .ion-color-{colorName} utility classes
+ */
+export const generateColorClasses = (theme: any): string => {
+  // Look for colors in the light palette first, then fallback to the
+  // direct color property if there is no light palette
+  const colors = theme?.palette?.light?.color || theme?.color;
+
+  if (!colors) {
+    return '';
```
thetaPC, line 180: Should we also include a warning when no colors are found

core/src/utils/test/theme.spec.ts
```diff
@@ -62,3 +78,706 @@ describe('getClassMap()', () => {
...
+
+  it('should work with real-world color examples', () => {
+    // Mix primary Ionic blue with white
+    expect(mix('#3880ff', '#ffffff', '10%')).toBe('#4c8dff');
+
+    // Mix primary Ionic blue with black for shade
+    expect(mix('#3880ff', '#000000', '12%')).toBe('#3171e0');
+  });
+
+  it('should handle edge cases', () => {
```
thetaPC, line 781: Should there be a test for when the first value has a hash and the second value doesn't? And vice versa? Or if the first value is 3 digits and the second is 6 (and vice versa)?

core/src/utils/test/theme.spec.ts
thetaPC: Understood! Please make a ticket to circle back to this.

core/src/utils/theme.ts
brandyscarney: No because this function currently gets called whenever theme is set and people don't have to define colors.

core/src/utils/test/theme.spec.ts
brandyscarney: This will be tackled as part of the Implement Dark Mode System work outlined in my modular plan document (I can resend it if you need the link). The task uses a placeholder ticket of `FW-4005` but the tickets have not been created that far out yet. I just added a note there to take a look at this property though so it will be in the ticket description once created.

core/src/utils/test/theme.spec.ts
brandyscarney: https://github.com/ionic-team/ionic-framework/pull/30692/commits/ca30f10dc548ab70c6103a428f6f8d0648b73a11

PR #30698, opened 2025-09-24
feat(input-otp): add helperText and errorText props

thetaPC:
Issue number: internal

---------

OTP doesn't have `helperText` and `errorText` properties.

## What is the new behavior?

- Added `helperText` and `errorText` properties.
- Added snapshots.
- Added tests for accessibility.

Description slot was not changed to prevent any breaking changes. `helperText` and `errorText` will render along with description. This would have OTP always show description (when provided) regardless if `helperText` or `errorText` is present.

- [ ] Yes
- [x] No

## Other information

[Preview](https://ionic-framework-git-fw-6762-ionic1.vercel.app/src/components/input-otp/test/bottom-content/)


thetaPC:
Closing since it's no longer being requested internally.

PR #30702, opened 2025-09-30
fix(tabs): respect stencil lifecycle order for tab selection

### LINKED ISSUE #30611: bug: Initial tab not selected when using memory router
### Prerequisites

### Ionic Framework Version

v7.x, v8.x

### Current Behavior

This issue stems from #30552. That was solved for location-based routing, but unfortunately the problem still exists when using Ionic's memory router.

Example: https://stackblitz.com/edit/lcx3ywfg-vadxa3sf?file=src%2Fmain.tsx

(Voyager for Lemmy uses the memory router for native and PWA to prevent certain other bugs, so unfortunately can't switch to normal location based router.)

### Expected Behavior

Initial tab should be selected

### Steps to Reproduce

Open repro. Observe "Listen Now" is not selected.

### Code Reproduction URL

https://stackblitz.com/edit/lcx3ywfg-vadxa3sf?file=src%2Fmain.tsx

### Ionic Info

N/A

### Additional Information

I think the problem is the fix uses the browser pathname for routing:

https://github.com/ionic-team/ionic-framework/commit/a2e803a553dc58fc0e1599e515a56180a7ab263a#diff-c81e7421844580393760b9b88460d44537261b56061b496988ff6d39647018bbR79

 However, Ionic should use the internal router pathname instead to support memory router

ShaneK:
Issue number: resolves #30611

---------

Currently, the way tabs are set in the tab bar abuses a bug that existed in older versions of Stencil where children would be rendered out of the correct order. This worked in the tab and tab bar's favor previously, but after the fix it broke our implementation so tabs would no longer correctly indicate the selected tab on direct navigation.

## What is the new behavior?

We had a temporary fix before we knew what actually caused this issue before, which was basically just a timeout. That blindly worked because it triggered after the child was fully rendered. This change embraces the new, and correct, way these components render and triggers tab updates correctly.

- [ ] Yes
- [X] No

## Other information

Current dev build:
```
8.7.6-dev.11759345401.165fca78
```

### DIFF
--- a/core/src/components/tab-button/tab-button.tsx
+++ b/core/src/components/tab-button/tab-button.tsx
@@ -85,15 +85,6 @@ export class TabButton implements ComponentInterface, AnchorInterface {
    */
   @Event() ionTabButtonClick!: EventEmitter<TabButtonClickEventDetail>;
 
-  @Listen('ionTabBarLoaded', { target: 'window' })
-  onTabBarLoaded(ev: Event) {
-    // Only respond to events from our direct parent tab-bar
-    const parent = this.el.parentElement;
-    if (parent && parent.tagName === 'ION-TAB-BAR' && ev.target === parent) {
-      this.syncWithTabBar();
-    }
-  }
-
   @Listen('ionTabBarChanged', { target: 'window' })
   onTabBarChanged(ev: CustomEvent<TabBarChangedEventDetail>) {
     const dispatchedFrom = ev.target as HTMLElement;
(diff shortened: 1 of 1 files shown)


core/src/components/tab-bar/test/translucent/tab-bar.e2e.ts-snapshots/tab-bar-translucent-container-ios-ltr-Mobile-Chrome-linux.png
ShaneK, line 1: These screenshots are correct now. They were broken before, but I guess we didn't notice when we approved them. The first tab is selected per the code, so this was committed in the wrong state.

core/src/components/tab-bar/test/translucent/tab-bar.e2e.ts-snapshots/tab-bar-translucent-container-ios-ltr-Mobile-Safari-linux.png
ShaneK, line 1: These screenshots are correct now. They were broken before, but I guess we didn't notice when we approved them. The first tab is selected per the code, so this was committed in the wrong state.

brandyscarney (approved):
Looks good! Nice work!

thetaPC (approved):
LGTM!

PR #30705, opened 2025-10-03
fix(header): prevent flickering upon page transition on iOS

thetaPC:
Issue number: internal

---------

The header flickers upon page transition when on iOS mode.

**Entering Page Two (P1 → P2):**
When navigating to Page Two, which has a collapsing header (intended to be hidden until scroll), the header briefly flashes into view. This happens because the header is initially rendered with full `opacity: 1` before the component's logic can apply `opacity: 0` to hide it, causing a visible flicker.

**Navigating Back (P2 → P1):**
When navigating back, Page One's header briefly bleeds through the top of Page Two. Although Page Two is on top (`z−index: 100`), its collapsing header is set to `opacity: 0`. This transparency allows Page One header (`z−index: 99`) to become visible underneath, as the transparent area cannot block the content below it.

## What is the new behavior?

- Added a transition-specific class that is applied to the ion-header element to override its default transparency.

This guarantees the header acts as an opaque block during the page transition, eliminating visual flickering caused by early `opacity: 0` or the header underneath bleeding through.

- [ ] Yes
- [x] No

## Other information

N/A

### DIFF
--- a/core/src/components/header/header.ios.scss
+++ b/core/src/components/header/header.ios.scss
@@ -65,8 +65,6 @@
  * since it needs to blend in with the header above it.
  */
 .header-collapse-condense ion-toolbar {
-  --background: var(--ion-background-color, #fff);
-
   z-index: 0;
 }
 
@@ -91,6 +89,28 @@
 .header-collapse-main ion-toolbar.in-toolbar ion-title,
 .header-collapse-main ion-toolbar.in-toolbar ion-buttons {
   transition: all 0.2s ease-in-out;
+}
+
+/**
+ * Large title toolbar should just use the content background
+ * since it needs to blend in with the header above it.
+ */
+.header-collapse-condense ion-toolbar,
+/**
+ * Override styles applied during the page transition to prevent
+ * header flickering.
+ */
+.header-collapse-condense-inactive.header-transitioning:not(.header-collapse-condense) ion-toolbar {
+  --background: var(--ion-background-color, #fff);
+}
+
+/**
+ * Override styles applied during the page transition to prevent
+ * header flickering.
+ */
+.header-collapse-condense-inactive.header-transitioning:not(.header-collapse-condense) ion-toolbar {
+  --border-style: none;
+  --opacity-scale: 1;
 }
 
 .header-collapse-condense-inactive:not(.header-collapse-condense) ion-toolbar.in-toolbar ion-title,
--- a/core/src/utils/transition/index.ts
+++ b/core/src/utils/transition/index.ts
@@ -18,34 +18,51 @@ const focusController = createFocusController();
 
 // TODO(FW-2832): types
 
+/**
+ * Executes the main page transition.
+ * It also manages the lifecycle of header visibility (if any)
+ * to prevent visual flickering in iOS. The flickering only
+ * occurs for a condensed header that is placed above the content.
+ *
+ * @param opts Options for the transition.
+ * @returns A promise that resolves when the transition is complete.
+ */
 export const transition = (opts: TransitionOptions): Promise<TransitionResult> => {
   return new Promise((resolve, reject) => {
     writeTask(() => {
-      beforeTransition(opts);
-      runTransition(opts).then(
-        (result) => {
-          if (result.animation) {
-            result.animation.destroy();
+      const transitioningInactiveHeader = getIosIonHeader(opts);
+      beforeTransition(opts, transitioningInactiveHeader);
+      runTransition(opts)
+        .then(
+          (result) => {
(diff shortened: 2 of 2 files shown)


core/src/components/header/header.ios.scss
```diff
@@ -93,6 +91,28 @@
   transition: all 0.2s ease-in-out;
 }
 
+/**
+ * Large title toolbar should just use the content background
+ * since it needs to blend in with the header above it.
+ */
+.header-collapse-condense ion-toolbar,
```
thetaPC, line 107: This is the large title that is tied to the inactive header.

core/src/components/header/header.ios.scss
```diff
@@ -93,6 +91,28 @@
...
+/**
+ * Override styles applied during the page transition to prevent
+ * header flickering.
+ */
+.header-collapse-condense-inactive.header-transitioning:not(.header-collapse-condense) ion-toolbar {
+  --background: var(--ion-background-color, #fff);
+}
```
thetaPC, line 114: I grouped these together so they will always be the same background color since they are related to the condensed header.

core/src/utils/transition/index.ts
```diff
@@ -18,34 +18,51 @@ const focusController = createFocusController();
...
+ */
 export const transition = (opts: TransitionOptions): Promise<TransitionResult> => {
   return new Promise((resolve, reject) => {
     writeTask(() => {
-      beforeTransition(opts);
-      runTransition(opts).then(
-        (result) => {
-          if (result.animation) {
-            result.animation.destroy();
+      const transitioningInactiveHeader = getIosIonHeader(opts);
```
thetaPC, line 33: I'm grabbing the header here so we don't have to do query it twice within `beforeTransition` and in the `finally`.

core/src/utils/transition/index.ts
```diff
@@ -18,34 +18,51 @@ const focusController = createFocusController();
...
-        },
-        (error) => {
-          afterTransition(opts);
-          reject(error);
-        }
-      );
+        )
+        .finally(() => {
+          // Ensure that the header is restored to its original state.
+          setHeaderTransitionClass(transitioningInactiveHeader, false);
```
thetaPC, line 51: Removed the class here since it should be removed regardless of the success of `runTransition`.

core/src/utils/transition/index.ts
```diff
@@ -278,6 +295,36 @@ const setZIndex = (
...
+ *
+ * @param header The header element to modify.
+ * @param isTransitioning Whether the transition is occurring.
+ */
+const setHeaderTransitionClass = (header: HTMLElement | null, isTransitioning: boolean) => {
+  if (!header) {
+    return;
+  }
+
+  const transitionClass = 'header-transitioning';
```
thetaPC, line 324: I'm open to a different name. I'm not too happy with it but not sure what else to call it to be specific.

core/src/utils/transition/index.ts
```diff
@@ -291,6 +338,32 @@ export const getIonPageElement = (element: HTMLElement) => {
...
+ */
+const getIosIonHeader = (opts: TransitionOptions): HTMLElement | null => {
+  const enteringEl = opts.enteringEl;
+  const leavingEl = opts.leavingEl;
+  const direction = opts.direction;
+  const mode = opts.mode;
+
+  if (mode !== 'ios') {
+    return null;
+  }
```
thetaPC, line 360: I didn't see the benefit of querying out the header if the mode is `md`. However, we would have to update this in `next` to use `theme` instead of `mode`.

core/src/utils/transition/index.ts
thetaPC, line 51: Removed class at the end so the header can go back to its default, preventing any potential regressions.

core/src/utils/transition/index.ts
brandyscarney: I am fine with this but yeah we will need to update it eventually so that we are using the theme (or a config option for it if one is added).

core/src/utils/transition/index.ts
brandyscarney: I think this makes sense. It's added to the header during the transition.

core/src/utils/transition/index.ts
thetaPC: Is there a specific ticket I can associate it to?

brandyscarney: No I don't think we've created tickets that far. We will just have to query for where `mode` is checked.

ShaneK:
Great work, Maria! Looks awesome!

PR #30708, opened 2025-10-06
feat(themes): add ionic theme tokens

brandyscarney:
Issue number: internal

---------

- The `ionic` theme tokens are undefined.
- The `dark` palette does not work with the `ionic` theme.
- The `ionic` theme uses different colors than `md` and `ios`.
- The modular theme tokens are not being used by any tests dependent on URL params. 

## What is the new behavior?
- Updates the token interface to accept a numeric scale - it is TBD if it will be the way forward but it is needed for the `ionic` theme currently.
- Updates the base tokens to use the new interface.
- Adds tokens for the `ionic` theme based on the existing design tokens.
- Excludes certain keys (`enabled`, `rippleEffect`, `formHighlight`) from the generated CSS variables.
- Removes the `--background` and `--ion-background-color` from the global `ionic` theme overrides, allowing it to use the dark palette.
- Removes the different Ionic colors from the `ionic` theme, ensuring all of the themes use the same named (primary, secondary, etc.) colors.
- Updates the testing script to pull in the tokens based on the theme & palette passed to the URL.
- Adds a new npm script, `build.themes`, to generate the theme files.

- [ ] Yes
- [x] No

## Other information

[Preview: Basic](https://ionic-framework-git-fw-6747-ionic1.vercel.app/src/themes/test/basic)
[Preview: Color](https://ionic-framework-git-fw-6747-ionic1.vercel.app/src/themes/test/color)
[Preview: Typography](https://ionic-framework-git-fw-6747-ionic1.vercel.app/src/themes/test/typography)

Add the URL parameters `?ionic:palette=dark` or `?ionic:theme={ios|md|ionic}` to see the palette or theme change and pull in the associated tokens. 

Note: most of the theme specific tokens are not applied yet because we haven't updated the components.

### DIFF
--- a/core/scripts/testing/scripts.js
+++ b/core/scripts/testing/scripts.js
@@ -1,3 +1,4 @@
+const DEFAULT_THEME = 'md';
 
 (function() {
 
@@ -20,7 +21,7 @@
   */
   const themeQuery = window.location.search.match(/ionic:theme=([a-z0-9]+)/i);
   const themeAttr = document.documentElement.getAttribute('theme');
-  const themeName = themeQuery?.[1] || themeAttr || 'md';
+  const themeName = themeQuery?.[1] || themeAttr || DEFAULT_THEME;
 
   // TODO(): Remove this when the tokens are working for all components
   // and the themes all use the same bundle
@@ -55,7 +56,7 @@
     loadThemeTokens(themeName, paletteName);
   } else if(themeName) {
     console.warn(
-      `Unsupported theme "${themeName}". Supported themes are: ${validThemes.join(', ')}.`
+      `Unsupported theme "${themeName}". Supported themes are: ${validThemes.join(', ')}. Defaulting to ${DEFAULT_THEME}.`
     );
   }
 


BenOsodrac:
@brandyscarney I have some doubts about the approach, as I see no relation was kept to the outsystems-design-tokens package in the code. 

I was expecting the interfaces and default values would be generated based on the info provided by the outsystems-design-tokens. Even more so, if now we are using the same structure for all themes, but at least keep that connection for the Ionic one.

What happens now if the UX team decides to change the scale or the palette, for example? We would need to manual review all interfaces, values and variables to match that? 

The advantage of tokens is the auto sync between design (figma) and code, and its different contexts (for example, the outsystems-design-tokens package is also used on the outsystems side, to make sure everything is synced, regarding css). 

If we are loosing that, then there's no use to call these design tokens 😄, just type-safe css structures, like https://vanilla-extract.style/

brandyscarney:
@BenOsodrac Yeah, this is definitely an open topic.

For now, the plan is to move forward with all themes using the same tokens, and the mobile UI would override those design tokens in the framework on your side. The idea is that if the Outsystems design tokens changed values it would require a breaking change release on our end, which we're trying to avoid.

This work is still in the early stages, and we haven't finalized the naming or process yet. I'll add this to the Confluence doc under open topics so we can discuss it further.

BenOsodrac:
> @BenOsodrac Yeah, this is definitely an open topic.
> 
> For now, the plan is to move forward with all themes using the same tokens, and the mobile UI would override those design tokens in the framework on your side. The idea is that if the Outsystems design tokens changed values it would require a breaking change release on our end, which we're trying to avoid.
> 
> This work is still in the early stages, and we haven't finalized the naming or process yet. I'll add this to the Confluence doc under open topics so we can discuss it further.

Ok thanks! 👍 

But would we keep the css variables with fallbacks from the token ones? 
Because if not, that could mean a big shift on how we are preparing the customization documentation for MobileUI

brandyscarney:
@BenOsodrac I think that decision still needs to be made also. I can send you the link to the open discussion document. It is still a WIP but feel free to add topics to it or add any thoughts to existing ones! 🙂

core/src/utils/theme.ts
```diff
@@ -81,6 +81,12 @@ export const generateCSSVars = (theme: any, prefix: string = CSS_PROPS_PREFIX):
         return [];
       }
 
+      // Do not generate CSS variables for excluded keys
+      const excludedKeys = ['enabled', 'ripple-effect', 'form-highlight'];
+      if (excludedKeys.includes(key)) {
+        return [];
+      }
+
```
ShaneK, line 94: Is it intended that this new code runs before the camelCase to kebab-case switch? This means these keys may exist in camelCase and not be excluded because they haven't been switched yet.

core/src/themes/base/default.tokens.ts
```diff
@@ -39,28 +64,52 @@ export const defaultTheme: DefaultTheme = {
...
   borderWidth: {
-    none: '0',
-    xxs: '1px',
-    xs: '2px',
-    sm: '4px',
-    md: '6px',
-    lg: '8px',
-    xl: '10px',
-    xxl: '12px',
+    25: '1px',
```
ShaneK, line 91: No 0 for border width?

brandyscarney: I accidentally left that off: https://github.com/ionic-team/ionic-framework/pull/30708/commits/1954189d1ab5bad6fcc392c9032377e4d722155e

core/src/utils/theme.ts
brandyscarney: Good catch! No I moved it around while I was adding it and didn't realize I left it there. I fixed it and also updated the test so it would catch if those get added unintentionally: https://github.com/ionic-team/ionic-framework/pull/30708/commits/050b9d0a840835fd82f190ea3926690d99db233e

core/scripts/testing/scripts.js
```diff
@@ -54,6 +41,46 @@
...
+  /**
+   * The `palette` query param is used to load a specific palette
+   * for the theme. This can be `light`, `dark`, `high-contrast`,
+   * or `high-contrast-dark`. Default to `light` for tests.
+   */
+  const paletteQuery = window.location.search.match(/palette=([a-z]+)/);
+  const paletteName = paletteQuery?.[1] || 'light';
+
+  // Load theme tokens if the theme is valid
+  if (themeName && ['ionic', 'ios', 'md'].includes(themeName)) {
```
thetaPC: Should it give an error for using an invalid theme?

core/src/themes/themes.interfaces.ts
```diff
@@ -3,20 +3,56 @@ export type PlatformTheme = Omit<BaseTheme, 'ios' | 'md'>;
...
+    750?: string;
+    800?: string;
+    850?: string;
+    900?: string;
+    950?: string;
+    1000?: string;
+    1050?: string;
+    1100?: string;
+    1150?: string;
+    1200?: string;
```
thetaPC, line 48: Jeez, didn't realize that OS uses so many! 😨

core/src/themes/themes.interfaces.ts
```diff
@@ -34,29 +70,54 @@ export type BaseTheme = {
...
+    2400?: string;
+    2800?: string;
+    3200?: string;
+    3400?: string;
+    3600?: string;
+    4000?: string;
+    5000?: string;
+    6200?: string;
+    7400?: string;
+    9000?: string;
```
thetaPC, line 92: Does OS really use this many? It would be great if we could just add them as needed.

core/src/themes/md/default.tokens.ts
```diff
@@ -0,0 +1,16 @@
+import { defaultTheme as baseDefaultTheme } from '../base/default.tokens';
+import type { DefaultTheme } from '../themes.interfaces';
+
+import { darkTheme } from './dark.tokens.js';
+import { lightTheme } from './light.tokens.js';
```
thetaPC:
```suggestion
import { darkTheme } from './dark.tokens';
import { lightTheme } from './light.tokens';
```

core/src/themes/base/dark.tokens.ts
```diff
@@ -161,4 +161,62 @@ export const darkTheme: DarkTheme = {
...
+    550: '#737373',
+    600: '#666666',
+    650: '#595959',
+    700: '#4d4d4d',
+    750: '#404040',
+    800: '#333333',
+    850: '#262626',
+    900: '#1a1a1a',
+    950: '#0d0d0d',
+  },
```
thetaPC, line 212: Why is only dark the only one that has these?

core/src/components/badge/test/shape/badge.e2e.ts-snapshots/badge-shape-soft-small-size-ionic-md-ltr-light-Mobile-Chrome-linux.png
thetaPC, line 1: Weird that this doesn't have a smaller width like the others.

thetaPC (commented):
1. For dark palette on the basic page, can the text be black on spacing? It's a bit hard to read.
2. When using dark palette with ionic theme, the header doesn't change to a darker shade like the other themes. Is that on purpose? This happens on all the previews.

core/src/themes/themes.interfaces.ts
brandyscarney: Yes: https://github.com/ionic-team/ionic-framework/blob/635edaf7333ca18072f4688869d03fc425e647bf/core/src/foundations/ionic.vars.scss#L498-L529

core/src/themes/themes.interfaces.ts
brandyscarney: I have an open topic to discuss these so we can evaluate them more later.

core/src/themes/md/default.tokens.ts
brandyscarney: https://github.com/ionic-team/ionic-framework/pull/30708/commits/13c6d3100376f9cbe3d0eeafaa983aaabd3d50e9

brandyscarney:
@thetaPC:

> 1. For dark palette on the basic page, can the text be black on spacing? It's a bit hard to read.

Done: https://github.com/ionic-team/ionic-framework/pull/30708/commits/6f2c0c3d2a54ba9ecc1b019f9f512db0abcbe691

> 2. When using dark palette with ionic theme, the header doesn't change to a darker shade like the other themes. Is that on purpose? This happens on all the previews.

The `ionic` theme has never worked with the dark palette. You can see this on `next`: [colors test](https://ionic-framework-git-next-ionic1.vercel.app/src/themes/ionic/test/colors?ionic:palette=dark)

My changes enable the global dark palette for the `ionic` theme, but that doesn't work on components that are [not using the global theme variables](https://github.com/ionic-team/ionic-framework/blob/635edaf7333ca18072f4688869d03fc425e647bf/core/src/components/toolbar/toolbar.ionic.scss#L8). The components will be updated to work as they are ported to the new modular system.

core/src/themes/base/dark.tokens.ts
brandyscarney: I copied them over from the [dark palette](https://github.com/ionic-team/ionic-framework/blob/e4fc33f331f7695935e1d29c98012e5880903ed2/core/src/css/palettes/dark.scss). We don't define stepped colors for the light palette right now. I have a task to revisit all of the stepped colors though so this is something we can address then.

core/scripts/testing/scripts.js
brandyscarney: Updated it to warn when you pass an invalid theme & also updated the regex so it would include numbers because if you passed `?ionic:theme=md3` it was dropping the `3` and reading it as `md`. https://github.com/ionic-team/ionic-framework/pull/30708/commits/994340f63a040d2028efcec7de08166d3c955539

core/scripts/testing/scripts.js
```diff
@@ -54,6 +41,51 @@
...
+  const paletteQuery = window.location.search.match(/palette=([a-z]+)/);
+  const paletteName = paletteQuery?.[1] || 'light';
+
+  // Load theme tokens if the theme is valid
+  const validThemes = ['ionic', 'ios', 'md'];
+  if (themeName && validThemes.includes(themeName)) {
+    loadThemeTokens(themeName, paletteName);
+  } else if(themeName) {
+    console.warn(
+      `Unsupported theme "${themeName}". Supported themes are: ${validThemes.join(', ')}.`
```
thetaPC: It would be great to mention that md has been defaulted.

brandyscarney: https://github.com/ionic-team/ionic-framework/pull/30708/commits/5be58381e274aacc93ac5ca96d975a256091c0e0

PR #30714, opened 2025-10-07
fix(checkbox, toggle, radio-group): improve screen reader announcement timing for validation errors

thetaPC:
Issue number: internal

---------

Currently, when an error text is shown, it may not announce itself to voice assistants. This is because the way error text currently works is by always existing in the DOM, but being hidden when there is no error. When the error state changes, the error text is shown, but as far as the voice assistant can tell it's always been there and nothing has changed.

## What is the new behavior?

- Updated aria attributes
- Added observer with an observer

- [ ] Yes
- [x] No

## Other information

[Checkbox preview](https://ionic-framework-git-fw-6757-ionic1.vercel.app/src/components/checkbox/test/validation/)
[Toggle preview](https://ionic-framework-git-fw-6757-ionic1.vercel.app/src/components/toggle/test/validation/)
[Radio Group preview](https://ionic-framework-git-fw-6757-ionic1.vercel.app/src/components/radio-group/test/validation/)

⚠️ Flakiness ⚠️ 

The flakiness on checkbox and toggle is because when a native input is present on the page, the browser will have the screen reader to be really fast when it comes to checking information. This speed ends up being too fast for `ion-checkbox` to be able to add the error text. This leads to the error text not being announce consistently. There's no issue when it comes to ion-input or ion-textarea because Ionic uses the native inputs so their arias are read. There's also no issue with ion-select because we don't have a native input. It's only an issue with checkbox and the others is because it has a [native input that is being hidden](https://github.com/ionic-team/ionic-framework/blob/8e884bd2cb73481acbb6eb4a4a507d6f6a8716b1/core/src/components/checkbox/checkbox.tsx#L368-L369). So the browser sees that and speeds up the screen reader.

The flakiness on radio group is because when you tab out of the last radio button, the ionBlur event is emitted by the child <ion-radio>. This event bubbles up, but the timing is still too early for the group.

### DIFF
--- a/core/src/components/radio-group/test/validation/index.html
+++ b/core/src/components/radio-group/test/validation/index.html
@@ -106,7 +106,7 @@
       // Validation functions
       const validators = {
         'fruits-radio-group': (value) => {
-          return value !== undefined;
+          return value !== '' && value !== undefined;
         },
         'optional-radio-group': () => true, // Always valid
       };


core/src/components/checkbox/checkbox.tsx
```diff
@@ -137,10 +146,64 @@ export class Checkbox implements ComponentInterface {
...
+           * priority queue regardless if it's called
+           * inside a Promise or not, thus causing
+           * the screen reader to potentially miss the
+           * change.
+           * By using a State variable inside a Promise,
+           * it guarantees a re-render immediately at
+           * a higher priority.
+           */
+          Promise.resolve().then(() => {
+            console.log('updating hint text id');
```
ShaneK: Oops! Stray console log!

core/src/components/checkbox/checkbox.tsx
```diff
@@ -120,6 +122,13 @@ export class Checkbox implements ComponentInterface {
    */
   @Prop() required = false;
 
+  /**
+   * Track validation state for proper aria-live announcements.
+   */
+  @State() isInvalid = false;
+
+  @State() private hintTextID?: string;
```
ShaneK:
```suggestion
  @State() private hintTextId?: string;
```

I'd suggest this just to align with the casing format of other IDs in this same file. If you take this suggestion, you'll need to update the places it's referenced too.

core/src/components/radio-group/test/validation/index.html
```diff
@@ -0,0 +1,195 @@
...
+            </ion-radio-group>
+          </div>
+
+          <div>
+            <h2>Optional Field (No Validation)</h2>
+            <ion-radio-group
+              id="optional-radio-group"
+              helper-text="You can skip this field"
+              allow-empty-selection="true"
+              required
```
ShaneK: I think this might have been a copy/paste error? Looks like you intended for this one not to have validation, but it's required.

core/src/components/radio-group/test/validation/index.html
```diff
@@ -0,0 +1,195 @@
...
+
+      // Track which fields have been touched
+      const touchedFields = new Set();
+
+      // Validation functions
+      const validators = {
+        'fruits-radio-group': (value) => {
+          return value !== undefined;
+        },
+        'optional-checkbox': () => true, // Always valid
```
ShaneK: I think this is a copy-paste issue from checkbox, `optional-checkbox` doesn't exist here (in radio group)

ShaneK (changes requested):
Generally looking really good, just a few minor change requests

core/src/components/checkbox/checkbox.tsx
thetaPC: https://github.com/ionic-team/ionic-framework/pull/30714/commits/7f84e841272e31772e0d5a8c1611105577f71a39

core/src/components/checkbox/checkbox.tsx
thetaPC: https://github.com/ionic-team/ionic-framework/pull/30714/commits/30327560a043350fa66770b3d5a81ea6b0489d56 and https://github.com/ionic-team/ionic-framework/pull/30714/commits/18b9e7253aa7609221ae54ee12cb570085d9f87f

core/src/components/radio-group/test/validation/index.html
thetaPC: https://github.com/ionic-team/ionic-framework/pull/30714/commits/22767e3c5f193a2b58b8ea6c5a35282eb379bd64

core/src/components/radio-group/test/validation/index.html
thetaPC: https://github.com/ionic-team/ionic-framework/pull/30714/commits/ec28127c12b11b1b58e8dd3406175f927524420d

core/src/components/radio-group/test/validation/index.html
brandyscarney, line 1: Minor issue with this test:

Clicking reset form after triggering the invalid state and then tabbing back and forth does not trigger the invalid state again:

https://github.com/user-attachments/assets/0ec74c38-4521-43ac-b25e-f74ddce55a48

This does not happen to checkbox or toggle.

thetaPC: https://github.com/ionic-team/ionic-framework/pull/30714/commits/f43433ffd3c2611508b8762d95c2b133bd0ac57f

PR #30718, opened 2025-10-08
fix(header): ensure one banner role in condensed header

thetaPC:
Issue number: internal

---------

As per accessibility guidelines, there should only be one banner landmark per page. A condensed header creates two banner landmarks since 2 `ion-header` components are required on the page. `ion-header` renders with a `role="banner"` by default (when not in `ion-menu`).

The visual effect of the condensed header is achieved by rendering two distinct header components. Because both components exist in the code at the same time and both have `role="banner"`, they create a duplicate landmark announcement for screen readers. This leads to a violation with the accessibility guidelines.

## What is the new behavior?

- The role is updated to either `none` or `banner` based off the header's active state.
- Added test.

- [ ] Yes
- [x] No

## Other information

[Preview](https://ionic-framework-git-fw-6767-ionic1.vercel.app/src/components/header/test/condense/?ionic%3Amode=ios)

### DIFF
--- a/core/src/components/header/header.utils.ts
+++ b/core/src/components/header/header.utils.ts
@@ -262,8 +262,23 @@ export const handleHeaderFade = (scrollEl: HTMLElement, baseEl: HTMLElement, con
  * Get the role type for the ion-header.
  *
  * @param isInsideMenu If ion-header is inside ion-menu.
- * @returns 'none' if inside ion-menu, otherwise 'banner'.
+ * @param isCondensed If ion-header has collapse="condense".
+ * @param mode The current mode.
+ * @returns 'none' if inside ion-menu or if condensed in md
+ * mode, otherwise 'banner'.
  */
-export const getRoleType = (isInsideMenu: boolean) => {
-  return isInsideMenu ? ROLE_NONE : ROLE_BANNER;
+export const getRoleType = (isInsideMenu: boolean, isCondensed: boolean, mode: 'ios' | 'md') => {
+  // If the header is inside a menu, it should not have the banner role.
+  if (isInsideMenu) {
+    return ROLE_NONE;
+  }
+  /**
+   * Only apply role="none" to `md` mode condensed headers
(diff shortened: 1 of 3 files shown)


core/src/components/header/test/condense/header.e2e.ts
```diff
@@ -40,5 +40,23 @@ configs({ modes: ['ios'], directions: ['ltr'] }).forEach(({ title, screenshot, c
 
       await expect(smallTitle).toHaveAttribute('aria-hidden', 'true');
     });
+
+    test('should only have the banner role on the active header', async ({ page }) => {
```
brandyscarney, line 44: This works great for `ios` but `md` still has two `banner` roles:

Image

It shouldn't be announced since the 2nd header is `display` none but it might still fail an audit.

core/src/components/header/header.utils.ts
```diff
@@ -244,3 +257,28 @@ export const handleHeaderFade = (scrollEl: HTMLElement, baseEl: HTMLElement, con
...
+/**
+ * Get the role type for the ion-header.
+ *
+ * @param isInsideMenu If ion-header is inside ion-menu.
+ * @param isCondensed If ion-header has collapse="condense".
+ * @param mode The current mode.
+ * @returns 'none' if inside ion-menu or if condensed in md
+ * mode, otherwise 'banner'.
+ */
+export const getRoleType = (isInsideMenu: boolean, isCondensed: boolean, mode: 'ios' | 'md') => {
```
thetaPC, line 270: We'll might have to update the `mode` to `theme` in `next`.

core/src/components/header/test/condense/header.e2e.ts
thetaPC: https://github.com/ionic-team/ionic-framework/pull/30718/commits/98fcbed8f57919a22c49ede7efaaac8453be0429

PR #30722, opened 2025-10-09
refactor(button): only check for undefined fill

brandyscarney:
Issue number: internal

---------

Button checks for undefined and null fill as a result of Stencil bug https://github.com/ionic-team/stencil/issues/3586

## What is the new behavior?
- Removes check for `null` with the release of Stencil v4.38.0
- No test needed as one exists already: https://github.com/ionic-team/ionic-framework/pull/26339/files

- [ ] Yes
- [x] No


thetaPC (approved):
LGTM

PR #30635, opened 2025-08-14
fix(input): improve error text accessibility

ShaneK:
Issue number: resolves internal

---------

## What is the current behavior?

Currently, when an error text is shown, it may not announce itself to voice assistants. This is because the way error text currently works is by always existing in the DOM, but being hidden when there is no error. When the error state changes, the error text is shown, but as far as the voice assistant can tell it's always been there and nothing has changed.

## What is the new behavior?

With these changes, both input and textarea have been updated so they'll properly announce error text when it shows up. We had to do this with a mutation observer and state because it's important in some frameworks, like Angular, that state changes to cause a re-render. This, combined with some minor aria changes, makes it so that when a field is declared invalid, it immediately announces the invalid state instead of waiting for the user to go back to the invalid field.

## Does this introduce a breaking change?

- [ ] Yes
- [X] No

## Other information

Current dev build:
```
8.7.4-dev.11756220757.185b8cbf
```

## Screens
[Textarea](https://ionic-framework-git-ionic-49-ionic1.vercel.app/src/components/textarea/test/validation)
[Input](https://ionic-framework-git-ionic-49-ionic1.vercel.app/src/components/input/test/validation)

PR #30723, opened 2025-10-10
fix(select): improve screen reader announcement timing for validation errors

thetaPC:
Issue number: internal

---------

Currently, when an error text is shown, it may not announce itself to voice assistants. This is because the way error text currently works is by always existing in the DOM, but being hidden when there is no error. When the error state changes, the error text is shown, but as far as the voice assistant can tell it's always been there and nothing has changed.

## What is the new behavior?

- Updated aria attributes
- Added observer with an observer

We had to do this with a mutation observer and state because it's important in some frameworks, like Angular, that state changes to cause a re-render. This, combined with some minor aria changes, makes it so that when a field is declared invalid, it immediately announces the invalid state instead of waiting for the user to go back to the invalid field.

- [ ] Yes
- [x] No

## Other information

[Preview](https://ionic-framework-git-fw-6797-ionic1.vercel.app/src/components/select/test/validation/)

### DIFF
--- a/core/src/components/textarea/textarea.tsx
+++ b/core/src/components/textarea/textarea.tsx
@@ -15,7 +15,7 @@ import {
   writeTask,
 } from '@stencil/core';
 import type { NotchController } from '@utils/forms';
-import { createNotchController } from '@utils/forms';
+import { createNotchController, checkInvalidState } from '@utils/forms';
 import type { Attributes } from '@utils/helpers';
 import { inheritAriaAttributes, debounceEvent, inheritAttributes, componentOnReady } from '@utils/helpers';
 import { createSlotMutationController } from '@utils/slot-mutation-controller';
@@ -335,16 +335,6 @@ export class Textarea implements ComponentInterface {
     }
   }
 
-  /**
-   * Checks if the textarea is in an invalid state based on Ionic validation classes
-   */
-  private checkValidationState(): boolean {
-    const hasIonTouched = this.el.classList.contains('ion-touched');
-    const hasIonInvalid = this.el.classList.contains('ion-invalid');
-
-    return hasIonTouched && hasIonInvalid;
-  }
-
   connectedCallback() {
     const { el } = this;
     this.slotMutationController = createSlotMutationController(el, ['label', 'start', 'end'], () => forceUpdate(this));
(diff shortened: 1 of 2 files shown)


core/src/components/select/select.tsx
```diff
@@ -298,10 +306,49 @@ export class Select implements ComponentInterface {
...
+           * This is due to `forceUpdate()` not being
+           * high priority enough to guarantee
+           * the DOM is updated before the screen reader
+           * announces the attribute change.
+           * By using a promise, it makes sure to
+           * announce the change before the next frame
+           * since promises are high priority.
+           */
+          Promise.resolve().then(() => {
+            this.hintTextID = this.getHintTextID();
```
brandyscarney, line 335: Why do we use a state here instead of calling [`forceUpdate`](https://github.com/ionic-team/ionic-framework/pull/30635/files#diff-a39147111a459c7d50e285f51eddf2d3fee983bda67654c842a2c645565af6bcR433)?

core/src/components/select/select.tsx
```diff
@@ -1115,6 +1168,17 @@ export class Select implements ComponentInterface {
     return <div class="select-bottom">{this.renderHintText()}</div>;
   }
 
+  /**
+   * Checks if the input is in an invalid state based
+   * on Ionic validation classes.
+   */
+  private checkInvalidState(): boolean {
```
brandyscarney: Can we move this to a shared util file since it's used by multiple components?

core/src/components/select/select.tsx
ShaneK: The comment looks like it describes why, though it may have been added after (I appreciate it either way). Maybe @thetaPC found this more reliable than my method of using forceUpdate? My attempt was not perfect, so a more reliable method would be great

core/src/components/select/select.tsx
thetaPC: https://github.com/ionic-team/ionic-framework/pull/30723/commits/3a5d48610a65911c47b36b3c68456954abfac7d0

core/src/components/select/select.tsx
thetaPC: @brandyscarney I've updated the comment to explain it a bit better: https://github.com/ionic-team/ionic-framework/pull/30723/commits/641f250bf55d9dbfbb65863140b6324ab1fcd6d8

PR #30734, opened 2025-10-17
feat(themes): add md theme tokens and update ionic theme to include semantic tokens

brandyscarney:
Issue number: internal

---------

- The theme interface defines only numeric tokens
- The config values are set at the top-level of the theme
- No tokens are defined for the `md` theme
- The `ionic` theme redefines the base numeric tokens
- Font family is being set in multiple places

## What is the new behavior?
- Updates the theme tokens interface to add more numeric tokens & semantic tokens
- Moves the `config` values into a separate `config` option which uses the `IonicConfig` interface
  - Adds a config option for `formHighlight` to `IonicConfig`
- Defines default & dark tokens for the `md` theme
- Removes the numeric tokens from the `ionic` theme and adds semantic tokens 
- Remove the numerous `--ion-font-family` overrides in favor of the tokens

- [ ] Yes
- [x] No

### DIFF
--- a/core/src/themes/md/dark.tokens.ts
+++ b/core/src/themes/md/dark.tokens.ts
@@ -32,7 +32,7 @@ export const darkTheme: DarkTheme = {
   },
 
   textColorStep: {
-    50: '#f33f33',
+    50: '#f3f3f3',
     100: '#e7e7e7',
     150: '#dbdbdb',
     200: '#d0d0d0',


core/src/css/ionic/core.ionic.scss
```diff
@@ -8,10 +8,6 @@
   --background: #{globals.$ion-bg-surface-default};
 }
 
-html {
-  --ionic-dynamic-font: -apple-system-body;
```
brandyscarney, line 12: This is no longer needed as it is defined by the base tokens: https://github.com/ionic-team/ionic-framework/blob/3fd1d5cab14a866354d5d21fe15e00661ef0f922/core/src/themes/base/default.tokens.ts#L116

core/src/css/ionic/typography.ionic.scss
```diff
@@ -2,7 +2,7 @@
 
 // TODO(ROU-10833): add font loading solution here, as a @font-face, base64 or cdn
 html {
-  font-family: globals.$ion-font-family;
+  font-family: var(--ion-font-family);
```
brandyscarney, line 5: `fontFamily` is now defined as a token but this CSS is required as the `ionic` theme is still using this file instead of `core.scss`.

core/scripts/testing/styles.css
```diff
@@ -32,19 +32,13 @@
...
-/* TODO(FW-6745): Remove this after adding the md tokens */
 :root,
 html,
-html.md.md:not(.ionic) {
+html.md.md {
   --ion-font-family: Roboto, "mdTestingFont", sans-serif;
   font-family: Roboto, "mdTestingFont", sans-serif;
 }
 
-/* TODO(FW-6744): Remove this after adding the ios tokens */
```
brandyscarney, line 47: I thought I could remove these but Playwright still needs the `iosTestingFont` and `mdTestingFont`.

core/scripts/testing/styles.css
```diff
@@ -32,19 +32,13 @@
...
 }
 
-:root, html {
-  --ion-dynamic-font: var(--ion-default-dynamic-font);
```
brandyscarney, line 36: This variable was removed in [v8](https://github.com/ionic-team/ionic-framework/blob/next/BREAKING_ARCHIVE/v8.md#dynamic-font).

core/src/css/core.scss
```diff
@@ -7,16 +7,9 @@
...
+// TODO(FW-6744): Remove this after adding the ios tokens
 html.ios {
-  --ion-default-font: -apple-system, BlinkMacSystemFont, "Helvetica Neue", "Roboto", sans-serif;
-}
-html.md {
-  --ion-default-font: "Roboto", "Helvetica Neue", sans-serif;
-}
-
-html {
-  --ion-dynamic-font: -apple-system-body;
```
brandyscarney, line 18: This is no longer needed as it is defined by the base tokens: https://github.com/ionic-team/ionic-framework/blob/3fd1d5cab14a866354d5d21fe15e00661ef0f922/core/src/themes/base/default.tokens.ts#L116

core/src/themes/md/dark.tokens.ts
```diff
@@ -3,4 +3,68 @@ import type { DarkTheme } from '../themes.interfaces';
...
+    700: '#b8b8b8',
+    750: '#c4c4c4',
+    800: '#d0d0d0',
+    850: '#dbdbdb',
+    900: '#e7e7e7',
+    950: '#f3f3f3',
+  },
+
+  textColorStep: {
+    50: '#f33f33',
```
ShaneK: So this is weird, but I noticed all of the other colors here seem to be a grey color (`#xyxyxy`) and this one doesn't follow that pattern, so I looked at it and this one is red. Was that intentional?

Maybe you meant `#f3f3f3`? That would be grey like the others.

core/scripts/testing/styles.css
thetaPC: Why does it need it?

brandyscarney: It doesn't apply Roboto without the testing font in this file:

```css
--ion-font-family: Roboto, "mdTestingFont", sans-serif;
```

and it doesn't apply the apple font without this either:

```css
 --ion-font-family: -apple-system, BlinkMacSystemFont, "iosTestingFont", sans-serif;
```

I thought maybe this code was outdated but it is still needed so I left it. We could look into removing it later.

core/src/themes/md/dark.tokens.ts
brandyscarney: Yes, nice catch. It should match this: https://github.com/ionic-team/ionic-framework/blob/9f013b7a5138f992cf310666ff190814e053bcb9/core/src/css/palettes/dark.scss#L164

Fixed md here: https://github.com/ionic-team/ionic-framework/pull/30734/commits/fafed4688f2b953c6019997f0b5790edee6bdb9d
I also reviewed the ios one and fixed it too: https://github.com/ionic-team/ionic-framework/pull/30736/commits/fbba94032179872d3ed2248d9debb04b0a691c00

core/scripts/testing/styles.css
thetaPC: That's weird, I wonder why it needs it. Thanks!

PR #30737, opened 2025-10-17
feat(toast): add wrapper and content parts

### LINKED ISSUE #30735: feat: add content part to toast
### Prerequisites

Add the attribute part="content" to toast-content to allow styling via css.

### Describe the Use Case

Currently, toast-content cannot be styled via css. For example, the padding around it cannot be changed.

_No response_

### Describe Alternatives

_No response_

### Related Code

_No response_

### Additional Information

_No response_

Hurubon:
Issue number: resolves #30735

---------

Toast's wrapper and content divs do not have part attributes.

## What is the new behavior?
- Adds `wrapper` and `content` parts to Toast.

- [ ] Yes
- [x] No


core/src/components/toast/test/custom/toast.e2e.ts
thetaPC, line 1: @brandyscarney Why are we checking the computed styles instead of using screenshots like how the other components test for parts?

brandyscarney: I started doing this in a few places because it's less likely to produce false diffs that require manual review. It's also faster and will fail when the colors aren't applied correctly, without needing manual review.

thetaPC: Good idea! I'll keep that in mind and do that too. Thanks!

ShaneK: That makes sense here where the colors are hardcoded in the test, but you're not doing that in places where colors and styles can be changed from outside, right? Like it's way easier to update screenshots if some CSS variable changes somewhere than update this type of check

brandyscarney: I'm just doing it in tests that are making sure that customization works via CSS shadow parts or variables. 

See:
- https://github.com/ionic-team/ionic-framework/blob/207d4858a624aad4d468606ba363471039488bf1/core/src/themes/native/test/colors/theme.e2e.ts#L257
- https://github.com/ionic-team/ionic-framework/blob/391d2f67af8cba8e0d466a96e370f5049abe78e6/core/src/components/textarea/test/custom/textarea.e2e.ts#L32

We don't need a screenshot test for this because if the color suddenly stops being applied it will fail. 

> That makes sense here where the colors are hardcoded in the test, but you're not doing that in places where colors and styles can be changed from outside, right?

I don't really understand the question here. Do you mean checking the colors that are set by the components? That would probably be updated too much for this to be worth doing, correct.

PR #30743, opened 2025-10-22
chore(ci): send alert message when stencil-nightly workflow finishes

OS-jacobbell:
Send alert to a Discord channel using a webhook.

---------

## What is the new behavior?

- If a step in the workflow fails, an alert message will be sent to a Discord channel using a webhook.

- [ ] Yes
- [X] No


.github/workflows/stencil-nightly.yml
```diff
@@ -225,3 +225,19 @@ jobs:
       - name: Check build matrix status
         if: ${{ needs.test-react-e2e.result != 'success' }}
         run: exit 1
+
+  send-results-massages:
```
brandyscarney:
```suggestion
  send-results-messages:
```

I think you meant messages here 😬

PR #30750, opened 2025-10-24
fix(select): remove item focus style when there are no selected items only at ionic mode

joselrio:
Issue number: resolves #

---------

This PR introduces improvements to the visual focus styling of Ionic items when inside a select modal. 

- The CSS variables `--background-focused` and `--background-focused-opacity` were missing from `item.ionic.scss`, which resulted in the native default outline focus style being applied to focused items.
- When no item is selected, the focus style is currently applied to the first list item by default, which we intend to change.

## What is the new behavior?
- Added missing focus css variables
- Hide the default focus style when there is no item selected

## NOTE
- This change will require an additional interaction to observe the focus behavior when navigating through keyboard, since tap-based navigation does not rely on focus styling.

- [ ] Yes
- [X] No


core/src/components/accordion/test/states/accordion.e2e.ts-snapshots/accordion-states-focused-ionic-md-ltr-light-Mobile-Chrome-linux.png
thetaPC, line 1: This snapshot is no longer correct. The point of it is to capture the focus ring, that's why it's called `states-focused`. Make sure to update the test to have a checked radio. We want to make sure that the focus ring is still being applied during the right conditions.

Also add a comment to the test of when the focus ring should appear.

core/src/components/accordion/test/states/accordion.e2e.ts-snapshots/accordion-states-inset-focused-ionic-md-ltr-light-Mobile-Chrome-linux.png
thetaPC, line 1: This snapshot is no longer correct. The point of it is to capture the focus ring, that's why it's called `states-inset-focused`. Make sure to update the test to have a checked radio. We want to make sure that the focus ring is still being applied during the right conditions.

Also add a comment to the test of when the focus ring should appear.

core/src/components/item/test/states/item.e2e.ts-snapshots/item-states-diff-ionic-md-ltr-light-Mobile-Chrome-linux.png
thetaPC, line 1: This snapshot is no longer correct. The point of it is to capture the focus ring, that's why it has a few focused states. Make sure to update the test to have a checked radios. We want to make sure that the focus ring is still being applied during the right conditions.

Also add a comment to the test of when the focus ring should appear.

core/src/components/toolbar/test/basic/toolbar.e2e.ts-snapshots/toolbar-basic-nested-slotted-images-ionic-md-ltr-light-Mobile-Firefox-linux.png
thetaPC, line 1: These snapshots need to be reverted. There shouldn't be a reason why these changes were made.

core/src/components/item/test/states/item.e2e.ts-snapshots/item-states-diff-ionic-md-ltr-light-Mobile-Chrome-linux.png
joselrio: Fixed

core/src/components/accordion/test/states/accordion.e2e.ts-snapshots/accordion-states-focused-ionic-md-ltr-light-Mobile-Chrome-linux.png
joselrio: Updated :)

core/src/components/accordion/test/states/accordion.e2e.ts-snapshots/accordion-states-inset-focused-ionic-md-ltr-light-Mobile-Chrome-linux.png
joselrio: Updated :)

core/src/components/toolbar/test/basic/toolbar.e2e.ts-snapshots/toolbar-basic-nested-slotted-images-ionic-md-ltr-light-Mobile-Firefox-linux.png
joselrio: Updated :)

joselrio:
> 3\. What's the ticket to add in the work for "This change will require an additional interaction to observe the focus behavior when navigating through keyboard, since tap-based navigation does not rely on focus styling."? Because the options are no longer showing the focus rings when navigating with keyboard.

Basically we must add this behaviour just when keyboard navigation is used to navigate through the elements, before this implementation it was defined on top of tap navigation which causes this visual issue, now focus style is missing when there are no pre-selected option when keyboard navigation is in use.

core/src/components/datetime-button/test/overlays/datetime-button.e2e.ts-snapshots/datetime-overlay-modal-ios-ltr-Mobile-Safari-linux.png
thetaPC, line 1: This should be reverted. The snapshot might have only updated due to being flaky? Let's see if we can revert and not include it in this PR.

core/src/components/item/item.ionic.scss
```diff
@@ -112,3 +111,13 @@ slot[name="end"]::slotted(*) {
 :host(.item-lines-none) {
   --inner-border-width: #{globals.$ion-border-size-0};
 }
+
+// Inside a Select Modal
```
thetaPC: Let's follow the structure.

```suggestion
// Item in Select Modal
// --------------------------------------------------
```

core/src/components/toolbar/test/basic/toolbar.e2e.ts-snapshots/toolbar-basic-slotted-images-ionic-md-ltr-light-Mobile-Chrome-linux.png
thetaPC, line 1: This should be reverted. The snapshot might have only updated due to being flaky? Let's see if we can revert and not include it in this PR especially since the others are no longer included.

thetaPC (changes requested):
Just realized that we should add a test under `select-modal` to capture the focus states when an option is selected and when it's not.

thetaPC:
> > 3. What's the ticket to add in the work for "This change will require an additional interaction to observe the focus behavior when navigating through keyboard, since tap-based navigation does not rely on focus styling."? Because the options are no longer showing the focus rings when navigating with keyboard.
> 
> Basically we must add this behaviour just when keyboard navigation is used to navigate through the elements, before this implementation it was defined on top of tap navigation which causes this visual issue, now focus style is missing when there are no pre-selected option when keyboard navigation is in use.

@joselrio I understand that the behavior needs to be added for keyboard. What's the Jira ticket that implements it?

core/src/components/datetime-button/test/overlays/datetime-button.e2e.ts-snapshots/datetime-overlay-modal-ios-ltr-Mobile-Safari-linux.png
joselrio: Done ;)

core/src/components/toolbar/test/basic/toolbar.e2e.ts-snapshots/toolbar-basic-slotted-images-ionic-md-ltr-light-Mobile-Chrome-linux.png
joselrio: Done ;)

joselrio:
> Just realized that we should add a test under `select-modal` to capture the focus states when an option is selected and when it's not.

I agree, but let's do it under the context of the task I mentioned before since focus should be specially tackled under that context ;)

joselrio:
> > > 3. What's the ticket to add in the work for "This change will require an additional interaction to observe the focus behavior when navigating through keyboard, since tap-based navigation does not rely on focus styling."? Because the options are no longer showing the focus rings when navigating with keyboard.
> > 
> > 
> > Basically we must add this behaviour just when keyboard navigation is used to navigate through the elements, before this implementation it was defined on top of tap navigation which causes this visual issue, now focus style is missing when there are no pre-selected option when keyboard navigation is in use.
> 
> @joselrio I understand that the behavior needs to be added for keyboard. What's the Jira ticket that implements it?

Task do not exist yet, but it will be created since we already aligned with @jessicamendesOS and @gnbm about it.

PR #30755, opened 2025-10-27
fix(select): add missing filled type styles

joselrio:
Issue number: resolves #

---------

- Select text color was not according design
- Fill solid type was missing the background color

## What is the new behavior?
- Added the expected color to the select text
- Added missing background color variable to the fill solid type

- [ ] Yes
- [X] No


core/src/components/select/test/custom/select.e2e.ts-snapshots/select-custom-diff-ionic-md-ltr-light-Mobile-Firefox-linux.png
BenOsodrac, line 1: @joselrio This doesn't look to be right

core/src/components/select/select.ionic.scss
```diff
@@ -226,6 +228,12 @@
 // States
 // ----------------------------------------------------------------
 
+// Fill types
+// ----------------------------------------------------------------
+:host(.select-fill-solid) {
+  --background: #{globals.$ion-primitives-neutral-200};
```
BenOsodrac:
```suggestion
  --background: #{globals.$ion-bg-input-bold-default};
```

core/src/components/select/select.ionic.scss
```diff
@@ -38,6 +38,8 @@
 
 .select-text {
   min-width: globals.$ion-space-400;
+
+  color: globals.$ion-primitives-neutral-800;
```
BenOsodrac:
```suggestion
  color: globals.$ion-text-subtlest;
```

core/src/components/select/test/custom/select.e2e.ts-snapshots/select-custom-diff-ionic-md-ltr-light-Mobile-Firefox-linux.png
joselrio: This is the color assigned to the test.

core/src/components/select/select.ionic.scss
```diff
@@ -38,6 +38,8 @@
 
 .select-text {
   min-width: globals.$ion-space-400;
+
+  color: globals.$ion-text-subtlest;
```
brandyscarney, line 42: The placeholder should be this color but the text when it has a value should be `$ion-text-default` right? 

Image

It is using subtlest for both currently:

Image

joselrio: You're right! Thank you @brandyscarney

PR #30763, opened 2025-10-30
fix(select): add missing filled type styles

joselrio:
Issue number: resolves #

---------

This is based on https://github.com/ionic-team/ionic-framework/pull/30755 PR to try to clean up things.

- Select text color was not according design
- Fill solid type was missing the background color

## What is the new behavior?
- Added the expected color to the select text
- Added missing background color variable to the fill solid type

- [ ] Yes
- [X] No


core/src/components/select/test/custom/select.e2e.ts-snapshots/select-custom-parts-diff-ionic-md-ltr-light-Mobile-Chrome-linux.png
brandyscarney, line 1: This screenshot has a regression because the users are not able to customize the color of the select using `::part(container)` due to the specificity of this rule:

```css
:host(.has-value) .select-text {
```

We can fix this by adding the following [here](https://github.com/ionic-team/ionic-framework/blob/183a90ad84c1f456f4473aa49ab0ed9c1adb5e8c/core/src/components/select/test/custom/select.e2e.ts#L60-L62): 

```diff
+ion-select::part(text),
ion-select::part(container) {
  color: purple;
}
```


Or by setting the `color` on the `native-wrapper` instead of `select-text`: https://github.com/ionic-team/ionic-framework/blob/183a90ad84c1f456f4473aa49ab0ed9c1adb5e8c/core/src/components/select/select.ionic.scss#L165

joselrio: Done!

PR #30769, opened 2025-11-05
fix(select, action-sheet): use radio role for options

thetaPC:
Issue number: internal

---------

The screen reader does not announce when an option is selected within the action sheet interface. This is because the action sheet uses standard buttons, which do not support a detectable selected state via native properties or ARIA attributes like `aria-checked` or `aria-selected`, creating an inconsistent user experience across different interface types.

## What is the new behavior?

- Updated the action sheet buttons to accept `role="radio"`
- Added keyboard navigation to follow the pattern for radio group
- Added test

- [ ] Yes
- [x] No

## Other information

[Basic](https://ionic-framework-git-fw-6818-ionic1.vercel.app/src/components/select/test/basic/)

### DIFF
--- a/core/src/components/select/select.tsx
+++ b/core/src/components/select/select.tsx
@@ -559,7 +559,7 @@ export class Select implements ComponentInterface {
       const isSelected = isOptionSelected(selectValue, value, this.compareWith);
 
       return {
-        role: isOptionSelected(selectValue, value, this.compareWith) ? 'selected' : '',
+        role: isSelected ? 'selected' : '',
         text: option.textContent,
         cssClass: optClass,
         handler: () => {


core/src/components/select/test/a11y/select.e2e.ts
```diff
@@ -111,3 +111,39 @@ configs({ directions: ['ltr'] }).forEach(({ title, config, screenshot }) => {
...
+
+      const select = page.locator('ion-select');
+
+      await select.click();
+      await ionActionSheetDidPresent.next();
+
+      const selectedOption = page.locator('.action-sheet-selected');
+      await expect(selectedOption).toHaveAttribute('aria-description', 'selected');
+
+      // Check that the attribut is not added to non-selected option
```
ShaneK:
```suggestion
      // Check that the attribute is not added to non-selected option
```

core/src/components/select/test/a11y/select.e2e.ts
```diff
@@ -111,3 +111,39 @@ configs({ directions: ['ltr'] }).forEach(({ title, config, screenshot }) => {
     });
   });
 });
+
+/**
+ * This behavior does not vary across modes/directions.
+ */
+configs({ modes: ['ios'], directions: ['ltr'] }).forEach(({ title, config }) => {
+  test.describe(title('select: aria attributes'), () => {
+    test('should have a aria-description on the selected option when action sheet interface is open', async ({
```
ShaneK:
```suggestion
    test('should have an aria-description on the selected option when action sheet interface is open', async ({
```

core/src/components/action-sheet/action-sheet.tsx
```diff
@@ -355,8 +500,83 @@ export class ActionSheet implements ComponentInterface, OverlayInterface {
     this.triggerChanged();
   }
 
+  private renderActionSheetButtons(filteredButtons: ActionSheetButton[]) {
+    const mode = getIonMode(this);
+    const { activeRadioId } = this;
+    console.log('Rendering buttons with activeRadioId:', activeRadioId);
```
ShaneK:
```suggestion
```
Console log 👀

core/src/components/action-sheet/action-sheet.tsx
```diff
@@ -81,6 +88,19 @@ export class ActionSheet implements ComponentInterface, OverlayInterface {
    * An array of buttons for the action sheet.
    */
   @Prop() buttons: (ActionSheetButton | string)[] = [];
+  @Watch('buttons')
+  buttonsChanged() {
+    // Initialize activeRadioId when buttons change
+    if (this.hasRadioButtons) {
```
ShaneK, line 97: My only concern here is `this.hasRadioButtons` is only set in `connectedCallback`, I believe this means if buttons are set dynamically this value may become stale and this process may not trigger when it's supposed to. That's sort of an edge case for many, but it may be worth re-computing `this.hasRadioButtons` here, if you feel like this could be a concern. Up to you!

ShaneK (changes requested):
Looking good, just a couple of minor things

core/src/components/action-sheet/action-sheet.tsx
thetaPC: https://github.com/ionic-team/ionic-framework/pull/30769/commits/626370ba8047ad5ac3a40e62227e546993899081

core/src/components/action-sheet/action-sheet.tsx
thetaPC: https://github.com/ionic-team/ionic-framework/pull/30769/commits/c150a614958c38fd4e041d9850eecae419ec0935

core/src/components/select/select.tsx
```diff
@@ -556,14 +556,18 @@ export class Select implements ComponentInterface {
         .filter((cls) => cls !== 'hydrated')
         .join(' ');
       const optClass = `${OPTION_CLASS} ${copyClasses}`;
+      const isSelected = isOptionSelected(selectValue, value, this.compareWith);
 
       return {
-        role: isOptionSelected(selectValue, value, this.compareWith) ? 'selected' : '',
+        role: 'radio',
```
brandyscarney: Why did you decide to pass `role` here instead of providing it as an `htmlAttribute`?

I'm worried this might introduce a breaking change.

For example, if anyone is relying on the role being returned as `selected` when clicking it and handling it through the following event:

```js
window.addEventListener('ionActionSheetDidDismiss', function (e) {
  console.log('didDismiss', e.detail);
});
```

this would no longer work as expected.

core/src/components/action-sheet/action-sheet.tsx
```diff
@@ -277,12 +300,50 @@ export class ActionSheet implements ComponentInterface, OverlayInterface {
...
+  private getRadioButtons(): ActionSheetButton[] {
+    return this.getButtons().filter((b) => b.role === 'radio' && !isCancel(b.role));
+  }
+
+  /**
+   * Handle radio button selection and update aria-checked state.
+   *
+   * @param button The radio button that was selected.
+   */
+  private async selectRadioButton(button: ActionSheetButton) {
```
brandyscarney:
```suggestion
  private selectRadioButton(button: ActionSheetButton) {
```

core/src/components/action-sheet/action-sheet.tsx
```diff
@@ -295,6 +356,89 @@ export class ActionSheet implements ComponentInterface, OverlayInterface {
...
   };
 
+  /**
+   * When the action sheet has radio buttons, we want to follow the
+   * keyboard navigation pattern for radio groups:
+   * - Arrow Down/Right: Move to the next radio button (wrap to first if at end)
+   * - Arrow Up/Left: Move to the previous radio button (wrap to last if at start)
+   * - Space/Enter: Select the focused radio button and trigger its handler
+   */
+  @Listen('keydown')
```
brandyscarney, line 369: We don't have to do it on this PR, but I think a lot of this code could be moved to a shared util that is also used by radio group:

https://github.com/ionic-team/ionic-framework/blob/bf0f1e36e4dd73c968ade75c714b89b3f10dd4ef/core/src/components/radio-group/radio-group.tsx#L232-L294

core/src/components/action-sheet/action-sheet.tsx
```diff
@@ -295,6 +356,89 @@ export class ActionSheet implements ComponentInterface, OverlayInterface {
...
+        const buttonId = this.getButtonId(b, allButtons.indexOf(b));
+        return buttonId === nextEl?.id;
+      });
+
+      if (buttonIndex !== -1) {
+        this.selectRadioButton(radioButtons[buttonIndex]);
+        nextEl.focus();
+      }
+    }
+  }
```
brandyscarney, line 450:
```suggestion
  onKeydown(ev: KeyboardEvent) {
    // Only handle keyboard navigation if we have radio buttons
    if (!this.hasRadioButtons || !this.presented) {
      return;
    }

    const target = ev.target as HTMLElement;

    // Ignore if the target element is not within the action sheet or not a radio button
    if (
      !this.el.contains(target) ||
      !target.classList.contains('action-sheet-button') ||
      target.getAttribute('role') !== 'radio'
    ) {
      return;
    }

    // Get all radio button elements and filter out disabled ones
    const radios = Array.from(this.el.querySelectorAll('.action-sheet-button[role="radio"]')).filter(
      (el) => !(el as HTMLButtonElement).disabled
    ) as HTMLButtonElement[];

    const currentIndex = radios.findIndex((radio) => radio.id === target.id);
    if (currentIndex === -1) {
      return;
    }

    const allButtons = this.getButtons();
    const radioButtons = this.getRadioButtons();

    // Build a map of button element IDs to their ActionSheetButton config objects.
    // This allows us to quickly look up which button config corresponds to a DOM element
    // when handling keyboard navigation (e.g., when user presses Space/Enter or arrow keys).
    // The key is the ID that was set on the DOM element during render, and the value is
    // the ActionSheetButton config that contains the handler and other properties.
    const buttonIdMap = new Map<string, ActionSheetButton>();
    radioButtons.forEach((b) => {
      const allIndex = allButtons.indexOf(b);
      const buttonId = this.getButtonId(b, allIndex);
      buttonIdMap.set(buttonId, b);
    });

    let nextEl: HTMLButtonElement | undefined;

    if (['ArrowDown', 'ArrowRight'].includes(ev.key)) {
      ev.preventDefault();
      ev.stopPropagation();

      nextEl = currentIndex === radios.length - 1 ? radios[0] : radios[currentIndex + 1];
    } else if (['ArrowUp', 'ArrowLeft'].includes(ev.key)) {
      ev.preventDefault();
      ev.stopPropagation();

      nextEl = currentIndex === 0 ? radios[radios.length - 1] : radios[currentIndex - 1];
    } else if (ev.key === ' ' || ev.key === 'Enter') {
      ev.preventDefault();
      ev.stopPropagation();

      const button = buttonIdMap.get(target.id);
      if (button) {
        this.selectRadioButton(button);
        this.buttonClick(button);
      }
      return;
    }

    // Focus the next radio button
    if (nextEl) {
      const button = buttonIdMap.get(nextEl.id);
      if (button) {
        this.selectRadioButton(button);
        nextEl.focus();
      }
    }
  }
```

Simplify to avoid querying the buttons multiple times. You may want to make this change yourself instead of accepting mine as I probably messed up spacing when I pasted.

core/src/components/select/select.tsx
thetaPC: While it's technically possible, relying on the `role` from the action sheet closing event is a bit of a weird way to do things. The best practice is to just use the `<ion-select>`'s own `ionChange` event. That's why I decided to fix the role, especially because the dismissal event would *only* pass back `'selected'` if the user happened to click on an option that was already chosen.

Example:
1. No selection has occurred
2. User clicks on "apple"
3. `e.detail.role=""`
4. User clicks on "apple" again
5. `e.detail.role="selected"`

core/src/components/action-sheet/action-sheet.tsx
thetaPC: https://github.com/ionic-team/ionic-framework/pull/30769/commits/73149c94430e1588368a02e61a5a6169f0853ec3

core/src/components/action-sheet/action-sheet.tsx
thetaPC: https://github.com/ionic-team/ionic-framework/pull/30769/commits/8314a540f870f4647752f76366cd5ad08eab95f7

core/src/components/action-sheet/action-sheet.tsx
thetaPC: Agreed! I was trying to do that with action sheet, alert, and radio group but it was starting to get a bit more complex so I internally shelved it. Ticket made: FW-6936

core/src/components/select/select.tsx
brandyscarney: I agree it's unlikely someone is checking for this, but I don't really see any benefit to adding a new role here, and changing it introduces a potential breaking change. This is what I get when I log the dismiss event for each interface:

console log of select interfaces on dismiss with role radio

I'm actually thinking we should remove the role entirely in a future major release so it lines up with the other interfaces. Select seems to work fine without it and I don't see any clear benefit for users to listen for it anyway:

console log of select interfaces on dismiss with role undefined

thetaPC: I made the switch back to have radio role under the `htmlAttributes` as you suggested: https://github.com/ionic-team/ionic-framework/pull/30769/commits/97c672e261185e806ab0e519b13db758fd242520. The button role still [renders](https://github.com/ionic-team/ionic-framework/pull/30769/files#diff-7e9d82a562c47c0d2142c4283ce3300b809765b4ed9b9a4f020b6167f821d1f5R562) as `radio` since it's needed for the accessibility.

core/src/components/select/select.tsx
```diff
@@ -556,6 +556,7 @@ export class Select implements ComponentInterface {
         .filter((cls) => cls !== 'hydrated')
         .join(' ');
       const optClass = `${OPTION_CLASS} ${copyClasses}`;
+      const isSelected = isOptionSelected(selectValue, value, this.compareWith);
 
       return {
         role: isOptionSelected(selectValue, value, this.compareWith) ? 'selected' : '',
```
```suggestion
        role: isSelected ? 'selected' : '',
```

thetaPC: https://github.com/ionic-team/ionic-framework/pull/30769/commits/d3e0d534de170aa53b98792bcf0c2eed77e3d9cb

PR #30771, opened 2025-11-06
feat(select-modal): add select default modal styling for ionic theme

joselrio:
Issue number: resolves #

---------

Once IonSelect was using a modal-default (with no breakpoints) the styling of the modal do not fit with the expected ones for ionic theme.
Screenshot 2025-11-06 at 11 01 00

## What is the new behavior?
Under the same context new styling has been added.
Screenshot 2025-11-06 at 11 01 22

- [ ] Yes
- [X] No

### DIFF
--- a/core/src/components/button/button.ionic.scss
+++ b/core/src/components/button/button.ionic.scss
@@ -185,21 +185,21 @@
 // --------------------------------------------------
 
 :host(.button-soft) {
-  --border-radius: #{globals.$ion-border-radius-200};
+  --border-radius: #{globals.$ion-soft-xs};
 }
 
 // Round Button
 // --------------------------------------------------
 
 :host(.button-round) {
-  --border-radius: #{globals.$ion-border-radius-full};
+  --border-radius: #{globals.$ion-round-xs};
 }
 
 // Rectangular Button
 // --------------------------------------------------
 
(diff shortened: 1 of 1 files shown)


core/src/components/modal/modal.ionic.scss
```diff
@@ -27,6 +27,16 @@
...
 
+// Default Modal
+// --------------------------------------------------
+
+// Modal when inside a select default-modal
+:host(.select-modal.modal-default) {
+  --max-height: 45%;
+  --max-width: calc(100% - (#{globals.$ion-space-400} * 2));
+  --min-height: 340px;
+}
```
OS-giulianasilva, line 38: Shouldn't we be using tokens here? I didn't found them in Figma, but was it checked with the UX Team?

joselrio: > Shouldn't we be using tokens here? I didn't found them in Figma, but was it checked with the UX Team?

@OS-giulianasilva well, I did not use them because they are not there :P
`--max-width` is making usage of it. Will ask about others.

joselrio: After discussed this should be kept for now ;) 
CC @jessicamendesOS, @gnbm

PR #30773, opened 2025-11-06
fix(ionic): update light base color

JoaoFerreira-FrontEnd:
Issue number: internal

---------

## What is the new behavior?

- guarantee that light base is #ffffff;

- [ ] Yes
- [x] No


core/src/components/button/test/outline/button.e2e.ts-snapshots/button-outline-ionic-ios-ltr-light-Mobile-Chrome-linux.png
thetaPC, line 1: Should the text for the light button also be changed?

JoaoFerreira-FrontEnd: I don't think so. There is no need. This will be reviewed in the near future.

thetaPC (approved):
LGTM

PR #30775, opened 2025-11-07
fix(tab-bar): ui tweaks for ionic theme

rugoncalves:
Issue number: internal

---------

The current UI of the tab-bar and tab-button in the ionic theme is not matching the expected, in edge use cases:
- tab-button:
   - Long label: causes the size of the tab-button to increase and eventually making other tab-buttons, not visible.
   - size: in different scenarios the size of each tab-button could become different.
- tab-bar:
   - Many tab-buttons: would case some of them to become inaccessible.
   - Floating: distance from the bottom, was incorrect.

## What is the new behavior?

The changes made allow to fix the issues described above, specifically for ionic theme, containing this way, potential impacts:
- tab-button:
   - min-width: has been defined for the tab-button.
   - property was set to force each tab-button to have the same size.
- tab-bar:
   - allow the ability for the content to overflow on the x axis - a scroll will appear if needed.
   - increased the distance from the bottom, to comply with latest design trends.

- [ ] Yes
- [x] No

### DIFF
--- a/core/src/components/button/button.ionic.scss
+++ b/core/src/components/button/button.ionic.scss
@@ -185,21 +185,21 @@
 // --------------------------------------------------
 
 :host(.button-soft) {
-  --border-radius: #{globals.$ion-border-radius-200};
+  --border-radius: #{globals.$ion-soft-xs};
 }
 
 // Round Button
 // --------------------------------------------------
 
 :host(.button-round) {
-  --border-radius: #{globals.$ion-border-radius-full};
+  --border-radius: #{globals.$ion-round-xs};
 }
 
 // Rectangular Button
 // --------------------------------------------------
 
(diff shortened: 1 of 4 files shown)


core/src/components/tab-button/tab-button.ionic.scss
```diff
@@ -12,8 +12,12 @@
 
   position: relative;
 
+  flex: 1;
+
```
brandyscarney:
```suggestion
```

This isn't necessary since it is defined in [tab-button.common.scss](https://github.com/ionic-team/ionic-framework/blob/cac12b31fbf2d6da49250c12e8933c87e10fd435/core/src/components/tab-button/tab-button.common.scss#L21)

PR #30778, opened 2025-11-09
chore(npm): Update release npm action to stop using tokens

gnbm:
Issue number: internal

---------

- Release workflows still inject a long-lived `NPM_TOKEN` via `.npmrc`, so publishes do not use npm’s trusted OIDC flow.

## What is the new behavior?

- The shared `actions/publish-npm` composite action now configures `setup-node` with the npm registry, upgrades npm in place, and publishes with `--provenance` without writing `.npmrc`.
- `release-dev.yml`, `release-nightly.yml`, and `release-production.yml` call into that trusted flow by removing the token input and (for production) inlining the same OIDC setup before `npm run release.ci`.
    - Allows npm to authenticate through trusted publishing requirements [docs.npmjs.com/trusted-publishers](https://docs.npmjs.com/trusted-publishers). 
- Step names were refreshed with emojis, but there are no other behavioral changes.

- [ ] Yes
- [x] No

## Other information

- These changes align the Ionic release automation with npm’s trusted-publisher enforcement while keeping the existing Lerna build/publish process intact.

### DIFF
--- a/.github/workflows/actions/publish-npm/action.yml
+++ b/.github/workflows/actions/publish-npm/action.yml
@@ -17,7 +17,7 @@ inputs:
   node-version:
     description: 'Node.js version to use when publishing.'
     required: false
-    default: '20'
+    default: '24.x'
 runs:
   using: 'composite'
   steps:


ShaneK (approved):
This looks fine to me

.github/workflows/actions/build-angular-server/action.yml
```diff
@@ -3,23 +3,23 @@ description: 'Build Ionic Angular Server'
 runs:
   using: 'composite'
   steps:
-    - uses: actions/setup-node@v6
+    - uses: actions/setup-node@2028fbc5c25fe9cf00d9f06a71cc4710d4507903 # v6.0.0
```
thetaPC, line 6: Why are we pinning the version?

.github/workflows/actions/build-core-stencil-prerelease/action.yml
```diff
@@ -8,20 +8,20 @@ inputs:
 runs:
   using: 'composite'
   steps:
-    - uses: actions/checkout@v5
-    - uses: actions/setup-node@v6
+    - uses: actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8 # v5.0.0
+    - uses: actions/setup-node@2028fbc5c25fe9cf00d9f06a71cc4710d4507903 # v6.0.0
```
thetaPC, line 12: Why are we pinning versions? The same goes for the rest of the files.

.github/workflows/actions/publish-npm/action.yml
```diff
@@ -8,48 +8,55 @@ inputs:
...
     description: 'The directory of the package.'
   folder:
     default: './'
     description: 'A folder containing a package.json file.'
-  token:
-    description: 'The NPM authentication token required to publish.'
+  node-version:
+    description: 'Node.js version to use when publishing.'
+    required: false
+    default: '24.x'
```
thetaPC: Why can't we use the latest node? That way we don't have to manually update the default when the default node is deprecated. By doing this way, I'm pretty sure that [renovate](github.com/ionic-team/ionic-framework/issues/30726) can't upgrade the node and we'll end up forgetting to do this.

.github/workflows/actions/publish-npm/action.yml
```diff
@@ -8,48 +8,55 @@ inputs:
...
+  registry-url:
+    description: 'Registry URL used for npm publish.'
+    required: false
+    default: 'https://registry.npmjs.org'
```
thetaPC: What scenario would we need to change the registry URL?

.github/workflows/actions/publish-npm/action.yml
```diff
@@ -8,48 +8,55 @@ inputs:
...
       shell: bash
       working-directory: ${{ inputs.working-directory }}
-    - name: Update Version
-      run: npx lerna@5 version ${{ inputs.version }} --yes --exact --no-changelog --no-push --no-git-tag-version --preid=${{ inputs.preid }}
+    - name: 🏷️ Set Version
+      run: |
+        if [ -z "${{ inputs.preid }}" ]; then
+          npx lerna@5 version ${{ inputs.version }} --yes --exact --no-changelog --no-push --no-git-tag-version
+        else
+          npx lerna@5 version ${{ inputs.version }} --yes --exact --no-changelog --no-push --no-git-tag-version --preid=${{ inputs.preid }}
```
thetaPC, line 47: Does this command break if no `preid` is passed?

.github/workflows/release-ionic.yml
```diff
@@ -162,13 +158,18 @@ jobs:
         preid: ${{ inputs.preid }}
         working-directory: 'packages/angular-server'
         folder: './dist'
-        token: ${{ secrets.NPM_TOKEN }}
+    - name: Cache Built @ionic/angular-server
+      uses: ./.github/workflows/actions/upload-archive
+      with:
+        name: ionic-angular-server
+        output: packages/angular-server/AngularServerBuild.zip
+        paths: packages/angular-server/dist
```
thetaPC: What is this doing?

.github/workflows/release.yml
```diff
@@ -23,32 +23,25 @@ on:
           - latest
           - next
       preid:
-        type: choice
+        type: string
```
thetaPC: Can we revert this back to options? How do we prevent a typo from being entered? By using options, we lessen any user errors.

.github/workflows/actions/build-angular-server/action.yml
gnbm: This is a security best practice enforced by our company's security team, so it makes sense to me to follow that same rule on this team's repo

.github/workflows/actions/build-core-stencil-prerelease/action.yml
gnbm: This is a security best practice enforced by our company's security team, so it makes sense to me to follow that same rule on this team's repo

.github/workflows/actions/publish-npm/action.yml
gnbm: @thetaPC Nothing was changed in terms of node version, as you can see in `main` https://github.com/ionic-team/ionic-framework/blob/0a02e0f8cffe7b307a9f0c7da13ff6c048335e80/.github/workflows/actions/publish-npm/action.yml#L24 
Any additional changes to the action, I would like to keep separate since it's out of scope

.github/workflows/release-ionic.yml
gnbm: Removed the unused cache step for @ionic/angular-server; no downstream job ever consumed that archive, so keeping it would add work without benefit. 
Well spotted, since this was a leftover from a test I made, and didn't end up removing it

.github/workflows/actions/publish-npm/action.yml
gnbm: The lerna version guard already handles empty preid: when the workflow passes an empty string, the first branch runs, so no --preid flag is added; any non-empty selection (alpha/beta/etc.) still works through the else branch.
Am I missing something?

.github/workflows/release.yml
gnbm: I was trying to be more generic here, but since we'll not reuse the action, I accepted your suggestion and rolled back the type to `choice`

.github/workflows/actions/publish-npm/action.yml
thetaPC: If no errors are thrown then does this new if statement mainly used to run a cleaner command?

.github/workflows/actions/publish-npm/action.yml
thetaPC: @gnbm didn't see a response for this.

gnbm: It was deleted..you can see the Outdated badge

.github/workflows/actions/publish-npm/action.yml
gnbm: Correct!

PR #30793, opened 2025-11-18
fix(datetime): ensure datetime is shown when intersection observer fails to report visibility

### LINKED ISSUE #30706: bug: IonDateTime sometimes show invisible calander
### Prerequisites

### Ionic Framework Version

v8.x

### Current Behavior

I have a IonDateTime component, which I show inside a modal, as per the referenced example, where I have reproduced the issue, of the calendar being invisible, and the buttons not working correctly

I have never had any issues with this until upgrading and converting all to standalone,

Now, sometimes it will open with an invisible calendar. Here is what I see in the referenced example (and what I get in my production app)

Image

Debugging into Ionic code, and using copilot suggestions, inside of ion-datetime.js, the following is not called a second time

```
componentDidLoad() {
        const { el, intersectionTrackerRef } = this;
        /**
         * If a scrollable element is hidden using `display: none`,
         * it will not have a scroll height meaning we cannot scroll elements
         * into view. As a result, we will need to wait for the datetime to become
         * visible if used inside of a modal or a popover otherwise the scrollable
         * areas will not have the correct values snapped into place.
         */
        const visibleCallback = (entries) => {
            const ev = entries[0];
            if (!ev.isIntersecting) {
                // visibleCallback is always called once and exits here
                return;
            }
            
           // When no issues, this is called again, where we can then call initializeListeners
          // and also add the `datetime-ready` class to make the calendar visible

            this.initializeListeners();
            /**
             * TODO FW-2793: Datetime needs a frame to ensure that it
             * can properly scroll contents into view. As a result
             * we hide the scrollable content until after that frame
             * so users do not see the content quickly shifting. The downside
             * is that the content will pop into view a frame after. Maybe there
             * is a better way to handle this?
             */
            writeTask(() => {
                this.el.classList.add('datetime-ready');
            });
```

I have tried various workaround, such as adding

```
ion-datetime .calendar-body,
ion-datetime .datetime-year {
  opacity: 1 !important;
}
```

but even if we get the calendar visible, the buttons do not work correctly, assumingly as we have not had `initializeListeners` called.

It seems to be quite random sometimes it works, other times not .

Noe, this is from copilot, so may not be correct, but it has said...

``
GitHub Copilot

Why datetime-ready never appears:

Causes (any one is enough):

IntersectionObserver never reports visibility: The child <div class="intersection-tracker"> has zero size (no layout yet) so intersection ratio stays 0 (threshold = 0.01) → callback not fired.
Datetime created inside a container that is display:none (e.g. *ngIf false, hidden ion-modal before present()) at componentDidLoad. Observer attaches while hidden; when later shown, some WebViews fail to re-fire for a zero‑area target.
IntersectionObserver not available / polyfilled incorrectly in the WebView → no callback → no class.
Your custom CSS (or parent flex constraints) collapses ion-datetime height to 0 until the class would be added (deadlock).
Modal/popover quickly hides then shows (hidden observer fires first removing class; visible never re-fires).
Quick confirmations: Open DevTools console and run:

If width/height = 0 → root cause #1.
```

For some reason I lost the test suggested in the history, but it did come back as 0;

So the issue here is that I don't always have the visibleCallback called with ev.isIntersecting set to true, so this causes
- Buttons not working correct
- Calendar to be invisible

I have included a reference to a simple example where the issue can occur

### Expected Behavior

For the calendar to show, and buttons to work correct. When I have this issue, the months change when I click the empty calendar area.

### Steps to Reproduce

1. Run the referenced example, e.g. `npm run serve`
2. Click the SHOW DATE TOME PICKER button.
3. Click in the actual sate time text to open the date time component which contains the ion-datetime

<img width="326" height="201" alt="Image" src="https://github.com/user-attachments/assets/11e461d0-1e6c-4fe3-b6b1-0e069aa08dea" />

4. Sometimes as here, there will be no calendar

<img width="361" height="448" alt="Image" src="https://github.com/user-attachments/assets/a5c89c97-4cbb-47e9-b2e6-78f27328d070" />


### Code Reproduction URL

https://github.com/pjc2007/date-time-issue

### Ionic Info

$ ionic info
       
       Require stack:
       - C:\Users\Peter\AppData\Roaming\npm\node_modules\@ionic\cli\lib\project\index.js
       - C:\Users\Peter\AppData\Roaming\npm\node_modules\@ionic\cli\lib\index.js
       - C:\Users\Peter\AppData\Roaming\npm\node_modules\@ionic\cli\index.js
       - C:\Users\Peter\AppData\Roaming\npm\node_modules\@ionic\cli\bin\ionic
       
       Require stack:
       - C:\Users\Peter\AppData\Roaming\npm\node_modules\@ionic\cli\lib\project\index.js
       - C:\Users\Peter\AppData\Roaming\npm\node_modules\@ionic\cli\lib\index.js
       - C:\Users\Peter\AppData\Roaming\npm\node_modules\@ionic\cli\index.js
       - C:\Users\Peter\AppData\Roaming\npm\node_modules\@ionic\cli\bin\ionic

Ionic:

   Ionic CLI                     : 7.2.1 (C:\Users\Peter\AppData\Roaming\npm\node_modules\@ionic\cli)
   Ionic Framework               : @ionic/angular 8.7.5
   @angular-devkit/build-angular : 20.3.4
   @angular-devkit/schematics    : 20.3.4
   @angular/cli                  : 20.3.4

Capacitor:

   Capacitor CLI      : 7.4.3
   @capacitor/core    : 7.4.3

Utility:

   native-run  : 2.0.1

System:

   NodeJS : v22.17.1 (C:\Program Files\nodejs\node.exe)
   npm    : 10.9.2
   OS     : Windows 10

### Additional Information

I am actually on Windows 11 (not sure why the above report Windows 10)

In the above sample app, I reproduced running in Chrome, but it also occurred on other Mobile devices. when the issue occurs, it is due to the following not being executed.

<img width="717" height="612" alt="Image" src="https://github.com/user-attachments/assets/5ee83313-c9d7-4dae-8bb1-1d16d0ade265" />

Note the height of intersection-tracker is 0 (not sure is relevant, was just mentioned by copilot)

<img width="1291" height="713" alt="Image" src="https://github.com/user-attachments/assets/464e927b-06f1-43f3-ab0b-ae780d1a6310" />

The ion-datetime component has the following attributes.
```
<ion-datetime _ngcontent-ng-c2850345462="" presentation="time-date" hourcycle="h23" class="ion-color ion-color-date-time-light-mode md datetime-presentation-time-date datetime-size-fixed datetime-grid hydrated"><span _ngcontent-ng-c2850345462="" slot="time-label">Time</span><ion-buttons _ngcontent-ng-c2850345462="" slot="buttons" class="sc-ion-buttons-md-h md sc-ion-buttons-md-s hydrated"><ion-button _ngcontent-ng-c2850345462="" class="ion-color ion-color-primary md button button-clear in-buttons ion-activatable ion-focusable hydrated" color="primary">Cancel</ion-button><ion-button _ngcontent-ng-c2850345462="" class="ion-color ion-color-primary md button button-clear in-buttons ion-activatable ion-focusable hydrated" color="primary">Ok</ion-button></ion-buttons><input type="hidden" class="aux-input" name="ion-dt-1" value="2025-10-05T04:19:33.782Z"></ion-datetime>
```

i.e. no `datetime-ready` class.

### PR DESCRIPTION
**Pull request by ShaneK**
Issue number: resolves #30706

---------

Due to some recent unknown changes, the intersection observer for date time no longer reliably fires, especially in mobile views.

## What is the new behavior?

In this PR, we're adding a visibility check after everything has had a chance to render to make sure we're setting up properly even if the intersection observer has failed to trigger for some reason.

- [ ] Yes
- [X] No

## Other information

Since the intersection observer is being set up after a `raf`, it's possible something got introduced to make the initial setup slower for some reason, causing timing issues. I think we should do a more thorough investigation into the cause of this problem when we have more time.

This PR also adds tests to verify the new fallback works properly.

Current dev build:
```
8.7.10-dev.11763478209.1d9c4cd8
```

### DIFF
--- a/core/src/components/datetime/datetime.tsx
+++ b/core/src/components/datetime/datetime.tsx
@@ -1168,8 +1168,8 @@ export class Datetime implements ComponentInterface {
     raf(() => visibleIO?.observe(intersectionTrackerRef!));
 
     /**
-   * TODO(FW-6931): Remove this fallback upon solving the root cause
-   * Fallback: If IntersectionObserver never reports that the
+     * TODO(FW-6931): Remove this fallback upon solving the root cause
+     * Fallback: If IntersectionObserver never reports that the
      * datetime is visible but the host clearly has layout, ensure
      * we still initialize listeners and mark the component as ready.
      *

### CODE REVIEW

File: core/src/components/datetime/test/basic/datetime.e2e.ts [hunk 1]
```diff
@@ -394,6 +394,60 @@ configs({ modes: ['ios'], directions: ['ltr'] }).forEach(({ title, config }) =>
...
+      await page.waitForTimeout(100);
+
+      await expect(datetime).toHaveClass(/datetime-ready/);
+
+      const calendarBody = datetime.locator('.calendar-body');
+      await expect(calendarBody).toBeVisible();
+
+      const firstEnabledDay = datetime.locator('.calendar-day:not([disabled])').first();
+      await firstEnabledDay.click();
+      await page.waitForChanges();
```
thetaPC: Are these needed? Seems like we're not using them to test.

core/src/components/datetime/test/basic/datetime.e2e.ts
```diff
@@ -394,6 +394,60 @@ configs({ modes: ['ios'], directions: ['ltr'] }).forEach(({ title, config }) =>
...
 
+/**
+ * Synthetic IntersectionObserver fallback behavior.
+ *
+ * This test stubs IntersectionObserver so that the callback
+ * never reports an intersecting entry. The datetime should
+ * still become ready via its internal fallback logic.
+ */
+configs({ modes: ['md'], directions: ['ltr'] }).forEach(({ title, config }) => {
+  test.describe(title('datetime: IO fallback'), () => {
```
thetaPC, line 405: The GH issue is missing. Here's an [example](https://github.com/ionic-team/ionic-framework/blob/2bebbd7a3e50c030f80e0b031d7f40d3ca13e92b/core/src/components/alert/test/a11y/alert.e2e.ts#L87) of what I mean.

core/src/components/datetime-button/test/overlays/datetime-button.e2e.ts
```diff
@@ -176,5 +176,26 @@ configs({ modes: ['md'], directions: ['ltr'] }).forEach(({ title, config }) => {
       await ionModalDidPresent.next();
       await expect(datetime).toBeVisible();
     });
+    test('should set datetime ready state and keep calendar interactive when reopening modal', async ({ page }) => {
```
thetaPC: The GH issue is missing. Here's an [example](https://github.com/ionic-team/ionic-framework/blob/2bebbd7a3e50c030f80e0b031d7f40d3ca13e92b/core/src/components/alert/test/a11y/alert.e2e.ts#L87) of what I mean.

core/src/components/datetime-button/test/overlays/datetime-button.e2e.ts
```diff
@@ -176,5 +176,26 @@ configs({ modes: ['md'], directions: ['ltr'] }).forEach(({ title, config }) => {
...
+
+        const calendarBody = datetime.locator('.calendar-body');
+        await expect(calendarBody).toBeVisible();
+
+        const firstEnabledDay = datetime.locator('.calendar-day:not([disabled])').first();
+        await firstEnabledDay.click();
+        await page.waitForChanges();
+
+        await modal.evaluate((el: HTMLIonModalElement) => el.dismiss());
+        await ionModalDidDismiss.next();
```
thetaPC: This seems to be needed for the first `openAndInteract()`. Could we just move it after line 197?

ljanner:
@ShaneK The dev build does correctly resolve the problem for me 👍 Thanks providing a fix 🙇

core/src/components/datetime/test/basic/datetime.e2e.ts
ShaneK: Nope, copypasta 💀  - [c07b0e9](https://github.com/ionic-team/ionic-framework/pull/30793/commits/c07b0e959ebcf5ff17d42030b1591596fd287fb1)

core/src/components/datetime-button/test/overlays/datetime-button.e2e.ts
ShaneK: IDK, the code felt cleaner the other way, though I guess now playwright isn't doing something it doesn't need to the second time around - [c07b0e9](https://github.com/ionic-team/ionic-framework/pull/30793/commits/c07b0e959ebcf5ff17d42030b1591596fd287fb1)

core/src/components/datetime/test/basic/datetime.e2e.ts
ShaneK: Good catch, I always forget about these haha - [a6f0eb3](https://github.com/ionic-team/ionic-framework/pull/30793/commits/a6f0eb31a463f90d29fbc2835bf73ea3706711f9)

core/src/components/datetime-button/test/overlays/datetime-button.e2e.ts
ShaneK: [a6f0eb3](https://github.com/ionic-team/ionic-framework/pull/30793/commits/a6f0eb31a463f90d29fbc2835bf73ea3706711f9)

core/src/components/datetime-button/test/overlays/datetime-button.e2e.ts
ShaneK: Oops, [0e59e06](https://github.com/ionic-team/ionic-framework/pull/30793/commits/0e59e06f6189df803851448f52ce92ca489c01e1)

thetaPC: I appreciate it!

core/src/components/datetime/datetime.tsx
```diff
@@ -1141,6 +1166,17 @@ export class Datetime implements ComponentInterface {
      */
     raf(() => visibleIO?.observe(intersectionTrackerRef!));
 
+    /**
+     * Fallback: If IntersectionObserver never reports that the
```
brandyscarney:
```suggestion
   * TODO(FW-6931): Remove this fallback upon solving the root cause
   * Fallback: If IntersectionObserver never reports that the
```

Let's link to the follow-up JIRA ticket.

core/src/components/datetime/datetime.tsx
```diff
@@ -1101,6 +1101,31 @@ export class Datetime implements ComponentInterface {
     this.initializeKeyboardListeners();
   }
 
+  /**
+   * Fallback to ensure the datetime becomes ready even if
```
brandyscarney, line 1106:
```suggestion
   * TODO(FW-6931): Remove this fallback upon solving the root cause
   * Fallback to ensure the datetime becomes ready even if
```

Let's link to the follow-up JIRA ticket.

core/src/components/datetime/datetime.tsx
brandyscarney: Sorry didn't realize I messed up the spacing here 😂

PR #30803, opened 2025-11-21
fix(angular): use `ComponentRef.setInput()` API by default

### LINKED ISSUE #30784: bug: [angular] `ModalController` is not compatible with Angular signal inputs
### Prerequisites

### Ionic Framework Version

v8.x

### Current Behavior

`ModalController` is not compatible with Angular signal inputs

### Expected Behavior

`ModalController` is compatible with Angular signal inputs

### Steps to Reproduce

1. Create Angular component with signal inputs
2. Try to set its inputs using `ModalController.create()`.

### Code Reproduction URL

-

### Ionic Info

@ionic/angular: 8.7.9

### Additional Information

_No response_

hakimio:
Issue number: resolves #30784

---------

You have to manually enable `useSetInputAPI` setting to use Angular signal inputs.

## What is the new behavior?
The setting was removed since the oldest supported Angular version (`v16`) supports `ComponentRef.setInput()` API which was introduced in Angular `v14.1`.

- [x] Yes
- [ ] No

Existing applications which enable `useSetInputAPI` setting will have to delete the setting.

## Other information

The `useSetInputAPI` setting was introduced when Ionic supported Angular versions older than `v14.1`. That's not the case anymore and there is no need to keep the setting.

### DIFF
--- a/packages/angular/common/src/providers/angular-delegate.ts
+++ b/packages/angular/common/src/providers/angular-delegate.ts
@@ -20,8 +20,6 @@ import {
 
 import { NavParams } from '../directives/navigation/nav-params';
 
-import { ConfigToken } from './config';
-
 // Token for injecting the modal element
 export const IonModalToken = new InjectionToken<HTMLIonModalElement>('IonModalToken');
 
@@ -31,7 +29,6 @@ export const IonModalToken = new InjectionToken<HTMLIonModalElement>('IonModalTo
 export class AngularDelegate {
   private zone = inject(NgZone);
   private applicationRef = inject(ApplicationRef);
-  private config = inject(ConfigToken);
 
   create(
     environmentInjector: EnvironmentInjector,
@@ -43,8 +40,7 @@ export class AngularDelegate {
       injector,
       this.applicationRef,
       this.zone,
-      elementReferenceKey,
-      this.config.useSetInputAPI ?? false
+      elementReferenceKey
     );
   }
 }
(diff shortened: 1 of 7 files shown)


ShaneK:
Thanks for this PR! I've changed it from `feat` to `fix` since this should have been done when we bumped the [minimum Angular version](https://ionicframework.com/docs/reference/support#ionic-angular) from 14 to 16 some time ago.

ShaneK:
It looks like this PR is having some unexpected impacts on the angular e2e tests. I'm not sure what's up with those, seems like for some reason some modals aren't working like they're expected to. We'll need to get those fixed before we can proceed.

hakimio:
@ShaneK please re-run the e2e tests.

ShaneK:
For future reference, this PR has been moved to point to v9 as it has breaking changes. These changes will break any existing Angular app that passes in props to a modal that it doesn't have `@Inputs` bound to. These changes will also break typescript building for any app that has this property set (since it removes it from the options).

ShaneK:
@hakimio I'm not sure why you closed this? I think it's a valuable addition, we just need to rebase the major branch and it can be merged into it. It's just some breaking changes, but that's what majors are for.

hakimio:
It should be a new PR. When you changed the target branch it messed up commit history.
Do you have any idea when `v9` might be released?

ShaneK:
No, it doesn't need to be a new PR. The commit history would clean up when we merge main into the v9 branch

hakimio:
So, when are you planning to release `v9`?

PR #30816, opened 2025-11-26
chore(renovate): configure node version consistency

thetaPC:
Issue number: N/A

---------

Core does not indicate what node version we should be running, which can lead to inconsistent versions between local and workflows. These inconsistencies can lead to workflow failures.

## What is the new behavior?

- Provides the node version within package
- Updates the node version of workflows and package at the same time

- [ ] Yes
- [x] No

### DIFF
--- a/core/package.json
+++ b/core/package.json
@@ -2,6 +2,9 @@
   "name": "@ionic/core",
   "version": "8.7.11",
   "description": "Base components for Ionic",
+  "engines": {
+    "node": "24.x"
+  },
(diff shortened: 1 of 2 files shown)


emmaus-zam:
This breaks installing @ionic/core with node versions < 24. Is this intended? 

Tested with yarn:

```
error @ionic/core@8.7.12: The engine "node" is incompatible with this module. Expected version "24.x". Got "20.19.0"
```

PR #30817, opened 2025-11-27
test(set-content): use the proper theme keyterm

thetaPC:
Issue number: N/A

---------

The docs do not match the new structure for the `theme`. It still calls `mode` as `theme`.

## What is the new behavior?

- Switched `mode` to `theme`

- [ ] Yes
- [x] No

## Other information

N/A


core/src/utils/test/playwright/page/utils/set-content.ts
```diff
@@ -19,18 +19,18 @@ export const setContent = async (page: Page, html: string, testInfo: TestInfo, o
 
   let mode: Mode;
   let direction: Direction;
-  let theme: Theme;
+  let themeName: Theme;
```
brandyscarney: Why was this renamed?

thetaPC: I undid it. I realized that this change would be best in `ionic-modular`.

https://github.com/ionic-team/ionic-framework/pull/30817/commits/c998e2d2b0c9a92650b8a2574c25de607b02937a

PR #30826, opened 2025-12-01
chore(git): sync ionic-modular with next

thetaPC:
Syncing with next


core/package.json
```diff
@@ -59,19 +59,19 @@
     "chalk": "^5.3.0",
     "clean-css-cli": "^5.6.1",
     "domino": "^2.1.6",
+    "esbuild": "0.27.0",
```
thetaPC: Installed this package else it would give an error with `build.themes`.

PR #30835, opened 2025-12-03
feat(radio-group): remove radio-group-wrapper and convert to shadow DOM

OS-pedrolourenco:
Issue number: internal

---------

The radio-group's wrapper div is responsible for several issues that have been reported, specially due to making it so that radio elements are not direct children of radio-groups. Now it is also causing problems on the OutSystems side.

## What is the new behavior?

The solution involved:
- Removing the radio-group-wrapper div, which was first added as a workaround to a Stencil issue where radio children would incorrectly fire blur events on focus;
- Converting the radio-group into a shadow component as another workaround to the Stencil issue;
- Adjusting the radio-group's styling to the new changes. 

- [x] Yes
- [ ] No

### DIFF
--- a/BREAKING.md
+++ b/BREAKING.md
@@ -38,6 +38,12 @@ This is a comprehensive list of the breaking changes introduced in the major ver
 
 - The properties `pull` and `push` have been deprecated and no longer work. A similar look can be achieved with the newly added property `order`.
 
+<h4 id="version-9x-radio-group">Radio Group</h4>
+
+- Converted `ion-radio-group` to use [Shadow DOM](https://developer.mozilla.org/en-US/docs/Web/Web_Components/Using_shadow_DOM).<br/>
+If you were targeting the internals of `ion-radio-group` in your CSS, you will need to target the `supporting-text`, `helper-text` or `error-text` [Shadow Parts](https://ionicframework.com/docs/theming/css-shadow-parts) instead, or use the provided CSS Variables.<br/>
+Additionally, the `radio-group-wrapper` div element has been removed, causing slotted elements to be direct children of the `ion-radio-group`.
+
 <h5>Example 1: Swap two columns</h5>
 
 **Version up to 8.x**


core/src/components/radio-group/radio-group.ionic.scss
```diff
@@ -23,16 +23,20 @@
...
-  @include globals.rtl() {
-    padding-right: calc(globals.$ion-space-400 + var(--ion-safe-area-right, 0px));
-    padding-left: globals.$ion-space-400;
-  }
-  /* stylelint-enable */
+// Note: :host-context() cannot be nested inside ltr()/rtl() mixins
+// because the mixin tries to transform the selector, creating invalid CSS.
+// We work around this by using separate rules outside the mixins.
+/* stylelint-disable */
+:host-context(ion-list) .radio-group-top {
```
brandyscarney: `:host-context` doesn't work with Shadow DOM in Safari or Firefox. You'll have to do something like this:

```scss
:host(.in-list) .radio-group-top {
```

Then in `radio-group.tsx`:

```tsx
return (
  <Host
    class={{
      [theme]: true,
      'in-list': hostContext('ion-list', el),
    }}
```

See this example: https://github.com/ionic-team/ionic-framework/blob/3709bba41e3f8cb479bed851c57a4cbf00cb7c8f/core/src/components/checkbox/checkbox.tsx#L352

core/src/components/radio-group/radio-group.tsx
```diff
@@ -317,11 +318,11 @@ export class RadioGroup implements ComponentInterface {
     }
 
     return (
-      <div class="radio-group-top">
-        <div id={helperTextId} class="helper-text" aria-live="polite">
+      <div class="radio-group-top" part="top">
+        <div id={helperTextId} class="helper-text" part="helper-text" aria-live="polite">
```
brandyscarney:
```suggestion
        <div id={helperTextId} class="helper-text" part="supporting-text helper-text" aria-live="polite">
```

core/src/components/radio-group/radio-group.tsx
```diff
@@ -317,11 +318,11 @@ export class RadioGroup implements ComponentInterface {
...
           {!isInvalid ? helperText : null}
         </div>
-        <div id={errorTextId} class="error-text" role="alert">
+        <div id={errorTextId} class="error-text" part="error-text" role="alert">
```
brandyscarney:
```suggestion
        <div id={errorTextId} class="error-text" part="supporting-text error-text" role="alert">
```

core/src/components/radio-group/radio-group.tsx
brandyscarney, line 19: We need to document the parts like this:

```tsx
 *
 * @part supporting-text - Supporting text displayed above the radios.
 * @part helper-text - Supporting text displayed above the radios when the radio group is valid.
 * @part error-text - Supporting text displayed above the radios when the radio group is invalid and touched.
 */
```

core/src/components/radio-group/radio-group.tsx
```diff
@@ -317,11 +318,11 @@ export class RadioGroup implements ComponentInterface {
     }
 
     return (
-      <div class="radio-group-top">
-        <div id={helperTextId} class="helper-text" aria-live="polite">
+      <div class="radio-group-top" part="top">
```
brandyscarney:
```suggestion
      <div class="radio-group-top">
```

Let's not add this part right now since we don't expose this wrapper for checkbox or select.

brandyscarney (changes requested):
Looking good, just a few requests!

core/src/components/radio-group/radio-group.ionic.scss
OS-pedrolourenco: Right! Forgot about that, sorry. I'm on it.

core/src/components/radio-group/radio-group.common.scss
```diff
@@ -1,9 +1,8 @@
 // Radio Group: Common
 // --------------------------------------------------
 
-ion-radio-group {
-  // Prevents additional pixels from being rendered on top
-  vertical-align: top;
```
OS-pedrolourenco, line 6: This became unnecessary by introducing `display: block`

core/src/components/radio-group/radio-group.ionic.scss
```diff
@@ -6,8 +6,6 @@
 
 .radio-group-top {
   @include globals.typography(globals.$ion-body-sm-medium);
-
-  margin-bottom: globals.$ion-space-400;
```
OS-pedrolourenco, line 10: This margin was being overridden on the OS side, thus this style makes no sense to exist here. Will remove the override as well.

PR #30842, opened 2025-12-08
test(scripts): update palette query

thetaPC:
Issue number: N/A

---------

`high-contrast` and `high-contrast-dark` palettes were not working when requested through a URL query or hash for a test page. This was due to the `match` not accepting hyphens so it would only save `high` which is not a valid palette.

## What is the new behavior?

- Updated `match` to accept hyphens
- Added an error if an invalid palette is provided
- Added a palette fallback if an invalid palette is provided
- Added a class check for high contrast and high contrast dark

- [ ] Yes
- [x] No

## Other information

How to test:
1. Verify that `?palette=high-contrast`, `?palette=high-contrast-dark`, `#palette=high-contrast`, and `#palette=high-contrast-dark` render correctly (I recommend using [button basic page](https://ionic-framework-git-scripts-ionic1.vercel.app/src/components/button/test/basic/))

### DIFF
--- a/core/scripts/testing/scripts.js
+++ b/core/scripts/testing/scripts.js
@@ -67,8 +67,8 @@
   let paletteName = paletteQuery?.[1] || paletteHash?.[1] || highContrastDarkClass || darkClass || highContrastClass || 'light';
 
   if (!validPalettes.includes(paletteName)) {
+    console.warn(`Invalid palette name: '${paletteName}'. Falling back to 'light' palette.`);
     paletteName = 'light';
-    console.warn(`Invalid palette name: ${paletteName}. Falling back to 'light' palette.`);
   }
 
   if (paletteName !== 'light') {


core/scripts/testing/scripts.js
```diff
@@ -57,11 +57,19 @@
...
   const darkClass = document.body?.classList.contains('ion-palette-dark') ? 'dark' : null;
+  const highContrastClass = document.body?.classList.contains('ion-palette-high-contrast') ? 'high-contrast' : null;
+  const highContrastDarkClass = darkClass && highContrastClass ? 'high-contrast-dark' : null;
 
-  const paletteName = paletteQuery?.[1] || paletteHash?.[1] || darkClass || 'light';
+  let paletteName = paletteQuery?.[1] || paletteHash?.[1] || highContrastDarkClass || darkClass || highContrastClass || 'light';
+
+  if (!validPalettes.includes(paletteName)) {
+    paletteName = 'light';
+    console.warn(`Invalid palette name: ${paletteName}. Falling back to 'light' palette.`);
```
brandyscarney:
```suggestion
    console.warn(`Invalid palette name: '${paletteName}'. Falling back to 'light' palette.`);
    paletteName = 'light';
```

This currently prints:

```
Invalid palette name: light. Falling back to 'light' palette.
```

because you are setting `paletteName` to `light` in the line right before this. Switching the order and adding quotes reads as this:

```
Invalid palette name: 'hifdfds'. Falling back to 'light' palette.
```

Issue #24177, opened 2021-11-06
feat: React Router v6 support


ptmkenny:
### Prerequisites

- [X] I have read the [Contributing Guidelines](https://github.com/ionic-team/ionic-framework/blob/main/.github/CONTRIBUTING.md#creating-an-issue).
- [X] I agree to follow the [Code of Conduct](https://ionicframework.com/code-of-conduct).
- [X] I have searched for [existing issues](https://github.com/ionic-team/ionic-framework/issues) that already include this feature request, without success.

### Describe the Feature Request

Ionic React currently uses React Router 5, which is [tagged as classic on yarn](https://classic.yarnpkg.com/en/package/react-router-dom).

[React Router](https://reactrouter.com/) v6 received a stable release on [November 3, 2021](https://remix.run/blog/react-router-v6).  Three years passed and Ionic released Ionic React versions 6, 7, and 8, but none of these supported React Router 6.

It would be great to be able to use a version of React Router that is actively being updated and makes use of the new features in React 18 and 19.

### Describe the Use Case

React Router 6 adds features with React 18 and 19 that v5 does not.

### Describe Preferred Solution

`@ionic/react-router` should update to React Router 6.

### Issues Fixed in React Router 6 but not 5

React Router 5 is only getting critical security fixes since development has effectively ended.

* https://github.com/remix-run/react-router/issues/10814

fortesp:
I also have this question. 
Now that we have Ionic version 6, i was trying to upgrade to react router version 6 as well, but it seems not yet possible.

I get the following error:

```
Creating an optimized production build...
Failed to compile.

./node_modules/@ionic/react-router/dist/index.esm.js
Attempted import error: 'withRouter' is not exported from 'react-router-dom'
```

ptmkenny:
@fortesp This is a feature request, not a bug report.  Upgrading to react router v6 is not supported in Ionic now, and it will not be until the Ionic team adds code to support it; there's no "workaround" unless you want to write the code and contribute a PR yourself.

fortesp:
@ptmkenny not reporting any bug, just enforcing the idea. Not sure i understand the rest of your comment. The idea is to understand when it will be supported.

zhangriyueming:
Like to hear a timeline for this.

jrobeson:
[...]

PR #30844, opened 2025-12-08
fix(react-router): support React Router 6 style relative paths in IonRouterOutlet

ShaneK:
Issue number: resolves [an issue from a comment](https://github.com/ionic-team/ionic-framework/issues/24177#issuecomment-3624311206)

---------

Currently, Ionic's RR6 doesn't support relative routes in the same way RR6 does. Routes that do not start with a `/`do not work in the Ionic RR6 implementation in some cases.

## What is the new behavior?

With this change, we properly support these route styles and more closely align with normal RR6 route support.

- [ ] Yes
- [X] No

## Other information

Current dev build:
```
 8.7.12-dev.11765307927.1f491e92
```

This PR will be merged into the RR6 branch, which will be squash+merged into the major 9 branch. This will prevent major 9 from having commits in the change log on release that reference fixing things that are only available in major 9.

### DIFF
--- a/packages/react-router/src/ReactRouter/StackManager.tsx
+++ b/packages/react-router/src/ReactRouter/StackManager.tsx
@@ -207,9 +207,40 @@ export class StackManager extends React.PureComponent<StackManagerProps> {
     }
 
     if (routeInfo.routeAction === 'replace') {
-      return true;
+      const enteringRoutePath = enteringViewItem?.reactElement?.props?.path as string | undefined;
+      const leavingRoutePath = leavingViewItem?.reactElement?.props?.path as string | undefined;
+
+      // Never unmount the root path "/" - it's the main entry point for back navigation
+      if (leavingRoutePath === '/' || leavingRoutePath === '') {
+        return false;
+      }
+
+      if (enteringRoutePath && leavingRoutePath) {
+        // Get parent paths to check if routes share a common parent
+        const getParentPath = (path: string) => {
+          const normalized = path.replace(/\/\*$/, ''); // Remove trailing /*
+          const lastSlash = normalized.lastIndexOf('/');
+          return lastSlash > 0 ? normalized.substring(0, lastSlash) : '/';
+        };
+
+        const enteringParent = getParentPath(enteringRoutePath);
+        const leavingParent = getParentPath(leavingRoutePath);
+
+        // Unmount if:
+        // 1. Routes are siblings (same parent, e.g., /page1 and /page2, or /foo/page1 and /foo/page2)
+        // 2. Entering is a child of leaving (redirect, e.g., /tabs -> /tabs/tab1)
+        const areSiblings = enteringParent === leavingParent && enteringParent !== '/';
+        const isChildRedirect =
+          enteringRoutePath.startsWith(leavingRoutePath) ||
+          (leavingRoutePath.endsWith('/*') && enteringRoutePath.startsWith(leavingRoutePath.slice(0, -2)));
+
+        return areSiblings || isChildRedirect;
+      }
+
+      return false;
     }
 
+    // For non-replace actions, only unmount for back navigation (not forward push)
     const isForwardPush = routeInfo.routeAction === 'push' && (routeInfo as any).routeDirection === 'forward';
     if (!isForwardPush && routeInfo.routeDirection !== 'none' && enteringViewItem !== leavingViewItem) {
       return true;
@@ -317,9 +348,6 @@ export class StackManager extends React.PureComponent<StackManagerProps> {
     leavingViewItem: ViewItem | undefined,
     shouldUnmountLeavingViewItem: boolean
   ): void {
-    // Ensure the entering view is not hidden from previous navigations
-    showIonPageElement(enteringViewItem.ionPageElement);
-
     // Handle same view item case (e.g., parameterized route changes)
     if (enteringViewItem === leavingViewItem) {
       const routePath = enteringViewItem.reactElement?.props?.path as string | undefined;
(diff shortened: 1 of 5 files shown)


ptmkenny:
I think there is a minor documentation issue here: `Routes that start with a / do not work in the Ionic RR6 implementation.` should be `Routes that **do not** start with a / do not work in the Ionic RR6 implementation.`

ptmkenny:
I just installed the test branch for this:

```
 npm install @ionic/core@8.7.12-dev.11765219790.17cbe2e9 @ionic/react@8.7.12-dev.11765219790.17cbe2e9 @ionic/react-router@8.7.12-dev.11765219790.17cbe2e9
```

Unfortunately, this immediately breaks my app. `npm run dev` shows this error on the index page:

```
A <Route> is only ever to be used as the child of <Routes> element, never rendered directly. Please wrap your <Route> in a <Routes>.
```

I have all my `Route` components wrapped in `IonRouterOutlet`, not `Routes` (to match the structure of the code used in the tests). This was working with the other build, but in this build, I get the above error.

Changing all `IonRouterOutlet` components to `Routes` components works-- I can then navigate around my app normally.

ShaneK:
@ptmkenny Can you provide a reproduction of this? I can't see how this change would cause that and I can't see how the e2e tests would be passing (much less building) if this were happening. [For example, this newly added one in this PR, which uses `IonRouterOutlet` + `Route`](https://github.com/ionic-team/ionic-framework/pull/30844/files#diff-e6d2a996a02d93eae4481eae9ef51be56fcd3a2219fadbbe3812428d37ac5706)

ptmkenny:
@ShaneK Sorry about that, it seems there was some weird caching issue that caused that error.  I rebooted the machine, blew away all the caches, and I could build my app again with the new build.

However, it seems that not having a slash in a root-level route will still break the app. See the repro here: https://github.com/ptmkenny/ionic-react-router-6-test/tree/router-outlet

I put the reproduction instructions in the README.md. Also, note that in addition to adding the slash (for IonRouterOutlet), you can also replace IonRouterOutlet with Routes (without adding the slash) and the page will be shown.

ShaneK:
@ptmkenny Wow, that was a big one! I don't know how I didn't have that as a test case before, but I do now.

I have a dev build up that will hopefully work for you now, as well as having this case as [part of my E2E tests](https://github.com/ionic-team/ionic-framework/pull/30844/files#diff-dc711e1dd902302b5e24a10271a2fc7b5440b56684c4f94b05f18fab80df9de5):
```
 8.7.12-dev.11765307927.1f491e92
 ```

**Comment by ptmkenny** [comment]
@ShaneK Thank you for the speedy fix, but it seems to be a little more complicated. 

Follow-up: https://github.com/ionic-team/ionic-framework/issues/24177#issuecomment-3637401533

**Comment by ShaneK** [comment]
That would be a new issue, not the same one. You should report it in the main thread and I'll deal with it in another PR

File: packages/react-router/test/base/src/pages/relative-paths/RelativePaths.tsx [hunk 1]
```diff
@@ -0,0 +1,131 @@
...
+  IonBackButton,
+  IonButtons,
+} from '@ionic/react';
+import React from 'react';
+import { Route } from 'react-router-dom';
+
+/**
+ * This test page verifies that IonRouterOutlet correctly handles
+ * relative paths (paths without a leading slash) the same way
+ * React Router 6's Routes component does.
```
brandyscarney, line 20:
```suggestion
 * React Router 5's Routes component does.
```

Shouldn't this be a 5?

Do we even need to mention RR5?

packages/react-router/test/base/src/pages/relative-paths/RelativePaths.tsx
```diff
@@ -0,0 +1,131 @@
...
+ *
+ * Issue: https://github.com/ionic-team/ionic-framework/issues/24177#issuecomment-3624311206
```
brandyscarney:
```suggestion
```

I understand linking issues for regressions but I don't think we need to link to issue comments for basic functionality that existed in rr5.

packages/react-router/test/base/tests/e2e/specs/relative-paths.cy.js
```diff
@@ -0,0 +1,58 @@
...
+
+  it('should navigate to Page B (defined with relative path - no leading slash)', () => {
+    cy.visit(`http://localhost:${port}/relative-paths`);
+    cy.ionPageVisible('relative-paths-home');
+    cy.ionNav('ion-item', 'Go to Page B');
+    cy.ionPageVisible('relative-paths-page-b');
+    cy.get('[data-testid="page-b-content"]').should('contain', 'Page B');
+  });
+
+  it('should navigate to Page C (defined with relative path - no leading slash)', () => {
```
brandyscarney: What's the point of a 2nd test for a relative path?

brandyscarney (approved):
LGTM! 🔥

packages/react-router/test/base/src/pages/relative-paths/RelativePaths.tsx
ShaneK: Nope, this functionality did not exist in RR5. RR5 used all absolute links. This is intentionally referring to RR6.

packages/react-router/test/base/src/pages/relative-paths/RelativePaths.tsx
ShaneK: This didn't exist in RR5

packages/react-router/test/base/src/pages/relative-paths/RelativePaths.tsx
brandyscarney: Ahhh sorry I was reading it like it was referring to our own router 😮‍💨

packages/react-router/test/base/src/pages/relative-paths/RelativePaths.tsx
brandyscarney: Still - this is basic functionality that is provided by RR6 then.

ShaneK: Yeah, and we need to mimic it. I guess it doesn't need to be linked in a comment like this, but we do it with other regressions we find so 🤷‍♂️

brandyscarney: I think it would be more informative to link to the RR6 docs if we are just trying to make sure we support something. It's not a regression like "relative routing doesn't work in modals" it's an entire base concept that needs to be supported.

packages/react-router/test/base/tests/e2e/specs/relative-paths.cy.js
ShaneK: I thought I was testing another edge case, but it seems not. Probably something I ended up forgetting to clean up. Good catch! [f570837](https://github.com/ionic-team/ionic-framework/pull/30844/commits/f570837a6ede64b91860debf6ab699d2d6521815)

packages/react-router/src/ReactRouter/ReactRouterViewStack.tsx
```diff
@@ -469,7 +469,9 @@ export class ReactRouterViewStack extends ViewStacks {
     let parentPath: string | undefined = undefined;
     try {
       // Only attempt parent path computation for non-root outlets
-      if (outletId !== 'routerOutlet') {
+      // Root outlets have IDs like 'routerOutlet' or 'routerOutlet-2'
+      const isRootOutlet = outletId.startsWith('routerOutlet');
```
thetaPC, line 481: Since this is being used in several files, why not create a utility for it?

ShaneK: I mean, I haven't up until this point because it's such a simple single line of code, but I do agree that it probably should still be extracted into a utility - especially because it's not impossible that it changes at some point in the future. I think most of the refactoring I'll save until I have the RR6 branch more consolidated, I still have stuff pretty spread out and I'm not even sure how similar this file is in my next PR. But I think after the main RR6 branch is approved and all of the sub branches are merged into it, a single refactoring sweep would be a good idea to help clean up things like this and extremely large code blocks.

packages/react-router/test/base/tests/e2e/specs/relative-paths.cy.js
brandyscarney: Maybe you meant to navigate from one relative path to another? 🤔

PR #30845, opened 2025-12-08
fix(modal): prevent browser hang when using ModalController in Angular

ShaneK:

---------

When using ModalController to present a modal in Angular applications, the browser becomes non-responsive and hangs in some circumstances. This regression was introduced in #30544 with the addition of a MutationObserver that watches document.body with subtree: true to detect when a modal's parent element is removed from the DOM. For controller-based modals, this observer fires on every DOM mutation in the document, causing severe performance issues during Angular's change detection cycles.

## What is the new behavior?

The MutationObserver for parent removal detection is now skipped for controller-based modals and when the cached parent is the app root (document.body or ion-app). These parents are never removed from the DOM, so observing them is unnecessary. This prevents the performance issues while still maintaining the parent removal detection behavior for inline modals with meaningful parent elements. 

- [ ] Yes
- [X] No

## Other information

Current dev build:
```
8.7.12-dev.11765231260.1def96ab
```

### DIFF
--- a/core/src/components/modal/test/basic/modal.e2e.ts
+++ b/core/src/components/modal/test/basic/modal.e2e.ts
@@ -108,7 +108,7 @@ configs({ modes: ['ios'], directions: ['ltr'] }).forEach(({ title, config }) => 
     test('should not set up parentRemovalObserver for controller-created modals', async ({ page }, testInfo) => {
       testInfo.annotations.push({
         type: 'issue',
-        description: 'https://github.com/ionic-team/ionic-framework/pull/30845',
+        description: 'FW-6766',
       });
 
       await page.goto('/src/components/modal/test/basic', config);


core/src/components/modal/modal.tsx
thetaPC, line 1: The ticket mentions that it's hard to create reliable test. Would it be possible to create a controller-based modal, open, then verify that we can still interact with the Angular test app? If it can interact, then it passes?

ShaneK: Well, that's kind of the thing. This test already "exists", [there's e2e tests already doing this exact thing](https://github.com/ionic-team/ionic-framework/blob/def96ab3265c82f38dee106cb890b27474f4cdb5/core/src/components/modal/test/sheet/modal.e2e.ts#L47) and they didn't break before. That's because the scenario itself is a bit unique and I'm not sure what actually makes it that way.

However, I think this might be a suitable alternative: [da1911](https://github.com/ionic-team/ionic-framework/pull/30845/commits/bda1911a5365e2a3e2dc163133b9cfd67a7d4751)

In this commit I've created a test that ensures that controller-created modals do not have a mutation observer created for them, which should generally cover the bases here. Note: I had to use this PR as the "issue" because the actual issue is internal.

core/src/components/modal/test/basic/modal.e2e.ts
```diff
@@ -104,6 +104,28 @@ configs().forEach(({ title, screenshot, config }) => {
 });
 
 configs({ modes: ['ios'], directions: ['ltr'] }).forEach(({ title, config }) => {
+  test.describe(title('modal: parent removal observer'), () => {
+    test('should not set up parentRemovalObserver for controller-created modals', async ({ page }, testInfo) => {
+      testInfo.annotations.push({
+        type: 'issue',
+        description: 'https://github.com/ionic-team/ionic-framework/pull/30845',
```
thetaPC: Since the ticket is internal, then maybe we should do this instead:


```suggestion
        description: 'FW-6766',
```

It doesn't seem that description has to be a link.

core/src/components/modal/modal.tsx
thetaPC: I wonder if we can replicate it under the Angular test app instead of the core e2e tests? Regardless, I like your alternative! I'm happy with it. I only made one small suggestion.

PR #30849, opened 2025-12-09
fix(searchbar): clear timeouts

KillerCodeMonkey:
Issue number: resolves #

---------

We have flaky tests in an ionic angular project that root cause are not cleaned up timeouts.
I commented out the timeout in the searchbar componentWillLoad method. and after several runs no flaky tests at all.

My guess -> test runs faster than the 300ms it takes til the timeout runs. Everything is cleaned up, but not the ionic timeouts (i think i saw something similar in the "label" component)

## What is the new behavior?

- [ ] Yes
- [x] No

## Other information


KillerCodeMonkey:
closed in favor of #30851

PR #30850, opened 2025-12-09
fix(label): clear timeout

KillerCodeMonkey:
- clean up setTimeout in disconnectedCallback
- follow up to #30849
---------

timeouts are not cleaned up and may lead to unexpected behavior (especially in unit tests).

## What is the new behavior?

- [ ] Yes
- [x] No

## Other information

### DIFF
--- a/core/src/components/label/label.tsx
+++ b/core/src/components/label/label.tsx
@@ -18,6 +18,7 @@ import type { Color, StyleEventDetail } from '../../interface';
 })
 export class Label implements ComponentInterface {
   private inRange = false;
+  private loadTimeout?: number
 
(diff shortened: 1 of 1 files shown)


ShaneK:
Nice find, you'll need to clean up the build issue though, which in this case is stemming from the annoying typescript Timeout isn't a number issue.

KillerCodeMonkey:
Window as global is returning a number. Nodejs is returning a Timeout object as far as i know.

An clear timeout on window is accepting undefined|number.

So how do you type it in general?

KillerCodeMonkey:
closed in favor of #30851

PR #30851, opened 2025-12-09
fix(many): clear timeouts

### LINKED ISSUE #30860: fix: componentDidLoad setTimeouts not cleaned up
After checking the code of the `ion-searchbar` i saw that in multiple components timeouts that are defined in componentDidLoad are not cleared when disconnected.

In some circumstances this can lead to unwanted side effects or memory not cleaned up correctly.

Especially the `label` and `searchbar` components have in some situations "long" timeouts in ms.

This issue was created because it is needed for the PR: https://github.com/ionic-team/ionic-framework/pull/30851

@thetaPC

KillerCodeMonkey:
Issue number: resolves #30860

We have flaky tests in an ionic angular project that root cause are not cleaned up timeouts.
I commented out the timeout in the searchbar componentWillLoad method. and after several runs no flaky tests at all.

My guess -> test runs faster than the 300ms it takes til the timeout runs. Everything is cleaned up, but not the ionic timeouts (i think i saw something similar in other components)

## What is the new behavior?

- [ ] Yes
- [x] No

## Other information

### DIFF
--- a/core/src/components/app/app.tsx
+++ b/core/src/components/app/app.tsx
@@ -14,12 +14,13 @@ import { getIonMode } from '../../global/ionic-global';
 })
 export class App implements ComponentInterface {
   private focusVisible?: FocusVisibleUtility;
+  private loadTimeout?: ReturnType<typeof setTimeout> | undefined;
 
   @Element() el!: HTMLElement;
 
   componentDidLoad() {
     if (Build.isBrowser) {
-      rIC(async () => {
+      this.rIC(async () => {
         const isHybrid = isPlatform(window, 'hybrid');
         if (!config.getBoolean('_testing')) {
           import('../../utils/tap-click').then((module) => module.startTapClick(config));
@@ -60,6 +61,12 @@ export class App implements ComponentInterface {
     }
   }
 
+  disconnectedCallback() {
+    if (this.loadTimeout) {
+      clearTimeout(this.loadTimeout);
+    }
+  }
+
   /**
    * Used to set focus on an element that uses `ion-focusable`.
    * Do not use this if focusing the element as a result of a keyboard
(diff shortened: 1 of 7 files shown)


KillerCodeMonkey:
@ShaneK 

lint + build work locally.

but one test is failing:

```BASH
src/components/datetime/test/disabled/datetime.spec.tsx
  ● ion-datetime disabled › picker should be disabled in prefer wheel mode

    expect(received).toEqual(expected) // deep equality

    Expected: 4
    Received: 3

      31 |     const columns = datetime.shadowRoot!.querySelectorAll('ion-picker-column');
      32 |
    > 33 |     await expect(columns.length).toEqual(4);
         |                                  ^
      34 |
      35 |     columns.forEach((column) => {
      36 |       expect(column.disabled).toBe(true);

      at Object.<anonymous> (src/components/datetime/test/disabled/datetime.spec.tsx:33:34)
```

but it fails with clean master as well.

thetaPC:
@KillerCodeMonkey Thank you for the pull request!

In order for us to move forward, we need this PR to have a linked issue. Please [submit one](https://github.com/ionic-team/ionic-framework/issues) with a minimal repro, thanks!

KillerCodeMonkey:
How to create a repro repository on possible flaky tests and not cleaned up timeouts?

I mean this is just how to work with timeouts, intervals? Clean them up when not needed anymore to avoid side effects. And prevent memory leaks.

So how should i be able to create a repro repo?

There are already some cleanups in the code in other components...
It is just not done everywhere.

Greets

thetaPC:
@KillerCodeMonkey I agree that we should clean up the timeouts and we appreciate your work on this! However, if you can provide a reproduction of what you have even if it's flaky. By providing an official issue, we can verify that we haven't missed anything else. We might end up finding more issues related to timeouts during triage that we can keep tabs of.

thetaPC:
@KillerCodeMonkey don't worry too much on providing a repro. The most important thing is to create the issue, thanks!

KillerCodeMonkey:
i just created a clean issue and not a bug to avoid to set the required repo url field. i hope this is okay https://github.com/ionic-team/ionic-framework/issues/30860

thetaPC:
That should be fine! Thank you so much!!

KillerCodeMonkey:
any updates on this?

thetaPC:
Thank you for the pull request!! This will be available in an upcoming release of Ionic.

PR #30853, opened 2025-12-09
fix(popover): recalculate the content dimensions after the header has fully loaded

brandyscarney:
Issue number: internal

---------

A translucent header in a popover does not consistently render as translucent upon presenting due to the `offset-top` of the content being set to `0`. 

## What is the new behavior?
Watch the header for height changes using `ResizeObserver` and recalculate the content dimensions when the header height is greater than `0`.

- [ ] Yes
- [x] No

### DIFF
--- a/packages/angular/test/apps/ng17/package-lock.json
+++ b/packages/angular/test/apps/ng17/package-lock.json
@@ -40,6 +40,7 @@
         "@angular/language-service": "^17.0.2",
         "@playwright/test": "^1.54.2",
         "@types/express": "^4.17.7",
+        "@types/node": "^18.19.0",
         "@types/ws": "8.5.3",
         "@typescript-eslint/eslint-plugin": "^6.0.0",
         "@typescript-eslint/parser": "^6.0.0",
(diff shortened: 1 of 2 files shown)


core/src/components/popover/popover.tsx
```diff
@@ -540,6 +542,38 @@ export class Popover implements ComponentInterface, PopoverInterface {
...
+    const ro = new ResizeObserver(async () => {
+      if (header.offsetHeight > 0) {
+        ro.disconnect();
+        for (const contentEl of contentElements) {
+          await contentEl.recalculateDimensions();
+        }
+      }
+    });
+
+    ro.observe(header);
```
thetaPC: Shouldn't we [disconnect](https://developer.mozilla.org/en-US/docs/Web/API/ResizeObserver/disconnect) this when the component is unmounted?

core/src/components/content/content.tsx
```diff
@@ -254,6 +254,17 @@ export class Content implements ComponentInterface {
     }
   }
 
+  /**
+   * Recalculate content dimensions. Called by overlays (e.g., popover) when
+   * sibling elements like headers or footers have finished rendering and their
+   * heights are available, ensuring accurate offset-top calculations.
+   * @internal
```
thetaPC, line 261: Since this is an internal method, should it not show up in `components.d.ts`?

brandyscarney: The `@internal` flag just stops it from showing up in `api.txt`, the docs `core.json` file (which goes on to be used by the docs site), and the framework proxies, but it still gets added in `components.d.ts`. This is how other internal methods work as well: 

https://github.com/ionic-team/ionic-framework/blob/1c89cf06ac959f9c9a35a66f811227c244d3198b/core/src/components/accordion-group/accordion-group.tsx#L207-L215
https://github.com/ionic-team/ionic-framework/blob/1c89cf06ac959f9c9a35a66f811227c244d3198b/core/src/components.d.ts#L147-L149

core/src/components/popover/popover.tsx
brandyscarney: https://github.com/ionic-team/ionic-framework/pull/30853/commits/7fa3edeb476ef39c3b672b5b6b2fdeb5b3cbeedd

thetaPC (approved):
LGTM

ShaneK (approved):
Looks good to me!

PR #30854, opened 2025-12-10
fix(react-router): isolate tab history to prevent cross-tab back navigation

ShaneK:

---------

When switching between tabs using the tab bar in a React Router 6 application, clicking IonBackButton incorrectly navigates back to the previous tab. Each tab should have its own isolated navigation history stack, and the back button should only navigate within the current tab's history.

## What is the new behavior?

Tab navigation history is now properly isolated. When switching tabs via the tab bar:
  - The back button only navigates within the current tab's history
  - Pressing back on a tab root (with no navigation history) does nothing instead of navigating to the previous tab or default route
  - Within-tab navigation continues to work correctly (e.g., navigating to a details page and back)
  - Tab history is preserved when switching away and returning to a tab

- [ ] Yes
- [X] No

## Other information

Current dev build:
```
8.7.12-dev.11765377112.16762e5b
```

### DIFF
--- a/packages/react-router/test/base/tests/e2e/specs/tab-history-isolation.cy.js
+++ b/packages/react-router/test/base/tests/e2e/specs/tab-history-isolation.cy.js
@@ -1,0 +1,124 @@
+const port = 3000;
+
+describe('Tab History Isolation', () => {
+  it('should NOT navigate back to previous tab when using back button after tab bar switch', () => {
+    cy.visit(`http://localhost:${port}/tab-history-isolation/a`);
+    cy.ionPageVisible('tab-a');
+
+    cy.ionTabClick('Tab B');
+    cy.ionPageHidden('tab-a');
+    cy.ionPageVisible('tab-b');
+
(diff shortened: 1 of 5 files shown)


ptmkenny:
I installed build 8.7.12-dev.11765377112.16762e5b and manually tested by navigating around my app; this successfully restores the previous back button behavior (back button only works within a single tab; separate histories are kept per tab).  Thanks!

thetaPC (approved):
LGTM

packages/react-router/test/base/tests/e2e/specs/tab-history-isolation.cy.js
```diff
@@ -0,0 +1,124 @@
+const port = 3000;
```
brandyscarney, line 1: We should consider moving this to a global setting

ShaneK: Yeah, for sure. There's some things I'd like to clean up about this flow in general, and that one irks for a few reasons...but yeah. I guess we want to kill cypress anyway

PR #30831, opened 2025-12-02
feat(react-router): upgrade to react router 6

ShaneK:
Issue number: resolves #24177

---------

## What is the current behavior?
Currently, Ionic Framework React Router only supports React Router 5. This has many issues and unsupported/broken features.

## What is the new behavior?
With this change, Ionic Framework will support React Router 6 while still supporting transitions in the same way a native app does.

Most of what caused this change to take a long time is that React Router 5 and React Router 6 have fundamental differences in how they handle components once they're no longer part of the view. In this change, we move away from relying on React Router directly so much and have our own implementation for deciding how views get dealt with during navigation and when they're cleaned up, allowing for us to still transition between them like we need to while still using React Router as much as we possibly can.

This change will also lay the foundation for the migration to React Router 7, which will ideally be easier since most of the hard work has been dealt with here.

## Does this introduce a breaking change?

- [X] Yes
- [ ] No

## Other information

Current dev build (last updated 2026-04-27):

> **⚠️ WARNING:** If you're going to use this dev build on an existing react project, you'll need to migrate to React Router 6. Migrating a large project at this point might be a bad idea since this will not release until v9, which will require further migrations and have other breaking changes! I have a preview of migration documentation for this [here](https://ionic-docs-git-v9-react-router-ionic1.vercel.app/docs/updating/9-0#react-router).

```
8.8.4-dev.11777318673.18d001f6
```

The dev build linked above will be the last one for this branch alone. Everything going forward will be the major-9.0 branch, which may include large breaking changes.

PR #30861, opened 2025-12-11
fix(react-router): resolve relative route matching inside root-level splat routes

ShaneK:

---------

Routes with relative paths (e.g., `path="tab1/*"`) inside root-level splat routes (`path="*"`) do not match correctly. The parent path computation returns an incorrect value, causing routes to 404. Developers must use absolute paths (e.g., `path="/tab1/*"`) as a workaround.

## What is the new behavior?

Routes with relative paths now correctly match when the parent is a splat-only route. The `computeParentPath` function checks at root level for embedded wildcard routes (like `tab1/*`) as a fallback when no match is found at deeper levels, allowing relative paths to work correctly inside splat route parents.

- [ ] Yes
- [X] No

## Other information

Current dev build:
```
8.7.13-dev.11765477700.112ae0a3
```

### DIFF
--- a/packages/react-router/test/base/src/pages/nested-tabs-relative-links/NestedTabsRelativeLinks.tsx
+++ b/packages/react-router/test/base/src/pages/nested-tabs-relative-links/NestedTabsRelativeLinks.tsx
@@ -21,9 +21,9 @@ import { Link, Navigate, Route } from 'react-router-dom';
  * This test page verifies that relative links work correctly within
  * nested IonRouterOutlet components, specifically in a tabs-based layout.
  *
- * Issue: When using React Router's <Link to="relative"> inside nested
- * outlets with index routes, the relative path resolution can produce
- * incorrect URLs (e.g., /tab1/tab1/abc instead of /tab1/abc).
+ * Issue: When using React Router's <Link to="page-a"> inside the tab1 route
+ * with nested outlets and index routes, the relative path resolution can produce
+ * incorrect URLs (e.g., /tab1/tab1/page-a instead of /tab1/page-a).
  *
  * This test also verifies that absolute links work when a catch-all route
  * is present.


packages/react-router/src/ReactRouter/ReactRouterViewStack.tsx
```diff
@@ -384,30 +384,38 @@ export class ReactRouterViewStack extends ViewStacks {
...
-              absolutePathnameBase = parentPathnameBase;
-            }
+          const isSplatOnlyRoute = routePath === '*' || routePath === '/*';
+
+          // Get parent's pathnameBase for relative path resolution
+          const parentPathnameBase =
+            parentMatches.length > 0 ? parentMatches[parentMatches.length - 1].pathnameBase : '/';
+
+          // Start with the match's pathnameBase, falling back to routeInfo.pathname
+          // BUT: splat-only routes should use parent's base (v7_relativeSplatPath behavior)
```
thetaPC, line 397: Does v6 not account for this so you went with v7's approach?

packages/react-router/test/base/src/pages/nested-tabs-relative-links/NestedTabsRelativeLinks.tsx
```diff
@@ -0,0 +1,194 @@
...
+ * This test page verifies that relative links work correctly within
+ * nested IonRouterOutlet components, specifically in a tabs-based layout.
+ *
+ * Issue: When using React Router's <Link to="relative"> inside nested
+ * outlets with index routes, the relative path resolution can produce
+ * incorrect URLs (e.g., /tab1/tab1/abc instead of /tab1/abc).
+ *
+ * This test also verifies that absolute links work when a catch-all route
+ * is present.
+ */
```
thetaPC, line 30: Really liking these comments to the pages you've been adding. Thank you!!

packages/react-router/test/base/src/App.tsx
```diff
@@ -75,6 +77,8 @@ const App: React.FC = () => {
           <Route path="/nested-params/*" element={<NestedParams />} />
           {/* Test root-level relative path - no leading slash */}
           <Route path="relative-paths/*" element={<RelativePaths />} />
+          <Route path="/nested-tabs-relative-links/*" element={<NestedTabsRelativeLinks />} />
+          <Route path="/root-splat-tabs/*" element={<RootSplatTabs />} />
```
thetaPC, line 81: Let's also add them to `Main` so we can easily navigate to the from the home page.

packages/react-router/src/ReactRouter/ReactRouterViewStack.tsx
ShaneK: v6's approach is fairly inconsistent and causes some weird issues causing double path name segments. They've changed it in v7 to be more consistent and it didn't feel wise to introduce problematic behavior intentionally here

packages/react-router/test/base/src/App.tsx
ShaneK: Done! [04dd740](https://github.com/ionic-team/ionic-framework/pull/30861/commits/04dd7404ec1aeccdc46b0eb296ba18ae6e485bc6)

packages/react-router/src/ReactRouter/ReactRouterViewStack.tsx
thetaPC: That makes sense to me! Good call!

thetaPC (approved):
LGTM

packages/react-router/test/base/src/pages/nested-tabs-relative-links/NestedTabsRelativeLinks.tsx
```diff
@@ -0,0 +1,194 @@
...
+import React from 'react';
+import { Link, Navigate, Route } from 'react-router-dom';
+
+/**
+ * This test page verifies that relative links work correctly within
+ * nested IonRouterOutlet components, specifically in a tabs-based layout.
+ *
+ * Issue: When using React Router's <Link to="relative"> inside nested
+ * outlets with index routes, the relative path resolution can produce
+ * incorrect URLs (e.g., /tab1/tab1/abc instead of /tab1/abc).
```
brandyscarney: Nit: it might be clearer here if you use the same path name in the `to` field on `Link` as you do in the example (`relative` vs `abc`) and say that you are inside of the `tab1` route when navigating.

packages/react-router/test/base/src/pages/Main.tsx
brandyscarney, line 1: Doesn't need to be done now but we should consider grouping the pages together somehow, like the Angular test app does: https://github.com/ionic-team/ionic-framework/blob/2ee52d77c86ef29756370f1ed732998589ca2fd8/packages/angular/test/base/src/app/standalone/home-page/home-page.component.html#L13-L15

brandyscarney (approved):
Good fix! 🥇

packages/react-router/test/base/src/pages/Main.tsx
ShaneK: Yes, definitely! Or even the Vue test app looks really nice. I was thinking about doing that as a separate stand alone PR

packages/react-router/test/base/src/pages/nested-tabs-relative-links/NestedTabsRelativeLinks.tsx
ShaneK: Done! [918281b](https://github.com/ionic-team/ionic-framework/pull/30861/commits/918281bd619cf3ac7cff4a7c25f0a7b6e6b70dd4)

ShaneK:
@ptmkenny were you able to verify if this fixes your issue?

ptmkenny:
@ShaneK I haven't been able to confirm this in my app yet because of other issues, some of which I think have been fixed in other PRs. My testing is also a little delayed due to the holiday period. Is there a new build of the main react-router 6 PR that includes all the fixes committed so far?

ShaneK:
@ptmkenny Everything has been merged back into the [main RR6 PR](https://github.com/ionic-team/ionic-framework/pull/30831). The current dev build is `8.7.13-dev.11766090775.11e2bebb` as of this afternoon, but that thread will generally be updated with the current dev build when fixes are added in.

PR #30862, opened 2025-12-11
fix(content): automatically recalculate offset when header/footer changes

### LINKED ISSUE #26981: bug: ion-content set sometimes offsets wrong, sometimes correct
### Prerequisites

### Ionic Framework Version

v6.x

### Current Behavior

ion-content has the following CSS variables:

```
--offset-top: 56px;
--offset-bottom: 64px;
```

The values are set dynamically by Ionic, depending on the height of the content of ion-header and ion-footer. So far, so good. However, this only works correctly sometimes, and other times it does not. For example, we have a page like this:

```
<ion-header>
  <ion-toolbar>
    <ion-buttons slot="start">
      <ion-back-button defaultHref="/dashboard/home"></ion-back-button>
    </ion-buttons>
  </ion-toolbar>
</ion-header>

<ion-content fullscreen>My Cool Content</ion-content>

<ion-footer class="ion-no-border">
  <ion-toolbar style="min-height: initial; padding: 24px 24px 0 24px;">
    <p>Bla Bla Bla</p>
  </ion-toolbar>
  <ion-toolbar style="padding: 24px;">
    <button style="height: 86px;">Save</button>
  </ion-toolbar>
</ion-footer>
```

Since we assign a background image to our ion-content, the error in the incorrect offset calculation is very noticeable. When the appload event is fired again (`window.dispatchEvent(new Event('appload'));`) afterward, ion-content will correct the offsets, and they will be accurate. See in this video:

https://user-images.githubusercontent.com/4930169/226323071-38292ef3-e6fb-4197-8c05-400125f6313e.mp4

### Expected Behavior

By subsequently firing `window.dispatchEvent(new Event('appload'));` with a timeout, there is a brief flickering, as the user briefly sees the black area. I would wish that in any case, ion-content always determines the correct offset values by itself and automatically.

### Steps to Reproduce

I'm a bit confused right now... I have exactly the same structure of my app in StackBlitz, and the error does not occur there. Could this possibly have something to do with timings? Does this only happen in my app because there is significantly more data to load?

### Code Reproduction URL

https://stackblitz.com/edit/ionic6-angular13-jq74l5?file=src/app/features/feature-wallet/feature-wallet-overview/feature-wallet-overview.component.html

### Ionic Info

Ionic:

> Ionic CLI : 5.4.16

Utility:

> cordova-res : not installed
> native-run  : 1.7.2

System:

> NodeJS : v16.17.1
> npm    : 8.19.2
> OS     : Linux 5.15

### Additional Information

Here's what I've tried to solve the problem:

I have created a lifecycle.service.ts:
```ts
import { Injectable } from '@angular/core';
import { Subject } from 'rxjs';

@Injectable({
  providedIn: 'root',
})
export class LifecycleService {
  viewDidEnter$ = new Subject<void>();
  viewDidLeave$ = new Subject<void>();
}
```

app.component.ts
```ts
constructor(private lifecycleService: LifecycleService) {}

ngOnInit() {
  this.lifecycleService.viewDidEnter$.subscribe(() => {
    window.dispatchEvent(new Event('appload'));
  });
}
```

In ALL Views
```ts
constructor(private lifecycleService: LifecycleService) {}

ionViewDidEnter() {
  this.lifecycleService.viewDidEnter$.next();
}
```

As a result, the problem with the black bar occurs MUCH less frequently. However, it still happens occasionally.

HappyKnuckles:
Issue number:  resolves #26981

Currently, ion-content calculates its layout offsets (--offset-top/bottom) only on initialization or window resize.

If a developer dynamically changes the height of a header or footer (e.g., expanding a searchbar, wrapping text) or adds a toolbar conditionally to the DOM (e.g., via *ngIf), the ion-content does not adjust. This results in content being obscured behind the header/footer.

## What is the new behavior?
This PR adds a ResizeObserver and MutationObserver to ion-content to automatically handle dynamic layout changes.

ResizeObserver watches sibling <ion-header> and <ion-footer> elements and triggers a recalculation of --offset-top and --offset-bottom whenever their dimensions change.

MutationObserver watches the parent element to detect when <ion-header> or <ion-footer> elements are added to or removed from the DOM.

Removes the need for developers to manually listen for standard dynamic layout changes.

- [ ] Yes
- [x] No

## Other information

I have added a new E2E test suite in core/src/components/content/test/auto-offset which covers:

Visual Regression: Ensures no regressions on initial load.
Functional Header: Expands a header dynamically and asserts that content --offset-top update correctly.
Functional Footer: Expands a footer dynamically and asserts that content --offset-bottom update correctly.
Visual Confirmation: Snapshots verify that the content is physically pushed down/up when toolbars expand.

Before:
image

After:
image


core/src/components/content/content.tsx
```diff
@@ -420,6 +424,85 @@ export class Content implements ComponentInterface {
...
+    }
+
+    const parent = this.el.parentElement;
+    if (!parent) {
+      return;
+    }
+
+    if ('ResizeObserver' in window) {
+      this.resizeObserver = new ResizeObserver(() => {
+        this.resize();
```
ShaneK: Could you add a 100ms or so debounce to this? I'm worried about performance issues from this being called a ton during resizing

core/src/components/content/content.tsx
```diff
@@ -420,6 +424,85 @@ export class Content implements ComponentInterface {
...
+      });
+    }
+
+    if ('MutationObserver' in window) {
+      this.mutationObserver = new MutationObserver((mutations) => {
+        let shouldUpdate = false;
+
+        for (const mutation of mutations) {
+          if (mutation.type === 'childList') {
+            mutation.addedNodes.forEach((node: any) => {
```
ShaneK: I think you can just use `HTMLElement` for this? Shouldn't need to use `any`, preferably

core/src/components/content/content.tsx
```diff
@@ -420,6 +424,85 @@ export class Content implements ComponentInterface {
...
+              if (node.tagName === 'ION-HEADER' || node.tagName === 'ION-FOOTER') {
+                shouldUpdate = true;
+              }
+            });
+
+            mutation.removedNodes.forEach((node: any) => {
```
ShaneK: I think you can just use `HTMLElement` for this? Shouldn't need to use `any`, preferably

core/src/components/content/content.tsx
```diff
@@ -420,6 +424,85 @@ export class Content implements ComponentInterface {
...
+    if (!this.resizeObserver || !this.el.parentElement) {
+      return;
+    }
+
+    this.resizeObserver.disconnect();
+
+    const headers = this.el.parentElement.querySelectorAll('ion-header');
+    const footers = this.el.parentElement.querySelectorAll('ion-footer');
+
+    headers.forEach((header) => this.resizeObserver!.observe(header));
```
ShaneK: Please check `this.resizeObserver` and return out if it's not set, rather than using a non-null assertion in these two loops

ShaneK (changes requested):
Thanks for your PR! This is looking really good! Aside from the broken tests, this is what I would like to see changed

core/src/components/content/content.tsx
HappyKnuckles: I typed it as Node and checked for instanceOf HtmlElement in the loop.

HappyKnuckles:
@ShaneK any updates on this considering the review?

core/src/components/content/content.tsx
```diff
@@ -420,6 +424,93 @@ export class Content implements ComponentInterface {
...
+    if (!parent) {
+      return;
+    }
+
+    if ('ResizeObserver' in window) {
+      let timeout: any;
+      this.resizeObserver = new ResizeObserver(() => {
+        clearTimeout(timeout);
+        timeout = setTimeout(() => this.resize(), 100);
+      });
```
ShaneK, line 529: The `timeout` here lives inside the `ResizeObserver` callback closure, so `disconnectObservers` can't reach it to call `clearTimeout`. If a header or footer resizes within ~100ms of the content being removed from the DOM, the pending `setTimeout` still fires and calls `this.resize()`, which calls `forceUpdate` on a disconnected component.

Could you hoist this to a class field (something like `private observerResizeTimeout: ReturnType<typeof setTimeout> | null = null`) and clear it in `disconnectObservers`? The existing `resizeTimeout` field at the top of the class is doing the same thing for the `onResize` handler at line 192.

core/src/components/content/content.tsx
```diff
@@ -420,6 +424,93 @@ export class Content implements ComponentInterface {
...
+      return;
+    }
+
+    const parent = this.el.parentElement;
+    if (!parent) {
+      return;
+    }
+
+    if ('ResizeObserver' in window) {
+      let timeout: any;
```
ShaneK: Could you type this as `ReturnType<typeof setTimeout> | null` instead of `any`? It matches the `resizeTimeout` field at the top of the class. If you take the suggestion above and move this to a class field, the field declaration handles the typing for you.

core/src/components/content/content.tsx
```diff
@@ -420,6 +424,93 @@ export class Content implements ComponentInterface {
...
+    const footers = this.el.parentElement.querySelectorAll('ion-footer');
+
+    const observer = this.resizeObserver;
+
+    if (observer === null || observer === undefined) {
+      return;
+    }
+
+    headers.forEach((header) => observer.observe(header));
+    footers.forEach((footer) => observer.observe(footer));
```
ShaneK:
```suggestion
    const observer = this.resizeObserver;

    headers.forEach((header) => observer.observe(header));
    footers.forEach((footer) => observer.observe(footer));
```

The early return on line 495 already guarantees `this.resizeObserver` is non-null by the time we get here, so the `observer === null || observer === undefined` check can't ever be true. The `const observer` alias is still worth keeping because it narrows the type for TypeScript inside the `forEach` callbacks.

core/src/components/content/content.tsx
```diff
@@ -420,6 +424,93 @@ export class Content implements ComponentInterface {
...
+
+  private refreshResizeObserver() {
+    if (!this.resizeObserver || !this.el.parentElement) {
+      return;
+    }
+
+    this.resizeObserver.disconnect();
+
+    const headers = this.el.parentElement.querySelectorAll('ion-header');
+    const footers = this.el.parentElement.querySelectorAll('ion-footer');
```
ShaneK: `querySelectorAll('ion-header')` walks the whole descendant tree of the parent, not just direct children. If the page contains a custom component that itself wraps an `ion-header` or `ion-footer`, this observer will pick those up too, and resizing them will trigger spurious `--offset-top` / `--offset-bottom` recalcs on this content.

The `MutationObserver` above is already scoped to direct children via `{ childList: true }`, so it'd be consistent to do the same here. Something like `parent.querySelectorAll(':scope > ion-header, :scope > ion-footer')` would only match siblings.

core/src/components/content/test/auto-offset/content.e2e.ts
```diff
@@ -0,0 +1,48 @@
...
+
+      await expect(content).not.toHaveCSS('--offset-bottom', before);
+
+      const after = await content.evaluate((el: HTMLElement) =>
+        getComputedStyle(el).getPropertyValue('--offset-bottom')
+      );
+
+      expect(after).not.toBe(before);
+
+      await expect(page).toHaveScreenshot(screenshot(`content-auto-offset-footer-updated`));
```
ShaneK:
```suggestion
      await expect(page).toHaveScreenshot(screenshot(`content-auto-offset-footer-updated`));
```

The `not.toHaveCSS` on line 37 already verifies `--offset-bottom` changed, so reading the value again and re-asserting it's different is checking the same thing twice. The header test above doesn't have this extra block, would be nice to keep them consistent.

core/src/components/content/test/auto-offset/content.e2e.ts
```diff
@@ -0,0 +1,48 @@
+import { expect } from '@playwright/test';
+import { test, configs } from '@utils/test/playwright';
+
+configs({ modes: ['ios'] }).forEach(({ title, screenshot, config }) => {
```
ShaneK: Any reason to restrict this suite to iOS? The new observer code doesn't have mode-specific paths, so dropping `modes: ['ios']` would also cover MD. If there's a reason to keep it iOS-only (translucent header or similar), a quick comment explaining would help future readers.

ShaneK (changes requested):
I'm so sorry I didn't get back to this! I guess it was too close to Christmas and probably right before we took holiday break 🫠 I did notice some issues that I'm hoping won't be too hard to clean up, let me know if you need clarification

core/src/components/content/content.tsx
HappyKnuckles: I adjusted it in the latest commit.
However regarding the timeout: i noticed that the timeout is very noticeable visually.
Do you think it is worth switching to `requestAnimationFrame`?

core/src/components/content/content.tsx
HappyKnuckles: Adjusted in latest commit.

core/src/components/content/content.tsx
HappyKnuckles: Adjusted in latest commit.

core/src/components/content/content.tsx
HappyKnuckles: Adjusted in latest commit.

core/src/components/content/test/auto-offset/content.e2e.ts
HappyKnuckles: Adjusted in latest commit.

core/src/components/content/test/auto-offset/content.e2e.ts
HappyKnuckles: Originally I scoped it to iOS because the offset is most visible on translucent headers, but you're right that the observer code is mode-agnostic. Dropped the `modes: ['ios']` restriction so it covers MD too.

core/src/components/content/content.tsx
ShaneK: Awesome! Yeah, a rAF might be a better solution there, probably worth trying out. It should be a pretty quick test to see if it works better, basically the same structure, you'd even want to cache it and clear it out the same.

HappyKnuckles: rAF indeed seems to improve it, so i will go with that solution.

lincolnthree:
If it helps anyone, we are currently patching this in manually and it's working well for us! Thank you for the fix. Looking forward to seeing it merged.

PR #30866, opened 2025-12-12
refactor(core): use Capacitor safe-area CSS variables on older WebViews

thetaPC:
Issue number: internal

---------

The safe area variables are only reliant on `env` variables that are provided by devices.

## What is the new behavior?

Capacitor 8 has released [safe area variable fallbacks](https://capacitorjs.com/docs/apis/system-bars#android-note) to provide consistent behaviors with older Android devices:

> Due to a [bug](https://issues.chromium.org/issues/40699457) in some older versions of Android WebView (< 140), correct safe area values are not available via the safe-area-inset-x CSS env variables. This plugin will inject the correct inset values into a new CSS variable(s) named --safe-area-inset-x that you can use as a fallback in your frontend styles.

- Updated safe area variables to use the fallbacks provided by Capacitor.

- [ ] Yes
- [x] No

## Other information

Changes to the [core file](https://github.com/ionic-team/ionic-framework/blob/2ee52d77c86ef29756370f1ed732998589ca2fd8/core/src/css/core.scss#L253-L260) will be done on `main`.


brandyscarney:
I'm marking this as a refactor so it doesn't show up in the Changelog twice since the fix was made on `main` already.

core/src/components/app/test/safe-area/app.e2e.ts-snapshots/app-picker-diff-ionic-md-ltr-light-Mobile-Chrome-linux.png
brandyscarney, line 1: It looks like this test uses the legacy Picker and the `ionic` theme does not have styles for it so this is expected: https://ionic-framework-git-fw-6987-next-ionic1.vercel.app/src/components/app/test/safe-area?ionic:theme=ionic

brandyscarney (approved):

PR #30867, opened 2025-12-12
fix(react-router): reject orphaned IonPage registrations during navigation

### LINKED ISSUE #28878: bug: weird routing behavior when view content is changed while navigation
### Prerequisites

### Ionic Framework Version

v7.x

### Current Behavior

When the page content is changed while we push new route, an invalid view is rendered. 
In our project, we navigate to the shop menu when the cart is empty. Also we want to show an empty view for the cart if a user opens the cart URL manually. So somewhere in the code we have the following:
```
function Cart() {
  //...

  const clearCartAndNavigateToMenu = () => {
    clearCart().then(() => {
      push('/menu/')
    })
  }

  if(isCartEmpty) {
   return <EmptyView />
  }

  return (
    // cart content here
  )
}
```

After navigation the route changed to '/menu/' and we expect to see the menu page. But instead, we see EmptyView although the route has changed.

### Expected Behavior

We expect to see correct view when the route is changed. 

### Steps to Reproduce

Minimal reprodusible example provided in the repo.
STR:
1. Open Tab2
2. Click "Go to list page" button
3. Click "Remove all items and navigate to home" button
5. The view will flicker double times and you will see the correct URL ('/tab2') but an invalid view 'There no items'reproducible

### Code Reproduction URL

https://github.com/legendar/ionic-issues

### Ionic Info

Ionic:

   Ionic CLI       : 7.2.0 (/www/tmp/ionic-issues/node_modules/@ionic/cli)
   Ionic Framework : @ionic/react 7.6.6

Capacitor:

   Capacitor CLI      : 5.6.0
   @capacitor/core    : 5.6.0

Utility:

   native-run  : 2.0.0

System:

   NodeJS : v18.15.0 (/home/lg/.nvm/versions/node/v18.15.0/bin/node)
   npm    : 9.5.0
   OS     : Linux 5.15

### Additional Information

After further debugging I found that there is double animation and the reason why this happens is this line in Ionic Router: https://github.com/ionic-team/ionic-framework/blob/e86f4f1cc9f8ce583c751fc63d44e5f8ccf82ae6/packages/react-router/src/ReactRouter/StackManager.tsx#L209

Before clearing the list we have the following components structure:
```
Route -> Tab2Nested -> IonPage -> IonContent
```
when we clear the list `Tab2EmptyView` view will be rendered and the next components structure is:
```
Route -> Tab2Nested -> Tab2EmptyView-> IonPage -> IonContent
```
this change causes the `IonPage` component remounting, `registerIonPage` call in IonicRouter and extra page transition. 
Alongside we trigger navigation to the home/Tab2 page and as a result, we have two different transitions at the same time.
And this is why we have inconsistency between the route path and rendered view. 

So we need to somehow prevent transition to the same page and/or prevent to call several transitions at the same time. 

As a workaround we can move `IonPage` to `Tab2Nested ` so we will have a similar component structure before and after removing a cart:
```
Route -> CartDetail -> IonPage -> IonContent ->
Route -> CartDetail -> IonPage -> Tab2Nested -> IonContent
```
```
if(items.length === 0) {
  return <IonPage><Tab2EmptyView /></IonPage>
}
```
In this case IonPage will not be remounted and we will not have an extra transition.

ShaneK:
Issue number: resolves #28878

---------

When a React component conditionally renders different `IonPage` structures based on state (e.g., a list view vs. an empty state view), and a state change happens simultaneously with navigation, the wrong view is displayed. The URL shows the correct destination route, but the stale/orphaned `IonPage` from the re-rendered component remains visible instead. Additionally, swipe-to-go-back gestures fail to properly unmount the leaving view, causing pages to remain in the DOM and preventing correct navigation.

## What is the new behavior?

When an `IonPage` registration occurs for a route we're navigating away from, and the new page has a different `data-pageid` than the existing page element, the registration is rejected. The orphaned page is hidden and scheduled for removal, ensuring the correct destination view is displayed. Cross-outlet swipe back now works correctly by searching all outlets when the target view isn't found in the current outlet.

- [ ] Yes
- [X] No

## Other information

Current dev build:
```
8.7.13-dev.11765829391.14bc580c
```

### DIFF
--- a/packages/react-router/src/ReactRouter/StackManager.tsx
+++ b/packages/react-router/src/ReactRouter/StackManager.tsx
@@ -207,8 +207,6 @@ export class StackManager extends React.PureComponent<StackManagerProps> {
     }
 
     if (routeInfo.routeAction === 'replace') {
-      // For replace actions, decide whether to unmount the leaving view.
-      // The key question is: are these routes in the same navigation context?
       const enteringRoutePath = enteringViewItem?.reactElement?.props?.path as string | undefined;
       const leavingRoutePath = leavingViewItem?.reactElement?.props?.path as string | undefined;
 
--- a/packages/react-router/test/base/tests/e2e/specs/content-change-navigation.cy.js
+++ b/packages/react-router/test/base/tests/e2e/specs/content-change-navigation.cy.js
@@ -1,11 +1,13 @@
 /**
  * Verifies that when view content changes (causing IonPage to remount)
  * while navigation is happening, the correct view is displayed.
+ *
+ * @see https://github.com/ionic-team/ionic-framework/issues/28878
  */
 
 const port = 3000;
 
-describe('Content Change Navigation Tests (Issue #28878)', () => {
+describe('Content Change Navigation Tests', () => {
   it('should navigate to list page correctly', () => {
     cy.visit(`http://localhost:${port}/content-change-navigation`);
     cy.ionPageVisible('content-nav-home');


packages/react-router/src/ReactRouter/StackManager.tsx
```diff
@@ -207,9 +207,42 @@ export class StackManager extends React.PureComponent<StackManagerProps> {
     }
 
     if (routeInfo.routeAction === 'replace') {
-      return true;
+      // For replace actions, decide whether to unmount the leaving view.
+      // The key question is: are these routes in the same navigation context?
```
thetaPC: Is this a question for the team? Or a leftover comment that needs to be removed?

packages/react-router/test/base/tests/e2e/specs/content-change-navigation.cy.js
```diff
@@ -0,0 +1,52 @@
+/**
+ * Verifies that when view content changes (causing IonPage to remount)
+ * while navigation is happening, the correct view is displayed.
```
thetaPC, line 3: Since this PR resolves an existing issue that's not tied to RR6, then let's add the issue link here. It follows the structure we have for other tests.

```suggestion
 * while navigation is happening, the correct view is displayed.
 * 
 * @see https://github.com/ionic-team/ionic-framework/issues/28878
```

packages/react-router/test/base/tests/e2e/specs/content-change-navigation.cy.js
```diff
@@ -0,0 +1,52 @@
...
+ */
+
+const port = 3000;
+
+describe('Content Change Navigation Tests (Issue #28878)', () => {
```
thetaPC: By adding the link above, we don't need to mention it again.

```suggestion
describe('Content Change Navigation Tests, () => {
```

packages/react-router/src/ReactRouter/StackManager.tsx
ShaneK: Left over, good find - [93cca29](https://github.com/ionic-team/ionic-framework/pull/30867/commits/93cca29a16806b6b6dc51958f5c0d17a8b5c4abb)

packages/react-router/test/base/tests/e2e/specs/content-change-navigation.cy.js
ShaneK: Done! [93cca29](https://github.com/ionic-team/ionic-framework/pull/30867/commits/93cca29a16806b6b6dc51958f5c0d17a8b5c4abb)

packages/react-router/test/base/tests/e2e/specs/content-change-navigation.cy.js
ShaneK: Done! [93cca29](https://github.com/ionic-team/ionic-framework/pull/30867/commits/93cca29a16806b6b6dc51958f5c0d17a8b5c4abb)

brandyscarney (approved):

PR #30799, opened 2025-11-19
feat(tokens): add gray and text colors

thetaPC:
Issue number: internal

---------

## What is the current behavior?

Themes are still using the stepped color variables.

## What is the new behavior?

The CSS variable `--ion-background-color-step-{number}` is no longer recommended for use and has been replaced by a more intuitive and standardized naming convention as part of Ionic's new theming structure. The functionality provided by the old variable has been replaced with a new set of CSS variables that use the same color values but are named: `--ion-color-gray-{number}`.

This new structure aligns with how other popular design systems and frameworks are setting up their color palettes (e.g., using names like `gray`, `red`, `blue`, followed by a number to denote the shade). Additionally, it follows the proposal agreed upon within the design doc.

- Added gray tokens based on themes and palettes
- Replaced background stepped color variables with gray tokens

The CSS variable `--ion-text-color-step-*` is being replaced by `--ion-color-text-*` to adhere to the standardized naming convention for Ionic's new theming structure (`--ion-color-*-*`). This change aims for greater clarity and consistency within the framework's color palette.

- Added text tokens based on themes and palettes
- Replaced text stepped color variables with text tokens

The original thought was to use only the `--ion-color-gray-*` variables for all purposes (backgrounds, text, icons, etc.). However, it was quickly determined that text color requirements are not a 1:1 match with the gray steps designed for backgrounds and other elements. Text often requires a different shade to maintain proper contrast and readability across various theme palettes (especially between high contrast and high contrast dark modes). Introducing the dedicated functional variable `--ion-color-text-*` solves this issue by allowing text color to be defined independently based on the current palette's needs.

## Does this introduce a breaking change?

- [x] Yes
- [ ] No

You simply need to swap the text of the background stepped variables with the gray variables. The number suffix remains exactly the same:

| Old Variable (Deprecated) | New Variable |
|--------|--------|
| `--ion-background-color-step-50` | `--ion-color-gray-50` |
| `--ion-background-color-step-900` | `--ion-background-color-step-900` | 

Unlike the migration for background colors, the number suffix for text colors is not a direct swap. Due to the inverse nature required for contrast, the new text variable number is calculated by subtracting the old number from 1000.

| Old Variable (Deprecated) | Calculation | New Variable |
|--------|--------|--------|
| `--ion-text-color-step-50` | `1000−50` | `--ion-color-text-950` |
| `--ion-text-color-step-900` | `1000-900` | `--ion-color-text-100` | 

## Other information

[...]

PR #30874, opened 2025-12-15
feat(themes): enable all palettes

thetaPC:
Issue number: internal

---------

The new theming structure does not allow high contrast and high contrast dark to render properly. This can be seen when running a test that uses `set-content`.

## What is the new behavior?

- High contrast and high contrast dark have been enabled to the testing environments
- The theme files for high contrast and high contrast dark have been added under `themes`.

- [x] Yes
- [ ] No

## Other information

[Colors preview](https://ionic-framework-git-set-content-ionic1.vercel.app/src/themes/test/color/)

How to test:
1. Verify that tests related to palettes are passing especially those that use `setContent`
2. Verify that the preview page renders the right colors for each possible variation of palettes and themes

### DIFF
--- a/core/scripts/testing/scripts.js
+++ b/core/scripts/testing/scripts.js
@@ -93,6 +93,16 @@ const DEFAULT_PALETTE = 'light';
   const configDarkMode = window.Ionic?.config?.customTheme?.palette?.dark?.enabled === 'always' ? 'dark' : null;
   const configHighContrastMode = window.Ionic?.config?.customTheme?.palette?.highContrast?.enabled === 'always' ? 'high-contrast' : null;
   const configHighContrastDarkMode = window.Ionic?.config?.customTheme?.palette?.highContrastDark?.enabled === 'always' ? 'high-contrast-dark' : null;
+  /**
+   * Ensure window.Ionic.config is defined before importing 'testing/scripts'
+   * in the test HTML to properly initialize the palette configuration below.
+   * 
+   * Example:
+   * <script>
+   *    window.Ionic = { config: { customTheme: { palette: { ... } } } };
+   * </script>
+   * <script src="testing/scripts.js"></script>
+   */
   const configPalette = configDarkMode || configHighContrastMode || configHighContrastDarkMode;
   const paletteQuery = window.location.search.match(/palette=([a-z-]+)/);
   const paletteHash = window.location.hash.match(/palette=([a-z-]+)/);


core/src/components/toolbar/test/basic/toolbar.e2e.ts-snapshots/toolbar-basic-text-buttons-inside-content-ionic-md-ltr-dark-Mobile-Chrome-linux.png
thetaPC, line 1: The difference is the text color. The old version is showing `#1f1f1f` and the updated version is showing `#000`. The text color was not being displayed correctly until I updated the `set-content` file. If you compare the text color to `next` then you can verify that it should be `#000`.

core/src/components/toolbar/test/basic/toolbar.e2e.ts-snapshots/toolbar-basic-text-icon-buttons-ionic-md-ltr-dark-Mobile-Chrome-linux.png
thetaPC, line 1: The difference is the text color. The old version is showing `#1f1f1f` and the updated version is showing `#000`. The text color was not being displayed correctly until I updated the `set-content` file. If you compare the text color to `next` then you can verify that it should be `#000`.

core/scripts/testing/scripts.js
```diff
@@ -119,8 +126,15 @@ const DEFAULT_THEME = 'md';
...
       // to set the enabled property to 'always'
+      // TODO(FW-4004): Implement dark mode
       if (paletteName === 'dark' && theme.palette?.dark) {
         theme.palette.dark.enabled = 'always';
+      // TODO(FW-4005): Implement high contrast mode
+      } else if (paletteName === 'high-contrast' && theme.palette?.highContrast) {
+        theme.palette.highContrast.enabled = 'always';
+      // TODO(FW-4005): Implement high contrast dark mode
+      } else if (paletteName === 'high-contrast-dark' && theme.palette?.highContrastDark) {
+        theme.palette.highContrastDark.enabled = 'always';
```
thetaPC: We can decide at the time of dark and high contrast implementation if this order makes sense.

core/src/components/modal/test/dark-mode/index.html
```diff
@@ -10,6 +10,21 @@
...
     <link href="../../../../../scripts/testing/styles.css" rel="stylesheet" />
+    <script>
+      // Need to be called before loading Ionic else
+      // the scripts.js logic runs too early.
+      window.Ionic = {
+        config: {
+          customTheme: {
+            palette: {
+              dark: {
+                enabled: 'always',
```
thetaPC, line 23: This file is meant to be always be in dark mode as stated in the file path.

core/src/themes/md/dark.tokens.ts
```diff
@@ -9,6 +9,11 @@ export const darkTheme: DarkTheme = {
   textColor: '#ffffff',
   textColorRgb: '255, 255, 255',
 
+  // TODO(FW-6864): Remove once IonToolbar themes are added
+  toolbar: {
+    background: '#1f1f1f',
+  },
```
thetaPC, line 15: Tests fail when `--ion-toolbar-background` is not set in certain cases like `md/dark`. Since the goal of this PR is to enable the palettes, I didn't think create this variable under the components since as of now it would have rendered it as `--ion-components-ion-toolbar-background`. I would have had to either update the generate function to not add the `--ion-components` suffix or update all the `--ion-toolbar-background` to the new variable if I wanted to do it properly. That would have been outside of the scope so this approach was taken along with adding a TODO to come back.

core/src/themes/md/high-contrast-dark.tokens.ts
```diff
@@ -0,0 +1,58 @@
...
+export const highContrastDarkTheme: HighContrastDarkTheme = {
+  ...baseHighContrastDarkTheme,
+
+  backgroundColor: '#121212',
+  textColor: '#000000',
+
+  // TODO(FW-6864): Remove once IonToolbar themes are added
+  toolbar: {
+    background: '#1f1f1f',
+  },
```
thetaPC, line 13: Tests fail when `--ion-toolbar-background` is not set in certain cases like `md/dark`. Since the goal of this PR is to enable the palettes, I didn't think create this variable under the components since as of now it would have rendered it as `--ion-components-ion-toolbar-background`. I would have had to either update the generate function to not add the `--ion-components` suffix or update all the `--ion-toolbar-background` to the new variable if I wanted to do it properly. That would have been outside of the scope so this approach was taken along with adding a TODO to come back.

core/src/themes/themes.interfaces.ts
```diff
@@ -17,6 +17,9 @@ export type BaseTheme = {
     [key: string]: string;
   };
 
+  // TODO(FW-6864): Remove once IonToolbar themes are added
+  toolbar?: any;
+
```
thetaPC, line 22: Tests fail when `--ion-toolbar-background` is not set in certain cases like `md/dark`. Since the goal of this PR is to enable the palettes, I didn't think create this variable under the components since as of now it would have rendered it as `--ion-components-ion-toolbar-background`. I would have had to either update the generate function to not add the `--ion-components` suffix or update all the `--ion-toolbar-background` to the new variable if I wanted to do it properly. That would have been outside of the scope so this approach was taken along with adding a TODO to come back.

core/src/components/toast/test/a11y/toast.e2e.ts
```diff
@@ -236,7 +236,8 @@ configs({ directions: ['ltr'] }).forEach(({ title, screenshot, config }) => {
  */
 configs({ directions: ['ltr'], palettes: ['high-contrast-dark', 'high-contrast'] }).forEach(
   ({ title, config, screenshot }) => {
-    test.describe(title('toast: high contrast: buttons'), () => {
+    // TODO(FW-4005): Once high contrast themes are fully implemented in ionic modular, remove the skips from these tests
+    test.describe.skip(title('toast: high contrast: buttons'), () => {
```
thetaPC, line 240: This test fails because there are [certain styles](https://github.com/ionic-team/ionic-framework/blob/8af2f13e3cc10f61d455f4aca4fa94680fcdb943/core/src/css/palettes/high-contrast.scss#L145-L168) that are not available for toast.  The point of this PR is to enable the palettes to work within tests and to not meddle with any components. Those styles should be reworked when we are fully implementing high contrast into the new theming structure.

core/scripts/testing/scripts.js
```diff
@@ -87,18 +88,24 @@ const DEFAULT_THEME = 'md';
    * Values can be `light`, `dark`, `high-contrast`,
    * or `high-contrast-dark`. Default to `light` for tests.
    */
-  const validPalettes = ['light', 'dark', 'high-contrast', 'high-contrast-dark'];
+  const validPalettes = [DEFAULT_PALETTE, 'dark', 'high-contrast', 'high-contrast-dark'];
```
brandyscarney:
```suggestion
  const validPalettes = ['light', 'dark', 'high-contrast', 'high-contrast-dark'];
```

It doesn't make sense to use the default palette var here if it ever gets changed.

core/scripts/testing/scripts.js
```diff
@@ -119,8 +126,15 @@ const DEFAULT_THEME = 'md';
 
       // If a specific palette is requested, modify the palette structure
       // to set the enabled property to 'always'
+      // TODO(FW-4004): Implement dark mode
```
brandyscarney:
```suggestion
      // TODO(FW-4004): Implement dark palette
```

core/scripts/testing/scripts.js
```diff
@@ -119,8 +126,15 @@ const DEFAULT_THEME = 'md';
...
       if (paletteName === 'dark' && theme.palette?.dark) {
         theme.palette.dark.enabled = 'always';
+      // TODO(FW-4005): Implement high contrast mode
```
brandyscarney:
```suggestion
      // TODO(FW-4005): Implement high contrast palette
```

core/scripts/testing/scripts.js
```diff
@@ -119,8 +126,15 @@ const DEFAULT_THEME = 'md';
...
+      } else if (paletteName === 'high-contrast' && theme.palette?.highContrast) {
+        theme.palette.highContrast.enabled = 'always';
+      // TODO(FW-4005): Implement high contrast dark mode
```
brandyscarney:
```suggestion
      // TODO(FW-4005): Implement high contrast dark palette
```

core/src/components/modal/test/dark-mode/index.html
brandyscarney, line 1: We don't need to do it now but I think we could remove this test since we can just append the palette name to change it to the dark palette.

core/src/themes/base/default.tokens.ts
```diff
@@ -1,6 +1,8 @@
 import type { DefaultTheme } from '../themes.interfaces';
 
 import { darkTheme } from './dark.tokens';
+import { highContrastDarkTheme } from './high-contrast-dark.tokens';
+import { highContrastTheme } from './high-contrast.tokens';
```
brandyscarney:
```suggestion
import { highContrastTheme } from './high-contrast.tokens';
import { highContrastDarkTheme } from './high-contrast-dark.tokens';
```

nit

core/src/utils/test/playwright/page/utils/set-content.ts
```diff
@@ -68,6 +68,34 @@ export const setContent = async (page: Page, html: string, testInfo: TestInfo, o
...
+   * that use `setContent`.
+   */
+  const customTheme = {
+    palette: {
+      dark: {
+        enabled: palette === 'dark' ? 'always' : 'never',
+      },
+      highContrast: {
+        enabled: palette === 'high-contrast' ? 'always' : 'never',
+      },
```
brandyscarney, line 92: So when none are enabled it passes this:

```ts
{
  palette: {
    dark: { enabled: 'never' },
    highContrast: { enabled: 'never' },
    highContrastDark: { enabled: 'never' }
  }
}
```

Is this desired? Or should nothing be passed?

core/src/themes/base/high-contrast.tokens.ts
```diff
@@ -0,0 +1,213 @@
...
+        base: colors.primary,
+        contrast: '#fff',
+        foreground: colors.primary,
+        shade: mix(colors.primary, '#000', '12%'),
+        tint: mix(colors.primary, '#fff', '10%'),
+      },
+      subtle: {
+        base: mix('#fff', colors.primary, '8%'),
+        contrast: colors.primary,
+        foreground: mix(colors.primary, '#000', '12%'),
```
brandyscarney, line 30: Is there a reason foreground is 12% darker than the primary color?

core/src/themes/ios/high-contrast-dark.tokens.ts
```diff
@@ -0,0 +1,53 @@
+import { highContrastDarkTheme as baseHighContrastDarkTheme } from '../base/high-contrast-dark.tokens';
+import type { HighContrastDarkTheme } from '../themes.interfaces';
+
+export const highContrastDarkTheme: HighContrastDarkTheme = {
+  ...baseHighContrastDarkTheme,
+
+  backgroundColor: '#000000',
+  textColor: '#ffffff',
+
+  backgroundColorStep: {
```
brandyscarney: Why are we redefining the stepped colors for ios when they're the same as the base?

core/scripts/testing/scripts.js
```diff
@@ -87,18 +88,24 @@ const DEFAULT_THEME = 'md';
    * Values can be `light`, `dark`, `high-contrast`,
    * or `high-contrast-dark`. Default to `light` for tests.
    */
-  const validPalettes = ['light', 'dark', 'high-contrast', 'high-contrast-dark'];
+  const validPalettes = [DEFAULT_PALETTE, 'dark', 'high-contrast', 'high-contrast-dark'];
+
+  const configDarkMode = window.Ionic?.config?.customTheme?.palette?.dark?.enabled === 'always' ? 'dark' : null;
+  const configHighContrastMode = window.Ionic?.config?.customTheme?.palette?.highContrast?.enabled === 'always' ? 'high-contrast' : null;
+  const configHighContrastDarkMode = window.Ionic?.config?.customTheme?.palette?.highContrastDark?.enabled === 'always' ? 'high-contrast-dark' : null;
+  const configPalette = configDarkMode || configHighContrastMode || configHighContrastDarkMode;
```
brandyscarney, line 106: We should add somewhere that this needs to be defined before the `testing/scripts` import like:

```html
<script>
  window.Ionic = {
    config: {
      customTheme: {
        palette: {
          dark: {
            enabled: 'never'
          },
          highContrast: {
            enabled: 'always'
          },
          highContrastDark: {
            enabled: 'never'
          },
        },
      },
    },
  };
</script>
<script src="../../../../../scripts/testing/scripts.js"></script>
```

core/scripts/testing/scripts.js
thetaPC: Good point! https://github.com/ionic-team/ionic-framework/pull/30874/commits/4f562c4b03eeb97da0fa81d63f0cb40573f2a6da

core/scripts/testing/scripts.js
thetaPC: https://github.com/ionic-team/ionic-framework/pull/30874/commits/33875fc6c33f91fd48cc55987a00fb5931af5aa8

core/scripts/testing/scripts.js
thetaPC: https://github.com/ionic-team/ionic-framework/pull/30874/commits/10366c084b25e1ef3aebc19ce009f4befb100b5e

core/scripts/testing/scripts.js
thetaPC: https://github.com/ionic-team/ionic-framework/pull/30874/commits/b5b2576735ea0d4b686bc5665671bbabe0927fbc

core/src/components/modal/test/dark-mode/index.html
thetaPC: Added a todo: https://github.com/ionic-team/ionic-framework/pull/30874/changes/69d21ad066331bd3ab8998787bafc6dc5055f3be

core/src/utils/test/playwright/page/utils/set-content.ts
thetaPC: This is desired in my opinion. There will be devs that provide this same structure so I don't see any harm in doing it this way. Thoughts?

core/src/themes/base/high-contrast.tokens.ts
thetaPC: I was following the same [setup](https://github.com/ionic-team/ionic-framework/blob/69d21ad066331bd3ab8998787bafc6dc5055f3be/core/src/themes/base/light.tokens.ts#L26-L32) as the light tokens since we don't have subtle defined for high contrast. Do you have a suggestion of what it should be?

Also my intention was to verify the subtle colors when we implement high contrast so I simply followed how light had them for now. So we could just leave this as is for now and then circle back.

core/src/themes/ios/high-contrast-dark.tokens.ts
thetaPC: Fixed: https://github.com/ionic-team/ionic-framework/pull/30874/changes/5eb1fb57b25b1e81c1ac3ba12d6f29e082098564

core/scripts/testing/scripts.js
thetaPC: https://github.com/ionic-team/ionic-framework/pull/30874/commits/db405d42967bf703e46edf4e314400545ed861b2

core/src/utils/test/playwright/page/utils/set-content.ts
brandyscarney: Yeah it doesn't hurt we just need to make sure that it isn't required to pass all of them as `never`.

core/src/themes/base/high-contrast.tokens.ts
brandyscarney: Ah okay I wasn't sure where it came from, but that makes sense! It's been awhile since I worked on this so I forgot I did that. 😂 I am good with revisiting the actual values at a future time and following the way light works for now.

core/src/themes/ios/high-contrast-dark.tokens.ts
brandyscarney: Thinking about this more, I wonder if we need the base one defined at all? Should we just be defining them in the specific themes? We can always talk this through later too.

thetaPC: Hmm I would say that we should wait to see how the [gray tokens PR](https://github.com/ionic-team/ionic-framework/pull/30799) ends since that PR ends up removing `backgroundColorStep`.

brandyscarney: Sounds good!

core/src/utils/test/playwright/page/utils/set-content.ts
thetaPC: Verified that we don't to pass them all.

PR #30785, opened 2025-11-14
feat(textarea): convert to a form associated shadow component

brandyscarney:
Issue number: internal

---------

## What is the current behavior?
Textarea uses `scoped` encapsulation. This causes issues with CSP compatibility and is inconsistent with our goal of having all components use Shadow DOM.

## What is the new behavior?
- Converts `ion-textarea` to `shadow` with `formAssociated: true`
- Adds shadow parts for inner elements
- Adds and updates existing e2e tests in core for textarea
- Updated Angular test app to target textarea shadowRoot and updated lazy forms test to include textarea (standalone already has these)
- Updated React & Vue test apps to target textarea shadowRoot and added validation tests
- Improves focus behavior inside of a popover so that it is no longer required to tab twice to get to the textarea in any browser

## Does this introduce a breaking change?
- [x] Yes
- [ ] No

```
BREAKING CHANGE:

Textarea has been converted to use [Shadow DOM](https://developer.mozilla.org/en-US/docs/Web/Web_Components/Using_shadow_DOM).

If you were targeting the internals of `ion-textarea` in your CSS, you will need to target the `container`, `label`, `native`, `supporting-text`, `helper-text`, `error-text`, `counter`, or `bottom` [Shadow Parts](https://ionicframework.com/docs/theming/css-shadow-parts) instead, or use the provided CSS Variables.
```

## Other information

- [Form Preview](https://ionic-framework-git-fw-6912-textarea-ionic1.vercel.app/src/components/textarea/test/form)
- [Validation Preview](https://ionic-framework-git-fw-6912-textarea-ionic1.vercel.app/src/components/textarea/test/validation)
- [Popover Preview](https://ionic-framework-git-fw-6912-textarea-ionic1.vercel.app/src/components/popover/test/basic) (focus trap testing)

Dev build: `8.7.13-dev.11765925297.1bb45463`

PR #30884, opened 2025-12-19
feat(input-otp): convert to a form associated shadow component #30785

brandyscarney:
Issue number: internal

---------

Input Otp uses `scoped` encapsulation. This causes issues with CSP compatibility and is inconsistent with our goal of having all components use Shadow DOM.

## What is the new behavior?
- Converts `ion-input-otp` to `shadow` with `formAssociated: true`
- Adds shadow parts for inner elements

- [x] Yes
- [ ] No

```
BREAKING CHANGE:

Input Otp has been converted to use [Shadow DOM](https://developer.mozilla.org/en-US/docs/Web/Web_Components/Using_shadow_DOM).

If you were targeting the internals of `ion-input-otp` in your CSS, you will need to target the `group`, `container`, `native`, `separator` or `description` [Shadow Parts](https://ionicframework.com/docs/theming/css-shadow-parts) instead, or use the provided CSS Variables.
```

## Other information

- [Basic Preview](https://ionic-framework-git-fw-6912-input-otp-ionic1.vercel.app/src/components/input-otp/test/basic)
- [Form Preview](https://ionic-framework-git-fw-6912-input-otp-ionic1.vercel.app/src/components/input-otp/test/form)


core/src/components/input-otp/test/form/input-otp.e2e.ts
```diff
@@ -0,0 +1,148 @@
...
+        }
+        const formData = new FormData(form);
+        const entries: Record<string, string> = {};
+        for (const [key, value] of formData.entries()) {
+          entries[key] = value.toString();
+        }
+        return entries;
+      });
+      expect(formData).toBeDefined();
+      expect(formData?.['otp']).toBe('1234');
```
thetaPC: Do we need these checks when it's already being done through `should set formData when submit button is clicked`?

brandyscarney: Probably not - I just added it so that there would never be a false positive on this test where it passes because the value is never set, but it should be caught by the first test so I can remove it.

brandyscarney: Removed: https://github.com/ionic-team/ionic-framework/pull/30884/changes/eb323d77216b032636e989d08d1dfdfd1f5cbde0

core/src/components/input-otp/input-otp.tsx
```diff
@@ -356,6 +380,42 @@ export class InputOTP implements ComponentInterface {
...
+   */
+  private getValue(): string {
+    return this.value != null ? this.value.toString() : '';
+  }
+
+  /**
+   * Called when the form is reset.
+   * Resets the component's value.
+   */
+  formResetCallback() {
```
ShaneK, line 403: Nice! Should we also add support for [`formStateRestoreCallback`](https://stenciljs.com/docs/form-associated#formstaterestorecallback) for when the browser populates the form?

core/src/components/input-otp/input-otp.tsx
```diff
@@ -356,6 +380,42 @@ export class InputOTP implements ComponentInterface {
...
+    this.value = '';
+  }
+
+  /**
+   * Updates the form value and reports validity state to the browser via
+   * ElementInternals. This should be called when the component loads, when
+   * the required prop changes, when the disabled prop changes, and when the value
```
ShaneK: As far as I can tell there is no required prop? I'm assuming this was an accident, but I could also be missing something!

brandyscarney: You're right! I started adding the `required` prop to match textarea, but I removed it when it got too complicated and I realized we should do a design doc first to determine how it should work. 

https://github.com/ionic-team/ionic-framework/pull/30884/commits/8ea32da91521a590a4adf8de027ee5cd15732788

PR #30889, opened 2025-12-23
fix(modal, popover): respect safe area insets on popovers and modals

### LINKED ISSUE #28411: feat: make it possible to add a safe-area for select-popovers

ShaneK:
Issue number: resolves #28411

---------

When a modal is displayed on tablet-sized screens (>= 768px × >= 600px), the `--ion-safe-area-*` CSS variables are explicitly set to 0px. This was intended for inset modals that don't touch screen edges, but it breaks safe area handling on newer iPads with Face ID/home indicators, causing content to overlap with system UI elements.

## What is the new behavior?
 Modals now dynamically handle safe-area insets based on their type and position. This has to be done because modals that don't touch the edges cannot have a safe area applied (because it will add unnecessary padding), but modals that do touch the edges need to apply safe area correctly or the edges will be obstructed by whatever is in the safe area.

- [ ] Yes
- [X] No

## Other information

[Test page](https://ionic-framework-git-fw-6830-ionic1.vercel.app/src/components/modal/test/safe-area/index.html)

Current dev build:
```
 8.7.17-dev.11768239180.18ee1069
```


core/src/components/modal/test/custom/modal.e2e.ts-snapshots/modal-custom-present-ios-ltr-Mobile-Chrome-linux.png
brandyscarney, line 1: There is a weird shadow here where the original modal was:

Image

I verified it happens [on the test](https://ionic-framework-git-fw-6830-ionic1.vercel.app/src/components/modal/test/custom?ionic:mode=ios) too (`ios` mode only). Also, why was this shifted down?

core/src/components/modal/test/custom/modal.e2e.ts-snapshots/modal-custom-present-md-ltr-Mobile-Chrome-linux.png
brandyscarney, line 1: It looks like this style (specifically `box-sizing`) on `div role="dialog"` is causing the height shift:

Image

Is this expected?

core/src/components/modal/test/safe-area/index.html
```diff
@@ -0,0 +1,263 @@
...
+    <style>
+      /**
+       * Simulate safe-area insets for testing.
+       * These values represent typical iPad safe areas.
+       */
+      :root {
+        --ion-safe-area-top: 44px;
+        --ion-safe-area-bottom: 34px;
+        --ion-safe-area-left: 0px;
+        --ion-safe-area-right: 0px;
```
brandyscarney: Can we add some tests for the left & right values? Just to make sure it is shifting for these properly as well.

core/src/components/popover/test/safe-area/index.html
```diff
@@ -0,0 +1,159 @@
...
+    <style>
+      /**
+       * Simulate safe-area insets for testing.
+       * These values represent typical Android edge-to-edge safe areas.
+       */
+      :root {
+        --ion-safe-area-top: 44px;
+        --ion-safe-area-bottom: 34px;
+        --ion-safe-area-left: 0px;
+        --ion-safe-area-right: 0px;
```
brandyscarney: Can we add some tests for the left & right values? Just to make sure it is shifting for these properly as well.

PR #30891, opened 2025-12-26
fix(segment-button): correct color for button childs

JoaoFerreira-FrontEnd:
Issue number: internal

## What is the new behavior?

- guarantee that every child of segment-button has color: primary-color when checked;

- [ ] Yes
- [x] No

Sample: https://ionic-framework-git-rou-12439-ionic1.vercel.app/src/components/segment/test/basic?ionic:theme=ionic

### DIFF
--- a/core/src/components/segment-button/segment-button.ionic.scss
+++ b/core/src/components/segment-button/segment-button.ionic.scss
@@ -64,19 +64,18 @@
 }
 
 // Segment Button: Checked
+// All slot content should have the checked color even when it's not label or icon
 // --------------------------------------------------
 
-:host(.segment-button-checked) ::slotted(ion-label),
-:host(.segment-button-checked) ::slotted(ion-icon),
 :host(.segment-button-checked) ::slotted(*) {
   color: var(--color-checked);
 }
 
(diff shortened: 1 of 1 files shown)


core/src/components/segment-button/segment-button.ionic.scss
ShaneK: You should apply the same changes to disabled, so selected and disabled work the same.

core/src/components/segment-button/segment-button.ionic.scss
```diff
@@ -67,7 +67,8 @@
 // --------------------------------------------------
 
 :host(.segment-button-checked) ::slotted(ion-label),
-:host(.segment-button-checked) ::slotted(ion-icon) {
+:host(.segment-button-checked) ::slotted(ion-icon),
+:host(.segment-button-checked) ::slotted(*) {
```
ShaneK: You should remove the redundant selectors, since the star selector will select ion-label and ion-icon already. You should probably document what you're trying to select here, though.

ShaneK (changes requested):

PR #30895, opened 2025-12-29
fix(input): prevent Android TalkBack from focusing label separately

ShaneK:

---------

When using `ion-input` with a label on Android, TalkBack treats the visual label text as a separate focusable element. This causes the initial focus to land on the label instead of the input field, creating a confusing experience for screen reader users.

## What is the new behavior?
The label text wrapper is now hidden from the accessibility tree via `aria-hidden="true"`, while the native input maintains proper labeling through `aria-labelledby`. This ensures Android TalkBack focuses directly on the input field while still announcing the label correctly.

- [ ] Yes
- [X] No

## Other information

Current dev build:
```
8.7.16-dev.11767032989.1ae720d0
```

### DIFF
--- a/core/src/components/input/input.tsx
+++ b/core/src/components/input/input.tsx
@@ -48,6 +48,7 @@ export class Input implements ComponentInterface {
   private inputId = `ion-input-${inputIds++}`;
   private helperTextId = `${this.inputId}-helper-text`;
   private errorTextId = `${this.inputId}-error-text`;
+  private labelTextId = `${this.inputId}-label`;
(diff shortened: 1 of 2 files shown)


ShaneK:
@thetaPC 
> Where are we getting the expectation for this behavior? When I tested the native inputs from [W3 examples](https://www.w3.org/WAI/tutorials/forms/labels/) to verify that we match, but the screen reader is reading the labels first. Based on this, I would expect that the labels are focused first then the input.

AAA had an ADA audit that flagged this behavior as unexpected

PR #30899, opened 2025-12-30
feat(angular): add custom injector support for modal and popover controllers

### LINKED ISSUE #30638: bug: Seems to be no way to pass an Angular lazy loaded components providers to another component via ModalController
### Prerequisites

### Ionic Framework Version

v8.x

### Current Behavior

When using Modules, I could open another component in my lazy module using the ModalController, and it would have all providers from the lazy component available (as well as all root components).

However, now, I now get the error

Image

There in a minimal example [here](https://stackblitz.com/~/github.com/pjc2007/modal-from-lazy-component)

If you open feature 1, then click the "Open Popup button, and look in the devtools console, we will see the error I am getting in my application

Image

I can't find any way to "pass" the providers to the popup when using the controller.

I did put the providers in a exported variable, and tried to add this the popup component's providers, but you will then see a differrent error..

`@Component({
  selector: 'app-popup1',
  templateUrl: './popup1.component.html',
  styleUrls: ['./popup1.component.scss'],

  // Does not work
  //providers: feature1Providers,
})`

### Expected Behavior

To be able to open a popup in  lazy leaded feature module in s standalone project as I could previously in a modules based

### Steps to Reproduce

1. Goto https://stackblitz.com/~/github.com/pjc2007/modal-from-lazy-component
2. Click the GO TO FEATUE 1 button
3. Open devtools so can see any errors
4. Click the OPEN popup button, and will see the error
  ` .webcontainer@runtime.96435430.js:26 Error opening popup: ɵNotFound: NG0201: No provider found for `_Feature1Service1`. Source: Environment`

### Code Reproduction URL

https://stackblitz.com/~/github.com/pjc2007/modal-from-lazy-component

### Ionic Info

$ ionic info

       Require stack:
       - C:\Users\Peter\AppData\Roaming\npm\node_modules\@ionic\cli\lib\project\index.js
       - C:\Users\Peter\AppData\Roaming\npm\node_modules\@ionic\cli\lib\index.js
       - C:\Users\Peter\AppData\Roaming\npm\node_modules\@ionic\cli\index.js
       - C:\Users\Peter\AppData\Roaming\npm\node_modules\@ionic\cli\bin\ionic

       Require stack:
       - C:\Users\Peter\AppData\Roaming\npm\node_modules\@ionic\cli\lib\project\index.js
       - C:\Users\Peter\AppData\Roaming\npm\node_modules\@ionic\cli\lib\index.js
       - C:\Users\Peter\AppData\Roaming\npm\node_modules\@ionic\cli\index.js
       - C:\Users\Peter\AppData\Roaming\npm\node_modules\@ionic\cli\bin\ionic

Ionic:

   Ionic CLI                     : 7.2.1 (C:\Users\Peter\AppData\Roaming\npm\node_modules\@ionic\cli)
   Ionic Framework               : @ionic/angular 8.7.2
   @angular-devkit/build-angular : 20.1.6
   @angular-devkit/schematics    : 20.1.6
   @angular/cli                  : 20.1.6

Capacitor:

   Capacitor CLI      : 7.4.2
   @capacitor/core    : 7.4.2

Utility:

   native-run  : 2.0.1

System:

   NodeJS : v22.17.1 (C:\Program Files\nodejs\node.exe)
   npm    : 10.9.2
   OS     : Windows 10

(I am actually on Windows 11, not 10)

### Additional Information

https://stackoverflow.com/questions/79736250/how-to-give-a-lazy-loaded-angular-standalone-components-providers-to-another-co

https://forum.ionicframework.com/t/how-to-provide-di-services-in-a-angular-stand-alone-component-that-is-created-via-a-lazy-loaded-component-that-uses-modelcontroller-create/248992

ShaneK:
Issue number: resolves #30638

---------

When using `ModalController.create()` or `PopoverController.create()` in Angular, components rendered inside overlays cannot access non-global services or tokens from the component tree. For example, route-scoped services or Angular's Dir directive for bidirectional text support are not accessible from within a modal, requiring complex workarounds with wrapper components.

## What is the new behavior?
`ModalController.create()` and `PopoverController.create()` now accept an optional injector property that allows passing a custom Angular Injector. This enables overlay components to access services and tokens that are not available in the root injector, such as route-scoped services or the Dir directive from Angular CDK.

```typescript
const customInjector = Injector.create({
  providers: [{ provide: MyService, useValue: myServiceInstance }],
  parent: this.injector,
});
```

```typescript
const modal = await this.modalController.create({
  component: MyModalComponent,
  injector: customInjector,
});
```

- [ ] Yes
- [X] No

## Other information

Current dev build:
```
8.7.17-dev.11769628168.11eca7cd
```

### DIFF
--- a/core/api.txt
+++ b/core/api.txt
@@ -569,6 +569,7 @@ ion-datetime,part,calendar-day today
 ion-datetime,part,month-year-button
 ion-datetime,part,time-button
 ion-datetime,part,time-button active
+ion-datetime,part,wheel
 ion-datetime,part,wheel-item
 ion-datetime,part,wheel-item active
 
--- a/core/src/components/datetime/datetime.tsx
+++ b/core/src/components/datetime/datetime.tsx
@@ -79,6 +79,7 @@ import { checkForPresentationFormatMismatch, warnIfTimeZoneProvided } from './ut
  * @slot buttons - The buttons in the datetime.
  * @slot time-label - The label for the time selector in the datetime.
  *
+ * @part wheel - The wheel container when using a wheel style layout, or in the month/year picker when using a grid style layout.
  * @part wheel-item - The individual items when using a wheel style layout, or in the
  * month/year picker when using a grid style layout.
  * @part wheel-item active - The currently selected wheel-item.
@@ -1724,6 +1725,7 @@ export class Datetime implements ComponentInterface {
 
     return (
       <ion-picker-column
+        part={WHEEL_PART}
         aria-label="Select a date"
         class="date-column"
         color={this.color}
@@ -1844,6 +1846,7 @@ export class Datetime implements ComponentInterface {
 
     return (
       <ion-picker-column
+        part={WHEEL_PART}
         aria-label="Select a day"
         class="day-column"
         color={this.color}
@@ -1888,6 +1891,7 @@ export class Datetime implements ComponentInterface {
 
     return (
       <ion-picker-column
+        part={WHEEL_PART}
         aria-label="Select a month"
         class="month-column"
         color={this.color}
@@ -1931,6 +1935,7 @@ export class Datetime implements ComponentInterface {
 
     return (
       <ion-picker-column
+        part={WHEEL_PART}
         aria-label="Select a year"
         class="year-column"
         color={this.color}
@@ -2005,6 +2010,7 @@ export class Datetime implements ComponentInterface {
 
     return (
       <ion-picker-column
+        part={WHEEL_PART}
         aria-label="Select an hour"
         color={this.color}
         disabled={disabled}
@@ -2045,6 +2051,7 @@ export class Datetime implements ComponentInterface {
 
     return (
       <ion-picker-column
+        part={WHEEL_PART}
         aria-label="Select a minute"
         color={this.color}
         disabled={disabled}
@@ -2088,6 +2095,7 @@ export class Datetime implements ComponentInterface {
 
     return (
       <ion-picker-column
+        part={WHEEL_PART}
         aria-label="Select a day period"
         style={isDayPeriodRTL ? { order: '-1' } : {}}
         color={this.color}
@@ -2716,5 +2724,6 @@ export class Datetime implements ComponentInterface {
 let datetimeIds = 0;
 const CANCEL_ROLE = 'datetime-cancel';
 const CONFIRM_ROLE = 'datetime-confirm';
+const WHEEL_PART = 'wheel';
 const WHEEL_ITEM_PART = 'wheel-item';
 const WHEEL_ITEM_ACTIVE_PART = `active`;
--- a/core/src/components/datetime/test/custom/datetime.e2e.ts
+++ b/core/src/components/datetime/test/custom/datetime.e2e.ts
@@ -42,6 +42,140 @@ configs({ directions: ['ltr'] }).forEach(({ title, screenshot, config }) => {
       await expect(datetime).toHaveScreenshot(screenshot(`datetime-custom-calendar-days`));
     });
   });
+
+  test.describe(title('CSS shadow parts'), () => {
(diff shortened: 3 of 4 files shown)


packages/angular/common/src/types/overlay-options.ts
thetaPC, line 1: I'm not sure if using `Angular` as the prefix is consistent with how the other types are set up. Wouldn't it make more sense to call it `IonicModalOptions` and `IonicPopoverOptions`?

ShaneK: I'm not sure, honestly. We're kind of all over the place with our naming. For me, it makes sense that package-specific options would be named after our packages, but we have names kinda everywhere. Like:
- `IonicVueRouterOptions` does Ionic{Package}{Thing}
- Angular has `IonicGlobal` and `IonicWindow` which is just Ionic{Thing}
- Angular also has `NavigationOptions`, which doesn't have Ionic at all or Angular
- Angular also also has `AngularDelegate` and `AngularFrameworkDelegate` which is in line with this naming
- React has `IonicReactProps`, which is Ionic{Package}{Thing} and it also has `HookOverlayOptions` which has neither Ionic nor React.

Probably the most widely used of these is `IonicVueRouterOptions`, and I could see us going for example `IonicAngularModalOptions`. What do you think?

thetaPC: I lean towards using `IonicAngularModalOptions`... but it also feels a bit too longer. @brandyscarney you have a knack on naming, thoughts?

Also would be great if we can create a ticket to try to sync the naming especially since Ionic React router is being worked on.

brandyscarney: I don't think we should use a prefixed name at all. In my opinion, the import path already provides that information:

```ts
import { ModalOptions } from '@ionic/angular/standalone';
```

If you're importing from `@ionic/angular` it's already clear that this is the Ionic + Angular version. I understand that this could conflict if you ever wanted to import from `@ionic/core` but I don't see why anyone would. Since these are just extensions of what `@ionic/core` exports I'd think you'd always want that. What if we did this where we define the types in Angular:

```ts
import type { Injector } from '@angular/core';
import type { ModalOptions as CoreModalOptions, PopoverOptions as CorePopoverOptions } from '@ionic/core/components';

/**
 * Modal options with Angular-specific injector support.
 * Extends @ionic/core ModalOptions with an optional injector property.
 */
export type ModalOptions = CoreModalOptions & {
  injector?: Injector;
};

/**
 * Popover options with Angular-specific injector support.
 * Extends @ionic/core PopoverOptions with an optional injector property.
 */
export type PopoverOptions = CorePopoverOptions & {
  injector?: Injector;
};
```

Then the users would automatically get the version that includes `Injector` without having to change their existing Angular import.

thetaPC: This works for me!

ShaneK: Done here: [641f319](https://github.com/ionic-team/ionic-framework/pull/30899/commits/641f3194ad040bb7730243f645dcc5f3cd886249)

packages/angular/common/src/types/overlay-options.ts
```diff
@@ -0,0 +1,18 @@
+import type { Injector } from '@angular/core';
+import type { ModalOptions as CoreModalOptions, PopoverOptions as CorePopoverOptions } from '@ionic/core/components';
+
+/**
+ * Modal options with Angular-specific injector support.
+ * Extends @ionic/core ModalOptions with an optional injector property.
+ */
+export type ModalOptions = CoreModalOptions & {
+  injector?: Injector;
+};
```
Ionaru, line 10: @ShaneK This change did introduce a breaking change by re-exporting `ModalOptions` with different behaviour: It no longer accepts a Type argument.
It should be 

```ts
export type ModalOptions<T extends ComponentRef = ComponentRef> = CoreModalOptions<T> & ...
```

PR #30906, opened 2026-01-05
fix(tab-bar): prevent keyboard controller memory leak on rapid mount/unmount

ShaneK:

---------

When `ion-tab-bar` is rapidly mounted and unmounted, a race condition in connectedCallback can cause the keyboard controller to be created after the component has been disconnected. This results in orphaned event listeners (`keyboardWillShow`, `keyboardWillHide`) on the window object that are never cleaned up, causing a memory leak.

## What is the new behavior?
The keyboard controller is now properly destroyed in all scenarios:
- If the component is disconnected while createKeyboardController is pending, the promise is tracked and destroyed when it resolves
- If a new connectedCallback runs before the previous async completes, the stale controller is destroyed

The promise tracking pattern ensures only the most recent async operation assigns its result

- [ ] Yes
- [X] No

## Other information

Current dev build:
```
8.7.17-dev.11767895575.16ea7cef
```

I was unable to find a way to create tests that accurately identified if this problem was occurring. Memory leaks are notoriously difficult to created automated tests for. I ultimately removed my previous attempts because I didn't want to give a false sense of security.

### DIFF
--- a/core/src/components/footer/footer.tsx
+++ b/core/src/components/footer/footer.tsx
@@ -22,6 +22,7 @@ export class Footer implements ComponentInterface {
   private scrollEl?: HTMLElement;
   private contentScrollCallback?: () => void;
   private keyboardCtrl: KeyboardController | null = null;
+  private keyboardCtrlPromise: Promise<KeyboardController> | null = null;
 
   @State() private keyboardVisible = false;
 
@@ -52,7 +53,7 @@ export class Footer implements ComponentInterface {
   }
 
   async connectedCallback() {
-    this.keyboardCtrl = await createKeyboardController(async (keyboardOpen, waitForResize) => {
+    const promise = createKeyboardController(async (keyboardOpen, waitForResize) => {
       /**
        * If the keyboard is hiding, then we need to wait
        * for the webview to resize. Otherwise, the footer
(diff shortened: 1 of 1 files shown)


core/src/components/tab-bar/tab-bar.tsx
brandyscarney, line 1: Do we need to make these changes to [footer](https://github.com/ionic-team/ionic-framework/blob/3b3318da513b199128f3822bd8226797cd118b0f/core/src/components/footer/footer.tsx#L54-L73) also?

core/src/components/tab-bar/tab-bar.tsx
```diff
@@ -100,11 +101,32 @@ export class TabBar implements ComponentInterface {
...
+    if (this.keyboardCtrlPromise === promise) {
+      this.keyboardCtrl = keyboardCtrl;
+      this.keyboardCtrlPromise = null;
+    } else {
+      keyboardCtrl.destroy();
+    }
   }
 
   disconnectedCallback() {
+    if (this.keyboardCtrlPromise) {
```
brandyscarney, line 122: Maybe this doesn't need to be done on this PR, but I feel like we do this same code in so many places that we should consider making it a utility.

core/src/components/tab-bar/tab-bar.tsx
ShaneK: Yep! Good call [4d19d31](https://github.com/ionic-team/ionic-framework/pull/30906/commits/4d19d31aa8b92b47ffc987861d14b9506104fa35)

core/src/components/tab-bar/tab-bar.tsx
ShaneK: It looks like this particular pattern is only specifically done in tab bar and footer (unless I'm missing something?), so I'm not sure if that's worth extracting right now

brandyscarney: I mostly meant the code that is destroying like this:

https://github.com/ionic-team/ionic-framework/blob/1c89cf06ac959f9c9a35a66f811227c244d3198b/core/src/components/select/select.tsx#L376-L379

We also disconnect in a lot of places:

https://github.com/ionic-team/ionic-framework/blob/1c89cf06ac959f9c9a35a66f811227c244d3198b/core/src/components/select/select.tsx#L382-L385

It seems like we could make this into a utility or something, but that's not an issue for your PR.

thetaPC: Let's make a ticket to circle back on adding the utility.

ShaneK: @thetaPC Done! This will be FW-7030

PR #30926, opened 2026-01-19
feat(content): add content-fullscreen class when fullscreen is true

brandyscarney:
Issue number: N/A

---------

Content does not reflect the `fullscreen` property or add a class when it is enabled, making it harder to style.

## What is the new behavior?
- Adds `content-fullscreen` class to content when `fullscreen` is true
- Adds an e2e test verifying the class is applied

- [ ] Yes
- [x] No

## Other information

[Preview](https://ionic-framework-git-fw-6792-ionic1.vercel.app/src/components/content/test/fullscreen/)

### DIFF
--- a/core/src/components/range/range.tsx
+++ b/core/src/components/range/range.tsx
@@ -883,7 +883,7 @@ export class Range implements ComponentInterface {
   }
 
   render() {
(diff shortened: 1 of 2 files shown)


thetaPC:
@brandyscarney I realized this after approving. It might be a good idea to add a comment stating that the class is needed for iOS Liquid Glass. I can see us down the line forgetting why it was added and end up removing it.

PR #30927, opened 2026-01-19
feat(item): add inner and container parts

brandyscarney:
Issue number: N/A

---------

The inner structural elements of item are not exposed as shadow parts, preventing users from being able to customize their styles directly.

## What is the new behavior?
- Exposes `inner` and `container` shadow parts
- Adds e2e test coverage for customizing the shadow parts
- Removes css variables test, combining it with the shadow parts test

- [ ] Yes
- [x] No


core/src/components/item/test/css-variables/item.e2e.ts
brandyscarney, line 1: I deleted this test because it doesn't do anything except set CSS variables so I added the same check here: https://github.com/ionic-team/ionic-framework/pull/30927/changes#diff-ede714f4c0e943356ae72d2456093d9e6ccc4e6fc86b465f598231be62a3ae8fR111

core/src/components/item/item.tsx
```diff
@@ -18,6 +18,8 @@ import type { RouterDirection } from '../router/utils/interface';
  * @slot end - Content is placed to the right of the item text in LTR, and to the left in RTL.
  *
  * @part native - The native HTML button, anchor or div element that wraps all child elements.
+ * @part inner - The inner container element that wraps the item content.
```
ShaneK:
```suggestion
 * @part inner - The inner wrapper element that arranges the item content.
```
Nit: Since `container` has actual meaning when referring to parts and it's not what you actually mean here, it might make sense to try to avoid using it when describing parts just to avoid confusion

core/src/components/item/test/custom/item.e2e.ts
```diff
@@ -0,0 +1,174 @@
+import { expect } from '@playwright/test';
+import { configs, test } from '@utils/test/playwright';
+
+/**
+ * This behavior does not vary across modes/directions
+ */
+configs({ directions: ['ltr'], modes: ['md'] }).forEach(({ title, config }) => {
+  test.describe(title('item: custom'), () => {
+    test.describe(title('CSS shadow parts'), () => {
```
ShaneK:
```suggestion
    test.describe('CSS shadow parts', () => {
```

brandyscarney: https://github.com/ionic-team/ionic-framework/pull/30927/changes/8200036ae9b04ff83c88c80a4a4cf42135aeca21

core/src/components/item/item.tsx
brandyscarney: https://github.com/ionic-team/ionic-framework/pull/30927/changes/109420b39fa9dbb99dfe2947c4bdf57432a37003

PR #30929, opened 2026-01-19
feat(item-option): add inner and container parts

brandyscarney:
Issue number: N/A

---------

The inner structural elements of item-option are not exposed as shadow parts, preventing users from being able to customize their styles directly.

## What is the new behavior?
- Exposes `inner` and `container` shadow parts
- Adds e2e test coverage for customizing the shadow parts

- [ ] Yes
- [x] No


core/src/components/item-option/test/custom/item-option.e2e.ts
```diff
@@ -0,0 +1,80 @@
+import { expect } from '@playwright/test';
+import { configs, test } from '@utils/test/playwright';
+
+/**
+ * This behavior does not vary across modes/directions
+ */
+configs({ directions: ['ltr'], modes: ['md'] }).forEach(({ title, config }) => {
+  test.describe(title('item-option: custom'), () => {
+    test.describe(title('CSS shadow parts'), () => {
```
ShaneK:
```suggestion
    test.describe('CSS shadow parts', () => {
```

core/src/components/item-option/item-option.tsx
```diff
@@ -17,6 +17,8 @@ import type { Color } from '../../interface';
  * @slot end - Content is placed to the right of the option text in LTR, and to the left in RTL.
  *
  * @part native - The native HTML button or anchor element that wraps all child elements.
+ * @part inner - The inner container element that wraps the option content.
```
ShaneK:
```suggestion
 * @part inner - The inner wrapper element that arranges the option content.
```

Nit: Avoid confusing terminology since `container` is a part

core/src/components/item-option/test/custom/item-option.e2e.ts
brandyscarney: https://github.com/ionic-team/ionic-framework/pull/30929/changes/097b8bfa2f03b9ceffaf8b3a7bdb5b8673fdcc70

core/src/components/item-option/item-option.tsx
brandyscarney: https://github.com/ionic-team/ionic-framework/pull/30929/changes/c652911bd49641dd4f8933e0ac6ff8021f561164

PR #30930, opened 2026-01-19
feat(list-header): add inner part

brandyscarney:
Issue number: N/A

---------

The inner structural element of list-header is not exposed as a shadow part, preventing users from being able to customize its styles directly.

## What is the new behavior?
- Exposes `inner` shadow part
- Adds e2e test coverage for customizing the shadow part

- [ ] Yes
- [x] No


core/src/components/list-header/list-header.tsx
```diff
@@ -40,7 +42,7 @@ export class ListHeader implements ComponentInterface {
           [`list-header-lines-${lines}`]: lines !== undefined,
         })}
       >
-        <div class="list-header-inner">
+        <div class="list-header-inner" part="inner">
```
thetaPC, line 45: Why would this be considered `inner` instead of `content`?

brandyscarney: Based on our latest discussion this is correct due to https://github.com/ionic-team/ionic-framework/pull/30931

core/src/components/list-header/test/custom/list-header.e2e.ts
```diff
@@ -0,0 +1,34 @@
+import { expect } from '@playwright/test';
+import { configs, test } from '@utils/test/playwright';
+
+/**
+ * This behavior does not vary across modes/directions
+ */
+configs({ directions: ['ltr'], modes: ['md'] }).forEach(({ title, config }) => {
+  test.describe(title('list-header: custom'), () => {
+    test.describe(title('CSS shadow parts'), () => {
```
ShaneK:
```suggestion
    test.describe('CSS shadow parts', () => {
```

core/src/components/list-header/list-header.tsx
```diff
@@ -7,6 +7,8 @@ import type { Color } from '../../interface';
 
 /**
  * @virtualProp {"ios" | "md"} mode - The mode determines which platform styles to use.
+ *
+ * @part inner - The inner container element that wraps the list header content.
```
ShaneK:
```suggestion
 * @part inner - The inner wrapper element that arranges the list header content.
```
Nit: Avoid confusing terminology since container is a part

ShaneK (approved):
Just a couple nits, which you can ignore if you want. Looks good to me!

core/src/components/list-header/test/custom/list-header.e2e.ts
brandyscarney: https://github.com/ionic-team/ionic-framework/pull/30930/changes/2393cc8fc8e81d4c6d8667629058d8358455eac7

core/src/components/list-header/list-header.tsx
brandyscarney: https://github.com/ionic-team/ionic-framework/pull/30930/changes/3f838b2478df3978c19de92f927f306984616bf2

Issue #29862, opened 2024-09-11
feat: Extend ion-range CSS shadowparts for individual knob styling with dualKnobs active


Tminchew:
### Prerequisites

- [X] I have read the [Contributing Guidelines](https://github.com/ionic-team/ionic-framework/blob/main/docs/CONTRIBUTING.md#creating-an-issue).
- [X] I agree to follow the [Code of Conduct](https://ionicframework.com/code-of-conduct).
- [X] I have searched for [existing issues](https://github.com/ionic-team/ionic-framework/issues) that already include this feature request, without success.

### Describe the Feature Request

I would like the ability to style an individual knob on the ion-range slider when dualKnob=true using the CSS shadowparts knob field.

### Describe the Use Case

I want the knob to change color based on areas/values of the range slider. This unfortunately modifies both knobs colors currently.

### Describe Preferred Solution

using CSS shadowparts so it can be referenced like: 
```ion-range::part(knob-a) { <-- CSS Here --> }```
```ion-range::part(knob-b) { <-- CSS Here --> }```

### Describe Alternatives

Any currently known way on how to individually style a knob when dualKnob is true

rdlabo:
Great request — this will be really useful 👍  

One additional thought: it would be great if `ion-range` itself could receive a class when `dual-knobs` is active.  
That would make it easier to detect the dual-knob state directly from the DOM and would unlock even more styling and customization possibilities.

PR #30932, opened 2026-01-20
feat(range): add classes to the range when the value is at the min or max

brandyscarney:
Issue number: N/A

---------

Range adds classes to the knobs at `min` and `max`, but the host element doesn't reflect those states.

## What is the new behavior?
- Adds `range-value-min` and `range-value-max` when the value is at the `min` or `max`, respectively.
- Adds a spec test verifying the classes are applied properly.

- [ ] Yes
- [x] No

## Other information

Dev build: `8.7.17-dev.11768943212.1b316931`

### DIFF
--- a/core/src/components/range/range.tsx
+++ b/core/src/components/range/range.tsx
@@ -883,7 +883,7 @@ export class Range implements ComponentInterface {
   }
 
   render() {
-    const { disabled, el, hasLabel, rangeId, pin, pressedKnob, labelPlacement, label, min, max, dualKnobs } = this;
+    const { disabled, el, hasLabel, rangeId, pin, pressedKnob, labelPlacement, label, dualKnobs, min, max } = this;
 
     const inItem = hostContext('ion-item', el);
 


rdlabo:
Great work — thank you for this improvement!

I noticed one potential limitation: since `dual-knobs` is not reflected in the DOM, using `[dualKnobs]="true"` makes this approach difficult to use in practice. We may want to think about how `dual-knobs` should be handled.

Two possible options could be:
1. Also set a `.use-dual-knobs` class (this would be the quickest solution).
2. When `dual-knobs` is enabled, provide `part="knob knob-a"` / `part="knob knob-b"`.

ex:

```scss
  &:not(.dual-knobs).range-pressed {
    &.range-value-min::part(knob) {}
    &.range-value-max::part(knob) {}
  }

  &.dual-knobs.range-pressed {
    &.range-value-min::part(knob-a) {}
    &.range-value-max::part(knob-b) {}
  }
```

Either of these would greatly expand how this can be leveraged.

brandyscarney:
Thank you for the feedback @rdlabo! I am actually working on adding `knob-a` and `knob-b` parts this week to resolve #29862 and it will also go in the `v8.8` release. I will take a look at adding a `dual-knobs` class too though either on this PR or the one for that feature. I may reach out to you on Discord with a question on how the knobs should act when they switch places. 🙂

rdlabo:
This is a fantastic improvement — it really expands the customization possibilities a lot. 🙌  
Let’s discuss dual-knobs and how the knobs should behave separately!

PR #30946, opened 2026-02-05
feat(refresher): add ionPullStart and ionPullEnd events

### LINKED ISSUE #24524: feat: add ionEnd event to refresher
### Prerequisites

I would like the refresher to have an `ionEnd` event.

I need to know when the refresher has returned to the inactive state to do some jiggery-pokery with my backgrounds. But watching the refresher progress isn’t good enough because hitting zero doesn’t necessarily mean the user has completed the pull gesture. So, I need to know both when the user is no longer touching the screen AND when the refresher is ready to be pulled again.

I'll have a PR to look at shortly.

### Describe the Use Case

I have a background that bleeds from the header to the content (it looks like one continuous image before you scroll). When I scroll the content up, the bg that's in the content should scroll up under the header along with the content.

But when I pull down to refresh, I don't want the background in the content to separate from the header and leave a gap. I just want the content to move down.

I have a few solutions for this. One of them involves having two bgs in different places and swapping their visibility based on when the user is interacting with the refresher.

Watching the progress doesn't quite work, while having an `ionEnd` event should.

_No response_

### Describe Alternatives

_No response_

### Related Code

The issue can be seen in the "BG Swap" page of this Ionic application.

https://github.com/WhatsThatItsPat/pull-to-refresh-with-bg

### Additional Information

_No response_

ShaneK:
Issue number: resolves #24524

---------

There is no way to know when the refresher has fully returned to its inactive state after a pull gesture. The existing `ionStart` event fires when pulling begins, but there is no corresponding end event. Watching the progress property is insufficient because hitting zero doesn’t necessarily mean the user has completed the pull gesture.

## What is the new behavior?

Two new events are added to the refresher component:
- `ionPullStart`: Emitted when the user begins pulling down (same as `ionStart`, which is now deprecated)
- `ionPullEnd`: Emitted when the refresher returns to inactive state, with a `reason` property of `'complete'` or `'cancel'` indicating whether the refresh operation completed successfully or was cancelled

This allows you to know both when the user is no longer touching the screen AND when the refresher is ready to be pulled again.

- [ ] Yes
- [X] No

## Other information

Test page:
https://ionic-framework-git-fw-6591-ionic1.vercel.app/src/components/refresher/test/basic/index.html

Current dev build:
```
8.7.17-dev.11770319814.172b4f50
```

### DIFF
--- a/core/api.txt
+++ b/core/api.txt
@@ -1695,6 +1695,7 @@ ion-segment-content,shadow
 
 ion-segment-view,shadow
 ion-segment-view,prop,disabled,boolean,false,false,false
+ion-segment-view,prop,swipeGesture,boolean,true,false,false
 ion-segment-view,event,ionSegmentViewScroll,SegmentViewScrollEvent,true
 
 ion-select,shadow
@@ -1759,14 +1760,17 @@ ion-select,css-prop,--placeholder-opacity,ios
 ion-select,css-prop,--placeholder-opacity,md
 ion-select,css-prop,--ripple-color,ios
 ion-select,css-prop,--ripple-color,md
+ion-select,part,bottom
 ion-select,part,container
 ion-select,part,error-text
 ion-select,part,helper-text
 ion-select,part,icon
+ion-select,part,inner
 ion-select,part,label
 ion-select,part,placeholder
 ion-select,part,supporting-text
 ion-select,part,text
+ion-select,part,wrapper
 
 ion-select-modal,scoped
 ion-select-modal,prop,cancelText,string,'Close',false,false
@@ -1877,7 +1881,7 @@ ion-textarea,prop,cols,number | undefined,undefined,false,true
 ion-textarea,prop,counter,boolean,false,false,false
 ion-textarea,prop,counterFormatter,((inputLength: number, maxLength: number) => string) | undefined,undefined,false,false
 ion-textarea,prop,debounce,number | undefined,undefined,false,false
-ion-textarea,prop,disabled,boolean,false,false,false
+ion-textarea,prop,disabled,boolean,false,false,true
 ion-textarea,prop,enterkeyhint,"done" | "enter" | "go" | "next" | "previous" | "search" | "send" | undefined,undefined,false,false
 ion-textarea,prop,errorText,string | undefined,undefined,false,false
 ion-textarea,prop,fill,"outline" | "solid" | undefined,undefined,false,false
@@ -1890,7 +1894,7 @@ ion-textarea,prop,minlength,number | undefined,undefined,false,false
 ion-textarea,prop,mode,"ios" | "md",undefined,false,false
 ion-textarea,prop,name,string,this.inputId,false,false
 ion-textarea,prop,placeholder,string | undefined,undefined,false,false
-ion-textarea,prop,readonly,boolean,false,false,false
+ion-textarea,prop,readonly,boolean,false,false,true
 ion-textarea,prop,required,boolean,false,false,false
 ion-textarea,prop,rows,number | undefined,undefined,false,false
 ion-textarea,prop,shape,"round" | undefined,undefined,false,false
--- a/core/src/components.d.ts
+++ b/core/src/components.d.ts
@@ -3113,6 +3113,11 @@ export namespace Components {
           * @param smoothScroll : Whether to animate the scroll transition.
          */
         "setContent": (id: string, smoothScroll?: boolean) => Promise<void>;
+        /**
+          * If `true`, users will be able to swipe the segment view to navigate between segment contents.
+          * @default true
+         */
+        "swipeGesture": boolean;
     }
     interface IonSelect {
         /**
@@ -8435,6 +8440,11 @@ declare namespace LocalJSX {
           * Emitted when the segment view is scrolled.
          */
         "onIonSegmentViewScroll"?: (event: IonSegmentViewCustomEvent<SegmentViewScrollEvent>) => void;
+        /**
+          * If `true`, users will be able to swipe the segment view to navigate between segment contents.
+          * @default true
+         */
+        "swipeGesture"?: boolean;
     }
     interface IonSelect {
         /**
--- a/core/src/components/segment-view/segment-view.scss
+++ b/core/src/components/segment-view/segment-view.scss
@@ -21,7 +21,8 @@
   display: none;
 }
 
-:host(.segment-view-disabled) {
+:host(.segment-view-disabled),
+:host(.segment-view-swipe-disabled) {
   touch-action: none;
   overflow-x: hidden;
 }
--- a/core/src/components/segment-view/segment-view.tsx
+++ b/core/src/components/segment-view/segment-view.tsx
@@ -24,6 +24,11 @@ export class SegmentView implements ComponentInterface {
   @Prop() disabled = false;
 
   /**
+   * If `true`, users will be able to swipe the segment view to navigate between segment contents.
+   */
+  @Prop() swipeGesture = true;
+
+  /**
    * @internal
    *
    * If `true`, the segment view is scrollable.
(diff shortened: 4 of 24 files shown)


WhatsThatItsPat:
Thanks for the update. I look forward to playing with it.

I have something related on my mind.

On a page without a refresher, you can obviously still pull down and rubber band past the top (`scrollY` is less than 0). If you keep your figure on the screen and hold the content still in that rubber-banded state, `isScrolling` will change to `false` and `ionScrollEnd` will fire (I think), even though you're still in the middle of a scroll. When you continue the gesture or let go, it will go back to `true` until it animates to its normal resting place.

I think you should be still be considered scrolling when you're rubber banding / elastic scrolling, even if you manually pause the scroll. Only when `ion-content` comes to a rest **AND** your gesture is over should the `ionScrollEnd` fire.

Anyway, I haven't thought of that in a year+, but I'm pretty sure that's the issue. This one reminded me of it because of the start and end stuff.

Maybe I'll look at that again in my free time.

core/src/components/refresher/refresher.tsx
```diff
@@ -308,6 +328,7 @@ export class Refresher implements ComponentInterface {
         this.lastVelocityY = ev.velocityY;
       },
       onEnd: () => {
+        const hadStarted = this.didStart;
```
thetaPC, line 331: Why save the value into another variable when `hadStarted` is only used once?

core/src/components/refresher/refresher.tsx
```diff
@@ -182,6 +197,10 @@ export class Refresher implements ComponentInterface {
     this.progress = 0;
 
     this.state = RefresherState.Inactive;
+
+    this.ionPullEnd.emit({
+      reason: state === RefresherState.Completing ? 'complete' : 'cancel',
```
thetaPC, line 202: The events should also be an `enum` like `RefresherState` to minimize issues like misspellings.

core/src/components/refresher/test/basic/refresher.e2e.ts
```diff
@@ -22,6 +22,41 @@ configs({ directions: ['ltr'] }).forEach(({ title, config }) => {
 
         expect(await items.count()).toBe(60);
       });
+
+      test('should emit ionPullStart and ionPullEnd with reason complete', async ({ page }) => {
+        await page.locator('ion-refresher.hydrated').waitFor({ state: 'attached' });
```
thetaPC: Do we need this when it's being done in `pullToRefresh`?

core/src/components/refresher/test/basic/refresher.e2e.ts
```diff
@@ -22,6 +22,41 @@ configs({ directions: ['ltr'] }).forEach(({ title, config }) => {
...
+
+        const ionPullStartEvent = await page.spyOnEvent('ionPullStartFired');
+        const ionPullEndEvent = await page.spyOnEvent('ionPullEndFired');
+
+        await pullToRefresh(page);
+
+        // Wait for the close animation timeout (600ms) to complete
```
thetaPC: It would be better to not include the time when we might change it in `waitForTimeout` and forget to update the comment. Like in this case.

Edit: Seems you actually meant to remove it from the beginning based on the next test. 😆 


```suggestion
        // Wait for the close animation to complete
```

core/src/components/refresher/test/basic/refresher.e2e.ts
```diff
@@ -22,6 +22,41 @@ configs({ directions: ['ltr'] }).forEach(({ title, config }) => {
...
+
+        expect(ionPullStartEvent).toHaveReceivedEventTimes(1);
+        expect(ionPullEndEvent).toHaveReceivedEventTimes(1);
+        expect(ionPullEndEvent).toHaveReceivedEventDetail({ reason: 'complete' });
+      });
+
+      test('should emit ionPullEnd with reason cancel when pull is released early', async ({ page }) => {
+        const target = page.locator('body');
+
+        await page.locator('ion-refresher.hydrated').waitFor({ state: 'attached' });
```
thetaPC: Same [question](https://github.com/ionic-team/ionic-framework/pull/30946/changes#r2783854946) applies here.

core/src/components/refresher/test/basic/refresher.e2e.ts
```diff
@@ -41,6 +76,30 @@ configs({ directions: ['ltr'] }).forEach(({ title, config }) => {
...
+      test('should emit ionPullStart and ionPullEnd with reason complete', async ({ page }) => {
+        const refresherContent = page.locator('ion-refresher-content');
+        refresherContent.evaluateHandle((el: any) => {
+          // Resets the pullingIcon to enable the native refresher
+          el.pullingIcon = undefined;
+        });
+
+        await page.waitForChanges();
+
+        await page.locator('ion-refresher.hydrated').waitFor({ state: 'attached' });
```
thetaPC: Same [question](https://github.com/ionic-team/ionic-framework/pull/30946/changes#r2783854946) applies here.

core/src/components/refresher/refresher.tsx
ShaneK: Because two lines below, we set `this.didStart` to false (as part of the `onEnd` trigger), but we need to know the value as it was at the beginning of the call for the else if branch.

ShaneK: Technically, we could move that set to after all the branches, but this way feels a little cleaner and is probably the same post-compilation anyway.

core/src/components/refresher/refresher.tsx
ShaneK: This is typescript with an union string type, so there cannot be misspellings assuming the consumer uses typescript (if they don't, they won't get benefit either way). Also, are there any examples of us using custom events with an enum in the response interface? I don't think that is an established pattern at this point, using string union types is pretty common though.

core/src/components/refresher/test/basic/refresher.e2e.ts
ShaneK: Yeah, good call!

core/src/components/refresher/test/basic/refresher.e2e.ts
ShaneK: I thought I needed it for the spies to hook on right, but I don't think so. I'll remove, thanks!

core/src/components/refresher/test/basic/refresher.e2e.ts
ShaneK: [ea3b354](https://github.com/ionic-team/ionic-framework/pull/30946/commits/ea3b354dd8e88b8f9fa2c45f11499b9713b1fb88)

core/src/components/refresher/test/basic/refresher.e2e.ts
ShaneK: [ea3b354](https://github.com/ionic-team/ionic-framework/pull/30946/commits/ea3b354dd8e88b8f9fa2c45f11499b9713b1fb88)

core/src/components/refresher/test/basic/refresher.e2e.ts
ShaneK: [ea3b354](https://github.com/ionic-team/ionic-framework/pull/30946/commits/ea3b354dd8e88b8f9fa2c45f11499b9713b1fb88)

core/src/components/refresher/test/basic/refresher.e2e.ts
ShaneK: [ea3b354](https://github.com/ionic-team/ionic-framework/pull/30946/commits/ea3b354dd8e88b8f9fa2c45f11499b9713b1fb88)

core/src/components/refresher/refresher.tsx
thetaPC: Makes sense to me, thanks!

core/src/components/refresher/refresher.tsx
thetaPC: Got it! Then I'm good with it as is.

Issue #30900, opened 2025-12-31
bug: In Android API 36, the safe-area is not respected.


Marius-Romanus:
### Prerequisites

- [x] I have read the [Contributing Guidelines](https://github.com/ionic-team/ionic-framework/blob/main/docs/CONTRIBUTING.md#creating-an-issue).
- [x] I agree to follow the [Code of Conduct](https://ionicframework.com/code-of-conduct).
- [x] I have searched for [existing issues](https://github.com/ionic-team/ionic-framework/issues) that already report this problem, without success.

### Ionic Framework Version

v8.x

### Current Behavior

This is due to switching to Capacitor 8, which uses Android API 36, where `windowOptOutEdgeToEdgeEnforcement` no longer works:

https://capacitorjs.com/docs/apis/status-bar#android-16-behavior-change

Modals and popovers (that I've seen) are not respecting the safe area.

I've implemented the SystemBars plugin with `insetsHandling` in CSS, but it makes no difference:

https://capacitorjs.com/docs/apis/system-bars#android-note

It seems the only place where it's respected is the `top` element, if you have an `ion-header` in CSS:

```
ion-header ion-toolbar:first-of-type {
    padding-top: var(--ion-safe-area-top, 0);
}
```

I've tested it on the following emulators:

Pixel 9 Pro - API 36
Pixel 7 Pro - API 35
Pixel 6a - API 33

<img width="249" height="473" alt="Image" src="https://github.com/user-attachments/assets/57e19d6d-4d4a-46dc-8bd2-0bda15d18dc2" />
<img width="246" height="470" alt="Image" src="https://github.com/user-attachments/assets/f19186ea-28ac-44c6-be6e-cd2f8b3fc172" />

### Expected Behavior

The safe area should always be respected in all cases and at all edges.

### Steps to Reproduce

```
npm i
ionic capacitor build android
```
Enable "Navigation mode: Buttons" if it's set to "Gestures"
See the modal and popover in tab 1.

The global CSS contains:
```
html {
    --ion-safe-area-top: var(--safe-area-inset-top, env(safe-area-inset-top, 0px));
    --ion-safe-area-bottom: var(--safe-area-inset-bottom, env(safe-area-inset-bottom, 0px));
    --ion-safe-area-left: var(--safe-area-inset-left, env(safe-area-inset-left, 0px));
    --ion-safe-area-right: var(--safe-area-inset-right, env(safe-area-inset-right, 0px));
}
```

[...]

PR #30949, opened 2026-02-09
fix(modal, popover): respect safe area insets on popovers and modals

### LINKED ISSUE #28411: feat: make it possible to add a safe-area for select-popovers

ShaneK:
Issue number: resolves #28411, #30900

---------

When a modal is displayed on tablet-sized screens (>= 768px × >= 600px), the `--ion-safe-area-*` CSS variables are explicitly set to 0px. This was intended for inset modals that don't touch screen edges, but it breaks safe area handling on newer iPads with Face ID/home indicators, causing content to overlap with system UI elements.

## What is the new behavior?
 Modals now dynamically handle safe-area insets based on their type and position. This has to be done because modals that don't touch the edges cannot have a safe area applied (because it will add unnecessary padding), but modals that do touch the edges need to apply safe area correctly or the edges will be obstructed by whatever is in the safe area.

- [ ] Yes
- [X] No

## Other information

[Modals test page](https://ionic-framework-git-fw-6830-2-ionic1.vercel.app/src/components/modal/test/safe-area/index.html)
[Popovers test page](https://ionic-framework-git-fw-6830-2-ionic1.vercel.app/src/components/popover/test/safe-area/index.html)

Current dev build:
```
8.7.18-dev.11770674094.18396f54
```

### DIFF
--- a/core/src/components/modal/test/safe-area/index.html
+++ b/core/src/components/modal/test/safe-area/index.html
@@ -111,6 +111,19 @@
               <li>Bottom: 34px (simulates home indicator)</li>
             </ul>
             <div id="modal-safe-area-info"></div>
+            <p>
+            Lorem ipsum dolor sit amet, consectetur adipiscing elit. Suspendisse scelerisque dui vitae massa tincidunt, id volutpat sapien dapibus. Nunc libero ante, tristique vel imperdiet vitae, aliquet a arcu. Proin vestibulum augue neque, posuere posuere nisl mattis vel. Ut interdum quam ut nulla ultrices tristique vitae in arcu. Nulla ac magna tincidunt mauris tincidunt porttitor. Integer tincidunt imperdiet ante. Mauris quam arcu, dictum quis commodo a, tempus at nulla. Curabitur non tellus a leo tempor lobortis.
+            </p>
+            <p>
+            Lorem ipsum dolor sit amet, consectetur adipiscing elit. Suspendisse scelerisque dui vitae massa tincidunt, id volutpat sapien dapibus. Nunc libero ante, tristique vel imperdiet vitae, aliquet a arcu. Proin vestibulum augue neque, posuere posuere nisl mattis vel. Ut interdum quam ut nulla ultrices tristique vitae in arcu. Nulla ac magna tincidunt mauris tincidunt porttitor. Integer tincidunt imperdiet ante. Mauris quam arcu, dictum quis commodo a, tempus at nulla. Curabitur non tellus a leo tempor lobortis.
+            </p>
+            <p>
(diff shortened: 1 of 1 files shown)


core/src/components/modal/modal.tsx
```diff
@@ -1335,6 +1358,114 @@ export class Modal implements ComponentInterface, OverlayInterface {
...
+
+    const context = this.getSafeAreaContext();
+    if (context.isSheetModal || context.isCardModal) return;
+
+    // Check for standard Ionic layout children (ion-content, ion-footer),
+    // searching one level deep for wrapped components (e.g.,
+    // <app-footer><ion-footer>...</ion-footer></app-footer>).
+    let hasContent = false;
+    let hasFooter = false;
+    for (const child of Array.from(el.children)) {
```
brandyscarney, line 1471: Is there a reason we don't use `querySelectorAll` to get all content descendants?

core/src/components/modal/safe-area-utils.ts
```diff
@@ -0,0 +1,151 @@
...
+/**
+ * Configuration for safe-area CSS custom properties.
+ * Values can be 'inherit' (use root value) or '0px' (no safe-area).
+ */
+export interface SafeAreaConfig {
+  top: string;
+  bottom: string;
+  left: string;
+  right: string;
+}
```
brandyscarney, line 15: Why don't we use this instead of `string`?

```suggestion
type SafeAreaValue = "0px" | "inherit";

export interface SafeAreaConfig {
  top: SafeAreaValue;
  bottom: SafeAreaValue;
  left: SafeAreaValue;
  right: SafeAreaValue;
}
```

core/src/components/modal/modal.tsx
ShaneK: Yeah, I only want to do a shallow search. I tried to fix this in [6f216f0](https://github.com/ionic-team/ionic-framework/pull/30949/commits/6f216f0e92bf61f04bbf1f767f487c27b1d5886e) by using `:scope`, but it broke Stencil's mock doc and I had to roll back to the for loop

core/src/components/modal/safe-area-utils.ts
ShaneK: That makes sense, done in [6f216f0](https://github.com/ionic-team/ionic-framework/pull/30949/commits/6f216f0e92bf61f04bbf1f767f487c27b1d5886e)

core/src/components/modal/test/safe-area/index.html
brandyscarney, line 1: Optional: Could we add some Lorem Ipsum content to the "no footer" modal to make it overflow enough to scroll so we can see that the bottom is adjusted still?

ShaneK: Done! [89d177b](https://github.com/ionic-team/ionic-framework/pull/30949/commits/89d177b32377e25daf26d442f9c3076f0f6e70d7)

PR #30951, opened 2026-02-09
feat(select): add wrapper and bottom shadow parts

### LINKED ISSUE #29918: feat(select): add part attribute to ion-input-wrapper to allow more customizations
### Prerequisites

Currently `ion-select` is limit in customizations when using shadows or boxing the input since the width and overflow is set on a wrapping div without a `part`-attribute to access it. It is also different from the `ion-input` HTML structure, maybe because of the "dropdown" but I think it would also work without that wrapping div.

image

### Describe the Use Case

We are trying to make every input to look the same, this works for most Ionic input components but is hard to accomplish with `ion-select`. Box shadow is removed due to overflow and it cannot go full width.

image

Use the same input HTML structure like in the other input components, reduce the wrapping or add a new part attribute to allow more customizations.

### Describe Alternatives

_No response_

### Related Code

_No response_

### Additional Information

_No response_

thetaPC:
Issue number: resolves #29918 

---------

Developers can't customize elements within select like the bottom container because the component is a shadow component.

## What is the new behavior?

- Added `bottom`, `wrapper`,  and `wrapper-inner` parts
- Added a test

- [ ] Yes
- [x] No

## Other information

N/A

### DIFF
--- a/core/api.txt
+++ b/core/api.txt
@@ -1693,6 +1693,7 @@ ion-segment-content,shadow
 
 ion-segment-view,shadow
 ion-segment-view,prop,disabled,boolean,false,false,false
+ion-segment-view,prop,swipeGesture,boolean,true,false,false
 ion-segment-view,event,ionSegmentViewScroll,SegmentViewScrollEvent,true
 
 ion-select,shadow
@@ -1878,7 +1879,7 @@ ion-textarea,prop,cols,number | undefined,undefined,false,true
 ion-textarea,prop,counter,boolean,false,false,false
 ion-textarea,prop,counterFormatter,((inputLength: number, maxLength: number) => string) | undefined,undefined,false,false
 ion-textarea,prop,debounce,number | undefined,undefined,false,false
-ion-textarea,prop,disabled,boolean,false,false,false
+ion-textarea,prop,disabled,boolean,false,false,true
 ion-textarea,prop,enterkeyhint,"done" | "enter" | "go" | "next" | "previous" | "search" | "send" | undefined,undefined,false,false
 ion-textarea,prop,errorText,string | undefined,undefined,false,false
 ion-textarea,prop,fill,"outline" | "solid" | undefined,undefined,false,false
@@ -1891,7 +1892,7 @@ ion-textarea,prop,minlength,number | undefined,undefined,false,false
 ion-textarea,prop,mode,"ios" | "md",undefined,false,false
 ion-textarea,prop,name,string,this.inputId,false,false
 ion-textarea,prop,placeholder,string | undefined,undefined,false,false
-ion-textarea,prop,readonly,boolean,false,false,false
+ion-textarea,prop,readonly,boolean,false,false,true
 ion-textarea,prop,required,boolean,false,false,false
 ion-textarea,prop,rows,number | undefined,undefined,false,false
 ion-textarea,prop,shape,"round" | undefined,undefined,false,false
--- a/core/src/components.d.ts
+++ b/core/src/components.d.ts
@@ -3113,6 +3113,11 @@ export namespace Components {
           * @param smoothScroll : Whether to animate the scroll transition.
          */
         "setContent": (id: string, smoothScroll?: boolean) => Promise<void>;
+        /**
+          * If `true`, users will be able to swipe the segment view to navigate between segment contents.
+          * @default true
+         */
+        "swipeGesture": boolean;
     }
     interface IonSelect {
         /**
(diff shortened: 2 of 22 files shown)


ShaneK (approved):
This looks fine to me, the only minor concern I have is that there's no visual regression tests, but that's probably fine. No idea.

thetaPC:
@ShaneK there's [tests](https://github.com/ionic-team/ionic-framework/pull/30951/changes#diff-20a275779f609ba027286af5de0c4c090f0db124b65d98cd5346e32a251f344a) that check that the parts are working as intended.

ShaneK:
> @ShaneK there's [tests](https://github.com/ionic-team/ionic-framework/pull/30951/changes#diff-20a275779f609ba027286af5de0c4c090f0db124b65d98cd5346e32a251f344a) that check that the parts are working as intended.

That is true, but in many other tests for parts we test with visual regression tests specifically. The fact tests existed at all is why I went ahead and approved and called it a minor concern.

thetaPC:
> > @ShaneK there's [tests](https://github.com/ionic-team/ionic-framework/pull/30951/changes#diff-20a275779f609ba027286af5de0c4c090f0db124b65d98cd5346e32a251f344a) that check that the parts are working as intended.
> 
> That is true, but in many other tests for parts we test with visual regression tests specifically. The fact tests existed at all is why I went ahead and approved and called it a minor concern.

That makes sense, but @brandyscarney has made good points of why we shifted over to use this new approach to test parts. I just can't seem to remember what it was. Maybe less snapshots for us to maintain?

core/src/components/select/select.tsx
```diff
@@ -1246,9 +1253,9 @@ export class Select implements ComponentInterface {
           [`select-label-placement-${labelPlacement}`]: true,
         })}
       >
-        <label class="select-wrapper" id="select-label" onClick={this.onLabelClick}>
+        <label class="select-wrapper" id="select-label" onClick={this.onLabelClick} part="wrapper">
           {this.renderLabelContainer()}
-          <div class="select-wrapper-inner">
+          <div class="select-wrapper-inner" part="wrapper-inner">
```
brandyscarney: This should be named `inner` to go with the way we are naming the inner wrappers in item components.

core/src/components/select/select.tsx
```diff
@@ -1246,9 +1253,9 @@ export class Select implements ComponentInterface {
           [`select-label-placement-${labelPlacement}`]: true,
         })}
       >
-        <label class="select-wrapper" id="select-label" onClick={this.onLabelClick}>
+        <label class="select-wrapper" id="select-label" onClick={this.onLabelClick} part="wrapper">
```
brandyscarney, line 1256: This is likely fine, let's just make sure we all agree on this name.

core/src/components/select/select.tsx
```diff
@@ -45,6 +45,9 @@ import type { SelectChangeEventDetail, SelectInterface, SelectCompareFn } from '
  * @part supporting-text - Supporting text displayed beneath the select.
  * @part helper-text - Supporting text displayed beneath the select when the select is valid.
  * @part error-text - Supporting text displayed beneath the select when the select is invalid and touched.
+ * @part bottom - The container element for helper text, error text, and counter.
+ * @part wrapper - The clickable label element that wraps the entire form field (label text, slots, selected values or placeholder, and toggle icons).
+ * @part wrapper-inner - The inner element of the wrapper that does not include the label text.
```
brandyscarney: Can we change this to define what it wraps instead of what it doesn't?

brandyscarney:
Sorry I missed your previous comments.

We don't need screenshot tests here since we're defining the colors ourselves. If these were the default colors for the elements we would verify them visually to ensure they look correct because the defaults may change frequently. Because we're defining what color the token should use we can just verify that the right element is styled by it.

core/src/components/select/select.tsx
thetaPC: Based on the our internal conversation, this works.

core/src/components/select/select.tsx
thetaPC: https://github.com/ionic-team/ionic-framework/pull/30951/changes/04fdbc6f8fbfeef61b04b124980cdc5ea149f755

core/src/components/select/select.tsx
thetaPC: https://github.com/ionic-team/ionic-framework/pull/30951/changes/6c31bc88677f4a7d4fbd1ff8eda21dd4712b42c2

PR #30955, opened 2026-02-13
fix(nav-controller): reset direction state when navigation is canceled

ShaneK:

---------

When a CanDeactivate guard cancels a back navigation (like when it's initiated by ion-back-button), the NavController’s explicit direction state (back) is never consumed because consumeTransition() is not called for canceled navigations. This stale direction leaks into the next forward navigation, causing it to be incorrectly treated as a back/root navigation. 

## What is the new behavior?
The NavController now listens for NavigationCancel and NavigationError router events and resets direction, animated, and animationBuilder back to their defaults. This ensures stale state from a canceled navigation does not affect subsequent navigations. The reset puts the NavController into 'auto' mode, which correctly uses guessDirection for the next navigation.

- [ ] Yes
- [X] No

## Other information

Current dev build:
```
8.7.18-dev.11771020096.1ca03a6d
```


thetaPC (approved):
LGTM

PR #30958, opened 2026-02-17
fix(radio-group): prevent DOMException and NotFoundError when filtering radios

### LINKED ISSUE #30279: bug: Ionic 8.5 regression: DOMException: Node.removeChild: The node to be removed is not a child of this node
### Prerequisites

### Ionic Framework Version

v8.x

### Current Behavior

After upgrading to 8.5.0 I started encountering the following crash:

```
DOMException: Node.removeChild: The node to be removed is not a child of this node
```

### Expected Behavior

No crash

### Steps to Reproduce

View codepen, see app fails to load with error in console

### Code Reproduction URL

https://stackblitz.com/edit/13jjhrag-hfjp2apn

### Ionic Info

Ionic v8.5.0

### Additional Information

It does NOT happen in 8.4.x. See this working codepen:

https://stackblitz.com/edit/13jjhrag-p1q8ymbd

### LINKED ISSUE #30359: bug: radio-group-wrapper breaks removing options with react
### Prerequisites

### Ionic Framework Version

v8.x

### Current Behavior

The radio-group-wrapper which was introduced in [core/src/components/radio-group/radio-group.tsx](https://github.com/ionic-team/ionic-framework/commit/18e26acb0132234d13355d62d288e920edd11a88#diff-255a631940c66a86086e163c2a131842beb360c945c15835c84d3193d61416a2) for solving #30222 causes React to crash when options are added/removed dynamically like in the typeahead selection example from the [docs](https://ionicframework.com/docs/api/select#typeahead-component).

Workaround: revert Ionic version to 8.4.3.

### Expected Behavior

Ionic should not cause React to crash when radio options are added/removed dynamically.

### Steps to Reproduce

1. Create a React-based ionic app
2. Add a modal with radio group and typeahead functionality (similar to the one from the [docs](https://ionicframework.com/docs/api/select#typeahead-component), but should also work without a modal around the radio group
3. Run the app, type something into the searchbar for filtering the radio items

### Code Reproduction URL

https://stackblitz.com/edit/4cdat5jx

### Ionic Info

```
Ionic:

   Ionic CLI       : 7.2.1 (/home/reuter/.nvm/versions/node/v20.18.0/lib/node_modules/@ionic/cli)
   Ionic Framework : @ionic/react 8.4.3

Capacitor:

   Capacitor CLI      : 7.2.0
   @capacitor/android : 7.2.0
   @capacitor/core    : 7.2.0
   @capacitor/ios     : 7.2.0

Utility:

   native-run  : 2.0.1

System:

   NodeJS : v20.18.0 (/home/reuter/.nvm/versions/node/v20.18.0/bin/node)
   npm    : 10.8.2
   OS     : Linux 5.15
```

### Additional Information

_No response_

OS-jacobbell:
resolves #30279 resolves #30359

While Ionic's `stencil.config.ts` sets `experimentalSlotFixes: true`, the fixes never get applied at runtime. Ionic is using an external runtime, so Ionic components import `defineCustomElement` from `@stencil/core/internal/client` at runtime and this code has no awareness of the project's stencil configuration.

This leads to a `NotFoundError` (Failed to execute 'removeChild' on 'Node') when filtering or dynamically removing radios in an `ion-radio-group`. The error occurs because `ion-radio-group` wraps its slotted content in an internal `<div>`.

## What is the new behavior?
By setting `externalRuntime: false`, Stencil generates a project-specific file with `defineCustomElement` that components import. This file has the project's build settings baked in, correctly applying slot fixes.

Additionally, the internal wrapper `<div>` around the slotted content in `ion-radio-group` is removed. With slot fixes correctly applied and the wrapper removed, radios can be filtered or dynamically removed without triggering `NotFoundError` or `DOMExceptions`.

- [ ] Yes
- [X] No

## Other information
External Runtime is enabled by default and designed for projects that import Stencil components from multiple sources. This is flawed because those components will not be running with the runtime settings for which they were made.

### DIFF
--- a/core/src/components/radio-group/radio-group.scss
+++ b/core/src/components/radio-group/radio-group.scss
@@ -8,10 +8,6 @@ ion-radio-group {
   vertical-align: top;
 }
 
-.radio-group-wrapper {
-  display: inline;
-}
-
 // Radio Group: Top
 // --------------------------------------------------
 


core/src/components/radio-group/radio-group.tsx
```diff
@@ -353,14 +353,7 @@ export class RadioGroup implements ComponentInterface {
         class={mode}
       >
         {this.renderHintText()}
-        {/*
-          TODO(FW-6279): Wrapping the slot in a div is a workaround due to a
-          Stencil issue. Without the wrapper, the children radio will fire the
-          blur event on focus, instead of waiting for them to be blurred.
-        */}
-        <div class="radio-group-wrapper">
```
brandyscarney, line 361: Can we remove the associated css here as well: https://github.com/ionic-team/ionic-framework/blob/70b1237823dd0cdab852486a6b2cbbfe0d0aaae9/core/src/components/radio-group/radio-group.scss#L11-L14

OS-jacobbell: Yes, removed it!

PR #30980, opened 2026-02-27
fix(checkbox): re-evaluate label visibility when label is updated

### LINKED ISSUE #30632: bug: ion-checkbox sometimes does not show label from variable
### Prerequisites

### Ionic Framework Version

v8.x

### Reproduced test case in StackBlitz
In https://stackblitz.com/~/github.com/SandraCaci/test-case-ion-checkbox you see an `ion-checkbox` based on an Angular form control, with default value false and label from a variable, which does not show the label when opening the page. It only shows the label when you check the box and then it remains.
When you refresh the page or navigate to another page and come back, the label disappears again.
### When does the bug not occur?
This does not happen if the form control is disabled, if the default value is true, or if the label is hardcoded in the html (see the other check boxes in the test case).
The bug occurs with Ionic 8.6.6, 8.7.1 and 8.7.2, but not with Ionic 8.6.7 or 8.7.0.
It seems related to the upgrade to `@stencil/core` 4.36.2 which was done in 8.6.6 (reverted in 8.6.7) and again done in 8.7.1, see https://github.com/ionic-team/ionic-framework/releases , because if I force the use of `@stencil/core` 4.33.1 with Ionic 8.7.2 the bug is not there.

### Expected Behavior

The label should show immediately when opening the page. Now the users cannot see what is the purpose of the checkbox.

### Steps to Reproduce

1. Go to https://stackblitz.com/~/github.com/SandraCaci/test-case-ion-checkbox
2. Wait until the app is started and you see "Demo of demo-test-input-checkbox"
3. Notice that the 2nd checkbox does not show a label (the one with red text below it)
4. Click on that checkbox and notice that the label appears.
5. Click the home icon in the upper left icon to go to the home page, and then the checkbox icon just below it to go back to the checkbox page, and note that the label of the 2nd checkbox has disappeared again.

### Code Reproduction URL

https://github.com/SandraCaci/test-case-ion-checkbox

### Ionic Info

Ionic:

   Ionic CLI : 5.4.16

Utility:

   cordova-res : not installed
   native-run  : not installed

System:

   NodeJS : v22.18.0
   npm    : 10.9.3
   OS     : Windows 10

### Additional Information

_No response_

OS-jacobbell:
Issue: Closes #30632

Checkbox's render function applies the `label-text-wrapper-hidden` css class when there is no label text to prevent extra margin from being added. The render function is not re-evaluated if the label is updated. This causes a problem in Angular where dynamic variables get applied after the page is loaded, and a checkbox using a variable as a label gets stuck with its label hidden until something else triggers a re-render, e.g. ticking the box.
## What is the new behavior?

- The checkbox will be re-rendered, and css classes will be updated, when the label text is changed.
- Updated tests to check that the label is visible after changing from blank to having content.

- [ ] Yes
- [X] No

### DIFF
--- a/core/src/components/checkbox/checkbox.tsx
+++ b/core/src/components/checkbox/checkbox.tsx
@@ -1,5 +1,5 @@
 import type { ComponentInterface, EventEmitter } from '@stencil/core';
-import { Build, Component, Element, Event, Host, Method, Prop, State, h, forceUpdate } from '@stencil/core';
+import { Build, Component, Element, Event, Host, Method, Prop, State, h } from '@stencil/core';
 import { checkInvalidState } from '@utils/forms';
 import type { Attributes } from '@utils/helpers';
 import { inheritAriaAttributes, renderHiddenInput } from '@utils/helpers';
@@ -126,6 +126,8 @@ export class Checkbox implements ComponentInterface {
    * Track validation state for proper aria-live announcements.
    */
   @State() isInvalid = false;
+
+  @State() private hasLabelContent = false;
 
   @State() private hintTextId?: string;
 
@@ -266,7 +268,7 @@ export class Checkbox implements ComponentInterface {
   };
 
   private onSlotChange = () => {
-    forceUpdate(this);
+    this.hasLabelContent = this.el.textContent !== '';
   };
 
   private getHintTextId(): string | undefined {
@@ -330,7 +332,6 @@ export class Checkbox implements ComponentInterface {
     } = this;
     const mode = getIonMode(this);
     const path = getSVGPath(mode, indeterminate);
-    const hasLabelContent = el.textContent !== '';
 
     renderHiddenInput(true, el, name, checked ? value : '', disabled);
 
@@ -342,7 +343,7 @@ export class Checkbox implements ComponentInterface {
         aria-checked={indeterminate ? 'mixed' : `${checked}`}
         aria-describedby={this.hintTextId}
         aria-invalid={this.isInvalid ? 'true' : undefined}
-        aria-labelledby={hasLabelContent ? this.inputLabelId : null}
+        aria-labelledby={this.hasLabelContent ? this.inputLabelId : null}
         aria-label={inheritedAttributes['aria-label'] || null}
         aria-disabled={disabled ? 'true' : null}
         aria-required={required ? 'true' : undefined}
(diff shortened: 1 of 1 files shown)


core/src/components/checkbox/checkbox.tsx
```diff
@@ -382,7 +382,7 @@ export class Checkbox implements ComponentInterface {
             id={this.inputLabelId}
             onClick={this.onDivLabelClick}
           >
-            <slot></slot>
+            <slot onSlotchange={() => forceUpdate(this)}></slot>
```
ShaneK: Generally, we should avoid using inline anonymous functions in templates. Doing this both creates a new anonymous function on every render and can make debugging trickier (anonymous functions are harder to find in stack traces than named functions).

It would be better to declare an `onSlotChange` function in the component and use it here.

packages/angular/test/base/src/app/standalone/value-accessors/checkbox/checkbox.component.ts
```diff
@@ -15,6 +15,11 @@ import { ValueAccessorTestComponent } from "../value-accessor-test/value-accesso
   ]
 })
 export class CheckboxComponent {
+  dynamicLabel = '';
+
+  ngAfterViewChecked(): void {
```
ShaneK:
```suggestion
   ngAfterViewInit(): void {
```
This is a test and normally I wouldn't make change requests to tests that are just kinda minor architecture changes, but `ngAfterViewChecked` runs after every change detection cycle and is meant for reading, not writing. I'm concerned this test may be a bad example for our users to follow if they reference it.

brandyscarney (approved):

SandraCaci:
The original bug https://github.com/ionic-team/ionic-framework/issues/30632 seems to have been fixed at first glance, but now when you come to the page from a different page, none of the checkbox labels from variables show. I have updated the StackBlitz reproduction at https://stackblitz.com/~/github.com/SandraCaci/test-case-ion-checkbox to show that.

Steps to reproduce:


1. Go to https://stackblitz.com/~/github.com/SandraCaci/test-case-ion-checkbox
2. Wait until the app is started and you see "Demo of demo-test-input-checkbox" Notice that all checkboxes have a label (Ionic 8.8.2 is applied, the bugfix https://github.com/ionic-team/ionic-framework/pull/30980 ensured that also the 2nd checkbox shows a label).
3. Click the home icon in the upper left icon to go to the home page, and then the checkbox icon just below it to go back to the checkbox page, and note that now all checkbox labels have disappeared.

SandraCaci:
Created new bug https://github.com/ionic-team/ionic-framework/issues/31052

PR #30983, opened 2026-03-01
fix(input-otp): prevent deletion and paste when disabled or readonly

### LINKED ISSUE #30913: bug: input-otp readonly doesn't disable delete
### Prerequisites

### Ionic Framework Version

v8.x

### Current Behavior

when ion-input-otp has the readonly prop, user can not write but can delete characters

### Expected Behavior

input shouldn't allow interaction when readonly

### Steps to Reproduce

Click on any block on the input and press backspace key on keyboard

### Code Reproduction URL

https://ionicframework.com/docs/api/input-otp#states

### Ionic Info

```
   Ionic CLI                     : 7.2.1 (/Users/jongbonga/.nvm/versions/node/v22.16.0/lib/node_modules/@ionic/cli)
   Ionic Framework               : @ionic/angular 8.7.16
   @angular-devkit/build-angular : 20.3.14
   @angular-devkit/schematics    : 20.3.14
   @angular/cli                  : 20.3.14
```

### Additional Information

_No response_

KanhaiyaPandey:
Closes #30913

## Description

When `ion-input-otp` has the `readonly` prop set, typing is correctly blocked but users are still able to delete characters using the Backspace or Delete keys.

When `disabled` or `readonly` are set, users are still able to paste into the component.

## Changes

- Added guard for `disabled` and `readonly` inside `onInput` and `onPaste` handler
- Prevented default behavior for `Backspace` and `Delete` when `readonly` is `true`
- Return in `onKeydown` when `disabled` is `true`
- Added e2e tests verifying the behavior

## How to Test

1. Add `readonly` to `ion-input-otp`
2. Attempt to type → no input allowed
3. Press Backspace/Delete → no characters removed

Behavior now matches expected readonly semantics.

### DIFF
--- a/core/src/components/input-otp/input-otp.tsx
+++ b/core/src/components/input-otp/input-otp.tsx
@@ -620,7 +620,6 @@ export class InputOTP implements ComponentInterface {
     const previousValue = this.previousInputValues[index] || '';
 
     if (disabled || readonly) {
-      event.preventDefault();
       return;
     }
 


core/src/components/input-otp/input-otp.tsx
```diff
@@ -603,11 +614,16 @@ export class InputOTP implements ComponentInterface {
...
    */
   private onInput = (index: number) => (event: InputEvent) => {
-    const { length, validKeyPattern } = this;
+    const { disabled, length, readonly, validKeyPattern } = this;
     const input = event.target as HTMLInputElement;
     const value = input.value;
     const previousValue = this.previousInputValues[index] || '';
 
+    if (disabled || readonly) {
+      event.preventDefault();
```
ShaneK:
```suggestion
 
```

Per the spec, [input's event is not cancelable](https://w3c.github.io/uievents/#event-types-list), so this doesn't actually do anything. It would be best to remove this to prevent future readers from thinking it's required.

brandyscarney: https://github.com/ionic-team/ionic-framework/pull/30983/commits/c6bb32cabd68110300e388a32dbe007d38a1b0aa

PR #31002, opened 2026-03-08
feat(modal): expose isDismissing on ionDragEnd

### LINKED ISSUE #31001: feat: swipe-to-close should support `e.detail.snapBreakpoint`
### Prerequisites

Sheet modals support the new `snapBreakpoint` parameter. It would be great if `swipe-to-close` dialogs also supported this parameter, so we could know if the dialog intends to continue animating to dismiss or not.

### Describe the Use Case

Dismiss the keyboard when the user lifts their finger during a gesture, if the dialog will animate to dismissal.

`e.detail.snapBreakpoint` (would just be 1 or 0) or something like that on `ionDragEnd`

### Describe Alternatives

_No response_

### Related Code

_No response_

### Additional Information

_No response_

aeharding:
Issue number: resolves #31001

---------

The `ionDragEnd` event on card modals emits `currentY`, `deltaY`, `velocityY` and `progress`. None of these indicate whether the modal is about to dismiss, so consumers cannot react to the outcome of the gesture when the user lifts their finger.

## What is the new behavior?

- `ionDragEnd` now includes `isDismissing` on card modals. It is `true` when the gesture passed the dismiss threshold and the modal will animate away, and `false` when the modal will settle back open.
- Updated tests.

Sheet modals are unchanged and continue to report `snapBreakpoint`.

- [ ] Yes
- [X] No

## Other information

Testing with Voyager and works great! https://github.com/aeharding/voyager/pull/2227

Previews:
- [Card modal](https://ionic-framework-git-fork-aeharding-should-complete-ionic1.vercel.app/src/components/modal/test/card/?ionic:mode=ios)
- [Sheet modal](https://ionic-framework-git-fork-aeharding-should-complete-ionic1.vercel.app/src/components/modal/test/sheet)


core/src/components/modal/gestures/swipe-to-close.ts
```diff
@@ -308,6 +308,7 @@ export const createSwipeToCloseGesture = (
       deltaY: detail.deltaY,
       velocityY: detail.velocityY,
       progress: calculateProgress(el, detail.deltaY),
+      snapBreakpoint: shouldComplete ? 0 : 1,
```
thetaPC: `snapBreakpoint` is specific to sheet modals, and card modals have no breakpoint concept, so exposing a "breakpoint" here is a bit misleading. Let's call it `isDismissing` and return a plain boolean instead of `0`/`1`.


```suggestion
      isDismissing: shouldComplete,
```

This also follows the convention we use for other booleans on public event details, such as `isScrolling` on ion-content's `ScrollDetail` and `isAnimating` on the menu interface.

One more change to go with it: please add the field to `ModalDragEventDetail` in `modal-interface.ts` so it's part of the public type:

```ts
/**
   * Whether the card modal will dismiss when the drag gesture ends.
   * `true` if the gesture passed the dismiss threshold, `false` if the
   * modal will remain open.
   */
  isDismissing?: boolean;
```

core/src/components/modal/gestures/swipe-to-close.ts
thetaPC, line 1: Adding this variable changes the `ionDragEnd` detail length, so the [existing assertion](https://github.com/ionic-team/ionic-framework/blob/6251eb85db0e5b43b9e09604248dc8451e1664c0/core/src/components/modal/test/card/modal-card.e2e.ts#L110) in `modal-card.e2e.ts` will fail. Please update the expected length there.

thetaPC (changes requested):

core/src/components/modal/gestures/swipe-to-close.ts
```diff
@@ -315,6 +315,7 @@ export const createSwipeToCloseGesture = (
       deltaY: detail.deltaY,
       velocityY: detail.velocityY,
       progress: calculateProgress(el, detail.deltaY),
+      isDismissing: shouldComplete,
```
ShaneK: I think there's a problem here with `canDismiss`. `shouldComplete` gets forced to `false` whenever `canDismiss` is a function or `false`, so this reports `isDismissing: false` on a modal that's about to dismiss. I built the branch locally to check: card modal with `canDismiss: () => Promise.resolve(true)`, swipe the header down 450px, and `ionDragEnd` fires with `{"deltaY":450,"velocityY":3.94,"progress":0.322,"isDismissing":false}` immediately before `ionModalDidDismiss`.

I'd want this sorted before approving, because "confirm before dismiss" is the most common reason to set `canDismiss` in the first place, and there's no later drag event to correct the value.

core/src/components/modal/test/card/modal-card.e2e.ts
```diff
@@ -148,7 +148,13 @@ configs({ modes: ['ios'], directions: ['ltr'] }).forEach(({ title, screenshot, c
       expect(ionDragMove.length).toBeGreaterThan(0);
 
       expect(ionDragEnd.length).toBe(1);
-      expect(Object.keys(dragEndEvent.detail).length).toBe(4);
+      expect(Object.keys(dragEndEvent.detail).sort()).toEqual([
```
ShaneK, line 153: This only checks the key set, so it passes whether the implementation emits `true`, `false`, `null`, or a hardcoded constant. The value is the entire feature, and the `canDismiss` problem I flagged over in `swipe-to-close.ts` passes like this.

The fixture already handles both outcomes, and the suite already proves which distances land either side of the threshold: `swipeToCloseModal(selector)` defaults to 300px and asserts dismissal, and [the test further up](https://github.com/ionic-team/ionic-framework/blob/8d7013b2a96a3f2237859f246e4ed1bdcac2428b/core/src/components/modal/test/card/modal-card.e2e.ts#L99-L104) drags 20px with the modal staying open. So something like:

```ts
test('should report isDismissing true when the gesture dismisses the modal', async ({ page }) => {
  await dragElementBy(page.locator('.modal-card ion-header'), page, 0, 300);
  const ev = await ionDragEnd.next();
  expect(ev.detail.isDismissing).toBe(true);
});

test('should report isDismissing false when the modal settles back open', async ({ page }) => {
  await dragElementBy(page.locator('.modal-card ion-header'), page, 0, 20);
  const ev = await ionDragEnd.next();
  expect(ev.detail.isDismissing).toBe(false);
  await expect(page.locator('ion-modal')).toBeVisible();
});
```

I ran both against a local build with this patch applied and they pass. Worth pairing each with the visibility check so the test proves the field matches reality instead of matching itself. A third one for the `canDismiss` path would be good too once that's settled, `#card-can-dismiss-promise-true` in `test/can-dismiss/index.html` is already there for it.

Swapping `.length).toBe(4)` for `Object.keys(...).sort()` is a nice improvement by the way, a count wouldn't have caught a rename.

core/src/components/modal/modal-interface.ts
```diff
@@ -72,4 +72,10 @@ export interface ModalDragEventDetail {
    * the gesture.
    */
   snapBreakpoint?: number;
+  /**
+   * Whether the card modal will dismiss when the drag gesture ends.
+   * `true` if the gesture passed the dismiss threshold, `false` if the
+   * modal will remain open.
+   */
+  isDismissing?: boolean;
```
ShaneK, line 83: Sheet modals never set this, so a consumer writing the obvious `if (ev.detail.isDismissing)` gets `undefined` on a sheet that is dismissing. I dragged one down to breakpoint 0 locally and got `{"deltaY":180,"velocityY":10.76,"progress":0,"snapBreakpoint":0}` with `typeof isDismissing === 'undefined'`. That's falsy, so the check quietly does nothing.

Both gestures publish through the same `ionDragEnd`, and `sheet.ts` already has what it needs in scope where it builds the detail:

```ts
isDismissing: snapBreakpoint === 0 && !canDismissBlocksGesture,
```

Same two values `moveSheetToBreakpoint` derives dismissal from. The early-return `onDragEnd` a bit further down would need an explicit `false`, since it doesn't move the sheet at all.

Card `ionDragMove` is the other half of this, it gets nothing while sheet emits `snapBreakpoint` on both move and end, where the move-time meaning is exactly "what happens if you let go now". I think both are worth deciding now rather than after release, because the asymmetry is permanent once it ships. Up to you though, the issue only asked about card modals so I won't block on it.

core/src/components/modal/modal-interface.ts
```diff
@@ -72,4 +72,10 @@ export interface ModalDragEventDetail {
    * the gesture.
    */
   snapBreakpoint?: number;
```
ShaneK, line 74: The `ModalDragEventDetail` block in ionic-docs at `docs/api/modal.md` also needs to be updated, nothing over there picks this up automatically, so as it stands this ships undocumented.

ShaneK (changes requested):
Awesome work! I did catch an issue that I believe is blocking and some others that aren't necessarily but might be nice to fix now

core/src/components/modal/gestures/swipe-to-close.ts
thetaPC: Great catch! I did add a note regarding `canDismiss` https://github.com/ionic-team/ionic-framework/pull/31002/changes/ae4023f7426ded224243070daa9fc025f4f22314

core/src/components/modal/test/card/modal-card.e2e.ts
thetaPC: https://github.com/ionic-team/ionic-framework/pull/31002/changes/ed90f22d5a95b4e9e0a4146f9814fee3c53a283c

core/src/components/modal/modal-interface.ts
thetaPC: https://github.com/ionic-team/ionic-framework/pull/31002/changes/98fd35858546ed397fc57d2941b752c6abd0dfa5

core/src/components/modal/modal-interface.ts
thetaPC: https://github.com/ionic-team/ionic-docs/pull/4612

ShaneK (approved):
Looks good to me! Great work!

thetaPC (approved):
LGTM

thetaPC:
Thank you for submitting the PR! The feature will be released in v9.

PR #31005, opened 2026-03-09
test(angular): validate checkbox and toggle in lazy template-form

OS-jacobbell:
Checkbox and toggle components are not validated to be ticked/on in `packages/angular/test/base/src/app/lazy/template-form`. This prevents the error text from being displayed. While they have the `required` attribute, this only applies to accessibility for [checkbox](https://ionicframework.com/docs/api/checkbox#required) and [toggle](https://ionicframework.com/docs/api/toggle#required).

## What is the new behavior?
- Use an Angular validator directive for checkbox and toggle.
- Make template-form an Angular module so the validator directive can be imported.

- [ ] Yes
- [X] No


packages/angular/test/base/src/app/lazy/app-lazy/app.routes.ts
ShaneK:
```suggestion
```

Looks like you removed the reference to this

PR #30873, opened 2025-12-15
feat(chip): add recipe and variables

thetaPC:
Issue number: resolves internal

---------

## What is the current behavior?

Component styles for `ion-chip` are fragmented across multiple files (`ios`, `md`, `ionic`). Developers were restricted to those themes and how those themes structured the logic and styles.

## What is the new behavior?

- **Unified Style Architecture**: Combined theme-specific styling into a single `chip.scss` file that consumes CSS variables, ensuring a single source of truth for component logic.
- **Property-First Token Structure**: Refactored the `IonChip` type and tokens to prioritize a **Property-First approach**. Instead of flat or element-centric tokens, they are nested by element and then by property (e.g., `icon.firstChild.margin.end`). This ensures:
    - **Predictability**: Developers can determine token names based on the CSS property they wish to override.
    - **Consistency**: The structure matches other system keys like `padding` and `border`.
    - **Granularity**: Allows for specific overrides of individual directions without cluttering the top-level interface.
- **Defined TypeScript Interfaces**: Added `chip.interfaces.ts` with reusable utility types like `HueRef` to standardize the "Bold vs. Subtle" logic used in variants and interaction states.
- **Standardized Sizing**: Introduced a "small" to `md` and `ios`.
- **Standardized Hues**: Introduced a "subtle" to `md` and `ios`.
- **Defined Defaults**: Updated defaults to be consistent regardless of theme.

## Does this introduce a breaking change?

- [x] Yes
- [ ] No

This PR introduces a breaking change to how `IonChip` is styled. Existing manual CSS overrides targeting internal chip structures or old token names will no longer work due to the shift to a **Property-First** token hierarchy and a unified base SCSS file. 

**Migration Path:**
1.  **Token Updates**: Update any custom theme configurations to match the new nested structure.

2.  **CSS Overrides**: Ensure selectors align with the new slotted element logic and variable names (e.g., `--ion-chip-hue-bold-bg`).

`--background` should no longer be used. Setting the value with `IonChip.hue.bold.bg` and `IonChip.hue.subtle.bg` within the tokens file should be used to change the background based on the hue.

3.  **Properties**: Update any references to match old defaults based on the theme.

If `md` or `ios` with no shape, then update the component to default to `soft`.

## Other information

This PR significantly improves the developer experience for theming. By moving logic into `default.tokens.ts` and away from hardcoded SCSS functions, designers and developers can now override complex states (like an activated outline chip) purely through token configuration.

Preview:
[...]

PR #31009, opened 2026-03-11
feat(item-divider): add recipe and tokens

thetaPC:
Issue number: internal

---------

Component styles for `ion-item-divider` are fragmented across multiple files (ios, md). The `ionic` theme uses the `md` styles. Developers were restricted to those themes and how those themes structured the logic and styles.

## What is the new behavior?

- **Unified Style Architecture**: Combined theme-specific styling into a single `item-divider.scss` file that consumes CSS variables, ensuring a single source of truth for component logic.
- **Anchor-Edge model**: expanded the property-first approach into an anchor-edge model to account for the amount of slot usage
- **Defined TypeScript Interface**: Added `item-divider.interfaces.ts`
- **Defined Shared Base Defaults**: Added defaults that are being used regardless of theme
- **Defined Shared Theme Defaults**: Added defaults that are being used per theme
- **Updated Test Page**: Updated the dividers test page to include more cases

- [x] Yes
- [ ] No

This PR introduces a breaking change to how `IonItemDivider` is styled. Existing manual CSS overrides targeting internal item divider structures or old token names will no longer work due to the shift to thew new token hierarchy and a unified base SCSS file. 

**Migration Path:**
1.  **Token Updates**: Update any custom theme configurations to match the new nested structure.

2.  **CSS Overrides**: Ensure selectors align with the new slotted element logic and variable names (e.g., `--ion-item-divider-background`).

`--background` should no longer be used. Setting the value, `IonDivider.background`, within the tokens file should be used to change the background.

3.  **Theme classes**: Remove any instances that target the theme classes: `ion-divider.md`

## Other information

This PR significantly improves the developer experience for theming. By moving logic into `default.tokens.ts` and away from hardcoded SCSS functions, designers and developers can now override styles (like the size on a slotted avatar) purely through token configuration.

[Preview](https://ionic-framework-git-fw-6841-ionic1.vercel.app/src/components/item/test/dividers/)

### DIFF
--- a/core/src/components/item-divider/item-divider.scss
+++ b/core/src/components/item-divider/item-divider.scss
@@ -33,72 +33,72 @@
    *
    * Slotted Label
    * @prop --ion-item-divider-label-margin-top: Top margin of the slotted ion-label
-    * @prop --ion-item-divider-label-margin-end: Right margin if direction is left-to-right, and left margin if direction is right-to-left of the slotted ion-label
-    * @prop --ion-item-divider-label-margin-bottom: Bottom margin of the slotted ion-label
-    * @prop --ion-item-divider-label-margin-start: Left margin if direction is left-to-right, and right margin if direction is right-to-left of the slotted ion-label
-    *
-    * Slotted Icon
-    * @prop --ion-item-divider-icon-default-color: Color of the slotted ion-icon
-    * @prop --ion-item-divider-icon-semantic-default-color: Color of the slotted ion-icon when the item divider has a semantic color applied
-    * @prop --ion-item-divider-icon-font-size: Font size of the slotted ion-icon
-    * @prop --ion-item-divider-icon-start-slotted-margin-top: Top margin of the slotted ion-icon when it is in the start slot of the item divider
-    * @prop --ion-item-divider-icon-start-slotted-margin-end: Right margin if direction is left-to-right, and left margin if direction is right-to-left of the slotted ion-icon when it is in the start slot of the item divider
-    * @prop --ion-item-divider-icon-start-slotted-margin-bottom: Bottom margin of the slotted ion-icon when it is in the start slot of the item divider
-    * @prop --ion-item-divider-icon-start-slotted-margin-start: Left margin if direction is left-to-right, and right margin if direction is right-to-left of the slotted ion-icon when it is in the start slot of the item divider
-    * @prop --ion-item-divider-icon-end-slotted-margin-top: Top margin of the slotted ion-icon when it is in the end slot of the item divider
-    * @prop --ion-item-divider-icon-end-slotted-margin-end: Right margin if direction is left-to-right, and left margin if direction is right-to-left of the slotted ion-icon when it is in the end slot of the item divider
-    * @prop --ion-item-divider-icon-end-slotted-margin-bottom: Bottom margin of the slotted ion-icon when it is in the end slot of the item divider
-    * @prop --ion-item-divider-icon-end-slotted-margin-start: Left margin if direction is left-to-right, and right margin if direction is right-to-left of the slotted ion-icon when it is in the end slot of the item divider
-    *
-    * Slotted Note
-    * @prop --ion-item-divider-note-margin-top: Top margin of the slotted ion-note
-    * @prop --ion-item-divider-note-margin-end: Right margin if direction is left-to-right, and left margin if direction is right-to-left of the slotted ion-note
-    * @prop --ion-item-divider-note-margin-bottom: Bottom margin of the slotted ion-note
-    * @prop --ion-item-divider-note-margin-start: Left margin if direction is left-to-right, and right margin if direction is right-to-left of the slotted ion-note
-    * @prop --ion-item-divider-note-align-self: Align self of the slotted ion-note
-    * @prop --ion-item-divider-note-font-size: Font size of the slotted ion-note
-    * @prop --ion-item-divider-note-padding-top: Top padding of the slotted ion-note when it has a slot
-    * @prop --ion-item-divider-note-padding-end: Right padding if direction is left-to-right, and left padding if direction is right-to-left of the slotted ion-note when it has a slot
-    * @prop --ion-item-divider-note-padding-bottom: Bottom padding of the slotted ion-note when it has a slot
-    * @prop --ion-item-divider-note-padding-start: Left padding if direction is left-to-right, and right padding if direction is right-to-left of the slotted ion-note when it has a slot
-    *
-    * Slotted Avatar
-    * @prop --ion-item-divider-avatar-margin-top: Top margin of the slotted ion-avatar
-    * @prop --ion-item-divider-avatar-margin-end: Right margin if direction is left-to-right, and left margin if direction is right-to-left of the slotted ion-avatar
-    * @prop --ion-item-divider-avatar-margin-bottom: Bottom margin of the slotted ion-avatar
-    * @prop --ion-item-divider-avatar-margin-start: Left margin if direction is left-to-right, and right margin if direction is right-to-left of the slotted ion-avatar
-    * @prop --ion-item-divider-avatar-width: Width of the slotted ion-avatar
-    * @prop --ion-item-divider-avatar-height: Height of the slotted ion-avatar
-    * @prop --ion-item-divider-avatar-start-slotted-margin-start: Left margin if direction is left-to-right, and right margin if direction is right-to-left of the slotted ion-avatar when it is in the start slot of the item divider
-    * @prop --ion-item-divider-avatar-start-slotted-margin-end: Right margin if direction is left-to-right, and left margin if direction is right-to-left of the slotted ion-avatar when it is in the start slot of the item divider
-    * @prop --ion-item-divider-avatar-end-slotted-margin-start: Left margin if direction is left-to-right, and right margin if direction is right-to-left of the slotted ion-avatar when it is in the end slot of the item divider
-    * @prop --ion-item-divider-avatar-end-slotted-margin-end: Right margin if direction is left-to-right, and left margin if direction is right-to-left of the slotted ion-avatar when it is in the end slot of the item divider
-    *
-    * Slotted Thumbnail
-    * @prop --ion-item-divider-thumbnail-margin-top: Top margin of the slotted ion-thumbnail
-    * @prop --ion-item-divider-thumbnail-margin-end: Right margin if direction is left-to-right, and left margin if direction is right-to-left of the slotted ion-thumbnail
-    * @prop --ion-item-divider-thumbnail-margin-bottom: Bottom margin of the slotted ion-thumbnail
-    * @prop --ion-item-divider-thumbnail-margin-start: Left margin if direction is left-to-right, and right margin if direction is right-to-left of the slotted ion-thumbnail
-    * @prop --ion-item-divider-thumbnail-width: Width of the slotted ion-thumbnail
-    * @prop --ion-item-divider-thumbnail-height: Height of the slotted ion-thumbnail
-    * @prop --ion-item-divider-thumbnail-start-slotted-margin-start: Left margin if direction is left-to-right, and right margin if direction is right-to-left of the slotted ion-thumbnail when it is in the start slot of the item divider
-    * @prop --ion-item-divider-thumbnail-start-slotted-margin-end: Right margin if direction is left-to-right, and left margin if direction is right-to-left of the slotted ion-thumbnail when it is in the start slot of the item divider
-    * @prop --ion-item-divider-thumbnail-end-slotted-margin-start: Left margin if direction is left-to-right, and right margin if direction is right-to-left of the slotted ion-thumbnail when it is in the end slot of the item divider
-    * @prop --ion-item-divider-thumbnail-end-slotted-margin-end: Right margin if direction is left-to-right, and left margin if direction is right-to-left of the slotted ion-thumbnail when it is in the end slot of the item divider
-    *
-    * Slotted Paragraph
-    * @prop --ion-item-divider-paragraph-margin-top: Top margin of the slotted p
-    * @prop --ion-item-divider-paragraph-margin-end: Right margin if direction is left-to-right, and left margin if direction is right-to-left of the slotted p
-    * @prop --ion-item-divider-paragraph-margin-bottom: Bottom margin of the slotted p
-    * @prop --ion-item-divider-paragraph-margin-start: Left margin if direction is left-to-right, and right margin if direction is right-to-left of the slotted p
-    * @prop --ion-item-divider-paragraph-color: Color of the slotted p
-    * @prop --ion-item-divider-paragraph-font-size: Font size of the slotted p
-    * @prop --ion-item-divider-paragraph-text-overflow: Text overflow of the slotted p
-    * @prop --ion-item-divider-paragraph-overflow: Overflow of the slotted p
-    * @prop --ion-item-divider-paragraph-trailing-margin-top: Top margin of the slotted p when it is the last child
-    * @prop --ion-item-divider-paragraph-trailing-margin-end: Right margin if direction is left-to-right, and left margin if direction is right-to-left of the slotted p when it is the last child
-    * @prop --ion-item-divider-paragraph-trailing-margin-bottom: Bottom margin of the slotted p when it is the last child
-    * @prop --ion-item-divider-paragraph-trailing-margin-start: Left margin if direction is left-to-right, and right margin if direction is right-to-left of the slotted p when it is the last child
+   * @prop --ion-item-divider-label-margin-end: Right margin if direction is left-to-right, and left margin if direction is right-to-left of the slotted ion-label
+   * @prop --ion-item-divider-label-margin-bottom: Bottom margin of the slotted ion-label
+   * @prop --ion-item-divider-label-margin-start: Left margin if direction is left-to-right, and right margin if direction is right-to-left of the slotted ion-label
+   *
+   * Slotted Icon
+   * @prop --ion-item-divider-icon-default-color: Color of the slotted ion-icon
+   * @prop --ion-item-divider-icon-semantic-default-color: Color of the slotted ion-icon when the item divider has a semantic color applied
+   * @prop --ion-item-divider-icon-font-size: Font size of the slotted ion-icon
+   * @prop --ion-item-divider-icon-start-slotted-margin-top: Top margin of the slotted ion-icon when it is in the start slot of the item divider
+   * @prop --ion-item-divider-icon-start-slotted-margin-end: Right margin if direction is left-to-right, and left margin if direction is right-to-left of the slotted ion-icon when it is in the start slot of the item divider
+   * @prop --ion-item-divider-icon-start-slotted-margin-bottom: Bottom margin of the slotted ion-icon when it is in the start slot of the item divider
+   * @prop --ion-item-divider-icon-start-slotted-margin-start: Left margin if direction is left-to-right, and right margin if direction is right-to-left of the slotted ion-icon when it is in the start slot of the item divider
+   * @prop --ion-item-divider-icon-end-slotted-margin-top: Top margin of the slotted ion-icon when it is in the end slot of the item divider
+   * @prop --ion-item-divider-icon-end-slotted-margin-end: Right margin if direction is left-to-right, and left margin if direction is right-to-left of the slotted ion-icon when it is in the end slot of the item divider
+   * @prop --ion-item-divider-icon-end-slotted-margin-bottom: Bottom margin of the slotted ion-icon when it is in the end slot of the item divider
+   * @prop --ion-item-divider-icon-end-slotted-margin-start: Left margin if direction is left-to-right, and right margin if direction is right-to-left of the slotted ion-icon when it is in the end slot of the item divider
+   *
+   * Slotted Note
+   * @prop --ion-item-divider-note-margin-top: Top margin of the slotted ion-note
+   * @prop --ion-item-divider-note-margin-end: Right margin if direction is left-to-right, and left margin if direction is right-to-left of the slotted ion-note
+   * @prop --ion-item-divider-note-margin-bottom: Bottom margin of the slotted ion-note
+   * @prop --ion-item-divider-note-margin-start: Left margin if direction is left-to-right, and right margin if direction is right-to-left of the slotted ion-note
+   * @prop --ion-item-divider-note-align-self: Align self of the slotted ion-note
+   * @prop --ion-item-divider-note-font-size: Font size of the slotted ion-note
+   * @prop --ion-item-divider-note-padding-top: Top padding of the slotted ion-note when it has a slot
+   * @prop --ion-item-divider-note-padding-end: Right padding if direction is left-to-right, and left padding if direction is right-to-left of the slotted ion-note when it has a slot
+   * @prop --ion-item-divider-note-padding-bottom: Bottom padding of the slotted ion-note when it has a slot
+   * @prop --ion-item-divider-note-padding-start: Left padding if direction is left-to-right, and right padding if direction is right-to-left of the slotted ion-note when it has a slot
+   *
+   * Slotted Avatar
+   * @prop --ion-item-divider-avatar-margin-top: Top margin of the slotted ion-avatar
+   * @prop --ion-item-divider-avatar-margin-end: Right margin if direction is left-to-right, and left margin if direction is right-to-left of the slotted ion-avatar
+   * @prop --ion-item-divider-avatar-margin-bottom: Bottom margin of the slotted ion-avatar
+   * @prop --ion-item-divider-avatar-margin-start: Left margin if direction is left-to-right, and right margin if direction is right-to-left of the slotted ion-avatar
+   * @prop --ion-item-divider-avatar-width: Width of the slotted ion-avatar
+   * @prop --ion-item-divider-avatar-height: Height of the slotted ion-avatar
+   * @prop --ion-item-divider-avatar-start-slotted-margin-start: Left margin if direction is left-to-right, and right margin if direction is right-to-left of the slotted ion-avatar when it is in the start slot of the item divider
+   * @prop --ion-item-divider-avatar-start-slotted-margin-end: Right margin if direction is left-to-right, and left margin if direction is right-to-left of the slotted ion-avatar when it is in the start slot of the item divider
+   * @prop --ion-item-divider-avatar-end-slotted-margin-start: Left margin if direction is left-to-right, and right margin if direction is right-to-left of the slotted ion-avatar when it is in the end slot of the item divider
+   * @prop --ion-item-divider-avatar-end-slotted-margin-end: Right margin if direction is left-to-right, and left margin if direction is right-to-left of the slotted ion-avatar when it is in the end slot of the item divider
+   *
+   * Slotted Thumbnail
+   * @prop --ion-item-divider-thumbnail-margin-top: Top margin of the slotted ion-thumbnail
+   * @prop --ion-item-divider-thumbnail-margin-end: Right margin if direction is left-to-right, and left margin if direction is right-to-left of the slotted ion-thumbnail
+   * @prop --ion-item-divider-thumbnail-margin-bottom: Bottom margin of the slotted ion-thumbnail
+   * @prop --ion-item-divider-thumbnail-margin-start: Left margin if direction is left-to-right, and right margin if direction is right-to-left of the slotted ion-thumbnail
+   * @prop --ion-item-divider-thumbnail-width: Width of the slotted ion-thumbnail
+   * @prop --ion-item-divider-thumbnail-height: Height of the slotted ion-thumbnail
+   * @prop --ion-item-divider-thumbnail-start-slotted-margin-start: Left margin if direction is left-to-right, and right margin if direction is right-to-left of the slotted ion-thumbnail when it is in the start slot of the item divider
+   * @prop --ion-item-divider-thumbnail-start-slotted-margin-end: Right margin if direction is left-to-right, and left margin if direction is right-to-left of the slotted ion-thumbnail when it is in the start slot of the item divider
+   * @prop --ion-item-divider-thumbnail-end-slotted-margin-start: Left margin if direction is left-to-right, and right margin if direction is right-to-left of the slotted ion-thumbnail when it is in the end slot of the item divider
+   * @prop --ion-item-divider-thumbnail-end-slotted-margin-end: Right margin if direction is left-to-right, and left margin if direction is right-to-left of the slotted ion-thumbnail when it is in the end slot of the item divider
+   *
+   * Slotted Paragraph
+   * @prop --ion-item-divider-paragraph-margin-top: Top margin of the slotted p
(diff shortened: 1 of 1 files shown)


core/src/components/item-divider/item-divider.scss
```diff
@@ -113,3 +125,283 @@
...
+  @include mixins.margin-horizontal(
+    var(--ion-item-divider-thumbnail-trailing-edge-margin-start),
+    var(--ion-item-divider-thumbnail-trailing-edge-margin-end)
+  );
+}
+
+// Slotted Content
+// --------------------------------------------------
+
+::slotted(h1) {
```
thetaPC: My following comment applies to all the slotted headers:

I noticed that the margins are being overridden by a high level style for headers. I'm not sure when the margins declared in the item divider ever get applied. This is also being overridden in `main`. Can we just get rid of the slotted header styles?

core/src/components/chip/chip.scss
thetaPC, line 1: Noticed that color functions were never used.

core/src/components/item-divider/item-divider.scss
```diff
@@ -113,3 +125,283 @@
...
+    var(--ion-item-divider-avatar-trailing-edge-margin-end)
+  );
+}
+
+// Slotted Thumbnail
+// --------------------------------------------------
+
+::slotted(ion-thumbnail) {
+  // TODO(FW-6862): separate width and height tokens for thumbnails
+  --size: var(--ion-item-divider-thumbnail-width);
```
thetaPC, line 312: To maximize customization, we should split the size into height and width.

core/src/components/thumbnail/thumbnail.scss
```diff
@@ -8,15 +8,15 @@
    * @prop --border-radius: Border radius of the thumbnail
    * @prop --size: Size of the thumbnail
    */
-  --size: 48px;
+  --size: 48px; // TODO(FW-6862): separate width and height tokens for thumbnails
```
thetaPC, line 11: To maximize customization, we should split the size into height and width.

core/src/components/thumbnail/thumbnail.scss
```diff
@@ -8,15 +8,15 @@
...
   --border-radius: 0;
 
   @include border-radius(var(--border-radius));
 
   display: block;
 
-  width: var(--size);
-  height: var(--size);
+  width: var(--size, 48px);
+  height: var(--size, 48px);
```
thetaPC, line 19: We're using `--ion-item-divider-thumbnail-width` to let dividers control thumbnail sizes. The problem is that if a theme doesn't provide this token, the CSS breaks instead of falling back. We can't use `revert-layer` because the 'original' size we want to return to is actually another variable, not a fixed value. The cleanest fix is to just bake the default `48px` right into the variable's fallback.

core/src/components/item-divider/item-divider.scss
```diff
@@ -113,3 +125,283 @@
 
   overflow: hidden;
 }
+
+// Item Divider Slots
+// --------------------------------------------------
+
+:host([slot="start"]) {
+  @include mixins.margin(
+    var(--ion-item-divider-leading-anchor-margin-top, revert-layer),
```
thetaPC: `revert-layer` is being used so that if we don't provide a modular CSS variable, we give the component permission to just keep doing whatever it was doing originally.

core/src/components/item-divider/item-divider.interfaces.ts
```diff
@@ -0,0 +1,199 @@
...
+      margin?: IonMargin;
+    };
+  };
+
+  trailing?: {
+    // Targets `::slotted([slot="end"])`
+    edge?: {
+      margin?: IonMargin;
+    };
+  };
```
thetaPC: ### 💡 Naming Convention: Anchors & Edges

To maintain consistency with the Ionic Modular architecture established in `ion-chip`, this component adopts a shared spatial vocabulary. While chips use DOM order (`:first-child`), dividers use named slots (`slot="start/end"`). The following naming convention bridges that gap:

- **Anchors (`:host`)**: Refers to the component's own placement within a parent container.
    - `leading-anchor`: Targets `:host([slot="start"])`.
    - `trailing-anchor`: Targets `:host([slot="end"])`.
- **Edges (`::slotted`)**: Refers to content placed at the boundaries of the component itself.
    - `leading-edge`: Targets `::slotted([slot="start"])`.
    - `trailing-edge`: Targets `::slotted([slot="end"])`.

**Why this matters:**
This mirrors the `leading` (`:first-child`) and `trailing` (`:last-child`) pattern used in chips. By using `leading/trailing` instead of `left/right` or `start/end`, we create a less confusing structure that developers can use with ease.

core/src/components/item-divider/item-divider.interfaces.ts
```diff
@@ -0,0 +1,199 @@
...
+
+    default?: {
+      color?: string;
+    };
+
+    semantic?: {
+      default?: {
+        color?: string;
+      };
+    };
```
thetaPC, line 76: This follows the same structure established in chip when it came to default vs semantic colors.

core/src/components/item/test/dividers/item.e2e.ts-snapshots/item-dividers-diff-ios-ltr-Mobile-Chrome-linux.png
thetaPC, line 1: Added:
- Item divider with icons at the start and end slot
- Item divider with a thumbnail at the start
- Item divider with paragrahs

core/src/themes/ionic/shared.tokens.ts
```diff
@@ -0,0 +1,137 @@
+/**
+ * Common tokens utilized multiple times across `ionic` theme files.
+ */
+
+import { mix, rgba } from '../../utils/theme';
+import { colors as baseColors } from '../base/shared.tokens';
+
+export const components = {
+  item: {
```
thetaPC, line 14: Item divider depends on the values set on item. This file was created to keep track of those values and other components that need to be shared with each other.

core/src/components/item-divider/item-divider.scss
```diff
@@ -113,3 +126,277 @@
...
+// Note
+::slotted(ion-note) {
+  @include mixins.margin(
+    var(--ion-item-divider-note-margin-top),
+    var(--ion-item-divider-note-margin-end),
+    var(--ion-item-divider-note-margin-bottom),
+    var(--ion-item-divider-note-margin-start)
+  );
+
+  align-self: flex-start;
```
ShaneK: Shouldn't this use `--ion-item-divider-note-align-self`? You're hard coding it here, but there's a CSS variable prepared for it that's not being consumed?

core/src/themes/ionic/shared.tokens.ts
```diff
@@ -0,0 +1,137 @@
...
+
+    inner: {
+      padding: {
+        end: 'var(--ion-spacing-lg)',
+      },
+    },
+
+    border: {
+      color: `var(--ion-item-border-color, var(--ion-border-color, var(--ion-background-color-step-150, ${rgba(
+        '0, 0, 0',
```
ShaneK: Should this be using `baseColors.textColorRgb` instead of the hardcoded `'0, 0, 0'`? The md version uses `baseColors.textColorRgb` for this same border color which lets it respond to `--ion-text-color-rgb` customization, but this one won't. You're already using `baseColors.textColorRgb` correctly for the icon color on line 47 so I'm guessing this is just a copy-paste miss?

core/src/utils/theme.ts
```diff
@@ -633,3 +633,32 @@ export function currentColor(variation: string, alpha: number | string | null =
...
+ * - `dynamicFont(20, 'em')` returns `'1.25em'`
+ *
+ * @internal
+ * @param size - The numeric pixel value.
+ * @param unit - The CSS unit string.
+ * @returns The calculated CSS value string.
+ */
+export const dynamicFont = (size: number, unit: string | undefined = baselineUnit): string => {
+  let baselinePixelSize = 16;
+
```
ShaneK, line 680: Does this path actually work? `config` is a plain object without a `.get()` method, and this runs at import time before `Ionic.config` is initialized anyway. So `configRootFontSize` will always be `undefined` and you'll always fall back to 16. Is this meant to hook into something that doesn't exist yet, or should we just drop the window check and use the constant?

core/src/components/item-divider/item-divider.scss
```diff
@@ -113,3 +126,277 @@
...
+  @include mixins.margin(
+    var(--ion-item-divider-header6-trailing-margin-top),
+    var(--ion-item-divider-header6-trailing-margin-end),
+    var(--ion-item-divider-header6-trailing-margin-bottom),
+    var(--ion-item-divider-header6-trailing-margin-start)
+  );
+}
+
+// Paragraph
+::slotted(p) {
```
ShaneK, line 337: Was dropping `line-height: normal` intentional here? Both the old iOS and MD files had it on `::slotted(p)`. Without it paragraphs will just inherit from the parent, which is probably fine most of the time but it's a behavior change.

core/src/components/item-divider/item-divider.scss
```diff
@@ -51,19 +55,28 @@
...
+  // TODO(FW-6848): remove border-bottom and replace it with the `ion-divider`
+  border-bottom: var(--ion-item-divider-border-bottom);
+
   background: var(--background);
   color: var(--color);
 
-  font-family: $font-family-base;
+  font-family: var(--ion-font-family, inherit);
+  font-size: var(--ion-item-divider-font-size);
+  font-weight: var(--ion-item-divider-font-weight);
```
ShaneK, line 135: iOS previously set `font-weight: 600` explicitly but the iOS recipe doesn't define a `font.weight` token. Without a fallback value here this becomes invalid at computed-value time and falls to inherited/initial (probably `normal`/400). Should iOS define `font.weight` in its recipe, or should this have a fallback like `var(--ion-item-divider-font-weight, normal)`?

core/src/components/item-divider/item-divider.interfaces.ts
```diff
@@ -0,0 +1,201 @@
...
+      margin?: IonMargin;
+    };
+
+    // Targets `::slotted([slot="start"])`
+    edge?: {
+      margin?: IonMargin;
+    };
+  };
+
+  trailing?: {
```
ShaneK: Something I noticed: `trailing` means `slot="end"` here (and in `icon.trailing`, `avatar.trailing`, etc.), but down on `header2.trailing` (line 141) it means `:last-child` which is the same meaning chip uses for `trailing`. So the same word means two different things within the same recipe. The anchor/edge vocabulary helps but I think this could still trip people up when working across recipes. Worth thinking about whether there's a clearer way to distinguish these?

core/src/themes/ios/shared.tokens.ts
```diff
@@ -0,0 +1,68 @@
...
+    },
+
+    inner: {
+      padding: {
+        end: 'var(--ion-spacing-lg)',
+      },
+    },
+
+    // Targets `:host([slot="start"])`
+    slot: {
```
ShaneK, line 32: The nesting order here is `item.slot.start.margin` but ionic and md use `item.start.slot.margin` for the semantically equivalent thing. I know they target different CSS selectors (`:host([slot])` vs `::slotted([slot])`) so different naming is defensible, but if someone's adding a new theme or reading across implementations this is going to be confusing. Could we normalize the path structure across all three?

core/src/components/item-divider/item-divider.scss
thetaPC: Good catch! https://github.com/ionic-team/ionic-framework/pull/31009/changes/be24a4170ef118a52a258eb56a1ec0a69fee36d0

core/src/themes/ionic/shared.tokens.ts
thetaPC: Good catch! https://github.com/ionic-team/ionic-framework/pull/31009/changes/85953533c894ac87b9b8b1c8988d11211e4358e3

core/src/utils/theme.ts
thetaPC: https://github.com/ionic-team/ionic-framework/pull/31009/changes/bd9f016beb5e945fb1bef3ee993051669597ac32

core/src/components/item-divider/item-divider.scss
thetaPC: Added it back just in case. I removed it at first because I didn't see any changes when I removed it. https://github.com/ionic-team/ionic-framework/pull/31009/changes/ac4433886107b82fe576f23f2767dd8a7e2e4625

core/src/components/item-divider/item-divider.scss
thetaPC: It's right here: https://github.com/ionic-team/ionic-framework/blob/02515ef4a0dc44659e1269f198712f096cb6ab9d/core/src/themes/ios/default.tokens.ts#L428

core/src/themes/ios/shared.tokens.ts
thetaPC: How does this sound? https://github.com/ionic-team/ionic-framework/pull/31009/changes/fd73b43bc4bbb8a02d5e58bef0bbc743f945850d

core/src/components/item-divider/item-divider.interfaces.ts
thetaPC: What about this? I also had to update it on chip. https://github.com/ionic-team/ionic-framework/pull/31009/changes/b9c7961bf1e43e4ee628ffb1839c5e63a78bf32c

core/src/components/item-divider/item-divider.scss
```diff
@@ -1,46 +1,171 @@
...
+    * @prop --ion-item-divider-note-margin-end: Right margin if direction is left-to-right, and left margin if direction is right-to-left of the slotted ion-note
+    * @prop --ion-item-divider-note-margin-bottom: Bottom margin of the slotted ion-note
+    * @prop --ion-item-divider-note-margin-start: Left margin if direction is left-to-right, and right margin if direction is right-to-left of the slotted ion-note
+    * @prop --ion-item-divider-note-align-self: Align self of the slotted ion-note
+    * @prop --ion-item-divider-note-font-size: Font size of the slotted ion-note
+    * @prop --ion-item-divider-note-padding-top: Top padding of the slotted ion-note when it has a slot
+    * @prop --ion-item-divider-note-padding-end: Right padding if direction is left-to-right, and left padding if direction is right-to-left of the slotted ion-note when it has a slot
+    * @prop --ion-item-divider-note-padding-bottom: Bottom padding of the slotted ion-note when it has a slot
+    * @prop --ion-item-divider_note_padding_start: Left padding if direction is left-to-right, and right padding if direction is right-to-left of the slotted ion-note when it has a slot
```
ShaneK: Three `@prop` typos here and at lines 131 and 139 -- they're all using an underscore before `start` instead of a dash. This one should be `--ion-item-divider-note-padding-start`, line 131 should be `--ion-item-divider-header5-last-margin-start`, and line 139 should be `--ion-item-divider-header6-last-margin-start`. The actual CSS at lines 344, 480, and 498 is correct, so this just causes `api.txt` to have the wrong names.

thetaPC: If only they were generated, all manual until we can figure out how to do it: https://github.com/ionic-team/ionic-framework/pull/31009/changes/9af22f8bb502fa4b3eaced3ff0665496ea0e8e4e

ShaneK: Ah, I didn't realize they weren't generated. Bummer

core/src/components/item/test/dividers/item.e2e.ts-snapshots/item-dividers-diff-ios-ltr-Mobile-Chrome-linux.png
brandyscarney, line 1: We really should break these screenshots up so that we can add new use cases without updating existing screenshots. I am fine with doing it after the modular migration though to verify no regressions.

core/src/components/thumbnail/thumbnail.scss
brandyscarney: Why don't we split it now? We can use `--size` as a fallback if we don't want to introduce a breaking change.

core/src/themes/md/default.tokens.ts
```diff
@@ -356,10 +353,10 @@ export const defaultTheme: DefaultTheme = {
         color: rgba('var(--ion-text-color-rgb, 0, 0, 0)', 0.54),
 
         font: {
-          size: `${(20 / fontSizes.chipBase).toFixed(2)}em`,
+          size: `${(20 / components.chip.font.size).toFixed(2)}em`,
         },
 
-        leading: {
+        first: {
```
brandyscarney: Why were these renamed back to `first` and `last` when we updated them [here](https://github.com/ionic-team/ionic-framework/pull/30873/changes/BASE..81ec32ff059cb1f1287c10d571bd56c7deb522f6#diff-e6a6723453a8bda8c23f44a6bb5bef25c77e6b98485590310373c63723220c69)?

Image

core/src/components/chip/chip.scss
brandyscarney: What color functions?

core/src/themes/themes.interfaces.ts
```diff
@@ -1,4 +1,5 @@
-import type { IonChipConfig, IonChipRecipe } from '../components/chip/chip.interfaces';
+import type { IonChipRecipe, IonChipConfig } from '../components/chip/chip.interfaces';
```
brandyscarney:
```suggestion
import type { IonChipConfig, IonChipRecipe } from '../components/chip/chip.interfaces';
```

Revert this change

core/src/themes/ios/default.tokens.ts
```diff
@@ -131,7 +128,7 @@ export const defaultTheme: DefaultTheme = {
           minHeight: 'var(--ion-scaling-md)',
 
           font: {
-            size: clamp('13px', `${(fontSizes.chipBase / fontSizes.root).toFixed(2)}rem`, '22px'),
+            size: clamp('13px', `${(components.chip.font.size / global.root).toFixed(2)}rem`, '22px'),
```
brandyscarney, line 131: Why are we making changes to chip in this PR?

core/src/themes/ios/default.tokens.ts
```diff
@@ -405,6 +402,174 @@ export const defaultTheme: DefaultTheme = {
       },
     },
 
+    IonItemDivider: {
+      background: `var(--ion-background-color-step-100, ${mix(baseColors.black, baseColors.white, '90%')})`,
```
brandyscarney, line 406: We really need to have these stepped colors always defined so we aren't having to add fallbacks like this.

core/src/components/item-divider/item-divider.scss
brandyscarney: I went all of the way back to the [v1 docs](https://ionicframework.com/docs/v1/components/#item-dividers), checked the [v3 docs](https://ionicframework.com/docs/v3/components/#list-dividers), and checked the [v4 docs](https://ionicframework.com/docs/v4/api/item-divider#usage) and we have never shown code using headings in item divider so I don't think we need to keep these. It was probably just copied over from another component.

core/src/components/thumbnail/thumbnail.scss
brandyscarney: Should we define a size in the base tokens?

core/src/components/item-divider/item-divider.scss
```diff
@@ -1,46 +1,171 @@
...
-  --inner-padding-bottom: 0px;
-  --inner-padding-start: 0px;
-
-  @include font-smoothing();
-  @include margin(0);
-  @include padding(var(--padding-top), null, var(--padding-bottom), null);
+  @include mixins.font-smoothing();
+  @include mixins.margin(0);
+  @include mixins.padding(var(--ion-item-divider-padding-top), null, var(--ion-item-divider-padding-bottom), null);
+  @include mixins.border-radius(0);
```
brandyscarney: Do we need this rule? Isn't this the [default](https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Properties/border-radius#formal_definition)?

```suggestion
```

core/src/components/item-divider/item-divider.scss
```diff
@@ -1,46 +1,171 @@
...
-  --inner-padding-top: 0px;
-  --inner-padding-end: 0px;
-  --inner-padding-bottom: 0px;
-  --inner-padding-start: 0px;
-
-  @include font-smoothing();
-  @include margin(0);
-  @include padding(var(--padding-top), null, var(--padding-bottom), null);
+  @include mixins.font-smoothing();
+  @include mixins.margin(0);
```
brandyscarney: Do we need this rule? Isn't this the [default](https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Properties/margin#formal_definition)?

```suggestion
```

core/src/components/item-divider/item-divider.scss
```diff
@@ -51,19 +176,27 @@
...
 
-  font-family: $font-family-base;
+  font-family: var(--ion-font-family, inherit);
+  font-size: var(--ion-item-divider-font-size);
+  font-weight: var(--ion-item-divider-font-weight);
 
   overflow: hidden;
-  z-index: $z-index-item-divider;
+
+  z-index: 100;
```
brandyscarney, line 139: Does this need to be customizable or global?

core/src/components/item-divider/item-divider.scss
```diff
@@ -51,19 +176,27 @@
...
   box-sizing: border-box;
 }
 
 :host(.ion-color) {
-  background: current-color(base);
-  color: current-color(contrast);
+  background: colors.current-color("base");
+  color: colors.current-color("contrast");
```
brandyscarney: Why do we need to pass these as strings?

core/src/components/item-divider/item-divider.interfaces.ts
brandyscarney: The change between `leading` and `first` and `trailing` and `last` is confusing to me. I'm also not sure about `edge` vs `anchor` but I don't know a better name right now.

core/src/themes/md/default.tokens.ts
thetaPC: Based on this [conversation](https://github.com/ionic-team/ionic-framework/pull/31009#discussion_r2967013015), Shane mentioned that it could be confusing for devs to have leading be meant for multiple reasons.

core/src/components/item-divider/item-divider.interfaces.ts
thetaPC: The naming for this was a difficult process as I tried to distinguish terms based on function and position. I'm open to any suggestions! The sooner we clarify this the sooner we can set up the structure for the other components.

core/src/components/item/test/dividers/item.e2e.ts-snapshots/item-dividers-diff-ios-ltr-Mobile-Chrome-linux.png
thetaPC: Ticket created

core/src/components/thumbnail/thumbnail.scss
thetaPC: I would have to create the tokens for thumbnail under the defaults files. I would prefer to avoid adding on more work to this ticket by introducing components that have not been migrated. If you prefer that I do it now, let me know.

core/src/components/thumbnail/thumbnail.scss
thetaPC: Yes, that is my end goal. As mentioned in this [comment](https://github.com/ionic-team/ionic-framework/pull/31009#discussion_r3001114175), I would like to do that when we work on thumbnail.

core/src/components/chip/chip.scss
thetaPC: Good catch! https://github.com/ionic-team/ionic-framework/pull/31009/changes/f2b63931ff8f0b9fd7a53c86243e5b0c02d50c73

core/src/themes/themes.interfaces.ts
thetaPC: https://github.com/ionic-team/ionic-framework/pull/31009/changes/8a9a4b4de814787e0d29dcb27f92c216295640d0

core/src/themes/ios/default.tokens.ts
thetaPC: I added a shared tokens file that updated how [font sizes were being retrieved](https://github.com/ionic-team/ionic-framework/pull/31009/changes#diff-6ca7a4458286b0082b92d16c636ba96611453af628501116b0b4e4ec86eed054R8-R17). This new structure applies to item divider and chip.

core/src/themes/ios/default.tokens.ts
thetaPC: Updated a Jira ticket to include this work.

core/src/components/item-divider/item-divider.scss
thetaPC: https://github.com/ionic-team/ionic-framework/pull/31009/changes/918ecbef3d4a1675dbd129fe983efa108cccd3af

core/src/components/item-divider/item-divider.scss
thetaPC: https://github.com/ionic-team/ionic-framework/pull/31009/changes/fc978e81596f43fbd7f7f3693ab13c960c814ecb

core/src/components/item-divider/item-divider.scss
thetaPC: https://github.com/ionic-team/ionic-framework/pull/31009/changes/6e66739077c5b1587523a61a09f35fad1a832e6d

core/src/components/item-divider/item-divider.scss
thetaPC: I don't think so. This z index has been used by all three themes without issues plus `$z-index-item-divider;` was only used in this file. That's why I also [removed](https://github.com/ionic-team/ionic-framework/pull/31009/changes#diff-8a62e041948d67b564093c7980635d4df2cd6e3428345d0cf2cbad7c2e00c220L66) it from globals.

core/src/components/item-divider/item-divider.scss
thetaPC: Good catch! https://github.com/ionic-team/ionic-framework/pull/31009/changes/ed648e0cfd6118e7cd0768c1de745a8190daae85

core/src/components/thumbnail/thumbnail.scss
brandyscarney: No it's fine we can do that as part of the thumbnail migration I was just curious what the blocker was.

core/src/components/item-divider/item-divider.interfaces.ts
brandyscarney: I'm kind of confused why this rule exists:

```scss
// Item Divider Slots

:host([slot="start"]) {
  @include mixins.margin(
    var(--ion-item-divider-leading-anchor-margin-top, revert-layer),
    var(--ion-item-divider-leading-anchor-margin-end, revert-layer),
    var(--ion-item-divider-leading-anchor-margin-bottom, revert-layer),
    var(--ion-item-divider-leading-anchor-margin-start, revert-layer)
  );
}
```

This is only applied when the item-divider itself is in the `"start"` `slot`. When does this use case happen?

```html
<ion-item>
  <ion-item-divider slot="start">
    Item Divider in Start slot
  </ion-item-divider>
  <ion-label>Item</ion-label>
</ion-item>
```

I searched and did not see any examples of this. 

Besides that, I was mostly just thinking `leading` and `trailing` were better than `firstChild` and `lastChild`.

In the cases of slots, I think we need to rethink these names. The naming should probably apply more to what it is actually being applied to. Like this:

```ts
start?: {
  // Targets `:host([slot="start"])`
  host?: {
    margin?: IonMargin;
  };

  // Targets `::slotted([slot="start"])`
  slotted?: {
    margin?: IonMargin;
  };
};

end?: {
  // Targets `:host([slot="end"])`
  host?: {
    margin?: IonMargin;
  };

  // Targets `::slotted([slot="end"])`
  slotted?: {
    margin?: IonMargin;
  };
};
```

which becomes:

```scss
// Item Divider Slots

:host([slot="start"]) {
  @include mixins.margin(
    var(--ion-item-divider-start-host-margin-top, revert-layer),
    var(--ion-item-divider-start-host-margin-end, revert-layer),
    var(--ion-item-divider-start-host-margin-bottom, revert-layer),
    var(--ion-item-divider-start-host-margin-start, revert-layer)
  );
}

// Item Divider Slotted Elements

// Slotted Start
::slotted([slot="start"]) {
  @include mixins.margin-horizontal(
    var(--ion-item-divider-start-slotted-margin-start, revert-layer),
    var(--ion-item-divider-start-slotted-margin-end, revert-layer)
  );
}

// Slotted End
::slotted([slot="end"]) {
  @include mixins.margin-horizontal(
    var(--ion-item-divider-end-slotted-margin-start, revert-layer),
    var(--ion-item-divider-end-slotted-margin-end, revert-layer)
  );
}
```

Updating `icon` to match would be:

```ts
start?: {
  // Targets `:host([slot="start"])`
  host?: {
    margin?: IonMargin;
  };

  // Targets `::slotted([slot="start"])`
  slotted?: {
    margin?: IonMargin;
  };
};

end?: {
  // Targets `::slotted([slot="end"])`
  slotted?: {
    margin?: IonMargin;
  };
};

icon?: {
  font?: {
    size?: string;
  };

  start?: {
    // Targets `::slotted([slot="start"])`
    slotted?: {
      margin?: IonMargin;
    };
  };

  end?: {
    // Targets `::slotted([slot="end"])`
    slotted?: {
      margin?: IonMargin;
    };
  };
```

Or we could use `content` instead of `slotted`? Let me know your thoughts.

thetaPC: https://github.com/ionic-team/ionic-framework/pull/31009/changes/faf287aa53b7afdd872c9c29f10c0fb82845fb19 and https://github.com/ionic-team/ionic-framework/pull/31009/changes/d676a9121167922e7a5a7b3d07c64595f0991cc6

core/src/components/item-divider/item-divider.scss
```diff
@@ -1,46 +1,117 @@
...
+   * @prop --ion-item-divider-start-slotted-margin-start: Left margin if direction is left-to-right, and right margin if direction is right-to-left of the slotted element when it is in the start slot of the item divider
+   * @prop --ion-item-divider-start-slotted-margin-end: Right margin if direction is left-to-right, and left margin if direction is right-to-left of the slotted element when it is in the start slot of the item divider
+   *
+   * ::slotted([slot="end"])
+   * @prop --ion-item-divider-end-slotted-margin-start: Left margin if direction is left-to-right, and right margin if direction is right-to-left of the slotted element when it is in the end slot of the item divider
+   * @prop --ion-item-divider-end-slotted-margin-end: Right margin if direction is left-to-right, and left margin if direction is right-to-left of the slotted element when it is in the end slot of the item divider
+   *
+   * Slotted Label
+   * @prop --ion-item-divider-label-margin-top: Top margin of the slotted ion-label
+    * @prop --ion-item-divider-label-margin-end: Right margin if direction is left-to-right, and left margin if direction is right-to-left of the slotted ion-label
```
brandyscarney: The indentation starts to be off here.

PR #31014, opened 2026-03-17
feat(spinner): add recipe and tokens

thetaPC:
Issue number: internal

---------

Component styles for `ion-spinner` are fragmented across multiple files (native and `ionic`). The `md` and `ios` theme uses the native styles. Developers were restricted to those themes and how those themes structured the logic and styles.

## What is the new behavior?

- **Unified Style Architecture**: Combined theme-specific styling into a single `spinner.scss` file that consumes CSS variables, ensuring a single source of truth for component logic.
- **Property First model**: based the recipe on the properties
- **Updated TypeScript Interface**: Changed the filename to `spinner.interfaces.ts`
- **Defined Theme Defaults**: Added defaults that are being used per theme
- **Standardized Sizing**: Introduced sizes to `md` and `ios`.
- **Updated Test Pages**: Updated a couple of test pages to include more cases
- **Updated Tests**: Updated tests to include more cases
- **Contextual Coloring**: Implemented `color: currentColor` for slotted spinners within buttons. This ensures the spinner's visual weight matches the button's text/icon color dynamically across all interaction states (hover, active, disabled) without additional token mapping.

- [x] Yes
- [ ] No

This PR introduces a breaking change to how `IonSpinner` is styled. Existing manual CSS overrides targeting internal spinner structures or old token names will no longer work due to the shift to the new token hierarchy and a unified base SCSS file. 

**Migration Path:**
1.  **Token Updates**: Update any custom theme configurations to match the new nested structure.

2.  **CSS Overrides/Removals**: Ensure selectors align with the new slotted element logic and variable names (e.g., `--ion-spinner-color`). CSS variables like `--color` have been removed in favor of the new variables.

`--color` should no longer be used. Setting the value, `IonSpinner.color`, within the tokens file should be used to change the color. If the color needs to be set per component, then `--ion-spinner-color` should be used.

3.  **Theme classes**: Remove any instances that target the theme classes: `ion-spinner.md`

4.  **Spinner name class**: The class have been updated to include the term `name`.

`.spinner-lines`, `.spinner-circular`, etc should be updated to `.spinner-name-lines`, `.spinner-name-circular`, etc.

## Other information

Previews:
- [Basic](https://ionic-framework-git-fw-6860-ionic1.vercel.app/src/components/spinner/test/basic/)
- [Color](https://ionic-framework-git-fw-6860-ionic1.vercel.app/src/components/spinner/test/color/)
- [Size](https://ionic-framework-git-fw-6860-ionic1.vercel.app/src/components/spinner/test/size/)

### DIFF
--- a/core/src/components/chip/chip.scss
+++ b/core/src/components/chip/chip.scss
@@ -14,7 +14,7 @@
    * @prop --ion-chip-padding-top: Top padding of the chip
    * @prop --ion-chip-padding-end: Right padding if direction is left-to-right, and left padding if direction is right-to-left of the chip
    * @prop --ion-chip-padding-bottom: Bottom padding of the chip
-   * @prop --ion-chip-padding-start: Left padding if direction is left-to-right, and right margin if direction is right-to-left of the chip
+   * @prop --ion-chip-padding-start: Left padding if direction is left-to-right, and right padding if direction is right-to-left of the chip
    *
    * @prop --ion-chip-font-weight: Font weight of the chip
    * @prop --ion-chip-letter-spacing: Letter spacing of the chip
@@ -102,25 +102,25 @@
    * @prop --ion-chip-icon-color: Color of the slotted `ion-icon`
    * @prop --ion-chip-icon-font-size: Font size of the slotted `ion-icon`
    * @prop --ion-chip-icon-leading-margin-top: Top margin of the slotted `ion-icon` when it is the first child
-   * @prop --ion-chip-icon-leading-margin-end: End margin of the slotted `ion-icon` when it is the first child
+   * @prop --ion-chip-icon-leading-margin-end: Right margin if direction is left-to-right, and left margin if direction is right-to-left of the slotted `ion-icon` when it is the first child
    * @prop --ion-chip-icon-leading-margin-bottom: Bottom margin of the slotted `ion-icon` when it is the first child
-   * @prop --ion-chip-icon-leading-margin-start: Start margin of the slotted `ion-icon` when it is the first child
+   * @prop --ion-chip-icon-leading-margin-start: Left margin if direction is left-to-right, and right margin if direction is right-to-left of the slotted `ion-icon` when it is the first child
    * @prop --ion-chip-icon-trailing-margin-top: Top margin of the slotted `ion-icon` when it is the last child
-   * @prop --ion-chip-icon-trailing-margin-end: End margin of the slotted `ion-icon` when it is the last child
+   * @prop --ion-chip-icon-trailing-margin-end: Right margin if direction is left-to-right, and left margin if direction is right-to-left of the slotted `ion-icon` when it is the last child
    * @prop --ion-chip-icon-trailing-margin-bottom: Bottom margin of the slotted `ion-icon` when it is the last child
-   * @prop --ion-chip-icon-trailing-margin-start: Start margin of the slotted `ion-icon` when it is the last child
+   * @prop --ion-chip-icon-trailing-margin-start: Left margin if direction is left-to-right, and right margin if direction is right-to-left of the slotted `ion-icon` when it is the last child
    *
    * Slotted Avatar
    * @prop --ion-chip-avatar-width: Width of the slotted `ion-avatar`
    * @prop --ion-chip-avatar-height: Height of the slotted `ion-avatar`
    * @prop --ion-chip-avatar-leading-margin-top: Top margin of the slotted `ion-avatar` when it is the first child
-   * @prop --ion-chip-avatar-leading-margin-end: End margin of the slotted `ion-avatar` when it is the first child
+   * @prop --ion-chip-avatar-leading-margin-end: Right margin if direction is left-to-right, and left margin if direction is right-to-left of the slotted `ion-avatar` when it is the first child
    * @prop --ion-chip-avatar-leading-margin-bottom: Bottom margin of the slotted `ion-avatar` when it is the first child
-   * @prop --ion-chip-avatar-leading-margin-start: Start margin of the slotted `ion-avatar` when it is the first child
+   * @prop --ion-chip-avatar-leading-margin-start: Left margin if direction is left-to-right, and right margin if direction is right-to-left of the slotted `ion-avatar` when it is the first child
    * @prop --ion-chip-avatar-trailing-margin-top: Top margin of the slotted `ion-avatar` when it is the last child
-   * @prop --ion-chip-avatar-trailing-margin-end: End margin of the slotted `ion-avatar` when it is the last child
+   * @prop --ion-chip-avatar-trailing-margin-end: Right margin if direction is left-to-right, and left margin if direction is right-to-left of the slotted `ion-avatar` when it is the last child
    * @prop --ion-chip-avatar-trailing-margin-bottom: Bottom margin of the slotted `ion-avatar` when it is the last child
-   * @prop --ion-chip-avatar-trailing-margin-start: Start margin of the slotted `ion-avatar` when it is the last child
+   * @prop --ion-chip-avatar-trailing-margin-start: Left margin if direction is left-to-right, and right margin if direction is right-to-left of the slotted `ion-avatar` when it is the last child
    */
 
   @include mixins.font-smoothing();


core/src/components/spinner/spinner.scss
thetaPC, line 1: The `ionic` theme for spinner [overwrites the medium](https://github.com/ionic-team/ionic-framework/blob/966d6b230ea38aa9fb10dfc2fdcef89cc51580ba/core/src/components/spinner/spinner.ionic.scss#L14-L16) since it's one of those components that differs to the set value of medium. I decided to not implement it since it's specific to the project. A workaround would be to set it:

```css
ion-spinner {
  --ion-color-medium-bold: #626262;
}
```

core/src/components/spinner/test/basic/spinner.e2e.ts-snapshots/spinner-basic-ios-ltr-Mobile-Chrome-linux.png
thetaPC, line 1: Only removed the word `diff` from the test file name to line up with the naming conventions of the other components.

core/src/components/spinner/test/color/spinner.e2e.ts-snapshots/spinner-color-diff-md-ltr-Mobile-Chrome-linux.png
thetaPC, line 1: Only removed the word `diff` from the test file name to line up with the naming conventions of the other components.

core/src/components/spinner/test/resize/spinner.e2e.ts-snapshots/spinner-resize-md-ltr-Mobile-Chrome-linux.png
thetaPC, line 1: Only removed the word `diff` from the test file name to line up with the naming conventions of the other components.

core/src/components/spinner/spinner-interface.ts
thetaPC, line 1: This was renamed, not sure why it's not showing as such.

core/src/components/spinner/test/color/index.html
brandyscarney, line 1: Why did you remove the examples of different spinners with color? I believe those existed to show the color properly applied to all spinner types.

core/src/components/spinner/test/color/spinner.e2e.ts
brandyscarney, line 1: Same question:

Why did you remove the examples of different spinners with color? I believe those existed to show the color properly applied to all spinner types.

core/src/components/spinner/spinner.interfaces.ts
```diff
@@ -0,0 +1,78 @@
...
+
+export interface SpinnerConfig {
+  dur: number;
+  circles?: number;
+  lines?: number;
+  elmDuration?: boolean;
+  fn: (dur: number, index: number, total: number) => SpinnerData;
+}
+
+interface SpinnerData {
```
brandyscarney, line 13:
```suggestion
export interface SpinnerData {
```

core/src/themes/themes.interfaces.ts
```diff
@@ -1,4 +1,5 @@
 import type { IonChipRecipe, IonChipConfig } from '../components/chip/chip.interfaces';
+import type { IonSpinnerRecipe, IonSpinnerConfig } from '../components/spinner/spinner.interfaces';
```
brandyscarney: Nit: can we alphabetize imports like:

```suggestion
import type { IonChipConfig, IonChipRecipe } from '../components/chip/chip.interfaces';
import type { IonSpinnerConfig, IonSpinnerRecipe } from '../components/spinner/spinner.interfaces';
```

core/src/components/spinner/spinner.tsx
```diff
@@ -54,9 +49,9 @@ export class Spinner implements ComponentInterface {
    * Set to `"large"` for a large size.
    * Set to `"xlarge"` for the largest size.
    *
-   * Defaults to `"xsmall"` for the `ionic` theme, undefined for all other themes.
+   * Defaults to `"medium"`.
```
brandyscarney: Isn't this only the default for md and ios themes?

core/src/components/spinner/spinner.scss
```diff
@@ -45,86 +42,122 @@ svg {
   transform: translateZ(0);
 }
 
-// Spinner: lines / lines-small / lines-sharp / lines-sharp-small
+// Spinner Names
```
brandyscarney, line 64: Should we document the change in class names as a breaking change?

core/src/components/spinner/spinner.tsx
```diff
@@ -67,29 +62,21 @@ export class Spinner implements ComponentInterface {
...
-    // TODO(ROU-10912): Remove theme check when sizes are defined for all themes.
-    if (theme !== 'ionic') {
-      return undefined;
-    }
-
-    if (size === undefined) {
-      return 'xsmall';
-    }
+  /**
+   * Set the size based on the custom theme config
```
brandyscarney: Shouldn't this be _get_ the size, and also it's not just based on the config but also whether or not its set on the component.

core/src/components.d.ts
```diff
@@ -77,6 +78,7 @@ export { SegmentViewScrollEvent } from "./components/segment-view/segment-view-i
 export { SelectChangeEventDetail, SelectCompareFn, SelectInterface } from "./components/select/select-interface";
 export { SelectModalOption } from "./components/select-modal/select-modal-interface";
 export { SelectPopoverOption } from "./components/select-popover/select-popover-interface";
+export { IonSpinnerSize } from "./components/spinner/spinner.interfaces";
```
brandyscarney: We may need to rethink this name since it's exported here. If you look at all of the others ones they do not have `Ion` prefixed.

core/api.txt
brandyscarney, line 1: The PR description says:

> CSS Overrides: Ensure selectors align with the new slotted element logic and variable names (e.g., `--ion-spinner-color`).
>
> `--color` should no longer be used. Setting the value, `IonSpinner.color`, within the tokens file should be used to change the color.

But we still have this variable, why?

core/src/components/spinner/spinner.interfaces.ts
```diff
@@ -0,0 +1,78 @@
+export interface SpinnerConfigs {
+  [spinnerName: string]: SpinnerConfig;
+}
+
+export interface SpinnerConfig {
```
brandyscarney: Should we rename this? It's kind of confusing when we have `IonSpinnerConfig` also.

core/src/themes/ionic/default.tokens.ts
```diff
@@ -261,5 +265,73 @@ export const defaultTheme: DefaultTheme = {
...
+            stroke: {
+              width: 'var(--ion-scaling-100)',
+            },
+          },
+        },
+      },
+
+      circular: {
+        stroke: {
+          width: '5.6',
```
brandyscarney, line 298: Should this be a token or calculated somehow?

core/src/themes/ionic/default.tokens.ts
```diff
@@ -261,5 +265,73 @@ export const defaultTheme: DefaultTheme = {
         },
       },
     },
+
+    IonSpinner: {
+      color: '#626262',
```
brandyscarney, line 270: Shouldn't this use a token or are you waiting on the text / gray colors to be added? 

iOS has:

```
color: 'var(--ion-text-color, #000)',
```

core/src/themes/ios/default.tokens.ts
```diff
@@ -400,5 +404,73 @@ export const defaultTheme: DefaultTheme = {
         },
       },
     },
+
+    IonSpinner: {
+      color: 'var(--ion-text-color, #000)',
+
+      lines: {
+        stroke: {
+          width: '7px',
```
brandyscarney, line 413: Do we need to add variables for this, or should we use `border-width` ones?

core/src/themes/md/default.tokens.ts
```diff
@@ -397,5 +401,73 @@ export const defaultTheme: DefaultTheme = {
         },
       },
     },
+
+    IonSpinner: {
+      color: 'var(--ion-text-color, #000)',
+
+      lines: {
+        stroke: {
+          width: '7px',
```
brandyscarney, line 410: Same question as ios

core/src/components/spinner/test/color/index.html
thetaPC: I didn't see how colors and names needed to be tied together for a test. I would expect that it would work for all names regardless. But I've added them back: https://github.com/ionic-team/ionic-framework/pull/31014/changes/5f0020c30137a24c5260baa2523f3af4a6cfae24

core/src/components/spinner/test/color/spinner.e2e.ts
thetaPC: https://github.com/ionic-team/ionic-framework/pull/31014#discussion_r2962996515

core/src/components/spinner/spinner.interfaces.ts
thetaPC: `SpinnerData` isn't being used anywhere outside of this file so we don't need to export it.

core/src/themes/themes.interfaces.ts
thetaPC: https://github.com/ionic-team/ionic-framework/pull/31014/changes/1179d40c063728873cd6a7b802285e70052b9bfb

core/src/components/spinner/spinner.tsx
thetaPC: Yes, I chose `medium` because that one has the values that are currently being used by the community. I also changed the [default](https://github.com/ionic-team/ionic-framework/blob/5f0020c30137a24c5260baa2523f3af4a6cfae24/core/src/themes/ionic/default.tokens.ts#L30) for `ionic` to be `xsmall` through their config.

core/src/components/spinner/spinner.scss
thetaPC: Updated, it's now number four under the breaking change section of the PR description.

core/src/components/spinner/spinner.tsx
thetaPC: Updated it for chip too: https://github.com/ionic-team/ionic-framework/pull/31014/changes/4f5ee31a14c06a6493fe33d43caabf8e038f94ca

core/src/components.d.ts
thetaPC: https://github.com/ionic-team/ionic-framework/pull/31014/changes/4c1c6d2b1f9f085f1638971220b0ca5d2924bbce

core/api.txt
thetaPC: Because we [discussed](https://github.com/ionic-team/ionic-framework/pull/30873#discussion_r2829797452) that it would be revisited as we add more components. Should I just remove them?

core/src/components/spinner/spinner.interfaces.ts
thetaPC: Thoughts on this?: https://github.com/ionic-team/ionic-framework/pull/31014/changes/ee535287322e31a34b6a757ac0e7d3216f9bb783

core/src/themes/ionic/default.tokens.ts
thetaPC: Nope, it's [exactly the same value](https://github.com/ionic-team/ionic-framework/blob/cbfe7cce3be345eacbf9fe29e74438a927c16679/core/src/components/spinner/spinner.scss#L135) set for native.

core/src/themes/ionic/default.tokens.ts
thetaPC: No, this is due to `ionic` currently using their own [tokens](https://github.com/ionic-team/ionic-framework/blob/a877bb711ddd55c5c64b763a9ee58b87ba850f9c/core/src/components/spinner/spinner.ionic.scss#L8). Since we are no longer using them, I just grabbed the [raw value](https://github.com/ionic-team/ionic-framework/blob/ee535287322e31a34b6a757ac0e7d3216f9bb783/core/src/foundations/ionic.vars.scss#L16) that represent that token. They can always create their own theme and use their tokens.

core/src/themes/ios/default.tokens.ts
thetaPC: `border-width` doesn't have a 7px value. Even if it did, I still would expect to use `scaling` since the stroke width targets a svg not a border.

I don't think we should add a variable to account for 7px because it would break the [current implementation](https://github.com/ionic-team/ionic-framework/blob/ee535287322e31a34b6a757ac0e7d3216f9bb783/core/src/themes/base/default.tokens.ts#L53) we have for tokens like scaling. We are incrementing the keys by 50 so odd values aren't supported.

core/src/themes/md/default.tokens.ts
thetaPC: https://github.com/ionic-team/ionic-framework/pull/31014#discussion_r2963125215

core/src/components/spinner/test/color/index.html
brandyscarney: The different spinners might apply colors in different ways so it just makes sure it always applies properly.

core/src/components/spinner/test/color/spinner.e2e.ts
brandyscarney: You should add `paused` back to the e2e test so the screenshot doesn't look different every time it runs.

core/src/themes/ios/default.tokens.ts
brandyscarney: Maybe we should consider updating these to use a token then? Like 8px instead of 7px.

core/src/themes/ionic/default.tokens.ts
brandyscarney: Gotcha. I think we'll probably want to update this later when we have our primitives added but that's a future issue.

core/api.txt
brandyscarney: Oh I thought that was just about the focus variables. I don't see a reason they can't override the `--color` on components directly? Developers might want it to be different than the global theme.

core/src/components/spinner/spinner.tsx
brandyscarney: Oh I see it defaults to medium if the config doesn't define it:

```tsx
const sizeConfig = config.getObjectValue('IonSpinner.size', 'medium') as string;
```

I was missing that part.

core/src/components/spinner/spinner.tsx
brandyscarney: Can we reword this something like:

```tsx
/**
 * Gets the spinner size. Uses the `size` property if set, otherwise
 * checks the theme config and falls back to 'medium' if neither is provided.
 */
```

Could replace `'medium'` with `the default` if desired.

core/src/components/spinner/spinner.interfaces.ts
brandyscarney: What about `SpinnerDefinition(s)`?

core/src/components/spinner/spinner.scss
brandyscarney: We also need to add these to the breaking changes document: https://github.com/ionic-team/ionic-framework/blob/FW-6860/BREAKING.md#version-9x

core/src/components/spinner/test/color/spinner.e2e.ts
thetaPC: Weirdly that doesn't cause issues. `main` [doesn't](https://github.com/ionic-team/ionic-framework/blob/5657e7d064b31dd54cb4a50162564cff92b144d3/core/src/components/spinner/test/color/index.html) have `paused` in the [tests](https://github.com/ionic-team/ionic-framework/blob/5657e7d064b31dd54cb4a50162564cff92b144d3/core/src/components/spinner/test/color/spinner.e2e.ts).

core/src/components/spinner/spinner.tsx
thetaPC: https://github.com/ionic-team/ionic-framework/pull/31014/changes/801994195ab1473e6d4bcef3eff25a774ee2c5ed

core/api.txt
thetaPC: Then they can override them per component by using `--ion-spinner-color` instead of `--color`. I ended up removing the old CSS variables since having both might be confusing of when to use them, the same for chip: https://github.com/ionic-team/ionic-framework/pull/31014/changes/0aef08c72159d0ffe1669d66d98901682cc46e0e and https://github.com/ionic-team/ionic-framework/pull/31014/changes/7e1f630c8aa9cd86ed16c6293d50ffcb7ea946f9

core/src/components/chip/chip.scss
```diff
@@ -37,8 +25,6 @@
   align-items: center;
   justify-content: center;
 
-  color: var(--color);
```
thetaPC, line 40: This was never being applied. The color is done through the hue classes.

core/src/components/spinner/spinner.interfaces.ts
thetaPC: https://github.com/ionic-team/ionic-framework/pull/31014/changes/cb8032109759f6046ae4ac628a47bd53711e1bd2

core/src/themes/ios/default.tokens.ts
thetaPC: It's possible, but it might be good to just leave them as is for now so it's easier to compare what the values should be plus there would be less snapshots to review. It would keep the entire process as quick as possible and we can always do another pass of what needs tokens at the end. Thoughts?

core/src/components/spinner/spinner.scss
```diff
@@ -45,86 +37,122 @@ svg {
...
 }
 
-// Spinner: dots
-// --------------------------------------------------
-
-:host(.spinner-dots) circle {
-  stroke-width: 0;
+// dots
+:host(.spinner-name-dots) circle {
+  stroke-width: var(--ion-spinner-dots-stroke-width);
```
ShaneK, line 126: I'm having a hard time finding where this is defined. Is it actually set anywhere yet?

core/src/components/spinner/spinner.tsx
ShaneK: This says "Defaults to medium" but the ionic theme overrides it to `xsmall` via the config. Should the JSDoc mention that the default can vary by theme? Something like "Defaults to medium unless overridden by theme configuration." Chip has the same situation so not a blocker, just wondering if we want to be more accurate here.

core/src/components/spinner/spinner.scss
```diff
@@ -45,86 +37,122 @@ svg {
...
+  height: var(--ion-spinner-size-xsmall-height);
+}
+
+:host(.spinner-size-small) {
+  width: var(--ion-spinner-size-small-width);
+  height: var(--ion-spinner-size-small-height);
+}
+
+:host(.spinner-size-medium) {
+  width: var(--ion-spinner-size-medium-width);
```
ShaneK, line 162: The old code had `width: 28px; height: 28px` as defaults on `:host`. Now dimensions come entirely from the token system with no fallback values. If the tokens aren't loaded for some reason, the spinner collapses to zero size. Is that expected on this branch, or should we add fallbacks like `var(--ion-spinner-size-medium-width, 28px)`?

core/src/components/spinner/spinner.scss
thetaPC: https://github.com/ionic-team/ionic-framework/pull/31014/changes/538874903f981303048bc333f19ee2cf13d99b45

BREAKING.md
```diff
@@ -38,12 +40,6 @@ This is a comprehensive list of the breaking changes introduced in the major ver
 
 - The properties `pull` and `push` have been deprecated and no longer work. A similar look can be achieved with the newly added property `order`.
 
-<h4 id="version-9x-radio-group">Radio Group</h4>
```
thetaPC, line 41: This was moved to the correct location. It was originally added right in the middle of the grid changes.

BREAKING.md
```diff
@@ -19,6 +19,8 @@ This is a comprehensive list of the breaking changes introduced in the major ver
   - [Card](#version-9x-card)
   - [Chip](#version-9x-chip)
   - [Grid](#version-9x-grid)
+  - [Radio Group](#version-9x-radio-group)
```
thetaPC: This was never added.

core/src/components/spinner/spinner.scss
thetaPC: https://github.com/ionic-team/ionic-framework/pull/31014/changes/3c5afb3a1b9b5e10744bb0e72adc3c466ac037d8

core/src/components/spinner/spinner.tsx
thetaPC: How does this sound? https://github.com/ionic-team/ionic-framework/pull/31014/changes/8668db43030e54661d0c72e5ad7649c11843ca4b

core/src/components/spinner/spinner.scss
thetaPC: Yes, I've been purposely been avoiding setting hard coded values as defaults. Developers wouldn't experience this if they use any of the in built themes. If they decide to make their own, then I would expect that they would know to set their own values either by using their own or copying it from one of the in built themes. If we find out that we should be setting defaults, then we can always come back to add them.

core/src/components/button/button.common.scss
```diff
@@ -208,6 +208,10 @@
   pointer-events: none;
 }
 
+::slotted(ion-spinner) {
+  color: currentColor;
```
thetaPC, line 212: This is needed because we are no longer using `--color`.

core/src/components/spinner/test/color/spinner.e2e.ts
brandyscarney: Oh that is weird. Maybe there isn't enough of a difference to cause issues.

core/src/themes/ios/default.tokens.ts
brandyscarney: Yeah definitely. We can revisit this later!

BREAKING.md
brandyscarney: This does exist on [next](https://github.com/ionic-team/ionic-framework/blob/next/BREAKING.md) was it dropped on `ionic-modular`?

BREAKING.md
```diff
@@ -140,4 +136,19 @@ To reorder two columns where column 1 has `size="9" push="3"` and column 2 has `
...
+
+<h4 id="version-9x-radio-group">Radio Group</h4>
+
+- Converted `ion-radio-group` to use [Shadow DOM](https://developer.mozilla.org/en-US/docs/Web/Web_Components/Using_shadow_DOM).<br/>
+If you were targeting the internals of `ion-radio-group` in your CSS, you will need to target the `supporting-text`, `helper-text` or `error-text` [Shadow Parts](https://ionicframework.com/docs/theming/css-shadow-parts) instead, or use the provided CSS Variables.<br/>
+Additionally, the `radio-group-wrapper` div element has been removed, causing slotted elements to be direct children of the `ion-radio-group`.
+
+<h4 id="version-9x-spinner">Spinner</h4>
+
+- Component CSS variables have been removed. The component now utilizes the centralized Ionic Theming system. Global updates should be managed via the theme tokens file, while component-specific overrides are handled through localized CSS variables.
```
brandyscarney, line 155: Shouldn't we document the same thing for Chip?

core/src/components/chip/chip.scss
```diff
@@ -5,19 +5,7 @@
 // --------------------------------------------------
 
 :host {
-  /**
-   * @prop --border-radius: Border radius of the chip
-   * @prop --color: Color of the chip
-   * @prop --focus-ring-color: Color of the focus ring
-   * @prop --focus-ring-style: Style of the focus ring
-   * @prop --focus-ring-width: Width of the focus ring
-   */
```
brandyscarney, line 14: How do we plan to document the new CSS variables? Are we going to add them here or somewhere else?

core/src/components/spinner/spinner.common.scss
```diff
@@ -1,32 +1,24 @@
...
+@use "../../themes/functions.color" as colors;
 
-// Spinners
+// Spinner: Common Styles
 // --------------------------------------------------
 
 :host {
-  /**
-   * @prop --color: Color of the spinner
-   */
```
brandyscarney, line 11: How do we plan to document the new CSS variables? Are we going to add them here or somewhere else?

BREAKING.md
thetaPC: Ah, it seems that we're not synced to `next` so it still has the old version. I reverted it so it can be fixed automatically in the next sync. https://github.com/ionic-team/ionic-framework/pull/31014/changes/e00d1f051bce2c2523732dbaa2d9f52ff593f5a1

BREAKING.md
thetaPC: https://github.com/ionic-team/ionic-framework/pull/31014/commits/062a7c7f43b5191c4acab3f9530c9c216156591b

core/src/components/chip/chip.scss
thetaPC: https://github.com/ionic-team/ionic-framework/pull/31014/changes/95b1b41356b05cccf234b700d5d36ffb6846cd95 and https://github.com/ionic-team/ionic-framework/pull/31014/changes/a8892a99cffe9f5738753b660d62a16bb7cbfc0d

core/src/components/spinner/spinner.common.scss
thetaPC: https://github.com/ionic-team/ionic-framework/pull/31014#discussion_r2978103272

core/src/components/chip/chip.scss
brandyscarney: Is this auto-generated? If not we should really find a way to do so lol.

thetaPC: No, it was painstakingly annoying 😭 I'll make a ticket for us to look into a way to generate these.

PR #31021, opened 2026-03-19
fix(datetime): days keep in focus after changing the month

os-davidlourenco:
Issue number: internal

---------

In Android native, when a user places a day in focus and then changes the month by dragging the month, the div previous on focus remains in focus in the new month.

## What is the new behavior?
After changing the month, the focus is removed from the active element and placed on the calendar

- [ ] Yes
- [ ] No

### DIFF
--- a/core/src/components/datetime/datetime.tsx
+++ b/core/src/components/datetime/datetime.tsx
@@ -1035,9 +1035,8 @@ export class Datetime implements ComponentInterface {
           }
 
           const activeEl = this.el.shadowRoot!.activeElement as HTMLElement | null;
-          if (activeEl) {
-            activeEl.blur();
-            calendarBodyRef.focus();
+          if (activeEl && activeEl.classList.contains('calendar-day')) {
+            (activeEl.closest('.calendar-body') as HTMLElement | null)?.focus();
           }
         });
       };


core/src/components/datetime/datetime.tsx
```diff
@@ -1033,6 +1033,12 @@ export class Datetime implements ComponentInterface {
           if (this.resolveForceDateScrolling) {
             this.resolveForceDateScrolling();
           }
+
+          const activeEl = document.activeElement as HTMLElement | null;
```
ShaneK:
```suggestion
          const activeEl = this.el.shadowRoot!.activeElement as HTMLElement | null;
```

Since we're in the shadow DOM, you need to access the active element like this. There's another example of this selection in this same file for the keyboard handler.

core/src/components/datetime/datetime.tsx
```diff
@@ -1033,6 +1033,12 @@ export class Datetime implements ComponentInterface {
...
+          if (activeEl) {
```
ShaneK: If you don't add a guard so that this only triggers if the currently focused element is a day, you're going to cause a focus change no matter what, absolutely destroying accessibility in the cases where somebody is using a11y tools to move through months with arrows or something.

You'll need the previously identified activeEl change to implement this fix.

os-davidlourenco: Addressed and changed approach, Thank you Shane

PR #31023, opened 2026-03-19
fix(accordion): update styles for ionic theme

OS-susmitabhowmik:

---------

There were several issues in the ionic theme which are resolved in this PR:
- The last item of an accordion had a divider
- The pressed state color did not match design specifications
-  Nested accordions did not have not have horizontal padding, causing the focus state to be different from the parent accordion

## What is the new behavior?

The following changes were made in the ionic theme:
- Removed the divider from the final accordion item 
- Update the pressed state color to match design specifications
- Add additional horizontal padding for nested accordion items

- [ ] Yes
- [x] No

## Other information
- [Basic Accordion Preview](https://ionic-framework-git-rou-11979-fix-accordian-styles-ionic1.vercel.app/src/components/accordion/test/basic?ionic:theme=ionic)
- [Accordion States Preview](https://ionic-framework-git-rou-11979-fix-accordian-styles-ionic1.vercel.app/src/components/accordion/test/states?ionic:theme=ionic)
- [Nested Accordion Preview](https://ionic-framework-git-rou-11979-fix-accordian-styles-ionic1.vercel.app/src/components/accordion/test/nested?ionic:theme=ionic)


core/src/components/accordion/accordion.tsx
```diff
@@ -199,6 +205,16 @@ export class Accordion implements ComponentInterface {
...
+
+    /**
+     * Ionic theme: mark header items so `item.ionic.scss` can set pressed tokens.
+     * We cannot use `:host-context(ion-accordion)` for this — that selector is dropped
+     * in WebKit/Firefox (see `mixins.scss` notes on `:host-context`), which breaks
+     * screenshot tests and real Safari.
+     */
+    if (getIonTheme(this) === 'ionic') {
+      ionItem.classList.add('accordion-header-item');
+    }
```
thetaPC: We need to figure out a solution without relying on the theme flag because we are preparing to move all the components over to Ionic Modular, which strips out all of these types of if statements.

OS-susmitabhowmik: I removed the check for the theme in this commit 3cd999dc473457202edebf1b5154214a45563627

core/src/components/accordion/test/nested/index.html
```diff
@@ -11,7 +11,7 @@
     <style>
       .grid {
         display: grid;
-        grid-template-columns: repeat(auto-fill, minmax(250px, 1fr));
+        grid-template-columns: repeat(auto-fill, minmax(350px, 1fr));
```
thetaPC, line 14: Was this change necessary?

core/src/components/item/item.ionic.scss
```diff
@@ -122,3 +122,10 @@ slot[name="end"]::slotted(*) {
 :host(.in-select-modal.ion-focused) .item-native {
   --border-radius: #{globals.$ion-border-radius-400};
 }
+
+// Item in Accordion
+// --------------------------------------------------
+// Ensure button element in accordion header has proper horizontal padding specifically for nested accordions
+:host(.accordion-header-item) .item-native {
+  @include globals.padding-horizontal(globals.$ion-space-400);
+}
```
thetaPC: This style doesn't seem to be applied. I commend this out locally and the tests still passed. After inspecting the element, this is already being [applied through the slot styles](https://github.com/ionic-team/ionic-framework/blob/ROU-11979-fix-accordian-styles/core/src/components/accordion/accordion.ionic.scss#L55-L58).

Regardless, any padding changes must be done through the slot styles that I linked.

```suggestion
```

core/src/components/accordion/accordion.tsx
```diff
@@ -199,6 +200,12 @@ export class Accordion implements ComponentInterface {
     if (ionItem.lines === undefined) {
       ionItem.lines = 'full';
     }
+
+    /**
+     * Add a class to identify this item as an accordion header
+     * for specific styling purposes
+     */
+    ionItem.classList.add('accordion-header-item');
```
thetaPC: This can be removed based on my [comment](https://github.com/ionic-team/ionic-framework/pull/31023/changes#r2977569970).

```suggestion
```

core/src/components/accordion/accordion.tsx
```diff
@@ -431,6 +438,12 @@ export class Accordion implements ComponentInterface {
 
     const shouldExpand = Array.isArray(value) ? value.includes(accordionValue) : value === accordionValue;
 
+    /**
+     * Check if this is the last accordion in the group.
+     */
```
thetaPC: This comment is not needed because:
1. The following line is not associated to the last accordion, it's about the `nextAccordion`
2. It's very obvious what `isLast` is about


```suggestion
```

core/src/components/accordion/test/states/accordion.e2e.ts
thetaPC, line 1: These changes can be reverted based on my [comment](https://github.com/ionic-team/ionic-framework/pull/31023/changes#r2977569970).

thetaPC (changes requested):

thetaPC:
@OS-susmitabhowmik Also please update the PR title to be shorter, thanks!

core/src/components/accordion/test/nested/index.html
OS-susmitabhowmik: I added this change because previously on larger screens the nested accordion was cut off

Screenshot 2026-03-23 at 3 16 56 PM

core/src/components/item/item.ionic.scss
OS-susmitabhowmik: It's applied within nested accordions so that the padding of the child accordions match the padding for the parent. The slot styles don't seem to work as expected, I think perhaps because the nested accordion items are within the shadow DOM. Removing these styles and class removes the horizontal padding for the nested items leading to inconsistencies in the focus styles between the parent and children. I also don't think we previously had tests for nested accordions with the ionic theme, I included them as a part of this commit a3a76f13ddaa38d6a6e0f2449df033305006f949.

Screenshot 2026-03-23 at 5 57 01 PM

core/src/components/accordion/accordion.tsx
OS-susmitabhowmik: Unfortunately I don't think we can remove this class as it breaks the styling for nested accordions. Please see my comment [here](https://github.com/ionic-team/ionic-framework/pull/31023#discussion_r2978484867) but I am happy to try any alternative suggestions!

core/src/components/accordion/accordion.tsx
OS-susmitabhowmik: Great catch, removed d67d6e59b5afb26f36fa3db22e8f3f37371911b6

core/src/components/accordion/test/states/accordion.e2e.ts
OS-susmitabhowmik: I replied above (https://github.com/ionic-team/ionic-framework/pull/31023#discussion_r2978484867), but I think we may need to keep the new class.

OS-susmitabhowmik:
> @OS-susmitabhowmik Also please update the PR title to be shorter, thanks!

@thetaPC Updated!

core/src/components/item/item.ionic.scss
thetaPC: Since the goal is to line up the nested accordion header items, then we can simply update the following [code](https://github.com/ionic-team/ionic-framework/blob/a3a76f13ddaa38d6a6e0f2449df033305006f949/core/src/css/ionic/core.ionic.scss#L330-L340) to:

```diff
ion-accordion > [slot="content"] ion-item {
  @include globals.typography(globals.$ion-body-md-regular);

  color: globals.$ion-primitives-neutral-1000;

-   --padding-start: 0;
  --padding-top: #{globals.$ion-space-300};
  --padding-bottom: #{globals.$ion-space-300};
-  --padding-end: 0;
  --min-height: #{globals.$ion-scale-700};
}

+ ion-accordion > [slot="content"] ion-item:not([slot="header"]) {
+  --padding-start: 0;
+  --padding-end: 0;
+ }
```

I was able to get the same result, I verified by running tests locally and they all pass.

Make sure to restart the server if you're running it else it doesn't apply the changes within the core styles.

thetaPC (changes requested):
Based off my [reply](https://github.com/ionic-team/ionic-framework/pull/31023#discussion_r2984582590), this can be done without adding a class after testing it locally. Please let me know if I'm misunderstanding the issue.

core/src/components/item/item.ionic.scss
OS-susmitabhowmik: Thanks @thetaPC, this works! I updated the styles and removed the additional class in this commit ee059a1e41cf8fb7b4dc34e5d173ce5d96e99bc8

core/src/components/accordion/accordion.tsx
OS-susmitabhowmik: Updated to remove this ee059a1e41cf8fb7b4dc34e5d173ce5d96e99bc8

core/src/components/accordion/test/states/accordion.e2e.ts
OS-susmitabhowmik: Removed ee059a1e41cf8fb7b4dc34e5d173ce5d96e99bc8

PR #31033, opened 2026-03-24
fix(datetime): scroll failing for adjacent days on ios

os-davidlourenco:

---------

When the user swipes to a previous month that has adjacent days and selects one of the adjacent days from a previous month, the datetime selects the day but fails to scroll to the previous month.

## What is the new behavior?
When a user clicks on a previous adjacent day after swiping the month, the calendar scrolls to the previous month. The same is applied to the next month.

- [ ] Yes
- [x] No

## Other information

The previous behavior was only reproducible on iOS real devices.

### DIFF
--- a/core/src/components/datetime/datetime.tsx
+++ b/core/src/components/datetime/datetime.tsx
@@ -1525,7 +1525,13 @@ export class Datetime implements ComponentInterface {
       return;
     }
 
-    nextMonth.scrollIntoView({ behavior: 'smooth' });
+    const left = (nextMonth as HTMLElement).offsetWidth * 2;
+
+    calendarBodyRef.scrollTo({
+      top: 0,
+      left: left * (isRTL(this.el) ? -1 : 1),
+      behavior: 'smooth',
+    });
   };
 
   private prevMonth = () => {
@@ -1538,7 +1544,14 @@ export class Datetime implements ComponentInterface {
     if (!prevMonth) {
       return;
     }
-    prevMonth.scrollIntoView({ behavior: 'smooth' });
+
+    const left = (prevMonth as HTMLElement).offsetWidth * 2;
+
+    calendarBodyRef.scrollTo({
+      top: 0,
+      left: left * (isRTL(this.el) ? 1 : -1),
+      behavior: 'smooth',
+    });
   };
 
   private toggleMonthAndYearView = () => {
--- a/core/src/components/datetime/test/show-adjacent-days/datetime.e2e.ts
+++ b/core/src/components/datetime/test/show-adjacent-days/datetime.e2e.ts
@@ -1,9 +1,6 @@
 import { expect } from '@playwright/test';
 import { configs, test } from '@utils/test/playwright';
 
-/**
- * This behavior does not vary across directions
- */
 configs({ directions: ['ltr'] }).forEach(({ title, screenshot, config }) => {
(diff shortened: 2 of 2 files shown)


core/src/components/datetime/datetime.tsx
```diff
@@ -1525,13 +1525,7 @@ export class Datetime implements ComponentInterface {
...
     }
 
-    const left = (nextMonth as HTMLElement).offsetWidth * 2;
-
-    calendarBodyRef.scrollTo({
-      top: 0,
-      left: left * (isRTL(this.el) ? -1 : 1),
-      behavior: 'smooth',
-    });
+    nextMonth.scrollIntoView({ behavior: 'smooth' });
```
ShaneK: There's a comment at line 915 that was written to document why `scrollIntoView` gets avoided here: "scrollIntoView() will scroll entire page if element is not in viewport. Use scrollLeft instead." The concern is that `scrollIntoView` walks all scrollable ancestors, so if the datetime is inside a scrollable `ion-content` or a sheet modal, clicking next/prev could scroll the whole page instead of just the calendar container. Was this tested with the datetime embedded in those contexts? If the root cause was `offsetWidth` returning 0 on iOS before layout, would `clientWidth` fix it while keeping the change within `calendarBodyRef.scrollTo()`?

ShaneK: Every other scroll calculation in this file uses `isRTL(this.el) ? -1 : 1` explicitly (lines 919, 1030, 1278). Was RTL on iOS tested with this change? `scrollIntoView` RTL behavior in scroll-snap containers on Safari has been patchy historically, and that's the platform where the original bug was found.

core/src/components/datetime/datetime.tsx
```diff
@@ -1544,12 +1538,7 @@ export class Datetime implements ComponentInterface {
     if (!prevMonth) {
       return;
     }
-
-    calendarBodyRef.scrollTo({
-      top: 0,
-      left: 0,
-      behavior: 'smooth',
-    });
+    prevMonth.scrollIntoView({ behavior: 'smooth' });
```
ShaneK: `scrollIntoView({ behavior: 'smooth' })` defaults to `block: 'start'`, which scrolls ancestors vertically to align the element to the top of each scrollable viewport. The old `scrollTo({ top: 0 })` was scoped to the calendar container. Adding `block: 'nearest'` would at least reduce the vertical scroll disturbance here. Also, the existing adjacent-days test only exercises the `nextMonth()` path (clicking a November adjacent day from October). Would it be possible to add a test that covers this fix directly? Something like: navigate back a month first, then click an adjacent day from the month before that, and assert the header title and `ionChange` value.

core/src/components/datetime/datetime.tsx
os-davidlourenco: After some test it is impossible to make the calendar work as expected with the scrollIntoView. This was reverted to the previous behavior

os-davidlourenco: Reverted to previous behavior

core/src/components/datetime/datetime.tsx
os-davidlourenco: To ensure this works in iOS, a calculation and approach similar to the next was developed in the previous. I also added a test to validate this behavior

PR #31034, opened 2026-03-24
feat(toolbar): add titlePlacement prop to control title position relative to content

brandyscarney:
Issue number: internal

---------

The `ionic` theme currently only supports aligning the toolbar title to the center.

## What is the new behavior?
- Adds `titlePlacement` property to toolbar
  - This property currently only works for the `ionic` theme as it is the only theme that supports the new center implementation
  - Set to `"start"` to place the title to the left in LTR and to the right in RTL
  - Set to `"center"` to center the title within the toolbar
  - Set to `"end"` to place the title to the right in LTR and to the left in RTL
- Adds an e2e test for `title-placement` which takes screenshots for the `ionic` theme only

- [ ] Yes
- [x] No

## Other information

[Preview](https://ionic-framework-git-fw-7054-ionic1.vercel.app/src/components/toolbar/test/title-placement?ionic:theme=ionic)

### DIFF
--- a/core/src/components.d.ts
+++ b/core/src/components.d.ts
@@ -4464,7 +4464,7 @@ export namespace Components {
          */
         "theme"?: "ios" | "md" | "ionic";
         /**
-          * Where to place the title relative to the other toolbar content. `"start"`: The title will appear to the left of the toolbar content in LTR and to the right in RTL. `"center"`: The title will appear in the center of the toolbar. `"end"`: The title will appear to the right of the toolbar content in LTR and to the left in RTL.
+          * Where to place the title relative to the other toolbar content. `"start"`: The title will appear to the left of the toolbar content in LTR and to the right in RTL. `"center"`: The title will appear in the center of the toolbar. `"end"`: The title will appear to the right of the toolbar content in LTR and to the left in RTL.  Only applies in the `ionic` theme.
          */
         "titlePlacement"?: 'start' | 'center' | 'end';
     }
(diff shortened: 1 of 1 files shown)


core/src/components/toolbar/toolbar.tsx
```diff
@@ -67,6 +83,24 @@ export class Toolbar implements ComponentInterface {
...
+   * Gets the title placement.
+   * Returns the title placement if it is set, otherwise returns `"center"`
+   * for `ionic` and `ios`, and `"start"` for `md`.
+   */
+  private getTitlePlacement(): 'start' | 'center' | 'end' {
+    if (this.titlePlacement !== undefined) {
+      return this.titlePlacement;
+    }
+
+    return getIonTheme(this) === 'ionic' || getIonTheme(this) === 'ios' ? 'center' : 'start';
```
brandyscarney, line 103: While this property doesn't work with `ios` or `md` yet, this will be needed eventually when we update them to work with it.

core/src/components/toolbar/toolbar.tsx
```diff
@@ -40,6 +48,14 @@ export class Toolbar implements ComponentInterface {
    */
   @Prop({ reflect: true }) color?: Color;
 
+  /**
+   * Where to place the title relative to the other toolbar content.
+   * `"start"`: The title will appear to the left of the toolbar content in LTR and to the right in RTL.
+   * `"center"`: The title will appear in the center of the toolbar.
+   * `"end"`: The title will appear to the right of the toolbar content in LTR and to the left in RTL.
+   */
+  @Prop() titlePlacement?: 'start' | 'center' | 'end';
```
brandyscarney, line 59: I chose to add `titlePlacement` to the toolbar because it controls all the positioning logic, but if we would prefer to add `placement` as a property on `ion-title` we could assign it there and pass it up to `ion-toolbar`.

core/src/components/title/test/basic/title.e2e.ts-snapshots/title-basic-ionic-md-ltr-light-Mobile-Chrome-linux.png
brandyscarney, line 1: I think this fixed a bug. The only reason the title wasn't centered in the two cases where it was left-aligned is because it had `size="small"` on it. I don't think this should change the alignment but I can change it back if desired.

core/src/components/toolbar/toolbar.tsx
ShaneK, line 59: Does the JSDoc need a note that this currently only has a visual effect in the `ionic` theme? Setting `titlePlacement="start"` on an `ios` toolbar won't do anything visually, but there's nothing in the docs to explain why. The `presentingElement` and `collapse` props both use "Only applies in iOS mode." for the same kind of restriction.

brandyscarney: I think we will need to update this to work for all themes before release but I have added it: https://github.com/ionic-team/ionic-framework/pull/31034/changes/9b001ab1fc5b99a98a035611a131e59335a965e7

PR #31036, opened 2026-03-25
feat(item-sliding): add automatic full expand animation of items

OS-pedrolourenco:
Issue number: internal

---------

Dragging an ion-item with an expandable option would cause a full swipe event to be triggered but also cause the item to return to a position just after the item options once the mouse button was released.

## What is the new behavior?
- Now, once the new threshold is passed and the mouse button is released, a full expand animation will take place and, after it is finished, only then is the full swipe event triggered and the item returns to its initial state where no options are visible.

- [ ] Yes
- [x] No

## Other information

[Preview](https://ionic-framework-git-rou-12664-ionic1.vercel.app/src/components/item-sliding/test/full-swipe/)

### DIFF
--- a/core/src/components/item-sliding/test/basic/item-sliding.e2e.ts
+++ b/core/src/components/item-sliding/test/basic/item-sliding.e2e.ts
@@ -14,8 +14,7 @@ configs({ modes: ['ios', 'md', 'ionic-md'] }).forEach(({ title, screenshot, conf
       await page.goto(`/src/components/item-sliding/test/basic`, config);
     });
     test.describe('start options', () => {
-      // TODO(FW-7184): remove skip once issue is resolved
-      test.skip('should not have visual regressions', async ({ page }) => {
+      test('should not have visual regressions', async ({ page }) => {
         const item = page.locator('#item2');
 
         /**
@@ -24,9 +23,9 @@ configs({ modes: ['ios', 'md', 'ionic-md'] }).forEach(({ title, screenshot, conf
          * Positive dragByX value to drag element from the left to the right
          * to reveal the options on the left side.
          */
-        const dragByX = config.direction === 'rtl' ? -150 : 150;
+        const dragByX = config.direction === 'rtl' ? -100 : 100;
 
-        await dragElementBy(item, page, dragByX);
+        await dragElementBy(item, page, dragByX, 0, undefined, undefined, true, 20);
         await page.waitForChanges();
 
         await expect(item).toHaveScreenshot(screenshot('item-sliding-start'));
@@ -45,7 +44,7 @@ configs({ modes: ['ios', 'md', 'ionic-md'] }).forEach(({ title, screenshot, conf
          */
         const dragByX = config.direction === 'rtl' ? 150 : -150;
 
-        await dragElementBy(item, page, dragByX);
+        await dragElementBy(item, page, dragByX, 0, undefined, undefined, true, 20);
 
         await expect(item).toHaveScreenshot(screenshot('item-sliding-end'));
       });
@@ -62,7 +61,7 @@ configs({ modes: ['md'], directions: ['ltr'] }).forEach(({ title, screenshot, co
       await page.goto(`/src/components/item-sliding/test/basic`, config);
       const item = page.locator('#item2');
 
-      await dragElementBy(item, page, -150);
+      await dragElementBy(item, page, -150, 0, undefined, undefined, true, 20);
       await page.waitForChanges();
 
       // item-sliding doesn't have an easy way to tell whether it's fully open so just screenshot it
@@ -109,8 +108,7 @@ configs({ modes: ['md'], directions: ['ltr'] }).forEach(({ title, screenshot, co
 configs({ modes: ['ios', 'md', 'ionic-md'] }).forEach(({ title, screenshot, config }) => {
   test.describe(title('item-sliding: basic'), () => {
     test.describe('safe area left', () => {
-      // TODO(FW-7184): remove skip once issue is resolved
-      test.skip('should have padding on the left only', async ({ page }) => {
+      test('should have padding on the left only', async ({ page }) => {
         await page.setContent(
           `
           <style>
@@ -143,7 +141,7 @@ configs({ modes: ['ios', 'md', 'ionic-md'] }).forEach(({ title, screenshot, conf
         const item = page.locator('ion-item-sliding');
 
         const dragByX = direction == 'rtl' ? -150 : 150;
-        await dragElementBy(item, page, dragByX);
+        await dragElementBy(item, page, dragByX, 0, undefined, undefined, true, 20);
         await page.waitForChanges();
 
         await expect(item).toHaveScreenshot(screenshot(`item-sliding-safe-area-left`));
@@ -151,8 +149,7 @@ configs({ modes: ['ios', 'md', 'ionic-md'] }).forEach(({ title, screenshot, conf
     });
 
     test.describe('safe area right', () => {
-      // TODO(FW-7184): remove skip once issue is resolved
-      test.skip('should have padding on the right only', async ({ page }) => {
+      test('should have padding on the right only', async ({ page }) => {
         await page.setContent(
           `
           <style>
@@ -185,7 +182,7 @@ configs({ modes: ['ios', 'md', 'ionic-md'] }).forEach(({ title, screenshot, conf
         const item = page.locator('ion-item-sliding');
 
         const dragByX = direction == 'rtl' ? 150 : -150;
-        await dragElementBy(item, page, dragByX);
+        await dragElementBy(item, page, dragByX, 0, undefined, undefined, true, 20);
         await page.waitForChanges();
 
         await expect(item).toHaveScreenshot(screenshot(`item-sliding-safe-area-right`));
--- a/core/src/components/item-sliding/test/icons/item-sliding.e2e.ts
+++ b/core/src/components/item-sliding/test/icons/item-sliding.e2e.ts
@@ -25,7 +25,7 @@ configs({ modes: ['ionic-md', 'ios', 'md'] }).forEach(({ title, screenshot, conf
          */
         const dragByX = config.direction === 'rtl' ? 150 : -150;
 
-        await dragElementBy(item, page, dragByX);
+        await dragElementBy(item, page, dragByX, 0, undefined, undefined, true, 20);
         await page.waitForChanges();
 
         // Convert camelCase ids to kebab-case for screenshot file names
--- a/core/src/utils/test/playwright/drag-element.ts
+++ b/core/src/utils/test/playwright/drag-element.ts
@@ -17,7 +17,8 @@ export const dragElementBy = async (
   dragByY = 0,
   startXCoord?: number,
   startYCoord?: number,
-  releaseDrag = true
+  releaseDrag = true,
+  steps?: number
 ) => {
   const boundingBox = await el.boundingBox();
 
@@ -36,7 +37,7 @@ export const dragElementBy = async (
   await page.mouse.down();
 
   // Drag the element.
-  await moveElement(page, startX, startY, dragByX, dragByY);
+  await moveElement(page, startX, startY, dragByX, dragByY, steps);
 
   if (releaseDrag) {
     await page.mouse.up();
@@ -115,8 +116,7 @@ const validateDragByY = (startY: number, dragByY: number, viewportHeight: number
   }
 };
 
-const moveElement = async (page: E2EPage, startX: number, startY: number, dragByX = 0, dragByY = 0) => {
-  const steps = 10;
+const moveElement = async (page: E2EPage, startX: number, startY: number, dragByX = 0, dragByY = 0, steps = 10) => {
   const browser = page.context().browser()!.browserType().name();
 
   const viewport = page.viewportSize();


core/src/components/item-sliding/test/full-swipe/index.html
brandyscarney, line 1: There is some weirdness going on with the headings in `ios`:

Image

core/src/components/item-sliding/test/full-swipe/index.html
```diff
@@ -0,0 +1,120 @@
...
+    <ion-app>
+      <ion-header>
+        <ion-toolbar>
+          <ion-title>Item Sliding - Full Swipe</ion-title>
+        </ion-toolbar>
+      </ion-header>
+
+      <ion-content>
+        <div class="ion-padding-start" style="padding-top: 30px">
+          <ion-title>Full Swipe - Expandable Options</ion-title>
```
brandyscarney: You shouldn't use `ion-title` outside of a toolbar, this is what's causing `ios` to look strange. Either use an `h2` like this:

```html
<h2>Full Swipe - Expandable Options</h2>

<style>
  h2 {
    font-size: 12px;
    font-weight: normal;

    color: #6f7378;

    margin-top: 10px;
    margin-left: 5px;
  }
</style>
```

Or use an `ion-list-header` at the top of the list.

core/src/components/item-sliding/test/basic/item-sliding.e2e.ts
```diff
@@ -14,7 +14,8 @@ configs({ modes: ['ios', 'md', 'ionic-md'] }).forEach(({ title, screenshot, conf
       await page.goto(`/src/components/item-sliding/test/basic`, config);
     });
     test.describe('start options', () => {
-      test('should not have visual regressions', async ({ page }) => {
+      //TODO(FW-7184): remove skip once issue is resolved
```
brandyscarney:
```suggestion
      // TODO(FW-7184): remove skip once issue is resolved
```

core/src/components/item-sliding/test/basic/item-sliding.e2e.ts
```diff
@@ -108,7 +109,8 @@ configs({ modes: ['md'], directions: ['ltr'] }).forEach(({ title, screenshot, co
 configs({ modes: ['ios', 'md', 'ionic-md'] }).forEach(({ title, screenshot, config }) => {
   test.describe(title('item-sliding: basic'), () => {
     test.describe('safe area left', () => {
-      test('should have padding on the left only', async ({ page }) => {
+      //TODO(FW-7184): remove skip once issue is resolved
```
brandyscarney:
```suggestion
      // TODO(FW-7184): remove skip once issue is resolved
```

core/src/components/item-sliding/test/basic/item-sliding.e2e.ts
```diff
@@ -149,7 +151,8 @@ configs({ modes: ['ios', 'md', 'ionic-md'] }).forEach(({ title, screenshot, conf
     });
 
     test.describe('safe area right', () => {
-      test('should have padding on the right only', async ({ page }) => {
+      //TODO(FW-7184): remove skip once issue is resolved
```
brandyscarney:
```suggestion
      // TODO(FW-7184): remove skip once issue is resolved
```

core/src/components/item-sliding/test/full-swipe/item-sliding.e2e.ts
```diff
@@ -0,0 +1,118 @@
...
+ * off-screen, fires ionSwipe, and returns to its closed position.
+ */
+
+// Full animation cycle duration (100ms expand + 250ms off-screen + 300ms delay + 250ms return)
+const FULL_ANIMATION_MS = 1100;
+
+configs({ modes: ['ios', 'md', 'ionic-md'], directions: ['ltr'] }).forEach(({ title, config }) => {
+  test.describe(title('item-sliding: full swipe'), () => {
+    test.beforeEach(async ({ page }) => {
+      await page.goto(`/src/components/item-sliding/test/full-swipe`, config);
```
brandyscarney: Can we break this into individual examples using `page.setContent`? Like this:

https://github.com/ionic-team/ionic-framework/blob/abe3b3f1001d096eaafe15eaf822effd2bf66e39/core/src/components/avatar/test/shape/avatar.e2e.ts#L10-L36

core/src/components/item-sliding/test/full-swipe/item-sliding.e2e.ts
```diff
@@ -0,0 +1,118 @@
...
+    });
+
+    test('should fire ionSwipe when expandable option is swiped fully (end side)', async ({ page }) => {
+      const ionSwipe = await page.spyOnEvent('ionSwipe');
+      const item = page.locator('#expandable-end');
+
+      await dragElementBy(item, page, -190);
+      await page.waitForTimeout(FULL_ANIMATION_MS);
+
+      expect(ionSwipe.length).toBeGreaterThan(0);
```
brandyscarney:
```suggestion
      expect(ionSwipe.length).toBe(1);
```

Can we check for exactly once?

core/src/components/item-sliding/test/full-swipe/item-sliding.e2e.ts
```diff
@@ -0,0 +1,118 @@
...
+    });
+
+    test('should fire ionSwipe when expandable option is swiped fully (start side)', async ({ page }) => {
+      const ionSwipe = await page.spyOnEvent('ionSwipe');
+      const item = page.locator('#expandable-start');
+
+      await dragElementBy(item, page, 190);
+      await page.waitForTimeout(FULL_ANIMATION_MS);
+
+      expect(ionSwipe.length).toBeGreaterThan(0);
```
brandyscarney:
```suggestion
      expect(ionSwipe.length).toBe(1);
```

Can we check for exactly once?

core/src/components/item-sliding/test/full-swipe/item-sliding.e2e.ts
```diff
@@ -0,0 +1,118 @@
...
+configs({ modes: ['ios', 'md', 'ionic-md'], directions: ['ltr', 'rtl'] }).forEach(({ title, config }) => {
+  test.describe(title('item-sliding: full swipe'), () => {
+    test('should fire ionSwipe in the correct swipe direction', async ({ page }) => {
+      await page.goto(`/src/components/item-sliding/test/full-swipe`, config);
+
+      const ionSwipe = await page.spyOnEvent('ionSwipe');
+      const item = page.locator('#expandable-end');
+
+      // In RTL the "end" side is on the left, revealed by dragging right
+      const dragByX = config.direction === 'rtl' ? 190 : -190;
```
brandyscarney: Why do we need a separate test for RTL? We should be able to use this on all the other tests.

core/src/components/item-sliding/item-sliding.tsx
```diff
@@ -383,14 +506,7 @@ export class ItemSliding implements ComponentInterface {
       restingPoint = 0;
     }
 
-    const state = this.state;
     this.setOpenAmount(restingPoint, true);
```
ShaneK, line 555: The original `onEnd` fired `fireSwipeEvent()` for any full swipe that reached `SwipeEnd`/`SwipeStart` state, regardless of whether any option was `expandable`. Now that event emission is entirely inside `animateFullSwipe`, which only runs when `hasExpandableOptions` is true, non-expandable option sets will never emit `ionSwipe`. Apps using `(ionSwipe)` on a plain `ion-item-options` will silently stop receiving it. Shouldn't the fall-through path (when `shouldTriggerFullSwipe` is false) still fire the event when state has `SwipeEnd`/`SwipeStart`?

core/src/components/item-sliding/item-sliding.tsx
```diff
@@ -248,6 +258,102 @@ export class ItemSliding implements ComponentInterface {
...
+
+      // Reset state
+      if (this.item) {
+        this.item.style.transition = '';
+      }
+      this.openAmount = 0;
+      this.state = SlidingState.Disabled;
+      openSlidingItem = undefined;
+    } finally {
+      el.classList.remove('item-sliding-full-swipe');
```
ShaneK: `openSlidingItem = undefined` is inside the `try` block (line 348) so if anything throws before reaching it, the module-level reference stays pointing to this element. That prevents any other sliding item in the list from opening until something eventually clears it. Shouldn't this move into `finally`?

core/src/components/item-sliding/item-sliding.tsx
```diff
@@ -248,6 +258,102 @@ export class ItemSliding implements ComponentInterface {
...
+  private animateToPosition(position: number, duration: number): Promise<void> {
+    return new Promise((resolve) => {
+      if (!this.item) {
+        return resolve();
+      }
+
+      this.item.style.transition = `transform ${duration}ms ease-out`;
+      this.item.style.transform = `translate3d(${-position}px, 0, 0)`;
+
+      setTimeout(() => {
```
ShaneK: These `setTimeout` IDs aren't stored anywhere, so `disconnectedCallback` can't cancel them. If the component unmounts mid-animation, `options.fireSwipeEvent()` still fires when the timer resolves since `options` is captured by the closure before `this.leftOptions`/`this.rightOptions` gets cleared. There's already a `this.tmr` pattern for exactly this. Could the timer IDs get the same treatment?

core/src/components/item-sliding/item-sliding.tsx
```diff
@@ -248,6 +258,102 @@ export class ItemSliding implements ComponentInterface {
...
+   * This is used when an expandable option is swiped beyond the threshold.
+   */
+  private async animateFullSwipe(direction: 'start' | 'end') {
+    // Prevent interruption during animation
+    if (this.gesture) {
+      this.gesture.enable(false);
+    }
+
+    const { el } = this;
+    el.classList.add('item-sliding-full-swipe');
```
ShaneK: Is there a CSS rule for `item-sliding-full-swipe` somewhere? I don't see one in `item-sliding.scss` or the platform-specific variants. If it's needed so the item can animate past the list's clipping boundary (e.g. `overflow: visible` on the host), the rule is missing. If it's a hook for external consumers, worth a comment. Otherwise it could be dropped.

core/src/components/item-sliding/test/full-swipe/item-sliding.e2e.ts
ShaneK: The codebase has `toHaveReceivedEvent()`, `toHaveReceivedEventTimes(1)`, and `not.toHaveReceivedEvent()` in `core/src/utils/test/playwright/matchers/`. Could you use those instead of `.length` checks? The non-expandable test on line 62 would become `expect(ionSwipe).not.toHaveReceivedEvent()` as well.

core/src/components/item-sliding/test/full-swipe/item-sliding.e2e.ts
```diff
@@ -0,0 +1,118 @@
...
+ * child components (ion-item-options, ion-item-option) have
+ * mode-specific styling, so we test across all modes.
+ *
+ * When an item has at least one expandable option and the user swipes
+ * beyond the threshold (or with sufficient velocity), the item slides
+ * off-screen, fires ionSwipe, and returns to its closed position.
+ */
+
+// Full animation cycle duration (100ms expand + 250ms off-screen + 300ms delay + 250ms return)
+const FULL_ANIMATION_MS = 1100;
```
ShaneK, line 15: `FULL_ANIMATION_MS = 1100` is the sum of the hardcoded delays in `animateFullSwipe` (100 + 250 + 300 + 250). If any of those change, these tests start silently flaking. For the tests that check `ionSwipe`, could you use `await ionSwipe.next()` instead? That resolves as soon as the event fires without sleeping for the full cycle. The closed-state test could then follow up with `page.waitForChanges()`.

core/src/components/item-sliding/item-sliding.tsx
brandyscarney: We should add some tests for this. I forgot we didn't have tests for events firing.

core/src/components/item-sliding/item-sliding.tsx
OS-pedrolourenco: Oops. This was a leftover. I have removed it.

core/src/components/item-sliding/item-sliding.tsx
OS-pedrolourenco: Fixed this and added a new test.

core/src/components/item-sliding/test/full-swipe/item-sliding.e2e.ts
OS-pedrolourenco: I have made this improvement but wasn't able to get the `should return to closed state after full swipe animation completes` test to pass with the changes you suggested (`await ionSwipe.next()` followed by `await page.waitForChanges()`), unfortunately.

core/src/components/item-sliding/item-sliding.tsx
```diff
@@ -248,6 +263,111 @@ export class ItemSliding implements ComponentInterface {
...
+
+      // Trigger expandable state without moving the item
+      // Set state directly so expandable option fills its container, starting from
+      // the exact position where the user released, without any visual snap.
+      this.state =
+        direction === 'end'
+          ? SlidingState.End | SlidingState.SwipeEnd | SlidingState.AnimatingFullSwipe
+          : SlidingState.Start | SlidingState.SwipeStart | SlidingState.AnimatingFullSwipe;
+
+      await new Promise<void>((resolve) => {
```
ShaneK: If `disconnectedCallback` fires while we're awaiting one of these, `clearTimeout(this.tmr)` cancels the timeout but the Promise's `resolve` never gets called, so the `await` hangs and the `finally` block never executes. The cleanup in `disconnectedCallback` (lines 122-129) partially covers this, but now there are two separate cleanup paths that need to stay in sync. Could you use a cancellation pattern so clearing the timer also resolves/rejects the promise? That way `finally` runs deterministically and the duplicate cleanup in `disconnectedCallback` isn't needed.

core/src/components/item-sliding/item-sliding.tsx
```diff
@@ -248,6 +263,111 @@ export class ItemSliding implements ComponentInterface {
...
+      // Return to closed state
+      await this.animateToPosition(0, 250);
+    } finally {
+      // Reset state
+      if (this.item) {
+        this.item.style.transition = '';
+      }
+      this.openAmount = 0;
+      this.state = SlidingState.Disabled;
+      openSlidingItem = undefined;
```
ShaneK: This unconditionally clears `openSlidingItem`, but during the ~900ms animation a user could start swiping a different item. That new swipe sets `openSlidingItem` to the new element in `onStart`, and then this `finally` block blindly clears it, orphaning the newly-opened item. Should this use the same guard as `disconnectedCallback`? i.e. `if (openSlidingItem === this.el) { openSlidingItem = undefined; }`

core/src/components/item-sliding/item-sliding.tsx
```diff
@@ -248,6 +263,111 @@ export class ItemSliding implements ComponentInterface {
...
+  }
+
+  /**
+   * Animate the item through a full swipe sequence: off-screen → trigger action → return.
+   * This is used when an expandable option is swiped beyond the threshold.
+   */
+  private async animateFullSwipe(direction: 'start' | 'end') {
+    // Prevent interruption during animation
+    if (this.gesture) {
+      this.gesture.enable(false);
```
ShaneK, line 345: You're disabling the gesture here to prevent interruption, but `close()` and `open()` are public `@Method()` calls that go through `setOpenAmount` and aren't gated. If someone calls `close()` programmatically during the animation, `setOpenAmount` clears `this.tmr` (which kills the current animation step), snaps the item, and schedules its own 600ms closing timer. The animation's promise hangs, `finally` never runs, and the component ends up with two competing timers. Should `close()`/`open()` check for `AnimatingFullSwipe` state and bail out?

core/src/components/item-sliding/item-sliding.tsx
```diff
@@ -370,6 +490,23 @@ export class ItemSliding implements ComponentInterface {
...
+    const hasExpandable = this.hasExpandableOptions(options);
+
+    const shouldTriggerFullSwipe =
+      hasExpandable &&
+      (rawSwipeDistance > this.getSwipeThreshold(direction) ||
+        (Math.abs(gesture.velocityX) > 0.5 &&
+          rawSwipeDistance > (direction === 'end' ? this.optsWidthRightSide : this.optsWidthLeftSide) * 0.5));
+
+    if (shouldTriggerFullSwipe) {
+      this.animateFullSwipe(direction);
```
ShaneK: This is fire-and-forget with no `.catch()`. If the promise hangs (e.g., component disconnects mid-animation) or `finally` throws, the gesture stays permanently disabled from the `enable(false)` in `animateFullSwipe`. Worth adding a `.catch()` that re-enables the gesture as a safety net?

core/src/components/item-sliding/item-sliding.tsx
```diff
@@ -248,6 +263,111 @@ export class ItemSliding implements ComponentInterface {
...
+    });
+  }
+
+  /**
+   * Calculate the swipe threshold distance required to trigger a full swipe animation.
+   * Returns the maximum options width plus a margin to ensure it's achievable.
+   */
+  private getSwipeThreshold(direction: 'start' | 'end'): number {
+    const maxWidth = direction === 'end' ? this.optsWidthRightSide : this.optsWidthLeftSide;
+    return maxWidth + 30; // Slightly larger than SWIPE_MARGIN to be achievable
```
ShaneK: The comment says 'slightly larger than SWIPE_MARGIN' but this is exactly `SWIPE_MARGIN` (both 30). Should this just reference the constant? `return maxWidth + SWIPE_MARGIN;` That way they stay in sync if the margin ever changes.

PR #31041, opened 2026-03-26
chore(deps): update to react 18

OS-jacobbell:
`packages/react` uses React 17.

## What is the new behavior?
`packages/react` uses React 18. This will allow Ionic Framework to update to the latest Stencil react-output-target, which requires React 18 or 19.

- Types needed to be updated for compatibility with type changes in React 18.
- `@testing-library/react-hooks` was moved to `@testing-library/react` so this dependency has been removed.

- [ ] Yes
- [X] No


thetaPC:
@OS-jacobbell please delete the branch if it's not needed anymore, thanks!

PR #31051, opened 2026-03-30
fix(modal): prevent double safe-area in fullscreen modals with ion-content (#31015)

### LINKED ISSUE #31015: bug: applyFullscreenSafeArea() shrinks fullscreen modal wrapper when ion-content exists without ion-footer (8.8.1 regression)
### Prerequisites
- [x] I have read the [Contributing Guidelines](https://github.com/ionic-team/ionic-framework/blob/main/.github/CONTRIBUTING.md).
- [x] I agree to follow the [Code of Conduct](https://github.com/ionic-team/ionic-framework/blob/main/CODE_OF_CONDUCT.md).

### Ionic Framework Version

v8.x (8.8.1)

### Current Behavior

After upgrading from `@ionic/angular` 8.7.17 to 8.8.1, fullscreen modals (presented via `ModalController` with `cssClass: "fullscreen"`) that contain `ion-content` but **no** `ion-footer` display a **white gap at the bottom** on iOS devices with a home indicator (safe area).

This is caused by the new `applyFullscreenSafeArea()` method in [`modal.tsx`](https://github.com/ionic-team/ionic-framework/blob/main/core/src/components/modal/modal.tsx) (introduced in PR #30949, included in 8.8.1). The method contains this logic:

```typescript
let hasContent = false;
let hasFooter = false;
for (const child of Array.from(el.children)) {
  if (child.tagName === 'ION-CONTENT') hasContent = true;
  if (child.tagName === 'ION-FOOTER') hasFooter = true;
  for (const grandchild of Array.from(child.children)) {
    if (grandchild.tagName === 'ION-CONTENT') hasContent = true;
    if (grandchild.tagName === 'ION-FOOTER') hasFooter = true;
  }
}

if (!hasContent || hasFooter) return;

wrapperEl.style.setProperty('height', 'calc(var(--height) - var(--ion-safe-area-bottom, 0px))');
wrapperEl.style.setProperty('padding-bottom', 'var(--ion-safe-area-bottom, 0px)');
```

When `ion-content` is present without `ion-footer`, the wrapper height is reduced by `--ion-safe-area-bottom` and equivalent padding is added. This creates a visible white gap at the bottom of the modal because the flex content area shrinks but the modal's visual boundary doesn't move.

This is a **regression** — the same templates worked correctly on 8.7.17.

### Expected Behavior

Fullscreen modals with `ion-content` and no `ion-footer` should render identically to how they did in 8.7.x — filling the entire screen without a gap at the bottom. The content inside `ion-content` already handles its own safe-area padding via `--ion-safe-area-bottom` on the scroll container. Applying it a second time at the wrapper level causes double safe-area compensation.

### Steps to Reproduce

1. Create an Ionic Angular app with `@ionic/angular` 8.8.1
2. Create a modal component with this template:
```html
<ion-header>
  <ion-toolbar>
    <ion-title>Modal Title</ion-title>
  </ion-toolbar>
</ion-header>

<ion-content>
  <p>Some content here</p>
</ion-content>

<!-- No ion-footer -->
```
3. Present it as a fullscreen modal:
```typescript
const modal = await this.modalCtrl.create({
  component: MyModalComponent,
  cssClass: 'fullscreen'
});
await modal.present();
```
4. Run on an iOS device/simulator with a home indicator (iPhone X or later)
5. Observe a white gap at the bottom of the modal

### Code Reproduction URL

_Reproduction project can be provided upon request._

### Ionic Info

```
Ionic:
   Ionic CLI       : 7.2.1
   Ionic Framework : @ionic/angular 8.8.1
   @angular/cli    : 21.3.1
   @ionic/angular-toolkit : 12.3.0

Capacitor:
   Capacitor CLI      : 8.1.0
   @capacitor/android : 8.1.0
   @capacitor/core    : 8.1.0
   @capacitor/ios     : 8.1.0

System:
   NodeJS : v22.17.0
   npm    : 10.9.2
   OS     : macOS
```

### Additional Information

**Root cause:** PR #30949 (resolving #28411 and #30900) added `applyFullscreenSafeArea()` to handle safe-area insets dynamically for modals. The heuristic `if (!hasContent || hasFooter) return;` assumes that any modal with `ion-content` and no `ion-footer` needs wrapper-level safe-area compensation. However, `ion-content` already handles bottom safe-area internally via its scroll container, so this creates a double safe-area effect.

**Affected templates:** Any modal that uses `ion-content` without `ion-footer` — a very common pattern for settings modals, info modals, selection modals, etc.

**Current workaround:** Add an empty `<ion-footer></ion-footer>` after `</ion-content>` in the modal template. This triggers the `hasFooter` early-exit path, preventing the wrapper height reduction:

```html
<ion-content>
  <!-- modal content -->
</ion-content>

<!-- Workaround: empty ion-footer prevents applyFullscreenSafeArea() regression -->
<ion-footer></ion-footer>
```

**Suggested fix considerations:**
- `ion-content` already applies `--ion-safe-area-bottom` to its scroll padding. The wrapper-level adjustment creates double compensation.
- Consider only applying the wrapper adjustment when the modal does NOT contain `ion-content` (raw HTML modals), or when `ion-content` has `fullscreen="false"`, rather than when `ion-footer` is absent.
- Alternatively, check whether the modal's `--height` is `100%` (fullscreen) before reducing it.

Vitalcheffe:
## Problem

After upgrading from @ionic/angular 8.7.17 to 8.8.1, fullscreen modals with ion-content but no ion-footer display a white gap at the bottom on iOS devices with a home indicator.

This is a regression from PR #30949, which added applyFullscreenSafeArea(). The function applies wrapper-level padding-bottom for modals with ion-content and no ion-footer. However, ion-content already handles bottom safe-area internally via its scroll container, creating double compensation.

## Root cause

The condition "if (!hasContent || hasFooter) return" means padding is applied when hasContent && !hasFooter. But ion-content already manages its own safe-area insets, so the wrapper-level padding creates an extra offset.

## Fix

Changed the condition to "if (hasContent || hasFooter) return" — skip wrapper padding when ion-content is present (it handles safe area internally) or when ion-footer is present (it handles its own padding). Only apply wrapper padding for custom modals with raw HTML that lack both.

## Impact

- Fixes the white gap regression for the common pattern: modals with ion-content and no ion-footer
- Preserves safe-area handling for custom modals without ion-content
- No breaking changes to existing ion-footer behavior

Fixes #31015

### DIFF
--- a/core/src/components/modal/modal.tsx
+++ b/core/src/components/modal/modal.tsx
@@ -1522,10 +1522,13 @@ export class Modal implements ComponentInterface, OverlayInterface {
       }
     }
 
-    // Only apply wrapper padding for standard Ionic layouts (has ion-content
-    // but no ion-footer). Custom modals with raw HTML are fully
-    // developer-controlled and should not be modified.
-    if (!hasContent || hasFooter) return;
+    // Only apply wrapper padding when ion-content is absent. When
+    // ion-content is present, it already handles bottom safe-area via
+    // its own scroll container — applying it again at the wrapper level
+    // creates double compensation (visible as a white gap on iOS).
+    // Custom modals with raw HTML are fully developer-controlled and
+    // should not be modified.
+    if (hasContent || hasFooter) return;
 
     // Reduce wrapper height by safe-area and add equivalent padding so the
     // total visual size stays the same but the flex content area shrinks.
--- a/core/src/components/modal/modal.tsx
+++ b/core/src/components/modal/modal.tsx
@@ -1522,13 +1522,17 @@ export class Modal implements ComponentInterface, OverlayInterface {
       }
     }
 
-    // Only apply wrapper padding when ion-content is absent. When
-    // ion-content is present, it already handles bottom safe-area via
-    // its own scroll container — applying it again at the wrapper level
-    // creates double compensation (visible as a white gap on iOS).
-    // Custom modals with raw HTML are fully developer-controlled and
-    // should not be modified.
-    if (hasContent || hasFooter) return;
+    // ion-content handles bottom safe-area internally via its scroll
+    // container. Applying wrapper-level padding on top of that creates
+    // double compensation (visible as a white gap on iOS with home
+    // indicator). Skip wrapper padding whenever ion-content is present
+    // regardless of ion-footer. See #31015.
+    if (hasContent) return;
+
+    // For modals without ion-content: apply wrapper padding only for
+    // standard Ionic layouts (has ion-footer). Custom modals with raw
+    // HTML are developer-controlled and should not be modified.
+    if (!hasFooter) return;
 
     // Reduce wrapper height by safe-area and add equivalent padding so the
     // total visual size stays the same but the flex content area shrinks.


ShaneK:
Thanks for the investigation here, the root cause in #31015 is spot on. `ion-content` handles its own bottom safe-area, so the wrapper padding was doubling it.

There's an e2e test failing in CI from this though. Check shard 11, specifically `core/src/components/modal/test/safe-area/modal.e2e.ts:103` ("fullscreen modal without footer should have wrapper padding-bottom"). It was added in #30949 alongside `applyFullscreenSafeArea()` and expects wrapper padding for exactly this case. That test will need updating to match the new behavior.

In general when changing a condition in a function, it's worth checking if there are tests covering the other branches too, not just the scenario you're fixing. Running `npm run test.e2e src/components/modal/test/safe-area` from `core/` would've caught this locally.

core/src/components/modal/modal.tsx
```diff
@@ -1522,10 +1522,13 @@ export class Modal implements ComponentInterface, OverlayInterface {
...
-    // but no ion-footer). Custom modals with raw HTML are fully
-    // developer-controlled and should not be modified.
-    if (!hasContent || hasFooter) return;
+    // Only apply wrapper padding when ion-content is absent. When
+    // ion-content is present, it already handles bottom safe-area via
+    // its own scroll container — applying it again at the wrapper level
+    // creates double compensation (visible as a white gap on iOS).
+    // Custom modals with raw HTML are fully developer-controlled and
+    // should not be modified.
+    if (hasContent || hasFooter) return;
```
ShaneK: This fixes the `ion-content` double compensation, but the condition flip also changes what happens for modals with neither `ion-content` nor `ion-footer`. Previously those raw HTML modals were left alone. Now they're the only ones that actually get wrapper padding applied. Was that intentional?

The comment above it says "Custom modals with raw HTML are fully developer-controlled and should not be modified" but that's now contradicting the code. Either update the comment to explain why wrapper padding is appropriate for those modals, or rethink the condition to preserve the "don't touch raw HTML modals" behavior.

ShaneK (changes requested):
Thanks for your PR! There's a few issues with this, mostly in that it was untested and affects cases other than the intended one.

Vitalcheffe:
Good catch — the original flip was too aggressive. Updated the logic to be explicit:

1. If `ion-content` is present → skip wrapper padding entirely (content handles safe-area)
2. If no content AND has footer → apply wrapper padding (standard Ionic layout)
3. If no content AND no footer → skip (raw HTML, developer-controlled)

This correctly fixes the double compensation for `ion-content` modals (#31015) while preserving the original behavior for all other cases.

ShaneK:
The [existing e2e test](https://github.com/ionic-team/ionic-framework/blob/fd724169fc71eb28aa8e4fd9482489661c18149f/core/src/components/modal/test/safe-area/modal.e2e.ts#L103) still isn't updated and will fail in CI again once the full suite runs on this commit, same as last time. Your `if (hasContent) return;` correctly skips wrapper padding now, but the test still asserts it's set.

If you're able to run the tests locally (`npm run test.e2e src/components/modal/test/safe-area` from `core/`), that would help catch these before pushing. If setting up the test environment is a blocker, let us know and we can help. We will need to close this if we can't get the tests passing though, so just flag it and we'll figure something out.

iAMkVIN-S:
Any update on this PR that would fix the [issue](https://github.com/ionic-team/ionic-framework/issues/31015) reported last month? 🙏

ShaneK:
Hey! The issue this PR was trying to address was resolved in #31092 and deployed in [v8.8.5](https://github.com/ionic-team/ionic-framework/releases/tag/v8.8.5), so I'm going to go ahead and close this. Thanks for your PR, sorry it took me so long to get to this.

If you're noticing other related problems, please open a new issue with relevant reproduction.

PR #31053, opened 2026-03-31
fix(datetime): multiple month selected and flakiness display

os-davidlourenco:

---------

Currently, the Datepicker keeps all the months visited as selected instead of only the one that is really selected
On iOS devices, the month picker sometimes doesn't appear since the `datepicker-ready` class is not added

## What is the new behavior?
Only the selected month appears selected in the month picker
On iOS devices, an additional validation is done to add the `datepicker-ready` class

- [ ] Yes
- [x] No

## Other information

The [Basic](https://ionic-framework-7el9b9l3f-ionic1.vercel.app/src/components/datetime/test/basic/?ionic:theme=ionic) test should be used to test the multiple-month selection case

### DIFF
--- a/core/src/components/datetime/datetime.tsx
+++ b/core/src/components/datetime/datetime.tsx
@@ -1096,10 +1096,7 @@ export class Datetime implements ComponentInterface {
       this.clearFocusVisible();
       this.clearFocusVisible = undefined;
     }
-    if (this.loadTimeout) {
-      clearTimeout(this.loadTimeout);
-      this.loadTimeout = undefined;
-    }
+    this.loadTimeoutCleanup();
   }
 
   /**
@@ -1150,6 +1147,13 @@ export class Datetime implements ComponentInterface {
     });
   };
 
+  private loadTimeoutCleanup = () => {
+    if (this.loadTimeout) {
+      clearTimeout(this.loadTimeout);
+      this.loadTimeout = undefined;
+    }
+  };
+
   componentDidLoad() {
     const { el, intersectionTrackerRef } = this;
 
@@ -1197,7 +1201,10 @@ export class Datetime implements ComponentInterface {
      * we still initialize listeners and mark the component as ready.
      *
      * We schedule this after everything has had a chance to run.
+     *
+     * We also clean up the load timeout to ensure that we don't have multiple timeouts running.
      */
+    this.loadTimeoutCleanup();
     this.loadTimeout = setTimeout(() => {
       this.ensureReadyIfVisible();
     }, 100);
--- a/core/src/components/datetime/test/basic/datetime.e2e.ts
+++ b/core/src/components/datetime/test/basic/datetime.e2e.ts
@@ -270,39 +270,6 @@ configs().forEach(({ title, screenshot, config }) => {
       const datetime = page.locator('ion-datetime');
       await expect(datetime).toHaveScreenshot(screenshot(`datetime-footer-custom-buttons`));
     });
-    test('should only have one active month option after changing month twice', async ({ page }) => {
-      await page.setContent(
-        `
-        <ion-datetime value="2022-05-03"></ion-datetime>
-      `,
-        config
-      );
-
-      await page.locator('.datetime-ready').waitFor();
-
-      const nextMonthButton = page.locator('ion-datetime .calendar-next-prev ion-button').nth(1);
-      const monthYearButton = page.locator('ion-datetime .calendar-month-year');
-
-      await monthYearButton.click();
-      await page.waitForChanges();
-      await monthYearButton.click();
-      await page.waitForChanges();
-      await nextMonthButton.click();
-      await page.waitForChanges();
-      await page.waitForTimeout(2000);
-      await monthYearButton.click();
-      await page.waitForChanges();
-      await monthYearButton.click();
-      await page.waitForChanges();
-      await nextMonthButton.click();
-      await page.waitForChanges();
-      await page.waitForTimeout(2000);
-      await monthYearButton.click();
-      await page.waitForChanges();
-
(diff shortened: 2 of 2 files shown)


core/src/components/datetime/test/basic/datetime.e2e.ts
```diff
@@ -270,6 +270,29 @@ configs().forEach(({ title, screenshot, config }) => {
...
+      );
+
+      await page.locator('.datetime-ready').waitFor();
+
+      const nextMonthButton = page.locator('ion-datetime .calendar-next-prev ion-button').nth(1);
+      const monthYearButton = page.locator('ion-datetime .calendar-month-year');
+
+      await nextMonthButton.click();
+      await nextMonthButton.click();
+      await page.waitForChanges();
```
brandyscarney:
```suggestion
      await expect(monthYearButton).toHaveText(/May 2022/);

      await nextMonthButton.click();
      await expect(monthYearButton).toHaveText(/June 2022/);

      await nextMonthButton.click();
      await expect(monthYearButton).toHaveText(/July 2022/);
```

This test does not currently fail when the fix is undone. I was able to get it to fail by including these expects.

core/src/components/datetime/datetime.tsx
```diff
@@ -1086,6 +1086,9 @@ export class Datetime implements ComponentInterface {
 
   connectedCallback() {
     this.clearFocusVisible = startFocusVisible(this.el).destroy;
+    this.loadTimeout = setTimeout(() => {
+      this.ensureReadyIfVisible();
+    }, 100);
```
ShaneK, line 1091: This `loadTimeout` gets overwritten by the [identical assignment in the init method](https://github.com/ionic-team/ionic-framework/blob/6b853c52b19e1f40a7f37c3c5166ffd1b88dd037/core/src/components/datetime/datetime.tsx#L1201) without clearing the one you're setting here. On first mount, `connectedCallback` runs first, then the init code reassigns `this.loadTimeout`, orphaning this timer. It still fires (harmless because `ensureReadyIfVisible` has the early return guard), but it can't be canceled in `disconnectedCallback` anymore.

Should the init code clear the existing timeout before setting its own? Or is this only intended for reconnection where the init method won't fire again? If so, you could guard it so it only runs on re-connect.

core/src/components/datetime/test/basic/datetime.e2e.ts
```diff
@@ -270,6 +270,39 @@ configs().forEach(({ title, screenshot, config }) => {
...
+      const nextMonthButton = page.locator('ion-datetime .calendar-next-prev ion-button').nth(1);
+      const monthYearButton = page.locator('ion-datetime .calendar-month-year');
+
+      await monthYearButton.click();
+      await page.waitForChanges();
+      await monthYearButton.click();
+      await page.waitForChanges();
+      await nextMonthButton.click();
+      await page.waitForChanges();
+      await page.waitForTimeout(2000);
```
ShaneK: Two `waitForTimeout(2000)` calls add 4 seconds of hard-coded delay per run. No other datetime test uses a delay this long. The [swiping tests below](https://github.com/ionic-team/ionic-framework/blob/6b853c52b19e1f40a7f37c3c5166ffd1b88dd037/core/src/components/datetime/test/basic/datetime.e2e.ts#L280-L301) use deterministic waits like `waitForChanges()` and `toHaveText()` assertions instead.

Could you replace these with the `toHaveText` assertions Brandy suggested? Those would both wait for the month to update and verify it's advancing correctly, which also makes the test actually fail when the fix is reverted.

core/src/components/datetime/test/basic/datetime.e2e.ts
```diff
@@ -270,6 +270,39 @@ configs().forEach(({ title, screenshot, config }) => {
       const datetime = page.locator('ion-datetime');
       await expect(datetime).toHaveScreenshot(screenshot(`datetime-footer-custom-buttons`));
     });
+    test('should only have one active month option after changing month twice', async ({ page }) => {
```
ShaneK: This test is inside the `datetime: footer` screenshot block, but it's a functional test about month picker selection with no screenshots and nothing about the footer. It's also inheriting the full `configs()` matrix (4 runs: ios/md x ltr/rtl) when this behavior doesn't vary by mode or direction.

Could you move it to its own block with something like `configs({ modes: ['md'], directions: ['ltr'] })` and a title like `'datetime: month picker selection'`? The [swiping tests](https://github.com/ionic-team/ionic-framework/blob/6b853c52b19e1f40a7f37c3c5166ffd1b88dd037/core/src/components/datetime/test/basic/datetime.e2e.ts#L280) do this for similar behavioral tests.

core/src/components/datetime/test/basic/datetime.e2e.ts
```diff
@@ -349,6 +349,40 @@ configs({ modes: ['md'], directions: ['ltr'] }).forEach(({ title, config }) => {
   });
 });
 
+/**
+ * This behavior does not differ across
+ * modes/directions.
+ */
+configs({ modes: ['md'], directions: ['ltr'] }).forEach(({ config }) => {
```
ShaneK: Every other `configs().forEach` block in this file destructures `title` and wraps tests in `test.describe(title('datetime: ...')), () => { ... })`. This is the only one that doesn't, so it won't get the mode/direction suffix in test reporters.

Could you match the pattern from the [other blocks](https://github.com/ionic-team/ionic-framework/blob/aa3a4e583d2fa1fe701a1c952898e1a995881353/core/src/components/datetime/test/basic/datetime.e2e.ts#L280-L281)? Something like:

```ts
configs({ modes: ['md'], directions: ['ltr'] }).forEach(({ title, config }) => {
  test.describe(title('datetime: month picker selection'), () => {
    test('should only mark the selected month as active', async ({ page }) => {
```

PR #31055, opened 2026-04-01
feat(router-outlet): add swipeGesture prop to control swipe-to-go-back per outlet

ShaneK:

---------

The swipe-to-go-back gesture on ion-router-outlet is controlled differently across frameworks:

- React and Vue check the `swipeBackEnabled` config option on every swipe attempt in their framework wrappers (`StackManager.canStart()` and `IonRouterOutlet.canStart()`)
- Angular reads the config once at mount and controls the gesture through its directive's `swipeGesture` setter
- There is no public, per-outlet property to enable or disable the gesture. The only mechanism is the global `swipeBackEnabled` config option

## What is the new behavior?
`ion-router-outlet` now exposes a public `swipeGesture` property in core that controls the swipe-to-go-back gesture per outlet instance. It defaults to true in "ios" mode and false in "md" mode, matching existing behavior.

- The `swipeBackEnabled` config is read once when the outlet mounts (as the default for `swipeGesture`) rather than checked on every swipe attempt
- The config check has been removed from React's StackManager and Vue's IonRouterOutlet -- core now owns this state
- Angular's directive forwards the value to the core component's `swipeGesture` property for consistency
- Apps can disable the gesture on a specific outlet: `<IonRouterOutlet swipeGesture={false} />`
- Apps that set `swipeBackEnabled` once at startup require no changes

- [X] Yes
- [ ] No

## Other information

I tried to align the updates to the BREAKING.md documents with the RR6 PR's changes to it to prevent issues from merging as much as possible


BREAKING_ARCHIVE/v8.md
brandyscarney, line 1: Wasn't the breaking change versioning already done on another PR?

ShaneK: Yeah, one that's not merged. I actually explicitly called this out in my PR description and why I did it, though.

BREAKING.md
```diff
@@ -4,290 +4,49 @@ This is a comprehensive list of the breaking changes introduced in the major ver
...
 
-**Minimum Mobile Platform Versions**
-| Platform | Supported Version      |
-| -------- | ---------------------- |
-| iOS      | 15+                    |
-| Android  | 5.1+ with Chromium 89+ |
+In React and Vue, the `swipeBackEnabled` config option is now read once when the outlet mounts. Apps that dynamically toggle this config value at runtime should migrate to the `swipeGesture` property instead:
 
-Ionic Framework v8 removes backwards support for CSS Animations in favor of the [Web Animations API](https://developer.mozilla.org/en-US/docs/Web/API/Web_Animations_API). All minimum browser versions listed above support the Web Animations API.
+**React:**
```
brandyscarney, line 30: Do we need to add an Angular example or no since this was already supported?

packages/react-router/test/base/src/pages/swipe-to-go-back/SwipToGoBack.tsx
brandyscarney, line 1: You might want to rename this file to avoid conflicts with the one on the rr6 branch: https://github.com/ionic-team/ionic-framework/blob/sk/react-router-6/packages/react-router/test/base/src/pages/swipe-to-go-back/SwipeToGoBack.tsx

Not sure which will merge better.

BREAKING.md
ShaneK: I don't think so, since Angular was already supported.

packages/react-router/test/base/src/pages/swipe-to-go-back/SwipToGoBack.tsx
ShaneK: This test will have to be changed regardless, I'm already prepared to deal with the conflict depending on which one merges first

codeCraft-Ritik (commented):
This is a fantastic PR description. You've clearly outlined the discrepancies between the different framework implementations and explained exactly how the new core property standardizes everything. The detail about `swipeBackEnabled `being read once on mount vs. every swipe is excellent context for reviewers.

PR #31062, opened 2026-04-02
fix(checkbox): show labels after page navigation

### LINKED ISSUE #31052: bug: ion-checkbox doesn't show label from variable after navigating in Angular
### Prerequisites

### Ionic Framework Version

v8.x

### Current Behavior

In Ionic 8.8.2 bug https://github.com/ionic-team/ionic-framework/issues/30632 was fixed, but that introduced a new bug:
When an ion-checkbox gets its label from an Angular variable, and you navigate from another page to the page with the checkbox, the label doesn't show.

### Expected Behavior

The checkbox label always shows the contents of the Angular variable.

### Steps to Reproduce

1. Go to https://stackblitz.com/~/github.com/SandraCaci/test-case-ion-checkbox
2. Wait until the app is started and you see "Demo of demo-test-input-checkbox" 
3. Observe that all checkboxes have a label (Ionic 8.8.2 is applied, the bugfix https://github.com/ionic-team/ionic-framework/pull/30980 ensured that also the 2nd checkbox shows a label).
4. Click the home icon in the upper left navbar to go to the home page, and then click the checkbox icon in the navbar just below the home icon to go back to the checkbox page
5. Observe that now all checkbox labels have disappeared.

### Code Reproduction URL

https://stackblitz.com/~/github.com/SandraCaci/test-case-ion-checkbox

### Ionic Info

**StackBlitz:**

Ionic:

   Ionic CLI : 5.4.16

Utility:

   cordova-res : not installed
   native-run  : not installed

System:

   NodeJS : v22.22.0
   npm    : 10.8.2
   OS     : Linux 5.0

**Local (same behavior):**
Ionic:

   Ionic CLI : 5.4.16

Utility:

   cordova-res : not installed
   native-run  : not installed

System:

   NodeJS : v24.14.0
   npm    : 11.9.0
   OS     : Windows 10

### Additional Information

Related to the fix of bug https://github.com/ionic-team/ionic-framework/pull/30980

OS-jacobbell:
Issue number: resolves #31052

---------

After a page navigation, ion-checkbox's `onslotchange` event fires before the element's `textContent` has been updated. It is called again after `textContent` becomes readable on Safari, but is not called again after the `textContent` becomes readable on Chrome and Firefox. 

## What is the new behavior?

- Uses `MutationObserver` instead of `onslotchange` and fires specifically on character data changes. This ensures `hasLabelContent` is up to date.
- `MutationObserver` does not fire on load, so `hasLabelContent` is initialized in `connectedCallback`

- [ ] Yes
- [X] No

### DIFF
--- a/core/src/components/checkbox/checkbox.tsx
+++ b/core/src/components/checkbox/checkbox.tsx
@@ -154,11 +154,11 @@ export class Checkbox implements ComponentInterface {
     if (Build.isBrowser && typeof MutationObserver !== 'undefined') {
       this.validationObserver = new MutationObserver((mutations) => {
         // Watch for label content changes
-        if (mutations.some((mutation) => mutation.type === 'characterData')) {
+        if (mutations.some((mutation) => mutation.type === 'characterData' || mutation.type === 'childList')) {
           this.hasLabelContent = this.el.textContent !== '';
         }
         // Watch for class changes to update validation state.
-        if (mutations.some((mutation) => mutation.type === 'attributes')) {
+        if (mutations.some((mutation) => mutation.type === 'attributes' && mutation.target === el)) {
           const newIsInvalid = checkInvalidState(el);
           if (this.isInvalid !== newIsInvalid) {
             this.isInvalid = newIsInvalid;
@@ -191,6 +191,7 @@ export class Checkbox implements ComponentInterface {
         attributes: true,
         attributeFilter: ['class'],
         characterData: true,
+        childList: true,
         subtree: true,
       });
     }


brandyscarney (approved):
Looks good! Please update the PR title before merging. 🙂

core/src/components/checkbox/checkbox.tsx
```diff
@@ -151,44 +151,53 @@ export class Checkbox implements ComponentInterface {
...
-           * By using a State variable inside a Promise,
-           * it guarantees a re-render immediately at
-           * a higher priority.
-           */
-          Promise.resolve().then(() => {
-            this.hintTextId = this.getHintTextId();
-          });
+      this.validationObserver = new MutationObserver((mutations) => {
+        // Watch for label content changes
+        if (mutations.some((mutation) => mutation.type === 'characterData')) {
```
ShaneK: This only catches mutations to existing text nodes (`.data` changes). If the framework replaces slot children entirely, like an `*ngIf` toggling a label, or `el.textContent = 'new'` which removes all children and creates a new text node, that's a `childList` mutation and this won't fire. The original `onSlotchange` covered both cases.

Should `childList: true` go into the [observe config](https://github.com/ionic-team/ionic-framework/pull/31062/files/cba086a1d8fb5e2af5f67a4d1b97d84fb8f4c5a2#diff-480c2b0a087c089851e16f28f69d728afc0bfbddf1b5c19b0b8ee07bd072c24bR193-R194) alongside `characterData: true`? The condition here would just become `mutation.type === 'characterData' || mutation.type === 'childList'`. [`createSlotMutationController`](https://github.com/ionic-team/ionic-framework/blob/cba086a1d8fb5e2af5f67a4d1b97d84fb8f4c5a2/core/src/utils/slot-mutation-controller.ts) uses `childList` for the same reason.

Also, [`radio.tsx`](https://github.com/ionic-team/ionic-framework/blob/cba086a1d8fb5e2af5f67a4d1b97d84fb8f4c5a2/core/src/components/radio/radio.tsx#L202-L203) and [`toggle.tsx`](https://github.com/ionic-team/ionic-framework/blob/cba086a1d8fb5e2af5f67a4d1b97d84fb8f4c5a2/core/src/components/toggle/toggle.tsx#L392-L393) have the same `get hasLabel()` getter pattern that checkbox had before #30980, so they're vulnerable to the same bug. Probably follow-up tickets rather than scope creep here, but good to track.

core/src/components/checkbox/checkbox.tsx
```diff
@@ -151,44 +151,53 @@ export class Checkbox implements ComponentInterface {
...
+            });
+          }
         }
       });
 
       this.validationObserver.observe(el, {
         attributes: true,
         attributeFilter: ['class'],
+        characterData: true,
+        subtree: true,
```
ShaneK, line 195: `subtree: true` applies to all observation types, not just `characterData`. So the `attributeFilter: ['class']` now fires for descendant class changes too, not just the host element. It's harmless since the handler filters by `mutation.type`, but you could tighten the [attributes branch](https://github.com/ionic-team/ionic-framework/pull/31062/files/cba086a1d8fb5e2af5f67a4d1b97d84fb8f4c5a2#diff-480c2b0a087c089851e16f28f69d728afc0bfbddf1b5c19b0b8ee07bd072c24bR161) with `mutation.target === el` to match the original observer's scope.

PR #31072, opened 2026-04-09
feat(select): allow HTML within options

### LINKED ISSUE #29890: feat: allow slot on ion-select-option
### Prerequisites

Add the `slot` option to the ion-select-option component.

### Describe the Use Case

Examples of elements to pass as slots:
image

See the elements have diff styles

Guess it works if the ion-select-option allows to pass `slot`

### Describe Alternatives

_No response_

### Related Code

_No response_

### Additional Information

_No response_

thetaPC:
Issue number: resolves #29890 

---------

Select only allows plain text to be passed within it's options.

## What is the new behavior?

**`ion-select-option`**
- Added default slot support for custom HTML content (images, avatars, icons, etc.) when `innerHTMLTemplatesEnabled` is `true`
- Added `start` and `end` named slots for content that appears in the overlay interface but not in the selected text
- Added `description` prop for text rendered below the option label in the overlay

**`ion-select`**
- Selected text renders HTML content from the default slot, excluding `start`/`end` slot content
- `aria-label` derives plain text only from the default slot, ensuring screen readers don't read slotted or element content
- Added CSS variables for styling media within the selected text (`--select-text-media-width`, `--select-text-media-height`, etc.)

**Overlay interfaces (`alert`, `action-sheet`, `select-popover`, `select-modal`)**
- All four interfaces render rich content (`start`, `end`, `description`, HTML label) consistently via the shared `renderOptionLabel` utility
- Public interfaces (`ActionSheetButton`, `AlertInput`, `SelectPopoverOption`, `SelectModalOption`) are unchanged — rich content fields are on internal extended interfaces (`SelectActionSheetButton`, `SelectAlertInput`, `SelectOverlayOption`)
- Content is sanitized via `sanitizeDOMString` before DOM injection to prevent XSS

**Utilities**
- Added `select-option-render.tsx`: shared render utility for option labels across all overlay components
- Added `getOptionContent`: extracts and clones slot content from `ion-select-option` for overlays and selected text
- Added `getDefaultSlotPlainText` — extracts plain text from the default slot, used for labels and aria

**Tests**
- Added E2E tests for rich content rendering across all four interfaces (single and multiple selection)
- Added tests for slot positioning (start/end placement verification)
- Added tests for selected text content verification (excludes slotted content)

- [ ] Yes
- [x] No

No, developers will be able to continue using plain text for select options as usual.

## Other information

[Preview](https://ionic-framework-git-fw-7137-ionic1.vercel.app/src/components/select/test/rich-content-option/)


core/src/components/action-sheet/action-sheet.native.scss
```diff
@@ -0,0 +1,19 @@
+@use "../../themes/native/native.theme.default" as native;
+@use "../../themes/mixins" as mixins;
+@use "../../themes/functions.font" as font;
+@use "./action-sheet.common";
+
+// Action Sheet: Native
+// --------------------------------------------------
+
+.action-sheet-button-label {
+  gap: 12px;
```
thetaPC, line 10: I decided to use the same value that ionic uses.

core/src/components/alert/alert.tsx
```diff
@@ -569,33 +573,43 @@ export class Alert implements ComponentInterface, OverlayInterface {
...
-              'alert-checkbox': true,
-              'alert-checkbox-button': true,
-              'ion-focusable': true,
-              'alert-checkbox-button-disabled': i.disabled || false,
-            }}
-          >
-            <div class="alert-button-inner">
-              <div class="alert-checkbox-icon">
-                <div class="alert-checkbox-inner"></div>
+        {inputs.map((i) => {
```
thetaPC, line 569: The only change is the addition of `optionLabelOptions` and `renderOptionLabel`

core/src/components/alert/alert.tsx
```diff
@@ -609,32 +623,42 @@ export class Alert implements ComponentInterface, OverlayInterface {
...
-              'alert-radio': true,
-              'ion-focusable': true,
-              'alert-radio-button-disabled': i.disabled || false,
-            }}
-            role="radio"
-          >
-            <div class="alert-button-inner">
-              <div class="alert-radio-icon">
-                <div class="alert-radio-inner"></div>
+        {inputs.map((i) => {
```
thetaPC, line 626: The only change is the addition of `optionLabelOptions` and `renderOptionLabel`

core/src/components/select-modal/select-modal.tsx
```diff
@@ -92,66 +93,86 @@ export class SelectModal implements ComponentInterface {
...
-              onKeyUp={(ev) => {
-                if (ev.key === ' ') {
-                  /**
-                   * Selecting a radio option with keyboard navigation,
-                   * either through the Enter or Space keys, should
-                   * dismiss the modal.
-                   */
-                  this.closeModal();
-                }
+        {this.options.map((option, index) => {
```
thetaPC, line 97: The only change is the addition of `optionLabelOptions` and `renderOptionLabel`

core/src/components/select-modal/select-modal.tsx
```diff
@@ -92,66 +93,86 @@ export class SelectModal implements ComponentInterface {
...
-        <ion-checkbox
-          value={option.value}
-          disabled={option.disabled}
-          checked={option.checked}
-          justify="start"
-          labelPlacement="end"
-          onIonChange={(ev) => {
-            this.setChecked(ev);
-            this.callOptionHandler(ev);
+    return this.options.map((option, index) => {
```
thetaPC, line 149: The only change is the addition of `optionLabelOptions` and `renderOptionLabel`

core/src/components/select-option/select-option.tsx
```diff
@@ -27,16 +27,26 @@ export class SelectOption implements ComponentInterface {
    */
   @Prop() value?: any | null;
 
+  /**
+   * Text that is placed underneath the option text to provide additional details about the option.
+   */
+  @Prop() description?: string;
```
thetaPC, line 33: Description isn't being used here to display on the screen because `ion-select-option` isn't used to display the options, it's done through the respective interface. So description is passed to the interface.

core/src/components/select-option/select-option.tsx
```diff
@@ -27,16 +27,26 @@ export class SelectOption implements ComponentInterface {
...
         class={{
           [theme]: true,
         }}
         role="option"
         id={this.inputId}
-      ></Host>
+      >
+        <slot name="start"></slot>
+        <slot></slot>
+        <slot name="end"></slot>
```
thetaPC, line 48: Slots aren't being used here to display on the screen because `ion-select-option` isn't used to display the options, it's done through the respective interface. So slots are passed to the interface and the rendering is done with a new [utility](https://github.com/ionic-team/ionic-framework/pull/31072/changes#diff-9ad5352208e128f6b5223775f67cb0237f738828d06614e5436638abfa9cf3b1).

Even though it's not used to display here, it's still useful to include here so the docs can generate that "slots" are available for the options.

core/src/components/select-popover/select-popover.tsx
```diff
@@ -119,65 +120,85 @@ export class SelectPopover implements ComponentInterface {
...
-        <ion-checkbox
-          value={option.value}
-          disabled={option.disabled}
-          checked={option.checked}
-          justify="start"
-          labelPlacement="end"
-          onIonChange={(ev) => {
-            this.setChecked(ev);
-            this.callOptionHandler(ev);
+    return options.map((option, index) => {
```
thetaPC, line 124: The only change is the addition of `optionLabelOptions` and `renderOptionLabel`

core/src/components/select-popover/select-popover.tsx
```diff
@@ -119,65 +120,85 @@ export class SelectPopover implements ComponentInterface {
...
-              onKeyUp={(ev) => {
-                if (ev.key === ' ') {
-                  /**
-                   * Selecting a radio option with keyboard navigation,
-                   * either through the Enter or Space keys, should
-                   * dismiss the popover.
-                   */
-                  this.dismissParentPopover();
-                }
+        {options.map((option, index) => {
```
thetaPC, line 173: The only change is the addition of `optionLabelOptions` and `renderOptionLabel`

core/src/components/action-sheet/action-sheet-interface.ts
```diff
@@ -14,7 +14,7 @@ export interface ActionSheetOptions extends OverlayOptions {
 }
 
 export interface ActionSheetButton<T = any> {
-  text?: string;
+  text?: string | HTMLElement;
```
ShaneK: Widening `text` from `string` to `string | HTMLElement` here affects all action-sheet consumers, not just select-driven ones. Same with [`AlertInput.label`](https://github.com/ionic-team/ionic-framework/pull/31072/files/03b8351b772d52cf62356ffe5a5b545c7735e99d#diff-24493ac9d18ef5dcf2fd1d1c191310d0fb82cb547b2edef59b993510398b914cR28). Anyone calling `actionSheetController.create()` who does `button.text.toLowerCase()` or passes `text` to a string-typed function will get TS errors. Since this targets `next`, breaking changes are expected, but would it be cleaner to add a separate property (e.g., `richText` or `contentElement`) for the HTMLElement case? That way existing string usage stays untouched.

core/src/components/action-sheet/action-sheet.tsx
```diff
@@ -580,7 +589,7 @@ export class ActionSheet implements ComponentInterface, OverlayInterface {
         >
           <span class="action-sheet-button-inner">
             {b.icon && <ion-icon icon={b.icon} aria-hidden="true" lazy={false} class="action-sheet-icon" />}
-            {b.text}
+            {renderOptionLabel(optionLabelOptions, 'action-sheet-button-label')}
```
ShaneK: This changes the DOM structure for all action-sheet buttons, not just ones created through select. Previously `{b.text}` rendered directly inside `.action-sheet-button-inner`, now it's wrapped in `<span class="action-sheet-button-label"><span>...</span></span>`. Same thing happens in [alert for radio/checkbox labels](https://github.com/ionic-team/ionic-framework/pull/31072/files/03b8351b772d52cf62356ffe5a5b545c7735e99d#diff-0a99d229e49597ec3fc6c5527b81880e64208692448d361b5fc6c760156abbeeR600). Any app CSS targeting these internal elements will break. Should `renderOptionLabel` only wrap when rich content is actually present (when startContent/endContent/description exist), falling back to the original rendering for plain strings?

ShaneK: I do understand that breaking changes are acceptable here and this may be intended. Just double checking that's what you're going for.

core/src/utils/select-option-render.tsx
```diff
@@ -0,0 +1,107 @@
...
+        if (el) {
+          const cached = contentCache.get(el);
+          // Skip if this element already has clones from the same source
+          if (cached === content) {
+            return;
+          }
+
+          // Clear previous content and clone new nodes
+          el.innerHTML = '';
+          Array.from(content.childNodes).forEach((n) => el.appendChild(n.cloneNode(true)));
```
ShaneK: The select's closed-state display goes through `sanitizeDOMString` before `innerHTML` injection, but this overlay path clones nodes with `cloneNode(true)` without sanitization. Inline event handlers (`onclick`, `onerror`) on elements inside `ion-select-option` get faithfully cloned into the overlay DOM. The two rendering paths for the same content have different security postures.

core/src/components/select/test/basic/select.e2e.ts
```diff
@@ -605,3 +605,363 @@ configs({ modes: ['md'], directions: ['ltr'] }).forEach(({ title, config }) => {
...
+                <img src="/src/components/avatar/test/avatar.svg" />
+              </ion-avatar>
+              Apples
+            </ion-select-option>
+          </ion-select>
+
+          <script>
+            window.Ionic = {
+              config: {
+                innerHTMLTemplatesEnabled: true,
```
ShaneK: This `window.Ionic = { config: { innerHTMLTemplatesEnabled: true } }` replaces the entire config object that `setContent` already sets in the head (which includes `mode` and `theme`). I think it works by accident because md is the fallback on non-iOS platforms, but you're losing the explicit config. Should be merging instead: `window.Ionic = window.Ionic || {}; window.Ionic.config = { ...window.Ionic.config, innerHTMLTemplatesEnabled: true };`

core/src/components/select/select.tsx
```diff
@@ -588,6 +595,9 @@ export class Select implements ComponentInterface {
           'aria-checked': isSelected ? 'true' : 'false',
           role: 'radio',
         },
+        startContent: startContent ?? undefined,
+        endContent: endContent ?? undefined,
+        description: option.description,
```
ShaneK, line 604: You're gating `startContent`/`endContent` behind `this.customHTMLEnabled` but `description` gets passed through regardless. Is that intentional? If someone has `innerHTMLTemplatesEnabled` set to false (the default), they'd still see descriptions in overlays but not the slot content. Feels like it should be consistent either way.

core/src/components/select/select.native.scss
```diff
@@ -88,6 +88,15 @@
   min-width: 16px;
 }
 
+/**
+ * If the select text contains rich content, we want to add some
+ * spacing between the items without changing the display to prevent
+ * losing the ellipsis behavior.
+ */
+.select-text > * {
```
ShaneK: `.select-text > *` applies `margin-inline-start` to the first child too, which adds unwanted leading space. Should be `.select-text > * + *` or `.select-text > *:not(:first-child)` to only space between items. Same thing in the [ionic theme file](https://github.com/ionic-team/ionic-framework/pull/31072/files/03b8351b772d52cf62356ffe5a5b545c7735e99d#diff-030647f0cfb9ca6f042e9d7114611c7076a7c01af73076d1db05cc881b40ddf7R58).

core/src/components/select-popover/select-popover.ionic.scss
```diff
@@ -0,0 +1,22 @@
+@use "../../themes/ionic/ionic.globals.scss" as globals;
+@use "./select-popover.common";
+@use "./select-popover.md" as select-popover-md;
+
+// Ionic Select Popover
+// --------------------------------------------------
+
+// Select Modal: Select Option
```
ShaneK: This says "Select Modal" but it's the select-popover file.

core/src/components/select/test/basic/index.html
```diff
@@ -12,6 +12,27 @@
     <script src="../../../../../scripts/testing/scripts.js"></script>
     <script nomodule src="../../../../../dist/ionic/ionic.js"></script>
     <script type="module" src="../../../../../dist/ionic/ionic.esm.js"></script>
+
+    <script>
+      window.Ionic = {
+        config: {
+          innerHTMLTemplatesEnabled: true,
```
ShaneK: Setting `innerHTMLTemplatesEnabled: true` globally here means all existing tests that navigate to this page now run with a different config than they were written for. Could the existing overlay open/dismiss and focus tests be affected? Would it be safer to create a separate test page for the rich content tests, or set the config per-test in `page.setContent`?

ShaneK (changes requested):
This is looking really good! Just some feedback on some things I noticed

core/src/components/action-sheet/action-sheet-interface.ts
thetaPC: Great point! https://github.com/ionic-team/ionic-framework/pull/31072/changes/cc0af05678f33dea908595e6db344b77b1f804c6 and https://github.com/ionic-team/ionic-framework/pull/31072/changes/49e0e6cb625fc6a32952d31e55aae3f5087aa32d

core/src/components/action-sheet/action-sheet.tsx
thetaPC: There's no need for the extra wrappers so it's now back to plain: https://github.com/ionic-team/ionic-framework/pull/31072/changes/0c7e4d60efd4086fdbca7270166c5155344a27da and https://github.com/ionic-team/ionic-framework/pull/31072/changes/17e9e378770085efc9d182efb1c1f9b1110d2457

core/src/components/select-popover/select-popover.ionic.scss
thetaPC: https://github.com/ionic-team/ionic-framework/pull/31072/changes/6ef23559f58769f8d76d4e6425da2d6b0895c6ab

core/src/utils/select-option-render.tsx
thetaPC: https://github.com/ionic-team/ionic-framework/pull/31072/changes/22dace32917c117dfd0607ae9d1c378a86d9c221

core/src/components/select/select.tsx
thetaPC: It was intentional. Devs can only pass [string text](https://github.com/ionic-team/ionic-framework/blob/22dace32917c117dfd0607ae9d1c378a86d9c221/core/src/components/select-option/select-option.tsx) as a description. To me, it would feel weird that a dev would have to enable `innerHTMLTemplatesEnabled` when they only want to have a description (no start or end slots). It follows how other props are being handled.

core/src/components/select/select.native.scss
thetaPC: I ended up finding an issue with my approach and the proposed approached when it came to multiple values and the rich content option is second on the list. The styling isn't applied as intended so there's no spacing between the plain text and the rich content. I ended wrapping everyone in a `span` if it has rich content and if it's in the select text: https://github.com/ionic-team/ionic-framework/pull/31072/changes/f2695caddcde7dba640ff3d866c67ddb16f9e053

core/src/components/select/test/basic/select.e2e.ts
thetaPC: I moved them to their own page: https://github.com/ionic-team/ionic-framework/pull/31072/changes/8a4e59bd59d0e13ac1d7d5f314d5f7070dc75b48

core/src/components/select/test/basic/index.html
thetaPC: Moved it to their own page: https://github.com/ionic-team/ionic-framework/pull/31072/changes/8a4e59bd59d0e13ac1d7d5f314d5f7070dc75b48

core/src/components/action-sheet/action-sheet.native.scss
thetaPC, line 10: Based it off `ionic` theme

mfernando797-gif (approved):

core/src/utils/select-option-render.tsx
```diff
@@ -0,0 +1,131 @@
+import { h } from '@stencil/core';
+
+import { sanitizeDOMString } from './sanitization';
+
+interface RichContentOption {
```
ShaneK, line 5: `RichContentOption` is declared here with `id`, `label`, `startContent`, `endContent`, `description`, and a separate [`RichContentOption` in `select-interface.ts`](https://github.com/ionic-team/ionic-framework/pull/31072/files/deaafb9412620eb6b21fcbbe29340bf531bf9bbc#diff-f47455daf5584b6e99f1679dd1d610b82a5afea485df90363c254af3875651a9R33) has only the three content fields. Same name, different shape, no shared source. Could we export one from `select-interface.ts` and import it here so we can avoid drift?

core/src/components/select/test/rich-content-option/index.html
```diff
@@ -0,0 +1,401 @@
...
+        <ion-list>
+          <ion-list-header>
+            <ion-label>Single Value</ion-label>
+          </ion-list-header>
+
+          <ion-item>
+            <ion-select id="alert-select" label="Alert" placeholder="Select one" interface="alert">
+              <ion-select-option value="full" description="Choose me!">
+                <ion-badge slot="end">NEW</ion-badge>
+                <ion-avatar id="avatar" slot="start" size="large">
```
ShaneK, line 60: `id="avatar"` is repeated 35+ times on this page (lines 60, 67, 74, etc.) Duplicate IDs are invalid HTML and none of the tests query by this id. Can we drop the attribute? I assume it's just copy pasta at this point

core/src/utils/select-option-render.tsx
```diff
@@ -0,0 +1,131 @@
...
+  className: string,
+  useSpan = false
+): HTMLElement | string | undefined => {
+  const { id, label, startContent, endContent, description } = option;
+  const hasRichContent = !!startContent || !!endContent || !!description;
+  const Tag = useSpan ? 'span' : 'div';
+
+  // Render simple string label if there is no rich content to display
+  if (!hasRichContent && (typeof label === 'string' || !label)) {
+    // const Tag = useSpan ? 'span' : 'div';
```
ShaneK, line 88: Dead commented line, `Tag` is already declared at line 84. Can you drop this?

ShaneK (approved):
Looks good to me! I consider these comments to be optional side quests (nits), feel free to disregard

core/src/utils/select-option-render.tsx
thetaPC: https://github.com/ionic-team/ionic-framework/pull/31072/changes/96a4bb38432d3ce92ac2d3d7d8051d79d7c6c5c6

core/src/components/select/test/rich-content-option/index.html
thetaPC: https://github.com/ionic-team/ionic-framework/pull/31072/changes/7fad4906e55435b66a3f361de0770a2ffd0bb0bb

core/src/utils/select-option-render.tsx
thetaPC: https://github.com/ionic-team/ionic-framework/pull/31072/changes/d6332e74d45d649416f10229d13e2c65a7dc339c

core/src/components/select/test/rich-content-option/index.html
brandyscarney, line 1: The popover example has cut off elements inside of it in all themes:

| ionic | ios | md |
| --- | --- | --- |
| ionic | ios | md |

brandyscarney, line 1: In the `ionic` theme, when you open and close the modal from the icon it scrolls to the bottom of the view and also the focus indicator is cut off:

https://github.com/user-attachments/assets/6bc6ea72-352b-481d-82f8-be8f8613410e

This does not happen for:
- The single value modal example
- The ios theme
- The md theme
- The overflowing or multiple [basic select test](https://ionic-framework-git-next-ionic1.vercel.app/src/components/select/test/basic?ionic:theme=ionic) on `next`
  - However, the focus indicator issue is there on the multiple example

Note: I didn't test if this exact test has the issue on next since this test is new.

core/src/components/select/test/rich-content-option/select.e2e.ts
brandyscarney, line 1: Can we add screenshot tests for this? We need to check the visual rendering as well.

core/src/components/select/test/rich-content-option/index.html
brandyscarney, line 1: The checkbox for the multiple value select modal looks different than the UX requirements - do we have a way for them to make this change?

 | Our implementation | Figma requirements |
 | ------------------ | ------------------ |
 | Our implementation | Figma requirements |

core/src/components/select-option/select-option.tsx
brandyscarney: Adding the slots here isn't what generates the docs. You need to add this to the component:

https://github.com/ionic-team/ionic-framework/blob/0db5b4032778cc8dfb8378865a623b58d5692989/core/src/components/item/item.tsx#L16-L18

core/src/components/select-popover/select-popover.ionic.scss
```diff
@@ -0,0 +1,22 @@
+@use "../../themes/ionic/ionic.globals.scss" as globals;
+@use "./select-popover.common";
+@use "./select-popover.md" as select-popover-md;
```
brandyscarney, line 3:
```suggestion
```

Is this an accidental import? It is adding the native styles for the `ionic` theme.

You should add any shared styles to `common`.

core/src/components/select-popover/select-popover.ionic.scss
```diff
@@ -0,0 +1,22 @@
...
+// Select Popover: Select Option
+// --------------------------------------------------
+
+.select-option-label {
+  gap: globals.$ion-space-300;
+}
+
+.select-option-description {
+  @include globals.typography(globals.$ion-body-md-regular);
+  @include globals.padding(0);
```
brandyscarney, line 17:
```suggestion
```

This can be removed when the `md` import is removed.

core/src/components/action-sheet/action-sheet.ionic.scss
```diff
@@ -0,0 +1,22 @@
+@use "../../themes/ionic/ionic.globals.scss" as globals;
+@use "./action-sheet.common";
+@use "./action-sheet.md" as action-sheet-md;
```
brandyscarney, line 3:
```suggestion
```

Is this an accidental import? It is adding the native styles for the `ionic` theme.

You should add any shared styles to `common`.

core/src/components/action-sheet/action-sheet.ionic.scss
```diff
@@ -0,0 +1,22 @@
...
+// Action Sheet: Select Option
+// --------------------------------------------------
+
+.action-sheet-button-label {
+  gap: globals.$ion-space-300;
+}
+
+.select-option-description {
+  @include globals.typography(globals.$ion-body-md-regular);
+  @include globals.padding(0);
```
brandyscarney, line 17:
```suggestion
```

This can be removed when the `md` import is removed.

core/src/components/alert/alert.ionic.scss
```diff
@@ -0,0 +1,23 @@
+@use "../../themes/ionic/ionic.globals.scss" as globals;
+@use "./alert.common";
+@use "./alert.md" as alert-md;
```
brandyscarney, line 3:
```suggestion
```

Is this an accidental import? It is adding the native styles for the `ionic` theme.

You should add any shared styles to `common`.

core/src/components/alert/alert.ionic.scss
```diff
@@ -0,0 +1,23 @@
...
+// --------------------------------------------------
+
+.alert-radio-label,
+.alert-checkbox-label {
+  gap: globals.$ion-space-300;
+}
+
+.select-option-description {
+  @include globals.typography(globals.$ion-body-md-regular);
+  @include globals.padding(0);
```
brandyscarney, line 18:
```suggestion
```

This can be removed when the `md` import is removed.

core/src/components/select/select.tsx
```diff
@@ -1421,30 +1513,192 @@ const parseValue = (value: any) => {
...
+
+  if (useHTML) {
+    return sanitizeDOMString(container.innerHTML.trim()) || null;
+  }
+
+  // Already sanitized through `renderOptionLabel`
+  return container;
+};
+
+const getOptionDefaultSlot = (option: HTMLIonSelectOptionElement): Node[] | null => {
```
brandyscarney, line 1686: Can we add a comment describing this function since the others have comments?

core/src/utils/select-option-render.tsx
```diff
@@ -0,0 +1,125 @@
+import { h } from '@stencil/core';
+
+import type { RichContentOption as RichContentOpt } from '../components/select/select-interface';
+
+import { sanitizeDOMString } from './sanitization';
+
+interface RichContentOption extends RichContentOpt {
```
brandyscarney, line 7: Why did we redefine this here instead of adding those properties to `select-interface`?

core/src/utils/select-option-render.tsx
```diff
@@ -0,0 +1,125 @@
...
+ * depending on the select interface. Buttons can only have phrasing
+ * content to prevent accessibility issues.
+ *
+ * @param id - Unique identifier for generating stable virtual DOM keys.
+ * @param content - The HTMLElement container whose child nodes will be cloned.
+ * @param className - CSS class applied to the wrapper span.
+ * @param useSpan - Whether to use a span element instead of a div for the wrapper.
+ */
+const renderClonedContent = (id: string, content: HTMLElement, className: string, useSpan = false) => {
+  const Tag = useSpan ? 'span' : 'div';
```
brandyscarney, line 38: I understand why the `span` is required, but why do we ever need to render as a `div`?

core/src/utils/select-option-render.tsx
```diff
@@ -0,0 +1,125 @@
...
+ * @param useSpan - Whether to use a span element instead of a div for the label.
+ */
+export const renderOptionLabel = (
+  option: RichContentOption,
+  className: string,
+  useSpan = false
+): HTMLElement | string | undefined => {
+  const { id, label, startContent, endContent, description } = option;
+  const hasRichContent = !!startContent || !!endContent || !!description;
+  const Tag = useSpan ? 'span' : 'div';
```
brandyscarney, line 80: I understand why the `span` is required, but why do we ever need to render as a `div`?

core/src/components/select/test/rich-content-option/index.html
brandyscarney, line 1: The avatar size in `md` theme is way too large. Can we decrease it here to match the other themes?

brandyscarney (changes requested):

core/src/components/select/test/rich-content-option/index.html
thetaPC: That isn't part of the scope in my opinion. The ticket is meant to add HTML content to the options.

thetaPC: https://github.com/ionic-team/ionic-framework/pull/31072/changes/8009bfe9254f3df0b2cfe2e08b9c85727553c489

thetaPC: I was not able to replicate either issues on Chrome and Firefox. What are the steps you took?

core/src/components/select-option/select-option.tsx
thetaPC: https://github.com/ionic-team/ionic-framework/pull/31072/changes/91091c286a04a526997f4f0827c8a7ba4c74b9ee

core/src/components/select-popover/select-popover.ionic.scss
thetaPC: This was not. Before creating the ionic file, the ionic theme was using [md styles](https://github.com/ionic-team/ionic-framework/blob/69e5ee11758ea333079a16acd1085409b899a74b/core/src/components/select-popover/select-popover.tsx#L23). So it still needs the md styles, I just didn't think it was necessary to copy and paste code.

core/src/components/select-popover/select-popover.ionic.scss
thetaPC: https://github.com/ionic-team/ionic-framework/pull/31072#discussion_r3141301521

core/src/components/action-sheet/action-sheet.ionic.scss
thetaPC: This was not. Before creating the ionic file, the ionic theme was using [md styles](https://github.com/ionic-team/ionic-framework/blob/69e5ee11758ea333079a16acd1085409b899a74b/core/src/components/action-sheet/action-sheet.tsx#L40). So it still needs the md styles, I just didn't think it was necessary to copy and paste code.

core/src/components/action-sheet/action-sheet.ionic.scss
thetaPC: https://github.com/ionic-team/ionic-framework/pull/31072#discussion_r3141304990

core/src/components/alert/alert.ionic.scss
thetaPC: This was not. Before creating the ionic file, the ionic theme was using [md styles](https://github.com/ionic-team/ionic-framework/blob/69e5ee11758ea333079a16acd1085409b899a74b/core/src/components/alert/alert.tsx#L47). So it still needs the md styles, I just didn't think it was necessary to copy and paste code.

core/src/components/alert/alert.ionic.scss
thetaPC: https://github.com/ionic-team/ionic-framework/pull/31072#discussion_r3141306266

core/src/components/select/select.tsx
thetaPC: https://github.com/ionic-team/ionic-framework/pull/31072/changes/bfd3c523180d171e2067050f62aa2472351d83c8

core/src/utils/select-option-render.tsx
thetaPC: I wanted the types to be strict.

core/src/utils/select-option-render.tsx
thetaPC: It's based on the flow of the interface. When it comes to action sheet, that is using buttons to render the option so we should be using `span`. The other interfaces are using `div`s to render the option. So in order to be consistent with the flow, I have it using a `div`.

core/src/utils/select-option-render.tsx
thetaPC: https://github.com/ionic-team/ionic-framework/pull/31072#discussion_r3141321136

core/src/components/select/test/rich-content-option/index.html
thetaPC: I would prefer to not modify the width since we'll eventually add sizes to native avatar and we can adjust it at that point. Let me know if you would still prefer to make the change now.

core/src/components/select/test/rich-content-option/select.e2e.ts
thetaPC: The reason that I didn't create screenshots for this is because the content for the options are based on the project so they're not fixed. Would it be beneficial to create screenshots for something like that?

core/src/components/select-option/select-option.tsx
brandyscarney: Shouldn't it say select option text instead of item?

thetaPC: https://github.com/ionic-team/ionic-framework/pull/31072/changes/fabd52167007c5a48fb329e39ec286a685f446c4

core/src/components/item-sliding/test/shapes/item-sliding.e2e.ts
```diff
@@ -7,9 +7,12 @@ import { configs, test, dragElementBy } from '@utils/test/playwright';
  */
 configs({ modes: ['ionic-md'], directions: ['ltr'] }).forEach(({ title, screenshot, config }) => {
   test.describe(title('item-sliding: shapes'), () => {
-    test('should not have visual regressions when not expanded', async ({ page }) => {
+    test('should not have visual regressions when not expanded', async ({ page, skip }) => {
       await page.goto(`/src/components/item-sliding/test/shapes`, config);
 
+      // TODO(FW-7288): Remove skip once fix has been implemented
```
thetaPC, line 13: I'm currently fixing the test so I don't expect this skip to be here long. I only added it so it's not a blocker.

core/src/components/select/test/rich-content-option/index.html
brandyscarney: Using Chrome or Firefox (does not happen in Safari):

1. Navigate to https://ionic-framework-git-fw-7137-ionic1.vercel.app/src/components/select/test/rich-content-option?ionic:theme=ionic
2. Click in the middle of the `Modal` item under `Multiple Value`
3. See that the modal opens with the scroll at the top
4. Click on the select icon of the `Modal` item under `Multiple Value`
5. See that the modal is scrolled to the bottom and the last item is focused
6. Refer to my video for visual steps

core/src/components/select/test/rich-content-option/select.e2e.ts
brandyscarney: We should define some good default styles for the rich content. As with [slotted elements in input](https://ionicframework.com/docs/api/input#start-and-end-slots-experimental)￼, it should look good out of the box without requiring additional styling from developers. I think we need to do this and then capture screenshots of it: [Input: slot test](https://github.com/ionic-team/ionic-framework/blob/main/core/src/components/input/test/slot/input.e2e.ts)

core/src/components/select/test/rich-content-option/index.html
brandyscarney: I don't think we need to provide the styling out of the box, but we do need to make sure it's possible to achieve on this PR.

core/src/components/select-popover/select-popover.ionic.scss
brandyscarney: The `ionic` styles should not be importing the `md` styles. This causes unnecessary overrides. If this is something we aren't going to do on this PR then we need to add it to the ticket to migrate select-popover as part of the modular work (FW-6891).

core/src/components/action-sheet/action-sheet.ionic.scss
brandyscarney: The `ionic` styles should not be importing the `md` styles. This causes unnecessary overrides. If this is something we aren't going to do on this PR then we need to add it to the ticket to migrate action-sheet as part of the modular work (FW-6885).

core/src/components/alert/alert.ionic.scss
brandyscarney: The `ionic` styles should not be importing the `md` styles. This causes unnecessary overrides. If this is something we aren't going to do on this PR then we need to add it to the ticket to migrate alert as part of the modular work (FW-6886).

core/src/components/select/test/rich-content-option/index.html
brandyscarney: We need to have good defaults for the UI when rich content is passed. See my comment here: https://github.com/ionic-team/ionic-framework/pull/31072#discussion_r3156615535

If we don't plan to adjust the styles now, we should create a follow-up ticket.

brandyscarney (approved):
All of my feedback can be done in follow-up tickets so I am approving this. Great work! 💪

PR #31073, opened 2026-04-09
chore(build): add Vercel preview builds for framework test apps

ShaneK:

---------

Vercel preview deployments only build and serve core Stencil component test pages. Framework-specific test apps (Angular, React, Vue, React Router) are not included, requiring developers to pull and build locally to validate them.

## What is the new behavior?
Vercel preview deployments build and serve framework test apps alongside core component tests under a single preview URL.

- [ ] Yes
- [X] No

### DIFF
--- a/core/scripts/vercel-build.sh
+++ b/core/scripts/vercel-build.sh
@@ -29,6 +29,9 @@ done
 echo "=== Ionic Framework Preview Build ==="
 echo "Core dir: ${CORE_DIR}"
 echo "Repo root: ${REPO_ROOT:-NOT FOUND}"
+if [ -z "${REPO_ROOT}" ]; then
+  echo "(This is expected in some Vercel configs -- framework test apps will be skipped)"
+fi
 
 rm -rf "${OUTPUT_DIR}"
 mkdir -p "${OUTPUT_DIR}"
@@ -135,17 +138,17 @@ echo ""
 echo "--- Step 2: Building Framework Packages ---"
 
 build_angular_pkgs() {
-  (cd "${REPO_ROOT}/packages/angular" && npm install && npm run build) || return 1
+  (cd "${REPO_ROOT}/packages/angular" && npm install && npm run sync && npm run build) || return 1
   (cd "${REPO_ROOT}/packages/angular-server" && npm install && npm run build) || return 1
 }
 
 build_react_pkgs() {
-  (cd "${REPO_ROOT}/packages/react" && npm install && npm run build) || return 1
+  (cd "${REPO_ROOT}/packages/react" && npm install && npm run sync && npm run build) || return 1
   (cd "${REPO_ROOT}/packages/react-router" && npm install && npm run build) || return 1
 }
 
 build_vue_pkgs() {
-  (cd "${REPO_ROOT}/packages/vue" && npm install && npm run build) || return 1
+  (cd "${REPO_ROOT}/packages/vue" && npm install && npm run sync && npm run build) || return 1
   (cd "${REPO_ROOT}/packages/vue-router" && npm install && npm run build) || return 1
 }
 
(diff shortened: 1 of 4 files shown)


core/scripts/vercel-build.sh
```diff
@@ -0,0 +1,353 @@
...
+for candidate in /vercel/path0 /vercel/path1 "${CORE_DIR}/.."; do
+  if [ -d "${candidate}/packages" ]; then
+    REPO_ROOT="${candidate}"
+    break
+  fi
+done
+
+echo "=== Ionic Framework Preview Build ==="
+echo "Core dir: ${CORE_DIR}"
+echo "Repo root: ${REPO_ROOT:-NOT FOUND}"
```
thetaPC, line 31: We get a warning if `REPO_ROOT` is not found later in the script, I wonder if it would be beneficial if we can move it after this echo as to remind the dev that there is no need for concern.

core/scripts/vercel-build.sh
```diff
@@ -0,0 +1,353 @@
...
+
+build_react_pkgs() {
+  (cd "${REPO_ROOT}/packages/react" && npm install && npm run build) || return 1
+  (cd "${REPO_ROOT}/packages/react-router" && npm install && npm run build) || return 1
+}
+
+build_vue_pkgs() {
+  (cd "${REPO_ROOT}/packages/vue" && npm install && npm run build) || return 1
+  (cd "${REPO_ROOT}/packages/vue-router" && npm install && npm run build) || return 1
+}
```
thetaPC, line 153: These are missing the `npm run sync` command

core/scripts/vercel-build.sh
```diff
@@ -0,0 +1,353 @@
...
+      echo "${v}"
+      return 0
+    fi
+  done
+  return 1
+}
+
+build_angular_test() {
+  local APP
+  APP=$(pick_app "${REPO_ROOT}/packages/angular/test" ng20 ng19 ng18) || {
```
thetaPC: Is it possible to make this dynamic? We would have to manually update the versions otherwise.

core/scripts/vercel-build.sh
```diff
@@ -0,0 +1,353 @@
...
+    return 1
+  fi
+  mkdir -p "${OUTPUT_DIR}/angular"
+  cp -r dist/test-app/browser/* "${OUTPUT_DIR}/angular/"
+  echo "[angular] Done."
+}
+
+build_react_test() {
+  local APP
+  APP=$(pick_app "${REPO_ROOT}/packages/react/test" react19 react18) || {
```
thetaPC: Is it possible to make this dynamic? We would have to manually update the versions otherwise.

core/scripts/vercel-build.sh
```diff
@@ -0,0 +1,353 @@
...
+  npx vite build --base /react/
+
+  mkdir -p "${OUTPUT_DIR}/react"
+  cp -r dist/* "${OUTPUT_DIR}/react/"
+  echo "[react] Done."
+}
+
+build_vue_test() {
+  local APP
+  APP=$(pick_app "${REPO_ROOT}/packages/vue/test" vue3) || {
```
thetaPC: Is it possible to make this dynamic? We would have to manually update the versions otherwise.

packages/react/test/base/src/App.tsx
```diff
@@ -44,7 +44,7 @@ setupIonicReact();
 
 const App: React.FC = () => (
   <IonApp>
-    <IonReactRouter>
+    <IonReactRouter basename={import.meta.env?.BASE_URL?.replace(/\/$/, '') || undefined}>
```
thetaPC, line 48: It might be helpful to add a comment of why `basename` is added so we avoid removing it because we don't remember it's importance.

thetaPC (changes requested):
Mainly requesting for the sync command.

Question: Is it possible to add a way where we can determine if we want to build the frameworks? That way if we only need core then we don't have to wait for the others to build. It can be a label that is added to the PR or text command that is submitted through the comments, etc.

packages/react/test/base/src/App.tsx
ShaneK: [Done](https://github.com/ionic-team/ionic-framework/pull/31073/commits/ab81a7837b2a772150137e69e31fd59a6dbeb955)

core/scripts/vercel-build.sh
ShaneK: [Done](https://github.com/ionic-team/ionic-framework/pull/31073/commits/ab81a7837b2a772150137e69e31fd59a6dbeb955), good catch!

core/scripts/vercel-build.sh
ShaneK: [Done](https://github.com/ionic-team/ionic-framework/pull/31073/commits/ab81a7837b2a772150137e69e31fd59a6dbeb955)

core/scripts/vercel-build.sh
ShaneK: [Done](https://github.com/ionic-team/ionic-framework/pull/31073/commits/ab81a7837b2a772150137e69e31fd59a6dbeb955)

core/scripts/vercel-build.sh
ShaneK: [Done](https://github.com/ionic-team/ionic-framework/pull/31073/commits/ab81a7837b2a772150137e69e31fd59a6dbeb955)

core/scripts/vercel-build.sh
ShaneK: [Done](https://github.com/ionic-team/ionic-framework/pull/31073/commits/ab81a7837b2a772150137e69e31fd59a6dbeb955)

ShaneK:
RE: Your review comment:

Not really, unfortunately. If we relied on code-based suggestion, we'd be building most of the time anyway (we need to build them all any time core changes). If we went the tag route, we'd have to make some annoying changes to Vercel, including giving it a GitHub token to read that I think. I'm pretty sure what it currently has access to isn't enough to pull labels like that.

thetaPC (approved):
LGTM, thank you so much!!

PR #31078, opened 2026-04-13
feat(picker-legacy, picker-legacy-column): remove legacy components

thetaPC:
Issue number: internal

---------

`ion-picker-legacy` and `ion-picker-legacy-column` are still available. They are components that have been deprecated in v8.

## What is the new behavior?

- Removed `ion-picker-legacy` and `ion-picker-legacy-column` from core for v9.
- Removed `ion-picker-legacy` and `ion-picker-legacy-column` from the frameworks for v9.
- Removed tests related to these components

- [x] Yes
- [ ] No

- `ion-picker-legacy` and `ion-picker-legacy-column` have been removed. The legacy picker component has been replaced with a inline picker component.
   - Usages such as `ion-picker-legacy` or `IonPickerLegacy` should be changed to `ion-picker` and `IonPicker`, respectively.

## Other information

N/A


core/src/components/app/test/safe-area/index.html
```diff
@@ -115,26 +114,6 @@
         await menu.open();
       }
 
-      async function showPicker() {
```
thetaPC, line 118: This was not replaced by `ion-picker` because that component doesn't apply safe area padding.

packages/angular/test/base/e2e/src/lazy/providers.spec.ts
```diff
@@ -53,10 +53,4 @@ test.describe('Providers', () => {
 
     await expect(page.locator('ion-loading')).toBeVisible();
   });
-
-  test('should open a picker', async ({ page }) => {
```
thetaPC, line 57: The point of this test was to verify the overlay was working as intended so we don't need to migrate it over to use `ion-picker`.

packages/angular/test/base/e2e/src/standalone/overlay-controllers.spec.ts
```diff
@@ -23,12 +23,6 @@ test.describe('Overlay Controllers', () => {
     await expect(page.locator('ion-modal app-dialog-content')).toBeVisible();
   });
 
-  test('should present a picker', async ({ page }) => {
```
thetaPC, line 26: The point of this test was to verify the overlay was working as intended so we don't need to migrate it over to use `ion-picker`.

packages/react/test/base/tests/e2e/specs/overlay-components/IonPicker.cy.ts
thetaPC, line 1: The point of this test was to verify the overlay was working as intended so we don't need to migrate it over to use `ion-picker`.

packages/react/test/base/tests/e2e/specs/overlay-hooks/useIonPicker.cy.ts
thetaPC, line 1: The point of this test was to verify the overlay was working as intended so we don't need to migrate it over to use `ion-picker`.

BREAKING.md
```diff
@@ -15,10 +15,16 @@ This is a comprehensive list of the breaking changes introduced in the major ver
...
 
 - [Components](#version-9x-components)
+  - [Legacy Picker](#version-9x-legacy-picker)
   - [Router Outlet](#version-9x-router-outlet)
 
 <h2 id="version-9x-components">Components</h2>
 
+<h4 id="version-9x-legacy-picker">Legacy Picker</h4>
+
+- `ion-picker-legacy` and `ion-picker-legacy-column` have been removed. The legacy picker component has been replaced with a inline picker component.
```
```suggestion
- `ion-picker-legacy` and `ion-picker-legacy-column` have been removed. The legacy picker component has been replaced with an inline picker component.
```

packages/angular/src/index.ts
brandyscarney, line 1: Do we also need to remove the components here: https://github.com/ionic-team/ionic-framework/blob/main/packages/angular-server/src/ionic-server-module.ts

brandyscarney (approved):
Looks good!

BREAKING.md
thetaPC: https://github.com/ionic-team/ionic-framework/pull/31078/changes/761f8ac0707bf86ca8b19d5aa600972f25d88914

packages/angular/src/index.ts
thetaPC: Good catch! https://github.com/ionic-team/ionic-framework/pull/31078/changes/c791c40f2d95685e68378916b6dde3423460ae5b

ShaneK (approved):
Looks good to me! Though I had one comment on the docs version of this that might require changes here should you choose to

PR #31080, opened 2026-04-14
fix(range): update tick styling and border radius

OS-giulianasilva:

---------

- When the range has `ticks="true",` the active bar segment is forced to a 0 border radius, while the inactive segment remains rounded.
- Tick height, width and border radius do not match Figma specifications.
- The tick and bar backgrounds use different colors.

## What is the new behavior?
- Added the `has-ticks` class to the `bar` part of the range bar, replicating the behavior found in the `bar-active` part.
- Updated the CSS to apply the same border radius for both `bar` and `bar-active` parts.
- Updated the tick's width, height and border radius.
- Changed the background color of `.range-tick` from a hardcoded neutral value to a CSS variable `--bar-background`, ensuring the ticks and bar share the same color.

- [ ] Yes
- [x] No

### Before:
image

### After: 
image


core/src/components/range/range.ionic.scss
```diff
@@ -21,6 +21,10 @@
   z-index: globals.$ion-z-index-100;
 }
 
+:host(.ticks) {
```
thetaPC: How is `.ticks` being added to range? I couldn't find it under `Host`.

thetaPC (approved):
LGTM

core/src/components/range/range.ionic.scss
OS-giulianasilva: @thetaPC, good catch! You're absolutely right! The `.ticks` was being added manually in that test file. I’ve updated the approach to align with how we handle the active part of the range bar and adjusted the CSS accordingly. Let me know if this looks better to you! #[8ac3957](https://github.com/ionic-team/ionic-framework/pull/31080/commits/8ac3957f13659a89c6e6121b24b68a6cd5b8c6da)

core/src/components/range/range.tsx
```diff
@@ -914,7 +914,14 @@ export class Range implements ComponentInterface {
         ))}
 
         <div class="range-bar-container">
-          <div class="range-bar" role="presentation" part="bar" />
+          <div
```
thetaPC, line 917: Why are these changes being applied? These are [already](https://github.com/ionic-team/ionic-framework/blob/37dbb80a9cdd195c24f9c2336dc339c3a8fd7cd0/core/src/components/range/range.tsx#L922) in `next`. You might need to sync with `next` to get them.

core/src/components/range/range.ionic.scss
```diff
@@ -75,30 +75,35 @@
...
 
-.range-bar.range-bar-active {
-  bottom: 0;
+.range-bar {
+  &.has-ticks {
+    @include border-radius(0);
+  }
 
-  width: auto;
+  &.range-bar-active {
```
thetaPC: Please revert this to be `.range-bar.range-bar-active`. Nested styles like these can get hard to read.

core/src/components/range/range.ionic.scss
```diff
@@ -75,30 +75,35 @@
...
 
-  background: var(--bar-background-active);
+    width: auto;
 
-  &.has-ticks {
-    @include border-radius(0);
-    @include margin(null, calc(-1 * globals.$ion-scale-100 * 0.5), null, calc(-1 * globals.$ion-scale-100 * 0.5));
+    background: var(--bar-background-active);
+
+    &.has-ticks {
```
thetaPC: Please revert this to be `.range-bar-active.has-ticks`. Nested styles like these can get hard to read.

core/src/components/range/range.ionic.scss
```diff
@@ -75,30 +75,35 @@
   @include margin(null, null, calc(globals.$ion-space-200 + globals.$ion-font-size-300), null);
 }
 
-.range-bar.range-bar-active {
-  bottom: 0;
+.range-bar {
+  &.has-ticks {
```
thetaPC: Please revert this to be `.range-bar.has-ticks`. Nested styles like these can get hard to read.

thetaPC (changes requested):

core/src/components/range/range.tsx
OS-giulianasilva: The permalink you attached refers to the `bar-active` part of the range bar. What I changed was the `bar` part that was missing this logic.

core/src/components/range/range.ionic.scss
OS-giulianasilva: Done!

core/src/components/range/range.ionic.scss
OS-giulianasilva: Done!

core/src/components/range/range.ionic.scss
OS-giulianasilva: Done!

core/src/components/range/range.tsx
thetaPC: Thank you! I completely missed that. It might be better to have a class on host then. But we can make that decision during ionic modular so this PR can move faster.

core/src/components/range/range.ionic.scss
```diff
@@ -75,30 +75,33 @@
   @include margin(null, null, calc(globals.$ion-space-200 + globals.$ion-font-size-300), null);
 }
 
+.range-bar.has-ticks {
+  @include border-radius(0);
+}
+
+.range-bar.range-bar-active.has-ticks {
+  @include margin(null, calc(-1 * globals.$ion-scale-100 * 0.5), null, calc(-1 * globals.$ion-scale-100 * 0.5));
```
thetaPC, line 83: Is this style needed? I had to zoom in a lot to see it being applied and it doesn't seem like it really does much.

thetaPC (approved):
LGTM

core/src/components/range/range.ionic.scss
OS-giulianasilva: I am not sure, to be honest. So I think I prefer to keep it for now.

PR #31081, opened 2026-04-14
feat(TabBar): add Ionic theme customization

joselrio:
Issue number: internal

---------
This pull request adds support for a translucent tab bar style in the Ionic tab bar component for Ionic theme by introducing new SCSS variables and styles. The main changes include defining variables for the translucent background and filter, and updating the component's SCSS to apply these styles when the `tab-bar-translucent` class is present.

**Translucent Tab Bar Support:**

* Added `tab-bar.ionic.vars.scss` with new SCSS variables for the translucent tab bar background color, alpha, and blur filter.
* Updated `tab-bar.ionic.scss` to import the new variables and apply translucent styles using the `tab-bar-translucent` class and `backdrop-filter` support. [[1]](diffhunk://#diff-fe63470b419227d79e943ff7370a337e9f273fbcd76d1d420916292e13126db7R2) [[2]](diffhunk://#diff-fe63470b419227d79e943ff7370a337e9f273fbcd76d1d420916292e13126db7R89-R103)

Translucent style was only applied to Ios theme.

## What is the new behavior?
These changes enable the tab bar to have a blurred, semi-transparent background when the `tab-bar-translucent` class is used, improving visual appeal and consistency with modern UI trends.

- [ ] Yes
- [x] No


core/src/components/tab-bar/tab-bar.ionic.vars.scss
```diff
@@ -0,0 +1,13 @@
+// Tab Bar Ionic
+// --------------------------------------------------
+/// @prop - Alpha of translucent tab bar background color
+$tab-bar-ionic-translucent-background-color-alpha: 0.6;
```
BenOsodrac: @joselrio we don't create vars files for ionic theme.

Also, these variables don't seem needed, as they are used just once. I think we can add the values directly

BenOsodrac (changes requested):

core/src/components/tab-bar/tab-bar.ionic.vars.scss
joselrio: Done!

BenOsodrac (approved):

thetaPC:
@joselrio @BenOsodrac  Moving forward, please make sure the PR titles follow the standard set for Ionic. The scopes should be in skewer-case. Thank you!

PR #31043, opened 2026-03-27
feat(badge): add recipe and tokens

thetaPC:
Issue number: internal

---------

## What is the current behavior?

Component styles for `ion-badge` are fragmented across multiple files (ios, md). Developers were restricted to those themes and how those themes structured the logic and styles.

## What is the new behavior?

- **Unified Style Architecture**: Combined theme-specific styling into a single `badge.scss` file that consumes CSS variables, ensuring a single source of truth for component logic.
- **Defined TypeScript Interface**: Added `badge.interfaces.ts`
- **Updated Test Page**: Updated the badge test page to be simple since new test pages with more in depth use cases were created
- **Added Test Pages**: Badge has multiple use cases based on which component it's slotted into. Test pages were added due to that.

## Does this introduce a breaking change?

- [x] Yes
- [ ] No

This PR introduces a breaking change to how `IonBadge` is styled. Existing manual CSS overrides targeting internal badge structures or old token names will no longer work due to the shift to thew new token hierarchy and a unified base SCSS file. 

**Migration Path:**
1.  **Token Updates**: Update any custom theme configurations to match the new nested structure.

2.  **CSS Overrides**: Ensure selectors align with the new slotted element logic and variable names (e.g., `--ion-badge-hue-bold-default-background`).

`--background` should no longer be used. Setting the value, `Badge.hue.bold.background`, within the tokens file should be used to change the background for the bold hue. Setting the value, `Badge.hue.subtle.background`, within the tokens file should be used to change the background for the subtle hue. 

3.  **Theme classes**: Remove any instances that target the theme classes: `badge.md`

## Other information

This PR significantly improves the developer experience for theming. By moving logic into `default.tokens.ts` and away from hardcoded SCSS functions, designers and developers can now override styles (like the size on a slotted avatar) purely through token configuration.

- [Avatar badges](https://ionic-framework-git-fw-6837-ionic1.vercel.app/src/components/avatar/test/badge/)
- [Button badges](https://ionic-framework-git-fw-6837-ionic1.vercel.app/src/components/button/test/badge/)
- [Tab Button badges](https://ionic-framework-git-fw-6837-ionic1.vercel.app/src/components/tab-button/test/badge/)
- [Badge](https://ionic-framework-git-fw-6837-ionic1.vercel.app/src/components/badge/test/basic/)
- [Badge: sizes](https://ionic-framework-git-fw-6837-ionic1.vercel.app/src/components/badge/test/size/)
[...]

PR #31087, opened 2026-04-17
feat(progress-bar): add recipe and tokens

thetaPC:
Issue number: internal

---------

Component styles for `ion-progress-bar` are fragmented across multiple files (native and `ionic`). The `md` and `ios` themes rely on the native styles. Developers were restricted to those themes and how those themes structured the logic and styles.

## What is the new behavior?

- **Unified Style Architecture**: Combined theme-specific styling into a single `progress-bar.scss` file that consumes CSS variables, ensuring a single source of truth for component logic.
- **Element-First Token Structure**: Refactored the `IonProgressBar` type and tokens to use an element-first hierarchy (`{type}.{element}.{state}`, e.g., `determinate.progress.default.background`). This provides predictability and consistency with the rest of the system.
- **Defined TypeScript Interface**: Added `progress-bar.interfaces.ts` with reusable types for the recipe and config.
- **Defined Theme Defaults**: Added per-theme token defaults in `ios`, `md`, and `ionic` theme files.
- **Shape Support**: Introduced `round` and `rectangular` shape variants with per-theme defaults (`round` for `ios` and `ionic`, `rectangular` for `md`).
- **Solid State Styling**: Introduced dedicated styles for when the progress bar is fully buffered (`buffer = 1`), including separate rules for the buffer bar and buffer circles.
- **Type-Scoped CSS**: Split generic buffer bar styles into type-specific selectors (`progress-bar-type-determinate` / `progress-bar-type-indeterminate`) to prevent cross-type style bleed.
- **Updated CSS Variables**: Replaced `--background` and `--progress-background` with a structured set of semantic variables (e.g., `--ion-progress-bar-determinate-progress-default-background`).
- **Updated Tests**: Added shape snapshot tests.

- [x] Yes
- [ ] No

This PR introduces breaking changes to how `IonProgressBar` is styled.

**Migration Path:**

1. **Token Updates**: Update any custom theme configurations to match the new nested structure.

2. **CSS Variable Replacements**: `--background` and `--progress-background` have been removed. Use the new token structure instead:
   - `--background` → `IonProgressBar.determinate.buffer.bar.default.background` (or the indeterminate equivalent)
   - `--progress-background` → `IonProgressBar.determinate.progress.default.background` (or the indeterminate equivalent)

If per-component customization is needed, the CSS variables can be used directly (e.g., `--ion-progress-bar-determinate-progress-default-background`).

3. **Host class names**: The type classes on the host element have been renamed to include the prop name:
   - `.progress-bar-determinate` → `.progress-bar-type-determinate`
   - `.progress-bar-indeterminate` → `.progress-bar-type-indeterminate`

4. **Theme classes**: Remove any instances that target the theme classes: `ion-progress-bar.md`, `ion-progress-bar.ios`.

## Other information

Previews:
Previews:
- [Basic](https://ionic-framework-git-fw-6857-ionic1.vercel.app/src/components/progress-bar/test/basic/)
- [Buffer](https://ionic-framework-git-fw-6857-ionic1.vercel.app/src/components/progress-bar/test/buffer/)
- [Determinate](https://ionic-framework-git-fw-6857-ionic1.vercel.app/src/components/progress-bar/test/determinate/)
- [Shape](https://ionic-framework-git-fw-6857-ionic1.vercel.app/src/components/progress-bar/test/shape/)


core/src/utils/theme.ts
```diff
@@ -622,6 +622,45 @@ export function currentColor(variation: string, alpha: number | string | null =
   }
 }
 
+interface IonColorOptions {
+  alpha?: number | string | null;
+  rgb?: boolean;
+  subtle?: boolean;
+}
+
+export function ionColor(name: string, variation: string, options: IonColorOptions = {}): string {
```
thetaPC, line 631: This function is originally added through the [badge PR](https://github.com/ionic-team/ionic-framework/pull/31043) but it's still in progress so I added it here to not be blocked.

core/src/themes/ionic/default.tokens.ts
```diff
@@ -463,6 +467,80 @@ export const defaultTheme: DefaultTheme = {
       },
     },
 
+    IonProgressBar: {
+      height: 'var(--token-scale-100, var(--ion-scaling-100))',
```
ShaneK, line 471: `var(--token-scale-100, var(--ion-scaling-100))` wraps a `--token-*` variable I couldn't find defined anywhere in the repo. Same on lines 532 and 538 with `--token-round-xs` and `--token-rectangular-xs`. Is this intentional scaffolding for the ionic modular work? If not, the other component entries in this file (`IonChip`, `IonSpinner`, `IonItemDivider`) reference `--ion-*` directly, so maybe this should too? Not sure.

core/src/components/progress-bar/progress-bar.scss
```diff
@@ -0,0 +1,365 @@
...
+
+// Determinate
+:host(.progress-bar-solid.progress-bar-type-determinate) .progress-buffer-bar {
+  background: var(
+    --ion-progress-bar-determinate-buffer-bar-solid-default-background,
+    var(--ion-progress-bar-determinate-buffer-bar-default-background)
+  );
+}
+
+:host(.progress-bar-solid) .buffer-circles {
```
ShaneK, line 209: When `progress-bar-solid` is set, the [render function](https://github.com/ionic-team/ionic-framework/pull/31087/files/bda0296864a64f613f1a6f09cb144f80e34d5ef0#diff-8910df3544717c58803e03e42829516a856569ac4433ef588d7ac5504e0be20cR125) applies `ion-hide` (`display: none !important`) to the parent `.buffer-circles-container`, so this `.buffer-circles` rule inside it can't paint. Is the plan to drop the `ion-hide` toggle later and let CSS handle the hide, or is this rule dead?

ShaneK (changes requested):
Looking good! Just a couple things I noticed but I could be off base

core/src/themes/ionic/default.tokens.ts
thetaPC: Ah, I meant to add a comment here.

You're right that `--token-*` variables don't exist in this repo. They're defined in the OS project. The intent was to make it easier for the components team to adopt the Ionic token file in the future by referencing `--token-*` first with `--ion-*` as a fallback. The components team can override values at the token level without having to make as many changes here. If we approve this change, I plan to go back to the others and add the `--token-*` variables through a different PR. I figured if it gets denied, then no need to do that extra work. 😅

core/src/components/progress-bar/progress-bar.scss
thetaPC: Good catch! https://github.com/ionic-team/ionic-framework/pull/31087/changes/6a19f57c37bf1a6253440c09689f7d32dc824c24

It sounds like we will keep `ion-hide` but just a [different class name](https://github.com/ionic-team/ionic-framework/blob/6a19f57c37bf1a6253440c09689f7d32dc824c24/core/src/components/progress-bar/progress-bar.tsx#L123).

core/src/components/progress-bar/progress-bar.tsx
```diff
@@ -135,9 +122,13 @@ const renderProgress = (value: number, buffer: number) => {
      * When finalBuffer === 1, we use display: none
      * instead of removing the element to avoid flickering.
      */
-    // TODO(FW-6697): change `ion-hide` class to `ion-display-none` or another class
+    // TODO(FW-6697): remove `ion-hide` class in favor of `progress-buffer-circles-hidden`
     <div
-      class={{ 'buffer-circles-container': true, 'ion-hide': finalBuffer === 1 }}
+      class={{
+        'buffer-circles-container': true,
+        'buffer-circles-container-hidden': finalBuffer === 1,
```
thetaPC, line 129: I added this class now so it would be easier to replace when we remove `ion-hide`.

core/src/components/progress-bar/progress-bar.tsx
```diff
@@ -135,9 +122,13 @@ const renderProgress = (value: number, buffer: number) => {
      * When finalBuffer === 1, we use display: none
      * instead of removing the element to avoid flickering.
      */
-    // TODO(FW-6697): change `ion-hide` class to `ion-display-none` or another class
+    // TODO(FW-6697): remove `ion-hide` class in favor of `progress-buffer-circles-hidden`
```
ShaneK:
```suggestion
    // TODO(FW-6697): remove `ion-hide` class in favor of `buffer-circles-container-hidden`
```

Nit: the class name in the TODO drifted from what you're actually applying two lines down. Future-us searching for `progress-buffer-circles-hidden` when FW-6697 comes around won't find anything.

core/src/components/progress-bar/progress-bar.tsx
```diff
@@ -95,8 +83,7 @@ export class ProgressBar implements ComponentInterface {
         aria-valuemin="0"
         aria-valuemax="1"
         class={createColorClasses(color, {
-          [theme]: true,
-          [`progress-bar-${type}`]: true,
+          [`progress-bar-type-${type}`]: true,
           'progress-paused': paused,
           'progress-bar-reversed': document.dir === 'rtl' ? !reversed : reversed,
           'progress-bar-solid': progressSolid,
```
ShaneK, line 90: Nit: The line below this `[`progress-bar-shape-${shape}`]: shape !== undefined,` should be just set to `true`, since this refactor makes it so that shape is always defined as far as I can tell

Also, it's very annoying that GitHub doesn't let you comment on lines just outside of their arbitrary scope of selection of lines in a diff

ShaneK (approved):
Looks great, Maria! Awesome work! Just two minor nits, but non-blocking

core/src/components/progress-bar/progress-bar.tsx
thetaPC: https://github.com/ionic-team/ionic-framework/pull/31087/changes/069befed684aae9377b20ce229636e4e8b53e7c7

core/src/components/progress-bar/progress-bar.tsx
thetaPC: https://github.com/ionic-team/ionic-framework/pull/31087/changes/c48f5b9b4c27533c7c027b5f3c6ea8678b479445

core/src/components.d.ts
```diff
@@ -2977,6 +2979,9 @@ export namespace Components {
          */
         "triggerAction": TriggerAction;
     }
+    /**
+     * *
+     */
```
brandyscarney: Where are these empty comment lines coming from?

core/src/themes/md/default.tokens.ts
```diff
@@ -33,6 +33,10 @@ export const defaultTheme: DefaultTheme = {
         size: 'large',
       },
 
+      IonProgressBar: {
+        shape: 'rectangular',
```
brandyscarney, line 37: Should the default be round per the md2 spec? https://m2.material.io/components/sliders

core/src/themes/ionic/default.tokens.ts
brandyscarney: Don't we plan to have them override all of these? Like they will redefine this, changing both variables:

```ts
height: 'var(--token-space-100, var(--token-scale-100, 4px)',
```

core/src/components/progress-bar/progress-bar.tsx
```diff
@@ -4,12 +4,12 @@ import { clamp } from '@utils/helpers';
...
 import { config } from '../../global/config';
-import { getIonTheme } from '../../global/ionic-global';
 import type { Color } from '../../interface';
 
+import type { IonProgressBarShape } from './progress-bar.interfaces';
+
 /**
- * @virtualProp {"ios" | "md"} mode - The mode determines the platform behaviors of the component.
- * @virtualProp {"ios" | "md" | "ionic"} theme - The theme determines the visual appearance of the component.
+ * * @virtualProp {"ios" | "md"} mode - The mode determines the platform behaviors of the component.
```
brandyscarney:
```suggestion
 * @virtualProp {"ios" | "md"} mode - The mode determines the platform behaviors of the component.
```

I think this extra * may be causing issues

brandyscarney (approved):
Great work on this! 💪

core/src/themes/ionic/default.tokens.ts
thetaPC: I'm hoping that they don't plan on overriding them especially by having it first check for the `--token-*`. With the fact that their token system changing frequently, it might be best for us to use `--token-*` instead of the internal token system. So if the value of `--token-scale-100` was `4px` but then changes to `6px`, they won't have to make an update on this repo because the token is still accurate, just the value changed. This is my reason of why I want to add their CSS variables. Does that answer your question?

core/src/components/progress-bar/progress-bar.tsx
thetaPC: https://github.com/ionic-team/ionic-framework/pull/31087/changes/62f395ea9ec68dcc7a47dd205703090eec31ca4a

core/src/components.d.ts
thetaPC: From the extra asterisk in `virtualProp`: https://github.com/ionic-team/ionic-framework/pull/31087/commits/3e9a5bbacb2222a4850a376d2392c5f6dfa43b5b

core/src/themes/md/default.tokens.ts
thetaPC: No, MD2 sliders are Ionic ranges. We should be using the [progress indicators](https://m2.material.io/components/progress-indicators#specs) and the default is rectangular. This also lines up with why we [don't have any border radius](https://github.com/ionic-team/ionic-framework/blob/f3cd39b7fb291286374285c4a326ec6b9a8ea237/core/src/components/progress-bar/progress-bar.md.scss) on `main`.

core/src/themes/ionic/default.tokens.ts
brandyscarney: I thought the goal was to not have the themes behave differently from each other. By adding another layer of tokens here we are doing that.

core/src/themes/md/default.tokens.ts
brandyscarney: You're right. I had a brain glitch and looked at the wrong component. 🤦‍♀️

core/src/themes/ionic/default.tokens.ts
thetaPC: Understood! Then I'll revert to what we've been doing.

thetaPC: https://github.com/ionic-team/ionic-framework/pull/31087/changes/e7ecbe96994fbbbfe7263557d8fbb8d79ebbeddb

PR #31089, opened 2026-04-17
feat(select-modal): add cancel icon

OS-susmitabhowmik:

---------

Currently the select modal only accepts a string for the `Cancel` action.

## What is the new behavior?
This PR adds a boolean prop, `cancelIcon` to the `IonSelect` modal which defaults to false. When the `cancelIcon` prop is true, an `X` icon is added in place of the `Cancel` text.

- [ ] Yes
- [x] No

## Other information

**Preview Links:**
- https://ionic-framework-git-rou-12695-add-select-cancel-icon-ionic1.vercel.app/src/components/select-modal/test/basic/
- https://ionic-framework-git-rou-12695-add-select-cancel-icon-ionic1.vercel.app/src/components/select-modal/test/basic?ionic:theme=ionic


core/src/components/select-modal/select-modal.tsx
```diff
@@ -79,6 +88,16 @@ export class SelectModal implements ComponentInterface {
     }
   }
 
+  private get cancelButtonIcon(): string {
+    const theme = getIonTheme(this);
+    const icons: Record<string, string> = {
```
ShaneK: The `icons` map is typed as `Record<string, string>` but `Theme` is an exhaustive union (`'ios' | 'md' | 'ionic'`), so `icons[theme]` ends up typed as `string | undefined`. Would `Record<Theme, string>` work here to enforce exhaustiveness at compile time? The existing pattern in [`searchbar.tsx`](https://github.com/ionic-team/ionic-framework/blob/12e579740cac7e927f1a36c0846aa1fedc352f45/core/src/components/searchbar/searchbar.tsx#L706) and [`input.tsx`](https://github.com/ionic-team/ionic-framework/blob/12e579740cac7e927f1a36c0846aa1fedc352f45/core/src/components/input/input.tsx#L951) is a plain object literal plus a `|| defaultIcons.md` fallback. Right now this matches neither.

ShaneK (approved):
Looks good to me! I had one minor suggestion/comment, but it's not blocking, so approving. Awesome work!

core/src/components/select-modal/select-modal.tsx
OS-susmitabhowmik: Great catch! I updated this type to be `Record<Theme, string>` in this commit dbeed2e3615215184d4c71db57425bc918ed1de7

PR #31092, opened 2026-04-22
fix(modal): remove safe-area gap and flash in fullscreen modals

### LINKED ISSUE #31015: bug: applyFullscreenSafeArea() shrinks fullscreen modal wrapper when ion-content exists without ion-footer (8.8.1 regression)

ShaneK:
Issue number: resolves #31015

---------

Fullscreen modals without an `ion-footer` set `height` and `padding-bottom` on `.modal-wrapper` to reserve space for the bottom safe-area. Because `ion-content`'s background does not extend into that wrapper padding, a visible gap appears between the content and the bottom edge of the modal.

Separately, modals that declare custom `--width` and `--height` on phone viewports flash inherited safe-area values while animating in, then snap to `0px` once the post-animation position correction runs.

## What is the new behavior?
Bottom safe-area compensation now happens inside `ion-content` instead of on the wrapper, so the modal background stays edge-to-edge.

- New internal `--ion-content-safe-area-padding-bottom` CSS property is added to `.inner-scroll`'s `padding-bottom` calc.
- `modal.tsx` sets that property on `ion-content` for fullscreen modals without a footer, and clears it on resize and dismiss.
- Wrapper no longer has `height` or `padding-bottom` inline styles written to it.
- New `hasCustomModalDimensions()` helper detects modals that override both `--width` and `--height` to non-fullscreen values. These are zeroed from the initial safe-area prediction so there is no flash before the post-animation correction.
- Extracted `findContentAndFooter()` and `applyFullscreenSafeAreaTo()` helpers so the resize handler and initial apply share the same lookup.
- E2E tests cover scroll-padding placement, wrapper content area remaining full viewport height, `.inner-scroll` padding including safe-area, and custom-dimension modals starting at zero.

- [ ] Yes
- [x] No

## Other information

`--ion-content-safe-area-padding-bottom` is an internal property. It is additive with the existing `--padding-bottom` consumers can set, and it only takes effect when `modal.tsx` writes it on fullscreen modals without a footer.

Example pages:
  - iOS: https://ionic-framework-git-fw-7136-ionic1.vercel.app/src/components/modal/test/basic?ionic:mode=ios
  - MD: https://ionic-framework-git-fw-7136-ionic1.vercel.app/src/components/modal/test/basic?ionic:mode=md

Example iOS card modal safe-area:
  - iOS: https://ionic-framework-git-fw-7136-ionic1.vercel.app/src/components/modal/test/safe-area?ionic:mode=ios

## Dev Build:
```
8.8.5-dev.11776879142.101200b9
```


core/src/components/modal/test/safe-area/modal.e2e.ts
```diff
@@ -100,10 +104,12 @@ configs({ modes: ['ios', 'md'], directions: ['ltr'] }).forEach(({ title, config
...
     });
 
-    test('fullscreen modal without footer should have wrapper padding-bottom', async ({ page }, testInfo) => {
+    test('fullscreen modal without footer should set safe-area scroll padding on ion-content', async ({
+      page,
+    }, testInfo) => {
       testInfo.annotations.push({
         type: 'issue',
-        description: 'https://github.com/ionic-team/ionic-framework/issues/30900',
+        description: 'https://github.com/ionic-team/ionic-framework/issues/31015',
```
thetaPC, line 112: Why was the issue changed?

core/src/components/modal/modal.tsx
```diff
@@ -1496,59 +1503,80 @@ export class Modal implements ComponentInterface, OverlayInterface {
...
-    // <app-footer><ion-footer>...</ion-footer></app-footer>).
-    // Note: uses a manual loop instead of querySelector(':scope > ...') because
-    // Stencil's mock-doc (used in spec tests) does not support :scope.
-    let hasContent = false;
+    const { contentEl, hasFooter } = this.findContentAndFooter();
+    this.applyFullscreenSafeAreaTo(contentEl, hasFooter);
+  }
+
+  /**
+   * Sets --ion-content-safe-area-padding-bottom on the given ion-content
```
thetaPC, line 1520: Correct me if I'm wrong but it seems that `--ion-content-safe-area-padding-bottom` is a a new variable to the project? If that's the case, can we comment that it's meant for internal usages and when to use it.

thetaPC (approved):
LGTM, minor suggestion

brandyscarney (approved):
Looks good!

core/src/components/modal/test/safe-area/modal.e2e.ts
ShaneK: Old version of this test asserted on `wrapperEl` `padding-bottom`, which was the behavior #30949 added when fixing #30900. This PR moves that compensation into `ion-content`'s scroll padding to fix #31015 (the regression #30949 caused), so the assertions now verify the #31015 fix rather than the original #30900 work. I don't think we can have two annotations at once for the same issue? Or can we? Not sure. We don't typically do that, so I just replaced this one.

core/src/components/modal/modal.tsx
ShaneK: For sure! Done: [559103a](https://github.com/ionic-team/ionic-framework/pull/31092/commits/559103a4b5e626518f3350249052f286d9225279)

PR #31095, opened 2026-04-24
feat(content): add recipe and tokens

thetaPC:

---------

`ion-content` does not fragment styles based on themes. All 3 themes share one style. However, it's not configured to the Modular Ionic.

## What is the new behavior?

- **Defined TypeScript Interface**: Renamed `content-interface.ts` to `content.interfaces.ts` and added `IonContentRecipe` type.
- **Defined Theme Defaults**: Added per-theme token defaults in ios, md, and ionic theme files.
- I**nternal Variable Prefix**: Renamed `--offset-top` / `--offset-bottom` to `--internal-offset-top` / `--internal-offset-bottom` to clearly separate them from the public CSS API.
- **CSS Logical Properties for RTL**: Replaced physical-property RTL selectors on `.transition-effect` and `.transition-shadow` with the `position-horizontal` mixin.
- **Removed dead property**: Dropped `-webkit-overflow-scrolling: touch` from styles as this property has been a no-op since iOS 13.
- **Updated Tests**: Added unit specs for `transitionShadow` (present in `ios` mode, absent in `md` mode) and the `getScrollElement()` / `getBackgroundElement()` element ref methods.
- **New utility**: Replaced inline `new Promise((resolve) => componentOnReady(...))` patterns with a shared `waitForComponent` helper, reducing boilerplate and improving type safety.
- **Split default vs. semantic tokens**: Nested background/color under `default` and added a `semantic.default` group. 

- [x] Yes
- [ ] No

This PR introduces breaking changes to how `ion-content` is styled.

Migration Path:

1. **CSS Variable Replacements**:    `--background` and `--color` have been **removed**. Use the new token structure:
```
--background -> IonContent.default.background
--color -> IonContent.default.color
```

`--padding-*` is still supported. `.ion-padding`, `.ion-padding-*` classes in `css/padding.scss` set `--padding-*` directly on the host. `ion-content` continues to honor those values (falling back to the new token when unset). Existing usage of the utility classes and direct `--padding-*` overrides will keep working.

New code is encouraged to use the token-based API instead:
```
--padding-top -> IonContent.padding.top
--padding-end -> IonContent.padding.end
--padding-bottom -> IonContent.padding.bottom
--padding-start -> IonContent.padding.start
```
_Note: `core/api.txt` only lists the new `--ion-content-padding-*` variables. The `--padding-*` overrides remain functional but are no longer part of the documented public API._

If per-component customization is needed, the CSS variables can be used directly.
```
--background -> --ion-content-background
--color -> --ion-content-color
--padding-top -> --ion-content-padding-top
--padding-end -> --ion-content-padding-end
--padding-bottom -> --ion-content-padding-bottom
--padding-start -> --ion-content-padding-start
```
2. **Internal-only variables (no replacement)**: The following CSS variables were previously documented `@prop`s on `ion-content` and have been renamed to the `--internal-*` namespace, removing them from the public API:

```
--keyboard-offset -> --internal-keyboard-offset
--offset-top      -> --internal-offset-top
--offset-bottom   -> --internal-offset-bottom
```

These are managed by `ion-content` itself (keyboard avoidance and header/footer offsets) and were never intended for consumer override. There is no replacement - any code that was setting them directly should be removed.

3. **Theme classes**: Remove any instances that target the theme classes: `ion-content.md`, `ion-content.ios`.

## Other information

Previews:
- [Basic](https://ionic-framework-git-fw-6896-ionic1.vercel.app/src/components/content/test/basic/)
- [Fullscreen](https://ionic-framework-git-fw-6896-ionic1.vercel.app/src/components/content/test/fullscreen/)
- [Standalone](https://ionic-framework-git-fw-6896-ionic1.vercel.app/src/components/content/test/standalone/)

The framework (Angular, React, Vue) test apps are not being styled correctly anymore because the new tokens are not being passed to them. This is expected until we can export the tokens as mentioned in the design doc. The functionality in those test apps are still working.


core/src/components/action-sheet/action-sheet.common.scss
```diff
@@ -143,7 +143,6 @@
   flex-shrink: 2;
   overscroll-behavior-y: contain;
   overflow-y: auto;
-  -webkit-overflow-scrolling: touch;
```
thetaPC, line 146: `-webkit-overflow-scrolling` has become [obsolete](https://developer.apple.com/documentation/safari-release-notes/safari-13-release-notes#New-Features) since iOS 13.

core/src/components/content/content.scss
```diff
@@ -44,36 +37,26 @@
   padding: 0 !important;
   /* stylelint-enable */
 
-  font-family: $font-family-base;
+  font-family: var(--ion-content-font-family);
 
   contain: size style;
 }
 
-:host(.ion-color) .inner-scroll {
```
thetaPC, line 52: There seems to be a lot deleted, but they were just moved around in the file.

core/src/components/content/content.tsx
```diff
@@ -326,9 +324,9 @@ export class Content implements ComponentInterface {
   @Method()
   async getBackgroundElement(): Promise<HTMLElement> {
     if (!this.backgroundContentEl) {
-      await new Promise((resolve) => componentOnReady(this.el, resolve));
```
thetaPC, line 334: Created a util for this since a lot of components are using the same code

core/src/components/content/test/standalone/index.html
```diff
@@ -15,7 +15,7 @@
   </head>
   <body>
     <div class="content-height">
-      <ion-content class="ion-padding">
```
thetaPC, line 18: These were remove because they never applied the padding from the start.

core/src/components/card/test/shape/card.e2e.ts-snapshots/card-default-ionic-md-ltr-light-Mobile-Chrome-linux.png
thetaPC, line 1: This change was because I synced the padding between themes to be consistent. `ionic` wasn't using `padding-*` on `ion-content`.

core/src/components/card/test/theme-ionic/index.html
```diff
@@ -32,7 +44,7 @@
         </ion-toolbar>
       </ion-header>
 
-      <ion-content class="ion-padding-space-400" id="content" no-bounce>
```
thetaPC, line 35: Removed the class because it wasn't set anywhere.

core/src/css/ionic/core.ionic.scss
thetaPC, line 1: The goal is to delete this file but that won't be until we can consolidate everything into one base. In the mean time, we should make updates so screenshots don't fail.

core/src/components/card/test/shape/index.html
```diff
@@ -33,7 +50,7 @@
         </ion-toolbar>
       </ion-header>
 
-      <ion-content class="ion-padding ion-text-center" id="content" no-bounce>
+      <ion-content class="ion-text-center" id="content" no-bounce>
```
thetaPC, line 55: Added the padding through the config since I already had to pass the background value there. Figured that I would keep it all in one place by adding the padding.

core/src/components/content/test/basic/index.html
```diff
@@ -136,6 +136,16 @@
           background: yellow;
         }
 
+        #content {
```
thetaPC, line 139: The values are being passed for this specific instance only. The other content components are using the values provided by `scripts.js`.

core/src/components/content/content.scss
```diff
@@ -1,34 +1,27 @@
...
-   * @prop --color: Color of the content
+   * @prop --ion-content-color: Color of the content
    *
-   * @prop --padding-top: Top padding of the content
-   * @prop --padding-end: Right padding if direction is left-to-right, and left padding if direction is right-to-left of the content
-   * @prop --padding-bottom: Bottom padding of the content
-   * @prop --padding-start: Left padding if direction is left-to-right, and right padding if direction is right-to-left of the content
+   * @prop --ion-content-font-family: Font family of the content
    *
-   * @prop --keyboard-offset: Keyboard offset of the content
```
thetaPC, line 17: I removed the comments for the internal variables.

core/src/components/content/content.scss
```diff
@@ -172,11 +179,37 @@
...
+   * causing it to appear under the main scrollable content as a result.
+   * The fixed content is correctly composited in other browsers. Adding
+   * the translateZ forces the fixed content to be composited so it correctly
+   * shows on top of the scrollable content. Setting a negative z-index will
+   * still allow the fixed content to appear under the scroll content if specified.
+   */
+  transform: translateZ(0);
+}
+
+// Content: iOS Mode Transition
```
thetaPC, line 216: The design doc did not account for animations based on mode. Should we be aiming to convert them to be based on themes? If so, then we can allow devs to customize the certain parts of the animations. In this instance, they could provide their preferred background color during the transition.
If we do decide to convert, then a ticket should be made and a TODO should be added here.

core/src/components/content/content.tsx
```diff
@@ -454,25 +452,27 @@ export class Content implements ComponentInterface {
...
     const rtl = isRTL(el) ? 'rtl' : 'ltr';
-    const theme = getIonTheme(this);
     const mode = getIonMode(this);
     const forceOverscroll = this.shouldForceOverscroll(mode);
-    const transitionShadow = theme === 'ios';
+    /**
+     * The transition shadow effect is only animated by the iOS transition
+     * builder, so these elements are only rendered in iOS mode.
+     */
+    const transitionShadow = mode === 'ios';
```
thetaPC, line 466: As mentioned in the comment, this is only [available in the `ios` mode](https://github.com/ionic-team/ionic-framework/blob/7b8d89c056630a64f4d8e99d9ac6ffdb40441eb4/core/src/utils/transition/index.ts#L118-L121) as shown in the [`ios.transition.ts` file.](https://github.com/ionic-team/ionic-framework/blob/7b8d89c056630a64f4d8e99d9ac6ffdb40441eb4/core/src/utils/transition/ios.transition.ts#L558)

core/scripts/testing/scripts.js
```diff
@@ -138,14 +159,7 @@ const DEFAULT_PALETTE = 'light';
...
       if (customTheme) {
-        theme = {
-          ...theme,
-          ...customTheme,
-          palette: {
-            ...theme.palette,
-            ...customTheme.palette,
-          },
-        };
+        theme = deepMerge(theme, customTheme);
```
thetaPC, line 162: This change [mimics exactly](https://github.com/ionic-team/ionic-framework/blob/5d7fb5f670270b188aedfca6c1869c2c8c16dd1d/core/src/utils/theme.ts#L59-L63) how we are loading themes in the `theme.ts` file.

core/src/components/content/content.scss
```diff
@@ -44,36 +37,31 @@
...
-  @include position(calc(var(--offset-top) * -1), 0px, calc(var(--offset-bottom) * -1), 0px);
-  @include padding(
-    calc(var(--padding-top) + var(--offset-top)),
-    var(--padding-end),
-    calc(var(--padding-bottom) + var(--keyboard-offset) + var(--offset-bottom)),
-    var(--padding-start)
+  $padding-top: var(--padding-top, var(--ion-content-padding-top, 0px));
+  $padding-end: var(--padding-end, var(--ion-content-padding-end, 0px));
+  $padding-bottom: var(--padding-bottom, var(--ion-content-padding-bottom, 0px));
+  $padding-start: var(--padding-start, var(--ion-content-padding-start, 0px));
```
thetaPC, line 56: In order for the padding classes (`.ion-padding`) to still be functional, we should be incorporating them as the first check.

core/src/components/content/content.scss
```diff
@@ -89,10 +77,16 @@
   touch-action: pan-x pan-y pinch-zoom;
 }
 
+:host(.ion-color) .inner-scroll {
+  background: color.current-color(base);
+  color: color.current-color(contrast);
+}
```
thetaPC, line 89: I didn't make this customizable because all themes are using the same styles. If the community requests it, then we can update the recipe to include it.

core/src/components/content/content.scss
```diff
@@ -216,6 +242,8 @@
 }
 
 .transition-shadow {
+  @include mixins.position-horizontal(null, 0);
```
thetaPC, line 247: We can now use this mixin instead of having to use `left/right`!

core/src/components/content/content.scss
```diff
@@ -89,10 +77,16 @@
   touch-action: pan-x pan-y pinch-zoom;
 }
 
+:host(.ion-color) .inner-scroll {
+  background: color.current-color(base);
+  color: color.current-color(contrast);
```
thetaPC: I was going to switch this to use `foreground` as stated in the ticket but that lead to unwanted changes like:

Image

The text color is not readable in some components. Others, it's too light and I question if it would pass accessibility.

Image

We should look into the foreground values again.

core/src/components/content/content.scss
ShaneK, line 56: The nesting here is load-bearing because `.ion-padding` and friends in `core/src/css/padding.scss` set `--padding-top` directly on the host, so dropping the legacy var would silently break every utility-class consumer. Worth a one-line SCSS comment so this doesn't get "simplified" away in a future cleanup.

The PR description's migration also reads as a hard break (`--padding-top` -> `--ion-content-padding-top`), but in practice direct `--padding-*` overrides keep working. Could you note in the migration that utility classes still work and direct `--padding-*` is deprecated but functional? Right now `core/api.txt` lists only the new vars while the SCSS honors both.

core/src/css/core.scss
```diff
@@ -247,8 +247,8 @@ ion-card-header.ion-color .ion-inherit-color {
  * The code below accounts for both ion-content and then custom
  * scroll containers within ion-content (such as virtual scroll)
  */
-.menu-content-open ion-content {
-  --overflow: hidden;
+.menu-content-open ion-content::part(scroll) {
```
ShaneK: You're switching the menu scroll-lock to `::part(scroll) { overflow: hidden }`, but this PR also exposes `--ion-content-overflow` as public CSS API with `'auto'` defaults across all three themes. Should this rule consume the new var instead, or is `overflow` the wrong abstraction at the recipe layer since it lives on the inner part, not the host? Right now the public var exists but the framework's own scroll-lock bypasses it.

Same change in `core/src/css/ionic/core.ionic.scss:257`.

core/src/components/content/content.scss
```diff
@@ -1,34 +1,27 @@
...
-  --padding-bottom: 0px;
-  --padding-start: 0px;
-  --padding-end: 0px;
-  --keyboard-offset: 0px;
-  --offset-top: 0px;
-  --offset-bottom: 0px;
-  --overflow: auto;
+  --internal-keyboard-offset: 0px;
+  --internal-offset-top: 0px;
+  --internal-offset-bottom: 0px;
```
ShaneK, line 23: The `--internal-*` rename is the right call, but it's a silent break for anyone who set `--keyboard-offset`, `--offset-top`, or `--offset-bottom` directly. The PR's migration path only mentions `--background`, `--color`, and `--padding-*`. Could you add a line saying these three are now internal-only with no replacement? They were previously documented `@prop`s on `ion-content`.

core/src/components/content/content.scss
```diff
@@ -44,36 +37,31 @@
   padding: 0 !important;
   /* stylelint-enable */
 
-  font-family: $font-family-base;
+  font-family: var(--ion-content-font-family);
```
ShaneK: Was `font-family: $font-family-base;` before, always set. Now if `--ion-content-font-family` isn't defined the declaration is invalid and font inherits from the ancestor. Theme-loaded apps are fine because the token blocks set it, but consumers using `@ionic/core` without the modular theme bundle (dist-custom-elements, partial migrations) lose the iOS/MD baseline stack.

If the theme bundle is mandatory in v9 we can skip this and call out the requirement explicitly. Otherwise an SCSS fallback would keep both paths working (would need to bring `$font-family-base` back via `@use`).

core/src/themes/ios/default.tokens.ts
```diff
@@ -599,6 +599,24 @@ export const defaultTheme: DefaultTheme = {
       },
     },
 
+    IonContent: {
+      background: baseColors.backgroundColor,
+      color: baseColors.textColor,
+
+      font: {
+        family: 'var(--ion-font-family, inherit)',
```
ShaneK: You're using `'var(--ion-font-family, inherit)'` here, but `IonBadge` at line 110 uses plain `'var(--ion-font-family)'` with no `inherit` fallback. Was the divergence intentional so apps with a custom body font cascade into `ion-content`? If yes, worth applying the same to `IonBadge` for consistency. If accidental, the `IonBadge` form keeps the font tied to `--ion-font-family`.

Same in `core/src/themes/md/default.tokens.ts` and `core/src/themes/ionic/default.tokens.ts`.

ShaneK (changes requested):
Looking awesome! Mostly just have a few clarafication/documentation questions

core/src/components/content/content.scss
thetaPC: https://github.com/ionic-team/ionic-framework/pull/31095/changes/4704cc1db57fa92fb956e22068ea1ff85e520252

I updated the description. But it's not been determined if the padding classes will be deprecated so I worded it differently.

core/src/css/core.scss
thetaPC: https://github.com/ionic-team/ionic-framework/pull/31095/changes/94e7f36a7bb164b2b649038e4c07a2c5d146a85d

core/src/components/content/content.scss
thetaPC: I also updated the PR description: https://github.com/ionic-team/ionic-framework/pull/31095/changes/dddcd4e6fa36c67bf73cb0afa669975f0ea5676b and https://github.com/ionic-team/ionic-framework/pull/31095/changes/32ea3b752c4b1c14c8256ef1fc3005999f191f8e

core/src/components/content/content.scss
thetaPC: The theme bundle is not mandatory. We want devs to have the option to choose one of ours or create their own. After reading your [badge comment](https://github.com/ionic-team/ionic-framework/pull/31095#discussion_r3197333406), I realized that I meant to set content the same as badge. We want `inherit` to always be the fallback.

https://github.com/ionic-team/ionic-framework/pull/31095/changes/095398c2a291b2f3ea8d32c015fdb30bfd706f7f

core/src/themes/ios/default.tokens.ts
thetaPC: https://github.com/ionic-team/ionic-framework/pull/31095#discussion_r3198074930

core/src/global/config.ts
```diff
@@ -21,7 +23,7 @@ export class Config {
    * @param fallback Default value if the key is not found
    * @returns The value found at the nested key or the fallback
    */
-  getObjectValue(key: string, fallback?: string): string | undefined {
+  getObjectValue(key: string, fallback?: ObjectConfigValue): ObjectConfigValue | undefined {
```
ShaneK: Are there callers lined up that need the `boolean` return? I couldn't find any consumers of `getObjectValue`, so widening from `string | undefined` to `string | boolean | undefined` reads as pre-emptive. If a follow-up PR needs it, that's cool, just checking.

core/src/components/content/content.scss
```diff
@@ -216,6 +254,8 @@
 }
 
 .transition-shadow {
+  @include mixins.position-horizontal(null, 0);
```
ShaneK, line 247: **Nit**: `.transition-effect` and `.transition-shadow` got converted to `mixins.position-horizontal(...)`, but [`.transition-cover` at line 245](https://github.com/ionic-team/ionic-framework/pull/31095/files/32ea3b752c4b1c14c8256ef1fc3005999f191f8e#diff-ca32a7f956c8f0ec26416437c4d92f2a755b928f56e072ce50dcd3f9d380bb86R245) still uses literal `right: 0` with stylelint-disable. Full-width so RTL behavior is identical, but is leaving the asymmetric pattern intentional?

ShaneK (approved):
Looks good to me! Just a question and a nit, but non-blocking

core/src/global/config.ts
thetaPC: It's from an old version of `ion-content` that I forgot to revert. It'll eventually be needed but that can be added at a later time. https://github.com/ionic-team/ionic-framework/pull/31095/changes/6f20e80974ee04888165715ff2bd9fad9744fa25

core/src/components/content/content.scss
thetaPC: After some more testing, it seems that it's not necessary to keep that `right` property. I was hesitant because it was added a long time ago.  Content has been modified since adding `right` that it seems to have been forgotten and never verified if we needed it.

https://github.com/ionic-team/ionic-framework/pull/31095/changes/bc743b447a09ebe2c051ae6e0e5c3e93a10af14f

core/src/components/content/content.scss
brandyscarney: Why did we go this route instead of having the `.ion-padding` class set `--ion-content-padding-top`, `--ion-content-padding-end`, etc directly?

core/src/components/list-header/test/basic/index.html
```diff
@@ -10,6 +10,21 @@
...
     <link href="../../../../../scripts/testing/styles.css" rel="stylesheet" />
+    <script>
+      // Need to be called before loading Ionic else
+      // the scripts.js logic runs too early.
+      window.Ionic = {
+        config: {
+          customTheme: {
+            components: {
+              IonContent: {
+                background: '#f1f1f1',
```
brandyscarney: How do we determine when this should be used over just keeping the css that was here and updating the variable?

```html
<style>
  ion-content {
    --ion-content-background: #f1f1f1;
  }
</style>
```

BREAKING.md
```diff
@@ -50,6 +51,46 @@ This is a comprehensive list of the breaking changes introduced in the major ver
 - Specific theme classes (e.g., `ion-chip.md`) are no longer supported. Style modifications based on the active theme must be implemented using theme tokens rather than direct class targeting.
 - The `border-radius` of the `ios` and `md` chip now defaults to `10px` and `8px`, respectively, instead of `16px` in accordance with the iOS and Material Design 3 guidelines. To revert to the previous appearance, set the `shape` to `"round"`, or override the `IonChip.shape.round.border.radius` to specify a different value for global styles and `--ion-chip-shape-round-border-radius` for component-specific styles.
 
+<h4 id="version-9x-content">Content</h4>
```
brandyscarney, line 86: Could we clean this section up a bit? It's kind of hard to hard to read through. Maybe we can use a numbered list or dividers?

core/src/utils/helpers.ts
```diff
@@ -83,6 +83,15 @@ export const componentOnReady = (el: any, callback: any) => {
   }
 };
 
+/**
+ * Promise-based wrapper around componentOnReady. Use when you need to await
+ * component readiness before accessing internal refs (e.g. in early lifecycle
+ * hooks like Vue onMounted with the custom elements build).
+ */
+export const waitForComponent = <T extends Element>(el: T): Promise<T> => {
```
brandyscarney: Nit: should this be called `waitForComponentReady`?

core/src/components/content/content.scss
```diff
@@ -1,34 +1,34 @@
...
-  --padding-top: 0px;
-  --padding-bottom: 0px;
-  --padding-start: 0px;
-  --padding-end: 0px;
-  --keyboard-offset: 0px;
-  --offset-top: 0px;
-  --offset-bottom: 0px;
-  --overflow: auto;
+  /**
+   * TODO(FW-6698): Remove this comment in v10.
```
brandyscarney: Do we need this comment to begin with?

core/src/components/content/content.scss
brandyscarney: What is this referring to?

core/src/components/content/content.scss
```diff
@@ -115,15 +121,29 @@
 }
 
 .scroll-y {
-  overflow-y: var(--overflow);
+  overflow-y: var(--ion-content-overflow);
```
brandyscarney: Should this fall back to `auto`?

core/src/components/content/content.scss
```diff
@@ -115,15 +121,29 @@
...
   overscroll-behavior-y: contain;
 }
 
 .scroll-x {
-  overflow-x: var(--overflow);
+  overflow-x: var(--ion-content-overflow);
```
brandyscarney: Should this fall back to `auto`?

core/src/components/content/content.scss
brandyscarney: The `foreground` color should only be used when it's a text-only component. So if setting `color` on the content changes its text but _not_ its background. You can look at `ion-text` as an example of a text-only component that uses `color`. Since we are changing the background on this component, `contrast` is the correct text value.

core/src/components/content/content.scss
brandyscarney: Maybe we should add this on to FW-7295?

brandyscarney (commented):
Looking good! Minor changes requested.

core/src/components/content/content.scss
thetaPC: I did think of doing that initially but we have a lot of components that use `--padding-*` like `ion-button`. The class would end up having a lot of components:
```
.ion-padding: {
   --ion-content-padding-*: ???;

   --ion-button-padding-*: ???;
}
```

This just didn't feel right.

core/src/components/list-header/test/basic/index.html
thetaPC: Using the config is recommended to use globally. The styles is recommended to use per instance. In this case, I'm wanting ally the `ion-content` to have that background and that includes any future `ion-content` that may be added.

core/src/components/content/content.scss
thetaPC: Shane [mentioned](https://github.com/ionic-team/ionic-framework/pull/31095#discussion_r3197331674) that it would be a good idea to highlight it for any dev who is using the internal variables.

core/src/components/content/content.scss
thetaPC: It's already been added.

core/src/components/content/content.scss
thetaPC: This just changes the background and color when dev does `<ion-content color="danger">`. Examples can be seen on the [Colors section](https://ionicframework.com/docs/api/content#colors) in the docs.

BREAKING.md
thetaPC: https://github.com/ionic-team/ionic-framework/pull/31095/changes/a56e8ce574ea42cb1168c693b3d026ad055b6ced

core/src/utils/helpers.ts
thetaPC: https://github.com/ionic-team/ionic-framework/pull/31095/changes/e167474c2f2deecdae412f51237da490213c859b

core/src/components/content/content.scss
thetaPC: https://github.com/ionic-team/ionic-framework/pull/31095/changes/944296cdd8f662e1e8352123ea76e6d0e4fd879d

core/src/components/content/content.scss
thetaPC: https://github.com/ionic-team/ionic-framework/pull/31095/changes/944296cdd8f662e1e8352123ea76e6d0e4fd879d

core/src/components/content/content.scss
brandyscarney: Ah I see. That makes sense.

core/src/components/content/content.scss
brandyscarney: Sorry when I replied your comment was on an empty line for me, but now I see the highlighted code.

I thought we were using the `semantic` tokens for this?

core/src/components/content/content.scss
brandyscarney: I can't see the comment because Github isn't navigating to it for me but wouldn't they need to go to the code to see this? Why not just add it as a potential breaking change?

BREAKING.md
```diff
@@ -82,6 +83,53 @@ This is a comprehensive list of the breaking changes introduced in the major ver
...
+New code should use the token-based API:
+
+| Old (8.x) | New token (global) | New CSS variable (component-specific) |
+|---|---|---|
+| `--padding-top` | `IonContent.padding.top` | `--ion-content-padding-top` |
+| `--padding-end` | `IonContent.padding.end` | `--ion-content-padding-end` |
+| `--padding-bottom` | `IonContent.padding.bottom` | `--ion-content-padding-bottom` |
+| `--padding-start` | `IonContent.padding.start` | `--ion-content-padding-start` |
+
+> **Note:** The `--padding-*` overrides and `.ion-padding`, `.ion-padding-*` utility classes in `css/padding.scss` continue to work — `ion-content` honors them as a fallback when the new token is unset. They are no longer part of the documented public API (only `--ion-content-padding-*` is listed in `core/api.txt`), but existing usage will not break.
```
```suggestion
> [!NOTE]
> The `--padding-*` overrides and `.ion-padding`, `.ion-padding-*` utility classes in `css/padding.scss` continue to work — `ion-content` honors them as a fallback when the new token is unset. They are no longer part of the documented public API (only `--ion-content-padding-*` is listed in `core/api.txt`), but existing usage will not break.
```

Recommend using GitHub's blockquote highlighting.

brandyscarney (approved):
Left some non-blocking comments. 👍

core/src/components/content/content.scss
thetaPC: https://github.com/ionic-team/ionic-framework/pull/31095/changes/4de68664d0f2399fb6b076716bcbf3bb048a7fca

core/src/components/content/content.scss
thetaPC: It's already mentioned in the PR description and in BREAKING.md so I've removed the comment: https://github.com/ionic-team/ionic-framework/pull/31095/changes/5490912b8a98dbb59cbe0367c803f0d4657ec6f7

BREAKING.md
thetaPC: https://github.com/ionic-team/ionic-framework/pull/31095/changes/304139657de023018d993805cba443a6f1f444e2

core/src/components/content/content.interfaces.ts
```diff
@@ -19,3 +20,24 @@ export interface ScrollBaseCustomEvent extends CustomEvent {
...
   detail: ScrollDetail;
 }
+
+export interface IonContentRecipe {
+  default: {
+    background?: string;
+    color?: string;
+  };
+
+  semantic: {
```
brandyscarney, line 30: Should any component that supports semantic `color` also support `hue`?

brandyscarney (approved):

core/src/components/content/content.interfaces.ts
thetaPC: Discussing this offline. If work is needed, then we will add it through another ticket.

PR #31098, opened 2026-04-28
chore(deps): unpin @stencil/react-output-target to ^1.5.2

ShaneK:
Issue number: internal

---------

`@stencil/react-output-target` is pinned to exactly `1.5.0` in `core/package.json` and `packages/react/package.json`. This was a defensive pin after `1.5.1` shipped a regression that dropped the `Partial<C>` wrapper from `StencilProps`, which broke everything in our react apps by [requiring all props to be defined](https://github.com/ionic-team/ionic-framework/actions/runs/25007199736/job/73233599812?pr=30831). Exact-pinning meant we couldn't pick up future patches in the `1.5.x` line.

## What is the new behavior?
Both `package.json` ranges move to `^1.5.2`. Upstream fully reverted [PR #788](https://github.com/stenciljs/output-targets/pull/788) in [`1.5.2`](https://github.com/stenciljs/output-targets/blob/main/packages/react/CHANGELOG.md), so the generator emits `Components.${tag}` again and the runtime restores `Partial<C>`. Regenerating `packages/react/src/components/components.ts` against the `1.5.2` generator and running prettier produces a file byte-identical to the one committed under `1.5.0`, so the published shape is unchanged.

- [ ] Yes
- [x] No

## Other information

The original break, the upstream conversation, and the rationale for staying on the `1.5.x` line (rather than taking the new "errors on truly required props" behavior) are all in the [PR #788 thread](https://github.com/stenciljs/output-targets/pull/788).

### DIFF
--- a/core/package-lock.json
+++ b/core/package-lock.json
@@ -25,7 +25,7 @@
         "@rollup/plugin-node-resolve": "^8.4.0",
         "@rollup/plugin-virtual": "^2.0.3",
         "@stencil/angular-output-target": "^0.10.0",
-        "@stencil/react-output-target": "1.5.0",
+        "@stencil/react-output-target": "^1.5.2",
         "@stencil/sass": "^3.0.9",
         "@stencil/vue-output-target": "0.10.8",
         "@types/jest": "^29.5.6",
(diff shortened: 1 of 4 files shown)


thetaPC (approved):
LGTM, thank you!

PR #31105, opened 2026-04-30
chore(vercel): add react-router preview build and fix dir-index navigation

ShaneK:
Issue number: internal

---------

Currently, reviewing react router based PRs isn't possible using the vercel preview link

## What is the new behavior?

This PR adds support for the preview link to support react router 6 and tries to fix some weird edge cases where navigating to a route would sometimes not show the tests in that route

- [ ] Yes
- [x] No

## Other information

The new React Router preview app shares the React package gate (`REACT_PKG_OK`) because `@ionic/react-router` is built inside `build_react_pkgs()`. Test app builds run in parallel under that gate, then `wait` reports per-app failures.

Preview URLs (once Vercel builds this branch):

- React Router test app: https://ionic-framework-git-chore-vercel-add-rr-ionic1.vercel.app/react-router/
- Dir-index nav fix demo: https://ionic-framework-git-chore-vercel-add-rr-ionic1.vercel.app/src/components/progress-bar/test (no trailing slash; clicking `basic/`, `buffer/`, `determinate/` should now land on the test page instead of the root landing)

### DIFF
--- a/core/scripts/vercel-build.sh
+++ b/core/scripts/vercel-build.sh
@@ -3,12 +3,14 @@
 # Vercel preview build script
 #
 # Builds core component tests (same as before) plus framework test apps
-# (Angular, React, Vue) so they're all accessible from a single preview URL.
+# (Angular, React, React Router, Vue) so they're all accessible from a single
+# preview URL.
 #
-# Core tests:        /src/components/{name}/test/{scenario}
-# Angular test app:  /angular/
-# React test app:    /react/
-# Vue test app:      /vue/
+# Core tests:             /src/components/{name}/test/{scenario}
+# Angular test app:       /angular/
+# React test app:         /react/
+# React Router test app:  /react-router/
+# Vue test app:           /vue/
 #
 set -e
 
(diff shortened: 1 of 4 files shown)


core/scripts/vercel-build.sh
```diff
@@ -254,8 +262,29 @@ build_vue_test() {
...
+    echo "[react-router] No test app found, skipping."
+    return 0
+  }
+  echo "[react-router] Building ${APP}..."
+
+  cd "${REPO_ROOT}/packages/react-router/test"
+  ./build.sh "${APP}"
+  cd "build/${APP}"
+  # The react-router test apps require --legacy-peer-deps (mixed react versions
+  # across @ionic/react peer ranges).
```
thetaPC, line 277: Assuming there's a ticket to remove the `--legacy-peer-deps`, let's add the TODO here.

thetaPC (approved):
LGTM, thank you!

core/scripts/vercel-build.sh
ShaneK: There's not, and I don't really want to. Any time we upgrade react router we'd have to add it back and there's no harm in having it at this point.

PR #31107, opened 2026-04-30
chore(deps): update @stencil/vue-output-target

OS-jacobbell:

Ionic's vue-output-target uses Stencil's vue-output-target version 0.10.7 and 0.10.8, which are close to a year old.

## What is the new behavior?

- Update to `@stencil/vue-output-target@0.13.1`.
- Change Vue model event names from kebab-case to camelCase. Stencil vue-output-target no longer converts them automatically.
- Generated proxies file has been updated by latest vue-output-target.

- [ ] Yes
- [X] No


packages/vue/package.json
```diff
@@ -68,8 +68,8 @@
     "vue-router": "^4.0.16"
   },
   "dependencies": {
-    "@ionic/core": "^8.8.5",
-    "@stencil/vue-output-target": "0.10.7",
+    "@ionic/core": "^8.8.4",
```
ShaneK: This looks like a bad merge conflict resolution and should be moved back to 8.8.5 and npm install ran again to update package-lock

ShaneK (changes requested):
Just a minor issue, but it is a blocker unfortunately

ShaneK (approved):
LGTM!

brandyscarney (approved):
Looks good! ✅

PR #31110, opened 2026-05-01
chore(docker): avoid shell interpretation and uncontrolled path expansion

thetaPC:
Issue number: N/A

---------

When passing a Playwright flag with a space in its value (e.g. `--project='Mobile Safari'`), the shell strips the quotes before Node receives the argument. Since `docker.mjs` passed args to `execa` with `shell: true`, the unquoted space caused the argument to be split and Playwright never received the correct value.

Additionally, arguments like `-e DISPLAY=${display}`, `-v ${displayVolume}`, and `--mount` were constructed as combined strings and passed through shell interpretation, meaning special characters in those values (e.g. spaces in an absolute path) could cause the command to fail unexpectedly. This caused CodeQL to trigger with security issues.

## What is the new behavior?

Each Docker argument is now passed as a separate array element to `execa` without `shell: true`, so values are forwarded directly to Docker without shell re-interpretation. This preserves spaces within argument values (e.g. `--project='Mobile Safari'`) and prevents uncontrolled expansion of paths and environment variable values.

The security issues stated by CodeQL has been addressed.

- [ ] Yes
- [x] No

## Other information

Verify that the following command works: `npm run test.e2e.docker datetime/test/basic -- -g 'IO fallback' --project='Mobile Safari' --repeat-each=20`

### DIFF
--- a/core/scripts/docker.mjs
+++ b/core/scripts/docker.mjs
@@ -1,7 +1,7 @@
+import chalk from 'chalk';
 import { execa } from 'execa';
 import * as fs from 'fs';
 import { resolve } from 'path';
-import chalk from 'chalk';
 
 const removeNewline = (string) => {
   return string.replace(/(\r\n|\n|\r)/gm, "");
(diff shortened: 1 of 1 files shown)


brandyscarney (approved):
The command runs but the test fails. Approving with the expectation that this behavior is expected.

Image

thetaPC:
@brandyscarney I didn't have the latest changes from `main` at the time so it didn't have the fix that Shane added. I can confirm that it fails on my local before the sync. After I updated my branch, the tests are passing without any issues.

PR #31112, opened 2026-05-05
fix(segment): segment drag would set disabled segment button checked

os-davidlourenco:

---------

- When the first segment is disabled and the last item is selected, if the user swipes the `segment-view`, the first `segment-button` gets checked even if it is disabled, and the view is presenting other `segment-view`.

## What is the new behavior?

- A validation was added to ensure that the `segment-button` can only be checked if it is not disabled

- [ ] Yes
- [x] No


ShaneK (approved):
Looks good to me!

PR #31113, opened 2026-05-05
fix(segment-view): prevent NaN/Infinity scrollRatio when a single content item is present

OS-susmitabhowmik:

---------

Currently if there is a single content item present the `scrollRatio` is computed as `NaN` or `Infinity` resulting in a console error.

Screenshot 2026-05-05 at 1 42 51 PM

## What is the new behavior?
Updates `handleScroll` in `ion-segment-view` to skip `handleScroll` if `max <= 0`

- [ ] Yes
- [ ] No

### DIFF
--- a/core/src/components/segment-view/segment-view.tsx
+++ b/core/src/components/segment-view/segment-view.tsx
@@ -48,7 +48,12 @@ export class SegmentView implements ComponentInterface {
     const { scrollLeft, scrollWidth, clientWidth } = ev.target as HTMLElement;
     const max = scrollWidth - clientWidth;
     // When only one content item is present max is 0 — skip to avoid NaN/Infinity scrollRatio.
-    if (max <= 0) return;
+    // Still reset the timeout so isManualScroll isn't cleared prematurely if setContent
+    // started the timer and a stray scroll event arrives on a non-overflowing element.
+    if (max <= 0) {
+      this.resetScrollEndTimeout();
+      return;
+    }
     const scrollRatio = (isRTL(this.el) ? -1 : 1) * (scrollLeft / max);
 
     this.ionSegmentViewScroll.emit({
--- a/core/src/components/segment-view/segment-view.tsx
+++ b/core/src/components/segment-view/segment-view.tsx
@@ -47,6 +47,8 @@ export class SegmentView implements ComponentInterface {
   handleScroll(ev: Event) {
     const { scrollLeft, scrollWidth, clientWidth } = ev.target as HTMLElement;
     const max = scrollWidth - clientWidth;
+    // When only one content item is present max is 0 — skip to avoid NaN/Infinity scrollRatio.
+    if (max <= 0) return;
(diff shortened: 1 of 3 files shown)


core/src/components/segment-view/segment-view.tsx
```diff
@@ -47,6 +47,8 @@ export class SegmentView implements ComponentInterface {
   handleScroll(ev: Event) {
     const { scrollLeft, scrollWidth, clientWidth } = ev.target as HTMLElement;
     const max = scrollWidth - clientWidth;
+    // When only one content item is present max is 0 — skip to avoid NaN/Infinity scrollRatio.
+    if (max <= 0) return;
```
ShaneK: Mind adding a quick Playwright test for this in `test/basic/segment-view.e2e.ts`? A single `ion-segment-content` that scrolls and asserts `ionSegmentViewScroll` doesn't fire (or fires with a finite `scrollRatio`) would lock the regression in. Without one, the next refactor of `handleScroll` can put the `NaN` back silently.

ShaneK: The early return also skips `resetScrollEndTimeout()`. If `setContent` set `isManualScroll = false` and started the timer, then a scroll event lands in the `max <= 0` branch, the timer doesn't get extended on this event and `isManualScroll` could clear early. I think it's fine in practice since a non-overflowing element shouldn't really fire scroll events, but worth a sanity check to just go ahead and `resetScrollEndTimeout` in the if body.

ShaneK (changes requested):
Nice! Just a couple things to make this bullet-proof

core/src/components/segment-view/segment-view.tsx
OS-susmitabhowmik: Great call! Added a regression test in these commits: b761da8caa7f773604389e4335fed987a50cd65a and 0302f23e659eb2f2df2d11b869a64f47e8255b1b

OS-susmitabhowmik: Updated to include `resetScrollEndTimeout` within the if block as a part of this commit dcbf3917ffad61db62d4d9486bb58d43a659aa7e

core/src/components/segment/segment.tsx
```diff
@@ -388,6 +388,7 @@ export class Segment implements ComponentInterface {
 
     const dispatchedFrom = ev.target as HTMLElement;
     const segmentViewEl = this.segmentViewEl as EventTarget;
+
```
ShaneK:
```suggestion
```
This feels unrelated

ShaneK (approved):
Looks good! Thanks for adding tests 🚀

core/src/components/segment/segment.tsx
OS-susmitabhowmik: Fixed! c2f6f39e38e5678d4eed374ac3d95cc9e69f3532

PR #31115, opened 2026-05-05
feat(tab): modular ionic migration and lazy-load specs

thetaPC:

---------

`ion-tab` retains a `@virtualProp theme` JSDoc annotation that is no longer appropriate after the Modular Ionic migration. Also no spec tests exist for the component's lazy-load behavior.

## What is the new behavior?

- Removes the `@virtualProp theme` from `ion-tab`
- Adds spec tests for the lazy-load guard: verifies `attachComponent` is only called once across multiple `setActive()` calls, and documents the permanent-failure behavior when the delegate rejects on the first attempt

- [ ] Yes
- [x] No

## Other information

Since I was already here, I added spec tests that would be beneficial.

Previews
- [Tabs: Basic](https://ionic-framework-git-fw-6906-ionic1.vercel.app/src/components/tabs/test/basic/)


core/src/components/tab/test/tab.spec.ts
```diff
@@ -0,0 +1,45 @@
...
+    const attachViewToDom = jest.fn().mockRejectedValue(new Error('attach failed'));
+    tabEl.delegate = mockDelegate(attachViewToDom);
+
+    // First call rejects because the delegate throws. The loaded flag is already
+    // set to true before the attempt, so the failure is permanent.
+    await expect(tabEl.setActive()).rejects.toThrow('attach failed');
+    await tabEl.setActive();
+
+    expect(attachViewToDom).toHaveBeenCalledTimes(1);
+  });
```
ShaneK, line 43: Is the "permanent failure" behavior actually what we want here, or just what the code currently does? The synchronous `try/catch` in `prepareLazyLoaded` can't catch the async rejection from `attachComponent`, and `this.loaded = true` is set before the await, so any first-attempt failure locks the tab out for good. A passing test on this makes a future fix (resetting `loaded` on rejection, or moving the catch around an `await`) look like a regression. Worth either filing a follow-up and referencing it here, or renaming the test so it reads "documents current behavior" rather than "asserts a contract"?

ShaneK (approved):
Looks good to me! Just highlighting an issue with a test that I think could be improved

core/src/components/tab/test/tab.spec.ts
thetaPC: I added TODOs to go back and fix this on a separate ticket: https://github.com/ionic-team/ionic-framework/pull/31115/changes/575d35532d18182e6424a94b7601e6bfa526f816

brandyscarney (approved):
LGTM

PR #31116, opened 2026-05-06
feat(tabs): migrate to Modular Ionic

thetaPC:

---------

`ion-tabs` uses a legacy `@import` in its stylesheet, has a separate `tabs-interface.ts` file (hyphen naming), and carries a `@virtualProp theme` doc that is no longer applicable after the Modular Ionic migration. An error message in the `getTab` helper always logs `"undefined"` instead of the actual tab id that was passed.

## What is the new behavior?

- Renamed `tabs-interface.ts` -> `tabs.interfaces.ts` to match the format in Modular Ionic
- Converted `@import "native.globals"` to `@use "../../themes/mixins" as mixins`
- Removed the `@virtualProp theme` JSDoc
- Fixed error message in `getTab`. Now correctly logs the tab id passed by the caller instead of always logging `"undefined"`
- Added spec tests

- [ ] Yes
- [x] No

## Other information

Previews
- [Basic](https://ionic-framework-git-fw-6909-ionic1.vercel.app/src/components/tabs/test/basic/)
- [Placements](https://ionic-framework-git-fw-6909-ionic1.vercel.app/src/components/tabs/test/placements/)


core/src/components/tabs/test/basic/index.html
```diff
@@ -42,8 +42,6 @@
           </ion-header>
           <ion-content class="ion-padding">
             <h1>Tab One</h1>
-            <button class="expand" onclick="updateBadgeCount()">Update Badge Count</button>
-            <button class="expand" onclick="updateBadgeColor()">Update Badge Color</button>
```
thetaPC, line 46: These buttons weren't doing anything. I do plan on adding `updateBadgeCount` to the `tab-bar/test/badge` page when we migrate.

core/src/components/tabs/test/basic/index.html
```diff
@@ -129,7 +127,7 @@ <h1>Hidden Tab</h1>
             </ion-toolbar>
           </ion-header>
           <ion-content class="ion-padding">
-            Page Four
+            <h1>Page Four</h1>
```
thetaPC: Following the setup of the other 3 pages.

core/src/components/tabs/tabs.tsx
```diff
@@ -158,7 +157,7 @@ export class Tabs implements NavOutlet {
 
   private setActive(selectedTab: HTMLIonTabElement): Promise<void> {
     if (this.transitioning) {
-      return Promise.reject('transitioning already happening');
+      return Promise.reject(new Error('transitioning already happening'));
```
thetaPC, line 160: This change attaches a stack trace to the rejection, making it easier to identify where the error originated when debugging.

core/src/components/tabs/tabs.tsx
```diff
@@ -236,7 +233,7 @@ const getTab = (tabs: HTMLIonTabElement[], tab: string | HTMLIonTabElement): HTM
   const tabEl = typeof tab === 'string' ? tabs.find((t) => t.tab === tab) : tab;
 
   if (!tabEl) {
-    printIonError(`[ion-tabs] - Tab with id: "${tabEl}" does not exist`);
+    printIonError(`[ion-tabs] - Tab with id: "${tab}" does not exist`);
```
thetaPC, line 236: This kept showing as `[ion-tabs] - Tab with id: "undefined" does not exist` so I fixed it by passing the correct variable.

core/src/components/tabs/tabs.scss
```diff
@@ -12,7 +12,7 @@
   height: 100%;
 
   contain: layout size style;
-  z-index: $z-index-page-container;
+  z-index: 0;
```
ShaneK: Was inlining `0` here instead of using `$z-index-page-container` deliberate? `nav.scss` and `router-outlet.scss` are still on the variable, so `tabs` becomes the lone magic number until they migrate too. Is the plan to define `$z-index-page-container` somewhere reachable from `mixins.scss`, or to inline `0` everywhere once those two migrate as well? Either way it works, but a short comment noting it tracks the page-container z-index would keep the intent visible during the transition.

ShaneK (approved):
Looks good to me! Just a nit

core/src/components/tabs/tabs.scss
thetaPC: I create a common globals file. There's going to be duplicate imports of the z index until we can migrate the rest of the components. https://github.com/ionic-team/ionic-framework/pull/31116/changes/51f637f82d8b2c8ea90973191cbeb600e5d6e8a2

core/src/components/nav/nav.scss
```diff
@@ -1,3 +1,4 @@
+@use "../../themes/common.globals" as globals;
 @import "../../themes/native/native.globals";
```
brandyscarney, line 2: Why was this not changed to `@use`?

core/src/components/router-outlet/router-outlet.scss
```diff
@@ -1,3 +1,4 @@
+@use "../../themes/common.globals" as globals;
```
brandyscarney, line 1: Why was this not changed to `@use`?

core/src/components/tabs/tabs.interfaces.ts
brandyscarney, line 1: Was adding the `s` to make it `tabs.interfaces.ts` instead of `tabs.interface.ts` intentional?

core/src/themes/common.globals.scss
```diff
@@ -0,0 +1,8 @@
+// Common globals
+//
+// Variables shared across all themes.
+// --------------------------------------------------
+
+// Z-Index
+// --------------------------------------------------
+$z-index-page-container: 0;
```
brandyscarney, line 8: Shouldn't this be a token value?

core/src/components/tabs/test/placements/index.html
```diff
@@ -2,7 +2,7 @@
 <html lang="en" dir="ltr">
   <head>
     <meta charset="UTF-8" />
-    <title>Tab - Basic</title>
+    <title>Tab - Placements</title>
```
brandyscarney: Nit: shouldn't these titles technically be `Tabs - *`?

```suggestion
    <title>Tabs - Placements</title>
```

core/src/components/tabs/test/basic/index.html
brandyscarney: Nit: shouldn't these titles technically be `Tabs - *`?


```suggestion
    <title>Tabs - Basic</title>
```

core/src/components/tabs/test/basic/index.html
brandyscarney:
```suggestion
            <h1>Tab Four</h1>
```

Technically the other pages say "Tab" instead of "Page"

core/src/components/tabs/test/basic/index.html
brandyscarney:
```suggestion
              <ion-title>Tab Four</ion-title>
```

The other pages use `Tab` instead of `Page` as their toolbar title.

brandyscarney (approved):
Looks good! Left some comments but I don't think any are blockers. Approving. 👍

core/src/components/nav/nav.scss
thetaPC: I'm trying to minimize the changes in other components as much as possible. The `use` migration will be done when we migrate `nav` to ionic modular. This is just a precaution in case an issue occurs. I don't want to spend more time on components that are not in the scope of tickets.

core/src/components/router-outlet/router-outlet.scss
thetaPC: https://github.com/ionic-team/ionic-framework/pull/31116#discussion_r3259218114

core/src/components/tabs/tabs.interfaces.ts
thetaPC: Yes, we've been doing that with all the components we've migrated. Should we be using singular? If so, I can make those changes in a different PR.

core/src/themes/common.globals.scss
thetaPC: No, it should be a 1:1 of the [value](https://github.com/ionic-team/ionic-framework/blob/5fda2dd16578c2db583ac4e7a110ef6394033685/core/src/themes/native/native.globals.scss#L57) in `native.global.scss` as that file will eventually be removed once we are done with the migration. However, we need to keep some variables/values since they are being shared in all themes.

core/src/components/tabs/test/placements/index.html
thetaPC: https://github.com/ionic-team/ionic-framework/pull/31116/commits/54c71611ddd7749487d5b57ec5dabac28eaa91fe

core/src/components/tabs/test/basic/index.html
thetaPC: https://github.com/ionic-team/ionic-framework/pull/31116/changes/61b64bb0e2d7ea483a904e6b31549f2bc663f8b2

thetaPC: https://github.com/ionic-team/ionic-framework/pull/31116/changes/8c08957a3c41fcdd84e5b71fef9cdb27a055c288

core/src/components/tabs/test/basic/index.html
thetaPC: https://github.com/ionic-team/ionic-framework/pull/31116/changes/5d57b05ddd21928160716587ba2e99abf7d53788

PR #31125, opened 2026-05-07
fix(select): only emit ionChange when value changes

ShaneK:
Issue number: internal

---------

In certain circumstances, `ion-select` emits `ionChange` every time the overlay is confirmed, even when the user picks the option that was already selected. The `popover` and `modal` interfaces don't have this issue because they delegate selection to `ion-radio-group` / `ion-checkbox`, which already de-dupe. The mismatch is in `select.tsx`: the alert OK handler and each action-sheet button handler call `setValue()` unconditionally, and `setValue()` always emits.

This contradicts the documented behavior ("Emitted when the value has changed") and is inconsistent with `ion-input`, `ion-textarea`, and the other select interfaces.

## What is the new behavior?

`setValue()` now compares the incoming value against the current value and early-returns when they match, so `ionChange` only fires on an actual change. The comparison honors `compareWith` for object values and treats `multiple` arrays as equal when they contain the same elements regardless of order. Behavior across all four interfaces is now consistent.

- [x] Yes
- [ ] No

Apps that relied on `ionChange` firing on every alert/action-sheet confirmation, even without a value change, will need to listen for `ionDismiss` (or the alert/action-sheet's own dismiss event) instead. This aligns alert/action-sheet with how `popover` and `modal` already behave, and with how every other Ionic form control handles `ionChange`.

## Other information

Preview pages (open the alert or action-sheet, click OK or tap the already-selected option, watch the console):
- https://ionic-framework-git-fw-6983-ionic1.vercel.app/src/components/select/test/basic

This goes along with the breaking change documentation in ionic docs, here:
- https://github.com/ionic-team/ionic-docs/pull/4490

### DIFF
--- a/BREAKING.md
+++ b/BREAKING.md
@@ -18,6 +18,7 @@ This is a comprehensive list of the breaking changes introduced in the major ver
 - [Components](#version-9x-components)
   - [Legacy Picker](#version-9x-legacy-picker)
   - [Router Outlet](#version-9x-router-outlet)
+  - [Select](#version-9x-select)
 - [Framework Specific](#version-9x-framework-specific)
   - [React](#version-9x-react)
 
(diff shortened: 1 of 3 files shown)


thetaPC (approved):
LGTM

PR #31134, opened 2026-05-11
fix(vue-router): keep tab back button working after history rewrite

ShaneK:
Issue number: internal

---------

After the user navigates back in browser history and then pushes a new route, `ion-back-button` on a tab child page silently fails.

## What is the new behavior?

When the computed history delta would be non-negative, fall back to `router.replace` on `prevInfo.pathname` (or to `defaultHref` if `prevInfo` has no pathname). `incomingRouteParams` was already set to a pop/back above, so the view stack still animates the transition as a back navigation, and the params are consumed by the resulting `afterEach` hook instead of leaking into the next navigation.

- [ ] Yes
- [x] No

## Other information

### DIFF
--- a/packages/vue-router/src/router.ts
+++ b/packages/vue-router/src/router.ts
@@ -183,7 +183,32 @@ export const createIonRouter = (
            * is not good because we would have two /tabs/tab1/child1 entries
            * separated by a /tabs/tab1/child2 entry.
            */
-          router.go(prevInfo.position - routeInfo.position);
+          const positionDelta = prevInfo.position! - routeInfo.position!;
+          if (positionDelta < 0) {
+            router.go(positionDelta);
+          } else if (prevInfo.pathname) {
+            /**
+             * prevInfo's history position was wiped when the user went
+             * back then pushed a new route (FW-6472), so router.go can't
+             * reach it. Replace falls through to afterEach with the
+             * pop/back `incomingRouteParams` set above, which preserves
+             * the back animation and consumes the params so they don't
+             * leak into the next navigation. We replace even when
+             * `positionDelta === 0` for the same consumption reason.
+             */
+            router.replace({
+              path: prevInfo.pathname,
+              query: parseQuery(prevInfo.search),
+            });
+          } else {
+            /**
+             * prevInfo has no pathname (synthesized root entry). Route
+             * to `defaultHref` so the pop/back `incomingRouteParams`
+             * set above gets consumed instead of leaking into the next
+             * navigation.
+             */
+            handleNavigate(defaultHref, "pop", "back", routerAnimation);
+          }
         }
       } else {
         handleNavigate(defaultHref, "pop", "back", routerAnimation);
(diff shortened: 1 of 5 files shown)


packages/vue-router/src/router.ts
```diff
@@ -183,7 +183,32 @@ export const createIonRouter = (
...
            * separated by a /tabs/tab1/child2 entry.
            */
-          router.go(prevInfo.position - routeInfo.position);
+          const positionDelta = prevInfo.position! - routeInfo.position!;
+          if (positionDelta < 0) {
+            router.go(positionDelta);
+          } else if (prevInfo.pathname) {
+            /**
+             * prevInfo's history position was wiped when the user went
+             * back then pushed a new route (FW-6472), so router.go can't
```
thetaPC:
```suggestion
             * back then pushed a new route, so router.go can't
```

packages/vue/test/base/tests/e2e/playwright/tabs-back-button.spec.ts
```diff
@@ -0,0 +1,70 @@
+import { test, expect } from './utils/test-base';
+import { ionBackClick, ionPageVisible, ionPageHidden, ionPageDoesNotExist, tabClick } from './utils/test-utils';
+
+/**
+ * Regression coverage for FW-6472. When the user has child pages open in
```
thetaPC:
```suggestion
 * When the user has child pages open in
```

packages/vue/test/base/tests/e2e/playwright/tabs-back-button.spec.ts
```diff
@@ -0,0 +1,70 @@
...
+ * two tabs and switches between them, the back button on the active tab's
+ * child must pop within that tab's stack, not jump to the other tab.
+ */
+test.describe('Tabs: back button after switching between tabs with child pages', () => {
+  test('back button on tab2 child still returns to tab2 root after multiple tab switches (FW-6472 scenario A)', async ({ page }) => {
```
thetaPC:
```suggestion
  test('back button on tab2 child still returns to tab2 root after multiple tab switches', async ({ page }) => {
```

packages/vue/test/base/tests/e2e/playwright/tabs-back-button.spec.ts
```diff
@@ -0,0 +1,70 @@
...
+    await ionPageHidden(page, 'tab1childone');
+
+    await ionBackClick(page, 'tab2childone');
+
+    await expect.poll(() => new URL(page.url()).pathname).toBe('/tabs/tab2');
+    await ionPageVisible(page, 'tab2');
+    await ionPageDoesNotExist(page, 'tab2childone');
+  });
+
+  // Scenario B: go back on tab1's child first, then re-enter tab2. Back on
```
thetaPC:
```suggestion
  // Go back on tab1's child first, then re-enter tab2. Back on
```

thetaPC (approved):
LGTM, with minor nits

PR #31137, opened 2026-05-12
fix(vue-router): set pushedByRoute on same-tab push with direction none

### LINKED ISSUE #24074: bug: Ionic React routerDirection "none" causes BackButton to go to defaultHref
### Prerequisites

### Ionic Framework Version

- [ ] v4.x
- [X] v5.x
- [ ] v6.x

### Current Behavior

I think based on my understanding of the `routerDirection` prop, I think this may be a bug-

When routing via Item, Button, etc with `routerLink` and `routerDirection` if `routerDirection` is set to "none" the back button on the routed to page goes to the `defaultHref` instead of the page you came from. 

https://user-images.githubusercontent.com/14936212/137333122-58991d36-1226-410a-b52f-85be66c35d5d.mov

### Expected Behavior

BackButton should use history to determine where to go back to, even with "none" as `routerDirection`

### Steps to Reproduce

Example REPO

From Tab 1 - 
Button 1 routes as default with a forward direction and then you go back to Tab 1. 
Button 2 routed with direction "none" and will then take you back to Tab 2 (the `defaultHref` for the Back Button)

### Code Reproduction URL

https://github.com/babycourageous/ionic-router-direction-none-bug

### Ionic Info

Ionic:

   Ionic CLI       : 6.17.1 (/Users/renedellefont/.fnm/node-versions/v14.17.6/installation/lib/node_modules/@ionic/cli)
   Ionic Framework : @ionic/react 5.8.4

Capacitor:

   Capacitor CLI      : 3.2.5
   @capacitor/core    : 3.2.5

Utility:

   native-run  : 1.5.0

System:

   NodeJS : v14.17.6 (/Users/renedellefont/.fnm/node-versions/v14.17.6/installation/bin/node)
   npm    : 6.14.15
   OS     : macOS Big Sur

### Additional Information

If this is the intended behavior I can PR to the docs to perhaps clarify that section. Thanks!

ShaneK:
Issue number: resolves #24074

---------

In vue-router, a `router.push` performed with `routerDirection="none"` doesn't set `pushedByRoute` on the resulting route info. The next page's `ion-back-button` then can't find a previous entry and falls back to `defaultHref` instead of going back through history.

The same bug was previously fixed for react-router in v6, this is the vue version of the fix (Angular doesn't need this fix because the router works very differently)

## What is the new behavior?

`createIonRouter` now sets `pushedByRoute` from the leaving location when the navigation is `routerAction === "push"`, `routerDirection === "none"`, and stays within the same tab context. Cross-tab pushes with direction `none` still go through the existing tab-aware branch, which has different `pushedByRoute` semantics

After the fix, `ion-back-button` returns to the prior page through history and only falls back to `defaultHref` when there's genuinely no history to pop.

- [ ] Yes
- [X] No

## Other information

New Playwright spec `direction-none-back.spec.ts` covers both the `forward` and `none` paths and asserts the back button lands on Page A, not the `defaultHref` fallback

[Test Page](https://ionic-framework-git-fw-7145-ionic1.vercel.app/vue/direction-none-back/a)

### DIFF
--- a/packages/vue-router/src/router.ts
+++ b/packages/vue-router/src/router.ts
@@ -383,6 +383,25 @@ export const createIonRouter = (
             routeInfo.tab
           );
           routeInfo.pushedByRoute = lastRoute?.pushedByRoute;
+        } else if (
+          routeInfo.routerAction === "push" &&
+          routeInfo.routerDirection === "none" &&
+          routeInfo.tab === leavingLocationInfo.tab
+        ) {
+          /**
+           * Same-tab push with direction "none" still needs pushedByRoute so
+           * ion-back-button uses history instead of falling back to defaultHref.
+           * Cross-tab pushes hit the branch above.
+           *
+           * Skip when the candidate equals the current pathname (e.g. /a?x=1 ->
+           * /a?x=2) to avoid a self-loop on back. Same guard as the replace
+           * branch below.
+           */
+          const candidate = leavingLocationInfo.pathname;
+          routeInfo.pushedByRoute =
+            candidate !== "" && candidate !== routeInfo.pathname
+              ? candidate
+              : undefined;
         } else if (routeInfo.routerAction === "replace") {
           /**
            * When replacing a route, we want to make sure we select the current route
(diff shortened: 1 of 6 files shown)


thetaPC (approved):
LGTM

PR #31138, opened 2026-05-12
feat(select): add labelPlacement prop, justify prop, and rich content styles for select options

thetaPC:

---------

- `ion-select-option` doesn't expose `labelPlacement` or `justify` props
- The ionic theme's overlay rendering doesn't match the latest design
- `<ion-icon>` placed inside an `ion-select-option` renders empty in Vue/React (and other frameworks that bind custom-element props as DOM properties)
- Slotted children in `select-option-start` / `select-option-end` aren't size-capped consistently: oversized content
- Several theme style files (`alert.{ios,md,ionic}.scss`, `action-sheet.ionic.scss`, `select-{modal,popover}.{ios,md,ionic}.scss`) import the wrong shared partials 

## What is the new behavior?

- Add `labelPlacement` and `justify` props to `ion-select-option` and pass them through to alert, popover, and modal overlay paths
- Update the ionic theme's alert, popover, and modal overlay styles to match the current design 
- Add per-theme slot-size limitations so oversized slotted content is capped to keep option rows visually uniform within an overlay
- Audit and correct the `@use` / `@import` chains in the affected theme stylesheets so each theme file pulls in the right shared partials (mixins, theme globals, common base) and no longer relies on transitive imports that weren't guaranteed
- Add `reflectPropertiesToAttributes` to `core/src/utils/sanitization/` and call it from `getOptionContent` immediately before `cloneNode`. The helper mirrors a registered set of custom-element DOM properties (`icon`, `name`, `src`, `ios`, `md` on `ion-icon`) onto attributes so the cloned overlay copy renders correctly regardless of how the framework bound the prop.
- Add test pages and Playwright snapshots for the overlays

- [ ] Yes
- [x] No

## Other information

Dev build: `8.8.9-dev.11779411201.1a483e09`

Preview:
- [Select: basic](https://ionic-framework-git-fw-7290-ionic1.vercel.app/src/components/select/test/basic/)
- [Select: rich content option](https://ionic-framework-git-fw-7290-ionic1.vercel.app/src/components/select/test/rich-content-option/)
- [Select option: label placement](https://ionic-framework-git-fw-7290-ionic1.vercel.app/src/components/select-option/test/label-placement/)
- [Alert: basic](https://ionic-framework-git-fw-7290-ionic1.vercel.app/src/components/alert/test/basic/)
- [Alert: label placement](https://ionic-framework-git-fw-7290-ionic1.vercel.app/src/components/alert/test/label-placement/)
- [Alert: states](https://ionic-framework-git-fw-7290-ionic1.vercel.app/src/components/alert/test/states/)
- [Action sheet: states](https://ionic-framework-git-fw-7290-ionic1.vercel.app/src/components/action-sheet/test/states/)
- [Select modal: states](https://ionic-framework-git-fw-7290-ionic1.vercel.app/src/components/select-modal/test/states/)
- [Select popover: states](https://ionic-framework-git-fw-7290-ionic1.vercel.app/src/components/select-popover/test/states/)


core/src/components/action-sheet/test/states/action-sheet.e2e.ts-snapshots/action-sheet-states-diff-ionic-md-ltr-light-Mobile-Firefox-linux.png
thetaPC, line 1: Hovered styles in Playwright Firefox do not display because we aren't truly mimicking a [supported device](https://github.com/ionic-team/ionic-framework/blob/a483e09b3b7865d9eab7cf0fa9f727d6b729cf6e/core/playwright.config.ts#L18-L26).

core/src/components/action-sheet/test/states/action-sheet.e2e.ts
```diff
@@ -0,0 +1,40 @@
...
+configs({ directions: ['ltr'], modes: ['ios', 'md', 'ionic-md'] }).forEach(({ config, screenshot, title }) => {
+  test.describe(title('action sheet: states'), () => {
+    /**
+     * `(any-hover: hover)` evaluates to "none" in all three mobile-emulated
+     * projects, suppressing the hover rules:
+     *
+     * - Chromium and WebKit suppress it because of hasTouch and isMobile.
+     * - Headless Firefox doesn't detect input devices and reports no hover
+     *   capability regardless of context options, so override it via
+     *   launchOptions.firefoxUserPrefs. Bit values: 4 = FINE (mouse),
```
thetaPC, line 17: Overriding it would lead to a global change so I didn't want to attempt that in this PR.

core/src/components/action-sheet/action-sheet.common.scss
```diff
@@ -209,14 +209,20 @@
...
     background: var(--button-background-focused);
 
     opacity: var(--button-background-focused-opacity);
   }
+
+  &.action-sheet-selected::after {
+    background: var(--button-background-focused, var(--button-background-selected));
+
+    opacity: var(--button-background-focused-opacity, var(--button-background-selected-opacity));
+  }
```
thetaPC, line 226: By restructuring it this way, we are making sure that this style will be just a copy and paste for the Modular Ionic migration.

core/src/components/action-sheet/action-sheet.ios.vars.scss
```diff
@@ -161,3 +161,6 @@ $action-sheet-ios-group-translucent-filter: saturate(280%) blur(20px);
 
 /// @prop - Filter of the translucent action-sheet button
 $action-sheet-ios-button-translucent-filter: saturate(120%);
+
+/// @prop - Maximum size of slotted children rendered in a select option's start/end slot
+$action-sheet-ios-select-option-slot-size: dynamic-font-max(24px, 2);
```
thetaPC: Based on what I found through iOS code and that it scales up.

core/src/components/action-sheet/action-sheet.md.scss
```diff
@@ -110,3 +110,16 @@
...
+// --------------------------------------------------
+
+/**
+ * Cap slotted children so they can't stretch the option
+ * row out of proportion, keeping rows visually uniform
+ * regardless of the content.
+ */
+.select-option-start > *,
+.select-option-end > * {
+  max-height: 24px;
```
thetaPC: Based on MD3

core/src/components/alert/test/basic/alert.e2e.ts-snapshots/alert-basic-ionic-md-ltr-light-Mobile-Chrome-linux.png
thetaPC, line 1: Ionic theme didn't have any screenshots for alert.

core/src/components/item/item.ionic.scss
```diff
@@ -111,14 +108,3 @@ slot[name="end"]::slotted(*) {
...
-// Item in Select Modal
-// --------------------------------------------------
-:host(.in-select-modal) {
-  --background-focused: #{globals.$ion-bg-neutral-subtlest-press};
-  --background-focused-opacity: 0;
-}
-
-:host(.in-select-modal.ion-focused) .item-native {
-  --border-radius: #{globals.$ion-border-radius-400};
-}
```
thetaPC, line 124: These can live under the select-modal.ionic.scss file

core/src/components/item/item.ionic.scss
thetaPC, line 1: Updated to match latest designs

core/src/components/action-sheet/action-sheet.ionic.scss
```diff
@@ -1,22 +1,136 @@
...
+
+  min-height: 52px;
+
+  text-align: start;
+
+  contain: content;
+  overflow: hidden;
+}
+
+.action-sheet-icon {
```
brandyscarney, line 95: Is this the right size for icons? It looks kind of small:

Image

core/src/components/select/test/rich-content-option/select.e2e.ts-snapshots/select-rich-content-action-sheet-md-ltr-Mobile-Chrome-linux.png
brandyscarney, line 1: Why is there no padding around this when the local test shows padding:

Image

core/src/components/select/test/rich-content-option/select.e2e.ts-snapshots/select-rich-content-action-sheet-ios-ltr-Mobile-Chrome-linux.png
brandyscarney, line 1: Why is there no padding around this when the local test shows padding:

Image

core/src/components/select/test/rich-content-option/select.e2e.ts-snapshots/select-rich-content-action-sheet-ionic-md-ltr-light-Mobile-Chrome-linux.png
brandyscarney, line 1: Why is there no padding around this when the local test shows padding:

Image

core/src/components/action-sheet/action-sheet.common.scss
brandyscarney, line 1: The inner label text is not vertically aligned for all themes:

Image

You need to add the following:

```scss
.action-sheet-button-label-text {
    display: flex;
    align-items: center;
}
```

It would probably look nicer to have a gap here as well.

core/src/components/action-sheet/action-sheet.ios.scss
brandyscarney, line 1: iOS looks kind of weird:

Image

Should we left-align the label when it has rich content? It isn't perfectly centered due to the slots and in this example it doesn't even align the label and description the same way:

Image

core/src/components/alert/alert.common.scss
brandyscarney, line 1: The inner label text is not vertically aligned for all themes:

Image

You need to add the following:

```scss
.alert-checkbox-label-text,
.alert-radio-label-text {
    display: flex;
    align-items: center;
}
```

It would probably look nicer to have a gap here as well.

core/src/components/select-modal/select-modal.common.scss
brandyscarney, line 1: The inner label text is not vertically aligned for all themes:

Image

You need to add the following:

```scss
.select-option-label-text {
    display: flex;
    align-items: center;
}
```

It would probably look nicer to have a gap here as well.

core/src/components/alert/alert.ionic.scss
brandyscarney, line 1: From what I can find in Figma, the buttons should not be uppercase:

Image

All example buttons use normal casing:

Image

Image

Image

core/src/components/alert/alert.tsx
```diff
@@ -674,12 +693,19 @@ export class Alert implements ComponentInterface, OverlayInterface {
                 'alert-radio-button': true,
                 'alert-tappable': true,
                 'alert-radio': true,
-                'ion-focusable': true,
+                'ion-focusable': !i.disabled,
```
brandyscarney, line 696: Why do we only check disabled here for certain components?

core/src/components/alert/alert.ios.vars.scss
```diff
@@ -316,3 +316,6 @@ $alert-ios-translucent-filter: saturate(180%) blur(20px);
 
 /// @prop - Height of the tappable inputs in the checkbox alert
 $alert-ios-tappable-height: $item-ios-min-height;
+
+/// @prop - Maximum size of slotted children rendered in a select option's start/end slot
+$alert-ios-select-option-slot-size: dynamic-font-max(24px, 2);
```
brandyscarney: Does this need to be a variable?

core/api.txt
brandyscarney, line 1: Shouldn't this be a `feat` since it adds new properties to the api?

core/src/components/action-sheet/action-sheet.md.scss
brandyscarney: The avatar looks pretty small compared to [md3](https://m3.material.io/components/dialogs/specs):

| Ionic | MD3 |
| --- | --- |
| ionic | md3 |

MD2 is also a bit bigger:

Image

but maybe because they don't show it by checkboxes?

core/src/components/action-sheet/action-sheet.ios.vars.scss
brandyscarney: Do you have a screenshot of how iOS looks?

Here is an example I took from my Contacts app:

CleanShot 2026-05-28 at 18 09 36

It looks closer to 30px.

core/src/components/alert/test/states/alert.e2e.ts-snapshots/alert-checkbox-states-ios-ltr-Mobile-Chrome-linux.png
brandyscarney, line 1: Is it intentional that iOS does not have a hover state?

core/src/components/alert/test/states/alert.e2e.ts-snapshots/alert-checkbox-states-md-ltr-Mobile-Chrome-linux.png
brandyscarney, line 1: Is it intentional that MD does not have a hover state?

core/src/components/select/test/rich-content-option/select.e2e.ts-snapshots/select-rich-content-alert-ionic-md-ltr-light-Mobile-Chrome-linux.png
brandyscarney, line 1: Should we decrease the padding right when rich content options are here? There's so much space between the checkbox and badge.

core/src/components/select/test/rich-content-option/select.e2e.ts-snapshots/select-rich-content-alert-md-ltr-Mobile-Safari-linux.png
brandyscarney, line 1: Why does this one look so different from Chrome and Firefox?

| Chrome | Safari |
| ---| ---|
| Chrome | Safari |

core/src/components/select-option/test/label-placement/select-option.e2e.ts
```diff
@@ -0,0 +1,96 @@
...
+      </ion-select>
+    `,
+    config
+  );
+};
+
+configs({ modes: ['md', 'ios', 'ionic-md'] }).forEach(({ config, screenshot, title }) => {
+  test.describe(title('select-option: label placement'), () => {
+    test.describe('alert interface', () => {
+      for (const placement of ['start', 'end'] as const) {
```
brandyscarney: Can we assign this to a constant to reuse? Also can't we wrap all the describes in one for loop instead of having each one with its own?

core/src/utils/select-option-render.tsx
```diff
@@ -12,48 +11,80 @@ interface RichContentOption extends RichContentOpt {
...
+ * element nodes become VNodes, and any other node types are skipped.
+ * @param keyPrefix - String prefix used to build a stable VNode key,
+ * so Stencil's diff can preserve elements across re-renders.
+ * @param index - Position of this node among its siblings. Combined
+ * with `keyPrefix` to form the final unique key.
+ * @returns The converted VNode, a text string, or `null` if the node
+ * type isn't supported.
  */
-const contentCache = new WeakMap<HTMLElement, HTMLElement>();
+const cloneToVNode = (node: Node, keyPrefix: string, index: number): VNode | string | null => {
```
brandyscarney: Can we add spec tests for this?

core/src/components/select/test/basic/index.html
brandyscarney, line 1: I don't think this is new with your changes but why does the select modal interface with multiple options only have the option to cancel but it keeps your selection 😂 

Image

core/src/utils/overlay-control-label.ts
brandyscarney, line 1: Do we plan to remove this file and set this in the config during the modular migration?

core/src/utils/sanitization/index.ts
brandyscarney, line 1: We should add spec tests for this verifying the proper elements are stripped.

core/src/components/select/select.tsx
```diff
@@ -1655,6 +1654,18 @@ const getOptionContent = (
...
+   * Mirror known custom-element properties (e.g. ion-icon's `icon`)
+   * onto attributes before cloning. Frameworks like Vue set these as
+   * DOM properties, which `cloneNode` doesn't copy, so without this
+   * step the cloned overlay copy renders without the prop's value.
+   */
+  nodes.forEach((n) => {
+    if (n.nodeType === Node.ELEMENT_NODE) {
+      reflectPropertiesToAttributes(n as Element);
+    }
+  });
```
brandyscarney, line 1668: Can we add tests verifying this behavior works?

core/src/components/select-popover/select-popover.tsx
```diff
@@ -82,7 +83,11 @@ export class SelectPopover implements ComponentInterface {
    * Dismisses the host popover that the `ion-select-popover`
    * is rendered within.
    */
-  private dismissParentPopover() {
+  private dismissParentPopover(isOptionDisabled = false) {
+    if (isOptionDisabled) {
```
brandyscarney: Why is this check now needed? Do we have tests verifying this behavior?

core/src/components/select-popover/select-popover.md.scss
brandyscarney, line 1: I have the same concern with the avatar size as I commented on action sheet.

core/src/components/select-popover/select-popover.ios.scss
brandyscarney, line 1: I have the same concern with the avatar size as I commented on action sheet.

core/src/components/select-popover/select-popover.ios.vars.scss
```diff
@@ -3,3 +3,6 @@
 
 // iOS Select Popover
 // --------------------------------------------------
+
+/// @prop - Maximum size of slotted children rendered in a select option's start/end slot
+$select-popover-ios-select-option-slot-size: dynamic-font-max(24px, 2);
```
brandyscarney: Do we need this as a variable?

core/src/components/select-popover/select-popover.common.scss
brandyscarney, line 1: The inner label text is not vertically aligned for all themes:

Image

You need to add the following:

```scss
.select-option-label-text {
    align-items: center;
}
```

core/src/components/select-modal/select-modal.tsx
```diff
@@ -49,7 +50,11 @@ export class SelectModal implements ComponentInterface {
 
   @Prop() options: SelectModalOption[] = [];
 
-  private closeModal() {
+  private closeModal(isOptionDisabled = false) {
+    if (isOptionDisabled) {
```
brandyscarney: Same question as popover - why is this suddenly needed and do we have tests for it?

core/src/components/select-modal/select-modal.ios.scss
brandyscarney, line 1: I have the same concern with the avatar size as I commented on action sheet and popover.

core/src/components/select-modal/select-modal.md.scss
brandyscarney, line 1: I have the same concern with the avatar size as I commented on action sheet and popover.

brandyscarney (changes requested):
This is looking really good, great work! I mostly have questions about some styles and requests for tests. There were a lot of files on this so sorry if I repeated myself anywhere.

core/src/components/action-sheet/action-sheet.md.scss
thetaPC: ~~I used the designs from `input` since it made more sense to me to compare it to another form control. Thoughts?~~

core/src/components/action-sheet/action-sheet.ionic.scss
thetaPC: I'm basing it off the tab button icon size and the radio box size as it seems that 24px is the consistent size for elements like this: https://github.com/ionic-team/ionic-framework/pull/31138/changes/ccdd1361d06d8e0d2d76184ccb93e296df244e24

core/src/components/select/test/rich-content-option/select.e2e.ts-snapshots/select-rich-content-action-sheet-md-ltr-Mobile-Chrome-linux.png
thetaPC: https://github.com/ionic-team/ionic-framework/pull/31138/changes/d475f863bf668105f9c75d594d81d00d1221453a

core/src/components/select/test/rich-content-option/select.e2e.ts-snapshots/select-rich-content-action-sheet-ios-ltr-Mobile-Chrome-linux.png
thetaPC: https://github.com/ionic-team/ionic-framework/pull/31138/changes/d475f863bf668105f9c75d594d81d00d1221453a

core/src/components/select/test/rich-content-option/select.e2e.ts-snapshots/select-rich-content-action-sheet-ionic-md-ltr-light-Mobile-Chrome-linux.png
thetaPC: https://github.com/ionic-team/ionic-framework/pull/31138/changes/d475f863bf668105f9c75d594d81d00d1221453a

core/src/components/action-sheet/action-sheet.common.scss
thetaPC: https://github.com/ionic-team/ionic-framework/pull/31138/changes/3c3206651b3153cf656c23dd0714bda0d47ab6ba and https://github.com/ionic-team/ionic-framework/pull/31138/changes/fdf9226282198ebedae243db98c8235ded3f9f83

core/src/components/alert/alert.common.scss
thetaPC: https://github.com/ionic-team/ionic-framework/pull/31138/changes/e669e54e899442ac43d06e475578dbe5fdfcd672

core/src/components/select-modal/select-modal.common.scss
thetaPC: https://github.com/ionic-team/ionic-framework/pull/31138/changes/3c3206651b3153cf656c23dd0714bda0d47ab6ba and https://github.com/ionic-team/ionic-framework/pull/31138/changes/fdf9226282198ebedae243db98c8235ded3f9f83

core/src/components/alert/alert.ionic.scss
thetaPC: https://github.com/ionic-team/ionic-framework/pull/31138/changes/863ca6cecf45f36413ff4bd260fec810ed4aba67

core/src/components/alert/alert.ios.vars.scss
thetaPC: Found the right value and removed it from being a variable: https://github.com/ionic-team/ionic-framework/pull/31138/changes/8914cc35562d2b24eb91a75dbc48b864ffff0d23

core/api.txt
thetaPC: Good point!

core/src/components/action-sheet/action-sheet.md.scss
thetaPC: I was able to find the right values! I'll add a comment on the actual line with more info: https://github.com/ionic-team/ionic-framework/pull/31138/changes/8914cc35562d2b24eb91a75dbc48b864ffff0d23

core/src/components/action-sheet/action-sheet.ios.vars.scss
thetaPC: I found the right value. I'll add a comment for more details. https://github.com/ionic-team/ionic-framework/pull/31138/changes/8914cc35562d2b24eb91a75dbc48b864ffff0d23

core/src/components/alert/test/states/alert.e2e.ts-snapshots/alert-checkbox-states-ios-ltr-Mobile-Chrome-linux.png
thetaPC: It's intentional. iOS doesn't have styles, you can confirm this by checking `next`.

core/src/components/alert/test/states/alert.e2e.ts-snapshots/alert-checkbox-states-md-ltr-Mobile-Chrome-linux.png
thetaPC: It's intentional. md doesn't have styles, you can confirm this by checking `next`.

core/src/components/select/test/rich-content-option/select.e2e.ts-snapshots/select-rich-content-alert-md-ltr-Mobile-Safari-linux.png
thetaPC: https://github.com/ionic-team/ionic-framework/pull/31138/changes/8914cc35562d2b24eb91a75dbc48b864ffff0d23

core/src/components/select/test/rich-content-option/select.e2e.ts-snapshots/select-rich-content-alert-ionic-md-ltr-light-Mobile-Chrome-linux.png
thetaPC: https://github.com/ionic-team/ionic-framework/pull/31138/changes/54d2c0c8cfa4291a6f06efef69a52b64becf00ff

core/src/components/select-option/test/label-placement/select-option.e2e.ts
thetaPC: https://github.com/ionic-team/ionic-framework/pull/31138/changes/6a5517dc47f49add0c69aefa573a3594c29975be

core/src/utils/select-option-render.tsx
thetaPC: https://github.com/ionic-team/ionic-framework/pull/31138/changes/645214a3334da30feea5dc7b5c672a05e529c2cb

core/src/utils/overlay-control-label.ts
thetaPC: Yes, I added a todo so we don't lose track of it: https://github.com/ionic-team/ionic-framework/pull/31138/changes/b89f9ac5ae369ee94696f3c8172dbc59fc0f1222

core/src/utils/sanitization/index.ts
thetaPC: https://github.com/ionic-team/ionic-framework/pull/31138/changes/ae8d923a6af7ae49f26ad57f622085b840ca0410

core/src/components/select/select.tsx
thetaPC: https://github.com/ionic-team/ionic-framework/pull/31138/commits/bfaf49e8c839b06b0617111f3b9c0f2c03144af3

core/src/components/select-popover/select-popover.md.scss
thetaPC: https://github.com/ionic-team/ionic-framework/pull/31138#discussion_r3337654397

core/src/components/select-popover/select-popover.ios.scss
thetaPC: https://github.com/ionic-team/ionic-framework/pull/31138#discussion_r3337654397

core/src/components/select-popover/select-popover.common.scss
thetaPC: https://github.com/ionic-team/ionic-framework/commit/3c3206651b3153cf656c23dd0714bda0d47ab6ba

core/src/components/select-popover/select-popover.ios.vars.scss
thetaPC: Found the right value and removed it from being a variable: https://github.com/ionic-team/ionic-framework/pull/31138/changes/8914cc35562d2b24eb91a75dbc48b864ffff0d23

core/src/components/select-modal/select-modal.ios.scss
thetaPC: https://github.com/ionic-team/ionic-framework/pull/31138#discussion_r3337654397

core/src/components/select-modal/select-modal.md.scss
thetaPC: https://github.com/ionic-team/ionic-framework/pull/31138#discussion_r3337654397

core/src/components/select-popover/select-popover.tsx
thetaPC: After testing, I removed it since dismissal wasn't happening for disabled options anyway. I added tests to guard against regressions, since the dismissal is being prevented by `pointer-events: none` - that way if we accidentally remove the style, the tests will catch it.

https://github.com/ionic-team/ionic-framework/pull/31138/changes/26c24eaf6172e604e7e8fd90a38f40ffd49e518e

core/src/components/select-modal/select-modal.tsx
thetaPC: After testing, I removed it since dismissal wasn't happening for disabled options anyway. I added tests to guard against regressions, since the dismissal is being prevented by `pointer-events: none` - that way if we accidentally remove the style, the tests will catch it.

https://github.com/ionic-team/ionic-framework/pull/31138/changes/26c24eaf6172e604e7e8fd90a38f40ffd49e518e

core/src/components/action-sheet/action-sheet.ios.scss
thetaPC: We should keep it centered to stay consistent to how native action sheet behaves. I did make some style changes to make it look better. Let me know if you still prefer left align.

https://github.com/ionic-team/ionic-framework/blob/e2d8d1566b4beb8ee174c62328bfa8fd6c9e0a6a/core/src/components/select-option/select-option.ios.interface.scss#L39

core/src/components/alert/alert.tsx
thetaPC: I was following the structure that was implemented for [action sheet focusable and activatable](https://github.com/ionic-team/ionic-framework/blob/bcbf6e1208f20411a6121f58f3d738e3add058a1/core/src/components/action-sheet/action-sheet.tsx#L696-L697) about [2 years ago](https://github.com/ionic-team/ionic-framework/pull/28723). It makes sense that we would gate it behind disabled.

As for why other components might not follow it, I suspect that we just weren't consistent on the approach.

core/src/components/select/test/basic/index.html
thetaPC: It's being passed because the `cancelText` for all select interfaces is "Cancel". It doesn't suit it as you said. We have two options:
1. Continue using "Cancel" to be consistent with the other interfaces
2. Use a different term, making it inconsistent with other interfaces

Thoughts? If we go with option 2, I'll create a new PR.

core/src/components/select-option/select-option.common.overlay.scss
thetaPC, line 1: Created the select option style to make sure all interfaces are being targeted from one source of truth. By doing this, we reduce the chances of things getting lost with copy and paste.

core/src/components/select-option/select-option.ionic.overlay.scss
```diff
@@ -0,0 +1,60 @@
...
+  gap: globals.$ion-space-200;
+}
+
+/**
+ * Cap slotted children so they can't stretch the option
+ * row out of proportion, keeping rows visually uniform
+ * regardless of the content.
+ */
+
+// Avatar / Image / SVG / Thumbnail / Icon
```
thetaPC, line 45: I decided to only update the size for these components based on the Figma design appearing to only support images and icons. This would also follow how MD behaves.

core/src/components/select-option/select-option.ios.overlay.scss
```diff
@@ -0,0 +1,47 @@
+@use "./select-option.native.interface";
+@use "../../themes/mixins" as mixins;
+@use "../item/item.ios.vars" as item;
+
+// Select Option - Interface
+// --------------------------------------------------
+
+/**
+ * Cap slotted children so they can't stretch the option
```
thetaPC, line 9: The sizes for the slotted children within this file were based off the official [Sketch designs for iOS 18](https://developer.apple.com/design/resources/) for components that are supported for start and end slots. Since iOS does not have an official select option, I determined the values based on the list component it has.

core/src/components/select-option/select-option.ios.overlay.scss
```diff
@@ -0,0 +1,47 @@
...
+ * row out of proportion, keeping rows visually uniform
+ * regardless of the content.
+ */
+
+// Avatar / Image / SVG / Thumbnail
```
thetaPC, line 14: Since iOS does not have an official select option, I determined the values based on the list component it has.

Images are considered to be a supported component as shown in the [Sketch design](https://www.sketch.com/s/bb57439f-19da-4c7a-bfd2-a196cf51f766/v/OrLwzq/f/8E5D8CC1-194D-40A6-A62C-F2176089072C?layerPath=2694b233#Inspect). The same value can be seen being used in the [Menu list for images](https://www.sketch.com/s/bb57439f-19da-4c7a-bfd2-a196cf51f766/v/OrLwzq/f/CEFAFCA9-E2DE-49F0-9363-AD7E894F00FD?layerPath=51dc5840#Inspect).

core/src/components/select-option/select-option.ios.overlay.scss
```diff
@@ -0,0 +1,47 @@
...
+.select-option-start > svg,
+.select-option-end > ion-img,
+.select-option-end > ion-thumbnail,
+.select-option-end > img,
+.select-option-end > svg {
+  width: 44px;
+  height: 44px;
+}
+
+// Icon
```
thetaPC, line 29: Since iOS does not have an official select option, I determined the values based on the list component it has.

Icons are considered to be a supported component as shown in the [Sketch design for the Menu list](https://www.sketch.com/s/bb57439f-19da-4c7a-bfd2-a196cf51f766/v/OrLwzq/f/706230B3-4CA0-4803-BCDD-C5E02EE42FC2?layerPath=3ccd9ebc#Inspect).

core/src/components/select-option/select-option.md.overlay.scss
```diff
@@ -0,0 +1,44 @@
+@use "./select-option.native.interface";
+
+// Select Option - Interface
+// --------------------------------------------------
+
+/**
+ * Cap slotted children so they can't stretch the option
```
thetaPC, line 7: The sizes for the slotted children within this file were based off the official [specs for MD3](https://m3.material.io/components/lists/specs) for components that are supported for start and end slots. Since MD does not have an official select option, I determined the values based on the list component it has.

core/src/components/select-option/select-option.md.overlay.scss
```diff
@@ -0,0 +1,44 @@
...
+ * row out of proportion, keeping rows visually uniform
+ * regardless of the content.
+ */
+
+// Avatar
```
thetaPC, line 12: Since MD does not have an official select option, I determined the values based on the list component it has.

Avatars are considered to be a supported, the size can be found under "Tokens & specs > Token > Size and typography".

core/src/components/select-option/select-option.md.overlay.scss
```diff
@@ -0,0 +1,44 @@
...
+.select-option-start > ion-avatar,
+.select-option-end > ion-avatar {
+  width: 40px;
+  height: 40px;
+}
+
+// Icon
```
thetaPC, line 19: Since MD does not have an official select option, I determined the values based on the list component it has.

Icons are considered to be a supported, the size can be found under "Tokens & specs > Token > Size and typography".

core/src/components/select-option/select-option.md.overlay.scss
```diff
@@ -0,0 +1,44 @@
...
+.select-option-start > ion-icon,
+.select-option-end > ion-icon {
+  width: 24px;
+  height: 24px;
+}
+
+// Image / SVG / Thumbnail
```
thetaPC, line 26: Since MD does not have an official select option, I determined the values based on the list component it has.

Images are considered to be a supported, the size can be found under "Tokens & specs > Token > Size and typography".

core/src/components/select-option/select-option.md.overlay.scss
```diff
@@ -0,0 +1,44 @@
...
+.select-option-start > ion-thumbnail,
+.select-option-end > ion-img,
+.select-option-end > img,
+.select-option-end > svg,
+.select-option-end > ion-thumbnail {
+  width: 56px;
+  height: 56px;
+}
+
+// Video
```
thetaPC, line 39: Since MD does not have an official select option, I determined the values based on the list component it has.

Videos are considered to be a supported, the size can be found under "Tokens & specs > Token > Size and typography".

core/src/components/action-sheet/action-sheet.ios.scss
brandyscarney: Okay that's fine!

core/api.txt
brandyscarney: Can we update the title to mention support for the new properties too?

core/src/components/select/test/basic/index.html
brandyscarney: Wouldn't it make more sense to be "OK" like Alert?

core/src/components/select/test/rich-content-option/select.e2e.ts-snapshots/select-rich-content-alert-ionic-md-ltr-light-Mobile-Safari-linux.png
brandyscarney, line 1: Why is this one gray but Chrome and Firefox are not?

core/src/components/select-option/select-option.ionic.interface.scss
```diff
@@ -0,0 +1,60 @@
...
+
+/**
+ * Cap slotted children so they can't stretch the option
+ * row out of proportion, keeping rows visually uniform
+ * regardless of the content.
+ */
+
+// Avatar / Image / SVG / Thumbnail / Icon
+.select-option-start > ion-avatar,
+.select-option-end > ion-avatar,
```
brandyscarney: Nit: this is the only one you didn't put with the rest of the `.select-option-end` styles

core/api.txt
thetaPC: How does it sound now?

core/src/components/alert/alert.md.vars.scss
```diff
@@ -191,7 +191,7 @@ $alert-md-list-border-bottom: $alert-md-list-border-top;
 $alert-md-radio-top: 0;
 
 /// @prop - Left of the alert radio
-$alert-md-radio-left: 26px;
+$alert-md-radio-left: 16px;
```
brandyscarney: I don't think this new value is correct. If you look at the before & after:

Image

The radios were aligned with the text in the header. This matches how [Material Design 2 confirmation dialogs](https://m2.material.io/components/dialogs#usage) look:

Image

And [Material Design 3 dialogs](https://m3.material.io/components/dialogs/guidelines#97ac3858-3932-4084-ae8e-73e42b7cb752) continue this:

Image

core/src/components/select/test/rich-content-option/select.e2e.ts-snapshots/select-rich-content-modal-ios-ltr-Mobile-Chrome-linux.png
brandyscarney, line 1: Are these ones gray because they are getting the focused state?

core/src/components/select-option/select-option.common.overlay.scss
brandyscarney: Is interface the best name here? I know it's what we call the overlays that open but it seems weird when we're going to have [`interface.ts`](https://github.com/ionic-team/ionic-framework/blob/ionic-modular/core/src/components/chip/chip.interfaces.ts) files everywhere.

core/src/components/select-option/select-option.ios.interface.scss
```diff
@@ -0,0 +1,47 @@
...
+
+/**
+ * Cap slotted children so they can't stretch the option
+ * row out of proportion, keeping rows visually uniform
+ * regardless of the content.
+ */
+
+// Avatar / Image / SVG / Thumbnail
+.select-option-start > ion-avatar,
+.select-option-end > ion-avatar,
```
brandyscarney: Nit: this is the only one you didn't put with the rest of the `.select-option-end` styles

core/api.txt
brandyscarney: I think it would be better to target `select`:

```
feat(select): add labelPlacement prop, justify prop, and rich content styles for select options
```

core/src/components/alert/alert.tsx
brandyscarney: Further research tells me this is likely not necessary. You don't have to remove it, but we should probably create a follow-up for these inconsistencies.

`ion-focusable`/`ion-activatable` only take effect when an element is actually focused or pressed, so gating them on `disabled` only matters for components that use `aria-disabled` without removing focusability (e.g. ion-chip). When a component disables via the native `disabled`/`tabindex` the browser already removes it from the focus/press path, making this check redundant.

Worth noting: `ion-chip` gates `ion-focusable` but leaves `ion-activatable` ungated. Neither should need to be gated here since it doesn't set `tabindex` to begin with.

brandyscarney (changes requested):

core/src/components/select-option/select-option.ionic.interface.scss
thetaPC: https://github.com/ionic-team/ionic-framework/pull/31138/changes/8cfef6ba26e57cb4bf62caee2bd6902850f2cfd4

core/src/components/select/test/rich-content-option/select.e2e.ts-snapshots/select-rich-content-alert-ionic-md-ltr-light-Mobile-Safari-linux.png
thetaPC: https://github.com/ionic-team/ionic-framework/pull/31138/changes/ea996110719a8beeb44da973f11fd7a35bcb22b3

core/src/components/select/test/rich-content-option/select.e2e.ts-snapshots/select-rich-content-modal-ios-ltr-Mobile-Chrome-linux.png
thetaPC: Yes, it gets the focused class on open so we see the focus styles immediately. Your PR should fix this.

core/src/components/select-option/select-option.common.overlay.scss
thetaPC: https://github.com/ionic-team/ionic-framework/pull/31138/changes/38dd8749f303b6916f7992b9cad321802396d911

core/src/components/select-option/select-option.ios.interface.scss
thetaPC: https://github.com/ionic-team/ionic-framework/pull/31138/changes/bd39bda517f5bfcfa6a7e4833e9fdb2f982d6813

core/src/components/alert/alert.tsx
thetaPC: Ticket FW-7561 created!

core/src/components/alert/alert.md.vars.scss
thetaPC: https://github.com/ionic-team/ionic-framework/pull/31138/changes/43817c0bc036e21e6178066cc7f25efaea13628c

core/src/components/alert/alert.md.vars.scss
thetaPC, line 1: Values are based on [MD2 specs for dialogs](https://m2.material.io/components/dialogs#specs).

core/src/components/select-option/select-option.ionic.overlay.scss
brandyscarney, line 1: Doesn't this need to be renamed to `overlay` too?

core/src/components/select-option/select-option.ios.overlay.scss
brandyscarney, line 1: Doesn't this need to be renamed to `overlay` too?

core/src/components/select-option/select-option.md.overlay.scss
brandyscarney, line 1: Doesn't this need to be renamed to `overlay` too?

core/src/components/select-option/select-option.native.overlay.scss
brandyscarney, line 1: Doesn't this need to be renamed to `overlay` too?

brandyscarney (approved):
Great work!! Some subtle changes in here but they make a huge difference. I love it! 🎉

core/src/components/select-option/select-option.ionic.overlay.scss
thetaPC: https://github.com/ionic-team/ionic-framework/pull/31138/changes/4471b8759dca220333c073db5d7789c5fc22a558

core/src/components/select-option/select-option.ios.overlay.scss
thetaPC: https://github.com/ionic-team/ionic-framework/pull/31138/changes/4471b8759dca220333c073db5d7789c5fc22a558

core/src/components/select-option/select-option.md.overlay.scss
thetaPC: https://github.com/ionic-team/ionic-framework/pull/31138/changes/4471b8759dca220333c073db5d7789c5fc22a558

core/src/components/select-option/select-option.native.overlay.scss
thetaPC: https://github.com/ionic-team/ionic-framework/pull/31138/changes/4471b8759dca220333c073db5d7789c5fc22a558

brandyscarney (approved):

core/src/utils/sanitization/index.ts
```diff
@@ -114,9 +158,17 @@ const sanitizeElement = (element: any) => {
...
     // remove non-allowed attribs
-    if (!allowedAttributes.includes(attributeName.toLowerCase())) {
+    if (!allowSafeAttributes && !allowedAttributes.includes(lowerName)) {
+      element.removeAttribute(attributeName);
+      continue;
+    }
+
+    // strip event-handler attributes (already removed by the allowlist
+    // when !allowSafeAttributes; this guards the permissive path)
+    if (lowerName.startsWith('on')) {
```
ShaneK: The old rich-content path ran through `sanitizeDOMString`, whose allowlist kept only `class`/`id`/`href`/`src`/`name`/`slot`. `sanitizeDOMTree`'s permissive mode flips that to an everything-except-denylist: it keeps any attribute that isn't `on*`-prefixed or `javascript:`-valued. So `style` (CSS injection, `background:url()` beaconing, UI spoofing), `data:` URIs in `href`/`src`, `formaction`/`action`/`target`, and `xlink:href` on SVG `<a>` now survive where they were always stripped before. The `javascript:` guard also leans on `element[attributeName]` property reflection to catch entity-obfuscated payloads, and that's `undefined` for namespaced attrs, so something like `java&#9;script:` in `xlink:href` would only hit the literal-substring value check and slip through.

I know this path is opt-in (`innerHTMLTemplatesEnabled` defaults to `false`) and the content is the developer's own `ion-select-option` markup, not raw user input, so the real-world exposure is narrow. And I get that the goal is to keep `size`/`color`/`shape`. Mostly I want a conscious sign-off here: would keeping an allowlist even in permissive mode (or at minimum stripping `style` and normalizing the URL-bearing attrs) be safer than the denylist? Not blocking, but worth a deliberate call since it widens what reaches the `innerHTML`/VNode sinks.

core/src/utils/select-option-render.tsx
```diff
@@ -12,48 +11,83 @@ interface RichContentOption extends RichContentOpt {
...
-          }
-
-          const sanitized = sanitizeDOMString(content.innerHTML);
-          el.innerHTML = sanitized ?? '';
-          contentCache.set(el, content);
-        }
-      }}
-    ></Tag>
+    <Tag class={className} key={keyPrefix}>
+      {Array.from(content.childNodes).map((child, i) => cloneToVNode(child, keyPrefix, i))}
```
ShaneK, line 93: `cloneToVNode` filters nothing by design and the doc leans on callers to pre-sanitize. The `ion-select` path is fine since `getOptionContent` runs `sanitizeDOMTree` first. The gap is `renderOptionLabel`: it accepts `label: string | HTMLElement` and reads `startContent`/`endContent`, none of which are on the public `AlertInput` type. A vanilla-JS consumer doing `alertController.create({ inputs: [{ label: someEl, startContent: ... }] })` reaches `cloneToVNode` with DOM that never passed through `sanitizeDOMTree`, and nothing enforces the contract.

Could we call `sanitizeDOMTree` defensively inside `renderClonedContent`? It'd be idempotent for the already-sanitized select path, so it just closes the JS-consumer hole without changing the select behavior. Up to you though.

core/src/utils/overlay-control-label.ts
```diff
@@ -0,0 +1,58 @@
...
+
+import type { Theme } from '../interface';
+
+/**
+ * Returns the default `labelPlacement` for a radio or checkbox option
+ * rendered inside an overlay. Defaults follow each theme's established
+ * option-row layout:
+ * - `ionic`: always `"start"`.
+ * - `ios`: `"start"` for radio in `alert` and `popover`. The `modal`
+ *   interface flips iOS radio back to `"end"`s. Checkbox is always
```
ShaneK:
```suggestion
 *   interface flips iOS radio back to `"end"`. Checkbox is always
```

Stray trailing `s` after `` `"end"` ``

core/src/utils/overlay-control-label.ts
```diff
@@ -0,0 +1,58 @@
...
+ *   `"end"` on iOS.
+ * - everything else (e.g. `md`): `"end"`.
+ *
+ * `interfaceType` is optional; only `"modal"` changes the result, so
+ * callers that aren't a modal can omit it.
+ *`
```
ShaneK:
```suggestion
 *
```

Lone random backtick

ShaneK (changes requested):
Looking awesome! I just had some security-related concerns and a couple of nits. Let me know what you think

core/src/utils/sanitization/index.ts
thetaPC: https://github.com/ionic-team/ionic-framework/pull/31138/changes/33ac26634b06a6495f2232402bb8f3101f869471

core/src/utils/select-option-render.tsx
thetaPC: https://github.com/ionic-team/ionic-framework/pull/31138/changes/b82c7dbc6d0872b89ad6dc01bd43c151bff43d70

core/src/utils/overlay-control-label.ts
thetaPC: https://github.com/ionic-team/ionic-framework/pull/31138/changes/355e972f856b615679e3f507df2baa3e9bd01b41

core/src/utils/overlay-control-label.ts
thetaPC: https://github.com/ionic-team/ionic-framework/pull/31138/changes/9105453411fe50a9d84de631fb0244fbb7bece4e

ShaneK (approved):

PR #31142, opened 2026-05-13
fix(vue-router): honor caller-provided direction on replace navigation

### LINKED ISSUE #24995: bug: Incorrect transition animation on 'forward' using Vue useIonRouter navigate/replace
### Prerequisites

### Ionic Framework Version

- [ ] v4.x
- [ ] v5.x
- [X] v6.x

### Current Behavior

In order to completely disable native back swipe in IOS I’m having to use `replace` instead of `push` and its working fine except for one issue … disorienting page transitions.

I expect `IonRouter.navigate(url, "forward", "replace")` to show a `forward` animation but instead it shows a `back` animation. Also the other issue I've noticed is that is that the very first transition almost always is abrupt ie. without any animation at all.

Here’s it in action: [Screen Recording 2022 03 21 at 6 18 11 pm - YouTube 4](https://youtu.be/Y1GVPZGllMg)

### Expected Behavior

`forward` animation when using `forward`

### Steps to Reproduce

Navigate between pages using `IonRouter.navigate(url, "forward", "replace")` in IonicVue

### Code Reproduction URL

_No response_

### Ionic Info

```
Ionic:

   Ionic CLI       : 6.19.0
   Ionic Framework : @ionic/vue 6.0.12

Capacitor:

   Capacitor CLI      : 3.4.3
   @capacitor/android : 3.4.3
   @capacitor/core    : 3.4.3
   @capacitor/ios     : 3.4.3

Utility:

   native-run  : 1.5.0

System:

   NodeJS : v16.13.1 
   npm    : 8.4.0
   OS     : macOS Montere
```

### Additional Information

_No response_

ShaneK:
Issue number: resolves #24995

---------

In `createIonRouter`, the replace-action branch resolves `routeInfo.routerDirection` and `routeInfo.routerAnimation` by preferring the leaving route's values over the ones the caller passed in. The leaving route wins whenever it has any value set, so `useIonRouter.navigate(url, 'forward', 'replace')` called after a back navigation ends up with `routerDirection: 'back'`, and the same call from the initial route ends up with `'none'`. The caller's explicit direction is silently discarded.

## What is the new behavior?

The precedence is flipped: the caller's `routerDirection` and `routerAnimation` are kept, and the leaving route's values are only used as a fallback when nothing was provided. `forward + replace` now animates forward regardless of how the user got to the leaving page, and the default `useIonRouter.replace(url)` keeps the `'root'` direction it was constructed with instead of inheriting whatever the leaving route had recorded.

- [ ] Yes
- [X] No

## Other information

### DIFF
--- a/packages/vue-router/src/router.ts
+++ b/packages/vue-router/src/router.ts
@@ -459,10 +459,15 @@ export const createIonRouter = (
           routeInfo.lastPathname =
             currentRouteInfo?.pathname || routeInfo.lastPathname;
           routeInfo.pushedByRoute = pushedByRoute;
+          /**
+           * Prefer the direction/animation the caller specified on
+           * the navigate call; fall back to the leaving route's
+           * values only when none was provided.
+           */
           routeInfo.routerDirection =
-            currentRouteInfo?.routerDirection || routeInfo.routerDirection;
+            routeInfo.routerDirection || currentRouteInfo?.routerDirection;
           routeInfo.routerAnimation =
-            currentRouteInfo?.routerAnimation || routeInfo.routerAnimation;
+            routeInfo.routerAnimation || currentRouteInfo?.routerAnimation;
           routeInfo.prevRouteLastPathname = currentRouteInfo?.lastPathname;
         }
       }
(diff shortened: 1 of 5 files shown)


OS-jacobbell (approved):
Makes sense, good find!

PR #31144, opened 2026-05-13
fix(vue-router): tab button works on a deep-loaded tab child

### LINKED ISSUE #29705: bug: [Bug in Ionic Vue Tabs Project]Clicking the buttons within the page fails to navigate to the page
### Prerequisites

### Ionic Framework Version

v8.x

### Current Behavior

Hello, Ionic official. When I was using the Ionic framework to develop a Vue project, I encountered a rather troublesome bug. I created a tabs sample project. After clicking a certain Button on the page and using ionRouter to navigate to the second-level page, when refreshing this second-level page and clicking the Button again, I found that the page was stuck. However, the address in the address bar was correct and could not be clicked. This is one manifestation. The second manifestation is that on the second-level page, clicking a button to use windows.location.href = 'http://www.xx.com' to jump to an external link. After returning to the second-level page and then back to the first-level page, clicking the Button again will reproduce the first manifestation. But in the Tabs sample of Angular, this problem does not exist.

### Expected Behavior

I hope the route navigation can jump normally.

### Steps to Reproduce

1.Create a tabs sample project in the Ionic framework for Vue.
2.Click a Button on the first-level page to navigate to the second-level page using ionRouter.
3.Refresh the second-level page and try clicking the Button again.
4.On the second-level page, click a button to use windows.location.href = 'http://www.xx.com' to jump to an external link.
5.Return to the second-level page and then back to the first-level page.
6.Click the Button again.

### Code Reproduction URL

https://gitee.com/FSAnoymous/ionic-vue-bug-demo

### Ionic Info

https://github.com/ionic-team/ionic-framework/assets/17871590/08c4a306-871a-4945-901c-8ec8e69423cf

https://github.com/ionic-team/ionic-framework/assets/17871590/2a8b1217-34fd-4a6b-abc2-db17c36c3d50

### Additional Information

Sorry, my English is not very good. Please check the bug by looking at the performance in the video. This bug only exists in the Vue project, and there are no issues with the Angular demo. This has seriously affected the project experience. I hope it can be solved as soon as possible.

ShaneK:
Issue number: resolves #29705

---------

In a Vue tabs app, clicking the active tab button after landing directly on a tab child (browser refresh on `/tabs/tab1/child`, or browser back from an external `window.location.href`) is a silent no-op. `IonTabBar.checkActiveTab` only records the active tab's `currentHref` when there's a `prevActiveTab`, so the initial deep load leaves `currentHref` matching `originalHref` and the `originalHref !== currentHref` guard in `IonTabButton.onClick` skips `resetTab`. Even past that guard, `resetTab` issues `router.go(routeInfo.position - currentHistoryPosition)`, which resolves to `router.go(0)` when the tab's only history entry is the current page.

## What is the new behavior?

`IonTabBar.checkActiveTab` now records `currentHref` on initial setup so the tab state reflects the deep-loaded URL. `resetTab` takes the tab's `originalHref` and, when the first history entry for the tab is the current entry, falls back to `handleNavigate(originalHref, "pop", "back", undefined, tab)`. The `pop` action routes through `router.replace`, so the deep-loaded child entry is replaced rather than stacked on top and no stale child entry stays in browser history.

- [ ] Yes
- [X] No

## Other information

Manual test: [/vue/tabs/tab1/childone](https://ionic-framework-git-fw-6452-ionic1.vercel.app/vue/tabs/tab1/childone). Load that URL directly, click the Tab 1 button, you should land on `/vue/tabs/tab1`

### DIFF
--- a/packages/vue-router/src/router.ts
+++ b/packages/vue-router/src/router.ts
@@ -555,7 +555,7 @@ export const createIonRouter = (
     router.push(routerLink);
   };
 
-  const resetTab = (tab: string) => {
+  const resetTab = (tab: string, originalHref?: string) => {
     /**
      * Resetting the tab should go back
      * to the initial view in the tab stack.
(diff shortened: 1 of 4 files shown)


thetaPC (approved):
LGTM

PR #31145, opened 2026-05-13
fix(vue-router): prevent out-of-bounds index when popping across tabs

### LINKED ISSUE #29413: bug: tab sub-page navigation breaks occasionally with ionic-vue-router
### Prerequisites

### Ionic Framework Version

v8.x

### Current Behavior

Navigating into tab sub-page and clicking on the respective tab bar in order to get to the initial tab page breaks navigation occasionally.
Either nothing happens or the following error `viewItem2 is undefined` is shown, sometimes subsequently followed by an this error `enteringEl is undefined`.

When this happens the navigation for the view in question stays broken. Other routes may keep working but eventually they stop working as well. The broken navigation point may starts working again.

### Expected Behavior

Inside a tab navigating to a sub page and then to another tab and then back to the initial sub page and then clicking on the tab of the sub page again should lead me to the initial page, regardless of the order of navigation steps I took before.

### Steps to Reproduce

Unfortunately there I wasn't able to identify precise steps in order to reproduce this. But I think these are steps that _should_ work
1. On **Tab 1** click the _Next_ button in order to enter Tab 1 sub page
2. Click on **Tab 5**
3. Click on **Tab 3**
4. Click the _Next_ button on **Tab 3** in order to enter Tab 3 sub page
5. Click on **Tab 5** again
6. Click **Tab 1** (you should be on Tab 1 Sub Page)
7. Click **Tab 1** again in order to get to initial page of Tab 1
8. There it is `TypeError: viewItem2 is undefined @ionic_vue_router.js:562:7`
9. _(If it's not broken by now go to **Tab 3** (you should be on the sub page there), click **Tab 3** again in order to get to the inital page of Tab 3 - Tab 3 doesn't do anything anymore)_

### Code Reproduction URL

https://stackblitz.com/~/github.com/socialmedialabs/ionic-sample-tabs-app

### Ionic Info

Ionic:

   Ionic CLI : 5.4.16 (/home/ubuntu/.nvm/versions/node/v21.6.1/lib/node_modules/ionic)

Capacitor:

   Capacitor CLI   : 6.0.0
   @capacitor/core : 6.0.0

Utility:

   cordova-res : not installed
   native-run  : 2.0.1

System:

   NodeJS : v21.6.1 (/home/ubuntu/.nvm/versions/node/v21.6.1/bin/node)
   npm    : 10.2.4
   OS     : Linux 6.5

### Additional Information

_No response_

ShaneK:
Issue number: resolves #29413

---------

In `viewStacks.ts`, `unmountLeavingViews` and `mountIntermediaryViews` walk the outlet's view stack using `startIndex - delta` (or `startIndex + delta`) as the loop end, with no bound on `viewStack.length`. `delta` comes from the popstate event's history delta. Apps that mount `<ion-tabs>` at the root with no outer `<ion-router-outlet>` only have one outlet registered, so `usingLinearNavigation` is true and these helpers actually run. Each tab switch adds a browser history entry but reuses existing view items, so `|delta|` can easily exceed the stack depth above the entering view. The loop then reads `viewStack[i]` as `undefined` and throws `TypeError: viewItem is undefined` from `viewItem.mount = false`. The navigation aborts mid-transition, which is what surfaces the secondary `enteringEl is undefined` warning and leaves that route stuck

## What is the new behavior?

Both helpers bail when the entering view item isn't in the stack (`startIndex === -1`) and clamp the loop end to `Math.min(viewStack.length, ...)`, so a delta that overruns the stack stops at the last real view item instead of indexing past the end. A new Vitest spec at `packages/vue/test/base/tests/unit/tabs-single-outlet.spec.ts` mounts `<ion-tabs>` as the app root with flat routes, builds up history across tabs and sub-pages, then calls `router.go(-6)`. Without the fix the spec catches the unhandled `TypeError` from `viewStacks.ts`

- [ ] Yes
- [X] No

## Other information

Current dev build:
```
8.8.7-dev.11778707797.1723d277
```

### DIFF
--- a/packages/vue-router/src/viewStacks.ts
+++ b/packages/vue-router/src/viewStacks.ts
@@ -196,8 +196,15 @@ export const createViewStacks = (router: Router) => {
     if (!viewStack) return;
 
     const startIndex = viewStack.findIndex((v) => v === viewItem);
+    if (startIndex === -1) return;
 
-    for (let i = startIndex + 1; i < startIndex - delta; i++) {
+    // delta from popstate reflects browser history depth, which can exceed
+    // the outlet's view stack when tab switches build up history without
+    // adding new view items. Clamp to the stack length so we never index
+    // past the end of the array.
+    const endIndex = Math.min(viewStack.length, startIndex - delta);
+
+    for (let i = startIndex + 1; i < endIndex; i++) {
       const viewItem = viewStack[i];
       viewItem.mount = false;
       viewItem.ionPageElement = undefined;
@@ -233,8 +240,11 @@ export const createViewStacks = (router: Router) => {
     if (!viewStack) return;
 
     const startIndex = viewStack.findIndex((v) => v === viewItem);
+    if (startIndex === -1) return;
 
-    for (let i = startIndex + 1; i < startIndex + delta; i++) {
+    const endIndex = Math.min(viewStack.length, startIndex + delta);
+
+    for (let i = startIndex + 1; i < endIndex; i++) {
       viewStack[i].mount = true;
     }
   };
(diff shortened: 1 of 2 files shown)


thetaPC (approved):
LGTM

PR #31150, opened 2026-05-15
fix(vue-router): invoke beforeRouteEnter next() callbacks with component instance

### LINKED ISSUE #25013: bug: vue, router outlet does not support in-component guards next callbacks
### Prerequisites

### Ionic Framework Version

- [ ] v4.x
- [ ] v5.x
- [x] v6.x

### Current Behavior

Vue-Router's beforeRouteEnter's "next" parameter, which is used inside beforeRouteEnter like so:
```typescript
next(component => component.member = value)
```
Gives the following TS error:
`Property 'member' does not exist on type 'ComponentPublicInstance<{}, {}, {}, {}, {}, {}, {}, {}, false, ComponentOptionsBase<any, any, any, any, any, any, any, any, any, {}>>'`

After bypassing the type check, the page changes as expected but `component.member` is however not updated.

### Expected Behavior

Calling `next` like previously stated should continue navigation and set `component.member` to `value`.

### Steps to Reproduce

Add this component to any route and navigate to it:
```typescript
export default defineComponent({
    name: 'test'
    data() {
        return { member: 0 }
    },
    beforeRouteEnter(to, from, next) {
        next(e => e.member = 5);
        // or next(e => (a as any).member = 5); to bypass type check
    }
});
```

Assert with Vue devtool that `member != 5`.

### Code Reproduction URL

https://github.com/3NdR1m/bug_report_ionic_vuejs

### Ionic Info

[WARN] Error loading @capacitor/ios package.json: Error: Cannot find module '@capacitor/ios/package'
       
       Require stack:
       - /usr/lib64/node_modules/@ionic/cli/lib/project/index.js
       - /usr/lib64/node_modules/@ionic/cli/lib/index.js
       - /usr/lib64/node_modules/@ionic/cli/index.js
       - /usr/lib64/node_modules/@ionic/cli/bin/ionic

Ionic:

   Ionic CLI       : 6.19.0 (/usr/lib64/node_modules/@ionic/cli)
   Ionic Framework : @ionic/vue 6.0.11

Capacitor:

   Capacitor CLI      : 3.4.3
   @capacitor/android : 3.4.3
   @capacitor/core    : 3.4.3

Utility:

   native-run  : 1.5.0

System:

   NodeJS : v16.14.1 (/usr/bin/node)
   npm    : 8.5.0
   OS     : Linux 5.15

### Additional Information

[In-Component Navigation Guards, official doc](https://router.vuejs.org/guide/advanced/navigation-guards.html#in-component-guards)

ShaneK:
Issue number: resolves #25013

---------

A `beforeRouteEnter(to, from, next)` guard that calls `next((vm) => vm.member = 5)` never runs the callback. Vue Router's `<router-view>` invokes whatever ends up in `matchedRoute.enterCallbacks[name]` from a post-flush watcher once the view's instance is set, but `IonRouterOutlet` does its own rendering and skips that step. `matchedRoute.instances.default` gets populated in `registerIonPage`, so internal guards still fire, but the queued enter callbacks sit there unused

## What is the new behavior?

`createViewStacks`'s `registerIonPage` now reads `matchedRoute.enterCallbacks.default`, resets it, and runs each callback with the resolved instance. The reset happens before invocation so a callback that triggers a synchronous navigation can push fresh entries onto the next matched record without getting wiped. There's a new Vitest case in `routing.spec.ts` that navigates into a page whose `beforeRouteEnter` does `next((vm) => { vm.member = 5 })` and asserts both the callback ran with the instance and the mutation reached the DOM

- [ ] Yes
- [x] No

### DIFF
--- a/packages/vue-router/src/viewStacks.ts
+++ b/packages/vue-router/src/viewStacks.ts
@@ -34,9 +34,29 @@ export const createViewStacks = (router: Router) => {
      * and will not run route guards that
      * are written in the component.
      */
+    const instance = viewItem.vueComponentRef.value;
     viewItem.matchedRoute.instances = {
-      default: viewItem.vueComponentRef.value,
+      default: instance,
     };
+
+    /**
+     * Run any callbacks passed to `next()` inside a
+     * `beforeRouteEnter` guard now that the component
+     * instance is available. Vue Router's official
+     * `<router-view>` does this in a post-flush watcher;
+     * because IonRouterOutlet manages rendering itself,
+     * the wrapper must invoke them explicitly.
+     */
+    const enterCallbacks = viewItem.matchedRoute.enterCallbacks?.default;
+    if (instance && enterCallbacks && enterCallbacks.length > 0) {
+      /**
+       * Clear before invoking so any synchronous navigation
+       * triggered by a callback can push fresh callbacks onto
+       * the array without being wiped by a post-iteration reset.
+       */
+      viewItem.matchedRoute.enterCallbacks.default = [];
+      enterCallbacks.forEach((cb) => cb(instance));
+    }
   };
 
   const findViewItemByRouteInfo = (routeInfo: RouteInfo, outletId?: number) => {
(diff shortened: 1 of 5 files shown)


thetaPC (approved):
LGTM

Issue #25534, opened 2022-06-27
bug: React route navigate back with query string(URL search params) shows flash of default(non-query string) content


babycourageous:
### Prerequisites

- [X] I have read the [Contributing Guidelines](https://github.com/ionic-team/ionic-framework/blob/main/.github/CONTRIBUTING.md#creating-an-issue).
- [X] I agree to follow the [Code of Conduct](https://ionicframework.com/code-of-conduct).
- [X] I have searched for [existing issues](https://github.com/ionic-team/ionic-framework/issues) that already report this problem, without success.

### Ionic Framework Version

- [ ] v4.x
- [ ] v5.x
- [X] v6.x
- [ ] Nightly

### Current Behavior

Not sure if this is specifically an Ionic bug or related to react router, URL search params, or something entirely different - 
When working with Ionic React and a url search param in the route, navigating back from the route with a search param shows a flash of content for the route if the search param was not provided. For example navigating back from `todos?todo=1` briefly shows the contents of `todos` before popping the page from the stack and navigating back.

https://user-images.githubusercontent.com/14936212/175848414-886ec938-c423-426f-a277-2755beac72fd.mov

### Expected Behavior

the page should just pop off the stack w/o showing the flash of non-"query string" content.

### Steps to Reproduce

In the repo are three buttons. One goes to a base route and the other two are the base route with a url search param after. Navigate to the base route to see what is rendered when no search param is provided.
Then navigate back. 
Now, navigate to either of the other two routes. You will see a different ui rendered that uses the search param. When navigating back from either of these routes, you will see a flash of the default route content (w/o search params).

### Code Reproduction URL

https://github.com/babycourageous/ionic-routing-query-string-bug

### Ionic Info

Ionic:

   Ionic CLI       : 6.18.1 (/Users/renedellefont/.fnm/node-versions/v16.13.1/installation/lib/node_modules/@ionic/cli)
   Ionic Framework : @ionic/react 6.1.10

Capacitor:

   Capacitor CLI      : 3.5.1
   @capacitor/android : not installed
   @capacitor/core    : 3.5.1
   @capacitor/ios     : not installed

Utility:

   cordova-res : not installed globally
   native-run  : 1.6.0

System:

   NodeJS : v16.13.1 (/Users/renedellefont/.fnm/node-versions/v16.13.1/installation/bin/node)
   npm    : 8.1.2
   OS     : macOS Big Sur

  ─────────────────────────────────────────────────

[...]

PR #31153, opened 2026-05-19
fix(react-router): symmetric pathname+search url comparison

### LINKED ISSUE #31152: bug: IonRouter.handleHistoryChange inconsistently compares search-bearing leavingUrl to search-free location.pathname, causing spurious transitions on every popstate for routes with query params
### Prerequisites

### Ionic Framework Version

v8.x

### Current Behavior

In `packages/react-router/src/ReactRouter/IonRouter.tsx`, `handleHistoryChange` contains:

```ts
const leavingUrl = leavingLocationInfo.pathname + leavingLocationInfo.search;
if (leavingUrl !== location.pathname) {
  // ...transition / locationHistory mutation block...
}
```

The two sides of the comparison are not comparable: the left side is `pathname + search`, the right side is `pathname` only. For any route where `leavingLocationInfo.search` is non-empty (e.g. `?foo=1`), the comparison is **always** unequal, so the transition block runs on every history event — including no-op popstates that didn't actually change the URL.

Concretely, this causes a false `POP` transition to whatever `pushedByRoute` points at whenever:

1. The current route's tracked `routeInfo` has any `search` string, AND
2. A `popstate` fires that doesn't actually change the URL (e.g. browser-back over a history entry pushed at the same URL via `window.history.pushState`).

Because the block treats this as an `action === 'POP'`, it looks up `currentRoute.pushedByRoute`, sets `incomingRouteParams` to the previous routeInfo, and `setState({routeInfo})` → IonRouterOutlet swaps the view. The browser URL is unchanged, but the rendered route is now the *previous* entry in `locationHistory`.

### Expected Behavior

The transition block should only run when the URL actually changed. Comparing like with like:

```ts
const leavingUrl = leavingLocationInfo.pathname + leavingLocationInfo.search;
const currentUrl = location.pathname + (location.search || '');
if (leavingUrl !== currentUrl) {
  // ...
}
```

With that change:
- Popstates that don't actually change the URL: comparison equal → block skipped (correct).
- Pathname changes: comparison unequal → block runs (correct, same as today).
- Search-only changes via `routerPush(samePath + newSearch)`: comparison unequal → block runs → push correctly tracked (correct, same as today).

### Steps to Reproduce

Minimal manual repro in any `@ionic/react` app, no setup beyond a route that can receive a search param:

1. Add two routes that push each other so they have a `pushedByRoute` chain: e.g. `/a` and `/b?foo=1`. Navigate `/a` → `/b?foo=1` via an `IonRouterLink` so Ionic's `locationHistory` tracks both.

2. While on `/b?foo=1`, open the devtools console and run:
   ```js
   window.history.pushState({ marker: true }, '', window.location.href);
   ```
   This pushes a stub at the same URL — browser history grows by one entry; Ionic's `locationHistory` is unchanged (direct `window.history.pushState` doesn't fire React Router's listener, so `IonRouter.handleHistoryChange` doesn't see it).

3. In the console, run:
   ```js
   window.history.back();
   ```

**Observed:** Ionic transitions back to `/a`. The browser URL is still `/b?foo=1`, but the rendered view is `/a`. Now the URL and the rendered view are out of sync — `useLocation()` returns `/b?foo=1`, but the IRO is showing `/a`'s component.

**Expected:** No transition. Browser URL didn't change, so nothing should happen visually. (Steps 1–2 establish the precondition; step 3's popstate over a same-URL entry shouldn't trigger any UI change.)

If you swap step 1 for two routes without any search params (`/a` → `/b`, no query), step 3 produces no transition — the bug only manifests when the current route has search.

### Code Reproduction URL

I haven't built a hosted StackBlitz, but the bug is isolated enough that the file/line is the actual reproduction surface:  - **File:** `packages/react-router/src/ReactRouter/IonRouter.tsx` - **Method:** `handleHistoryChange` - **The comparison:** the line that reads `if (leavingUrl !== location.pathname)`  Current `main`: <https://github.com/ionic-team/ionic-framework/blob/main/packages/react-router/src/ReactRouter/IonRouter.tsx> — the buggy comparison is still present.

### Ionic Info

Ionic:

   Ionic Framework      : @ionic/react 8.5.5
   @ionic/react-router  : 8.5.5

Capacitor:

   Capacitor CLI      : 8.1.0
   @capacitor/android : 8.1.0
   @capacitor/core    : 8.1.0
   @capacitor/ios     : 8.1.0

Dependencies:

   react              : 18.3.1
   react-dom          : 18.3.1
   react-router-dom   : 5.3.4

System:

   NodeJS : v22.12.0
   npm    : 10.9.0
   OS     : macOS 26.4.1

### Additional Information

Related but distinct: #25534 — that issue describes the user-visible symptom on its own repro but the root cause was never named, and the maintainer's workaround was about how consumers read URL params, not about the comparison itself. This issue identifies the exact line and proposes a one-line fix.

### Why this matters in practice

We hit this building a back-button-handler for in-app modals: open a non-URL modal → `window.history.pushState({overlay: true}, '', href)` → on dismiss or browser-back, pop the stub. The pattern works perfectly on every search-free route. On routes with `?` search params (e.g. our chat route uses `?jumpTo=<messageId>` to deep-link to a message), the stub pop fires the false `POP` transition and the user is silently teleported to a different page in the IRO stack, while the URL stays put.

Any pattern that uses direct `window.history.pushState`/`back` for transient state on a search-bearing route has this problem. The same root cause is likely behind the user-visible behavior reported in #25534, though that issue was framed as a transition-rerender flash rather than as a comparison bug.

### Suggested Fix

One-line change in `handleHistoryChange`:

```diff
- const leavingUrl = leavingLocationInfo.pathname + leavingLocationInfo.search;
- if (leavingUrl !== location.pathname) {
+ const leavingUrl = leavingLocationInfo.pathname + leavingLocationInfo.search;
+ const currentUrl = location.pathname + (location.search || '');
+ if (leavingUrl !== currentUrl) {
```

Happy to open a PR if useful.

EricHech:
Issue number: resolves #31152

In `IonRouter.handleHistoryChange`, the URL-change guard compares mismatched operands:

```ts
const leavingUrl = leavingLocationInfo.pathname + leavingLocationInfo.search;
if (leavingUrl !== location.pathname) { ... }
```

The left side includes `search`, the right side does not. For any route with a non-empty query string, the comparison is **always** unequal, so the transition block runs on every history event — including no-op popstates over same-URL entries pushed via `window.history.pushState`.

Concretely: when a same-URL history entry on a search-bearing route is popped, `action === 'POP'` is processed and the IRO transitions to `currentRoute.pushedByRoute`. The browser URL doesn't change but the rendered view does — the user gets silently teleported to a different page in the stack.

Minimal repro: on any search-bearing route, run in the console:
```js
window.history.pushState({}, '', window.location.href);
window.history.back();
```
The current behavior swaps the rendered page to the previous entry in `locationHistory` while the URL stays put. Expected: no visible change.

Full repro and analysis in #31152. This is also the root cause behind the symptom reported in #25534 (framed there as a transition-rerender flash).

## What is the new behavior?

The right side of the comparison now also includes `search`, so the block only runs when the URL actually changed:

```ts
const leavingUrl = leavingLocationInfo.pathname + leavingLocationInfo.search;
const currentUrl = location.pathname + (location.search || '');
if (leavingUrl !== currentUrl) { ... }
```

Behavior matrix:
- Search-only navigations (e.g. `routerPush(samePath + newSearch)`): **unchanged** — pathname matches but search differs, block still runs.
- Pathname changes: **unchanged** — pathnames differ, block still runs.
- No-op popstates on search-bearing routes: **fixed** — pathname + search both match, block correctly skipped.

A new Cypress regression test is included (`packages/react-router/test/base/tests/e2e/specs/routing.cy.js`) that pushes a same-URL state on a search-bearing route, calls `history.back()`, and asserts the page does not teleport.

- [ ] Yes
- [x] No

The comparison becomes stricter (skips the block in more cases than before), but only for the cases where the URL didn't actually change. All cases where the URL *did* change are still routed through the existing transition logic unchanged.

## Other information

Happy to iterate on the fix shape if there's a preferred alternative — e.g. gating behind a flag for backward compat, or restructuring the guard differently. The one-line change above is the smallest possible fix that keeps existing behavior for every "real URL change" case.

### DIFF
--- a/packages/react-router/src/ReactRouter/IonRouter.tsx
+++ b/packages/react-router/src/ReactRouter/IonRouter.tsx
@@ -107,7 +107,8 @@ class IonRouterInner extends React.PureComponent<IonRouteProps, IonRouteState> {
     }
 
     const leavingUrl = leavingLocationInfo.pathname + leavingLocationInfo.search;
-    if (leavingUrl !== location.pathname) {
+    const currentUrl = location.pathname + (location.search || '');
+    if (leavingUrl !== currentUrl) {
       if (!this.incomingRouteParams) {
         if (action === 'REPLACE') {
           this.incomingRouteParams = {
--- a/packages/react-router/src/ReactRouter/IonRouter.tsx
+++ b/packages/react-router/src/ReactRouter/IonRouter.tsx
@@ -84,10 +84,15 @@ class IonRouterInner extends React.PureComponent<IonRouteProps, IonRouteState> {
         this.incomingRouteParams.routeOptions = routeOptions;
         this.props.history.push(routeInfo.pathname + (routeInfo.search || ''));
       } else {
+        /**
+         * The recorded search has to match the URL we push, otherwise
+         * handleHistoryChange sees a URL change where there is none.
+         */
+        const normalizedSearch = search ? '?' + search : '';
         this.incomingRouteParams.pathname = pathname;
-        this.incomingRouteParams.search = search ? '?' + search : undefined;
+        this.incomingRouteParams.search = normalizedSearch;
         this.incomingRouteParams.routeOptions = routeOptions;
-        this.props.history.push(pathname + (search ? '?' + search : ''));
+        this.props.history.push(pathname + normalizedSearch);
       }
     } else {
       this.handleNavigate(pathname, 'push', 'none', undefined, routeOptions, tab);
--- a/packages/react-router/test/base/tests/e2e/specs/routing.cy.js
+++ b/packages/react-router/test/base/tests/e2e/specs/routing.cy.js
@@ -344,6 +344,25 @@ describe('Routing Tests', () => {
     cy.get('div.ion-page[data-pageid=home-details-page-1] [data-testid="details-input"]').should('have.value', '1');
   });
 
+  it('Details 1 with Query Params > pop a same-URL history entry, should stay on details 1', () => {
+    // Regression: popstate over a same-URL entry on a search-bearing route used to trigger a false POP transition.
+    cy.visit(`http://localhost:${port}/routing`);
+    cy.ionNav('ion-item', 'Details 1 with Query Params');
+    cy.ionPageVisible('home-details-page-1');
+    cy.location('search').should('eq', '?hello=there');
+
+    cy.window().then((win) => {
+      win.history.pushState({ marker: true }, '', win.location.href);
+    });
+
+    cy.go('back');
+
+    // No teleport: still on details 1, URL unchanged.
+    cy.ionPageVisible('home-details-page-1');
+    cy.location('pathname').should('eq', '/routing/tabs/home/details/1');
+    cy.location('search').should('eq', '?hello=there');
+  });
+
   /*
     Tests to add:
     Test that lifecycle events fire
--- a/packages/react-router/test/base/tests/e2e/specs/routing.cy.js
+++ b/packages/react-router/test/base/tests/e2e/specs/routing.cy.js
@@ -250,7 +250,7 @@ describe('Routing Tests', () => {
   it('/routing/ > Details 1 > Details 2 > Details 3 > Back > Settings Tab > Home Tab > Should be at details 2 page', () => {
     // fixes an issue where route history was being lost after starting to go back, switching tabs
     // and switching back to the same tab again
-    // for bug https://github.com/ionic-team/ionic-framework/issues/21834
+    // For bug https://github.com/ionic-team/ionic-framework/issues/21834
     cy.visit(`http://localhost:${port}/routing`);
     cy.ionPageVisible('home-page');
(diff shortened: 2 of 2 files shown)


EricHech:
Hoping to get some eyes on this, it's a one-line fix that solves a real routing bug in Ionic.

ShaneK:
Hey! Thank you for your PR! Sorry this has taken so long to get to. 

FWIW this issue does not exist in V9, it sort of incidentally got taken care of during the rewrite for RR6. 

I've made a couple of small changes to your PR to fix a couple of issues. I found an edge case and also noticed that your test didn't consistently fail without the fix applied. I've addressed those and we'll get this merged soon and hopefully it can go out this week or next in our normal patch release.

brandyscarney (approved):
Thank you for the PR! 🙂

Issue #931, opened 2014-03-27
Scrolling broken on long forms (focus cancels swipe gesture)


rvanbaalen:
When I have a long form in an `<ion-content>` element, I can't swipe down the list of form fields. As soon as I touch the area, it's recognized as a tap and the iPhone (4) focusses on the field. But all I wanted was to scroll down, since I actually made a swipe gesture and not a specific touch.

To sum it up:
- Long form in scroll view
- Scroll down the long form
- As soon as finger hits the form (to start a swipe gesture), device focusses on that form field
- Scroll 'cancels' out or continues with unexpected / weird results (focused cursor somewhere between two fields)

I hope I made things clear enough.

PR #31164, opened 2026-05-22
fix(split-pane): inline default literal for when

gnbm:
Issue number: resolves N/A

---------

`IonSplitPane.when` is initialized to `QUERY['lg']`, where `QUERY` is a local `const` mapping shortcut keys to CSS media query strings.

An upcoming Stencil compiler change ([ionic-team/stencil#6728](https://github.com/ionic-team/stencil/pull/6728)) resolves `@Prop` default initializers that reference identifiers / property accesses / element accesses to their underlying primitive literal when generating `components.d.ts`. Once that change ships, a clean rebuild of `@ionic/core` will produce a `components.d.ts` whose `@default` JSDoc reads `'(min-width: 992px)'` instead of the committed `QUERY['lg']`, failing the `git diff --exit-code` guard in `test-core-clean-build`, `build-angular`, `build-vue`, and `build-react`.

The [Stencil Nightly Build (#931)](https://github.com/ionic-team/ionic-framework/actions/runs/26273945131) — which is pointed at Stencil's `main` — is already failing on exactly this diff:

```diff
-          * @default QUERY['lg']
+          * @default '(min-width: 992px)'
```

## What is the new behavior?

- `core/src/components/split-pane/split-pane.tsx`: inline the literal `'(min-width: 992px)'` as the prop default instead of dereferencing `QUERY['lg']`.
- `core/src/components.d.ts`: regenerate the two `@default` JSDoc lines (one in `Components`, one in `LocalJSX`) to match.

Both the current and upcoming Stencil compiler now emit identical output. Runtime behavior is unchanged: the shortcut lookup `QUERY[query] || query` continues to resolve user-supplied shortcuts (`'lg'`, `'md'`, etc.) via the `QUERY` map at the same place as before.

-
-
-

- [ ] Yes
- [x] No

The default value is identical (`'(min-width: 992px)'` is exactly what `QUERY['lg']` evaluated to). Consumers passing the shortcut `'lg'`/`'md'`/etc. to `when` keep working unchanged.

## Other information

- Author note: I'm also the author of the Stencil PR ([ionic-team/stencil#6728](https://github.com/ionic-team/stencil/pull/6728)) that surfaced this drift. This Ionic-side change is forward-compatible — it works against both the current and the new Stencil compiler — so it can land independently of any Stencil release.
- Audit follow-up: a grep across `core/src/components` did not find any other `@Prop` default that uses an identifier / property / element access pattern, so `IonSplitPane` was the only affected component.

### DIFF
--- a/core/api.txt
+++ b/core/api.txt
@@ -1833,7 +1833,7 @@ ion-spinner,css-prop,--color
 ion-split-pane,shadow
 ion-split-pane,prop,contentId,string | undefined,undefined,false,true
 ion-split-pane,prop,disabled,boolean,false,false,false
-ion-split-pane,prop,when,boolean | string,QUERY['lg'],false,false
+ion-split-pane,prop,when,boolean | string,'(min-width: 992px)',false,false
 ion-split-pane,event,ionSplitPaneVisible,{ visible: boolean; },true
 ion-split-pane,css-prop,--border,ios
 ion-split-pane,css-prop,--border,md
--- a/core/src/components.d.ts
+++ b/core/src/components.d.ts
@@ -3313,7 +3313,7 @@ export namespace Components {
         "isVisible": () => Promise<boolean>;
         /**
           * When the split-pane should be shown. Can be a CSS media query expression, or a shortcut expression. Can also be a boolean expression.
-          * @default QUERY['lg']
+          * @default '(min-width: 992px)'
          */
         "when": string | boolean;
     }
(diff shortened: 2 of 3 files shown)


gnbm:
Closing this — thanks for the review feedback.

Summary of why this is not the right fix:

- The Stencil Nightly Build failing on a `components.d.ts` diff is **working as intended**: it's a manual-review gate that surfaces every Stencil change that would alter generated output, so the change can be reviewed and accepted (or reverted) consciously rather than landing silently.
- Resolving `@Prop` defaults to their underlying literal in the generated JSDoc (Stencil PR ionic-team/stencil#6728) is an **improvement** to the docs comments in `components.d.ts`, not a regression to work around.
- The correct path is to wait until `@ionic/core` bumps `@stencil/core` to a release containing that change, then commit the regenerated `components.d.ts` together with the bumped `package.json` / `package-lock.json` as one intentional review-able change.

The approach in this PR (inlining the literal so neither old nor new Stencil produces a diff) papers over the review signal and strips the readability of the `QUERY['lg']` named constant in `split-pane.tsx`, so it's the wrong direction.

Closing and deleting the branch — no action needed in core today.

PR #31165, opened 2026-05-22
fix(overlays): show focus indicators only during keyboard navigation and improve focus management

brandyscarney:
Issue number: internal

---------

The focus indicator is displayed on the option when opening certain interfaces by clicking on the Select component:

| Modal | Popover |
| --- | --- |
| select-modal | select-popover |

## What is the new behavior?
- Hide the focus indicator when the Select component is opened using pointer events
- Display the focus indicator when the Select component is opened using keyboard navigation (Space or Enter)
- Cycle focus through the Select elements (radios, handle, cancel buttons) in the correct order and allow `Shift + Tab` to go backwards in order
- Removes the native focus from Checkbox and does not allow focusable when inside of an item (this matches Radio's behavior)
- Adds e2e tests for focus functionality in both Select and Modal to avoid future regressions

- [ ] Yes
- [x] No

## Other information

- [Select](https://ionic-framework-git-fw-6922-ionic1.vercel.app/src/components/select/test/basic/)
- [Modal: Sheet](https://ionic-framework-git-fw-6922-ionic1.vercel.app/src/components/modal/test/sheet/)


core/src/components/item/item.ionic.scss
```diff
@@ -114,10 +114,6 @@ slot[name="end"]::slotted(*) {
 
 // Item in Select Modal
 // --------------------------------------------------
-:host(.in-select-modal) {
```
brandyscarney: This was added to hide the focused state.

core/src/components/select-modal/select-modal.ionic.scss
```diff
@@ -15,11 +15,6 @@ ion-item {
   --border-width: 0;
 }
 
-ion-item.ion-focused::part(native)::after {
```
brandyscarney, line 27: This was added to hide the focused state.

core/src/components/select/test/basic/select.e2e.ts-snapshots/select-basic-action-sheet-scroll-to-selected-ionic-md-ltr-light-Mobile-Chrome-linux.png
brandyscarney, line 1: The ionic theme wasn't added to this test so now it is generating screenshots.

core/src/components/select/test/basic/select.e2e.ts-snapshots/select-basic-modal-scroll-to-selected-ios-ltr-Mobile-Chrome-linux.png
brandyscarney, line 1: The focus indicator is no longer displayed when opening with a mouse click.

core/src/components/select/test/basic/select.e2e.ts-snapshots/select-basic-modal-scroll-to-selected-md-ltr-Mobile-Chrome-linux.png
brandyscarney, line 1: The focus indicator is no longer displayed when opening with a mouse click.

core/src/components/select/test/basic/select.e2e.ts-snapshots/select-basic-popover-scroll-to-selected-ios-ltr-Mobile-Chrome-linux.png
brandyscarney, line 1: The focus indicator is no longer displayed when opening with a mouse click.

core/src/components/select/test/basic/select.e2e.ts-snapshots/select-basic-popover-scroll-to-selected-md-ltr-Mobile-Chrome-linux.png
brandyscarney, line 1: The focus indicator is no longer displayed when opening with a mouse click.

core/src/components/select/test/basic/select.e2e.ts
```diff
@@ -8,7 +8,7 @@ import { configs, test } from '@utils/test/playwright';
  * does not. The overlay rendering is already tested in the respective
  * test files.
  */
-configs({ directions: ['ltr'] }).forEach(({ title, config, screenshot }) => {
+configs({ modes: ['ios', 'md', 'ionic-md'], directions: ['ltr'] }).forEach(({ title, config, screenshot }) => {
```
brandyscarney, line 11: This now checks for the ionic theme and captures screenshots to avoid future focus behavior regressions.

core/src/components/checkbox/checkbox.tsx
```diff
@@ -375,10 +376,11 @@ export class Checkbox implements ComponentInterface {
         onClick={this.onClick}
         class={createColorClasses(color, {
           [theme]: true,
-          'in-item': hostContext('ion-item', el),
+          'in-item': inItem,
           'checkbox-checked': checked,
           'checkbox-disabled': disabled,
-          'ion-focusable': true,
+          // Focus styling should not apply when the checkbox is in an item
+          'ion-focusable': !inItem,
```
brandyscarney, line 383: This matches [radio](https://github.com/ionic-team/ionic-framework/blob/ece483c32256c3f10ebab73411bf53a86428c497/core/src/components/radio/radio.tsx#L239) behavior. Otherwise when a checkbox in an item has focus (such as inside of a select popover) the item and checkbox will both show a focus indicator.

core/src/components/checkbox/test/basic/checkbox.e2e.ts-snapshots/checkbox-focus-ios-ltr-Mobile-Chrome-linux.png
brandyscarney, line 1: This was removed intentionally because we don't want to show the native focus when we are adding our own. We need to add some `ion-focused`  styles to checkbox for `ios` and `md` to fix this. Should I make a follow-up?

core/src/components/select/test/basic/select.e2e.ts
```diff
@@ -318,10 +667,7 @@ configs({ directions: ['ltr'] }).forEach(({ title, config, screenshot }) => {
         await expect(popover).toBeVisible();
       });
 
-      test('holding Enter to open a popover should not immediately dismiss it', async ({ page, skip }, testInfo) => {
-        // TODO (ROU-5437)
-        skip.browser('webkit', 'Safari 16 only allows text fields and pop-up menus to be focused.');
-
```
brandyscarney, line 324: I updated the keyboard presses to this:

```diff
-await page.keyboard.press('Enter');
+await pageUtils.pressKeys('Enter');
``` 

This handles Safari tabbing so we are able to check Safari again in most of these tests.

core/src/utils/overlays.ts
```diff
@@ -44,8 +44,146 @@ type OverlayWithFocusTrapProps = HTMLIonOverlayElement & {
...
+};
+
+/**
+ * Sheet modals can have visual order that differs from DOM order.
+ * Without sorting, Tab can skip the handle after option traversal.
+ *
+ * Order: option controls (radio/checkbox) → sheet handle → end-slot
+ * header button. Any other focusables are appended after.
+ */
+const sortSheetModalFocusables = (overlay: HTMLElement, elements: HTMLElement[]): HTMLElement[] => {
```
ShaneK: This adds a fair bit of component-specific structure to the generic overlay trap: `sortSheetModalFocusables` knows the sheet layout, and elsewhere the trap matches `ion-select-modal`, `.action-sheet-button`, `.modal-handle`, `select-interface-option`, and the header Cancel button by their private class names. Before this PR the trap only knew to exclude `ion-toast`, so each of these selectors is now an implicit contract that breaks Tab order if those components get restyled. The roving-tabindex resolution feels general enough to live here, but would it be cleaner for the components to tag their participating elements with a shared marker (a `data-` attr or one documented class) the trap reads generically? Or I guess we could at least leave a note at each class definition that it's load-bearing for the focus trap?

core/src/components/checkbox/checkbox.tsx
ShaneK, line 383: Bringing checkbox to parity with radio here makes sense. One edge case worth confirming: in a multi-input item `hasCover()` is false, so the item won't carry `ion-focusable` either, and the checkbox was the `.ion-focusable` child the item relied on. Does a checkbox sitting next to a second input still show a focus ring somewhere when you tab to it? Radio has the same shape today so I don't think it's a blocker, just want to make sure it's intentional.

core/src/components/select/test/basic/select.e2e.ts
ShaneK, line 11: You're widening this to `ionic-md` but not `ionic-ios`, even though the `item.ionic.scss` and `select-modal.ionic.scss` changes apply to both Ionic modes. Was leaving out `ionic-ios` deliberate?

ShaneK (changes requested):
This is looking really good! I tested the functionality and the changes here seem to work fine, but I noticed some things and just wanted a bit of clarification. Good work so far, though!

core/src/utils/overlays.ts
brandyscarney: Thank you for calling this out! I went back and forth and decided on adding three new `data-` attributes that the focus trap can read to determine the order: https://github.com/ionic-team/ionic-framework/pull/31165/commits/9e200edbfdc8be2d7544b64fbc15d091bb2b5af7

I also added some documentation explaining these new markers: https://github.com/ionic-team/ionic-framework/pull/31165/changes/272c7f72f87d9309a343315f04b29225602656cc

Let me know if you think anything should be changed!

core/src/components/checkbox/checkbox.tsx
brandyscarney: This is indeed an issue but it is a bigger issue than I think this ticket should handle. After my change, an item with two checkboxes shows **no** focus state on the item or the individual checkboxes, but before this change it didn't show a focus state for the individual checkboxes either. Radio seems to have the same pre-existing issue. 

The bigger problem here seems to be Checkbox lacks focus styles entirely for `ios` and `md`. Radio has them with the `ion-focused` class, see: 

Radios

But Checkbox doesn't have any at all:

Checkboxes

As a result I have created two tickets to follow up on this: 
- Add checkbox focus indicator: `FW-7585`
- Multiple inputs in item lack individual focus: `FW-7586`

core/src/components/select/test/basic/select.e2e.ts
brandyscarney: Yes, I think adding `ionic-ios` would add a lot of screenshots with no benefit. The visual difference between the `ios` and `md` modes is nonexistent because the `ionic` theme applies the same structure and styles regardless of the mode. 

Here is the same screenshot for the `ionic` theme in each mode:

| `ionic-md` | `ionic-ios` |
| --- | --- |
| select-basic-popover-opened-with-value-focused-ionic-md-ltr-light-Mobile-Chrome-linux 1 | select-basic-popover-opened-with-value-focused-ionic-ios-ltr-light-Mobile-Chrome-linux 1 |

I also confirmed there are no behavioral differences between the two modes across all four select interfaces:

- Identical focusable elements rendered in both modes.
- The focus sequence through each interface is the same, as the focus-trap and roving focus logic are not mode-dependent.

So `ionic-ios` coverage here would only duplicate `ionic-md`. If we add different behavior in the future based on the mode I could see a reason to add it.

core/src/components/modal/test/sheet/modal.e2e.ts
```diff
@@ -391,6 +391,75 @@ configs({ modes: ['ios', 'ionic-ios'], directions: ['ltr'] }).forEach(({ title,
 
       await expect(dragHandle).toBeFocused();
     });
+
+    test('it should preserve the last arrow-focused radio when tabbing', async ({ page, pageUtils }) => {
+      await page.goto('/src/components/modal/test/sheet', config);
```
ShaneK: Is this useful since you call page.setContent right after it? I thought setContent fully replaced the page's content, but I may be mistaken and am just curious

ShaneK (approved):
This looks great to me! Thanks for fixing that issue and addressing the questions I left before! Just a very minor non-blocking question

core/src/components/checkbox/test/basic/checkbox.e2e.ts-snapshots/checkbox-focus-ios-ltr-Mobile-Chrome-linux.png
ShaneK: Probably? Just so we don't forget it in the future

thetaPC: Yes, let's create a ticket and then can you link it to the checkbox modular migration ticket? If it's not done by then, we can just add it during the migration.

core/src/components/item/item.tsx
```diff
@@ -259,7 +259,12 @@ export class Item implements ComponentInterface, AnchorInterface, ButtonInterfac
 
   private isFocusable(): boolean {
     const focusableChild = this.el.querySelector('.ion-focusable');
-    return this.canActivate() || focusableChild !== null;
+    // An item is focusable when it can receive keyboard focus: when it is
```
thetaPC, line 262: Is the colon supposed to be a comma?


```suggestion
    // An item is focusable when it can receive keyboard focus, when it is
```

core/src/components/modal/test/sheet/modal.e2e.ts
thetaPC: I'm under the same assumption. I'm assuming it was from a copy and paste.

core/src/components/select/test/basic/select.e2e.ts-snapshots/select-basic-action-sheet-opened-focused-ionic-md-ltr-light-Mobile-Chrome-linux.png
thetaPC, line 1: Do we need a snapshot of the entire screen? It seems that a snapshot of the overlay would work. It would be great if we can get in the habit of only capturing the overlay if that's all we need. It limits the need to update snapshots if the main page changes.

core/src/components/select/test/basic/select.e2e.ts
```diff
@@ -60,12 +64,26 @@ configs({ directions: ['ltr'] }).forEach(({ title, config, screenshot }) => {
...
+        // Tabbing moves focus into the radio group, focusing the first option
+        // and showing the focus ring.
+        await pageUtils.pressKeys('Tab');
+        await page.waitForChanges();
+
+        const firstOption = alert.locator('.alert-radio-button').nth(0);
+        await expect(firstOption).toBeFocused();
+        await expect(firstOption).toHaveClass(/ion-focused/);
+
+        await expect(alert).toHaveScreenshot(screenshot(`select-basic-alert-opened-focused`));
```
thetaPC: Do we need a snapshot when you've already tested for focus? The snapshots seem redundant in my opinion especially since we have Alert state tests.

core/src/components/select/test/basic/select.e2e.ts
```diff
@@ -90,9 +108,101 @@ configs({ directions: ['ltr'] }).forEach(({ title, config, screenshot }) => {
...
+
+        // Tabbing moves focus to the selected option and shows the focus ring.
+        await pageUtils.pressKeys('Tab');
+        await page.waitForChanges();
+
+        const selectedOption = alert.locator('.alert-radio-button').nth(1);
+        await expect(selectedOption).toBeFocused();
+        await expect(selectedOption).toHaveClass(/ion-focused/);
+
+        await expect(alert).toHaveScreenshot(screenshot(`select-basic-alert-opened-with-value-focused`));
```
thetaPC: Do we need a snapshot when you've already tested for focus? The snapshots seem redundant in my opinion especially since we have Alert state tests.

core/src/components/select/test/basic/select.e2e.ts
```diff
@@ -142,12 +255,27 @@ configs({ directions: ['ltr'] }).forEach(({ title, config, screenshot }) => {
...
+
+        // Tabbing moves focus to the first option and shows the focus ring.
+        await pageUtils.pressKeys('Tab');
+        await page.waitForChanges();
+
+        const firstOption = actionSheet.locator('.action-sheet-button').nth(0);
+        await expect(firstOption).toBeFocused();
+        await expect(firstOption).toHaveClass(/ion-focused/);
+
+        await expect(actionSheet).toHaveScreenshot(screenshot(`select-basic-action-sheet-opened-focused`));
```
thetaPC: Do we need a snapshot when you've already tested for focus? The snapshots seem redundant in my opinion especially since we have Action Sheet state tests.

core/src/components/select/test/basic/select.e2e.ts
```diff
@@ -172,9 +300,19 @@ configs({ directions: ['ltr'] }).forEach(({ title, config, screenshot }) => {
...
+        await expect(actionSheet.locator('.action-sheet-button.ion-focused')).toHaveCount(0);
+
+        // Tabbing moves focus to the Cancel button and shows the focus ring.
+        await pageUtils.pressKeys('Tab');
+        await page.waitForChanges();
+        const cancelButton = actionSheet.getByRole('button', { name: 'Cancel' });
+        await expect(cancelButton).toBeFocused();
+        await expect(cancelButton).toHaveClass(/ion-focused/);
+
+        await expect(actionSheet).toHaveScreenshot(screenshot(`select-basic-action-sheet-opened-with-value-focused`));
```
thetaPC: Do we need a snapshot when you've already tested for focus? The snapshots seem redundant in my opinion especially since we have Action Sheet state tests.

core/src/components/select/test/basic/select.e2e.ts
```diff
@@ -213,36 +351,160 @@ configs({ directions: ['ltr'] }).forEach(({ title, config, screenshot }) => {
...
+        await page.waitForChanges();
 
-      test('it should focus the second option when opened with a value', async ({ page, skip }) => {
-        // TODO (ROU-5437)
-        skip.browser('webkit', 'Safari 16 only allows text fields and pop-up menus to be focused.');
+        const secondOption = popover.locator('.select-interface-option').nth(1);
+        await expect(secondOption.locator('ion-radio')).toBeFocused();
+        await expect(secondOption).toHaveClass(/ion-focused/);
 
+        await expect(popover).toHaveScreenshot(screenshot(`select-basic-popover-opened-focused`));
```
thetaPC: Do we need a snapshot when you've already tested for focus? The snapshots seem redundant in my opinion especially since we have Select Popover state tests.

core/src/components/select/test/basic/select.e2e.ts
```diff
@@ -267,25 +529,112 @@ configs({ directions: ['ltr'] }).forEach(({ title, config, screenshot }) => {
...
+
+        // Arrowing moves focus to the next option and shows the focus ring.
+        await pageUtils.pressKeys('ArrowDown');
+        await page.waitForChanges();
+
+        const orangesOption = popover.locator('.select-interface-option:has-text("Oranges")');
+        await expect(orangesOption.locator('ion-radio')).toBeFocused();
+        await expect(orangesOption).toHaveClass(/ion-focused/);
+
+        await expect(popover).toHaveScreenshot(screenshot(`select-basic-popover-opened-with-value-focused`));
```
thetaPC: Do we need a snapshot when you've already tested for focus? The snapshots seem redundant in my opinion especially since we have Select Popover state tests.

core/src/components/select/test/basic/select.e2e.ts
```diff
@@ -267,25 +529,112 @@ configs({ directions: ['ltr'] }).forEach(({ title, config, screenshot }) => {
...
+        // The selected option (Bananas, the 2nd), not the first, is focused and
+        // shows the focus ring.
+        const firstOption = popover.locator('.select-interface-option').nth(0);
+        const selectedOption = popover.locator('.select-interface-option').nth(1);
+        await expect(selectedOption.locator('ion-radio')).toBeFocused();
+        await expect(selectedOption).toHaveClass(/ion-focused/);
+        await expect(firstOption).not.toHaveClass(/ion-focused/);
+      });
+
+      test('it should scroll to selected option when opened', async ({ page }) => {
```
thetaPC, line 615: This test feels like it should live under `select-popover` instead.

core/src/components/select/test/basic/select.e2e.ts
```diff
@@ -371,16 +717,61 @@ configs({ directions: ['ltr'] }).forEach(({ title, config, screenshot }) => {
...
-        await expect(modal).toBeVisible();
+        // Arrowing moves focus to the next option and shows the focus ring.
+        await pageUtils.pressKeys('ArrowDown');
+        await page.waitForChanges();
+
+        const secondOption = modal.locator('.select-interface-option').nth(1);
+        await expect(secondOption.locator('ion-radio')).toBeFocused();
+        await expect(secondOption).toHaveClass(/ion-focused/);
+
+        await expect(modal).toHaveScreenshot(screenshot(`select-basic-modal-opened-focused`));
```
thetaPC: Do we need a snapshot when you've already tested for focus? The snapshots seem redundant in my opinion especially since we have Select Modal state tests.

core/src/components/select/test/basic/select.e2e.ts
```diff
@@ -405,9 +796,99 @@ configs({ directions: ['ltr'] }).forEach(({ title, config, screenshot }) => {
...
+
+        // Arrowing moves focus to the next option and shows the focus ring.
+        await pageUtils.pressKeys('ArrowDown');
+        await page.waitForChanges();
+
+        const orangesOption = modal.locator('.select-interface-option:has-text("Oranges")');
+        await expect(orangesOption.locator('ion-radio')).toBeFocused();
+        await expect(orangesOption).toHaveClass(/ion-focused/);
+
+        await expect(modal).toHaveScreenshot(screenshot(`select-basic-modal-opened-with-value-focused`));
```
thetaPC: Do we need a snapshot when you've already tested for focus? The snapshots seem redundant in my opinion especially since we have Select Modal state tests.

core/src/components/checkbox/test/basic/checkbox.e2e.ts-snapshots/checkbox-focus-ios-ltr-Mobile-Chrome-linux.png
brandyscarney: I actually created FW-7585 already for this https://github.com/ionic-team/ionic-framework/pull/31165#discussion_r3406040214 which requests the same thing. I linked it to the modular migration ticket.

thetaPC (changes requested):
Questions:
1. Multiple-select interfaces: if a checked option is not the first one, should that checked option receive focus when the overlay is opened with the keyboard, rather than the first option?

Issues:
1. Multiple-select on popover: I can only Tab to the first two options.
2. Multiple-select on popover: when I Tab into the second option and press Space to select it, it selects the first option instead.
3. Action sheet: when I Tab to open it, I do not see the focus indicator on the wrapper the way I do with the alert. This happens both with no value and with a value selected.

Comments:
1. The action sheet and alert behave differently from each other, and I find the wrapper ring surprising in general. I would have expected the focus indicator to be on the selected option when an overlay opens with a value, not on the wrapper. You mentioned the wrapper ring is the intended behavior: is that intended based on web standards, or is it inherited from the existing overlay focus-trap?

core/src/components/modal/test/sheet/modal.e2e.ts
brandyscarney: Oops, yes this shouldn't also be there! https://github.com/ionic-team/ionic-framework/pull/31165/commits/8341bd5820dcbc5ced6abcb197fd4c4e0dae4f7f

core/src/components/item/item.tsx
brandyscarney: No, everything after the colon is describing when it can receive keyboard focus.

core/src/components/select/test/basic/select.e2e.ts
brandyscarney: Removed: https://github.com/ionic-team/ionic-framework/pull/31165/changes/312857cb7033d88709081e6535ab15bdcf9046f5

core/src/components/select/test/basic/select.e2e.ts
brandyscarney: Removed: https://github.com/ionic-team/ionic-framework/pull/31165/changes/312857cb7033d88709081e6535ab15bdcf9046f5

core/src/components/select/test/basic/select.e2e.ts
brandyscarney: Removed: https://github.com/ionic-team/ionic-framework/pull/31165/changes/312857cb7033d88709081e6535ab15bdcf9046f5

core/src/components/select/test/basic/select.e2e.ts
brandyscarney: Removed: https://github.com/ionic-team/ionic-framework/pull/31165/changes/312857cb7033d88709081e6535ab15bdcf9046f5

core/src/components/select/test/basic/select.e2e.ts
brandyscarney: Removed: https://github.com/ionic-team/ionic-framework/pull/31165/changes/312857cb7033d88709081e6535ab15bdcf9046f5

core/src/components/select/test/basic/select.e2e.ts
brandyscarney: Removed: https://github.com/ionic-team/ionic-framework/pull/31165/changes/312857cb7033d88709081e6535ab15bdcf9046f5

core/src/components/select/test/basic/select.e2e.ts
brandyscarney: Removed: https://github.com/ionic-team/ionic-framework/pull/31165/changes/312857cb7033d88709081e6535ab15bdcf9046f5

core/src/components/select/test/basic/select.e2e.ts-snapshots/select-basic-action-sheet-opened-focused-ionic-md-ltr-light-Mobile-Chrome-linux.png
brandyscarney: Removed: https://github.com/ionic-team/ionic-framework/pull/31165/changes/312857cb7033d88709081e6535ab15bdcf9046f5

core/src/components/select/test/basic/select.e2e.ts
brandyscarney: Removed: https://github.com/ionic-team/ionic-framework/pull/31165/changes/312857cb7033d88709081e6535ab15bdcf9046f5

core/src/components/select/test/basic/select.e2e.ts
brandyscarney: This test is under each interface here:

Alert: https://github.com/ionic-team/ionic-framework/blob/9fb3990d396c49c55eb7ac1d6229af64318c70c9/core/src/components/select/test/basic/select.e2e.ts#L28
Action Sheet: https://github.com/ionic-team/ionic-framework/blob/9fb3990d396c49c55eb7ac1d6229af64318c70c9/core/src/components/select/test/basic/select.e2e.ts#L110
Popover: https://github.com/ionic-team/ionic-framework/blob/9fb3990d396c49c55eb7ac1d6229af64318c70c9/core/src/components/select/test/basic/select.e2e.ts#L275
Modal: https://github.com/ionic-team/ionic-framework/blob/9fb3990d396c49c55eb7ac1d6229af64318c70c9/core/src/components/select/test/basic/select.e2e.ts#L413

I think having some in a different place would be more confusing when they're all related to Select. These were also [added prior to my PR](https://github.com/ionic-team/ionic-framework/commit/8eaeb22e7a967100ffaadae8c8221e2e4888a3b6) so I don't think it's within the scope to move them.

thetaPC: That works for me!

thetaPC (approved):
LGTM

PR #31166, opened 2026-05-22
feat(thumbnail): add recipe and tokens

thetaPC:

---------

`ion-thumbnail` is already a single SCSS file (no per-theme fragmentation), but its styles are hardcoded rather than driven by the Modular Ionic token system. The TSX also still calls `getIonTheme` and applies a theme class to the host even though all three themes render identically.

In addition, `--size` controls both width and height with a single token, so non-square thumbnails are not possible without a separate `width`/`height` override.

## What is the new behavior?

- **Defined TypeScript Interface**: Added a new `thumbnail.interfaces.ts` with the `IonThumbnailRecipe` type.
- **Defined Theme Defaults**: Added per-theme token defaults in ios, md, and ionic theme files.
- **Split `--size` into `--width` + `--height`**: Non-square thumbnails are now possible.
- **Public CSS variables**: Replaced `--size` / `--border-radius` with the namespaced `--ion-thumbnail-width`, `--ion-thumbnail-height`, and `--ion-thumbnail-border-radius`.
- **Removed theme classes**: Thumbnails won't render with a theme defined as a class.
- **Updated consumers internally**: `ion-item-divider`, `ion-item` (ios + md theme files) now set `--ion-thumbnail-width` and `--ion-thumbnail-height` directly instead of `--size`.

- [x] Yes
- [ ] No

This PR introduces breaking changes to how `ion-thumbnail` is styled.

Migration Path:

1. **CSS Variable Replacements**: `--size` and `--border-radius` have been **removed**. Use the new token structure:

```
--size -> IonThumbnail.width + IonThumbnail.height
--border-radius -> IonThumbnail.border.radius
```

Code that previously relied on a single `--size` override must now set both `--ion-thumbnail-width` and `--ion-thumbnail-height`:

```html
// Before
<ion-thumbnail style="--size: 48px"></ion-thumbnail>

// After
<ion-thumbnail style="--ion-thumbnail-width: 48px; --ion-thumbnail-height: 48px"></ion-thumbnail>
```

If per-component customization is needed, the CSS variables can be used directly.

```
--size -> --ion-thumbnail-width / --ion-thumbnail-height
--border-radius -> --ion-thumbnail-border-radius
```

_Note: slotted thumbnails inside `ion-item` and `ion-item-divider` continue to render at their previous sizes — these components have been updated internally to set the new variables._

2. **Theme classes**: Remove any instances that target the theme classes: `ion-thumbnail.md`, `ion-thumbnail.ios`.

## Other information

Preview:
- [Thumbnail: basic](https://ionic-framework-git-fw-6862-ionic1.vercel.app/src/components/thumbnail/test/basic/)


core/scripts/vercel-build.sh
```diff
@@ -48,6 +48,7 @@ echo "Copying core output..."
 cp -r "${CORE_DIR}/src" "${OUTPUT_DIR}/src"
 cp -r "${CORE_DIR}/dist" "${OUTPUT_DIR}/dist"
 cp -r "${CORE_DIR}/css" "${OUTPUT_DIR}/css"
+cp -r "${CORE_DIR}/themes" "${OUTPUT_DIR}/themes"
```

core/src/components/item-divider/item-divider.scss
```diff
@@ -308,15 +308,15 @@
...
-
   @include mixins.margin(
     var(--ion-item-divider-thumbnail-margin-top),
     var(--ion-item-divider-thumbnail-margin-end),
     var(--ion-item-divider-thumbnail-margin-bottom),
     var(--ion-item-divider-thumbnail-margin-start)
   );
+
+  width: var(--ion-item-divider-thumbnail-width, revert-layer);
+  height: var(--ion-item-divider-thumbnail-height, revert-layer);
```
thetaPC, line 319: If these tokens are not set then default to the size that was set prior like the original `--ion-thumbnail-width`/`--ion-thumbnail-height`.

core/src/components/item/item.ios.scss
```diff
@@ -161,7 +161,9 @@
 // --------------------------------------------------
 
 ::slotted(ion-thumbnail) {
-  --size: #{$item-ios-thumbnail-size};
+  // TODO(FW-6847): replace $item-md-thumbnail-* to revert-layer pattern used in item-divider.scss. Assuming the --ion-item-thumbnail-width/height tokens will be used in the migration, else update them.
```
ShaneK:
```suggestion
  // TODO(FW-6847): replace $item-ios-thumbnail-* to revert-layer pattern used in item-divider.scss. Assuming the --ion-item-thumbnail-width/height tokens will be used in the migration, else update them.
```

Copy-paste from `item.md.scss` left this referencing `$item-md-thumbnail-*` even though the file uses `$item-ios-thumbnail-*` on the next two lines. Since both files carry the same note, a shorter `// TODO(FW-6847): adopt the revert-layer pattern from item-divider.scss` could also work

core/src/components/thumbnail/test/basic/thumbnail.e2e.ts
```diff
@@ -22,8 +22,15 @@ configs({ modes: ['ios'], directions: ['ltr'] }).forEach(({ title, screenshot, c
 
       await page.setContent(
         `
+          <style>
+            ion-thumbnail {
+              --ion-item-thumbnail-width: 20px;
+              --ion-item-thumbnail-height: 20px;
+            }
+          </style>
```
ShaneK, line 30: Should this set `--ion-item-thumbnail-width` and `--ion-item-thumbnail-height` on `ion-item` instead of `ion-thumbnail`? The `<ion-item-divider>` test right below sets the divider variant on `ion-item-divider`, so the two would line up. It works either way because the property cascades, but the variable name implies it belongs on the parent.

core/src/components/thumbnail/thumbnail.scss
```diff
@@ -1,27 +1,25 @@
...
-  --border-radius: 0;
-
-  @include border-radius(var(--border-radius));
+  @include mixins.border-radius(var(--ion-thumbnail-border-radius));
 
   display: block;
 
-  width: var(--size, 48px);
-  height: var(--size, 48px);
+  width: var(--ion-thumbnail-width);
```
ShaneK: `var(--ion-thumbnail-width)` and `var(--ion-thumbnail-height)` lost the literal fallback that `var(--size, 48px)` used to give you. I think the token system sets these globally so it's fine in practice, but if a consumer opts out of the default theme or the tokens haven't applied yet, the thumbnail collapses to `auto` and renders 0x0. Worth keeping a fallback, or is there a guarantee I'm missing?

ShaneK (approved):
Minor non-blocking nits, great work!

core/src/components/item/item.ios.scss
thetaPC: https://github.com/ionic-team/ionic-framework/pull/31166/changes/6b24703fb9e716d563d4d5a49b47614ba786c502

core/src/components/thumbnail/test/basic/thumbnail.e2e.ts
thetaPC: https://github.com/ionic-team/ionic-framework/pull/31166/changes/2b4b30ba0291f12d5d17859138539ecdfce41a78

core/src/components/thumbnail/thumbnail.scss
thetaPC: Good catch! https://github.com/ionic-team/ionic-framework/pull/31166/changes/987a13160ade5fa2bfcfd53f5088e180e046c383

BREAKING.md
```diff
@@ -294,3 +295,39 @@ Additionally, the `radio-group-wrapper` div element has been removed, causing sl
 Converted `ion-textarea` to use [Shadow DOM](https://developer.mozilla.org/en-US/docs/Web/Web_Components/Using_shadow_DOM).
 
 If you were targeting the internals of `ion-textarea` in your CSS, you will need to target the `wrapper`, `container`, `label`, `native`, `supporting-text`, `helper-text`, `error-text`, `counter`, or `bottom` [Shadow Parts](https://ionicframework.com/docs/theming/css-shadow-parts) instead, or use the provided CSS Variables.
+
+<h4 id="version-9x-thumbnail">Thumbnail</h4>
+
+The following breaking changes apply to `ion-thumbnail`:
+
+1. `--size` has been split into separate `--ion-thumbnail-width` and `--ion-thumbnail-height` CSS variables.
+2. `--border-radius` has been replaced.
```
brandyscarney: Should this say replaced with `--ion-thumbnail-border-radius`?

BREAKING.md
```diff
@@ -294,3 +295,39 @@ Additionally, the `radio-group-wrapper` div element has been removed, causing sl
...
+3. Theme classes (`ion-thumbnail.md`, `ion-thumbnail.ios`) are no longer supported.
+
+<h5>Removed CSS variables</h5>
+
+`--size` and `--border-radius` have been removed. Use the new token structure for global styles, or the corresponding CSS variable for component-specific overrides:
```
brandyscarney: Should it say `replaced` instead of `removed`?

brandyscarney (approved):
Looks great! Left some non-blocking comments on the breaking changes. 🎉

BREAKING.md
thetaPC: My intention with the numbered list is a quick summary of the changes. That way a dev can quickly scan to determine if it pertains to them. If it does then they can proceed to the respective section like `<h5>Removed CSS variables</h5>` for more details or they can just skip it.

BREAKING.md
thetaPC: https://github.com/ionic-team/ionic-framework/pull/31166/changes/e5746b81a39ae2f7c128742f4caf8087dfda168a

BREAKING.md
thetaPC: https://github.com/ionic-team/ionic-framework/pull/31166/changes/ac75dd16739c632174808f63959f500fdd066d9f

core/src/components/tab-bar/test/basic/tab-bar.e2e.ts-snapshots/tab-bar-default-ionic-md-rtl-light-Mobile-Firefox-linux.png
thetaPC, line 1: The old version was accidentally added during a sync. The original version was generated through a flaky test so it's not the true snapshot.

PR #31167, opened 2026-05-23
feat(img): add recipe and tokens

thetaPC:

---------

`ion-img` reads the active theme via `getIonTheme()` and applies it as a host class even though no SCSS rule depends on the theme class. The `@virtualProp theme` is also exposed publicly.

## What is the new behavior?

- **Removed theme classes**: Images won't render with a theme defined as a class.

- [x] Yes
- [ ] No

This PR introduces breaking changes to how `ion-img` is styled.

Migration Path:

1. **Theme classes**: Remove any instances that target the theme classes: `ion-img.md`, `ion-img.ios`.

## Other information

Previews
- [Image: basic](https://ionic-framework-git-fw-6845-ionic1.vercel.app/src/components/img/test/basic/)


core/scripts/vercel-build.sh
```diff
@@ -48,6 +48,7 @@ echo "Copying core output..."
 cp -r "${CORE_DIR}/src" "${OUTPUT_DIR}/src"
 cp -r "${CORE_DIR}/dist" "${OUTPUT_DIR}/dist"
 cp -r "${CORE_DIR}/css" "${OUTPUT_DIR}/css"
+cp -r "${CORE_DIR}/themes" "${OUTPUT_DIR}/themes"
```

core/src/components/img/test/basic/img.e2e.ts
```diff
@@ -4,8 +4,7 @@ import { configs, test } from '@utils/test/playwright';
 
 configs({ modes: ['md'], directions: ['ltr'] }).forEach(({ title, config }) => {
   test.describe(title('img: basic'), () => {
-    // TODO FW-3596
-    test.describe.skip('image successfully loads', () => {
+    test.describe('image successfully loads', () => {
```
thetaPC, line 7: This test was originally marked skip under `FW-3596` because it was flaky on CI. That predates the move to Docker based e2e runs, where the test environment is now deterministic (pinned browser, pinned OS, controlled timing). The flake was almost certainly caused by host environment variance.

Verified stable by running `npm run test.e2e.docker img/test/basic -- --repeat-each=80` locally several times. Zero failures.

If the test ends up flaking in the future, we can skip and create a new ticket to address it.

ShaneK (approved):
Looks good to me! Great work 🎉

core/src/components/tab-bar/test/basic/tab-bar.e2e.ts-snapshots/tab-bar-default-ionic-md-rtl-light-Mobile-Firefox-linux.png
thetaPC, line 1: The old version was accidentally added during a sync. The original version was generated through a flaky test so it's not the true snapshot.

brandyscarney (approved):

PR #31169, opened 2026-05-26
refactor(many): delete extra files and update vercel build

thetaPC:
Issue number: N/A

---------

- Some files were not deleted during the last sync.
- Tokens are no longer being loaded into the pages.

## What is the new behavior?

- Deleted files that are no longer being used within the Modular Ionic
- Tokens are now being loaded into the pages

- [ ] Yes
- [x] No

## Other information

N/A


core/scripts/vercel-build.sh
```diff
@@ -48,6 +48,7 @@ echo "Copying core output..."
 cp -r "${CORE_DIR}/src" "${OUTPUT_DIR}/src"
 cp -r "${CORE_DIR}/dist" "${OUTPUT_DIR}/dist"
 cp -r "${CORE_DIR}/css" "${OUTPUT_DIR}/css"
+cp -r "${CORE_DIR}/themes" "${OUTPUT_DIR}/themes"
```

brandyscarney (approved):

ShaneK (approved):
🚀

PR #31171, opened 2026-05-27
feat(menu-toggle): migrate to Modular Ionic

thetaPC:

---------

`ion-menu-toggle` does not fragment styles based on themes. All 3 themes share one style. However, it's not configured to the Modular Ionic.

## What is the new behavior?

- The virtual `theme` prop has been removed. The `ios` and`md` classes are no longer applied to the `ion-menu-toggle` element. 

- [x] Yes
- [ ] No

This PR introduces breaking changes to how `ion-menu-toggle` is styled.

Migration Path:

1. **Theme classes**: Remove any instances that target the theme classes: `ion-menu-toggle.md`, `ion-menu-toggle.ios`.

## Other information

Previews
- [Menu toggle: basic](https://ionic-framework-git-fw-6854-ionic1.vercel.app/src/components/menu-toggle/test/basic/)
- [Menu toggle: button](https://ionic-framework-git-fw-6854-ionic1.vercel.app/src/components/menu-toggle/test/button/)
- [Menu toggle: list](https://ionic-framework-git-fw-6854-ionic1.vercel.app/src/components/menu-toggle/test/list/)


core/src/components/menu-toggle/menu-toggle.tsx
```diff
@@ -45,23 +42,25 @@ export class MenuToggle implements ComponentInterface {
   @Listen('ionMenuChange', { target: 'body' })
   @Listen('ionSplitPaneVisible', { target: 'body' })
   async visibilityChanged() {
-    this.visible = await updateVisibility(this.menu);
+    try {
+      this.visible = await updateVisibility(this.menu);
+    } catch {
+      this.visible = false;
+    }
```
thetaPC, line 49: `connectedCallback` and the `@Listen` handlers discard the Promise returned by `visibilityChanged()`, so any rejection would surface as unhandled. The current chain can't reject, but catching here makes the method safe against future changes to `menuController.get()` (from `@utils/menu-controller`) or `ion-menu`'s `isActive`.

ShaneK, line 49: Nit: This try/catch is a nice change, but do we want a test covering that path? The basic e2e only exercises open/close right now, so nothing locks in the rejection fallback.

BREAKING.md
```diff
@@ -272,6 +273,16 @@ If you were targeting the internals of `ion-input-otp` in your CSS, you will nee
...
 
+<h4 id="version-9x-menu-toggle">Menu Toggle</h4>
+
+The following breaking changes apply to `ion-menu-toggle`:
+
+1. Theme classes (`ion-menu-toggle.md`, `ion-menu-toggle.ios`) are no longer supported.
+
+<h5>Theme classes</h5>
+
+Remove any instances that target the theme classes: `ion-menu-toggle.md`, `ion-menu-toggle.ios`.
```
ShaneK, line 297: Nit: This reads a little inconsistent with the other single-change entries. `Item Divider` and `Spinner` both state the theme-class removal as one bullet, no numbered list or `<h5>`. The numbered-list-plus-`<h5>` shape is what `Content` uses, but Content has four separate changes; here item 1 and the `<h5>Theme classes</h5>` body say nearly the same thing. Maybe we should collapse it to a single bullet to match item-divider? Totally up to you, probably doesn't matter much

ShaneK (approved):
Looks great! I have a couple of nits but feel free to disregard

brandyscarney (approved):
LGTM!

core/src/components/menu-toggle/menu-toggle.tsx
thetaPC: https://github.com/ionic-team/ionic-framework/pull/31171/changes/e6c78b4e178959b4d7a2c22939ce2b88992855d2

BREAKING.md
thetaPC: I plan to update the other components to follow this structure regardless of how many there are. That way it's always consistent.

PR #31172, opened 2026-05-27
fix(router): support anchor fragments in href

### LINKED ISSUE #19365: bug: Support for Angular's 'anchorScrolling' router option
**Ionic version:**

[x] **4.x**

**Current behavior:**
Angular has options to enable anchor handling using the following router config:

```
RouterModule.forRoot(routes, {
      anchorScrolling: 'enabled',
      onSameUrlNavigation: 'reload', // required to handle clicking same anchor twice
      scrollPositionRestoration: 'enabled'
    })
```
These options do not work when Ionic is used.

**Expected behavior:**
The options should be supported or alternate solution provided.

**Steps to reproduce:**
1. Create a page with anchors using `routerLink` with `fragment` directive to different sections. 
2. Try clicking on anchors.

**Ionic info:**

```
Ionic:

   Ionic CLI                     : 5.2.4 (/Users/[user]/.nvm/versions/node/v10.15.3/lib/node_modules/ionic)
   Ionic Framework               : @ionic/angular 4.9.0
   @angular-devkit/build-angular : 0.13.9
   @angular-devkit/schematics    : 7.3.9
   @angular/cli                  : 7.3.9
   @ionic/angular-toolkit        : 1.5.1

Cordova:

   Cordova Platforms : ios 5.0.1
   Cordova Plugins   : cordova-plugin-ionic-keyboard 2.1.3, cordova-plugin-ionic-webview 2.5.1, (and 14 other plugins)

Utility:

   cordova-res : 0.6.0
   native-run  : 0.2.8
```

### LINKED ISSUE #19566: bug:
# Bug Report

**Ionic version:** 4.9.0

**Current behavior:**
When clicking on a `ion-router-link` whose `href` attribute contains a hash (e.g. `/catalog#pens`) the app doesn't navigate to the `/catalog` URL, and an error is printed to the console :

image

**Expected behavior:**
The app should navigate to the `/catalog` page and scroll down to the `#pens` heading.

**Related code:**

```jsx
// app-root.tsx

<ion-app>
  <ion-router useHash={false}>
    <ion-route url='/' component='home-page' />
    <ion-route url='/catalog' component='catalog-page' />
  </ion-router>

  <ion-nav animated={false} />
</ion-app>
```

```jsx
// app-footer.tsx

<ion-router-link href='/catalog#pens'>Pens</ion-router-link>
```

**Other information:** It's a `ionic-pwa` Stencil app so I don't have the `ionic` CLI to provide additional info.

ShaneK:
Issue number: resolves #19566, resolves #19365

---------

Clicking an `ion-router-link` whose `href` contains a `#fragment` (e.g. `/catalog#pens`) fails to navigate and logs `not part of the routing set`. `parsePath` in `core/src/components/router/utils/path.ts` only splits on `?`, so the fragment gets folded into the last path segment and no registered route matches. Even when navigation succeeds, the matching anchor is never scrolled into view.

## What is the new behavior?

`parsePath` extracts the fragment as a separate field, and `writeSegments` writes it back onto the URL via `pushState`. Routes match against the path segments alone, so `/catalog#pens` resolves to the `/catalog` route as expected.

After `writeNavStateRoot` resolves, the router polls a few animation frames for an element matching the fragment inside the active page, then smooth-scrolls to it via `ion-content.scrollToPoint`, falling back to `Element.scrollIntoView` when the target sits outside `ion-content`. A `fragmentScrollToken` cancels any in-flight scroll the moment a newer `setSegments` runs.

- [ ] Yes
- [x] No

## Other information

The fragment lookup scopes to the last `.ion-page:not(.ion-page-hidden)` in the document. Nested outlets can have multiple non-hidden pages mounted at once, and an identically-named anchor on a parent page would otherwise win over the leaf the user is viewing.

Preview pages:
- [router/basic](https://ionic-framework-git-fw-6958-ionic1.vercel.app/src/components/router/test/basic#/two)

### DIFF
--- a/core/src/components/router/test/basic/router.e2e.ts
+++ b/core/src/components/router/test/basic/router.e2e.ts
@@ -1,5 +1,20 @@
 import { expect } from '@playwright/test';
+import type { E2EPage } from '@utils/test/playwright';
 import { configs, test } from '@utils/test/playwright';
+
+/**
+ * Waits until `page-two`'s `ion-content` has scrolled past the fixture's 2000px
+ * spacer. The anchor target sits below the spacer, so a successful fragment
+ * scroll must move `scrollTop` well past it; a regression that scrolled by
+ * only a handful of pixels would fail this threshold.
+ */
+const waitForAnchorScrolled = (page: E2EPage) =>
+  page.waitForFunction(async () => {
+    const content = document.querySelector('page-two ion-content') as HTMLIonContentElement | null;
+    if (!content) return false;
+    const scrollEl = await content.getScrollElement();
+    return scrollEl.scrollTop > 1500;
+  });
 
 /**
  * This behavior does not vary across modes/directions.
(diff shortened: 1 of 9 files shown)


core/src/components/router/test/basic/router.e2e.ts
```diff
@@ -27,6 +42,188 @@ configs({ modes: ['ios'], directions: ['ltr'] }).forEach(({ title, config }) =>
 
       expect(page.url()).toContain('#/two/three/absolute');
     });
+
+    test('should route when ion-router-link href contains a fragment', async ({ page }, testInfo) => {
+      testInfo.annotations.push({
+        type: 'issue',
+        description: 'https://github.com/ionic-team/ionic-framework/issues/19365',
```
thetaPC, line 49: Is this issue number correct? The PR description has `19566`.

thetaPC (approved):
LGTM, left a question

core/src/components/router/test/basic/router.e2e.ts
ShaneK: I forgot to link that in the description, this PR addresses both. One is technically a duplicate of the other, but they're both pretty old at this point

thetaPC: I updated the PR description to have "resolves" in front of [each issue](https://docs.github.com/en/issues/tracking-your-work-with-issues/using-issues/linking-a-pull-request-to-an-issue). Else it won't close them all out.

core/src/components/router/test/basic/router.e2e.ts
```diff
@@ -27,6 +42,188 @@ configs({ modes: ['ios'], directions: ['ltr'] }).forEach(({ title, config }) =>
...
+      const errors: string[] = [];
+      page.on('pageerror', (e) => errors.push(e.message));
+      page.on('console', (msg) => {
+        if (msg.type() === 'error') errors.push(msg.text());
+      });
+
+      await page.goto(`/src/components/router/test/basic#/two`, config);
+      await page.click('#link-with-fragment');
+
+      await expect(page.locator('page-two')).toBeVisible();
```
brandyscarney, line 60: Do you think we should check for specific text on this page or is this sufficient?

core/src/components/router/test/basic/router.e2e.ts
```diff
@@ -27,6 +42,188 @@ configs({ modes: ['ios'], directions: ['ltr'] }).forEach(({ title, config }) =>
...
+
+    test('should route when ion-router-link href contains both query and fragment', async ({ page }, testInfo) => {
+      testInfo.annotations.push({
+        type: 'issue',
+        description: 'https://github.com/ionic-team/ionic-framework/issues/19365',
+      });
+      await page.goto(`/src/components/router/test/basic#/two`, config);
+      await page.click('#link-with-query-and-fragment');
+
+      await expect(page.locator('page-three')).toBeVisible();
```
brandyscarney, line 73: Do you think we should check for specific text on this page or is this sufficient?

core/src/components/router/test/dom.spec.tsx
```diff
@@ -0,0 +1,91 @@
...
+    `;
+  });
+
+  afterEach(() => {
+    document.body.innerHTML = '';
+  });
+
+  it('should return false for null/undefined/empty fragment', async () => {
+    expect(await scrollToFragment(undefined)).toBe(false);
+    expect(await scrollToFragment('')).toBe(false);
```
brandyscarney, line 18: Should you add a check here for a similarly named fragment like:

```tsx
expect(await scrollToFragment('target2')).toBe(false);
expect(await scrollToFragment('targe')).toBe(false);
```

brandyscarney (approved):
Looks really good and passed my testing. Just a few optional comments on the tests but overall a lot of great coverage was added!

core/src/components/router/test/basic/router.e2e.ts
ShaneK: The only thing I think that would help catch here would be catch by the scroll test, which verifies the anchor exists on the page, so I don't think this would be valuable

core/src/components/router/test/basic/router.e2e.ts
ShaneK: The only thing I think that would help catch here would be catch by the scroll test, which verifies the anchor exists on the page, so I don't think this would be valuable

core/src/components/router/test/dom.spec.tsx
ShaneK: Done here: [637eab1](https://github.com/ionic-team/ionic-framework/pull/31172/commits/637eab14c8a7ac5c3a16693dd80ac54f6584214a)

PR #31174, opened 2026-05-28
chore(ci): bump @playwright/test to ^1.60.0

ShaneK:

---------

Playwright tests are hanging because of a [bug that breaks it for node 24.16+](https://github.com/microsoft/playwright/issues/40724)

## What is the new behavior?
Upgrading Playwright to 1.16.0 to fix the issue

- [ ] Yes
- [X] No


thetaPC (approved):
LGTM

PR #31175, opened 2026-05-28
feat(toolbar): add hideOnScroll for header and footer toolbars

joselrio:

---------

This pull request introduces the new `hideOnScroll` feature for `ion-toolbar`, enabling toolbars in header and footer to automatically hide and show based on scroll direction. 

**Hide on Scroll Feature for Toolbars:**

* Added `hideOnScroll` property to `ion-toolbar`, updating the API documentation and exposing it in Angular and Vue proxies, allowing toolbars to hide on scroll in both headers and footers. [[1]](diffhunk://#diff-5e21becb925e949068c3cbefae90d040401eaf39874bfd1336d93a3cd8d64809R2788) [[2]](diffhunk://#diff-1fd76195dbb521bd6fc15558ddb677c06cd4a94d7541ecdea25409b875ea71daL2661-R2668) [[3]](diffhunk://#diff-23bd4dda0490183f78acea6eb747bfab3091a16ce2e1d4710f4c194669597628L2326-R2333) [[4]](diffhunk://#diff-e8cbb9c3666e073d4a751535b38a9f64291e2643c76108796cf6d1e5799a90d9L1138-R1139)
* Implemented logic in `ion-header` to prioritize `hideOnScroll` over collapsible large titles, and to skip collapse behavior when `hideOnScroll` is present. [[1]](diffhunk://#diff-2310daf8d40f7af1cde67dd4577f96893d660c3b5bdeff03aec1d3aa7aa08fbbR48-R50) [[2]](diffhunk://#diff-2310daf8d40f7af1cde67dd4577f96893d660c3b5bdeff03aec1d3aa7aa08fbbR86-R102)
* Added tests and demo page for the hide-on-scroll toolbar functionality.

**Styling and Layout Enhancements:**

* Introduced new CSS and SCSS for header, footer, toolbar, and content to support smooth hide/show transitions and proper overlay behavior when `hideOnScroll` is active. [[1]](diffhunk://#diff-ca32a7f956c8f0ec26416437c4d92f2a755b928f56e072ce50dcd3f9d380bb86R148-R220) [[2]](diffhunk://#diff-e867ed36d2ac34f224ab75d1adb5cfc81c2912644ac3ec5c4aaab001fcfca7a8L14-R38) [[3]](diffhunk://#diff-955328e924080d4388ac3836fcc637997828f555e4265a9154a5340f72e42f60R18-R38) [[4]](diffhunk://#diff-dbf4ccf7c3209b0b39d7e40f7a569a1f0aa07146b4df5482ab42e852d509ac77R100-R124)

- [ ] Yes
- [x] No

### DIFF
--- a/core/src/components/header/header.common.scss
+++ b/core/src/components/header/header.common.scss
@@ -1,3 +1,5 @@
+@use "../../themes/native/native.globals" as globals;
+
 // Header
 // --------------------------------------------------
 
@@ -12,4 +14,25 @@ ion-header {
 
 ion-header ion-toolbar:first-of-type {
   padding-top: var(--ion-safe-area-top, 0);
+}
+
+// Header Hide on Scroll
+// --------------------------------------------------
+// Added by `ion-toolbar` when `hideOnScroll` is enabled. Overlays `ion-content`.
+
+ion-header.ion-header-hide-on-scroll {
+  @include globals.position(0, 0, null, 0);
+  position: absolute;
+
+  transition: transform var(--token-transition-time-300, 300ms)
+    var(--token-transition-curve-quick, cubic-bezier(0, 0, 0.2, 1));
+
+  z-index: 10;
+}
+
+ion-header.ion-header-hide-on-scroll.header-scroll-hidden {
+  transform: translateY(calc(-100% - var(--ion-safe-area-top, 0)));
+
+  transition: transform var(--token-transition-time-200, 200ms)
+    var(--token-transition-curve-base, cubic-bezier(0.4, 0, 1, 1));
 }
--- a/core/src/components/toolbar/test/hide-on-scroll/index.html
+++ b/core/src/components/toolbar/test/hide-on-scroll/index.html
@@ -1,0 +1,169 @@
+<!DOCTYPE html>
+<html lang="en" dir="ltr" mode="ionic">
+  <head>
+    <meta charset="UTF-8" />
+    <title>Toolbar - Hide on Scroll</title>
+    <meta
+      name="viewport"
+      content="width=device-width, initial-scale=1.0, minimum-scale=1.0, maximum-scale=1.0, user-scalable=no, viewport-fit=cover"
+    />
+    <script src="../../../../../scripts/testing/scripts.js"></script>
+    <script nomodule src="../../../../../dist/ionic/ionic.js"></script>
+    <script type="module" src="../../../../../dist/ionic/ionic.esm.js"></script>
+    <link rel="stylesheet" href="../../../../../css/ionic.bundle.css" />
+    <link rel="stylesheet" href="../../../../../scripts/testing/styles.css" />
+  </head>
+
+  <body>
+    <template id="header-template">
+      <ion-header id="test-header">
+        <ion-toolbar hide-on-scroll="true">
+          <ion-title>IonToolbar inside IonHeader</ion-title>
+        </ion-toolbar>
+      </ion-header>
+    </template>
+
+    <template id="footer-template">
+      <ion-footer id="test-footer">
+        <ion-toolbar hide-on-scroll="true">
+          <ion-title>IonToolbar inside IonFooter</ion-title>
+        </ion-toolbar>
+      </ion-footer>
+    </template>
+
+    <ion-app>
+      <div class="ion-page">
+        <ion-content>
+          <div class="spacer">
+            <div class="info">
+              <p>Scroll down to hide the toolbars, scroll up to show them.</p>
+              <p><strong>Requirements:</strong></p>
+              <ul>
+                <li><code>hide-on-scroll</code> must be set on <code>ion-toolbar</code></li>
+                <li>Toolbar must be inside an <code>ion-header</code> or <code>ion-footer</code></li>
+                <li>A sibling <code>ion-content</code> must exist for scroll detection</li>
+              </ul>
+            </div>
+            <div class="toolbar-layout">
+              <ion-toggle id="toggle-header" justify="space-between" checked>Header toolbar</ion-toggle>
+              <ion-toggle id="toggle-footer" justify="space-between" checked>Footer toolbar</ion-toggle>
+              <p class="toolbar-layout-hint" id="layout-hint">Header and footer toolbars</p>
+            </div>
+          </div>
+        </ion-content>
+      </div>
+    </ion-app>
+
+    <style>
+      .spacer {
+        position: relative;
+        padding: 16px;
+        height: 2000px;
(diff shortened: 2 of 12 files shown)


core/src/components/toolbar/toolbar.tsx
```diff
@@ -266,6 +353,403 @@ export class Toolbar implements ComponentInterface {
...
+  }
+
+  /** Update `scrollTop` after each scroll event for the given scroll element. */
+  private static setScrollLastTop(scrollEl: HTMLElement, scrollTop: number) {
+    Toolbar.scrollLastTopByElement.set(scrollEl, scrollTop);
+  }
+
+  /** Get element with the scrollable content inside `ion-content` and register it to the scroll listener. */
+  private async initScrollListener(contentEl: HTMLElement) {
+    const scrollEl = (this.scrollEl = await getScrollElement(contentEl));
```
ShaneK, line 460: `registerScrollListener` runs right after `await getScrollElement` with nothing re-checking state. If the toolbar disconnects or `hideOnScroll` toggles during the await, `destroyHideOnScroll` has already unregistered and nulled `scrollEl`, then this resumes and re-adds `this` to the static `scrollListeners` set with no teardown left to remove it. That leaks the instance on fast mount/unmount. Can you capture the in-flight promise and bail if it's superseded or `hideOnScrollActive` is false after the await, like the pattern in `.claude/docs/components.md`?

core/src/components/toolbar/toolbar.tsx
```diff
@@ -266,6 +353,403 @@ export class Toolbar implements ComponentInterface {
...
+    if (this.scrollHidden) {
+      this.setScrollHidden(false);
+    } else {
+      this.scrollHidden = false;
+    }
+
+    if (this.hideOnScrollActive && this.containerEl) {
+      const containerClass = this.position === 'top' ? 'ion-header-hide-on-scroll' : 'ion-footer-hide-on-scroll';
+      this.containerEl.classList.remove(containerClass);
+      this.containerEl.removeAttribute(Toolbar.TOOLBAR_HIDE_COUNT_ATTR);
```
ShaneK, line 553: This strips `data-toolbar-hide-count` and the positioning class unconditionally, but a sibling `hideOnScroll` toolbar in the same header or footer may still be hidden and counting on it. With two stacked toolbars both hidden (count 2), removing one decrements to 1 via `setScrollHidden(false)`, then this wipes the attribute entirely, so the survivor can never get back to `newCount === 0` and the container stays desynced. Can you decrement here like the show path does and only clear the classes and attribute at 0?

core/src/components/toolbar/test/hide-on-scroll/index.html
```diff
@@ -0,0 +1,169 @@
+<!DOCTYPE html>
```
ShaneK, line 1: This adds a manual demo but no `toolbar.e2e.ts`, while `basic/`, `colors/`, and `title-placement/` each ship a spec. CONTRIBUTING.md asks for tests covering changed behavior, and the hide/show logic (the 24px threshold, refcounting, scroll compensation, the iOS collapse-precedence branch in `header.tsx`) is exactly what needs regression coverage. Can you add a spec for header hide/show, footer independence, collapse precedence, and the two-toolbars-in-one-header case? The failing screenshot CI may be tied to missing snapshots here too.

core/src/components/header/header.common.scss
```diff
@@ -13,3 +15,24 @@ ion-header {
...
+// Added by `ion-toolbar` when `hideOnScroll` is enabled. Overlays `ion-content`.
+
+ion-header.ion-header-hide-on-scroll {
+  @include globals.position(0, 0, null, 0);
+  position: absolute;
+
+  transition: transform var(--token-transition-time-300, 300ms)
+    var(--token-transition-curve-quick, cubic-bezier(0, 0, 0.2, 1));
+
+  z-index: 10;
```
ShaneK, line 30:
```suggestion
  z-index: globals.$z-index-toolbar;
```
[`footer.scss`](https://github.com/ionic-team/ionic-framework/pull/31175/files/bca677841c459d362e70fd986819201fa7ae06c8#diff-e867ed36d2ac34f224ab75d1adb5cfc81c2912644ac3ec5c4aaab001fcfca7a8R14) in this PR uses `globals.$z-index-toolbar` and `header.native.scss` does too, both resolving to `10`. This file already `@use`s globals, and sass-guidelines lists `z-index` as a global that shouldn't be hardcoded.

core/src/components/header/header.common.scss
```diff
@@ -13,3 +15,24 @@ ion-header {
...
+
+// Header Hide on Scroll
+// --------------------------------------------------
+// Added by `ion-toolbar` when `hideOnScroll` is enabled. Overlays `ion-content`.
+
+ion-header.ion-header-hide-on-scroll {
+  @include globals.position(0, 0, null, 0);
+  position: absolute;
+
+  transition: transform var(--token-transition-time-300, 300ms)
```
ShaneK, line 27: These inline the transition-token fallbacks by hand here and in `content.scss`, `footer.scss`, and `toolbar.common.scss`. The repo already wraps them as [`$ion-transition-*` Sass vars](https://github.com/ionic-team/ionic-framework/blob/bca677841c459d362e70fd986819201fa7ae06c8/core/src/foundations/ionic.vars.scss#L938) in foundations, though they live in the `ionic` globals rather than `native.globals`, which is probably why you inlined them. Could we forward them into `native.globals` and consume the vars, so the fallback values aren't hand-copied across four files?

core/src/components/toolbar/toolbar.tsx
```diff
@@ -310,6 +795,9 @@ export class Toolbar implements ComponentInterface {
             'in-toolbar': hostContext('ion-toolbar', this.el),
             [`toolbar-title-placement-${titlePlacement}`]: true,
           }),
+          'toolbar-hide-on-scroll': hideOnScroll,
+          'toolbar-scroll-hidden': scrollHidden,
+          'toolbar-hide-on-scroll-bottom': hideOnScroll && position === 'bottom',
```
ShaneK, line 800: Nothing in SCSS reads `toolbar-hide-on-scroll-bottom`, only `.toolbar-hide-on-scroll` and `.toolbar-scroll-hidden`. Is it leftover? If it's unused, dropping it also sidesteps the fact that `position` is a plain field set after first render, so it wouldn't be reflected here until `scrollHidden` next flips anyway.

ShaneK (changes requested):
This will probably need to be re-reviewed depending on what you decide about Brandy's requested change

joselrio:
This was closed due the changes on the implementation... new PR has been opened!

PR #31178, opened 2026-05-28
test(many): update controller tests to use correct import path

brandyscarney:
Many of the controller tests are using the wrong path to import the overlay controllers. This results in those tests not working in Vercel.

To reproduce the issue:

1. Open the [Action Sheet (main)](https://ionic-framework-git-main-ionic1.vercel.app/src/components/action-sheet/test/basic) preview
2. Attempt to open any Action Sheet
3. See the Action Sheet does not open and there is the following console error: `ReferenceError: actionSheetController is not defined`

To see it working:

1. Open the [Action Sheet (branch)](https://ionic-framework-git-test-vercel-controllers-ionic1.vercel.app/src/components/action-sheet/test/basic) preview
2. Attempt to open any Action Sheet
3. See the Action Sheet opens without any errors

This should be able to reproduce on any of the updated tests.

### DIFF
--- a/core/src/components/action-sheet/test/a11y/index.html
+++ b/core/src/components/action-sheet/test/a11y/index.html
@@ -11,7 +11,7 @@
     <script type="module" src="../../../../../dist/ionic/ionic.esm.js"></script>
   </head>
   <script type="module">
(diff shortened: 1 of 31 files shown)


ShaneK (approved):
Looks good to me! Great work 🚀

PR #31183, opened 2026-05-30
fix(react): type overlay hook component props

### LINKED ISSUE #28680: feat: Improved types for react useIonModal

SSDWGG:
Issue number: resolves #28680

---------

`useIonModal` and `useIonPopover` accept `componentProps` as `any`, so props passed to typed React components are not checked.

## What is the new behavior?

- Adds a generic `ReactComponent<Props = any>` helper while keeping `ReactComponentOrElement` backward-compatible.
- Adds overloads for `useIonModal` and `useIonPopover` so component props are inferred from function/class components.
- Keeps JSX element usage permissive for compatibility.
- Adds type coverage for valid props, missing required props, and unknown props.

- [ ] Yes
- [x] No

The generic defaults to `any`, and JSX element overloads keep the previous permissive behavior.

## Testing

- `npm ci` in `packages/react`
- `npm run test.spec -- --runTestsByPath src/hooks/__tests__/overlay-hook-types.spec.tsx`
- `npx tsc --noEmit --project tsconfig.json --skipLibCheck`
- `npx prettier --check src/hooks/useIonModal.ts src/hooks/useIonPopover.ts src/hooks/__tests__/overlay-hook-types.spec.tsx src/models/ReactComponentOrElement.ts`
- `npx eslint src/hooks/useIonModal.ts src/hooks/useIonPopover.ts src/hooks/__tests__/overlay-hook-types.spec.tsx src/models/ReactComponentOrElement.ts`

Note: plain `npx tsc --noEmit --project tsconfig.json` is blocked by the existing `AbortSignal` declaration conflict between DOM and `@types/node`; the same command passes with `--skipLibCheck`. ESLint exits successfully, with the repository ignore warning for `src/hooks/__tests__/overlay-hook-types.spec.tsx`.

### DIFF
--- a/packages/react/src/hooks/__tests__/overlay-hook-types.spec.tsx
+++ b/packages/react/src/hooks/__tests__/overlay-hook-types.spec.tsx
@@ -1,0 +1,42 @@
+import type React from 'react';
+
+import type { useIonModal } from '../useIonModal';
+import type { useIonPopover } from '../useIonPopover';
+
+type OverlayComponentProps = {
+  title: string;
+  count?: number;
+};
+
+declare const ModalComponent: React.FC<OverlayComponentProps>;
+declare const PopoverComponent: React.ComponentClass<OverlayComponentProps>;
+declare const modalElement: JSX.Element;
+declare const useIonModalForTypeTest: typeof useIonModal;
+declare const useIonPopoverForTypeTest: typeof useIonPopover;
+
+function expectOverlayHookTypes() {
+  useIonModalForTypeTest(ModalComponent, { title: 'Modal', count: 1 });
+  useIonPopoverForTypeTest(PopoverComponent, { title: 'Popover', count: 1 });
+
+  useIonModalForTypeTest(ModalComponent);
+  useIonPopoverForTypeTest(PopoverComponent);
+
+  useIonModalForTypeTest(modalElement, { anything: true });
+
+  // @ts-expect-error component props should match the modal component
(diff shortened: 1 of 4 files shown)


ShaneK:
Hey @SSDWGG! Thanks for picking this up, and sorry for the slow reply.

The direction is right, but we went with a different implementation, so I'm going to close this one.

The `componentProps` argument stays optional, so a component that declares required props still type checks when you pass nothing at all, which is the most common way to get this wrong. It's the same gap [the review on #29058 settled](https://github.com/ionic-team/ionic-framework/pull/29058#discussion_r1509878180). The signature also infers the props type from `componentProps`, which makes the self-referencing dismiss pattern circular, so it fails with `TS7022`. Both v9 controller examples in our docs use that pattern, so they would stop compiling.

I opened [a replacement against `major-9.0`](https://github.com/ionic-team/ionic-framework/pull/31362), which is also where this needs to go rather than `main`, as we [concluded on #29158](https://github.com/ionic-team/ionic-framework/pull/29158), though this may not actually be able to be merged until v10 due to how close we are to v9 and it being a breaking change. You're a co-author on the new PR along with @aeharding and @liamdebeasi, since it builds on all three attempts at solving this issue. Thanks again for your contribution!

PR #31186, opened 2026-06-02
feat(gallery): support CSS variables being passed to gap

brandyscarney:
Issue number: internal

---------

The Gallery component currently only supports CSS length strings and numbers being passed to `gap`.

## What is the new behavior?
- Adds support for CSS variables being passed to `gap` as a string.
- Adds spec & e2e tests verifying the behavior.

- [ ] Yes
- [x] No

### DIFF
--- a/core/src/components/gallery/test/basic/gallery.e2e.ts
+++ b/core/src/components/gallery/test/basic/gallery.e2e.ts
@@ -213,5 +213,28 @@ configs({ directions: ['ltr'], modes: ['md'] }).forEach(({ config, screenshot, t
           .toBe(expectedGap[breakpoint.name]);
       });
     });
+
+    test('should resolve the gap CSS variable fallback when the variable is not defined', async ({ page }) => {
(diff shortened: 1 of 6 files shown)


ShaneK (approved):
Looks good to me, awesome work 🚀

core/src/components/gallery/test/basic/gallery.e2e.ts
thetaPC, line 1: A test to verify that the fallback is used when the CSS variable is empty would be beneficial. Not sure if we need to add it to `layout/gallery.e2e.ts`.

thetaPC (approved):
LGTM, with a minor suggestion

core/src/components/gallery/test/basic/gallery.e2e.ts
brandyscarney: https://github.com/ionic-team/ionic-framework/pull/31186/changes/caf9565923007cbbf77386d12aa5069f9d4a2e8e

PR #31189, opened 2026-06-03
feat(grid, row, col): add recipe and tokens

thetaPC:

---------

`ion-grid`, `ion-row`, and `ion-col` do not fragment styles based on themes.

## What is the new behavior?

- Migrates `ion-grid`, `ion-row`, and `ion-col` to Modular Ionic.
- Adds `IonGridRecipe`, `IonColRecipe`, and `IonRowRecipe`, plus per-theme token defaults in the ios, md, and ionic theme files.
- Reworks grid and column padding and the fixed-grid width into the per-breakpoint, per-side `--ion-grid-breakpoint-*` / `--ion-col-breakpoint-*` naming, and adds `--ion-row-gap` for row spacing.
- Makes `ion-col` spans adjustable: `size`, `offset`, and `order` are written as inline custom properties (`--internal-col-span`, `--internal-col-margin`, `order`) so any `--ion-grid-columns` value resolves correctly, removing the build-time 12-column cap.
- Renames the internal classes to `col-size`, `col-auto`, and `col-offset`, and adds `IonColProperty` / `IonColStyle` types.

- [x] Yes
- [ ] No

This PR introduces breaking changes to how `ion-grid`, `ion-row`, `ion-col` are styled.

Migration Path:

1. **Grid padding**: was a single value per breakpoint, now per-side:

|-----|---|---|
| `--ion-grid-padding-{bp}` | `IonGrid.breakpoint.{bp}.padding.{top\|end\|bottom\|start}` | `--ion-grid-breakpoint-{bp}-padding-{top\|end\|bottom\|start}` |

2. **Grid fixed width**

|-----|---|---|
| `--ion-grid-width-{bp}` | `IonGrid.breakpoint.{bp}.width` | `--ion-grid-breakpoint-{bp}-width` |

3. **Column padding**: renamed into the `col` namespace and now per-side:

|-----|---|---|
| `--ion-grid-column-padding-{bp}` | `IonCol.breakpoint.{bp}.padding.{top\|end\|bottom\|start}` | `--ion-col-breakpoint-{bp}-padding-{top\|end\|bottom\|start}` |

4. **Theme classes**: Remove any instances that target the theme classes: ion-grid.md, ion-grid.ios, ion-row.md, ion-row.ios, ion-col.md, ion-col.ios

## Other information

Previews
- [Grid: basic](https://ionic-framework-git-fw-6868-ionic1.vercel.app/src/components/grid/test/basic/)
- [Grid: fixed](https://ionic-framework-git-fw-6868-ionic1.vercel.app/src/components/grid/test/fixed/)
- [Grid: offsets](https://ionic-framework-git-fw-6868-ionic1.vercel.app/src/components/grid/test/offsets/)
- [Grid: padding](https://ionic-framework-git-fw-6868-ionic1.vercel.app/src/components/grid/test/padding/)
- [Grid: sizes](https://ionic-framework-git-fw-6868-ionic1.vercel.app/src/components/grid/test/sizes/)


core/src/components/col/col.scss
```diff
@@ -1,27 +1,43 @@
...
+   *
+   * @prop --ion-col-breakpoint-xl-padding-top: Top padding on xl screens
+   * @prop --ion-col-breakpoint-xl-padding-end: End padding on xl screens
+   * @prop --ion-col-breakpoint-xl-padding-bottom: Bottom padding on xl screens
+   * @prop --ion-col-breakpoint-xl-padding-start: Start padding on xl screens
    */
-  --col-unit-size: calc(
-    (100% - (var(--ion-grid-columns, 12) - 1) * var(--ion-grid-gap, 0px)) / var(--ion-grid-columns, 12)
+  --internal-col-unit-size: calc(
+    (100% - (var(--ion-grid-columns, 12) - 1) * var(--ion-row-gap, 0px)) / var(--ion-grid-columns, 12)
```
thetaPC, line 36: I changed `--ion-grid-gap` to `--ion-row-gap` since the value is adding the gap on row, not gap. No need to add it to the breaking changes since this variable was introduced in `next`.

core/src/components/col/col.scss
```diff
@@ -34,48 +50,29 @@
   overflow: auto;
 }
 
-:host(.ion-grid-col-auto) {
+:host(.col-auto) {
```
thetaPC, line 53: No need to add it to the breaking changes since this class was introduced in `next`.

core/src/components/col/col.scss
```diff
@@ -34,48 +50,29 @@
...
-    --ion-grid-col-span: #{$i};
-  }
-
-  :host(.ion-grid-order-col--#{$i}) {
-    order: #{$i};
-  }
-
-  :host(.ion-grid-offset-col--#{$i}) {
-    --ion-grid-col-margin: #{$i};
-  }
```
thetaPC, line 80: Previously, a `@for 1 through 12` loop generated a fixed set of per-column classes (`ion-grid-col--1…--12`, plus order/offset variants), each hardcoding the span/order/offset multiplier. Because the loop ran at Sass compile time, the supported column count was capped at 12 — yet the width math already divided by the runtime `--ion-grid-columns`, so a grid with more than 12 columns (e.g. `--ion-grid-columns: 16` + `size="16"`) produced a class with no matching rule, leaving `--internal-col-span` unset and invalidating the `calc()`. `col.tsx` now writes the span, offset, and order as inline custom properties from the parsed attribute value, so any column count composes with the `--ion-grid-columns` divisor without a build-time class cap. The generated classes (and `$grid-col-number`) are no longer needed.

core/src/components/col/col.tsx
```diff
@@ -3,22 +3,22 @@ import { Component, Element, Host, Listen, Prop, forceUpdate, h } from '@stencil
...
  * @virtualProp {"ios" | "md"} mode - The mode determines the platform behaviors of the component.
- * @virtualProp {"ios" | "md" | "ionic"} theme - The theme determines the visual appearance of the component.
  */
 @Component({
   tag: 'ion-col',
   styleUrl: 'col.scss',
   shadow: true,
 })
 export class Col implements ComponentInterface {
+  private resizeTimeout: ReturnType<typeof setTimeout> | null = null;
```
thetaPC, line 28: Added a debounced window-resize handler so the column re-renders when the viewport crosses a responsive breakpoint — Stencil won't re-render on media-query changes on its own, so `getColumns()`'s `matchBreakpoint()` result would otherwise go stale. `resizeTimeout` holds the debounce handle so rapid resize events collapse into a single `forceUpdate`, and so the pending timer can be cleared in `disconnectedCallback`. This mirrors the established pattern in other components such as ion-content, which uses the same `resizeTimeout` + debounced `forceUpdate` approach on window resize.

core/src/components/grid/test/offsets-pull-push/index.html
```diff
@@ -57,43 +58,8 @@ <h2 class="ion-padding-start">Push and Pull</h2>
             </ion-col>
           </ion-row>
         </ion-grid>
-
-        <h2 class="ion-padding-start">Offset</h2>
```
thetaPC, line 61: These were removed because the same copy exists in the basic test.

core/src/components/grid/grid.interface.ts
```diff
@@ -0,0 +1,16 @@
...
+
+export type IonGridRecipe = {
+  breakpoint?: {
+    [K in IonGridBreakpoint]?: {
+      padding?: IonPadding;
+      width?: string;
+    };
+  };
+
+  columns?: number;
```
thetaPC, line 11: Should this be a prop instead to match `ion-gallery`? I think we should to be consistent. If so, then a new ticket should be created to do the overhaul.

core/src/components/row/row.interface.ts
```diff
@@ -0,0 +1,3 @@
+export type IonRowRecipe = {
+  gap?: string;
```
thetaPC, line 2: Should this be a prop instead to match `ion-gallery`? I think we should to be consistent. If so, then a new ticket should be created to do the overhaul.

packages/angular/src/directives/proxies.ts
```diff
@@ -572,14 +572,14 @@ export declare interface IonChip extends Components.IonChip {}
 
 
 @ProxyCmp({
-  inputs: ['mode', 'offset', 'offsetLg', 'offsetMd', 'offsetSm', 'offsetXl', 'offsetXs', 'order', 'orderLg', 'orderMd', 'orderSm', 'orderXl', 'orderXs', 'pull', 'pullLg', 'pullMd', 'pullSm', 'pullXl', 'pullXs', 'push', 'pushLg', 'pushMd', 'pushSm', 'pushXl', 'pushXs', 'size', 'sizeLg', 'sizeMd', 'sizeSm', 'sizeXl', 'sizeXs', 'theme']
+  inputs: ['mode', 'offset', 'offsetLg', 'offsetMd', 'offsetSm', 'offsetXl', 'offsetXs', 'order', 'orderLg', 'orderMd', 'orderSm', 'orderXl', 'orderXs', 'pull', 'pullLg', 'pullMd', 'pullSm', 'pullXl', 'pullXs', 'push', 'pushLg', 'pushMd', 'pushSm', 'pushXl', 'pushXs', 'size', 'sizeLg', 'sizeMd', 'sizeSm', 'sizeXl', 'sizeXs']
```
thetaPC, line 575: I think we should convert `order` and `offset` to accept an object like how `columns` does with `ion-gallery`. This would allow us to set the breakpoints through one value instead. Thoughts? I can create a ticket if we agree.

core/src/components/action-sheet/test/translucent/action-sheet.e2e.ts-snapshots/action-sheet-translucent-ios-ltr-Mobile-Chrome-linux.png
thetaPC, line 1: I replaced `ion-grid` to native grid so we don't have to always update snapshots if `ion-grid` changed.

core/src/components/action-sheet/test/translucent/index.html
```diff
@@ -14,10 +14,16 @@
     <script type="module" src="../../../../../dist/ionic/ionic.esm.js"></script>
 
     <style>
-      f {
-        display: block;
-        width: 100%;
+      .color-grid {
```
thetaPC, line 17: I replaced `ion-grid` to native grid so we don't have to always update snapshots if `ion-grid` changed.

core/src/components/fab/test/translucent/index.html
```diff
@@ -23,18 +23,16 @@
...
-            <ion-col size="6"><f class="green"></f></ion-col>
-            <ion-col size="6"><f class="blue"></f></ion-col>
-            <ion-col size="6"><f class="yellow"></f></ion-col>
-            <ion-col size="6"><f class="pink"></f></ion-col>
-            <ion-col size="6"><f class="purple"></f></ion-col>
-            <ion-col size="6"><f class="black"></f></ion-col>
-            <ion-col size="6"><f class="orange"></f></ion-col>
-          </ion-row>
-        </ion-grid>
+        <div class="color-grid">
```
thetaPC, line 26: I replaced `ion-grid` to native grid so we don't have to always update snapshots if `ion-grid` changed.

core/src/components/loading/test/basic/loading.e2e.ts-snapshots/loading-basic-diff-ios-ltr-Mobile-Firefox-linux.png
thetaPC, line 1: I replaced `ion-grid` to native grid so we don't have to always update snapshots if `ion-grid` changed.

core/src/components/loading/test/basic/index.html
```diff
@@ -30,13 +30,19 @@
...
       }
 
-      f {
-        display: block;
-        background: blue;
+      /* A simple multi-colored backdrop so the loading overlay (and its
+         translucent variant) is visible over varied colors. Uses native CSS
+         grid instead of ion-grid so this page's snapshots don't churn when
+         grid/row/col change. */
+      .color-grid {
```
thetaPC, line 33: I replaced `ion-grid` to native grid so we don't have to always update snapshots if `ion-grid` changed.

core/src/components/row/row.scss
```diff
@@ -1,14 +1,14 @@
...
+
   display: flex;
 
   flex-wrap: wrap;
 
-  /**
-   * @prop --ion-grid-gap: The gap between grid items
-   */
-  gap: var(--ion-grid-gap, 0px);
+  gap: var(--ion-row-gap);
```
ShaneK, line 13: Just calling out that the fallback was lost here, was this intentional?

core/src/components/col/col.tsx
```diff
@@ -203,20 +215,36 @@ export class Col implements ComponentInterface {
 
   @Listen('resize', { target: 'window' })
   onResize() {
-    forceUpdate(this);
+    if (this.resizeTimeout) {
+      clearTimeout(this.resizeTimeout);
+      this.resizeTimeout = null;
+    }
+
+    this.resizeTimeout = setTimeout(() => {
```
ShaneK, line 231: Could you add a short comment on why `onResize` debounces at 100ms now? It matches the `content.tsx` resize pattern, but the bare `100` reads as a magic number we may not want to lose during later refactoring

ShaneK (approved):
Only a couple of nits, looks awesome to me! Great work 🚀

core/src/components/row/row.scss
thetaPC: Yes, that was intentional. Since gap isn't used inside a `calc()`, an unset `--ion-row-gap` simply falls back to the property's initial value (`normal`, which resolves to `0`) rather than invalidating the declaration. It also matches how we handle the custom variables in the other migrated components (fallbacks only where a `calc()` would otherwise be invalidated). Hence why we have [fallbacks](https://github.com/ionic-team/ionic-framework/blob/b96af96ce9c5d3f53364439b63ed38ab99e521d7/core/src/components/col/col.scss#L35) for `var(--ion-row-gap)` in `col.scss`.

core/src/components/col/col.tsx
thetaPC: https://github.com/ionic-team/ionic-framework/pull/31189/changes/cd6678a19a9e15fb68af2b6f92c584b8f1440c50

core/src/components/grid/grid.interface.ts
brandyscarney: Yes. Grid needs a complete overhaul.

core/src/components/row/row.interface.ts
brandyscarney: Yes. Grid needs a complete overhaul.

packages/angular/src/directives/proxies.ts
brandyscarney: Yes. Grid needs a complete overhaul.

core/src/components/grid/test/basic/index.html
```diff
@@ -205,16 +224,55 @@
...
-      <style>
-        .grid-demo ion-col div {
-          background-color: #f7f7f7;
-          border: 1px solid #ddd;
-          font-size: 0.8em;
-          padding: 10px 5px;
-        }
-      </style>
+        <h2>Custom Column Count</h2>
+        <ion-grid id="custom-columns" style="--ion-grid-columns: 16">
```
brandyscarney, line 229: Why did we add this example but not an e2e test for it?

core/src/components/grid/test/basic/grid.e2e.ts
```diff
@@ -2,16 +2,26 @@ import { expect } from '@playwright/test';
...
 
 /**
- * ion-grid does not have different styling per-mode
+ * This behavior does not vary across modes.
  */
 configs({ modes: ['md'] }).forEach(({ title, screenshot, config }) => {
   test.describe(title('grid: basic'), () => {
-    test('should not have visual regressions', async ({ page }) => {
+    test.beforeEach(async ({ page }) => {
       await page.goto(`/src/components/grid/test/basic`, config);
```
brandyscarney: Shouldn't we be converting the e2e tests to use `page.setContent`?

core/src/components/col/test/col.spec.ts
```diff
@@ -0,0 +1,118 @@
...
+    const page = await newSpecPage({
+      components: [Col],
+      html: `<ion-col size="auto"></ion-col>`,
+    });
+
+    const col = page.body.querySelector('ion-col')!;
+    expect(col.classList.contains('col-auto')).toBe(true);
+  });
+
+  it('applies no sizing class for a value less size attribute', async () => {
```
brandyscarney:
```suggestion
  it('applies no sizing class for a valueless size attribute', async () => {
```

core/src/components/col/test/col.spec.ts
```diff
@@ -0,0 +1,118 @@
...
+      html: `<ion-col size="banana"></ion-col>`,
+    });
+
+    const col = page.body.querySelector('ion-col')!;
+    const sizeClasses = Array.from(col.classList).filter((c) => c.startsWith('col-'));
+    expect(sizeClasses).toEqual([]);
+  });
+});
+
+describe('ion-col: deprecated push/pull props', () => {
```
brandyscarney, line 72:
```suggestion
TODO(FW-7557): Remove this when the push/pull props are removed.
describe('ion-col: deprecated push/pull props', () => {
```

core/src/components/col/col.interface.ts
```diff
@@ -0,0 +1,22 @@
...
+
+// TODO(FW-7285): Replace with global breakpoints
+export const ION_COL_BREAKPOINTS = ION_GRID_BREAKPOINTS;
+export type IonColBreakpoint = (typeof ION_COL_BREAKPOINTS)[number];
+
+export type IonColProperty = 'size' | 'order' | 'offset';
+
+export type IonColStyle = {
+  '--internal-col-margin'?: string;
+  '--internal-col-span'?: string;
```
brandyscarney, line 20: Why is the css variable a property?

core/src/components/col/col.scss
brandyscarney: This was [added for Mobile UI](https://github.com/ionic-team/ionic-framework/pull/30658) so we likely need a follow-up ticket for them to update it if they are using it.

core/src/components/grid/test/offsets-pull-push/grid.e2e.ts
```diff
@@ -11,7 +12,7 @@ configs({ modes: ['md'] }).forEach(({ title, screenshot, config }) => {
 
       await page.setIonViewport();
 
-      await expect(page).toHaveScreenshot(screenshot(`grid-offsets`));
+      await expect(page).toHaveScreenshot(screenshot('grid-offsets'));
```
brandyscarney: Should this be renamed since a different folder named `offsets` exists?


```suggestion
      await expect(page).toHaveScreenshot(screenshot('grid-pull-push'));
```

core/src/components/grid/test/sizes/grid.e2e.ts-snapshots/grid-sizes-md-ltr-Mobile-Chrome-linux.png
brandyscarney, line 1: Ideally these would be broken up into separate tests:

- 10 column layout
- responsive sm
- size auto
- breakpoint sizes

We could probably remove some screenshot tests and just check values.

Note: I am fine with making this a task on the follow-up overhaul ticket.

core/src/components/grid/test/offsets/grid.e2e.ts-snapshots/grid-offsets-md-ltr-Mobile-Firefox-linux.png
brandyscarney, line 1: Ideally these would be broken up into separate tests:

- offset 1
- order
- offset 2
- dynamic offset

We could probably remove some screenshot tests and just check values.

Note: I am fine with making this a task on the follow-up overhaul ticket.

core/src/components/grid/test/padding/grid.e2e.ts-snapshots/grid-padding-md-ltr-Mobile-Chrome-linux.png
brandyscarney, line 1: Ideally these would be broken up into separate tests:

- no padding
- no padding / nested
- default padding
- custom grid padding
- custom column padding

We could probably remove some screenshot tests and just check values.

Note: I am fine with making this a task on the follow-up overhaul ticket.

BREAKING.md
```diff
@@ -141,9 +162,32 @@ Remove any instances that target the theme classes: `ion-content.md`, `ion-conte
 
 <h4 id="version-9x-grid">Grid</h4>
 
-- The properties `pull` and `push` have been deprecated and no longer work. A similar look can be achieved with the newly added property `order`.
+The following breaking changes apply to `ion-grid`:
+
+1. `--ion-grid-padding-*` CSS variables have been replaced with per-side, per-breakpoint tokens. <sup>[1](#version-9x-grid-padding-variables)</sup>
+2. `--ion-grid-width-*` CSS variables for the fixed grid have been replaced with per-breakpoint tokens. <sup>[2](#version-9x-grid-fixed-width-variables)</sup>
+3. The `push` and `pull` properties have been deprecated and no longer work. <sup>[3](#version-9x-grid-deprecated-push-and-pull-properties)</sup>
```
brandyscarney: "no longer work" - is this accurate? It isn't "deprecated" if they don't work.

BREAKING.md
```diff
@@ -141,9 +162,32 @@ Remove any instances that target the theme classes: `ion-content.md`, `ion-conte
...
+
+<h5 id="version-9x-grid-fixed-width-variables">Fixed width variables</h5>
+
+| Old (8.x) | New token (global) | New CSS variable (component-specific) |
+|---|---|---|
+| `--ion-grid-width-{bp}` | `IonGrid.breakpoint.{bp}.width` | `--ion-grid-breakpoint-{bp}-width` |
+
+<h5 id="version-9x-grid-deprecated-push-and-pull-properties">Deprecated <code>push</code> and <code>pull</code> properties</h5>
+
+The `push` and `pull` properties have been deprecated and no longer work. A similar look can be achieved with the newly added property `order`.
```
brandyscarney: Same concern here with "deprecated" and "no longer work"

brandyscarney (approved):
Looks good! A few non-blocking questions and comments.

core/src/components/grid/test/basic/grid.e2e.ts
thetaPC: No? I didn't see the need to use `setContent` for this test. The test page covers different scenarios that are good to capture. I don't see the need to duplicate it through `setContent`.

core/src/components/col/test/col.spec.ts
thetaPC: https://github.com/ionic-team/ionic-framework/pull/31189/changes/4b31f33b657e57d8c7c78f72af72f18eb2466ee8

core/src/components/col/test/col.spec.ts
thetaPC: https://github.com/ionic-team/ionic-framework/pull/31189/changes/7fb5865cafdd0f27f266fd6560f4a4243e2bde4d

core/src/components/grid/test/basic/index.html
thetaPC: The e2e was [added](https://github.com/ionic-team/ionic-framework/blob/7fb5865cafdd0f27f266fd6560f4a4243e2bde4d/core/src/components/grid/test/basic/grid.e2e.ts#L21-L25) and can be found under the basic folder.

core/src/components/col/col.interface.ts
thetaPC: It's the shape of the host's inline `style` object. Setting a CSS variable from TSX means adding it to that object keyed by its name (`style={{ '--internal-col-span': '3' }}`), so the variable names have to be properties here. I gave it explicit keys rather than a generic `{ [key: string]: string }` so the type documents exactly which variables col writes and catches typos. col.scss reads `--internal-col-span` / `--internal-col-margin` in its `calc()`; `order` is a normal inline style alongside them.

core/src/components/col/col.scss
thetaPC: FW-7563 created!

core/src/components/grid/test/offsets-pull-push/grid.e2e.ts
thetaPC: You're not wrong but since it's going to be deleted, I didn't do it in the first place: https://github.com/ionic-team/ionic-framework/pull/31189/changes/378fb7614c82bae2b4ae8ec9eb343aaaf9f80349

BREAKING.md
thetaPC: I just copied it from the previous version but I've updated it: https://github.com/ionic-team/ionic-framework/pull/31189/changes/c881b5a500acbd05831b044d18d7865f263e4101

BREAKING.md
thetaPC: https://github.com/ionic-team/ionic-framework/pull/31189/changes/c881b5a500acbd05831b044d18d7865f263e4101

core/src/components/grid/test/sizes/grid.e2e.ts-snapshots/grid-sizes-md-ltr-Mobile-Chrome-linux.png
thetaPC: https://github.com/ionic-team/ionic-framework/pull/31189/changes/789fb43d42016abbea1daf0a1fca2d2aa92f6d6e and https://github.com/ionic-team/ionic-framework/pull/31189/changes/6703be00c6e542c4a42fbce906f760cdfc11cb7d

core/src/components/grid/test/offsets/grid.e2e.ts-snapshots/grid-offsets-md-ltr-Mobile-Firefox-linux.png
thetaPC: https://github.com/ionic-team/ionic-framework/pull/31189/changes/b215bb30849162512378c81ce9a054ecea5639b4 and https://github.com/ionic-team/ionic-framework/pull/31189/changes/e6869a88d2f162bb98a70646dfce7c75049506ed

core/src/components/grid/test/padding/grid.e2e.ts-snapshots/grid-padding-md-ltr-Mobile-Chrome-linux.png
thetaPC: https://github.com/ionic-team/ionic-framework/pull/31189/changes/2a796a1af8ae09fb19fe20c00915a78861cfb4af

core/src/components/grid/test/basic/grid.e2e.ts
brandyscarney: Sorry I meant instead of using `page.goto` we have been migrating away from full page screenshots and using `setContent`.

BREAKING.md
brandyscarney: We need to fix it if they don't work at all, that's not deprecation. If push or pull is set it should change to offset.

brandyscarney (approved):

core/src/components/grid/test/basic/grid.e2e.ts
thetaPC: https://github.com/ionic-team/ionic-framework/pull/31189/changes/52c3618f0b6db593a510d90a34406ac63abc251a

core/src/components/grid/test/offsets/grid.e2e.ts-snapshots/grid-offsets-md-ltr-Mobile-Firefox-linux.png
thetaPC: https://github.com/ionic-team/ionic-framework/pull/31189/changes/5df838d1c02b5cc11600a7b012aa2ed2968347ec

core/src/components/grid/test/padding/grid.e2e.ts-snapshots/grid-padding-md-ltr-Mobile-Chrome-linux.png
thetaPC: https://github.com/ionic-team/ionic-framework/pull/31189/changes/5df838d1c02b5cc11600a7b012aa2ed2968347ec

core/src/components/grid/test/sizes/grid.e2e.ts-snapshots/grid-sizes-md-ltr-Mobile-Chrome-linux.png
thetaPC: https://github.com/ionic-team/ionic-framework/pull/31189/changes/6ccf91352159f84ebdf838e0c5de9d7581fc7709

BREAKING.md
thetaPC: FW-7565 created!

PR #31194, opened 2026-06-04
fix(range): initialize ratios before first render

### LINKED ISSUE #31026: bug: `ion-range`: value attribute doesn't render correctly in a modal
### Prerequisites

### Ionic Framework Version

v8.x

### Current Behavior

The value attribute isn't causing the knobs to be placed where they should be if the `ion-range` is in a modal

### Expected Behavior

I would expect that the lower knob is placed at range 20 and the upper knob is placed at range 80

### Steps to Reproduce

1. Add an ion range to a modal
2. Assign the value attribute to something that you expect

### Code Reproduction URL

https://stackblitz.com/edit/trarread-zarwym83?file=src%2Fapp%2Fexample.component.ts

### Ionic Info

Ionic:

   Ionic CLI                     : 7.2.1 (C:\Users\test_user\AppData\Roaming\npm\node_modules\@ionic\cli)
   Ionic Framework               : @ionic/angular 8.8.1
   @angular-devkit/build-angular : not installed
   @angular-devkit/schematics    : 20.3.21
   @angular/cli                  : 21.2.3

Utility:

   native-run  : 2.0.3

System:

   NodeJS : v25.8.1 (C:\Program Files\nodejs\node.exe)
   npm    : 11.11.0
   OS     : Windows 10

### Additional Information

_No response_

puneetdixit200:
Issue number: resolves #31026

`ion-range` can render from its default knob ratios on the first render when it is created with an initial object value inside a modal. The issue report's dual-knob range starts with `{ lower: 20, upper: 80 }`, but both knobs are initially painted at the default positions.

## What is the new behavior?

- Initialize the range ratios during `componentWillLoad` after min/max/step are normalized.
- Preserve the existing `componentDidLoad` ratio update for layout-dependent work.
- Add a modal e2e regression fixture that presents a dual-knob range and verifies the lower and upper knob handle positions are `20%` and `80%`.

- [ ] Yes
- [x] No

## Other information

Testing:
- `npm run build.css`
- `npx stencil build --dev`
- `git diff --check`
- `npx prettier --check src/components/range/range.tsx src/components/range/test/modal/index.html src/components/range/test/modal/range.e2e.ts`
- Manual Playwright verification with system Chrome against `src/components/range/test/modal?ionic:_testing=true&ionic:mode=ios`, observed `{"lower":"20%","upper":"80%","a":"20%","b":"80%","value":{"lower":20,"upper":80}}`

I also tried `npx playwright test src/components/range/test/modal/range.e2e.ts --project="Mobile Chrome"`, but this machine is missing Playwright's managed Chromium binary (`chromium_headless_shell-1217`), so the test could not start here.

### DIFF
--- a/core/src/components/range/range.tsx
+++ b/core/src/components/range/range.tsx
@@ -401,6 +401,8 @@ export class Range implements ComponentInterface {
     this.min = isSafeNumber(this.min) ? this.min : 0;
     this.max = isSafeNumber(this.max) ? this.max : 100;
     this.step = isSafeNumber(this.step) ? this.step : 1;
+
+    this.updateRatio();
   }
 
   componentDidLoad() {
--- a/core/src/components/range/test/modal/index.html
+++ b/core/src/components/range/test/modal/index.html
@@ -1,0 +1,50 @@
+<!DOCTYPE html>
+<html lang="en" dir="ltr">
+  <head>
+    <meta charset="UTF-8" />
+    <title>Range - Modal</title>
+    <meta
+      name="viewport"
+      content="width=device-width, initial-scale=1.0, minimum-scale=1.0, maximum-scale=1.0, user-scalable=no"
+    />
+    <link href="../../../../../css/ionic.bundle.css" rel="stylesheet" />
+    <script src="../../../../../scripts/testing/scripts.js"></script>
+    <script nomodule src="../../../../../dist/ionic/ionic.js"></script>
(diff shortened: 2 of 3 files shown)


brandyscarney:
Thank you for the PR! I am going to close this because I have fixed the issue in this PR: https://github.com/ionic-team/ionic-framework/pull/31365

However, I gave you co-author credit on the commit: https://github.com/ionic-team/ionic-framework/commit/7d64d62a31d59078560575ad9c4fc67a8e49e524

PR #31195, opened 2026-06-05
feat(gallery): add new required gallery-item component for child items

brandyscarney:
Issue number: internal

---------

The Gallery component queries all direct children and uses them to build both the uniform and masonry layouts. This causes issues when a wrapper element exists between the gallery and the gallery items. For example:

```html
<ion-gallery>
  <div class="some-wrapper">
    <img src="/src/components/gallery/test/assets/01.png" alt="One" />
    <img src="/src/components/gallery/test/assets/02.png" alt="Two" />
    <img src="/src/components/gallery/test/assets/03.png" alt="Three" />
    <img src="/src/components/gallery/test/assets/04.png" alt="Four" />
  </div>
</ion-gallery>
```

In this scenario, the gallery treats the wrapper as a single item and does not correctly lay out the images contained within it.

## What is the new behavior?

Several approaches were considered:

1. Add a class or attribute to wrapper elements that the Gallery could use to locate its items.
2. Query for items using a specific class or attribute when present, and fall back to querying direct children otherwise.
3. Introduce a new `ion-gallery-item` component that encapsulates all item-related styling and can contain arbitrary content (img, div, etc.).

We ultimately chose option 3 and introduced a new `ion-gallery-item` component. The `ion-gallery` component now queries for `ion-gallery-item` elements instead of relying on direct children. 

As such, this PR does the following:
- Adds a new `ion-gallery-item` component.
- Updates `ion-gallery` to query for `ion-gallery-item` elements.
- Updates existing E2E and spec tests to use the new structure.
- Adds new E2E and spec tests for `ion-gallery-item`.
- Adds an E2E test verifying that wrapper elements around gallery items are effectively transparent to gallery layout calculations.

- [ ] Yes
- [x] No

## Other information

Dev build: `8.8.9-dev.11781201980.1b6e8398`

Previews:
- [Basic](https://ionic-framework-git-fw-7301-ionic1.vercel.app/src/components/gallery/test/basic)
- [Layout](https://ionic-framework-git-fw-7301-ionic1.vercel.app/src/components/gallery/test/layout)
- [Wrapper](https://ionic-framework-git-fw-7301-ionic1.vercel.app/src/components/gallery/test/wrapper)


core/src/components/gallery/test/layout/gallery.e2e.ts-snapshots/gallery-masonry-best-fit-divs-dynamically-appended-md-ltr-Mobile-Chrome-linux.png
brandyscarney, line 1: This change is actually more accurate. In both images the blue rectangle I added is 16px tall. You can see on the left there is a half pixel above and below it.

Image

core/src/components/gallery/test/utils.ts
```diff
@@ -3,57 +3,56 @@ export const sharedStyles = `
     width: 343px;
   }
 
-  div {
+  ion-gallery-item {
     color: #fff;
-    height: 150px;
```
brandyscarney, line 8: I moved this to be on the items themselves because I can't apply them to all items or the ones containing `img` will also get an explicit height.

core/src/components/gallery/test/layout/gallery.e2e.ts
```diff
@@ -58,7 +58,7 @@ configs({ directions: ['ltr'], modes: ['md'] }).forEach(({ config, screenshot, t
           await expect(gallery).toHaveScreenshot(screenshot(`gallery-${layout}${orderSuffix}-divs-same-height`));
```
brandyscarney, line 59: I didn't change the name of this to replace `divs` with `items` so you could see there are not diffs on these but I can update them after it's been reviewed if desired.

core/src/components/gallery/gallery.tsx
```diff
@@ -450,20 +472,60 @@ export class Gallery implements ComponentInterface {
...
+      // components. Collapse the wrapper's box so the items participate in the
+      // gallery grid.
+      const nestedItems = Array.from(child.querySelectorAll<HTMLIonGalleryItemElement>(GALLERY_ITEM_SELECTOR));
+
+      if (nestedItems.length === 0) {
+        this.warnInvalidItems();
+        return;
+      }
+
+      (child as HTMLElement).style.display = 'contents';
```
ShaneK: `getItems()` sets `display: contents` on a wrapper when it contains items, but nothing ever clears it. If that wrapper later loses its `ion-gallery-item` children (removed or moved out), it falls into the `nestedItems.length === 0` branch, which just warns and returns, so the stale `display: contents` stays on the now-empty wrapper and its remaining content keeps rendering with no box. Since `getItems()` runs on every slotchange, resize, and masonry pass, this is reachable in a dynamic gallery. There's no equivalent of `clearItemStyles()` for wrapper display. I think the empty-wrapper branch needs to reset `child.style.display`.

core/src/components/gallery/gallery.tsx
```diff
@@ -450,20 +472,60 @@ export class Gallery implements ComponentInterface {
...
+      // Standard path: <ion-gallery-item> is a direct child of <ion-gallery>.
+      if (child.matches(GALLERY_ITEM_SELECTOR)) {
+        items.push(child as HTMLIonGalleryItemElement);
+        return;
+      }
+
+      // Compatibility path: a wrapper element may contain <ion-gallery-item>
+      // components. Collapse the wrapper's box so the items participate in the
+      // gallery grid.
+      const nestedItems = Array.from(child.querySelectorAll<HTMLIonGalleryItemElement>(GALLERY_ITEM_SELECTOR));
```
ShaneK: `querySelectorAll('ion-gallery-item')` matches descendants at any depth, so a wrapper that contains a nested `ion-gallery` will pull the inner gallery's items into this gallery's grid. Both galleries then write conflicting `gridColumn`/`gridRowStart` onto the same elements and both observe them via `ResizeObserver`. I think this should filter to items whose nearest `ion-gallery` ancestor is `this.el`.

core/src/components/gallery-item/gallery-item.tsx
```diff
@@ -0,0 +1,107 @@
...
+      this.el
+    );
+    this.hasWarnedInvalidParent = true;
+  }
+
+  /**
+   * Watch the parent gallery's class list so the item can react to layout
+   * changes (the gallery reflects its layout as a `gallery-layout-*` class).
+   */
+  private watchGalleryLayoutClasses() {
```
ShaneK: Every other core child that reflects parent state as an `in-*` class computes it inline in `render()` via `hostContext()`, for example `segment-button.tsx:169` does `'in-segment': hostContext('ion-segment', this.el)`, and the same pattern shows up in `item`, `badge`, `checkbox`/`radio`/`range`. `hostContext()` already runs a light-DOM `closest()` against class selectors, so `'in-gallery-layout-uniform': hostContext('ion-gallery.gallery-layout-uniform', this.el)` would read the gallery's reflected class with no `MutationObserver`, no `@State`, and no `componentDidLoad`/`disconnectedCallback`. Is there a reason this hand-rolls the observer instead? The only thing the inline approach doesn't cover is reacting to a runtime `layout` change, and the gallery already has `@Watch('layout')`, so it could emit an event the item subscribes to, the way `ion-accordion` subscribes to the group's `ionValueChange`. That matches the existing convention and drops a per-item observer.

core/src/components/gallery-item/gallery-item.tsx
```diff
@@ -0,0 +1,107 @@
...
+  @Element() el!: HTMLIonGalleryItemElement;
+
+  /**
+   * The layout of the parent `ion-gallery`, mirrored as a class so the item
+   * can apply layout-specific styles (e.g. a square aspect ratio in the
+   * `uniform` layout, a bottom margin in the `masonry` layout).
+   */
+  @State() galleryLayout?: 'uniform' | 'masonry';
+
+  componentWillLoad() {
```
ShaneK, line 31: `galleryEl` is resolved in `componentWillLoad` and the observer is wired in `componentDidLoad`, but `disconnectedCallback` tears both down. Stencil only re-fires `connectedCallback`/`disconnectedCallback` on a DOM move, not the `componentWill`/`DidLoad` pair, so after a detach and reattach `galleryEl` stays undefined, the observer is never rebuilt, and the item silently loses its `in-gallery-layout-*` styling. Framework integrations and conditional rendering move hosts around routinely, so this is worth handling. Moving to the `hostContext` approach from the other comment would sidestep it entirely, since the class gets recomputed every render.

core/src/components/gallery/gallery.tsx
```diff
@@ -450,20 +472,60 @@ export class Gallery implements ComponentInterface {
...
   /**
-   * Return all directly slotted children of the gallery that can be grid items
-   * with inline placement styles (HTML elements and SVG elements).
+   * Return the `ion-gallery-item` elements to place in the grid. Each item is a
+   * direct grid cell. A direct child that is not an `ion-gallery-item` is
+   * treated as a pass-through wrapper (e.g. a layout `<div>`): its box is
+   * collapsed with `display: contents` so the nested items participate in the
+   * gallery grid. Children that contain no `ion-gallery-item` are ignored.
+   */
+  private getItems(): HTMLIonGalleryItemElement[] {
```
ShaneK, line 490: `getItems()` reads like a pure accessor, and the other `get*` helpers in this file are pure, but this one mutates the DOM with `display: contents` and emits a warning, and it runs on every masonry frame through `resizeMasonryGrid`. `clearMasonryStyles()` even re-applies the wrapper collapse while it's supposedly clearing. I think it'd read better split into a pure `getItems()` plus a `normalizeWrappers()` (or fold the flattening into the slotchange handler) so the mutation happens once when the structure changes. That also gives a natural home for resetting the stale `display: contents`. Up to you, though

core/src/components/gallery-item/gallery-item.tsx
```diff
@@ -0,0 +1,107 @@
...
+      attributeFilter: ['class'],
+    });
+  }
+
+  private syncLayoutClasses() {
+    const layout = this.galleryEl?.layout;
+    this.galleryLayout = layout === 'masonry' || layout === 'uniform' ? layout : undefined;
+  }
+
+  render() {
```
ShaneK, line 72: `ion-gallery-item` is `shadow: true` wrapping slotted content but exposes no shadow part and no `@part` JSDoc. Is a part intended on the host box, or is it intentionally left off?

core/src/components/gallery-item/gallery-item.tsx
```diff
@@ -0,0 +1,107 @@
...
+
+/**
+ * @virtualProp {"ios" | "md"} mode - The mode determines the platform behaviors of the component.
+ * @virtualProp {"ios" | "md" | "ionic"} theme - The theme determines the visual appearance of the component.
+ *
+ * @slot - The content placed inside of the gallery item. This is typically an
+ * `img`, but can be any element (e.g. a `figure` wrapping an image and caption).
+ */
+@Component({
+  tag: 'ion-gallery-item',
```
ShaneK, line 15: `ion-gallery-item` is a public component but only has `gallery-item.spec.ts`, no `test/basic/index.html`, and its `in-gallery-layout-*` visual styles (the square aspect ratio and the masonry margin) have no e2e or screenshot coverage, only the spec asserting the class is present. Is the intent to test it only through the parent gallery? If so that's fine, just highlighting this now

core/src/components/gallery/gallery.spec.ts
```diff
@@ -746,53 +746,78 @@ describe('gallery', () => {
...
 
-      it('should apply masonry grid placement styles to slotted SVG elements', () => {
-        const div = document.createElement('div');
-        const svg = document.createElementNS('http://www.w3.org/2000/svg', 'svg');
-        el.appendChild(div);
-        el.appendChild(svg);
+      it('should warn and ignore children that do not contain an ion-gallery-item', () => {
+        const warningSpy = jest.spyOn(logging, 'printIonWarning').mockImplementation(() => {});
+
+        el.appendChild(document.createElement('img'));
```
ShaneK: Nit: `getItems()` only sets `display: contents` on wrappers that contain items, so the `img`/`span` ignore path should leave the child's style alone. This test checks the warning and the empty array but not that the ignored child's `display` is untouched. A quick `expect(img.style.display).toBe('')` would guarantee that going forward

ShaneK (changes requested):
This is looking really cool! I just had some change requests and some nits, but I'm loving the way it's looking!

core/src/components/gallery/gallery.tsx
brandyscarney: https://github.com/ionic-team/ionic-framework/pull/31195/changes/d4032af5ab0fe1ca6a3c7a57c7e3e91fd3dd6262

core/src/components/gallery/gallery.tsx
brandyscarney: https://github.com/ionic-team/ionic-framework/pull/31195/changes/20a1e7f79504ffe1f9aacadef08c532f25800e41

core/src/components/gallery/gallery.tsx
brandyscarney: https://github.com/ionic-team/ionic-framework/pull/31195/changes/253ff44e4fb5d3fb25254010368fb35b936ccec3

core/src/components/gallery-item/gallery-item.tsx
brandyscarney: It does not have any parts because the only inner element it has is the default slot. This matches how other shadow components with no parts behave, such as [Avatar](https://github.com/ionic-team/ionic-framework/blob/next/core/src/components/avatar/avatar.tsx).

core/src/components/gallery-item/gallery-item.tsx
brandyscarney: Good catch! I actually meant to circle back around to updating this but I forgot. I went through a few different approaches but landed on updating the state by calling a method. Let me know if you think I should do it differently: https://github.com/ionic-team/ionic-framework/pull/31195/changes/9a25d30e305c179e74e1d1fb0a55f2325e107c52 and https://github.com/ionic-team/ionic-framework/pull/31195/commits/b6e8398406ad35470a64d420ab63c3b97a35df94

core/src/components/gallery-item/gallery-item.tsx
brandyscarney: https://github.com/ionic-team/ionic-framework/pull/31195/changes/9a25d30e305c179e74e1d1fb0a55f2325e107c52 and https://github.com/ionic-team/ionic-framework/pull/31195/commits/b6e8398406ad35470a64d420ab63c3b97a35df94

core/src/components/gallery/gallery.spec.ts
brandyscarney: https://github.com/ionic-team/ionic-framework/pull/31195/commits/6cd971ea61e2984dfa18acf393790871ff3a0946

core/src/components/gallery-item/gallery-item.tsx
brandyscarney: Good question! Yes this was intentional. This matches how we treat other components where the parent is doing most of the layout work. See [segment](https://github.com/ionic-team/ionic-framework/tree/main/core/src/components/segment) and [segment-button](https://github.com/ionic-team/ionic-framework/tree/main/core/src/components/segment-button).

The visuals for the `gallery-item` layout is covered by the [parent `gallery` tests](https://github.com/ionic-team/ionic-framework/tree/FW-7301/core/src/components/gallery/test/layout). Any regressions with the gallery item styles will be caught by those tests.

ShaneK (approved):
Awesome work! Looks good to me 🚀

core/src/components/gallery-item/gallery-item.scss
```diff
@@ -0,0 +1,51 @@
+@use "../../themes/native/native.globals" as globals;
```
thetaPC: Since this is shared by all themes, we should be using the mixin path instead.

```suggestion
@use "../../themes/mixins" as mixins;
```

core/src/components/gallery-item/gallery-item.scss
```diff
@@ -0,0 +1,51 @@
...
+  width: 100%;
+}
+
+// Slotted content
+// --------------------------------------------------
+
+// Reset the default margin for slotted elements so wrapper elements
+// (such as <figure>) align properly with other gallery items.
+::slotted(*) {
+  @include globals.margin(0);
```
thetaPC:
```suggestion
  @include mixins.margin(0);
```

core/src/components/gallery-item/gallery-item.tsx
```diff
@@ -0,0 +1,88 @@
...
+  render() {
+    const { galleryLayout } = this;
+    const theme = getIonTheme(this);
+
+    return (
+      <Host
+        class={{
+          [theme]: true,
+          'in-gallery-layout-uniform': galleryLayout === 'uniform',
+          'in-gallery-layout-masonry': galleryLayout === 'masonry',
```
thetaPC: Couldn't we do the following so it follows how we set classes?

```suggestion
          `in-gallery-layout-${galleryLayout}`: galleryLayout != undefined,
```

core/src/components/gallery/test/wrapper/gallery.e2e.ts
```diff
@@ -0,0 +1,84 @@
+import { expect } from '@playwright/test';
+import { configs, test } from '@utils/test/playwright';
+
+import { sharedGalleryStyles } from '../utils';
+
+const LAYOUT_OPTIONS = ['uniform', 'masonry'];
```
thetaPC, line 6: We might want to create const like we've been doing in the Modular Ionic so we can pass them to tests and the component itself. Doesn't have to happen now, but it'll have to during the migration.

```suggestion
export const ION_GALLERY_LAYOUT_OPTIONS = ['uniform', 'masonry']as const;
export type IonGalleryLayout = (typeof ION_GALLERY_LAYOUT_OPTIONS)[number];
```

thetaPC (approved):
LGTM, left suggestions but not blockers

core/src/components/gallery/test/wrapper/gallery.e2e.ts
brandyscarney: I'd rather just do that during the migration if that's okay since we will need to create the [interfaces](https://github.com/ionic-team/ionic-framework/blob/0ca882dce1d2f235505fd9b43f13a97f5103b696/core/src/components/badge/badge.interfaces.ts#L88) file and everything.

core/src/components/gallery-item/gallery-item.scss
brandyscarney: https://github.com/ionic-team/ionic-framework/pull/31195/changes/3c1dc324f01527913727c11c4c7876e246591236

core/src/components/gallery-item/gallery-item.scss
brandyscarney: https://github.com/ionic-team/ionic-framework/pull/31195/changes/3c1dc324f01527913727c11c4c7876e246591236

core/src/components/gallery-item/gallery-item.tsx
brandyscarney: https://github.com/ionic-team/ionic-framework/pull/31195/changes/e96350b827565866fb883601babcd33e4ebea56f

core/src/components/gallery/test/wrapper/gallery.e2e.ts
thetaPC: Works for me!

PR #31196, opened 2026-06-05
feat(angular): default to zoneless change detection

### LINKED ISSUE #30804: feat(angular): full support zoneless
### Prerequisites

With the release of Angular 21, zoneless applications are now the default configuration. While zone.js can still be enabled, it is expected to be deprecated in the future.

### Describe the Use Case

Adding zoneless support would let Ionic work out-of-the-box with Angular 21’s default setup and future Angular versions.

_No response_

### Describe Alternatives

_No response_

### Related Code

_No response_

### Additional Information

_No response_

ShaneK:
Issue number: resolves #30804

---------

Currently, on `major-9.0`, `@ionic/angular` still defaulted to Zone.js. The `ng add` schematic registered `provideZoneChangeDetection()`, and `zone.js` was a required peer dependency on both `@ionic/angular` and `@ionic/angular-server`. Angular 21 bootstraps zoneless out of the box, so Ionic was actively opting new apps back into Zone.js.

## What is the new behavior?

Ionic 9 now defaults to zoneless change detection while keeping Zone.js as a supported opt-in. The `ng add` schematic no longer registers `provideZoneChangeDetection()`, and `zone.js` is now an optional peer dependency on both `@ionic/angular` and `@ionic/angular-server`.

- [X] Yes
- [ ] No

Apps that want to keep Zone.js on Angular 21 must opt back in with `provideZoneChangeDetection()` AND keep `import 'zone.js'` in their polyfills (Angular 21's default scaffold omits it). Under zoneless, state updated from async callbacks Angular doesn't wrap needs a signal or a `markForCheck()` call. Template bindings, `@HostListener`, reactive forms, and synchronous lifecycle-hook state are unaffected. Angular 18 through 20 keep Angular's Zone.js default and need no change. BREAKING.md is updated with the full migration path.

## Other information

- Test apps run dual-mode for coverage: ng21 is zoneless, ng18-20 stay on Zone.js. Shared `base/` test pages use the new `assertZoneContext()` helper (`base/src/app/zone-assert.util.ts`), which asserts in-zone only when Zone.js is present, so the same pages pass in both modes.

Current dev build (2026-06-08):
```
8.8.9-dev.11781013468.1ed93da9
```

Test app: https://ionic-framework-git-feat-zoneless-ionic1.vercel.app/angular/

Docs PR: https://github.com/ionic-team/ionic-docs/pull/4541
Docs Preview URL: https://ionic-docs-git-feat-zoneless-v2-ionic1.vercel.app/docs/updating/9-0#zoneless-change-detection


thetaPC (changes requested):
Issues found during testing:
- On the `/angular/lazy/modals` page, when I reopen the modal after closing, I get `Error: ngOnInit was not called`.

Image

- On the `/angular/lazy/providers` page, `isResized` is already set to `true` after the page loads even after a hard refresh.

ShaneK:
@thetaPC Good find on the first one, fixed in [c55bd21](https://github.com/ionic-team/ionic-framework/pull/31196/commits/c55bd2119f065cbc976bebd047c5cfd8e3dbe188)

The second one isn't actually an issue, I just did a bad job of explaining the test. The page itself emits an artificial resize event, the fact that it says "true" when you load it shows that the change detection is working properly despite being zoneless. I've updated that to a counter so you can see resize events incrementing it properly too, but it'll always start at 1 because the page itself does the first one.

thetaPC (approved):
LGTM

BREAKING.md
```diff
@@ -128,9 +128,17 @@ Apps that relied on `ionChange` firing on every confirmation (for example, to de
...
+Ionic 9 defaults to zoneless change detection. Angular 21 bootstraps zoneless out of the box, so a new Ionic 9 app on Angular 21 runs without Zone.js and requires no change-detection provider. The `ng add @ionic/angular` schematic no longer registers `provideZoneChangeDetection()`.
+
+Because Zone.js no longer triggers change detection automatically, component state that you update from an asynchronous callback Angular doesn't wrap (awaiting an overlay result such as `modal.onWillDismiss()`, `setTimeout`, RxJS subscriptions, `Platform` events) no longer re-renders on its own. Update a signal or call `ChangeDetectorRef.markForCheck()` in those callbacks. Template event bindings, `@HostListener`, reactive forms, and Ionic lifecycle hooks (`ionViewWillEnter`, etc.) that set state synchronously are unaffected. Refer to the [Zoneless Change Detection guide](https://ionicframework.com/docs/angular/zoneless) for the patterns.
```
brandyscarney: This is missing a word or something:

> Because Zone.js no longer triggers change detection automatically, component state that you update from an asynchronous callback Angular

brandyscarney (approved):
Great work!

BREAKING.md
ShaneK: Nice catch! Fixed here: https://github.com/ionic-team/ionic-framework/pull/31196/commits/4cbb0cb219762c51fe829e1ec9f49db35f644a6b

PR #31197, opened 2026-06-05
feat(select): set modal interface cancel button to "Ok" by default

thetaPC:
Issue number: N/A

---------

The select modal commits each option the moment it's tapped, so there's nothing to cancel. A "Cancel" button is misleading because it implies the selection can be undone, when in fact the change has already taken effect. The button should use wording that reflects that it simply closes the modal.

## What is the new behavior?

- Changed the cancel text to be more appropriate.
- Updated snapshots

- [ ] Yes
- [x] No

## Other information

Preview
- Select: basic


core/src/components/select/select.tsx
```diff
@@ -94,8 +94,10 @@ export class Select implements ComponentInterface {
 
   /**
    * The text to display on the cancel button.
+   * Defaults to `'Cancel'` for the `alert` and `action-sheet`
+   * interfaces, and `'Okay'` for the `modal` interface.
```
```suggestion
   * interfaces, and `'Ok'` for the `modal` interface.
```

core/src/components/select/select.tsx
```diff
@@ -492,6 +494,18 @@ export class Select implements ComponentInterface {
     return overlay;
   }
 
+  /**
+   * Resolves the cancel button text, falling back to the
+   * per-interface default when `cancelText` is not set:
+   * `'Okay'` for the `modal` interface, `'Cancel'` otherwise.
```
```suggestion
   * `'Ok'` for the `modal` interface, `'Cancel'` otherwise.
```

brandyscarney (changes requested):

core/src/components/select/select.tsx
```diff
@@ -525,6 +527,18 @@ export class Select implements ComponentInterface {
     return overlay;
   }
 
+  /**
+   * Resolves the cancel button text, falling back to the
+   * per-interface default when `cancelText` is not set:
+   * `'Ok'` for the `modal` interface, `'Cancel'` otherwise.
+   */
```
brandyscarney, line 534: This is really confusing. The cancel button should be closing the overlay without selecting anything, but that is not how the "Ok" button works in modal. If it is going to be tied to `cancelText` it should behave the same way as the other buttons: not committing their selections when pressed. Then a separate "Ok" button would be added.

PR #30839, opened 2025-12-05
fix(modal): allow interaction with parent content through sheet modals in child routes

ShaneK:
Issue number: resolves #30700

---------

## What is the current behavior?

When a sheet modal with showBackdrop=false is rendered in a child route (nested ion-router-outlet), the parent content becomes non-interactive. Clicks on buttons or other interactive elements in the parent component are blocked, even though showBackdrop=false should allow background interaction.

Two separate issues contributed to this bug:
1. **Root locking with `backdropBreakpoint`**: The `shouldLockRoot` logic in `overlays.ts` didn't account for `backdropBreakpoint`. Modals with `backdropBreakpoint > 0` were still locking the root with `aria-hidden`, even though developers expect background interaction when the modal is below the backdrop breakpoint.
2. **Child route wrapper blocking**: When a modal is in a child route, the child route's page wrapper (`ion-page`) and its parent `ion-router-outlet` remain in the DOM with `position: absolute` covering the viewport. Even after the modal is moved to `ion-app` and has `pointer-events: none`, these wrapper elements block clicks to the parent page's content.

This issue stems from [#30563](https://github.com/ionic-team/ionic-framework/pull/30563), which added root-locking behavior that didn't account for modals that allow background interaction. A partial fix in [#30689](https://github.com/ionic-team/ionic-framework/pull/30689) partially addressed `showBackdrop=false` and `focusTrap=false`, but missed `backdropBreakpoint`.

## What is the new behavior?

Sheet modals with showBackdrop=false or focusTrap=false now correctly allow interaction with parent content when the modal is in a child route.
Improvements:
- Recalculates isSheetModal in present() to handle Angular binding timing
- Sets pointer-events: none on the modal element and its original parent elements when background interaction should be allowed
- Cleans up pointer-events on dismiss
- Adds regression tests

## Does this introduce a breaking change?

- [ ] Yes
- [X] No

## Other information

Dev build:
```
8.7.12-dev.11765060985.14ad27fb
```

PR #31202, opened 2026-06-09
fix(modal): prevent sheet gesture crash with late-bound breakpoints

ShaneK:
Issue number: internal

---------

Currently, when an `ion-modal`'s `breakpoints` are set after the component has loaded, `sortedBreakpoints` stays empty. `breakpointsChanged` had no `@Watch`, so it only ran once from the manual call in `componentDidLoad`. `present()` still recalculates `isSheetModal` to `true` from the late-set `breakpoints` (added in #30839), so the sheet gesture initializes with an empty breakpoints array and throws `Reduce of empty array with no initial value` on the `breakpoints.reduce(...)` in `sheet.ts` when the sheet is swiped, which breaks swiping entirely.

This shows up with framework bindings that land after the web component loads, e.g., an inline `ion-modal` inside an Angular component rendered through a plain `router-outlet` with zoneless change detection. It doesn't reproduce with zone change detection, with `ModalController.create`, or with the modal inside `ion-router-outlet`, since those set `breakpoints` before the modal presents.

## What is the new behavior?

Now `breakpointsChanged` has `@Watch('breakpoints')`, so `breakpoints` set after the component loads update `sortedBreakpoints`. This mirrors how `trigger`/`triggerChanged` already handles late binding (a watch plus a manual call in `componentDidLoad`). The sheet gesture then gets the real breakpoints and swiping works.

- [ ] Yes
- [X] No

## Other information

Sheet modal preview:
- https://ionic-framework-git-fix-modal-late-breakpoints-ionic1.vercel.app/src/components/modal/test/sheet

Note that the preview won't show *this* working, but will show the modal at least isn't broken from this change.

Current dev build:
```
8.8.10-dev.11781019195.14a7eeaf
```

### DIFF
--- a/core/src/components/modal/modal.tsx
+++ b/core/src/components/modal/modal.tsx
@@ -438,6 +438,7 @@ export class Modal implements ComponentInterface, OverlayInterface {
    */
   @Event() ionDragEnd!: EventEmitter<ModalDragEventDetail>;
 
+  @Watch('breakpoints')
   breakpointsChanged(breakpoints: number[] | undefined) {
     if (breakpoints !== undefined) {
       this.sortedBreakpoints = breakpoints.sort((a, b) => a - b);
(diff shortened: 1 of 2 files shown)


thetaPC (approved):
LGTM

PR #31203, opened 2026-06-09
feat(ripple-effect): add recipe and tokens

thetaPC:

---------

`ion-ripple-effect` fragments its styles across separate theme stylesheets.

## What is the new behavior?

- Migrates `ion-ripple-effect` to Modular Ionic.
- Consolidates `ripple-effect.common.scss` and `ripple-effect.ionic.scss` into a single `ripple-effect.scss`.
- Adds `IonRippleEffectRecipe` with `--ion-ripple-effect-opacity`, plus per-theme token defaults in the ios, md, and ionic theme files.
- Renames the previously undocumented `--ripple-opacity` to `--ion-ripple-effect-opacity`.
- Prefixes the JS-driven ripple variables `--translate-end` and `--final-scale` as `--internal-translate-end` and `--internal-final-scale`.
- Removes the `theme` virtual prop and the `getIonTheme` runtime read.
- Extracts the scale duration into a `SCALE_DURATION` constant tied to the `$scale-duration` Sass variable.
- Adds `ION_RIPPLE_EFFECT_TYPES` plus the `IonRippleEffectType` type.

- [x] Yes
- [ ] No

This PR introduces breaking changes to how `ion-ripple-effect` is styled.

Migration Path:

1. **Opacity variable**: the previously undocumented `--ripple-opacity` is renamed.

|-----|---|---|
| `--ripple-opacity` | `IonRippleEffect.opacity` | `--ion-ripple-effect-opacity` |

2. **Theme classes**: Remove any instances that target the theme classes: ion-ripple-effect.md, ion-ripple-effect.ios

## Other information

Preview
- [Ripple Effect: basic](https://ionic-framework-git-fw-6856-ionic1.vercel.app/src/components/ripple-effect/test/basic/)


core/src/components/ripple-effect/test/basic/index.html
```diff
@@ -129,10 +129,6 @@ <h2>Ripple at full expansion (auto-triggered, persists for inspection)</h2>
       </ion-content>
     </ion-app>
     <script>
-      function blockClicked() {
```
thetaPC, line 94: Was never used

core/src/components/ripple-effect/test/basic/index.html
thetaPC, line 1: Cleaned up the layout and added ripple demos at full trigger to make it easier to debug.

core/src/components/ripple-effect/test/ripple-effect.spec.ts
thetaPC, line 1: This can be tested manually by commenting out the [disconnection flag](https://github.com/ionic-team/ionic-framework/pull/31203/changes#diff-1082cfd297cac1fca40802cb4e4c688f1a2b49c68824de80a176bfa60e16fec1R63).

core/src/components/ripple-effect/ripple-effect.ionic.scss
thetaPC, line 1: While consolidating the ripple-effect theme files, the only difference in the `rippleAnimation` keyframe between ionic and ios/md was the easing curve:
* ios/md used `cubic-bezier(0.4, 0, 0.2, 1)` directly
* ionic used `$ion-transition-curve-expressive`

That ionic variable is defined as `var(--token-transition-curve-expressive, cubic-bezier(0.4, 0, 0.2, 1))`, so it resolves to the exact same curve by default. All three themes animate identically.

I chose to hardcode `cubic-bezier(0.4, 0, 0.2, 1)` in the consolidated file rather than introduce a `--ion-ripple-effect-*` token for it, because:
1. No visual difference to express. A token's purpose is to let a value vary per theme or be customized. Here the value is identical across all three themes, so a token would add public API surface and a recipe/token entry in every theme file for zero behavioral gain.
2. Consistency cost. If we tokenized this easing, we'd owe matching tokens for the other keyframe easings (`fadeInAnimation`/`fadeOutAnimation` both use linear) to avoid an arbitrary split — more surface area, still no visual difference.

If we ever need the ripple easing to be theme-customizable, then we can create the token at that time.

core/src/components/ripple-effect/test/basic/index.html
brandyscarney:
```suggestion
    <title>Ripple Effect - Basic</title>
```

brandyscarney:
```suggestion
          <ion-title>Ripple Effect - Basic</ion-title>
```

brandyscarney (approved):
Looks good!

ShaneK (approved):
Awesome work, looks good to me!

core/src/components/ripple-effect/test/basic/index.html
thetaPC: https://github.com/ionic-team/ionic-framework/pull/31203/changes/0791dd9db6cd224296a2a3967e80ed532de1dfe9

thetaPC: https://github.com/ionic-team/ionic-framework/pull/31203/changes/0791dd9db6cd224296a2a3967e80ed532de1dfe9

PR #31210, opened 2026-06-10
feat(skeleton-text): add recipe and tokens

thetaPC:

---------

`ion-skeleton-text` is not integrated with the Modular Ionic token system.

## What is the new behavior?

- Migrates `ion-skeleton-text` to Modular Ionic, consolidating its styles into a single token-driven stylesheet.
- Adds `IonSkeletonTextRecipe` plus per-theme token defaults in the ios, md, and ionic theme files.
- Replaces `--background` and `--background-rgb` with per-state background tokens. Each state (resting and animated) exposes an RGB channel and an alpha, so the resting fill and the animated shimmer can be recolored and have their opacity set independently (`--ion-skeleton-text-default-background-rgb` / `-alpha` and `--ion-skeleton-text-animated-background-rgb` / `-alpha`).
- Renames `--border-radius` to `--ion-skeleton-text-border-radius`, and adds `--ion-skeleton-text-line-height` and `--ion-skeleton-text-margin-top` / `--ion-skeleton-text-margin-bottom` tokens.
- Removes the runtime theme branching (`getIonTheme` and the `theme` virtual prop); visual differences now come entirely from tokens.
- Rounds a slotted skeleton to match its media parent: `ion-thumbnail` sets the radius on `::slotted(ion-skeleton-text)`, and `ion-avatar` does the same through its existing `--border-radius`.

- [x] Yes
- [ ] No

This PR introduces breaking changes to how `ion-skeleton-text` is styled.

Migration Path:

1. **Background**: `--background` and `--background-rgb` are replaced with per-state RGB and alpha variables. The single `--background-rgb` previously tinted both the resting fill and the shimmer, so set both state variables to recolor both. `--background` (the resolved resting color) has no one-to-one replacement; set the resting RGB and alpha instead.

|-----|---|---|
| `--background-rgb` | `IonSkeletonText.default.background.rgb` | `--ion-skeleton-text-default-background-rgb` |
| `--background-rgb` | `IonSkeletonText.animated.background.rgb` | `--ion-skeleton-text-animated-background-rgb` |

2. **Border radius**

|-----|---|---|
| `--border-radius` | `IonSkeletonText.border.radius` | `--ion-skeleton-text-border-radius` |

3. **Theme classes**: Remove any instances that target the theme classes: ion-skeleton-text.md, ion-skeleton-text.ios

## Other information

Previews
- [Skeleton Text: basic](https://ionic-framework-git-fw-6859-ionic1.vercel.app/src/components/skeleton-text/test/basic/)
- [Skeleton Text: custom](https://ionic-framework-git-fw-6859-ionic1.vercel.app/src/components/skeleton-text/test/custom/)


core/src/components/avatar/avatar.common.scss
```diff
@@ -25,3 +25,7 @@
 
   overflow: hidden;
 }
+
+::slotted(ion-skeleton-text) {
+  @include border-radius(var(--border-radius, revert-layer));
```
thetaPC, line 30: This will eventually look like `@include border-radius(var(--ion-avatar-border-radius, revert-layer));` when avatar is migrated.

core/src/components/skeleton-text/test/basic/skeleton-text.e2e.ts-snapshots/skeleton-text-basic-ios-ltr-Mobile-Chrome-linux.png
thetaPC, line 1: Deleted snapshots because they didn't differ in behavior when it came to modes or direction.

core/src/components/skeleton-text/test/basic/skeleton-text.e2e.ts-snapshots/skeleton-text-basic-md-ltr-Mobile-Chrome-linux.png
thetaPC, line 1: Now using `setContent` and simplified it to show one of each use case that the basic test page is showing.

core/src/components/skeleton-text/test/custom/skeleton-text.e2e.ts-snapshots/skeleton-text-custom-md-ltr-Mobile-Chrome-linux.png
thetaPC, line 1: Now using `setContent` and simplified it to show one of each use case that the custom test page is showing.

core/src/components/skeleton-text/test/skeleton-text.spec.ts
thetaPC, line 1: Since animations are turned off on e2e, it was easier to test out in a spec.

core/src/components/skeleton-text/skeleton-text.interface.ts
thetaPC, line 1: I decided to split the background into two states and also introduced alpha. This would allow further flexibility for devs to customize their skeletons.

ShaneK (approved):

core/src/components/skeleton-text/test/basic/skeleton-text.e2e.ts-snapshots/skeleton-text-basic-md-ltr-Mobile-Chrome-linux.png
brandyscarney, line 1: I believe part of this screenshot was also capturing the width being calculated properly. Do we have test coverage for this?

core/src/components/skeleton-text/skeleton-text.scss
```diff
@@ -1,30 +1,54 @@
-@import "./skeleton-text.vars";
+@use "../../themes/mixins" as mixins;
 
-// Skeleton Text
+// Skeleton Text: Common Styles
 // --------------------------------------------------
 
+// RGB fallback used when no theme provides the skeleton's background
+$background-rgb-fallback: var(--ion-text-color-rgb, 0, 0, 0);
```
brandyscarney: Is this the desired way to add variables instead of a vars file?

thetaPC: Yes, at least in my opinion. The goal is to limit the need for variables, and only use them if they are going to be constantly used throughout or the value is very complex.

brandyscarney: I think this kind of goes against how we've been separating out constants into interfaces files. I would expect the variables to be in vars files still.

core/src/components/skeleton-text/test/basic/skeleton-text.e2e.ts-snapshots/skeleton-text-basic-md-ltr-Mobile-Chrome-linux.png
thetaPC: https://github.com/ionic-team/ionic-framework/pull/31210/changes/17cdf0330e91f4d23ccbecae16582c1d76ba8d09

core/src/components/skeleton-text/skeleton-text.scss
thetaPC: I went with plural (*.vars.scss) since it sounded weird singular. But let me know if it should be singular like the interfaces file and I'll change it. https://github.com/ionic-team/ionic-framework/pull/31210/changes/0ed659d7e1e255d22d7821f9e1509a48e7488d18

brandyscarney (approved):
Looks great!

PR #31212, opened 2026-06-11
fix(searchbar): fix search icon sometimes being offset incorrectly

### LINKED ISSUE #30434: bug: ion-searchbar style is broken inside of ion-modal.
### Prerequisites

### Ionic Framework Version

v8.x

### Current Behavior

Image

### Expected Behavior

Image

### Steps to Reproduce

Add the ion-searchbar inside of the modal.

I suspect it must wait for modal to finish animating.

### Code Reproduction URL

https://stackblitz.com/edit/wercss8s

### Ionic Info

N/A as i used component playground on Stackblitz

### Additional Information

_No response_

droc101:
Issue number: resolves #30434

---------
The search icon in an ion-searchbar is sometimes positioned incorrectly.

## What is the new behavior?
If the `positionPlaceholder` function in `searchbar.tsx` detects that `iconEl.clientWidth` is 0, it will use a `ResizeObserver` to wait until the element has a width > 0, then run `positionPlaceholder` again.

- [ ] Yes
- [x] No

### DIFF
--- a/core/src/components/searchbar/searchbar.tsx
+++ b/core/src/components/searchbar/searchbar.tsx
@@ -32,6 +32,7 @@ export class Searchbar implements ComponentInterface {
   private inheritedAttributes: Attributes = {};
   private loadTimeout: ReturnType<typeof setTimeout> | undefined;
   private clearTimeout: ReturnType<typeof setTimeout> | undefined;
+  private searchIconResizeObserver?: ResizeObserver;
 
   /**
    * The value of the input when the textarea is focused.
@@ -302,6 +303,7 @@ export class Searchbar implements ComponentInterface {
     if (this.clearTimeout) {
       clearTimeout(this.clearTimeout);
     }
+    this.searchIconResizeObserver?.disconnect();
   }
 
   private emitStyle() {
@@ -547,13 +549,14 @@ export class Searchbar implements ComponentInterface {
          * If it is zero, set up a ResizeObserver and call this function when it observes a width greater than 0.
          */
         if (iconWidth === 0) {
-          const observer = new ResizeObserver((entries) => {
+          this.searchIconResizeObserver?.disconnect();
+          this.searchIconResizeObserver = new ResizeObserver((entries) => {
             if (entries[0].contentRect.width > 0) {
-              observer.disconnect();
+              this.searchIconResizeObserver?.disconnect();
               this.positionPlaceholder();
             }
           });
-          observer.observe(iconEl);
+          this.searchIconResizeObserver?.observe(iconEl);
           return;
         }
 
--- a/core/src/components/searchbar/searchbar.tsx
+++ b/core/src/components/searchbar/searchbar.tsx
@@ -506,7 +506,7 @@ export class Searchbar implements ComponentInterface {
   /**
    * Positions the input placeholder
    */
-  private positionPlaceholder() {
+  private positionPlaceholder(numTries = 0) {
     const inputEl = this.nativeInput;
     if (!inputEl) {
       return;
@@ -539,7 +539,17 @@ export class Searchbar implements ComponentInterface {
          * such as Dynamic Type on iOS as well as 8px
          * of padding.
          */
-        const iconLeft = 'calc(50% - ' + (textWidth / 2 + iconEl.clientWidth + 8) + 'px)';
+        const iconWidth = iconEl.clientWidth;
+        if (iconWidth == 0 && numTries < 10) {
+          /*
+           * fix for https://github.com/ionic-team/ionic-framework/issues/30434
+           * iconEl.clientWidth can very briefly be 0 when this is called from componentDidLoad
+           * this will try again until it is nonzero
+           */
+          this.positionPlaceholder(numTries + 1);
+          return;
+        }
+        const iconLeft = 'calc(50% - ' + (textWidth / 2 + iconWidth + 8) + 'px)';
 
         // Set the input padding start and icon margin start
         if (rtl) {
--- a/core/src/components/searchbar/searchbar.tsx
+++ b/core/src/components/searchbar/searchbar.tsx
@@ -506,7 +506,7 @@ export class Searchbar implements ComponentInterface {
   /**
    * Positions the input placeholder
    */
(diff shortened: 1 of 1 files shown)


core/src/components/searchbar/searchbar.tsx
```diff
@@ -539,7 +539,17 @@ export class Searchbar implements ComponentInterface {
          * such as Dynamic Type on iOS as well as 8px
          * of padding.
          */
-        const iconLeft = 'calc(50% - ' + (textWidth / 2 + iconEl.clientWidth + 8) + 'px)';
+        const iconWidth = iconEl.clientWidth;
+        if (iconWidth == 0 && numTries < 10) {
```
ShaneK:
```suggestion
        if (iconWidth === 0 && numTries < 10) {
```

Quick JS thing that's easy to miss, not a big deal: `===` is a stricter version of `==` that skips JavaScript's type coercion, and we use it just about everywhere in this codebase. Our linter (the `eqeqeq` rule) only allows `==` for the `== null` special case, so this line would trip up `npm run lint`. `clientWidth` is always a number, so `===` is exactly what you want here. Worth running `npm run lint` from `core/` before you push, it catches these automatically.

core/src/components/searchbar/searchbar.tsx
```diff
@@ -506,7 +506,7 @@ export class Searchbar implements ComponentInterface {
   /**
    * Positions the input placeholder
    */
-  private positionPlaceholder() {
+  private positionPlaceholder(numTries = 0) {
```
ShaneK: I'm not super opposed to the current approach, the `raf` is doing real work here since it spaces the retries out a frame at a time rather than spamming layout reads. The thing I'd want to check first: did you get a chance to test against the StackBlitz repro from the issue? My worry is that 10 frames is ~160ms, the modal enter animation is around 300ms, and that search icon is an `<ion-icon>` whose SVG loads asynchronously, so on a slower device the width could still be 0 when we run out of tries. When that happens it falls through and positions with `iconWidth` at 0, which is the original bug, so it'd come back intermittently and be a pain to track down later.

Maybe worth exploring a `ResizeObserver` on the icon instead. It reacts the moment the icon gets a width rather than guessing how many frames to wait, so there's no magic number and it handles an animation or icon load of any length. We already use it in a few spots like `alert` and `popover`. Roughly:

```ts
const iconWidth = iconEl.clientWidth;
if (iconWidth === 0) {
  const ro = new ResizeObserver((entries) => {
    const width = entries[0].contentRect.width;
    if (width > 0) {
      ro.disconnect();
      // position the placeholder using width
    }
  });
  ro.observe(iconEl);
} else {
  // position the placeholder using iconWidth
}
```

I haven't tested that, it's just a sketch. Totally fine to leave the current version, but I do think you should explore the ResizeObserver version and see if it feels cleaner to you.

core/src/components/searchbar/searchbar.tsx
```diff
@@ -539,7 +539,17 @@ export class Searchbar implements ComponentInterface {
          * such as Dynamic Type on iOS as well as 8px
          * of padding.
          */
-        const iconLeft = 'calc(50% - ' + (textWidth / 2 + iconEl.clientWidth + 8) + 'px)';
+        const iconWidth = iconEl.clientWidth;
+        if (iconWidth == 0 && numTries < 10) {
+          /*
```
ShaneK: Nit, totally optional. The other explanatory comments in this method use the `/** */` JSDoc style with full capitalized sentences (like the Dynamic Type one just above), so matching that would keep things consistent. You could also pull the `10` out into a named const like `MAX_POSITION_RETRIES` so it documents itself. The latter is generally preferred to prevent magic numbers.

core/src/components/searchbar/searchbar.tsx
ShaneK: Tiny nit, take it or leave it: `numTries` reads a little awkwardly to me, something like `retries` or `retryCount` would be a bit clearer. Not a big deal at all.

ShaneK (changes requested):
Hey! Awesome first try at this problem! I left some comments - some need action, others are just suggestions (nits). I've also started the tests, but I do think they might fail at the lint step for reasons I detailed below

core/src/components/searchbar/searchbar.tsx
```diff
@@ -539,7 +539,25 @@ export class Searchbar implements ComponentInterface {
...
-        const iconLeft = 'calc(50% - ' + (textWidth / 2 + iconEl.clientWidth + 8) + 'px)';
+        const iconWidth = iconEl.clientWidth;
+
+        /**
+         * Fix for https://github.com/ionic-team/ionic-framework/issues/30434:
+         * iconEl.clientWidth can very briefly be 0 when this is called from componentDidLoad.
+         * If it is zero, set up a ResizeObserver and call this function when it observes a width greater than 0.
+         */
+        if (iconWidth === 0) {
+          const observer = new ResizeObserver((entries) => {
```
ShaneK: The observer is a local `const` and only disconnects on the success path inside the callback. If the searchbar gets removed or hidden before the icon ever hits a non-zero width (the exact race we're guarding against), it never disconnects and leaks. `positionPlaceholder` also gets called from `onFocus`, `onBlur`, `showCancelButtonChanged`, and the `shouldAlignLeft` watcher, so calling it again while the icon's still 0-width stacks up a new observer each time, and they each fire their own `positionPlaceholder` later.

Both of these go away if you hoist the observer onto the class as a field (something like `private iconResizeObserver?: ResizeObserver`), `disconnect()` any existing one before creating a new one, and disconnect in `disconnectedCallback`. `alert.tsx` and `popover.tsx` both do it this way, and you've already got a `disconnectedCallback` here clearing the timeouts so there's a natural spot for it.

ShaneK (changes requested):
Nice refactor! Just a small thing that's easy to overlook

ShaneK (approved):
Looks good to me and I've verified it works against the user reproduction by using this dev build:
```
8.8.11-dev.11781283724.13318488
```

Awesome work!

PR #31217, opened 2026-06-15
feat(angular): make standalone imports default

OS-jacobbell:

The default path for imports is `@ionic/angular` and imports the lazy-loaded angular components. `@ionic/angular/standalone` imports the standalone components.

## What is the new behavior?

The default path, `@ionic/angular`, imports standalone components. Lazy-loaded components are imported from `@ionic/angular/lazy`.

- [X] Yes
- [ ] No

Applications that use lazy-loaded angular components will need to update their import paths from `@ionic/angular` to `@ionic/angular/lazy`.
Applications that use standalone angular components will need to update their import paths from `@ionic/angular/standalone` to `@ionic/angular`.

## Other information

Ionic starters will need their import paths updated to match these changes.


packages/angular/schematics/add/index.ts
```diff
@@ -86,7 +86,7 @@ function addProvideIonicAngular(projectName: string, projectSourceRoot: Path): R
     if (host.exists(appConfig)) {
       return addRootProvider(
         projectName,
-        ({ code, external }) => code`${external('provideIonicAngular', '@ionic/angular/standalone')}({})`
+        ({ code, external }) => code`${external('provideIonicAngular', '@ionic/angular')}({})`
```
ShaneK, line 89: You updated this path to `@ionic/angular`, but the sibling NgModule path got missed. `addIonicModuleImportToNgModule` in [`schematics/utils/ast.ts:27`](https://github.com/ionic-team/ionic-framework/blob/521cffcc0f005638d068162ed80a86429c260c6a/packages/angular/schematics/utils/ast.ts#L27) still does `insertImport(..., 'IonicModule', '@ionic/angular')`, and after this PR `IonicModule` only lives in `@ionic/angular/lazy` (the standalone default `@ionic/angular` doesn't export it). So `ng add @ionic/angular` on an NgModule app, the branch that runs when `app.module.ts` exists, generates an import that won't compile. That line should pull `IonicModule` from `@ionic/angular/lazy`. CI doesn't catch it because `ng add` isn't run against a generated app.

ShaneK (changes requested):
Awesome work! Just a few problems below, but I also noticed a couple of other problems outside the diff scope:
- BREAKING.md wasn't updated, and it definitely should be for this. This is a massive breaking change. Worth a migration note: @ionic/angular/standalone becomes @ionic/angular, and @ionic/angular becomes @ionic/angular/lazy. The existing Module Resolution section also uses @ionic/angular/standalone as its example subpath (BREAKING.md:203), which no longer exists after this change.
- The Project Structure section in packages/angular/README.md (lines 96-114) is now stale. It still describes a src submodule imported from @ionic/angular and standalone imported from @ionic/angular/standalone, both reversed and renamed now.

BREAKING.md
```diff
@@ -139,6 +139,9 @@ Apps that relied on `ionChange` firing on every confirmation (for example, to de
 
 Ionic 9 requires Angular 18 or later. Angular 16 and 17 are no longer supported.
 
+**Standalone Components Imported by Default**
+Following industry standards, Ionic 9 makes standalone components the default import path. Standalone component imports have changed from `@ionic/angular/standalone` to `@ionic/angular`. Lazy-loaded component imports have changed from `@ionic/angular` to `@ionic/angular/lazy`.
```
ShaneK, line 152: Thanks for adding this section! One follow-up from my earlier review is still open: the Module Resolution paragraph further down (line 206) still uses `@ionic/angular/standalone` as its example subpath, which this PR removes. Can you point it at something that still exists, like `@ionic/angular/lazy` or `@ionic/angular/common`?

packages/angular/README.md
```diff
@@ -99,16 +99,16 @@ The local Ionic Framework build is now active in the Angular app. Changes to the
...
 
 **standalone**
 
-This is where the standalone component implementations live. It was added as a separate entry point to avoid any lazy loaded logic from accidentally being pulled in to the final build. Having a separate directory allows the lazy loaded implementation to remain accessible from `@ionic/angular` for backwards compatibility.
+This is where the standalone component implementations live. It was added as a separate entry point to avoid any lazy loaded logic from accidentally being pulled in to the final build. Having a separate directory allows the lazy loaded implementation to remain accessible from `@ionic/angular/lazy` for backwards compatibility.
 
-Ionic developers can access this by importing from `@ionic/angular/standalone`.
+Ionic developers can access this by importing from `@ionic/angular`.
 
 **src**
```
ShaneK:
```suggestion
**lazy**
```
The body text reads right now, but the heading still says `**src**`. This PR renamed that directory to `lazy/`, so the heading names a directory that no longer exists.

ShaneK (approved):
Awesome! Thanks! Two minor doc change requests, but not blocking

packages/angular/test/base/src/app/standalone/standalone.component.ts
```diff
@@ -1,6 +1,6 @@
 import { Component } from "@angular/core";
 
-import { IonicModule } from '@ionic/angular';
+import { IonicModule } from '@ionic/angular/lazy';
```
thetaPC: This file is under the standalone folder. So shouldn't this just be `@ionic/angular`?

packages/angular/test/base/src/app/app.module.ts
```diff
@@ -1,7 +1,7 @@
 import { APP_ID, NgModule } from '@angular/core';
 import { BrowserModule } from '@angular/platform-browser';
 import { RouteReuseStrategy } from '@angular/router';
-import { IonicModule, IonicRouteStrategy } from '@ionic/angular';
+import { IonicModule, IonicRouteStrategy } from '@ionic/angular/lazy';
```
thetaPC, line 4: Since the default is now standalone, shouldn't this file be updated to handle standalone?

BREAKING.md
OS-jacobbell: Woops, forgot about that one

packages/angular/test/base/src/app/standalone/standalone.component.ts
OS-jacobbell: Looks like IonicModule does nothing here and can be removed. I'm not aware of an IonicModule for standalone.

packages/angular/test/base/src/app/app.module.ts
OS-jacobbell: I think IonicModule is needed here to initialize the app for the lazy half of the test suite.

thetaPC (approved):
LGTM

BREAKING.md
brandyscarney, line 152:
```suggestion

Following industry standards, Ionic 9 makes standalone components the default import path. Standalone component imports have changed from `@ionic/angular/standalone` to `@ionic/angular`. Lazy-loaded component imports have changed from `@ionic/angular` to `@ionic/angular/lazy`.
```

Fixes missing line break:
Image

brandyscarney (approved):
Looks good to me!

PR #31218, opened 2026-06-15
fix(datetime-button): respect datetime constraints in initial text

### LINKED ISSUE #30183: bug: ion-datetime-button does not respect minuteValues and others
### Prerequisites

### Ionic Framework Version

v8.x

### Current Behavior

A `ion-datetime-button` element will default to an unselectable date or time when these properties are used on the `ion-datetime` element:
- monthValues 
- dayValues
- yearValues
- hourValues
- minuteValues

Image

Image

### Expected Behavior

The `ion-datetime-button` contents should default to show the closest available valid value. Like the example for `ion-datetime` does.

Image

### Steps to Reproduce

1. Use this template code

```html
<ion-modal [keepContentsMounted]="true">
  <ng-template>
    <ion-datetime id="datetime" minuteValues="0" />
  </ng-template>
</ion-modal>
```

2. See the button show any date or time outside the values allowed by `minuteValues` (or the other properties)

### Code Reproduction URL

https://stackblitz.com/edit/sgkfctaz?file=src%2Fapp%2Fexample.component.html

### Ionic Info

```
Ionic:

   Ionic CLI                     : 7.2.0 (C:\Users\<NAME>\AppData\Local\nvs\node\22.12.0\x64\node_modules\@ionic\cli)
   Ionic Framework               : @ionic/angular 8.4.2 (C:\Users\<NAME>\Projects\<PROJECT>\node_modules\@ionic\angular)
   @angular-devkit/build-angular : 19.1.4 (C:\Users\<NAME>\Projects\<PROJECT>\node_modules\@angular-devkit\build-angular)
   @angular-devkit/schematics    : 19.1.4 (C:\Users\<NAME>\Projects\<PROJECT>\node_modules\@angular-devkit\schematics)
   @angular/cli                  : 19.1.4 (C:\Users\<NAME>\Projects\<PROJECT>\node_modules\@angular\cli)
   @ionic/angular-toolkit        : 12.1.1 (C:\Users\<NAME>\Projects\<PROJECT>\node_modules\@ionic\angular-toolkit)

Capacitor:

   Capacitor CLI      : not installed
   @capacitor/android : 6.2.0 (C:\Users\<NAME>\Projects\<PROJECT>\node_modules\@capacitor\android)
   @capacitor/core    : 6.2.0 (C:\Users\<NAME>\Projects\<PROJECT>\node_modules\@capacitor\core)
   @capacitor/ios     : 6.2.0 (C:\Users\<NAME>\Projects\<PROJECT>\node_modules\@capacitor\ios)

Utility:

System:

   NodeJS : v22.12.0 (C:\Users\<NAME>\AppData\Local\nvs\default\node.exe)
   npm    : 11.1.0
   OS     : Windows 10
```

### Additional Information

Possibly related to https://github.com/ionic-team/ionic-framework/issues/24722

droc101:
Issue number: resolves #30183

---------

an `IonDatetimeButton`'s default value (such as when .value is undefined) always uses the exact current date & time, regardless of any restrictions requested by the associated `IonDatetime` (such as only allowing 5 minute increments)

## What is the new behavior?
- The default display value of an `IonDatetimeButton` now follows the constraints on its associated `IonDatetime`
- The actual default `.value` of the `IonDatetimeButton` is unchanged

- [ ] Yes
- [x] No


core/src/components/datetime/datetime.tsx
```diff
@@ -604,6 +610,45 @@ export class Datetime implements ComponentInterface {
...
+
+  /**
+   * Get the closest valid Date according to the restrictions on this Datetime
+   * @param date The Date to find the closest valid value for
+   */
+  @Method()
+  async getClosestDate(date: Date) {
+    const closest = this.getClosestDatetimeParts({
+      month: date.getMonth(),
+      day: date.getDay(),
```
thetaPC: `DatetimeParts` and the native `Date` API don't use the same conventions, and two fields are mismatched here:

1. `month` is off by one. `Date.getMonth()` is 0-indexed (January is `0`), but `DatetimeParts.month` is 1-indexed (January is `1`). So today, June, comes through as month 5, which `DatetimeParts` reads as May.

2. `day` is the wrong field entirely. `Date.getDay()` returns the _day of the week_ (0–6, Sun–Sat), not the day of the month. The day of the month comes from `Date.getDate()`.

You can reproduce it with a plain no-value date picker (no constraints needed): the calendar lands on Jun 17, 2026 but the button reads "May 3, 2026".

Both of these can be seen with:
```html
<ion-item>
  <ion-label>Start Date</ion-label>
  <ion-datetime-button slot="end" datetime="no-value-date"></ion-datetime-button>
</ion-item>

<!-- keep-contents-mounted makes the datetime (and the button text) compute on load, so the mismatch shows immediately -->
<ion-modal keep-contents-mounted="true">
  <ion-datetime locale="en-US" presentation="date" id="no-value-date"></ion-datetime>
</ion-modal>
```

Fix:
```suggestion
      month: date.getMonth() + 1,
      day: date.getDate(),
```

core/src/components/datetime-button/datetime-button.tsx
```diff
@@ -204,7 +203,8 @@ export class DatetimeButton implements ComponentInterface {
      * Both ion-datetime and ion-datetime-button default
      * to today's date and time if no value is set.
      */
-    const parsedDatetimes = parseDate(parsedValues.length > 0 ? parsedValues : [getToday()]);
+    const defaultDatetime = [(await datetimeEl.getClosestDate(new Date())).toISOString()];
+    const parsedDatetimes = parseDate(parsedValues.length > 0 ? parsedValues : defaultDatetime);
```
thetaPC: `getClosestDate` is awaited on every call, but the result is thrown away whenever `parsedValues` is non-empty (i.e. whenever the datetime has a value). Since `getClosestDate` is a Stencil `@Method`, this is an async cross-component round-trip on every `ionValueChange`, just to discard the result. Move it into the no-value branch and also update the comment:

```suggestion
     /**
     * Both ion-datetime and ion-datetime-button default to today's date
     * and time if no value is set. The datetime is queried for the closest
     * valid value so the button respects the same constraints (min, max,
     * minuteValues, etc.) that the datetime applies to its own default.
     */
     let valuesToParse = parsedValues;
    if (valuesToParse.length === 0) {
      const closestDate = await datetimeEl.getClosestDate(new Date());
      valuesToParse = [closestDate.toISOString()];
    }

    const parsedDatetimes = parseDate(valuesToParse);
```

core/src/components/datetime/datetime.tsx
```diff
@@ -604,6 +610,45 @@ export class Datetime implements ComponentInterface {
...
+      year: date.getFullYear(),
+      dayOfWeek: date.getDay(),
+      hour: date.getHours(),
+      minute: date.getMinutes(),
+    });
+    return removeDateTzOffset(new Date(convertDataToISO(closest)));
+  }
```
thetaPC, line 617: I'd recommend an `@internal` method instead of a new public one.

`componentWillLoad` already computes `this.defaultParts` (today snapped to the closest valid value via `getClosestValidDate`, [datetime.tsx:1540](https://github.com/droc101/ionic-framework/blob/ba0ebeb6315a2b9de6f4940bccc219b52d7c5092/core/src/components/datetime/datetime.tsx#L1540)) — exactly what the button is trying to reconstruct. So the button can just read that value rather than recomputing it from a Date, which also guarantees the button and picker never disagree.

Making it `@internal` follows how these two already communicate (the button listens to the `@internal` `ionValueChange` event), and avoids adding public API to document and support. It also avoids reconstructing parts from a `Date` (the ISO -> `removeDateTzOffset` round-trip), which is where the `getMonth()`/`getDay()` mistakes lived.

```suggestion
  /**
   * Returns the default parts the datetime falls back to when no value is set:
   * today's date and time snapped to the closest value allowed by the
   * component's constraints (`min`, `max`, and the `*Values` props).
   *
   * @internal
   */
  @Method()
  async getDefaultPart(): Promise<DatetimeParts> {
    return this.defaultParts;
  }
```

The `datetime-button.tsx` then reads it in the no-value branch
```tsx
/**
     * Both ion-datetime and ion-datetime-button default to today's date and
     * time if no value is set. We read the datetime's computed default so the
     * button respects the same constraints (min, max, minuteValues, etc.) that
     * the datetime applies to its own fallback, instead of using a raw "now".
     */
    const parsedDatetimes =
      parsedValues.length > 0 ? parseDate(parsedValues) : [await datetimeEl.getDefaultPart()];
```

thetaPC (changes requested):
Please also add tests to prevent any regressions in the future. I would recommend adding them to `datetime-button/test/basic/datetime-button.e2e.ts`:
- `presentation="date"` with no constraints — guards the plain day/month fallback (this is the case the current bug breaks).
- `presentation="time"` with `minute-values="0"` — the originally reported case from #30183.
- At least one other constraint, e.g. `hour-values` or `month-values`, to confirm it's not minute-specific.

droc101:
Tests will be coming in a separate future commit, have to figure out how that side of the code works.

thetaPC:
@droc101 I highly recommend reviewing the [E2E testing doc](https://github.com/ionic-team/ionic-framework/blob/2ac98512c78f6d47e3a6a17ff0159047ba14a5cd/docs/core/testing/usage-instructions.md).

core/src/components/datetime/datetime.tsx
```diff
@@ -604,6 +604,40 @@ export class Datetime implements ComponentInterface {
     }
   }
 
+  /**
+   * Get the closest valid DatetimeParts according to the restrictions on this Datetime
+   * @param parts The DatetimeParts to find the closest valid value for
+   */
+  private getClosestDatetimeParts(parts: DatetimeParts) {
```
thetaPC: This helper can go away. It only exists because the earlier `getClosestDate(date)` method needed reusable logic, but now that `getDefaultPart()` just returns the already-computed `this.defaultParts`, this has a single caller (`componentWillLoad`). I'd revert it back to how it was originally.

core/src/components/datetime-button/datetime-button.tsx
```diff
@@ -201,10 +201,14 @@ export class DatetimeButton implements ComponentInterface {
...
-     * to today's date and time if no value is set.
+     * Both ion-datetime and ion-datetime-button default to today's date and
+     * time if no value is set. We read the datetime's computed default so the
+     * button respects the same constraints (min, max, minuteValues, etc.) that
+     * the datetime applies to its own fallback, instead of using a raw "now".
      */
-    const parsedDatetimes = parseDate(parsedValues.length > 0 ? parsedValues : [getToday()]);
+    const parsedDatetimes = parseDate(
+      parsedValues.length > 0 ? parsedValues : [convertDataToISO(await datetimeEl.getDefaultPart())]
+    );
```
thetaPC: `getDefaultPart()` already returns `DatetimeParts`, so wrapping it in `convertDataToISO(...)` only to feed it back through `parseDate` is a parts → ISO → parts round-trip. You can read it directly and skip both the conversion and the `convertDataToISO` import:


```suggestion
    const parsedDatetimes =
      parsedValues.length > 0 ? parseDate(parsedValues) : [await datetimeEl.getDefaultPart()];
```

Unless there's a reason of why you went with this approach?

core/src/components/datetime-button/datetime-button.tsx
```diff
@@ -125,7 +125,7 @@ export class DatetimeButton implements ComponentInterface {
       overlayEl.classList.add('ion-datetime-button-overlay');
     }
 
-    componentOnReady(datetimeEl, () => {
+    componentOnReady(datetimeEl, async () => {
```
thetaPC: Why is this change necessary?

core/src/components/datetime-button/datetime-button.tsx
```diff
@@ -138,7 +138,7 @@ export class DatetimeButton implements ComponentInterface {
        * to re-render the displayed
        * text in the buttons.
        */
-      this.setDateTimeText();
+      await this.setDateTimeText();
```
thetaPC: Why is this change necessary?

core/src/components/datetime-button/datetime-button.tsx
```diff
@@ -189,7 +189,7 @@ export class DatetimeButton implements ComponentInterface {
    * ion-datetime and then format it according
    * to the locale specified on ion-datetime.
    */
-  private setDateTimeText = () => {
+  private setDateTimeText = async () => {
```
thetaPC, line 191: Why is this change necessary?

core/src/components/datetime/datetime.tsx
```diff
@@ -1495,27 +1529,12 @@ export class Datetime implements ComponentInterface {
       warnIfTimeZoneProvided(el, formatOptions);
     }
 
-    const hourValues = (this.parsedHourValues = convertToArrayOfNumbers(this.hourValues));
```
thetaPC, line 1498: As mentioned in https://github.com/ionic-team/ionic-framework/pull/31218/changes#r3463042864, I would revert this.

core/src/components/datetime/datetime.tsx
```diff
@@ -1495,27 +1529,12 @@ export class Datetime implements ComponentInterface {
...
-    const monthValues = (this.parsedMonthValues = convertToArrayOfNumbers(this.monthValues));
-    const yearValues = (this.parsedYearValues = convertToArrayOfNumbers(this.yearValues));
-    const dayValues = (this.parsedDayValues = convertToArrayOfNumbers(this.dayValues));
-
     const todayParts = (this.todayParts = parseDate(getToday())!);
 
     this.processMinParts();
     this.processMaxParts();
 
-    this.defaultParts = getClosestValidDate({
```
thetaPC, line 1509: As mentioned in https://github.com/ionic-team/ionic-framework/pull/31218/changes#r3463042864, I would revert this.

core/src/components/datetime-button/test/basic/datetime-button.e2e.ts
```diff
@@ -344,4 +344,154 @@ configs({ modes: ['md'], directions: ['ltr'] }).forEach(({ title, config }) => {
...
+              minute: "2-digit"
+            }
+          }
+        </script>
+      `,
+        config
+      );
+      await page.locator('.datetime-ready').waitFor();
+
+      await page.pause();
```
thetaPC: Please remove the await `page.pause()` calls from these tests. `page.pause()` is a debugging only tool that halts the run and opens the Playwright Inspector, so it shouldn't ship in the suite.

core/src/components/datetime-button/test/basic/datetime-button.e2e.ts
```diff
@@ -344,4 +344,154 @@ configs({ modes: ['md'], directions: ['ltr'] }).forEach(({ title, config }) => {
       await expect(page.locator('ion-datetime-button')).toContainText('Thu, November 02 01:22 AM');
     });
   });
+
+  test.describe(title('datetime-button: datetime constraints'), () => {
+    test('should default to exact current time with no constraints', async ({ page }) => {
```
thetaPC: These constraint tests should be annotated with the issue they're covering, like the existing #27797 tests in this file do.

core/src/components/datetime-button/test/basic/datetime-button.e2e.ts
```diff
@@ -344,4 +344,154 @@ configs({ modes: ['md'], directions: ['ltr'] }).forEach(({ title, config }) => {
...
+          }
+        </script>
+      `,
+        config
+      );
+      await page.locator('.datetime-ready').waitFor();
+
+      await page.pause();
+
+      const expectedTime = fixedTime;
```
thetaPC: `expectedTime = fixedTime` copies the reference, so `setMinutes`/`setHours`/`setMonth` mutate `fixedTime` itself. It's harmless here only because `fixedTime` isn't read again.

```suggestion
      const expectedTime = new Date(fixedTime);
```

core/src/components/datetime-button/test/basic/datetime-button.e2e.ts
```diff
@@ -344,4 +344,154 @@ configs({ modes: ['md'], directions: ['ltr'] }).forEach(({ title, config }) => {
       await expect(page.locator('ion-datetime-button')).toContainText('Thu, November 02 01:22 AM');
     });
   });
+
+  test.describe(title('datetime-button: datetime constraints'), () => {
```
thetaPC, line 348: Optional / nice-to-have, take it or leave it. These four tests repeat the same clock setup and formatter definitions. We already share per-describe setup with `beforeEach` elsewhere in this file (see [datetime-button.e2e.ts:27](https://github.com/ionic-team/ionic-framework/blob/41952e6ab5bc5a4217a84694a8bf674518d1f5ce/core/src/components/datetime-button/test/basic/datetime-button.e2e.ts#L27) in the "switching to correct view" describe), so we could do the same here. Hoist the constant time and the formatters to the describe scope and set the clock in a `beforeEach`, then each test only carries its own `setContent` and expected value:

```ts
  test.describe(title('datetime-button: datetime constraints'), () => {
    const fixedTime = new Date('2026-06-18T17:54:54.518Z');

    const dateFormat = new Intl.DateTimeFormat('en-US', {
      weekday: 'short',
      month: 'long',
      day: '2-digit',
    });
    const timeFormat = new Intl.DateTimeFormat('en-US', {
      hour: '2-digit',
      minute: '2-digit',
    });

    test.beforeEach(async ({ page }) => {
      await page.clock.setFixedTime(fixedTime);
    });
```

Then each test drops its local `fixedTime`, `setFixedTime`, and `Intl.DateTimeFormat` lines and just uses the shared ones. For the expected value, clone before mutating so the shared `fixedTime` isn't altered between tests:

```ts
const expectedTime = new Date(fixedTime);
expectedTime.setMinutes(0);.
```

One caveat if you do take this: once `fixedTime` is shared across tests, cloning it before mutating becomes necessary (not just tidy), otherwise a mutation in one test leaks into the others via `beforeEach`.

thetaPC (changes requested):

thetaPC:
Small nit on the PR title: the `fix(...)` prefix already signals this is a fix, so the second "fix" is redundant. Could you reword the description part to just say what changed? Something like: `fix(datetime-button): respect datetime constraints in initial text`

droc101:
Thanks for the review! The changes making some things use async/await are there because `@Method()` seems to require the method to be async. Please let me know if I'm missing some way to omit the async, I will gladly change it.

core/src/components/datetime-button/datetime-button.tsx
thetaPC: @droc101 responding to your [comment](https://github.com/ionic-team/ionic-framework/pull/31218#issuecomment-4791435226)

> Thanks for the review! The changes making some things use async/await are there because `@Method()` seems to require the method to be async. Please let me know if I'm missing some way to omit the async, I will gladly change it.

`setDateTimeText` needs to stay async since it awaits `getDefaultPart()`, but the `componentOnReady` callback doesn't need to be. You can call it without await (fire-and-forget), like the original did:

    componentOnReady(datetimeEl, () => {

Nothing after this call reads the button text, so there's no reason to wait for it. Dropping the `await` on line 140 means the callback no longer needs `async`, which keeps the change minimal.

core/src/components/datetime-button/test/basic/datetime-button.e2e.ts
```diff
@@ -344,4 +344,146 @@ configs({ modes: ['md'], directions: ['ltr'] }).forEach(({ title, config }) => {
...
+      hour: '2-digit',
+      minute: '2-digit',
+    });
+
+    test.beforeEach(async ({ page }) => {
+      await page.clock.setFixedTime(fixedTime);
+    });
+
+    test('should default to exact current time with no constraints', async ({ page }) => {
+      test.info().annotations.push({
```
thetaPC: For consistency with the existing annotated tests in this file (the #27797 ones at [line 142](https://github.com/ionic-team/ionic-framework/blob/9d784dd37a3eb0e992e754df9d39c1bb60dec1f8/core/src/components/datetime-button/test/basic/datetime-button.e2e.ts#L142) use the testInfo parameter), use testInfo here instead of test.info(). Both work, but matching the surrounding file reads cleaner. Add testInfo as the second callback arg and reference it directly:


```suggestion
    test('should default to exact current time with no constraints', async ({ page }, testInfo) => {
      testInfo.annotations.push({
```

Same change applies to the other three constraint test.

core/src/components/datetime/datetime.tsx
```diff
@@ -1495,17 +1507,16 @@ export class Datetime implements ComponentInterface {
...
     const todayParts = (this.todayParts = parseDate(getToday())!);
 
     this.processMinParts();
     this.processMaxParts();
 
+    const hourValues = (this.parsedHourValues = convertToArrayOfNumbers(this.hourValues));
+    const minuteValues = (this.parsedMinuteValues = convertToArrayOfNumbers(this.minuteValues));
+    const monthValues = (this.parsedMonthValues = convertToArrayOfNumbers(this.monthValues));
+    const yearValues = (this.parsedYearValues = convertToArrayOfNumbers(this.yearValues));
+    const dayValues = (this.parsedDayValues = convertToArrayOfNumbers(this.dayValues));
```
thetaPC, line 1514: This block was just reordered from main (the `convertToArrayOfNumbers` lines used to sit above `todayParts`/`processMinParts`/`processMaxParts`). It has no functional effect, so restoring the original order removes this hunk entirely and leaves `getDefaultPart` as the only change to this file.

thetaPC (changes requested):

thetaPC (approved):
LGTM

PR #31221, opened 2026-06-16
feat(select): remove role from action sheet interface

thetaPC:

---------

A `role` value is being passed within the action sheet interface when it's being opened through the `ion-select`. This does not match with how the other interfaces behave, which can be seen through `event.detail`.

## What is the new behavior?

- Removed the `role` from the action sheet interface within the `ion-select`

- [x] Yes
- [ ] No

When using `interface="action-sheet"`, `ion-select` no longer assigns the `selected` role to the action sheet button for the currently selected option. This aligns the `action-sheet` interface with the `alert`, `popover`, and `modal` interfaces, none of which assign this role. The selected option's styling is unchanged because it is driven by the option's checked state.

Because the option button no longer carries the `selected` role, the action sheet's dismiss role for a selected option is no longer `"selected"`. Apps that inspected this role to detect that a value was chosen, such as reading `role` from the underlying action sheet's `onDidDismiss` result, should listen for `ion-select`'s `ionChange` event instead, which emits the selected value when the selection changes.

## Other information

Preview
- [Select: basic](https://ionic-framework-git-fw-6982-ionic1.vercel.app/src/components/select/test/basic/)


BREAKING.md
brandyscarney, line 121: Should we add a header above this now since there are two sections?

BREAKING.md
```diff
@@ -120,6 +120,12 @@ The `ionChange` event on `ion-select` now only fires when the selected value act
 
 Apps that relied on `ionChange` firing on every confirmation (for example, to detect overlay dismissal without a value change) should listen for `ionDismiss` instead, or use the `didDismiss` event on the underlying alert or action sheet.
 
+**Action Sheet Interface `selected` Role Removed**
+
+When using `interface="action-sheet"`, `ion-select` no longer assigns the `selected` role to the action sheet button for the currently selected option. This aligns the `action-sheet` interface with the `alert`, `popover`, and `modal` interfaces, none of which assign this role. The selected option's styling is unchanged because it is driven by the option's checked state.
+
+Because the option button no longer carries the `selected` role, the action sheet's dismiss role for a selected option is no longer `"selected"`. Apps that inspected this role to detect that a value was chosen, such as reading `role` from the underlying action sheet's `onDidDismiss` result, should listen for `ion-select`'s `ionChange` event instead, which emits the selected value when the selection changes.
```
brandyscarney: When was the `"selected"` role being output? Even on `main` it just shows as `''`:

Image

BREAKING.md
```diff
@@ -120,6 +120,12 @@ The `ionChange` event on `ion-select` now only fires when the selected value act
 
 Apps that relied on `ionChange` firing on every confirmation (for example, to detect overlay dismissal without a value change) should listen for `ionDismiss` instead, or use the `didDismiss` event on the underlying alert or action sheet.
 
+**Action Sheet Interface `selected` Role Removed**
+
+When using `interface="action-sheet"`, `ion-select` no longer assigns the `selected` role to the action sheet button for the currently selected option. This aligns the `action-sheet` interface with the `alert`, `popover`, and `modal` interfaces, none of which assign this role. The selected option's styling is unchanged because it is driven by the option's checked state.
```
```suggestion
When using `interface="action-sheet"`, `ion-select` no longer assigns the `selected` role to the action sheet button for the currently selected option. This aligns the `action-sheet` interface with the `alert`, `popover`, and `modal` interfaces, none of which assign this role. This does not change the selected option's styling.
```

I think this might be giving too much information. We don't need to explain what controls the styling.

brandyscarney (approved):
Looks good! Some non-blocking comments on the breaking changes.

BREAKING.md
thetaPC: https://github.com/ionic-team/ionic-framework/pull/31221/changes/376fef9b72af478a005110039b18a76785322b85

BREAKING.md
thetaPC: This would happen only when selecting the same value. Exact steps can be found in the Jira ticket. I updated the Breaking doc to clarify it: https://github.com/ionic-team/ionic-framework/pull/31221/changes/e546e913bd42fedc0289169d748fd91bdbd31d73

BREAKING.md
thetaPC: https://github.com/ionic-team/ionic-framework/pull/31221/changes/9962d4023afe36608f7e20618af79c5ddedf2463

BREAKING.md
brandyscarney: Ahh okay. Shouldn't it just say the role is now `undefined` in both cases?

thetaPC: Better? https://github.com/ionic-team/ionic-framework/pull/31221/changes/a53543faf304d79a8a354f972abc38ccee01c305

brandyscarney: Yeah looks good!

PR #31222, opened 2026-06-16
feat(modal): default handleBehavior to "cycle" for sheet modals

thetaPC:

---------

`handleBehavior` on `ion-modal` defaults to `"none"`. For sheet modals that show a handle, the handle is purely decorative: it is not focusable, it is skipped in the tab order, and activating it does nothing. Keyboard and screen reader users cannot resize the sheet unless the developer explicitly opts in with `handleBehavior="cycle"`.

## What is the new behavior?

- `handleBehavior` now defaults to `"cycle"`.
- Sheet modal handles are focusable by default and activating them (click, keyboard, or screen reader) cycles the sheet through its available breakpoints. This mirrors native iOS sheet behavior and keeps the sheet operable with assistive technology out of the box.
- Rewrote the `handleBehavior` property documentation to lead with what each value does and how `"cycle"` helps accessibility, improving findability and learnability.
- Developers can opt out with `handleBehavior="none"` to restore the previous decorative handle.

- [x] Yes
- [ ] No

This changes the default interaction and focus behavior of sheet modals that use a handle. The handle becomes focusable and enters the tab order, and activating it now cycles breakpoints where it previously did nothing. Sheet modals that relied on the handle being inert should set `handle-behavior="none"`:

```html
<ion-modal handle-behavior="none"></ion-modal>
```

## Other information

Preview
- [Modal: sheet](https://ionic-framework-git-fw-6657-ionic1.vercel.app/src/components/modal/test/sheet/)

### DIFF
--- a/BREAKING.md
+++ b/BREAKING.md
@@ -19,6 +19,7 @@ This is a comprehensive list of the breaking changes introduced in the major ver
 - [Components](#version-9x-components)
   - [Input](#version-9x-input)
   - [Legacy Picker](#version-9x-legacy-picker)
+  - [Modal](#version-9x-modal)
   - [Router Outlet](#version-9x-router-outlet)
   - [Searchbar](#version-9x-searchbar)
   - [Select](#version-9x-select)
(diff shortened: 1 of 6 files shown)


ShaneK (approved):
Looks good to me! Great work 🚀

PR #31159, opened 2026-05-20
fix(react): bind events properly for overlays rendered within a nav

ShaneK:
Issue number: resolves #27843

---------

## What is the current behavior?

`createInlineOverlayComponent` renders inline overlays (modal, popover, etc.) inside a `<template>` at their declared JSX position. When the overlay presents, `CoreDelegate` teleports the DOM node into `ion-app`, but React's synthetic event delegation root stays at the original JSX parent. Once the overlay lives outside that subtree, React no longer dispatches events to children inside it, so `onClick`, `onChange`, and other handlers inside an `IonModal` rendered within an `IonNav` silently stop firing

## What is the new behavior?

Top-level inline overlays now render through `createPortal` into the same `ion-app` container that `CoreDelegate` teleports into, so React's event root follows the DOM

## Does this introduce a breaking change?

- [ ] Yes
- [x] No

## Other information

This issue also reports the same problem in Vue, but that was fixed in #30227

Preview: [navigation-modal](https://ionic-framework-git-fw-6314-ionic1.vercel.app/react/navigation-modal)

Dev build:
```
8.8.8-dev.11779302602.17decfbf
```

PR #31223, opened 2026-06-16
fix(react): remove relocated inline overlays orphaned on unmount

ShaneK:
Issue number: internal

---------

Currently, when an inline overlay (`ion-modal`, `ion-popover`) is rendered through React and `CoreDelegate` teleports its host out of the React-rendered location, unmounting the component can leave the host orphaned in the DOM. `componentWillUnmount` only recovered a relocated host when `this.state.isOpen` was `true`, so a host that was moved while still closed was never reattached for React to remove. That happens during the React 18 StrictMode mount/unmount cycle, or any re-render before the present events flip `isOpen`. The result is duplicate `<ion-modal>` elements stacking up in `document.body`. This came from the portal change in #31159

## What is the new behavior?

This splits the unmount cleanup into two independent steps. The DOM recovery now runs whenever the host is still connected and has been relocated, regardless of open state: portaled overlays get reattached to `portalTarget` so React's `removeChild` can find them, and nested overlays moved out of their `<template>` are removed directly. The dismiss-lifecycle teardown, detaching the `didDismiss` listener and props, stays gated on `isOpen` as before. So a relocated-but-closed host is now cleaned up instead of orphaned.

- [ ] Yes
- [X] No

### DIFF
--- a/packages/react/package.json
+++ b/packages/react/package.json
@@ -77,6 +77,14 @@
   "jest": {
     "testEnvironment": "jsdom",
     "preset": "ts-jest",
+    "transform": {
+      "^.+\\.(ts|tsx)$": [
+        "ts-jest",
+        {
+          "isolatedModules": true
+        }
+      ]
+    },
     "setupFilesAfterEnv": [
       "<rootDir>/jest.setup.js"
     ],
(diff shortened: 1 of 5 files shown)


thetaPC (approved):
LGTM

PR #31225, opened 2026-06-17
fix(button): sync disabled state and type in renderHiddenButton

### LINKED ISSUE #30968: bug: ion-button with type="submit" does not submit form in some situations
### Prerequisites

### Ionic Framework Version

v8.x

### Current Behavior

`<ion-button>` with `type="submit"` does not cause the associated `<form>` to be submitted even though it is not disabled

### Expected Behavior

`<ion-button>` with `type="submit"` should always submit the associated `<form>` when it is clicked and not disabled

### Steps to Reproduce

1. Open the [reproduction URL](https://stackblitz.com/edit/ion-button-submit-bug)
2. Click the submit button (**nothing happens**)
3. Clear the text of the input field (the button becomes disabled)
4. Enter any value into the input field (the button becomes enabled again)
5. Click the button again (the text "Form submitted!" appears on the page)

### Code Reproduction URL

https://stackblitz.com/edit/ion-button-submit-bug

### Ionic Info

Ionic:

   Ionic CLI                     : 7.2.1
   Ionic Framework               : @ionic/angular 8.7.17
   @angular-devkit/build-angular : 21.1.4
   @angular-devkit/schematics    : 21.1.4
   @angular/cli                  : 21.1.4

Utility:

System:

   NodeJS : v20.19.1 (/usr/local/bin/node)
   npm    : 10.8.2
   OS     : Linux 5.0

### Additional Information

This issue is a regression, caused by the Stencil update to 4.29.3 in Ionic 8.6.0 (Copy of the reproduction with Ionic 8.5.9: https://stackblitz.com/edit/ion-button-submit-before-regression). This Stencil update changed some things in component initialization order (that I don't fully understand). Other issues that were caused by the same update which have already been resolved: https://github.com/ionic-team/ionic-framework/issues/30613, https://github.com/ionic-team/ionic-framework/issues/30552

The root cause is the "hidden button" in ion-button, which is used to submit the form. The disabled state of the hidden button is supposed to be kept in sync with the disabled state of the ion-button, but this doesn't work at initialization.

The button gets created on first render with the current disabled state of the ion-button here:
https://github.com/ionic-team/ionic-framework/blob/682a17ebb754da7714989623cf84b75e715e20e1/core/src/components/button/button.tsx#L209
In the reproduction, the button is initially disabled, so the hidden button is also disabled.

After that, it's supposed to be updated using the `@Watch` decorator:
https://github.com/ionic-team/ionic-framework/blob/main/core/src/components/button/button.tsx#L62-L68

Some time after the first render the disabled state changes without triggering the `@Watch` listener (probably because the ion-button isn't fully initialized yet?).

In my reproduction I caused this update using async form validation, because that's how I encountered this issue in production, but I'm sure there's other situations that could trigger the same behavior.

Zac-Smucker-Bryan:
Issue number: resolves #30968

---------

`<ion-button>` with type="submit" does not cause the associated `<form>` to be submitted even though it is not disabled. When the text of the input field is cleared and then input re-added, you can submit the input.

## What is the new behavior?

The visible and hidden button should always have the same disabled state. `<ion-button>` with type="submit" should always submit the associated `<form>` when it is enabled.

- Add `formButtonEl.disabled = this.disabled;` and `formButtonEl.type = this.type;` to button.tsx in `renderHiddenButton()`. This addition syncs the disabled status and type of the button when that happens.
- Add a test in to to `packages/angular/test/base/e2e/src/lazy/form.spec.ts` which should ensure any changes in button type are synced. The test gets the hidden button type, changes the visible button state, and then expects the hidden button state to change as well.
- Per conversations around this PR, testing disabled state sync will need a dev build due to the nature of the issue.

- [ ] Yes
- [x] No

## Other information

Related to a stencil update that affected when @Watch('disabled') fires.

### DIFF
--- a/core/src/components/button/button.tsx
+++ b/core/src/components/button/button.tsx
@@ -196,7 +196,7 @@ export class Button implements ComponentInterface, AnchorInterface, ButtonInterf
       /**
        * If the form already has a rendered form button
        * then do not append a new one again. Sync the
-       * disabled state and type in it changes after button
+       * disabled state and type if it changes after button
        * creation (e.g., runtime property updates).
        */
       if (formButtonEl !== null && formEl.contains(formButtonEl)) {
--- a/core/src/components/button/button.tsx
+++ b/core/src/components/button/button.tsx
@@ -195,8 +195,8 @@ export class Button implements ComponentInterface, AnchorInterface, ButtonInterf
 
       /**
        * If the form already has a rendered form button
-       * then do not append a new one again. Sync the 
-       * disabled state and type in it changes after button 
+       * then do not append a new one again. Sync the
+       * disabled state and type in it changes after button
        * creation (e.g., runtime property updates).
        */
       if (formButtonEl !== null && formEl.contains(formButtonEl)) {
--- a/core/src/components/button/button.tsx
+++ b/core/src/components/button/button.tsx
@@ -198,6 +198,7 @@ export class Button implements ComponentInterface, AnchorInterface, ButtonInterf
        * then do not append a new one again.
        */
       if (formButtonEl !== null && formEl.contains(formButtonEl)) {
+        formButtonEl.disabled = this.disabled;
         return;
       }
 
--- a/core/src/components/button/button.tsx
+++ b/core/src/components/button/button.tsx
@@ -195,10 +195,13 @@ export class Button implements ComponentInterface, AnchorInterface, ButtonInterf
 
       /**
        * If the form already has a rendered form button
-       * then do not append a new one again.
+       * then do not append a new one again. Sync the 
+       * disabled state and type in it changes after button 
+       * creation (e.g., runtime property updates).
        */
       if (formButtonEl !== null && formEl.contains(formButtonEl)) {
-        formButtonEl.disabled = this.disabled;
+        // formButtonEl.disabled = this.disabled;
+        // formButtonEl.type = this.type;
         return;
       }
 
--- a/core/src/components/button/button.tsx
+++ b/core/src/components/button/button.tsx
@@ -200,8 +200,8 @@ export class Button implements ComponentInterface, AnchorInterface, ButtonInterf
        * creation (e.g., runtime property updates).
        */
       if (formButtonEl !== null && formEl.contains(formButtonEl)) {
-        // formButtonEl.disabled = this.disabled;
-        // formButtonEl.type = this.type;
+        formButtonEl.disabled = this.disabled;
+        formButtonEl.type = this.type;
         return;
       }
 
--- a/core/src/components/button/test/form-reference/button.e2e.ts
+++ b/core/src/components/button/test/form-reference/button.e2e.ts
@@ -154,42 +154,6 @@ configs({ directions: ['ltr'], modes: ['ios'] }).forEach(({ title, config }) => 
       expect(submitEvent).toHaveReceivedEvent();
     });
 
-    test('should submit the form when disabled state changes asynchronously', async ({ page }, testInfo) => {
-      testInfo.annotations.push({
-        type: 'issue',
-        description: 'https://github.com/ionic-team/ionic-framework/issues/30968',
-      });
-
-      await page.setContent(
-        `
-        <form>
-          <input type="text" />
-          <ion-button type="submit" disabled>Submit</ion-button>
-        </form>
-        `,
-        config
-      );
-
-      const submitEvent = await page.spyOnEvent('submit');
-      const button = page.locator('ion-button');
-
-      // Simulate async disabled state change — e.g. async validator resolving
-      await button.evaluate((el: HTMLIonButtonElement) => {
-        el.disabled = true;
-        setTimeout(() => {
-          el.disabled = false;
-        }, 0);
-      });
-
-      // Wait for the async change to settle
-      await page.waitForTimeout(100);
-
-      await page.click('ion-button');
-
-      expect(submitEvent).toHaveReceivedEvent();
-    });
-  });
-
   test.describe(title('should throw a warning if the form cannot be found'), () => {
     test('form is a string selector', async ({ page }) => {
       const logs: string[] = [];
--- a/core/src/components/button/test/form-reference/button.e2e.ts
+++ b/core/src/components/button/test/form-reference/button.e2e.ts
@@ -153,6 +153,41 @@ configs({ directions: ['ltr'], modes: ['ios'] }).forEach(({ title, config }) => 
 
       expect(submitEvent).toHaveReceivedEvent();
     });
+
+    test('should keep hidden button type in sync with visible button', async ({ page }) => {
+      // Set the form to have a submit button
+      await page.setContent(
+        `
+        <form id="myForm"></form>
+        <ion-button form="myForm" type="submit">
+          Submit
+        </ion-button>
+      `,
+        config
+      );
+
+      // Get visible button
+      const button = page.locator('ion-button');
+
+      // Get type of the hidden button
+      const getHiddenType = () =>
+        page.evaluate(() => {
+          const hidden = document.querySelector('form button[style*="display: none"]') as HTMLButtonElement;
+
+          return hidden?.type;
+        });
+
+      // Type of hidden button should be submit to start
+      expect(await getHiddenType()).toBe('submit');
+
+      // Set type of visible button to reset
+      await button.evaluate((el: HTMLIonButtonElement) => {
+        el.type = 'reset';
+      });
+
+      // Expect hidden button type to be reset
+      await expect.poll(async () => await getHiddenType()).toBe('reset');
+    });
   });
 
   test.describe(title('should throw a warning if the form cannot be found'), () => {
--- a/core/src/components/button/test/form-reference/button.e2e.ts
+++ b/core/src/components/button/test/form-reference/button.e2e.ts
@@ -154,7 +154,7 @@ configs({ directions: ['ltr'], modes: ['ios'] }).forEach(({ title, config }) => 
       expect(submitEvent).toHaveReceivedEvent();
     });
   });
-  
+
   test.describe(title('should throw a warning if the form cannot be found'), () => {
     test('form is a string selector', async ({ page }) => {
       const logs: string[] = [];
--- a/core/src/components/button/test/form-reference/button.e2e.ts
+++ b/core/src/components/button/test/form-reference/button.e2e.ts
@@ -153,6 +153,42 @@ configs({ directions: ['ltr'], modes: ['ios'] }).forEach(({ title, config }) => 
 
       expect(submitEvent).toHaveReceivedEvent();
     });
+
+    test('should submit the form when disabled state changes asynchronously', async ({ page }, testInfo) => {
+      testInfo.annotations.push({
+        type: 'issue',
+        description: 'https://github.com/ionic-team/ionic-framework/issues/30968',
+      });
+
+      await page.setContent(
+        `
+        <form>
+          <input type="text" />
+          <ion-button type="submit" disabled>Submit</ion-button>
+        </form>
+        `,
+        config
+      );
+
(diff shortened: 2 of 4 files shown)


core/src/components/button/button.tsx
thetaPC, line 1: We need to update the following comment to reflect the changes and why they were added.

```
/**
 * If the form already has a rendered form button
 * then do not append a new one again.
 */
```

core/src/components/button/button.tsx
```diff
@@ -198,6 +198,7 @@ export class Button implements ComponentInterface, AnchorInterface, ButtonInterf
        * then do not append a new one again.
        */
       if (formButtonEl !== null && formEl.contains(formButtonEl)) {
+        formButtonEl.disabled = this.disabled;
```
thetaPC, line 203: We should also keep `type` in sync on the hidden button. The early return path never updates it, so a runtime change from `submit` to `reset` (or vice versa) leaves the hidden button with its original type and it performs the wrong form action.


```suggestion
        formButtonEl.disabled = this.disabled;
        formButtonEl.type = this.type;
```

core/src/components/button/test/form-reference/button.e2e.ts
```diff
@@ -153,6 +153,41 @@ configs({ directions: ['ltr'], modes: ['ios'] }).forEach(({ title, config }) =>
 
       expect(submitEvent).toHaveReceivedEvent();
     });
+
+    test('should submit the form when disabled state changes asynchronously', async ({ page }, testInfo) => {
```
thetaPC: This test does not actually catch the regression. It still passes without the fix. I verified by reverting the `formButtonEl.disabled = this.disabled;` line, keeping the test, and rebuilding: it stays green.

The reason is timing. The bug only happens when the `disabled` change races with Stencil's initialization, which is the window where `@Watch('disabled')` gets missed. This test changes `disabled` with `el.evaluate` and `setTimeout` after the component has fully loaded, and by that point the watch is active again and syncs normally. So it reduces to "set disabled to false after load," which is the same as the manual workaround in the issue (clear the input, then retype) that already worked before your fix. Also worth noting that `el.disabled = true` is a no-op there since the button already starts disabled in the markup.

There is no post-load DOM API to suppress the watch, so core e2e cannot reproduce this one deterministically. The faithful reproduction needs a framework binding that sets `disabled` during init, which is exactly the Angular async validator scenario from the issue.

Please move the regression test to `packages/angular/test/base/e2e/src/lazy/form.spec.ts`. Set up a submit button that starts disabled and has its disabled state flip to false asynchronously during init (an async validator resolving, or a `Promise.resolve().then(...)` in `ngOnInit`), then assert the form submits on click. Confirm it fails without the fix before pushing, since that is the whole point of the test.

If you are feeling adventurous, add the same coverage to the standalone test app too.

thetaPC (changes requested):
Small thing on the PR title: `fix(ion-button): sync disabled state in ion-button renderHiddenButton`

Two tweaks. The scope convention here uses the component name without the `ion-` prefix (for example `fix(button)`, `fix(modal)`, `fix(checkbox)`), so the scope should be `button`. And since the scope already says it's about the button, repeating `ion-button` in the description is redundant. Suggested:

`fix(button): sync disabled state in renderHiddenButton`

Small thing on the PR description:

The issue link in the description won't auto close the issue. GitHub's closing keywords only fire when the keyword is followed directly by `#` and the number, like `resolves #30968`. Here it's `resolves #[30968](...)`, so the `#` is followed by a `[` and GitHub doesn't parse it as an issue reference. The markdown link renders fine visually but the linkage and auto close on merge won't happen.

Please change it to:

`Issue number: resolves #30968`

GitHub will turn that into a link on its own, so the markdown link isn't needed.

Once you push your latest changes, make sure that the PR description is still valid.

core/src/components/button/button.tsx
Zac-Smucker-Bryan: That makes sense. I have included that in the latest round of commits.

core/src/components/button/test/form-reference/button.e2e.ts
Zac-Smucker-Bryan: I added a test to `packages/angular/test/base/e2e/src/lazy/form.spec.ts` that compares both states and validates them as expected. The test fails if you comment out the suggested changes AND if you comment out the watch decorator for the disabled state in line 62 of `button.tsx`. It then works if you leave the watch decorator commented out but apply the changes OR if you add the decorator back in and also apply the changes. It seemed like the watch decorator was firing successfully always in the test environment, and as a consequence, that the test is at least good enough way to verify that if the watch decorator doesn't fire successfully, the fix should catch it.

I had a hard time replicating the race condition and wasn't sure if you wanted form.component.ts to be updated as well with ngOnInit (or else, a separate test component and ts file for this particular issue). When I was experimenting with other forms of testing and trying to use more of the `Promise.resolve().then(...)` in `ngOnInit` (and a couple of other things I tried to do with https://angular.dev/api/forms/AsyncValidatorFn), I kept getting timeout errors in the test. Let me know if you want want me to keep at your original suggestion and I will take another look.

core/src/components/button/button.tsx
Zac-Smucker-Bryan: I added a another line of explanatory text with the latest round of commits.

core/src/components/button/test/form-reference/button.e2e.ts
thetaPC: Thanks for digging into this so thoroughly! Your experiments confirmed what the mechanism suggests: the `disabled` change is masked by `@Watch('disabled')` in any environment where the watch fires, so the only thing that exposes the bug is the rare init-timing fluke, and that is not something we can trigger reliably. Your timeouts with `ngOnInit` and `AsyncValidatorFn` are that flakiness showing up.

When a regression is this hard to reproduce reliably, our approach is to skip the automated test rather than ship a flaky or misleading one. Instead we create a dev build and verify the fix directly against the reproduction from the issue. A test that passes whether or not the fix is present gives us false confidence, so we would rather have no test and a confirmed manual verification on the original repro. So please remove the disabled test.

One note on the dev build: only an Ionic team member can create one, so it isn't something you can run on your end. I'll see if I can get one going, though since this PR is on a fork it might not work, and I'll follow up here either way.

The `type` sync is a different case. There is no `@Watch('type')`, so the line you added is the only thing keeping the hidden button in sync, which means we can test it reliably. Please add a small e2e test that creates an `ion-button` with `type="submit"` inside a form, changes `type` to `"reset"` after load, and asserts the hidden form button reflects the new type. That test fails without the fix and passes with it, so it is worth keeping.

So the split is: no test for `disabled` (verified manually on a dev build instead), one deterministic test for `type`.

packages/angular/test/base/e2e/src/lazy/form.spec.ts
```diff
@@ -282,7 +282,52 @@ test.describe('Form', () => {
 
       await expect(control).toHaveClass(/ng-invalid/);
     });
-  });
+
+    test('should keep hidden submit button disabled state in sync', async ({ page }) => {
```
thetaPC: https://github.com/ionic-team/ionic-framework/pull/31225#discussion_r3463402248

thetaPC (changes requested):
Two small description fixes.

Title: the PR now syncs both the disabled state and the type, but the title only mentions disabled. Please update it to cover both, for example:

`fix(button): sync disabled state and type in renderHiddenButton`

Issue linking: the `resolves` line still won't auto-close the issue on merge. It's currently `resolves #[30968](...)`, and GitHub only treats it as a closing reference when the keyword is followed directly by `#` and the number. Please change it to:

`Issue number: resolves #30968`

GitHub turns that into a link on its own, so the markdown link isn't needed.

One process note for next time. The commit history here is hard to follow because it has several `revert:` commits that undo and redo the same changes, plus a merge commit from pulling the branch into itself. When you want to undo something you just committed, it's cleaner to manually remove that change and commit the corrected state, rather than adding a `revert:` on top, since each revert leaves both the original and the undo in the log and muddies the history.

It's not blocking and the final state is what matters once this is squashed, but a clean history makes the PR much easier to review as we go.

packages/angular/test/base/e2e/src/lazy/form.spec.ts
Zac-Smucker-Bryan: Thanks for the feedback @thetaPC. I will make my commits much more clean for next time. I appreciate the review. I believe I have the PR written the way it should be per your comments. So marking this resolved for now.

core/src/components/button/test/form-reference/button.e2e.ts
Zac-Smucker-Bryan: Thanks @thetaPC. I removed the test for `disabled` and added one for `type`. I verified the test fails when formButtonEl.type = this.type is removed and passes when it is present. Browser logging confirmed that renderHiddenButton() re-runs after the button type changes and enters the existing-hidden-button sync branch with this.type === 'reset'. Without the fix, the hidden native form button remains "submit" and the test times out waiting for "reset". With the fix, the hidden button updates to "reset" and the test passes. Let me know if anything else needs to change on this.

packages/angular/test/base/e2e/src/lazy/form.spec.ts
```diff
@@ -282,7 +282,33 @@ test.describe('Form', () => {
 
       await expect(control).toHaveClass(/ng-invalid/);
     });
-  });
+
+    test('should keep hidden button type in sync with visible button', async ({ page }) => {
```
thetaPC: The `type` test is solid and the logic is exactly right. Nice call leaving the `testInfo.annotations` off it too. That's correct: the issue annotation links to #30968, which is specifically the `disabled` regression, and there's no linked issue for the `type` behavior, so the test shouldn't have one.

One change: please move it out of the Angular lazy-form suite and into core e2e at `core/src/components/button/test/form-reference/button.e2e.ts`, inside the existing `button: form` describe block, and remove the version in `form.spec.ts`.

It doesn't belong in the Angular suite because this isn't a framework-specific problem. The `type` sync lives in the core `ion-button` component, and the test just sets a property and reads the DOM, so we can verify it entirely from within core with no Angular involved. Keeping it in core also means it runs in the core pipeline rather than only under the Angular tests. Have a look at the existing tests in that file for the `setContent` and `config` pattern to follow.

thetaPC (dismissed):

packages/angular/test/base/e2e/src/lazy/form.spec.ts
Zac-Smucker-Bryan: I removed the test from `form.spect.ts` and moved it to `core/src/components/button/test/form-reference/button.e2e.ts`, matching the conventions of the other tests for creating the form with `setContent` and `config`. Tests fail without the fix and pass with them.

packages/angular/test/base/e2e/src/lazy/form.spec.ts
```diff
@@ -282,7 +282,8 @@ test.describe('Form', () => {
 
       await expect(control).toHaveClass(/ng-invalid/);
     });
-  });
+
+    });
```
ShaneK:
```suggestion
  });
```
This file shouldn't be in the diff at all now that the `type` test moved to core `button.e2e.ts`. Removing the test here left a stray blank line and re-indented the closing `});` to 4 spaces. Could you revert `form.spec.ts` back to its original state so the PR only touches `button.tsx` and the core test?

ShaneK (changes requested):
A fairly minor nit, but after that I think this will be good to go!

packages/angular/test/base/e2e/src/lazy/form.spec.ts
Zac-Smucker-Bryan: Thanks for catching that. Let me know if you need anything else!

ShaneK (approved):
Awesome work 🎉 Thanks!

PR #31230, opened 2026-06-17
fix(angular): honor modifier-click on routerLink

### LINKED ISSUE #26394: feat: Have the feature of "href" and "router-link" in a unified way for web and mobile apps

ShaneK:
Issue number: resolves #26394

---------

When `routerLink` is applied to a non-anchor Ionic component (`ion-item`, `ion-button`), `RouterLinkDelegateDirective` always routes in-app on click. A ctrl/meta/shift/alt click, or a host with `target="_blank"`, gets swallowed by the in-app navigation, so the browser never opens the link in a new tab. A plain `<a routerLink>` honors these intents, but the Ionic delegate doesn't, even though these components render a native anchor in their shadow DOM that's capable of the native new-tab behavior.

## What is the new behavior?

`RouterLinkDelegateDirective` now adds a capture-phase click listener so it runs before Angular's `RouterLink` handler and the directive's own bubble-phase `onClick`. When the click should be handled natively (a ctrl/meta/shift/alt modifier is held, or the host `target` is set to anything other than `_self`), it calls `stopImmediatePropagation()` to cancel the in-app navigation but leaves `preventDefault` alone, so the shadow-DOM anchor can still open the tab. This mirrors the modifier set Angular's own `RouterLink` guards on. Normal clicks are unaffected and continue to navigate in-app. The listener is removed in `ngOnDestroy`.

- [ ] Yes
- [x] No

## Other information

This supersedes #28976 by @j-oppenhuis, which first surfaced the target and modifier-click gap. Credited as co-author below.

To verify manually in the Angular preview, ctrl/cmd-click the two `ion-item` rows (a plain one and a `target="_blank"` one) and confirm each opens a new tab instead of navigating in place:
https://ionic-framework-git-fix-angular-routerlink-modifi-896c58-ionic1.vercel.app/angular/standalone/router-link

If you compare that against the main version of the same component, you'll be able to see how it's fixed:
https://ionic-framework-git-main-ionic1.vercel.app/angular/standalone/router-link

Co-authored-by: Jelle Oppenhuis <jelle@siyou.nl>

### DIFF
--- a/packages/angular/test/base/e2e/src/standalone/router-link-modifier-click.spec.ts
+++ b/packages/angular/test/base/e2e/src/standalone/router-link-modifier-click.spec.ts
@@ -1,0 +1,73 @@
+import { test, expect, type Page } from '@playwright/test';
+
+/**
+ * Issue #26394: a modifier click (ctrl/meta/shift) or a non-`_self` target on a
+ * `routerLink` over a non-anchor Ionic component (ion-item, ion-button) must
+ * open a new tab instead of navigating the current page in-app.
+ *
+ * Playwright headless can't dispatch a real modifier-click or observe the
+ * browser's native new-tab default, so we dispatch a synthetic click on the
+ * host (not the shadow anchor, which would fire its own default navigation) and
+ * assert the JS handler chain leaves the page unchanged with `preventDefault`
+ * uncalled.
+ */
+test.describe('RouterLink: modifier click', () => {
+  const dispatchClick = (page: Page, selector: string, modifiers: Record<string, boolean>) =>
+    page.evaluate(
+      ({ selector, mods }: { selector: string; mods: Record<string, boolean> }) => {
+        const item = document.querySelector(selector) as HTMLElement;
+        const ev = new MouseEvent('click', { bubbles: true, composed: true, cancelable: true, button: 0, ...mods });
+        item.dispatchEvent(ev);
+        return { defaultPrevented: ev.defaultPrevented };
+      },
+      { selector, mods: modifiers }
+    );
+
+  test.beforeEach(async ({ page }) => {
+    await page.goto('/standalone/router-link');
+  });
+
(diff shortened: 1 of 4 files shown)


packages/angular/test/base/e2e/src/standalone/router-link-modifier-click.spec.ts
thetaPC, line 1: Should it also check for `OPTION` to verify that it behaves as intended since it was [brought up](https://github.com/ionic-team/ionic-framework/pull/28976#pullrequestreview-1868902765) in the original conversation?

thetaPC (approved):
LGTM, please provide co-author for the associated PR

packages/angular/test/base/e2e/src/standalone/router-link-modifier-click.spec.ts
ShaneK: [d86e78c](https://github.com/ionic-team/ionic-framework/pull/31230/commits/d86e78cd8b1529e54f1933248ed32f61847eed8b)

Co-author has been added to the body so it'll be attributed when squash + merged

PR #31235, opened 2026-06-18
fix(angular): run change detection on attached overlay views

### LINKED ISSUE #31220: bug: Angular: Modal does not open when using ModalController and the component declares an InputSignal with the root
### Prerequisites

### Ionic Framework Version

v8.x

### Current Behavior

When using ionic/angular: ^8.8. When we use ModalController with a dynamic root input that is given to the nav. It will open an empty nav.

### Expected Behavior

When ModalController opens a component that contains an ion-nav, and that nav's root is set dynamically via an InputSignal, the nav should render the root component.

### Steps to Reproduce

1. Go to https://stackblitz.com/edit/kodxyzx4?file=src%2Fmain.ts
2. Click on open modal. (This opens with an empty nav)
3. Refresh
4. Click on 'Open nav with if' (This opens with a filled nav)

### Code Reproduction URL

https://stackblitz.com/edit/kodxyzx4?file=src%2Fmain.ts

### Ionic Info

Ionic:

Ionic CLI : 7.2.1
Ionic Framework : @ionic/angular 8.8.10

Capacitor:

Capacitor CLI : 8.4.0
@capacitor/android : 8.4.0
@capacitor/core : 8.4.0
@capacitor/ios : 8.4.0

Utility:

native-run : 2.0.3

System:

NodeJS : v22.18.0
npm : 10.9.3
OS : macOS Unknown

### Additional Information

_No response_

ShaneK:
Issue number: resolves #31220

---------

Currently, a component opened through `ModalController` that binds a value from `componentProps` in its template, like `<ion-nav [root]="rootPage()">`, can render empty. `attachView` only runs the creation pass, so the `[root]` binding is applied on a later tick, after `ion-nav` has already finished loading and read `root` as undefined. Wrapping the nav in `@if (rootPage())` worked around it.

## What is the new behavior?

After this fix, `attachView` runs `detectChanges()` on the view right after attaching it, so template bindings apply before the web component loads. `ion-nav` now sees its `root` during load and renders it. This matches the `detectChanges` call already on `major-9.0`.

- [ ] Yes
- [X] No

## Other information

The actual fix for this PR is fairly small, most of it is just testing. It's also important to note that this is already fixed in v9 in the same exact way, we're just backporting it to v8 here.

Dev build:
```
8.8.12-dev.11781813448.16e7a064
```

### DIFF
--- a/packages/angular/common/src/providers/angular-delegate.ts
+++ b/packages/angular/common/src/providers/angular-delegate.ts
@@ -233,6 +233,14 @@ export const attachView = (
 
   applicationRef.attachView(componentRef.hostView);
 
+  /**
+   * Run change detection so template bindings (e.g. `<ion-nav [root]="rootPage">`)
+   * apply during this synchronous pass, before the web component runs its load
+   * lifecycle. `createComponent` only runs the creation pass, so without this the
+   * binding lands after the element has loaded, too late for `ion-nav` to read it.
+   */
+  componentRef.changeDetectorRef.detectChanges();
+
   elRefMap.set(hostElement, componentRef);
   elEventsMap.set(hostElement, unbindEvents);
   return hostElement;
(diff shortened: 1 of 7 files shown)


thetaPC (approved):
LGTM

PR #31241, opened 2026-06-23
feat(select): add labelPlacement prop, justify prop, and rich content for select options

### LINKED ISSUE #29890: feat: allow slot on ion-select-option

thetaPC:
Issue number: resolves #29890 

---------

Select only allows plain text to be passed within it's options.

## What is the new behavior?

**`ion-select-option`**
- Added default slot support for custom HTML content (images, avatars, icons, etc.) when `innerHTMLTemplatesEnabled` is `true`
- Added `start` and `end` named slots for content that appears in the overlay interface but not in the selected text
- Added `description` prop for text rendered below the option label in the overlay
- Add `labelPlacement` and `justify` props to `ion-select-option` and pass them through to alert, popover, and modal overlay paths
- Add per-theme slot-size limitations so oversized slotted content is capped to keep option rows visually uniform within an overlay

**`ion-select`**
- Selected text renders HTML content from the default slot, excluding `start`/`end` slot content
- `aria-label` derives plain text only from the default slot, ensuring screen readers don't read slotted or element content
- Added CSS variables for styling media within the selected text (`--select-text-media-width`, `--select-text-media-height`, etc.)

**Overlay interfaces (`alert`, `action-sheet`, `select-popover`, `select-modal`)**
- All four interfaces render rich content (`start`, `end`, `description`, HTML label) consistently via the shared `renderOptionLabel` utility
- Public interfaces (`ActionSheetButton`, `AlertInput`, `SelectPopoverOption`, `SelectModalOption`) are unchanged — rich content fields are on internal extended interfaces (`SelectActionSheetButton`, `SelectAlertInput`, `SelectOverlayOption`)
- Content is sanitized via `sanitizeDOMString` before DOM injection to prevent XSS

**Utilities**
- Added `select-option-render.tsx`: shared render utility for option labels across all overlay components
- Added `getOptionContent`: extracts and clones slot content from `ion-select-option` for overlays and selected text
- Added `getDefaultSlotPlainText` — extracts plain text from the default slot, used for labels and aria
- Add `reflectPropertiesToAttributes` to `core/src/utils/sanitization/` and call it from `getOptionContent` immediately before `cloneNode`. The helper mirrors a registered set of custom-element DOM properties (`icon`, `name`, `src`, `ios`, `md` on `ion-icon`) onto attributes so the cloned overlay copy renders correctly regardless of how the framework bound the prop.

**Tests**
- Added E2E tests for rich content rendering across all four interfaces (single and multiple selection)
- Added tests for slot positioning (start/end placement verification)
- Added tests for selected text content verification (excludes slotted content)

- [ ] Yes
- [x] No

No, developers will be able to continue using plain text for select options as usual.

## Other information

Dev build: `8.8.12-dev.11782342444.186b6e9c`

[Preview](https://ionic-framework-git-fw-7599-ionic1.vercel.app/src/components/select/test/rich-content-option/)


core/src/utils/sanitization/index.ts
```diff
@@ -175,8 +255,183 @@ const isSanitizerEnabled = (): boolean => {
...
+ * `aria-*` and `data-*` are allowed separately by prefix (refer
+ * `allowedAttributePrefixes`) since they are inert and not worth
+ * enumerating. URL-bearing names (`href`, `src`) are allowed here, but
+ * their values are still scrubbed for script schemes in `sanitizeElement`.
+ *
+ * Notably absent: `style`, event handlers (`on*`), and the
+ * form/navigation-hijack attributes (`formaction`, `action`, `target`),
+ * which are therefore stripped.
+ */
+const allowedAttributes = [
```
ShaneK: This `allowedAttributes` array is shared with `sanitizeDOMString`, so growing it from 6 entries to ~70 (plus `aria-*`/`data-*` wholesale) changes the sanitization output for every existing consumer, not just the new select path. `ion-toast`, `ion-loading`, the `ion-alert` message, `ion-refresher-content`, and `ion-infinite-scroll-content` all run innerHTML through this when `innerHTMLTemplatesEnabled` is on, so names like `type`, `value`, `width`, `height`, `mode`, and `theme` now survive where they were stripped before. The dangerous-vector handling is actually tighter than before so I don't think this is unsafe, but is the wider allowlist intended for those other components too? Worth a regression test covering one of them, and a line in the PR description so anyone bisecting a future toast/loading rendering difference knows where to look.

core/src/utils/select-option-render.tsx
```diff
@@ -0,0 +1,163 @@
...
+ * @param id - Unique identifier for generating stable virtual DOM keys.
+ * @param content - The HTMLElement container whose child nodes will be cloned.
+ * @param className - CSS class applied to the wrapper element.
+ * @param useSpan - Whether to use a span element instead of a div for the wrapper.
+ */
+const renderClonedContent = (id: string, content: HTMLElement, className: string, useSpan = false) => {
+  const Tag = useSpan ? 'span' : 'div';
+  const keyPrefix = `${className}-${id}`;
+
+  sanitizeDOMTree(content);
```
ShaneK, line 97: `renderClonedContent` calls `sanitizeDOMTree(content)` here, and the new "should sanitize an HTMLElement label that bypassed ion-select" spec shows this is the only sanitization pass for vanilla-JS callers that hand an `HTMLElement` straight to `renderOptionLabel` without going through `getOptionContent`. But the `cloneToVNode` comment just above says "pure structural conversion, no security filtering" and "Highly recommended to pre-sanitize the source DOM", which reads like sanitization always happens upstream. Someone could delete this call as redundant and reopen an XSS hole on the direct-`HTMLElement` path. Could you add a short note above it explaining that this is the only sanitize pass for direct `HTMLElement` callers and shouldn't be removed?

core/src/components/select-option/select-option.tsx
```diff
@@ -23,6 +30,43 @@ export class SelectOption implements ComponentInterface {
    */
   @Prop() value?: any | null;
```
ShaneK, line 31: The `disabled` JSDoc just above (`This property does not apply when interface="action-sheet" as ion-action-sheet does not allow for disabled buttons.`) is now stale. This PR wires `option.disabled` through to the action-sheet button: `createActionSheetButtons` sets it, the button renders `disabled={b.disabled}`, there's new disabled styling in `action-sheet.scss`, and an e2e test covers it. The comment says the opposite of what ships now. Can you update it?

ShaneK (changes requested):
Looking really good! Fairly minor change requests with one intention verification request

core/src/utils/sanitization/index.ts
thetaPC: https://github.com/ionic-team/ionic-framework/pull/31241/changes/7f7d86702cddd72db874d020e3ba7efff5846e01

core/src/utils/select-option-render.tsx
thetaPC: https://github.com/ionic-team/ionic-framework/pull/31241/changes/0426878b5c18df1e3f25563aecfba78fd538adfb

core/src/components/select-option/select-option.tsx
thetaPC: https://github.com/ionic-team/ionic-framework/pull/31241/changes/f4d45a006e05d2a6972cafc75abe3b4f59b595b7

ShaneK (approved):
LGTM! Great work 🚀

PR #31245, opened 2026-06-25
fix(sanitization): scope allowlist to the rich content path

thetaPC:
Issue number: N/A

---------

The attribute allowlist in `core/src/utils/sanitization` is shared between `sanitizeDOMString` and `sanitizeDOMTree`. The rich content work broadened it from 6 attributes to the full component and SVG set plus `aria-*` and `data-*`. Because `sanitizeDOMString` is also used by `ion-toast`, `ion-loading`, the `ion-alert` message, `ion-refresher-content`, and `ion-infinite-scroll-content`, those components keep attributes that were previously stripped when `innerHTMLTemplatesEnabled` is on.

`renderClonedContent` is the only place that sanitizes content for callers that pass an `HTMLElement` straight to `renderOptionLabel`, but the surrounding `cloneToVNode` comment implies sanitization always happens upstream, so the call looks redundant and could be removed.

The `ion-select-option` `disabled` JSDoc states the property does not apply to `interface="action-sheet"`, which is no longer accurate.

## What is the new behavior?

- Scope the broadened allowlist to `sanitizeDOMTree` (the select rich content path). `sanitizeDOMString` reverts to its original narrow allowlist, so sanitization output is unchanged for toast, loading, the alert message, refresher content, and infinite scroll content. The shared dangerous vector scrubbing (style removal, script scheme checks, `data:` URI filtering) still applies to both paths.
- Add a regression test asserting `sanitizeDOMString` strips attributes that are only allowed on the rich content path.
- Add a note above the `sanitizeDOMTree` call in `renderClonedContent` explaining it is the only sanitization pass for direct `HTMLElement` callers and must not be removed.
- Remove the stale action sheet note from the `ion-select-option` `disabled` prop JSDoc, since action sheet options now honor `disabled`.

- [ ] Yes
- [x] No

## Other information

Follow-up to the select rich content work ([FW-7599](https://github.com/ionic-team/ionic-framework/pull/31241)) addressing review feedback. The allowlist change restores the prior sanitization behavior for the non-select consumers rather than changing it, so there is no user facing behavior change for those components.

[FW-7599]: https://outsystemsrd.atlassian.net/browse/FW-7599?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


ShaneK (approved):

Issue #30266, opened 2025-03-18
bug: ion-datetime auto-selecting date when month/year changed (more then one month) when you can do multiple dates


beckyhilton:
### Prerequisites

- [x] I have read the [Contributing Guidelines](https://github.com/ionic-team/ionic-framework/blob/main/docs/CONTRIBUTING.md#creating-an-issue).
- [x] I agree to follow the [Code of Conduct](https://ionicframework.com/code-of-conduct).
- [x] I have searched for [existing issues](https://github.com/ionic-team/ionic-framework/issues) that already report this problem, without success.

### Ionic Framework Version

v8.x

### Current Behavior

https://www.loom.com/share/4fd796af5d4d4250942a7a8cc85fb3b2?sid=0ae84e29-8334-417b-a5d3-6475e9191449

When changing the month & year (not using the arrows) it automatically selects a date in the new month

### Expected Behavior

Nothing should be automatically selected and it should look like when you first come to a blank calendar

### Steps to Reproduce

1. Go to a page with the ion-datetime displayed as a month with multi-select
2. Select the Month/Year in the upper left hand corner
3. Change to a different month and select the Month/Year at the top to close
4. The new month is displayed and a date is automatically selected

### Code Reproduction URL

https://ionicframework.com/docs/api/datetime

### Ionic Info

Ionic:

   Ionic CLI                     : 7.2.0 (/usr/local/lib/node_modules/@ionic/cli)
   Ionic Framework               : @ionic/angular 8.4.3
   @angular-devkit/build-angular : 17.3.11
   @angular-devkit/schematics    : 17.3.11
   @angular/cli                  : 17.3.11
   @ionic/angular-toolkit        : 11.0.1

Capacitor:

   Capacitor CLI      : 7.0.1
   @capacitor/android : 7.0.1
   @capacitor/core    : 7.0.1
   @capacitor/ios     : 7.0.1

perrygovier:
I believe this issue should be closed now as #31248 resolves it

thetaPC:
Thanks for reporting this! This is a duplicate of #29673, which was fixed in [v8.8.13](https://github.com/ionic-team/ionic-framework/releases/tag/v8.8.13), so I am closing it in favor of that issue.

PR #31248, opened 2026-06-26
fix(datetime): don't setActiveParts from month/year wheels when multiple=true

### LINKED ISSUE #29673: bug: ion-datetime, error when selecting the year and month when [multiple] is true
### Prerequisites

### Ionic Framework Version

v7.x, v8.x

### Current Behavior

If you switch to the year and month selection mode, then when the year and month are alternately changed, the value of the month or year is reset, respectively, to a value equal to the current date

https://github.com/ionic-team/ionic-framework/assets/44109078/a05226eb-053e-446c-9bbd-b833aeffd004

### Expected Behavior

The values of the year and month should change independently of each other. Changing the value of the year should not affect the change in the value of the month and vice versa

### Steps to Reproduce

1. Open the example [Stackblitz url](https://stackblitz.com/run?file=src%2Fapp%2Fexample.component.ts) of the ion-datetime.
2. Switch to the year and month selection mode 
image
3. Alternately change the year and month in this window

### Code Reproduction URL

https://stackblitz.com/run?file=src%2Fapp%2Fexample.component.ts

### Ionic Info

[@angular/animations](https://www.npmjs.com/package/@angular/animations)
17.0.1

[@angular/common](https://www.npmjs.com/package/@angular/common)
17.0.1

[@angular/compiler](https://www.npmjs.com/package/@angular/compiler)
17.0.1

[@angular/core](https://www.npmjs.com/package/@angular/core)
17.0.1

[@angular/forms](https://www.npmjs.com/package/@angular/forms)
17.0.1

[@angular/platform-browser](https://www.npmjs.com/package/@angular/platform-browser)
17.0.1

[@angular/platform-browser-dynamic](https://www.npmjs.com/package/@angular/platform-browser-dynamic)
18.0.1

[@angular/router](https://www.npmjs.com/package/@angular/router)
17.0.1

[@ionic/angular](https://www.npmjs.com/package/@ionic/angular)
8.2.2

[@ionic/core](https://www.npmjs.com/package/@ionic/core)
8.2.2

[rxjs](https://www.npmjs.com/package/rxjs)
7.8.1

[tslib](https://www.npmjs.com/package/tslib)
2.6.2

[zone.js](https://www.npmjs.com/package/zone.js)
0.14.2

### Additional Information

_No response_

droc101:
Issue number: resolves #29673

---------

Using the month/year selection wheels on an `ion-datetime` with `multiple=true` changes the selection in an unintuitive manner and changing the value of one of the wheels often reverts the value of the other wheel.

## What is the new behavior?
- The month and year wheels in an `ion-datetime` no longer change the selection when `multiple=true`
- The behavior of a single-select `ion-datetime` is unchanged.
- The fix has only been applied to `renderMonthPickerColumn` and `renderYearPickerColumn`, as [per the docs](https://ionicframework.com/docs/api/datetime#multiple-date-selection), the `multiple` property is only supported on `presentation=date preferWheel=false`, which only shows those two wheels.
- Four new datetime e2e tests have been added to test these changes

- [ ] Yes
- [x] No


core/src/components/datetime/test/basic/datetime.e2e.ts
```diff
@@ -816,3 +816,121 @@ configs({ modes: ['md'], directions: ['ltr'] }).forEach(({ title, config }) => {
     });
   });
 });
+
+configs({ modes: ['md'], directions: ['ltr'] }).forEach(({ title, config }) => {
+  test.describe(title("datetime: interaction between month/year wheels and selected date(s) with different values of multiple"), () => {
```
ShaneK: These four cover the `multiple` prop's interaction with the month/year wheels, which isn't really default behavior. The testing best-practices doc says `basic/` is for default behavior and that tests should be broken up per-feature, closest to the feature. There's already a `test/multiple/datetime.e2e.ts` with a "multiple date selection (functionality)" block. Could these move there instead? Easier for the next person to find.

core/src/components/datetime/test/basic/datetime.e2e.ts
```diff
@@ -816,3 +816,121 @@ configs({ modes: ['md'], directions: ['ltr'] }).forEach(({ title, config }) => {
...
+        type: 'issue',
+        description: 'https://github.com/ionic-team/ionic-framework/issues/29673',
+      });
+
+      await page.setContent(
+        `
+        <ion-datetime multiple></ion-datetime>
+        <script>
+          const datetime = document.querySelector('ion-datetime');
+          datetime.value = [ "2026-06-26T10:36:00", "2026-06-28T10:36:00", "2026-01-1T10:36:00" ];
```
ShaneK: `"2026-01-1T10:36:00"` has an unpadded day, which isn't valid ISO 8601. It only round-trips here because `multiple=true` skips `setActiveParts`, so the value never gets re-normalized. I'd pad it to `"2026-01-01T10:36:00"` in both the input arrays and the assertions so it doesn't read as an intentional edge case.

core/src/components/datetime/test/basic/datetime.e2e.ts
```diff
@@ -816,3 +816,121 @@ configs({ modes: ['md'], directions: ['ltr'] }).forEach(({ title, config }) => {
...
+      await page.setContent(`<ion-datetime value="2026-06-26T10:36:00"></ion-datetime>`, config);
+      const datetime = page.locator('ion-datetime');
+
+      await page.waitForSelector('.datetime-ready');
+
+      const dropdownButton = page.locator('.calendar-month-year-toggle');
+      await dropdownButton.click();
+      await page.waitForChanges();
+
+      const monthOption = page.locator('.year-column ion-picker-column-option').nth(0);
```
ShaneK:
```suggestion
      const yearOption = page.locator('.year-column ion-picker-column-option').nth(0);
```
The selector's right, but the variable's named `monthOption`, looks like a copy/paste leftover from the month test. Same thing in the other year-wheel test below.

core/src/components/datetime/datetime.tsx
```diff
@@ -1963,10 +1963,12 @@ export class Datetime implements ComponentInterface {
             month: ev.detail.value,
           });
 
-          this.setActiveParts({
-            ...activePart,
-            month: ev.detail.value,
-          });
+          if (!this.multiple) {
```
ShaneK, line 1967: Could you add a one-line comment on why active parts get skipped in `multiple` mode? It's not obvious that the wheels are navigation-only here and that touching active parts would corrupt the selection (#29673), especially with `setWorkingParts` running unconditionally right above. There's similar `multiple`-conditional logic commented in the calendar-grid branch for reference. Same applies to the year guard.

core/src/components/datetime/test/basic/datetime.e2e.ts
```diff
@@ -816,3 +816,121 @@ configs({ modes: ['md'], directions: ['ltr'] }).forEach(({ title, config }) => {
...
+
+configs({ modes: ['md'], directions: ['ltr'] }).forEach(({ title, config }) => {
+  test.describe(title("datetime: interaction between month/year wheels and selected date(s) with different values of multiple"), () => {
+    test('using month wheel should change selection when multiple=false', async ({page}, testInfo) => {
+      testInfo.annotations.push({
+        type: 'issue',
+        description: 'https://github.com/ionic-team/ionic-framework/issues/29673',
+      });
+
+      await page.setContent(`<ion-datetime value="2026-06-26T10:36:00"></ion-datetime>`, config);
```
ShaneK: Nit: the testing best-practices doc calls out adding `locale="en-US"` to `ion-datetime` test instances specifically, and the existing `basic/` tests do it. These assert on the ISO `value` rather than rendered labels so they pass either way, but worth matching the convention. Up to you!

ShaneK (changes requested):
Awesome, straightforward fix. Just a few change requests

ShaneK (approved):
LGTM, nice work 🎉

perrygovier:
Note, this closes #30266

PR #31260, opened 2026-07-01
fix(modal): focus the dialog wrapper on present so screen readers can enter the modal

gnbm:
Issue number: internal

---------

When a modal is opened with Android TalkBack enabled, the screen reader cannot enter or navigate the modal's content. On present, focus is moved to the modal host, but the host is role-less. The `role="dialog"`, `aria-modal`, and the label live on the inner `.ion-overlay-wrapper` so assistive technologies have no dialog to land on.

## What is the new behavior?
- `present()` now focuses the overlay's `[role="dialog"]` element (the wrapper) instead of the role-less host, falling back to the host when the role is on the host (action-sheet, loading).
- The modal wrapper gets `tabIndex="-1"` so it can receive that programmatic focus (an element with only `role="dialog"` is not focusable). Its focus ring is suppressed since the focus is programmatic, not keyboard-driven.
- Clears text selection when gestures start to prevent Firefox from interrupting swipe-to-close and sheet gestures.
- Adds an e2e test to `modal/test/a11y/modal.e2e.ts` to verify no visible outline added to the wrapper.
- Adds e2e tests to `utils/test/overlays/overlays.e2e.ts` to verify the focus is on the wrapper instead of the host for all modals.
- Fixes Playwright test helper to clear text selection immediately after `mouse.down()` to prevent the selection from blocking gesture detection in the test environment.


core/src/components/modal/modal.tsx
```diff
@@ -1644,10 +1644,23 @@ export class Modal implements ComponentInterface, OverlayInterface {
...
+            host) the thing that `present()` in overlays.ts focuses when
+            the modal opens, so assistive technologies get a focus target
+            that actually carries the dialog's role/label. It is not part
+            of the Tab order since -1 is never Tab-reachable.
+
+            This is only applied to default modals. For sheet and iOS card
+            modals the wrapper doubles as the drag-gesture surface, and
+            leaving focus on that surface interferes with the pointer
+            gesture (most visibly on Firefox). Those modals keep focusing
+            the host as before. See FW-7611.
```
gnbm:
```suggestion
            the host as before. 
```

core/src/components/modal/test/a11y/modal.e2e.ts
```diff
@@ -19,5 +19,22 @@ configs({ modes: ['ios'], directions: ['ltr'] }).forEach(({ title, config }) =>
       const results = await new AxeBuilder({ page }).analyze();
       expect(results.violations).toEqual([]);
     });
+
+    // role="dialog" lives on .modal-wrapper, not the host, so focus must land
+    // there on present for screen readers (e.g. TalkBack) to enter the dialog
+    // (IONIC-91 / FW-7611).
+    test('should move focus to the dialog wrapper, not the role-less host, on present', async ({ page }) => {
+      await page.goto(`/src/components/modal/test/a11y`, config);
```
brandyscarney: We should use `testInfo.annotations` to link an issue.

```suggestion
    // Focus the wrapper with role="dialog" on present so screen readers
    // (e.g. TalkBack) enter the dialog.
    test('should focus the modal wrapper on present', async ({ page }, testInfo) => {
      testInfo.annotations.push({
        type: 'issue',
        description: 'FW-7611',
      });
      await page.goto(`/src/components/modal/test/a11y`, config);
```

core/src/components/modal/modal.tsx
```diff
@@ -1644,10 +1644,23 @@ export class Modal implements ComponentInterface, OverlayInterface {
...
+            host) the thing that `present()` in overlays.ts focuses when
+            the modal opens, so assistive technologies get a focus target
+            that actually carries the dialog's role/label. It is not part
+            of the Tab order since -1 is never Tab-reachable.
+
+            This is only applied to default modals. For sheet and iOS card
+            modals the wrapper doubles as the drag-gesture surface, and
+            leaving focus on that surface interferes with the pointer
+            gesture (most visibly on Firefox). Those modals keep focusing
+            the host as before.
```
brandyscarney: This comment is a bit verbose, could we say something like this:

```suggestion
            tabIndex={-1} makes this element (rather than the role-less host) 
            receive focus when present() is called, so assistive technologies 
            get a proper focus target. This only applies to default modals. 
            Sheet and card modals continue to focus the host to preserve 
            gesture handling.
```

core/src/components/modal/modal.tsx
```diff
@@ -1644,10 +1644,23 @@ export class Modal implements ComponentInterface, OverlayInterface {
...
+            This is only applied to default modals. For sheet and iOS card
+            modals the wrapper doubles as the drag-gesture surface, and
+            leaving focus on that surface interferes with the pointer
+            gesture (most visibly on Firefox). Those modals keep focusing
+            the host as before. 
           */
           role="dialog"
           {...inheritedAttributes}
           aria-modal="true"
+          tabIndex={!isCardModal && !isSheetModal ? -1 : undefined}
```
brandyscarney: I don't think excluding card and sheet modals is the right solution here. I can see it working for the default modal but I have no way to get to the input inside of a sheet modal:

https://github.com/user-attachments/assets/c37dbeb9-30f3-405e-9e87-e3c69ff5d05e

gnbm: Removed the exclusion — the wrapper now carries `tabindex="-1"` for all modal types (default, sheet, and card), so `present()` moves focus to the element that declares `role="dialog"` for every modal, and a screen reader can enter the content.

Why sheet/card were excluded originally: making the wrapper focusable made the `sheet modal: drag events` e2e hang on Firefox. Root cause is a Gecko quirk — an element with `tabindex` is treated as a selection root, so a pointer press over a text caret inside it can begin a native drag-and-drop session (`dragstart` → `pointercancel`) that swallows the pointer stream, and the gesture never receives its final `pointerup`. Real user gestures weren't affected (the failing sequence requires pressing an already-pressed button), but it's a real footgun. The sheet and swipe-to-close gestures now cancel any native `dragstart` while a drag is active; native drag-and-drop is untouched when the modal isn't being dragged.

Tests added:
- Wrapper (with `role="dialog"`) is focused on present for default, sheet, and card modals.
- Native drag-and-drop is cancelled only while the gesture is active, for both sheet and card.
- The focused wrapper doesn't render a native focus ring when the modal is opened via keyboard (`:focus-visible`), matching the host.
- `sheet modal: drag events` now passes on Chromium, Firefox, and WebKit.

On whether we need a test for "can reach the input inside the sheet modal": I prototyped one (present a sheet modal with an input, then Tab) and it turned out to be redundant. With the old host-focus behavior the keyboard focus-trap already moved focus into the content and reached the input — so keyboard reachability was never the failure. The thing that actually changed is where initial/accessibility focus lands (the `role="dialog"` wrapper), which is exactly what the "wrapper is focused on present" tests assert. TalkBack navigation itself isn't drivable in Playwright, so those focus-target tests are the closest automatable proxy. Happy to add the input-reach test anyway if you'd prefer the explicit scenario documented.

Could you re-check the sheet modal (including the input) with TalkBack on a fresh dev build when you have a moment?

core/src/components/modal/modal.scss
```diff
@@ -84,6 +84,14 @@ ion-backdrop {
   z-index: 10;
 }
 
+/**
+ * The wrapper is focused on present (see modal.tsx); suppress its focus
+ * ring, matching the host above.
```
brandyscarney:
```suggestion
 * The wrapper receives programmatic focus for screen readers but should not
 * show a visible focus ring, which is meant only for keyboard navigation.
```

core/src/utils/overlays.ts
```diff
@@ -593,7 +593,15 @@ export const present = async <OverlayPresentOptions>(
...
-    overlay.el.focus();
+    /**
+     * Some overlays (e.g. modal) put the dialog role and label on the
+     * `.ion-overlay-wrapper` rather than the host. Screen readers like
+     * TalkBack need focus to land there, so focus the wrapper when it is
+     * focusable (has a `tabindex`); otherwise focus the host as before.
+     */
+    const overlayWrapper = getElementRoot(overlay.el).querySelector<HTMLElement>('.ion-overlay-wrapper');
+    const focusTarget = overlayWrapper?.hasAttribute('tabindex') ? overlayWrapper : overlay.el;
+    focusTarget.focus({ preventScroll: true });
```
brandyscarney:
```suggestion
    /**
     * Some overlays (e.g. modal) put the dialog role and label on the
     * `.ion-overlay-wrapper` rather than the host. Screen readers need
     * focus on the element with role="dialog" to properly announce and
     * navigate the dialog.
     */
    const overlayWrapper = getElementRoot(overlay.el).querySelector<HTMLElement>('[role="dialog"]');
    const focusTarget = overlayWrapper ?? overlay.el;
    focusTarget.focus({ preventScroll: true });
```

Why do we need to add `tabindex` at all? Can't we just query for `role="dialog"` if that is the only difference between the overlays?

core/src/components/modal/modal.tsx
```diff
@@ -1644,10 +1644,14 @@ export class Modal implements ComponentInterface, OverlayInterface {
...
             shadow DOM otherwise ion-button will not be highlighted
             when using VoiceOver: https://bugs.webkit.org/show_bug.cgi?id=247134
+
+            tabIndex={-1} lets present() focus this element (which carries
+            the dialog role) instead of the role-less host.
           */
           role="dialog"
           {...inheritedAttributes}
           aria-modal="true"
+          tabIndex={-1}
```
brandyscarney, line 1657: See my comment on `core/src/utils/overlays.ts`

```suggestion

```

core/src/components/modal/modal.tsx
```diff
@@ -1644,10 +1644,14 @@ export class Modal implements ComponentInterface, OverlayInterface {
             same element. They must also be set inside the
             shadow DOM otherwise ion-button will not be highlighted
             when using VoiceOver: https://bugs.webkit.org/show_bug.cgi?id=247134
+
+            tabIndex={-1} lets present() focus this element (which carries
+            the dialog role) instead of the role-less host.
```
brandyscarney: See my comment on `core/src/utils/overlays.ts`

```suggestion

```

core/src/components/modal/gestures/sheet.ts
```diff
@@ -298,7 +298,16 @@ export const createSheetGesture = (
     return true;
   };
 
+  /**
+   * In Firefox, pressing over a text caret in the focusable wrapper can
+   * start a native drag that swallows the gesture's pointer events.
+   * Cancel native drag while the gesture is active.
+   */
+  const preventNativeDragStart = (ev: DragEvent) => ev.preventDefault();
```
brandyscarney: How do I test this?

core/src/components/modal/gestures/swipe-to-close.ts
```diff
@@ -113,9 +113,18 @@ export const createSwipeToCloseGesture = (
     return false;
   };
 
+  /**
+   * In Firefox, pressing over a text caret in the focusable wrapper can
+   * start a native drag that swallows the gesture's pointer events.
+   * Cancel native drag while the gesture is active.
+   */
+  const preventNativeDragStart = (ev: DragEvent) => ev.preventDefault();
```
brandyscarney: How do I test this?

core/src/components/modal/test/sheet/modal.e2e.ts
```diff
@@ -398,6 +398,46 @@ configs({ modes: ['ios'], directions: ['ltr'] }).forEach(({ title, config }) =>
       expect(ionDragEnd.length).toBe(1);
       expect(Object.keys(dragEndEvent.detail).length).toBe(5);
     });
+
+    // Native drag is cancelled during the gesture (see gestures/sheet.ts)
+    // but must work again once it ends.
```
brandyscarney:
```suggestion
    // Native drag can interfere with the swipe gesture when initiated over
    // text in the focusable wrapper. Suppress it during the gesture only.
```

Typically we don't mention file names in tests.

core/src/components/modal/test/card/modal-card.e2e.ts
```diff
@@ -150,5 +150,45 @@ configs({ modes: ['ios'], directions: ['ltr'] }).forEach(({ title, screenshot, c
       expect(ionDragEnd.length).toBe(1);
       expect(Object.keys(dragEndEvent.detail).length).toBe(4);
     });
+
+    // Native drag is cancelled during the gesture (see gestures/swipe-to-close.ts)
+    // but must work again once it ends.
```
brandyscarney:
```suggestion
    // Native drag can interfere with the swipe gesture when initiated over
    // text in the focusable wrapper. Suppress it during the gesture only.
```

Typically we don't mention file names in tests.

core/src/components/modal/test/card/modal-card.e2e.ts
```diff
@@ -150,5 +150,45 @@ configs({ modes: ['ios'], directions: ['ltr'] }).forEach(({ title, screenshot, c
...
+    test('should cancel native drag and drop only while the gesture is active', async ({ page }, testInfo) => {
+      testInfo.annotations.push({
+        type: 'issue',
+        description: 'modal content is inaccessible when TalkBack is enabled',
+      });
```
brandyscarney:
```suggestion
    test('should cancel native drag and drop only while the gesture is active', async ({ page }) => {
```

Test annotations like this should only be used when the test was added to solve an issue (Jira ticket or GH issue).

core/src/components/modal/test/sheet/modal.e2e.ts
```diff
@@ -398,6 +398,46 @@ configs({ modes: ['ios'], directions: ['ltr'] }).forEach(({ title, config }) =>
...
+    test('should cancel native drag and drop only while the gesture is active', async ({ page }, testInfo) => {
+      testInfo.annotations.push({
+        type: 'issue',
+        description: 'modal content is inaccessible when TalkBack is enabled',
+      });
```
brandyscarney:
```suggestion
    test('should cancel native drag and drop only while the gesture is active', async ({ page }) => {
```

Test annotations like this should only be used when the test was added to solve an issue (Jira ticket or GH issue).

core/src/components/modal/test/a11y/modal.e2e.ts
```diff
@@ -19,5 +19,101 @@ configs({ modes: ['ios'], directions: ['ltr'] }).forEach(({ title, config }) =>
       const results = await new AxeBuilder({ page }).analyze();
       expect(results.violations).toEqual([]);
     });
+
+    // Focus the role="dialog" wrapper on present so screen readers can enter.
+    test('should focus the modal wrapper on present', async ({ page }, testInfo) => {
+      testInfo.annotations.push({
+        type: 'issue',
+        description: 'modal content is inaccessible when TalkBack is enabled',
```
brandyscarney:
```suggestion
        description: 'FW-7611',
```

Issue description should point to a ticket number so we can get all of the information related to it.

core/src/components/modal/test/a11y/modal.e2e.ts
```diff
@@ -19,5 +19,101 @@ configs({ modes: ['ios'], directions: ['ltr'] }).forEach(({ title, config }) =>
...
+      await ionModalDidPresent.next();
+
+      await expect(wrapper).toBeFocused();
+      await expect(wrapper).toHaveCSS('outline-style', 'none');
+    });
+
+    test('should focus the sheet modal wrapper on present', async ({ page }, testInfo) => {
+      testInfo.annotations.push({
+        type: 'issue',
+        description: 'modal content is inaccessible when TalkBack is enabled',
```
brandyscarney:
```suggestion
        description: 'FW-7611',
```

Issue description should point to a ticket number so we can get all of the information related to it.

core/src/components/modal/test/a11y/modal.e2e.ts
```diff
@@ -19,5 +19,101 @@ configs({ modes: ['ios'], directions: ['ltr'] }).forEach(({ title, config }) =>
...
+      await ionModalDidPresent.next();
+
+      await expect(wrapper).toHaveAttribute('role', 'dialog');
+      await expect(wrapper).toBeFocused();
+    });
+
+    test('should focus the card modal wrapper on present', async ({ page }, testInfo) => {
+      testInfo.annotations.push({
+        type: 'issue',
+        description: 'modal content is inaccessible when TalkBack is enabled',
```
brandyscarney:
```suggestion
        description: 'FW-7611',
```

Issue description should point to a ticket number so we can get all of the information related to it.

core/src/components/modal/modal.scss
```diff
@@ -84,6 +84,14 @@ ion-backdrop {
...
+ */
+.modal-wrapper {
+  outline: none;
```
brandyscarney, line 92: It looks like this is no longer required with my suggested change to query for `role="dialog"`.

core/src/components/modal/test/a11y/modal.e2e.ts
```diff
@@ -19,5 +19,101 @@ configs({ modes: ['ios'], directions: ['ltr'] }).forEach(({ title, config }) =>
...
+      await expect(wrapper).toHaveAttribute('role', 'dialog');
+      await expect(wrapper).toBeFocused();
+    });
+
+    // The focused wrapper must not show a focus ring when opened via keyboard.
+    test('should not render a focus ring on the wrapper when presented via keyboard', async ({ page }, testInfo) => {
+      testInfo.annotations.push({
+        type: 'issue',
+        description: 'modal content is inaccessible when TalkBack is enabled',
+      });
```
brandyscarney:
```suggestion
    test('should not render a focus ring on the wrapper when presented via keyboard', async ({ page }) => {
```

Test annotations like this should only be used when the test was added to solve an issue (Jira ticket or GH issue).

brandyscarney (changes requested):
I'll add more feedback on the Jira ticket! 🙂

core/src/utils/overlays.ts
gnbm: Adopted the `[role="dialog"]` query — thanks, it reads much cleaner. On "why tabindex at all?": we still need `tabIndex="-1"` on the wrapper because `role="dialog"` alone does **not** make an element focusable, so `focus()` on it is a no-op. I verified this in an isolated test across light + shadow DOM: `<div role="dialog">` → `document.activeElement` unchanged; `<div role="dialog" tabindex="-1">` → focused. So the two are complementary: the `role="dialog"` query selects the correct element, and the `tabindex` lets it actually receive the programmatic focus. Action-sheet/loading keep the role on the host, where `querySelector` won't match the root, so they correctly fall back to focusing the host as before.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

core/src/components/modal/modal.tsx
gnbm: Keeping `tabIndex={-1}` here. Without it, `focus()` on the `role="dialog"` wrapper is a no-op (an ARIA role does not make an element focusable), so the fix would silently stop working — focus would stay on the role-less host and screen readers would have nothing to land on. Verified in light + shadow DOM. See the overlays.ts thread for detail.

core/src/components/modal/modal.tsx
gnbm: Kept this comment (clarified it slightly) since the `tabIndex={-1}` it documents is required for focusability — see the overlays.ts thread.

core/src/components/modal/modal.scss
gnbm: This is still required. Because `tabIndex={-1}` stays on the wrapper (needed for focusability), the wrapper is the element that receives programmatic focus on present, so it can render a focus ring — most visibly when the modal is opened via the keyboard. This rule suppresses that ring to match the host. The `should not render a focus ring on the wrapper when presented via keyboard` e2e covers it.

core/src/components/modal/gestures/sheet.ts
gnbm: Manual repro (Firefox): present a **sheet** modal and drag the header immediately. Because the wrapper is now focusable and contains text, Firefox starts a native text drag that swallows the gesture's pointer events, so `ionDragStart` never fires and the sheet cannot be dragged. Cancelling `dragstart` while the gesture is active fixes it.

Automated: the existing `should emit ionDragStart, ionDragMove, and ionDragEnd events` e2e is the fail/pass gate (I confirmed it fails 3/3 on Firefox with this removed), and `should cancel native drag and drop only while the gesture is active` asserts `dragstart` is `defaultPrevented` during the gesture and not after.

core/src/components/modal/gestures/swipe-to-close.ts
gnbm: Manual repro (Firefox): present a **sheet** modal and drag the header immediately. Because the wrapper is now focusable and contains text, Firefox starts a native text drag that swallows the gesture's pointer events, so `ionDragStart` never fires and the sheet cannot be dragged. Cancelling `dragstart` while the gesture is active fixes it.

Automated: the existing `should emit ionDragStart, ionDragMove, and ionDragEnd events` e2e is the fail/pass gate (I confirmed it fails 3/3 on Firefox with this removed), and `should cancel native drag and drop only while the gesture is active` asserts `dragstart` is `defaultPrevented` during the gesture and not after.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

core/src/components/modal/test/sheet/modal.e2e.ts
gnbm: Applied — updated the comment to drop the file reference.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

core/src/components/modal/test/card/modal-card.e2e.ts
gnbm: Applied — updated the comment to drop the file reference.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

core/src/components/modal/test/sheet/modal.e2e.ts
gnbm: Applied — removed the issue annotation from this workaround test.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

core/src/components/modal/test/card/modal-card.e2e.ts
gnbm: Applied — removed the issue annotation from this workaround test.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

core/src/utils/overlays.ts
```diff
@@ -593,7 +593,15 @@ export const present = async <OverlayPresentOptions>(
...
-    overlay.el.focus();
+    /**
+     * Some overlays (e.g. modal) put the dialog role and label on the
+     * `.ion-overlay-wrapper` rather than the host. Screen readers need
+     * focus on the element with role="dialog" to properly announce and
+     * navigate the dialog.
+     */
+    const overlayWrapper = getElementRoot(overlay.el).querySelector<HTMLElement>('[role="dialog"]');
+    const focusTarget = overlayWrapper ?? overlay.el;
+    focusTarget.focus({ preventScroll: true });
```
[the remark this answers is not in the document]
gnbm: Good catch — fixed in 5ddc39ce78. The query is now `[role="dialog"][tabindex]`, so it only targets a dialog wrapper that is actually focusable (the modal wrapper, which has `tabindex="-1"`). picker-legacy's `role="dialog"` wrapper has no tabindex, so it no longer matches and we fall back to focusing the host — preserving the prior behavior instead of a no-op. Verified the modal/sheet/card focus e2e still pass.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

core/src/utils/overlays.ts
```diff
@@ -593,7 +593,18 @@ export const present = async <OverlayPresentOptions>(
...
+     * `.ion-overlay-wrapper` rather than the host. Screen readers need
+     * focus on the element with role="dialog" to properly announce and
+     * navigate the dialog. The `[tabindex]` qualifier ensures we only
+     * target such a wrapper when it is actually focusable; otherwise
+     * (e.g. picker-legacy, whose role="dialog" wrapper has no tabindex)
+     * `focus()` would be a no-op, so we fall back to the host.
+     */
+    const overlayWrapper = getElementRoot(overlay.el).querySelector<HTMLElement>('[role="dialog"][tabindex]');
+    const focusTarget = overlayWrapper ?? overlay.el;
+    focusTarget.focus({ preventScroll: true });
```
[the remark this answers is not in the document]
gnbm: Added a try/catch fallback to a plain `focus()` in 7449639fed. In practice older engines ignore the unknown options object rather than throwing, but since this runs during `present()` the guard removes any chance a focus call could reject presentation, while preserving the no-scroll behavior on modern browsers.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

core/src/components/modal/modal.scss
```diff
@@ -84,6 +84,14 @@ ion-backdrop {
   z-index: 10;
 }
 
+/**
+ * The wrapper receives programmatic focus for screen readers but should not
+ * show a visible focus ring, which is meant only for keyboard navigation.
+ */
+.modal-wrapper {
+  outline: none;
+}
```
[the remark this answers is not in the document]
gnbm: Keeping this as-is — it preserves existing behavior rather than adding a regression. The modal `:host` already sets `outline: none` (modal.scss line 48), and the host was the element `present()` focused *before* this PR, so the focused overlay container never rendered a ring. Moving that focus to `.modal-wrapper` just moves where the same suppression applies.

The wrapper is `tabindex="-1"`, so it is never a keyboard tab stop — users never navigate *to* it; when they Tab, focus moves to the interactive controls inside, which keep their own `:focus-visible` rings. A ring around the entire dialog container isn't a meaningful focus cue, and showing one only on keyboard-open (via `:focus-visible`) would be new visual noise versus the long-standing host behavior. The `should not render a focus ring on the wrapper when presented via keyboard` e2e locks this in.

joselrio (approved):
After looking into the suggested stuff and also run tests around this directly on the android device through the https://github.com/OutSystems/ui-comp-ionic-sandboxes repo and everything's running as expected.

codeCraft-Ritik (commented):
This is a well-considered fix. Updating focus management to target the dialog element makes the modal much more accessible for screen reader users, while preserving compatibility with overlays that keep the role on the host. The cross-browser consideration is a nice touch.

brandyscarney (approved):

thetaPC (approved):
LGTM

PR #31261, opened 2026-07-10
feat(datetime, datetime-button, picker-column, picker-column-option): improve styling flexibility by removing default color prop

thetaPC:

---------

`ion-datetime` and related components (`ion-datetime-button`, `ion-picker-column`, `ion-picker-column-option`) have a default `color="primary"` property. This forces users who want to customize datetime styling to override the entire primary color system or set explicit colors on every instance, rather than being able to style the component normally.

## What is the new behavior?

- Removed default `color="primary"` property assignments from all four components
- Components now apply primary color styling via SCSS defaults (`ion-color(primary, base)`) instead of relying on a property default
- Added `:host(.ion-color)` overrides in SCSS to apply explicit colors via `current-color()` when a color prop is set
- Action buttons (cancel, done, clear) have explicit primary text color styling to ensure consistent appearance
- Color now cascades from datetime → picker-column → picker-column-option when explicitly set. Previously the selected wheel option (active time picker text) stayed primary regardless of color; it now inherits the value passed to datetime.

| `main` | `FW-6996` |
| --- | --- |
| main | branch |

- [ ] Yes
- [x] No

**Backwards compatible:** Visual appearance is unchanged. Components still appear with primary color by default. Existing code that explicitly sets `color` props continues to work identically.

## Other information

Dev build: `8.8.14-dev.11783706459.1d5a7aae`

Previews:
- [Datetime: basic](https://ionic-framework-git-fw-6996-ionic1.vercel.app/src/components/datetime/test/basic/)
- [Datetime: color](https://ionic-framework-git-fw-6996-ionic1.vercel.app/src/components/datetime/test/color/)
- [Datetime: display](https://ionic-framework-git-fw-6996-ionic1.vercel.app/src/components/datetime/test/display/)
- [Datetime: presentation](https://ionic-framework-git-fw-6996-ionic1.vercel.app/src/components/datetime/test/presentation/)
- [Picker: basic](https://ionic-framework-git-fw-6996-ionic1.vercel.app/src/components/picker/test/basic/)

### DIFF
--- a/core/api.txt
+++ b/core/api.txt
@@ -513,7 +513,7 @@ ion-content,part,scroll
 ion-datetime,shadow
 ion-datetime,prop,cancelText,string,'Cancel',false,false
 ion-datetime,prop,clearText,string,'Clear',false,false
-ion-datetime,prop,color,"danger" | "dark" | "light" | "medium" | "primary" | "secondary" | "success" | "tertiary" | "warning" | string & Record<never, never> | undefined,'primary',false,false
+ion-datetime,prop,color,"danger" | "dark" | "light" | "medium" | "primary" | "secondary" | "success" | "tertiary" | "warning" | string & Record<never, never> | undefined,undefined,false,false
 ion-datetime,prop,dayValues,number | number[] | string | undefined,undefined,false,false
 ion-datetime,prop,disabled,boolean,false,false,false
 ion-datetime,prop,doneText,string,'Done',false,false
@@ -582,7 +582,7 @@ ion-datetime,part,wheel-item
 ion-datetime,part,wheel-item active
 
 ion-datetime-button,shadow
-ion-datetime-button,prop,color,"danger" | "dark" | "light" | "medium" | "primary" | "secondary" | "success" | "tertiary" | "warning" | string & Record<never, never> | undefined,'primary',false,true
+ion-datetime-button,prop,color,"danger" | "dark" | "light" | "medium" | "primary" | "secondary" | "success" | "tertiary" | "warning" | string & Record<never, never> | undefined,undefined,false,true
 ion-datetime-button,prop,datetime,string | undefined,undefined,false,false
 ion-datetime-button,prop,disabled,boolean,false,false,true
 ion-datetime-button,prop,mode,"ios" | "md",undefined,false,false
@@ -1269,7 +1269,7 @@ ion-picker,css-prop,--highlight-border-radius,ios
 ion-picker,css-prop,--highlight-border-radius,md
 
 ion-picker-column,shadow
-ion-picker-column,prop,color,"danger" | "dark" | "light" | "medium" | "primary" | "secondary" | "success" | "tertiary" | "warning" | string & Record<never, never> | undefined,'primary',false,true
+ion-picker-column,prop,color,"danger" | "dark" | "light" | "medium" | "primary" | "secondary" | "success" | "tertiary" | "warning" | string & Record<never, never> | undefined,undefined,false,true
 ion-picker-column,prop,disabled,boolean,false,false,false
 ion-picker-column,prop,mode,"ios" | "md",undefined,false,false
 ion-picker-column,prop,value,number | string | undefined,undefined,false,false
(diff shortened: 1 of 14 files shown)


core/src/components/datetime/datetime.tsx
```diff
@@ -1813,6 +1813,7 @@ export class Datetime implements ComponentInterface {
       >
         {items.map((item) => (
           <ion-picker-column-option
+            color={this.color}
```
thetaPC, line 1816: The month/year picker was displaying selected values in the datetime's color (for example, red for `color="danger"`), but the time picker was ignoring this and showing selected values in primary color instead, creating an inconsistent experience on `md`. This was wrong because the datetime's color prop should control all picker text colors uniformly.

core/src/components/datetime/test/color/datetime.e2e.ts
```diff
@@ -29,3 +29,106 @@ configs({ directions: ['ltr'], palettes: ['light', 'dark'] }).forEach(({ title,
     });
   });
 });
+
```
thetaPC, line 32: Tests were added based on my [findings](https://github.com/ionic-team/ionic-framework/pull/31261/changes#r3560968268).

core/api.txt
brandyscarney, line 1: Are there any breaking changes associated with this, like CSS that may start overriding where it previously wasn't?

core/src/components/datetime/datetime.scss
```diff
@@ -366,9 +373,15 @@
...
+  background: ion-color(primary, base, 0.2);
+
+  box-shadow: 0px 0px 0px 4px ion-color(primary, base, 0.2);
+}
+
+:host(.ion-color) .calendar-day:not(.calendar-day-adjacent-day):focus {
+  background: #{current-color(base, 0.2)};
 
-  box-shadow: 0px 0px 0px 4px current-color(base, 0.2);
+  box-shadow: 0px 0px 0px 4px #{current-color(base, 0.2)};
```
brandyscarney:
```suggestion
  background: ion-color(primary, base, 0.2);

  box-shadow: 0px 0px 0px 4px ion-color(primary, base, 0.2);
}

:host(.ion-color) .calendar-day:not(.calendar-day-adjacent-day):focus {
  background: current-color(base, 0.2);

  box-shadow: 0px 0px 0px 4px current-color(base, 0.2);
```

Nit: you don't need to escape unless it's in a CSS variable

core/src/components/datetime/test/color/datetime.e2e.ts
brandyscarney, line 1: While this is actually probably a bug fix, should we document this text color change somewhere in case it's unexpected?

| `main` | `FW-6996` |
| --- | --- |
| main | branch |

brandyscarney (approved):
Looks great! Left some minor feedback.

core/api.txt
thetaPC: I can't think of any. Devs shouldn't be able to tell the difference with these changes even if they are customizing the primary variables themselves.

core/src/components/datetime/datetime.scss
thetaPC: https://github.com/ionic-team/ionic-framework/pull/31261/changes/9eaee49ec227211a3ab413d3d5dd3e3d52a1bcce

core/src/components/datetime/test/color/datetime.e2e.ts
thetaPC: I updated the PR description to reflect it better. Let me know what you think. I can also fix it another PR if you prefer.

core/src/components/datetime/datetime.scss
```diff
@@ -366,7 +373,13 @@
...
 
 .calendar-day:not(.calendar-day-adjacent-day):focus {
-  background: current-color(base, 0.2);
+  background: ion-color(primary, base, 0.2);
+
+  box-shadow: 0px 0px 0px 4px ion-color(primary, base, 0.2);
+}
+
+:host(.ion-color) .calendar-day:not(.calendar-day-adjacent-day):focus {
+  background: #{current-color(base, 0.2)};
```
ShaneK:
```suggestion
  background: current-color(base, 0.2);
```

Nit: looks like this one got missed when you removed the escapes. The `box-shadow` right below it and the default rule above both drop the `#{}`, so this can too. Up to you!

ShaneK (approved):
Looks good to me, awesome work! Had a nit that's just a follow up on Brandy's earlier comment, but just a nit

PR #31262, opened 2026-07-10
fix(tab-bar): fire tab change events on tab click

### LINKED ISSUE #30145: bug: ionTabsDidChange event not triggering in Ionic 8 (React)
### Prerequisites

### Ionic Framework Version

v8.x

### Current Behavior

The `ionTabsDidChange` event does not trigger in Ionic 8. This issue does not occur in Ionic 7, where the event triggers as expected.

### Expected Behavior

The `ionTabsDidChange` event should trigger in Ionic 8 when a tab is changed, just as it does in Ionic 7.

### Steps to Reproduce

1. Initialize a new Ionic project using the tabs template:
```
ionic start project tabs
```
This installs Ionic 7 by default.

2. Add a listener for the `ionTabsDidChange` event in the IonTabs component and include a `console.log()` to verify the event is triggered.
```
<IonTabs onIonTabsDidChange={(e) => console.log('onIonTabsDidChange', e.detail.tab)}>
...
</IonTabs>
```
 
3. Confirm that the event works as expected by running the project.
 
4. Upgrade to Ionic 8 by running:
```
npm i @ionic/react@8 @ionic/react-router@8
```

5. Run the project again and observe that the `ionTabsDidChange` event no longer triggers when switching tabs.

### Code Reproduction URL

https://codesandbox.io/p/sandbox/purple-hooks-tsmpmp

### Ionic Info

Ionic:

   Ionic CLI       : 5.4.16 (C:\Users\sergio\AppData\Local\npm-cache\_npx\864a9e3c2cd0cf50\node_modules\ionic)
   Ionic Framework : @ionic/react 8.4.1

Utility:

   cordova-res : not installed
   native-run  : not installed

System:

   NodeJS : v18.19.0 (C:\Program Files\nodejs\node.exe)
   npm    : 10.2.3
   OS     : Windows 11

### Additional Information

_No response_

droc101:
Issue number: resolves #30145

---------

`ionic-react`'s `IonTabBar` tries to reference `this.props.onIonTabsWillChange` and `this.props.onIonTabsDidChange`, which do not exist, causing the `onIonTabsWillChange` and `onIonTabsDidChange` events to never fire.

## What is the new behavior?

- Update usage of `this.props.onIonTabsWillChange` to `this.props.tabsContext.tabBarProps.onIonTabsWillChange`
- Update usage of `this.props.onIonTabsDidChange` to `this.props.tabsContext.tabBarProps.onIonTabsDidChange`
- Add check to make sure  `this.props.tabsContext` is truthy before using it
- `onIonTabsWillChange` and `onIonTabsDidChange` fire as expected

- [ ] Yes
- [x] No

### DIFF
--- a/packages/react/src/components/navigation/IonTabBar.tsx
+++ b/packages/react/src/components/navigation/IonTabBar.tsx
@@ -224,18 +224,20 @@ class IonTabBarUnwrapped extends React.PureComponent<InternalProps, IonTabBarSta
         this.context.resetTab(e.detail.tab, originalHref, tappedTab.originalRouteOptions);
       }
     } else {
+      let onWillChange = this.props.onIonTabsWillChange;
+      let onDidChange = this.props.onIonTabsDidChange;
       if (this.props.tabsContext) {
-        if (this.props.tabsContext.tabBarProps.onIonTabsWillChange) {
-          this.props.tabsContext.tabBarProps.onIonTabsWillChange(
-            new CustomEvent('ionTabWillChange', { detail: { tab: e.detail.tab } })
-          );
-        }
-        if (this.props.tabsContext.tabBarProps.onIonTabsDidChange) {
-          this.props.tabsContext.tabBarProps.onIonTabsDidChange(
-            new CustomEvent('ionTabDidChange', { detail: { tab: e.detail.tab } })
-          );
-        }
+        onWillChange = this.props.tabsContext?.tabBarProps.onIonTabsWillChange ?? this.props.onIonTabsWillChange;
+        onDidChange = this.props.tabsContext?.tabBarProps.onIonTabsDidChange ?? this.props.onIonTabsDidChange;
       }
+
+      if (onWillChange) {
+        onWillChange(new CustomEvent('ionTabWillChange', { detail: { tab: e.detail.tab } }));
+      }
+      if (onDidChange) {
+        onDidChange(new CustomEvent('ionTabDidChange', { detail: { tab: e.detail.tab } }));
+      }
+
       if (hasRouterOutlet) {
         this.setActiveTabOnContext(e.detail.tab);
         this.context.changeTab(e.detail.tab, currentHref, e.detail.routeOptions);
--- a/packages/react/src/components/navigation/IonTabBar.tsx
+++ b/packages/react/src/components/navigation/IonTabBar.tsx
@@ -224,11 +224,17 @@ class IonTabBarUnwrapped extends React.PureComponent<InternalProps, IonTabBarSta
         this.context.resetTab(e.detail.tab, originalHref, tappedTab.originalRouteOptions);
       }
     } else {
-      if (this.props.onIonTabsWillChange) {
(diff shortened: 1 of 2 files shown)


packages/react/src/components/navigation/IonTabBar.tsx
```diff
@@ -224,11 +224,17 @@ class IonTabBarUnwrapped extends React.PureComponent<InternalProps, IonTabBarSta
         this.context.resetTab(e.detail.tab, originalHref, tappedTab.originalRouteOptions);
       }
     } else {
-      if (this.props.onIonTabsWillChange) {
-        this.props.onIonTabsWillChange(new CustomEvent('ionTabWillChange', { detail: { tab: e.detail.tab } }));
-      }
-      if (this.props.onIonTabsDidChange) {
-        this.props.onIonTabsDidChange(new CustomEvent('ionTabDidChange', { detail: { tab: e.detail.tab } }));
+      if (this.props.tabsContext) {
```
ShaneK, line 229: This fixes the `IonTabs`-level handlers, but I think it drops the ones set directly on `IonTabBar`? They're typed on `IonTabBarProps`, and `IonTabBarContainer` builds `updatedTabBarProps` as `{...tabsContext.tabBarProps, ref}` without merging the direct props, so `this.props.tabsContext.tabBarProps.onIonTabsWillChange` won't see them. Before this change `this.props.onIonTabsWillChange` was read, so a handler on `<IonTabBar onIonTabsDidChange={fn}>` fired and now silently won't.

`IonTabs` at lines 153-172 deliberately avoids overriding a handler set on the tab bar (the comment there spells it out), so that path looks intentional. Could we fall back to the direct prop so both work? Something like:

```ts
const onWillChange = this.props.tabsContext?.tabBarProps.onIonTabsWillChange ?? this.props.onIonTabsWillChange;
const onDidChange = this.props.tabsContext?.tabBarProps.onIonTabsDidChange ?? this.props.onIonTabsDidChange;
```

ShaneK, line 229: Could we add a regression test for this? It's a behavior fix with nothing guarding it, which is how it slipped in originally. The harness is already there: [`TabsBasic.tsx`](https://github.com/ionic-team/ionic-framework/blob/07c230469cc8e256fcab739d66890605c370cf40/packages/react/test/base/src/pages/TabsBasic.tsx) sets both handlers on `<IonTabs>` at `/tabs-basic`, and the `Without IonRouterOutlet` block in [`tabs.cy.ts`](https://github.com/ionic-team/ionic-framework/blob/07c230469cc8e256fcab739d66890605c370cf40/packages/react/test/base/tests/e2e/specs/tabs/tabs.cy.ts) already visits it and clicks `ion-tab-button[tab="tab2"]`. Asserting the callbacks fire on that click should cover it.

ShaneK (changes requested):
Looking good, just a change request and test request

ShaneK (approved):

